PR #304, opened 2021-11-24
[master < T0774-MG] Update stats with query modules

antonio2368:
[master < Task] PR


jbajic (approved):

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -2113,7 +2138,12 @@ mgp_error mgp_graph_detach_delete_vertex(struct mgp_graph *graph, mgp_vertex *ve
       }
     }
 
-    auto *trigger_ctx_collector = graph->ctx->trigger_context_collector;
+    auto &ctx = graph->ctx;
+
+    ctx->execution_stats[query::ExecutionStats::Key::DELETED_NODES] += 1;
+    ctx->execution_stats[query::ExecutionStats::Key::DELETED_EDGES] += (*result)->second.size();
```
antaljanosbenjamin, line 2144: At this point `result` can be `std::nullopt`, therefore I would push this after the following `if` statement after which `result` holds a valid value.

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -2087,8 +2108,12 @@ mgp_error mgp_graph_delete_vertex(struct mgp_graph *graph, mgp_vertex *vertex) {
           throw SerializationException{"Cannot serialize removing a vertex."};
       }
     }
-    if (graph->ctx->trigger_context_collector && *result) {
-      graph->ctx->trigger_context_collector->RegisterDeletedObject(**result);
+    auto &ctx = graph->ctx;
+
+    ctx->execution_stats[query::ExecutionStats::Key::DELETED_NODES] += 1;
```
antaljanosbenjamin, line 2113: I would put this into the inner body of the following `if` statement, because result can be `std::nullopt` here, which means no deletion.

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -2176,8 +2210,13 @@ mgp_error mgp_graph_delete_edge(struct mgp_graph *graph, mgp_edge *edge) {
           throw SerializationException{"Cannot serialize removing an edge."};
       }
     }
-    if (graph->ctx->trigger_context_collector && *result) {
-      graph->ctx->trigger_context_collector->RegisterDeletedObject(**result);
+
+    auto &ctx = graph->ctx;
+
+    ctx->execution_stats[query::ExecutionStats::Key::DELETED_EDGES] += 1;
```
antaljanosbenjamin, line 2216: Same as for `delete_vertex`.

antaljanosbenjamin (changes requested):

kostasrim (approved):

kostasrim:
I think once @antaljanosbenjamin comments are addressed this should be good to merge

antaljanosbenjamin (approved):

PR #306, opened 2021-11-29
[master < T0775-MG] Use default values instantly for batch size/interval

antonio2368:
[master < Task] PR


tests/unit/integrations_kafka_consumer.cpp
```diff
@@ -40,6 +41,10 @@ int SpanToInt(std::span<const char> span) {
   std::memcpy(&result, span.data(), sizeof(int));
   return result;
 }
+
+constexpr std::chrono::milliseconds kDefaultBatchInterval{100};
+constexpr int64_t kDefaultBatchSize{1000};
```
antaljanosbenjamin, line 46: Why not to use `streams::kDefaultBatchInterval` and `streams::kDefaultBatchSzie`?

antonio2368: I have to link against mg-query and this seemed like the fastest way because it's not that critical IMO. 
I can do that if you think it's better.

antaljanosbenjamin: Ohh I see. We don't use the `Streams` object here, so I was wrong. Even though this value is happened to be the same as in `Stream`, they are logically separate values.

antaljanosbenjamin (approved):

kostasrim (approved):

jbajic (approved):

PR #309, opened 2021-12-01
[master < MG] Update to newest spdlog and fmt

antonio2368:
[master < Task] PR


jbajic (approved):

src/utils/logging.hpp
```diff
@@ -22,31 +22,33 @@
...
       "\n\tExpression: '{}'"
       "{}",
-      file_name, line_num, expr, message ? fmt::format("\n\tMessage: '{}'", *message) : "");
+      file_name, line_num, expr, !message.empty() ? fmt::format("\n\tMessage: '{}'", message) : "");
   std::terminate();
 }
 
+#define GET_MESSAGE(...) \
+  BOOST_PP_IF(BOOST_PP_EQUAL(BOOST_PP_VARIADIC_SIZE(__VA_ARGS__), 0), "", fmt::format(__VA_ARGS__))
+
```
antaljanosbenjamin, line 49: What about using two functions?

```c++

std::string_view GetMessage() { return ""; }

template <typename... Args>
std::sting GetMessage(const Args&...msg_args) { return fmt::format(msg_args...)); }
```
This way we can avoid using even more macros.

src/utils/logging.hpp
```diff
@@ -22,31 +22,33 @@
...
 // TODO (antonio2368): Replace with std::source_location when it's supported by
 // compilers
-template <typename... Args>
-void AssertFailed(const char *file_name, int line_num, const char *expr, const Args &...msg_args) {
-  std::optional<std::string> message;
-  if constexpr (sizeof...(msg_args) > 0) {
-    message.emplace(fmt::format(msg_args...));
-  }
-
+inline void AssertFailed(const char *file_name, int line_num, const char *expr, const std::string &message) {
```
antaljanosbenjamin, line 38: This change is necessary to utilize the compile time checks in fmt, right?

tests/unit/utils_temporal.cpp
```diff
@@ -24,11 +24,12 @@
 namespace {
 
 std::string ToString(const utils::DateParameters &date_parameters) {
-  return fmt::format("{:04d}-{:02d}-{:02d}", date_parameters.year, date_parameters.month, date_parameters.day);
+  return fmt::format(fmt::runtime("{:04d}-{:02d}-{:02d}"), date_parameters.year, date_parameters.month,
```
antaljanosbenjamin, line 27: Why is `fmt::runtime` necessary here?

src/utils/logging.hpp
antonio2368: Here's the deal:
`fmt::format(msg, args...)` creates `fmt::format_string` from the `msg` argument. That constructor is `consteval` so `msg` should be `constant value`. Arguments of function that is executed during runtime has all the arguments as runtime argument, so the `consteval` requirement will fail. I could avoid that problem by making `AssertFailed` accept `fmt::format_string` as argument (look at `BasicExceptions`), but there is no overload which accepts no msg, so I had to do it with macros.

src/utils/CMakeLists.txt
```diff
@@ -5,17 +5,30 @@ set(utils_src_files
...
     thread.cpp
     thread_pool.cpp
     uuid.cpp)
 
+find_package(Boost REQUIRED)
+
 add_library(mg-utils STATIC ${utils_src_files})
-target_link_libraries(mg-utils mg-kvstore mg-slk stdc++fs Threads::Threads spdlog fmt gflags uuid rt)
+target_link_libraries(mg-utils PUBLIC Boost::headers fmt::fmt spdlog::spdlog)
+target_link_libraries(mg-utils PRIVATE stdc++fs Threads::Threads gflags uuid rt)
```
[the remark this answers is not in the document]
antonio2368: If it's used to compile only source files inside the `mg-utils` why should we implicitly expose them to other libs?
I mean with static libs shouldn't this only expose includes and flags/defs? For symbols, t should use only the ones it needs and throw everything else anyways.

src/utils/logging.hpp
antonio2368: yes

tests/unit/utils_temporal.cpp
antonio2368: good question, need to check it out.

kostasrim (approved):

src/utils/CMakeLists.txt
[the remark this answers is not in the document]
antaljanosbenjamin: I think no. Lets consider the following case:

```cmake
add_library(A STATIC A.hpp A.cpp)

add_library(B STATIC B.hpp B.cpp)
# A.hpp in only included in B.cpp, not in the header
target_link_libraries(B PRIVATE A)

...

# somewhere else
add_executable(E E.hpp E.cpp)
target_link_libraries(E B)
```

This would fail, because when it tries to link `E`, then the objects in `A` cannot be resolved, as `E` is not linked against `A`. So basically the private dependencies are not propagated and this can cause a problem. Or at least that's how [I interpreted it](https://cmake.org/cmake/help/latest/command/target_link_libraries.html#libraries-for-a-target-and-or-its-dependents). In my opinion this is not a good thing.

> I mean with static libs shouldn't this only expose includes and flags/defs?
And also transitively links the dependents to the libs that are necessary for the static lib.

> For symbols, t should use only the ones it needs and throw everything else anyways.
I think the symbols from the dependencies are not included into the dependent static lib.

I will try this out to be 100% sure, because I am in doubt.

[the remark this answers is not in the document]
antonio2368: Based on what I got with Pulsar, cmake link keywords don't affect the symbols or anything on the linking itself, but it defines which libraries can CMake use for building a specific lib.
E.g.
If E used something from A, `PUBLIC` would allow cmake to get A as dependency transitively from B. If it links and uses only symbols from B it should be able to get them, while B as a static lib should have everything it needs from A included in itself. `PRIVATE` won't hide the symbols from A that B incorporated in itself, but it won't have all the symbols, only the ones it required.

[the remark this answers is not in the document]
antonio2368: But definitely check it out!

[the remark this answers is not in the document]
antaljanosbenjamin: You are right, sorry.

antaljanosbenjamin (approved):

PR #314, opened 2021-12-08
[E0108-MG < T0780-MG] Add websocket server and log stream

antonio2368:
[master < Task] PR

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -36,8 +36,10 @@ set(mg_single_node_v2_sources
     glue/auth.cpp
 )
 
+find_package(Boost REQUIRED)
+
 set(mg_single_node_v2_libs stdc++fs Threads::Threads
-    telemetry_lib mg-query mg-communication mg-memory mg-utils mg-auth mg-license mg-settings)
+    telemetry_lib mg-query mg-communication mg-memory mg-utils mg-auth mg-license mg-settings Boost::headers)
 if (MG_ENTERPRISE)
   # These are enterprise subsystems
   set(mg_single_node_v2_libs ${mg_single_node_v2_libs} mg-audit)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -36,10 +36,8 @@ set(mg_single_node_v2_sources
     glue/auth.cpp
 )
 
-find_package(Boost REQUIRED)
-
 set(mg_single_node_v2_libs stdc++fs Threads::Threads
-    telemetry_lib mg-query mg-communication mg-memory mg-utils mg-auth mg-license mg-settings Boost::headers)
+    telemetry_lib mg-query mg-communication mg-memory mg-utils mg-auth mg-license mg-settings)
 if (MG_ENTERPRISE)
   # These are enterprise subsystems
   set(mg_single_node_v2_libs ${mg_single_node_v2_libs} mg-audit)
--- a/src/communication/websocket/listener.cpp
+++ b/src/communication/websocket/listener.cpp
@@ -1,0 +1,82 @@
+// Copyright 2021 Memgraph Ltd.
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
+// License, and you may not use this file except in compliance with the Business Source License.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#include "communication/websocket/listener.hpp"
+
+namespace communication::websocket {
+namespace {
+void LogError(boost::beast::error_code ec, const std::string_view what) {
+  spdlog::warn("Websocket listener failed on {}: {}", what, ec.message());
+}
+}  // namespace
+void Listener::Run() { DoAccept(); }
+
+void Listener::WriteToAll(const std::string_view message) {
+  auto sessions_ptr = sessions_.Lock();
+  for (auto &session : *sessions_ptr) {
+    session->Write(message);
+  }
+}
+
+Listener::Listener(boost::asio::io_context &ioc, tcp::endpoint endpoint) : ioc_(ioc), acceptor_(ioc) {
+  boost::beast::error_code ec;
+
+  // Open the acceptor
+  acceptor_.open(endpoint.protocol(), ec);
+  if (ec) {
+    LogError(ec, "open");
+    return;
+  }
+
+  // Allow address reuse
+  acceptor_.set_option(boost::asio::socket_base::reuse_address(true), ec);
+  if (ec) {
+    LogError(ec, "set_option");
+    return;
+  }
+
+  // Bind to the server address
+  acceptor_.bind(endpoint, ec);
+  if (ec) {
+    LogError(ec, "bind");
+    return;
+  }
+
+  acceptor_.listen(boost::asio::socket_base::max_listen_connections, ec);
+  if (ec) {
+    LogError(ec, "listen");
+    return;
+  }
+}
+
+void Listener::DoAccept() {
+  // The new connection gets its own strand
+  acceptor_.async_accept(boost::asio::make_strand(ioc_), [shared_this = shared_from_this()](auto ec, auto socket) {
+    shared_this->OnAccept(ec, std::move(socket));
+  });
+}
+
+void Listener::OnAccept(boost::beast::error_code ec, tcp::socket socket) {
+  if (ec) {
+    return LogError(ec, "accept");
+  }
+
+  {
+    auto sessions_ptr = sessions_.Lock();
+    sessions_ptr->emplace_back(Session::CreateSession(std::move(socket)))->Run();
+
+    // Clean disconnected clients
+    std::erase_if(*sessions_ptr, [](const auto &elem) { return !elem->Connected(); });
+  }
+
+  DoAccept();
+}
+}  // namespace communication::websocket
--- a/src/communication/websocket/listener.cpp
+++ b/src/communication/websocket/listener.cpp
@@ -60,7 +60,6 @@ Listener::Listener(boost::asio::io_context &ioc, tcp::endpoint endpoint) : ioc_(
 }
 
 void Listener::DoAccept() {
-  // The new connection gets its own strand
   acceptor_.async_accept(
       ioc_, [shared_this = shared_from_this()](auto ec, auto socket) { shared_this->OnAccept(ec, std::move(socket)); });
 }
--- a/src/communication/websocket/listener.cpp
+++ b/src/communication/websocket/listener.cpp
@@ -55,6 +55,8 @@ Listener::Listener(boost::asio::io_context &ioc, tcp::endpoint endpoint) : ioc_(
     LogError(ec, "listen");
     return;
   }
+
+  spdlog::info("WebSocket server is listening on {}:{}", endpoint.address(), endpoint.port());
 }
 
 void Listener::DoAccept() {
--- a/src/communication/websocket/listener.cpp
+++ b/src/communication/websocket/listener.cpp
@@ -19,7 +19,7 @@ void LogError(boost::beast::error_code ec, const std::string_view what) {
 }  // namespace
 void Listener::Run() { DoAccept(); }
 
-void Listener::WriteToAll(const std::string_view message) {
+void Listener::WriteToAll(std::shared_ptr<std::string> message) {
   auto sessions_ptr = sessions_.Lock();
   for (auto &session : *sessions_ptr) {
     session->Write(message);
@@ -61,9 +61,8 @@ Listener::Listener(boost::asio::io_context &ioc, tcp::endpoint endpoint) : ioc_(
 
 void Listener::DoAccept() {
   // The new connection gets its own strand
-  acceptor_.async_accept(boost::asio::make_strand(ioc_), [shared_this = shared_from_this()](auto ec, auto socket) {
-    shared_this->OnAccept(ec, std::move(socket));
-  });
+  acceptor_.async_accept(
+      ioc_, [shared_this = shared_from_this()](auto ec, auto socket) { shared_this->OnAccept(ec, std::move(socket)); });
 }
 
 void Listener::OnAccept(boost::beast::error_code ec, tcp::socket socket) {
@@ -73,10 +72,10 @@ void Listener::OnAccept(boost::beast::error_code ec, tcp::socket socket) {
 
   {
     auto sessions_ptr = sessions_.Lock();
-    sessions_ptr->emplace_back(Session::CreateSession(std::move(socket)))->Run();
+    sessions_ptr->emplace_back(Session::Create(std::move(socket)))->Run();
 
     // Clean disconnected clients
-    std::erase_if(*sessions_ptr, [](const auto &elem) { return !elem->Connected(); });
+    std::erase_if(*sessions_ptr, [](const auto &elem) { return !elem->IsConnected(); });
   }
 
   DoAccept();
--- a/src/communication/websocket/listener.hpp
+++ b/src/communication/websocket/listener.hpp
@@ -11,15 +11,14 @@
 
 #pragma once
 
-#include <boost/asio/io_context.hpp>
 #include <list>
 #include <memory>
 
+#include <spdlog/spdlog.h>
+#include <boost/asio/io_context.hpp>
 #include <boost/asio/ip/tcp.hpp>
 #include <boost/asio/strand.hpp>
 #include <boost/beast/core.hpp>
-
-#include <spdlog/spdlog.h>
 
 #include "communication/websocket/session.hpp"
 #include "utils/spin_lock.hpp"
@@ -31,13 +30,13 @@ class Listener : public std::enable_shared_from_this<Listener> {
 
  public:
   template <typename... Args>
-  static std::shared_ptr<Listener> CreateListener(Args &&...args) {
+  static std::shared_ptr<Listener> Create(Args &&...args) {
     return std::shared_ptr<Listener>{new Listener(std::forward<Args>(args)...)};
   }
 
   // Start accepting incoming connections
   void Run();
-  void WriteToAll(std::string_view message);
+  void WriteToAll(std::shared_ptr<std::string> message);
 
  private:
   Listener(boost::asio::io_context &ioc, tcp::endpoint endpoint);
--- a/src/communication/websocket/listener.hpp
+++ b/src/communication/websocket/listener.hpp
@@ -1,0 +1,111 @@
+// Copyright 2021 Memgraph Ltd.
+//
+// Use of this software is governed by the Business Source License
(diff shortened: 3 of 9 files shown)


src/memgraph.cpp
```diff
@@ -1200,20 +1204,35 @@ int main(int argc, char **argv) {
...
     // After the server is notified to stop accepting and processing
     // connections we tell the execution engine to stop processing all pending
     // queries.
     query::Shutdown(&interpreter_context);
+    websocket_server.Shutdown();
   };
-  InitSignalHandlers(shutdown);
 
   MG_ASSERT(server.Start(), "Couldn't start the Bolt server!");
+  websocket_server.Start();
```
antonio2368, line 1229: Do we have websocket server as optional? :thinking:

gitbuda: If we have the server compiled, we should enable it by default (otherwise, it's just consuming the memory and people have to make additional effort to start it). Being optional vs. what's the default value are two different questions. From my perspective, the more exciting thing is to make the server "compile-time option" for the environments where you have less memory and don't need WebSockets (I'm not sure what the binary size / actual memory consumption is). That seems to me as an exciting option 😄 But, even though that's interesting, I would ignore that for now 😂

antonio2368: I agree with the compile out option, something we can consider if it proves too heavy for some env even though I doubt that. 
My question was for the assert, what happens if it fails to bind to a port? 
Do we terminate as is the case for bolt port or do we just print out an error?

gitbuda: Aha, that's another point. Since WebSockets are not required for Memgraph to be operational, I think it makes more sense to print out an error and move on 🤔

gitbuda: But yea, if the WebSockets-related code can deal with that? If no, than we have to assert here 😄

src/communication/websocket/listener.hpp
```diff
@@ -0,0 +1,111 @@
...
+    if (ec) {
+      return LogError(ec, "accept");
+    }
+
+    {
+      auto sessions_ptr = sessions_.Lock();
+      sessions_ptr->emplace_back(std::make_shared<WebSocketSession>(std::move(socket)))->Run();
+
+      // Clean disconnected clients
+      for (auto it = sessions_ptr->begin(); it != sessions_ptr->end();) {
```
kostasrim, line 91:
```suggestion
     std::erase_if(sessions_ptr->begin(), sessions_ptr->end(), [](auto& elem) { return !elem.Connected(); });
```

src/communication/websocket/session.hpp
```diff
@@ -0,0 +1,119 @@
...
+#pragma once
+
+#include <deque>
+#include <memory>
+
+#include <boost/asio/dispatch.hpp>
+#include <boost/asio/ip/tcp.hpp>
+#include <boost/beast/websocket.hpp>
+
+class WebSocketSession : public std::enable_shared_from_this<WebSocketSession> {
```
kostasrim, line 21: Consider the same comment as above.

src/communication/websocket/session.hpp
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  void OnAccept(boost::beast::error_code ec) {
+    if (ec) {
+      return LogError(ec, "accept");
+    }
+    connected_.store(true, std::memory_order_relaxed);
+
+    // run on the strand
+    boost::asio::dispatch(ws_.get_executor(), [shared_this = shared_from_this()] { shared_this->DoRead(); });
```
kostasrim, line 91: `executors` everywhere ;)

kostasrim (commented):
Really nice work, big thumbs up for Beast!

src/communication/websocket/listener.hpp
```diff
@@ -0,0 +1,111 @@
...
+#include "communication/websocket/session.hpp"
+#include "utils/spin_lock.hpp"
+#include "utils/synchronized.hpp"
+
+namespace communication::websocket {
+class WebSocketListener : public std::enable_shared_from_this<WebSocketListener> {
+  using tcp = boost::asio::ip::tcp;
+
+ public:
+  WebSocketListener(boost::asio::io_context &ioc, tcp::endpoint endpoint) : ioc_(ioc), acceptor_(ioc) {
```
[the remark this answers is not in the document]
antonio2368: That's a very nice suggestion! :smile:

src/communication/websocket/listener.cpp
```diff
@@ -0,0 +1,84 @@
...
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#include "communication/websocket/listener.hpp"
+
+namespace communication::websocket {
+namespace {
+void LogError(boost::beast::error_code ec, const std::string_view what) {
```
jbajic, line 16:
```suggestion
void LogError(boost::beast::error_code ec, std::string_view what) {
```

antonio2368: `what` doesn't change throughout the function definition, why not have it as a const?

src/communication/websocket/listener.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+#include <boost/asio/io_context.hpp>
+#include <list>
+#include <memory>
+
+#include <boost/asio/ip/tcp.hpp>
+#include <boost/asio/strand.hpp>
+#include <boost/beast/core.hpp>
+
+#include <spdlog/spdlog.h>
```
antonio2368, line 22:
```suggestion
#include <list>
#include <memory>

#include <boost/asio/io_context.hpp>
#include <boost/asio/ip/tcp.hpp>
#include <boost/asio/strand.hpp>
#include <boost/beast/core.hpp>
#include <spdlog/spdlog.h>
```

src/communication/websocket/server.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+void Server::LoggingSink::sink_it_(const spdlog::details::log_msg &msg) {
+  const auto listener = listener_.lock();
+  if (!listener) {
+    return;
+  }
+  using memory_buf_t = fmt::basic_memory_buffer<char, 250>;
+  memory_buf_t formatted;
+  base_sink<std::mutex>::formatter_->format(msg, formatted);
+  listener->WriteToAll(std::string_view{formatted.data(), formatted.size()});
```
antaljanosbenjamin, line 45: I couldn't find out what happens if 250 character is not enough? It is truncated? if yes, then I would go with some dynamic buffer if possible.

src/memgraph.cpp
```diff
@@ -1200,20 +1204,35 @@ int main(int argc, char **argv) {
...
-  auto shutdown = [&server, &interpreter_context] {
+  auto shutdown = [&websocket_server, &server, &interpreter_context] {
     // Server needs to be shutdown first and then the database. This prevents
     // a race condition when a transaction is accepted during server shutdown.
     server.Shutdown();
     // After the server is notified to stop accepting and processing
     // connections we tell the execution engine to stop processing all pending
     // queries.
     query::Shutdown(&interpreter_context);
+    websocket_server.Shutdown();
```
antaljanosbenjamin, line 1225: Shouldn't we wait for the server to shut down properly? Or why not stop and wait for the server in the destructor of `websoket::Server`?

antaljanosbenjamin:
Are you planning to introduce some kind of tests? I assume boost::beast and boost::asio has well known ways to write unit tests. For e2e tests I think it would be nice to at least try to connect to the socket and read some logs (e.g.: with some regex check the logs are well formatted).

src/communication/websocket/server.cpp
antonio2368: https://fmt.dev/latest/api.html#_CPPv4I0_6size_t0EN3fmt19basic_memory_bufferE
It's growing.

src/communication/websocket/session.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+ private:
+  explicit Session(tcp::socket &&socket) : ws_(std::move(socket)) {}
+
+  void OnRun();
+  void DoWrite();
+  void OnWrite(boost::beast::error_code ec, size_t bytest_transferred);
+  void OnAccept(boost::beast::error_code ec);
+  void DoRead();
+  void OnRead(boost::beast::error_code ec, size_t bytest_transferred);
```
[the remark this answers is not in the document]
antonio2368: We don't have a `Read` from exposed, we have `DoRead`/`OnRead` which are for the internal reading from the websocket. We need to read to notice closed connections.
It's same as `Write` vs `OnWrite`/`DoWrite`.

src/communication/websocket/server.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+void Server::Shutdown() { ioc_.stop(); }
+
+void Server::AwaitShutdown() {
+  if (background_thread_ && background_thread_->joinable()) {
+    background_thread_->join();
+  }
+}
+
+bool Server::IsRunning() { return !background_thread_ || ioc_.stopped(); }
```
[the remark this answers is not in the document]
antonio2368: you are correct it should be the opposite of this...

src/memgraph.cpp
antonio2368: We are waiting for it bellow (`websocket_server.AwaitShutdown()`)
I went with the same solution as for the Bolt Server, so that's why we don't wait in the destructor.

src/communication/websocket/session.hpp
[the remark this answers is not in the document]
antonio2368: Yes, this depends that everything is being run on a strand (sequentially).
Notice in the listener when I accept a session, I call `boost::asio::make_strand`, so each session gets its own strand.
Strand is a synchronization technique meaning that the sessions are thread-safe that way.

All the operations in session are dispatched on the executor we sent to the websocket which is the strand created in the listener.

But, I should write a comment which explains that the Session isn't thread-safe by itself, and strand should be used for synchronization.

src/communication/websocket/listener.cpp
jbajic: Sry, you are right

src/communication/websocket/listener.cpp
```diff
@@ -0,0 +1,84 @@
...
+  spdlog::warn("Websocket listener failed on {}: {}", what, ec.message());
+}
+}  // namespace
+void Listener::Run() { DoAccept(); }
+
+void Listener::WriteToAll(const std::string_view message) {
+  auto sessions_ptr = sessions_.Lock();
+  for (auto &session : *sessions_ptr) {
+    session->Write(message);
+  }
```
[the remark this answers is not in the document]
antonio2368: Yeah yeah, you're right, will change it.

src/communication/websocket/session.hpp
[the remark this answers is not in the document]
antonio2368: I went with a slightly different solution, I don't like how we depend for one part on the caller for this class to be thread safe (make a strand) and for other we don't (atomic `is_connected_`). So I made the class explicitly thread_safe by creating manually the strand from whatever executor we receive in the websocket.

src/communication/websocket/server.hpp
```diff
@@ -0,0 +1,63 @@
...
+  Server(Server &&) = delete;
+  Server &operator=(const Server &) = delete;
+  Server &operator=(Server &&) = delete;
+
+  ~Server();
+
+  void Start();
+  void Shutdown();
+  void AwaitShutdown();
+  bool IsRunning();
```
kostasrim, line 41:
```suggestion
  bool IsRunning() const;
```

src/communication/websocket/session.cpp
```diff
@@ -0,0 +1,101 @@
...
+    }
+    shared_this->DoWrite();
+  });
+}
+
+bool Session::Connected() { return connected_.load(std::memory_order_relaxed); }
+
+void Session::DoWrite() {
+  const auto next_message = messages_.front();
+  ws_.async_write(boost::asio::buffer(*next_message), [message_string = next_message, shared_this = shared_from_this()](
```
kostasrim, line 59: There is way to avoid an allocation here. I would move the element from the front to the lambda (as a sink) and in the case of `if(ec)` on `OnWrite` I would move it back at the front.

src/communication/websocket/session.cpp
```diff
@@ -0,0 +1,101 @@
...
+  ws_.set_option(boost::beast::websocket::stream_base::timeout::suggested(boost::beast::role_type::server));
+
+  ws_.set_option(boost::beast::websocket::stream_base::decorator(
+      [](boost::beast::websocket::response_type &res) { res.set(boost::beast::http::field::server, "Memgraph WS"); }));
+
+  // Accept the websocket handshake
+  boost::beast::error_code ec;
+  ws_.accept(ec);
+  OnAccept(ec);
+  // ws_.async_accept([shared_this = shared_from_this()](boost::beast::error_code ec) { shared_this->OnAccept(ec); });
```
kostasrim: What is this?

src/communication/websocket/session.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+ public:
+  template <typename... Args>
+  static std::shared_ptr<Session> CreateSession(Args &&...args) {
+    return std::shared_ptr<Session>{new Session{std::forward<Args>(args)...}};
+  }
+
+  void Run();
+  void Write(std::string_view message);
+  bool Connected();
```
kostasrim:
```suggestion
  bool Connected() const;
```

src/communication/websocket/session.cpp
antonio2368: What allocation exactly? It's a `shared_ptr` so nothing much should happen?

src/communication/websocket/session.cpp
antonio2368: Previous version, it's removed now.

src/communication/websocket/session.cpp
kostasrim: I am talking about the string copy, `message_string = next_message`

src/communication/websocket/session.hpp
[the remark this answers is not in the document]
antaljanosbenjamin: Yes, I think that is so much better!

antonio2368:
> Are you planning to introduce some kind of tests? I assume boost::beast and boost::asio has well known ways to write unit tests. For e2e tests I think it would be nice to at least try to connect to the socket and read some logs (e.g.: with some regex check the logs are well formatted).

At a later point, I'll try to do that but for now, I would like to finalize it first because we still are not sure how it will look like exactly.

src/communication/websocket/session.cpp
antonio2368: `next_message` is a `shared_ptr` so no string copy going on here.

jbajic (approved):
I think it's ok to merge this since we can expect more changes on this

src/CMakeLists.txt
```diff
@@ -36,8 +36,10 @@ set(mg_single_node_v2_sources
     glue/auth.cpp
 )
 
+find_package(Boost REQUIRED)
```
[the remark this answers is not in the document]
antonio2368: We can add that independently as it's called from multiple places.
I would like to avoid inconsistency for now.

src/communication/websocket/session.cpp
antaljanosbenjamin: Actually copying a shared_ptr requires locking (to modify the ref counter), so move does make sense here in my opinion.

src/communication/websocket/session.cpp
```diff
@@ -0,0 +1,95 @@
...
+  });
+}
+
+bool Session::IsConnected() const { return connected_.load(std::memory_order_relaxed); }
+
+void Session::DoWrite() {
+  const auto next_message = messages_.front();
+  ws_.async_write(boost::asio::buffer(*next_message), [message_string = next_message, shared_this = shared_from_this()](
+                                                          boost::beast::error_code ec, const size_t bytes_transferred) {
+    shared_this->OnWrite(ec, bytes_transferred);
```
antaljanosbenjamin, line 61: This won't execute on `strand_`, so I think this might cause some data race between line 66 and line 46.

antaljanosbenjamin (changes requested):

src/communication/websocket/session.cpp
antonio2368: why do you think it won't execute on `strand_`? It's DoWrite is dispatched on it.

antonio2368: Oh I see, it doesn't fetch automatically the executor of the write, I should bind the executor of the callback to strand_.

antonio2368: It seems I was right the first time:
https://stackoverflow.com/a/15879971
It will execute everything on the strand. YOu also get a compiler error if you try to bind strand to the completion handler explicitly.

src/memgraph.cpp
```diff
@@ -1200,20 +1204,35 @@ int main(int argc, char **argv) {
...
   };
-  InitSignalHandlers(shutdown);
 
   MG_ASSERT(server.Start(), "Couldn't start the Bolt server!");
+  websocket_server.Start();
+
+  InitSignalHandlers(shutdown);
```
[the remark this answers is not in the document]
antonio2368: That is true, but IntepreterContext cannot receive queries without the Bolt Server running.
I could move it before starting WebSocket server but could be confusing.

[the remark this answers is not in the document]
antaljanosbenjamin: After some discussion on slack we figured out the following things:
1. The signal handler should be registered before the BOLT server is started, because otherwise (theoretically) the interpreter can be interrupted without gracefully shutting down (because of the default signal handler just terminates the application).
2. Furthermore in case of the application is terminated, the destructor of `Storage` won't be called, therefore the snapshots/wal file can end up in a corrupted state. 
3.  If the Memgraph receives SIGINT after the server was created and before it was started shutdown will be called on the server and then the server will be started. The start function creates a thread with infinite loop and the only way to stop it is from the shutdown, thus Memgraph won't shutdown in that case because the thread will never join.

src/communication/websocket/listener.cpp
```diff
@@ -0,0 +1,83 @@
...
+  if (ec) {
+    LogError(ec, "listen");
+    return;
+  }
+
+  spdlog::info("WebSocket server is listening on {}:{}", endpoint.address(), endpoint.port());
+}
+
+void Listener::DoAccept() {
+  // The new connection gets its own strand
```
antonio2368, line 63: remove comment

src/communication/websocket/session.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+  if (!messages_.empty()) {
+    DoWrite();
+  }
+}
+
+void Session::DoRead() {
+  ws_.async_read(buffer_,
+                 [shared_this = shared_from_this()](boost::beast::error_code ec, const size_t bytes_transferred) {
+                   shared_this->OnRead(ec, bytes_transferred);
```
antonio2368, line 86: callback should be executed on a strand.

antaljanosbenjamin (approved):

PR #316, opened 2021-12-14
[E0108-MG < T0781-MG] Send logs over websocket

antonio2368:


src/communication/websocket/listener.hpp
```diff
@@ -11,6 +11,8 @@
 
 #pragma once
 
+#define BOOST_ASIO_USE_TS_EXECUTOR_AS_DEFAULT
```
antaljanosbenjamin, line 14: As currently this is only contained in the `websocket` folder, I think it is fine. But as soon as we are going to include asio in the bolt server and its tests, I think it is better to wrap boost asio in some way.

antaljanosbenjamin (approved):

jbajic (approved):

src/communication/websocket/listener.hpp
kostasrim: I was about to make exact same comment +1

kostasrim (approved):

PR #322, opened 2022-01-07
[E108-MG < T0783-MG] Add websocket authentication

jbajic:
[master < Task] PR

### DIFF
--- a/src/auth/CMakeLists.txt
+++ b/src/auth/CMakeLists.txt
@@ -8,7 +8,7 @@ find_package(Seccomp REQUIRED)
 
 add_library(mg-auth STATIC ${auth_src_files})
 target_link_libraries(mg-auth json libbcrypt gflags fmt::fmt)
-target_link_libraries(mg-auth mg-utils mg-kvstore)
+target_link_libraries(mg-auth mg-utils mg-kvstore mg-license )
 
 target_link_libraries(mg-auth ${Seccomp_LIBRARIES})
 target_include_directories(mg-auth SYSTEM PRIVATE ${Seccomp_INCLUDE_DIRS})
--- a/src/auth/models.cpp
+++ b/src/auth/models.cpp
@@ -68,6 +68,8 @@ std::string PermissionToString(Permission permission) {
       return "AUTH";
     case Permission::STREAM:
       return "STREAM";
+    case Permission::WEBSOCKET:
+      return "WEBSOCKET";
   }
 }
 
--- a/src/auth/models.cpp
+++ b/src/auth/models.cpp
@@ -31,7 +31,7 @@ DEFINE_string(auth_password_strength_regex, default_password_regex.data(),
 namespace auth {
 
 // Constant list of all available permissions.
-constexpr std::array<Permission, 19> kPermissionsAll = {
+constexpr std::array kPermissionsAll = {
     Permission::MATCH,      Permission::CREATE,    Permission::MERGE,       Permission::DELETE,
     Permission::SET,        Permission::REMOVE,    Permission::INDEX,       Permission::STATS,
     Permission::CONSTRAINT, Permission::DUMP,      Permission::AUTH,        Permission::REPLICATION,
--- a/src/auth/models.cpp
+++ b/src/auth/models.cpp
@@ -30,6 +30,14 @@ DEFINE_string(auth_password_strength_regex, default_password_regex.data(),
 
 namespace auth {
 
+// Constant list of all available permissions.
+constexpr std::array<Permission, 19> kPermissionsAll = {
+    Permission::MATCH,      Permission::CREATE,    Permission::MERGE,       Permission::DELETE,
+    Permission::SET,        Permission::REMOVE,    Permission::INDEX,       Permission::STATS,
+    Permission::CONSTRAINT, Permission::DUMP,      Permission::AUTH,        Permission::REPLICATION,
+    Permission::DURABILITY, Permission::READ_FILE, Permission::FREE_MEMORY, Permission::TRIGGER,
+    Permission::CONFIG,     Permission::STREAM,    Permission::WEBSOCKET};
+
 std::string PermissionToString(Permission permission) {
   switch (permission) {
     case Permission::MATCH:
--- a/src/auth/models.hpp
+++ b/src/auth/models.hpp
@@ -36,17 +36,18 @@ enum class Permission : uint64_t {
   TRIGGER     = 1U << 14U,
   CONFIG      = 1U << 15U,
   AUTH        = 1U << 16U,
-  STREAM      = 1U << 17U
+  STREAM      = 1U << 17U,
+  WEBSOCKET   = 1U << 18U
 };
 // clang-format on
 
 // Constant list of all available permissions.
-const std::vector<Permission> kPermissionsAll = {Permission::MATCH,      Permission::CREATE,    Permission::MERGE,
-                                                 Permission::DELETE,     Permission::SET,       Permission::REMOVE,
-                                                 Permission::INDEX,      Permission::STATS,     Permission::CONSTRAINT,
-                                                 Permission::DUMP,       Permission::AUTH,      Permission::REPLICATION,
-                                                 Permission::DURABILITY, Permission::READ_FILE, Permission::FREE_MEMORY,
-                                                 Permission::TRIGGER,    Permission::CONFIG,    Permission::STREAM};
+const std::vector<Permission> kPermissionsAll = {
+    Permission::MATCH,      Permission::CREATE,    Permission::MERGE,       Permission::DELETE,
+    Permission::SET,        Permission::REMOVE,    Permission::INDEX,       Permission::STATS,
+    Permission::CONSTRAINT, Permission::DUMP,      Permission::AUTH,        Permission::REPLICATION,
+    Permission::DURABILITY, Permission::READ_FILE, Permission::FREE_MEMORY, Permission::TRIGGER,
+    Permission::CONFIG,     Permission::STREAM,    Permission::WEBSOCKET};
 
 // Function that converts a permission to its string representation.
 std::string PermissionToString(Permission permission);
--- a/src/auth/models.hpp
+++ b/src/auth/models.hpp
@@ -41,14 +41,6 @@ enum class Permission : uint64_t {
 };
 // clang-format on
 
-// Constant list of all available permissions.
-const std::vector<Permission> kPermissionsAll = {
-    Permission::MATCH,      Permission::CREATE,    Permission::MERGE,       Permission::DELETE,
-    Permission::SET,        Permission::REMOVE,    Permission::INDEX,       Permission::STATS,
-    Permission::CONSTRAINT, Permission::DUMP,      Permission::AUTH,        Permission::REPLICATION,
-    Permission::DURABILITY, Permission::READ_FILE, Permission::FREE_MEMORY, Permission::TRIGGER,
-    Permission::CONFIG,     Permission::STREAM,    Permission::WEBSOCKET};
-
 // Function that converts a permission to its string representation.
 std::string PermissionToString(Permission permission);
 
--- a/src/communication/websocket/auth.cpp
+++ b/src/communication/websocket/auth.cpp
@@ -11,13 +11,15 @@
 
 #include "communication/websocket/auth.hpp"
 
+#include <string>
+
 namespace communication::websocket {
 
 bool SafeAuth::Authenticate(const std::string &username, const std::string &password) const {
   return auth_->Lock()->Authenticate(username, password).has_value();
 }
 
-bool SafeAuth::DoesUserHasPermission(const std::string &username, const auth::Permission permission) const {
+bool SafeAuth::HasUserPermission(const std::string &username, const auth::Permission permission) const {
   if (const auto user = auth_->Lock()->GetUser(username); user) {
     return user->GetPermissions().Has(permission) == auth::PermissionLevel::GRANT;
   }
--- a/src/communication/websocket/auth.cpp
+++ b/src/communication/websocket/auth.cpp
@@ -25,4 +25,6 @@ bool SafeAuth::HasUserPermission(const std::string &username, const auth::Permis
   }
   return false;
 }
+
+bool SafeAuth::HasAnyUsers() const { return auth_->ReadLock()->HasUsers(); }
 }  // namespace communication::websocket
--- a/src/communication/websocket/auth.cpp
+++ b/src/communication/websocket/auth.cpp
@@ -13,7 +13,7 @@
 
 namespace communication::websocket {
 
-bool SafeAuth::Authenticate(const std::string &username, const std::string &password) {
+bool SafeAuth::Authenticate(const std::string &username, const std::string &password) const {
   std::optional<auth::User> maybe_user;
   {
     // TODO change lock to readlock, auth now creates user...
--- a/src/communication/websocket/auth.cpp
+++ b/src/communication/websocket/auth.cpp
@@ -1,0 +1,25 @@
+// Copyright 2021 Memgraph Ltd.
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
+// License, and you may not use this file except in compliance with the Business Source License.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#include "communication/websocket/auth.hpp"
+
+namespace communication::websocket {
+
+bool SafeAuth::Authenticate(const std::string &username, const std::string &password) {
+  std::optional<auth::User> maybe_user;
+  {
+    // TODO change lock to readlock, auth now creates user...
+    auto locked_auth = auth_->Lock();
+    maybe_user = locked_auth->Authenticate(username, password);
+  }
+  return maybe_user.has_value();
+}
+}  // namespace communication::websocket
--- a/src/communication/websocket/auth.cpp
+++ b/src/communication/websocket/auth.cpp
@@ -14,12 +14,6 @@
 namespace communication::websocket {
 
 bool SafeAuth::Authenticate(const std::string &username, const std::string &password) const {
-  std::optional<auth::User> maybe_user;
-  {
-    // TODO change lock to readlock, auth now creates user...
-    auto locked_auth = auth_->Lock();
-    maybe_user = locked_auth->Authenticate(username, password);
-  }
-  return maybe_user.has_value();
+  return auth_->Lock()->Authenticate(username, password).has_value();
 }
 }  // namespace communication::websocket
--- a/src/communication/websocket/auth.cpp
+++ b/src/communication/websocket/auth.cpp
@@ -16,4 +16,11 @@ namespace communication::websocket {
 bool SafeAuth::Authenticate(const std::string &username, const std::string &password) const {
   return auth_->Lock()->Authenticate(username, password).has_value();
 }
+
+bool SafeAuth::DoesUserHasPermission(const std::string &username, const auth::Permission permission) const {
+  if (const auto user = auth_->Lock()->GetUser(username); user) {
+    return user->GetPermissions().Has(permission) == auth::PermissionLevel::GRANT;
+  }
+  return false;
+}
 }  // namespace communication::websocket
--- a/src/communication/websocket/auth.hpp
+++ b/src/communication/websocket/auth.hpp
@@ -23,7 +23,7 @@ class SafeAuth {
  public:
   explicit SafeAuth(utils::Synchronized<auth::Auth, utils::WritePrioritizedRWLock> *auth) : auth_{auth} {}
 
-  bool Authenticate(const std::string &username, const std::string &password);
+  bool Authenticate(const std::string &username, const std::string &password) const;
 
  private:
   utils::Synchronized<auth::Auth, utils::WritePrioritizedRWLock> *auth_;
--- a/src/communication/websocket/auth.hpp
+++ b/src/communication/websocket/auth.hpp
@@ -25,7 +25,7 @@ class SafeAuth {
 
   bool Authenticate(const std::string &username, const std::string &password) const;
 
-  bool DoesUserHasPermission(const std::string &username, auth::Permission permission) const;
+  bool HasUserPermission(const std::string &username, auth::Permission permission) const;
 
  private:
   utils::Synchronized<auth::Auth, utils::WritePrioritizedRWLock> *auth_;
--- a/src/communication/websocket/auth.hpp
+++ b/src/communication/websocket/auth.hpp
@@ -1,0 +1,31 @@
+// Copyright 2021 Memgraph Ltd.
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
+// License, and you may not use this file except in compliance with the Business Source License.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#pragma once
+
+#include <string>
+
+#include "auth/auth.hpp"
+#include "utils/spin_lock.hpp"
+#include "utils/synchronized.hpp"
+
+namespace communication::websocket {
+
+class SafeAuth {
+ public:
+  explicit SafeAuth(utils::Synchronized<auth::Auth, utils::WritePrioritizedRWLock> *auth) : auth_{auth} {}
+
+  bool Authenticate(const std::string &username, const std::string &password);
+
+ private:
+  utils::Synchronized<auth::Auth, utils::WritePrioritizedRWLock> *auth_;
+};
+}  // namespace communication::websocket
--- a/src/communication/websocket/auth.hpp
+++ b/src/communication/websocket/auth.hpp
@@ -27,6 +27,8 @@ class SafeAuth {
 
   bool HasUserPermission(const std::string &username, auth::Permission permission) const;
 
+  bool HasAnyUsers() const;
+
  private:
   utils::Synchronized<auth::Auth, utils::WritePrioritizedRWLock> *auth_;
 };
--- a/src/communication/websocket/auth.hpp
+++ b/src/communication/websocket/auth.hpp
@@ -25,6 +25,8 @@ class SafeAuth {
 
   bool Authenticate(const std::string &username, const std::string &password) const;
 
+  bool DoesUserHasPermission(const std::string &username, auth::Permission permission) const;
+
  private:
   utils::Synchronized<auth::Auth, utils::WritePrioritizedRWLock> *auth_;
 };
--- a/src/communication/websocket/listener.cpp
+++ b/src/communication/websocket/listener.cpp
@@ -27,7 +27,7 @@ void Listener::WriteToAll(std::shared_ptr<std::string> message) {
   }
 }
 
-Listener::Listener(boost::asio::io_context &ioc, tcp::endpoint endpoint, SafeAuth *auth)
+Listener::Listener(boost::asio::io_context &ioc, tcp::endpoint endpoint, SafeAuth &auth)
     : ioc_(ioc), acceptor_(ioc), auth_(auth) {
   boost::beast::error_code ec;
 
--- a/src/communication/websocket/listener.cpp
+++ b/src/communication/websocket/listener.cpp
@@ -27,7 +27,7 @@ void Listener::WriteToAll(std::shared_ptr<std::string> message) {
   }
 }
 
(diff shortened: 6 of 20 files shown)


src/communication/websocket/session.hpp
```diff
@@ -37,18 +39,21 @@ class Session : public std::enable_shared_from_this<Session> {
...
   void DoRead();
   void OnRead(boost::beast::error_code ec, size_t bytest_transferred);
+  bool Authenticate();
 
   boost::beast::websocket::stream<boost::beast::tcp_stream> ws_;
   boost::beast::flat_buffer buffer_;
   std::deque<std::shared_ptr<std::string>> messages_;
   boost::asio::strand<decltype(ws_)::executor_type> strand_;
   std::atomic<bool> connected_{false};
+  std::atomic<bool> authenticated_{false};
```
antonio2368, line 56: we use strand to make the Session thread-safe so you don't have to use `atomic`. `connected_` is atomic because it's accessed from outside of the implementation (`IsConnected`) function.

jbajic: aha I was confused why was `connected_` atomic, and did not want to risk it so I've made `authenticated_` as well, that makes sense

src/communication/websocket/session.cpp
```diff
@@ -86,13 +91,41 @@ void Session::DoRead() {
...
 
-void Session::OnRead(boost::beast::error_code ec, size_t /*bytes_transferred*/) {
+bool Session::Authenticate() {
+  try {
+    const auto creds = nlohmann::json::parse(boost::beast::buffers_to_string(buffer_.data()));
+    if (auth_->Authenticate(creds.at("username").get<std::string>(), creds.at("password").get<std::string>())) {
+      authenticated_ = true;
+      return true;
+    }
+    return false;
```
jbajic, line 101: redundant return

src/communication/websocket/auth.cpp
```diff
@@ -0,0 +1,25 @@
...
+// licenses/APL.txt.
+
+#include "communication/websocket/auth.hpp"
+
+namespace communication::websocket {
+
+bool SafeAuth::Authenticate(const std::string &username, const std::string &password) {
+  std::optional<auth::User> maybe_user;
+  {
+    // TODO change lock to readlock, auth now creates user...
```
antaljanosbenjamin, line 19: What about this?

src/communication/websocket/auth.hpp
```diff
@@ -0,0 +1,31 @@
+// Copyright 2021 Memgraph Ltd.
```
antaljanosbenjamin, line 1:
```suggestion
// Copyright 2022 Memgraph Ltd.
```

src/communication/websocket/session.hpp
```diff
@@ -11,6 +11,9 @@
 
 #pragma once
 
+#include "utils/synchronized.hpp"
```
antaljanosbenjamin, line 14: This is in wrong place

src/communication/websocket/session.cpp
```diff
@@ -88,13 +91,40 @@ void Session::DoRead() {
...
+  try {
+    const auto creds = nlohmann::json::parse(boost::beast::buffers_to_string(buffer_.data()));
+    if (auth_->Authenticate(creds.at("username").get<std::string>(), creds.at("password").get<std::string>())) {
+      authenticated_ = true;
+      return true;
+    }
+  } catch (const nlohmann::json::out_of_range &out_of_range) {
+    spdlog::error("Missing username or password: {}!", out_of_range.what());
+  } catch (const nlohmann::json::parse_error &parse_error) {
+    spdlog::error("Parsing error: {}!", parse_error.what());
```
antaljanosbenjamin, line 104:
```suggestion
    spdlog::error("Cannot parse JSON containing username and password: {}!", parse_error.what());
```

Or something similar that has some information about what we were trying to parse.

antaljanosbenjamin (changes requested):

src/auth/CMakeLists.txt
```diff
@@ -8,7 +8,7 @@ find_package(Seccomp REQUIRED)
 
 add_library(mg-auth STATIC ${auth_src_files})
 target_link_libraries(mg-auth json libbcrypt gflags fmt::fmt)
-target_link_libraries(mg-auth mg-utils mg-kvstore)
+target_link_libraries(mg-auth mg-utils mg-kvstore mg-license )
```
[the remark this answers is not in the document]
jbajic: No problem!

src/communication/websocket/auth.cpp
```diff
@@ -0,0 +1,25 @@
+// Copyright 2021 Memgraph Ltd.
```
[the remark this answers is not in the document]
jbajic: Currently the license checker will fail, I will correct these when we merge new license checker

src/communication/websocket/auth.cpp
jbajic: Well, this is more to do with this method from auth module:
https://github.com/memgraph/memgraph/blob/master/src/auth/auth.cpp#L70-L162
And the issue is I am not sure everything that I might affect by refactoring this, and not sure how much time would it take me to do it...

src/communication/websocket/auth.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+#include <string>
+
+#include "auth/auth.hpp"
+#include "utils/spin_lock.hpp"
+#include "utils/synchronized.hpp"
+
+namespace communication::websocket {
+
+class SafeAuth {
```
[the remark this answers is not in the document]
jbajic: I will leave it for tests, and do it next iteration

src/communication/websocket/auth.cpp
antaljanosbenjamin: No, it doesn't worth it now. I check the auth module, and I just don't like how it is written. Especially the `FLAGS_auth_module_create_missing_user` variable is super scary to me. I think that shouldn't be part of the module, because if that is true, then you cannot enforce the authentication without creating new user.

So right into my next question which is also meant for @antonio2368 and @kostasrim: do you think enabling a newly created user reading all the log messages is fine? I would say the user should have very high privileges in order to do this.

jbajic: I think it makes sense to add another role for log reading over websocket

src/memgraph.cpp
```diff
@@ -1204,7 +1205,8 @@ int main(int argc, char **argv) {
                             []() -> nlohmann::json { return query::plan::CallProcedure::GetAndResetCounters(); });
   }
 
-  communication::websocket::Server websocket_server{{"0.0.0.0", 7444}};
+  communication::websocket::SafeAuth safe_auth{&auth};
+  communication::websocket::Server websocket_server{{"0.0.0.0", 7444}, safe_auth};
```
kostasrim, line 1209: `safe_auth` is `copyable` and very cheap i.e., a copy of the ptr, therefore you don't really need to have a member reference. Ideally this will look like:

```
  communication::websocket::Server websocket_server{{"0.0.0.0", 7444}, {&auth}};
```

src/communication/websocket/session.hpp
```diff
@@ -35,18 +40,21 @@ class Session : public std::enable_shared_from_this<Session> {
...
   void OnRead(boost::beast::error_code ec, size_t bytest_transferred);
+  bool Authenticate();
 
   boost::beast::websocket::stream<boost::beast::tcp_stream> ws_;
   boost::beast::flat_buffer buffer_;
   std::deque<std::shared_ptr<std::string>> messages_;
   boost::asio::strand<decltype(ws_)::executor_type> strand_;
   std::atomic<bool> connected_{false};
+  bool authenticated_{false};
+  SafeAuth &auth_;
```
kostasrim, line 58:
```suggestion
  SafeAuth auth_;
```

src/communication/websocket/server.hpp
```diff
@@ -26,10 +28,10 @@ class Server final {
   using tcp = boost::asio::ip::tcp;
 
  public:
-  explicit Server(io::network::Endpoint endpoint)
+  explicit Server(io::network::Endpoint endpoint, SafeAuth &auth)
```
kostasrim, line 31:
```suggestion
  explicit Server(io::network::Endpoint endpoint, SafeAuth auth)
```

src/communication/websocket/listener.cpp
```diff
@@ -26,7 +27,8 @@ void Listener::WriteToAll(std::shared_ptr<std::string> message) {
   }
 }
 
-Listener::Listener(boost::asio::io_context &ioc, tcp::endpoint endpoint) : ioc_(ioc), acceptor_(ioc) {
+Listener::Listener(boost::asio::io_context &ioc, tcp::endpoint endpoint, SafeAuth &auth)
```
kostasrim, line 30:
```suggestion
Listener::Listener(boost::asio::io_context &ioc, tcp::endpoint endpoint, SafeAuth auth)
```

src/communication/websocket/listener.hpp
```diff
@@ -39,13 +41,14 @@ class Listener : public std::enable_shared_from_this<Listener> {
   void WriteToAll(std::shared_ptr<std::string> message);
 
  private:
-  Listener(boost::asio::io_context &ioc, tcp::endpoint endpoint);
+  Listener(boost::asio::io_context &ioc, tcp::endpoint endpoint, SafeAuth &auth);
```
kostasrim, line 44:
```suggestion
  Listener(boost::asio::io_context &ioc, tcp::endpoint endpoint, SafeAuth auth);
```

src/communication/websocket/listener.hpp
```diff
@@ -39,13 +41,14 @@ class Listener : public std::enable_shared_from_this<Listener> {
...
 
   void DoAccept();
   void OnAccept(boost::beast::error_code ec, tcp::socket socket);
 
   boost::asio::io_context &ioc_;
   tcp::acceptor acceptor_;
   utils::Synchronized<std::list<std::shared_ptr<Session>>, utils::SpinLock> sessions_;
+  SafeAuth &auth_;
```
kostasrim, line 52:
```suggestion
  SafeAuth auth_;
```

src/memgraph.cpp
jbajic: I don't think that SafeAuth is saved to copy since it holds a pointer to auth, which has to be synced across all modules that want to use auth (pointer to lock on `auth` => https://github.com/memgraph/memgraph/blob/00cf230b427f639247d6407a20f6285a3318bd8f/src/communication/websocket/auth.hpp#L29). I guess to signify that it makes sense to make SageAuth uncopiable, what do you think @kostasrim ?

src/communication/websocket/auth.cpp
antonio2368: As far as I know, modules are a solution for a really specific problem we solved for one of our customers, so I wouldn't put too much thought into it now. All the flags seem really specific and have questionable side effects. We should protect access to logs with some kind of permission, that's for sure.

@jbajic can you try and add that as part of this PR?

src/communication/websocket/session.cpp
```diff
@@ -88,13 +91,40 @@ void Session::DoRead() {
...
+  if (!authenticated_) {
+    auto response = nlohmann::json();
+    if (Authenticate()) {
+      response["success"] = true;
+      response["message"] = "User has been successfully authenticated!";
+    } else {
+      response["success"] = false;
+      response["message"] = "Authentication failed!";
+    }
+    Write(std::make_shared<std::string>(response.dump()));
```
antonio2368, line 125: I would close the connection here.
I see the authentication JSON as part of the handshake. Even though the WS passed, we still don't know if the user is allowed to interact with us using that connection in any way.
But I cannot say anything for sure, maybe it's just an extra inconvenience for the client-side without extra benefits.

@tonilastre What do you think?
Lab <-> Memgraph make a WebSocket connection. 
Lab sends a JSON with auth details as the first message, but it's invalid. 
Would you destroy the WS connection or keep it alive so Memgraph gives Lab multiple chances to send valid credentials on the same established connection?

src/memgraph.cpp
antonio2368, line 1209:
```suggestion
  communication::websocket::Server websocket_server{{"0.0.0.0", 7444}, communication::websocket::SafeAuth{&auth}};
```

src/communication/websocket/session.cpp
tonilastre: Hmm. Both ways would work. The Lab will send the credentials only once because we will establish a WS connection after successfully connecting via bolt (our standard connection).
From that point, closing a connection or keeping it alive will be the same because we won't try it out again. The user needs to log out in the Lab and try out new credentials which will definitely yield a new WS connection.

Even though the purpose of the WS connection is the reuse it as much as you can, this is somehow a special case because of the handshake.

**Future-proof**
In the future, I am seeing Lab as purely a web product that won't use bolt but WS/WSS protocol and everything will go through it - with this in mind, it still makes more sense to kill the connection server-side after incorrect credentials. Here is why: Let's say I have a public Memgraph, then someone can just create WS connections towards my Memgraph and send incorrect credentials. I will soon go out of memory because every WS connection (unauthorized) will be keep-alive. We don't want that.

**Recap**
I would suggest destroying the WS connection after the invalid credentials.

antonio2368: Oh, the `Future-proof` point makes a lot of sense. We would like to give users the ability to connect to that WS server by themselves, not just through Lab. So a scenario you described can easily happen.

@jbajic We already agreed on that but just to have it here, we should try destroying the connection if the authorization fails.

src/communication/websocket/session.cpp
```diff
@@ -9,11 +9,14 @@
...
-#define BOOST_ASIO_USE_TS_EXECUTOR_AS_DEFAULT
-
 #include "communication/websocket/session.hpp"
 
+#include <spdlog/spdlog.h>
 #include <boost/asio/bind_executor.hpp>
+#include <boost/beast/core/buffers_to_string.hpp>
+#include <exception>
+#include <memory>
+#include <stdexcept>
```
antonio2368, line 19: Fix the include groups (std / external libs / our headers)

src/communication/websocket/session.cpp
```diff
@@ -88,13 +91,43 @@ void Session::DoRead() {
...
+    auto response = nlohmann::json();
+    if (Authenticate()) {
+      response["success"] = true;
+      response["message"] = "User has been successfully authenticated!";
+      Write(std::make_shared<std::string>(response.dump()));
+    } else {
+      response["success"] = false;
+      response["message"] = "Authentication failed!";
+      Write(std::make_shared<std::string>(response.dump()));
+      ws_.close("Authentication failed!");
```
antonio2368, line 126:
```suggestion
      ws_.close("Authentication failed!");
      connected_.store(false, std::memory_order_relaxed);
```

src/communication/websocket/session.cpp
```diff
@@ -88,13 +91,43 @@ void Session::DoRead() {
...
+}
+
+void Session::OnRead(const boost::beast::error_code ec, const size_t /*bytes_transferred*/) {
   if (ec == boost::beast::websocket::error::closed) {
     messages_.clear();
     connected_.store(false, std::memory_order_relaxed);
     return;
   }
 
+  if (!authenticated_) {
```
antonio2368, line 116: You need to check before Write if it's authenticated.

antonio2368 (changes requested):

src/communication/websocket/session.cpp
jbajic: What do you mean before `Write`? I am checking if the user is authenticated and if not I try to authenticate him, is there something wrong here?

antonio2368: You don't want to send the logs to the user that isn't authorized and authenticated for that.
So Write should check if the user is authorized before writing anything on the socket.

jbajic: Aha ok then I need to add the same logic to the Write method, I was confused why you commented this here?

antonio2368: Sorry didn't see a better place to write it as you didn't modify the method.

src/communication/websocket/auth.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+#include "communication/websocket/auth.hpp"
+
+namespace communication::websocket {
+
+bool SafeAuth::Authenticate(const std::string &username, const std::string &password) const {
+  return auth_->Lock()->Authenticate(username, password).has_value();
+}
+
+bool SafeAuth::DoesUserHasPermission(const std::string &username, const auth::Permission permission) const {
```
antonio2368, line 20:
```suggestion
bool SafeAuth::UserHasPermission(const std::string &username, const auth::Permission permission) const {
```

src/communication/websocket/session.cpp
```diff
@@ -46,14 +50,30 @@ void Session::Write(std::shared_ptr<std::string> message) {
   }
   boost::asio::dispatch(strand_, [message = std::move(message), shared_this = shared_from_this()]() mutable {
     shared_this->messages_.push_back(std::move(message));
-
+    if (!shared_this->authenticated_) {
+      return;
+    }
```
antonio2368, line 55: The check should go before the push_back, `DoWrite` is just the operation to write message from the queue. You still pushed the message to the queue and it will be written to websocket, just not immediatelly. That is not something we want, we want to forbid pushing the message to the queue if the user is not authorized.

Also, I would move the `connected_` check also inside this lambda, just to have a better thread-safety.

src/communication/websocket/session.cpp
```diff
@@ -86,14 +106,68 @@ void Session::DoRead() {
...
+  connected_.store(false, std::memory_order_relaxed);
+}
+
+bool Session::Authenticate(const nlohmann::json &creds) {
+  return auth_.Authenticate(creds.at("username").get<std::string>(), creds.at("password").get<std::string>());
+}
+
+bool Session::Authorize(const nlohmann::json &creds) {
+  return auth_.DoesUserHasPermission(creds.at("username").get<std::string>(), auth::Permission::WEBSOCKET);
+}
```
antonio2368, line 129: This can be a single method `Authenticate` which simply checks if the user exists and, if it's enterprise build, checks if he has the correct permissions.

src/communication/websocket/session.cpp
```diff
@@ -86,14 +106,68 @@ void Session::DoRead() {
...
+    auto response = nlohmann::json();
+    try {
+      const auto creds = nlohmann::json::parse(boost::beast::buffers_to_string(buffer_.data()));
+      buffer_.consume(buffer_.size());
+
+      if (!Authenticate(creds)) {
+        response["success"] = false;
+        response["message"] = "Authentication failed!";
+        close_ = true;
+        QuickWrite(response.dump());
```
antonio2368, line 148: why don't you just do 
```
MG_ASSERT(messages_.empty());
messages_.push_back(response.dump()); 
DoWrite()
```
you can even extract it into a lambda.
And the call explicitly DoClose where it needs to be called.

src/communication/websocket/session.cpp
jbajic: Whoops, you are right... Will move the connected check as well

src/communication/websocket/session.cpp
jbajic: I made it like this since we want to form different response messages depending on if the Authentication or Authorization fails. Do you think it better to pass ref to reponse and let them change the message?

antonio2368: You can use `utils::BasicResult<std::string>` for that (returns an error message if something went wrong).

src/auth/models.hpp
```diff
@@ -36,17 +36,18 @@ enum class Permission : uint64_t {
...
 // clang-format on
 
 // Constant list of all available permissions.
-const std::vector<Permission> kPermissionsAll = {Permission::MATCH,      Permission::CREATE,    Permission::MERGE,
-                                                 Permission::DELETE,     Permission::SET,       Permission::REMOVE,
-                                                 Permission::INDEX,      Permission::STATS,     Permission::CONSTRAINT,
-                                                 Permission::DUMP,       Permission::AUTH,      Permission::REPLICATION,
-                                                 Permission::DURABILITY, Permission::READ_FILE, Permission::FREE_MEMORY,
-                                                 Permission::TRIGGER,    Permission::CONFIG,    Permission::STREAM};
+const std::vector<Permission> kPermissionsAll = {
```
kostasrim, line 45: This has `internal_linkage` by default because of `const` which means that a vector is created on each translation unit. I know this is a little beyond the scope of this PR but would you mind adding a more idiomatic way like:
```
inline const auto = []() -> const std::array<Permision, 19>& { 
  static std::array<Permission, 19> cache;
  static bool once = true;
  if(once) [[unlikely]] {
    std::generate(cache.begin(), cache.end(), [n = 0] { return static_cast<Permission>(n++); //watch post increment });
    once = false;
  }
  return cache_;
};
```

p.s. with reflection we could extract the total number of fields in the enum and abstract away the size of the array (not hardcoded with 19) which would then introduce a factory method for every use of the above pattern.

src/communication/websocket/auth.cpp
kostasrim:
```suggestion
bool SafeAuth::HasUserPermission(const std::string &username, const auth::Permission permission) const {
```

kostasrim: :eyes: :eyes:

src/communication/websocket/session.cpp
```diff
@@ -86,14 +106,68 @@ void Session::DoRead() {
...
   if (ec == boost::beast::websocket::error::closed) {
     messages_.clear();
     connected_.store(false, std::memory_order_relaxed);
     return;
   }
 
-  buffer_.consume(buffer_.size());
+  if (!authenticated_) {
+    auto response = nlohmann::json();
+    try {
```
kostasrim, line 140: If this throws it will happily call `DoRead`, is this intented?

src/communication/websocket/auth.cpp
jbajic: I am always for functions that begin with verb

src/communication/websocket/session.cpp
jbajic: Good catch, I will close the connection here as well

src/auth/models.hpp
jbajic: This approach requires that I need to call `kPermissionsAll()` in order to get the ref to it, it's kind of weird and complicated to me. Can't we just make it `cosntexpr std::array`?

antaljanosbenjamin: Maybe using `constexpr std::array` instead? Or as this vector is only used in the source file, move it to the source file.

src/communication/websocket/auth.cpp
```diff
@@ -16,4 +16,11 @@ namespace communication::websocket {
 bool SafeAuth::Authenticate(const std::string &username, const std::string &password) const {
   return auth_->Lock()->Authenticate(username, password).has_value();
 }
+
+bool SafeAuth::DoesUserHasPermission(const std::string &username, const auth::Permission permission) const {
+  if (const auto user = auth_->Lock()->GetUser(username); user) {
+    return user->GetPermissions().Has(permission) == auth::PermissionLevel::GRANT;
+  }
+  return false;
```
antaljanosbenjamin, line 26: What if memgraph doesn't have any users? Shouldn't we allow then everything similarly to `AuthChecker::IsUserAuthorized` in `memgraph.cpp`? I think it would the best to push down that AuthChecker, because that is meant to do this exactly: check if a user has permission or not.

src/communication/websocket/session.cpp
antaljanosbenjamin: Then don't forget to close the socket in that case.

src/communication/websocket/session.cpp
```diff
@@ -41,19 +50,32 @@ void Session::Run() {
...
     shared_this->messages_.push_back(std::move(message));
-
     if (shared_this->messages_.size() > 1) {
       return;
     }
     shared_this->DoWrite();
   });
 }
 
+void Session::QuickWrite(std::string message) {
```
antaljanosbenjamin, line 68: Not needed anymore in my opinion.

antaljanosbenjamin (changes requested):

src/communication/websocket/auth.cpp
jbajic: Yes that makes sense, to only authenticate/authorize if there are any users

src/auth/models.cpp
```diff
@@ -30,6 +30,14 @@ DEFINE_string(auth_password_strength_regex, default_password_regex.data(),
...
 namespace auth {
 
+// Constant list of all available permissions.
+constexpr std::array<Permission, 19> kPermissionsAll = {
+    Permission::MATCH,      Permission::CREATE,    Permission::MERGE,       Permission::DELETE,
+    Permission::SET,        Permission::REMOVE,    Permission::INDEX,       Permission::STATS,
+    Permission::CONSTRAINT, Permission::DUMP,      Permission::AUTH,        Permission::REPLICATION,
+    Permission::DURABILITY, Permission::READ_FILE, Permission::FREE_MEMORY, Permission::TRIGGER,
+    Permission::CONFIG,     Permission::STREAM,    Permission::WEBSOCKET};
+
```
antonio2368, line 40:
```suggestion
constexpr std::array kPermissionsAll = {
    Permission::MATCH,      Permission::CREATE,    Permission::MERGE,       Permission::DELETE,
    Permission::SET,        Permission::REMOVE,    Permission::INDEX,       Permission::STATS,
    Permission::CONSTRAINT, Permission::DUMP,      Permission::AUTH,        Permission::REPLICATION,
    Permission::DURABILITY, Permission::READ_FILE, Permission::FREE_MEMORY, Permission::TRIGGER,
    Permission::CONFIG,     Permission::STREAM,    Permission::WEBSOCKET};

```

src/communication/websocket/session.cpp
```diff
@@ -86,14 +100,78 @@ void Session::DoRead() {
...
+
+void Session::OnRead(const boost::beast::error_code ec, const size_t /*bytes_transferred*/) {
   if (ec == boost::beast::websocket::error::closed) {
     messages_.clear();
     connected_.store(false, std::memory_order_relaxed);
     return;
   }
 
-  buffer_.consume(buffer_.size());
+  if (!authenticated_ && auth_.HasAnyUsers()) {
```
antonio2368, line 136:
```suggestion
  if (auth_.HasAnyUsers() && !authenticated_) {
```
I think it's better to do the simpler operation first because of short circuting

antonio2368 (commented):
Add test to `CypherMainVisitorTest, GrantPrivilege` in the cypher_main_visitor.cpp for `WEBSOCKET` permission.

src/communication/websocket/session.cpp
```diff
@@ -86,14 +100,78 @@ void Session::DoRead() {
...
-  buffer_.consume(buffer_.size());
+  if (auth_.HasAnyUsers() && !authenticated_) {
+    auto response = nlohmann::json();
+    auto auth_failed = [this, &response]() {
+      response["success"] = false;
+      MG_ASSERT(messages_.empty());
+      messages_.push_back(make_shared<std::string>(response.dump()));
+      close_ = true;
+      DoWrite();
+    };
```
antonio2368, line 144: Why don't you accept the error message as an argument to lambda? That way you are sure that they are stored under the same key (`message`).

antonio2368 (approved):
Tested it out, works nicely!
Good job!

antonio2368:
Don't forget to fix the Community build error.

src/communication/websocket/session.cpp
```diff
@@ -86,14 +100,76 @@ void Session::DoRead() {
...
+    return LogError(ec, "close");
+  }
+  connected_.store(false, std::memory_order_relaxed);
+}
+
+utils::BasicResult<std::string> Session::Authorize(const nlohmann::json &creds) {
+  if (!auth_.Authenticate(creds.at("username").get<std::string>(), creds.at("password").get<std::string>())) {
+    return {"Authentication failed!"};
+  }
+#ifndef MG_ENTERPRISE
```
antonio2368, line 121:
```suggestion
#ifdef MG_ENTERPRISE
```
Is what you probably meant.

jbajic: Can you remind me why do we have `MG_ENTERPRISE`, when we have only one binary?

antonio2368: If people insist on having pure open-source package, they can build it with disabled `MG_ENTERPRISE`.
Authentication is part of the binary regardless, so we don't actually need to protect this with the flag.

src/communication/websocket/session.cpp
antaljanosbenjamin: Isn't checking `authenticated_` the simpler, as it is just a simple bool member?

antonio2368: That is right, I thought it was a call to Authenticate for some reason.... Sorry @jbajic

src/communication/websocket/auth.cpp
antaljanosbenjamin: @jbajic pointed out that `AuthChecker` works with privileges from the query engine point of view, and not with permissions. I agree with him that the two shouldn't be mixed, so for now it is okay to go with a different solution.

antaljanosbenjamin (approved):

PR #323, opened 2022-01-10
Adapt header-checker

jbajic:
[master < Task] PR

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -32,7 +32,7 @@ for file in $modified_files; do
   fi
 
   echo "Running header checker..."
-  $project_folder/tools/header-checker.py $tmpdir/$file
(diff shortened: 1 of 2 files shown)


antonio2368 (approved):
Awesome!

antaljanosbenjamin (approved):
Maybe, if the only difference is the year, the script could update it by replacing `Copyright \d\d\d\d Memgraph Ltd.` with `Copyright {year} Memgraph Ltd`. Or it is too much?

jbajic:
> Maybe, if the only difference is the year, the script could update it by replacing `Copyright \d\d\d\d Memgraph Ltd.` with `Copyright {year} Memgraph Ltd`. Or it is too much?

I think that makes sense, I would do that for the pre-commit hook, and we could still use the check in the job

antaljanosbenjamin (approved):

PR #324, opened 2022-01-12
[master < MG-fix] Unsafe git protocol fix

antonio2368:


kostasrim (approved):

antaljanosbenjamin (approved):
I like the conversion to patch file!

PR #326, opened 2022-01-13
[master < MG] Use libs from toolchain

antonio2368:


libs/CMakeLists.txt
```diff
@@ -8,6 +8,15 @@ if (NPROC EQUAL 0)
...
 endif()
 
+find_package(Boost REQUIRED)
+find_package(BZip2 REQUIRED)
+find_package(Threads REQUIRED)
+set(GFLAGS_NOTHREADS OFF)
+find_package(gflags REQUIRED)
+find_package(fmt)
+find_package(Jemalloc REQUIRED)
+find_package(ZLIB REQUIRED)
```
gitbuda, line 18: @antonio2368 did you consider specifying versions here?

antonio2368: That could introduce some way of verifying that the correct packages are used so I see no reason not to use it.
@antaljanosbenjamin what do you think?
I know `Jemalloc` is custom logic, but everything should be fine.

jbajic (approved):

gitbuda (approved):

PR #327, opened 2022-01-17
[E0108-MG < T0783-MG] WSS support

antonio2368:


src/communication/websocket/listener.cpp
```diff
@@ -71,12 +71,18 @@ void Listener::OnAccept(boost::beast::error_code ec, tcp::socket socket) {
     return LogError(ec, "accept");
   }
 
-  {
+  auto session = Session::Create(std::move(socket), *context_, auth_);
+
+  if (auto maybe_error = session->Run(); maybe_error.HasError()) {
+    spdlog::warn(maybe_error.GetError());
+  } else {
+    spdlog::critical("Adding new session");
```
antonio2368, line 79: remove this

src/communication/websocket/session.cpp
```diff
@@ -19,42 +19,79 @@
...
+  auto result = ExecuteForWebsocket([](auto &&ws) -> bool {
+    // Accept the websocket handshake
+    boost::beast::error_code ec;
+    ws.accept(ec);
+    if (ec) {
+      LogError(ec, "accept");
+      return false;
+    }
+    return true;
+    ;
```
antonio2368, line 75: extra

src/communication/websocket/session.cpp
```diff
@@ -19,42 +19,78 @@
...
+bool Session::Run() {
+  ExecuteForWebsocket([](auto &&ws) {
+    ws.set_option(boost::beast::websocket::stream_base::timeout::suggested(boost::beast::role_type::server));
+
+    ws.set_option(boost::beast::websocket::stream_base::decorator([](boost::beast::websocket::response_type &res) {
+      res.set(boost::beast::http::field::server, "Memgraph WS");
+    }));
+  });
+
+  if (auto *ssl_ws = std::get_if<SSLWebSocket>(&ws_)) {
```
antaljanosbenjamin, line 57: I think with an explicit condition it would be better:

```suggestion
  if (auto *ssl_ws = std::get_if<SSLWebSocket>(&ws_); ssl_ws != nullptr) {
```

or at least
```suggestion
  if (auto *ssl_ws = std::get_if<SSLWebSocket>(&ws_); ssl_ws) {
```
I know that the result is the same, but these two make it more readable in my opinion (I like explicit more than implicit, so I might be alone.)

src/communication/websocket/session.cpp
```diff
@@ -173,4 +211,21 @@ void Session::OnRead(const boost::beast::error_code ec, const size_t /*bytes_tra
   DoRead();
 }
 
+bool Session::IsAuthenticated() const noexcept { return authenticated_ || !auth_.HasAnyUsers(); }
```
antaljanosbenjamin, line 214: `auth_.HasAnyUsers()` is not `noexcept`, so I think this function also cannot be `noexcept`.

src/communication/websocket/session.cpp
```diff
@@ -173,4 +211,21 @@ void Session::OnRead(const boost::beast::error_code ec, const size_t /*bytes_tra
...
+
+void Session::DoShutdown() {
+  std::visit(utils::Overloaded{[this](SSLWebSocket &ssl_ws) {
+                                 boost::beast::get_lowest_layer(ssl_ws).expires_after(std::chrono::seconds(30));
```
antaljanosbenjamin, line 218: Why `expires_after` is necessary to be called on the `tcp_stream` here, but not in case of `PlainText`?

antaljanosbenjamin (changes requested):
Tomorrow I will double check the session, because it has a lot of important details.

src/communication/context.cpp
```diff
@@ -73,11 +73,9 @@ SSL_CTX *ClientContext::context() { return ctx_; }
 
 bool ClientContext::use_ssl() { return use_ssl_; }
 
-ServerContext::ServerContext() : use_ssl_(false), ctx_(nullptr) {}
-
 ServerContext::ServerContext(const std::string &key_file, const std::string &cert_file, const std::string &ca_file,
                              bool verify_peer)
-    : use_ssl_(true),
+    :
 #if OPENSSL_VERSION_NUMBER < 0x10100000L
```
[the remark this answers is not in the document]
antonio2368: I would need to add `:` to `else` also. Do you still prefer it that way?

src/communication/context.hpp
```diff
@@ -95,16 +92,13 @@ class ServerContext final {
...
-
   SSL_CTX *context();
+  SSL_CTX *context_clone();
 
   bool use_ssl();
 
  private:
-  bool use_ssl_;
-  SSL_CTX *ctx_;
+  std::optional<SSL_CTX *> ctx_;
```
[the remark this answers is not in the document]
antonio2368: It's the `use_ssl` that pulled me to `std::optional` but you're right both `optional` and `use_ssl` are unecessary.

src/communication/context.cpp
```diff
@@ -107,46 +105,44 @@ ServerContext::ServerContext(const std::string &key_file, const std::string &cer
...
-  other.use_ssl_ = false;
-  other.ctx_ = nullptr;
+  other.ctx_.reset();
 
   return *this;
 }
 
-ServerContext::~ServerContext() {
-  if (use_ssl_) {
-    SSL_CTX_free(ctx_);
```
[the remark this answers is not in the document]
antonio2368: I would say it's a mistake. I removed this before figuring out the `ref` stuff (`context_clone` function). 
I'll check it a bit more but I think it should be returned. Nice catch!

src/communication/websocket/session.cpp
antonio2368: I based it on this example:
https://www.boost.org/doc/libs/1_77_0/libs/beast/example/advanced/server-flex/advanced_server_flex.cpp
I would say that the timeout options from `ws` don't apply to the TCP operations, so you should call it before doing anything on the layers below `Websocket`. That reminded me that we are missing the timeout for the handshake.

src/communication/context.hpp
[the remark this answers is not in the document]
antonio2368: I wanted to touch this part as minimally as possible because the focus of the PR is the WSS. 
The context is used by multiple servers so I would need to change all of the interactions with it.

src/communication/context.cpp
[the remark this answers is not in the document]
antaljanosbenjamin: Sorry, my mistake. I somehow missed the else branch.

src/memgraph.cpp
```diff
@@ -349,39 +349,54 @@ DEFINE_VALIDATED_string(log_level, "WARNING", log_level_help_string.c_str(), {
...
-void ConfigureLogging() {
-  std::vector<spdlog::sink_ptr> loggers;
+struct MemgraphLogger {
+  MemgraphLogger() : log_level(ParseLogLevel()) {
+    std::vector<spdlog::sink_ptr> sinks;
 
-  if (FLAGS_also_log_to_stderr) {
-    loggers.emplace_back(std::make_shared<spdlog::sinks::stderr_color_sink_mt>());
-  }
+    if (FLAGS_also_log_to_stderr) {
```
antaljanosbenjamin, line 365: What about passing this option as a parameter of the constructor? This way I think the responsibilities are clearly separated, the logger no longer depends on the config flag.

src/memgraph.cpp
```diff
@@ -349,39 +349,54 @@ DEFINE_VALIDATED_string(log_level, "WARNING", log_level_help_string.c_str(), {
...
-    local_time = localtime(&current_time);
+      time(&current_time);
+      local_time = localtime(&current_time);
 
-    loggers.emplace_back(std::make_shared<spdlog::sinks::daily_file_sink_mt>(
-        FLAGS_log_file, local_time->tm_hour, local_time->tm_min, false, log_retention_count));
+      sinks.emplace_back(std::make_shared<spdlog::sinks::daily_file_sink_mt>(
+          FLAGS_log_file, local_time->tm_hour, local_time->tm_min, false, log_retention_count));
+    }
+    CreateLoggerFromSink(sinks);
```
antaljanosbenjamin, line 380: I would also move this to a separate function, because by creating the object, we can easily make another one obsolete.

In general I am not sure whether this object is a good abstraction. Multiple instances can be created, but all of them operate over a global resource (default logger). What I am thinking is tweaking this a bit:
1. Make the logger to be a singleton: this way it is explicitly expressed that only one object can exists.
2. Wrap the logic into a function: using function local static variables and `std::call_once` for initialization should be fine. As there is not need for destructor, this might be the easiest solution.

antaljanosbenjamin (changes requested):
I did a second check and I found one more thing.

src/communication/context.hpp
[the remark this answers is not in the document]
kostasrim: Then don't, it's not really worth it!

src/memgraph.cpp
antonio2368: Check it out now. What I wanted to avoid was fetching the logger every time we want to add a new sink because it takes a lock (same for the level). But it actually makes no sense as we shouldn't do that in any possible hot path + we do it only once now and it's during startup.

jbajic (approved):
Looks good to me!

kostasrim (approved):

src/memgraph.cpp
antaljanosbenjamin: I like this much better!

src/communication/context.hpp
```diff
@@ -11,6 +11,7 @@
 
 #pragma once
 
+#include <optional>
```
antaljanosbenjamin, line 14: Unnecessary include.

antaljanosbenjamin (approved):
Just a tiny comment, but everything else looks good. The modified approach to add logger sink is much better in my opinion!

PR #328, opened 2022-01-19
[master < T0791-MG] Add configs for kafka streams

antaljanosbenjamin:
[master < Task] PR

[Changlog + docs update](https://github.com/memgraph/docs/pull/288)


src/query/frontend/opencypher/grammar/MemgraphCypher.g4
```diff
@@ -309,6 +311,8 @@ createStream : kafkaCreateStream | pulsarCreateStream ;
 kafkaCreateStreamConfig : TOPICS topicNames
                         | CONSUMER_GROUP consumerGroup=symbolicNameWithDotsAndMinus
                         | BOOTSTRAP_SERVERS bootstrapServers=literal
+                        | CONFIGS configsMap=literal
+                        | CREDENTIALS credentialsMap=literal
```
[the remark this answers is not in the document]
antonio2368: There is a `visitLiteral` method which returns `PrimitiveLiteral` cast to `Expression *` and `visitMapLiteral` which returns an `std::unordered_map` in `cypher_main_visitior.cpp`. If you use `literal`, the first method is called, `and for `mapLiteral`, `visitMapLiteral` is called.

antaljanosbenjamin:
I have to update the related docs with the new options/results.

kostasrim (commented):
I got a few more comments, will complete this later :)

src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -584,6 +588,21 @@ void MapCommonStreamConfigs(auto &memory, StreamQuery &stream_query) {
...
 
+antlrcpp::Any CypherMainVisitor::visitConfigKeyValuePair(MemgraphCypher::ConfigKeyValuePairContext *ctx) {
+  MG_ASSERT(ctx->literal().size() == 2);
+  return std::pair{ctx->literal(0)->accept(this).as<Expression *>(), ctx->literal(1)->accept(this).as<Expression *>()};
+}
+
+antlrcpp::Any CypherMainVisitor::visitConfigMap(MemgraphCypher::ConfigMapContext *ctx) {
+  std::unordered_map<Expression *, Expression *> map;
+  for (auto *key_value_pair : ctx->configKeyValuePair()) {
+    // If the queries are cached, then only the stripped query is parsed, so the actual keys are cannot be determined
```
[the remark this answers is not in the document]
antaljanosbenjamin: I am happy to elaborate here, but I intentionally kept the comment short.
Basically when a query is parsed in case of the queries are cached, the first step is "stripping" the query, and the caching is based on stripped queries. During stripping, the query is transformed in a way that white-spaces (use only single white-spaces), literals (one value per literal type), casing and many other things are unified. Multiple queries result in the same stripped query, therefore the plan can be reused for them.

For further read please check [`ParseQuery` in cypher_query_interpreter.cpp](https://github.com/memgraph/memgraph/blob/bb1308acc7faeb010764ec06eeb587847bc47610/src/query/cypher_query_interpreter.cpp#L26-L30) and [`stripped.hpp`](https://github.com/memgraph/memgraph/blob/master/src/query/frontend/stripped.hpp).

tests/unit/query_streams.cpp
```diff
@@ -183,6 +197,13 @@ TEST_F(StreamsTest, RestoreStreams) {
       stream_info.common_info.batch_interval = std::chrono::milliseconds((i + 1) * 10);
       stream_info.common_info.batch_size = 1000 + i;
       stream_check_data.owner = std::string{"owner"} + iteration_postfix;
+
+      if (i == 1 || i == 3) {
+        stream_info.configs.emplace(std::string{"sasl.username"}, std::string{"username"} + iteration_postfix);
+      }
+      if (i == 2 || i == 3) {
+        stream_info.credentials.emplace(std::string{"sasl.password"}, std::string{"password"} + iteration_postfix);
+      }
```
jbajic, line 206: Could you elaborate a bit on these conditions?

jbajic (approved):
Good job!

tests/unit/query_streams.cpp
antaljanosbenjamin: Yes, those are just "random" numbers, to make the CONFIGS and CREDENTIALS map vary between consumers. Basically:
- 0 means no config, no credential
- 1 means only config
- 2 means only credential
- 3 means both configuration and credential is set

src/query/frontend/ast/cypher_main_visitor.cpp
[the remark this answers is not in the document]
kostasrim: Wow thank you so much for the explanation! This perfect sense. (I thought that queries are cached the way they are, but it makes sense to strip them because as you say multiple queries result in the same stripped query -- and probably you can do all sorts of optimizations here beyond the scope of this discussion). 

Thank you so much for this <3

tests/e2e/streams/kafka_streams_tests.py
```diff
@@ -421,24 +421,35 @@ def comparison_check(a, b):
...
+        f"CONFIGS {configs} "
+        f"CREDENTIALS {credentials}"
     )
 
-    stream_info = common.execute_and_fetch_all(cursor, f"CALL mg.kafka_stream_info('{stream_name}') YIELD *")
+    stream_info = common.execute_and_fetch_all(
+        cursor, f"CALL mg.kafka_stream_info('{stream_name}') YIELD *")
+
+    reducted_credentials = {key: "<REDUCTED>" for (
+        key, value) in credentials.items()}
```
antonio2368, line 447: why not just `credentials.keys()`

tests/unit/cypher_main_visitor.cpp
```diff
@@ -3803,6 +3842,44 @@ TEST_P(CypherMainVisitorTest, CreateKafkaStream) {
...
+
+    const auto map_as_str = std::invoke([&config_map] {
+      std::stringstream buffer;
+      buffer << '{';
+      if (!config_map.empty()) {
+        const auto &first_item = *config_map.begin();
+        buffer << fmt::format("'{}': '{}'", first_item.first, first_item.second);
+        for (auto it = ++config_map.begin(); it != config_map.end(); ++it) {
+          buffer << fmt::format(", '{}': '{}'", it->first, it->second);
+        }
```
antonio2368, line 3857:
```suggestion
        auto it = config_map.begin();
        buffer << fmt::format("'{}': '{}'", it->first, it->second);
        for (; it != config_map.end(); ++it) {
          buffer << fmt::format(", '{}': '{}'", it->first, it->second);
        }
```
maybe something like this

tests/unit/cypher_main_visitor.cpp
```diff
@@ -3803,6 +3842,44 @@ TEST_P(CypherMainVisitorTest, CreateKafkaStream) {
...
+                                   config_map, {});
+
+    ValidateCreateKafkaStreamQuery(ast_generator,
+                                   fmt::format("CREATE KAFKA STREAM {} TOPICS {} TRANSFORM {} CREDENTIALS {}",
+                                               kStreamName, kTopicName, kTransformName, map_as_str),
+                                   kStreamName, {kTopicName}, kTransformName, "", std::nullopt, std::nullopt, {}, {},
+                                   config_map);
+  };
+
+  const std::array<std::unordered_map<std::string, std::string>, 3> config_maps = {
```
antonio2368, line 3876:
```suggestion
  const std::array config_maps = {
```
Or you don't define type for each element, no need for both

antonio2368 (commented):
Some small comments in the tests, but the main logic looks good!

tests/unit/cypher_main_visitor.cpp
antaljanosbenjamin: Removing the types from each element makes the second (and probably the third) one fail to compile (with error `error: excess elements in struct initializer`). I don't know the exact reason, but as it is not critical, I removed the template types of the array.

tests/unit/cypher_main_visitor.cpp
antonio2368: What about this?

tests/unit/query_streams.cpp
jbajic: Maybe put that into comments so it can be easily understood by glossing over.

kostasrim (approved):

src/integrations/kafka/exceptions.hpp
```diff
@@ -22,37 +22,46 @@ class KafkaStreamException : public utils::BasicException {
...
 class ConsumerFailedToInitializeException : public KafkaStreamException {
  public:
-  ConsumerFailedToInitializeException(const std::string &consumer_name, const std::string &error)
+  ConsumerFailedToInitializeException(std::string_view consumer_name, std::string_view error)
       : KafkaStreamException("Failed to initialize Kafka consumer {} : {}", consumer_name, error) {}
 };
 
+class SettingCustomConfigFailed : public ConsumerFailedToInitializeException {
+ public:
+  SettingCustomConfigFailed(std::string_view consumer_name, std::string_view error, std::string_view key,
```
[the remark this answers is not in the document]
antaljanosbenjamin: You are right, I removed the `const` by reflex, but it was a mistake, will add them.

PR #330, opened 2022-01-24
[master < T0798-MG] Procedures for handling modules

antonio2368:
[master < Task] PR

### DIFF
--- a/src/query/frontend/semantic/required_privileges.cpp
+++ b/src/query/frontend/semantic/required_privileges.cpp
@@ -89,11 +89,11 @@ class PrivilegeExtractor : public QueryVisitor<void>, public HierarchicalTreeVis
         AddPrivilege(privilege);
       }
     };
-    constexpr std::array<std::string_view, 2> module_read_procedures{"mg.get_module", "mg.get_module_files"};
+    constexpr std::array<std::string_view, 2> module_read_procedures{"mg.get_module_file", "mg.get_module_files"};
     add_privilege(module_read_procedures, AuthQuery::Privilege::MODULE_READ);
 
-    constexpr std::array<std::string_view, 3> module_write_procedures{"mg.create_module", "mg.update_module",
-                                                                      "mg.delete_module"};
+    constexpr std::array<std::string_view, 3> module_write_procedures{"mg.create_module_file", "mg.update_module_file",
+                                                                      "mg.delete_module_file"};
     add_privilege(module_write_procedures, AuthQuery::Privilege::MODULE_WRITE);
     return false;
   }
--- a/src/query/frontend/semantic/required_privileges.cpp
+++ b/src/query/frontend/semantic/required_privileges.cpp
@@ -11,6 +11,8 @@
 
 #include "query/frontend/ast/ast.hpp"
 #include "query/frontend/ast/ast_visitor.hpp"
+#include "query/procedure/module.hpp"
+#include "utils/memory.hpp"
 
 namespace query {
 
@@ -83,18 +85,11 @@ class PrivilegeExtractor : public QueryVisitor<void>, public HierarchicalTreeVis
     return false;
   }
   bool PreVisit(CallProcedure &procedure) override {
-    const auto add_privilege = [&](const auto &procedures, const auto privilege) {
-      if (std::any_of(procedures.begin(), procedures.end(),
-                      [&](const auto procedure_name) { return procedure_name == procedure.procedure_name_; })) {
-        AddPrivilege(privilege);
-      }
-    };
-    constexpr std::array<std::string_view, 2> module_read_procedures{"mg.get_module_file", "mg.get_module_files"};
-    add_privilege(module_read_procedures, AuthQuery::Privilege::MODULE_READ);
-
-    constexpr std::array<std::string_view, 3> module_write_procedures{"mg.create_module_file", "mg.update_module_file",
-                                                                      "mg.delete_module_file"};
-    add_privilege(module_write_procedures, AuthQuery::Privilege::MODULE_WRITE);
+    const auto maybe_proc =
+        procedure::FindProcedure(procedure::gModuleRegistry, procedure.procedure_name_, utils::NewDeleteResource());
+    if (maybe_proc && maybe_proc->second->info.required_privilege) {
+      AddPrivilege(*maybe_proc->second->info.required_privilege);
+    }
     return false;
   }
   bool PreVisit(Delete & /*unused*/) override {
--- a/src/query/frontend/semantic/required_privileges.cpp
+++ b/src/query/frontend/semantic/required_privileges.cpp
@@ -89,7 +89,7 @@ class PrivilegeExtractor : public QueryVisitor<void>, public HierarchicalTreeVis
         AddPrivilege(privilege);
       }
     };
-    constexpr std::array<std::string_view, 1> module_read_procedures{"mg.get_module"};
+    constexpr std::array<std::string_view, 2> module_read_procedures{"mg.get_module", "mg.get_module_files"};
     add_privilege(module_read_procedures, AuthQuery::Privilege::MODULE_READ);
 
     constexpr std::array<std::string_view, 3> module_write_procedures{"mg.create_module", "mg.update_module",
--- a/src/query/frontend/semantic/required_privileges.cpp
+++ b/src/query/frontend/semantic/required_privileges.cpp
@@ -1,4 +1,4 @@
-// Copyright 2021 Memgraph Ltd.
+// Copyright 2022 Memgraph Ltd.
 //
 // Use of this software is governed by the Business Source License
 // included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
@@ -82,8 +82,19 @@ class PrivilegeExtractor : public QueryVisitor<void>, public HierarchicalTreeVis
     AddPrivilege(AuthQuery::Privilege::CREATE);
     return false;
   }
-  bool PreVisit(CallProcedure & /*unused*/) override {
-    // TODO: Corresponding privilege
+  bool PreVisit(CallProcedure &procedure) override {
+    const auto add_privilege = [&](const auto &procedures, const auto privilege) {
+      if (std::any_of(procedures.begin(), procedures.end(),
+                      [&](const auto procedure_name) { return procedure_name == procedure.procedure_name_; })) {
+        AddPrivilege(privilege);
+      }
+    };
+    constexpr std::array<std::string_view, 1> module_read_procedures{"mg.get_module"};
+    add_privilege(module_read_procedures, AuthQuery::Privilege::MODULE_READ);
+
+    constexpr std::array<std::string_view, 3> module_write_procedures{"mg.create_module", "mg.update_module",
+                                                                      "mg.delete_module"};
+    add_privilege(module_write_procedures, AuthQuery::Privilege::MODULE_WRITE);
     return false;
   }
   bool PreVisit(Delete & /*unused*/) override {
--- a/src/query/procedure/module.cpp
+++ b/src/query/procedure/module.cpp
@@ -84,6 +84,19 @@ void BuiltinModule::AddTransformation(std::string_view name, mgp_trans trans) {
 
 namespace {
 
+auto WithUpgradedLock(auto *lock, const auto &function) {
+  lock->unlock_shared();
+  try {
+    function();
+    // There's no finally in C++, but we have to return our original READ lock
+    // state in any possible case.
+  } catch (...) {
+    lock->lock_shared();
+    throw;
+  }
+  lock->lock_shared();
+};
+
 void RegisterMgLoad(ModuleRegistry *module_registry, utils::RWLock *lock, BuiltinModule *module) {
   // Loading relies on the fact that regular procedure invocation through
   // CallProcedureCursor::Pull takes ModuleRegistry::lock_ with READ access. To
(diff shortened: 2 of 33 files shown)


src/query/procedure/module.cpp
```diff
@@ -560,6 +906,9 @@ bool PythonModule::Close() {
     py_module_ = py::Object(nullptr);
     return false;
   }
+
+  // Remove the cached bytecode if it's present
+  std::filesystem::remove_all(file_path_.parent_path() / "__pycache__");
```
antaljanosbenjamin, line 911: Just to be sure: `__pycache__` is only used when the interpreter loads a module and it is not used after that, right? It would be awful to delete a file while the interpreter is using it.

tests/e2e/module_file_manager/module_file_manager.cpp
```diff
@@ -0,0 +1,170 @@
+// Copyright 2022 Memgraph Ltd.
```
antaljanosbenjamin, line 1: I would add some cases where at least the error handling is checked, because it wouldn't be nice if because of a bug memgraph would crash in invalid cases.

antaljanosbenjamin (changes requested):
You did a looot of work!

src/query/frontend/semantic/required_privileges.cpp
```diff
@@ -80,8 +80,19 @@ class PrivilegeExtractor : public QueryVisitor<void>, public HierarchicalTreeVis
...
+                      [&](const auto procedure_name) { return procedure_name == procedure.procedure_name_; })) {
+        AddPrivilege(privilege);
+      }
+    };
+    constexpr std::array<std::string_view, 2> module_read_procedures{"mg.get_module_file", "mg.get_module_files"};
+    add_privilege(module_read_procedures, AuthQuery::Privilege::MODULE_READ);
+
+    constexpr std::array<std::string_view, 3> module_write_procedures{"mg.create_module_file", "mg.update_module_file",
+                                                                      "mg.delete_module_file"};
+    add_privilege(module_write_procedures, AuthQuery::Privilege::MODULE_WRITE);
```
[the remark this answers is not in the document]
antonio2368: What do you mean? Store it with the `Module` object?
Different procedures from the same module could have different Privileges as is the case here.
Also, it made more sense to me because of the separation of concerns. I know it's a bit weird to reference the name of procedures here, but you know if you want to change the privileges where to check.
Tell me if you think something different would make more sense! :smile:

src/query/procedure/module.cpp
```diff
@@ -320,6 +322,350 @@ void RegisterMgTransformations(const std::map<std::string, std::unique_ptr<Modul
...
+
+    if (!std::filesystem::exists(path)) {
+      static_cast<void>(mgp_result_set_error_msg(result, "The specified file doesn't exist."));
+      return;
+    }
+
+    if (!IsAllowedExtension(path.extension())) {
+      static_cast<void>(mgp_result_set_error_msg(result, "The specified file isn't in the supported format."));
+      return;
+    }
```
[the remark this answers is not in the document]
antonio2368: Ah I see, that makes sense. My thoughts were somewhere along those lines, but it went what's the point of checking the extension of a file that doesn't exist, but I always forget that `path` object doesn't have to point to an existing file.

src/query/procedure/module.cpp
```diff
@@ -320,6 +322,350 @@ void RegisterMgTransformations(const std::map<std::string, std::unique_ptr<Modul
...
+    const auto path_value = GetStringValueOrSetError(std::filesystem::canonical(file_path).c_str(), memory, result);
+    if (!path_value) {
+      return;
+    }
+
+    if (!InsertResultOrSetError(result, record, "path", path_value.get())) {
+      return;
+    }
+
+    WithUpgradedLock(lock, [&]() { module_registry->UnloadAndLoadModulesFromDirectories(); });
```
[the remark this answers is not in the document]
antonio2368: If you updated a module file which is just a utility file without any procedures/transformations, you would like to reload all the modules that use it.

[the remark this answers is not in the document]
antonio2368: Remember that these functions can be used for any python file, not just the ones being loaded by memgraph directly.

src/query/procedure/module.cpp
antonio2368: These operations require the write lock on module registry (`Unload/LoadModules`), while the interpreter gets read lock on each procedure call so I would say it's safe.

src/query/procedure/module.cpp
```diff
@@ -320,6 +322,350 @@ void RegisterMgTransformations(const std::map<std::string, std::unique_ptr<Modul
...
+    std::error_code ec;
+    if (!std::filesystem::remove(path, ec)) {
+      static_cast<void>(
+          mgp_result_set_error_msg(result, fmt::format("Failed to delete the module: {}", ec.message()).c_str()));
+      return;
+    }
+
+    auto parent_path = path.parent_path();
+    while (std::filesystem::is_empty(parent_path) &&
+           !std::any_of(module_directories.begin(), module_directories.end(), [&](const auto &module_directory) {
```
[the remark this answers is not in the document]
antonio2368: I would find the module directory of which the deleted file is sub path, but agree with the idea.

src/query/procedure/module.cpp
```diff
@@ -320,6 +322,350 @@ void RegisterMgTransformations(const std::map<std::string, std::unique_ptr<Modul
...
+  return std::any_of(allowed_extensions.begin(), allowed_extensions.end(),
+                     [&](const auto allowed_extension) { return allowed_extension == extension; });
+}
+
+bool IsSubPath(const auto &base, const auto &destination) {
+  const std::string relative = std::filesystem::relative(destination, base);
+  // if it starts with '..', base is not subpath of destination
+  return relative.size() == 1 || relative[0] != '.' || relative[1] != '.';
+}
+
```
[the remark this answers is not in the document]
antonio2368: This is such an annoying problem that seems easy at first...
You are correct, I need to check if it's empty, some examples in the docs show how it can happen (even though I'm not sure how to trigger it on Linux - `/etc` relative to `/other` will return `../other`)

Your `IsSubPath2` looks okay, the current solution tries to shortcircuit as much as possible but there is no reason to do be so pedantic here.

tests/e2e/module_file_manager/module_file_manager.cpp
antonio2368: For this I need to update mgclient to support exceptions, so you'll see the changes tomorrow :smile:

src/query/frontend/semantic/required_privileges.cpp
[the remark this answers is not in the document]
antaljanosbenjamin: Yeah, maybe not the module, but the procedure could hold the required privileges. However I agree with separation of concerns, so that's is a bad approach.

My biggest concern is referencing the names as pure strings. Maybe we could expose those names as `constexpr std::string_view` (`get_module_file`, `delete_module_file`) and the builtin module name (`mg`) also. That way we can use those names and concatenate them everywhere in the codebase where we need them. Not the best, but I think it is less prone to bugs than pure strings.

src/query/procedure/module.cpp
```diff
@@ -347,6 +346,20 @@ std::optional<std::string> ReadFile(const auto &path) {
   file.close();
   return std::move(content);
 }
+
+// Return the module directory that contains the `path`
+utils::BasicResult<const char *, std::filesystem::path> ParentModuleDirectory(const ModuleRegistry &module_registry,
```
antaljanosbenjamin, line 351: Really nice! :ok_hand:

src/query/procedure/module.cpp
```diff
@@ -634,11 +656,9 @@ void RegisterMgDeleteModuleFile(ModuleRegistry *module_registry, utils::RWLock *
       return;
     }
 
-    const auto &module_directories = module_registry->GetModulesDirectory();
-    if (std::all_of(module_directories.begin(), module_directories.end(),
-                    [&](const auto &base_directory) { return !IsSubPath(base_directory, path); })) {
-      static_cast<void>(
-          mgp_result_set_error_msg(result, "The specified file isn't contained in one of the module directories."));
+    const auto parent_module_directory = ParentModuleDirectory(*module_registry, path);
```
antaljanosbenjamin, line 659: I mentioned the longest common prefix, because if somebody uses nested query module directories (both `a` and `a/b` is specified) then we might end up deleting an empty query module directory. Although I am not sure if that is a problem or not.

src/query/procedure/module.cpp
```diff
@@ -320,6 +322,368 @@ void RegisterMgTransformations(const std::map<std::string, std::unique_ptr<Modul
...
+
+    if (!std::filesystem::exists(path)) {
+      static_cast<void>(mgp_result_set_error_msg(result, "The specified file doesn't exist."));
+      return;
+    }
+
+    if (!IsAllowedExtension(path.extension())) {
+      static_cast<void>(mgp_result_set_error_msg(result, "The specified file isn't in the supported format."));
+      return;
+    }
```
antaljanosbenjamin, line 657: Yesterday I somehow slipped through this, but the order should be also changed here in my opinion.

antaljanosbenjamin (changes requested):

src/query/frontend/semantic/required_privileges.cpp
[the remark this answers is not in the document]
antonio2368: I see, I only have a small problem with where to put the constants. Do you have any suggestions?

src/query/frontend/semantic/required_privileges.cpp
```diff
@@ -80,8 +80,19 @@ class PrivilegeExtractor : public QueryVisitor<void>, public HierarchicalTreeVis
     AddPrivilege(AuthQuery::Privilege::CREATE);
     return false;
   }
-  bool PreVisit(CallProcedure & /*unused*/) override {
-    // TODO: Corresponding privilege
+  bool PreVisit(CallProcedure &procedure) override {
+    const auto add_privilege = [&](const auto &procedures, const auto privilege) {
+      if (std::any_of(procedures.begin(), procedures.end(),
+                      [&](const auto procedure_name) { return procedure_name == procedure.procedure_name_; })) {
```
kostasrim, line 86:
```suggestion
                      [&](const auto &procedure_name) { return procedure_name == procedure.procedure_name_; })) {
```

src/query/frontend/semantic/required_privileges.cpp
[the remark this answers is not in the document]
kostasrim: At first, I would say inside modules but since these are completely separate we could add an extra header for that. However, for this simple case, I would be personally fine to keep it as is unless we got something that is really better. Open for suggestions here as well :D

src/query/procedure/module.cpp
```diff
@@ -83,6 +84,19 @@ void BuiltinModule::AddTransformation(std::string_view name, mgp_trans trans) {
 
 namespace {
 
+auto WithUpgradedLock(auto *lock, const auto &function) {
+  lock->unlock_shared();
+  try {
+    function();
+    // There's no finally in C++, but we have to return our original READ lock
```
kostasrim, line 91: Why don't you use `OnScopeExit` that passes a lambda which calls `lock->lock_shared()`

src/query/procedure/module.cpp
```diff
@@ -320,6 +322,368 @@ void RegisterMgTransformations(const std::map<std::string, std::unique_ptr<Modul
   module->AddProcedure("transformations", std::move(procedures));
 }
 
+namespace {
+bool IsAllowedExtension(const auto &extension) {
+  constexpr std::array<std::string_view, 1> allowed_extensions{".py"};
+  return std::any_of(allowed_extensions.begin(), allowed_extensions.end(),
```
kostasrim, line 328: I was about to write `std::find` but I really like the idea of using `std::any_of` to return a bool. Really nice :fire:

src/query/procedure/module.cpp
```diff
@@ -320,6 +322,368 @@ void RegisterMgTransformations(const std::map<std::string, std::unique_ptr<Modul
...
+std::optional<std::string> ReadFile(const auto &path) {
+  std::ifstream file(path);
+  if (!file.is_open()) {
+    return std::nullopt;
+  }
+
+  const auto size = std::filesystem::file_size(path);
+  std::string content(size, '\0');
+  file.read(content.data(), static_cast<std::streamsize>(size));
+  file.close();
```
kostasrim, line 346: Does `ifstream` close the file on its destructor? What if there is an exception before the call to close?

src/query/frontend/semantic/required_privileges.cpp
kostasrim: Ignore me if this is a `string_view`

tests/e2e/module_file_manager/module_file_manager.cpp
```diff
@@ -0,0 +1,170 @@
...
+// License, and you may not use this file except in compliance with the Business Source License.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#include <filesystem>
+#include <fstream>
+#include <ranges>
```
kostasrim, line 14: :fire:

tests/e2e/module_file_manager/module_file_manager.cpp
```diff
@@ -0,0 +1,170 @@
...
+    result.emplace_back(row[0].ValueString());
+  }
+
+  return result;
+}
+
+bool ModuleFileExists(auto &client, const auto &path) {
+  const auto module_files = GetModuleFiles(client);
+
+  return std::find_if(module_files.begin(), module_files.end(),
```
kostasrim, line 54: `std::any_of` ? ;) (you don't have to change this, just poking a little bit)

tests/e2e/module_file_manager/module_file_manager.cpp
kostasrim: You include it but you don't use it! <3

src/query/procedure/module.cpp
antonio2368: that absolutely works, nice catch!

src/query/procedure/module.cpp
antonio2368: it seems it does call `close` on destruction :+1:

src/query/procedure/module.cpp
```diff
@@ -313,13 +316,388 @@ void RegisterMgTransformations(const std::map<std::string, std::unique_ptr<Modul
...
+      }
+    }
+  }
+
+  if (longest_parent_directory == module_directories.end()) {
+    return "The specified file isn't contained in any of the module directories.";
+  }
+
+  return *longest_parent_directory;
+}
```
antonio2368, line 373: @antaljanosbenjamin check it

src/query/procedure/module.hpp
```diff
@@ -121,8 +122,14 @@ class ModuleRegistry final {
...
   bool RegisterMgProcedure(std::string_view name, mgp_proc proc);
 
+  const std::filesystem::path &InternalModuleDir() const noexcept;
+
  private:
   std::vector<std::filesystem::path> modules_dirs_;
+
+  std::filesystem::path internal_module_dir_;
+
+  std::unordered_map<std::string, AuthQuery::Privilege> privileges_;
```
antonio2368, line 132: remove this

tests/e2e/module_file_manager/module_file_manager.cpp
antonio2368: I really tried but nothing warranted all the extra characters :smiling_face_with_tear:

src/query/procedure/module.cpp
```diff
@@ -313,13 +308,386 @@ void RegisterMgTransformations(const std::map<std::string, std::unique_ptr<Modul
...
+    MG_ASSERT(Call<size_t>(mgp_list_size, args) == 1U, "Should have been type checked already");
+    auto *arg = Call<mgp_value *>(mgp_list_at, args, 0);
+    MG_ASSERT(CallBool(mgp_value_is_string, arg), "Should have been type checked already");
+    const char *path_str{nullptr};
+    {
+      const auto success = TryOrSetError([&] { return mgp_value_get_string(arg, &path_str); }, result);
+      if (!success) {
+        return;
+      }
+    }
```
jbajic, line 429:
```suggestion
      if (const auto success = TryOrSetError([&] { return mgp_value_get_string(arg, &path_str); }, result); !success) {
        return;
      }
```

antonio2368: oh wow, How did I miss this and so many times, great catch!
I can just do `if (!TryOrSetError..)`

antaljanosbenjamin (approved):
Just a tiny error with the include order, everything else looks super nice!

jbajic (approved):
Amazing job! :tada:

PR #331, opened 2022-01-24
[master < T0800-MG] Add microseconds precision to temporal types now functions

kostasrim:
[master < Task] PR

Don't merge this unless I update(tick) the update doc above.


src/utils/temporal.cpp
```diff
@@ -79,36 +79,15 @@ Date::Date(const DateParameters &date_parameters) {
...
-}
+LocalTime UtcLocalTime() { return UtcLocalDateTime().local_time; }
 
 LocalDateTime UtcLocalDateTime() {
-  const auto utc_today = GetUtcFromSystemClockOrThrow();
-  return LocalDateTime({TMYearToUtcYear(utc_today.tm_year), TMMonthToUtcMonth(utc_today.tm_mon), utc_today.tm_mday},
-                       {utc_today.tm_hour, utc_today.tm_min, utc_today.tm_sec});
+  namespace chrono = std::chrono;
+  using TpMicros = chrono::time_point<chrono::system_clock, chrono::microseconds>;
+  TpMicros ts = chrono::time_point_cast<chrono::microseconds>(chrono::system_clock::now());
```
antaljanosbenjamin, line 89:
```suggestion
  auto ts = chrono::time_point_cast<chrono::microseconds>(chrono::system_clock::now());
```

antaljanosbenjamin (changes requested):
Is it possible to write/extend some of the tests? If that is hard (because it is very time dependent), then it doesn't worth the effort.

jbajic (approved):

src/utils/temporal.cpp
kostasrim: hahaha, :eyes: how did I miss that :D :D +100

kostasrim: I will check for the tests and update

antonio2368 (approved):

antonio2368:
Rebase this so we can merge it next

PR #334, opened 2022-01-24
[E108-MG < T0787-MG] Websocket tests

jbajic:
[master < Task] PR

### DIFF
--- a/src/communication/websocket/auth.hpp
+++ b/src/communication/websocket/auth.hpp
@@ -1,4 +1,4 @@
-// Copyright 2021 Memgraph Ltd.
+// Copyright 2022 Memgraph Ltd.
 //
 // Use of this software is governed by the Business Source License
 // included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
@@ -19,15 +19,24 @@
 
 namespace communication::websocket {
 
-class SafeAuth {
+class IAuthentication {
+ public:
+  virtual bool Authenticate(const std::string &username, const std::string &password) const = 0;
+
+  virtual bool HasUserPermission(const std::string &username, auth::Permission permission) const = 0;
+
+  virtual bool HasAnyUsers() const = 0;
+};
+
+class SafeAuth : public IAuthentication {
  public:
   explicit SafeAuth(utils::Synchronized<auth::Auth, utils::WritePrioritizedRWLock> *auth) : auth_{auth} {}
 
-  bool Authenticate(const std::string &username, const std::string &password) const;
+  bool Authenticate(const std::string &username, const std::string &password) const override;
 
-  bool HasUserPermission(const std::string &username, auth::Permission permission) const;
+  bool HasUserPermission(const std::string &username, auth::Permission permission) const override;
 
-  bool HasAnyUsers() const;
+  bool HasAnyUsers() const override;
 
  private:
   utils::Synchronized<auth::Auth, utils::WritePrioritizedRWLock> *auth_;
--- a/src/communication/websocket/auth.hpp
+++ b/src/communication/websocket/auth.hpp
@@ -19,7 +19,7 @@
 
 namespace communication::websocket {
 
-class IAuthentication {
+class AuthenticationInterface {
  public:
   virtual bool Authenticate(const std::string &username, const std::string &password) const = 0;
 
@@ -28,7 +28,7 @@ class IAuthentication {
   virtual bool HasAnyUsers() const = 0;
 };
 
-class SafeAuth : public IAuthentication {
+class SafeAuth : public AuthenticationInterface {
  public:
   explicit SafeAuth(utils::Synchronized<auth::Auth, utils::WritePrioritizedRWLock> *auth) : auth_{auth} {}
 
--- a/src/communication/websocket/listener.hpp
+++ b/src/communication/websocket/listener.hpp
@@ -42,7 +42,7 @@ class Listener : public std::enable_shared_from_this<Listener> {
   tcp::endpoint GetEndpoint() const;
 
  private:
-  Listener(boost::asio::io_context &ioc, tcp::endpoint endpoint, IAuthentication *auth);
+  Listener(boost::asio::io_context &ioc, tcp::endpoint endpoint, IAuthentication &auth);
 
   void DoAccept();
   void OnAccept(boost::beast::error_code ec, tcp::socket socket);
@@ -50,6 +50,6 @@ class Listener : public std::enable_shared_from_this<Listener> {
   boost::asio::io_context &ioc_;
   tcp::acceptor acceptor_;
   utils::Synchronized<std::list<std::shared_ptr<Session>>, utils::SpinLock> sessions_;
-  IAuthentication *auth_;
+  IAuthentication &auth_;
 };
 }  // namespace communication::websocket
--- a/src/communication/websocket/listener.hpp
+++ b/src/communication/websocket/listener.hpp
@@ -1,4 +1,4 @@
-// Copyright 2021 Memgraph Ltd.
+// Copyright 2022 Memgraph Ltd.
 //
 // Use of this software is governed by the Business Source License
 // included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
@@ -39,9 +39,10 @@ class Listener : public std::enable_shared_from_this<Listener> {
   // Start accepting incoming connections
   void Run();
   void WriteToAll(std::shared_ptr<std::string> message);
+  tcp::endpoint GetEndpoint() const;
 
  private:
-  Listener(boost::asio::io_context &ioc, tcp::endpoint endpoint, SafeAuth auth);
+  Listener(boost::asio::io_context &ioc, tcp::endpoint endpoint, IAuthentication *auth);
 
   void DoAccept();
   void OnAccept(boost::beast::error_code ec, tcp::socket socket);
@@ -49,6 +50,6 @@ class Listener : public std::enable_shared_from_this<Listener> {
   boost::asio::io_context &ioc_;
   tcp::acceptor acceptor_;
   utils::Synchronized<std::list<std::shared_ptr<Session>>, utils::SpinLock> sessions_;
-  SafeAuth auth_;
+  IAuthentication *auth_;
 };
 }  // namespace communication::websocket
--- a/src/communication/websocket/listener.hpp
+++ b/src/communication/websocket/listener.hpp
@@ -42,7 +42,7 @@ class Listener : public std::enable_shared_from_this<Listener> {
   tcp::endpoint GetEndpoint() const;
 
  private:
-  Listener(boost::asio::io_context &ioc, tcp::endpoint endpoint, IAuthentication &auth);
+  Listener(boost::asio::io_context &ioc, tcp::endpoint endpoint, AuthenticationInterface &auth);
 
   void DoAccept();
   void OnAccept(boost::beast::error_code ec, tcp::socket socket);
@@ -50,6 +50,6 @@ class Listener : public std::enable_shared_from_this<Listener> {
   boost::asio::io_context &ioc_;
   tcp::acceptor acceptor_;
   utils::Synchronized<std::list<std::shared_ptr<Session>>, utils::SpinLock> sessions_;
-  IAuthentication &auth_;
+  AuthenticationInterface &auth_;
 };
 }  // namespace communication::websocket
--- a/tests/unit/CMakeLists.txt
+++ b/tests/unit/CMakeLists.txt
@@ -369,9 +369,7 @@ add_test(test_lcp ${CMAKE_CURRENT_BINARY_DIR}/test_lcp)
 add_dependencies(memgraph__unit test_lcp)
 
 # Test websocket
-if (MG_ENTERPRISE)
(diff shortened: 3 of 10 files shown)


src/communication/websocket/auth.hpp
```diff
@@ -19,15 +19,24 @@
 
 namespace communication::websocket {
 
-class SafeAuth {
+class IAuthentication {
```
kostasrim, line 22: IAuthentication? What's the `I`? :D

kostasrim: Since this is a pure abstract class maybe you can name it as `AuthenticatioInterface`

src/communication/websocket/auth.hpp
```diff
@@ -19,15 +19,24 @@
...
+ public:
+  virtual bool Authenticate(const std::string &username, const std::string &password) const = 0;
+
+  virtual bool HasUserPermission(const std::string &username, auth::Permission permission) const = 0;
+
+  virtual bool HasAnyUsers() const = 0;
+};
+
+class SafeAuth : public IAuthentication {
```
kostasrim, line 31: Why not `CRTP` here? You don't really need runtime polymorphism here.

tests/unit/websocket.cpp
```diff
@@ -0,0 +1,262 @@
...
+  EXPECT_FALSE(websocket_server.IsRunning());
+
+  EXPECT_NO_THROW(websocket_server.AwaitShutdown());
+  EXPECT_FALSE(websocket_server.IsRunning());
+}
+
+TEST(WebSocketServer, WebsocketLoggingSink) {
+  MockAuth auth{};
+  communication::websocket::Server websocket_server({"0.0.0.0", 0}, &auth);
+  EXPECT_NO_THROW(websocket_server.GetLoggingSink());
```
antaljanosbenjamin, line 125: Maybe try to write something to the log and read on the other side.

tests/unit/websocket.cpp
```diff
@@ -0,0 +1,262 @@
...
+  }
+  EXPECT_NO_THROW(websocket_server.Shutdown());
+  EXPECT_NO_THROW(websocket_server.AwaitShutdown());
+}
+
+TEST(WebSocketServer, WebsocketAuthorizationFails) {
+  MockAuth auth{true, false, true};
+  communication::websocket::Server websocket_server({"0.0.0.0", 0}, &auth);
+  constexpr auto host = "0.0.0.0";
+  const auto port = std::to_string(websocket_server.GetEndpoint().port());
```
antaljanosbenjamin, line 247: I think this part and the destruction of the server could be extracted to a [test fixture](https://github.com/google/googletest/blob/main/docs/primer.md#test-fixtures-using-the-same-data-configuration-for-multiple-tests-same-data-multiple-tests). I think providing full access to all members of `MockAuth` doesn't harm anybody.

Using a test fixture might make it possible to put the test cases of `WebsocketAuthenticationWhenAuthPasses` into different tests.

antaljanosbenjamin (changes requested):
Please also consider adding some unit tests for `Session`. because the authentication and authorization part is important, therefore it is better to be tested with as narrow scope as it is possible.

src/communication/websocket/auth.hpp
jbajic: I stands for the interface, but I think your point is valid I am looking and haven't seen this naming convention anywhere in codebase

src/communication/websocket/auth.hpp
jbajic: I guess the only reason I would not prefer CRTP here is that I would have to have only header files for the listener and session. And I don't think that inheritance here will ever be a bottleneck since authentication is a straightforward process. What do you think about that?

tests/unit/CMakeLists.txt
```diff
@@ -368,3 +368,8 @@ add_custom_target(test_lcp ALL DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/test_lcp)
 add_test(test_lcp ${CMAKE_CURRENT_BINARY_DIR}/test_lcp)
 add_dependencies(memgraph__unit test_lcp)
 
+# Test websocket
+find_package(Boost REQUIRED)
+
+add_unit_test(websocket.cpp)
+target_link_libraries(${test_prefix}websocket mg-communication gtest Boost::headers Threads::Threads)
```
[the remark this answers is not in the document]
jbajic: Should we rely on transitive dependencies, isn't it better to link all the stuff that the source file uses?

tests/unit/websocket.cpp
```diff
@@ -0,0 +1,262 @@
...
+TEST(WebSocketServer, WebSockerServerCreation) {
+  /**
+   * Notice how there is not port management for the clients
+   * and servers, that is because when using "0.0.0.0" and address
+   * and 0 as port number we delegate port assignment to the os
+   * and he is the keeper of all available port numbers and
+   * assignees is automaticaly to us.
+   */
+  MockAuth auth{};
+  EXPECT_NO_THROW(communication::websocket::Server websocket_server({"0.0.0.0", 0}, &auth));
```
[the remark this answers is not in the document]
jbajic: That check is in the test case below since here only the instantiation is being tested

tests/unit/websocket.cpp
jbajic: I've actually discovered a bug here, thanks

tests/unit/websocket.cpp
```diff
@@ -0,0 +1,298 @@
...
+    spdlog::default_logger()->flush();
+  };
+  {
+    auto client = Client();
+    client.Connect(ServerAddress(), ServerPort());
+
+    std::thread(log_message, spdlog::level::err, "Sending error message!").detach();
+    const auto received_message1 = client.Read();
+    EXPECT_EQ(received_message1,
+              "{\"event\": \"log\", \"level\": \"error\", \"message\": \"Sending error message!\"}\n");
```
antonio2368, line 173: this could be extracted into a lambda (what if we change the format slightly)

tests/unit/websocket.cpp
```diff
@@ -0,0 +1,298 @@
...
+    EXPECT_NO_THROW(client2.Write(R"({"username": "user", "password": "123"})"));
+
+    const auto response1 = nlohmann::json::parse(client1.Read());
+    const auto message_header1 = response1[kResponseMessage].get<std::string>();
+    const auto message_first_part1 = message_header1.substr(0, message_header1.find(": "));
+    EXPECT_FALSE(response1[kResponseSuccess]);
+    EXPECT_EQ(message_first_part1, auth_fail) << "Message was: " << message_header1;
+
+    const auto response2 = client2.Read();
+    EXPECT_TRUE(response2 == auth_success) << "Response was: " << response2;
```
antonio2368, line 268:
```suggestion
    {
      const auto response = nlohmann::json::parse(client1.Read());
      const auto message_header = response[kResponseMessage].get<std::string>();
      const auto message_first_part = message_header.substr(0, message_header.find(": "));
      EXPECT_FALSE(response[kResponseSuccess]);
      EXPECT_EQ(message_first_part, auth_fail) << "Message was: " << message_header;
    }
    {
      const auto response = client2.Read();
      EXPECT_TRUE(response == auth_success) << "Response was: " << response;
    }
```
And you can also add `SCOPED_TRACE` for more information.

antonio2368 (changes requested):

src/communication/websocket/auth.hpp
kostasrim: personally, I think if something can be expressed at compile time it should be done at compile time. However, that's idealistic and it's not going to make you more productive nor will it help you on this PR. I agree it won't be a bottleneck so plz feel free to keep things as is

src/communication/websocket/listener.hpp
```diff
@@ -39,16 +39,17 @@ class Listener : public std::enable_shared_from_this<Listener> {
...
+  Listener(boost::asio::io_context &ioc, tcp::endpoint endpoint, IAuthentication *auth);
 
   void DoAccept();
   void OnAccept(boost::beast::error_code ec, tcp::socket socket);
 
   boost::asio::io_context &ioc_;
   tcp::acceptor acceptor_;
   utils::Synchronized<std::list<std::shared_ptr<Session>>, utils::SpinLock> sessions_;
-  SafeAuth auth_;
+  IAuthentication *auth_;
```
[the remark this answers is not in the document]
kostasrim: I really don't like having member references but I think it's fine for this case.

tests/unit/CMakeLists.txt
```diff
@@ -368,3 +368,10 @@ add_custom_target(test_lcp ALL DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/test_lcp)
 add_test(test_lcp ${CMAKE_CURRENT_BINARY_DIR}/test_lcp)
 add_dependencies(memgraph__unit test_lcp)
 
+# Test websocket
+if (MG_ENTERPRISE)
```
kostasrim, line 372: Why on `enterprise`?

tests/unit/websocket.cpp
```diff
@@ -0,0 +1,298 @@
...
+
+  bool HasUserPermission(const std::string & /*username*/, auth::Permission /*permission*/) const override {
+    return authorization_;
+  }
+
+  bool HasAnyUsers() const override { return has_any_users_; }
+
+  bool authentication_{true};
+  bool authorization_{true};
+  bool has_any_users_{true};
```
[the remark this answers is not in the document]
kostasrim: Or if you don't make them private, remove the getters :) (thinking of temporal types now -- my mistake there :D :D )

tests/unit/websocket.cpp
jbajic: I did it deliberately this way to have two sessions simultaneously connected

tests/unit/CMakeLists.txt
jbajic: Since on community the tests will not work for the authorization part, that part only works in enterprise

tests/unit/websocket.cpp
```diff
@@ -0,0 +1,305 @@
...
+  constexpr auto create_log_message = [](std::string_view event, std::string_view level, std::string_view message) {
+    return fmt::format("{{\"event\": \"{}\", \"level\": \"{}\", \"message\": \"{}\"}}\n", event, level, message);
+  };
+  {
+    auto client = Client();
+    client.Connect(ServerAddress(), ServerPort());
+
+    std::thread(log_message, spdlog::level::err, "Sending error message!").detach();
+    const auto received_message1 = client.Read();
+    EXPECT_EQ(received_message1, create_log_message("log", "error", "Sending error message!"));
```
antonio2368, line 176: I meant having these 3 lines in a lambda as they're almost identical.

antonio2368 (changes requested):

tests/unit/websocket.cpp
antonio2368: but the sessions will still be active as the clients are not part of the scope.
You just process them in a independent scopes.

jbajic: but if they are in a different scope the client will disconnect, since in the clines destructor we are disconnecting from the server

tests/unit/websocket.cpp
```diff
@@ -0,0 +1,262 @@
...
+};
+
+TEST(WebSocketServer, WebSockerServerCreation) {
+  /**
+   * Notice how there is not port management for the clients
+   * and servers, that is because when using "0.0.0.0" and address
+   * and 0 as port number we delegate port assignment to the os
+   * and he is the keeper of all available port numbers and
+   * assignees is automaticaly to us.
+   */
```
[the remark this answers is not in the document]
antaljanosbenjamin: There is still a typo: `assignees` -> `assigns`

tests/unit/CMakeLists.txt
jbajic: Will remove since a better solution was suggested, by just skipping the authorization test

antaljanosbenjamin (approved):
Just a few tiny comments. I think the result is really nice.

kostasrim (approved):
Well done :taco:

antaljanosbenjamin (approved):

tests/unit/websocket.cpp
```diff
@@ -0,0 +1,298 @@
...
+  {
+    auto client = Client();
+    client.Connect(ServerAddress(), ServerPort());
+
+    auto log_message = [](spdlog::level::level_enum level, std::string message) {
+      spdlog::log(level, message);
+      spdlog::default_logger()->flush();
+    };
+    auto log_and_check = [log_message, &client](spdlog::level::level_enum level, std::string message,
+                                                std::string log_level_received) {
```
antonio2368, line 168:
```suggestion
    auto log_and_check = [log_message, &client](spdlog::level::level_enum level, std::string_view message,
                                                std::string_view log_level_received) {
```

tests/unit/websocket.cpp
antonio2368: I think we didn't understand each other. I didn't mean to put the clients in a different scope, just the temp variables you used (`response`, `message_header`...) so you don't have to enumerate them. But the current implementation also works!

jbajic: Aha sorry I've overlook that when reading your suggestion

antonio2368 (approved):

PR #335, opened 2022-01-25
[master < fix-symbol-loading] Fix module symbol loading

antonio2368:
[master < Task] PR


src/query/procedure/module.cpp
```diff
@@ -832,10 +832,28 @@ PythonModule::~PythonModule() {
...
+    MG_ASSERT(setdl);
+    auto *arg = PyTuple_New(1);
+    MG_ASSERT(arg);
+    MG_ASSERT(PyTuple_SetItem(arg, 0, flag) == 0);
+    PyObject_CallObject(setdl, arg);
+    Py_DECREF(flag);
+    Py_DECREF(setdl);
+    Py_DECREF(arg);
+    return true;
+  });
```
jbajic, line 850: This cannot be moved in the constructor since the `generator.py` script runs memgraph binary during build time and it fails to run it, if this is set in the ctor

src/query/procedure/module.hpp
```diff
@@ -125,6 +126,21 @@ class ModuleRegistry final {
...
+    SharedLibraryHandle operator=(const SharedLibraryHandle &) = delete;
+    SharedLibraryHandle operator=(SharedLibraryHandle &&) = delete;
+
+    ~SharedLibraryHandle() { dlclose(handle_); }
+
+   private:
+    void *handle_;
+  };
+
+  SharedLibraryHandle libstd_handle{"libstdc++.so.6", RTLD_NOW | RTLD_LOCAL | RTLD_DEEPBIND};
```
gitbuda, line 143: Could you put an extensive comment on why `.so.6` is required here?

src/query/procedure/module.cpp
```diff
@@ -715,7 +715,7 @@ bool SharedLibraryModule::Load(const std::filesystem::path &file_path) {
   spdlog::info("Loading module {}...", file_path);
   file_path_ = file_path;
   dlerror();  // Clear any existing error.
-  handle_ = dlopen(file_path.c_str(), RTLD_NOW | RTLD_LOCAL);
+  handle_ = dlopen(file_path.c_str(), RTLD_NOW | RTLD_LOCAL | RTLD_DEEPBIND);
```
gitbuda, line 718: Can you use `SharedLibraryHandle` also here?

src/query/procedure/module.hpp
jbajic: Yes definitely

src/query/procedure/module.cpp
jbajic: Hmm I wouldn't since the library that's been explicitly loaded and also explicitly closed in `Close` method, and that's been handled by `SharedLibraryModule`

src/query/procedure/module.cpp
```diff
@@ -832,10 +833,30 @@ PythonModule::~PythonModule() {
...
+    MG_ASSERT(setdl);
+    auto *arg = PyTuple_New(1);
+    MG_ASSERT(arg);
+    MG_ASSERT(PyTuple_SetItem(arg, 0, flag) == 0);
+    PyObject_CallObject(setdl, arg);
+    Py_DECREF(flag);
+    Py_DECREF(setdl);
+    Py_DECREF(arg);
+    is_python_initialized = true;
+  }
```
antaljanosbenjamin, line 853: We have a Python initialization related code in [memgraph.cpp](https://github.com/memgraph/memgraph/blob/bf01c58ed9950a1b2d05a884aad9544e8193b7f3/src/memgraph.cpp#L1050-L1059): https://github.com/memgraph/memgraph/blob/bf01c58ed9950a1b2d05a884aad9544e8193b7f3/src/memgraph.cpp#L1050-L1059

I think this and the existing python initialization should be in one place. Maybe you can move this to `main`?

src/query/procedure/module.hpp
```diff
@@ -125,6 +126,44 @@ class ModuleRegistry final {
...
+      if (handle_) {
+        dlclose(handle_);
+      }
+    }
+
+   private:
+    void *handle_;
+  };
+
+#if defined(__has_feature)
```
antaljanosbenjamin, line 147: I think as we are fully in control over which compiler do we use (we only support a specific version of clang) I think we are safe to assume `__has_feature` is supported by the compiler. This way we can spare one of the `#else` branches.

kostasrim (approved):

src/query/procedure/module.hpp
gitbuda: Great!

src/query/procedure/module.cpp
gitbuda: Makes sense 😄

gitbuda (approved):
More cases are covered, looks legit, let's merge 😄

PR #339, opened 2022-01-31
[master < T0806-MG] Fix nested transaction error

antaljanosbenjamin:
[master < Task] PR

**[Docs PR](https://github.com/memgraph/docs/pull/295)**


antaljanosbenjamin:
I will update the docs soon.

src/query/stream/streams.cpp
```diff
@@ -680,8 +680,7 @@ TransformationResult Streams::Check(const std::string &stream_name, std::optiona
           CallCustomTransformation(transformation_name, messages, result, accessor, *memory_resource, stream_name);
 
           for (auto &row : result.rows) {
-            auto [query, parameters] =
-                ExtractTransformationResult(std::move(row.values), transformation_name, stream_name);
+            auto [query, parameters] = ExtractTransformationResult(row.values, transformation_name, stream_name);
```
jbajic, line 683: How does this solve the nested transaction error?

antaljanosbenjamin: This doesn't solve it. The solution is on line 486.

This is just a bugfix that was hidden by the original bug: if the data is moved out here, then in the retry theere is no data, so we cannot retry the transaction. That's why I changed this to not move.

jbajic: aha, I see now it so if the transaction that has a serialization error it aborts. So why do we have both abort every time and throw after a certain amount of retries, should it not abort when it has tried several times? Does this make the counter redundant?

kostasrim: @jbajic That's a good question. If you call `abort()` it means that the whole transaction gets aborted and a new one must start by calling `begin()`. This is satisfied on line 462 (if you see this is wrapped in a while(true) loop. If there is a `serialization` error the whole transaction will be replayed - see the inner for-loop). Therefore the counter is for the total amount of retries the whole transaction must replayed either completing successfully or throwing

kostasrim (commented):
Shiny and clean. Thank you for taking care of this -- I introduced this small bug ;)

kostasrim (approved):

jbajic (approved):
Amazing work! :tada:

antonio2368 (approved):

PR #340, opened 2022-01-31
Build for ARM64

antonio2368:


src/storage/v2/property_store.hpp
```diff
@@ -19,6 +19,7 @@
 namespace storage {
 
 class PropertyStore {
+  static_assert(std::endian::native == std::endian::little, "PropertyStore supports only architectures using little-endian.");
```
gitbuda, line 22: Nice!!

environment/toolchain/folly.diff
```diff
@@ -0,0 +1,2 @@
+31d30
```
antaljanosbenjamin, line 1: Maybe consider switching to patch files as I did for the updated v4 toolchain.

src/utils/tsc.hpp
```diff
@@ -11,124 +11,34 @@
...
-#include <sys/prctl.h>
-#include <x86intrin.h>
-
 #include <chrono>
+#include <functional>
 #include <optional>
 #include <thread>
 
+#include "utils/logging.hpp"
 #include "utils/timer.hpp"
```
antaljanosbenjamin, line 20: I think we don't need these, also `functional`,`thread` and `chrono` is not necessary.

antaljanosbenjamin (changes requested):
Nice!

environment/toolchain/folly.diff
antonio2368: For which projects? Do you mean git patch files?
As we get the archives, we don't get the git folder so `git apply` doesn't work.

release/arm64/build_env.dockerfile
```diff
@@ -0,0 +1,18 @@
...
+RUN yum update && yum install -y curl git
+
+RUN /${env_folder}/os/centos-9.sh install MEMGRAPH_BUILD_DEPS
+RUN /${env_folder}/os/centos-9.sh install TOOLCHAIN_RUN_DEPS
+
+RUN rm -rf /env_folder
+
+RUN yum clean all
+
+RUN curl https://s3.eu-west-1.amazonaws.com/deps.memgraph.io/toolchain-v4/toolchain-v4-binaries-centos-9-arm64.tar.gz -o /tmp/toolchain.tar.gz \
```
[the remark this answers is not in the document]
antonio2368: Can you give me an example of what it would look like?
Would you then manually send the version on `docker build` or have a default value that will be used?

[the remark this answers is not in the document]
jbajic: toolchainc-v4Something similar exists here https://github.com/memgraph/memgraph/blob/master/release/package/ubuntu-20.04/Dockerfile, I would not provide a default, since default values will change, just pass build-arg in bash script as you do for `env_folder`. In this case that would be `toolchain-v4`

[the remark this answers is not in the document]
antonio2368: sure, makes sense

jbajic (approved):

environment/toolchain/folly.diff
antaljanosbenjamin: I meant something like [this](https://github.com/memgraph/memgraph/blob/T0804-MG-base-thrift-project/environment/toolchain/folly.patch), but this is not that important. I just copied the original source folder, made the changes and used `diff -ur`. If you name the original/modified folder as `a` and `b` then you get nice paths.

My reasoning to switch to patch files (not git patch, simple patch) is it can contain changes to multiple files, and also contain the path for the files to be modified.

antonio2368: oh okay, I would leave it as is for this patch as it's only one file but I'm definitely for it when there are multiple of them.

antaljanosbenjamin (approved):

kostasrim (approved):

gitbuda (approved):

PR #341, opened 2022-01-31
[master < ] Memgraph Docker Init Script

jmatak:
Simple working example of `init-script`. To use this kind of format, the user would need to bind `/usr/lib/memgraph/init` volume to the *.cypherl files that would be executed at the start of the Memgraph execution.

Work in progress: There should be automatic loading once mgconsole is connected, not custom sleep time.

- [ ] Add multiple cypher files for the init script
- [ ] When flags parsing, insert the possibility of having multiple directories

[master < Task] PR


Josipmrden (approved):
Good job!

gitbuda:
Taka a look here https://github.com/memgraph/memgraph/blob/046cd800542305f3a1f43d9e8c8dba4bd7428821/tests/drivers/run.sh for `wait_for_server` implementation + the correct port has to be deduced from both runtime flags and gflags (maybe we can ignore gflags for now).

gitbuda (changes requested):
The wait loop has to be improved 😄

gitbuda:
Yea, I'm just looking at that. The whole `perl` part is 🤯Having the arg parsing logic (or part of it) in 2 places doesn't feel right (we'll forget to update). An option is to isolate the parsing part in standalone binary, but more and more, I think that's an even bigger hack to the code now. But moving the whole part inside Memgraph seems better and better because you have more control, everything can be done in perfect time, and it will work everywhere, so that's what I would do 😄 (current thinking)

jmatak:
> Yea, I'm just looking at that. The whole `perl` part is exploding_headHaving the arg parsing logic (or part of it) in 2 places doesn't feel right (we'll forget to update). An option is to isolate the parsing part in standalone binary, but more and more, I think that's an even bigger hack to the code now. But moving the whole part inside Memgraph seems better and better because you have more control, everything can be done in perfect time, and it will work everywhere, so that's what I would do smile (current thinking)

I agree with you 100%. Going further and covering each edge case would end up in a disaster.

release/docker/init_script.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+
+# Read the user given arguments
+ARGS=$@
+# Read the configuration @ /etc/memgraph/memgraph.conf
+FLAGS="$(perl -ne 'print "$1=$2"  while /^(-{2}\b[a-z][a-z\d]{2,}\b(?:-[a-z\d]+)*)=((?:[a-zA-Z][a-zA-Z0-9_]*$)|(?:(?:(?:[\/]*\/)*)(?:.*))|(?:[0-9]*))/g' /etc/memgraph/memgraph.conf | sed '/^\s*$/d')"
```
antonio2368, line 6: what

jmatak: The most useless two hours spent in Memgraph. The goal was extracting a port from the gflags  but I have extracted all gflags arguments from config. Reason being checking if Memgraph is active during init script. Ended up with this disaster, and yeah, now I have no idea why I needed ALL arguments, at the point of making it was pretty clear for me...

aabrodskiy:
If I may suggest couple of additions, starting with description of use cases we have.
In my mind, there are two requirements similar to this issue: 
1. Very first initialization of the database when no database is found at all
2. Continuous migration/ upgrade of the database, when docker image upgraded - check and compare version / checksum on every startup.

These two tasks could be written very similar, but also could be very different, with the second one being a conditional script able to handle version jump, not only direct upgrade to the next version.
Many other databases/graphs/systems handle both scenarios with /init.d being shell scripts rather than language (cypher, sql, etc) specific scripts.

For example, postgresql container allows placing any arbitrary shell script in the init folder, which will be executed upon initialization.
In there, one can use 
`psql -d database -u username -p password -f filename.sql` to execute any SQL script and initialize the database.
This has proven to be quite flexible and would be great to see something along the same lines embedded into memgraph docker containers.

Josipmrden:
> If I may suggest couple of additions, starting with description of use cases we have. In my mind, there are two requirements similar to this issue:
> 
> 1. Very first initialization of the database when no database is found at all
> 2. Continuous migration/ upgrade of the database, when docker image upgraded - check and compare version / checksum on every startup.
> 
> These two tasks could be written very similar, but also could be very different, with the second one being a conditional script able to handle version jump, not only direct upgrade to the next version. Many other databases/graphs/systems handle both scenarios with /init.d being shell scripts rather than language (cypher, sql, etc) specific scripts.
> 
> For example, postgresql container allows placing any arbitrary shell script in the init folder, which will be executed upon initialization. In there, one can use `psql -d database -u username -p password -f filename.sql` to execute any SQL script and initialize the database. This has proven to be quite flexible and would be great to see something along the same lines embedded into memgraph docker containers.

@aabrodskiy  Memgraph is currently schemaless and version numbers are not in our core part of the database. However, we have added the integration of executing arbitrary scripts and initialization files with the latest commits :)

aabrodskiy:
> @aabrodskiy Memgraph is currently schemaless and version numbers are not in our core part of the database. However, we have added the integration of executing arbitrary scripts and initialization files with the latest commits :)

Awesome, I'll give it a try! When I mentioned schema, I meant a set of indexes and constraints that we apply to pre-defined labels and relantionships as well as some pre-created objects serving as a pre-defined starting point in our graph implementation.

Josipmrden:
> > @aabrodskiy Memgraph is currently schemaless and version numbers are not in our core part of the database. However, we have added the integration of executing arbitrary scripts and initialization files with the latest commits :)
> 
> Awesome, I'll give it a try! When I mentioned schema, I meant a set of indexes and constraints that we apply to pre-defined labels and relantionships as well as some pre-created objects serving as a pre-defined starting point in our graph implementation.

Yes, I think I understand what you're trying to say. However I'd just like to ask you a few more questions, just to solve your problem without any overhead or workarounds. Currently, in Memgraph all indexes/constraints and other configurations are done via queries. They are in same hierarchy just as is inserting nodes, relationships etc. It is advised to insert indexes, constraints at the beginning of your project since then everything is verified on the fly. That is better than updating all that if we have a present dataset in place. 
All of the queries in the init script will be executed if you bind the *.cypherl files as your docker volume. Do you have multiple cypherl files, e.g. one for creating indexes, constraints and other one for inserting a dataset? Are you having any trouble with priorities on executing the same?

gitbuda:
Closing this is favor of #974

Issue #231, opened 2021-09-10
How to extend Memgraph with a custom function?


gitbuda:
Query modules (`CALL procedure()`) are great to extend Memgraph capabilities but is there a way to inject an arbitrary function that could be used in an expression (`MATCH (n) WHERE my_fun(n) = True RETURN n;`).

gitbuda:
Done in #345

PR #345, opened 2022-02-03
[master < T168-MAGE] Memgraph magic functions

jmatak:
**Note:** When pulling the latest openCypher grammar from https://opencypher.org/resources/, there might be a conflict with the current functionName resolving. Have it in mind when updating the whole grammar.

- [x] Add arguments type checking
- [x] Implement an error checking

[master < Epic] PR
- [ ] Write E2E tests - [memgraph/docs PR](https://github.com/memgraph/docs/pull/346)


include/mg_procedure.h
```diff
@@ -1290,6 +1290,9 @@ struct mgp_module;
 /// Describes a procedure of a query module.
 struct mgp_proc;
 
+/// Describes a Memgraph magic function.
+struct mgp_func;
+
```
jmatak, line 1295: Since there are no write capabilities in functions, and vertex/edge has an instance on "non-writable" mgp_graph, do I actually need this in the API? Pretty sure I don't :sweat_smile: 

@antaljanosbenjamin

include/mg_procedure.h
```diff
@@ -1502,6 +1505,71 @@ typedef void (*mgp_trans_cb)(struct mgp_messages *, struct mgp_graph *, struct m
...
+/// Return MGP_ERROR_INVALID_ARGUMENT if `name` is not a valid argument name.
+/// RETURN MGP_ERROR_LOGIC_ERROR if the procedure already has any optional argument.
+enum mgp_error mgp_func_add_arg(struct mgp_func *func, const char *name, struct mgp_type *type);
+
+/// Entry-point for a custom Memgraph awesome function.+.
+///
+/// Passed in arguments will not live longer than the callback's execution.
+/// Therefore, you must not store them globally or use the passed in mgp_memory
+/// to allocate global resources.
+typedef struct mgp_value *(*mgp_func_cb)(struct mgp_list *, struct mgp_func_context *, struct mgp_memory *);
```
jmatak, line 1557: Since there are no write capabilities in functions, and vertex/edge has an instance on "non-writable" mgp_graph, do I actually need this (`mgp_func_context`) in the API? Pretty sure I don't sweat_smile

@antaljanosbenjamin

src/query/procedure/module.cpp
```diff
@@ -300,6 +304,89 @@ void RegisterMgTransformations(const std::map<std::string, std::unique_ptr<Modul
...
+
+        if (!InsertResultOrSetError(result, record, "path", path_value.get())) {
+          return;
+        }
+
+        if (!InsertResultOrSetError(result, record, "is_editable", is_editable_value.get())) {
+          return;
+        }
+      }
+    }
```
antaljanosbenjamin, line 381: I think this could be unified with the procedures ones with some template magic:
```

template <typename T>
const std::map<std::string, T, std::less<>> * GetCallables(Module& module) {
   if constexpr (std::is_same_v<TCall, mgp_proc>) {
      return module.Procedures();
    } else if constexpr (std::is_same_v<TCall, mgp_func>) {
      return module.Functions();
    } 
}
```

If you need help with that, let me know.

src/query/procedure/module.hpp
```diff
@@ -147,4 +149,9 @@ std::optional<std::pair<procedure::ModulePtr, const mgp_proc *>> FindProcedure(
 std::optional<std::pair<procedure::ModulePtr, const mgp_trans *>> FindTransformation(
     const ModuleRegistry &module_registry, const std::string_view fully_qualified_transformation_name,
     utils::MemoryResource *memory);
+
+/// TODO: Write documentation
```
antaljanosbenjamin, line 153: TODO

tests/e2e/magic_functions/common.py
```diff
@@ -0,0 +1,35 @@
+# Copyright 2022 Memgraph Ltd.
```
antaljanosbenjamin, line 1: I don't think that is necessary for now, just as an idea for the future: as we are started to copy these functions, what about making a strictly internal library for testing utilities? That would be a nice way of unifying our e2e tests and the python code that they use.

@jbajic, as you are our relative python expert, how hard would this be?

tests/e2e/magic_functions/function_example.py
```diff
@@ -0,0 +1,85 @@
...
+    result_null = result[0][0]
+    assert result_null is None
+
+
+@pytest.mark.parametrize("function_type", ["py", "c"])
+def test_too_many_arguments(connection, function_type):
+    cursor = connection.cursor()
+    assert has_n_result_row(cursor, "MATCH (n) RETURN n", 0)
+    # Should raise too many arguments
+    with pytest.raises(Exception) as e_info:
```
antaljanosbenjamin, line 64: Could you please restrict the exception type? Probably it is `mgclient.DatabaseError`. Also either add some checks on `e_info` or do not name it, because currently it is unused.

tests/e2e/magic_functions/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+# Set up C++ functions for e2e tests
+add_library(mg_procedure INTERFACE)
+target_include_directories(mg_procedure INTERFACE ${CMAKE_SOURCE_DIR}/include)
+
+function(add_query_module target_name src)
+  add_library(${target_name} SHARED ${src})
+  SET_TARGET_PROPERTIES(${target_name} PROPERTIES PREFIX "")
+  target_link_libraries(${target_name} PRIVATE mg_procedure)
+endfunction()
```
antaljanosbenjamin, line 9: As `add_library` would introduce a new global target, I would either remove that and put the `target_include_directories` in `add_query_module` or I would move this extra lib to the root `CMakeLists.txt` and name it `mg-procedure` (dash instead of underscore).

tests/e2e/magic_functions/functions/c_read.cpp
```diff
@@ -0,0 +1,134 @@
...
+  if (err_code != MGP_ERROR_NO_ERROR) {
+    mgp_func_result_error("Failed to construct return value!", memory, &result);
+    return result;
+  }
+  return result;
+}
+
+double get_element_from_arg(struct mgp_list *args, int index) {
+  mgp_value *value{};
+  mgp_list_at(args, index, &value);
```
antaljanosbenjamin, line 35: I think it would be nice to check the return type here as things can go really wrong if something doesn't match the expected structure/type.

tests/e2e/magic_functions/functions/c_read.cpp
```diff
@@ -0,0 +1,134 @@
...
+  return result;
+}
+
+static mgp_func_result *add_two_numbers(struct mgp_list *args, mgp_func_context *ctx, struct mgp_memory *memory) {
+  mgp_func_result *result{};
+
+  auto first = get_element_from_arg(args, 0);
+  auto second = get_element_from_arg(args, 1);
+
+  mgp_value *value{};
```
antaljanosbenjamin, line 57:
```suggestion
  mgp_value *value{nullptr};
```

I think it is not initialized to `nullptr` by default, but I might be wrong.

src/query/interpret/awesome_memgraph_functions.cpp
```diff
@@ -1259,6 +1263,91 @@ std::function<TypedValue(const TypedValue *, int64_t, const FunctionContext &ctx
...
+        elems.emplace_back(std::get<2>(func_opt_args[i]), &graph);
+      }
+
+      auto argslist = mgp_list(std::move(elems), ctx.memory);
+
+      auto *maybe_res = func_cb(&argslist, &functx, &memory);
+      if (maybe_res->error_msg) {
+        auto error_msg = *(maybe_res->error_msg);
+        throw QueryRuntimeException(error_msg);
+      }
```
antaljanosbenjamin, line 1342: Setting the error message could fail, so there is a chance that the user cannot indicate there is something went wrong. I think in that case the default behavior should be making sure the function call fails. E.g. the user want to set the error, but cannot, and then we try to read the `mgp_value` from the result which might be uninitialized or `nullptr`. I don't know how to solve it yet, but if you need help, feel free to ask, I am happy to help to achieve this behavior.

tests/e2e/magic_functions/functions/py_write.py
```diff
@@ -0,0 +1,19 @@
...
+# by the Apache License, Version 2.0, included in the file
+# licenses/APL.txt.
+
+import mgp
+
+
+@mgp.function
+def try_to_write(
+    ctx: mgp.FuncCtx, argument: mgp.Vertex, name: str, value: mgp.Nullable[mgp.Any]
+) -> mgp.Record():
```
antaljanosbenjamin, line 18: I think it shouldn't return an `mgp.Record`.

tests/unit/query_function_mgp_module.cpp
```diff
@@ -0,0 +1,52 @@
...
+  mgp_func *func{nullptr};
+  EXPECT_EQ(module.functions.find("same_name"), module.functions.end());
+  EXPECT_EQ(mgp_module_add_function(&module, "same_name", DummyCallback, &func), MGP_ERROR_NO_ERROR);
+  EXPECT_NE(module.functions.find("same_name"), module.functions.end());
+  EXPECT_EQ(mgp_module_add_function(&module, "same_name", DummyCallback, &func), MGP_ERROR_LOGIC_ERROR);
+  EXPECT_EQ(mgp_module_add_function(&module, "same_name", DummyCallback, &func), MGP_ERROR_LOGIC_ERROR);
+  EXPECT_NE(module.functions.find("same_name"), module.functions.end());
+}
+
+TEST(Module, CaseSensitiveFunctionNames) {
```
antaljanosbenjamin, line 45: I think the case insensitivity wouldn't work in e2e tests. Could you please create an e2e test for that just to have a definite answer for that?

antaljanosbenjamin (changes requested):

include/mg_procedure.h
```diff
@@ -1290,6 +1292,9 @@ struct mgp_module;
 /// Describes a procedure of a query module.
 struct mgp_proc;
 
+/// Describes a Memgraph magic function.
```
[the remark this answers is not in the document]
jmatak: I guess that is settled right now :smile:, since nobody had nothing against that name

include/mg_procedure.h
```diff
@@ -1502,6 +1507,76 @@ typedef void (*mgp_trans_cb)(struct mgp_messages *, struct mgp_graph *, struct m
...
+/// Therefore, you must not store them globally or use the passed in mgp_memory
+/// to allocate global resources.
+typedef struct mgp_func_result *(*mgp_func_cb)(struct mgp_list *, struct mgp_func_context *, struct mgp_memory *);
+
+/// Register a Memgraph magic function
+///
+/// The `name` must be a sequence of digits, underscores, lowercase and
+/// uppercase Latin letters. The name must begin with a non-digit character.
+/// Note that Unicode characters are not allowed. Additionally, names are
+/// case-sensitive.
```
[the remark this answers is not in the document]
jmatak: Well, I think it should be case-insensitive since we are converting them to upper case. However, modules could have a different case?

src/query/procedure/mg_procedure_impl.hpp
```diff
@@ -545,14 +545,38 @@ struct mgp_result {
   std::optional<utils::pmr::string> error_msg;
 };
 
+struct mgp_func_result {
+  explicit mgp_func_result(mgp_value *value, std::optional<utils::pmr::string> error_msg)
+      : value(value), error_msg(error_msg) {}
+
+  explicit mgp_func_result() {}
+  /// Return Magic function result with potential error
+  mgp_value *value;
```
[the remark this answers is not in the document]
jmatak: This is where I am stuck currently. I've chosen `*value` here because in the lambda which is executing the external function I am defining `mgp_function_result` on the stack, and this value gets deleted when returned. After returning the value, this fails: `MG_ASSERT(res.GetMemoryResource() == ctx_->memory);`

include/mg_procedure.h
```diff
@@ -1502,6 +1507,76 @@ typedef void (*mgp_trans_cb)(struct mgp_messages *, struct mgp_graph *, struct m
...
+/// Return MGP_ERROR_INVALID_ARGUMENT if `name` is not a valid argument name.
+/// RETURN MGP_ERROR_LOGIC_ERROR if the procedure already has any optional argument.
+enum mgp_error mgp_func_add_arg(struct mgp_func *func, const char *name, struct mgp_type *type);
+
+/// Entry-point for a custom Memgraph awesome function.+.
+///
+/// Passed in arguments will not live longer than the callback's execution.
+/// Therefore, you must not store them globally or use the passed in mgp_memory
+/// to allocate global resources.
+typedef struct mgp_func_result *(*mgp_func_cb)(struct mgp_list *, struct mgp_func_context *, struct mgp_memory *);
```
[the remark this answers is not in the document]
jmatak: I agree with you 100%, therefore I've changed this

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -2376,31 +2391,54 @@ mgp_error mgp_module_add_write_procedure(mgp_module *module, const char *name, m
   return WrapExceptions([=] { return mgp_module_add_procedure(module, name, cb, {.is_write = true}); }, result);
 }
 
-mgp_error mgp_proc_add_arg(mgp_proc *proc, const char *name, mgp_type *type) {
+namespace {
+template <typename TCall>
+mgp_error mgp_add_arg(TCall *callable, const char *name, mgp_type *type) {
+  static_assert(std::is_same_v<TCall, mgp_proc> || std::is_same_v<TCall, mgp_func>,
+                "Argument can be added only to mgp_proc or mgp_func");
```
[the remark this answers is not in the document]
jmatak: Was looking for that but couldn't find :joy:

include/mgp.py
```diff
@@ -40,27 +40,31 @@ class InvalidContextError(Exception):
     """
     Signals using a graph element instance outside of the registered procedure.
     """
+
```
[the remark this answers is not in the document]
jmatak: Yes, I think VSCode found a different config :+1:

src/query/procedure/mg_procedure_impl.hpp
```diff
@@ -545,14 +545,38 @@ struct mgp_result {
...
+  }
+
+  static mgp_graph NonWritableGraph(query::DbAccessor &acc, storage::View view) {
+    return mgp_graph{&acc, view, nullptr};
+  }
 };
 
+struct mgp_func_context {
+  query::DbAccessor *impl;
+  storage::View view;
```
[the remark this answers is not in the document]
jmatak: I think we agreed on preventing the user from using the `mgp_graph` at all. Also I don't think that it would be clean to have a `mgp_graph`  wrapper for functions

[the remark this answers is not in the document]
jmatak: This can be solved by unified `mgp_context` structure.

src/query/procedure/py_module.cpp
```diff
@@ -1031,6 +1101,87 @@ void CallPythonTransformation(const py::Object &py_cb, mgp_messages *msgs, mgp_g
...
+  auto error_to_msg = [](const std::optional<py::ExceptionInfo> &exc_info) -> std::optional<std::string> {
+    if (!exc_info) return std::nullopt;
+    // Here we tell the traceback formatter to skip the first line of the
+    // traceback because that line will always be our wrapper function in our
+    // internal `mgp.py` file. With that line skipped, the user will always
+    // get only the relevant traceback that happened in his Python code.
+    return py::FormatException(*exc_info, /* skip_first_line = */ true);
+  };
+
+  auto call = [&](py::Object py_graph) -> std::pair<mgp_value *, std::optional<py::ExceptionInfo>> {
```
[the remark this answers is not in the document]
jmatak: This was awesome, I changed it a bit, but we can discuss further

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -186,12 +186,15 @@ template <typename TFunc, typename... Args>
...
-bool MgpGraphIsMutable(const mgp_graph &graph) noexcept { return graph.view == storage::View::NEW; }
+// Graph mutations
+bool MgpGraphIsMutable(const mgp_graph &graph) noexcept {
+  return graph.view == storage::View::NEW && graph.ctx != nullptr;
+}
 
 bool MgpVertexIsMutable(const mgp_vertex &vertex) { return MgpGraphIsMutable(*vertex.graph); }
 
 bool MgpEdgeIsMutable(const mgp_edge &edge) { return MgpVertexIsMutable(edge.from); }
+
```
kostasrim, line 197:
```suggestion
```

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -2539,6 +2593,19 @@ void PrintProcSignature(const mgp_proc &proc, std::ostream *stream) {
...
+    stream << arg.first << " :: " << arg.second->GetPresentableName();
+  });
+  if (!proc.args.empty() && !proc.opt_args.empty()) (*stream) << ", ";
+  utils::PrintIterable(*stream, proc.opt_args, ", ", [](auto &stream, const auto &arg) {
+    stream << std::get<0>(arg) << " = ";
+    PrintValue(std::get<2>(arg), &stream) << " :: " << std::get<1>(arg)->GetPresentableName();
+  });
+  (*stream) << ")";
+}
+
```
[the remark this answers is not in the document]
kostasrim: `std::ostream *stream` -> `std::ostream &` 

Plz replace the pointer!

src/query/procedure/module.cpp
```diff
@@ -300,6 +304,89 @@ void RegisterMgTransformations(const std::map<std::string, std::unique_ptr<Modul
...
+      static_assert(std::is_same_v<decltype(module->Functions()), const std::map<std::string, mgp_func, std::less<>> *>,
+                    "Expected module procedures to be sorted by name");
+
+      const auto path = module->Path();
+      const auto path_string = GetPathString(path);
+      const auto is_editable = IsFileEditable(path);
+
+      for (const auto &[proc_name, proc] : *module->Functions()) {
+        mgp_result_record *record{nullptr};
+        {
```
kostasrim, line 327: I would say remove the `{}` and make `success` non const`. Based on the context, it will be easier to understand and read without the explosion of curly braces. Trust the curly guy ;) :D @antaljanosbenjamin what's your opinion on this?

src/query/procedure/module.hpp
```diff
@@ -147,4 +150,58 @@ std::optional<std::pair<procedure::ModulePtr, const mgp_proc *>> FindProcedure(
...
+template <typename T>
+concept IsCallable = utils::SameAsAnyOf<T, mgp_proc, mgp_func>;
+
+template <IsCallable TCall>
+void ConstructArguments(const std::vector<TypedValue> &args, const TCall callable,
+                        const std::string_view &fully_qualified_name, std::size_t n_args, mgp_list &args_list,
+                        const mgp_graph &graph) {
+  args_list.elems.reserve(n_args);
+  if (n_args < callable.args.size() ||
+      // Rely on `||` short circuit so we can avoid potential overflow of
```
kostasrim, line 168: This code is repeated is `awesome_memgraph_functions`. Make it a function!

src/query/procedure/py_module.cpp
```diff
@@ -1031,6 +1089,92 @@ void CallPythonTransformation(const py::Object &py_cb, mgp_messages *msgs, mgp_g
...
+  auto call = [&](py::Object py_graph) -> utils::BasicResult<std::optional<py::ExceptionInfo>, mgp_value *> {
+    py::Object py_args(MgpListToPyTuple(args, py_graph.Ptr()));
+    if (!py_args) return {py::FetchError()};
+    auto py_res = py_cb.Call(py_graph, py_args);
+    if (!py_res) return {py::FetchError()};
+    mgp_value *ret_val = PyObjectToMgpValueWithPythonExceptions(py_res.Ptr(), memory);
+    if (ret_val == nullptr) {
+      return {py::FetchError()};
+    }
+    return std::move(ret_val);
```
kostasrim, line 1114: Moving a pointer? What does this do exactly? :D

src/query/procedure/py_module.cpp
```diff
@@ -1031,6 +1089,92 @@ void CallPythonTransformation(const py::Object &py_cb, mgp_messages *msgs, mgp_g
...
+    py::Object py_graph(MakePyGraph(graph, memory));
+    if (py_graph) {
+      try {
+        auto maybe_result = call(py_graph);
+        if (!maybe_result.HasError()) {
+          static_cast<void>(mgp_func_result_set_value(result, maybe_result.GetValue(), memory));
+          return;
+        }
+        maybe_msg = error_to_msg(maybe_result.GetError());
+        cleanup(py_graph);
```
kostasrim, line 1162: This is dangerous, use `utils::OnScopeExit` to do RAII

kostasrim (changes requested):
Sorry for the long list of comments! Plz, qualify with `const` on nonmutable objects and simplify the logic in some places! You are almost there, this is a big PR :) 

Good job :)

tests/e2e/magic_functions/common.py
jbajic: Actually, it would be fairly easy, since we can already use parts of gqlalchemy for our test, especially setup and teardown. I agree this part is not needed unless it would be used across all e2e tests (that are using python)

src/query/procedure/mg_procedure_impl.hpp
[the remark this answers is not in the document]
jmatak: Holding the `TypedValue` in the `std::optional` container now.

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -2376,31 +2391,54 @@ mgp_error mgp_module_add_write_procedure(mgp_module *module, const char *name, m
...
+  static_assert(std::is_same_v<TCall, mgp_proc> || std::is_same_v<TCall, mgp_func>,
+                "Argument can be added only to mgp_proc or mgp_func");
+
   return WrapExceptions([=] {
+    std::string type_name;
+    if constexpr (std::is_same_v<TCall, mgp_proc>) {
+      type_name = "procedure";
+    } else if constexpr (std::is_same_v<TCall, mgp_func>) {
+      type_name = "function";
+    }
```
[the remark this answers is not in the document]
jmatak: Introduced std::invoke as @jbajic proposed :+1:

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -2376,31 +2391,54 @@ mgp_error mgp_module_add_write_procedure(mgp_module *module, const char *name, m
   return WrapExceptions([=] { return mgp_module_add_procedure(module, name, cb, {.is_write = true}); }, result);
 }
 
-mgp_error mgp_proc_add_arg(mgp_proc *proc, const char *name, mgp_type *type) {
+namespace {
+template <typename TCall>
+mgp_error mgp_add_arg(TCall *callable, const char *name, mgp_type *type) {
```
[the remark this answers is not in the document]
jmatak: We have this in more places in our `mg_procedure_impl.cpp`. I'll fix ones related to functions and another PR for the ones not related.

include/mgp.py
```diff
@@ -1414,27 +1458,84 @@ def graph(self) -> Graph:
         return self._graph
 
 
-def transformation(func: typing.Callable[..., Record]):
+def transformation(func: typing.Callable):
```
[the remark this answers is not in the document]
jmatak: No reason, I think this happened by accident...

src/query/procedure/mg_procedure_impl.cpp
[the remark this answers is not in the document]
jmatak: Similar situation in `PrintProcSignature`. Should I change both in this PR or should I make a separate one?

include/mg_procedure.h
```diff
@@ -1502,6 +1507,76 @@ typedef void (*mgp_trans_cb)(struct mgp_messages *, struct mgp_graph *, struct m
...
+/// must receive them through openCypher. Optional arguments must follow the
+/// required arguments.
+///
+/// The `name` must be a valid identifier, following the same rules as the
+/// function `name` in mgp_module_add_function.
+///
+/// Passed in `type` describes what kind of values can be used as the argument.
+///
+/// `default_value` is copied and set as the default value for the argument.
+/// Don't forget to call mgp_value_destroy when you are done using
```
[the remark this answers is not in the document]
jmatak: Why is it redundant?

include/mg_procedure.h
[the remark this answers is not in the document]
antaljanosbenjamin: I think this is not clear right now, I would even say we have some bug related to this. When calling a procedures/function/transformation, we convert its name to lowercase from the invocation, but when registering them from query modules, we don't care if they are not lowercase, so registering a "UpperCase" procedures/function/transformation is possible, but calling it isn't. I might be wrong here though.

src/query/procedure/mg_procedure_impl.cpp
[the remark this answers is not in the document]
antaljanosbenjamin: Even better :rocket:

src/query/procedure/module.cpp
antaljanosbenjamin: What Jure said!

tests/unit/query_function_mgp_module.cpp
jmatak: Since it is obvious that Functions are fully upper case, I don't think it is necessary to create an E2E test to confirm that, but rather change this test to prevent adding functions with similar upper-case names?

jmatak:
@kostasrim Just the reminder to try the unit tests locally, since they are working for me but failing on Actions

antaljanosbenjamin (changes requested):
There is one thing I don't like: the upper/lower case conversion of function names. Why? Three reasons:
1. transformations and procedures doesn't work this way
2. if the name of the function or module contains an uppercase letter, then the function cannot be called (but it can be loaded without any issue)
3. there can two functions that are only differ in casing, then always the lowercase function is called.

@gitbuda What do you think about this? 

In my opinion there is only one way to fix it: do not capitalize the function names in the AST, just in the `NameToFunction` function, so for user defined functions we can use the original casing.

Everything else would introduce inconsistent behavior, strange edge cases, etc.

gitbuda:
> There is one thing I don't like: the upper/lower case conversion of function names. Why? Three reasons:
> 
> 1. transformations and procedures doesn't work this way
> 2. if the name of the function or module contains an uppercase letter, then the function cannot be called (but it can be loaded without any issue)
> 3. there can two functions that are only differ in casing, then always the lowercase function is called.
> 
> @gitbuda What do you think about this?
> 
> In my opinion there is only one way to fix it: do not capitalize the function names in the AST, just in the `NameToFunction` function, so for user defined functions we can use the original casing.
> 
> Everything else would introduce inconsistent behavior, strange edge cases, etc.

I think your idea makes sense... In theory, nothing against the change, a few questions to help us decide:
1. Time to implement?
2. What will break?

tests/e2e/magic_functions/functions/c_write.cpp
```diff
@@ -0,0 +1,79 @@
...
+      return 1;
+    }
+
+    mgp_type *type_string{};
+    mgp_type_string(&type_string);
+    err_code = mgp_func_add_arg(func, "name", type_string);
+    if (err_code != MGP_ERROR_NO_ERROR) {
+      return 1;
+    }
+
```
kostasrim, line 63: plz test `mgp_func_add_opt_arg` as well!

tests/e2e/magic_functions/functions/c_read.cpp
```diff
@@ -0,0 +1,135 @@
...
+  auto err_code = mgp_func_result_set_value(result, value, memory);
+  if (err_code != MGP_ERROR_NO_ERROR) {
+    mgp_func_result_set_error(result, "Failed to construct return value!", memory);
+  }
+}
+
+static void return_null(struct mgp_list *args, mgp_func_context *ctx, mgp_func_result *result,
+                        struct mgp_memory *memory) {
+  mgp_value *value{};
+  mgp_value_make_null(memory, &value);
```
kostasrim, line 77: free this at the end! Consider, our or `OnScopeExit` util function! (same change for all the occurrences of mgp_value_make_*** functions around this file.

tests/e2e/magic_functions/functions/c_read.cpp
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  double result;
+  int is_int;
+  mgp_value_is_int(value, &is_int);
+
+  if (is_int) {
+    int64_t result_int;
+    mgp_value_get_int(value, &result_int);
+    result = static_cast<double>(result_int);
```
kostasrim, line 45: Why are you casting an `int` to a `double`?

src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -2104,8 +2104,6 @@ antlrcpp::Any CypherMainVisitor::visitFunctionInvocation(MemgraphCypher::Functio
         storage_->Create<Aggregation>(expressions[1], expressions[0], Aggregation::Op::COLLECT_MAP));
   }
 
-  auto function = NameToFunction(function_name);
-  if (!function) throw SemanticException("Function '{}' doesn't exist.", function_name);
```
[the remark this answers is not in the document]
jmatak: The reason this was removed is that the check was transferred to the `Function` object constructor at `ast.clp` line #854.  Before this, two different function pointers were created, which wasn't a problem before introducing user functions, however, with the introduction we are moving lambdas and unnecessarily creating memory.

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -2716,3 +2777,21 @@ mgp_error mgp_module_add_transformation(mgp_module *module, const char *name, mg
     module->transformations.emplace(name, mgp_trans(name, cb, memory));
   });
 }
+
+namespace {
+inline mgp_func *MgpModuleAddFunction(mgp_module &module, const std::string name, const mgp_func_cb cb) {
```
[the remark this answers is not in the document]
jmatak: Today I learned: https://stackoverflow.com/questions/1759300/when-should-i-write-the-keyword-inline-for-a-function-method
:exploding_head: , tnx

tests/e2e/magic_functions/functions/c_read.cpp
jmatak: It was going to be cast to double anyway. Just wanted to make sure it happens. I guess that is not the best practice?

tests/e2e/magic_functions/functions/c_write.cpp
jmatak: Added this to the `c_read.cpp`

src/query/procedure/py_module.cpp
```diff
@@ -1100,6 +1178,39 @@ PyObject *PyQueryModuleAddTransformation(PyQueryModule *self, PyObject *cb) {
...
+  const auto *name = PyUnicode_AsUTF8(py_name.Ptr());
+  if (!name) return nullptr;
+  if (!IsValidIdentifierName(name)) {
+    PyErr_SetString(PyExc_ValueError, "Function name is not a valid identifier");
+    return nullptr;
+  }
+  auto *memory = self->module->functions.get_allocator().GetMemoryResource();
+  mgp_func func(
+      name,
+      [py_cb](mgp_list *args, mgp_func_context *func_ctx, mgp_func_result *result, mgp_memory *memory) {
```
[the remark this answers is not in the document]
jmatak: Fixed with https://github.com/memgraph/memgraph/pull/345/commits/697cbe55b9222a517d102c32d960d57f3f4efd4d

jmatak:
What's left is to define the Functions name resolving. After that is defined, a new review could be done.

src/query/procedure/module.hpp
```diff
@@ -147,4 +150,58 @@ std::optional<std::pair<procedure::ModulePtr, const mgp_proc *>> FindProcedure(
...
+  if (n_args < c_args_sz || (n_args - c_args_sz > c_opt_args_sz)) {
+    if (callable.args.empty() && callable.opt_args.empty()) {
+      throw QueryRuntimeException("'{}' requires no arguments.", fully_qualified_name);
+    } else if (callable.opt_args.empty()) {
+      throw QueryRuntimeException("'{}' requires exactly {} {}.", fully_qualified_name, c_args_sz,
+                                  c_args_sz == 1U ? "argument" : "arguments");
+    } else {
+      throw QueryRuntimeException("'{}' requires between {} and {} arguments.", fully_qualified_name, c_args_sz,
+                                  c_args_sz + c_opt_args_sz);
+    }
```
jbajic, line 181:
```suggestion
    if (callable.args.empty() && callable.opt_args.empty()) {
      throw QueryRuntimeException("'{}' requires no arguments.", fully_qualified_name);
    } 
    if (callable.opt_args.empty()) {
      throw QueryRuntimeException("'{}' requires exactly {} {}.", fully_qualified_name, c_args_sz,
                                  c_args_sz == 1U ? "argument" : "arguments");
    }
     throw QueryRuntimeException("'{}' requires between {} and {} arguments.", fully_qualified_name, c_args_sz,
                                  c_args_sz + c_opt_args_sz);
    }
```

src/query/procedure/module.hpp
```diff
@@ -147,4 +150,58 @@ std::optional<std::pair<procedure::ModulePtr, const mgp_proc *>> FindProcedure(
...
+  }
+  args_list.elems.reserve(n_args);
+  for (size_t i = 0; i < n_args; ++i) {
+    auto arg = args[i];
+    std::string_view name;
+    const query::procedure::CypherType *type;
+    if (c_args_sz > i) {
```
jbajic, line 188: Just a suggestion to make code more readable, create a lambda outside of the loop that will determine if the argument is optional or not, e.g.:
```
auto is_optional = [c_args_sz](int i) { return c_args_sz > i; };
```
and use it like this
```suggestion
    if (!is_optional(i)) {
```

src/query/interpret/awesome_memgraph_functions.cpp
```diff
@@ -1174,6 +1177,41 @@ TypedValue Duration(const TypedValue *args, int64_t nargs, const FunctionContext
   MapNumericParameters<Number>(parameter_mappings, args[0].ValueMap());
   return TypedValue(utils::Duration(duration_parameters), ctx.memory);
 }
+
+std::function<TypedValue(const TypedValue *, int64_t, const FunctionContext &)> UserFunction(
```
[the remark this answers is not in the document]
jmatak: Why make primitive argument `const`?

src/query/interpret/awesome_memgraph_functions.cpp
```diff
@@ -1174,6 +1177,41 @@ TypedValue Duration(const TypedValue *args, int64_t nargs, const FunctionContext
...
+    const mgp_func &func, std::string &fully_qualified_name) {
+  return [func, fully_qualified_name](const TypedValue *args, int64_t nargs, const FunctionContext &ctx) -> TypedValue {
+    const auto &func_cb = func.cb;
```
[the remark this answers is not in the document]
jmatak: This inserted would check whether the argument to the function is List, and that is not always the case, it could be whatever we want. Also one thing, I have the argument checking in  `procedure::ConstructArguments`

src/query/interpret/awesome_memgraph_functions.cpp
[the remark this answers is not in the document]
jbajic: Since you are not changing the passed value in the given scope

src/query/interpret/awesome_memgraph_functions.cpp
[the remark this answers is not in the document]
jbajic: You are right, sorry I misunderstood the check that is performed here

tests/e2e/magic_functions/functions/py_read.py
```diff
@@ -0,0 +1,32 @@
...
+# by the Apache License, Version 2.0, included in the file
+# licenses/APL.txt.
+
+import mgp
+
+
+@mgp.function
+def return_function_argument(ctx: mgp.FuncCtx, argument: mgp.Any):
+    return argument
+
```
jbajic, line 18: For me this function does not work:
```python
from typing import List
@mgp.function
def reverse_list(ctx: mgp.FuncCtx, first: List):
    return reversed(first)
```
I get `ValueError: Unsupported PyObject conversion` on running:
`return py_read.reverse_list([1,2,3,4]);`
I've tried the same with procedures and it works there. Not sure what's going on :thinking:

jmatak: I've tried this code, and get the `Unsupported conversion` (I am pretty sure that generators are not supported as the resulting type):

```
@mgp.read_proc
def reverse_list(ctx: mgp.ProcCtx, listarg: mgp.List[mgp.Any]) -> mgp.Record(list=mgp.List[mgp.Any]):
    return mgp.Record(list=reversed(listarg))
```

However I would like to know the example that worked for you?

jmatak: I am not even able to import the query modules if I don't specify the arguments for List objects type

jbajic: aha sry you are right, I've forgotten reversed returns a generator, and in procedures, I used list instead of the generator (returning just the argument for testing purposes)...

jbajic (approved):
It all looks good to me, good job! :tada:

tests/unit/query_function_mgp_module.cpp
jmatak: Added e2e test that checks the case sensitivity of user-written functions.

src/query/procedure/mg_procedure_impl.cpp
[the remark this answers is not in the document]
antaljanosbenjamin: Both way is fine, even if you leave the other as is. As soon as we modify that, we will change it.

antaljanosbenjamin (approved):
LGTM! Thank you for the tremendous amount of work, we really appreciate it!

src/query/interpret/awesome_memgraph_functions.cpp
```diff
@@ -1259,6 +1297,14 @@ std::function<TypedValue(const TypedValue *, int64_t, const FunctionContext &ctx
   if (function_name == "LOCALDATETIME") return LocalDateTime;
   if (function_name == "DURATION") return Duration;
 
+  const auto &maybe_found =
+      procedure::FindFunction(procedure::gModuleRegistry, function_name, utils::NewDeleteResource());
+
+  if (maybe_found) {
+    const auto *func = (*maybe_found).second;
+    return UserFunction(*func, function_name);
```
antaljanosbenjamin, line 1305: Ohh, I forgot to comment on this one! Your forgot to save the ModulePtr. Without that it is not guaranteed that the module won't be unloaded. I think we have talked about this previously, but if you need help, let me know!

antaljanosbenjamin (changes requested):
Sorry for lying before. As soon as I hit the submit button, I realized that I forgot to comment on this.

src/query/interpret/awesome_memgraph_functions.cpp
```diff
@@ -1174,6 +1177,46 @@ TypedValue Duration(const TypedValue *args, int64_t nargs, const FunctionContext
...
   return TypedValue(utils::Duration(duration_parameters), ctx.memory);
 }
+
+std::function<TypedValue(const TypedValue *, const int64_t, const FunctionContext &)> UserFunction(
+    const procedure::ModulePtr &module_ptr, const mgp_func &func, const std::string &fully_qualified_name) {
+  /// Module pointer should be propagated because of the lock aquired. It prevents reloading module while function is
+  /// executing.
+
+  // NOLINTNEXTLINE(clang-diagnostic-unused-lambda-capture)
+  return [func, fully_qualified_name, &module_ptr](const TypedValue *args, int64_t nargs,
```
antaljanosbenjamin, line 1187:
```suggestion
std::function<TypedValue(const TypedValue *, const int64_t, const FunctionContext &)> UserFunction(
    procedure::ModulePtr &&module_ptr, const mgp_func &func, const std::string &fully_qualified_name) {
  /// Module pointer should be propagated because of the lock aquired. It prevents reloading module while function is
  /// executing.

  // NOLINTNEXTLINE(clang-diagnostic-unused-lambda-capture)
  return [func, fully_qualified_name, module_ptr = std::move(module_ptr)](const TypedValue *args, int64_t nargs,
```

You have to move the `module_ptr`, because a raw reference won't prevent destruction.

src/query/interpret/awesome_memgraph_functions.cpp
```diff
@@ -1259,6 +1302,15 @@ std::function<TypedValue(const TypedValue *, int64_t, const FunctionContext &ctx
   if (function_name == "LOCALDATETIME") return LocalDateTime;
   if (function_name == "DURATION") return Duration;
 
+  const auto &maybe_found =
+      procedure::FindFunction(procedure::gModuleRegistry, function_name, utils::NewDeleteResource());
+
+  if (maybe_found) {
+    const auto &module_ptr = (*maybe_found).first;
+    const auto *func = (*maybe_found).second;
+    return UserFunction(module_ptr, *func, function_name);
```
antaljanosbenjamin, line 1311:
```suggestion
    return UserFunction(std::move(module_ptr), *func, function_name);
```

antaljanosbenjamin (changes requested):

src/query/procedure/mg_procedure_impl.hpp
```diff
@@ -562,14 +562,37 @@ struct mgp_result {
...
+  static mgp_graph NonWritableGraph(memgraph::query::DbAccessor &acc, memgraph::storage::View view) {
+    return mgp_graph{&acc, view, nullptr};
+  }
 };
 
+struct mgp_func_context {
+  memgraph::query::DbAccessor *impl;
+  memgraph::storage::View view;
+
+  // Prevents user to use ExecutionContext in writable callables
```
kostasrim, line 594: This lonely comment looks a bit weird. Consider moving it on line 589

tests/e2e/magic_functions/functions/c_write.cpp
```diff
@@ -0,0 +1,79 @@
...
+  mgp_list_at(args, 0, &value);
+  mgp_value_get_vertex(value, &vertex);
+
+  const char *name;
+  mgp_list_at(args, 1, &value);
+  mgp_value_get_string(value, &name);
+
+  mgp_list_at(args, 2, &value);
+
+  auto err_code = mgp_vertex_set_property(vertex, name, value);  // This should set an error
```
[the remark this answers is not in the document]
jmatak: It's there because the method should raise an error if trying to write to the graph while being in functions.

tests/e2e/magic_functions/functions/c_read.cpp
jmatak: Ohh, now I understand why would you want that. However, I just wanted to return the double precision result without caring about the arguments since this was only for the purpose of testing.

antaljanosbenjamin (approved):

tests/e2e/magic_functions/functions/c_read.cpp
```diff
@@ -0,0 +1,185 @@
...
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#include <functional>
+#include <stdexcept>
+
+#include "mg_procedure.h"
+
+namespace {
+class OnScopeExit {
```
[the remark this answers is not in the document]
kostasrim: This is not fixed

src/query/procedure/mg_procedure_impl.hpp
kostasrim: This is not fixed either

kostasrim (changes requested):

tests/e2e/magic_functions/functions/c_write.cpp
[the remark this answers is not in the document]
kostasrim: I meant to remove the comment at the end of the line. Is really no needed.

kostasrim (approved):

src/query/interpret/awesome_memgraph_functions.cpp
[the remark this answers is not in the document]
antaljanosbenjamin: This might not be the best, but because it doesn't affect the functionality/behavior of the function, I think we can merge this. The `const` qualifier for primitive by value parameters is only important in the definition, not in the declaration. In other words the `const` for primitive by value parameters is important only at the implementation of the function, in this case in the lambda.

gitbuda:
Closes #231

PR #346, opened 2022-02-03
[master < MG] Fix folly in toolchain-v4

antaljanosbenjamin:
[master < Task] PR

No need to update the docs, as it just changes the toolchain.


gitbuda:
I had a strange issue with building gdb, somehow `/usr/include/ncurses.h` (referenced while building gdb) was a broken symlink on my system (Ubuntu 20.04), I've managed to fix the problem by:
```
apt purge libncurses-dev
apt install libncurses-dev
```
We don't reference it in the `TOOLCHAIN_BUILD_DEPS` (I assume it comes installed on the system) so, nothing has to really change in the scripts... @antaljanosbenjamin I can add `libncurses-dev` to the Ubuntu script if you think that's a good idea?

gitbuda (approved):
I've managed to build the toolchain on Ubuntu 20.04, so this looks good 😄

antonio2368:
> 

I had to update the GDB to 11.2 from 11.1 to make it work on ARM so maybe that introduced some new changes on deps for deb based machines. 
@jbajic can add missing deps when he notices them while building toolchain on each OS.

antonio2368 (approved):

PR #351, opened 2022-02-09
[master < T0808-MG] Implement foreach clause

### LINKED ISSUE #308: The FOREACH language construct for performing an operation on every list element
While working on different query examples I was frequently missing the FOREACH construct. Also, noticed that our developer users feel that would be a nice addition.

kostasrim:
[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR

Closes #308


tests/unit/query_plan.cpp
```diff
@@ -1553,4 +1553,10 @@ TYPED_TEST(TestPlanner, LabelPropertyInListWhereLabelPropertyOnRight) {
   }
 }
 
+TYPED_TEST(TestPlanner, Foreach) {
+  AstStorage storage;
+  auto *i = NEXPR("i", IDENT("i"));
+  auto *query = QUERY(SINGLE_QUERY(FOREACH(i, {})));
+  CheckPlan<TypeParam>(query, storage, ExpectForeach());
+}
```
gitbuda, line 1561: Add a couple of more tests here, or at least one which is a bit more complex 🧪

gitbuda (commented):
The PR is still in the draft phase, but I put a couple of comments 😄 It has a bunch of stuff, mostly adding new stuff, so the whole PR is mainly about the correctness of the newly added stuff rather than breaking the existing functionalities.

In addition to the comments in the code, please start writing docs pages under `memgraph/docs` and link them here. Outline basic things and then docs people can take over at some point and build examples and how-to guides 💪

Not far from being ready for 🚀

tests/unit/query_plan_read_write_typecheck.cpp
```diff
@@ -246,3 +246,9 @@ TEST_F(ReadWriteTypeCheckTest, ConstructNamedPath) {
 
   CheckPlanType(last_op.get(), RWType::R);
 }
+
+TEST_F(ReadWriteTypeCheckTest, Foreach) {
```
kostasrim, line 250: Double-check this.

src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -1030,6 +1037,9 @@ antlrcpp::Any CypherMainVisitor::visitClause(MemgraphCypher::ClauseContext *ctx)
   if (ctx->loadCsv()) {
     return static_cast<Clause *>(ctx->loadCsv()->accept(this).as<LoadCsv *>());
   }
+  if (ctx->foreach ()) {
+    return static_cast<Clause *>(ctx->foreach ()->accept(this).as<Foreach *>());
```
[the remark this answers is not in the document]
kostasrim: clang-format :D

src/query/plan/preprocess.cpp
```diff
@@ -546,6 +548,16 @@ std::vector<SingleQueryPart> CollectSingleQueryParts(SymbolTable &symbol_table,
       if (auto *merge = utils::Downcast<query::Merge>(clause)) {
         query_part->merge_matching.emplace_back(Matching{});
         AddMatching({merge->pattern_}, nullptr, symbol_table, storage, query_part->merge_matching.back());
+      } else if (auto *foreach = utils::Downcast<query::Foreach>(clause)) {
+        // todo kostasrim
+        // fix preprocessing
+        // add algorithm for this
```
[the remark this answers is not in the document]
kostasrim: This was an old comment. I added the preprocessing for Merge + I decided not to use `ranges::foreach` to implement the algorithm (if you look the for-loop is actually a for_each algorithm)

tests/benchmark/query/execution.cpp
```diff
@@ -496,5 +496,42 @@ BENCHMARK_TEMPLATE(Unwind, MonotonicBufferResource)
     ->Unit(benchmark::kMicrosecond);
 
 BENCHMARK_TEMPLATE(Unwind, PoolResource)->Ranges({{4, 1U << 7U}, {512, 1U << 13U}})->Unit(benchmark::kMicrosecond);
+/*
```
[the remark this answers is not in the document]
kostasrim: Test modified :)

tests/unit/query_plan.cpp
kostasrim: Although I think the tests in `gql_behave` suffice (since they test the planning implicitly) I added a few more for completeness!

src/query/frontend/semantic/symbol_generator.cpp
```diff
@@ -264,6 +264,20 @@ bool SymbolGenerator::PostVisit(Match &) {
...
+bool SymbolGenerator::PreVisit(Foreach &for_each) {
+  scope_.in_foreach = true;
+  const auto &name = for_each.named_expression_->name_;
+  if (HasSymbol(name)) {
+    throw RedeclareVariableError(name);
+  }
+  for_each.named_expression_->MapTo(CreateSymbol(name, true));
+  return true;
+}
+bool SymbolGenerator::PostVisit(Foreach &) {
```
kostasrim, line 276: Check that symbols don't leak here.

tests/benchmark/query/execution.cpp
```diff
@@ -497,4 +497,39 @@ BENCHMARK_TEMPLATE(Unwind, MonotonicBufferResource)
 
 BENCHMARK_TEMPLATE(Unwind, PoolResource)->Ranges({{4, 1U << 7U}, {512, 1U << 13U}})->Unit(benchmark::kMicrosecond);
 
+template <class TMemory>
+// NOLINTNEXTLINE(google-runtime-references)
+static void Foreach(benchmark::State &state) {
+  query::AstStorage ast;
+  query::Parameters parameters;
```
jbajic, line 504: How does foreach works with parameters?

src/query/plan/preprocess.cpp
```diff
@@ -546,6 +548,13 @@ std::vector<SingleQueryPart> CollectSingleQueryParts(SymbolTable &symbol_table,
       if (auto *merge = utils::Downcast<query::Merge>(clause)) {
         query_part->merge_matching.emplace_back(Matching{});
         AddMatching({merge->pattern_}, nullptr, symbol_table, storage, query_part->merge_matching.back());
+      } else if (auto *foreach = utils::Downcast<query::Foreach>(clause)) {
+        for (auto *clause : foreach->clauses_) {
+          if (auto *merge = utils::Downcast<query::Merge>(clause)) {
+            query_part->merge_matching.emplace_back(Matching{});
+            AddMatching({merge->pattern_}, nullptr, symbol_table, storage, query_part->merge_matching.back());
+          }
+        }
```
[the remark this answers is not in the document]
kostasrim: There are no match clauses here and this is enforced by the grammar? Am I missing something here?

src/query/plan/operator.cpp
```diff
@@ -4024,4 +4026,76 @@ UniqueCursorPtr LoadCsv::MakeCursor(utils::MemoryResource *mem) const {
...
+      cache_ = expr_result.ValueList();
+      index_ = cache_.begin();
+    }
+
+    if (index_ == cache_.end()) {
+      if (!is_nested_ || !input_cursor_->Pull(frame, context)) {
+        return false;
+      }
+      index_ = cache_.begin();
+    }
```
antaljanosbenjamin, line 4062: I think checking two different things for pulling is not good. I think `index_` should be initialized in the ctor with `cache_.end()`, then `index_ == cache_.end()` can be used to check whether another pull is necessary or not.
```suggestion
    while (index_ == cache_.end()) {
      if (!input_cursor_->Pull(frame, context)) {
        return false;
      }

      ExpressionEvaluator evaluator(&frame, context.symbol_table, context.evaluation_context, context.db_accessor,
                                    storage::View::NEW);
      TypedValue expr_result = expression->Accept(evaluator);
      if (!expr_result.IsList()) {
        throw QueryRuntimeException("FOREACH expression must resolve to a list, but got '{}'.", expr_result.type());
      }
      cache_ = expr_result.ValueList();
      index_ = cache_.begin();
    }
```

Furthermore I don't see why it is important to know whether we are in a nested foreach clause or not. How does that should alter the behavior?

antaljanosbenjamin: Some queries that I think point out the problems:
```
memgraph> MATCH (n) DETACH DELETE n;
Empty set (0.001 sec)
memgraph> CREATE (:V {prop: [[1, 2], []]});
Empty set (0.000 sec)
memgraph> MATCH (n) RETURN n;
+---------------------------+
| n                         |
+---------------------------+
| (:V {prop: [[1, 2], []]}) |
+---------------------------+
1 row in set (0.027 sec)
memgraph> MATCH (n) FOREACH ( li IN n.prop | FOREACH ( value IN li | CREATE (m:F {prop:value})));
Empty set (0.001 sec)
memgraph> MATCH (n) RETURN n;
+---------------------------+
| n                         |
+---------------------------+
| (:V {prop: [[1, 2], []]}) |
| (:F {prop: 1})            |
| (:F {prop: 2})            |
| (:F {prop: 1})            |
| (:F {prop: 2})            |
+---------------------------+
5 rows in set (0.001 sec)
```
If the original node is created as `CREATE (:V {prop: [[], [1, 2]]});`, then memgraph crashes.

In my opinion rewriting the cursor with the proposed changes fixes this issue.

src/query/plan/rule_based_planner.hpp
```diff
@@ -530,6 +535,28 @@ class RuleBasedPlanner {
...
+    bound_symbols.insert(symbol);
+    input_op = std::make_unique<plan::Foreach>(std::move(input_op), foreach->named_expression_->expression_, symbol,
+                                               is_nested);
+    for (auto *clause : foreach->clauses_) {
+      if (auto *nested_for_each = utils::Downcast<query::Foreach>(clause)) {
+        HandleForeachClause(nested_for_each, input_op, symbol_table, bound_symbols, query_part, merge_id, is_write,
+                            true);
+        continue;
+      } else if (auto *merge = utils::Downcast<query::Merge>(clause)) {
+        input_op = GenMerge(*merge, std::move(input_op), query_part.merge_matching[merge_id++]);
```
antaljanosbenjamin, line 553: This fails in the following query: `EXPLAIN MATCH (n) FOREACH ( li IN n.prop | CREATE (:F4 {prop: li}) FOREACH ( value IN li | MERGE (m:F {prop:value})));`.

src/query/plan/preprocess.cpp
```diff
@@ -546,6 +548,13 @@ std::vector<SingleQueryPart> CollectSingleQueryParts(SymbolTable &symbol_table,
       if (auto *merge = utils::Downcast<query::Merge>(clause)) {
         query_part->merge_matching.emplace_back(Matching{});
         AddMatching({merge->pattern_}, nullptr, symbol_table, storage, query_part->merge_matching.back());
+      } else if (auto *foreach = utils::Downcast<query::Foreach>(clause)) {
+        for (auto *clause : foreach->clauses_) {
+          if (auto *merge = utils::Downcast<query::Merge>(clause)) {
+            query_part->merge_matching.emplace_back(Matching{});
+            AddMatching({merge->pattern_}, nullptr, symbol_table, storage, query_part->merge_matching.back());
```
antaljanosbenjamin, line 555: Based on the comments of `merge_matching` I think this is not good, because you are modifying `merge_matching` without adding anything to `remaining_clauses`.

To be honest I don't see the whole picture, but I think we should check whether we have any similar clause that defines its scope and do something similar here.

src/query/plan/rule_based_planner.hpp
```diff
@@ -530,6 +535,28 @@ class RuleBasedPlanner {
     }
     return std::make_unique<plan::Merge>(std::move(input_op), std::move(on_match), std::move(on_create));
   }
+
+  void HandleForeachClause(query::Foreach *foreach, std::unique_ptr<LogicalOperator> &input_op,
```
antaljanosbenjamin, line 539: I don't like that `input_op` is an out parameter. Why don't the function has a return value instead?

src/query/plan/rule_based_planner.hpp
```diff
@@ -530,6 +535,28 @@ class RuleBasedPlanner {
...
+    for (auto *clause : foreach->clauses_) {
+      if (auto *nested_for_each = utils::Downcast<query::Foreach>(clause)) {
+        HandleForeachClause(nested_for_each, input_op, symbol_table, bound_symbols, query_part, merge_id, is_write,
+                            true);
+        continue;
+      } else if (auto *merge = utils::Downcast<query::Merge>(clause)) {
+        input_op = GenMerge(*merge, std::move(input_op), query_part.merge_matching[merge_id++]);
+        continue;
+      }
+      auto op = HandleWriteClause(clause, input_op, symbol_table, bound_symbols);
```
antaljanosbenjamin, line 556: It also hurts me that `HandleWriteClause` moves out `input_op`, but it is not your issue, so don't bother with it. I just had to vent :smile:

antaljanosbenjamin (changes requested):
I still have to understand the rule based planner and check the tests, but my brain is burnt out.

src/query/plan/operator.cpp
```diff
@@ -4024,4 +4026,76 @@ UniqueCursorPtr LoadCsv::MakeCursor(utils::MemoryResource *mem) const {
...
+      if (!input_cursor_->Pull(frame, context)) {
+        return false;
+      }
+
+      ExpressionEvaluator evaluator(&frame, context.symbol_table, context.evaluation_context, context.db_accessor,
+                                    storage::View::NEW);
+      TypedValue expr_result = expression->Accept(evaluator);
+      if (!expr_result.IsList()) {
+        throw QueryRuntimeException("FOREACH expression must resolve to a list, but got '{}'.", expr_result.type());
+      }
```
antaljanosbenjamin, line 4052: Don't we want to handle the null values in a nice way? 

```
CREATE (n);
MATCH (n) FOREACH ( li IN n.prop | CREATE (:F4 {prop: li}) CREATE (:F5 {prop2: li}) SET n.li = li);
```
Does notthing in neo4j, but throws and exception in our case. I think this is a nice little addition that really makes users' life easier.

antaljanosbenjamin:
After today's daily we discussed this PR and we found a good example why the current approach is not good:
```
CREATE (n {prop: [1,2]});
MATCH (n) FOREACH ( i IN n.prop | CREATE (:V { i: i})) RETURN n;
```
the second query should return one row, but for us it returns two. That is the main reason we should somehow embed another plane inside FOREACH instead of making one, "linear" plan.

Maybe we don't need the embed a plan, we can build one plan, but the `FOREACH` should have an additional side branch which contains all the inside clauses. Let's call them `main` and `side`. When the `FOREACH` is pulled, it should pull `main` once, populate the frame with everything that is necessary for `side`, and then it should exhaust `side` and "pass through" the original row from `main`. I don't see the exact implementation, but I think this is doable.

Tomorrow we will discuss this in more details.

antaljanosbenjamin:
Another query that doesn't behave correctly in my opinion: `MATCH (n) FOREACH ( li IN n.prop | CREATE (:F4 {prop: li}) CREATE (m:F5 {prop2: li}) SET n.li = li) RETURN m;`

It should fail because m shouldn't be defined in RETURN, but it succeeds:
```
memgraph> MATCH (n) DETACH DELETE n;
Empty set (0.001 sec)
memgraph> CREATE (n {prop: [1,2]});
Empty set (0.001 sec)
memgraph> MATCH (n) FOREACH ( li IN n.prop | CREATE (:F4 {prop: li}) CREATE (m:F5 {prop2: li}) SET n.li = li) RETURN m;
+------------------+
| m                |
+------------------+
| (:F5 {prop2: 1}) |
| (:F5 {prop2: 2}) |
+------------------+
2 rows in set (0.001 sec)
```

src/query/plan/preprocess.cpp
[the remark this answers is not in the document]
jbajic: Sorry, my bad I didn't fully understand the meaning of this part

src/query/plan/preprocess.cpp
```diff
@@ -526,6 +528,18 @@ void Filters::AnalyzeAndStoreFilter(Expression *expr, const SymbolTable &symbol_
   // as `expr1 < n.prop AND n.prop < expr2`.
 }
 
+static void ParseForeach(query::Foreach &foreach, SingleQueryPart &query_part, AstStorage &storage,
+                         SymbolTable &symbol_table) {
+  for (auto *clause : foreach.clauses_) {
+    if (auto *merge = utils::Downcast<query::Merge>(clause)) {
+      query_part.merge_matching.emplace_back(Matching{});
+      AddMatching({merge->pattern_}, nullptr, symbol_table, storage, query_part.merge_matching.back());
```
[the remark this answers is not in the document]
antaljanosbenjamin: So after deeply understanding what is happening I think you are right, `merge_matching` is handled properly. However, the changes you introduced with `FOREACH`, the above mentioned comment should be modified (and maybe a few other ones around that), because `merge_matching` contains part from clauses that are directly stored in `remaining_clauses`, but inside a clause.

tests/gql_behave/tests/memgraph_V1/features/foreach.feature
```diff
@@ -0,0 +1,240 @@
...
+  Scenario: Foreach set
+    Given an empty graph
+    And having executed
+    """
+    CREATE (n1 { marked: false })-[:RELATES]->(n2 { marked: false })
+    """
+    And having executed 
+    """
+    MATCH p=(n1)-[*]->(n2)
+    FOREACH (n IN nodes(p) | set n.marked = true)
```
antaljanosbenjamin, line 61:
```suggestion
    FOREACH (n IN nodes(p) | SET n.marked = true)
```

tests/gql_behave/tests/memgraph_V1/features/foreach.feature
```diff
@@ -0,0 +1,240 @@
...
+  Scenario: Foreach remove
+    Given an empty graph
+    And having executed
+    """
+    CREATE (n1 { marked: false })-[:RELATES]->(n2 { marked: false })
+    """
+    And having executed 
+    """
+    MATCH p=(n1)-[*]->(n2)
+    FOREACH (n IN nodes(p) | remove n.marked)
```
antaljanosbenjamin, line 83:
```suggestion
    FOREACH (n IN nodes(p) | REMOVE n.marked)
```

tests/gql_behave/tests/memgraph_V1/features/foreach.feature
```diff
@@ -0,0 +1,240 @@
...
+  Scenario: Foreach delete
+    Given an empty graph
+    And having executed
+    """
+    CREATE (n1 { marked: false })-[:RELATES]->(n2 { marked: false })
+    """
+    And having executed 
+    """
+    MATCH p=(n1)-[*]->(n2)
+    FOREACH (n IN nodes(p) | detach delete n)
```
antaljanosbenjamin, line 105:
```suggestion
    FOREACH (n IN nodes(p) | DETACH DELETE n)
```

tests/unit/cypher_main_visitor.cpp
```diff
@@ -4088,3 +4088,88 @@ TEST_P(CypherMainVisitorTest, VersionQuery) {
   TestInvalidQuery("SHOW VERSIONS", ast_generator);
   ASSERT_NO_THROW(ast_generator.ParseQuery("SHOW VERSION"));
 }
+
+TEST_P(CypherMainVisitorTest, ForeachThrow) {
+  auto &ast_generator = *GetParam();
+  EXPECT_THROW(ast_generator.ParseQuery("FOREACH(i IN [1, 2] | "), SyntaxException);
+  EXPECT_THROW(ast_generator.ParseQuery("FOREACH(i IN [1, 2] | UNWIND"), SyntaxException);
+  EXPECT_THROW(ast_generator.ParseQuery("FOREACH(i IN [1, 2] | LOADCSV"), SyntaxException);
```
antaljanosbenjamin, line 4096: These test cases can be improved in my opinion, because it is possible they throw because the nested statement is not valid. Therefore I would extend them to be valid: `UNWIND [1,2,3] AS j CREATE (n)`, and maybe switch `LOADCSV` to something else. If I missed the point of these tests, then adding new test cases are also fine to me.

tests/unit/plan_pretty_print.cpp
```diff
@@ -911,3 +911,34 @@ TEST_F(PrintToJsonTest, CallProcedure) {
...
+      std::make_shared<plan::Foreach>(nullptr, std::move(create), LIST(LITERAL(1)), x);
+
+  Check(foreach.get(), R"sep(
+          {
+           "expression": "(ListLiteral [1])",
+           "input": {
+            "name": "Once"
+           },
+           "name": "Foreach",
+           "output_symbol": "x",
```
antaljanosbenjamin, line 929: "output_symbol" definitely looks bad here, I an sure it should be changed something else as I mentioned in my previous comment.

antaljanosbenjamin (changes requested):
The improvement is quite good, however there are a few small things to change in my opinion.

src/query/frontend/ast/ast.lcp
```diff
@@ -711,6 +711,7 @@ cpp<#
 (lcp:define-class identifier (expression)
   ((name "std::string" :scope :public)
    (user-declared :bool :initval "true" :scope :public)
+   (in-foreach :bool :initval "false" :scope :public)
```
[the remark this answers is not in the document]
kostasrim: removed!

src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -974,6 +975,12 @@ antlrcpp::Any CypherMainVisitor::visitSingleQuery(MemgraphCypher::SingleQueryCon
       }
       check_write_procedure("WITH");
       has_update = has_return = has_optional_match = false;
+    } else if (utils::IsSubtype(clause_type, Foreach::kType)) {
+      check_write_procedure("FOREACH");
+      if (has_return) {
+        throw SemanticException("FOREACH can't be put after RETURN clause or after an update.");
+      }
+      has_update = true;
```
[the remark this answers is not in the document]
kostasrim: good catch!

src/query/frontend/semantic/symbol_generator.cpp
```diff
@@ -39,23 +41,26 @@ std::unordered_map<std::string, Identifier *> GeneratePredefinedIdentifierMap(
...
 auto SymbolGenerator::GetOrCreateSymbol(const std::string &name, bool user_declared, Symbol::Type type) {
-  auto search = scope_.symbols.find(name);
-  if (search != scope_.symbols.end()) {
-    auto symbol = search->second;
-    // Unless we have `ANY` type, check that types match.
-    if (type != Symbol::Type::ANY && symbol.type() != Symbol::Type::ANY && type != symbol.type()) {
-      throw TypeMismatchError(name, Symbol::TypeToString(symbol.type()), Symbol::TypeToString(type));
+  for (auto scope = scope_.rbegin(); scope != scope_.rend(); ++scope) {
+    if (scope->symbols.contains(name)) {
+      auto symbol = scope->symbols.find(name)->second;
```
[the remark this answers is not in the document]
kostasrim: I really like `contains` from an API point of view. However, the underlying container is `std::map` and the second lookup is redundant. +1 for the suggestion

src/query/frontend/semantic/symbol_generator.cpp
```diff
@@ -39,23 +41,26 @@ std::unordered_map<std::string, Identifier *> GeneratePredefinedIdentifierMap(
 }  // namespace
 
 SymbolGenerator::SymbolGenerator(SymbolTable *symbol_table, const std::vector<Identifier *> &predefined_identifiers)
-    : symbol_table_(symbol_table), predefined_identifiers_{GeneratePredefinedIdentifierMap(predefined_identifiers)} {}
+    : symbol_table_(symbol_table), predefined_identifiers_{GeneratePredefinedIdentifierMap(predefined_identifiers)} {
+  scope_.push_back(Scope());
```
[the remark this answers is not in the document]
kostasrim: very nice!

src/query/frontend/semantic/symbol_generator.cpp
```diff
@@ -248,67 +253,94 @@ bool SymbolGenerator::PostVisit(Unwind &unwind) {
...
     if (!HasSymbol(ident.name_) && !ConsumePredefinedIdentifier(ident.name_)) throw UnboundVariableError(ident.name_);
-    symbol = scope_.symbols[ident.name_];
+    if (HasSymbol(ident.name_) && !scope_.back().in_foreach) {
+      for (auto scope = scope_.rbegin(); scope != scope_.rend(); ++scope) {
+        if (scope->symbols.contains(ident.name_) &&
+            scope->in_foreach) {  // && !scope->symbols[ident.name_].in_foreach) {
+          throw UnboundVariableError(ident.name_);
+        }
+      }
+    }
```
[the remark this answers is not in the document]
kostasrim: I completely forgot about this. This is really not needed.

src/query/frontend/semantic/symbol_generator.cpp
```diff
@@ -248,67 +253,94 @@ bool SymbolGenerator::PostVisit(Unwind &unwind) {
...
+  const auto &name = for_each.named_expression_->name_;
+  scope_.push_back(Scope());
+  scope_.back().in_foreach = true;
+  if ((++scope_.rbegin())->in_foreach && HasSymbolInParentScope(name)) {
+    throw RedeclareVariableError(name);
   }
-  scope_.identifiers_in_match.clear();
+  for_each.named_expression_->MapTo(CreateSymbol(name, true));
+  return true;
+}
```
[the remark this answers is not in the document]
antaljanosbenjamin: We ended up allowing shadowing variables in `FOREACH`.

tests/unit/cypher_main_visitor.cpp
```diff
@@ -4088,3 +4088,89 @@ TEST_P(CypherMainVisitorTest, VersionQuery) {
   TestInvalidQuery("SHOW VERSIONS", ast_generator);
   ASSERT_NO_THROW(ast_generator.ParseQuery("SHOW VERSION"));
 }
+
+TEST_P(CypherMainVisitorTest, ForeachThrow) {
+  auto &ast_generator = *GetParam();
+  EXPECT_THROW(ast_generator.ParseQuery("FOREACH(i IN [1, 2] | UNWIND [1,2,3] AS j CREATE (n))"), SyntaxException);
+  EXPECT_THROW(ast_generator.ParseQuery("FOREACH(i IN [1, 2] CREATE (:Foo {prop : i}))"), SyntaxException);
+  EXPECT_THROW(ast_generator.ParseQuery("FOREACH(i IN [1, 2] | MATCH(n)"), SyntaxException);
+  EXPECT_THROW(ast_generator.ParseQuery("FOREACH(i IN x | MATCH(n)"), SyntaxException);
```
antaljanosbenjamin, line 4097:
```suggestion
  EXPECT_THROW(ast_generator.ParseQuery("FOREACH(i IN [1, 2] | MATCH (n)"), SyntaxException);
  EXPECT_THROW(ast_generator.ParseQuery("FOREACH(i IN x | MATCH (n)"), SyntaxException);
```

antaljanosbenjamin (changes requested):
Looking really nice. I commented on small things and plan to have another review tomorrow in the morning, to check SymbolGenerator/CypherMainVisitor again, to make sure I didn't miss anything.

src/query/plan/rewrite/index_lookup.hpp
```diff
@@ -433,6 +433,15 @@ class IndexLookupRewriter final : public HierarchicalLogicalOperatorVisitor {
     return true;
   }
 
+  bool PreVisit(Foreach &op) override {
+    prev_ops_.push_back(&op);
+    return false;
+  }
+  bool PostVisit(Foreach &) override {
```
antaljanosbenjamin, line 440:
```suggestion

  bool PostVisit(Foreach &) override {
```

antaljanosbenjamin (changes requested):

antaljanosbenjamin (changes requested):

kostasrim:
docs PR  memgraph/docs#355

antaljanosbenjamin (approved):

jbajic (approved):

PR #353, opened 2022-02-11
[E108-MG < T0810-MG] Add Websocket e2e tests

jbajic:
[master < Task] PR


src/communication/websocket/session.cpp
```diff
@@ -80,6 +80,7 @@ bool Session::Run() {
     return false;
   }
 
+  authentication_process_ = auth_.HasAnyUsers();
```
antonio2368, line 83: You can just use `authenticated_ = !auth_.HasAnyUsers()` here.
And then the function `IsAuthenticated` can just be `return authenticated_`.

jbajic: Aha that is true, it makes everything simpler

tests/e2e/websocket/common.hpp
```diff
@@ -0,0 +1,110 @@
...
+}
+
+void AssertLogMessage(const std::string &log_message) {
+  const auto json_message = nlohmann::json::parse(log_message);
+  if (json_message.contains("success")) {
+    spdlog::info("Received auth message: {}", json_message.dump());
+    AssertAuthMessage(json_message);
+    return;
+  }
+  MG_ASSERT(json_message.at("event").is_string(), "Event is not a string!");
```
antonio2368, line 106: You can assert that the value of event is `log`.

tests/e2e/websocket/websocket.cpp
```diff
@@ -0,0 +1,214 @@
...
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
+// License, and you may not use this file except in compliance with the Business Source License.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#include <gflags/gflags.h>
```
antonio2368, line 12: this goes in the second group

tests/e2e/websocket/websocket_ssl.cpp
```diff
@@ -0,0 +1,255 @@
...
+  const auto received_messages = websocket_client.GetReceivedMessages();
+  spdlog::info("Received {} messages.", received_messages.size());
+
+  MG_ASSERT(received_messages.size() < 2, "There is no more than one received message!");
+  if (!received_messages.empty()) {
+    auto json_message = nlohmann::json::parse(received_messages[0]);
+    AssertAuthMessage(json_message, false);
+  }
+  spdlog::info("Finishing websocket connection with users but without being authenticated.");
+}
```
antonio2368, line 239: These are the same tests, maybe extract it and send the client as the argument.

antonio2368 (changes requested):

tests/e2e/websocket/websocket.cpp
```diff
@@ -0,0 +1,214 @@
...
+  explicit Session(net::io_context &ioc, std::vector<std::string> &expected_messages)
+      : resolver_(net::make_strand(ioc)), ws_(net::make_strand(ioc)), received_messages_{expected_messages} {}
+
+  explicit Session(net::io_context &ioc, std::vector<std::string> &expected_messages, Credentials creds)
+      : resolver_(net::make_strand(ioc)),
+        ws_(net::make_strand(ioc)),
+        received_messages_{expected_messages},
+        creds_{creds} {}
+
+  void Run(std::string host, std::string port) {
```
antaljanosbenjamin, line 39:
```suggestion
  void Run(const std::string& host, const std::string& port) {
```

tests/e2e/websocket/websocket.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+    ws_.async_handshake(host_, "/", beast::bind_front_handler(&Session::OnHandshake, shared_from_this()));
+  }
+
+  void OnHandshake(beast::error_code ec) {
+    if (ec) {
+      return Fail(ec, "handshake");
+    }
+    ws_.async_write(net::buffer(GetAuthenticationJSON(creds_)),
+                    beast::bind_front_handler(&Session::OnWrite, shared_from_this()));
```
antaljanosbenjamin, line 78: What happens if `creds_` is empty?

tests/e2e/websocket/websocket.cpp
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  void Close() { ioc_.stop(); }
+
+  std::vector<std::string> GetReceivedMessages() { return received_messages_; }
+
+ private:
+  std::vector<std::string> received_messages_{};
+  net::io_context ioc_{};
+  std::thread bg_thread_;
```
antaljanosbenjamin, line 135:
```suggestion
  std::jthread bg_thread_;
```

Just to make sure we join the thread on destruction.

tests/e2e/websocket/websocket.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+  websocket_client.Close();
+  const auto received_messages = websocket_client.GetReceivedMessages();
+  spdlog::info("Received {} messages.", received_messages.size());
+
+  MG_ASSERT(received_messages.size() < 2, "There is no more than one received message!");
+  if (!received_messages.empty()) {
+    auto json_message = nlohmann::json::parse(received_messages[0]);
+    AssertAuthMessage(json_message, false);
+  }
```
antaljanosbenjamin: Why not expecting exactly one message? If the authentication fails, we expect the server to send a failed message, don't we?

tests/e2e/websocket/websocket_ssl.cpp
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    beast::get_lowest_layer(ws_).expires_never();
+    ws_.set_option(websocket::stream_base::timeout::suggested(beast::role_type::client));
+    ws_.set_option(websocket::stream_base::decorator([](websocket::request_type &req) {
+      req.set(http::field::user_agent, std::string(BOOST_BEAST_VERSION_STRING) + " websocket-client-async-ssl");
+    }));
+
+    ws_.async_handshake(host_, "/", beast::bind_front_handler(&Session::OnHandshake, shared_from_this()));
+  }
```
antaljanosbenjamin, line 91: Maybe consider executing this conditionally and in this way the two clients might be merged into one.

tests/e2e/websocket/workloads.yaml
```diff
@@ -0,0 +1,32 @@
...
+      ssl: true
+
+workloads:
+  - name: "Websocket"
+    binary: "tests/e2e/websocket/memgraph__e2e__websocket"
+    args: ["--bolt-port", "7687"]
+    <<: *template_cluster
+  - name: "Websocket SSL"
+    binary: "tests/e2e/websocket/memgraph__e2e__websocket_ssl"
+    args: ["--bolt-port", "7686"]
```
antaljanosbenjamin, line 31: As we discussed, the same port can be used in both tests.

src/communication/websocket/auth.cpp
```diff
@@ -16,11 +16,12 @@
 namespace communication::websocket {
 
 bool SafeAuth::Authenticate(const std::string &username, const std::string &password) const {
+  // TODO: Make ReadLock after dealing with Authenticate
```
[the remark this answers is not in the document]
jbajic: Yes that what I meant, will create an issue then

tests/e2e/memgraph.py
```diff
@@ -70,18 +75,21 @@ def start(self, restart=False, args=[]):
             return
         self.stop()
         self.args = copy.deepcopy(args)
+        self.args = [replace_paths(arg) for arg in self.args if isinstance(arg, str)]
```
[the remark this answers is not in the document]
jbajic: You are right, I missed that

tests/e2e/websocket/websocket.cpp
jbajic: JSON will be empty and authentication will fail if there are any users, will use optional

tests/e2e/websocket/common.hpp
```diff
@@ -0,0 +1,309 @@
...
+constexpr std::array kSupportedLevels{"debug", "trace", "info", "warning", "error", "critical"};
+
+struct Credentials {
+  std::string_view username;
+  std::string_view passsword;
+};
+
+inline void Fail(beast::error_code ec, char const *what) { std::cerr << what << ": " << ec.message() << "\n"; }
+
+std::string GetAuthenticationJSON(const Credentials &creds) {
```
antonio2368, line 53: inline

tests/e2e/websocket/common.hpp
```diff
@@ -0,0 +1,309 @@
...
+#include "utils/logging.hpp"
+
+namespace beast = boost::beast;
+namespace http = beast::http;
+namespace websocket = beast::websocket;
+namespace net = boost::asio;
+using tcp = boost::asio::ip::tcp;
+namespace ssl = boost::asio::ssl;
+
+constexpr std::array kSupportedLevels{"debug", "trace", "info", "warning", "error", "critical"};
```
antonio2368, line 44:
```suggestion
constexpr std::array kSupportedLogLevels{"debug", "trace", "info", "warning", "error", "critical"};
```

tests/e2e/websocket/common.hpp
```diff
@@ -0,0 +1,309 @@
...
+      : resolver_(net::make_strand(ioc)),
+        ws_(net::make_strand(ioc), ctx),
+        received_messages_{expected_messages},
+        creds_{creds} {}
+
+  explicit Session(net::io_context &ioc, std::vector<std::string> &expected_messages, Credentials creds) requires(!ssl)
+      : resolver_(net::make_strand(ioc)),
+        ws_(net::make_strand(ioc)),
+        received_messages_{expected_messages},
+        creds_{creds} {}
```
antonio2368, line 82: These 2 constructors can be replaced by a single one using Parameter pack: https://en.cppreference.com/w/cpp/language/parameter_pack

I think it's a fun exercise to try to write it yourself. :smile:

antonio2368 (changes requested):

tests/e2e/websocket/common.hpp
jbajic: Whoops, I didn't see that one, sorry.

tests/e2e/websocket/websocket_ssl.cpp
```diff
@@ -0,0 +1,74 @@
...
+#include <unistd.h>
+
+#include "common.hpp"
+#include "utils/logging.hpp"
+
+DEFINE_uint64(bolt_port, 7687, "Bolt port");
+
+class WebsocketSSLClient {
+ public:
+  WebsocketSSLClient() { session_ = {std::make_shared<Session<true>>(ioc_, ctx_, received_messages_)}; }
```
antonio2368, line 36:
```suggestion
  WebsocketSSLClient() { session_ = std::make_shared<Session<true>>(ioc_, ctx_, received_messages_); }
```
What is going on here with the brackets?

tests/e2e/runner.py
```diff
@@ -36,18 +37,17 @@ def load_args():
 
 def load_workloads(root_directory):
     workloads = []
-    for file in Path(root_directory).rglob('*.yaml'):
+    for file in Path(root_directory).rglob("*.yaml"):
```
kostasrim, line 40: @jbajic you mentioned yesterday that `.yaml` is the wrong extension. Shall we create a task for it that replaces the extension with the proper one?

tests/e2e/websocket/common.hpp
```diff
@@ -0,0 +1,309 @@
...
+
+  void Run(std::string host, std::string port) {
+    host_ = host;
+    resolver_.async_resolve(host, port, beast::bind_front_handler(&Session::OnResolve, shared_from_this()));
```
kostasrim, line 86:
```suggestion
    resolver_.async_resolve(std::move(host), std::move(port), beast::bind_front_handler(&Session::OnResolve, shared_from_this()));
```

tests/e2e/websocket/common.hpp
```diff
@@ -0,0 +1,309 @@
...
+}
+
+template <typename TWebsocketClient>
+void TestWebsocketWithoutAnyUsers(std::unique_ptr<mg::Client> &mg_client) {
+  spdlog::info("Starting websocket connection without any users.");
+  auto websocket_client = TWebsocketClient();
+  websocket_client.Connect("127.0.0.1", "7444");
+
+  RunQueries(mg_client);
+  std::this_thread::sleep_for(std::chrono::seconds(1));
```
kostasrim, line 255: Why  sleep  here?

tests/e2e/websocket/common.hpp
```diff
@@ -0,0 +1,309 @@
...
+
+  websocket_client.Close();
+  const auto received_messages = websocket_client.GetReceivedMessages();
+  spdlog::info("Received {} messages.", received_messages.size());
+  MG_ASSERT(!received_messages.empty(), "There are no received messages!");
+  for (const auto &log_message : received_messages) {
```
kostasrim, line 261: ranges for-each  ? :D  :D

tests/e2e/websocket/websocket.cpp
```diff
@@ -0,0 +1,66 @@
...
+    bg_thread_.detach();
+  }
+
+  void Close() { ioc_.stop(); }
+
+  std::vector<std::string> GetReceivedMessages() { return received_messages_; }
+
+ private:
+  std::vector<std::string> received_messages_{};
+  net::io_context ioc_{};
```
kostasrim, line 47:
```suggestion
  net::io_context ioc_;
```

tests/e2e/runner.py
jbajic: I think its fine the way it is, it is not a big deal :smile:

tests/e2e/websocket/common.hpp
antonio2368: delegate is also what I had in mind actually (delegate + pack) but @jbajic will figure it out :smile:

tests/e2e/websocket/common.hpp
jbajic: So we give some time to the client to receive messages

tests/e2e/websocket/common.hpp
```diff
@@ -0,0 +1,302 @@
...
+  explicit Session(net::io_context &ioc, ssl::context &ctx, std::vector<std::string> &expected_messages) requires(ssl)
+      : resolver_(net::make_strand(ioc)), ws_(net::make_strand(ioc), ctx), received_messages_{expected_messages} {}
+
+  explicit Session(net::io_context &ioc, std::vector<std::string> &expected_messages) requires(!ssl)
+      : resolver_(net::make_strand(ioc)), ws_(net::make_strand(ioc)), received_messages_{expected_messages} {}
+
+  template <typename... Args>
+  explicit Session(Credentials creds, Args &&...args) : Session<ssl>(std::forward<Args>(args)...) {
+    creds_.emplace(creds);
+  }
```
antonio2368, line 74:
```suggestion
  explicit Session(Credentials creds, Args &&...args) : Session<ssl>(std::forward<Args>(args)...), creds_(creds) {}
```

tests/e2e/websocket/common.hpp
```diff
@@ -0,0 +1,302 @@
...
+
+  void Run(std::string host, std::string port) {
```
antonio2368, line 76:
```suggestion
  void Run(std::string_view host, std::string_view port) {
```

tests/e2e/websocket/common.hpp
```diff
@@ -0,0 +1,302 @@
...
+namespace websocket = beast::websocket;
+namespace net = boost::asio;
+using tcp = boost::asio::ip::tcp;
+namespace ssl = boost::asio::ssl;
+
+constexpr std::array kSupportedLogLevels{"debug", "trace", "info", "warning", "error", "critical"};
+
+struct Credentials {
+  std::string_view username;
+  std::string_view passsword;
```
antonio2368, line 48: `string_view` works for this example because you use string literals for all your use cases but remember that it can be dangerous as it doesn't increase the lifetime of the `std::string` object.
E.g. if you had something like
```
int i = 1;
auto user = fmt::format("user{}", i);
Credenntials  creds{user, "password");
```
The lifetime of `username` inside the `creds` depends on the `user` variable.

tests/e2e/websocket/websocket.cpp
antonio2368: Why didn't you use `jthread` here but you use it in the SSL version.

antonio2368: I would leave the `jthread` so it's closed correctly if `AwaitClose` wasn't called (safer).
Also, `AwaitClose` will allow you to stop the processing manually without relying on RAII so it doesn't lose its functionality.

PR #355, opened 2022-02-14
[E108-MG < T0811-MG] Add websocket port argument

jbajic:
[master < Task] PR


src/memgraph.cpp
```diff
@@ -134,6 +134,10 @@ std::optional<Enum> StringToEnum(const auto &value, const auto &mappings) {
 DEFINE_string(bolt_address, "0.0.0.0", "IP address on which the Bolt server should listen.");
 DEFINE_VALIDATED_int32(bolt_port, 7687, "Port on which the Bolt server should listen.",
                        FLAG_IN_RANGE(0, std::numeric_limits<uint16_t>::max()));
+DEFINE_VALIDATED_int32(websocket_port, 7444,
+                       "Port for websocket connection on which "
+                       "logs will be served.",
+                       FLAG_IN_RANGE(0, std::numeric_limits<uint16_t>::max()));
```
antaljanosbenjamin, line 140: I think Antonio [phrased the docs](https://github.com/memgraph/docs/pull/307) in a way that websocket is not specific to logs, but in general websocket is used to provide "certain" information about Memgraph. Therefore I think here we should also not constrain the websockets only for logs.

```suggestion
DEFINE_VALIDATED_int32(websocket_port, 7444, "Port on which the WebSocket server should listen.",
                       FLAG_IN_RANGE(0, std::numeric_limits<uint16_t>::max()));
```

src/memgraph.cpp
```diff
@@ -134,6 +134,10 @@ std::optional<Enum> StringToEnum(const auto &value, const auto &mappings) {
 DEFINE_string(bolt_address, "0.0.0.0", "IP address on which the Bolt server should listen.");
```
antaljanosbenjamin, line 134: Should we also introduce a `websocket-adress` flag?

antaljanosbenjamin (approved):

antonio2368 (approved):

PR #356, opened 2022-02-16
[master < T0813-MG] Add docker release action

jbajic:
[master < Task] PR

This PR also introduces a naming change to our packages as follows:
- For DEB packages: amd64/arm64 or all (according to https://www.debian.org/doc/debian-policy/ch-controlfields.html#architecture and https://wiki.debian.org/SupportedArchitectures)
- For RPM packages: x84_64/aarch64 or  noarch (according to https://docs.fedoraproject.org/en-US/Fedora_Draft_Documentation/0.1/html/RPM_Guide/ch01s03.html)


release/docker/memgraph_deb.dockerfile
```diff
@@ -1,7 +1,8 @@
 FROM debian:bullseye
 # NOTE: If you change the base distro update release/package as well.
 
-ARG release
+ARG BINARY_NAME
+ARG TARGETARCH
```
antonio2368, line 5: shouldn't the `package_docker` script be changed then?

.github/workflows/release_docker.yaml
```diff
@@ -0,0 +1,65 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      memgraph_arm_binary_url:
+        description: "A URL to download memgraph arm binary you wish to release."
+        required: true
+      memgraph_amd_binary_url:
+        description: "A URL to download memgraph amd binary you wish to release."
+        required: true
```
jbajic, line 11: Not needed since we know the URL for binaries

init
```diff
@@ -65,8 +65,14 @@ else
 fi
 
 DISTRO=$(operating_system)
-echo "ALL BUILD PACKAGES: $($DIR/environment/os/$DISTRO.sh list MEMGRAPH_BUILD_DEPS)"
-$DIR/environment/os/$DISTRO.sh check MEMGRAPH_BUILD_DEPS
+ARCHITECTURE=$(architecture)
+if [ "${ARCHITECTURE}" = "arm64" ]; then
+  echo "ALL BUILD PACKAGES: $($DIR/environment/os/$DISTRO-arm.sh list MEMGRAPH_BUILD_DEPS)"
+  $DIR/environment/os/$DISTRO.sh check MEMGRAPH_BUILD_DEPS
```
antaljanosbenjamin, line 71:
```suggestion
  $DIR/environment/os/$DISTRO-arm.sh check MEMGRAPH_BUILD_DEPS
```

Shouldn't this be the correct script?

init
```diff
@@ -65,8 +65,14 @@ else
...
+else
+  echo "ALL BUILD PACKAGES: $($DIR/environment/os/$DISTRO.sh list MEMGRAPH_BUILD_DEPS)"
+  $DIR/environment/os/$DISTRO.sh check MEMGRAPH_BUILD_DEPS
+fi
```
antaljanosbenjamin, line 75: What about extracting the path to the correct script into a variable and using it? I think the arm branch has a bug in the checking, because it calls `$DISTRO.sh` instead of `$DISTRO-arm.sh`.

release/docker/memgraph_deb.dockerfile
```diff
@@ -1,19 +1,20 @@
...
   && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 RUN pip3 install networkx==2.4 numpy==1.21.4 scipy==1.7.3
 
-COPY ${release} /
+COPY "${BINARY_NAME}${TARGETARCH}.deb" /
 
 # Install memgraph package
-RUN dpkg -i ${release}
+RUN dpkg -i "${BINARY_NAME}${TARGETARCH}.deb"
```
antaljanosbenjamin, line 17: First of all, I think `package_docker` and this is not compatible, because `package_docker` passes `${package_name}.${extension}` as binary name, so I think `${BINARY_NAME}${TARGETARCH}.deb` will be expanded to `<memgraph-package-name>.deb<targetarch>.deb`, which is not good.

Is there any reason, why we cannot pass the whole package name as before? I think `TARGET_ARCH` is only used to figure out the installer package.

environment/os/debian-11-arm.sh
```diff
@@ -0,0 +1,104 @@
...
+                apt-get install -y apt-transport-https dotnet-sdk-3.1
+            fi
+            continue
+        fi
+        apt install -y "$pkg"
+    done
+}
+
+deps=$2"[*]"
+"$1" "${!deps}"
```
antaljanosbenjamin, line 104:
```suggestion
"$1" "${!deps}"

```

antaljanosbenjamin (changes requested):

init
jbajic: You are right will do that

release/docker/memgraph_deb.dockerfile
jbajic: The issue is that `TARGETARCH` is automatically passed by buildx. I've tried running separate build with different arguments but then the images get overriden on dockerhub, and the only way to differentiate architecture is in dockerfile.

jbajic: I've tested the change script and it should work now

antaljanosbenjamin: Ohh, I see. What about unifying all the package names to contain the architecture? So instead of `memgraph.deb` it has to be either `memgraph-amd64.deb` or `memgraph-arm64.deb` (or aarch64, not sure which is the correct).

I know this can be "abused" to work with both `memgraph.deb` and `memgraph-arm64.deb`, but I don't like the concept to use the script in different ways.

antaljanosbenjamin: After the changes I think the current shape is not bad. Changing all the binary names can be a complex task, so let this be like that now.

release/docker/package_docker
```diff
@@ -55,7 +55,10 @@ image_name="memgraph:${version}"
 image_package_name="memgraph-${version}-docker.tar.gz"
 
 # Build docker image.
-docker build -t ${image_name} ${tag_latest} -f ${dockerfile_path} --build-arg release=${package_name}.${extension} .
+docker build -t ${image_name} ${tag_latest} -f ${dockerfile_path} \
+--build-arg BINARY_NAME=${package_name} \
+--build-arg EXTENSION=${extension} \
+--build-arg TARGETARCH="".
```
antaljanosbenjamin, line 61: After consistently using the architecture related postfix in CMake, I am suspicious about the empty `TARGETARCH` parameter here. I think the name of package will be the default one coming from CMake. If I am wrong, could you please describe how that this works?

antaljanosbenjamin: Furthermore this script is called from two places:
1. release_docker.yaml which download the binary packages from our website.
2. from `release/package/run.sh` which uses a freshly built binary package that is named by CMake.

Wouldn't that cause an issue?

jbajic: All that you said is valid but in this script during extraction of `package_name` and `extension`, the architecture is a part of binary name => `package_name`, therefore we pass an empty string to ignore it, but in the docker case the architecture is passed by the docker build extension `build` and that how we decide internally inside Dockerfile for which architecture we are building. E.g. running this script like this:
` ./package_docker --latest /home/bajic/Downloads/memgraph_2.2.1-1_amd64.deb`
`package_name` = `memgraph_2.2.1-1_amd64`
`extension` = `deb`

jbajic: The test run can be seen here https://github.com/memgraph/memgraph/actions/runs/2076704439

antaljanosbenjamin (approved):

kostasrim (approved):

PR #360, opened 2022-02-23
[master < T0817-MG] Small io network socket fixes

sinisa-susnjar:
[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR


src/io/network/addrinfo.hpp
```diff
@@ -11,21 +11,39 @@
...
+
   ~AddrInfo();
 
-  static AddrInfo Get(const char *addr, const char *port);
+  AddrInfo(const std::string &addr, uint16_t port);
+  explicit AddrInfo(const Endpoint &endpoint);
 
-  operator struct addrinfo *();
+  auto begin() const { return Iterator(info); }
+  auto end() const { return Iterator(nullptr); }
```
kostasrim, line 46: Once you exhaust  `ptr->as_next`, is it guranteed by the implementation that it will point to `nulltpr`? Also, isn't  the case that `end` should always return the address after the last element? I am not sure what the C++ standard mandates but if  it does, then we should follow that.

src/io/network/addrinfo.hpp
```diff
@@ -11,21 +11,39 @@
...
-
  public:
+  struct Iterator {
+    explicit Iterator(addrinfo *p) : ptr(p) {}
+    addrinfo &operator*() const { return *ptr; }
+    Iterator &operator++() {
+      ptr = ptr->ai_next;
+      return *this;
+    }
+    bool operator!=(const Iterator &rhs) { return rhs.ptr != ptr; };
```
kostasrim, line 34: This break `operator symmetry` meaning that for two iterators `a && b` `a != b` is not the same as `b != a` because  the lhs is subject to different conversion rules than the right (by C++ standard). Define it an `inline friend` :)

kostasrim (changes requested):
A few first comments! 

P.s. I like the initiative of this and congratz for your first PR 🎉

src/io/network/addrinfo.hpp
antaljanosbenjamin: I couldn't find the exact change in the standard, but the spaceship operator makes "reversed ordering" of comparison operators [possible](https://godbolt.org/z/vhn9Txjsd).

.gitignore
```diff
@@ -66,3 +66,6 @@ src/raft/storage_info_rpc_messages.hpp
 src/stats/stats_rpc_messages.hpp
 src/storage/distributed/rpc/concurrent_id_mapper_rpc_messages.hpp
 src/transactions/distributed/engine_rpc_messages.hpp
+
+# Clangd cache
+.cache
```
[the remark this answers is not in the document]
antaljanosbenjamin:
```
echo -e ".cache/\n" > ~/.gitignore_global
git config --global core.excludesFile ~/.gitignore_global
```

src/io/network/addrinfo.cpp
```diff
@@ -10,33 +10,25 @@
...
-  auto status = getaddrinfo(addr, port, &hints, &result);
+AddrInfo::AddrInfo(const Endpoint &endpoint) : AddrInfo(endpoint.address, endpoint.port) {}
 
+AddrInfo::AddrInfo(const std::string &addr, uint16_t port) {
+  addrinfo hints{
+      .ai_flags = AI_PASSIVE,
+      .ai_family = AF_UNSPEC,     // IPv4 and IPv6
+      .ai_socktype = SOCK_STREAM  // TCP socket
+  };
+  auto status = getaddrinfo(addr.c_str(), std::to_string(port).c_str(), &hints, &info);
```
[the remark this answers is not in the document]
antaljanosbenjamin: Are you sure? Based on [cppreference](https://en.cppreference.com/w/cpp/language/lifetime):

> All temporary objects are destroyed as the last step in evaluating the full-expression that (lexically) contains the point where they were created, and if multiple temporary objects were created, they are destroyed in the order opposite to the order of creation. This is true even if that evaluation ends in throwing an exception.

So I think the temporary object won't be destroyed until the function returns.

src/io/network/socket.hpp
```diff
@@ -42,7 +42,7 @@ class Socket {
   /**
    * Shutdown the socket if it is open.
    */
-  void Shutdown();
+  void Shutdown() const;
```
antaljanosbenjamin, line 45: Even though technically these functions could be const, in my opinion they shouldn't. The class represents a socket object (by wrapping some pointers), and these functions are modifying the abstract state of the object. Currently these function can only be const, because they are using a C API underneath, which doesn't enforce const correctness.

There is another reason why not to do: when we are using some kind of read-write locking, then a read lock operation can return a const object, making sure the object doesn't get modified. Because you can have multiple read locks, all of the const functions should be thread-safe from the outside (not modifying the state of the object or using some synchronization internally). Everything that is not thread-safe, shouldn't be const.

What do you think about it? What was your idea, why to make these functions `const`?

src/io/network/socket.cpp
```diff
@@ -135,37 +113,34 @@ bool Socket::Bind(const Endpoint &endpoint) {
   return true;
 }
 
-void Socket::SetNonBlocking() {
-  int flags = fcntl(socket_, F_GETFL, 0);
+void Socket::SetNonBlocking() const {
+  int flags = fcntl(socket_, F_GETFL);
```
antaljanosbenjamin, line 117:
```suggestion
  const int flags = fcntl(socket_, F_GETFL);
```

src/io/network/addrinfo.hpp
antaljanosbenjamin: I guess in a linked list you have to set `next` to `nullptr` to mark the end of the list.

I think the standard doesn't require an exact address that is after the last element, but more likely it meant in a way that the end operator should be exclusive.

src/io/network/addrinfo.hpp
```diff
@@ -11,21 +11,39 @@
 
 #pragma once
 
+#include <netdb.h>
+#include <string>
+
 namespace io::network {
 
+struct Endpoint;
```
antaljanosbenjamin, line 19: That means we start to use  forward declarations! FYI: it is not common in the codebase, but I am totally up for it.

antaljanosbenjamin (changes requested):

src/io/network/addrinfo.hpp
kostasrim: PLz don't unless required :D

src/io/network/addrinfo.cpp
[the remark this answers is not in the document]
kostasrim: You are right Benjo! You learn every day ;)

src/io/network/addrinfo.hpp
kostasrim: Yes but it's still not `symmetrical` unless defined as friend. Simply put the expression 
`a == b` is transformed to `a.operator==(b)` which  is not symmetrical to `b.operartor==(a)`. However if you make it a function then both `a == b` and `b == a` are the exact same thing from the standards point of view

src/io/network/socket.cpp
```diff
@@ -9,38 +9,23 @@
...
 
-Socket::Socket(Socket &&other) {
-  socket_ = other.socket_;
-  endpoint_ = std::move(other.endpoint_);
+Socket::Socket(Socket &&other) noexcept : socket_(other.socket_), endpoint_(std::move(other.endpoint_)) {
   other.socket_ = -1;
 }
 
-Socket &Socket::operator=(Socket &&other) {
+Socket &Socket::operator=(Socket &&other) noexcept {
```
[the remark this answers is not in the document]
sinisa-susnjar: Yes! OMG how did I not see this lol. It never blew up because the op is actually never used :sweat_smile:

src/io/network/addrinfo.hpp
sinisa-susnjar: Yes, getaddrinfo() returns the info as a single linked list, using `ai_next` as the pointer to the next address element in the list or `null` in case there are no more list entries, thus the `nullptr` in `end()`.

src/io/network/addrinfo.hpp
```diff
@@ -11,21 +11,39 @@
...
+
 /**
  * Wrapper class for getaddrinfo.
  * see: man 3 getaddrinfo
  */
 class AddrInfo {
-  explicit AddrInfo(struct addrinfo *info);
-
  public:
+  struct Iterator {
```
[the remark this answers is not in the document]
sinisa-susnjar: Hm, yes, I'm aware - I actually wanted to keep the code as small as possible by only implementing the bare minimum to make range based for working, but I can add the full forward iterator interface.

src/io/network/addrinfo.hpp
sinisa-susnjar: Forward declarations reduce compile time, but if it's not common or you guys don't like it I can use `#include` instead :smile:

src/io/network/addrinfo.hpp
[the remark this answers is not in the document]
sinisa-susnjar: Thinking of it, only an input iterator is required.

src/io/network/addrinfo.hpp
sinisa-susnjar: cool, I have made the op it a friend function

src/io/network/addrinfo.hpp
sinisa-susnjar: Hm, interesting... clang-format reorders the include files in the cpp file, which in turn can cause problems with the forward declared identifiers, i.e. if the include file that contains the forward declaration is included before the actual declaration.

src/io/network/addrinfo.hpp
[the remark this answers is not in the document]
kostasrim: Two things here. I could be fine for "just making it work" for something more complex (like using runtime polymorphism vs static -- because the  second requires CRTP and a little bit more templated work for a small gain in performance for most of the time). However, for this case we opted in to make AddrInfo iterable which means that we either do that and implement the whole interface (since we opted in for that) or not at all. Half baked glue code should be avoided. 

For the input iterator, I am not sure, I quote from cppreference on LegacyInputIterator:
>once a LegacyInputIterator i has been incremented, all copies of its previous value may be invalidated.
Is that the case for `AddrInfo` ? If the answer is yes, then `InputIterator` suffices for this  case.

src/io/network/addrinfo.hpp
kostasrim: @sinisa-susnjar Something very very interesting! So take a look here: https://godbolt.org/z/qbYsfrz7a 
It turns out that C++ 20 added reversed operators for comparisons so the above code is actually symmetrical (remove the C++20 standard from godbolt and you should get an error). 
p.s.  special thanks for discussing this with @antaljanosbenjamin my mind is bend! 
p.s. 2. My comment still applies for ALL other operators, so keep that in mind!

src/io/network/addrinfo.hpp
kostasrim: I would need a benchmark for the argument  "Forward decls reduce compile time". There are so much more to optimize out at compile time that I hardly believe this would make a difference :D

antaljanosbenjamin: In 99% percent of the time (assuming forward declarations are used consistently), forward declarations are faster. Check [this SO answer](https://stackoverflow.com/a/61901220).

kostasrim: > It generates 200 headers. Each header has a struct with 100 fields and a comment 5000 bytes long.
I am not convinced.

How does this represent a codebase? What are the dependencies between the headers? Which file includes what? How similar is our codebase to this example (basically how's our dependency DAG vs the one in the example)? This is an overly generalized benchmark/example that does not have any meaningful information or relation to an actual codebase. 

Moreover, there are two nice answers above, and one hint about why forward decls decrease compilation time (the source is John Lakos book). However, that book is old and compilers have changed significantly since then which makes me question if that's still the case. 

All in all, unless forward decls provably solve your builds compile time for a given codebase (that is you benchmarked and you actually proven that the drag on the builds is not from SFINAE or other compile intensive computations), I would stick away from that as I don't think they make your code more readable (quite the opposite). And even then. there are other solutions, precompiled headers, and C++20 modules :P

src/io/network/addrinfo.cpp
```diff
@@ -9,34 +9,49 @@
...
-
 #include "io/network/addrinfo.hpp"
 
 #include "io/network/network_error.hpp"
+#include "utils/iterator_concepts.hpp"
 
 namespace memgraph::io::network {
 
-AddrInfo::AddrInfo(struct addrinfo *info) : info(info) {}
+static_assert(utils::LegacyForwardIterator<AddrInfo::Iterator>);
```
antaljanosbenjamin, line 19: Iterator concepts FTW!

antaljanosbenjamin (commented):
As I took over the PR, I cannot review it.

jbajic (approved):
Wow on the concepts stuff! :+1:

src/io/network/addrinfo.cpp
```diff
@@ -9,34 +9,50 @@
...
```
kostasrim, line 19: Let's slightly change this so we can get rid of `iterator_concepts.hpp` all alone. 
1. delete `iterator_concepts.hpp` 
2. Use this assert (see explanation below)
```
static_assert(std::input_iterator<AddrInfo::Iterator> && std::equality_comparable<AddrInfo::Iterator>);
```

`std::input_iterator` is a slightly different concept, i.e., it has slightly different requirements than `LegacyInputIterator`. The most important is that `ranges` work with sentinels to denote the end of the range and these are not part of the iterator interface but rather added as extra requirements to the ranges algorithms. Hence, `std::equality_comparable` is required in the assertion to mimic the requirements of `LegacyInputIterator`.

Note, that `std::input_iterator` has also stronger type requirements that `LegacyInputIterator` (see cpp ref) but are satisfied and not an insue in our case :)

kostasrim (approved):

PR #371, opened 2022-03-21
[E-109-MG < T0785-MG] Use asio socket

jbajic:
[master < Task] PR

### DIFF
--- a/src/communication/context.cpp
+++ b/src/communication/context.cpp
@@ -78,17 +78,18 @@ bool ClientContext::use_ssl() { return use_ssl_; }
 
 ServerContext::ServerContext(const std::string &key_file, const std::string &cert_file, const std::string &ca_file,
                              bool verify_peer) {
-  ctx_.emplace(boost::asio::ssl::context::tls_server);
+  namespace ssl = boost::asio::ssl;
+  ctx_.emplace(ssl::context::tls_server);
   // NOLINTNEXTLINE(hicpp-signed-bitwise)
-  ctx_->set_options(boost::asio::ssl::context::default_workarounds | boost::asio::ssl::context::no_sslv2 |
-                    boost::asio::ssl::context::no_sslv3 | boost::asio::ssl::context::single_dh_use);
+  ctx_->set_options(ssl::context::default_workarounds | ssl::context::no_sslv2 | ssl::context::no_sslv3 |
+                    ssl::context::single_dh_use);
   ctx_->set_default_verify_paths();
   // TODO: add support for encrypted private keys
   // TODO: add certificate revocation list (CRL)
   boost::system::error_code ec;
   ctx_->use_certificate_chain_file(cert_file, ec);
   MG_ASSERT(!ec, "Couldn't load server certificate from file: {}", cert_file);
-  ctx_->use_private_key_file(key_file, boost::asio::ssl::context::pem, ec);
+  ctx_->use_private_key_file(key_file, ssl::context::pem, ec);
   MG_ASSERT(!ec, "Couldn't load server private key from file: {}", key_file);
 
   ctx_->set_options(SSL_OP_NO_SSLv3, ec);
@@ -103,7 +104,7 @@ ServerContext::ServerContext(const std::string &key_file, const std::string &cer
     if (verify_peer) {
       // Enable verification of the client certificate.
       // NOLINTNEXTLINE(hicpp-signed-bitwise)
-      ctx_->set_verify_mode(boost::asio::ssl::verify_peer | boost::asio::ssl::verify_fail_if_no_peer_cert, ec);
+      ctx_->set_verify_mode(ssl::verify_peer | ssl::verify_fail_if_no_peer_cert, ec);
       MG_ASSERT(!ec, "Setting SSL verification mode failed!");
     }
   }
--- a/src/communication/context.cpp
+++ b/src/communication/context.cpp
@@ -79,6 +79,7 @@ bool ClientContext::use_ssl() { return use_ssl_; }
 ServerContext::ServerContext(const std::string &key_file, const std::string &cert_file, const std::string &ca_file,
                              bool verify_peer) {
   ctx_.emplace(boost::asio::ssl::context::sslv23_server);
+  // NOLINTNEXTLINE(hicpp-signed-bitwise)
   ctx_->set_options(boost::asio::ssl::context::default_workarounds | boost::asio::ssl::context::no_sslv2 |
                     boost::asio::ssl::context::single_dh_use);
   ctx_->set_default_verify_paths();
--- a/src/communication/context.cpp
+++ b/src/communication/context.cpp
@@ -78,7 +78,7 @@ bool ClientContext::use_ssl() { return use_ssl_; }
 
 ServerContext::ServerContext(const std::string &key_file, const std::string &cert_file, const std::string &ca_file,
                              bool verify_peer) {
-  ctx_.emplace(boost::asio::ssl::context::tls_server);
+  ctx_.emplace(boost::asio::ssl::context::sslv23_server);
   ctx_->set_default_verify_paths();
   // TODO: add support for encrypted private keys
   // TODO: add certificate revocation list (CRL)
--- a/src/communication/context.cpp
+++ b/src/communication/context.cpp
@@ -79,6 +79,8 @@ bool ClientContext::use_ssl() { return use_ssl_; }
 ServerContext::ServerContext(const std::string &key_file, const std::string &cert_file, const std::string &ca_file,
                              bool verify_peer) {
   ctx_.emplace(boost::asio::ssl::context::sslv23_server);
+  ctx_->set_options(boost::asio::ssl::context::default_workarounds | boost::asio::ssl::context::no_sslv2 |
+                    boost::asio::ssl::context::single_dh_use);
   ctx_->set_default_verify_paths();
   // TODO: add support for encrypted private keys
   // TODO: add certificate revocation list (CRL)
--- a/src/communication/context.cpp
+++ b/src/communication/context.cpp
@@ -78,10 +78,10 @@ bool ClientContext::use_ssl() { return use_ssl_; }
 
 ServerContext::ServerContext(const std::string &key_file, const std::string &cert_file, const std::string &ca_file,
                              bool verify_peer) {
-  ctx_.emplace(boost::asio::ssl::context::sslv23_server);
+  ctx_.emplace(boost::asio::ssl::context::tls_server);
   // NOLINTNEXTLINE(hicpp-signed-bitwise)
   ctx_->set_options(boost::asio::ssl::context::default_workarounds | boost::asio::ssl::context::no_sslv2 |
-                    boost::asio::ssl::context::single_dh_use);
+                    boost::asio::ssl::context::no_sslv3 | boost::asio::ssl::context::single_dh_use);
   ctx_->set_default_verify_paths();
   // TODO: add support for encrypted private keys
   // TODO: add certificate revocation list (CRL)
--- a/src/communication/v2/listener.hpp
+++ b/src/communication/v2/listener.hpp
@@ -54,12 +54,12 @@ class Listener final : public std::enable_shared_from_this<Listener<TSession, TS
   void Start() { DoAccept(); }
 
  private:
-  Listener(IOContextPool &io_context_pool, TSessionData *data, ServerContext *server_context, tcp::endpoint &endpoint,
-           const std::string_view service_name, const int inactivity_timeout_sec)
-      : io_context_pool_(io_context_pool),
+  Listener(IOContextThreadPool &io_thread_context_pool, TSessionData *data, ServerContext *server_context,
+           tcp::endpoint &endpoint, const std::string_view service_name, const int inactivity_timeout_sec)
+      : io_thread_context_pool_(io_thread_context_pool),
         data_(data),
         server_context_(server_context),
-        acceptor_(io_context_pool_.GetIOContext()),
+        acceptor_(io_thread_context_pool.GetIOContext()),
         endpoint_{endpoint},
         service_name_{service_name},
         inactivity_timeout_sec_{inactivity_timeout_sec} {
@@ -95,7 +95,7 @@ class Listener final : public std::enable_shared_from_this<Listener<TSession, TS
   }
 
   void DoAccept() {
-    acceptor_.async_accept(io_context_pool_.GetIOContext(),
+    acceptor_.async_accept(io_thread_context_pool_.GetIOContext(),
                            [shared_this = this->shared_from_this()](auto ec, auto &&socket) {
                              shared_this->OnAccept(ec, std::forward<decltype(socket)>(socket));
                            });
@@ -118,7 +118,7 @@ class Listener final : public std::enable_shared_from_this<Listener<TSession, TS
     DoAccept();
   }
 
-  IOContextPool &io_context_pool_;
+  IOContextThreadPool &io_thread_context_pool_;
   TSessionData *data_;
   ServerContext *server_context_;
   tcp::acceptor acceptor_;
--- a/src/communication/v2/listener.hpp
+++ b/src/communication/v2/listener.hpp
@@ -39,6 +39,7 @@ template <class TSession, class TSessionData>
 class Listener final : public std::enable_shared_from_this<Listener<TSession, TSessionData>> {
   using tcp = boost::asio::ip::tcp;
   using SessionHandler = Session<TSession, TSessionData>;
+  using std::enable_shared_from_this<Listener<TSession, TSessionData>>::shared_from_this;
 
  public:
   Listener(const Listener &) = delete;
@@ -99,8 +100,8 @@ class Listener final : public std::enable_shared_from_this<Listener<TSession, TS
 
   void DoAccept() {
     acceptor_.async_accept(io_thread_context_pool_.GetIOContext(),
-                           [shared_this = this->shared_from_this()](auto ec, auto &&socket) {
-                             shared_this->OnAccept(ec, std::forward<decltype(socket)>(socket));
+                           [shared_this = shared_from_this()](auto ec, boost::asio::ip::tcp::socket &&socket) {
+                             shared_this->OnAccept(ec, std::move(socket));
                            });
   }
 
--- a/src/communication/v2/listener.hpp
+++ b/src/communication/v2/listener.hpp
@@ -24,6 +24,7 @@
 #include <boost/beast/core.hpp>
 
 #include "communication/context.hpp"
+#include "communication/v2/pool.hpp"
 #include "communication/v2/session.hpp"
 #include "utils/spin_lock.hpp"
 #include "utils/synchronized.hpp"
@@ -53,12 +54,12 @@ class Listener final : public std::enable_shared_from_this<Listener<TSession, TS
   void Start() { DoAccept(); }
 
  private:
-  Listener(boost::asio::io_context &ioc, TSessionData *data, ServerContext *server_context, tcp::endpoint &endpoint,
+  Listener(IOContextPool &io_context_pool, TSessionData *data, ServerContext *server_context, tcp::endpoint &endpoint,
            const std::string_view service_name, const int inactivity_timeout_sec)
-      : ioc_(ioc),
+      : io_context_pool_(io_context_pool),
         data_(data),
         server_context_(server_context),
-        acceptor_(ioc),
+        acceptor_(io_context_pool_.GetIOContext()),
         endpoint_{endpoint},
         service_name_{service_name},
         inactivity_timeout_sec_{inactivity_timeout_sec} {
@@ -94,9 +95,10 @@ class Listener final : public std::enable_shared_from_this<Listener<TSession, TS
   }
 
   void DoAccept() {
-    acceptor_.async_accept(ioc_, [shared_this = this->shared_from_this()](auto ec, auto socket) {
-      shared_this->OnAccept(ec, std::move(socket));
-    });
+    acceptor_.async_accept(io_context_pool_.GetIOContext(),
+                           [shared_this = this->shared_from_this()](auto ec, auto &&socket) {
+                             shared_this->OnAccept(ec, std::forward<decltype(socket)>(socket));
+                           });
   }
 
   void OnAccept(boost::system::error_code ec, tcp::socket socket) {
@@ -116,7 +118,7 @@ class Listener final : public std::enable_shared_from_this<Listener<TSession, TS
     DoAccept();
   }
 
-  boost::asio::io_context &ioc_;
+  IOContextPool &io_context_pool_;
   TSessionData *data_;
   ServerContext *server_context_;
   tcp::acceptor acceptor_;
--- a/src/communication/v2/listener.hpp
+++ b/src/communication/v2/listener.hpp
@@ -105,11 +105,11 @@ class Listener final : public std::enable_shared_from_this<Listener<TSession, TS
       return LogError(ec, "accept");
     }
 
-    spdlog::info("Accepted a connection from {}:", service_name_, socket.local_endpoint().address(),
-                 socket.local_endpoint().port());
+    spdlog::info("Accepted a connection from {}:", service_name_, socket.remote_endpoint().address(),
+                 socket.remote_endpoint().port());
 
     auto session = SessionHandler::Create(std::move(socket), this->data_, *this->server_context_, this->endpoint_,
-                                          this->inactivity_timeout_sec_);
+                                          this->inactivity_timeout_sec_, service_name_);
     sessions_.WithLock([session = session](auto &sessions) {
       // Clean disconnected clients
       std::erase_if(sessions, [](const auto &elem) { return !elem->IsConnected(); });
--- a/src/communication/v2/listener.hpp
+++ b/src/communication/v2/listener.hpp
@@ -58,12 +58,12 @@ class Listener final : public std::enable_shared_from_this<Listener<TSession, TS
   bool IsRunning() const noexcept { return alive_.load(std::memory_order_relaxed); }
 
  private:
-  Listener(IOContextThreadPool &io_thread_context_pool, TSessionData *data, ServerContext *server_context,
+  Listener(boost::asio::io_context &io_context, TSessionData *data, ServerContext *server_context,
            tcp::endpoint &endpoint, const std::string_view service_name, const uint64_t inactivity_timeout_sec)
-      : io_thread_context_pool_(io_thread_context_pool),
+      : io_context_(io_context),
         data_(data),
         server_context_(server_context),
-        acceptor_(io_thread_context_pool.GetIOContext()),
+        acceptor_(io_context_),
         endpoint_{endpoint},
         service_name_{service_name},
         inactivity_timeout_{inactivity_timeout_sec} {
@@ -99,7 +99,7 @@ class Listener final : public std::enable_shared_from_this<Listener<TSession, TS
   }
 
   void DoAccept() {
-    acceptor_.async_accept(io_thread_context_pool_.GetIOContext(),
+    acceptor_.async_accept(io_context_,
                            [shared_this = shared_from_this()](auto ec, boost::asio::ip::tcp::socket &&socket) {
                              shared_this->OnAccept(ec, std::move(socket));
                            });
@@ -127,7 +127,7 @@ class Listener final : public std::enable_shared_from_this<Listener<TSession, TS
     alive_.store(false, std::memory_order_relaxed);
   }
 
-  IOContextThreadPool &io_thread_context_pool_;
+  boost::asio::io_context &io_context_;
   TSessionData *data_;
   ServerContext *server_context_;
   tcp::acceptor acceptor_;
--- a/src/communication/v2/listener.hpp
+++ b/src/communication/v2/listener.hpp
@@ -1,0 +1,144 @@
+// Copyright 2022 Memgraph Ltd.
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
+// License, and you may not use this file except in compliance with the Business Source License.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#pragma once
+
+#include <cstddef>
+#include <list>
+#include <memory>
+#include <thread>
+
+#include <spdlog/spdlog.h>
+#include <boost/asio/io_context.hpp>
+#include <boost/asio/ip/tcp.hpp>
+#include <boost/asio/strand.hpp>
+#include <boost/beast/core.hpp>
+
+#include "communication/context.hpp"
+#include "communication/v2/session.hpp"
+#include "communication/websocket/session.hpp"
+#include "utils/spin_lock.hpp"
+#include "utils/synchronized.hpp"
+
+namespace memgraph::communication {
+inline void LogError(boost::beast::error_code ec, const std::string_view what) {
+  spdlog::warn("Websocket listener failed on {}: {}", what, ec.message());
+}
+
+template <class TSession, class TSessionData>
+class Listener final : public std::enable_shared_from_this<Listener<TSession, TSessionData>> {
+  using tcp = boost::asio::ip::tcp;
+  using SessionHandler = Session<TSession, TSessionData>;
+
+ public:
+  Listener(const Listener &) = delete;
+  Listener(Listener &&) = delete;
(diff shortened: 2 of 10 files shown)


src/communication/context.cpp
```diff
@@ -78,7 +78,10 @@ bool ClientContext::use_ssl() { return use_ssl_; }
 
 ServerContext::ServerContext(const std::string &key_file, const std::string &cert_file, const std::string &ca_file,
                              bool verify_peer) {
-  ctx_.emplace(boost::asio::ssl::context::tls_server);
+  ctx_.emplace(boost::asio::ssl::context::sslv23_server);
+  // NOLINTNEXTLINE(hicpp-signed-bitwise)
+  ctx_->set_options(boost::asio::ssl::context::default_workarounds | boost::asio::ssl::context::no_sslv2 |
```
antaljanosbenjamin, line 83: Don't we need TLS instead of SSL?

src/communication/context.cpp
```diff
@@ -78,7 +78,10 @@ bool ClientContext::use_ssl() { return use_ssl_; }
...
+                    boost::asio::ssl::context::single_dh_use);
```
antaljanosbenjamin, line 84: If you check their definition, they are removed in OpenSSL 1.1.0. Don't we use only OpenSSL 1.1.0 and later versions? If so, then this is not necessary.

src/memgraph.cpp
```diff
@@ -140,11 +141,6 @@ DEFINE_VALIDATED_int32(monitoring_port, 7444,
                        "Port on which the websocket server for Memgraph monitoring should listen.",
                        FLAG_IN_RANGE(0, std::numeric_limits<uint16_t>::max()));
 // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
-DEFINE_VALIDATED_int32(bolt_num_workers, std::max(std::thread::hardware_concurrency(), 1U),
-                       "Number of workers used by the Bolt server. By default, this will be the "
-                       "number of processing units available on the machine.",
-                       FLAG_IN_RANGE(1, INT32_MAX));
-// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
```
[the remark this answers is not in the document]
jbajic: You are right I missed that

src/communication/context.cpp
jbajic: I think we want to support OpenSSL 1.0.0 and 1.1.0 since we also support native OpenSSL on Centos 7.

src/communication/context.cpp
jbajic: I am not entirely sure what doe flags imply, so I will check it out.

src/communication/context.cpp
antaljanosbenjamin: Khm... CentOS... :smile: Sorry then, you are right.

antaljanosbenjamin (changes requested):
After we have results of the measurement of the `thread-per-core` vs `io_context-per-core` approaches, we should discuss how to eliminate the current bottlenecks introduced by the `io_context`.

src/communication/v2/session.hpp
```diff
@@ -0,0 +1,331 @@
...
+          boost::asio::bind_executor(strand_, std::bind_front(&Session::OnHandshake, this->shared_from_this())));
+    }
+  }
+
+  void OnHandshake(const boost::system::error_code &ec) {
+    DoRead();
+    if (ec) {
+      OnError(ec);
+      return;
+    }
```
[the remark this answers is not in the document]
jbajic: You are right I don't know how I missed that

src/communication/v2/session.hpp
```diff
@@ -0,0 +1,331 @@
...
+                            socket.lowest_layer().set_option(
+                                boost::asio::socket_base::keep_alive(true));  // enable SO_KEEPALIVE
+                            socket.lowest_layer().non_blocking(false);
+                          },
+                          [](TCPSocket &socket) {
+                            socket.set_option(boost::asio::ip::tcp::no_delay(true));        // enable PSH
+                            socket.set_option(boost::asio::socket_base::keep_alive(true));  // enable SO_KEEPALIVE
+                            socket.non_blocking(false);
+                          }},
+        socket_);
```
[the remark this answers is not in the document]
jbajic: I've tried doing something like that but I've encountered issues, so let me explain:
1. `lowest_layer()` does not return TCP socket but the lowest layer type, so the function LowestLayer does not have the same return type
2. ok then I've tried letting the compiler deduce which return type it is
```c++
  auto &LowestLayer() {
    return std::visit(utils::Overloaded{[](TCPSocket &socket) { return socket; },
                                        [](SSLSocket &socket) { return socket.lowest_layer(); }},
                      socket_);
  }
```
and here I get the error that I am trying to copy socket, which is understandable since that can happen from the Overload lambda call,
3. Then I've tried:
```c++
  auto &LowestLayer() {
    return std::visit(utils::Overloaded{[](TCPSocket &socket) -> auto& { return socket; },
                                        [](SSLSocket &socket) -> auto& { return socket.lowest_layer(); }},
                      socket_);
  }
```
I've also tried with `TCPSocket&` and `SSLSocket::lowest_layer_type&` as return types but I get  `non-void constexpr function 'visit...` errors. Not sure if I am missing something?

src/communication/context.cpp
jbajic: I've left our boost::asio::context suffix, but here are the flags and what I think which ones we should apply:
- `sslv23_server` will be converted to `tls_server` since we want to use a modern version of SSL a TLS
- `default_workarounds` sets a group of default options for a specific OpenSSL version, I would keep this since the options were defined by the maintainer themselves
- `no_sslv2`  translates to `SSL_OP_NO_SSLv2` which disables ssl v2 and that's recommended, I suggest disabling ssl v3 as well since that's no longer secure
- `single_dh_use` translates to `SSL_OP_SINGLE_DH_USE` and that forces the older OpenSSL version (prior 1.0.2) to use  a new key for each session, we will leave for centos 7

src/communication/v2/listener.hpp
```diff
@@ -0,0 +1,141 @@
...
+    if (ec) {
+      OnError(ec, "listen");
+      return;
+    }
+  }
+
+  void DoAccept() {
+    acceptor_.async_accept(io_thread_context_pool_.GetIOContext(),
+                           [shared_this = this->shared_from_this()](auto ec, auto &&socket) {
+                             shared_this->OnAccept(ec, std::forward<decltype(socket)>(socket));
```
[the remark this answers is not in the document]
kostasrim: Some explanations here:

The signature of `OnAccept` is: `OnAccept(boost::system::error_code ec, tcp::socket socket)`

If socket is an `rvalue` reference, then `std::forward` casts it to an rvalue and then moved to the `socket` parameter of `OnAccept`. 

If the socket is an `lvalue` reference, then std::forward does nothing, the result is an lvalue reference and the `socket` gets copied to the socket parameter of `OnAccept`. 

Therefore, with the above-suggested changes, if you now accept by value, then if the result is an `rvalue` reference, the move constructor will be called, and then the socket will be moved to `OnAccept`. That's two moves. 

If the the passed value is an lvalue, then socket will be initialized with a copy which will then be moved to `OnAccept`. 

Therefore, I think `std::forward` is not necessarry for this case.

src/communication/v2/listener.hpp
```diff
@@ -0,0 +1,141 @@
...
+                           });
+  }
+
+  void OnAccept(boost::system::error_code ec, tcp::socket socket) {
+    if (ec) {
+      return OnError(ec, "accept");
+    }
+
+    auto session = SessionHandler::Create(std::move(socket), this->data_, *this->server_context_, this->endpoint_,
```
kostasrim, line 110:
```suggestion
    auto session = SessionHandler::Create(std::move(socket), data_, *server_context_, endpoint_,
```

src/communication/v2/listener.hpp
```diff
@@ -0,0 +1,141 @@
...
+                                          this->inactivity_timeout_sec_, service_name_);
```
kostasrim, line 111:
```suggestion
                                          inactivity_timeout_sec_, service_name_);
```

src/communication/v2/listener.hpp
```diff
@@ -0,0 +1,141 @@
...
+    alive_.store(false, std::memory_order_relaxed);
+  }
+
+  IOContextThreadPool &io_thread_context_pool_;
+  TSessionData *data_;
+  ServerContext *server_context_;
+  tcp::acceptor acceptor_;
+
+  tcp::endpoint endpoint_;
+  std::string_view service_name_;
```
kostasrim, line 133:
```suggestion
  std::string service_name_;
```

kostasrim: Since service name is owned by the listener, I think it should also own the `service_name` instead of a view.

src/communication/v2/listener.hpp
```diff
@@ -0,0 +1,141 @@
...
+  int inactivity_timeout_sec_;
+
+  // OnAccept is not performed within strand, and it can be performed concurrently
```
kostasrim, line 136: Maybe I am missing something here, but shouldn't we use a strand here? A connection has a state, which is transitioned based on the data send/received. Therefore, logically, each step, is a continuation of the previous one (there is an implicit dependency on each step to the previous) which implies a strand? What am I missing here?

src/communication/v2/server.hpp
```diff
@@ -0,0 +1,122 @@
...
+    spdlog::info("{} shutting down...", service_name_);
+  }
+
+  void AwaitShutdown() { context_thread_pool_.AwaitShutdown(); }
+
+  bool IsRunning() const noexcept { return context_thread_pool_.IsRunning() && listener_->IsRunning(); }
+
+ private:
+  ServerEndpoint endpoint_;
+  std::string_view service_name_;
```
kostasrim, line 116: I think the server should own the `service_name_` instead of being a view.

src/communication/v2/server.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+  bool Start() {
+    MG_ASSERT(!IsRunning(), "The server was already started!");
+    listener_->Start();
+
+    spdlog::info("{} server is fully armed and operational", service_name_);
+    spdlog::info("{} listening on {}", service_name_, endpoint_.address());
+    context_thread_pool_.Run();
+
+    return true;
```
kostasrim, line 102: This `function` always returns `true` and never `false`. Consider making it `void` or handle the `false` case.

src/communication/v2/session.hpp
```diff
@@ -0,0 +1,331 @@
...
+                                   boost::asio::write(socket, boost::asio::buffer(data, len));
+                                 },
+                                 [shared_this = this->shared_from_this(), data, len](SSLSocket &socket) {
+                                   boost::asio::write(socket, boost::asio::buffer(data, len));
+                                 }},
+               socket_);
+  }
+
+  void DoRead() {
+    if (!IsConnected()) {
```
kostasrim, line 181: Do we want this to be a `no-op` if there is no active connection?

src/communication/v2/session.hpp
```diff
@@ -0,0 +1,331 @@
...
+        socket_);
+    timeout_timer_.expires_at(boost::asio::steady_timer::time_point::max());
+    spdlog::info("Accepted a connection from {}:", service_name_, remote_endpoint_.address(), remote_endpoint_.port());
+  }
+
+  void DoWrite(const uint8_t *data, size_t len, bool /*have_more*/) {
+    if (!IsConnected()) {
+      return;
+    }
+    std::visit(utils::Overloaded{[shared_this = this->shared_from_this(), data, len](TCPSocket &socket) {
```
kostasrim, line 171:
```suggestion
    std::visit(utils::Overloaded{[shared_this = shared_from_this(), data, len](TCPSocket &socket) {
```

src/communication/v2/session.hpp
```diff
@@ -0,0 +1,331 @@
...
+                                   boost::asio::write(socket, boost::asio::buffer(data, len));
+                                 },
+                                 [shared_this = this->shared_from_this(), data, len](SSLSocket &socket) {
```
kostasrim, line 174:
```suggestion
                                 [shared_this = shared_from_this(), data, len](SSLSocket &socket) {
```

src/communication/v2/session.hpp
```diff
@@ -0,0 +1,331 @@
...
+                                   boost::asio::write(socket, boost::asio::buffer(data, len));
+                                 }},
+               socket_);
+  }
+
+  void DoRead() {
+    if (!IsConnected()) {
+      return;
+    }
+    timeout_timer_.expires_after(std::chrono::seconds(this->timeout_seconds_));
```
kostasrim, line 184:
```suggestion
    timeout_timer_.expires_after(std::chrono::seconds(timeout_seconds_));
```

src/communication/v2/session.hpp
```diff
@@ -0,0 +1,331 @@
...
+    ExecuteForSocket([this](auto &&socket) {
+      auto buffer = input_buffer_.write_end()->Allocate();
+      socket.async_read_some(
+          boost::asio::buffer(buffer.data, buffer.len),
+          boost::asio::bind_executor(strand_, std::bind_front(&Session::OnRead, this->shared_from_this())));
```
kostasrim, line 189:
```suggestion
          boost::asio::bind_executor(strand_, std::bind_front(&Session::OnRead, shared_from_this())));
```

src/communication/v2/session.hpp
```diff
@@ -0,0 +1,331 @@
...
+  }
+
+  void DoHandshake() {
+    if (!IsConnected()) {
+      return;
+    }
+    if (auto *socket = std::get_if<SSLSocket>(&socket_); socket) {
+      socket->async_handshake(
+          boost::asio::ssl::stream_base::server,
+          boost::asio::bind_executor(strand_, std::bind_front(&Session::OnHandshake, this->shared_from_this())));
```
kostasrim, line 262:
```suggestion
          boost::asio::bind_executor(strand_, std::bind_front(&Session::OnHandshake, shared_from_this())));
```

kostasrim:
Also, are you planning to write units on this PR?

kostasrim (commented):
Great work, minor comments. Will do another review on the weekend to make sure I am onboard with everything.

src/communication/v2/listener.hpp
```diff
@@ -0,0 +1,141 @@
...
+    acceptor_.listen(boost::asio::socket_base::max_listen_connections, ec);
+    if (ec) {
+      OnError(ec, "listen");
+      return;
+    }
+  }
+
+  void DoAccept() {
+    acceptor_.async_accept(io_thread_context_pool_.GetIOContext(),
+                           [shared_this = this->shared_from_this()](auto ec, auto &&socket) {
```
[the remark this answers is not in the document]
jbajic: The usage of `this` is mandatory in this case since we are referring to the template Base class whose parameter depends Session parameter, and in that case we have to sue this (read more here https://stackoverflow.com/questions/17853212/using-shared-from-this-in-templated-classes)

src/communication/v2/listener.hpp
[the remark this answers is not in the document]
jbajic: The reason I would keep `std::forward` is that we are dealing with forwarding reference (socket&&), and I would always expect from `async_accept` to make the right call do we need to copy the socket or do we just reuse the given one since if we do `std::move` we can steal the ownership unintentionally, like its shown in this example (https://clang.llvm.org/extra/clang-tidy/checks/bugprone-move-forwarding-reference.html) (clang-tidy told me that)

src/communication/v2/listener.hpp
jbajic: This I can remove :+1:

src/communication/v2/listener.hpp
jbajic: Aha you are right nobody owns it, I will move it to the server since it makes more sense that the server owns it than the listener, since they both need it.

src/communication/v2/session.hpp
jbajic: What do you mean by that?

src/communication/v2/listener.hpp
jbajic: I have to think about this one

src/communication/v2/session.hpp
```diff
@@ -0,0 +1,331 @@
...
+    if (!IsConnected()) {
+      return;
+    }
+    timeout_timer_.expires_after(std::chrono::seconds(this->timeout_seconds_));
+    ExecuteForSocket([this](auto &&socket) {
+      auto buffer = input_buffer_.write_end()->Allocate();
+      socket.async_read_some(
+          boost::asio::buffer(buffer.data, buffer.len),
+          boost::asio::bind_executor(strand_, std::bind_front(&Session::OnRead, this->shared_from_this())));
+    });
```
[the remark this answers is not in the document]
jbajic: so funny thing the `boost::asio::async_read` does not work in our case and here is my reasoning behind this. In order to ensure that we have read everything we need to know how much is this everything therefore the `boost::asio::async_read` must have and indication of the size of the data it will read, there are three options:
- `transfer_all()` => which is a default one in which we need to receive the whole buffer in order to confirm that we have read anything, the issue is that bolt messages rarely fill the whole bugger, they are dynamic in size
- `transfer_at_least(size_t)` => to which we pass the minimum size of the buffer that signals that we have read something, this in my opinion ends up being same as `async_read_some`
- `transfer_exactly(size_t)` => to which we need to pass to exact size we want to read, which again is unknown
the

src/communication/v2/listener.hpp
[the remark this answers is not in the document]
kostasrim: That's true but you can bring the dependent base locally in the class definition by doing:
```
using Base::shared_from_this;
```
Now the name is available :)

src/communication/v2/session.hpp
kostasrim: I mean that `DoRead` doesn't really tell you if it read anything because if the connection is false it returns immediately. The question is, are these the right semantics? (I am not sure that's why I am asking)

src/communication/v2/listener.hpp
jbajic: From what I understood the reason for using strand is to have the operations run in the order you want, (to synchronize them) e.g. when you read data you want to write something back and that has to be done one after the other. And I would say the same case applies to accept, the order of accepts does not matter, there is no need to sync them, and there is only one `accept` method.

src/communication/v2/session.hpp
jbajic: Aha, I get the question, so we don't want to do anything if for some reason we initiated a shutdown, but maybe there will be read/write as the next action, we want to skip it since we will be closing the connection soon.

src/communication/v2/session.hpp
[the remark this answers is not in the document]
antaljanosbenjamin: Actually it is even easier! `lowest_layer` is defined for both `TCPSocket` and `SSLSocket`, so it can be written in this way:
```
    std::visit(utils::Overloaded{[](auto &socket) {
                 auto &lowest_layer = socket.lowest_layer();
                 lowest_layer.set_option(boost::asio::ip::tcp::no_delay(true));        // enable PSH
                 lowest_layer.set_option(boost::asio::socket_base::keep_alive(true));  // enable SO_KEEPALIVE
                 lowest_layer.non_blocking(false);
               }},
               socket_);
```

Proof:
`boost::asio::ip::tcp::socket` is defined as `basic_stream_socket<tcp>`, and `basic_stream_socket` is derived from `basic_socket`, so `basic_socket::lowest_layer` can be used for `TCPSocket`.
```
  /// Get a reference to the lowest layer.
  /**
   * This function returns a reference to the lowest layer in a stack of
   * layers. Since a basic_socket cannot contain any further layers, it simply
   * returns a reference to itself.
   *
   * @return A reference to the lowest layer in the stack of layers. Ownership
   * is not transferred to the caller.
   */
  lowest_layer_type& lowest_layer()
  {
    return *this;
  }
```
For `SSLSocket` which is `boost::asio::ssl::stream<TCPSocket>`, the `lowest_layer` is defined in the following way:
```
  /// Get a reference to the lowest layer.
  /**
   * This function returns a reference to the lowest layer in a stack of
   * stream layers.
   *
   * @return A reference to the lowest layer in the stack of stream layers.
   * Ownership is not transferred to the caller.
   */
  lowest_layer_type& lowest_layer()
  {
    return next_layer_.lowest_layer();
  }
```

I tried it, and it worked for me, so I think it should work in general. What do you think?

src/communication/v2/listener.hpp
```diff
@@ -53,28 +55,30 @@ class Listener final : public std::enable_shared_from_this<Listener<TSession, TS
 
   void Start() { DoAccept(); }
 
+  bool IsRunning() const noexcept { return alive_.load(std::memory_order_relaxed); }
+
  private:
-  Listener(IOContextPool &io_context_pool, TSessionData *data, ServerContext *server_context, tcp::endpoint &endpoint,
-           const std::string_view service_name, const int inactivity_timeout_sec)
-      : io_context_pool_(io_context_pool),
+  Listener(IOContextThreadPool &io_thread_context_pool, TSessionData *data, ServerContext *server_context,
```
antaljanosbenjamin, line 61: As `Listener` doesn't actually use the the functionalities of the thread pool, I think passing in an `io_context` reference is enough, that is what the `Listener` actually need.

src/communication/v2/listener.hpp
```diff
@@ -83,53 +87,58 @@ class Listener final : public std::enable_shared_from_this<Listener<TSession, TS
...
 
   tcp::endpoint endpoint_;
   std::string_view service_name_;
-  int inactivity_timeout_sec_;
+  std::chrono::seconds inactivity_timeout_;
 
   // OnAccept is not performed within strand, and it can be performed concurrently
   // therefore a lock is needed.
-  utils::Synchronized<std::unordered_set<std::shared_ptr<SessionHandler>>, utils::SpinLock> sessions_;
+  utils::Synchronized<std::vector<std::shared_ptr<SessionHandler>>, utils::SpinLock> sessions_;
```
antaljanosbenjamin, line 141: I like this! :rocket:

src/communication/v2/pool.hpp
```diff
@@ -22,42 +22,32 @@
...
-  IOContextPool(const IOContextPool &) = delete;
-  IOContextPool &operator=(const IOContextPool &) = delete;
-  IOContextPool(IOContextPool &&) = delete;
-  IOContextPool &operator=(IOContextPool &&) = delete;
-  ~IOContextPool() { MG_ASSERT(background_threads_.empty(), "Error while destructing thread pool"); };
+  IOContextThreadPool(const IOContextThreadPool &) = delete;
+  IOContextThreadPool &operator=(const IOContextThreadPool &) = delete;
+  IOContextThreadPool(IOContextThreadPool &&) = delete;
+  IOContextThreadPool &operator=(IOContextThreadPool &&) = delete;
+  ~IOContextThreadPool() { MG_ASSERT(background_threads_.empty(), "Error while destructing context thread pool"); };
```
antaljanosbenjamin, line 39: What is the reason for this assert? I think in an RAII class the destructor should take care of this. A good example is how the standard now contains `std::jthread`.

src/communication/v2/pool.hpp
```diff
@@ -69,23 +59,16 @@ class IOContextPool {
...
+  IOContext &GetIOContext() noexcept { return io_context_; }
+
  private:
   /// The pool of io_context.
-  std::vector<std::shared_ptr<IOContext>> io_contexts_;
-  std::vector<std::shared_ptr<IOContextGuard>> guards_;
+  IOContext io_context_;
+  IOContextGuard guard_;
+  size_t pool_size_;
   std::vector<std::thread> background_threads_;
```
antaljanosbenjamin, line 71: Maybe `std::jthread` can be used, so the joining would be done in a RAII manner, `AwaitShutdown` should become much simpler, just `background_threads_.clear()`.

src/communication/v2/server.hpp
```diff
@@ -75,11 +77,11 @@ class Server final {
...
         service_name_{service_name},
-        context_pool_{workers_count},
-        listener_{Listener<TSession, TSessionData>::Create(context_pool_, session_data, server_context, endpoint_,
-                                                           service_name_, inactivity_timeout_sec)} {}
+        context_thread_pool_{workers_count},
+        listener_{Listener<TSession, TSessionData>::Create(context_thread_pool_, session_data, server_context,
+                                                           endpoint_, service_name_, inactivity_timeout_sec)} {}
 
-  ~Server() { MG_ASSERT(!context_pool_.IsRunning(), "Server wasn't shutdown properly"); }
+  ~Server() { MG_ASSERT(!IsRunning(), "Server wasn't shutdown properly"); }
```
antaljanosbenjamin, line 84: Here I still question the decision, but I am more okay with this, as this server is quite specific compared to `IOContextThreadPool`. Also this worked in the same way as before and because we want to maintain "plug-and-play" compatibility, this is kind of necessary.

src/communication/v2/session.hpp
```diff
@@ -0,0 +1,347 @@
...
+      spdlog::info("Shutting down session after {} of inactivity", this->timeout_seconds_);
+      DoShutdown();
+    } else {
+      // Put the actor back to sleep.
+      timeout_timer_.async_wait(
+          boost::asio::bind_executor(this->strand_, std::bind(&Session::OnTimeout, shared_from_this())));
+    }
+  }
+
+  std::variant<TCPSocket, SSLSocket> CreateWebSocket(tcp::socket &&socket, ServerContext &context) {
```
antaljanosbenjamin, line 309:
```suggestion
  std::variant<TCPSocket, SSLSocket> CreateSocket(tcp::socket &&socket, ServerContext &context) {
```

src/communication/v2/session.hpp
```diff
@@ -0,0 +1,347 @@
...
+  Session &operator=(const Session &) = delete;
+  Session &operator=(Session &&) = delete;
+  ~Session() {
+    if (IsConnected()) {
+      spdlog::error("Session: Destructor called while execution is active");
+    }
+  }
+
+  bool Start() {
+    if (IsConnected()) {
```
antaljanosbenjamin, line 105: I thing here the `execution_active_` flag is better, because in theory `IsConnected` might return false after `Start` returns, because the socket might not be properly connected (however I couldn't find what exactly "open" means in `boost::asio`). This might cause some issue if `Start` is called at least twice fast enough.

src/communication/v2/session.hpp
```diff
@@ -0,0 +1,347 @@
...
+    }
+    execution_active_ = true;
+    timeout_timer_.async_wait(boost::asio::bind_executor(strand_, std::bind(&Session::OnTimeout, shared_from_this())));
+
+    if (auto *socket = std::get_if<SSLSocket>(&socket_); socket) {
+      boost::asio::dispatch(shared_from_this()->strand_,
+                            [shared_this = shared_from_this()] { shared_this->DoHandshake(); });
+    } else {
+      boost::asio::dispatch(this->shared_from_this()->strand_,
+                            [shared_this = shared_from_this()] { shared_this->DoRead(); });
```
antaljanosbenjamin, line 116:
```suggestion
      boost::asio::dispatch(strand_,
                            [shared_this = shared_from_this()] { shared_this->DoHandshake(); });
    } else {
      boost::asio::dispatch(strand_,
                            [shared_this = shared_from_this()] { shared_this->DoRead(); });
```

We don't need to create a `shared_ptr` just to get the strand.

src/communication/v2/session.hpp
```diff
@@ -0,0 +1,347 @@
...
+    // deadline before this actor had a chance to run.
+    if (timeout_timer_.expiry() <= boost::asio::steady_timer::clock_type::now()) {
+      // The deadline has passed. Stop the session. The other actors will
+      // terminate as soon as possible.
+      spdlog::info("Shutting down session after {} of inactivity", this->timeout_seconds_);
+      DoShutdown();
+    } else {
+      // Put the actor back to sleep.
+      timeout_timer_.async_wait(
+          boost::asio::bind_executor(this->strand_, std::bind(&Session::OnTimeout, shared_from_this())));
```
antaljanosbenjamin, line 305:
```suggestion
      spdlog::info("Shutting down session after {} of inactivity", timeout_seconds_);
      DoShutdown();
    } else {
      // Put the actor back to sleep.
      timeout_timer_.async_wait(
          boost::asio::bind_executor(strand_, std::bind(&Session::OnTimeout, shared_from_this())));
```

src/communication/v2/session.hpp
```diff
@@ -0,0 +1,347 @@
...
+      shared_this->DoWrite(data, len, have_more);
+    });
+    return true;
+  }
+
+  bool IsConnected() const {
+    return std::visit(
+        utils::Overloaded{
+            [this](const SSLSocket &socket) { return execution_active_ && socket.lowest_layer().is_open(); },
+            [this](const TCPSocket &socket) { return execution_active_ && socket.is_open(); }},
```
antaljanosbenjamin, line 135:
```suggestion
            [this](const auto &socket) { return execution_active_ && socket.lowest_layer().is_open(); }},
```

src/communication/v2/session.hpp
```diff
@@ -0,0 +1,347 @@
...
+                                 },
+                                 [](SSLSocket &ssl_socket) {
+                                   boost::system::error_code ec;
+                                   ssl_socket.lowest_layer().shutdown(boost::asio::ip::tcp::socket::shutdown_both, ec);
+                                   if (ec) {
+                                     spdlog::error("Session shutdown failed: {}", ec.what());
+                                   }
+                                   ssl_socket.lowest_layer().close();
+                                 }},
+               socket_);
```
antaljanosbenjamin, line 269:
```suggestion
    std::visit(utils::Overloaded{[](auto &socket) {
                                   boost::system::error_code ec;
                                   auto &lowest_layer = socket.lowest_layer(); lowest_layer.shutdown(boost::asio::ip::tcp::socket::shutdown_both, ec);
                                   if (ec) {
                                     spdlog::error("Session shutdown failed: {}", ec.what());
                                   }
                                   lowest_layer.close();
                                 }},
               socket_);
```

src/communication/v2/session.hpp
```diff
@@ -0,0 +1,347 @@
...
+    return TCPSocket{std::move(socket)};
+  }
+
+  auto GetExecutor() {
+    return std::visit(utils::Overloaded{[](auto &&socket) { return socket.get_executor(); }}, socket_);
+  }
+
+  auto GetRemoteEndpoint() {
+    return std::visit(utils::Overloaded{[](TCPSocket &socket) { return socket.remote_endpoint(); },
+                                        [](SSLSocket &socket) { return socket.lowest_layer().remote_endpoint(); }},
```
antaljanosbenjamin, line 324:
```suggestion
    return std::visit(utils::Overloaded{[](auto &socket) { return socket.lowest_layer().remote_endpoint(); }},
```

antaljanosbenjamin (changes requested):

src/communication/v2/pool.hpp
```diff
@@ -0,0 +1,74 @@
...
+namespace memgraph::communication::v2 {
+
+class IOContextThreadPool {
+ private:
+  using IOContext = boost::asio::io_context;
+  using IOContextGuard = boost::asio::executor_work_guard<boost::asio::io_context::executor_type>;
+
+ public:
+  explicit IOContextThreadPool(size_t pool_size) : guard_{io_context_.get_executor()}, pool_size_{pool_size} {
+    MG_ASSERT(pool_size != 0, "Pool size must be bigger then 0!");
```
kostasrim, line 32:
```suggestion
    MG_ASSERT(pool_size != 0, "Pool size must be greater than 0!");
```

src/communication/v2/pool.hpp
kostasrim: This is a gem :gem:

src/communication/v2/pool.hpp
```diff
@@ -0,0 +1,74 @@
...
+#include <vector>
+
+#include <boost/asio/executor_work_guard.hpp>
+#include <boost/asio/io_context.hpp>
+
+#include "utils/logging.hpp"
+
+namespace memgraph::communication::v2 {
+
+class IOContextThreadPool {
```
kostasrim, line 25:
```suggestion
class IOContextThreadPool final {
```

kostasrim (commented):
minor comments. Also, one more question, are you planning to add the binary protocol in this PR?

src/communication/v2/session.hpp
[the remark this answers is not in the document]
jbajic: Aha, you are right, I like this approach much better!

jbajic:
> minor comments. Also, one more question, are you planning to add the binary protocol in this PR?

That comes in the next PR

kostasrim (approved):

antaljanosbenjamin (approved):

PR #376, opened 2022-04-16
[master < T0830-MG] Fix SHOW REPLICATION ROLE and SHOW REPLICAS

gitbuda:
- [x] Write E2E tests
- [x] Check, and update documentation if necessary
- [x] Update [changelog](https://docs.memgraph.com/memgraph/changelog) -> https://github.com/memgraph/docs/pull/381

PR #379, opened 2022-04-23
[master < T0833-MG] Add replica state to SHOW REPLICAS

gitbuda:
- [x] Write E2E tests (depends on #376 because the test there has to be extended and #380 because the `FrequentHeartbeat` has to be tested, first merge those two branches then finish this one):
    - [x] Expand the `Show` e2e test
    - [x] Implement `FrequentHeartbeat` test
- [x] Update [changelog](https://docs.memgraph.com/memgraph/changelog) -> https://github.com/memgraph/docs/pull/381


jbajic (approved):

tests/e2e/replication/show.py
```diff
@@ -30,14 +30,14 @@ def test_show_replicas(connection):
...
     actual_column_names = {x.name for x in cursor.description}
     assert expected_column_names == actual_column_names
 
     expected_data = {
-        ("replica_1", "127.0.0.1:10001", "sync", 0),
-        ("replica_2", "127.0.0.1:10002", "sync", 1.0),
-        ("replica_3", "127.0.0.1:10003", "async", None),
+        ("replica_1", "127.0.0.1:10001", "sync", 0, "ready"),
+        ("replica_2", "127.0.0.1:10002", "sync", 1.0, "ready"),
+        ("replica_3", "127.0.0.1:10003", "async", None, "ready"),
```
antaljanosbenjamin, line 40: Maybe try to add at least an invalid state to make sure it works. I guess a non-existing replica would do the trick, but if not, then feel free to leave as is, I can check it manually.

antaljanosbenjamin (commented):
You mentioned `Implement FrequentHeartbeat test` in the task list. Are you still planning to do that?

src/storage/v2/storage.cpp
```diff
@@ -306,6 +306,13 @@ Storage::Storage(Config config)
       uuid_(utils::GenerateUUID()),
       epoch_id_(utils::GenerateUUID()),
       global_locker_(file_retainer_.AddLocker()) {
+  if (config_.durability.snapshot_wal_mode == Config::Durability::SnapshotWalMode::DISABLED &&
+      replication_role_ == ReplicationRole::MAIN) {
+    spdlog::warn(
+        "The instance has the MAIN replication role, but durability logs and snapshots are disabled. Please consider "
+        "enabling durability by using --storage-snapshot-interval-sec and --storage-wal-enabled flags because "
+        "without write-ahead logs this instance is not replicating any data.");
+  }
```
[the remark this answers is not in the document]
gitbuda: No, because it's valid to have an instance of Memgraph on a single machine without persistence, that also means without any replication 😄

src/query/interpreter.hpp
```diff
@@ -127,6 +127,7 @@ class ReplicationQueryHandler {
     std::string socket_address;
     ReplicationQuery::SyncMode sync_mode;
     std::optional<double> timeout;
+    ReplicationQuery::ReplicaState state;
```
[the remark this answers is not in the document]
42jeremy: It is used. For instance:
https://github.com/memgraph/memgraph/blob/5ed98d82aecbd29d8e1123d2d86260bc28bcb214/src/query/interpreter.cpp#L234-L246

tests/e2e/replication/show.py
42jeremy: I have adapted the test:
-We drop a replica and check it is not anymore in the list. 
-We kill the two other replicas and check that their state is invalid.

tests/e2e/replication/show.py
```diff
@@ -27,17 +32,49 @@ def test_show_replication_role(port, role, connection):
...
+    execute_and_fetch_all(cursor, "DROP REPLICA replica_2")
+    actual_data = set(execute_and_fetch_all(cursor, "SHOW REPLICAS;"))
+    expected_data = {
+        ("replica_1", "127.0.0.1:10001", "sync", 0, "ready"),
+        ("replica_3", "127.0.0.1:10003", "async", None, "ready"),
+    }
+    assert expected_data == actual_data
+
+    # 3/
+    for line in os.popen("ps ax | grep -E '7688|7690' | grep -E 'replica1.log|replica3.log' | grep -v grep"):
```
42jeremy, line 64: I want opinion on that one. Here I want to kill the replicas and check that the main realizes it and mark their state as invalid.
I came up with this solution but am not completely convinced by it. Works on Linux, can't ensure for other OS. 
Here are the other options I have though about:

- When setting up the test from the workload file/runner.py, keeping the PID of MemgraphInstances corresponding to the replicas in a global variable to kill them afterwards. However given the setup and fact that a python script call a bash script which call python etc... This is not possible/easy.

- When setting up the test from the workload file/runner.py, passing instances of MemgraphInstances corresponding to the replicas to the test file. However given the setup and fact that a python script call a bash script which call python etc... This is not possible/easy.

- Writing a new test file (show2.py, but better naming) where I manually start the replicas myself then kill them. I think this is the "best" but requires more devel time.

kostasrim: I think there is a general way to do this with `[psutil](https://pypi.org/project/psutil/)`. It should be something like:

`
for conn in filter(lambda conn: conn.laddr and conn.laddr.port == 7688, psutil.net_connections()):
  os.kill(con.pid, signal.SIGKILL)
`
I haven't tested the above, if it doesn't work, it should be simple to modify based on the docs :)

42jeremy: I will give it a try

42jeremy: @kostasrim as we discussed I have tried the approach where I start the instances myself. It works as is and then we do not risk to kill something we are not supposed to.

tests/e2e/CMakeLists.txt
```diff
@@ -1,25 +1,32 @@
...
-        COMMAND ${CMAKE_COMMAND} -E copy
+        add_custom_target(memgraph__e2e__${TARGET_PREFIX}__${FILE_NAME} ALL
+                COMMAND ${CMAKE_COMMAND} -E copy
                 ${CMAKE_CURRENT_SOURCE_DIR}/${FILE_NAME}
                 ${CMAKE_CURRENT_BINARY_DIR}/${FILE_NAME}
-        DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/${FILE_NAME})
+                DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/${FILE_NAME})
+endfunction()
+
+function(copy_e2e_python_files_from_parent_folder TARGET_PREFIX EXTRA_PATH FILE_NAME)
```
42jeremy, line 16: That deserves an explanation: from build/tests/e2e/replication/my_new_test.py I want to have access to the memgraph.py.MemgraphInstanceRunner.

Structure of tests folder is:
 - memgraph/tests/e2e/memgraph.py this is what we want to access. Note that this file is not copied yet to the /build folder as it is currently only used in memgraph/tests/e2e/runner.py
 - memgraph/tests/replication/show_while_creating_invalid_state.py: this is where I want to access the MemgraphInstanceRunner present in memgraph.py.

I have considered:
 - Not copying memgraph.py and use it directly from memgraph/tests/e2e/. I am not such fan of directly using the "source".
 - Copying memgraph.py from memgraph/tests/e2e/ to memgraph/build/tests/e2e/. However my understanding is that, in Python, it is easy to import function from another file from the same directory but it becomes more tricky when the file is in a different folder.
 - Copying memgraph.py from memgraph/tests/e2e/ to memgraph/build/tests/e2e/replication.

Let me know if anyone else has a better idea, I am open for suggestions.

42jeremy, line 16: That deserves an explanation: from build/tests/e2e/replication/my_new_test.py I want to have access to the memgraph.py.MemgraphInstanceRunner.

Structure of tests folder is:

memgraph/tests/e2e/memgraph.py this is what we want to access. Note that this file is not copied yet to the /build folder as it is currently only used in memgraph/tests/e2e/runner.py
memgraph/tests/replication/show_while_creating_invalid_state.py: this is where I want to access the MemgraphInstanceRunner present in memgraph.py.

I have considered:

Not copying memgraph.py and use it directly from memgraph/tests/e2e/. I am not such fan of directly using the "source".
Copying memgraph.py from memgraph/tests/e2e/ to memgraph/build/tests/e2e/. However my understanding is that, in Python, it is easy to import function from another file from the same directory but it becomes more tricky when the file is in a different folder.
Copying memgraph.py from memgraph/tests/e2e/ to memgraph/build/tests/e2e/replication.

Let me know if anyone else has a better idea, I am open for suggestions.

42jeremy (commented):
.

tests/e2e/CMakeLists.txt
antaljanosbenjamin: > The best would be to re-organize the test so we can use them as module and share some code between all the tests. One will most likely need to execute a main.py that will execute the test himself. All tests would then have access to the module loaded via main.py.

100% agree, however that requires a lot of time. That's why I think the current approach is fine, as we are not really duplicating code.

src/storage/v2/storage.cpp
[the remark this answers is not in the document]
antaljanosbenjamin: Ahh, so there is always a replication role. I see.

tests/e2e/CMakeLists.txt
```diff
@@ -1,25 +1,32 @@
 # Set up C++ functions for e2e tests
 function(add_query_module target_name src)
-  add_library(${target_name} SHARED ${src})
```
antaljanosbenjamin, line 3: We have to introduce [cmake-format](https://cmake-format.readthedocs.io/en/latest/), because all of our CMake files has different formatting...

antaljanosbenjamin (approved):

kostasrim (approved):

PR #380, opened 2022-04-23
[master < T0834-MG] Add frequent replica ping

gitbuda:
The intention here is to help the user to get the most accurate replica state possible.

- [x] Update [changelog](https://docs.memgraph.com/memgraph/changelog) -> https://github.com/memgraph/docs/pull/381


.clang-tidy
```diff
@@ -60,6 +60,7 @@ Checks: '*,
          -readability-implicit-bool-conversion,
          -readability-magic-numbers,
          -readability-named-parameter,
+         -readability-convert-member-functions-to-static,
```
gitbuda, line 63: Not sure if this is the right approach to fix the clang-tidy error? When I run
```
clang-tidy -p build src/storage/v2/replication/replication_server.cpp -- -I./src -I/opt/toolchain-v4/include -I./libs/spdlog/include -std=c++20
```
There are a bunch of warnings but no errors...

What's the best approach here?

antaljanosbenjamin: So first of all, you don't have to fix all the warnings. If you check [our CI](https://github.com/memgraph/memgraph/blob/b782271be87a04173ebe7c724358ba107b15dd9a/.github/workflows/diff.yaml#L114-L115), then you can see only the modified lines are checked in the codebase. So if you don't modify a specific function/line, you don't have fix that. However if you touch it, then the result should be clang-tidy warning free. I think you can use the same command that we are using in our CI to check the warnings locally.

After pointing this out, I think it understandable why I am against of disabling this check. This check is quite reliable and useful. Why not to make such functions static member functions?

src/storage/v2/replication/replication_client.cpp
```diff
@@ -41,12 +41,50 @@ Storage::ReplicationClient::ReplicationClient(std::string name, Storage *storage
...
+  // If success && INVALID -> [it's possible that replica came back to life] -> TryInitializeClient.
+  // If fail -> [replica is not reachable at all] -> INVALID state.
+  // NOTE: TryInitializeClient might return nothing if there is a branching point.
+  if (is_success && replica_state_.load() != replication::ReplicaState::INVALID) {
+    // Pass because all seems fine, replica is in some valid state.
+  } else if (is_success && replica_state_.load() == replication::ReplicaState::INVALID) {
+    TryInitializeClientAsync();
+  } else if (!is_success) {
+    replica_state_.store(replication::ReplicaState::INVALID);
+  }
```
jbajic, line 87:
```suggestion
  if (!is_success) {
    replica_state_.store(replication::ReplicaState::INVALID);
    return;
  }
  if (replica_state_.load() == replication::ReplicaState::INVALID) {
    TryInitializeClientAsync();
  }
```

jbajic (changes requested):

antaljanosbenjamin (changes requested):

src/memgraph.cpp
```diff
@@ -252,6 +252,12 @@ DEFINE_double(query_execution_timeout_sec, 600,
               "Maximum allowed query execution time. Queries exceeding this "
               "limit will be aborted. Value of 0 means no limit.");
 
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
+DEFINE_uint64(
+    replication_replica_check_delay_sec, 1,
+    "The dealy between the end of a replica check/ping and start of another one. If < 1 replicas will NOTE be checket "
```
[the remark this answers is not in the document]
gitbuda: I kept the plural there because there could be more of them.

src/memgraph.cpp
```diff
@@ -252,6 +252,11 @@ DEFINE_double(query_execution_timeout_sec, 600,
               "Maximum allowed query execution time. Queries exceeding this "
               "limit will be aborted. Value of 0 means no limit.");
 
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
+DEFINE_uint64(replication_replica_check_delay_sec, 1,
+              "The time duration between two replica checks/pings. If < 1, replicas will NOT be checked at all. NOTE: "
+              "The MAIN instance allocates a new thread for each REPLICA.");
```
gitbuda, line 258: @antaljanosbenjamin is this an ok resolution (the flag name and the description), I think it doesn't make sense here to explain all the details of the Scheduler implementation...

src/storage/v2/replication/replication_client.cpp
gitbuda: Yep, it should be the same but simplified 😄

jbajic (approved):

src/query/config.hpp
```diff
@@ -21,6 +21,8 @@ struct InterpreterConfig {
 
   // The default execution timeout is 10 minutes.
   double execution_timeout_sec{600.0};
+  // The same as \ref memgraph::storage::replication::ReplicationClientConfig
+  uint64_t replication_replica_check_delay{1};
```
kostasrim, line 25:
```suggestion
  std::chrono::seconds replication_replica_ping_frequency{1};
```

src/query/interpreter.cpp
```diff
@@ -160,7 +160,8 @@ class ReplQueryHandler final : public query::ReplicationQueryHandler {
 
   /// @throw QueryRuntimeException if an error ocurred.
   void RegisterReplica(const std::string &name, const std::string &socket_address,
-                       const ReplicationQuery::SyncMode sync_mode, const std::optional<double> timeout) override {
+                       const ReplicationQuery::SyncMode sync_mode, const std::optional<double> timeout,
+                       const uint64_t replica_check_delay) override {
```
kostasrim, line 164:
```suggestion
                       const std::chrono::seconds replica_check_delay) override {
```

src/query/interpreter.hpp
```diff
@@ -137,7 +137,8 @@ class ReplicationQueryHandler {
 
   /// @throw QueryRuntimeException if an error ocurred.
   virtual void RegisterReplica(const std::string &name, const std::string &socket_address,
-                               const ReplicationQuery::SyncMode sync_mode, const std::optional<double> timeout) = 0;
+                               const ReplicationQuery::SyncMode sync_mode, const std::optional<double> timeout,
+                               const uint64_t replica_check_delay) = 0;
```
kostasrim, line 141:
```suggestion
                               const std::chrono::seconds replica_check_delay) = 0;
```

src/storage/v2/replication/config.hpp
```diff
@@ -16,6 +16,10 @@
 namespace memgraph::storage::replication {
 struct ReplicationClientConfig {
   std::optional<double> timeout;
+  // The default delay between main checking/pinging replicas is 1s because
+  // that seems like a reasonable timeframe in which main should notice a
+  // replica is down.
+  uint64_t replica_check_delay{1};
```
kostasrim, line 22:
```suggestion
  std::chrono::seconds replica_ping_frequency{1};
```

src/memgraph.cpp
```diff
@@ -252,6 +252,11 @@ DEFINE_double(query_execution_timeout_sec, 600,
               "Maximum allowed query execution time. Queries exceeding this "
               "limit will be aborted. Value of 0 means no limit.");
 
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
+DEFINE_uint64(replication_replica_check_delay_sec, 1,
```
kostasrim, line 256:
```suggestion
DEFINE_uint64(replication_replica_ping_frequency_seconds, 1,
```

antaljanosbenjamin: I think `sec` is consistent with the other flags, so I would keep it as is.

antaljanosbenjamin (approved):

gitbuda:
I think the final name for the flag will be `replication_replica_check_frequency_sec`

PR #381, opened 2022-04-23
[master < T0835-MG] Fix abort when sync replica is down

gitbuda:
[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR


src/query/interpreter.cpp
```diff
@@ -2206,25 +2206,39 @@ void RunTriggersIndividually(const utils::SkipList<Trigger> &triggers, Interpret
...
-          const auto &label_name = db_accessor.LabelToName(constraint_violation.label);
-          MG_ASSERT(constraint_violation.properties.size() == 1U);
-          const auto &property_name = db_accessor.PropertyToName(*constraint_violation.properties.begin());
-          spdlog::warn("Trigger '{}' failed to commit due to existence constraint violation on :{}({})", trigger.Name(),
-                       label_name, property_name);
+    auto maybe_commit_error = db_accessor.Commit();
+    if (maybe_commit_error.HasError()) {
+      const auto &commit_error = maybe_commit_error.GetError();
+      switch (commit_error.type) {
+        case storage::CommitError::Type::UNABLE_TO_REPLICATE: {
```
gitbuda, line 2213: Rename the error type to `UNABLE_TO_SYNC_REPLICATE`

src/query/interpreter.cpp
```diff
@@ -2209,25 +2209,39 @@ void RunTriggersIndividually(const utils::SkipList<Trigger> &triggers, Interpret
...
-          const auto &property_name = db_accessor.PropertyToName(*constraint_violation.properties.begin());
-          spdlog::warn("Trigger '{}' failed to commit due to existence constraint violation on :{}({})", trigger.Name(),
-                       label_name, property_name);
+    auto maybe_commit_error = db_accessor.Commit();
+    if (maybe_commit_error.HasError()) {
+      const auto &commit_error = maybe_commit_error.GetError();
+      switch (commit_error.type) {
+        case storage::CommitError::Type::UNABLE_TO_SYNC_REPLICATE: {
+          // TODO(gitbuda): This is tricky because this is an internal
+          // operation. Consider stopping main Memgraph instance here.
```
gitbuda, line 2218: Or we can just pause somehow, "when in doubt don't do anything"

src/storage/v2/storage.cpp
```diff
@@ -1124,6 +1165,18 @@ EdgeTypeId Storage::NameToEdgeType(const std::string_view &name) {
...
+// because there is no an abort op for that yet, one idea is to just apply
+// reverse operation, e.g., CreateIndex <-> DropIndex.
+//
+// Another idea is to double check that all replicas have relevant data prior
+// to calling MarkFinished. That approach would work in both data replication
+// and global operation cases.
+//
+// EDGE CASE 1: What if the first SYNC replica is alive, receives the delta
+// object, while the second SYNC replica is dead? (replication clients are
+// stored in a vector and accessed one by one)
```
gitbuda, line 1178: Partial failure, again it's better to do nothing, if we try to do undo, we actually make the situation worse, so don't do anything here.

spacejam:
Today during the replication sync with @gitbuda and @42jeremy a couple subtle correctness issues came up which I'll capture here. The high-level take-away is that it is incorrect to abort / roll back any update in response to when a sync replica becomes unavailable if there is a chance that the sync replica may have actually received and committed that transaction despite its response to main failing to arrive. The correct action is to keep trying to send the transaction to the replica, and if any other transactions are processed, they must be replicated to the replica in the transactional WAL order. Only after we reestablish communication with the replica (or an operator reconfigures replication to exclude that replica) can we start responding to clients with successful committed responses.

The most important feature of replication is that it allows us to have machines fail without sacrificing the transactional guarantees that we have communicated to users. There are other nice features, like providing more opportunities for scaling up stale read throughput, but correctness in the face of machine failures is the most important one. In the case of synchronous replication, we actually trade lower availability for higher correctness guarantees, because the main server must stop making forward progress any time any of its replicas becomes unavailable. If any server can be expected to be unavailable for writes 1% of the time, and we have a main server with two replica servers operating in sync mode, the overall system will be unavailable for writes ~2.98% of the time, because the the overall system cannot make progress unless the entire system can make progress. This will be fixed later on with a consensus protocol that is able to make progress despite >=1 machine failures.

As part of our consistency guarantees that we make to users (currently snapshot isolation by default, combined with synchronous replication in this case), if a user ever receives a response that their transaction has committed, any data that was created as part of their transaction must have been present (unless it was changed or overwritten by a later transaction) on a replica that is promoted to main. If a user ever receives a response that their transaction has aborted, no data that was created as part of that transaction may ever become visible to a future transaction. If data from an aborted transaction becomes visible, that is a "dirty read" anomaly that snapshot isolation is supposed to prevent.

The situation to avoid (there are multiple correctness bugs that occur below):
1. client submits transaction that writes item 1, W1
2. main logs the transaction to WAL and the WAL is shipped to the replica
3. the replica receives the transaction and applies it to its local state
4. the connection between the replica and main goes away due to normal network weather
5. (bug 1) main doesn't hear back from the replica, so it aborts or rolls-back the transaction that is currently committed on the replica 
6. (bug 2) the client receives a response saying their transaction failed
7. (bug 3) main fails, the previously unavailable replica is chosen as the new main
8. the client that expected its previous transaction to have been aborted now reads the new main, only to find that the previously written data is actually present, resulting in a dirty read anomaly

* bug 1 is a bug because it causes history to diverge between main and the replica
* bug 2 is a bug because the client was told a lie about the state of their transaction given the reality of the overall distributed system
* bug 3 is a bug because it's assumed that the other replica actually received the abort/rollback, so the replica with a more out of date history was chosen as the new main

Given the fact that we have agreed to sacrifice availability for correctness with sync replication, it's totally fine to simply do nothing as we wait for the unavailable replica to come back. We can also accept new work, as long as any replica receiving the replicated WAL will receive all transactions in-order, and as long as we do not allow later transactions to be committed before earlier ones. "Being committed" does not mean that the client itself receives a message that its message is committed, it means that any effects of the transaction at all have started to become visible, possibly only after a replica promotion to main.

It's OK for a client to receive nothing, as long as they don't receive an abort that is later visible anyway due to replica promotion. 

From a performance perspective, we don't have to completely stop accepting transactions. But we should not ever tell any client that their transaction has committed until their corresponding replicated WAL entries have reached the configured sync replica servers, as well as the WAL entries for all previous transactions. We should not undo/roll-back a commit that may have already been replicated. It's fine to let client connections time-out, but it's probably better to let them hang as we try to reestablish connectivity with the replica. The important part is that whenever we restart the entire cluster at once, there is never a recovered transaction that was actually aborted, and there is never a transaction that a client received a committed response for that is no longer present. This means that when we choose who should be main and who should be the replicas, we go with the main that has recovered the highest contiguous WAL coordinates. And ideally a batched fsync happens on the WAL at some point before the client is sent any success responses. These conditions don't need to be met if the user has configured the system to run in async replication mode, but fsync needs to happen for sync replication before the client is communicated a successful response.

42jeremy:
Closing PR as https://github.com/memgraph/memgraph/pull/435 is ready.

PR #388, opened 2022-04-29
[master <MG] Add improve docker-compose setup for running streams e2e tests locally

antaljanosbenjamin:
[master < Task] PR
- [ ] Check, and update documentation if necessary - **NOT NEEDED**
- [ ] Update [changelog](https://docs.memgraph.com/memgraph/changelog) - **NOT NEEDED**


kostasrim (approved):

tests/e2e/streams/kafka.yml
```diff
@@ -1,13 +1,13 @@
-version: "3"
+version: '3.7'
```
kostasrim, line 1: Why `3.7` ?

tests/e2e/streams/kafka.yml
```diff
@@ -1,13 +1,13 @@
...
 services:
   zookeeper:
-    image: 'bitnami/zookeeper:3.6.3-debian-10-r33'
+    image: 'bitnami/zookeeper:latest'
```
jbajic, line 4: I don't think it's good to pull the latest versions of services, especially with docker service since the latest can easily introduce breaking changes. Plus we don't do that in our infrastructure as well we have them pinned.

tests/e2e/streams/kafka.yml
antaljanosbenjamin: It was like this in redpanda docs, I thought specifying a stricter version is okay.

tests/e2e/streams/kafka.yml
antaljanosbenjamin: yes, in our infrastructure we have them pinned. However these docker files are intended to be used by developers locally, it doesn't matter if it breaks from time to time in my opinion. However it has the possibility to catch bugs in our streams implementations with newer versions of the services.

jbajic (approved):

PR #392, opened 2022-05-05
[master < T0642-MG]  add limit batches option to start stream query

42jeremy:
[master < Task] PR

**[Docs PR](https://github.com/memgraph/docs/pull/406)**


include/mg_procedure.h
```diff
@@ -16,6 +16,9 @@
 
 #ifdef __cplusplus
 extern "C" {
+#define MGP_ENUM_CLASS enum class
```
42jeremy, line 19: This file should be reverted, changes belong to another branch?

src/integrations/kafka/consumer.cpp
```diff
@@ -13,6 +13,7 @@
 
 #include <algorithm>
 #include <chrono>
+#include <cstdint>
```
42jeremy, line 16: Is this really needed?

42jeremy, line 16: Is this really needed?

src/integrations/kafka/consumer.cpp
```diff
@@ -367,7 +367,15 @@ void Consumer::StartConsuming() {
       }
       const auto &batch = maybe_batch.GetValue();
 
-      if (batch.empty()) continue;
+      if (batch.empty()) {
+        continue;
+      }
+
+      if (limit_batches.has_value() && nOfBatchesProcessed >= limit_batches.value()) {
```
42jeremy, line 374: This is probably better after the try{}.

42jeremy, line 374: This is probably better after the try{}.

src/integrations/pulsar/consumer.cpp
```diff
@@ -251,6 +253,12 @@ void Consumer::StartConsuming() {
         continue;
       }
 
+      if (limit_batches.has_value() && nOfBatchesProcessed >= limit_batches.value()) {
+        spdlog::info("Kafka consumer {} has reached the number of batches to process.", info_.consumer_name);
```
42jeremy, line 257: "Kafka" -> "Pulsar"

42jeremy, line 257: "Kafka" -> "Pulsar"

src/integrations/pulsar/consumer.cpp
```diff
@@ -251,6 +253,12 @@ void Consumer::StartConsuming() {
         continue;
       }
 
+      if (limit_batches.has_value() && nOfBatchesProcessed >= limit_batches.value()) {
```
42jeremy, line 256: This is probably better after the try{}.

42jeremy, line 256: This is probably better after the try{}.

src/integrations/kafka/consumer.hpp
```diff
@@ -111,13 +111,12 @@ class Consumer final : public RdKafka::EventCb {
   /// Starts consuming messages.
   ///
   /// This method will start a new thread which will poll all the topics for messages.
+  ////
+  /// @param limit_batches the consumer will only consume the given number of batches. If not present, a default value
```
42jeremy, line 115: Which default?

src/integrations/pulsar/consumer.cpp
```diff
@@ -274,11 +286,13 @@ void Consumer::StartConsuming() {
       spdlog::info("Pulsar consumer {} finished processing", info_.consumer_name);
     }
     is_running_.store(false);
+    remaining_nof_batches_to_read_.reset();
```
42jeremy, line 289: As for the Kafka consumer this can give data race issue. This should be removed.

src/query/stream/common.hpp
```diff
@@ -51,7 +53,7 @@ concept Stream = requires(TStream stream) {
   typename TStream::StreamInfo;
   typename TStream::Message;
   TStream{std::string{""}, typename TStream::StreamInfo{}, ConsumerFunction<typename TStream::Message>{}};
-  { stream.Start() } -> std::same_as<void>;
+  { stream.Start(std::optional<int64_t>{}) } -> std::same_as<void>;
```
42jeremy, line 56: Is this really needed?

src/query/stream/streams.cpp
```diff
@@ -571,12 +571,13 @@ void Streams::RestoreStreams() {
       MG_ASSERT(status.name == stream_name, "Expected stream name is '{}', but got '{}'", status.name, stream_name);
 
       try {
+        const auto tmp = status.info.common_info.batch_limit;
```
42jeremy, line 574: Better naming?
Can't we use directly from "it"?

42jeremy: -> No need anymore, stream with batch_limit won't be restored

src/integrations/kafka/consumer.cpp
```diff
@@ -367,37 +366,58 @@ void Consumer::StartConsuming() {
...
+      spdlog::warn("Error happened in consumer {} while fetching messages: {}!", info_.consumer_name,
+                   maybe_batch.GetError());
+      break;
+    }
+    const auto &batch = maybe_batch.GetValue();
+
+    if (batch.empty()) {
+      continue;
+    }
+    ++i;
```
antaljanosbenjamin, line 408:
```suggestion
  for (int64_t batch_count = 0; i < limit_batches;) {
    auto maybe_batch = GetBatch(*consumer_, info_, is_running_);
    if (maybe_batch.HasError()) {
      spdlog::warn("Error happened in consumer {} while fetching messages: {}!", info_.consumer_name,
                   maybe_batch.GetError());
      break;
    }
    const auto &batch = maybe_batch.GetValue();

    if (batch.empty()) {
      continue;
    }
    ++batch_count;
```

src/integrations/kafka/consumer.cpp
```diff
@@ -450,4 +470,40 @@ void Consumer::ConsumerRebalanceCb::rebalance_cb(RdKafka::KafkaConsumer *consume
...
+    }
+    if (const auto err = consumer_->position(partitions); err != RdKafka::ERR_NO_ERROR) {
+      throw ConsumerCheckFailedException(info_.consumer_name,
+                                         fmt::format("Couldn't get offsets from librdkafka {}", RdKafka::err2str(err)));
+    }
+    if (const auto err = consumer_->commitSync(partitions); err != RdKafka::ERR_NO_ERROR) {
+      spdlog::warn("Committing offset of consumer {} failed: {}", info_.consumer_name, RdKafka::err2str(err));
+      return false;
+    }
+  } catch (const std::exception &e) {
```
antaljanosbenjamin, line 502: And of course I wouldn't catch the exceptions here, but only in `StartConsuming`. In `StartConsumingWithLimit` you can leave the exceptions to propagate into the query engine to make it visible to the users.

src/integrations/pulsar/consumer.cpp
```diff
@@ -277,11 +275,66 @@ void Consumer::StartConsuming() {
...
+}
+
 void Consumer::StopConsuming() {
   is_running_.store(false);
   if (thread_.joinable()) {
     thread_.join();
   }
 }
 
+bool Consumer::TryToConsumeBatch(const std::vector<Message> &batch) const {
```
antaljanosbenjamin, line 318: I would use exception based error handling here also because of the same reason as in the Kafka producer (it propagates to the users in the blocking calls).

src/query/interpreter.cpp
```diff
@@ -652,12 +652,25 @@ Callback HandleStreamQuery(StreamQuery *stream_query, const Parameters &paramete
...
+                                    fmt::format("Started stream {}.", stream_query->stream_name_));
+
+      } else {
+        callback.fn = [interpreter_context, stream_name = stream_query->stream_name_, batch_limit]() {
+          interpreter_context->streams.StartWithLimit(stream_name, batch_limit.value());
+          return std::vector<std::vector<TypedValue>>{};
+        };
+        notifications->emplace_back(
+            SeverityLevel::INFO, NotificationCode::START_STREAM,
+            fmt::format("Started stream {} with batch_limit {}.", stream_query->stream_name_, batch_limit.value()));
```
antaljanosbenjamin, line 672: I think this is ont necessary, maybe confusing. This notification will be delivered to the users when the query execution is done. They will see the results, they will know the stream has started. Therefore I think this notification might be a bit confusing.

src/query/interpreter.cpp
```diff
@@ -695,7 +708,8 @@ Callback HandleStreamQuery(StreamQuery *stream_query, const Parameters &paramete
       return callback;
     }
     case StreamQuery::Action::SHOW_STREAMS: {
-      callback.header = {"name", "type", "batch_interval", "batch_size", "transformation_name", "owner", "is running"};
+      callback.header = {"name", "type", "batch_size", "batch_interval", "transformation_name", "owner", "is running"};
```
antaljanosbenjamin, line 711: The `batch_interval` is pushed to the result first, so I don't understand this change.

src/query/stream/streams.cpp
```diff
@@ -639,6 +639,18 @@ void Streams::Start(const std::string &stream_name) {
...
 }
 
+void Streams::StartWithLimit(const std::string &stream_name, int64_t batch_limit) const {
+  auto locked_streams = streams_.ReadLock();
+  auto it = GetStream(*locked_streams, stream_name);
+
+  std::visit(
+      [&](auto &&stream_data) {
+        auto stream_source_ptr = stream_data.stream_source->ReadLock();
+        stream_source_ptr->StartWithLimit(batch_limit);
```
antaljanosbenjamin, line 649:
```suggestion
        locked_streams.reset();
        stream_source_ptr->StartWithLimit(batch_limit);
```

You have to release the lock on `streams_`, otherwise it might block deleting/adding other streams. See how `Check` is implemented.

tests/e2e/streams/common.py
```diff
@@ -129,28 +131,25 @@ def validate_info(actual_stream_info, expected_stream_info):
...
+    return f"persistent://public/default/{topic}"
 
 
-def test_start_and_stop_during_check(
-        operation,
-        connection,
-        stream_creator,
-        message_sender,
-        already_stopped_error):
+def test_start_and_stop_during_check(operation, connection, stream_creator, message_sender, already_stopped_error):
```
antaljanosbenjamin, line 152: I think it is easy to parameterize this test to call `CHECK STREAM` and `START STREAM .. LIMIT_BATCHES`. Maybe you can give it a try.

tests/e2e/streams/common.py
```diff
@@ -255,31 +245,56 @@ def call_operation(counter):
...
     end = time.time()
 
-    assert (end - start) < 1.3 * \
-        timeout_ms, "The START STREAM was blocked too long"
+    assert (end - start) < 1.3 * timeout_ms, "The START STREAM was blocked too long"
     assert get_is_running(cursor, "test_stream")
     stop_stream(cursor, "test_stream")
+
+
+def test_start_stream_with_batch_limit(connection, stream_creator, messages_sender):
```
antaljanosbenjamin, line 271: I would add one or two more tests to make sure calling `CHECK` while `START STREAM ... LIMIT_BATCHES` is resulting in the correct exception and not a crash. The second test would be the other way around (calling `START STREAM ... LIMIT_BATCHES` while `CHECK` is running), but it might be done in one tests with parameters.

tests/e2e/streams/kafka_streams_tests.py
```diff
@@ -439,27 +372,39 @@ def test_info_procedure(kafka_topics, connection):
...
+    query = "CALL mg.transformations() YIELD * WITH name WHERE name STARTS WITH '" + transformation + "' RETURN name"
+    result = common.execute_and_fetch_all(cursor, query)
     assert len(result) == 1
-    assert result[0][0] == "c_transformations." + transformation
-    
+    assert result[0][0] == transformation
+
+
+@pytest.mark.parametrize("transformation", TRANSFORMATIONS_TO_CHECK_PY)
+def test_start_stream_with_batch_limit(kafka_producer, kafka_topics, connection, transformation):
```
antaljanosbenjamin, line 396:
```suggestion
def test_start_stream_with_batch_limit(kafka_producer, kafka_topics, connection):
```

Does this test need to be parameterized based on the transformation? I think different transformations shouldn't make a difference in the test case.

tests/e2e/streams/pulsar_streams_tests.py
```diff
@@ -380,13 +330,30 @@ def test_service_url(pulsar_client, pulsar_topics, connection, transformation):
...
         producer.send(common.SIMPLE_MSG)
 
     for topic in pulsar_topics:
-        check_vertex_exists_with_topic_and_payload(
-            cursor, topic, common.SIMPLE_MSG)
+        check_vertex_exists_with_topic_and_payload(cursor, topic, common.SIMPLE_MSG)
+
+
+@pytest.mark.parametrize("transformation", TRANSFORMATIONS_TO_CHECK)
+def test_start_stream_with_batch_limit(pulsar_client, pulsar_topics, connection, transformation):
```
antaljanosbenjamin, line 342:
```suggestion
def test_start_stream_with_batch_limit(pulsar_client, pulsar_topics, connection):
```

tests/unit/integrations_kafka_consumer.cpp
```diff
@@ -178,13 +178,7 @@ TEST_F(ConsumerTest, BatchInterval) {
...
-  auto start = [&consumer](const bool use_conditional) {
-    if (use_conditional) {
-      consumer.StartIfStopped();
-    } else {
-      consumer.Start();
-    }
-  };
+  auto start = [&consumer]() { consumer.Start(); };
 
   auto stop = [&consumer](const bool use_conditional) {
```
antaljanosbenjamin, line 183: I think this can be eliminated also. Why have you kept it? Just forgot it, or I missed something?

tests/unit/integrations_kafka_consumer.cpp
```diff
@@ -178,13 +178,7 @@ TEST_F(ConsumerTest, BatchInterval) {
...
   Consumer consumer{CreateDefaultConsumerInfo(), kDummyConsumerFunction};
 
-  auto start = [&consumer](const bool use_conditional) {
-    if (use_conditional) {
-      consumer.StartIfStopped();
-    } else {
-      consumer.Start();
-    }
-  };
+  auto start = [&consumer]() { consumer.Start(); };
```
antaljanosbenjamin, line 181: It might make sense to fully remove this lambda and just use `consumer.Start()` instead.

tests/unit/integrations_kafka_consumer.cpp
```diff
@@ -496,8 +484,68 @@ TEST_F(ConsumerTest, ConsumerStatus) {
...
+TEST_F(ConsumerTest, LimitBatches_SendingMoreThanLimit) {
+  /*
+  We send more messages than the BatchSize*LimitBatches:
+  -Consumer should receive 2*3=6 messages.
+  -Consumer should not be running afterwards.
+  */
+  static constexpr auto kBatchSize = 2;
+  static constexpr auto kLimitBatches = 3;
+  static constexpr auto kNOfMessagesToSend = 20;
+  static constexpr auto kNOfMessagesExpected = kBatchSize * kLimitBatches;
```
antaljanosbenjamin, line 520:
```suggestion
  static constexpr auto kNumberOfMessagesToSend = 20;
  static constexpr auto kNumberOfMessagesExpected = kBatchSize * kLimitBatches;
```

tests/unit/integrations_kafka_consumer.cpp
```diff
@@ -496,8 +484,68 @@ TEST_F(ConsumerTest, ConsumerStatus) {
...
+  for (auto sent_messages = 0; sent_messages <= kNOfMessagesToSend; ++sent_messages) {
+    cluster.SeedTopic(kTopicName, kMessage);
+  }
+
+  consumer->StartWithLimit(kLimitBatches);
+
+  EXPECT_FALSE(consumer->IsRunning());
+  EXPECT_EQ(nOfMessagesReceived, kNOfMessagesExpected);
+  EXPECT_TRUE(expected_messages_received) << "Some unexpected message have been received";
+}
```
antaljanosbenjamin, line 551: I think you pre-commit hook is not set up properly or you are not on the latest master. It should fix this missing new line.

antaljanosbenjamin (changes requested):

antaljanosbenjamin (changes requested):
One thing I forgot to mention: we need to be able to set a timeout for the blocking call similarly to `CHECK STREAM` to make sure the command doesn't hang forever if there are no messages arriving.

src/integrations/kafka/consumer.cpp
```diff
@@ -367,37 +366,58 @@ void Consumer::StartConsuming() {
...
-        }
-        if (const auto err = consumer_->position(partitions); err != RdKafka::ERR_NO_ERROR) {
-          throw ConsumerCheckFailedException(
-              info_.consumer_name, fmt::format("Couldn't get offsets from librdkafka {}", RdKafka::err2str(err)));
-        }
-        if (const auto err = consumer_->commitSync(partitions); err != RdKafka::ERR_NO_ERROR) {
-          spdlog::warn("Committing offset of consumer {} failed: {}", info_.consumer_name, RdKafka::err2str(err));
-          break;
-        }
-      } catch (const std::exception &e) {
```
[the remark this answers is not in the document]
42jeremy: Normally I have not changed the behavior, the catch is still present in `TryToConsumeBatch()` which has the same exceptions/catching logic that the previous `StartConsuming` had

tests/unit/integrations_kafka_consumer.cpp
42jeremy: Short version: it checks that both `Consumer::Stop` and `Consumer::StopIfRunning`work as expected.
Longer version: my understanding is that there are two flavors to stop a stream:
- `Consumer::Stop`: 
  - Raises an exception if the stream is not running. 
  - Stops the stream if it is running.
  - Available to the outside world.
-`Consumer::StopIfRunning`: this stops the stream if it is running. 
  - Stops the stream if it is running.
  - Available to the outside world.
  - Only used in the destructor of `Consumer` and some unit tests.
`StopIfRunning` could be removed if we move its logic directly in the desctructor, then there is no need to check the `StopIfRunning` flavor and this test can be simplified.

antaljanosbenjamin: I missed its usage in the destructor. Then it is fine.

src/query/interpreter.cpp
42jeremy: No idea what was behind this change, I'll revert.

tests/e2e/streams/common.py
42jeremy: Test `test_start_and_stop_during_check` seems quite complex already, I am wondering/afraid parameterizing it would make it even more complex and hard to maintain.
I can still give it a try however. What's your call on that?

src/query/stream/common.hpp
```diff
@@ -52,6 +52,7 @@ concept Stream = requires(TStream stream) {
   typename TStream::Message;
   TStream{std::string{""}, typename TStream::StreamInfo{}, ConsumerFunction<typename TStream::Message>{}};
   { stream.Start() } -> std::same_as<void>;
+  { stream.StartWithLimit(int64_t{}, std::optional<std::chrono::milliseconds>{}) } -> std::same_as<void>;
```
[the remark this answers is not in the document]
kostasrim: and all the other occurrences  of it

tests/unit/integrations_kafka_consumer.cpp
```diff
@@ -496,8 +482,88 @@ TEST_F(ConsumerTest, ConsumerStatus) {
...
 }
+
+TEST_F(ConsumerTest, LimitBatches_CannotStartIfAlreadyRunning) {
+  static constexpr auto kLimitBatches = 3;
+
+  auto info = CreateDefaultConsumerInfo();
+
+  auto consumer_function = [](const std::vector<Message> &messages) mutable {};
+
+  auto consumer = CreateConsumer(std::move(info), std::move(consumer_function));
```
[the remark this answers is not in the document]
kostasrim:
```
class consumer_function {
public:
    void operator()(const std::vector<Message> &messages) { };
};
```

consumer_function fun;

Does calling  std::move on this case have  any effect ?

src/integrations/pulsar/consumer.cpp
```diff
@@ -277,6 +307,49 @@ void Consumer::StartConsuming() {
...
+  const auto timeout_to_use = timeout.value_or(kDefaultCheckTimeout);
+  const auto start = std::chrono::steady_clock::now();
+
+  for (int64_t batch_count = 0; batch_count < limit_batches;) {
+    const auto now = std::chrono::steady_clock::now();
+    if (now - start >= timeout_to_use) {
+      throw ConsumerCheckFailedException(info_.consumer_name, "Timeout reached");
+    }
+
+    auto maybe_batch = GetBatch(consumer_, info_, is_running_, last_message_id_);
```
kostasrim, line 325:
```suggestion
    const auto maybe_batch = GetBatch(consumer_, info_, is_running_, last_message_id_);
```

kostasrim (changes requested):
Can you also add a test for cases where the limit literal is negative?

tests/unit/integrations_kafka_consumer.cpp
```diff
@@ -496,8 +482,88 @@ TEST_F(ConsumerTest, ConsumerStatus) {
...
+  info.batch_size = kBatchSize;
+  info.batch_interval = kBatchInterval;
+
+  static constexpr std::string_view kMessage = "LimitBatchesTestMessage";
+
+  auto expected_messages_received = true;
+  auto number_Of_Messages_Received = 0;
+  auto consumer_function = [&expected_messages_received,
+                            &number_Of_Messages_Received](const std::vector<Message> &messages) mutable {
+    number_Of_Messages_Received += messages.size();
```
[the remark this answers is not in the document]
jbajic:
```suggestion
    number_of_messages_received += messages.size();
```

tests/e2e/streams/common.py
```diff
@@ -255,31 +249,181 @@ def call_operation(counter):
...
+    kStreamName = "test"
+    kBatchLimit = 5
```
antaljanosbenjamin, line 277: Could you please sneak_casify the constant names? Similarly to the other PR, you can use `FULL_CAP` to emphasize the constness, but it is not required.

tests/e2e/streams/common.py
```diff
@@ -255,31 +249,181 @@ def call_operation(counter):
...
+
+def test_start_stream_with_batch_limit_timeout(connection, stream_creator):
+    # We check that we get the expected exception when trying to run START STREAM while providing TIMEOUT and not BATCH_LIMIT
+    kStreamName = "test"
+
+    cursor = connection.cursor()
+    execute_and_fetch_all(cursor, stream_creator(kStreamName))
+
+    with pytest.raises(mgclient.DatabaseError):
+        execute_and_fetch_all(cursor, f"START STREAM {kStreamName} TIMEOUT 3000")
```
antaljanosbenjamin, line 315: Maybe consider putting some asserts on the elapsed time, so it won't be like 2-3-4 times more than the timeout.

tests/e2e/streams/common.py
```diff
@@ -255,31 +249,181 @@ def call_operation(counter):
...
+def test_start_stream_with_batch_limit_reaching_timeout(connection, stream_creator):
+    # We check that we get the expected exception when running START STREAM while providing TIMEOUT and BATCH_LIMIT
+    kStreamName = "test"
+    kBatchLimit = 5
+
+    cursor = connection.cursor()
+    execute_and_fetch_all(cursor, stream_creator(kStreamName))
+
+    with pytest.raises(mgclient.DatabaseError):
+        execute_and_fetch_all(cursor, f"START STREAM {kStreamName} BATCH_LIMIT {kBatchLimit} TIMEOUT 3000")
```
antaljanosbenjamin, line 327: And here put an assert on only a small interval has been spent. I think without it this test might succeed by getting an exception because of timeout.

tests/e2e/streams/common.py
```diff
@@ -255,31 +249,181 @@ def call_operation(counter):
...
+        execute_and_fetch_all(cursor, f"CHECK STREAM {stream_name} BATCH_LIMIT {batch_limit} TIMEOUT {timeout}")
+
+    def start_new_stream_with_limit(stream_name, batch_limit, timeout):
+        connection = connect()
+        cursor = connection.cursor()
+        start_stream_with_limit(cursor, stream_name, batch_limit, timeout=timeout)
+
+    kStreamName = "test_check_and_batch_limit"
+    kBatchLimit = 1
+    kTimeout = 10000
```
antaljanosbenjamin, line 347: Same here as above.

tests/e2e/streams/common.py
```diff
@@ -255,31 +249,181 @@ def call_operation(counter):
...
+    kBatchLimit = 5
+
+    cursor = connection.cursor()
+    execute_and_fetch_all(cursor, stream_creator(kStreamName))
+
+    with pytest.raises(mgclient.DatabaseError):
+        execute_and_fetch_all(cursor, f"START STREAM {kStreamName} BATCH_LIMIT {kBatchLimit} TIMEOUT 3000")
+
+
+def test_start_stream_with_batch_limit__while_check_running(
```
antaljanosbenjamin, line 330:
```suggestion
def test_start_stream_with_batch_limit_while_check_running(
```
Could you please also do the same for the rest of the tests?

tests/e2e/streams/common.py
antaljanosbenjamin: Please do the same for the rest of the tests also.

tests/e2e/streams/common.py
```diff
@@ -255,31 +249,181 @@ def call_operation(counter):
...
+    thread_stream_running = Process(
+        target=start_new_stream_with_limit, daemon=True, args=(kStreamName, kBatchLimit, kTimeout)
+    )
+    thread_stream_running.start()
+    time.sleep(2)
+    assert get_is_running(cursor, kStreamName)
+
+    with pytest.raises(mgclient.DatabaseError):
+        execute_and_fetch_all(cursor, f"CHECK STREAM {kStreamName} BATCH_LIMIT {kBatchLimit} TIMEOUT {kTimeout}")
+
```
antaljanosbenjamin, line 414: Maybe assert on the elapsed time to increase the chance that the exception is really thrown because the stream is already running, and not because of timeout.

tests/e2e/streams/common.py
```diff
@@ -255,31 +249,181 @@ def call_operation(counter):
...
+    # 2/ Afterwards, we terminate the START STREAM with BATCH_LIMIT and start a CHECK STREAM
+    def start_new_stream_with_limit(stream_name, batch_limit, timeout):
+        connection = connect()
+        cursor = connection.cursor()
+        start_stream_with_limit(cursor, stream_name, batch_limit, timeout=timeout)
+
+    def start_check_stream(stream_name, batch_limit, timeout):
+        connection = connect()
+        cursor = connection.cursor()
+        execute_and_fetch_all(cursor, f"CHECK STREAM {stream_name} BATCH_LIMIT {batch_limit} TIMEOUT {timeout}")
```
antaljanosbenjamin, line 395: Based on the comment above I assume this should be `START STREAM ... BATCH_LIMIT ...`, and not `CHECK STREAM`.

tests/e2e/streams/common.py
```diff
@@ -255,31 +249,181 @@ def call_operation(counter):
...
+    kStreamName = "test"
+
+    cursor = connection.cursor()
+    execute_and_fetch_all(cursor, stream_creator(kStreamName))
+
+    with pytest.raises(mgclient.DatabaseError):
+        execute_and_fetch_all(cursor, f"START STREAM {kStreamName} TIMEOUT 3000")
+
+
+def test_start_stream_with_batch_limit_reaching_timeout(connection, stream_creator):
```
antaljanosbenjamin, line 318: As these tests are depending on the fact that `BATCH_SIZE` is `1` for the created streams, I would either put a comment about this expectation around every `stream_creator` for which this stands or make the batch size into a parameter.

tests/e2e/streams/common.py
```diff
@@ -255,31 +249,181 @@ def call_operation(counter):
...
+    # 2/
+    thread_stream_check = Process(target=start_check_stream, daemon=True, args=(kStreamName, kBatchLimit, kTimeout))
+    thread_stream_check.start()
+    time.sleep(2)
+    assert get_is_running(cursor, kStreamName)
+
+    message_sender(SIMPLE_MSG)
+    time.sleep(2)
+
+    assert not get_is_running(cursor, kStreamName)
```
antaljanosbenjamin, line 429: Maybe add an assert to make sure that the stream didn't die because of a timeout.

antaljanosbenjamin (changes requested):
Nicely done! There are small changes needs to be done in my opinion, but the big picture looks good!

src/integrations/pulsar/consumer.cpp
```diff
@@ -137,6 +156,23 @@ void Consumer::Start() {
   StartConsuming();
 }
 
+void Consumer::StartWithLimit(int64_t limit_batches, std::optional<std::chrono::milliseconds> timeout) const {
+  if (is_running_) {
+    throw ConsumerRunningException(info_.consumer_name);
+  }
+  if (limit_batches < kDefaultStartBatchLimit) {
+    throw ConsumerStartFailedException(
+        info_.consumer_name, fmt::format("Batch limit has to be greater than or equal to {}", kDefaultStartBatchLimit));
```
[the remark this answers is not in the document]
42jeremy: Actually there was a typo. I was referring to timeout in the second case.
Thanks for pointing it out!

tests/e2e/streams/kafka_streams_tests.py
```diff
@@ -360,10 +305,7 @@ def test_set_offset(kafka_producer, kafka_topics, connection, transformation):
     cursor = connection.cursor()
     common.execute_and_fetch_all(
         cursor,
-        "CREATE KAFKA STREAM test "
-        f"TOPICS {kafka_topics[0]} "
-        f"TRANSFORM {transformation} "
-        "BATCH_SIZE 1",
+        "CREATE KAFKA STREAM test " f"TOPICS {kafka_topics[0]} " f"TRANSFORM {transformation} " "BATCH_SIZE 1",
```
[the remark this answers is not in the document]
42jeremy: (These changes were automatically done by the pre-commit reformatting) 

I have already adapted the f-string in the PR https://github.com/memgraph/memgraph/pull/394

tests/e2e/streams/pulsar_streams_tests.py
```diff
@@ -37,30 +34,23 @@ def test_simple(pulsar_client, pulsar_topics, connection, transformation):
     cursor = connection.cursor()
     common.execute_and_fetch_all(
         cursor,
-        "CREATE PULSAR STREAM test "
-        f"TOPICS '{','.join(pulsar_topics)}' "
-        f"TRANSFORM {transformation}",
+        "CREATE PULSAR STREAM test " f"TOPICS '{','.join(pulsar_topics)}' " f"TRANSFORM {transformation}",
```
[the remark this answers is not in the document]
42jeremy: Same as https://github.com/memgraph/memgraph/pull/392#discussion_r882803598

tests/e2e/streams/pulsar_streams_tests.py
```diff
@@ -70,9 +60,7 @@ def test_separate_consumers(
         stream_names.append(stream_name)
         common.execute_and_fetch_all(
             cursor,
-            f"CREATE PULSAR STREAM {stream_name} "
-            f"TOPICS {topic} "
-            f"TRANSFORM {transformation}",
+            f"CREATE PULSAR STREAM {stream_name} " f"TOPICS {topic} " f"TRANSFORM {transformation}",
```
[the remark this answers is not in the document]
42jeremy: Same as https://github.com/memgraph/memgraph/pull/392#discussion_r882803598

42jeremy:
> Can you also add a test for cases where the limit literal is negative?

-> `test_start_stream_with_batch_limit_with_invalid_batch_limit`

src/integrations/kafka/consumer.cpp
```diff
@@ -74,6 +74,36 @@ utils::BasicResult<std::string, std::vector<Message>> GetBatch(RdKafka::KafkaCon
 
   return std::move(batch);
 }
+
+void CheckAndDestroyLastAssignmentIfNeeded(RdKafka::KafkaConsumer &consumer, const ConsumerInfo &info,
```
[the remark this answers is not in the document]
kostasrim: it does but I did not see it :P

src/query/interpreter.cpp
```diff
@@ -655,12 +655,26 @@ Callback HandleStreamQuery(StreamQuery *stream_query, const Parameters &paramete
...
+
+      if (!batch_limit.has_value()) {
+        callback.fn = [interpreter_context, stream_name = stream_query->stream_name_]() {
+          interpreter_context->streams.Start(stream_name);
+          return std::vector<std::vector<TypedValue>>{};
+        };
+        notifications->emplace_back(SeverityLevel::INFO, NotificationCode::START_STREAM,
+                                    fmt::format("Started stream {}.", stream_query->stream_name_));
+      } else {
+        if (batch_limit.has_value() && batch_limit.value() < 0) {
```
antaljanosbenjamin, line 669: Because you are in the else branch of the if, `batch_limit.has_value()` is always `true`. Maybe if you rephrase the condition it is a bit more readable:
```
if (batch_limit.has_value()) {
  if (batch_limit.value() < 0) {
  // ...
  }
} else {
  // handle the case when batch limit doesn't have a value.
}
```

src/query/interpreter.cpp
```diff
@@ -729,10 +743,16 @@ Callback HandleStreamQuery(StreamQuery *stream_query, const Parameters &paramete
...
+
+      const auto batch_limit = GetOptionalValue<int64_t>(stream_query->batch_limit_, evaluator);
+      if (batch_limit.has_value() && batch_limit.value() < 0) {
+        throw utils::BasicException("Parameter BATCH_LIMIT cannot hold negative value");
+      }
+
       callback.fn = [interpreter_context, stream_name = stream_query->stream_name_,
                      timeout = GetOptionalValue<std::chrono::milliseconds>(stream_query->timeout_, evaluator),
-                     batch_limit = GetOptionalValue<int64_t>(stream_query->batch_limit_, evaluator)]() mutable {
+                     &batch_limit]() mutable {
```
antaljanosbenjamin, line 755: To be honest I am not sure why this haven't caused any issues, but in my opinion this would be a danling reference.
```suggestion
                     batch_limit]() mutable {
```

src/integrations/pulsar/consumer.cpp
```diff
@@ -14,6 +14,7 @@
 #include <fmt/format.h>
 #include <pulsar/Client.h>
 #include <pulsar/InitialPosition.h>
+#include <algorithm>
 
 #include <chrono>
 #include <thread>
```
antaljanosbenjamin, line 20:
```suggestion
#include <algorithm>
#include <chrono>
#include <thread>

#include <fmt/format.h>
#include <pulsar/Client.h>
#include <pulsar/InitialPosition.h>
```
Sorry, I haven't noticed this before, but it is quite straightforward.

src/integrations/pulsar/consumer.cpp
```diff
@@ -97,6 +98,24 @@ pulsar_client::Client CreateClient(const std::string &service_url) {
   conf.setLogger(new SpdlogLoggerFactory);
   return {service_url, conf};
 }
+
+template <PulsarConsumer TConsumer>
+bool TryToConsumeBatch(TConsumer &consumer, const ConsumerInfo &info, const ConsumerFunction &consumer_function,
```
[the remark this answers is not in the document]
antaljanosbenjamin: I think you haven't addressed this comment, or I missed something.

tests/e2e/streams/common.py
antaljanosbenjamin: Can you please help me here, because I am not sure whether you addressed this or not?

antaljanosbenjamin (changes requested):
Almost done!

tests/unit/integrations_kafka_consumer.cpp
```diff
@@ -496,8 +482,88 @@ TEST_F(ConsumerTest, ConsumerStatus) {
...
+  EXPECT_TRUE(expected_messages_received) << "Some unexpected message have been received";
+}
+
+TEST_F(ConsumerTest, LimitBatches_Timeout_Reached) {
+  // We do not send any messages, we expect an exeption to be thrown.
+  static constexpr auto kLimitBatches = 3;
+
+  auto info = CreateDefaultConsumerInfo();
+
+  auto consumer_function = [](const std::vector<Message> &messages) {};
```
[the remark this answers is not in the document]
42jeremy: There's already one existing, I'll replace in the new tests as well as the test(s) I inspired from.

src/integrations/pulsar/consumer.cpp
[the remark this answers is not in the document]
42jeremy: You are right, I missed it. I only adapted the caller side so far. I'll update the method itself.

src/query/interpreter.cpp
```diff
@@ -652,12 +652,26 @@ Callback HandleStreamQuery(StreamQuery *stream_query, const Parameters &paramete
       return callback;
     }
     case StreamQuery::Action::START_STREAM: {
-      callback.fn = [interpreter_context, stream_name = stream_query->stream_name_]() {
-        interpreter_context->streams.Start(stream_name);
-        return std::vector<std::vector<TypedValue>>{};
-      };
-      notifications->emplace_back(SeverityLevel::INFO, NotificationCode::START_STREAM,
-                                  fmt::format("Started stream {}.", stream_query->stream_name_));
+      const auto batch_limit = GetOptionalValue<int64_t>(stream_query->batch_limit_, evaluator);
```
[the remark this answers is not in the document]
42jeremy: Reason explained: https://github.com/memgraph/memgraph/pull/392#discussion_r882504219
I've made the changes so limit_batches it cast to an unsigned int later.

kostasrim (approved):

PR #394, opened 2022-05-10
[master < T0829-MG] Adding "raw message" column to the result return by CHECK STREAM query

42jeremy:
[master < Task] PR

**[Docs PR](https://github.com/memgraph/docs/pull/400)**


src/query/stream/streams.cpp
```diff
@@ -43,6 +43,7 @@ extern const Event MessagesConsumed;
 namespace memgraph::query::stream {
 namespace {
 inline constexpr auto kExpectedTransformationResultSize = 2;
+inline constexpr auto kCheckStreamResultSize = 3;
```
antaljanosbenjamin, line 46: I would write
```suggestion
inline constexpr auto kCheckStreamResultSize = kExpectedTransformationResultSize + 1;
```

src/query/stream/streams.cpp
```diff
@@ -724,12 +725,16 @@ TransformationResult Streams::Check(const std::string &stream_name, std::optiona
           auto accessor = interpreter_context->db->Access();
           CallCustomTransformation(transformation_name, messages, result, accessor, *memory_resource, stream_name);
 
-          for (auto &row : result.rows) {
+          for (auto idx = 0; idx < result.rows.size(); ++idx) {
+            const auto &row = result.rows[idx];
+            const auto &message = messages[idx];
```
antaljanosbenjamin, line 730: This made me think. If I am not mistaken, it is not guaranteed that the number of queries are exactly the same as the number of messages:
 - in case of merging multiple updates into one query, the number of queries might be less than the number of messages
 - in case of "multi step" updates, the number of queries might be greater than the number of messages
 
I propose to yield `max(result.rows.size(), messages.size())` rows from the `CHECK` query properly handling the possible differences in the number of messages/queries.

antaljanosbenjamin (changes requested):
I think you mentioned that you checked how mgconsole/mgclient handles this. Can you show an example?

src/query/stream/streams.cpp
tonilastre: Correct, it is not guaranteed. There can be `N` messages and `M` queries. It all depends on the user and how the transformation is defined.

42jeremy:
> I think you mentioned that you checked how mgconsole/mgclient handles this. Can you show an example?

Examples while giving {"\x04", "\0x04","^D", "\^D", ":quit"} as raw messages attached.

test_with_0x04
test_with_^D
test_with_^D2
test_with_:quit
test_with_x04

Let me know if you have other tests ideas.

jbajic (approved):
Look good to me!

src/query/stream/streams.cpp
42jeremy: So I have implemented as follow:
1. size(queries) == size(messages): We display 1 message per query (current implementation).
2. We create a list from all messages and show only for the first query.

An example of what it would look like if the transformation would do some filtering (thus create less queries than it has received messages):
example_transformation_filtering_messages

kostasrim: @42jeremy I don't really like this because in the second case the last column of the first row is filled with a bunch of messages and the rest is plain null -- it doesn't reason well to me to have the last column of each row (other than the first) being null. Consider Benjo's approach.

@antaljanosbenjamin even with your approach, if `result.rows.size() < messages.size()` then max would produce as many rows as the size of the messages. Why not limit the number of raw messages to the size of the result queries? 

All in all, I think there is a general problem with the current approach: we don't know (and we can't know) the relationship between the number of messages and the queries. Therefore, why don't we separate these by adding a `CHECK STREAM RAW` statement instead (which basically returns N rows containing N raw messages from the stream)?

antaljanosbenjamin: > Why not limit the number of raw messages to the size of the result queries?

Two reasons:
1. Most probably the query and the message on the same line are not related to each other.
2. It might show a query that was created from a message that is not shown, thus really hard to debug. In this way we at least show everything in the "same" order, so the user can debug it based on the actual data.

> All in all, I think there is a general problem with the current approach: we don't know (and we can't know) the relationship between the number of messages and the queries.

100 percent agree.

> Therefore, why don't we separate these by adding a CHECK STREAM RAW statement instead (which basically returns N rows containing N raw messages from the stream)?

We can do that, if that is better in your opinion, however that would require the user to run two queries always, not talking about taking care of the same parameters for the `CHECK STREAM` and `CHECK STREAM RAW` queries.

With the current approach, if the `message -> query` is 1:1 mapping, the user gets a super nice view. If not, than get a not that good view, but at least it is done by one query (without caring about the parameters).

Maybe you can discuss this tomorrow or on Friday on daily, or on weekly.

antaljanosbenjamin: During discussing this on daily the main talking points were the following:

1. Even when the number of messages and queries are same, the 1-to-1 mapping is not guaranteed. It is possible that one message might not result in any query and another message might result in two queries. Also, considering 1-to-1 mapping, the order of the messages and queries might not be the same. Therefore showing 1 query and 1 message in a single row might not make sense.
2. Instead of having the messages aggregated into a list when the number of messages and queries don't match, we might want to yield `max(result.rows.size(), messages.size())` rows, but we couldn't decide on that as we didn't see any clear advantage of them.
3. Having two queries (`CHECK STREAM` and `CHECK STREAM RAW`) does not guarantee that the queries are generated from the messages the user sees. If the batch "closes" because of batch interval (the longest time until which a message should wait for other messages to include in the same batch), then the `CHECK STREAM` and `CHECK STREAM RAW` queries might experience different batches. The lower the batch interval, the likely it will happen.

Because of these we have the following questions:
1. Does showing 1 message and query in 1 line makes sense when their numbers are equal?
2. Should we **always** show one message per line, or should we aggregate them? See example below.
3. Should we have separate queries for getting the message and getting the queries from the transformation?

### Example for #2 question

Always show 1 message per line:
```
+--------------------------+------------+----------------------+
| query                    | parameters | raw messages         |
+--------------------------+------------+----------------------+
| "CREATE (v {prop: 'a'})" | NULL       | "a"                  |
+--------------------------+------------+----------------------+
| "CREATE (v {prop: 'b'})" | NULL       | "b"                  |
+--------------------------+------------+----------------------+
| "CREATE (v {prop: 'c'})" | NULL       | "don't create query" |
+--------------------------+------------+----------------------+
| NULL                     | NULL       | "c"                  |
+--------------------------+------------+----------------------+
```
Or when the number of message is less than the queries:

```
+--------------------------+------------+----------------------+
| query                    | parameters | raw messages         |
+--------------------------+------------+----------------------+
| "CREATE (v {prop: 'a'})" | NULL       | "a"                  |
+--------------------------+------------+----------------------+
| "CREATE (v {prop: 'a'})" | NULL       | "b"                  |
+--------------------------+------------+----------------------+
| "CREATE (v {prop: 'b'})" | NULL       | NULL                 |
+--------------------------+------------+----------------------+
| "CREATE (v {prop: 'b'})" | NULL       | NULL                 |
+--------------------------+------------+----------------------+
```

Versus the aggregated versions:
```
+--------------------------+------------+--------------------------------------+
| query                    | parameters | raw messages         |
+--------------------------+------------+--------------------------------------+
| "CREATE (v {prop: 'a'})" | NULL       | ["a", "b", "don't create query", "c] |
+--------------------------+------------+--------------------------------------+
| "CREATE (v {prop: 'b'})" | NULL       | NULL                                 |
+--------------------------+------------+--------------------------------------+
| "CREATE (v {prop: 'c'})" | NULL       | NULL                                 |
+--------------------------+------------+--------------------------------------+
```
and
```
+--------------------------+------------+----------------------+
| query                    | parameters | raw messages         |
+--------------------------+------------+----------------------+
| "CREATE (v {prop: 'a'})" | NULL       | ["a", "b"]           |
+--------------------------+------------+----------------------+
| "CREATE (v {prop: 'a'})" | NULL       | NULL                 |
+--------------------------+------------+----------------------+
| "CREATE (v {prop: 'b'})" | NULL       | NULL                 |
+--------------------------+------------+----------------------+
| "CREATE (v {prop: 'b'})" | NULL       | NULL                 |
+--------------------------+------------+----------------------+
```

42jeremy:
Example when filtering out some values in the transformation:
filtering_a_bit

Example when filtering out all values in the transformation:
filtering_everything

src/query/stream/streams.cpp
antaljanosbenjamin: We have settled to use the format @42jeremy presented in [his comment](https://github.com/memgraph/memgraph/pull/394#issuecomment-1126184318). The main reason is even with the same number of queries and messages, we cannot be sure about the mapping. By providing batches of queries and messages in each row, we trust the user to do the mapping by themself.

tests/e2e/streams/kafka_streams_tests.py
```diff
@@ -117,30 +103,27 @@ def test_start_from_last_committed_offset(
...
 @pytest.mark.parametrize("transformation", TRANSFORMATIONS_TO_CHECK_PY)
-def test_check_stream(
-        kafka_producer,
-        kafka_topics,
-        connection,
-        transformation):
+def test_check_stream(kafka_producer, kafka_topics, connection, transformation):
     assert len(kafka_topics) > 0
+    kBatchSize = 1
+    kIndexOfFirstBatch = 0
```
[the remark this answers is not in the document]
antaljanosbenjamin: Could you please rename the rest of the constants you added in the same way (regardless of the version you picked)?

tests/e2e/streams/common.py
```diff
@@ -255,31 +248,156 @@ def call_operation(counter):
...
     end = time.time()
 
-    assert (end - start) < 1.3 * \
-        timeout_ms, "The START STREAM was blocked too long"
+    assert (end - start) < 1.3 * timeout_ms, "The START STREAM was blocked too long"
     assert get_is_running(cursor, "test_stream")
     stop_stream(cursor, "test_stream")
+
+
+def test_check_stream__same_nOf_queries_than_messages(connection, stream_creator, message_sender):
```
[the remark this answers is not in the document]
antaljanosbenjamin: Could you please also do the same changes in the regarding kafka/pulsar tests?

antaljanosbenjamin (changes requested):

src/query/stream/streams.cpp
```diff
@@ -724,15 +725,27 @@ TransformationResult Streams::Check(const std::string &stream_name, std::optiona
...
-            result_row.push_back(std::move(query));
-            result_row.push_back(std::move(parameters));
+          auto result_row = std::vector<TypedValue>();
+          result_row.reserve(kCheckStreamResultSize);
 
-            test_result.push_back(std::move(result_row));
-          }
+          auto queries_and_parameters = std::vector<TypedValue>(result.rows.size());
+          std::transform(
+              result.rows.cbegin(), result.rows.cend(), queries_and_parameters.begin(), [&](const auto &row) {
```
kostasrim, line 733: This is a bug. `queries_and_parameters.size()` will be zero. Add a `back_inserter` :)

src/query/stream/streams.cpp
```diff
@@ -724,15 +725,27 @@ TransformationResult Streams::Check(const std::string &stream_name, std::optiona
...
+                auto [query, parameters] = ExtractTransformationResult(row.values, transformation_name, stream_name);
+
+                return std::map<std::string, TypedValue>{{"query", std::move(query)},
+                                                         {"parameters", std::move(parameters)}};
+              });
+          result_row.emplace_back(std::move(queries_and_parameters));
+
+          auto messages_list = std::vector<TypedValue>(messages.size());
+          std::transform(messages.cbegin(), messages.cend(), messages_list.begin(), [](const auto &message) {
+            return std::string_view(message.Payload().data(), message.Payload().size());
```
kostasrim, line 743: `return std::string(....) `;

kostasrim (changes requested):

tests/e2e/streams/common.py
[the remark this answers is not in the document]
42jeremy: Will do!

src/query/stream/streams.cpp
antaljanosbenjamin: Why `std::string`? It copies the data from payload, then when the TypedValue is constructed, it will also copy it into a `std::pmr::string`. With the `string_view` you can avoid the first copy.

tests/e2e/streams/kafka_streams_tests.py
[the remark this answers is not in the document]
42jeremy: (I have capitalized all constants in the file as well as in common.py and pulsarèstreams_tests.py)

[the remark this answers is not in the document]
42jeremy: (I have capitalized all constants in the file as well as in common.py and pulsarèstreams_tests.py)

src/query/stream/streams.cpp
42jeremy: Conclusion is we'll go with `std::string_view`.
From @antaljanosbenjamin -> `TypedValue` doesn't support construction with not null terminated strings, so we have to construct the a `pmr::string` , that will be moved into the `TypedValue` that sits inside the vector. However if somebody changes that vector to pmr::vector, that might still result in copying `pmr::string`. With `std::string_view`, we copy a small object (if compiler doesn't optimize the whole stuff away, as `std::transform`, `TypedValue` and also `pmr::string` is a template type implemented in header files), but get rid of the complexity of allocators. In my view this is not that a big issue (we are not on the hot path), so I think the gained safety worth more than the performance we could gain with `pmr::string`.

kostasrim (approved):

antaljanosbenjamin (approved):

PR #398, opened 2022-05-14
Add default commented values for Bolt SSL flags

gitbuda:
The idea here is to simplify how someone can start playing with Bolt SSL config (providing where are the autogenerated SSL files).


gitbuda:
The only issue I see here is related to the different operating systems. If Memgraph starts supporting, e.g., Windows natively, the generator will have to change to contain the valid path. Since Memgraph is not there yet, I would leave the config as-is (there are other Linux-specific paths).

How can we start documenting all the changes that should be done once we decide to support the native Windows build?

antaljanosbenjamin (approved):

antaljanosbenjamin:
I think documenting what should be done for a native windows build might be a bit overwhelming without being sure that we will get there. Currently we have tons of POSIX/Linux specific stuff that I think we will have troubles to go native on Windows.

jbajic (approved):

PR #399, opened 2022-05-19
[master < T0869-MG] Kafka check stream incorrectly committing offset 

42jeremy:
[master < Task] PR


42jeremy:
The problem is that Kafka cannot provide any assignment until it has consumed at least one message. 

Issue won't be solve now as it is pretty complex and minor issue.
The problem

There is an implementation that solve the specific issue of the test. This is not enough however as setting the offset to 0 is not correct.
There could be a way to solve this by doing a reset when exiting the Check by resetting the offset of each partitions: we know how many messages have been consumed, we can know from which partition/topic; so for each partition we could set the offset being "currentOffset-nOfMessagesConsumedForThatPartition". This requires some changes in the Check method as well as the GetBatch method.

PR #400, opened 2022-05-19
[E112-MG < T0848-MG] Add initial schema implementation

jbajic:
Introducing only in-memory representation of schema, without any support for creating or validating schema against vertices.

[epic < Task] PR

### DIFF
--- a/src/storage/v2/schemas.cpp
+++ b/src/storage/v2/schemas.cpp
@@ -9,7 +9,6 @@
 // by the Apache License, Version 2.0, included in the file
 // licenses/APL.txt.
 
-#include <tuple>
 #include <unordered_map>
 #include <utility>
 #include <vector>
@@ -19,40 +18,53 @@
 
 namespace memgraph::storage {
 
-Schemas::CreationStatus Schemas::CreateSchema(
-    const LabelId label, const std::vector<std::pair<PropertyId, PropertyValue::Type>> &property_types) {
-  auto res = schemas_.insert({label, property_types}).second;
+SchemaViolation::SchemaViolation(ValidationStatus status, LabelId label) : status{status}, label{label} {}
+
+SchemaViolation::SchemaViolation(ValidationStatus status, LabelId label, SchemaType violated_type)
+    : status{status}, label{label}, violated_type{violated_type} {}
+
+SchemaViolation::SchemaViolation(ValidationStatus status, LabelId label, SchemaType violated_type,
+                                 PropertyValue violated_property_value)
+    : status{status}, label{label}, violated_type{violated_type}, violated_property_value{violated_property_value} {}
+
+Schemas::CreationStatus Schemas::CreateSchema(const LabelId primary_label,
+                                              const std::vector<SchemaType> &schemas_types) {
+  auto res = schemas_.insert({primary_label, schemas_types}).second;
   return res ? Schemas::CreationStatus::SUCCESS : Schemas::CreationStatus::FAIL;
 }
 
-Schemas::DeletionStatus Schemas::DeleteSchema(const LabelId label) {
-  auto res = schemas_.erase(label);
+Schemas::DeletionStatus Schemas::DeleteSchema(const LabelId primary_label) {
+  auto res = schemas_.erase(primary_label);
   return res != 0 ? Schemas::DeletionStatus::SUCCESS : Schemas::DeletionStatus::FAIL;
 }
 
-Schemas::ValidationStatus Schemas::ValidateVertex(const LabelId primary_label, const Vertex &vertex) {
+[[nodiscard]] std::optional<SchemaViolation> Schemas::ValidateVertex(const LabelId primary_label,
+                                                                     const Vertex &vertex) {
+  // TODO Check for multiple defined primary labels
+  std::vector<std::pair<SchemaType, PropertyValue>> type_violations;
   if (!schemas_.contains(primary_label)) {
-    return Schemas::ValidationStatus::NO_SCHEMA_DEFINED_FOR_LABEL;
+    return SchemaViolation(SchemaViolation::ValidationStatus::NO_SCHEMA_DEFINED_FOR_LABEL, primary_label);
   }
   if (!utils::Contains(vertex.labels, primary_label)) {
-    return Schemas::ValidationStatus::VERTEX_HAS_NO_PRIMARY_LABEL;
+    return SchemaViolation(SchemaViolation::ValidationStatus::VERTEX_HAS_NO_PRIMARY_LABEL, primary_label);
   }
 
-  for (auto &[property_id, property_type] : schemas_[primary_label]) {
-    // Property existence check
-    if (!vertex.properties.HasProperty(property_id)) {
-      return Schemas::ValidationStatus::VERTEX_HAS_NO_PROPERTY;
+  for (auto &schema_type : schemas_[primary_label]) {
+    if (!vertex.properties.HasProperty(schema_type.property_id)) {
+      return SchemaViolation(SchemaViolation::ValidationStatus::VERTEX_HAS_NO_PROPERTY, primary_label, schema_type);
     }
     // Property type check
-    if (vertex.properties.GetProperty(property_id).type() != property_type) {
-      return Schemas::ValidationStatus::VERTEX_PROPERTY_WRONG_TYPE;
+    //  TODO Can this be replaced with just property id check?
+    if (auto vertex_property = vertex.properties.GetProperty(schema_type.property_id);
+        PropertyValueTypeToSchemaType(vertex_property) != schema_type.type) {
+      return SchemaViolation(SchemaViolation::ValidationStatus::VERTEX_PROPERTY_WRONG_TYPE, primary_label, schema_type,
+                             vertex_property);
     }
   }
-
   // TODO after the introduction of vertex hashing introduce check for vertex
   // primary key uniqueness
 
-  return Schemas::ValidationStatus::SUCCESS;
+  return std::nullopt;
 }
 
 Schemas::SchemasList Schemas::ListSchemas() const {
--- a/src/storage/v2/schemas.cpp
+++ b/src/storage/v2/schemas.cpp
@@ -19,9 +19,9 @@
 
 namespace memgraph::storage {
 
-Schemas::CreationStatus Schemas::CreateSchema(const LabelId label,
-                                              const std::pair<PropertyId, PropertyValue::Type> &property_ids) {
-  auto res = schemas_.insert({label, property_ids}).second;
+Schemas::CreationStatus Schemas::CreateSchema(
+    const LabelId label, const std::vector<std::pair<PropertyId, PropertyValue::Type>> &property_types) {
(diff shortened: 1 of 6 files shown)


src/storage/v2/schemas.cpp
```diff
@@ -0,0 +1,79 @@
...
+SchemaViolation::SchemaViolation(ValidationStatus status, LabelId label, SchemaType violated_type)
+    : status{status}, label{label}, violated_type{violated_type} {}
+
+SchemaViolation::SchemaViolation(ValidationStatus status, LabelId label, SchemaType violated_type,
+                                 PropertyValue violated_property_value)
+    : status{status}, label{label}, violated_type{violated_type}, violated_property_value{violated_property_value} {}
+
+Schemas::CreationStatus Schemas::CreateSchema(const LabelId primary_label,
+                                              const std::vector<SchemaType> &schemas_types) {
+  auto res = schemas_.insert({primary_label, schemas_types}).second;
```
42jeremy, line 32:
```suggestion
  const auto res = schemas_.insert({primary_label, schemas_types}).second;
```

src/storage/v2/schemas.cpp
```diff
@@ -0,0 +1,79 @@
...
+  return res ? Schemas::CreationStatus::SUCCESS : Schemas::CreationStatus::FAIL;
+}
+
+Schemas::DeletionStatus Schemas::DeleteSchema(const LabelId primary_label) {
+  auto res = schemas_.erase(primary_label);
```
42jeremy, line 37:
```suggestion
  const auto res = schemas_.erase(primary_label);
```

src/storage/v2/schemas.cpp
```diff
@@ -0,0 +1,79 @@
...
+  return res != 0 ? Schemas::DeletionStatus::SUCCESS : Schemas::DeletionStatus::FAIL;
+}
+
+[[nodiscard]] std::optional<SchemaViolation> Schemas::ValidateVertex(const LabelId primary_label,
+                                                                     const Vertex &vertex) {
+  // TODO Check for multiple defined primary labels
+  std::vector<std::pair<SchemaType, PropertyValue>> type_violations;
```
42jeremy, line 44: type_violations is not used

src/storage/v2/schemas.cpp
```diff
@@ -0,0 +1,79 @@
...
+  if (!schemas_.contains(primary_label)) {
+    return SchemaViolation(SchemaViolation::ValidationStatus::NO_SCHEMA_DEFINED_FOR_LABEL, primary_label);
+  }
+  if (!utils::Contains(vertex.labels, primary_label)) {
+    return SchemaViolation(SchemaViolation::ValidationStatus::VERTEX_HAS_NO_PRIMARY_LABEL, primary_label);
+  }
+
+  for (auto &schema_type : schemas_[primary_label]) {
```
42jeremy, line 52:
```suggestion
  for (const auto &schema_type : schemas_[primary_label]) {
```

42jeremy (changes requested):

src/storage/v2/storage.hpp
```diff
@@ -402,6 +414,14 @@ class Storage final {
 
   ConstraintsInfo ListAllConstraints() const;
 
+  /// @throw std::bad_alloc
+  bool CreateSchema(LabelId primary_label, std::vector<SchemaType> &schemas_types);
+
+  /// @throw std::bad_alloc
+  bool DeleteSchema(LabelId primary_label);
```
antaljanosbenjamin, line 421:
```suggestion
  bool CreateSchema(LabelId primary_label, std::vector<SchemaType> &schemas_types);

  bool DeleteSchema(LabelId primary_label);
```

I know that other functions mention bad alloc, but I think that is not that useful. I mean you cannot handle `bad_alloc` in any proper way in my opinion.

src/storage/v2/schemas.hpp
```diff
@@ -0,0 +1,176 @@
...
+class SchemaViolationException : public utils::BasicException {
+  using utils::BasicException::BasicException;
+};
+
+struct SchemaType {
+  enum class Type : uint8_t { Bool, Int, Double, String, List, Map, Date, LocalTime, LocalDateTime, Duration };
+
+  Type type;
+  PropertyId property_id;
+  std::shared_ptr<SchemaType> nested_value;
```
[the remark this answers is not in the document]
jbajic: Hmm but how we will support Map and List then? e.g. Map[Map[int]], List[DATETIME]

src/storage/v2/schemas.hpp
```diff
@@ -0,0 +1,176 @@
...
+    }
+  }
+}
+
+inline std::string PropertyTypeToString(const PropertyValue &property_value) {
+  if (const auto schema_type = PropertyValueTypeToSchemaType(property_value); schema_type) {
+    return SchemaTypeToString(*schema_type);
+  }
+  return "Unknown";
+}
```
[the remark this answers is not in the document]
jbajic: Currently its unused, therefore I will remove it, but it is neccessary during validation, when property type does not match schema property, to inform user what exactly went wrong

src/storage/v2/schemas.hpp
```diff
@@ -0,0 +1,176 @@
...
+
+  Schemas() = default;
+  Schemas(const Schemas &) = delete;
+  Schemas(Schemas &&) = delete;
+  Schemas &operator=(const Schemas &) = delete;
+  Schemas &operator=(Schemas &&) = delete;
+  ~Schemas() = default;
+
+  enum class CreationStatus : uint8_t { SUCCESS, FAIL };
+  enum class DeletionStatus : uint8_t { SUCCESS, FAIL };
```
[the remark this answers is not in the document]
jbajic: I will replace it with bool, since I don't see any need for better messaging at this moment

src/storage/v2/schemas.cpp
```diff
@@ -0,0 +1,79 @@
...
+    if (!vertex.properties.HasProperty(schema_type.property_id)) {
+      return SchemaViolation(SchemaViolation::ValidationStatus::VERTEX_HAS_NO_PROPERTY, primary_label, schema_type);
+    }
+    // Property type check
+    //  TODO Can this be replaced with just property id check?
+    if (auto vertex_property = vertex.properties.GetProperty(schema_type.property_id);
+        PropertyValueTypeToSchemaType(vertex_property) != schema_type.type) {
+      return SchemaViolation(SchemaViolation::ValidationStatus::VERTEX_PROPERTY_WRONG_TYPE, primary_label, schema_type,
+                             vertex_property);
+    }
```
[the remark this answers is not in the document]
jbajic: Why would we keep the knowledge of which properties of schema have been matched? This way the first violation it finds it stops searcing for others and it returns a `SchemaViolation`. I am not sure I understand what you meant by deserializing only property values once?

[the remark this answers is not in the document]
jbajic: Why would we keep the knowledge of which schema properties have been matched? This way, the first violation it finds stops searching for others and returns a `SchemaViolation`. I am not sure I understand what you meant by deserializing only property values once?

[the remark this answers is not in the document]
antaljanosbenjamin: Currently every `GetProperty` call might deserialize some of the properties because they are serialized into a single buffer and if you want to deserialize the `n`th one, you have to kind of deserialize all of the properties before the `n`th. There I would only call `GetProperties` once and use the result of that. However I think this might be rewritten in a better when we modified the vertex creation (basically the schema can be checked before the vertex is created, so we might don't have to deal with `PropertyValue`, but only a single map of `TypedValue`/`PropertyValue`. So for now maybe it is better to leave it as is, just keep the serialization in mind.

src/storage/v2/schemas.hpp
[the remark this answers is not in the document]
antaljanosbenjamin: I see your point, but wasn't clear from the code. My biggest issue is the `shared_ptr`: it requires locking etc, so instead I would go with `unique_ptr` instead. However, there are other possibilities:
1. Using some kind of vector to store the nested types. `Map[List[str]]` would translate into a vector of `{Map, List, Str}`. The type to the right is the nested type. If we concerned about the heap allocation of the vector, we can be strict on the size with an enum class defined as `uint8_t` and using an `array<Type, 4>` to restrict ourselves up to 3 nested types (+1 root type) and keep the size under 32 bits.
2. Just dropping the support for complex types. To be honest I didn't see this as an issue, but if it makes our life harder in any sense, drop it. We can always reiterate on this in the future.
3. Using some kind of a global map and storing pointer kind of things similar to what is done with `gListTypes` in `mgp_type_list` in the MGP API.

In my mind the two best option is using an array or dropping support for complex types. What do you think?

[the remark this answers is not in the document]
jbajic: I am more in favor of dropping support for `Map` and `List`, since whatever is in these structures can always be flattened by the user.

antaljanosbenjamin (approved):
Apart from the extra member, it looks good!

src/storage/v2/schemas.cpp
```diff
@@ -0,0 +1,76 @@
...
+SchemaViolation::SchemaViolation(ValidationStatus status, LabelId label, SchemaProperty violated_type)
+    : status{status}, label{label}, violated_type{violated_type} {}
+
+SchemaViolation::SchemaViolation(ValidationStatus status, LabelId label, SchemaProperty violated_type,
+                                 PropertyValue violated_property_value)
+    : status{status}, label{label}, violated_type{violated_type}, violated_property_value{violated_property_value} {}
+
+bool Schemas::CreateSchema(const LabelId primary_label, const std::vector<SchemaProperty> &schemas_types) {
+  const auto res = schemas_.insert({primary_label, schemas_types}).second;
+  return res;
```
kostasrim, line 31:
```suggestion
  return schemas_.insert({primary_label, schemas_types}).second;
```

src/storage/v2/schemas.cpp
```diff
@@ -0,0 +1,76 @@
...
+}
+
+bool Schemas::DeleteSchema(const LabelId primary_label) {
+  const auto res = schemas_.erase(primary_label);
+  return res;
```
kostasrim, line 36:
```suggestion
  return schemas_.erase(primary_label);
```

src/storage/v2/storage.hpp
```diff
@@ -402,6 +414,12 @@ class Storage final {
 
   ConstraintsInfo ListAllConstraints() const;
 
+  bool CreateSchema(LabelId primary_label, std::vector<SchemaProperty> &schemas_types);
```
kostasrim, line 417: This is not implemented. Same as `DeleteSchema` below. Is this intended?

jbajic: Yes, it is intentional, it will be implemented in the next PR, since it will be used then

kostasrim: Then `vector` should be const? Anyway, leave it for now. On the next PR should be ok!

antaljanosbenjamin (approved):

42jeremy (approved):

PR #402, opened 2022-05-26
[master < MG] Run clang-tidy also on our header files

antaljanosbenjamin:
[master < Task] PR

All the credits goes to @jbajic noticing that we are not running clang-tidy on our headers.


antaljanosbenjamin:
This reverts #141. I don't know the original reason why the header files were ignored, but based on the branch `origin/MG-test-analysis` where Antonio printed a bunch of build folder related thing I assume the issue was clang-tidy didn't find the `compile_commands.json` file. However I don't remember the exact reason.

jbajic (approved):

42jeremy (approved):

spacejam (approved):

kostasrim (approved):

antaljanosbenjamin:
I realized why this is wrong. So `clang-tidy` understands the `compile_commands.json` file generated by CMake. Based on that, it can figure out the correct config for source files, but not for header files. Therefore clang-tidy is run on a header file, it will most probably fail, as it cannot figure out the config (compiler flags which includes extra include paths), so the check will fail. However if we change a source file that includes that header file, then the header file will be checked properly. So the solution in my opinion is do not filter out header files but using the `-regex` option of clang-tidy. This way the line filter that passed to clang-tidy (which determines which warnings should be suppressed or emitted) will contain the lines from the headers, but won't run the check on individual header files. This means if we only change a header file, but not any of the sources files that includes that, then we might not notice issues in headers. However I think the possibility for this is very low.

PR #404, opened 2022-05-26
[E112-MG < T0849-MG] Create schema DDL expressions

jbajic:
Add query expressions for schema manipulation.
[E112-MG < Task] PR


src/query/interpreter.cpp
```diff
@@ -820,6 +821,108 @@ Callback HandleSettingQuery(SettingQuery *setting_query, const Parameters &param
...
+          std::vector<std::string> primary_key_properties;
+          primary_key_properties.reserve(schema_types.size());
+          std::transform(schema_types.begin(), schema_types.end(), std::back_inserter(primary_key_properties),
+                         [&db](const auto &schema_type) {
+                           return db->PropertyToName(schema_type.property_id) +
+                                  "::" + storage::SchemaTypeToString(schema_type.type);
+                         });
+
+          schema_info_row.emplace_back(utils::Join(primary_key_properties, ", "));
+          schema_info_row.emplace_back(schema_types.size() == 1 ? "Single" : "Composite");
```
jbajic, line 850: Maybe remove this altogether, I don't think this provides any value to the user.

src/storage/v2/schemas.cpp
```diff
@@ -19,21 +19,39 @@
...
+
+bool Schemas::CreateSchema(const LabelId primary_label, const std::vector<SchemaPropertyType> &schemas_types) {
+  if (schemas_.contains(primary_label)) {
+    return false;
+  }
+  schemas_.insert({primary_label, schemas_types});
+  return true;
 }
 
+bool Schemas::DropSchema(const LabelId primary_label) { return schemas_.erase(primary_label); }
```
jbajic, line 53: Opted for Drop, since we use Create/Delete for data, and Create/Delete for indexes and constraints, it makes sense to use the same for schema

src/storage/v2/schemas.hpp
```diff
@@ -31,10 +32,8 @@ class SchemaViolationException : public utils::BasicException {
   using utils::BasicException::BasicException;
 };
 
-struct SchemaProperty {
-  enum class Type : uint8_t { Bool, Int, Double, String, Date, LocalTime, LocalDateTime, Duration };
-
-  Type type;
+struct SchemaPropertyType {
+  common::SchemaType type;
```
jbajic, line 36: SInce `SchemaProperty` has been removed from Schema definition, this was renamed to `SchemaType`, we had this discussion @antaljanosbenjamin

src/storage/v2/schemas.hpp
```diff
@@ -73,48 +73,52 @@ class Schemas {
   Schemas &operator=(Schemas &&) = delete;
   ~Schemas() = default;
 
-  [[nodiscard]] bool CreateSchema(LabelId label, const std::vector<SchemaProperty> &schemas_types);
+  [[nodiscard]] SchemasList ListSchemas() const;
 
-  [[nodiscard]] bool DeleteSchema(LabelId label);
+  [[nodiscard]] SchemasList GetSchema(LabelId primary_label) const;
```
jbajic, line 78: Don't return `SchemaList`, that is an interface lie, return a single schema, and the caller should create a list if he wants to

src/storage/v2/schemas.cpp
```diff
@@ -19,21 +19,39 @@
...
+SchemaViolation::SchemaViolation(ValidationStatus status, LabelId label, SchemaPropertyType violated_type,
                                  PropertyValue violated_property_value)
     : status{status}, label{label}, violated_type{violated_type}, violated_property_value{violated_property_value} {}
 
-bool Schemas::CreateSchema(const LabelId primary_label, const std::vector<SchemaProperty> &schemas_types) {
-  return schemas_.insert({primary_label, schemas_types}).second;
+Schemas::SchemasList Schemas::ListSchemas() const {
+  Schemas::SchemasList ret;
+  ret.reserve(schemas_.size());
+  for (const auto &[label_props, schema_property] : schemas_) {
```
42jeremy, line 32: std::transform instead of for loop

42jeremy (approved):

42jeremy (approved):

src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -2336,6 +2338,94 @@ antlrcpp::Any CypherMainVisitor::visitForeach(MemgraphCypher::ForeachContext *ct
...
+  }
+  return map;
+}
+
+antlrcpp::Any CypherMainVisitor::visitCreateSchema(MemgraphCypher::CreateSchemaContext *ctx) {
+  auto *schema_query = storage_->Create<SchemaQuery>();
+  schema_query->action_ = SchemaQuery::Action::CREATE_SCHEMA;
+  schema_query->label_ = AddLabel(ctx->labelName()->accept(this));
+  if (!ctx->schemaTypeMap()) {
+    throw SemanticException("Schema property map must exist!");
```
antaljanosbenjamin, line 2410: I think this error message might not be clear for the users. What about `Schema must contain at least one property and its type!`?

antaljanosbenjamin: Isn't the existence of schemaTypeMap is enforced by the grammar?

tests/unit/interpreter.cpp
```diff
@@ -1465,3 +1473,148 @@ TEST_F(InterpreterTest, LoadCsvClauseNotification) {
...
+    ASSERT_TRUE(primary_key_split == result_table[key1].first);
+    ASSERT_EQ(result[2].ValueString(), result_table[key1].second);
+
+    const auto &result2 = stream.GetResults().front();
+    ASSERT_EQ(result2.size(), 3U);
+    const auto key2 = result2[0].ValueString();
+    ASSERT_TRUE(result_table.contains(key2));
+    const auto primary_key_split2 = StringToUnorderedSet(result2[1].ValueString(), 2);
+    ASSERT_TRUE(primary_key_split2 == result_table[key2].first);
+    ASSERT_EQ(result2[2].ValueString(), result_table[key2].second);
```
antaljanosbenjamin, line 1567: I know this might be a bit nitpicking, so feel free to ignore this comment.

I think the lot of indices and `xxx2` variables might be a good reason to introduce a function for validating a single result row against the possible results (`result_table` in this case).

antaljanosbenjamin (changes requested):

src/storage/v2/schemas.hpp
jbajic: I would really like to avoid the name `Property` since it is heavily used it always represents a value, not a type. How about `DataType`? (inspired by postgres docs https://www.postgresql.org/docs/current/sql-createtable.html)

src/storage/v2/storage.cpp
```diff
@@ -1234,6 +1234,20 @@ SchemasInfo Storage::ListAllSchemas() const {
...
+    return {{*schema}};
+  }
+  return {};
+}
+
+bool Storage::CreateSchema(const LabelId primary_label, const std::vector<SchemaPropertyType> &schemas_types) {
+  return schemas_.CreateSchema(primary_label, schemas_types);
+}
+
+bool Storage::DropSchema(const LabelId primary_label) { return schemas_.DropSchema(primary_label); }
```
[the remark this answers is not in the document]
jbajic: I will create an airtable task for this

src/query/frontend/ast/cypher_main_visitor.cpp
jbajic: Ah yes you are right, this check is redundant

src/query/interpreter.cpp
```diff
@@ -2015,6 +2118,32 @@ PreparedQuery PrepareConstraintQuery(ParsedQuery parsed_query, bool in_explicit_
...
+                         if (!pull_plan) {
+                           auto results = handler();
+                           pull_plan = std::make_shared<PullPlanVector>(std::move(results));
+                         }
+
+                         if (pull_plan->Pull(stream, n)) {
+                           return action;
+                         }
+                         return std::nullopt;
+                       },
```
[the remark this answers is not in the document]
jbajic: I agree this has become quite a callback

src/storage/v2/schemas.hpp
jbajic: Sorry I misunderstood, I agree SchemaProperty is a better name for this struct

tests/unit/cypher_main_visitor.cpp
```diff
@@ -4234,47 +4235,82 @@ TEST_P(CypherMainVisitorTest, TestShowSchema) {
...
+                             std::vector<std::pair<std::string, memgraph::common::SchemaType>> properties_type,
+                             auto &ast_generator) {
+  EXPECT_EQ(schema_property_map.size(), properties_type.size());
+  for (const auto &[property_name, type] : properties_type) {
+    const auto property_ix = ast_generator.Prop(property_name);
+    const auto element = std::ranges::find_if(schema_property_map,
+                                              [property_ix](const auto &elem) { return elem.first == property_ix; });
+    EXPECT_TRUE(element != schema_property_map.end());
+    EXPECT_EQ(element->second, type);
+  }
```
antaljanosbenjamin, line 4251: Is the order of the types are expected to be the same as in the query? I know we discussed that this can be exploited by us in the early stages, but we didn't want to expose this knowledge to the users. I am asking because if it is an expected thing (even only for us), then we should be strict about the ordering here, and obviously if this is not an expected thing, then we don't care about the order (as it is implemented right now).

jbajic: I agree, I will also assert the ordering here

antaljanosbenjamin (approved):
LGTM!

PR #412, opened 2022-06-17
[master < T0880-MG] Add information on show replicas to express how uptodate is replica

42jeremy:

Documentation [PR](https://github.com/memgraph/docs/pull/439)

### DIFF
--- a/src/query/interpreter.cpp
+++ b/src/query/interpreter.cpp
@@ -232,6 +232,10 @@ class ReplQueryHandler final : public query::ReplicationQueryHandler {
         replica.timeout = *repl_info.timeout;
       }
 
+      replica.current_timestamp = repl_info.timestamp_info.current_timestamp;
+      replica.current_number_of_timestamp_behind_master =
+          repl_info.timestamp_info.current_number_of_timestamp_behind_master;
+
       return replica;
     };
 
@@ -497,7 +501,8 @@ Callback HandleReplicationQuery(ReplicationQuery *repl_query, const Parameters &
       return callback;
     }
     case ReplicationQuery::Action::SHOW_REPLICAS: {
-      callback.header = {"name", "socket_address", "sync_mode", "timeout"};
+      callback.header = {"name",    "socket_address",    "sync_mode",
+                         "timeout", "current_timestamp", "number_of_timestamp_behind_master"};
       callback.fn = [handler = ReplQueryHandler{interpreter_context->db}, replica_nfields = callback.header.size()] {
         const auto &replicas = handler.ShowReplicas();
         auto typed_replicas = std::vector<std::vector<TypedValue>>{};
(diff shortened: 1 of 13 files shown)


src/storage/v2/replication/replication_client.cpp
```diff
@@ -537,6 +537,33 @@ std::vector<Storage::ReplicationClient::RecoveryStep> Storage::ReplicationClient
...
+  try {
+    auto stream{rpc_client_->Stream<replication::TimestampRpc>()};
+    const auto response = stream.AwaitResponse();
+    const auto is_success = response.success;
+    if (!is_success) {
+      replica_state_.store(replication::ReplicaState::INVALID);
+      HandleRpcFailure();
+    }
+    auto main_time_stamp = storage_->last_commit_timestamp_.load();
+    info.current_timestamp = response.current_commit_timestamp;
```
antaljanosbenjamin, line 554:
```suggestion
    info.current_timestamp = main_time_stamp;
```
```suggestion
    info.current_timestamp = response.current_commit_timestamp;
```

src/query/interpreter.cpp
```diff
@@ -512,7 +518,8 @@ Callback HandleReplicationQuery(ReplicationQuery *repl_query, const Parameters &
     }
 
     case ReplicationQuery::Action::SHOW_REPLICAS: {
-      callback.header = {"name", "socket_address", "sync_mode", "timeout", "state"};
+      callback.header = {"name",    "socket_address",    "sync_mode",
+                         "timeout", "current_timestamp", "number_of_timestamp_behind_master", "state"};
```
antaljanosbenjamin, line 522:
```suggestion
                         "timeout", "current_timestamp_of_replica", "number_of_timestamp_behind_master", "state"};
```

As you suggested on slack, we should rename this to make it more clear to users.

antaljanosbenjamin (changes requested):

src/storage/v2/replication/replication_client.cpp
42jeremy: Original idea was to have info.current_timestamp containing the current timestamp of the replica (that returned the response). I think I disagree with the suggestion here.
(Note: attribute will be renamed for clarity as `current_timestamp` is definitely confusing)

antaljanosbenjamin (approved):

src/storage/v2/replication/replication_client.cpp
```diff
@@ -537,6 +537,33 @@ std::vector<Storage::ReplicationClient::RecoveryStep> Storage::ReplicationClient
...
+    const auto is_success = response.success;
+    if (!is_success) {
+      replica_state_.store(replication::ReplicaState::INVALID);
+      HandleRpcFailure();
+    }
+    auto main_time_stamp = storage_->last_commit_timestamp_.load();
+    info.current_timestamp_of_replica = response.current_commit_timestamp;
+    info.current_number_of_timestamp_behind_master = response.current_commit_timestamp - main_time_stamp;
+  } catch (const rpc::RpcFailedException &) {
+    {
```
kostasrim, line 557: Why the extra scope?

42jeremy: Nice catch there!

kostasrim (approved):

PR #414, opened 2022-06-20
[master < T0882-MG] Registering a replica with timeout 0 should not be allowed

42jeremy:

Doc [PR](https://github.com/memgraph/docs/pull/442)


tests/unit/utils_math.cpp
```diff
@@ -21,3 +21,127 @@ TEST(UtilsMath, Log2) {
...
+  const auto b = 1.0 - 2 * memgraph::utils::GetEpsilon();
+
+  EXPECT_TRUE(memgraph::utils::IsStrictlyGreater(a, b));
+}
+
+TEST(UtilsMath, IsStrictlyGreaterTooSmallDistance) {
+  const auto a = 1.0;
+  const auto b = 1.0 - memgraph::utils::GetEpsilon() / 2;
+
+  EXPECT_TRUE(!memgraph::utils::IsStrictlyGreater(a, b));
```
antaljanosbenjamin, line 111:
```suggestion
  EXPECT_FALSE(memgraph::utils::IsStrictlyGreater(a, b));
```

It is easy to not notice the exclamation mark, so being verbose here and using `EXPECT_FALSE(...)` instead of `EXPECT_TRUE(!...)` can be beneficial.

Even though in this case it doesn't really matter because of the simplicity of the tests, but if you have to modify something else on the PR, it might also worth to update these checks.

tests/unit/utils_math.cpp
```diff
@@ -21,3 +21,127 @@ TEST(UtilsMath, Log2) {
...
+  const auto b = 1.0;
+
+  EXPECT_TRUE(memgraph::utils::IsStrictlyLower(a, b));
+}
+
+TEST(UtilsMath, IsStrictlyLowerWithTooSmallDistance) {
+  const auto a = 1.0 - memgraph::utils::GetEpsilon() / 2;
+  const auto b = 1.0;
+
+  EXPECT_TRUE(!memgraph::utils::IsStrictlyLower(a, b));
```
antaljanosbenjamin, line 139:
```suggestion
  EXPECT_FALSE(memgraph::utils::IsStrictlyLower(a, b));
```

antaljanosbenjamin (approved):

src/query/interpreter.cpp
```diff
@@ -489,6 +490,10 @@ Callback HandleReplicationQuery(ReplicationQuery *repl_query, const Parameters &
       } else if (timeout.IsInt()) {
         maybe_timeout = static_cast<double>(timeout.ValueInt());
       }
+      if (maybe_timeout && memgraph::utils::IsLowerOrEqual(*maybe_timeout, 0.0)) {
+        throw utils::BasicException("Parameter TIMEOUT must be strictly greater than 0.0.");
```
kostasrim, line 494:
```suggestion
        throw utils::BasicException("Parameter TIMEOUT must be strictly greater than 0.");
```

src/utils/math.hpp
```diff
@@ -64,4 +67,44 @@ constexpr std::optional<uint64_t> RoundUint64ToMultiple(uint64_t val, uint64_t m
...
+  if (std::abs(dist) > epsilon) {
+    if (dist > 0) {
+      result = -1;
+    } else {
+      MG_ASSERT(dist < 0);
+      result = 1;
+    }
+  }
+
+  return result;
```
kostasrim, line 101:
```suggestion
  if (!(std::abs(dist) > epsilon)) {
    return 0;
  }
  if (dist > 0) {
     return -1;
  } 
  MG_ASSERT(dist < 0);
  return 1;
```

src/utils/math.hpp
```diff
@@ -64,4 +67,44 @@ constexpr std::optional<uint64_t> RoundUint64ToMultiple(uint64_t val, uint64_t m
   return (numerator / multiple) * multiple;
 }
 
+/*!
+ * @return The default precision used in the project. Two doubles are considered different if the distance between them
+ * is strictly greater than Epsilon.
+ */
+constexpr double GetEpsilon() noexcept { return 0.0000001; }
```
kostasrim, line 74: This is hardcoded. Why did you choose `0.0000001`?

src/utils/math.hpp
```diff
@@ -64,4 +67,44 @@ constexpr std::optional<uint64_t> RoundUint64ToMultiple(uint64_t val, uint64_t m
...
+ *
+ * @param a
+ * @param b
+ * @param epsilon The precision to use when comparing the two double. If dist(a,b)<=epsilon then a==b.
+ * @return an integer indicating whether a and b are equal.
+ *          +0: dist(a,b) <= epsilon
+ *          +1: a > b+epsilon
+ *          -1: a < b-epsilon
+ */
+constexpr int16_t CompareDouble(double a, double b, double epsilon = GetEpsilon()) noexcept {
```
kostasrim, line 86: Also, could you plz expand on why we need this? I get there is a precision error somewhere, but what exactly is it and how does it happen?

src/utils/math.hpp
42jeremy: Why `0.0000001`: To have a starting point.
To be 100% correct we should have 1 precision per unit (ex: Time, Distance, Weight, etc...) we handle in the application and 1 default configurable by end-user:
- If we know which unit we are handling, we should use the corresponding precision.
- If we can't guess which unit we are handling, use the default one.

I felt it was a bit too much for now. I'd suggest we handle this on `T0886: Double comparison should always include an epsilon. I only wanted to introduce the structure so I could use it.`

src/utils/math.hpp
42jeremy: > Also, could you plz expand on why we need this? I get there is a precision error because of the `doubles`, but what exactly is it and how does it happen? Can you give me an example?

Interesting question.

1. Overall one should not compare double as follow `myDouble == myOtherDouble`  because of the precision behind double. The following UT can fail for that reason:
```
TEST(UtilsMath, example) {
  // x = (value in arithmetic, value on computer) - https://www.binaryconvert.com/convert_double.html
  double a = 42.424242424242424242;           // a = (42.424242424242422, 42.4242424242424220892644370906)
  int b = 1 / memgraph::utils::GetEpsilon();  // b = 10000000
  double c = a / b;                           // c = (0,000004242, 0.00000424200000000000021249972267934 )
  double d = 0.0;                             // d = (0.0, 0.0)
  for (auto idx = 0; idx < b; ++idx) {
    d += c;
  }
  // d = (42.424242424242422, 42.4242424201029919572647486348)
  EXPECT_TRUE(memgraph::utils::CompareDouble(a, d) == 0);
  EXPECT_TRUE(a == d); // can fail
}
```
2. I do not need to check whether `a == b` in my case but `a <= b`, which can lead to the same problem:
-   `a= 1.0; b = 100.0;` so  `a < b` is `true`. Expected.
-   `a = 1.000000000000001; b = 1.0000000000000011;` so `a < b` is `true`. Is it really expected, is the difference due to the double rounding? Maybe we want to consider them equal in that case.
-   `a = 1.0000000000000011; b = 1.000000000000001;` so `a < b is false`. Same as above.

3. I realize we have problematic comparison in the code-base, for instance:
https://github.com/memgraph/memgraph/blob/7a2bbd4bb37920322a0097d964c2751027602e55/src/query/typed_value.cpp#L742

But this goes beyond the scope of this PR, I have created a task on Airtable to tackle it properly and independently.

kostasrim (approved):
Let's chat tomorrow for the doubles on daily!

src/utils/math.hpp
42jeremy: After discussion with the whole team: we will not keep it for now until it is proven to be a problem. The responsibility of comparing double correctly is on the side of the user and Memgraph only mimic the c++ comparison.

tests/e2e/interactive_mg_runner.py
```diff
@@ -96,6 +96,9 @@ def load_args():
 
 
 def _start_instance(name, args, log_file, queries, use_ssl, procdir):
+    assert not (
+        name in MEMGRAPH_INSTANCES.keys()
+    )  # If this raises, you are trying to start an instance with the same name than one already running.
```
jbajic, line 101: This is a bit strange ordering, why not:
```suggestion
    assert name not in MEMGRAPH_INSTANCES.keys()
      # If this raises, you are trying to start an instance with the same name than one already running.
```

tests/e2e/interactive_mg_runner.py
```diff
@@ -131,6 +134,14 @@ def stop(context, name):
     stop_all()
 
 
+def kill(context, name):
+    for key, _ in context.items():
+        if key != name:
+            continue
```
jbajic, line 140:
```suggestion
    for key in context.keys():
        if key != name:
            continue
```

PR #415, opened 2022-06-21
[master < T0837-MG] Save replication settings

42jeremy:

Doc [PR](https://github.com/memgraph/docs/pull/445)

### DIFF
--- a/src/storage/v2/storage.cpp
+++ b/src/storage/v2/storage.cpp
@@ -1944,13 +1944,12 @@ utils::BasicResult<Storage::RegisterReplicaError> Storage::RegisterReplica(
 
   auto client = std::make_unique<ReplicationClient>(std::move(name), this, endpoint, replication_mode, config);
 
-  utils::BasicResult<Storage::RegisterReplicaError> return_value;
   if (client->State() == replication::ReplicaState::INVALID) {
     if (replication::RegistrationMode::CAN_BE_INVALID != registration_mode) {
       return RegisterReplicaError::CONNECTION_FAILED;
     }
 
-    return_value = RegisterReplicaError::CONNECTION_FAILED;
+    spdlog::warn("Connection failed when registering replica {}. Replica will still be registered.", client->Name());
   }
 
   return replication_clients_.WithLock([&](auto &clients) -> utils::BasicResult<Storage::RegisterReplicaError> {
@@ -1967,7 +1966,7 @@ utils::BasicResult<Storage::RegisterReplicaError> Storage::RegisterReplica(
     }
 
     clients.push_back(std::move(client));
-    return return_value;
+    return {};
   });
 }
 
@@ -2043,16 +2042,10 @@ void Storage::RestoreReplicas() {
                         });
 
     if (ret.HasError()) {
-      if (RegisterReplicaError::CONNECTION_FAILED != ret.GetError()) {
-        // At recovery, only connection failed is allowed. Other errors are fatal.
-        LOG_FATAL("Failure when restoring replica {}: {}.", replica_name, ret.GetError());
-      } else {
-        spdlog::error("Failure when restoring replica {}: {}.", replica_name,
-                      RegisterReplicaErrorToString(ret.GetError()));
-      }
-    } else {
-      spdlog::info("Replica {} restored.", replica_name);
+      MG_ASSERT(RegisterReplicaError::CONNECTION_FAILED != ret.GetError());
+      LOG_FATAL("Failure when restoring replica {}: {}.", replica_name, RegisterReplicaErrorToString(ret.GetError()));
     }
+    spdlog::info("Replica {} restored.", replica_name);
   }
 }
 


42jeremy:
This needs to be merged after [#414](https://github.com/memgraph/memgraph/pull/414/)!

42jeremy:
With current state of code, application(s) crash(es) in the following scenario:
- Start two instances of Memgraph

antaljanosbenjamin (changes requested):

src/storage/v2/storage.cpp
```diff
@@ -50,6 +52,92 @@ using OOMExceptionEnabler = utils::MemoryTracker::OutOfMemoryExceptionEnabler;
...
+
+    const auto &key_file = data.at(kSSLKeyFile);
+    const auto &cert_file = data.at(kSSLCertFile);
+
+    MG_ASSERT(key_file.is_null() == cert_file.is_null());
+
+    if (!key_file.is_null()) {
+      replica_status.ssl = replication::ReplicationClientConfig::SSL{};
+      replica_status.ssl->key_file = key_file.get<typename decltype(replica_status.ssl->key_file)::value_type>();
+      replica_status.ssl->cert_file = cert_file.get<typename decltype(replica_status.ssl->cert_file)::value_type>();
```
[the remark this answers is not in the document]
antaljanosbenjamin: Ohh, I think e2e tests are only passing because we are not using SSL for replicas. Maybe consider extracting the logic for persisting/reloading replicas into a class that can be well unit tested without spinning up multiple processes. If that is not suitable in your opinion, than at least make the json <-> ReplicaStatus conversion functions public so they can be tested separately.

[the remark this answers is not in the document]
42jeremy: > This can be rewritten using `get_to` in my opinion, so we can avoid the `decltype` part.
> 
> Also, what does `typename decltype(replica_status.ssl->key_file)::value_type` mean? As far as I understodd:
> 
>     1. `decltype(replica_status.ssl)` is `std::optional<replication::ReplicationClientConfig::SSL>`
> 
>     2. Thus `decltype(replica_status.ssl->key_file)` is `std::string`
> 
>     3. Thus `decltype(replica_status.ssl->cert_file)::value_type` is `std::string::value_type`, which is basically some kind of a character
> 
> 
> So my question is what did I miss? Of if I didn't miss anything, then how does this work?

You're right here. Originally they were both optional but I have changed it to directly use `replication::ReplicationClientConfig::SSL` afterwards.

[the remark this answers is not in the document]
42jeremy: > Ohh, I think e2e tests are only passing because we are not using SSL for replicas. Maybe consider extracting the logic for persisting/reloading replicas into a class that can be well unit tested without spinning up multiple processes. If that is not suitable in your opinion, than at least make the json <-> ReplicaStatus conversion functions public so they can be tested separately.

The logic to parse has been moved to a separate file and is unit-tested: `src/storage/v2/replication/replication_persistance_helper.hpp` and `tests/unit/replication_persistance_helper.cpp`.
I have also added two unit-tests in `tests/unit/storage_v2_replication.cpp` that cover the registration and "un-registration" of replicas.

src/storage/v2/CMakeLists.txt
```diff
@@ -12,8 +12,7 @@ set(storage_v2_src_files
     vertex_accessor.cpp
     storage.cpp)
 
-##### Replication #####
-
+# #### Replication #####
```
kostasrim, line 15:
```suggestion
##### Replication #####
```

src/storage/v2/replication/replication_persistance_helper.cpp
```diff
@@ -0,0 +1,94 @@
...
+// licenses/APL.txt.
+
+#include "storage/v2/replication/replication_persistance_helper.hpp"
+#include "utils/logging.hpp"
+
+namespace {
+const std::string kReplicaName = "replica_name";
+const std::string kIpAddress = "replica_ip_address";
+const std::string kPort = "replica_port";
+const std::string kSyncMode = "replica_sync_mod";
```
kostasrim, line 19:
```suggestion
const std::string kSyncMode = "replica_sync_mode";
```

src/storage/v2/storage.cpp
```diff
@@ -1928,8 +1954,14 @@ utils::BasicResult<Storage::RegisterReplicaError> Storage::RegisterReplica(
   });
 }
 
-bool Storage::UnregisterReplica(const std::string_view name) {
+bool Storage::UnregisterReplica(const std::string &name) {
   MG_ASSERT(replication_role_.load() == ReplicationRole::MAIN, "Only main instance can unregister a replica!");
+  if (ShouldStoreAndRestoreReplicas()) {
+    if (!kvstorage_->Delete(name)) {
+      spdlog::error("Issue when removing replica {} from settings.", name);
```
kostasrim, line 1961:
```suggestion
      spdlog::error("Error when removing replica {} from settings.", name);
```

kostasrim (commented):
I left a few cosmetic comments. I do have more to add here on the behavioural side but I will do it tomorrow :)

src/storage/v2/CMakeLists.txt
```diff
@@ -26,10 +25,10 @@ set(storage_v2_src_files
     replication/replication_server.cpp
     replication/serialization.cpp
     replication/slk.cpp
+    replication/replication_persistance_helper.cpp
     ${lcp_storage_cpp_files})
 
-#######################
-
+# ######################
```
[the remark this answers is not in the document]
42jeremy: This was automatically done.  Next time someone will modify the file this will happen again.

tests/unit/replication_persistance_helper.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+                all_tests_succeeded = CheckEquality(replicas_status, *replicas_status_converted) && all_tests_succeeded;
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+
```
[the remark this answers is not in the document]
42jeremy: @jbajic Should I remove it?

src/storage/v2/replication/replication_persistance_helper.cpp
```diff
@@ -0,0 +1,94 @@
...
+const std::string kSyncMode = "replica_sync_mode";
+const std::string kTimeout = "replica_timeout";
+const std::string kCheckFrequency = "replica_check_frequency";
+const std::string kSSLKeyFile = "replica_ssl_key_file";
+const std::string kSSLCertFile = "replica_ssl_cert_file";
+}  // namespace
+
+namespace memgraph::storage::replication {
+
+nlohmann::json replica_status_to_json(ReplicaStatus &&status) {
```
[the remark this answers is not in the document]
42jeremy: Indeed, you are right. They were originally free functions and I have not renamed them afterwards.

tests/unit/replication_persistance_helper.cpp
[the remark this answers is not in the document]
jbajic: Well I don't have a strong opinion, maybe @antaljanosbenjamin and @kostasrim  could share theirs

tests/unit/replication_persistance_helper.cpp
```diff
@@ -0,0 +1,199 @@
...
+      if (status.ssl->key_file != other_status.ssl->key_file) {
+        return false;
+      }
+      if (status.ssl->cert_file != other_status.ssl->cert_file) {
+        return false;
+      }
+    }
+
+    return true;
+  }
```
kostasrim, line 107: These are not needed, just default the operator of the struct `ReplicaStatus` like:

```
friend bool operator==(const ReplicaStatus&, const ReplicaStatus&) const = default;
```

See https://en.cppreference.com/w/cpp/language/default_comparisons

kostasrim, line 107:
```suggestion

```

tests/unit/replication_persistance_helper.cpp
```diff
@@ -0,0 +1,199 @@
...
+                                                         .port = port,
+                                                         .sync_mode = sync_mode,
+                                                         .timeout = timeout,
+                                                         .replica_check_frequency = replica_check_frequency,
+                                                         .ssl = ssl};
+  }
+
+  bool CheckEquality(memgraph::storage::replication::ReplicaStatus &status,
+                     memgraph::storage::replication::ReplicaStatus &other_status) {
+    if (!CheckEqualityImpl(status, other_status)) {
```
kostasrim, line 45:
```suggestion
    if (!(status == other_status)) {
```

tests/unit/replication_persistance_helper.cpp
```diff
@@ -0,0 +1,199 @@
...
+              for (auto &ssl : ssls_) {
+                auto replicas_status =
+                    CreateReplicaStatus(name, ip_address, port, sync_mode, timeout, replica_check_frequency, ssl);
+
+                auto json_status = memgraph::storage::replication::ReplicaStatusToJSON(
+                    memgraph::storage::replication::ReplicaStatus(replicas_status));
+                auto replicas_status_converted =
+                    memgraph::storage::replication::JSONToReplicaStatus(std::move(json_status));
+
+                all_tests_succeeded = CheckEquality(replicas_status, *replicas_status_converted) && all_tests_succeeded;
```
kostasrim, line 189:
```suggestion
                ASSERT_TRUE(CheckEquality(replicas_status, *replicas_status_converted));
```

tests/unit/replication_persistance_helper.cpp
```diff
@@ -0,0 +1,199 @@
...
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+
+  ASSERT_TRUE(all_tests_succeeded);
```
kostasrim, line 198:
```suggestion
```

tests/unit/replication_persistance_helper.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+  auto json_status = memgraph::storage::replication::ReplicaStatusToJSON(
+      memgraph::storage::replication::ReplicaStatus(replicas_status));
+  auto replicas_status_converted = memgraph::storage::replication::JSONToReplicaStatus(std::move(json_status));
+
+  ASSERT_TRUE(CheckEquality(replicas_status, *replicas_status_converted));
+}
+
+TEST_F(ReplicationPersistanceHelperTest, RandomizedTests) {
+  auto all_tests_succeeded = true;
```
kostasrim, line 172:
```suggestion
```

tests/unit/replication_persistance_helper.cpp
```diff
@@ -0,0 +1,199 @@
...
+      if (status.ssl->cert_file != other_status.ssl->cert_file) {
+        return false;
+      }
+    }
+
+    return true;
+  }
+
+  //! Method to automatically generate failing UT.
+  void ExportFailingTest(memgraph::storage::replication::ReplicaStatus &status) const {
```
kostasrim, line 110: I think that this is mostly noise. You can just `ASSERT` and fail on a single iteration of the for loop below (look my comment). I understand that we lose information (that is we now fail on a single assertion) but it makes the code in the test easier to understand.

tests/unit/replication_persistance_helper.cpp
```diff
@@ -0,0 +1,199 @@
...
+TEST_F(ReplicationPersistanceHelperTest, RandomizedTests) {
+  auto all_tests_succeeded = true;
+
+  for (auto &name : names_) {
+    for (auto &ip_address : ip_addresses_) {
+      for (auto port : ports_) {
+        for (auto sync_mode : sync_modes_) {
+          for (auto timeout : timeouts_) {
+            for (auto replica_check_frequency : replica_check_frequencies_) {
+              for (auto &ssl : ssls_) {
```
kostasrim, line 180: This is not random at all, so I think the test name shall be renamed. Moreover, I would put these on a vector of `ReplicaStatus`. I know that the corresponding vectors (ports_, etc) have all different sizes but I would hard code them. It doesn't add much value to test different `ssl` values with all the possible `names`

tests/unit/storage_v2_replication.cpp
```diff
@@ -709,3 +709,91 @@ TEST_F(ReplicationTest, ReplicationReplicaWithExistingEndPoint) {
...
+  ASSERT_EQ(replicaInfos[1].name, "REPLICA2");
+  ASSERT_EQ(replicaInfos[1].endpoint.address, "127.0.0.1");
+  ASSERT_EQ(replicaInfos[1].endpoint.port, 20000);
+}
+
+TEST_F(ReplicationTest, RestoringReplicationAtStartup) {
+  auto main_config = configuration;
+  main_config.durability.restore_replicas_on_startup = true;
+  auto main_store = std::make_unique<memgraph::storage::Storage>(main_config);
+
```
kostasrim, line 758:
```suggestion
const std::string local_host("127.0.0.1");
```

tests/unit/storage_v2_replication.cpp
```diff
@@ -709,3 +709,91 @@ TEST_F(ReplicationTest, ReplicationReplicaWithExistingEndPoint) {
...
+  memgraph::storage::Storage replica_store1(configuration);
+  replica_store1.SetReplicaRole(memgraph::io::network::Endpoint{"127.0.0.1", 10000});
+
+  memgraph::storage::Storage replica_store2(configuration);
+  replica_store2.SetReplicaRole(memgraph::io::network::Endpoint{"127.0.0.1", 20000});
```
kostasrim, line 763:
```suggestion
  replica_store2.SetReplicaRole(memgraph::io::network::Endpoint{local_host, port[1]});
```

kostasrim: similar to other references of `127.0.01` and `10000`, `2000`

tests/unit/storage_v2_replication.cpp
```diff
@@ -709,3 +709,91 @@ TEST_F(ReplicationTest, ReplicationReplicaWithExistingEndPoint) {
...
+
+  auto res = main_store->RegisterReplica("REPLICA1", memgraph::io::network::Endpoint{"127.0.0.1", 10000},
+                                         memgraph::storage::replication::ReplicationMode::SYNC);
+  ASSERT_FALSE(res.HasError());
+  res = main_store->RegisterReplica("REPLICA2", memgraph::io::network::Endpoint{"127.0.0.1", 20000},
+                                    memgraph::storage::replication::ReplicationMode::SYNC);
+  ASSERT_FALSE(res.HasError());
+
+  auto replicaInfos = main_store->ReplicasInfo();
```
kostasrim, line 772:
```suggestion
  auto replica_infos = main_store->ReplicasInfo();
```

kostasrim: Maybe add `const` and introduce blocks `{}`. However, this is not that important, you can leave this as is.

tests/unit/storage_v2_replication.cpp
kostasrim, line 758: const std::array<size_t, 2>ports{10000, 20000}`

tests/unit/replication_persistance_helper.cpp
42jeremy: Test is removed, no need for that anymore.

tests/unit/replication_persistance_helper.cpp
42jeremy: Test is removed, no need for that anymore.

tests/unit/replication_persistance_helper.cpp
42jeremy: Test is removed, no need for that anymore.

tests/unit/replication_persistance_helper.cpp
42jeremy: Test is removed (and the Export method as well), no need for that anymore.

tests/unit/replication_persistance_helper.cpp
42jeremy: Test is removed, no need for that anymore.

tests/unit/storage_v2_replication.cpp
42jeremy: I'll also update the rest of the file, as this is everywhere over the file.

src/storage/v2/storage.cpp
```diff
@@ -1966,4 +1998,39 @@ void Storage::SetIsolationLevel(IsolationLevel isolation_level) {
...
+    return;
+  }
+  spdlog::info("Restoring replicas.");
+
+  for (const auto &[replica_name, replica_data] : *kvstorage_) {
+    spdlog::info("Restoring replica {}.", replica_name);
+
+    const auto maybe_replica_status = replication::JSONToReplicaStatus(nlohmann::json::parse(replica_data));
+    if (!maybe_replica_status.has_value()) {
+      LOG_FATAL("Could parse previously saved configuration of replica {}.", replica_name);
```
antaljanosbenjamin, line 2013:
```suggestion
LOG_FATAL("Cannot parse previously saved configuration of replica {}.", replica_name);
```

Maybe "Couldn't" also good, I am not sure. Either way, please negate the verb.

src/storage/v2/storage.cpp
```diff
@@ -1905,6 +1917,20 @@ utils::BasicResult<Storage::RegisterReplicaError> Storage::RegisterReplica(
...
+    auto data = replication::ReplicaStatusToJSON(
+        replication::ReplicaStatus{.name = name,
+                                   .ip_address = endpoint.address,
+                                   .port = endpoint.port,
+                                   .sync_mode = replication_mode,
+                                   .timeout = config.timeout,
+                                   .replica_check_frequency = config.replica_check_frequency,
+                                   .ssl = config.ssl});
+    if (!kvstorage_->Put(name, data.dump())) {
+      spdlog::error("Issue when saving replica {} in settings.", name);
```
antaljanosbenjamin, line 1930: Just thinking, not sure, but shouldn't this result in a error reported via the return value? If we couldn't persist the replica, then it means Memgraph cannot pick it up again when it starts. That also mean the error would be turned into an exception in the interpreter which can be easily seen by the user.

src/storage/v2/storage.cpp
```diff
@@ -1928,8 +1954,14 @@ utils::BasicResult<Storage::RegisterReplicaError> Storage::RegisterReplica(
   });
 }
 
-bool Storage::UnregisterReplica(const std::string_view name) {
+bool Storage::UnregisterReplica(const std::string &name) {
   MG_ASSERT(replication_role_.load() == ReplicationRole::MAIN, "Only main instance can unregister a replica!");
+  if (ShouldStoreAndRestoreReplicas()) {
+    if (!kvstorage_->Delete(name)) {
+      spdlog::error("Error when removing replica {} from settings.", name);
```
antaljanosbenjamin, line 1961: This is also interesting, however I am not sure what are the implications if we attach another replica that the user didn't want to.

antaljanosbenjamin (changes requested):

src/storage/v2/storage.cpp
42jeremy: I think returning false is better than not doing anything. Otherwise user won't be aware of it (except if he pays attention to the error).
Not returning would mean Memgraph would pick up again the replica when starting again.

42jeremy:
Note to myself to revert commit fd55287996888ac076e6dd79df746be3bd61d110

src/storage/v2/replication/replication_persistence_helper.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+#pragma once
+
+#include <json/json.hpp>
+
+#include <chrono>
+#include <optional>
+#include <string>
+#include "storage/v2/replication/config.hpp"
+#include "storage/v2/replication/enums.hpp"
```
antaljanosbenjamin, line 20:
```suggestion
#include <chrono>
#include <compare>
#include <optional>
#include <string>

#include <json/json.hpp>

#include "storage/v2/replication/config.hpp"
#include "storage/v2/replication/enums.hpp"
```

The `compare` header is mandatory for defaulting the `operator==`.

tests/unit/storage_v2_replication.cpp
```diff
@@ -689,23 +693,110 @@ TEST_F(ReplicationTest, ReplicationReplicaWithExistingEndPoint) {
...
+  ASSERT_EQ(replica_infos[1].name, replicas[1]);
+  ASSERT_EQ(replica_infos[1].endpoint.address, local_host);
+  ASSERT_EQ(replica_infos[1].endpoint.port, ports[1]);
+}
+
+TEST_F(ReplicationTest, RestoringReplicationAtStartup) {
+  auto main_config = configuration;
+  main_config.durability.restore_replicas_on_startup = true;
+  auto main_store = std::make_unique<memgraph::storage::Storage>(main_config);
+  memgraph::storage::Storage replica_store1(configuration);
```
antaljanosbenjamin, line 762: You don't have to act on that, as this is functionality is already there, I just would like to raise the attention here: the two storages are using the same data directory and the lock file doesn't fail because the two storage are in the same process. Thus it might cause some issues in the future, so it is good if more people is aware of this.

antaljanosbenjamin (approved):
Apart from that single header issue it looks good to me!

tests/unit/replication_persistance_helper.cpp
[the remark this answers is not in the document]
42jeremy: -> Test was removed

tests/e2e/replication/show_while_creating_invalid_state.py
```diff
@@ -154,5 +154,208 @@ def test_add_replica_invalid_timeout(connection):
...
+    # 1/
+    expected_data = {
+        ("replica_1", "127.0.0.1:10001", "sync", 2.0, 0, 0, "ready"),
+        ("replica_2", "127.0.0.1:10002", "sync", 1.0, 0, 0, "ready"),
+        ("replica_3", "127.0.0.1:10003", "async", None, 0, 0, "ready"),
+        ("replica_4", "127.0.0.1:10004", "async", None, 0, 0, "ready"),
+    }
+    actual_data = set(execute_and_fetch_all(cursor, "SHOW REPLICAS;"))
+
+    assert expected_data == actual_data
```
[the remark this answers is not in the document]
42jeremy: (Done in: e97ae10ffcfe7af952e1e937ea0147ce4aaec71c)

42jeremy (commented):
Suggestion needed to be done before merging.

tests/unit/storage_v2_replication.cpp
42jeremy: Good remark, I think it's good to know for everybody. That could explain some weird behavior in unit test.
Remark regarding the duplication of config however: here the main_Memgraph wants to allow `restore_replicas_on_startup`, whereas we do not want to have it enable for the replicas themselves (attribute `restore_replicas_on_startup` is `false` by default)

src/storage/v2/config.hpp
```diff
@@ -49,7 +49,7 @@ struct Config {
     uint64_t wal_file_flush_every_n_tx{100000};
 
     bool snapshot_on_exit{false};
-
+    bool restore_replicas_on_startup{false};
```
spacejam, line 52: should this default to true? in memgraph.cpp it is set to true. I don't know what our conventions are but this feels like an opportunity for misunderstandings if it sort of has two defaults, but the one in memgraph.cpp is "actually default" but bound later.

src/storage/v2/storage.cpp
```diff
@@ -400,6 +402,16 @@ Storage::Storage(Config config)
...
   }
+
+  if (config_.durability.restore_replicas_on_startup) {
+    spdlog::info("Replica's configuration will be stored and will be automatically restored in case of a crash.");
+    utils::EnsureDirOrDie(config_.durability.storage_directory / durability::kReplicationDirectory);
+    kvstorage_ =
+        std::make_unique<kvstore::KVStore>(config_.durability.storage_directory / durability::kReplicationDirectory);
+    RestoreReplicas();
+  } else {
+    spdlog::warn("Replicas' configuration will NOT be stored.");
```
spacejam, line 413:
```suggestion
    spdlog::warn("Replicas' configuration will NOT be stored. When the server restarts, replicas will be forgotten.");
```

src/storage/v2/config.hpp
spacejam, line 52:
```suggestion
    bool restore_replicas_on_startup{true};
```
should this default to true? in memgraph.cpp it is set to true. I don't know what our conventions are but this feels like an opportunity for misunderstandings if it sort of has two defaults, but the one in memgraph.cpp is "actually default" but bound later.

src/storage/v2/storage.cpp
```diff
@@ -1962,4 +1995,38 @@ void Storage::SetIsolationLevel(IsolationLevel isolation_level) {
   isolation_level_ = isolation_level;
 }
 
+void Storage::RestoreReplicas() {
+  MG_ASSERT(memgraph::storage::ReplicationRole::MAIN == GetReplicationRole());
```
spacejam, line 1999: I'm not sure if we want to support it, but it can be operationally legitimate to make one replica be the main for another replica in a chain. I think it's fine to assert this though and explicitly not support chain replication, because that causes a lot of additional correctness considerations (what happens if a replica is sync, and that replica has a sync replica, and they basically have to replicate all the way down the chain before main ever gets a success response on the happy path).

src/storage/v2/config.hpp
42jeremy: The struct `Config::Durability` (where the parameter `restore_replicas_on_startup` is) is used at more places than `memgraph.cpp`, for instance in `main` from `mg_import.csv.cpp` at: https://github.com/memgraph/memgraph/blob/b737e53456e0bfb225ff9967b5379f9d3927beb8/src/mg_import_csv.cpp#L702-L708 
Idea was that I wanted these changes to be as transparent as possible excepted for `memgraph.cpp` where we explicitly say we want to restore replica(s) on startup.

src/storage/v2/storage.cpp
42jeremy: That can make sense in the new architecture. I wouldn't change it here though, since we have the same checks for registering/"un-registering" replicas:
https://github.com/memgraph/memgraph/blob/b737e53456e0bfb225ff9967b5379f9d3927beb8/src/storage/v2/storage.cpp#L1883-L1886

src/storage/v2/storage.hpp
```diff
@@ -534,6 +544,7 @@ class Storage final {
   std::filesystem::path wal_directory_;
   std::filesystem::path lock_file_path_;
   utils::OutputFile lock_file_handle_;
+  std::unique_ptr<kvstore::KVStore> kvstorage_;
```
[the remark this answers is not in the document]
42jeremy: Agreed.
What made me decided specifically here was that the name of the class itself is already Storage, I found adding an attribute storage_ could be even more confusing.

tests/e2e/replication/show_while_creating_invalid_state.py
```diff
@@ -117,7 +118,283 @@ def test_show_replicas(connection):
...
+    # 3/
+    interactive_mg_runner.start(CONFIGURATION, "main")
+    cursor = connection(7687, "main").cursor()
+    check_roles()
+
+    # 4/
+    # We leave some time for the main to recover.
+    time.sleep(2)
+    actual_data = set(execute_and_fetch_all(cursor, "SHOW REPLICAS;"))
+    assert actual_data == expected_data
```
spacejam, line 215: replicas may not instantly be in a good state - they might take some time to reconnect and reestablish their configuration. it should be retried a few times in a loop, up to some "definitely-will-never-flake" retry limit, with like 1 second sleeps in between attempts.

tests/e2e/replication/show_while_creating_invalid_state.py
```diff
@@ -117,7 +118,283 @@ def test_show_replicas(connection):
...
+    interactive_mg_runner.MEMGRAPH_INSTANCES["main"].query("REGISTER REPLICA replica_2 SYNC TO '127.0.0.1:10002';")
+
+    # 1/
+    expected_data = {
+        ("replica_1", "127.0.0.1:10001", "sync", 0, 0, "ready"),
+        ("replica_2", "127.0.0.1:10002", "sync", 0, 0, "ready"),
+    }
+    actual_data = set(interactive_mg_runner.MEMGRAPH_INSTANCES["main"].query("SHOW REPLICAS;"))
+
+    assert actual_data == expected_data
```
spacejam, line 371: same as above - we should have some retrying logic that gives the test some time for nodes to connect and set things up without relying on sleeps, which will cause both flakes and slower tests than necessary over time.

spacejam (changes requested):
Everything looks generally good except for the "quiescence" behavior in the python tests - we just need something that dynamically determines whether the cluster has reached a desired state without having to use sleep statements. sleeps cause flaky tests on bad days and slower tests than necessary on good days, so we can avoid flakes and run tests more quickly by having some simple retry logic that asserts that a machine is reachable and has finished trying to set up replication if it is expected to do so. It's OK for a test to hang for a bit before failing, after exhausting a maximum number of retries, because it's better to optimize for the test latency when there are no bugs than the case where one gets introduced, for this kind of test. Even if it's a little slower to flag the bug, it still does so, and avoids false positives or slowness in the absence of bugs as happens with simple sleeps.

tests/e2e/replication/show_while_creating_invalid_state.py
42jeremy: I have created task T0899-MG to tackle the sleep(). We could replace it overall in the tests but that would go out of scope of this task.

42jeremy:
> Everything looks generally good except for the "quiescence" behavior in the python tests - we just need something that dynamically determines whether the cluster has reached a desired state without having to use sleep statements. sleeps cause flaky tests on bad days and slower tests than necessary on good days, so we can avoid flakes and run tests more quickly by having some simple retry logic that asserts that a machine is reachable and has finished trying to set up replication if it is expected to do so. It's OK for a test to hang for a bit before failing, after exhausting a maximum number of retries, because it's better to optimize for the test latency when there are no bugs than the case where one gets introduced, for this kind of test. Even if it's a little slower to flag the bug, it still does so, and avoids false positives or slowness in the absence of bugs as happens with simple sleeps.

I have created task T0899-MG to tackle the sleep(). We could replace it overall in the tests but that would go out of scope of this task.

tests/e2e/replication/show_while_creating_invalid_state.py
```diff
@@ -117,7 +118,283 @@ def test_show_replicas(connection):
...
+    time.sleep(2)
+
+    # 3/
+    interactive_mg_runner.start(CONFIGURATION, "main")
+    cursor = connection(7687, "main").cursor()
+    check_roles()
+
+    # 4/
+    # We leave some time for the main to recover.
+    time.sleep(2)
```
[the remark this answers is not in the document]
42jeremy: I have created task T0899-MG to tackle the sleep(). We could replace it overall in the tests but that would go out of scope of this task.

src/storage/v2/storage.cpp
```diff
@@ -2014,16 +2030,19 @@ void Storage::RestoreReplicas() {
...
+                        {
+                            .replica_check_frequency = replica_status.replica_check_frequency,
+                            .ssl = replica_status.ssl,
+                        });
     if (ret.HasError()) {
-      LOG_FATAL("Failure when restoring replica {}: {}.", replica_name, ret.GetError());
+      spdlog::error("Failure when restoring replica {}: {}.", replica_name,
+                    RegisterReplicaErrorToString(ret.GetError()));
+    } else {
+      spdlog::info("Replica {} restored.", replica_name);
```
antaljanosbenjamin, line 2044: As we just discussed this should still be a `LOG_FATAL`. The above `RegisterReplica` call won't fail because of connection issues, thus every possible error means our knowledge of the replicas is corrupted, we shouldn't continue our startup.

antaljanosbenjamin (changes requested):

src/storage/v2/replication/enums.hpp
```diff
@@ -16,4 +16,7 @@ namespace memgraph::storage::replication {
 enum class ReplicationMode : std::uint8_t { SYNC, ASYNC };
 
 enum class ReplicaState : std::uint8_t { READY, REPLICATING, RECOVERY, INVALID };
+
+enum class RegistrationMode : std::uint8_t { RECOVERY_AT_STARTUP, OTHER };
```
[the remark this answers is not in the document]
42jeremy: I think we won't need it anymore given [this comment](https://github.com/memgraph/memgraph/pull/415#discussion_r914476817). I will resolve the comment if not needed anymore.

src/storage/v2/storage.cpp
42jeremy: Done

src/storage/v2/config.hpp
spacejam: sounds good!

spacejam (approved):

src/storage/v2/replication/enums.hpp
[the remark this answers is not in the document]
42jeremy: I renamed:
- `MUST_BE_INSTANTLY_VALID`
- `CAN_BE_INVALID`

antaljanosbenjamin (approved):

PR #417, opened 2022-06-23
[master < T243-MAGE] Add logging API

antoniofilipovic:
[master < Task] PR
- [x] Add tests

Once PR is ready, I will update the changelog and docs.


include/mg_procedure.h
```diff
@@ -1581,6 +1584,13 @@ enum mgp_error mgp_func_result_set_error_msg(struct mgp_func_result *result, con
 /// mgp_func_result.
 enum mgp_error mgp_func_result_set_value(struct mgp_func_result *result, struct mgp_value *value,
                                          struct mgp_memory *memory);
+
+/// Get an day value for the mgp_date and add 1
+enum mgp_error mgp_date_get_day_add1(struct mgp_date *date, int *day);
```
gitbuda, line 1589: What's the underlying intention of this function? 👀

include/mg_procedure.h
```diff
@@ -164,6 +164,9 @@ enum mgp_value_type {
   MGP_VALUE_TYPE_DURATION,
 };
 
+/// All available log levels that can be use in mgp_log function
+enum mgp_log_level { Trace, Debug, Info, Warn, Error, Critical };
```
antaljanosbenjamin, line 168:
```suggestion
MGP_ENUM_CLASS mgp_log_level {
  TRACE,
  DEBUG,
  INFO,
  WARN,
  ERROR,
  CRITICAL,
};
```

Use `enum class` in C++ and use uppercase enum names. In C the single `PascalCase` enum values can be misleading.

src/query/procedure/py_module.cpp
```diff
@@ -2052,6 +2052,86 @@ PyObject *PyPathMakeWithStart(PyTypeObject *type, PyObject *vertex) {
...
+struct PyLogger {
+  PyObject_HEAD
+};
+// clang-format on
+
+void PyLoggerAddLog(const enum mgp_log_level level, const char *out) { mgp_log(level, out); }
+
+PyObject *PyLoggerAddInfoLog(PyLogger *self, PyObject *args) {
+  const char *out = nullptr;
+  if (!PyArg_ParseTuple(args, "s", &out)) return nullptr;
```
antaljanosbenjamin, line 2065:
```suggestion
  if (!PyArg_ParseTuple(args, "s", &out)) {
    return nullptr;
  }
```

I am not 100% sure about formatting, so directly applying this commit might not be correct.

The same is true for the rest of the functions.

src/query/procedure/py_module.cpp
```diff
@@ -2052,6 +2052,86 @@ PyObject *PyPathMakeWithStart(PyTypeObject *type, PyObject *vertex) {
   return py_path;
 }
 
+// clang-format off
+struct PyLogger {
+  PyObject_HEAD
+};
+// clang-format on
+
+void PyLoggerAddLog(const enum mgp_log_level level, const char *out) { mgp_log(level, out); }
```
antaljanosbenjamin, line 2061: I think here the return value of `mgp_log` is ignored, so it should result in a compilation error in my opinion.

include/mg_procedure.h
```diff
@@ -1581,6 +1584,10 @@ enum mgp_error mgp_func_result_set_error_msg(struct mgp_func_result *result, con
 /// mgp_func_result.
 enum mgp_error mgp_func_result_set_value(struct mgp_func_result *result, struct mgp_value *value,
                                          struct mgp_memory *memory);
+
+// Log a message on certain level with spdlog
+enum mgp_error mgp_log(const enum mgp_log_level log_level, const char *output);
```
antaljanosbenjamin, line 1589:
```suggestion
enum mgp_error mgp_log(enum mgp_log_level log_level, const char *output);
```

As `log_level` is passed by value, the `const` specifier is a noise for the caller and only important in the definition of the function, not in the declaration.

include/mgp.py
```diff
@@ -1610,4 +1610,28 @@ def defined_in_this_module(obj: object):
...
+class Logger:
+    """Represents a Logger."""
+
+    __slots__ = ("_logger",)
+
+    def __init__(self):
+        self._logger = _mgp._LOGGER
+
+    def info(self, out: str) -> None:
+        _mgp._LOGGER.info(out)
```
antaljanosbenjamin, line 1622:
```suggestion
        self._logger.info(out)
```

src/query/procedure/py_module.cpp
antoniofilipovic: I changed mgp_log to be void as it doesn't return value, so all good here now I think

include/mg_procedure.h
antoniofilipovic: Didn't know that, thanks for the explanation!

include/mg_procedure.h
antoniofilipovic: Changed it 👍🏻

include/mg_procedure.h
```diff
@@ -164,6 +164,12 @@ enum mgp_value_type {
   MGP_VALUE_TYPE_DURATION,
 };
 
+/// All available log levels that can be use in mgp_log function
+MGP_ENUM_CLASS mgp_log_level{
+    MGP_LOG_LEVEL_TRACE, MGP_LOG_LEVEL_DEBUG, MGP_LOG_LEVEL_INFO,
+    MGP_LOG_LEVEL_WARN,  MGP_LOG_LEVEL_ERROR, MGP_LOG_LEVEL_CRITICAL,
+};
+
```
antaljanosbenjamin, line 172: I this file up to my knowledge the functions are organized into different groups. The smallest group is `Error Codes`, check out the start and end comments of the group. It is important because the docs page are generated based on these comments.
```

/// @name Error Codes
///
///@{

/// All functions return an error code that can be used to figure out whether the API call was successful or not. In
/// case of failure, the specific error code can be used to identify the reason of the failure.
MGP_ENUM_CLASS MGP_NODISCARD mgp_error{
    MGP_ERROR_NO_ERROR,           MGP_ERROR_UNKNOWN_ERROR,
    MGP_ERROR_UNABLE_TO_ALLOCATE, MGP_ERROR_INSUFFICIENT_BUFFER,
    MGP_ERROR_OUT_OF_RANGE,       MGP_ERROR_LOGIC_ERROR,
    MGP_ERROR_DELETED_OBJECT,     MGP_ERROR_INVALID_ARGUMENT,
    MGP_ERROR_KEY_ALREADY_EXISTS, MGP_ERROR_IMMUTABLE_OBJECT,
    MGP_ERROR_VALUE_CONVERSION,   MGP_ERROR_SERIALIZATION_ERROR,
};
///@}
```

Currently this `mgp_log_level` enum is in the `Operations on mgp_value` group, which is not good in my opinion. I would this into the `Query Module & Procedures`, together with the function.

include/mg_procedure.h
```diff
@@ -1581,6 +1587,10 @@ enum mgp_error mgp_func_result_set_error_msg(struct mgp_func_result *result, con
 /// mgp_func_result.
 enum mgp_error mgp_func_result_set_value(struct mgp_func_result *result, struct mgp_value *value,
                                          struct mgp_memory *memory);
+
+// Log a message on certain level with spdlog
```
antaljanosbenjamin, line 1591:
```suggestion
/// Log a message on a certain level.
```

Is `spdlog` an important factor here? As it is kind of an internal thing, I wouldn't mention this on the public docs, unless it is important. E.g.: if the format strings and parameter would be exposed, then it is important to know what is supported. As only a single string is passed into the function, I think it is irrelevant which logger do we use.

antaljanosbenjamin (changes requested):
Please also considering adding tests, at least calling all of the functions to make sure it works.

include/mgp.py
```diff
@@ -1610,4 +1610,28 @@ def defined_in_this_module(obj: object):
...
+        self._logger.warning(out)
+
+    def critical(self, out: str) -> None:
+        self._logger.critical(out)
+
+    def error(self, out: str) -> None:
+        self._logger.error(out)
+
+    def trace(self, out: str) -> None:
+        self._logger.trace(out)
```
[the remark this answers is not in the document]
jmatak: I would also opt for using it statically rather than defining a Logger object. The reason is that there is no difference in logging functionality, and it is all going through a single logging source in Memgraph.

src/query/procedure/py_module.cpp
```diff
@@ -2052,6 +2052,102 @@ PyObject *PyPathMakeWithStart(PyTypeObject *type, PyObject *vertex) {
   return py_path;
 }
 
+// clang-format off
+struct PyLogger {
+  PyObject_HEAD
+};
```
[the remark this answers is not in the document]
jmatak: Agreed with the above :arrow_up:

src/query/procedure/py_module.cpp
```diff
@@ -2052,6 +2052,102 @@ PyObject *PyPathMakeWithStart(PyTypeObject *type, PyObject *vertex) {
...
+// clang-format on
+
+void PyLoggerAddLog(const enum mgp_log_level level, const char *out) { mgp_log(level, out); }
```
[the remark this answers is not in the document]
antoniofilipovic: I agree, too much repetition in my approach, this one does it better 👍🏻

include/mgp.py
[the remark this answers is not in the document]
antoniofilipovic: The reason behind this approach was to make logging the same experience as with the python [`logging` ](https://docs.python.org/3/library/logging.html#logging.Logger)module, where the user can create a logger and set the global level of logger  (we can add this functionality), and  log with just `logger.log("message")`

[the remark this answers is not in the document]
antoniofilipovic: Furthermore, I know now it is not yet implemented, but having different  loggers for different procedures, we could add something like the following:
`[datetime][log_level][procedure_name] <message>`
I as a user would really love this, I know this is not yet implemented, but we can then easily add that part later on when we have different loggers for different procedures.

src/query/procedure/py_module.cpp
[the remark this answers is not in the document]
antoniofilipovic: It was python related reason, but if we decide to go in a different direction, I will change it to be static.

include/mg_procedure.h
antoniofilipovic: Yes, I completely agree 👍🏻 I didn't know actually that the docs page is generated based on these comments. Fixed 👍🏻

include/mg_procedure.h
antoniofilipovic: Now that you say, it is not important 👍🏻 So as you say it is internal thing and I will remove it 👍🏻

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -2796,3 +2796,26 @@ mgp_error mgp_module_add_function(mgp_module *module, const char *name, mgp_func
...
+    case mgp_log_level::MGP_LOG_LEVEL_WARN:
+      spdlog::warn(output);
+      return;
+    case mgp_log_level::MGP_LOG_LEVEL_ERROR:
+      spdlog::error(output);
+      return;
+    case mgp_log_level::MGP_LOG_LEVEL_CRITICAL:
+      spdlog::critical(output);
+      return;
+  }
```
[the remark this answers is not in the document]
antoniofilipovic: Thanks for the explanation! Loved it. Okay, I will add this `throw` after the switch case.

antoniofilipovic:
> Please also considering adding tests, at least calling all of the functions to make sure it works.

I will add tests, immediately after we reach the decision on whether the `PyLogger` object is necessary.

include/mg_procedure.h
antaljanosbenjamin: You missed the "a" article before "certain level".

include/mg_procedure.h
antaljanosbenjamin: Sorry for not being complete, I also missed that previously, but the `mgp_log` function is in the `Memgraph Magic Functions API` group, so that should be moved also.

include/mgp.py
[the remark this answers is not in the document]
antaljanosbenjamin: If you have plans for the future, then I am okay to go with it!

src/query/procedure/py_module.cpp
```diff
@@ -2052,6 +2052,73 @@ PyObject *PyPathMakeWithStart(PyTypeObject *type, PyObject *vertex) {
   return py_path;
 }
 
+// clang-format off
+struct PyLogger {
+  PyObject_HEAD
+};
+// clang-format on
+
+PyObject *PyLoggerLog(PyLogger *self, PyObject *args, const enum mgp_log_level level) {
```
antaljanosbenjamin, line 2061:
```suggestion
PyObject *PyLoggerLog(PyLogger *self, PyObject *args, const mgp_log_level level) {
```

include/mg_procedure.h
```diff
@@ -1292,6 +1292,12 @@ struct mgp_proc;
 /// Describes a Memgraph magic function.
 struct mgp_func;
 
+/// All available log levels that can be use in mgp_log function
```
antaljanosbenjamin, line 1295:
```suggestion
/// All available log levels that can be used in mgp_log function
```

antaljanosbenjamin (changes requested):
I also reopened two comments of mine as I think you missed small things regarding of them.

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -2796,3 +2796,26 @@ mgp_error mgp_module_add_function(mgp_module *module, const char *name, mgp_func
       },
       result);
 }
+
+void mgp_log(const enum mgp_log_level log_level, const char *output) {
```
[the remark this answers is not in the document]
antaljanosbenjamin: You deleted the enum from the header, not from the source.

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -2796,3 +2796,27 @@ mgp_error mgp_module_add_function(mgp_module *module, const char *name, mgp_func
...
+      spdlog::warn(output);
+      return;
+    case mgp_log_level::MGP_LOG_LEVEL_ERROR:
+      spdlog::error(output);
+      return;
+    case mgp_log_level::MGP_LOG_LEVEL_CRITICAL:
+      spdlog::critical(output);
+      return;
+  }
+  throw std::invalid_argument{fmt::format("Invalid log level: {}", log_level)};
```
antaljanosbenjamin, line 2821: I just noticed this throw is wrong here. If you want to throw, then wrap the implementation of the function, as `mgp_log` is a C API function, it cannot let exceptions go over the API, as exceptions are not existing in C.

tests/unit/CMakeLists.txt
```diff
@@ -151,6 +151,10 @@ target_include_directories(${test_prefix}query_procedure_py_module PRIVATE ${CMA
 add_unit_test(query_procedures_mgp_graph.cpp)
 target_link_libraries(${test_prefix}query_procedures_mgp_graph mg-query storage_test_utils)
 target_include_directories(${test_prefix}query_procedures_mgp_graph PRIVATE ${CMAKE_SOURCE_DIR}/include)
+
+add_unit_test(mgp_log.cpp)
+target_link_libraries(${test_prefix}mgp_log mg-query)
+#target_include_directories(${test_prefix}mgp_log PRIVATE ${CMAKE_SOURCE_DIR}/include)
```
antaljanosbenjamin, line 157: As the function is in MGP API, I think this unit tests is not good. Check out `tests/e2e/write_procedures` how to write e2e tests for procedures.

antaljanosbenjamin (changes requested):

antaljanosbenjamin:
It looks good to me to be merged, however I don't want to approve until at least the PR is created on the docs repo, but nothing more is left in my opinion.

antaljanosbenjamin:
For the formatter don't worry, sooner or later they would have been formatted, so you don't need to revert them.

antejavor:
@antoniofilipovic It would be great if added comments to code matches to mgp.py previous style (there are some parts that I need to update), but you can use this function/class as reference: https://github.com/memgraph/memgraph/blob/T243-MAGE-add-logging-api/include/mgp.py#L756

antaljanosbenjamin (approved):

jmatak (approved):

antejavor (approved):

PR #428, opened 2022-07-04
[master <  T0897-MG] Improve error prone OS setup

gitbuda:


environment/os/debian-11-arm.sh
```diff
@@ -5,6 +5,8 @@ set -Eeuo pipefail
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 source "$DIR/../util.sh"
 
+check_operating_system "debian-11-arm"
```
jbajic, line 8: `operating_system` return only debian not architecure, you need to call `architecture` function to determine if its arm or nor

gitbuda: I've included the architecture check, will resolve this and put the PR in the Open review state 😄

gvolfing (approved):

jbajic (approved):

PR #422, opened 2022-06-29
[master < T0889-MG] Throw exceptions when null is used in the simple form of CASE clause

gvolfing:
[master < Epic] PR
- [ ] Check, and update documentation if necessary
- [ ] Update [changelog](https://docs.memgraph.com/memgraph/changelog)
- [ ] Write E2E tests
- [ ] Compare the [benchmarking results](https://bench-graph.memgraph.com/) between the master branch and the Epic branch

[master < Task] PR
- [ ] Check, and update documentation if necessary
- [ ] Update [changelog](https://docs.memgraph.com/memgraph/changelog)

PR #432, opened 2022-07-05
[master < T0889-MG] Modify logaical operators to conform openCyper policies regarding checking against NULL in CASE expressions

gvolfing:
[master < Task] PR


gvolfing:
Replacement PR for https://github.com/memgraph/memgraph/pull/422

tests/gql_behave/tests/memgraph_V1/features/case.feature
```diff
@@ -57,3 +57,33 @@ Feature: Case
...
+
+    Scenario: Simple CASE nullcheck does not have match:
+        Given an empty graph
+        When executing query:
+            """
+            WITH 2 AS name RETURN CASE name WHEN 3 THEN 'works' WHEN null THEN "doesn't work" ELSE 'something went wrong' END
+            """
+        Then the result should be:
+            |  CASE name WHEN 3 THEN 'works' WHEN null THEN "doesn't work" ELSE 'something went wrong' END |
+            |  'something went wrong'                                                                      |
```
antaljanosbenjamin, line 69:
```suggestion
            WITH 2 AS name RETURN CASE name WHEN 3 THEN 'something went wrong' WHEN null THEN "doesn't work" ELSE 'works' END
            """
        Then the result should be:
            |  CASE name WHEN 3 THEN 'something went wrong' WHEN null THEN "doesn't work" ELSE 'works' END |
            |  'works'                                                                      |
```

Seeing the expected output it is a bit confusing to see `something went wrong` as expected result, that's why I proposed these changes.

tests/gql_behave/tests/memgraph_V1/features/case.feature
```diff
@@ -57,3 +57,33 @@ Feature: Case
...
+
+    Scenario: Generic CASE nullcheck does have match:
+        Given an empty graph
+        When executing query:
+            """
+            WITH 2 AS name RETURN CASE WHEN name is NULL THEN "doesn't work" WHEN name = 2 THEN "works" ELSE "something went wrong" END
+            """
+        Then the result should be:
+            |  CASE WHEN name is NULL THEN "doesn't work" WHEN name = 2 THEN "works" ELSE "something went wrong" END |
+            |  'works'                                                                                               |
```
antaljanosbenjamin, line 89: I was thinking about adding a test case where the test expression itself is `null`. So `WITH null AS name RETURN CASE name WHEN null THEN "doesn't work" WHEN 2 "doesn't work" ELSE 'works' END`.

antaljanosbenjamin (changes requested):

antaljanosbenjamin (approved):

jbajic (approved):

antaljanosbenjamin:
[Docs PR](https://github.com/memgraph/docs/pull/461)

PR #434, opened 2022-07-06
[master < T0899-MG]  Avoid usage of time dot sleep

42jeremy:
- Final commit message: e2e python: added tooling function around time.sleep() that stops as soon as condition is fulfilled and will raise assert if timeout is reached 

e2e testing, no documentation needed


jbajic (approved):
Nice!

tests/e2e/mg_utils.py
```diff
@@ -0,0 +1,16 @@
+import time
+
+
+def mg_sleep_and_assert(expected_value, function_to_retrieve_data, max_duration=20, time_between_attempt=0.5):
```
spacejam, line 4:
```suggestion
def mg_sleep_and_assert(expected_value, function_to_retrieve_data, max_duration=20, time_between_attempt=0.05):
```
I think 2hz is a bit too long to wait for everything, but 20hz should keep tests feeling zippier without causing much looping that could slow things down if it's too aggressive. Many of these operations will have their own timeouts anyway, which will need to be hit before a retry is done after this higher level sleep.

Issue #343, opened 2022-02-02
[BUG] Different results returned when executing the same queries on Memgraph and Neo4j, respectively


abc123000111:
**Memgraph version**
2.1.1 Docker

**Environment**
Windows 11
Java API / Cypher

**Describe the bug**
I discovered that when I executed the following queries on Memgraph and Neo4j, respectively, different results were returned:
```
CREATE (n0 :L2{k17:''});
CREATE (n1 :L3{k20:330281481, k22:false});
CREATE (n2 :L0);
CREATE (n3 :L2);
CREATE (n4 :L3{k18:'8'});
CREATE (n5 :L1{k8:-2090394048, k11:false});
CREATE (n6 :L1{k9:'O4Kc5v'});
CREATE (n7 :L1);
CREATE (n8 :L1);
CREATE (n9 :L0{k1:1490197806, k2:true, k4:false, k5:1490197806});
CREATE (n10 :L1{k10:-2090394048});
MATCH (n10 :L1{k10:-2090394048}), (n8 :L1) MERGE (n10)-[r0 :T5{k63:true}]->(n8);
MATCH (n1) WHERE (0 = 1) MATCH (n1) RETURN sum(0) AS a0;
```

**To Reproduce**
1. Start Memgraph 2.1.1 and Neo4j 4.3.10, respectively.
2. Execute the mentioned queries on each graph database.
3. Compare the result sets.

**Expected behavior**
The result sets are the same. But the actual behavior is:
Memgraph returns: [{a0:null}]
Neo4j returns: [{a0:0}]

**Logs**
Memgraph:
![1](https://user-images.githubusercontent.com/34190313/152110595-c4f0aa7c-c0ab-4d19-9612-85fb0adeed06.png)

Neo4j:
![2](https://user-images.githubusercontent.com/34190313/152110598-d3743553-0567-4633-885b-bd29ae936577.png)

gitbuda:
Hi @abc123000111! I think both are actually valid based on the Cypher semantic. It's quite obvious that `count(nothing)=0`, but with an aggregate function, it's a bit trickier because the query part before returns nothing, and then there are two options:
1. `sum(nothing)=nothing`
2. `sum(nothing)=0`

I personally think returning `null` is more aligned with the semantic because of how Cypher works and what's semantic of the `sum` function is, but returning 0 is also a viable option.

Is this something critical for you? And could you please explain what should be correct behavior from your perspective and why? 😄

abc123000111:
[...]

PR #448, opened 2022-07-12
[master < T0906-MG] different results returned when executing the same queries on memgraph and neo4j respectively

gvolfing:
[master < Task] PR
- [ ] Check, and update documentation if necessary - **Not needed.**

Docs PR:
https://github.com/memgraph/docs/pull/479

### DIFF
--- a/tests/gql_behave/tests/memgraph_V1/features/aggregations.feature
+++ b/tests/gql_behave/tests/memgraph_V1/features/aggregations.feature
@@ -73,6 +73,16 @@ Feature: Aggregations
             | n |
             | 5 |
 
+    Scenario: Count test 07:
+        Given an empty graph
+        When executing query:
+        """
+        RETURN count(null)
+        """
+        Then the result should be:
+            | count(null) |
+            | 0           |
+
     Scenario: Sum test 01:
         Given an empty graph
         And having executed
@@ -113,6 +123,16 @@ Feature: Aggregations
             | n | a.x |
             | 4 | 0   |
             | 4 | 1   |
+
+    Scenario: Sum test 04:
+        Given an empty graph
+        When executing query:
+        """
+        RETURN sum(null)
+        """
+        Then the result should be:
+            | sum(null) |
+            | 0         |
 
     Scenario: Avg test 01:
         Given an empty graph
@@ -155,6 +175,16 @@ Feature: Aggregations
             | 2.0 | 0   |
             | 4.0 | 1   |
 
+    Scenario: Avg test 04:
+        Given an empty graph
+        When executing query:
+        """
+        RETURN avg(null)
+        """
+        Then the result should be:
+            | avg(null) |
+            | 0         |
+
     Scenario: Min test 01:
         Given an empty graph
         And having executed
@@ -195,6 +225,16 @@ Feature: Aggregations
             | n | a.x |
             | 1 | 0   |
             | 4 | 1   |
+
+    Scenario: Min test 04:
+        Given an empty graph
+        When executing query:
+        """
+        RETURN min(null)
+        """
+        Then the result should be:
+            | min(null) |
+            | 0         |
 
     Scenario: Max test 01:
         Given an empty graph
@@ -237,6 +277,16 @@ Feature: Aggregations
             | 3 | 0   |
             | 4 | 1   |
 
+    Scenario: Max test 04:
+        Given an empty graph
+        When executing query:
+        """
+        RETURN max(null)
+        """
+        Then the result should be:
+            | max(null) |
+            | 0         |
+
     Scenario: Collect test 01:
         Given an empty graph
         And having executed
@@ -278,4 +328,3 @@ Feature: Aggregations
         Then the result should be
             | n                                 |
             | {a_key: 13, b_key: 11, c_key: 12} |
-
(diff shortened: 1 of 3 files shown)


gvolfing:
I will also investigate the `count` function in this PR based on this [comment](https://github.com/memgraph/memgraph/issues/343#issuecomment-1041470777).

tests/gql_behave/tests/memgraph_V1/features/aggregations.feature
```diff
@@ -237,6 +277,16 @@ Feature: Aggregations
...
+    Scenario: Max test 04:
+        Given an empty graph
+        When executing query:
+        """
+        RETURN max(null)
+        """
+        Then the result should be:
+            | max(null) |
+            | null      |
+
```
antaljanosbenjamin, line 289: I like the new test cases! First I thought I can construct a case which breaks the compatibility with openCypher/Neo4j:

```
memgraph> MATCH (n) DELETE n;
Empty set (0.001 sec)
1 nodes have been deleted.
memgraph> CREATE (n {property: 1});
Empty set (0.001 sec)
1 nodes have been created.
memgraph> CREATE (n);
Empty set (0.001 sec)
1 nodes have been created.
memgraph> MATCH (n) RETURN count(n) < n.property;
+-----------------------+
| count(n) < n.property |
+-----------------------+
| Null                  |
| false                 |
+-----------------------+
2 rows in set (0.001 sec)
```
But, fortunately I was shocked that is the exact result Neo4j returns. So this test is also passed. However I would introduce one more test that checks the case when the `group_by` expression sometimes gets evaluated to `Null`, something with the following queries:
```
CREATE (n);
CREATE (n {property: 1});
MATCH (n) RETURN count(n) < n.property, count(n.property), count(n), avg(n.property), min(n.property), max(n.property), sum(n.property);
```
And the expected result should be:
```
+-----------------------+-----------------------+-----------------------+-----------------------+-----------------------+-----------------------+-----------------------+
| count(n) < n.property | count(n.property)     | count(n)              | avg(n.property)       | min(n.property)       | max(n.property)       | sum(n.property)       |
+-----------------------+-----------------------+-----------------------+-----------------------+-----------------------+-----------------------+-----------------------+
| Null                  | 0                     | 1                     | Null                  | Null                  | Null                  | 0                     |
| false                 | 1                     | 1                     | 1                     | 1                     | 1                     | 1                     |
+-----------------------+-----------------------+-----------------------+-----------------------+-----------------------+-----------------------+-----------------------+
```

tests/unit/query_plan_accumulate_aggregate.cpp
```diff
@@ -217,31 +217,31 @@ TEST_F(QueryPlanAggregateOps, WithData) {
 TEST_F(QueryPlanAggregateOps, WithoutDataWithGroupBy) {
   {
     auto results = AggregationResults(true, {Aggregation::Op::COUNT});
-    EXPECT_EQ(results.size(), 0);
+    EXPECT_EQ(results.size(), 1);
```
antaljanosbenjamin, line 220: Maybe, could you please check their type and value also?

tests/unit/query_plan_accumulate_aggregate.cpp
```diff
@@ -257,13 +257,13 @@ TEST_F(QueryPlanAggregateOps, WithoutDataWithoutGroupBy) {
...
   EXPECT_TRUE(results[0][3].IsNull());
-  // sum
-  EXPECT_TRUE(results[0][4].IsNull());
-  // avg
+  //  sum
+  EXPECT_EQ(results[0][4].ValueInt(), 0);
+  //  avg
   EXPECT_TRUE(results[0][5].IsNull());
-  // collect list
+  //  collect list
```
antaljanosbenjamin, line 266:
```suggestion
  // collect list
```

antaljanosbenjamin (changes requested):

tests/gql_behave/tests/memgraph_V1/features/aggregations.feature
gvolfing: I am thinking if the ordering of the rows might cause some problems, because for me its the other way around, so:
+-----------------------+-----------------------+-----------------------+-----------------------+-----------------------+-----------------------+-----------------------+
                               │| count(n) < n.property | count(n.property)     | count(n)              | avg(n.property)       | min(n.property)       | max(n.property)       | sum(n.property)       |
                               │+-----------------------+-----------------------+-----------------------+-----------------------+-----------------------+-----------------------+-----------------------+
                               │| false                 | 1                     | 1                     | 1                     | 1                     | 1                     | 1                     |
                               │| Null                  | 0                     | 1                     | Null                  | Null                  | Null                  | 0                     |
                               │+-----------------------+-----------------------+-----------------------+-----------------------+-----------------------+-----------------------+-----------------------+

jbajic (approved):

tests/gql_behave/tests/memgraph_V1/features/aggregations.feature
antaljanosbenjamin: I copied it from Memgraph, but I might created the vertices in reversed order.

antaljanosbenjamin (approved):

PR #449, opened 2022-07-12
[master <  T0958-MG] Use CentOS 9 instead of 8

antaljanosbenjamin:
[master < Task] PR


antaljanosbenjamin:
This is not that easy, because we don't have CentOS 9 runners.

PR #455, opened 2022-07-18
[E118-MG < T0887-MG] Make storage use `KeyStore`

antaljanosbenjamin:
Currently `KeyStore` stores only the global id of the vertex, but with the introduction of schema to storage, we can utilize KeyStore to store multiple values.


tests/unit/storage_v3_test_utils.cpp
```diff
@@ -0,0 +1,23 @@
...
+#include "storage_v3_test_utils.hpp"
+
+namespace memgraph::storage::v3::tests {
+
+size_t CountVertices(Storage::Accessor &storage_accessor, View view) {
+  auto vertices = storage_accessor.Vertices(view);
+  size_t count = 0U;
+  for (auto it = vertices.begin(); it != vertices.end(); ++it, ++count)
+    ;
+  return count;
```
kostasrim, line 21:
```suggestion
  return std::distance(vertices.begin(), vertices.end());
```

kostasrim: you can also try `std::ranges::distance` but the iterator concepts have slightly changed so it might not work

tests/unit/storage_v3_key_store.cpp
```diff
@@ -0,0 +1,46 @@
...
+#include <rapidcheck.h>
+#include <rapidcheck/gtest.h>
+
+#include "storage/v3/id_types.hpp"
+#include "storage/v3/key_store.hpp"
+#include "storage/v3/property_value.hpp"
+
+namespace memgraph::storage::v3::test {
+
+RC_GTEST_PROP(KeyStore, KeyStore, (std::vector<std::string> values)) {
```
kostasrim, line 29: Just to make sure that I understand `RC_GTEST_PROP`.  This is a google macro that effectively uses `RapidCheck` to generate a random list of properties seeded/passed to the `values` parameter. Is that correct?

src/storage/v3/vertex_accessor.hpp
```diff
@@ -94,7 +94,10 @@ class VertexAccessor final {
 
   Result<size_t> OutDegree(View view) const;
 
-  Gid Gid() const noexcept { return vertex_->gid; }
+  Gid Gid() const noexcept {
+    // TODO(antaljanosbenjamin): remove this whole function.
```
kostasrim, line 98: Is this going to be part of this PR?

src/storage/v3/vertex.hpp
```diff
@@ -13,24 +13,28 @@
...
-  Vertex(Gid gid, Delta *delta) : gid(gid), deleted(false), delta(delta) {
+  Vertex(Gid gid, Delta *delta) : keys{{PropertyValue{gid.AsInt()}}}, deleted(false), delta(delta) {
     MG_ASSERT(delta == nullptr || delta->action == Delta::Action::DELETE_OBJECT,
               "Vertex must be created with an initial DELETE_OBJECT delta!");
   }
 
-  Gid gid;
+  Gid Gid() const { return Gid::FromInt(keys.GetKey(0).ValueInt()); }
 
+  KeyStore keys{{}};
```
kostasrim, line 37: Why not add a default constructor?

src/storage/v3/key_store.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+#include "storage/v3/property_store.hpp"
+#include "storage/v3/property_value.hpp"
+
+namespace memgraph::storage::v3 {
+
+class KeyStore {
+ public:
+  explicit KeyStore(const std::vector<PropertyValue> &key_values);
+
```
kostasrim, line 24:
```suggestion
 KeyStore() = default;
```

src/storage/v3/key_store.hpp
```diff
@@ -0,0 +1,62 @@
...
+  PropertyValue GetKey(size_t index) const;
+
+  std::vector<PropertyValue> Keys() const;
+
+  friend bool operator<(const KeyStore &lhs, const KeyStore &rhs) {
+    // TODO(antaljanosbenjamin): also compare the schema
+    return std::ranges::lexicographical_compare(
+        lhs.Keys(), rhs.Keys(),
+        [](const PropertyValue &lhs_value, const PropertyValue &rhs_value) { return lhs_value < rhs_value; });
```
kostasrim, line 40:
```suggestion
        lhs.Keys(), rhs.Keys(), std::less<PropertyValue>, std::less<PropertyValue>{});
```

kostasrim: Also this is implementation-defined behaviour, because `ranges::lexicographical_compare` is defined in the `<algorithm>` header. Plz include that

src/storage/v3/key_store.hpp
```diff
@@ -0,0 +1,62 @@
...
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#pragma once
+
+#include <compare>
+
```
kostasrim, line 15:
```suggestion
#include <algorithm>
#include <functional>
```

src/storage/v3/key_store.cpp
```diff
@@ -0,0 +1,40 @@
...
+  for (auto i = 0; i < key_values.size(); ++i) {
+    MG_ASSERT(!key_values[i].IsNull());
+    store_.SetProperty(PropertyId::FromInt(i), key_values[i]);
+  }
+}
+
+PropertyValue KeyStore::GetKey(const size_t index) const { return store_.GetProperty(PropertyId::FromUint(index)); }
+
+std::vector<PropertyValue> KeyStore::Keys() const {
+  auto keys_map = store_.Properties();
```
kostasrim, line 31: Basically, you get a `copy` of the properties here which you then move to the `keys`. Can the `properties` change by the end of this function call (through some other thread which sets the properties)? If that's the case, by the end of this function the properties will be outdated. Is this desirable?

src/storage/v3/key_store.hpp
```diff
@@ -0,0 +1,62 @@
...
+  friend bool operator==(const KeyStore &lhs, const KeyStore &rhs) {
+    return std::ranges::equal(lhs.Keys(), rhs.Keys());
+  }
+
+  friend bool operator<(const KeyStore &lhs, const std::vector<PropertyValue> &rhs) {
+    // TODO(antaljanosbenjamin): also compare the schema
+    return std::ranges::lexicographical_compare(
+        lhs.Keys(), rhs,
+        [](const PropertyValue &lhs_value, const PropertyValue &rhs_value) { return lhs_value < rhs_value; });
+  }
```
kostasrim, line 52:
```suggestion
    return std::ranges::lexicographical_compare(
        lhs.Keys(), rhs, std::less<PropertyValue>{} );
  }
```

src/storage/v3/key_store.hpp
```diff
@@ -0,0 +1,62 @@
...
+        lhs.Keys(), rhs.Keys(),
+        [](const PropertyValue &lhs_value, const PropertyValue &rhs_value) { return lhs_value < rhs_value; });
+  }
+
+  friend bool operator==(const KeyStore &lhs, const KeyStore &rhs) {
+    return std::ranges::equal(lhs.Keys(), rhs.Keys());
+  }
+
+  friend bool operator<(const KeyStore &lhs, const std::vector<PropertyValue> &rhs) {
+    // TODO(antaljanosbenjamin): also compare the schema
```
kostasrim, line 48: Is this going to be part of this PR? I assume not correct?

src/storage/v3/key_store.hpp
```diff
@@ -0,0 +1,62 @@
...
+    return std::ranges::lexicographical_compare(
+        lhs.Keys(), rhs,
+        [](const PropertyValue &lhs_value, const PropertyValue &rhs_value) { return lhs_value < rhs_value; });
+  }
+
+  friend bool operator==(const KeyStore &lhs, const std::vector<PropertyValue> &rhs) {
```
kostasrim, line 54: Maybe also add: `friend bool operator==(const std::vector<PropertyValue> &rhs, const KeyStore &lhs)` for symmetry?

src/storage/v3/key_store.hpp
```diff
@@ -0,0 +1,62 @@
...
+    return std::ranges::lexicographical_compare(
+        lhs.Keys(), rhs.Keys(),
+        [](const PropertyValue &lhs_value, const PropertyValue &rhs_value) { return lhs_value < rhs_value; });
+  }
+
+  friend bool operator==(const KeyStore &lhs, const KeyStore &rhs) {
+    return std::ranges::equal(lhs.Keys(), rhs.Keys());
+  }
+
+  friend bool operator<(const KeyStore &lhs, const std::vector<PropertyValue> &rhs) {
```
kostasrim, line 47: Maybe add `friend bool operator<(const std::vector<PropertyValue> &rhs, const KeyStore &lhs)` for symmetry?

src/storage/v3/lexicographically_ordered_vertex.hpp
```diff
@@ -0,0 +1,57 @@
...
+  friend bool operator==(const LexicographicallyOrderedVertex &lhs, const LexicographicallyOrderedVertex &rhs) {
+    return lhs.vertex.keys == rhs.vertex.keys;
+  }
+
+  friend bool operator<(const LexicographicallyOrderedVertex &lhs, const LexicographicallyOrderedVertex &rhs) {
+    return lhs.vertex.keys < rhs.vertex.keys;
+  }
+
+  // TODO(antaljanosbenjamin): maybe it worth to overload this for std::array to avoid heap construction of the vector
+  friend bool operator==(const LexicographicallyOrderedVertex &lhs, const std::vector<PropertyValue> &rhs) {
```
kostasrim, line 32: Consider symmetry as the above comments. Same for below.

src/storage/v3/lexicographically_ordered_vertex.hpp
```diff
@@ -0,0 +1,57 @@
...
+    return lhs.vertex.keys < rhs;
+  }
+};
+
+template <typename T>
+concept IsLexicographicallyOrderedVertex = std::same_as<LexicographicallyOrderedVertex, std::remove_cvref_t<T>>;
+
+template <typename T>
+concept IsLexicographicallyOrderedVertexHolder =
+    std::same_as<LexicographicallyOrderedVertex, std::remove_cvref_t<decltype(*(std::declval<T>()))>>;
```
kostasrim, line 46: You are applying `*` to a `T` that might not be `dereferenceable` and I think this would generate a bunch of annoying mesasges. Maybe introduce a concept `is_dereferenceable` to check the existence of the `operator*`. As far as I am aware, there is no standard type_trait (or concept) for that and we can't really use https://en.cppreference.com/w/cpp/experimental/is_detected since it's experimental. It should be simple to write with a concept though.

tests/unit/storage_v3_property_store.cpp
```diff
@@ -0,0 +1,629 @@
+// Copyright 2022 Memgraph Ltd.
```
kostasrim, line 1: I haven;t checked this file yet (I will do it tomorrow morning).

src/storage/v3/key_store.cpp
antaljanosbenjamin: No, the `KeyStore` together with everything else inside a `StorageObject` are not meant to be used from multiple threads, strictly single threaded logic will be here. The parallelization will be done by using multiple shards per storage engine.

src/storage/v3/key_store.hpp
antaljanosbenjamin: Actually this is an outdated comment. Inside a skiplist we will only store vertices from a single primary label, that means comparing the schema/primary label is totally unnecessary. I will remove this comment on the PR when the label spaces will be introduced.

src/storage/v3/lexicographically_ordered_vertex.cpp
```diff
@@ -0,0 +1,14 @@
...
+// included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
+// License, and you may not use this file except in compliance with the Business Source License.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#include "storage/v3/lexicographically_ordered_vertex.hpp"
+
```
jbajic, line 13: Why have this file?

src/storage/v3/key_store.hpp
antaljanosbenjamin: From [C++20](https://isocpp.org/files/papers/N4860.pdf) you don't have to. During overload resolution the compiler also consider the operators with reversed order of parameters. Check chapter 12.4.1.2 [over.match.oper] paragraph 3.4.4. The even nicer thing is by defining `operator==`, you also get `!=`. Really cool!

src/storage/v3/vertex.hpp
jbajic: Also every vertex will have KeyStore and every `KeyStore` has a `PropertyStore`, isn't `PropertyStore` a global object and local for every vertex, or am I missing something?

src/storage/v3/replication/replication_server.cpp
```diff
@@ -167,9 +167,10 @@ void Storage::ReplicationServer::SnapshotHandler(slk::Reader *req_reader, slk::B
       LabelPropertyIndex(&storage_->indices_, &storage_->constraints_, storage_->config_.items);
   try {
     spdlog::debug("Loading snapshot");
-    auto recovered_snapshot = durability::LoadSnapshot(*maybe_snapshot_path, &storage_->vertices_, &storage_->edges_,
-                                                       &storage_->epoch_history_, &storage_->name_id_mapper_,
-                                                       &storage_->edge_count_, storage_->config_.items);
+    auto recovered_snapshot = durability::RecoveredSnapshot{};
+
+    durability::LoadSnapshot(*maybe_snapshot_path, &storage_->vertices_, &storage_->edges_, &storage_->epoch_history_,
+                             &storage_->name_id_mapper_, &storage_->edge_count_, storage_->config_.items);
```
jbajic, line 173: Maybe just place TODO to get nack to this, or is there a task for this?

src/storage/v3/lexicographically_ordered_vertex.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+template <IsLexicographicallyOrderedVertex T>
+auto &GetVertex(T &wrapper) {
+  return wrapper.vertex;
+}
+template <IsLexicographicallyOrderedVertexHolder T>
+auto &GetVertex(T &wrapper) {
+  return (*wrapper).vertex;
```
jbajic, line 54: Since this is being called everywhere, why not make it a part of `LexicographicallyOrderedVertex`?

jbajic: It is kind of strange to see `GetVertex` being called on vertex

tests/unit/storage_v3.cpp
```diff
@@ -12,10 +12,2575 @@
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
 
+#include <limits>
```
jbajic, line 15: limits is std, and should go abveo gtest library

tests/unit/storage_v3.cpp
```diff
@@ -12,10 +12,2575 @@
...
+#include "storage/v3/vertex_accessor.hpp"
+#include "storage_v3_test_utils.hpp"
+
+using testing::UnorderedElementsAre;
+
+namespace memgraph::storage::v3::tests {
+
+// NOLINTNEXTLINE(hicpp-special-member-functions)
+TEST(StorageV3, Commit) {
+  Storage store;
```
jbajic, line 28: You have storage in every test, and there are a lot of them, maybe move it to fixture jsut to avoid repetition

jbajic: Plus another question are all of these tests from v2, and is there something new here

tests/unit/storage_v3_test_utils.cpp
jbajic: Why not call `ApproximateVertexCount`?

tests/unit/storage_v3_property_store.cpp
```diff
@@ -0,0 +1,629 @@
...
+
+#include "storage/v3/property_store.hpp"
+#include "storage/v3/property_value.hpp"
+#include "storage/v3/temporal.hpp"
+
+namespace memgraph::storage::v3::tests {
+
+using testing::UnorderedElementsAre;
+
+const PropertyValue kSampleValues[] = {
```
jbajic, line 25: This will ve visible in our namespace now and the name is kind of generic, maybe add it as static variable inside fixture, where you can also instantitae `PropertyStore`?

tests/unit/storage_v3_property_store.cpp
```diff
@@ -0,0 +1,629 @@
...
+    PropertyValue(
+        std::vector<PropertyValue>{PropertyValue(33), PropertyValue(std::string("sample")), PropertyValue(-33.33)}),
+    PropertyValue(std::vector<PropertyValue>{PropertyValue(), PropertyValue(false)}),
+    PropertyValue(std::map<std::string, PropertyValue>{{"sample", PropertyValue()}, {"key", PropertyValue(false)}}),
+    PropertyValue(std::map<std::string, PropertyValue>{
+        {"test", PropertyValue(33)}, {"map", PropertyValue(std::string("sample"))}, {"item", PropertyValue(-33.33)}}),
+    PropertyValue(TemporalData(TemporalType::Date, 23)),
+};
+
+void TestIsPropertyEqual(const PropertyStore &store, PropertyId property, const PropertyValue &value) {
```
jbajic, line 54: `AssertPropertyIsEqual`

tests/unit/storage_v3_property_store.cpp
jbajic, line 25: What do you think about moving this to fixture, I would prefer to have this encapsulated, plus you can instantiate `PropertyStore` in fixture

tests/unit/storage_v3_property_store.cpp
```diff
@@ -0,0 +1,629 @@
...
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+#include <limits>
```
jbajic, line 15: Move limits above gtest includes

src/storage/v3/vertex.hpp
kostasrim: What do you mean `PropertyStore` being global and local? The `Vertex` class has a `PropertyStore` which basically is the list of the properties in compressed form?

src/storage/v3/key_store.cpp
kostasrim: :gem: completely forgot about that. Second time you mention this ;)

src/storage/v3/key_store.hpp
kostasrim: :gem:: completely forgot about that. Second time you mention this ;)

tests/unit/storage_v3_test_utils.cpp
antaljanosbenjamin: @kostasrim [`std::distance`](https://en.cppreference.com/w/cpp/iterator/distance) requires iterator to be LegacyInputIterator and `std::ranges::difference` requires `std::weakly_incrementable`. Both of them are unsatisfied for this iterator. As we have to probably rewrite these iterators, I wouldn't spend time on improving them.

@jbajic I guess because it doesn't consider the transactions, meaning it might "count" vertices that are not accessible from the transaction. This iterator takes care of that.

tests/unit/storage_v3_key_store.cpp
antaljanosbenjamin: I am not sure if this come with GTest of RapidCheck, but you are right about the functionality of the macro.

src/storage/v3/vertex_accessor.hpp
antaljanosbenjamin: No, this can only be removed when we have schema.

src/storage/v3/vertex.hpp
antaljanosbenjamin: @kostasrim Because I don't want to have a default constructor, because the `KeyStore` is meant to be an immutable type. I will remove this though, as it was just a temporary hack.

antaljanosbenjamin: After discussion this with Jure, the important things to note:
1. `PropertyStore` is not a global object at any level, it is somewhat similar to flatmap (property id -> property value) that tries to minimize the used memory. So both `KeyStore` and `Vertex` will have its own `PropertyStore`. Th former one will contain the immutable primary key, the latter one will contain all of the other properties.

src/storage/v3/key_store.hpp
antaljanosbenjamin: I wouldn't like to, see my other comment in vertex.hpp.

src/storage/v3/key_store.hpp
antaljanosbenjamin: Same as above.

src/storage/v3/lexicographically_ordered_vertex.cpp
antaljanosbenjamin: Forgot to remove, done now.

src/storage/v3/replication/replication_server.cpp
antaljanosbenjamin: I think this should actually work with this changes, because the vertices are still "referenced" by Gid, thus I don't think it requires any TODO for now.

tests/unit/storage_v3.cpp
antaljanosbenjamin: I think they are 1-to-1 copies of v2 tests. I think there was no issue with compilation etc, so I only "shortened" the source by putting it into the `memgraph::storage::v3::tests` namespace and remove the extraneous prefixes.

src/storage/v3/lexicographically_ordered_vertex.hpp
antaljanosbenjamin: Removed `GetVertex`. `LexicographicallyOrderedVertex::vertex` can be accessed directly.

src/storage/v3/vertex.hpp
```diff
@@ -13,24 +13,28 @@
...
 
 struct Vertex {
-  Vertex(Gid gid, Delta *delta) : gid(gid), deleted(false), delta(delta) {
+  Vertex(Gid gid, Delta *delta) : keys{{PropertyValue{gid.AsInt()}}}, deleted(false), delta(delta) {
     MG_ASSERT(delta == nullptr || delta->action == Delta::Action::DELETE_OBJECT,
               "Vertex must be created with an initial DELETE_OBJECT delta!");
   }
 
-  Gid gid;
+  Gid Gid() const { return Gid::FromInt(keys.GetKey(0).ValueInt()); }
```
jbajic, line 35: You don't think its valuable to store gid for quicker retrival?

tests/unit/storage_v3_property_store.cpp
jbajic: Just a reminder about this comment

jbajic (approved):
Everything looks good in my opinion, left just few questions/comments

src/storage/v3/vertex.hpp
antaljanosbenjamin: After we have schema, gid will be gone. Currently the "implicit schema" has only one property, the gid.

tests/unit/storage_v3_property_store.cpp
antaljanosbenjamin: Somehow I put :+1: on this comment without addressing, I will address this. Good that you cought it!

kostasrim (approved):

PR #457, opened 2022-07-18
[E112-MG < T0851-MG] Enforce schema on vertex creation

jbajic:
[master < Task] PR

This PR introduces a breaking change of enforcing schema validation upon vertex manipulation:
- Vertex creation,
- Vertex properties update and
- Vertex adding/removing labels
- Major changes to unit tests
Also, the following features have been excluded from vertex validation:
- `mg_import`
- query modules
- replication

This PR won't fix any e2e tests, integration tests, or mgbench tests.

Commit Message:
- Separating schema definition from schema validation
- Updating `vertex_accessor` and `db_accessors` with necessary methods
- Adding a primary label to `Vertex`
- Adding schema tests
- Updating existing tests for storage v3, and deprecating old:
  -  `interpreter` => `interpreter_v2`
  -  `query_plan_accumulate_aggregate` => `storage_v3_query_plan_accumulate_aggregate`
  - `query_plan_create_set_remove_delete` => `storage_v3_query_plan_create_set_remove_delete`
  - `query_plan_bag_semantics` => `storage_v3_query_plan_bag_semantics`
  - `query_plan_edge_cases` => `storage_v3_query_plan_edge_cases`
  - `query_plan_v2_create_set_remove_delete` => `storage_v3_query_plan_v2_create_set_remove_delete`
  - `query_plan_match_filter_return` => `storage_v3_query_plan_match_filter_return`


src/storage/v2/vertex_accessor.cpp
```diff
@@ -118,7 +120,8 @@ Result<bool> VertexAccessor::HasLabel(LabelId label, View view) const {
   {
     std::lock_guard<utils::SpinLock> guard(vertex_->lock);
     deleted = vertex_->deleted;
-    has_label = std::find(vertex_->labels.begin(), vertex_->labels.end(), label) != vertex_->labels.end();
+    has_label = vertex_->primary_label == label ||
+                std::find(vertex_->labels.begin(), vertex_->labels.end(), label) != vertex_->labels.end();
```
antaljanosbenjamin, line 124:
```
    has_label = VertexHasLabel(vertex_, label);
```

src/storage/v2/vertex_accessor.cpp
```diff
@@ -158,6 +161,8 @@ Result<bool> VertexAccessor::HasLabel(LabelId label, View view) const {
   return has_label;
 }
 
+LabelId VertexAccessor::PrimaryLabel() const noexcept { return vertex_->primary_label; }
```
antaljanosbenjamin, line 164: After checking the PR, two questions here:
1. Do we need to expose the priamry label of the vertex? I think until we don't use it, we can emit this function.
2. I think I forgot about the existing and deleted stuff. So if we want to expose the primary label, then maybe it worth to check whether the vertex exists/deleted.

src/storage/v2/schema_validator.cpp
```diff
@@ -0,0 +1,107 @@
...
+      violated_schema_property{violated_schema_property},
+      violated_property_value{violated_property_value} {}
+
+SchemaValidator::SchemaValidator(Schemas &schemas) : schemas_{schemas} {}
+
+[[nodiscard]] std::optional<SchemaViolation> SchemaValidator::ValidateVertexCreate(
+    LabelId primary_label, const std::vector<LabelId> &labels,
+    const std::vector<std::pair<PropertyId, PropertyValue>> &properties) const {
+  // Schema on primary label
+  const auto schema = schemas_.GetSchema(primary_label);
```
antaljanosbenjamin, line 46: As `Schemas::GetSchema` is called in the hot path and it contains a vector, consider using a pointer instead. If the function returns a `nullptr`, it means the schema doesn't exists. This way you can avoid copying the vector inside the schema.

src/storage/v2/schema_validator.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+  // Check only properties defined by schema
+  for (const auto &schema_type : schema->second) {
+    // Check schema property existence
+    auto property_pair = std::ranges::find_if(
+        properties, [schema_property_id = schema_type.property_id](const auto &property_type_value) {
+          return property_type_value.first == schema_property_id;
+        });
+    if (property_pair == properties.end()) {
+      return SchemaViolation(SchemaViolation::ValidationStatus::VERTEX_HAS_NO_PROPERTY, primary_label, schema_type);
```
antaljanosbenjamin, line 66:
```
      return SchemaViolation(SchemaViolation::ValidationStatus::VERTEX_HAS_NO_PRIMARY_PROPERTY, primary_label, schema_type);
```

or 

```
      return SchemaViolation(SchemaViolation::ValidationStatus::SCHEMA_HAS_NO_PROPERTY, primary_label, schema_type);
```

Would be a better name in my opinion.

src/storage/v2/schema_validator.cpp
```diff
@@ -0,0 +1,107 @@
...
+  return std::nullopt;
+}
+
+[[nodiscard]] std::optional<SchemaViolation> SchemaValidator::ValidateVertexUpdate(const LabelId primary_label,
+                                                                                   const PropertyId property_id) const {
+  // Verify existence of schema on primary label
+  const auto schema = schemas_.GetSchema(primary_label);
+  if (!schema) {
+    return SchemaViolation(SchemaViolation::ValidationStatus::NO_SCHEMA_DEFINED_FOR_LABEL, primary_label);
+  }
```
antaljanosbenjamin, line 86: This should be an assert in my opinion. This function shouldn't be called with non existing primary label, it is only valid for `ValidateCreateVertex`.

src/storage/v2/schema_validator.cpp
```diff
@@ -0,0 +1,107 @@
...
+    return SchemaViolation(SchemaViolation::ValidationStatus::VERTEX_UPDATE_PRIMARY_KEY, primary_label,
+                           *schema_property);
+  }
+  return std::nullopt;
+}
+
+[[nodiscard]] std::optional<SchemaViolation> SchemaValidator::ValidateLabelUpdate(const LabelId label) const {
+  const auto schema = schemas_.GetSchema(label);
+  if (schema) {
+    return SchemaViolation(SchemaViolation::ValidationStatus::VERTEX_ALREADY_HAS_PRIMARY_LABEL, label);
```
antaljanosbenjamin, line 102:
```suggestion
    return SchemaViolation(SchemaViolation::ValidationStatus::VERTEX_MODIFY_PRIMARY_LABEL, label);
```

The name `VERTEX_ALREADY_HAS_PRIMARY_LABEL` suggest for me that a vertex might exist without a primary label, which shouldn't be allowed in my opinion. Also this function is also called from `RemoveLabel`, which doesn't want to add any new label.

tests/unit/interpreter_v2.cpp
```diff
@@ -0,0 +1,1636 @@
+// Copyright 2022 Memgraph Ltd.
```
antaljanosbenjamin, line 1: When and how do you plan to migrate these changes to query-v2/storage-v3? I know we talked about it, but forgot the conclusion.

tests/unit/storage_v3_query_plan_accumulate_aggregate.cpp
```diff
@@ -0,0 +1,634 @@
+// Copyright 2022 Memgraph Ltd.
```
antaljanosbenjamin, line 1: Could you please explain why these tests has `storage_v3` in their name but using `storage-v2`?

antaljanosbenjamin (changes requested):

src/query/db_accessor.hpp
```diff
@@ -106,8 +108,34 @@ class VertexAccessor final {
...
+        maybe_violation_error) {
+      return {*maybe_violation_error};
+    }
+    auto maybe_error = impl_.AddLabel(label);
+    if (maybe_error.HasValue()) {
+      return *maybe_error;
+    }
+    return {maybe_error.GetError()};
+  }
+
```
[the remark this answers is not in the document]
jbajic: It is not a problem at all but my question is why not keep this validation here, isn't it better than t the validation happens as soon as possible?

src/query/db_accessor.hpp
```diff
@@ -11,10 +11,11 @@
 
 #pragma once
 
-#include <optional>
-
 #include <cppitertools/filter.hpp>
 #include <cppitertools/imap.hpp>
+#include <optional>
+#include <variant>
+#include <vector>
```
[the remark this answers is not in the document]
jbajic: I will check includes in other files as well then, add this info to our c++ convention page

src/query/common.hpp
```diff
@@ -81,15 +86,54 @@ concept AccessorWithSetProperty = requires(T accessor, const storage::PropertyId
...
+          fmt::format("Cannot create vertex with secondary label :{}", dba.LabelToName(schema_violation.label)));
+    }
+  }
+}
+
+template <AccessorWithSetProperty T>
+storage::PropertyValue PropsSetChecked(T *record, const DbAccessor &dba, const storage::PropertyId &key,
+                                       const TypedValue &value) {
+  return PropsSetChecked(record, dba, key, storage::PropertyValue(value));
+}
```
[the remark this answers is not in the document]
jbajic: This was something I was trying out, but it is not needed anymore, will remove it

tests/unit/storage_v3_query_plan_accumulate_aggregate.cpp
jbajic: Because the schema is currently in storage-v2 and not in v3, and that will be all moved in the next PR, both query and storage to their next version, since the schema epic began before storage-v3 was introduced

tests/unit/interpreter_v2.cpp
jbajic: In the PR after we merge this one

src/query/db_accessor.hpp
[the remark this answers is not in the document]
jbajic: This was resolved in conversation between Benjamin and me, and the conclusion is we want to keep this logic in the storage engine for now because of these reasons:
 - query engine might send data that are not part of the schema defined  in the storage engine, because of some bug  especially in the beggining
 - validating in the query engine can be considered and optimization that will be possibly implemented in the later stages

tests/unit/storage_v3_schema.cpp
```diff
@@ -0,0 +1,310 @@
...
+ protected:
+  SchemaTest()
+      : prop1{storage.NameToProperty("prop1")},
+        prop2{storage.NameToProperty("prop2")},
+        label1{storage.NameToLabel("label1")},
+        label2{storage.NameToLabel("label2")},
+        schema_prop1{prop1, memgraph::common::SchemaType::STRING},
+        schema_prop2{prop2, memgraph::common::SchemaType::INT} {}
+
+  memgraph::storage::Storage storage;
```
antaljanosbenjamin, line 42: I am wondering whether we need this storage at all. I think it is not necessary. If you want to generate the property/label ids, then I think a NameIdMapper + two helper function are enough. The reason why I think this can be beneficial is it makes harder to introduce any dependency on the storage itself within these tests.

antaljanosbenjamin (changes requested):

src/storage/v2/schema_validator.cpp
jbajic: I will for the first one, since all status already have vertex suffix

tests/unit/CMakeLists.txt
```diff
@@ -282,42 +278,67 @@ target_link_libraries(${test_prefix}commit_log_v2 gflags mg-utils mg-storage-v2)
...
-add_unit_test(storage_v2_wal_file.cpp)
-target_link_libraries(${test_prefix}storage_v2_wal_file mg-storage-v2 fmt)
-
-add_unit_test(storage_v2_replication.cpp)
-target_link_libraries(${test_prefix}storage_v2_replication mg-storage-v2 fmt)
+# add_unit_test(storage_v2_wal_file.cpp)
+# target_link_libraries(${test_prefix}storage_v2_wal_file mg-storage-v2 fmt)
 
+# add_unit_test(storage_v2_replication.cpp)
+# target_link_libraries(${test_prefix}storage_v2_replication mg-storage-v2 fmt)
```
spacejam, line 309: does it make more sense to remove these, rather than have them commented out?

src/storage/v2/vertex_accessor.cpp
jbajic: 1. This is used throughout tests to confirm that the correct `PrimaryLabel` is set
2. I agree I will add that logic for fetching `PrimaryLabel`

tests/unit/interpreter_v2.cpp
```diff
@@ -0,0 +1,1636 @@
...
+//     ASSERT_EQ(stream.GetResults()[1][0].ValueInt(), 2);
+//     ASSERT_EQ(stream.GetResults()[2][0].ValueInt(), 3);
+//     Pull(&stream);
+//     ASSERT_EQ(stream.GetSummary().count("has_more"), 1);
+//     ASSERT_FALSE(stream.GetSummary().at("has_more").ValueBool());
+//     ASSERT_EQ(stream.GetResults().size(), 5U);
+//     ASSERT_EQ(stream.GetResults()[3][0].ValueInt(), 4);
+//     ASSERT_EQ(stream.GetResults()[4][0].ValueInt(), 5);
+//   }
+// }
```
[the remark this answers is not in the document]
jbajic: I forgot about it, when fixing tets I've commented out unrelated parts to spped up

tests/unit/storage_v3_query_plan_bag_semantics.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+using namespace memgraph::query;
+using namespace memgraph::query::plan;
+
+class QueryPlanBagSemanticsTest : public testing::Test {
+ protected:
+  QueryPlanBagSemanticsTest() {
+    EXPECT_TRUE(
+        db.CreateSchema(label, {memgraph::storage::SchemaProperty{property, memgraph::common::SchemaType::INT}}));
+  }
```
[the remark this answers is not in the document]
jbajic: That makes sense

tests/unit/storage_v3_query_plan_create_set_remove_delete.cpp
```diff
@@ -0,0 +1,1116 @@
...
+    dba.InsertVertex();
+    dba.AdvanceCommand();
+    auto produce = MakeProduce(advance, NEXPR("res", LITERAL(42))->MapTo(res_sym));
+    auto context = MakeContext(storage, symbol_table, &dba);
+    EXPECT_EQ(1, PullAll(*produce, &context));
+  }
+  {
+    auto storage_dba = db.Access();
+    memgraph::query::DbAccessor dba(&storage_dba);
+    dba.InsertVertex();
```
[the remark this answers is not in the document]
jbajic: You are right

tests/unit/storage_v3_schema.cpp
```diff
@@ -0,0 +1,310 @@
...
+
+class SchemaTest : public testing::Test {
+ protected:
+  SchemaTest()
+      : prop1{storage.NameToProperty("prop1")},
+        prop2{storage.NameToProperty("prop2")},
+        label1{storage.NameToLabel("label1")},
+        label2{storage.NameToLabel("label2")},
+        schema_prop1{prop1, memgraph::common::SchemaType::STRING},
+        schema_prop2{prop2, memgraph::common::SchemaType::INT} {}
```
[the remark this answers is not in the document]
jbajic: Yes will move this to initialization

tests/unit/CMakeLists.txt
jbajic: Hmm I don't see why, unless we want to explicitly track for some time which tests we have moved and which we haven't

jbajic: What do you say @antaljanosbenjamin?

tests/unit/query_v2_query_plan_match_filter_return.cpp
```diff
@@ -0,0 +1,2062 @@
...
+TEST_F(QueryPlanMatchFilterTest, ScanAllByLabelPropertyRangeNull) {
+  // Add 2 vertices with the same label, but one has a property value while
+  // the other does not. Checking if the value is between nulls, should
+  // yield no results.
+  auto label1 = db.NameToLabel("label1");
+  auto prop = db.NameToProperty("prop");
+  {
+    auto storage_dba = db.Access();
+    DbAccessor dba(&storage_dba);
+    auto vertex = dba.InsertVertex();
```
antaljanosbenjamin, line 1938: Please replace this with `InsertVertexAndValidate`.

tests/unit/CMakeLists.txt
antaljanosbenjamin: > unless we want to explicitly track for some time which tests we have moved and which we haven't

How does this connected to this? I don't see it.

In general I think we should barely touch things that are on master, because we will constantly merging master to project pineapples. So maybe it make sense revert all the changes in v2 to leave it untouched.

src/storage/v2/durability/snapshot.cpp
```diff
@@ -713,8 +713,9 @@ void CreateSnapshot(Transaction *transaction, const std::filesystem::path &snaps
       // type and invalid from/to pointers because we don't know them here,
       // but that isn't an issue because we won't use that part of the API
       // here.
-      auto ea =
-          EdgeAccessor{edge_ref, EdgeTypeId::FromUint(0UL), nullptr, nullptr, transaction, indices, constraints, items};
+      // TODO Fix snapshot with new schema rules @jbajic
```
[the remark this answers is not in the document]
jbajic: I didn't know there was a style guideline for TODOs

tests/unit/query_v2_query_plan_match_filter_return.cpp
jbajic: I've missed that sorry

tests/unit/CMakeLists.txt
jbajic: But we don't seperate tests, and we agreed that the new tests would have prefix storage_v3 or query_v2, but we would keep the old tests as well, the issues is that we cannot leave them running since they will be always failing. So we can either comment them out and always take a quick look to see which ones we had to convert or just delete them, is that clear?

src/storage/v2/vertex_accessor.hpp
```diff
@@ -29,20 +31,35 @@ struct Constraints;
 
 class VertexAccessor final {
  private:
+  struct VertexValidator {
+    explicit VertexValidator(SchemaValidator *schema_validator, const Vertex *vertex)
```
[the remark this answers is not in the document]
jbajic: Currently, it is not possible due to `VertexAccessot` being copied, and if we have a `VertexValidator` that holds a reference object it cannot be copied, take a look at void `LabelIndex::Iterable::Iterator::AdvanceUntilValid() {` in `indices.cpp`. Do you think its worth doing something about it?

src/storage/v2/durability/snapshot.cpp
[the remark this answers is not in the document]
antaljanosbenjamin: I think even clang-tidy warns on this.

src/storage/v2/vertex_accessor.hpp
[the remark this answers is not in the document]
antaljanosbenjamin: I meant only as a constructor parameter, not as a member. So take a reference int the constructor, and then store the address of the object in a member variable. That way I think:
1. it is much more clear for the outside world that they cannot pass into nullptr
2. maybe it also gives a strong hint for the developer that they don't have to deal with nullptr in the member variable.

As I am checking it out, `SchemaValiator` can be passed as `const&`, because all of the interesting member functions are const qualified.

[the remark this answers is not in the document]
jbajic: Aha ok that makes sense, that's why you only marked the constructor, sorry for the misunderstanding

spacejam (approved):
AFAICT this looks good, but I defer to @antaljanosbenjamin wrt the `tests/unit/CMakeLists.txt` structuring feedback which is the main thing that caught my eye

tests/unit/CMakeLists.txt
jbajic: This will be handled in the next PR which moves all the changes do appropriate versions of storage and query engine

antaljanosbenjamin (approved):
LGTM.

src/storage/v2/vertex_accessor.hpp
```diff
@@ -29,20 +31,35 @@ struct Constraints;
...
+        : schema_validator_{schema_validator}, vertex_{vertex} {}
+
+    [[nodiscard]] std::optional<SchemaViolation> ValidatePropertyUpdate(PropertyId property_id) const;
+
+    [[nodiscard]] std::optional<SchemaViolation> ValidateAddLabel(LabelId label) const;
+
+    [[nodiscard]] std::optional<SchemaViolation> ValidateRemoveLabel(LabelId label) const;
+
+    SchemaValidator *schema_validator_;
+    const Vertex *vertex_;
```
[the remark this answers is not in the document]
antaljanosbenjamin: It isn't, because the `VertexAccessor` sometimes instantiated with a `Vertex` pointer that is `nullptr`.

jbajic:
Next PR will move all the changes to query v2 and storage v3!

PR #459, opened 2022-07-19
[master < T0976-MG] Add command to return startup config

gvolfing:

Add command to return startup config

Add a new command that is able to return the set of configurations that that the given instance of memgraph was started up with. The returned information currently consists of the name, the default and the current value of each flag. The hidden property of three flags were removed, namely --query-cost-planner, --query-vertex-count-to-expand-existing and --query-max-plans. The flag --log-link-basename was completely removed since it is not used.

[/docs PR](https://github.com/memgraph/docs/pull/495)


src/query/interpreter.cpp
```diff
@@ -791,6 +791,31 @@ Callback HandleStreamQuery(StreamQuery *stream_query, const Parameters &paramete
...
 }
 
+Callback HandleConfigQuery() {
+  Callback callback;
+  callback.header = {"name", "default_value", "current_value"};
+
+  callback.fn = [] {
+    // Query the flags that the currently running instance was started up with.
+    std::vector<GFLAGS_NAMESPACE::CommandLineFlagInfo> flags;
+    GetAllFlags(&flags);
```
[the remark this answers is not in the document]
gvolfing: I think -currently- it is everything.

+-------------------------------------------------+-------------------------------------------+-------------------------------------------+
| name                                                  | default_value                             | current_value                             |
+-------------------------------------------------+-------------------------------------------+-------------------------------------------+
| "auth_module_create_missing_role"  | "true"                                    | "true"                                    |
| "auth_module_create_missing_user" | "true"                                    | "true"                                    |
| "auth_module_executable"                 | ""                                        | ""                                        |
| "auth_module_manage_roles"           | "true"                                    | "true"                                    |
| "auth_module_timeout_ms"               | "10000"                                   | "10000"                                   |
| "auth_password_permit_null"            | "true"                                    | "true"                                    |
| "auth_password_strength_regex"            | ".+"                                      | ".+"                                      |
| "allow_load_csv"                          | "true"                                    | "true"                                    |
| "also_log_to_stderr"                      | "false"                                   | "true"                                    |
| "audit_buffer_flush_interval_ms"          | "200"                                     | "200"                                     |
| "audit_buffer_size"                       | "100000"                                  | "100000"                                  |
| "audit_enabled"                           | "false"                                   | "false"                                   |
| "auth_user_or_role_name_regex"            | "[a-zA-Z0-9_.+-@]+"                       | "[a-zA-Z0-9_.+-@]+"                       |
| "bolt_address"                            | "0.0.0.0"                                 | "0.0.0.0"                                 |
| "bolt_cert_file"                          | ""                                        | ""                                        |
| "bolt_key_file"                           | ""                                        | ""                                        |
| "bolt_num_workers"                        | "12"                                      | "12"                                      |
| "bolt_port"                               | "7687"                                    | "7687"                                    |
| "bolt_server_name_for_init"               | ""                                        | ""                                        |
| "bolt_session_inactivity_timeout"         | "1800"                                    | "1800"                                    |
| "data_directory"                          | "mg_data"                                 | "mg_data"                                 |
| "isolation_level"                         | "SNAPSHOT_ISOLATION"                      | "SNAPSHOT_ISOLATION"                      |
| "kafka_bootstrap_servers"                 | ""                                        | ""                                        |
| "license_key"                             | ""                                        | ""                                        |
| "log_file"                                | ""                                        | ""                                        |
| "log_level"                               | "WARNING"                                 | "TRACE"                                   |
| "log_link_basename"                       | ""                                        | ""                                        |
| "memory_limit"                            | "0"                                       | "0"                                       |
| "memory_warning_threshold"                | "1024"                                    | "1024"                                    |
| "monitoring_address"                      | "0.0.0.0"                                 | "0.0.0.0"                                 |
| "monitoring_port"                         | "7444"                                    | "7444"                                    |
| "organization_name"                       | ""                                        | ""                                        |
| "pulsar_service_url"                      | ""                                        | ""                                        |
| "query_execution_timeout_sec"             | "600"                                     | "600"                                     |
| "query_modules_directory"                 | ""                                        | ""                                        |
| "replication_replica_check_frequency_sec" | "1"                                       | "1"                                       |
| "storage_gc_cycle_sec"                    | "30"                                      | "30"                                      |
| "storage_properties_on_edges"             | "false"                                   | "false"                                   |
| "storage_recover_on_startup"              | "false"                                   | "false"                                   |
| "storage_restore_replicas_on_startup"     | "true"                                    | "true"                                    |
| "storage_snapshot_interval_sec"           | "0"                                       | "0"                                       |
| "storage_snapshot_on_exit"                | "false"                                   | "false"                                   |
| "storage_snapshot_retention_count"        | "3"                                       | "3"                                       |
| "storage_wal_enabled"                     | "false"                                   | "false"                                   |
| "storage_wal_file_flush_every_n_tx"       | "100000"                                  | "100000"                                  |
| "storage_wal_file_size_kib"               | "20480"                                   | "20480"                                   |
| "stream_transaction_conflict_retries"     | "30"                                      | "30"                                      |
| "stream_transaction_retry_interval"       | "500"                                     | "500"                                     |
| "telemetry_enabled"                       | "false"                                   | "false"                                   |
| "query_cost_planner"                      | "true"                                    | "true"                                    |
| "query_plan_cache_ttl"                    | "60"                                      | "60"                                      |
| "query_vertex_count_to_expand_existing"   | "10"                                      | "10"                                      |
| "query_max_plans"                         | "1000"                                    | "1000"                                    |
| "flag_file"                               | ""                                        | ""                                        |
| "help"                                    | "false"                                   | "false"                                   |
| "help_xml"                                | "false"                                   | "false"                                   |
| "version"                                 | "false"                                   | "false"                                   |
+-------------------------------------------+-------------------------------------------+-------------------------------------------+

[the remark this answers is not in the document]
gvolfing: **Conclusion:**
We will not show hidden flags either way.

src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -2342,6 +2342,12 @@ antlrcpp::Any CypherMainVisitor::visitForeach(MemgraphCypher::ForeachContext *ct
   return for_each;
 }
 
+antlrcpp::Any CypherMainVisitor::visitShowConfigQuery(MemgraphCypher::ShowConfigQueryContext * /*ctx*/) {
+  auto *show_config_query = storage_->Create<ShowConfigQuery>();
```
kostasrim, line 2346:
```suggestion
  query_ = storage_->Create<ShowConfigQuery>();
  return query_
```

src/query/interpreter.cpp
```diff
@@ -791,6 +791,33 @@ Callback HandleStreamQuery(StreamQuery *stream_query, const Parameters &paramete
   }
 }
 
+Callback HandleConfigQuery() {
+  Callback callback;
+  callback.header = {"name", "default_value", "current_value"};
```
kostasrim, line 796: I think `SHOW CONFIG` printing both the default and current config is redundant. A couple of options to consider: 

1. Introduce `SHOW DEFAULT CONFIG` which basically prints the default values
2. Remove the `default_value` completely
3. Keep things as is. 

Any comments on this @gitbuda and @antaljanosbenjamin?

kostasrim, line 796:
```suggestion
  callback.header = {"name", "default_configuration", "current_configuration"};
```

src/query/interpreter.cpp
```diff
@@ -791,6 +791,33 @@ Callback HandleStreamQuery(StreamQuery *stream_query, const Parameters &paramete
...
+
+  callback.fn = [] {
+    // Query the flags that the currently running instance was started up with.
```
kostasrim, line 799:
```suggestion
```

src/query/interpreter.cpp
```diff
@@ -1521,6 +1548,28 @@ PreparedQuery PrepareFreeMemoryQuery(ParsedQuery parsed_query, const bool in_exp
...
+  if (in_explicit_transaction) {
+    throw ShowConfigModificationInMulticommandTxException();
+  }
+
+  auto callback = HandleConfigQuery();
+
+  return PreparedQuery{std::move(callback.header), std::move(parsed_query.required_privileges),
+                       [callback_fn = std::move(callback.fn), pull_plan = std::shared_ptr<PullPlanVector>{nullptr}](
+                           AnyStream *stream, std::optional<int> n) mutable -> std::optional<QueryHandlerResult> {
+                         if (UNLIKELY(!pull_plan)) {
```
kostasrim, line 1561: Use the builtin C++ attribute https://en.cppreference.com/w/cpp/language/attributes/likely 

@antaljanosbenjamin I will make a quick PR when I get a chance to remove those 2 from the codebase.

src/query/interpreter.cpp
```diff
@@ -1521,6 +1548,28 @@ PreparedQuery PrepareFreeMemoryQuery(ParsedQuery parsed_query, const bool in_exp
...
 
+PreparedQuery PrepareShowConfigQuery(ParsedQuery parsed_query, const bool in_explicit_transaction) {
+  if (in_explicit_transaction) {
+    throw ShowConfigModificationInMulticommandTxException();
+  }
+
+  auto callback = HandleConfigQuery();
+
+  return PreparedQuery{std::move(callback.header), std::move(parsed_query.required_privileges),
+                       [callback_fn = std::move(callback.fn), pull_plan = std::shared_ptr<PullPlanVector>{nullptr}](
```
kostasrim, line 1559: `std::move(callback.fn)` is redundant. Why are you moving? The lambda does not have a state and therefore is trivially copyable.

tests/unit/cypher_main_visitor.cpp
```diff
@@ -4124,6 +4124,19 @@ TEST_P(CypherMainVisitorTest, VersionQuery) {
   ASSERT_NO_THROW(ast_generator.ParseQuery("SHOW VERSION"));
 }
 
+TEST_P(CypherMainVisitorTest, ConfigQuery) {
```
kostasrim, line 4127: I think this is nice but plz add an `e2e` test to confirm the expected behavior, that is, that the values returned are the config values used when memgraph was launched,

src/query/interpreter.cpp
```diff
@@ -791,6 +791,33 @@ Callback HandleStreamQuery(StreamQuery *stream_query, const Parameters &paramete
   }
 }
 
+Callback HandleConfigQuery() {
```
kostasrim, line 794: Currently, this includes the `license_key`. Is this an issue?

src/query/interpreter.cpp
kostasrim: Actually I think I am wrong on this one because fn is a `std::function` which dynamically allocates memory. I will take a look in the morning, I am from my phone :D

src/query/interpreter.cpp
gvolfing: Just to add to this question, if modified, is there anything else we would like to add?

- std::string name;            // the name of the flag
- std::string type;            // the type of the flag: int32, etc
- std::string description;     // the "help text" associated with the flag
- std::string current_value;   // the current value, as a string
- std::string default_value;   // the default value, as a string
- std::string filename;        // 'cleaned' version of filename holding the flag

description might make sense.

src/query/interpreter.cpp
gvolfing: Currently we return all 56 start-up flags. Used to be 57, 4 of them used to be hidden we unhide 3 of them and deleted one.

Since the license and the org. name is allready available through `SHOW DATABASE SETTINGS;` I am guessing they should be made hidden. Is there anything else we would like to hide from the user? 

| **"auth_module_create_missing_role"**                                                                                                                                                                                                          | "Set to false to disable creation of missing roles."                                                                                                                                                                                                   |
| **"auth_module_create_missing_user"**                                                                                                                                                                                                                      | "Set to false to disable creation of missing users."                                                                                                                                                                                                   |
| **"auth_module_executable"**                                                                                                                                                                                                                               | "Absolute path to the auth module executable that should be used."                                                                                                                                                                                     |
| **"auth_module_manage_roles"**                                                                                                                                                                                                                             | "Set to false to disable management of roles through the auth module."                                                                                                                                                                                 |
| **"auth_module_timeout_ms"**                                                                                                                                                                                                                               | "Timeout (in milliseconds) used when waiting for a response from the auth module."                                                                                                                                                                     |
| **"auth_password_permit_null"**                                                                                                                                                                                                                            | "Set to false to disable null passwords."                                                                                                                                                                                                              |
| **"auth_password_strength_regex"**                                                                                                                                                                                                                         | "The regular expression that should be used to match the entire entered password to ensure its strength."                                                                                                                                              |
| **"allow_load_csv"**                                                                                                                                                                                                                                       | "Controls whether LOAD CSV clause is allowed in queries."                                                                                                                                                                                              |
| **"also_log_to_stderr"**                                                                                                                                                                                                                                   | "Log messages go to stderr in addition to logfiles"                                                                                                                                                                                                    |
| **"audit_buffer_flush_interval_ms"**                                                                                                                                                                                                                       | "Interval (in milliseconds) used for flushing the audit log buffer."                                                                                                                                                                                   |
| **"audit_buffer_size"**                                                                                                                                                                                                                                    | "Maximum number of items in the audit log buffer."                                                                                                                                                                                                     |
| **"audit_enabled"**                                                                                                                                                                                                                                        | "Set to true to enable audit logging."                                                                                                                                                                                                                 |
| **"auth_user_or_role_name_regex"**                                                                                                                                                                                                                         | "Set to the regular expression that each user or role name must fulfill."                                                                                                                                                                              |
| **"bolt_address"**                                                                                                                                                                                                                                         | "IP address on which the Bolt server should listen."                                                                                                                                                                                                   |
| **"bolt_cert_file"**                                                                                                                                                                                                                                       | "Certificate file which should be used for the Bolt server."                                                                                                                                                                                           |
| **"bolt_key_file"**                                                                                                                                                                                                                                        | "Key file which should be used for the Bolt server."                                                                                                                                                                                                   |
| **"bolt_num_workers"**                                                                                                                                                                                                                                     | "Number of workers used by the Bolt server. By default, this will be the number of processing units available on the machine."                                                                                                                         |
| **"bolt_port"**                                                                                                                                                                                                                                            | "Port on which the Bolt server should listen."                                                                                                                                                                                                         |
| **"bolt_server_name_for_init"**                                                                                                                                                                                                                            | "Server name which the database should send to the client in the Bolt INIT message."                                                                                                                                                                   |
| **"bolt_session_inactivity_timeout"**                                                                                                                                                                                                                      | "Time in seconds after which inactive Bolt sessions will be closed."                                                                                                                                                                                   |
| **"data_directory"**                                                                                                                                                                                                                                       | "Path to directory in which to save all permanent data."                                                                                                                                                                                               |
| **"isolation_level"**                                                                                                                                                                                                                                      | "Default isolation level used for the transactions. Allowed values: SNAPSHOT_ISOLATION, READ_COMMITTED, READ_UNCOMMITTED"                                                                                                                              |
| **"kafka_bootstrap_servers"**                                                                                                                                                                                                                              | "List of default Kafka brokers as a comma separated list of broker host or host:port."                                                                                                                                                                 |
| **"license_key"**                                                                                                                                                                                                                                          | "License key for Memgraph Enterprise."                                                                                                                                                                                                                 |
| **"log_file"**                                                                                                                                                                                                                                             | "Path to where the log should be stored."                                                                                                                                                                                                              |
| **"log_level"**                                                                                                                                                                                                                                            | "Minimum log level. Allowed values: TRACE, DEBUG, INFO, WARNING, ERROR, CRITICAL"                                                                                                                                                                      |
| **"memory_limit"**                                                                                                                                                                                                                                         | "Total memory limit in MiB. Set to 0 to use the default values which are 100% of the phyisical memory if the swap is enabled and 90% of the physical memory otherwise."                                                                                |
| **"memory_warning_threshold"**                                                                                                                                                                                                                             | "Memory warning threshold, in MB. If Memgraph detects there is less available RAM it will log a warning. Set to 0 to disable."                                                                                                                         |
| **"monitoring_address"**                                                                                                                                                                                                                                   | "IP address on which the websocket server for Memgraph monitoring should listen."                                                                                                                                                                      |
| **"monitoring_port"**                                                                                                                                                                                                                                      | "Port on which the websocket server for Memgraph monitoring should listen."                                                                                                                                                                            |
| **"organization_name"**                                                                                                                                                                                                                                    | "Organization name."                                                                                                                                                                                                                                   |
| **"pulsar_service_url"**                                                                                                                                                                                                                                   | "Default URL used while connecting to Pulsar brokers."                                                                                                                                                                                                 |
| **"query_execution_timeout_sec"**                                                                                                                                                                                                                          | "Maximum allowed query execution time. Queries exceeding this limit will be aborted. Value of 0 means no limit."                                                                                                                                       |
| **"query_modules_directory"**                                                                                                                                                                                                                              | "Directory where modules with custom query procedures are stored. NOTE: Multiple comma-separated directories can be defined."                                                                                                                          |
| **"replication_replica_check_frequency_sec"**                                                                                                                                                                                                              | "The time duration between two replica checks/pings. If < 1, replicas will NOT be checked at all. NOTE: The MAIN instance allocates a new thread for each REPLICA."                                                                                    |
| **"storage_gc_cycle_sec"**                                                                                                                                                                                                                                 | "Storage garbage collector interval (in seconds)."                                                                                                                                                                                                     |
| **"storage_properties_on_edges"**                                                                                                                                                                                                                          | "Controls whether edges have properties."                                                                                                                                                                                                              |
| **"storage_recover_on_startup"**                                                                                                                                                                                                                           | "Controls whether the storage recovers persisted data on startup."                                                                                                                                                                                     |
| **"storage_restore_replicas_on_startup"**                                                                                                                                                                                                                  | "Controls replicas should be restored automatically."                                                                                                                                                                                                  |
| **"storage_snapshot_interval_sec"**                                                                                                                                                                                                                        | "Storage snapshot creation interval (in seconds). Set to 0 to disable periodic snapshot creation."                                                                                                                                                     |
| **"storage_snapshot_on_exit"**                                                                                                                                                                                                                             | "Controls whether the storage creates another snapshot on exit."                                                                                                                                                                                       |
| **"storage_snapshot_retention_count"**                                                                                                                                                                                                                     | "The number of snapshots that should always be kept."                                                                                                                                                                                                  |
| **"storage_wal_enabled"**                                                                                                                                                                                                                                  | "Controls whether the storage uses write-ahead-logging. To enable WAL periodic snapshots must be enabled."                                                                                                                                             |
| **"storage_wal_file_flush_every_n_tx"**                                                                                                                                                                                                                    | "Issue a \'fsync\' call after this amount of transactions are written to the WAL file. Set to 1 for fully synchronous operation."                                                                                                                      |
| **"storage_wal_file_size_kib"**                                                                                                                                                                                                                            | "Minimum file size of each WAL file."                                                                                                                                                                                                                  |
| **"stream_transaction_conflict_retries"**                                                                                                                                                                                                                  | "Number of times to retry when a stream transformation fails to commit because of conflicting transactions"                                                                                                                                            |
| **"stream_transaction_retry_interval"**                                                                                                                                                                                                                    | "Retry interval in milliseconds when a stream transformation fails to commit because of conflicting transactions"                                                                                                                                      |
| **"telemetry_enabled"**                                                                                                                                                                                                                                    | "Set to true to enable telemetry. We collect information about the running system (CPU and memory information) and information about the database runtime (vertex and edge counts and resource usage) to allow for easier improvement of the product." |
| **"query_cost_planner"**                                                                                                                                                                                                                                   | "Use the cost-estimating query planner."                                                                                                                                                                                                               |
| **"query_plan_cache_ttl"**                                                                                                                                                                                                                                 | "Time to live for cached query plans, in seconds."                                                                                                                                                                                                     |
| **"query_vertex_count_to_expand_existing"**                                                                                                                                                                                                                | "Maximum count of indexed vertices which provoke indexed lookup and then expand to existing, instead of a regular expand. Default is 10, to turn off use -1."                                                                                          |
| **"query_max_plans"**                                                                                                                                                                                                                                      | "Maximum number of generated plans for a query."                                                                                                                                                                                                       |
| **"flag_file"**                                                                                                                                                                                                                                            | "load flags from file"                                                                                                                                                                                                                                 |
| **"help"**                                                                                                                                                                                                                                                 | "show help on all flags [tip: all flags can have two dashes]"                                                                                                                                                                                          |
| **"help_xml"**                                                                                                                                                                                                                                             | "produce an xml version of help"                                                                                                                                                                                                                       |
| **"version"**                                                                                                                                                                                                                                              | "show version and build info and exit"                                                                                                                                                                                                                 |

src/query/frontend/opencypher/grammar/MemgraphCypher.g4
```diff
@@ -79,6 +79,7 @@ memgraphCypherKeyword : cypherKeyword
                       | SESSION
                       | SETTING
                       | SETTINGS
+                      | SHOW_CONFIG
```
antaljanosbenjamin, line 82: I think this is wrong. In the query you use them as separate words (`SHOW CONFIG`), and not as one (like `BATCH_LIMIT`), thus I think you need `SHOW` and `CONFIG` as separate words. I think both words are there (`SHOW` is in `Cypher.g4` in the `cypherKeyword` rule, `CONFIG` is here). So I think you don't have to modify the keywords unless you want to introduce another word.

src/query/frontend/opencypher/grammar/MemgraphCypherLexer.g4
```diff
@@ -93,6 +93,7 @@ SERVICE_URL         : S E R V I C E UNDERSCORE U R L ;
 SESSION             : S E S S I O N ;
 SETTING             : S E T T I N G ;
 SETTINGS            : S E T T I N G S ;
+SHOW_CONFIG         : S H O W UNDERSCORE C O N F I G ;
```
antaljanosbenjamin, line 96: Similar to my other comment: as you are using them separately, you have to have the separate words, but `SHOW` and `CONFIG` are already there (CypherLexer.g4 and in this file).

src/query/frontend/stripped_lexer_constants.hpp
```diff
@@ -204,6 +204,7 @@ const trie::Trie kKeywords = {"union",
                               "pulsar",
                               "service_url",
                               "version",
+                              "config",
```
antaljanosbenjamin, line 207: It was no here, good catch.

src/query/interpreter.cpp
```diff
@@ -791,6 +791,32 @@ Callback HandleStreamQuery(StreamQuery *stream_query, const Parameters &paramete
...
+  callback.header = {"name", "default_configuration", "current_configuration"};
+
+  callback.fn = [] {
+    std::vector<GFLAGS_NAMESPACE::CommandLineFlagInfo> flags;
+    GetAllFlags(&flags);
+
+    std::vector<std::vector<TypedValue>> results;
+
+    for (const auto &flag : flags) {
+      if (flag.hidden) continue;
```
antaljanosbenjamin, line 805:
```suggestion
      if (flag.hidden) {
        continue;
      }
```

In general I think we prefer to write out the braces to make sure we are safe if we want to extend the condition.

src/query/interpreter.cpp
antaljanosbenjamin: Yes, you are wrong. I think std::move here is completely fine.

tests/unit/cypher_main_visitor.cpp
```diff
@@ -4124,6 +4124,19 @@ TEST_P(CypherMainVisitorTest, VersionQuery) {
...
+  auto &ast_generator = *GetParam();
+
+  TestInvalidQuery("SHOW CF", ast_generator);
+  TestInvalidQuery("SHOW CFG", ast_generator);
+  TestInvalidQuery("SHOW CFGS", ast_generator);
+  TestInvalidQuery("SHOW CONF", ast_generator);
+  TestInvalidQuery("SHOW CONFIGS", ast_generator);
+  TestInvalidQuery("SHOW CONFIGURATION", ast_generator);
+  TestInvalidQuery("SHOW CONFIGURATIONS", ast_generator);
+  ASSERT_NO_THROW(ast_generator.ParseQuery("SHOW CONFIG"));
```
antaljanosbenjamin, line 4137: Maybe do a typecheck just to be sure the correct query is created.

src/query/interpreter.cpp
antaljanosbenjamin: @kostasrim I think showing the default value is okay, it is something people might not know from top of their heart.

@gvolfing I think the description makes sense to output. Another thing to consider (it is just an idea), is to format the flags into `--lower-dash-case-format` instead of `current_sneak_case`. I think users might happen to copy these flags and it is easy to miss the difference between `_` and `-`.

antaljanosbenjamin (changes requested):

src/query/interpreter.cpp
gvolfing: The returned value of the first column is the _name_ of the given flag not the flag itself.
`snake_case_flag` is the way they are used inside the `std::string`s of gFlag's `CommandLineFlagInfo` and in other places such as `memgraph/config/flags.yaml`.
That being said, I can change it if you insist.

src/query/interpreter.cpp
```diff
@@ -791,6 +791,37 @@ Callback HandleStreamQuery(StreamQuery *stream_query, const Parameters &paramete
...
+  callback.fn = [] {
+    std::vector<GFLAGS_NAMESPACE::CommandLineFlagInfo> flags;
+    GetAllFlags(&flags);
+
+    std::vector<std::vector<TypedValue>> results;
+
+    for (const auto &flag : flags) {
+      if (flag.hidden ||
+          // These flags are not defined with gflags macros but are specified in config/flags.yaml
+          flag.name == "flag_file" || flag.name == "help" || flag.name == "help_xml" || flag.name == "version") {
```
antaljanosbenjamin, line 807: Why the `flag_file`? Cannot that have a value?

src/query/interpreter.cpp
antaljanosbenjamin: I wasn't aware of that, haven't used the flag files more than once or two. I think based on what you said, it is okay to have them in `sneak_case`.

tests/e2e/configuration/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+function(copy_configuration_check_e2e_python_files FILE_NAME)
+    copy_e2e_python_files(write_procedures ${FILE_NAME})
+endfunction()
+
+copy_write_procedures_e2e_python_files(default_config.py)
+copy_write_procedures_e2e_python_files(configuration_check.py)
```
antaljanosbenjamin, line 6:
```suggestion
copy_configuration_check_e2e_python_files(default_config.py)
copy_configuration_check_e2e_python_files(configuration_check.py)
```

tests/e2e/configuration/default_config.py
```diff
@@ -0,0 +1,187 @@
...
+# Use of this software is governed by the Business Source License
+# included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
+# License, and you may not use this file except in compliance with the Business Source License.
+#
+# As of the Change Date specified in that file, in accordance with
+# the Business Source License, use of this software will be governed
+# by the Apache License, Version 2.0, included in the file
+# licenses/APL.txt.
+
+startup_config = [
```
antaljanosbenjamin, line 12: :+1: for separating this behemot into a file!

antaljanosbenjamin (changes requested):

tests/e2e/configuration/configuration_check.py
```diff
@@ -0,0 +1,33 @@
...
+def test_does_default_config_match():
+    connection = mgclient.connect(host="localhost", port=7687)
+    connection.autocommit = True
+
+    cursor = connection.cursor()
+    cursor.execute("SHOW CONFIG")
+    config = cursor.fetchall()
+
+    for flag in config:
+        assert config[1] == default_config.startup_config[1]
```
antaljanosbenjamin, line 29: This check is not good, it only checks the first config option. Maybe 1.) check their sizes are equal; 2.) iterator over them by index.

src/query/interpreter.cpp
gvolfing: I probably included that one by accident. Will remove.

tests/e2e/configuration/configuration_check.py
gvolfing: Good catch, I don't know what I was thinking :smile:
Also added the size check.

tests/e2e/configuration/configuration_check.py
```diff
@@ -25,8 +24,23 @@ def test_does_default_config_match():
...
+                # This will always depend on the number of processing units available on the given machine
+                if flag[0] == "bolt_num_workers":
+                    continue
+
+                current_flag = flag
+                break
+
+        if current_flag != ():
+            assert len(current_flag) == len(default_config.startup_config[idx])
+            assert current_flag[1] == default_config.startup_config[idx][1]
```
antaljanosbenjamin, line 43: In default config you have listed the current values and descriptions. I either simplify that structure or add checks here.

antaljanosbenjamin (changes requested):

tests/e2e/configuration/configuration_check.py
gvolfing: I am not sure if I'm understand the reasoning behind this. The purpose of this test is to check if the default config have changed or not, as it was requested by @kostasrim.
"either simplify that structure or add checks here" - Do you mean to simplify the list of tuples that is in `/configuration/default_config.py` maybe into just a list of values? Should it just be a list of `default_values`?

I could check against the description, if that changed or not,
I see no point in checking against the name of the flags because I need them to make sure I find the individual tuples in case the different machines shuffle the order of them(happens),
So basically I should only remove the `current_value`s from the `default_config`.

To sum it up, on individual flags
1. Check if the `default_config` has changed or not
2. Check if the `description` has changed or not
3. Remove the `current_config` from the `default_config`

is that correct? @antaljanosbenjamin

src/query/interpreter.cpp
kostasrim: This looks good

tests/e2e/configuration/configuration_check.py
kostasrim: The test is not about checking if the default config has changed or not but rather if the implementation of `SHOW CONFIG` does what is intended to do. You start memgraph with a certain configuration and the values of that configuration are displayed in the `current_config` column of the `SHOW CONFIG` + the default values and the description. I would make `startup_config` a dictionary (description will be the key), and then in the test I would look the `default` && `current` values of each description. That way you can assert/cross-reference the expected configuration and the one you fetched from `SHOW CONFIG`

gvolfing: **Conlusion**:
I will create a dictionary out of the `startup_config`, use the name of the flag as the key, and a tuple of `default_value`, `current_value`, `description` as the key.

antaljanosbenjamin (approved):

kostasrim (approved):
Good job!

gitbuda (approved):
🚀 💪

PR #463, opened 2022-07-21
[master <  T0986-MG] Add parser stress test

gitbuda:
- [x] Simplify the test
- [x] Introduce and upgrade mgclient
- [x] Compare `mgclient` and `memgraph::communication::bolt::Client`


tests/stress/continuous_integration
```diff
@@ -26,6 +26,11 @@ SMALL_DATASET = [
         "options": ["--vertex-count", "40000", "--create-pack-size", "100"],
         "timeout": 5,
     },
+    {
+        "test": "parser.cpp",
+        "options": ["--client-count", "12", "--per-client-query-count", "1000"],
+        "timeout": 5,
+    },
```
gitbuda, line 33: This is the only change in this file, everything else was the pre-commit `black` hook

gitbuda:
Running `./tests/stress/parser --client-count 8 --per-client-query-count 1000` takes:
* with `memgraph::communication::bolt::Client` 0.42s
* with `mgclient` 0.48s

which is fine in my view, we should use `mgclient` everywhere in the codebase.

tests/stress/parser.cpp
```diff
@@ -0,0 +1,81 @@
...
+  spdlog::info(fmt::format("Starting parser stress test with {} clients and {} queries per client...",
+                           FLAGS_client_count, FLAGS_per_client_query_count));
+  std::vector<std::thread> threads;
+  memgraph::utils::Timer timer;
+  for (int i = 0; i < FLAGS_client_count; ++i) {
+    threads.push_back(std::thread([]() {
+      auto client = make_client();
+      std::mt19937 generator{std::random_device{}()};
+      std::uniform_int_distribution<uint64_t> distribution{std::numeric_limits<uint64_t>::min(),
+                                                           std::numeric_limits<uint64_t>::max()};
```
jbajic, line 58: Move outside of loop, or did you intentionally do this to get a different seed every time?

tests/stress/parser.cpp
```diff
@@ -0,0 +1,81 @@
...
+          LOG_FATAL("One of the parser stress test queries failed.");
+        }
+      }
+    }));
+  }
+
+  for (int i = 0; i < FLAGS_client_count; ++i) {
+    threads[i].join();
+  }
+  auto elapsed = timer.Elapsed();
```
jbajic, line 76:
```suggestion
  const auto elapsed = timer.Elapsed();
```

tests/stress/parser.cpp
```diff
@@ -0,0 +1,81 @@
...
+        } catch (const std::exception &e) {
+          LOG_FATAL("One of the parser stress test queries failed.");
+        }
+      }
+    }));
+  }
+
+  for (int i = 0; i < FLAGS_client_count; ++i) {
+    threads[i].join();
+  }
```
jbajic, line 75:
```suggestion
std::ranges::for_each(threads, [](auto &thread) { thread.join(); });
```

tests/stress/parser.cpp
gitbuda: I actually don't need that variable at all, I'll just inline the call in the next line 😄

tests/stress/parser.cpp
gitbuda: This is nice! Done!

tests/stress/parser.cpp
jbajic: Oh sorry didn't noticed that is in lambda while commenting

jbajic (approved):
Looks good!

PR #466, opened 2022-07-21
[project-pineapples < T0879-MG] Transport prototype

spacejam:
[master < Epic] PR
- [x] Write E2E tests

[master < Task] PR


src/io/simulator/simulator_handle.hpp
```diff
@@ -0,0 +1,453 @@
...
+  }
+};
+
+class OpaquePromise {
+  const std::type_info *ti_;
+  void *ptr_;
+  std::function<void(void *)> dtor_;
+  std::function<bool(void *)> is_awaited_;
+  std::function<void(void *, OpaqueMessage)> fill_;
+  std::function<void(void *)> time_out_;
```
antaljanosbenjamin, line 108: I think 4 `std::function` is a good indicator to introduce virtual inheritance :smiley:
Something like this should work:
```
class OpaquePromiseTraitBase {
    virtual const std::type_info * TypeInfo() const = 0;
    virtual void Destroy(void* ptr) const = 0;
    virtual bool IsAwaited(void* ptr) const = 0;
    virtual bool Fill(void* ptr, OpaqueMessage) const = 0;
    virtual void Timeout() const = 0;
};

template <typename T>
class OpaquePromieTrait : public OpaquePromiseTraitBase {
     const std::type_info * TypeInfo() const override { return &typeid(T);};

     void Destroy(void* ptr) const override {
        // I think you need to call delete on the pointer,, as the destructor won't free the memory.
        delete static_cast<ResponsePromise<T> *>(ptr);
     };

     bool IsAwaited(void* ptr) const override {
        return static_cast<ResponsePromise<T> *>(ptr)->IsAwaited(); 
     };
     
     bool Fill(void* ptr, OpaqueMessage msg) const override {
        // ...
     };
     
     void Timeout() const override {
        // ...
     }
};


template <typename T>
class OpaquePromise {
  // Of course, ctor is necessary but left it out now
  void *ptr_;
  std::unique_ptr<OpaquePromiseTraitBase> trait_{std::make_unique<OpaquePromieTrait<T>>()};
};
```

antaljanosbenjamin: As a bonus, you can also default the move operators.

src/io/simulator/simulator_handle.hpp
```diff
@@ -0,0 +1,453 @@
...
+
+  ~OpaquePromise() {
+    if (nullptr != ptr_) {
+      dtor_(ptr_);
+    }
+  }
+};
+
+struct DeadlineAndOpaquePromise {
+  uint64_t deadline;
```
antaljanosbenjamin, line 198: Please use `std::chrono`.

src/io/future.hpp
```diff
@@ -0,0 +1,257 @@
...
+        simulator_progressed = (*simulator_notifier_)();
+        lock.lock();
+        if (item_) {
+          // item may have been filled while we
+          // had dropped our mutex while notifying
+          // the simulator of our waiting_ status.
+          break;
+        }
+      }
+      if (!simulator_progressed) {
```
antaljanosbenjamin, line 77:
```suggestion
      if (!simulator_progressed) [[likely]] {
```

src/io/simulator/simulator_handle.hpp
```diff
@@ -0,0 +1,453 @@
...
+  // messages that have not yet been scheduled or dropped
+  std::vector<std::pair<Address, OpaqueMessage>> in_flight_;
+
+  // the responsese to requests that are being waited on
+  std::map<PromiseKey, DeadlineAndOpaquePromise> promises_;
+
+  // messages that are sent to servers that may later receive them
+  std::map<Address, std::vector<OpaqueMessage>> can_receive_;
+
+  uint64_t cluster_wide_time_microseconds_;
```
antaljanosbenjamin, line 215: I won't write it more, but please use `std::chrono` everywhere for time related things.

src/io/simulator/simulator_handle.hpp
```diff
@@ -0,0 +1,453 @@
...
+};
+
+class SimulatorHandle {
+  std::mutex mu_{};
+  std::condition_variable cv_;
+
+  // messages that have not yet been scheduled or dropped
+  std::vector<std::pair<Address, OpaqueMessage>> in_flight_;
+
+  // the responsese to requests that are being waited on
```
antaljanosbenjamin, line 209:
```suggestion
  // the responses to requests that are being waited on
```

src/io/simulator/simulator_handle.hpp
```diff
@@ -0,0 +1,453 @@
...
+    while (true) {
+      size_t blocked_servers = blocked_on_receive_;
+
+      for (auto &[promise_key, opaque_promise] : promises_) {
+        if (opaque_promise.promise.IsAwaited()) {
+          if (server_addresses_.contains(promise_key.requester_address)) {
+            blocked_servers++;
+          }
+        }
+      }
```
antaljanosbenjamin, line 240: Maybe extract this logic to a member function because of two reasons:
1. It can be reused in `MaybeTickSimulator`
2. You can give it a nice name to explain its meaning: `GetBlockedServersOnWait()`, if I am not mistaken.

src/io/simulator/simulator_handle.hpp
```diff
@@ -0,0 +1,453 @@
...
+      if (all_servers_blocked) {
+        return;
+      }
+
+      cv_.wait(lock);
+    }
+  }
+
+  void TimeoutPromisesPastDeadline() {
+    uint64_t now = cluster_wide_time_microseconds_;
```
antaljanosbenjamin, line 253: Please use const.

src/io/simulator/simulator_handle.hpp
```diff
@@ -0,0 +1,453 @@
...
+
+    for (auto &[promise_key, dop] : promises_) {
+      // TODO(tyler) queue this up and drop it after its deadline
+      if (dop.deadline < now) {
+        std::cout << "timing out request" << std::endl;
+        DeadlineAndOpaquePromise dop = std::move(promises_.at(promise_key));
```
antaljanosbenjamin, line 259:
```suggestion
        DeadlineAndOpaquePromise dop = std::move(dop);
```
Shouldn't this work also apart from variable shadowing? Please be careful with shadowing.

antaljanosbenjamin (changes requested):
I will finish the review on Monday morning.

src/io/simulator/simulator_handle.hpp
```diff
@@ -0,0 +1,453 @@
...
+                "Cluster has executed beyond its configured abort_time, and something may be failing to make progress "
+                "in an expected amount of time.");
+      return true;
+    }
+
+    if (config_.scramble_messages) {
+      // scramble messages
+      std::uniform_int_distribution<size_t> swap_distrib(0, in_flight_.size() - 1);
+      size_t swap_index = swap_distrib(rng_);
+      std::swap(in_flight_[swap_index], in_flight_.back());
```
antaljanosbenjamin, line 314: Have you thought about using [`std::shuffle`](https://en.cppreference.com/w/cpp/algorithm/random_shuffle)? With that you can scramble all the messages easily. Or is there a reason why to change only one message?

src/io/transport.hpp
```diff
@@ -0,0 +1,130 @@
...
+  std::variant<Ms...> message;
+  uint64_t request_id;
+  Address from_address;
+};
+
+template <Message... Ms>
+using RequestResult = BasicResult<TimedOut, RequestEnvelope<Ms...>>;
+
+template <typename I>
+class Io {
```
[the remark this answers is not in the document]
antaljanosbenjamin: Maybe add a small explanation about what does Request/Receive/Send is used for:
1. Request is to request something from the server on the other side of the `Io` class
2. Receive is to wait for incoming messages from any other servers
3. Send is to send a message to a specific server without expecting any response

Even though we talked a lot about transport layer, it was not clear for me which one should be used on the "client" and "server" side, or what exactly is their goal.

src/io/simulator/simulator_handle.hpp
```diff
@@ -0,0 +1,453 @@
...
+    }
+
+    auto [to_address, opaque_message] = std::move(in_flight_.back());
+    in_flight_.pop_back();
+
+    std::uniform_int_distribution<int> drop_distrib(0, 99);
+    int drop_threshold = drop_distrib(rng_);
+    bool should_drop = drop_threshold < config_.drop_percent;
```
antaljanosbenjamin, line 322: Please apply const where it is possible. It helps us to read the code and the compiler to understand your intention better.

src/io/simulator/simulator_handle.hpp
```diff
@@ -0,0 +1,453 @@
...
+    if (promises_.contains(promise_key)) {
+      // complete waiting promise if it's there
+      DeadlineAndOpaquePromise dop = std::move(promises_.at(promise_key));
+      promises_.erase(promise_key);
+
+      bool normal_timeout = config_.perform_timeouts && (dop.deadline < cluster_wide_time_microseconds_);
+
+      if (should_drop || normal_timeout) {
+        stats_.timed_out_requests++;
+        dop.promise.TimeOut();
```
antaljanosbenjamin, line 341: So this means if a message is dropped, that can cause a promise to timeout before the timeout of the request, right? I find this tricky, because in real world dropping would result in some network error like "disconnected" or "connection is broken" or something like that.

src/io/simulator/simulator_handle.hpp
```diff
@@ -0,0 +1,453 @@
...
+        std::cout << "timing out request" << std::endl;
+        DeadlineAndOpaquePromise dop = std::move(promises_.at(promise_key));
+        promises_.erase(promise_key);
+
+        stats_.timed_out_requests++;
+
+        dop.promise.TimeOut();
+      }
+    }
+  }
```
antaljanosbenjamin, line 267: This should be a private function in my opinion.

src/io/simulator/simulator_handle.hpp
```diff
@@ -0,0 +1,453 @@
...
+};
+
+struct DeadlineAndOpaquePromise {
+  uint64_t deadline;
+  OpaquePromise promise;
+};
+
+class SimulatorHandle {
```
antaljanosbenjamin, line 202: Could you please move the non-template member function into a source file? They are quite big.

src/io/simulator/simulator_handle.hpp
```diff
@@ -0,0 +1,453 @@
...
+  void ShutDown() {
+    std::unique_lock<std::mutex> lock(mu_);
+    should_shut_down_ = true;
+    cv_.notify_all();
+  }
+
+  bool ShouldShutDown() {
+    std::unique_lock<std::mutex> lock(mu_);
+    return should_shut_down_;
+  }
```
antaljanosbenjamin, line 366: You can mark `mu_` mutable, so this function can be turned into a `const` member function.

src/io/simulator/simulator_handle.hpp
```diff
@@ -0,0 +1,453 @@
...
+
+  template <Message... Ms>
+  requires(sizeof...(Ms) > 0) RequestResult<Ms...> Receive(const Address &receiver, uint64_t timeout_microseconds) {
+    std::unique_lock<std::mutex> lock(mu_);
+
+    uint64_t deadline = cluster_wide_time_microseconds_ + timeout_microseconds;
+
+    while (!should_shut_down_ && (cluster_wide_time_microseconds_ < deadline)) {
+      if (can_receive_.contains(receiver)) {
+        std::vector<OpaqueMessage> &can_rx = can_receive_.at(receiver);
```
antaljanosbenjamin, line 400: I wonder whether a `std::queue` wouldn't be a better fit here. Currently this is LIFO, but maybe a FIFO queue should be better.

src/io/simulator/simulator_handle.hpp
```diff
@@ -0,0 +1,453 @@
...
+
+    stats_.total_messages++;
+
+    cv_.notify_all();
+  }
+
+  uint64_t Now() {
+    std::unique_lock<std::mutex> lock(mu_);
+    return cluster_wide_time_microseconds_;
+  }
```
antaljanosbenjamin, line 440: This can be also `const`.

src/io/simulator/simulator_handle.hpp
```diff
@@ -0,0 +1,453 @@
...
+  template <class D = std::poisson_distribution<>, class Return = uint64_t>
+  Return Rand(D distrib) {
+    std::unique_lock<std::mutex> lock(mu_);
+    return distrib(rng_);
+  }
+
+  SimulatorStats Stats() {
+    std::unique_lock<std::mutex> lock(mu_);
+    return stats_;
+  }
```
antaljanosbenjamin, line 451: This can be also `const`.

src/io/simulator/simulator_transport.hpp
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  template <Message M>
+  void Send(Address address, uint64_t request_id, M message) {
+    return simulator_handle_->template Send<M>(address, address_, request_id, message);
+  }
+
+  uint64_t Now() { return simulator_handle_->Now(); }
+
+  bool ShouldShutDown() { return simulator_handle_->ShouldShutDown(); }
```
antaljanosbenjamin, line 55: Make them const also.

antaljanosbenjamin (changes requested):
I would add some unit tests that run on the CI to make sure the simulator and its features work.

src/io/simulator/simulator.hpp
```diff
@@ -0,0 +1,44 @@
...
+#include <memory>
+
+#include "io/address.hpp"
+#include "io/simulator/simulator_config.hpp"
+#include "io/simulator/simulator_handle.hpp"
+#include "io/simulator/simulator_transport.hpp"
+
+namespace memgraph::io::simulator {
+class Simulator {
+  std::mt19937 rng_{};
```
jbajic, line 23: Add `#include<random>`, and remove redunadn `{}` since the member variable will be initialized by default. https://en.cppreference.com/w/cpp/language/default_initialization.
Also true for `seed_distrib` below

src/io/simulator/simulator.hpp
```diff
@@ -0,0 +1,44 @@
...
+  std::shared_ptr<SimulatorHandle> simulator_handle_;
+
+ public:
+  explicit Simulator(SimulatorConfig config)
+      : rng_(std::mt19937{config.rng_seed}), simulator_handle_{std::make_shared<SimulatorHandle>(config)} {}
+
+  void ShutDown() { simulator_handle_->ShutDown(); }
+
+  Io<SimulatorTransport> Register(Address address) {
```
jbajic, line 32: Why pass by value?

src/io/simulator/simulator_config.hpp
```diff
@@ -0,0 +1,23 @@
...
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#pragma once
+
+namespace memgraph::io::simulator {
+struct SimulatorConfig {
+  int drop_percent = 0;
```
jbajic, line 16: This can be `uint8_t`, since it's never gonna go below 0 and above 100, right?

src/io/simulator/simulator_handle.hpp
```diff
@@ -0,0 +1,453 @@
...
+    }
+  }
+
+  void TimeoutPromisesPastDeadline() {
+    uint64_t now = cluster_wide_time_microseconds_;
+
+    for (auto &[promise_key, dop] : promises_) {
+      // TODO(tyler) queue this up and drop it after its deadline
+      if (dop.deadline < now) {
+        std::cout << "timing out request" << std::endl;
```
jbajic, line 258: Use `spdlog` for logging

src/io/simulator/simulator_transport.hpp
```diff
@@ -0,0 +1,62 @@
...
+  std::mt19937 rng_{};
+
+ public:
+  SimulatorTransport(std::shared_ptr<SimulatorHandle> simulator_handle, Address address, uint64_t seed)
+      : simulator_handle_(simulator_handle), address_(address), rng_(std::mt19937{seed}) {}
+
+  template <Message Request, Message Response>
+  ResponseFuture<Response> Request(Address address, uint64_t request_id, Request request,
+                                   uint64_t timeout_microseconds) {
+    std::function<bool()> maybe_tick_simulator = [=] { return simulator_handle_->MaybeTickSimulator(); };
```
jbajic, line 33:
```suggestion
    std::function<bool()> maybe_tick_simulator = [this] { return simulator_handle_->MaybeTickSimulator(); };
```

tests/simulation/future.cpp
```diff
@@ -0,0 +1,56 @@
...
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#include <string>
+#include <thread>
+
+#include "io/future.hpp"
+#include "utils/logging.hpp"
+
+using namespace memgraph::io;
```
jbajic, line 18: Not sure how much testing you meant to introduce but I would like to see a small benchmark test for `future` implementation, unless you think that this implementation is trivial enough that it does not require that

src/io/future.hpp
spacejam: I think anything related to testing can avoid noisiness around branch hints

src/io/future.hpp
```diff
@@ -0,0 +1,257 @@
...
+  Shared &operator=(const Shared &) = delete;
+  ~Shared() = default;
+
+  T Wait() {
+    std::unique_lock<std::mutex> lock(mu_);
+    waiting_ = true;
+
+    while (!item_) {
+      bool simulator_progressed = false;
+      if (simulator_notifier_) {
```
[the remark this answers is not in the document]
spacejam: fyi this broke formatting and removed the last brace

src/io/future.hpp
```diff
@@ -0,0 +1,257 @@
...
+
+#include "io/errors.hpp"
+
+namespace memgraph::io {
+
+// Shared is in an anonymous namespace, and the only way to
+// construct a Promise or Future is to pass a Shared in. This
+// ensures that Promises and Futures can only be constructed
+// in this translation unit.
+namespace {
```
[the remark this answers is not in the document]
spacejam: the value is preventing anyone from constructing a Future or Promise without going through our correctness-critical constructor function

src/io/future.hpp
```diff
@@ -0,0 +1,257 @@
...
+  }
+
+  /// Block on the corresponding promise to be filled,
+  /// returning the inner item when ready.
+  T Wait() {
+    MG_ASSERT(!consumed_or_moved_, "Future should only be consumed with Wait once!");
+    T ret = shared_->Wait();
+    consumed_or_moved_ = true;
+    return ret;
+  }
```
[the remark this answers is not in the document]
spacejam: in general I like the idea of enforcing move-only semantics around Future and Promise as much as possible. Is there a reason not to simply make TryGet also an rvalue reference function?

src/io/future.hpp
```diff
@@ -0,0 +1,257 @@
...
+
+  /// Marks this Future as canceled.
+  void Cancel() {
+    MG_ASSERT(!consumed_or_moved_, "Future::Cancel called on a future that was already moved or consumed!");
+    consumed_or_moved_ = true;
+  }
```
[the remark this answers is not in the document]
spacejam: So that we can gracefully discard work when it becomes unnecessary, like if there are 8 promises related to various requests, and several of them don't make it back, so we can just cancel the rest of them and return an error to the client. Over time, cancellation may have additional resource clean-ups associated with it, depending on concrete transport abstractions etc...

src/io/future.hpp
```diff
@@ -0,0 +1,257 @@
...
+
+template <typename T>
+std::pair<Future<T>, Promise<T>> FuturePromisePairWithNotifier(std::function<bool()> simulator_notifier) {
+  std::shared_ptr<Shared<T>> shared = std::make_shared<Shared<T>>(simulator_notifier);
+
+  Future<T> future = Future<T>(shared);
+  Promise<T> promise = Promise<T>(shared);
+
+  return std::make_pair(std::move(future), std::move(promise));
+}
```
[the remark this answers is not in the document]
spacejam: this is accomplished with the anonymous namespace. the previous use of friends seemed to make several people uncomfortable.

src/io/simulator/simulator.hpp
spacejam: because it's just a few bytes on the stack, and can be correctly memcpy'd without loss of correctness. My assumption is that anything that does not have associated allocations or other resources, and memcpy-friendly, is better to pass by value and let the compiler narrow down the specifics that it should push into the stackframe.

src/io/transport.hpp
```diff
@@ -0,0 +1,130 @@
...
+  template <Message... Ms>
+  RequestResult<Ms...> ReceiveWithTimeout(uint64_t timeout_microseconds) {
+    return implementation_.template Receive<Ms...>(timeout_microseconds);
+  }
+
+  /// Wait the default number of microseconds for a request of one of the
+  /// provided types to arrive.
+  template <Message... Ms>
+  requires(sizeof...(Ms) > 0) RequestResult<Ms...> Receive() {
+    uint64_t timeout_microseconds = default_timeout_microseconds_;
```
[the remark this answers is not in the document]
spacejam: I find the explicitness nice. it's trivially eliminable by the compiler, so it's basically just free explicitness IMO

spacejam:
it's surprising that the github diff view is showing a couple unrelated commits as part of this branch after merging in `project-pineapples`, because `git branch --contains  eb0b3141d51e679cecf6d85588696b039677b9bf ` shows:
```
T0879-MG-transport-prototype
  T0912-MG-in-memory-shard-map
  T0941-MG-implement-basic-raft-version
  master
  project-pineapples
  ```

But I think it's harmless

File: src/io/CMakeLists.txt [hunk 32]
```diff
@@ -2,7 +2,11 @@ set(io_src_files
     network/addrinfo.cpp
     network/endpoint.cpp
     network/socket.cpp
-    network/utils.cpp)
+    network/utils.cpp
+    future.hpp
+    address.hpp
+    errors.hpp
+    transport.hpp)
```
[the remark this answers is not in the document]
antaljanosbenjamin: I had a typo affecting understanding: Instead of "Source files are currently not listed" I wanted to write "Header files are currently not listed.

src/io/future.hpp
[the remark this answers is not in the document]
antaljanosbenjamin: Yeah, I messed that up, sorry.

src/io/future.hpp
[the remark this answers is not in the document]
antaljanosbenjamin: My only reasoning for this `TryGet` _might_ consume the object, but not certainly. Because of that we might and up finding with the compiler, because it cannot figure out in compilation time whether the `Future` will be moved or not, thus popping up a lot of "use after move" warnings.

src/io/time.hpp
```diff
@@ -0,0 +1,21 @@
...
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#pragma once
+
+#include <chrono>
+
+namespace memgraph::io {
+
+using Duration = std::chrono::duration<int64_t, std::ratio<1, 1000000>>;
```
antaljanosbenjamin, line 18:
```suggestion
using Duration = std::chrono::microseconds;
```

antaljanosbenjamin, line 18:
```
using Duration = std::chrono::microseconds;
```

src/io/transport.hpp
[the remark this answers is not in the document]
antaljanosbenjamin: I think this case it is fine, however in general I think we shouldn't assume compiler can elide these thing, especially if they are not const or not a primitive type. C++ compilers are not that smart as rustc.

src/io/simulator/simulator_handle.hpp
```diff
@@ -0,0 +1,453 @@
...
+
+  /// High level "user-facing" conversion function that lets
+  /// people interested in conversion only supply a single
+  /// parameter pack for the types that they want to compare
+  /// with the any and potentially include in the returned
+  /// variant.
+  template <Message... Ms>
+  requires(sizeof...(Ms) > 0) std::optional<std::variant<Ms...>> VariantFromAny(std::any &&a) {
+    return Unpack<std::variant<Ms...>, Ms...>(std::move(a));
+  }
```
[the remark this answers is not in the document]
antaljanosbenjamin: I think you kept this as non-static member functions, is there any reason for this?

src/io/simulator/simulator_handle.hpp
```diff
@@ -0,0 +1,453 @@
...
+
+  template <Message... Ms>
+  requires(sizeof...(Ms) > 0) std::optional<RequestEnvelope<Ms...>> Take() {
```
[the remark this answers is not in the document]
antaljanosbenjamin: I think also missed this comment.

src/io/simulator/simulator_handle.hpp
antaljanosbenjamin: You also missed this in my opinion.

src/io/simulator/simulator_handle.hpp
antaljanosbenjamin: I think they remain non-const.

src/io/simulator/simulator_handle.hpp
antaljanosbenjamin: What is your opinion here?

src/io/simulator/simulator.hpp
antaljanosbenjamin: We have a thumb of rule is to pass objects by reference if they are bigger than `2 * sizeof(pointer)`. If you specify to pass an object by value, then the compiler will always copy, it cannot optimize out the copying. However this case is fine, as
 the [size of Address](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:10,endLineNumber:26,positionColumn:10,positionLineNumber:26,selectionStartColumn:10,selectionStartLineNumber:26,startColumn:10,startLineNumber:26),source:'%23include+%3Ciostream%3E%0A%23include+%3Cmemory%3E%0A%23include+%3Cstring%3E%0A%23include+%3Cvariant%3E%0A%0A%0A%23include+%3Cboost/asio/ip/tcp.hpp%3E%0A%23include+%3Cboost/uuid/uuid.hpp%3E%0A%23include+%3Cboost/uuid/uuid_io.hpp%3E%0A%0Astruct+Address+%7B%0A++//+It!'s+important+for+all+participants+to+have+a%0A++//+unique+identifier+-+IP+and+port+alone+are+not%0A++//+enough,+and+may+change+over+the+lifecycle+of%0A++//+the+nodes.+Particularly+storage+nodes+may+change%0A++//+their+IP+addresses+over+time,+and+the+system%0A++//+should+gracefully+update+its+information%0A++//+about+them.%0A++boost::uuids::uuid+unique_id%3B%0A++boost::asio::ip::address+last_known_ip%3B%0A++uint16_t+last_known_port%3B%0A%7D%3B%0A%0Aint+main()+%7B%0Astd::cout+%3C%3C+sizeof(Address)+%3C%3C+std::endl%3B%0Areturn+0%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:50,l:'4',m:50,n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:compiler,i:(compiler:clang1301,filters:(b:'0',binary:'0',commentOnly:'0',demangle:'1',directives:'0',execute:'0',intel:'1',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!((name:boost,ver:'179')),options:'--std%3Dc%2B%2B20+',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1,tree:'1'),l:'5',n:'0',o:'x86-64+clang+13.0.1+(C%2B%2B,+Editor+%231,+Compiler+%231)',t:'0')),header:(),l:'4',m:50,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+gcc+12.1',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+clang+13.0.1+(Compiler+%231)',t:'0')),header:(),k:50,l:'4',m:50,n:'0',o:'',s:0,t:'0')),k:50,l:'3',n:'0',o:'',t:'0')),l:'2',m:100,n:'0',o:'',t:'0')),version:4) is only 48 bytes.
 
And I lied about not being able to optimize out: for function that in can inline (mostly for functions defined in the same translation unit) the compiler can actually super smart, but not in general.

src/io/simulator/simulator_handle.hpp
antaljanosbenjamin: You missed this also.

src/io/simulator/simulator_handle.hpp
antaljanosbenjamin: .

antaljanosbenjamin (changes requested):
I resolved all of the conversations that are either addressed, and unresolved all of the conversations that are not addressed. Please check them.

I tried to provide a single review for them, but somehow it didn't work out, not sure what did I do wrong.

tests/simulation/trial_query_storage/query_storage_test.cpp
```diff
@@ -0,0 +1,86 @@
...
+  Io<SimulatorTransport> srv_io = simulator.Register(srv_addr);
+
+  auto srv_thread = std::jthread(run_server, std::move(srv_io));
+  simulator.IncrementServerCountAndWaitForQuiescentState(srv_addr);
+
+  auto req = ScanVerticesRequest{2, std::nullopt};
+
+  auto res_f = cli_io.Request<ScanVerticesRequest, VerticesResponse>(srv_addr, req);
+  auto res_rez = std::move(res_f).Wait();
+  // MG_ASSERT(res_rez.HasError());
```
42jeremy, line 83: Either remove the line or the comment
(feel free to resolve conversation yourself)

src/io/address.hpp
```diff
@@ -0,0 +1,61 @@
...
+    return ret;
+  }
+
+  bool operator==(const Address &other) const {
+    return ((unique_id == other.unique_id) && last_known_ip == other.last_known_ip) &&
+           (last_known_port == other.last_known_port);
+  }
+
+  /// unique_id is most dominant for ordering, then last_known_ip, then last_known_port
+  bool operator<(const Address &other) const {
```
42jeremy, line 46: For this types of operator, I prefer the following from maintenance point of view:
```
bool operator<(const Address &other) const {
  if (unique_id != other.unique_id) {
    return unique_id < other.unique_id;
  } 

  if (last_known_ip != other.last_known_ip) {
      return last_known_port < other.last_known_port;
  }

  MG_ASSERT(last_known_ip < other.last_known_ip);
  return last_known_ip < other.last_known_ip;
}
```
Feel free to resolve the conversation if you disagree, this is personal opinion

src/io/address.hpp
```diff
@@ -0,0 +1,61 @@
...
+  uint16_t last_known_port;
+
+  static Address TestAddress(uint16_t port) {
+    Address ret;
+    ret.last_known_port = port;
+    return ret;
+  }
+
+  bool operator==(const Address &other) const {
+    return ((unique_id == other.unique_id) && last_known_ip == other.last_known_ip) &&
```
42jeremy, line 41: Minor remark: parenthesis are confusing.
```
return (unique_id == other.unique_id) && (last_known_ip == other.last_known_ip) &&
           (last_known_port == other.last_known_port);
```
or 
```
return unique_id == other.unique_id && last_known_ip == other.last_known_ip &&
           last_known_port == other.last_known_port;
```

tests/simulation/trial_query_storage/messages.hpp
```diff
@@ -0,0 +1,36 @@
...
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#pragma once
+
+#include <optional>
+#include <string>
+#include <vector>
+
+// header
```
42jeremy, line 18:
```suggestion

```

42jeremy:
Questions about the tests under /simulation
-Why aren't they ([basic_request.cpp](https://github.com/memgraph/memgraph/pull/466/files#diff-41baac3e78a6e323c5bfc035df86f8ec7ff8ce62ecceb3593f0fd3a9e44904e9) and [query_storage_test.cpp](https://github.com/memgraph/memgraph/pull/466/files#diff-7e04e5a1373642c251926dd0294a8a7c97238f387c61939082877496ad97ce3a)) written as unit test? 
-How will the CI knows whether this is a success or failure? Will a sudden crash (MG_ASSERT for instance) implies a failure?

src/io/address.hpp
spacejam: thanks for catching this! I aggree, and will go with option 1

src/io/address.hpp
spacejam: I agree, and I've applied the same change to simulator_handle.hpp where PromiseKey had a similar nesting issue

src/io/future.hpp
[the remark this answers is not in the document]
spacejam: I've changed this to use a details namespace as suggested

src/io/simulator/simulator_handle.hpp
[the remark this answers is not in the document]
spacejam: applied, thanks

src/io/simulator/simulator_handle.cpp
```diff
@@ -0,0 +1,158 @@
...
+    size_t swap_index = swap_distrib(rng_);
+    std::swap(in_flight_[swap_index], in_flight_.back());
+  }
+
+  auto [to_address, opaque_message] = std::move(in_flight_.back());
+  in_flight_.pop_back();
+
+  std::uniform_int_distribution<int> drop_distrib(0, 99);
+  int drop_threshold = drop_distrib(rng_);
+  bool should_drop = drop_threshold < config_.drop_percent;
```
jbajic, line 114:
```suggestion
  const int drop_threshold = drop_distrib(rng_);
  const bool should_drop = drop_threshold < config_.drop_percent;
```

src/io/simulator/simulator_handle.cpp
```diff
@@ -0,0 +1,158 @@
...
+      stats_.total_responses++;
+      dop.promise.Fill(std::move(opaque_message));
+    }
+  } else if (should_drop) {
+    // don't add it anywhere, let it drop
+  } else {
+    // add to can_receive_ if not
+    const auto &[om_vec, inserted] = can_receive_.try_emplace(to_address, std::vector<OpaqueMessage>());
+    om_vec->second.emplace_back(std::move(opaque_message));
+  }
```
jbajic, line 144: Did you intentionally make the `should_drop` branch to place a comment?
```suggestion
  } else if (!should_drop) {
    // add to can_receive_ if not
    const auto &[om_vec, inserted] = can_receive_.try_emplace(to_address, std::vector<OpaqueMessage>());
    om_vec->second.emplace_back(std::move(opaque_message));
  }
```

src/io/simulator/simulator_handle.hpp
spacejam: applied, thank you

src/io/simulator/simulator_handle.hpp
spacejam: thanks, extracted and pushed

src/io/simulator/simulator_handle.hpp
spacejam: applied, thank you

src/io/simulator/simulator_handle.hpp
spacejam: I think the more chaos introduced in terms of timing, the better. Any code that depends on a certain duration of time for correctness has a bug.  To this end, the simulator will eventually move clocks backwards as well, getting buggy math related to timeout handling to jump out. This is "real" because NTP can do the same thing if the delta is above a few minutes. 

But I think in terms of realism, there should only be one place that times out requests, and that should be in the TimeoutPromisesPastDeadline method. I'll make this specific case a no-op and just keep the associated promise in the promises_ map until timeouts get dealt with in the single path.

src/io/address.hpp
```diff
@@ -0,0 +1,63 @@
...
+  boost::asio::ip::address last_known_ip;
+  uint16_t last_known_port;
+
+  static Address TestAddress(uint16_t port) {
+    Address ret;
+    ret.last_known_port = port;
+    return ret;
+  }
+
+  bool operator==(const Address &other) const {
```
42jeremy, line 40: I think you can use a default == operator here.
E.G.: bool operator==(const Address &) const = default;
(https://en.cppreference.com/w/cpp/language/default_comparisons)

42jeremy (commented):
Nothing to add on top of what the others have already said.

src/io/future.hpp
```diff
@@ -0,0 +1,262 @@
...
+    return ret;
+  }
+
+  T Wait() {
+    std::unique_lock<std::mutex> lock(mu_);
+    waiting_ = true;
+
+    while (!item_) {
+      bool simulator_progressed = false;
+      if (simulator_notifier_) [[unlikely]] {
```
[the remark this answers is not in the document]
spacejam: it checks whether there is a simulator notifier specified in the Future implementation, which will basically just communicate to the simulator during testing that this thread is blocked on future message delivery / timeout. If there is no notification function specified, this will be a null function pointer, and cast to false.

src/io/simulator/simulator_handle.cpp
spacejam: this was applied on the branch

src/io/simulator/simulator_handle.cpp
spacejam: I think the original is more clear, and has zero runtime performance implications

42jeremy (approved):

jbajic (approved):

.github/workflows/diff.yaml
```diff
@@ -262,6 +262,15 @@ jobs:
           cd build
           ctest -R memgraph__unit --output-on-failure -j$THREADS
 
+      - name: Run simulation tests
```
antaljanosbenjamin, line 265: Could you please also add the simulation tests to the code analysis build? It runs the tests using the sanitizers, thus I think it might point out memory issues using the simulation tests.

src/io/simulator/message_conversion.hpp
```diff
@@ -0,0 +1,172 @@
...
+class OpaquePromise {
+  void *ptr_;
+  std::unique_ptr<OpaquePromiseTraitBase> trait_;
+
+ public:
+  OpaquePromise(OpaquePromise &&old) noexcept : ptr_(old.ptr_), trait_(std::move(old.trait_)) { old.ptr_ = nullptr; }
+
+  OpaquePromise &operator=(OpaquePromise &&old) noexcept {
+    MG_ASSERT(this != &old);
+    ptr_ = old.ptr_;
```
antaljanosbenjamin, line 125: Won't this leak the object pointed by `ptr_`?

src/io/simulator/simulator_handle.hpp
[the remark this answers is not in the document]
antaljanosbenjamin: This is still not addressed but I don't want to keep the PR back because of this.

src/io/simulator/simulator_handle.cpp
```diff
@@ -0,0 +1,156 @@
...
+    cv_.wait(lock);
+  }
+}
+
+size_t SimulatorHandle::BlockedServers() {
+  size_t blocked_servers = blocked_on_receive_;
+
+  for (auto &[promise_key, opaque_promise] : promises_) {
+    if (opaque_promise.promise.IsAwaited()) {
+      if (server_addresses_.contains(promise_key.requester_address)) {
```
kostasrim, line 58:
```suggestion
    if (opaque_promise.promise.IsAwaited() && server_addresses_.contains(promise_key.requester_address)) {
```

spacejam: fyi this broke compilation

kostasrim (approved):

antaljanosbenjamin (approved):

PR #472, opened 2022-07-22
T0982 mg implement edge type filtering

niko4299:


gitbuda (changes requested):
The code is changed on the `storage` side, which is unnecessary for this feature. Please revert/move all changes under `src/storage` somewhere else, probably under `src/query` 😄

PR #475, opened 2022-07-26
[master < T0990-MG] Fix handling of the `ROUTE` Bolt message

antaljanosbenjamin:
[master < Task] PR
- [ ] Check, and update documentation if necessary - [Docs PR](https://github.com/memgraph/docs/pull/503)

> The fields of `ROUTE` message were not read from the input buffer, thus the
> imput buffer got corrupted. Sending a new message to the server would result
> reading the remaining fields from the buffer, which means reading some values
> instead of message signature. Because of this unmet expectation, Memgraph closed
> the connection. With this fix, the fields of the `ROUTE` message are properly
> read and ignored.


src/communication/bolt/client.cpp
```diff
@@ -0,0 +1,247 @@
+// Copyright 2022 Memgraph Ltd.
```
antaljanosbenjamin, line 1: I mostly copied this file over, but after that changed a few things, check the commit history.

src/communication/bolt/v1/states/handlers.hpp
```diff
@@ -403,12 +404,33 @@ State HandleGoodbye() {
...
+
+  if (!session.decoder_.ReadValue(&bookmarks, Value::Type::List)) {
+    spdlog::trace("Couldn't read bookmarks field!");
+    return State::Close;
+  }
+
+  if (!session.decoder_.ReadValue(&db)) {
+    spdlog::trace("Couldn't read db field!");
+    return State::Close;
+  }
```
antaljanosbenjamin, line 430: This is the gist of the bugfix.

src/communication/bolt/v1/states/handlers.hpp
```diff
@@ -403,12 +404,33 @@ State HandleGoodbye() {
...
   session.encoder_buffer_.Clear();
   bool fail_sent =
-      session.encoder_.MessageFailure({{"code", 66}, {"message", "Route message not supported in Memgraph!"}});
+      session.encoder_.MessageFailure({{"code", "66"}, {"message", "Route message is not supported in Memgraph!"}});
```
antaljanosbenjamin, line 433: The code is a string field in the `FAILURE` message.

tests/e2e/magic_functions/functions/c_read.cpp
```diff
@@ -22,13 +22,13 @@ static void ReturnFunctionArgument(struct mgp_list *args, mgp_func_context *ctx,
   mgp_value *value{nullptr};
   auto err_code = mgp_list_at(args, 0, &value);
   if (err_code != mgp_error::MGP_ERROR_NO_ERROR) {
-    mgp_func_result_set_error_msg(result, "Failed to fetch list!", memory);
+    static_cast<void>(mgp_func_result_set_error_msg(result, "Failed to fetch list!", memory));
```
antaljanosbenjamin, line 25: These are some unrelated compiler warning fixes.

42jeremy:
Do we have any end to end tests using Bolt? If not, is it easy to introduce?

src/communication/bolt/v1/codes.hpp
```diff
@@ -16,7 +16,8 @@
 namespace memgraph::communication::bolt {
 
 inline constexpr uint8_t kPreamble[4] = {0x60, 0x60, 0xB0, 0x17};
-inline constexpr uint8_t kProtocol[4] = {0x00, 0x00, 0x00, 0x01};
+inline constexpr uint8_t kProtocol[4] = {0x00, 0x00, 0x03, 0x04};
```
jbajic, line 19: I think it is confusing to see `kProtocol` being defined here since it has nothing to do with the server but only the client. Maybe it makes sense to move this to the client code, plus it is only used in the client code

src/communication/bolt/v1/encoder/client_encoder.hpp
```diff
@@ -38,22 +44,20 @@ class ClientEncoder : private BaseEncoder<Buffer> {
...
    *          when flushing, false otherwise
    */
-  bool MessageInit(const std::string client_name, const std::map<std::string, Value> &auth_token) {
-    WriteRAW(utils::UnderlyingCast(Marker::TinyStruct2));
+  bool MessageInit(const std::map<std::string, Value> &extra) {
+    WriteRAW(utils::UnderlyingCast(Marker::TinyStruct1));
     WriteRAW(utils::UnderlyingCast(Signature::Init));
-    WriteString(client_name);
-    WriteMap(auth_token);
+    WriteMap(extra);
```
jbajic, line 60: If you are changing how we handle messages to fit v3 bolt version, does this mean we will be incompatible with v1? Have you tested these changes so that clients using v1 are not affected?

src/communication/bolt/v1/states/handlers.hpp
```diff
@@ -403,12 +404,33 @@ State HandleGoodbye() {
 }
 
 template <typename TSession>
-State HandleRoute(TSession &session) {
-  // Route message is not implemented since it is neo4j specific, therefore we
-  // will receive it an inform user that there is no implementation.
+State HandleRoute(TSession &session, const Marker marker) {
+  // Route message is not implemented since it is Neo4j specific, therefore we will receive it an inform user that there
```
jbajic, line 408:
```suggestion
  // Route message is not implemented since it is Neo4j specific, therefore we will receive it and inform user that there
```

src/communication/bolt/v1/states/handlers.hpp
jbajic, line 430:
```suggestion
  Value routing;
  if (!session.decoder_.ReadValue(&routing, Value::Type::Map)) {
    spdlog::trace("Couldn't read routing field!");
    return State::Close;
  }
  
  Value bookmarks;
  if (!session.decoder_.ReadValue(&bookmarks, Value::Type::List)) {
    spdlog::trace("Couldn't read bookmarks field!");
    return State::Close;
  }

  Value db;
  if (!session.decoder_.ReadValue(&db)) {
    spdlog::trace("Couldn't read db field!");
    return State::Close;
  }
```

tests/integration/transactions/tester.cpp
```diff
@@ -100,20 +100,20 @@ TEST_F(BoltClient, DoubleRollbackWithoutTransaction) {
 TEST_F(BoltClient, DoubleBegin) {
   EXPECT_TRUE(Execute("begin"));
   EXPECT_THROW(Execute("begin", kNestedTransactions), ClientQueryException);
-  EXPECT_TRUE(TransactionActive());
+  EXPECT_FALSE(TransactionActive());
```
jbajic, line 103: Why did this change?

antaljanosbenjamin: Because from Bolt v3 `ACK_FAILURE` is deprecated, so after receiving a `FAILURE` message, we have to send `RESET` to the server, which will reset the server state, meaning aborting the ongoing transactions of that session.

src/communication/bolt/v1/encoder/client_encoder.hpp
antaljanosbenjamin: This `ClientEncoder` is only meant to support one specific Bolt version. Previously it supported only Bolt v1 and I upgraded this to Bolt v4.3. So you are right, this client will be incompatible with Bolt v1, but Memgraph will work well with Bolt v1, as these changes only affect the C++ client and not the Bolt server we use in Memgraph.

src/communication/bolt/client.cpp
```diff
@@ -0,0 +1,262 @@
...
+    spdlog::error("Couldn't send preamble!");
+    throw ServerCommunicationException();
+  }
+
+  if (!client_.Write(kBoltV43Version, sizeof(kBoltV43Version), true)) {
+    spdlog::error("Couldn't send protocol version!");
+    throw ServerCommunicationException();
+  }
+
+  for (int i = 0; i < 3; ++i) {
```
42jeremy, line 42: I am a little bit confused about this for-loop here. What's the goal behind it?
I guess we are trying several attempts, is it correct? Isn't there another way?

jbajic (approved):
Amazing job on the fix, and the addition of new tests!

42jeremy (approved):
Just one question, feel free to resolve once we've discussed it.

src/communication/bolt/client.cpp
antaljanosbenjamin: No. As part of the Bolt protocol negotiation, the client has to send `128 bytes` which includes at least one, at most four Bolt versions, and the rest remaining bytes are filled with zeros. So here we send one Bolt version and zeros for the rest of the 3 possible versions. Read more about this on the [Bolt documentation](https://neo4j.com/docs/bolt/current/bolt/handshake/#_version_negotiation).

42jeremy: ok, got it. Thanks!

PR #478, opened 2022-07-28
Merge old Label Based Auth Epic branch into new one because of commits with bad checks on the old epic branch

BorisTasevski:


BorisTasevski:
@antaljanosbenjamin Can You tell me why is code analysis failing here?

antaljanosbenjamin:
As far as I see there are two issues:
```
/opt/actions-runner/_work/memgraph/memgraph/src/auth/models.hpp:94:3: warning: constructors that are callable with a single argument must be marked explicit to avoid unintentional implicit conversions [hicpp-explicit-conversions]
Use -header-filter=.* to display errors from all non-system headers. Use -system-headers to display errors from system headers as well.

  FineGrainedAccessPermissions(const std::unordered_set<std::string> &grants = {},
  ^
  explicit
```

The ctor of `FineGrainedAccessPermissions` can be called with a single argument. In order to avoid accidental conversion, the constructor should be marked explicit, so an `std::unordered_set<std::string>` cannot be converted to a `FineGrainedAccessPermissions` object without explicitly writing out `FineGrainedAccessPermissions`.

The other thing is a header issue:
```
/opt/actions-runner/_work/memgraph/memgraph/src/query/frontend/ast/cypher_main_visitor.hpp:18:10: error: 'antlr4-runtime.h' file not found [clang-diagnostic-error]
Error while processing /opt/actions-runner/_work/memgraph/memgraph/src/query/context.hpp.
Suppressed 167109 warnings (163440 in non-user code, 3563 due to line filter, 106 NOLINT).
Use -header-filter=.* to display errors from all non-system headers. Use -system-headers to display errors from system headers as well.
Found compiler error(s).

#include <antlr4-runtime.h>
```

So basically clang-tidy cannot find `antlr4-runtime.h`. The question is good why. I will have a quick look at it today, maybe @gitbuda faced this issue regarding to Antlr 4.10 upgrade.

BorisTasevski:
explicit ctor is a warning, and i didn't pay much attention to it, because I've seen that throughout the codebase. But i can "mitigate" that. Even so, it's still a warning

Josipmrden (approved):

PR #481, opened 2022-07-29
[master < MG] Declare `mgp_func_context` outside the callback funciton

antaljanosbenjamin:
[master < Task] PR

> Declare `mgp_func_context` outside the callback function

We missed to declare this struct, so it was implicitly declared in the type definition of `mg_func_cb`, but we introduced this for future improvements, currently it doesn't have any value (apart from making sure we can add new functionalities to functions without breaking the API). I added the declaration just to be 100% correct.

As this didn't mean any error to the users, because the struct is currently unused, I am not sure whether we should add this to the changelog or not. What do you think?

### DIFF
--- a/include/mg_procedure.h
+++ b/include/mg_procedure.h
@@ -1512,6 +1512,10 @@ enum mgp_error mgp_module_add_transformation(struct mgp_module *module, const ch
 ///
 ///@{
 
+/// State of the database that is exposed to magic functions. Currently it is unused, but it enables extending the
+/// functionalities of magic functions in future without breaking the API.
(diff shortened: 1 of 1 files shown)


42jeremy (approved):

gvolfing (approved):

gvolfing:
If this did not cause any problems so far, and no user had any problem with this whatsoever then I don't think it should be added to the changelog. From the users perspective nothing has changed.

antaljanosbenjamin:
I got the same feedback from @42jeremy and @jmatak, so I won't create a changelog for this.

PR #482, opened 2022-08-01
[E112-MG < T0950-MG] Move schema to storage v3 and query v2

jbajic:
[master < Task] PR

__What was done?__
- Schema implementation in query engine was moved from query v1 was moved to query v2
- Schema implementation in the storage engine was moved from storage v2 to storage v3
- Make glue v2 because the test `query_v2_query_plan_edge_cases` requires `result_faker_stream.hpp` which requires glue, also moved `result_faker_stream.hpp` to tests 
- Make newer tests depend on storage v3 and query v2, also uncommented tests on older versions of engines
- Fix clang-tidy errors

__Why was this done?__
Since scheme implementation was started before the newer version of query and storage existed this was necessary to do to align with the rest of the `project-pineapples`

__Commit message:__
- Move schema implementation to query v2 and storage v3
- Make glue v2
- Fix unit tests that use new version of query and storage engine


antaljanosbenjamin:
I haven't checked every line in the PR, but scrolled through the changes of each commit and they make sense.

I searched for `schema` in the codebase and it mostly used in storage-v3 and query-v2 and their tests. However it is also used in auth, but I think that small overlap is okay for now, we shouldn't create an auth-v2 just for this.

antaljanosbenjamin (approved):

kostasrim (approved):

PR #492, opened 2022-08-04
[E129 < T0956] Filtering nodes in ScanAll cursor [Niko]

BorisTasevski:


src/query/plan/operator.cpp
```diff
@@ -405,11 +406,28 @@ class ScanAllCursor : public Cursor {
...
+    if (vertices_it_.value() == vertices_.value().end()) return false;
+#endif
+
     frame[output_symbol_] = *vertices_it_.value();
     ++vertices_it_.value();
     return true;
   }
 
+  void FilterNodes(const FineGrainedAccessChecker *fine_grained_access_checker, DbAccessor *db_accessor) {
+    if (!fine_grained_access_checker) return;
```
Josipmrden, line 420: is it better to keep this if outside the method?

Josipmrden: and then just leave the logic in the method, that actually does the filtering

Josipmrden: Maybe it would be good if this method was returning a vertex. that way we could even test it (if that's even possible). 
if iterator is exhausted, return null, else return a vertex.
The method would be then called GetVertex or something like that?
And yes, we can maybe consider calling here everything as Vertex instead of nodes since here it is mostly about vertices naming in the operator

Josipmrden: it's just my preference that I don't like methods that return void :') you do as you please

Josipmrden (approved):

PR #494, opened 2022-08-08
[E129 < T0996] C-API: Implement using Fine Grained Access Checker in iterator over vertices

BorisTasevski:


src/query/procedure/mg_procedure_impl.hpp
```diff
@@ -675,8 +675,27 @@ struct mgp_vertices_iterator {
   /// @throw anything VerticesIterable may throw
   mgp_vertices_iterator(mgp_graph *graph, memgraph::utils::MemoryResource *memory)
       : memory(memory), graph(graph), vertices(graph->impl->Vertices(graph->view)), current_it(vertices.begin()) {
-    if (current_it != vertices.end()) {
+    memgraph::utils::OnScopeExit clean_up([this] { this->current_v = std::nullopt; });
+    auto *checker = graph->ctx->fine_grained_access_checker;
```
antaljanosbenjamin, line 679:
```suggestion
    const auto *checker = graph->ctx->fine_grained_access_checker;
```

Furthermore I think this is very similar to `mgp_vertices_iterator_next`, it might make sense to extract the logic into a common function.

antaljanosbenjamin: `graph->ctx->fine_grained_access_checker` can be `nullptr`, so a check is necessary. One more reason to extract this logic into a common function.

src/query/procedure/mg_procedure_impl.hpp
```diff
@@ -675,8 +675,27 @@ struct mgp_vertices_iterator {
...
+
+    while (current_it != vertices.end()) {
       current_v.emplace(*current_it, graph, memory);
+
+      if (!current_v.has_value()) {
+        break;
+      }
```
antaljanosbenjamin, line 686: You just emplaced a value into this optional, therefore I think this check is extraneous.

src/query/procedure/mg_procedure_impl.hpp
```diff
@@ -675,8 +675,27 @@ struct mgp_vertices_iterator {
   /// @throw anything VerticesIterable may throw
   mgp_vertices_iterator(mgp_graph *graph, memgraph::utils::MemoryResource *memory)
       : memory(memory), graph(graph), vertices(graph->impl->Vertices(graph->view)), current_it(vertices.begin()) {
-    if (current_it != vertices.end()) {
+    memgraph::utils::OnScopeExit clean_up([this] { this->current_v = std::nullopt; });
+    auto *checker = graph->ctx->fine_grained_access_checker;
+
+    while (current_it != vertices.end()) {
       current_v.emplace(*current_it, graph, memory);
```
antaljanosbenjamin, line 682: I would avoid the construction of `mgp_vertex` until we are sure we can use that vertex. `current_it` should be enough to check whether the use has access to a vertex or not.

src/query/procedure/mg_procedure_impl.hpp
```diff
@@ -675,8 +675,27 @@ struct mgp_vertices_iterator {
   /// @throw anything VerticesIterable may throw
   mgp_vertices_iterator(mgp_graph *graph, memgraph::utils::MemoryResource *memory)
       : memory(memory), graph(graph), vertices(graph->impl->Vertices(graph->view)), current_it(vertices.begin()) {
-    if (current_it != vertices.end()) {
+    memgraph::utils::OnScopeExit clean_up([this] { this->current_v = std::nullopt; });
```
antaljanosbenjamin, line 678: What about `mgp_edge_get_from` and `mgp_edge_get_to`?

antaljanosbenjamin (changes requested):

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -2250,6 +2250,42 @@ mgp_error mgp_graph_delete_edge(struct mgp_graph *graph, mgp_edge *edge) {
   });
 }
 
+#ifdef MG_ENTERPRISE
+namespace {
+void NextPermitted(mgp_vertices_iterator *it) {
```
antaljanosbenjamin, line 2255:
```suggestion
void NextPermitted(mgp_vertices_iterator &it) {
```

On top of that I don't see why should it be protected with the `MG_ENTERPRISE` macro. The regarding field (`it->graph->ctx->fine_grained_access_checker`) is always there, even in community versions. I think this extra function call and `nullptr` check is a reasonable price to pay for getting rid of some `#ifdef`s.

antaljanosbenjamin (changes requested):

jbajic (approved):

Josipmrden (approved):

src/query/procedure/mg_procedure_impl.hpp
BorisTasevski: This is something that we are addressing in one of the next PRs

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -2250,6 +2250,39 @@ mgp_error mgp_graph_delete_edge(struct mgp_graph *graph, mgp_edge *edge) {
...
+
+  if (!checker) {
+    return;
+  }
+
+  while (it.current_it != it.vertices.end()) {
+    auto labels = (*it.current_it).impl_.Labels(it.graph->view);
+    if (!labels.HasValue()) {
+      break;
+    }
```
antaljanosbenjamin, line 2265: What does it mean that we cannot access the labels of the vertex? It means we don't have access to the vertex at that point. So instead of breaking, I think we should just move to the next vertex. I think the iterator takes care of this, so this should be an assert, like `MG_ASSERT(labels.HasValue(), "Unexpected error while getting the labels of a vertex!")`, or something, the error message is just the first that popped into my mind.

```suggestion
    MG_ASSERT(labels.HasValue(), "Unexpected error while getting the labels of a vertex!");
```

antaljanosbenjamin (approved):
Only one comment from me, sorry for not spotting this earlier. If you agree, please apply!

src/query/procedure/mg_procedure_impl.cpp
BorisTasevski: I will fix this in the next PR, if You are OK with that

PR #460, opened 2022-07-20
[E118-MG < T0974-MG] Remove multi-threaded related logic and variables

antaljanosbenjamin:
[master < Epic] PR
- [ ] Check, and update documentation if necessary
- [ ] Update [changelog](https://docs.memgraph.com/memgraph/changelog)
- [ ] Write E2E tests
- [ ] Compare the [benchmarking results](https://bench-graph.memgraph.com/) between the master branch and the Epic branch
- [ ] Provide the full content or a guide for the final git message

[master < Task] PR
- [ ] Check, and update documentation if necessary
- [ ] Update [changelog](https://docs.memgraph.com/memgraph/changelog)
- [x] Provide the full content or a guide for the final git message

> * Remove logic that was necessary for optimal multi-threaded performance, such
>  as accumulating deleted object in local containers and appending them to a
>  global one, handling overlapping locking.
> * Remove background GC thread.
> * Remove mutexes, locks and atomics throughout storage.

PR #497, opened 2022-08-09
[E118-MG < T0960-MG] Integrate schema and keystore

jbajic:
[master < Task] PR

__What is being done?__
- Separate properties and primary properties at vertex creation
- Remove Gid, and introduce Primary Key as unique identifier
- Integrate schema and keystore on vertex creation
**- Breaking query modules, triggers, durability and replication**

__Why is this being done?__
This is an essential part of schema usage in vertex since this is where schemas come into play to differentiate between primary properties and other properties.

__Commit message:__
- Integrate schema and keystore on vertex creation
- Add GC test for storage v3
- Add tests for accessors
- Fix all tests related to this except for query v2
- Fix labels not returning primary label


src/storage/v3/storage.cpp
```diff
@@ -56,36 +63,78 @@ using OOMExceptionEnabler = utils::MemoryTracker::OutOfMemoryExceptionEnabler;
...
+    if (labels_it_ != self_->labelspace_->end()) {
+      self_->vertex_accessor.emplace(labels_it_->second.access());
+      vertex_it_.emplace(self_->vertex_accessor->begin());
+      vertex_it_ = AdvanceToVisibleVertex(*vertex_it_, self_->vertex_accessor->end(), &self_->vertex_,
+                                          self_->transaction_, self_->view_, self_->indices_, self_->constraints_,
+                                          self_->config_, *self_->schema_validator_, *self_->schemas_);
+      return *this;
+    }
+  }
+  end_ = true;
```
jbajic, line 130: Just want to point out this for the reviewers, I was not sure if we should have all accessors opened at once or like this, one at the time. Curerently we open them when we pass to next skiplist, this iterator iterates through map of skiplists, and everytime we jump to new skiplist we need to access it by creating accessor object

src/glue/v2/communication.cpp
```diff
@@ -152,9 +153,10 @@ storage::v3::Result<communication::bolt::Vertex> ToBoltVertex(const storage::v3:
 
 storage::v3::Result<communication::bolt::Edge> ToBoltEdge(const storage::v3::EdgeAccessor &edge,
                                                           const storage::v3::Storage &db, storage::v3::View view) {
-  auto id = communication::bolt::Id::FromUint(edge.Gid().AsUint());
-  auto from = communication::bolt::Id::FromUint(edge.FromVertex().Gid().AsUint());
-  auto to = communication::bolt::Id::FromUint(edge.ToVertex().Gid().AsUint());
+  // TODO(jbajic) Fix bolt communication
+  auto id = communication::bolt::Id::FromUint(0);
+  auto from = communication::bolt::Id::FromUint(1);
+  auto to = communication::bolt::Id::FromUint(2);
```
antaljanosbenjamin, line 159: I would say let's use only `0`, no real meaning for `1` and `2`, then why do use them?

src/query/v2/procedure/mg_procedure_impl.cpp
```diff
@@ -2085,24 +2093,25 @@ mgp_error mgp_graph_is_mutable(mgp_graph *graph, int *result) {
...
+//         auto &ctx = graph->ctx;
+//         ctx->execution_stats[memgraph::query::v2::ExecutionStats::Key::CREATED_NODES] += 1;
+
+//         if (ctx->trigger_context_collector) {
+//            ctx->trigger_context_collector->RegisterCreatedObject(vertex);
+//         }
+//         return NewRawMgpObject<mgp_vertex>(memory, nullptr, graph);
+//       },
+//       result);
+// }
```
antaljanosbenjamin, line 2114: Why did you commented out the whole function here, while just part of the functions in other places? I would be consistent with that is possible.

src/query/v2/plan/operator.cpp
```diff
@@ -2198,23 +2197,26 @@ void SetPropertiesOnRecord(TRecordAccessor *record, const TypedValue &rhs, SetPr
...
-          case storage::v3::Error::DELETED_OBJECT:
-            throw QueryRuntimeException("Trying to set properties on a deleted graph element.");
-          case storage::v3::Error::SERIALIZATION_ERROR:
-            throw TransactionSerializationException();
-          case storage::v3::Error::PROPERTIES_DISABLED:
-            throw QueryRuntimeException("Can't set property because properties on edges are disabled.");
-          case storage::v3::Error::VERTEX_HAS_EDGES:
-          case storage::v3::Error::NONEXISTENT_OBJECT:
-            throw QueryRuntimeException("Unexpected error when setting properties.");
+      if constexpr (std::is_same_v<TRecordAccessor, VertexAccessor>) {
```
antaljanosbenjamin, line 2200: Cannot you use `AccessorWithSetPropertyAndValidate<TRecordAccessor>` here? I think it is better to use that, otherwise you might use the other branch for accessors that have `SetPropertyAndValidate` functions.

src/query/v2/db_accessor.hpp
```diff
@@ -241,8 +246,12 @@ class DbAccessor final {
  public:
   explicit DbAccessor(storage::v3::Storage::Accessor *accessor) : accessor_(accessor) {}
 
-  std::optional<VertexAccessor> FindVertex(storage::v3::Gid gid, storage::v3::View view) {
-    auto maybe_vertex = accessor_->FindVertex(gid, view);
+  // TODO(jbajic) Fix Remove Gid
+  // NOLINTNEXTLINE(readability-convert-member-functions-to-static)
+  std::optional<VertexAccessor> FindVertex(uint64_t /*unused*/) { return std::nullopt; }
```
antaljanosbenjamin, line 251: Why did you keep this?

src/query/v2/trigger_context.cpp
```diff
@@ -265,7 +265,7 @@ void TriggerContext::AdaptForAccessor(DbAccessor *accessor) {
     // adapt created_vertices_
     auto it = created_vertices_.begin();
     for (auto &created_vertex : created_vertices_) {
-      if (auto maybe_vertex = accessor->FindVertex(created_vertex.object.Gid(), storage::v3::View::OLD); maybe_vertex) {
+      if (auto maybe_vertex = accessor->FindVertex(kFakeVertexGid); maybe_vertex) {
         *it = detail::CreatedObject{*maybe_vertex};
```
antaljanosbenjamin, line 269: The triggers will be completely broken, so the only goal here is to minize the time you spend on fixing this.

src/query/v2/trigger_context.hpp
```diff
@@ -223,8 +227,14 @@ class TriggerContextCollector {
   struct HashPairWithAccessor {
     template <detail::ObjectAccessor TAccessor, typename T2>
     size_t operator()(const std::pair<TAccessor, T2> &pair) const {
-      using GidType = decltype(std::declval<TAccessor>().Gid());
-      return utils::HashCombine<GidType, T2>{}(pair.first.Gid(), pair.second);
+      // TODO(jbajic) Fix Remove Gid
+      if constexpr (std::is_same_v<TAccessor, VertexAccessor>) {
+        static uint64_t i{0};
+        return utils::HashCombine<uint64_t, T2>{}(i++, pair.second);
```
antaljanosbenjamin, line 233: Why not just `kFakeVertexGid` instead of `i++`?

src/storage/v3/durability/wal.cpp
```diff
@@ -486,93 +486,93 @@ void EncodeDelta(BaseEncoder *encoder, NameIdMapper *name_id_mapper, Config::Ite
...
-      encoder->WriteUint(delta.vertex_edge.vertex->Gid().AsUint());
-      break;
-    }
-    case Delta::Action::ADD_IN_EDGE:
-    case Delta::Action::REMOVE_IN_EDGE:
-      // These actions are already encoded in the *_OUT_EDGE actions. This
-      // function should never be called for this type of deltas.
-      LOG_FATAL("Invalid delta action!");
-  }
+  // encoder->WriteMarker(Marker::SECTION_DELTA);
```
antaljanosbenjamin, line 489: Perfect handling of compilation issues! No pain, ~no~ really nice gain!

src/storage/v3/vertex_accessor.hpp
```diff
@@ -146,12 +134,27 @@ class VertexAccessor final {
...
+  /// @throw std::bad_alloc
+  Result<PropertyValue> SetProperty(PropertyId property, const PropertyValue &value);
+
   Vertex *vertex_;
   Transaction *transaction_;
   Indices *indices_;
   Constraints *constraints_;
   Config::Items config_;
   VertexValidator vertex_validator_;
+  const Schemas *schemas_;
```
antaljanosbenjamin, line 157: Where is this used? I cannot find any usage of this.

src/storage/v3/schemas.cpp
```diff
@@ -47,6 +49,15 @@ bool Schemas::CreateSchema(const LabelId primary_label, const std::vector<Schema
 
 bool Schemas::DropSchema(const LabelId primary_label) { return schemas_.erase(primary_label); }
 
+bool Schemas::IsPropertyKey(const LabelId primary_label, const PropertyId property_id) const {
+  if (const auto schema_map = schemas_.find(primary_label); schema_map != schemas_.end()) {
+    return std::ranges::find_if(schema_map->second, [property_id](const auto &elem) {
+             return elem.property_id == property_id;
+           }) != schema_map->second.end();
+  }
```
antaljanosbenjamin, line 57:
```suggestion
  if (const auto schema = schemas_.find(primary_label); schema != schemas_.end()) {
    return std::ranges::find_if(schema->second, [property_id](const auto &elem) {
             return elem.property_id == property_id;
           }) != schema->second.end();
  }

File: src/storage/v3/storage.cpp [hunk 11]
```diff
@@ -515,38 +491,46 @@ ResultSchema<VertexAccessor> Storage::Accessor::CreateVertexAndValidate(
...
+  for (const auto &[property_id, property_value] : properties) {
+    if (!storage_->schemas_.IsPropertyKey(primary_label, property_id)) {
+      secondary_properties.emplace_back(property_id, property_value);
+    }
+  }
+
   auto acc = storage_->vertices_.access();
   auto *delta = CreateDeleteObjectDelta(&transaction_);
-  auto [it, inserted] = acc.insert({Vertex{Gid::FromUint(gid), delta, primary_label}});
+  auto [it, inserted] = acc.insert({Vertex{delta, primary_label, primary_properties, labels, secondary_properties}});
```
antaljanosbenjamin, line 512: Why not to move the `primary_properties` and `secondary_properties` into the vertex? They are already copied to the stack, let's let the `Vertex` steal the resources.

Usually this wouldn't matter, however I think `CreateVertex` can be on a hot path, especially for loading data into Memgraph.

src/storage/v3/storage.cpp
```diff
@@ -515,38 +491,46 @@ ResultSchema<VertexAccessor> Storage::Accessor::CreateVertexAndValidate(
...
   }
   OOMExceptionEnabler oom_exception;
-  auto gid = storage_->vertex_id_.fetch_add(1, std::memory_order_acq_rel);
+  // Extract key properties
+  std::vector<PropertyValue> primary_properties;
+  for (const auto &[property_id, property_value] : properties) {
+    if (storage_->schemas_.IsPropertyKey(primary_label, property_id)) {
+      primary_properties.push_back(property_value);
+    }
+  }
```
antaljanosbenjamin, line 500: What about the order of the properties? I think this way it only works if the primary keys are sorted in the same order in `properties` as the regarding property ids in the schema.

tests/unit/storage_v3_gc.cpp
```diff
@@ -0,0 +1,224 @@
...
+                                       const std::vector<LabelId> &labels,
+                                       const std::vector<std::pair<PropertyId, PropertyValue>> &properties) {
+  auto vtx = acc.CreateVertexAndValidate(primary_label, labels, properties);
+  EXPECT_TRUE(vtx.HasValue());
+  return *vtx;
+}
+
+// TODO: The point of these is not to test GC fully, these are just simple
+// sanity checks. These will be superseded by a more sophisticated stress test
+// which will verify that GC is working properly in a multithreaded environment.
```
antaljanosbenjamin, line 43: I am not sure what do you mean `GC is working properly in a multithreaded environment`, but as we discussed the GC has to be streamlined into the single threaded access to the shards, as we don't want to deal with accesses from multiple threads simultaneously.

tests/unit/CMakeLists.txt
```diff
@@ -334,33 +334,39 @@ target_link_libraries(${test_prefix}storage_v3_property_store mg-storage-v3 fmt)
 add_unit_test(storage_v3_key_store.cpp)
 target_link_libraries(${test_prefix}storage_v3_key_store mg-storage-v3 rapidcheck rapidcheck_gtest)
 
+add_unit_test(storage_v3_gc.cpp)
```
antaljanosbenjamin, line 337: I see what you did with the tests, and I have to say nice!

antaljanosbenjamin (changes requested):

src/query/v2/procedure/mg_procedure_impl.cpp
jbajic: Because I get `static_assert failed due to requirement '!std::uses_allocator_v<mgp_vertex, memgraph::utils::Allocator<mgp_vertex>>' "Class declares std::uses_allocator but has no valid constructor overload. Refer to 'Uses-allocator construction' rules in C++ reference."` if commenting out vertex creation, not sure why I can just return `mgp_error` here

src/query/v2/db_accessor.hpp
jbajic: Just for the sake of triggers, so I can minimize changes done there

src/query/v2/trigger_context.hpp
jbajic: It wasn't much of a difference just wanted t omake code compilable, but will change it

src/storage/v3/vertex_accessor.hpp
jbajic: Schemas will be removed

tests/unit/storage_v3_gc.cpp
jbajic: That's not me these comments were left from the file I've copied, but I will remove this file, this is just to prove that we did not break anything in this PR

src/storage/v3/storage.cpp
```diff
@@ -506,7 +508,8 @@ ResultSchema<VertexAccessor> Storage::Accessor::CreateVertexAndValidate(
 
   auto acc = storage_->vertices_.access();
   auto *delta = CreateDeleteObjectDelta(&transaction_);
-  auto [it, inserted] = acc.insert({Vertex{delta, primary_label, primary_properties, labels, secondary_properties}});
+  auto [it, inserted] = acc.insert(
+      {Vertex{delta, primary_label, std::move(primary_properties), labels, std::move(secondary_properties)}});
```
antaljanosbenjamin, line 512: I made a mistake as I didn't consider that these property values cannot be moved into the property stores. Feel free to keep this as is (we can fix that later on) or revert it.

tests/unit/storage_v3_gc.cpp
antaljanosbenjamin: As I disable the background thread for GC in my PR, I think this file is unnecessary to add for this short period.

antaljanosbenjamin (approved):

src/storage/v3/vertex_accessor.cpp
```diff
@@ -237,6 +238,44 @@ Result<LabelId> VertexAccessor::PrimaryLabel(const View view) const {
...
 }
 
+Result<PrimaryKey> VertexAccessor::PrimaryKey(const View view) const {
+  bool exists = true;
+  bool deleted = false;
+  Delta *delta = nullptr;
+  {
+    std::lock_guard<utils::SpinLock> guard(vertex_->lock);
+    deleted = vertex_->deleted;
+    delta = vertex_->delta;
```
spacejam, line 248: can we avoid acquiring this mutex?

spacejam (approved):

src/storage/v3/vertex_accessor.cpp
antaljanosbenjamin: That mutex will be eliminated in [my PR](https://github.com/memgraph/memgraph/pull/460) completely.

PR #501, opened 2022-08-11
[T1007-MG < T0997-MG] Authorization on paths

niko4299:


Josipmrden:
Also, you need to set the target branch to be edge filtering, since I branched out of edge filtering, not from the epic. Please fix this before further review, so I don't comment again on 2 PRs same comments

src/query/plan/operator.cpp
```diff
@@ -855,13 +855,27 @@ auto ExpandFromVertex(const VertexAccessor &vertex, EdgeAtom::Direction directio
...
 
   if (direction != EdgeAtom::Direction::IN) {
     auto edges = UnwrapEdgesResult(vertex.OutEdges(view, edge_types));
+    if (context.auth_checker) {
+      (void)std::remove_if(edges.begin(), edges.end(), [&context, &view](const auto &edge) {
+        return !context.auth_checker->Accept(context.user, *context.db_accessor, edge) ||
+               !context.auth_checker->Accept(context.user, *context.db_accessor, edge.From(), view);
+      });
+    }
+
```
Josipmrden, line 878: check if this can be made into one logic, seems like a duplicate

src/query/plan/operator.cpp
```diff
@@ -1043,17 +1059,23 @@ class ExpandVariableCursor : public Cursor {
...
-      AppendEdge(current_edge.first, &edges_on_frame);
       VertexAccessor current_vertex =
           current_edge.second == EdgeAtom::Direction::IN ? current_edge.first.From() : current_edge.first.To();
 
+      if (context.auth_checker &&
+          !context.auth_checker->Accept(context.user, *context.db_accessor, current_vertex, storage::View::OLD))
+        continue;
+
+      AppendEdge(current_edge.first, &edges_on_frame);
+
```
Josipmrden, line 1078: why is this moved, I know that Benjo commented something on this one, but what was the initial ordering? (I'm not sure where you first introduced this, in edge type filtering or here?)

src/query/plan/operator.cpp
```diff
@@ -1258,9 +1290,15 @@ class STShortestPathCursor : public query::plan::Cursor {
...
       for (const auto &vertex : sink_frontier) {
+        if (context.auth_checker &&
+            !context.auth_checker->Accept(context.user, *context.db_accessor, vertex, storage::View::OLD))
+          continue;
+
         if (self_.common_.direction != EdgeAtom::Direction::OUT) {
           auto out_edges = UnwrapEdgesResult(vertex.OutEdges(storage::View::OLD, self_.common_.edge_types));
           for (const auto &edge : out_edges) {
+            if (context.auth_checker && !context.auth_checker->Accept(context.user, *context.db_accessor, edge))
+              continue;
```
Josipmrden, line 1301: Something doesn't feel right here
I think source and sink frontier are first {start_vertex}, and {and_vertex}. Which means, they are already processed in scan all. Which means, we should only check the edge, and then the vertex the edge is expanding to. And then we will have it correct through the whole algorithm?

Josipmrden (changes requested):

src/query/plan/operator.cpp
niko4299: Ordering before was AppendEdge then get current_vertex but here I don't want to AppendEdge on frame if vertex I'm expanding to is not authorized.

tests/unit/query_required_privileges.cpp
```diff
@@ -104,6 +104,7 @@ TEST_F(TestPrivilegeExtractor, AuthQuery) {
       std::vector<std::unordered_map<AuthQuery::FineGrainedPrivilege, std::vector<std::string>>>{};
   auto *query = AUTH_QUERY(AuthQuery::Action::CREATE_ROLE, "", "role", "", nullptr, std::vector<AuthQuery::Privilege>{},
                            label_privileges, edge_type_privileges);
+
```
Josipmrden, line 107: remove empty line as it is the only change in the file

tests/e2e/fine_grained_access/path_filtering_tests.py
```diff
@@ -0,0 +1,432 @@
...
+    total_paths_results = common.execute_and_fetch_all(
+        user_connnection.cursor(), "MATCH p=(n)-[r *wShortest (r, n | r.weight)]->(m) RETURN p;"
+    )
+    path_length_result = common.execute_and_fetch_all(
+        user_connnection.cursor(),
+        "MATCH p=(n:label0)-[r *wShortest (r, n | r.weight) path_length]->(m:label4) RETURN path_length;",
+    )
+
+    assert len(total_paths_results) == 16
+    assert path_length_result[0][0] == 20
```
Josipmrden, line 23: it would be good if you had the list of expected results in an array and then do something like this:
```
expected results = [...]
assert len(total_paths_results) = len(expected_results)
for res in total_paths_results:
  assert res in expected_results
```

tests/e2e/fine_grained_access/path_filtering_tests.py
```diff
@@ -0,0 +1,432 @@
...
+    total_paths_results = common.execute_and_fetch_all(
+        user_connnection.cursor(), "MATCH p=(n)-[r *wShortest (r, n | r.weight)]->(m) RETURN p;"
+    )
+    path_length_result = common.execute_and_fetch_all(
+        user_connnection.cursor(),
+        "MATCH p=(n:label0)-[r *wShortest (r, n | r.weight) path_length]->(m:label4) RETURN path_length;",
+    )
+
+    assert len(total_paths_results) == 11
+    assert path_length_result[0][0] == 30
```
Josipmrden, line 99: same here

tests/e2e/fine_grained_access/path_filtering_tests.py
```diff
@@ -0,0 +1,432 @@
...
+    total_paths_results = common.execute_and_fetch_all(
+        user_connnection.cursor(), "MATCH p=(n)-[r *wShortest (r, n | r.weight)]->(m) RETURN p;"
+    )
+    path_length_result = common.execute_and_fetch_all(
+        user_connnection.cursor(),
+        "MATCH p=(n:label0)-[r *wShortest (r, n | r.weight) path_length]->(m:label4) RETURN path_length;",
+    )
+
+    assert len(total_paths_results) == 16
+    assert path_length_result[0][0] == 25
```
Josipmrden, line 122: same here

tests/e2e/fine_grained_access/path_filtering_tests.py
```diff
@@ -0,0 +1,432 @@
...
+    common.execute_and_fetch_all(admin_connection.cursor(), "REVOKE LABELS * FROM user;")
+    common.execute_and_fetch_all(admin_connection.cursor(), "REVOKE EDGE_TYPES * FROM user;")
+    common.execute_and_fetch_all(admin_connection.cursor(), "GRANT READ ON LABELS * TO user;")
+    common.execute_and_fetch_all(admin_connection.cursor(), "GRANT READ ON EDGE_TYPES * TO user;")
+
+    source_destination_path = common.execute_and_fetch_all(
+        user_connnection.cursor(), "MATCH p=(n:label0)-[*]->(m:label4) RETURN p;"
+    )
+
+    assert len(source_destination_path) == 15
```
Josipmrden, line 137: same here

tests/e2e/fine_grained_access/path_filtering_tests.py
```diff
@@ -0,0 +1,432 @@
...
+    common.execute_and_fetch_all(
+        admin_connection.cursor(), "GRANT READ ON LABELS :label0, :label2, :label3, :label4 TO user;"
+    )
+    common.execute_and_fetch_all(admin_connection.cursor(), "DENY READ ON LABELS :label1 TO user;")
+    common.execute_and_fetch_all(admin_connection.cursor(), "GRANT READ ON EDGE_TYPES * TO user;")
+    source_destination_paths = common.execute_and_fetch_all(
+        user_connnection.cursor(), "MATCH p=(n:label0)-[*]->(m:label4) RETURN nodes(p);"
+    )
+
+    assert len(source_destination_paths) == 6
```
Josipmrden, line 203: same here

tests/e2e/fine_grained_access/path_filtering_tests.py
```diff
@@ -0,0 +1,432 @@
...
+    common.execute_and_fetch_all(
+        admin_connection.cursor(), "GRANT READ ON EDGE_TYPES :edge_type_1, :edge_type_2, :edge_type_4 TO user;"
+    )
+    common.execute_and_fetch_all(admin_connection.cursor(), "DENY READ ON EDGE_TYPES :edge_type_3 TO user;")
+
+    source_destination_paths = common.execute_and_fetch_all(
+        user_connnection.cursor(), "MATCH p=(n:label0)-[r *]->(m:label4) RETURN r;"
+    )
+
+    assert len(source_destination_paths) == 6
```
Josipmrden, line 223: same here and every path below that you tested. sorry but it's bulletproof like that

tests/e2e/streams/conftest.py
```diff
@@ -37,29 +37,22 @@ def connection():
...
     for topic in topics:
-        topics_to_create.append(
-            NewTopic(
-                name=topic,
-                num_partitions=1,
-                replication_factor=1))
+        topics_to_create.append(NewTopic(name=topic, num_partitions=1, replication_factor=1))
 
     admin_client.create_topics(new_topics=topics_to_create, timeout_ms=5000)
     yield topics
```
Josipmrden, line 58: delete changes on this file as it is not modified within the scope of the PR

tests/unit/bfs_common.hpp
```diff
@@ -440,3 +450,213 @@ void BfsTest(Database *db, int lower_bound, int upper_bound, memgraph::query::Ed
...
+            EXPECT_EQ(results.size(), 12);
+          }
+          break;
+        case memgraph::query::EdgeAtom::Direction::BOTH:
+          if (known_sink) {
+            EXPECT_EQ(results.size(), 12);
+          } else {
+            EXPECT_EQ(results.size(), 16);
+          }
+          break;
```
Josipmrden, line 656: I think you should as well here make lists and check for individual inputs, and not just sizes

tests/unit/bfs_fine_grained.cpp
```diff
@@ -0,0 +1,107 @@
...
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#include <gtest/internal/gtest-param-util-generated.h>
+#include <optional>
+#include <string>
+#include <unordered_map>
+#include "auth/models.hpp"
```
Josipmrden, line 16: empty line between. + gtest comes below of unordered map and above of auth models

Josipmrden (changes requested):

src/auth/models.cpp
```diff
@@ -280,27 +282,27 @@ const std::unordered_map<std::string, uint64_t> &FineGrainedAccessPermissions::G
 }
 const std::optional<uint64_t> &FineGrainedAccessPermissions::GetGlobalPermission() const { return global_permission_; };
 
-uint64_t FineGrainedAccessPermissions::CalculateGrant(FineGrainedPermission label_permission) {
+uint64_t FineGrainedAccessPermissions::CalculateGrant(FineGrainedPermission fine_grained_permission) {
   uint64_t shift{1};
   uint64_t result{0};
-  auto uint_label_permission = static_cast<uint64_t>(label_permission);
+  auto uint_fine_grained_permission = static_cast<uint64_t>(fine_grained_permission);
```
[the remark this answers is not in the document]
BorisTasevski: No, we've discussed this. Binary operations do implicit conversion to uint64_t, so i wanted to avoid that

tests/e2e/streams/conftest.py
```diff
@@ -37,29 +37,22 @@ def connection():
 
 
 def get_topics(num):
-    return [f'topic_{i}' for i in range(num)]
+    return [f"topic_{i}" for i in range(num)]
```
BorisTasevski, line 40: Outside of scope of this PR. Please revert

tests/unit/bfs_common.hpp
```diff
@@ -440,3 +450,213 @@ void BfsTest(Database *db, int lower_bound, int upper_bound, memgraph::query::Ed
...
+
+void BfsTestWithFineGrainedFiltering(Database *db, int lower_bound, int upper_bound,
+                                     memgraph::query::EdgeAtom::Direction direction,
+                                     std::vector<std::string> edge_types, bool known_sink,
+                                     FineGrainedTestType fine_grained_test_type) {
+  auto storage_dba = db->Access();
+  memgraph::query::DbAccessor dba(&storage_dba);
+  memgraph::query::AstStorage storage;
+  memgraph::query::ExecutionContext context{&dba};
+  memgraph::query::Symbol blocked_sym = context.symbol_table.CreateSymbol("blocked", true);
```
BorisTasevski, line 462: Please don't use abbreviations

tests/unit/query_plan_match_filter_return.cpp
```diff
@@ -1006,6 +1651,8 @@ class QueryPlanExpandWeightedShortestPath : public testing::Test {
     for (int i = 0; i < 5; i++) {
       v.push_back(dba.InsertVertex());
       ASSERT_TRUE(v.back().SetProperty(prop.second, memgraph::storage::PropertyValue(i)).HasValue());
+      auto label = "l" + std::to_string(i);
```
BorisTasevski, line 1654: use += instead of +

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -2257,7 +2257,8 @@ void NextPermitted(mgp_vertices_iterator &it) {
   }
 
   while (it.current_it != it.vertices.end()) {
-    if (it.graph->ctx->auth_checker->Accept(*it.graph->ctx->db_accessor, *it.current_it, it.graph->view)) {
+    if (it.graph->ctx->auth_checker->Accept(*it.graph->ctx->db_accessor, *it.current_it, it.graph->view,
+                                            memgraph::auth::FineGrainedPermission::READ)) {
```
antaljanosbenjamin, line 2261: I guess the `WRITE` part will be checked in the function that are modifying the vertices, right?

Also one question popped up into my mind: shouldn't be `it.graph->ctx->auth_checker` checked against `nullptr`? I mgiht be wrong here.

antaljanosbenjamin (changes requested):

src/query/procedure/mg_procedure_impl.cpp
BorisTasevski: > I guess the `WRITE` part will be checked in the function that are modifying the vertices, right?
> 
> Also one question popped up into my mind: shouldn't be `it.graph->ctx->auth_checker` checked against `nullptr`? I mgiht be wrong here.

It is checked at the top of the method

BorisTasevski: > I guess the `WRITE` part will be checked in the function that are modifying the vertices, right?

Correct, but still this discussion is on change that is out of scope of this PR :)

tests/unit/query_plan_match_filter_return.cpp
niko4299: so code should look like
```auto label = "l";
   label+=std::to_string(i);``` ?

File: src/query/procedure/mg_procedure_impl.cpp (hunk 18 above)
Comment (niko4299) [reply src/query/procedure/mg_procedure_impl.cpp:2261]: I changed accept method in auth_checker so I needed to change it here also.

File: src/query/plan/operator.cpp [hunk 19]
```diff
@@ -33,6 +31,7 @@
 #include "query/frontend/semantic/symbol_table.hpp"
 #include "query/interpret/eval.hpp"
 #include "query/path.hpp"
+#include "query/plan/operator.hpp"
```
antaljanosbenjamin, line 34: Actually this header was in the right spot, because this is the main header for the source file. That always goes first. Search "Includes" in our [public](https://memgraph.notion.site/C-Code-Conventions-7497766e0e1647b987733b9e94171423) code style guide.

antaljanosbenjamin (approved):
Only one comment from me!

tests/unit/query_plan_match_filter_return.cpp
BorisTasevski: Or use some sort of string formatting, such as stringstream

BorisTasevski (changes requested):
Not all comments have been addressed, please look into all proposed changes. 

And once again, in future please don't change code outside of the scope of the PR. For example passing of the `memgraph::storage::View` enum as const reference has been fixed, which in general is something that should be done, but it is outside of the scope of this particular PR. It pollutes the code review and introduces unnecessary overhead in a PR that is already very large and long-running. 

If these issues are discovered, put them in some sort of public list on notion, we can create tech debt task and fix them, but after the main task is completed.

tests/unit/bfs_fine_grained.cpp
niko4299: I fixed this but for some reason this is not outdated, you can see how I fixed it here https://github.com/memgraph/memgraph/blob/T0997-MG-authorization-on-paths/tests/unit/bfs_fine_grained.cpp

tests/unit/bfs_common.hpp
niko4299: https://github.com/memgraph/memgraph/blob/T0997-MG-authorization-on-paths/tests/unit/bfs_common.hpp also i changed this and git didn't catch differences

Josipmrden (approved):

PR #502, opened 2022-08-13
[master < ] Fix RETURN true, false query

### LINKED ISSUE #470: Not executable query with two boolean values
**Memgraph version**
2.3.1

**Environment**
Windows, Docker

**Describe the bug**
If you run query `RETURN true, false;` you get error `Client received exception: Multiple results with the same name 'true' are not allowed.`.

**To Reproduce**
1. Run the following query 'RETURN true, false;'

**Expected behavior**
I expect to get result with two values `true` and `false`.

gitbuda:
[master < Task] PR
- [x] Provide the full content or a guide for the final git message -> `Fix RETURN true, false query`

Fixes #470

The fix only applies to this specific case. To solve the problem in general, some additional alias logic is required. It's possible to solve a problem by manually giving aliases to the expressions that boil down to the same stripped value, e.g.:
* `RETURN 1+2, 3+4;           // fails because that's replaced during the query stripping with 0+0, 0+0`
* `RETURN 1+2 AS x, 3+4 AS y; // OK`

### DIFF
--- a/src/query/frontend/stripped.cpp
+++ b/src/query/frontend/stripped.cpp
@@ -124,9 +124,9 @@ StrippedQuery::StrippedQuery(const std::string &query) : original_(query) {
         // We don't strip NULL, since it can appear in special expressions
         // like IS NULL and IS NOT NULL, but we strip true and false keywords.
         if (utils::IEquals(token.second, "true")) {
-          replace_stripped(token_index, true, kStrippedBooleanToken);
+          replace_stripped(token_index, true, kStrippedBooleanTrueToken);
         } else if (utils::IEquals(token.second, "false")) {
-          replace_stripped(token_index, false, kStrippedBooleanToken);
+          replace_stripped(token_index, false, kStrippedBooleanFalseToken);
         } else {
           token_strings.push_back(token.second);
         }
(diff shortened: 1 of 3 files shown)


42jeremy:
Is the workflow common/important enough so we want a e2e test to cover it?

42jeremy (approved):

42jeremy (commented):
@gitbuda Is the workflow common/important enough so we want a e2e test to cover it?

jbajic (changes requested):
This is a specific solution to a more general problem. The issue is that we replace all returns of constant expressions with default values, `true`, `0`, `0.0`, and this solves only the bool part, but e.g. `RETURN 1,2;` and `RETURN 0.1, 0.2;` still fails. I would close this issue when all of the cases are covered.

gitbuda:
> This is a specific solution to a more general problem. The issue is that we replace all returns of constant expressions with default values, `true`, `0`, `0.0`, and this solves only the bool part, but e.g. `RETURN 1,2;` and `RETURN 0.1, 0.2;` still fails. I would close this issue when all of the cases are covered.

You're right and good catch! I would explore a bit more to find more details on what actually has to be done and then decide, stay tuned... + moving this PR to the Draft stage

gitbuda:
Closing this in the end, because the solution is not full + it's easy to write an alternative query that will work (using aliases).

PR #506, opened 2022-08-17
[E129-MG < T997-MG] Show label privileges

Josipmrden:
[E129-MG-label-based-authorization  < T997 Show Privileges] PR

The PR represents showing label based privileges in the same command as `SHOW PRIVILEGES;` query. It is expanded to show what is the visibility of node labels and edge types in the stored user and role fine grained privileges.

- [x] Write E2E tests


src/auth/models.cpp
```diff
@@ -101,6 +101,35 @@ std::string PermissionLevelToString(PermissionLevel level) {
   }
 }
 
+LabelPermission PermissionToLabelPermission(uint64_t permission) {
+  if (permission & LabelPermission::CREATE_DELETE) {
```
BorisTasevski, line 105: Here a temporary object for which we are not really interested in is created, maybe use `&&` to produce boolean.

src/auth/models.cpp
```diff
@@ -370,6 +399,14 @@ Permissions &Role::permissions() { return permissions_; }
 const FineGrainedAccessHandler &Role::fine_grained_access_handler() const { return fine_grained_access_handler_; }
 FineGrainedAccessHandler &Role::fine_grained_access_handler() { return fine_grained_access_handler_; }
 
+const FineGrainedAccessPermissions &Role::GetFineGrainedAccessLabelPermissions() const {
```
BorisTasevski, line 402: Do we need GetFineGrainedAccessLabelPermissions or GetFineGrainedAccessEdgeTypePermissions?

src/glue/auth_handler.cpp
```diff
@@ -0,0 +1,581 @@
...
+// included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
+// License, and you may not use this file except in compliance with the Business Source License.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#include "glue/auth_handler.hpp"
+
```
BorisTasevski, line 13: remove empty line

src/glue/auth_handler.cpp
```diff
@@ -0,0 +1,581 @@
...
+    permission_description.append(" PERMISSION ");
+    permission_description.append(permission_level_representation);
+    permission_description.append(" TO ");
+    permission_description.append(actor);
+
+    fine_grained_permissions.push_back(
+        FineGrainedPermissionForPrivilegeResult{permission_representation, permission_level, permission_description});
+  }
+
+  for (const auto &permission : permissions.GetPermissions()) {
```
BorisTasevski, line 139: use unpacking here

const auto &[label, permission]...

src/glue/auth_handler.cpp
```diff
@@ -0,0 +1,581 @@
...
+    const auto label = permission.first;
+    auto permission_level = memgraph::auth::PermissionToLabelPermission(permission.second);
+
+    const auto &permission_representation = permission_type + " :" + permission.first;
+    const auto &permission_level_representation =
+        permission_level == memgraph::auth::LabelPermission::NO_PERMISSION ? "DENIED" : "GRANTED";
+
+    std::string permission_description;
+    permission_description.append(permission_type);
```
BorisTasevski, line 148: replace these appends with some sort of string creation, ie `stringstream`

tests/e2e/lba_procedures/show_privileges.py
```diff
@@ -0,0 +1,75 @@
...
+        ("LABEL :Label6", "EDIT", "LABEL PERMISSION GRANTED TO USER"),
+        ("LABEL :Label7", "NO_PERMISSION", "LABEL PERMISSION DENIED TO USER"),
+    ]
+
+    cursor = connect(username="Josip", password="").cursor()
+    result = execute_and_fetch_all(cursor, "SHOW PRIVILEGES FOR Josip;")
+
+    assert len(result) == 30
+
+    fine_privilege_results = []
```
BorisTasevski, line 63:
```suggestion
    fine_privilege_results = [res for res in result if res[0] not in BASIC_PRIVILEGES]
```

BorisTasevski (changes requested):

BorisTasevski (approved):

tests/e2e/lba_procedures/show_privileges.py
```diff
@@ -0,0 +1,93 @@
...
+    result = execute_and_fetch_all(cursor, "SHOW PRIVILEGES FOR Josip;")
+
+    assert len(result) == 30
+
+    fine_privilege_results = [res for res in result if res[0] not in BASIC_PRIVILEGES]
+
+    assert len(fine_privilege_results) == len(expected_assertions_josip)
+
+    for assertion in expected_assertions_josip:
+        assert assertion in fine_privilege_results
```
jbajic, line 68:
```suggestion
    assert set(fine_privilege_results) == set(expected_assertions_josip)
```

jbajic (changes requested):

tests/unit/auth_handler.cpp
```diff
@@ -0,0 +1,724 @@
...
+TEST_F(AuthQueryHandlerFixture, GivenUserWhenNoDeniesOrGrantsThenNothingIsReturned) {
+  memgraph::auth::User user = memgraph::auth::User{user_name, "", perms, handler};
+  auth->SaveUser(user);
+
+  { ASSERT_TRUE(auth_handler.GetUsernames().size()); }
+
+  {
+    auto privileges = auth_handler.GetPrivileges(user_name);
+
+    ASSERT_FALSE(privileges.size());
```
antaljanosbenjamin, line 63: Consider using `ASSERT_EQ`, because then it will print out the actual value in case of a failure. By implicitly converting `size_t` to `bool`, it will only print `true` for `false`.

tests/unit/auth_handler.cpp
```diff
@@ -0,0 +1,724 @@
...
+  }
+}
+
+TEST_F(AuthQueryHandlerFixture, GivenUserWhenAddedGrantPermissionThenItIsReturned) {
+  perms.Grant(memgraph::auth::Permission::MATCH);
+  memgraph::auth::User user = memgraph::auth::User{user_name, "", perms, handler};
+  auth->SaveUser(user);
+
+  auto privileges = auth_handler.GetPrivileges(user_name);
+  ASSERT_TRUE(privileges.size() == 1);
```
antaljanosbenjamin, line 73: `ASSERT_EQ(privileges.size(), 1)` will produce a better error message, because the actual value of `privileges.size()` will be printed also.

jbajic (approved):

PR #513, opened 2022-08-24
[T1007-MG < T1016-MG] Added authorization in create and delete operators

niko4299:


tests/unit/query_plan_v2_create_set_remove_delete.cpp
```diff
@@ -11,10 +11,12 @@
...
 #include <gtest/gtest.h>
 
-#include "query_plan_common.hpp"
-
+#include "auth/models.hpp"
+#include "glue/auth_checker.hpp"
 #include "query/frontend/semantic/symbol_table.hpp"
 #include "query/plan/operator.hpp"
+#include "query/typed_value.hpp"
+#include "query_plan_common.hpp"
```
BorisTasevski, line 19: This include is not sorted properly

BorisTasevski (changes requested):
Very good job. Please fix minor issues so that we can merge this as soon as possible

src/query/plan/operator.cpp
```diff
@@ -3203,9 +3231,7 @@ bool Merge::MergeCursor::Pull(Frame &frame, ExecutionContext &context) {
       if (pull_input_) {
         // if we have just now pulled from the input
         // and failed to pull from merge_match, we should create
-        __attribute__((unused)) bool merge_create_pull_result = merge_create_cursor_->Pull(frame, context);
```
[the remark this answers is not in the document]
niko4299: This was removed because it was protection that merge_create always returns true and if it didn't it would throw an error but now if user doesn't have permissions to create then there is a way that merge_create could return false.

BorisTasevski (approved):

src/query/plan/operator.cpp
```diff
@@ -237,6 +239,10 @@ CreateNode::CreateNodeCursor::CreateNodeCursor(const CreateNode &self, utils::Me
 bool CreateNode::CreateNodeCursor::Pull(Frame &frame, ExecutionContext &context) {
   SCOPED_PROFILE_OP("CreateNode");
 
+  if (context.auth_checker && !context.auth_checker->Accept(*context.db_accessor, self_.node_info_.labels,
+                                                            memgraph::auth::FineGrainedPermission::CREATE_DELETE))
+    return false;
```
Josipmrden, line 244: Check with Boris if we need to rather throw an exception here.

src/query/plan/operator.cpp
```diff
@@ -321,6 +327,12 @@ bool CreateExpand::CreateExpandCursor::Pull(Frame &frame, ExecutionContext &cont
 
   if (!input_cursor_->Pull(frame, context)) return false;
 
+  if (context.auth_checker && (!context.auth_checker->Accept(*context.db_accessor, self_.edge_info_.edge_type,
+                                                             memgraph::auth::FineGrainedPermission::CREATE_DELETE) ||
+                               !context.auth_checker->Accept(*context.db_accessor, self_.node_info_.labels,
```
Josipmrden, line 332: Check with Boris if we need to rather throw an exception here.

src/query/plan/operator.cpp
```diff
@@ -2003,7 +2015,16 @@ bool Delete::DeleteCursor::Pull(Frame &frame, ExecutionContext &context) {
   for (TypedValue &expression_result : expression_results) {
     if (MustAbort(context)) throw HintedAbortError();
     if (expression_result.type() == TypedValue::Type::Edge) {
-      auto maybe_value = dba.RemoveEdge(&expression_result.ValueEdge());
+      auto &ea = expression_result.ValueEdge();
+      if (context.auth_checker &&
+          !(context.auth_checker->Accept(*context.db_accessor, ea, auth::FineGrainedPermission::CREATE_DELETE) &&
+            context.auth_checker->Accept(*context.db_accessor, ea.To(), storage::View::NEW,
+                                         auth::FineGrainedPermission::CREATE_DELETE) &&
```
Josipmrden, line 2022: Check with Boris if we need to rather throw an exception here.

src/query/plan/operator.cpp
```diff
@@ -3257,9 +3282,7 @@ bool Merge::MergeCursor::Pull(Frame &frame, ExecutionContext &context) {
       if (pull_input_) {
         // if we have just now pulled from the input
         // and failed to pull from merge_match, we should create
-        __attribute__((unused)) bool merge_create_pull_result = merge_create_cursor_->Pull(frame, context);
-        DMG_ASSERT(merge_create_pull_result, "MergeCreate must never fail");
```
Josipmrden, line 3261: you deleted here some things?

src/query/plan/operator.cpp
Josipmrden: I think Buda mentioned about some feedback message that some of the vertices were not removed, like an info message that comes with the query

Josipmrden (commented):
Minor comments

Josipmrden (approved):

tests/unit/query_plan_create_set_remove_delete.cpp
```diff
@@ -188,6 +291,127 @@ TEST(QueryPlan, CreateExpand) {
...
+  auto test_create_path = [&](bool cycle, int expected_nodes_created, int expected_edges_created,
+                              memgraph::auth::User &user) {
+    memgraph::storage::Storage db;
+    auto storage_dba = db.Access();
+    memgraph::query::DbAccessor dba(&storage_dba);
+
+    memgraph::storage::LabelId label_node_1 = dba.NameToLabel("Node1");
+    memgraph::storage::LabelId label_node_2 = dba.NameToLabel("Node2");
+    auto property = PROPERTY_PAIR("property");
+    memgraph::storage::EdgeTypeId edge_type = dba.NameToEdgeType("edge_type");
```
jbajic, line 304:
```suggestion
    const auto label_node_1 = dba.NameToLabel("Node1");
    const auto label_node_2 = dba.NameToLabel("Node2");
    const auto property = PROPERTY_PAIR("property");
    const auto edge_type = dba.NameToEdgeType("edge_type");
```

src/query/plan/operator.cpp
```diff
@@ -237,6 +239,12 @@ CreateNode::CreateNodeCursor::CreateNodeCursor(const CreateNode &self, utils::Me
 bool CreateNode::CreateNodeCursor::Pull(Frame &frame, ExecutionContext &context) {
   SCOPED_PROFILE_OP("CreateNode");
 
+  if (context.auth_checker && !context.auth_checker->Accept(*context.db_accessor, self_.node_info_.labels,
+                                                            memgraph::auth::FineGrainedPermission::CREATE_DELETE)) {
+    spdlog::error("Error while creating a vertex due to not having enough permission!");
+    return false;
+  }
+
```
antaljanosbenjamin, line 247: do we really need to check this in every pull? not enough to do it once we know the labels?

src/query/plan/operator.cpp
```diff
@@ -26,6 +26,8 @@
 #include <cppitertools/chain.hpp>
 #include <cppitertools/imap.hpp>
 
+#include "auth/models.hpp"
```
antaljanosbenjamin, line 29: `mg-auth` is not a dependency of `mg-query`, thus you cannot include anything from auth.

src/query/plan/operator.cpp
```diff
@@ -321,6 +329,13 @@ bool CreateExpand::CreateExpandCursor::Pull(Frame &frame, ExecutionContext &cont
 
   if (!input_cursor_->Pull(frame, context)) return false;
 
+  if (context.auth_checker && !(context.auth_checker->Accept(*context.db_accessor, self_.edge_info_.edge_type,
+                                                             memgraph::auth::FineGrainedPermission::CREATE_DELETE) &&
+                                context.auth_checker->Accept(*context.db_accessor, self_.node_info_.labels,
+                                                             memgraph::auth::FineGrainedPermission::CREATE_DELETE))) {
+    spdlog::error("Error while creating an edge due to not having enough permission!");
+    return false;
+  }
```
antaljanosbenjamin, line 338: Same here as for create? Isn't this enough once when we know the labels and edge type?

src/query/plan/operator.cpp
```diff
@@ -321,6 +329,13 @@ bool CreateExpand::CreateExpandCursor::Pull(Frame &frame, ExecutionContext &cont
 
   if (!input_cursor_->Pull(frame, context)) return false;
 
+  if (context.auth_checker && !(context.auth_checker->Accept(*context.db_accessor, self_.edge_info_.edge_type,
+                                                             memgraph::auth::FineGrainedPermission::CREATE_DELETE) &&
+                                context.auth_checker->Accept(*context.db_accessor, self_.node_info_.labels,
+                                                             memgraph::auth::FineGrainedPermission::CREATE_DELETE))) {
+    spdlog::error("Error while creating an edge due to not having enough permission!");
```
antaljanosbenjamin, line 336: I would avoid logging this on any higher level than info or debug. Because it is done in every pull it can be easily produce thousands of log messages. Also, it is not an error from memgraphs perspective: it is the expected behavior that we prevent users from accessing vertices/edge to which they don't have access.

antaljanosbenjamin: This is also true for the rest of occurrences.

src/query/plan/operator.cpp
```diff
@@ -2003,7 +2018,17 @@ bool Delete::DeleteCursor::Pull(Frame &frame, ExecutionContext &context) {
...
     if (MustAbort(context)) throw HintedAbortError();
     if (expression_result.type() == TypedValue::Type::Edge) {
-      auto maybe_value = dba.RemoveEdge(&expression_result.ValueEdge());
+      auto &ea = expression_result.ValueEdge();
+      if (context.auth_checker &&
+          !(context.auth_checker->Accept(*context.db_accessor, ea, auth::FineGrainedPermission::CREATE_DELETE) &&
+            context.auth_checker->Accept(*context.db_accessor, ea.To(), storage::View::NEW,
+                                         auth::FineGrainedPermission::CREATE_DELETE) &&
+            context.auth_checker->Accept(*context.db_accessor, ea.From(), storage::View::NEW,
+                                         auth::FineGrainedPermission::CREATE_DELETE))) {
```
antaljanosbenjamin, line 2027: Does the user need `CREATE_DELETE` permission for a vertex in order to delete one of its edges?

antaljanosbenjamin: If this is done in order to make sure the the user has access to the vertices they want to delete, then I might run the authorization check for the vertices before deleting the edges and use `UPDATE` permission here for vertices.

tests/e2e/fine_grained_access/common.py
```diff
@@ -11,6 +11,14 @@
 import mgclient
 
 
+def reset_permissions(admin_cursor):
+    execute_and_fetch_all(admin_cursor, "REVOKE LABELS * FROM user;")
+    execute_and_fetch_all(admin_cursor, "REVOKE EDGE_TYPES * FROM user;")
+    execute_and_fetch_all(admin_cursor, "MATCH(n) DETACH DELETE n;")
+    execute_and_fetch_all(admin_cursor, "CREATE (n:test_delete {name: 'test1'});")
+    execute_and_fetch_all(admin_cursor, "CREATE (n:test_delete_1)-[r:edge_type_delete]->(m:test_delete_2);")
```
antaljanosbenjamin, line 19: It is confusing why `MATCH` and `CREATE` are used in a function called `reset_permissions`, they don't hahve to anything with permissions.

src/query/plan/operator.cpp
antaljanosbenjamin: I think this not the best approach as running a create query without the correct permissions will run without any errors but the users won't see any result. Compared to other privileges, the query will fail with the error that hey, you don't have permission to do that. However this is only my opinion, feel free to discard it.

src/query/plan/operator.cpp
niko4299: I talked about adding checks if user can do certain operation on nodes and edges while preparing query in interpreter.cpp with @Josipmrden and he told me that for this iteration we are going with changes only in operators.

Josipmrden: Yes, due to lack of time and complexity the cypher query can get, we decided to roll with only operators as we have taken a good look at them and possibly extend the functionality in the upcoming months in the interpreter if needed.

antaljanosbenjamin: And what about having a bool member which can be changed from `false` to `true` in the first pull and checking the privileges at the same time? I have concerns about this really plummeting the performance of create when we have privileges.

src/query/auth_checker.hpp
```diff
@@ -37,39 +37,39 @@ class FineGrainedAuthChecker {
 
   [[nodiscard]] virtual bool Accept(const memgraph::query::DbAccessor &dba, const query::VertexAccessor &vertex,
                                     memgraph::storage::View view,
-                                    auth::FineGrainedPermission fine_grained_permission) const = 0;
+                                    query::AuthQuery::FineGrainedPrivilege fine_grained_permission) const = 0;
```
antaljanosbenjamin, line 39: Could you please also remove the `auth/models.hpp` include?

src/query/plan/operator.cpp
```diff
@@ -39,6 +38,7 @@
 #include "query/procedure/cypher_types.hpp"
 #include "query/procedure/mg_procedure_impl.hpp"
 #include "query/procedure/module.hpp"
+#include "spdlog/spdlog.h"
```
antaljanosbenjamin, line 41: This should go one block above as `<spdlog/spdlog.h>`

antaljanosbenjamin (approved):

PR #515, opened 2022-08-28
[T1006-MG < T1017-MG] Add LBA checks to all read procedures in C API

BorisTasevski:


Josipmrden (approved):

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -1966,11 +2006,21 @@ mgp_error mgp_edge_get_type(mgp_edge *e, mgp_edge_type *result) {
...
   *result = &e->from;
   return mgp_error::MGP_ERROR_NO_ERROR;
 }
 
 mgp_error mgp_edge_get_to(mgp_edge *e, mgp_vertex **result) {
+  if (e->from.graph->ctx && e->from.graph->ctx->auth_checker &&
+      !e->from.graph->ctx->auth_checker->Accept(*e->from.graph->ctx->db_accessor, e->from.impl, e->from.graph->view)) {
+    return mgp_error::MGP_ERROR_NO_ERROR;
+  }
+
```
Josipmrden, line 2023: what are you checking here?

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -1827,6 +1828,31 @@ mgp_error mgp_vertex_iter_properties(mgp_vertex *v, mgp_memory *memory, mgp_prop
...
+void NextPermittedEdge(mgp_edges_iterator &it, const bool for_in) {
+  if (!it.source_vertex.graph->ctx || !it.source_vertex.graph->ctx->auth_checker) return;
+
+  auto &impl_it = for_in ? it.in_it : it.out_it;
+  const auto &end = for_in ? it.in->end() : it.out->end();
+
+  if (impl_it) {
+    const auto &auth_checker = it.source_vertex.graph->ctx->auth_checker;
+    const auto db_accessor = *it.source_vertex.graph->ctx->db_accessor;
+    const auto &view = it.source_vertex.graph->view;
```
antaljanosbenjamin, line 1841:
```suggestion
    const auto *auth_checker = it.source_vertex.graph->ctx->auth_checker;
    const auto db_accessor = *it.source_vertex.graph->ctx->db_accessor;
    const auto view = it.source_vertex.graph->view;
```
Reasoning:
1. `auth_checker` is a pointer, by writing `const auto &auth_checker = ...` you introduce another indirection (reference to a pointer), which doesn't have any advantage, but has one drawback: one more indirection. Just copy the pointer and use that.
2. `view` fits into 64 bits, so it makes more sense to copy it, rather then hold a 64bit long pointer and carry the burden of indirection.

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -1910,24 +1942,32 @@ mgp_error mgp_edges_iterator_next(mgp_edges_iterator *it, mgp_edge **result) {
   return WrapExceptions(
       [it] {
         MG_ASSERT(it->in || it->out);
-        auto next = [&](auto *impl_it, const auto &end) -> mgp_edge * {
+        auto next = [it](const bool for_in) -> mgp_edge * {
+          auto impl_it = for_in ? it->in_it : it->out_it;
+          const auto end = for_in ? it->in->end() : it->out->end();
```
antaljanosbenjamin, line 1947: I like what you did here, really nice!

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -1966,11 +2006,21 @@ mgp_error mgp_edge_get_type(mgp_edge *e, mgp_edge_type *result) {
 }
 
 mgp_error mgp_edge_get_from(mgp_edge *e, mgp_vertex **result) {
+  if (e->from.graph->ctx && e->from.graph->ctx->auth_checker &&
+      !e->from.graph->ctx->auth_checker->Accept(*e->from.graph->ctx->db_accessor, e->from.impl, e->from.graph->view)) {
+    return mgp_error::MGP_ERROR_NO_ERROR;
+  }
+
   *result = &e->from;
   return mgp_error::MGP_ERROR_NO_ERROR;
```
antaljanosbenjamin, line 2015: If you keep this check (see the discussion @Josipmrden started), then use `WrapExceptions` as the `Accept` function is not `noexcept`.

src/query/procedure/mg_procedure_impl.cpp
antaljanosbenjamin: I was thinking the same. As the edge iterators are checking to have access to the other side of the edge, I am wondering how you pass in an edge for which this will filter out any vertices.

BorisTasevski: Edge Iterators are filtering out edges if user cant see both sides of the edge

src/query/procedure/mg_procedure_impl.cpp
BorisTasevski: I took me some time to wrap my head around it, but it seems much much nicer to pass a single boolean as a parameter, rather that working with those sausage-like decltype/declvalues mumbo-jumbo. The iterators are of the same type, and it turned out i only needed to decide which one to iterate on... So, iterate on first? No? Ok, on second then. BTW, You missed a subtle bug, that originates in line 1946 :) The bug was actually the biggest drawback of using this approach, since it's not so obvious at first. When i created impl_it, i actually created a copy of the underlying iterator, so when i iterated on impl_it, i actually iterated on the copy. The member iterator remained stationary. I've fixed this in subsequent PRs, with making `auto impl_it` -> `auto &impl_it`

src/query/procedure/mg_procedure_impl.cpp
BorisTasevski: This one is going out, mgp_edges for which the user can't see both vertices are filtered out. So, this check is redundant.

PR #516, opened 2022-08-28
[T1006-MG < T1018-MG] Add LBA checks to all update procedures in C API

BorisTasevski:


src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -1618,6 +1624,11 @@ mgp_error mgp_vertex_set_property(struct mgp_vertex *v, const char *property_nam
 
 mgp_error mgp_vertex_add_label(struct mgp_vertex *v, mgp_label label) {
   return WrapExceptions([=] {
+    if (v->graph->ctx && v->graph->ctx->auth_checker &&
+        !v->graph->ctx->auth_checker->Accept(*v->graph->ctx->db_accessor, v->impl, v->graph->view)) {
+      return;
```
Josipmrden, line 1629: Do we return or throw

Josipmrden (approved):

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -1579,9 +1579,16 @@ memgraph::storage::PropertyValue ToPropertyValue(const mgp_value &value) {
 
 mgp_error mgp_vertex_set_property(struct mgp_vertex *v, const char *property_name, mgp_value *property_value) {
   return WrapExceptions([=] {
+    if (v->graph->ctx && v->graph->ctx->auth_checker &&
+        !v->graph->ctx->auth_checker->Accept(*v->graph->ctx->db_accessor, v->impl, v->graph->view,
+                                             memgraph::query::AuthQuery::FineGrainedPrivilege::UPDATE)) {
+      return;
+    }
+
```
antaljanosbenjamin, line 1587: I would put this after the mutability check, but it is not an important change. My reasoning is using a "write" operation in a read procedure is a "bigger problem" than not having the privilege to modify a certain vertex.

And this will also result in an error in the future, right?

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -2041,6 +2049,11 @@ mgp_error mgp_edge_get_property(mgp_edge *e, const char *name, mgp_memory *memor
 
 mgp_error mgp_edge_set_property(struct mgp_edge *e, const char *property_name, mgp_value *property_value) {
   return WrapExceptions([=] {
+    if (e->from.graph->ctx && e->from.graph->ctx->auth_checker &&
```
antaljanosbenjamin, line 2052: Same as for vertex_set_property.

src/query/procedure/mg_procedure_impl.cpp
BorisTasevski: > I would put this after the mutability check, but it is not an important change. My reasoning is using a "write" operation in a read procedure is a "bigger problem" than not having the privilege to modify a certain vertex.

I've placed the check here as a way of sort of protecting the information about the entity. We can discuss a bit more if You'd like

> And this will also result in an error in the future, right?

Yes this is addressed in the 3rd PR

antaljanosbenjamin (approved):

PR #518, opened 2022-08-29
Add Unwind, Aggregate, Optional and Filter

gitbuda:
[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR

### DIFF
--- a/src/query/v2/plan/operator_distributed.cpp
+++ b/src/query/v2/plan/operator_distributed.cpp
@@ -45,6 +45,7 @@ extern const Event ScanAllByLabelPropertyValueOperator;
 extern const Event ScanAllByIdOperator;
 extern const Event ExpandOperator;
 extern const Event ProduceOperator;
+extern const Event UnwindOperator;
 }  // namespace EventCounter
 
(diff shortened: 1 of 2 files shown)


src/query/v2/plan/operator_distributed.cpp
```diff
@@ -595,4 +597,149 @@ void Produce::ProduceCursor::Shutdown() { input_cursor_->Shutdown(); }
...
+      if (!frame.IsValid()) {
+        continue;
+      }
+      ExpressionEvaluator evaluator(&frame, context.symbol_table, context.evaluation_context, context.db_accessor,
+                                    storage::v3::View::OLD);
+      if (!EvaluateFilter(evaluator, self_.expression_)) {
+        frame.MakeInvalid();
+      }
+    }
+    return true;
```
42jeremy, line 736: Note: you should probably just invalidate frames if we have at least one valid (ie: that we return true), if all frame are invalid then we want to do an extra pull.
For example you can look at the current implementation of expand, there is somehow a similar logic (with var `invalid_frames` )

PR #526, opened 2022-08-31
[project-pineapples < T1026-MG] Add shard requests responses

kostasrim:
[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR

### DIFF
--- a/src/query/v2/requests.hpp
+++ b/src/query/v2/requests.hpp
@@ -1,0 +1,201 @@
+// Copyright 2022 Memgraph Ltd.
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
+// License, and you may not use this file except in compliance with the Business Source License.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#pragma once
+
+#include <iostream>
+#include <optional>
+#include <unordered_map>
+#include <variant>
+#include <vector>
+
+// TODO(kostasrim) update this with CompoundKey, same for the rest of the file.
+using VertexId = size_t;
+using Gid = size_t;
+
+using Hlc = memgraph::coordinator::Hlc;
+
+struct Label {
+  size_t id;
+};
+
+struct EdgeType {
+  std::string name;
+};
+
+struct EdgeId {
+  VertexId id;
+  Gid gid;
+};
+
+struct Vertex {
+  VertexId id;
+  std::vector<Label> labels;
+};
+
+struct Edge {
+  VertexId src;
+  VertexId dst;
+  EdgeType type;
+};
+
+struct PathPart {
+  Vertex dst;
+  Edge edge;
+};
+
+struct Path {
+  Vertex src;
+  std::vector<PathPart> parts;
+}
+
+using Value = TypedValue<Vertex, Edge, Path>;
+
+struct Null {};
+
+struct ValuesMap {
+  std::unordered_map<size_t, Value> values_map;
+};
+
+struct MappedValues {
+  std::vector<ValuesMap> values_map;
+};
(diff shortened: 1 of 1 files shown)


src/query/v2/requests.hpp
```diff
@@ -0,0 +1,201 @@
...
+  //  TODO(antaljanosbenjamin): All of the special values should be communicated through a single vertex object
+  //                            after schema is implemented
+  //  Special values are accepted:
+  //  * __mg__labels
+  std::optional<std::vector<std::string>> src_vertex_properties;
+  //  TODO(antaljanosbenjamin): All of the special values should be communicated through a single vertex object
+  //                            after schema is implemented
+  //  Special values are accepted:
+  //  * __mg__dst_id (Vertex, but without labels)
+  //  * __mg__type (binary)
```
antaljanosbenjamin, line 148:
```suggestion
  //  Special values are accepted:
  //  * __mg__labels
  std::optional<std::vector<std::string>> src_vertex_properties;
  //  TODO(antaljanosbenjamin): All of the special values should be communicated through a single vertex object
  //                            after schema is implemented
  //  Special values are accepted:
  //  * __mg__dst_id (Vertex, but without labels)
  //  * __mg__type (binary)
```

The special values should also be communicated through property ids. I think we can hardcode these values (either start the mapping in coordinator from an offset, or hardcode that the first few values are fixed).

src/query/v2/requests.hpp
```diff
@@ -0,0 +1,201 @@
...
+#include <optional>
+#include <unordered_map>
+#include <variant>
+#include <vector>
+
+// TODO(kostasrim) update this with CompoundKey, same for the rest of the file.
+using VertexId = size_t;
+using Gid = size_t;
+
+using Hlc = memgraph::coordinator::Hlc;
```
spacejam, line 24: this file needs to be pulled into this pr

src/query/v2/requests.hpp
```diff
@@ -0,0 +1,201 @@
...
+  Expression expression;
+  OrderingDirection direction;
+};
+
+enum class StorageView { OLD = 0, NEW = 1 };
+
+struct ScanVerticesRequest {
+  Hlc transaction_id;
+  size_t start_id;
+  std::optional<std::vector<std::string>> props_to_return;
```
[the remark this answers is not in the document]
spacejam: PropertyId is already defined in `storage/v3/id_types.hpp` so it can be pulled in from that

src/query/v2/requests.hpp
```diff
@@ -0,0 +1,201 @@
...
+  // in order to avoid duplicated keys (two threads might map the same
+  // property with different numbers) and multiple passes (to unify the
+  // mapping amond result returned from different shards).
+  std::vector<ExpandOneResultRow> result;
+  std::optional<std::unordered_map<size_t, std::string>> property_name_map;
+};
+
+struct NewVertex {
+  std::vector<size_t> label_ids;
+  std::map<size_t, Value> properties;
```
spacejam, line 183: should this be PropertyId instead of size_t?

src/query/v2/requests.hpp
```diff
@@ -0,0 +1,201 @@
...
+};
+
+struct ExpandOneResponse {
+  // This approach might not suit the expand with per shard parrallelization,
+  // because the property_name_map has to be accessed from multiple threads
+  // in order to avoid duplicated keys (two threads might map the same
+  // property with different numbers) and multiple passes (to unify the
+  // mapping amond result returned from different shards).
+  std::vector<ExpandOneResultRow> result;
+  std::optional<std::unordered_map<size_t, std::string>> property_name_map;
```
spacejam, line 178: can this map be removed?

src/query/v2/requests.hpp
kostasrim: yes and the above vector should be `std::vector<Label>` which is basically the label_id

spacejam (approved):

antaljanosbenjamin (approved):

PR #528, opened 2022-09-01
[E118-MG < T1027-MG] Enable indices for storage v3

jbajic:
[master < Task] PR


src/storage/v3/shard.cpp
```diff
@@ -1065,8 +1065,10 @@ const std::string &Shard::EdgeTypeToName(EdgeTypeId edge_type) const {
 }
 
 bool Shard::CreateIndex(LabelId label, const std::optional<uint64_t> desired_commit_timestamp) {
-  // TODO Fix Index
-  return false;
+  // Do label indexes make sense on primary labels?
```
antaljanosbenjamin, line 1068: No, because the shards are the label indices over primary labels.

src/storage/v3/shard.cpp
```diff
@@ -1075,9 +1077,14 @@ bool Shard::CreateIndex(LabelId label, const std::optional<uint64_t> desired_com
 }
 
 bool Shard::CreateIndex(LabelId label, PropertyId property, const std::optional<uint64_t> desired_commit_timestamp) {
-  // TODO Fix Index
-  // if (!indices_.label_property_index.CreateIndex(label, property, labelspace.access())) return false;
-  return false;
+  if (label == primary_label_ && schemas_.GetSchema(primary_label_)->second.size() == 1 &&
+      schemas_.GetSchema(primary_label_)->second[0].property_id == property) {
+    // Index already exists on primary key
```
antaljanosbenjamin, line 1082: It is a good question whether we say it exists in this case or the user cannot create such an index. I am not sure which is better. For M1 I guess this is fine, but later on we have to make this clear. Let's put a TODO here I think.

src/storage/v3/shard.cpp
```diff
@@ -1075,9 +1077,14 @@ bool Shard::CreateIndex(LabelId label, const std::optional<uint64_t> desired_com
...
+    return false;
+  }
```
antaljanosbenjamin, line 1084: Based on this logic it is possible to create an index on a secondary label + primary property. Because the indices are only "checked against" in label and property updates, it is possible to construct a vertex with predefined secondary labels. In that case the vertex will be left out from the index if I am not mistaken. Do we have any test to check this exact use case? If it complicates things to much, then we can "fall back" to the old way of creating a vertex: only primary label (not even that after the other PR of yours) and the primary key is provided, everything else has to be set using `SetProperties` or `AddLabel`.

src/storage/v3/shard.cpp
jbajic: Will add that check as well then

src/storage/v3/shard.cpp
jbajic: Will cover this with test cases

tests/unit/storage_v3_indices.cpp
```diff
@@ -0,0 +1,989 @@
...
+          {{primary_property, PropertyValue(primary_key_id++)}, {prop_id, PropertyValue(vertex_id++)}});
+      ASSERT_TRUE(vertex.HasValue());
+    }
+
+    EXPECT_THAT(GetPrimaryKeyIds(acc.Vertices(label1, prop_id, View::NEW), View::NEW),
+                UnorderedElementsAre(0, 1, 2, 3, 4, 5, 6, 7, 8, 9));
+    EXPECT_THAT(GetPrimaryKeyIds(acc.Vertices(label1, prop_id, View::OLD), View::OLD),
+                UnorderedElementsAre(0, 1, 2, 3, 4));
+  }
+}
```
jbajic, line 988: These are new tests

tests/unit/storage_v3_indices.cpp
```diff
@@ -860,9 +873,117 @@ TEST_F(IndexTest, LabelPropertyIndexCreateWithExistingPrimaryKey) {
...
+      auto acc = storage.Access();
+      EXPECT_THAT(GetPrimaryKeyIds(acc.Vertices(label1, prop_id, View::OLD), View::OLD),
+                  UnorderedElementsAre(0, 1, 2, 3, 4));
+    }
+  }
+  {
+    EXPECT_TRUE(storage.DropIndex(label1, prop_id));
+    {
+      auto acc = storage.Access();
+      EXPECT_FALSE(acc.LabelIndexExists(label1));
```
antaljanosbenjamin, line 966: Shouldn't be this `LabelPropertyIndexExists`?
```suggestion
      EXPECT_FALSE(acc.LabelPropertyIndexExists(label1, prop_id));
```

tests/unit/storage_v3_indices.cpp
```diff
@@ -860,9 +873,117 @@ TEST_F(IndexTest, LabelPropertyIndexCreateWithExistingPrimaryKey) {
...
+    }
+    EXPECT_EQ(storage.ListAllIndices().label.size(), 0);
```
antaljanosbenjamin, line 968:
```suggestion
    EXPECT_EQ(storage.ListAllIndices().label_property.size(), 0);
```

antaljanosbenjamin (approved):
Two small changes, after that I think we can merge this.

PR #537, opened 2022-09-07
[E129-MG < T1041-MG] Add authorization in setLabels, removeLabels, Allshortestpath cursor

niko4299:


tests/e2e/fine_grained_access/workloads.yaml
```diff
@@ -68,6 +68,7 @@ workloads:
     binary: "tests/e2e/pytest_runner.sh"
     args: ["fine_grained_access/edge_type_filtering_tests.py"]
     <<: *edge_type_filtering_cluster
+
```
BorisTasevski, line 71: please remove newline

src/query/plan/operator.cpp
```diff
@@ -25,6 +25,7 @@
 
 #include <cppitertools/chain.hpp>
 #include <cppitertools/imap.hpp>
+#include "auth/models.hpp"
```
BorisTasevski, line 28: This import should be removed

src/query/plan/operator.cpp
```diff
@@ -2740,6 +2755,11 @@ SetLabels::SetLabelsCursor::SetLabelsCursor(const SetLabels &self, utils::Memory
 bool SetLabels::SetLabelsCursor::Pull(Frame &frame, ExecutionContext &context) {
   SCOPED_PROFILE_OP("SetLabels");
 
+  if (context.auth_checker && !context.auth_checker->Accept(*context.db_accessor, self_.labels_,
+                                                            memgraph::query::AuthQuery::FineGrainedPrivilege::UPDATE)) {
```
BorisTasevski, line 2759: This should be Create_Delete not Update

src/query/plan/operator.cpp
```diff
@@ -2880,6 +2900,10 @@ RemoveLabels::RemoveLabelsCursor::RemoveLabelsCursor(const RemoveLabels &self, u
 
 bool RemoveLabels::RemoveLabelsCursor::Pull(Frame &frame, ExecutionContext &context) {
   SCOPED_PROFILE_OP("RemoveLabels");
+  if (context.auth_checker && !context.auth_checker->Accept(*context.db_accessor, self_.labels_,
```
BorisTasevski, line 2903: This should be Create_Delete not Update

BorisTasevski (changes requested):

src/query/plan/operator.cpp
```diff
@@ -2740,13 +2754,23 @@ SetLabels::SetLabelsCursor::SetLabelsCursor(const SetLabels &self, utils::Memory
 bool SetLabels::SetLabelsCursor::Pull(Frame &frame, ExecutionContext &context) {
   SCOPED_PROFILE_OP("SetLabels");
 
+  if (context.auth_checker &&
+      !context.auth_checker->Accept(*context.db_accessor, self_.labels_,
+                                    memgraph::query::AuthQuery::FineGrainedPrivilege::CREATE_DELETE)) {
+    throw QueryRuntimeException("Couldn't remove label due to not having enough permission!");
```
BorisTasevski, line 2760: Wrong error message

src/query/plan/operator.cpp
```diff
@@ -2740,13 +2754,23 @@ SetLabels::SetLabelsCursor::SetLabelsCursor(const SetLabels &self, utils::Memory
...
   if (!input_cursor_->Pull(frame, context)) return false;
 
   TypedValue &vertex_value = frame[self_.input_symbol_];
   // Skip setting labels on Null (can occur in optional match).
   if (vertex_value.IsNull()) return true;
   ExpectType(self_.input_symbol_, vertex_value, TypedValue::Type::Vertex);
   auto &vertex = vertex_value.ValueVertex();
+  if (context.auth_checker && !context.auth_checker->Accept(*context.db_accessor, vertex, storage::View::OLD,
+                                                            memgraph::query::AuthQuery::FineGrainedPrivilege::UPDATE)) {
+    throw QueryRuntimeException("Couldn't remove label due to not having enough permission!");
```
BorisTasevski, line 2772: Wrong error message

src/query/plan/operator.cpp
```diff
@@ -2740,13 +2754,23 @@ SetLabels::SetLabelsCursor::SetLabelsCursor(const SetLabels &self, utils::Memory
 bool SetLabels::SetLabelsCursor::Pull(Frame &frame, ExecutionContext &context) {
   SCOPED_PROFILE_OP("SetLabels");
 
+  if (context.auth_checker &&
```
BorisTasevski, line 2757: Move this if and merge it with the one that checks update on vertex

src/query/plan/operator.cpp
```diff
@@ -2880,6 +2904,11 @@ RemoveLabels::RemoveLabelsCursor::RemoveLabelsCursor(const RemoveLabels &self, u
 
 bool RemoveLabels::RemoveLabelsCursor::Pull(Frame &frame, ExecutionContext &context) {
   SCOPED_PROFILE_OP("RemoveLabels");
+  if (context.auth_checker &&
```
BorisTasevski, line 2907: Move this if and merge it with the one that checks update on vertex

BorisTasevski (changes requested):

Josipmrden (commented):
Fine for me, just discuss the ongoing changes with Boris

src/query/plan/operator.cpp
BorisTasevski: I missed an important part. This will result in two ifdefs and two licence checks

BorisTasevski (changes requested):

src/query/plan/operator.cpp
niko4299: But with this I will save a lot of checks for example let say we are setting/removing 2 labels and we have 1 milion nodes in database if we do it like this we will have only one check for both labels at start but if we do it how you propose we will have  1 milion checks for those same labels for each node that we pull. I would say that one additonal licence check and one additonal ifdef that is in compiling time is nothing compared to one milion check.

Josipmrden: Actually I think Boris is right. If you look at the interpreter.cpp file, you have this line of code

```
  for (; !n || i < n; ++i) {
    if (!pull_result()) {
      break;
    }

    if (!output_symbols.empty()) {
      stream_values();
    }
  }
```
if we have for example a query (1M nodes)
`MATCH (n) SET n:L2:L3 RETURN n;`
The operators would be
`Produce -> SetLabels -> ScanAll -> Once`

And the outer loop which I pasted above is responsible for doing it iteratively.
In the case of 1M nodes, when scanning the first node and going back to `SetLabels` operator, the query would fail either way at the check of `CREATE_DELETE` of `L2` and `L3`. The exception would propagate through the code mentioned above and it will no longer pull the other 999,999 nodes since the query is stopped.

Therefore it is good to group it since we are in fact not saving any significant performance.
Either way, benchmark results should tell us whether we missed something or not

niko4299: Okay but lets say for example you  have read permission on 500k node and you dont have create delete permission on any labels. In node cursor pull method you will have to check 500k nodes before you find that one node you can see and then you will check do you even have permission for creating deleting label while now how I did it you wont even start pulling nodes.

Josipmrden (approved):

gitbuda (approved):

BorisTasevski (approved):

PR #532, opened 2022-09-05
[E118-MG < T0971-MG] Reference vertices by id in edge

antaljanosbenjamin:
[master < Epic] PR
- [ ] Check, and update documentation if necessary
- [ ] Update [changelog](https://docs.memgraph.com/memgraph/changelog)
- [ ] Write E2E tests
- [ ] Compare the [benchmarking results](https://bench-graph.memgraph.com/) between the master branch and the Epic branch
- [ ] Provide the full content or a guide for the final git message

[master < Task] PR
- [ ] Check, and update documentation if necessary
- [ ] Update [changelog](https://docs.memgraph.com/memgraph/changelog)
- [ ] Provide the full content or a guide for the final git message

PR #538, opened 2022-09-08
[E118 < T919] Implement message based actions

gvolfing:
[E118 < T919] PR

Create shard-side handlers for basic messages

Implement the handlers for CreateVertices, CreateEdges and ScanAll. Use
or modify the defined messages to interact with individual Shards and
test their behavior. Shard is currently being owned by ShardRsm
instances. The two top level dispatching functions Read() and Apply()
are responsible for read- and write operations respectively. Currently
there are a handful of messages that are defined but not utilized, these
will be used in the near future, as well as a couple of handler
functions with empty implementations.


src/storage/v3/shard_rsm.cpp
```diff
@@ -0,0 +1,443 @@
...
+    /// should have a predetermined primary label, so there is no point in
+    /// specifying it in the accessor functions. Their signature will
+    /// change.
+    /// TODO(gvolfing) Consider other methods than converting. Change either
+    /// the way that the property map is stored in the messages, or the
+    /// signature of CreateVertexAndValidate.
+    auto converted_property_map = ConvertPropertyMap(new_vertex.properties);
+
+    // TODO(gvolfing) make sure you don't create vectors bigger than they actually should be, e.g.
+    // std::vector<memgraph::storage::v3::LabelId> converted_label_ids(new_vertex.label_ids.size());
```
antaljanosbenjamin, line 223: Is this comment still relevant? I think `reserve` + `emplace_back` is better than creating an explicitly sized `vector` with default value and then overriding them.

src/storage/v3/shard_rsm.cpp
```diff
@@ -0,0 +1,443 @@
...
+}
+
+WriteResponses ShardRsm::ApplyWrite(DeleteVerticesRequest &&req) {
+  bool action_successful = true;
+  auto acc = shard_->Access();
+
+  for (auto &propval : req.primary_keys) {
+    if (!action_successful) {
+      break;
+    }
```
antaljanosbenjamin, line 277: Instead break you set `action_successful` to false.

src/storage/v3/shard_rsm.cpp
```diff
@@ -0,0 +1,443 @@
...
+    if (!vertex_from_acc || !vertex_to_acc) {
+      action_successful = false;
+      spdlog::debug(
+          &"Error while trying to insert edge, vertex does not exist. Transaction id: "[req.transaction_id.logical_id]);
+      continue;
+    }
+
+    auto edge_type_id = EdgeTypeId::FromUint(edge.type.id);
+    auto edge_acc = acc.CreateEdge(&vertex_from_acc.value(), &vertex_to_acc.value(), edge_type_id);
+
```
antaljanosbenjamin, line 355: I forgot to merge #532 and that is necessary here, as it might happen that only one of the vertices are present. I will let you know as soon as I merged the PR, so you can update your PR.

src/io/message_conversion.hpp
```diff
@@ -12,6 +12,7 @@
 #pragma once
 
 #include "io/transport.hpp"
+#include "spdlog/spdlog.h"
```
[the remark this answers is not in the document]
gvolfing: Actually this was some left-over from debugging, I will remove this.

src/storage/v3/shard_rsm.cpp
gvolfing: I will for the most cases, however, in the case of `ShardRsm::ApplyWrite(DeleteVerticesRequest &&req)` the intent is clearer this way, since the setting of `action_successful = false;` is happening inside a `switch` statement, where calling `break` would not effect the loop.

src/query/v2/requests.hpp
```diff
@@ -125,17 +335,23 @@ enum class StorageView { OLD = 0, NEW = 1 };
...
   std::optional<std::vector<std::string>> filter_expressions;
   std::optional<size_t> batch_limit;
   StorageView storage_view;
 };
 
+struct ScanResultRow {
+  Value vertex;
+  // empty() is no properties returned
+  std::map<PropertyId, Value> props;
+};
```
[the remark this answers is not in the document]
gvolfing: I think we should merge as it is, there are other messages that will probably have to change until we have everything in place. Guessing what should be the final form of may not worth the effort we put in because they will possibly change.

src/query/v2/requests.hpp
```diff
@@ -202,9 +418,24 @@ struct ExpandOneResponse {
...
 };
 
+// Update related messages
+struct UpdateVertexProp {
+  VertexId primary_key;
+  std::vector<std::pair<PropertyId, Value>> property_updates;
+};
+
+struct UpdateEdgeProp {
+  Edge edge;
```
jbajic, line 428: Why not `EdgeId`?

src/storage/v3/shard_rsm.cpp
```diff
@@ -0,0 +1,432 @@
...
+memgraph::storage::v3::PropertyValue ToPropertyValue(Value &&value) {
+  using PV = memgraph::storage::v3::PropertyValue;
+  PV ret;
+  switch (value.type) {
+    case Value::Type::NILL:
+      return PV();
+    case Value::Type::BOOL:
+      return PV(value.bool_v);
+    case Value::Type::INT64:
+      return PV(static_cast<int>(value.int_v));
```
jbajic, line 34:
```suggestion
      return PV(static_cast<int64_t>(value.int_v));
```

src/storage/v3/shard_rsm.cpp
```diff
@@ -0,0 +1,432 @@
...
+    ret.emplace(prop_key, ToValue(prop_val));
+  }
+
+  return ret;
+}
+
+Value ConstructValueVertex(const memgraph::storage::v3::VertexAccessor &acc, memgraph::storage::v3::View view) {
+  // Get the vertex id
+  auto prim_label = acc.PrimaryLabel(view).GetValue();
+  Label value_label = {.id = prim_label};
```
jbajic, line 175:
```suggestion
  Label value_label{.id = prim_label};
```

src/storage/v3/shard_rsm.cpp
```diff
@@ -0,0 +1,432 @@
...
+    if (!vertex_from_acc || !vertex_to_acc) {
+      action_successful = false;
+      spdlog::debug(
+          &"Error while trying to insert edge, vertex does not exist. Transaction id: "[req.transaction_id.logical_id]);
+      break;
+    }
+
+    auto edge_type_id = EdgeTypeId::FromUint(edge.type.id);
+    // auto edge_acc = acc.CreateEdge(&vertex_from_acc.value(), &vertex_to_acc.value(), edge_type_id,
+    // Gid::FromUint(edge.id.gid));
```
jbajic, line 339: Remove

src/query/v2/requests.hpp
```diff
@@ -75,11 +74,224 @@ struct Path {
...
+  }
+  explicit Value(std::map<std::string, Value> &&val) noexcept : type(Type::MAP) {
+    new (&map_v) std::map<std::string, Value>(std::move(val));
+  }
+
+  ~Value() { DestroyValue(); }
+
+  void DestroyValue() noexcept {
+    switch (type) {
+      case Type::NILL:
```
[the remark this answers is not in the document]
gvolfing: `NULL` is conflicting the C++ keyword. I will use `Null` instead (just like in the case of the `PropertyValue` types) and I will change all the types to CamelCase.

src/query/v2/requests.hpp
gvolfing: The corresponding handle, that actually interacts with the `Shard` is not yet implemented. It will probably make sense to use just the `EdgeId` instead, however I think that should be done in the PR that will contain the implementation of said handle. I don't think we should set the message's format into stone now because the exact way we might use them can change and we can only be sure once the `Shard`-side handle is in place.

src/query/v2/requests.hpp
```diff
@@ -16,52 +16,51 @@
...
-  VertexId id;
+  VertexId src;
+  VertexId dst;
   Gid gid;
 };
 
+struct Edge {
+  EdgeId id;
+  EdgeType type;
+};
```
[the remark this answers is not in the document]
gvolfing: Since the `Gid` should uniquely identify the `Edge`, I think to move it out from the struct makes perfect sense.

src/storage/v3/shard_rsm.hpp
```diff
@@ -0,0 +1,58 @@
...
+  WriteResponses ApplyWrite(UpdateVerticesRequest &&req);
+
+  WriteResponses ApplyWrite(CreateEdgesRequest &&req);
+  WriteResponses ApplyWrite(DeleteEdgesRequest &&req);
+  WriteResponses ApplyWrite(UpdateEdgesRequest &&req);
+
+ public:
+  explicit ShardRsm(std::unique_ptr<Shard> &&shard) : shard_(std::move(shard)){};
+
+  // NOLINTNEXTLINE(readability-convert-member-functions-to-static
```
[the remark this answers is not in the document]
gvolfing: This one is needed for clang-tidy to pass.

src/storage/v3/shard_rsm.cpp
```diff
@@ -0,0 +1,439 @@
...
+    case PropertyValue::Type::TemporalData: {
+      // TBD -> we need to specify this in the messages, not a priority.
+      MG_ASSERT(false, "Temporal datatypes are not yet implemented on Value!");
+      return Value{};
+    }
+  }
+}
+
+std::vector<std::pair<memgraph::storage::v3::PropertyId, memgraph::storage::v3::PropertyValue>> ConvertPropertyMap(
+    std::vector<std::pair<PropertyId, Value>> &properties) {
```
antaljanosbenjamin, line 102: I think the `properties` parameter should be also rvalue, to clearly express that the content of the vector will be moved.

jbajic (approved):

PR #546, opened 2022-09-11
[master < T271-MG] New C++ query module API

antepusic:
(Code transcluded from [MAGE #158](https://github.com/memgraph/mage/pull/158) - New C++ query module API)

**[master < Task] PR**
- [x] Add example module & converted tests


include/mage.hpp
```diff
@@ -0,0 +1,2780 @@
+// Copyright 2022 Memgraph Ltd.
```
gitbuda, line 1: @antepusic why is there `mage.hpp` and `mgp.hpp` shouldn't that be one file (probably the same applies to `mg_exceptions.hpp`)?

Josipmrden:
@vpavicic fixed and added just 2 sentences with NOTHING permission level

query_modules/example.cpp
```diff
@@ -0,0 +1,104 @@
...
+// included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
+// License, and you may not use this file except in compliance with the Business Source License.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#include <mg_exceptions.hpp>
+#include <mgp.hpp>
```
gitbuda, line 13: What about moving the exceptions into the `mgp.hpp` because when you include the header you include one thing? That's I would say standard for library code 😄

gitbuda: If `mgp.hpp` includes the exceptions file, this will work out of the box, not the best practice but that's I think an ok solution here (just to include exception inside mgp.hpp, if possible) 😄

Josipmrden: It actually works without. It's all in `mgp.hpp`

Josipmrden: I will delete the include

gitbuda (approved):

gitbuda (changes requested):
We have to add the release stuff, under `src/CMakeLists.txt` and `query_modules/CMakeLists.txt` to actually ship the API code to users!

gitbuda (approved):
All required files are packaged.

PR #548, opened 2022-09-13
[master < E157-MG] Add run_id as connection server name

brunos252:


src/communication/bolt/v1/session.hpp
```diff
@@ -99,6 +99,8 @@ class Session {
    * message. */
   virtual std::optional<std::string> GetServerNameForInit() = 0;
 
+  virtual std::optional<std::string> GetRunIdForInit() = 0;
```
gitbuda, line 102: I think this function should just be called `GetRunId`?

src/telemetry/telemetry.cpp
```diff
@@ -41,13 +42,17 @@ std::string GetMachineId() {
...
+Telemetry::Telemetry(std::string url, std::filesystem::path storage_directory, std::filesystem::path run_id_directory,
                      std::chrono::duration<int64_t> refresh_interval, const uint64_t send_every_n)
     : url_(std::move(url)),
       uuid_(utils::GenerateUUID()),
       machine_id_(GetMachineId()),
       send_every_n_(send_every_n),
       storage_(std::move(storage_directory)) {
+  memgraph::kvstore::KVStore run_id_storage(run_id_directory);
+  run_id_storage.Put("run_id", uuid_);
+  run_id_storage.~KVStore();
```
gitbuda, line 54: Aha, you used kvstore as a way to pass/communicate the `run_id`? That seems like a hack, can you pass that info in some other way because relying on disk for this type of thing doesn't make much sense to me.

gitbuda: E.g. putting that into `SessionData` object seems much better 😄

brunos252:
Do not merge this yet

brunos252:
Remember to squash and merge

gitbuda (approved):

PR #556, opened 2022-09-17
Add MacOS and Windows compilation

gitbuda:
The goal is to do all the necessary things to compile Memgraph on both Mac and Windows machines. Since there are a lot of common changes, I think it makes sense to do all the work within a single PR. The reason to compile Memgraph natively is to become independent of environments like Docker because it should be easier to install and use Memgraph.

Once the setup and compilation adjustments are made, packaging for different platforms should be done in additional PRs.

Related
- https://github.com/memgraph/memgraph/pull/870

TODOs:
- [ ] Take a look at [an equivalent example](https://github.com/facebook/rocksdb/blob/main/CMakeLists.txt#L494) 

Bigger Picture
File (7)


CMakeLists.txt
```diff
@@ -235,6 +235,12 @@ set(MG_ARCH "x86_64" CACHE STRING "Host architecture to build Memgraph on. Suppo
 
 # setup external dependencies -------------------------------------------------
 
+set(CMAKE_THREAD_LIBS_INIT "-lpthread")
```
gitbuda, line 238: Add the appropriate if

vpavicic:
We have a +1 on this feature from a user, commenting it would be very helpful :)

gitbuda:
@karrtikiyer do you mean `memgraph` or some of the client libraries like e.g. [pymgclient](https://github.com/memgraph/pymgclient)? I assume Memgraph, this will take a bit more time because there is actually a bunch of stuff to port on Mac 👀 

Not sure if this is a viable alternative, but it's possible to build Memgraph on Mac under [LimaVM](https://github.com/lima-vm/lima), instructions are located under:
* https://memgraph.notion.site/Quick-Start-82a99a85e62a4e3d89f6a9fb6d35626d
* https://memgraph.notion.site/MacOS-Lima-Compilation-Guide-eae1e9dcef5740579c5a41075b8f499b

gitbuda:
Calling bankruptcy here and closing this PR because:
* the changes just scratch the surface of all the things that have to be done
* it got in conflict with other changes
* it's not a priority at the moment

MironAtHome:
I can look into porting on Windows.
I don't have Mac :)

PR #563, opened 2022-09-20
[project-pineapples < T0919] Implement handles for missing messages

gvolfing:

Implement missing message handles

Implement the missing message handles with basic functionality. The
implementation does not include any capabilities to filter based on
expressions. In this PR All but the GetPropertiesRequest message
handles have been implemented.


src/storage/v3/shard_rsm.cpp
```diff
@@ -109,6 +111,299 @@ Value ConstructValueVertex(const memgraph::storage::v3::VertexAccessor &acc, mem
...
+      break;
+    }
+    case memgraph::msgs::EdgeDirection::BOTH: {
+      PropertyValueT ret;
+      ret.resize(ret_out.size() + ret_in.size());
+      ret.insert(ret.end(), ret_in.begin(), ret_in.end());
+      ret.insert(ret.end(), ret_out.begin(), ret_out.end());
+
+      properties_to_value = ret;
+      break;
```
antaljanosbenjamin, line 290: What about moving/forwarding? I can assume copying these datastructures might be expensive. Same true for other places in the new code.

src/query/v2/requests.hpp
```diff
@@ -57,24 +57,25 @@ struct EdgeId {
...
-  std::vector<Label> labels;
-  friend bool operator==(const Vertex &lhs, const Vertex &rhs) {
-    return (lhs.id == rhs.id) && (lhs.labels == rhs.labels);
-  }
-};
-
 struct Edge {
   VertexId src;
   VertexId dst;
+  std::optional<std::vector<std::pair<PropertyId, Value>>> properties;
```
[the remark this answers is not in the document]
gvolfing: Currently, we are inconsistent with this one.
At some places we are using `std::map<...>` and at others `std::vector<std::pair<...>>`. When the `Shard` class was reworked it took some parameters as `std::vector<std::pair<...>>` and so this was just how some fields were defined in the messages. This is not great of course, I have already talked with @kostasrim about that. Once the exact communication type is in place (or at least decided), we plan on reworking the messages.
Currently I would rather not spend time with that since the modification here would affect the `ShardRequestManager` as well.
If you insist of course I can change that in this PR.

src/query/v2/requests.hpp
```diff
@@ -458,21 +459,30 @@ struct ExpandOneResultRow {
...
-  Values edges;
+  std::optional<std::map<PropertyId, Value>> src_vertex_properties;
+
+  // NOTE: If the desired edges are specified in the request,
+  // edges_with_specific_properties will have a value and it will
+  // return the properties as a vector of property values. The order
+  // of the values returned should be the same as the PropertyIds
+  // were defined in the request.
+  std::optional<std::vector<std::tuple<VertexId, Gid, std::map<PropertyId, Value>>>> edges_with_all_properties;
+  std::optional<std::vector<std::tuple<VertexId, Gid, std::vector<Value>>>> edges_with_specific_properties;
```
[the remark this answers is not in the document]
gvolfing: I assume this was not a concern when the message type was defined. I am not sure how much gain in performance that would yield to us. This is a struct what is representing the request for one source vertex in the `ExpandOne` request, I think the reasoning behind the two possible forms of returning edge properties was because it is possible that one vertex would have a lot of connecting edges.

My opinion about changing the interface at this stage is the same as above,
but again, if you insist I will of course change this.

src/query/v2/requests.hpp
```diff
@@ -57,24 +57,25 @@ struct EdgeId {
...
   friend bool operator==(const Edge &lhs, const Edge &rhs) {
     return (lhs.src == rhs.src) && (lhs.dst == rhs.dst) && (lhs.type == rhs.type);
   }
 };
 
+struct Vertex {
+  VertexId id;
+  std::vector<Label> labels;
+  friend bool operator==(const Vertex &lhs, const Vertex &rhs) {
+    return (lhs.id == rhs.id) && (lhs.labels == rhs.labels);
```
jbajic, line 75: How is the order of labels ensured, since two vectors of the same labels can yield false here?

src/storage/v3/shard_rsm.cpp
```diff
@@ -109,6 +112,295 @@ Value ConstructValueVertex(const memgraph::storage::v3::VertexAccessor &acc, mem
...
+    auto edge_props = std::get<EdgePropertyDataStructure>(edge_props_var);
+    memgraph::msgs::Gid gid = edge.Gid().AsUint();
+
+    ret.emplace_back(EdgeProperties{other_vertex, gid, edge_props});
+  }
+
+  return ret;
+}
+
+template <typename PropertyValueT, typename PropertyNulloptT>
```
jbajic, line 270:
```suggestion
template <typename TPropertyValue, typename TPropertyNullopt>
```

src/query/v2/requests.hpp
kostasrim: I think this is a bug. `VertexId` should be `unique` since it's the primary label + primary key. I think for the comparison to work it sufficies to compare only the id's (`VertexId`)

gvolfing: I did change the comparison.

antaljanosbenjamin (approved):

jbajic (approved):
Left a minor comment!

PR #566, opened 2022-09-21
[master < T0945DX] Add Neo4j client and mixed workload to mgbench

antejavor:
In this PR, the following things has been done: 

- Added support for running Neoj4 (this introduces `--vendor-name` and change  from `--memgraph-binary` to `--vendor-binary`
- Flag for support when does authorization runs with 
- Support for and flag for warmup run
- Support and flag for mixed workload, run mixed queries from different categories. 
- Support and flag for tail latency 
- Support and flag for performance tracking (logs information about RES, through query/workload execution)
- `compare_results` support for mixed tail latency
- Dataset setup (index + delete) and import are separated. 
- Added `graph_bench.py` for executing multiple workloads on single vendor, sample:
```
graph_bench.py
--vendor memgraph /home/memgraph/binary
--dataset-group basic
--dataset-size small
--realistic 100 30 70 0 0
--realistic 100 50 50 0 0
--realistic 100 70 30 0 0
--realistic 100 30 40 10 20
--mixed 100 30 0 0 0 70
```


tests/mgbench/benchmark.py
```diff
@@ -201,17 +209,25 @@ def filter_benchmarks(generators, patterns):
...
         args.memgraph_binary,
         args.temporary_directory,
         not args.no_properties_on_edges,
         args.bolt_port,
     )
+    # vendor = runners.Neo4j(
+    #     args.memgraph_binary, 
+    #     args.temporary_directory, 
+    #     args.bolt_port, 
+    #     )
```
gitbuda, line 222: Put this under if or some other nicer concept, since you have 2 vendors, if seems quite ok from my perspective 😄

tests/mgbench/benchmark.py
```diff
@@ -201,17 +209,25 @@ def filter_benchmarks(generators, patterns):
...
+ 
     client = runners.Client(args.client_binary, args.temporary_directory, args.bolt_port)
-    memgraph.start_preparation()
+    vendor.start_preparation()
+    # client.execute(queries=[
+    #     ("DROP INDEX ON:User(id)", {}),
+    # ])
```
gitbuda, line 228: 0 commented code please 😄

tests/mgbench/benchmark.py
```diff
@@ -243,9 +259,9 @@ def filter_benchmarks(generators, patterns):
 
     # Run all benchmarks in all available groups.
 
-    for with_fine_grained_authorization in [False, True]:
+    for with_fine_grained_authorization in [False]:
```
gitbuda, line 262: This is very strange (from before), that hole block could/should be massively simplified

tests/mgbench/runners.py
```diff
@@ -122,6 +127,78 @@ def stop(self):
...
+        self._proc_neo4j = None
+        self._bolt_port = bolt_port
+        atexit.register(self._cleanup)
+
+    def __del__(self):
+        self._cleanup()
+        atexit.unregister(self._cleanup)
+
+    # def _get_args(self, **kwargs):
+    #     #kwargs["bolt_port"] = self._bolt_port
```
gitbuda, line 147: make the code work or remove the commented code completly

gitbuda (changes requested):
I've put a few comments +

@42jeremy it seems to me that the best place for queries that are different for different vendors is the `Dataset` class, the derived one because all the queries are in the same spot and you can see the different immediately, "centralization" makes total sense in this case?

tests/mgbench/benchmark.py
antejavor: I updated this and now changed is with arguments.

tests/mgbench/benchmark.py
antejavor: When setting trigger on neo4j, if it is there you get an exepction, I uncommented this, but I will updated dataset with this line.

tests/mgbench/benchmark.py
antejavor: Yes, I have an idea how to make it a bit different, I will change tomorrow/or next week.

tests/mgbench/datasets.py
```diff
@@ -291,4 +332,176 @@ def benchmark__match__vertex_on_property(self):
...
+    def benchmark__mixed__single_vertex_read_read(self):
+        return ("MATCH (n:User {id : $id}) RETURN n", {"id": self._get_random_vertex()})
+
+    def benchmark__mixed__single_vertex_write_write(self):
+        return (
+            "CREATE (n:UserTemp {id : $id}) RETURN n",
+            {"id": random.randint(1, self._num_vertices * 10)},
+        )
+
+    def benchmark__mixed__single_vertex_property_update_update(self):
```
gitbuda, line 489: two updates at the end, one is enough

tests/mgbench/datasets.py
```diff
@@ -291,4 +332,176 @@ def benchmark__match__vertex_on_property(self):
...
+    def benchmark__basic__pattern_short_analytical(self):
+        return (
+            "MATCH (n:User {id: $id})-[e]->(m) " "RETURN m LIMIT 1",
+            {"id": self._get_random_vertex()},
+        )
+
+    def benchmark__mixed__single_vertex_read_read(self):
+        return ("MATCH (n:User {id : $id}) RETURN n", {"id": self._get_random_vertex()})
+
+    def benchmark__mixed__single_vertex_write_write(self):
```
gitbuda, line 483: one write is fine as well 😄 (maybe there are more queries like that)

tests/mgbench/benchmark.py
```diff
@@ -144,45 +104,427 @@ def filter_benchmarks(generators, patterns):
...
+    config_distribution = "_".join(s)
+
+    print("Generating mixed workload")
+    write = list()
+    read = list()
+    update = list()
+    analytical = list()
+
+    for (
+        query,
```
gitbuda, line 270: query in not used in the loop, so use `_`

tests/mgbench/benchmark.py
```diff
@@ -144,45 +104,427 @@ def filter_benchmarks(generators, patterns):
...
+        )
+    s = [str(i) for i in args.mixed_workload]
+
+    config_distribution = "_".join(s)
+
+    print("Generating mixed workload")
+    write = list()
+    read = list()
+    update = list()
+    analytical = list()
```
gitbuda, line 267: `analytical, read, update, write = [], [], [], []`

tests/mgbench/benchmark.py
```diff
@@ -144,45 +104,427 @@ def filter_benchmarks(generators, patterns):
...
+
+    if (
+        len(write) == 0
+        and percentage_distribution[0] != 0
+        or len(read) == 0
+        and percentage_distribution[1] != 0
+        or len(update) == 0
+        and percentage_distribution[2] != 0
+        or len(analytical) == 0
+        and percentage_distribution[3] != 0
```
gitbuda, line 290: Ideas:
* I would write a function for a single check (`len(write) == 0 and distribution != 0`) and then use https://docs.python.org/3/library/functions.html#any
* This whole block of code (list init + for + this if) seems very repetitive and error-prone + since there are `ops`, `distributions`, maybe there is a nice/short abstraction for that? But that might be too much for this simple piece of logic here...

tests/mgbench/benchmark.py
```diff
@@ -144,45 +104,427 @@ def filter_benchmarks(generators, patterns):
...
+    ):
+        raise Exception(
+            "There is a missing query in group (write, read, update or analytical) for given workload distribution."
+        )
+
+    random.seed(config_distribution)
+
+    # Executing mixed workload for each test
+    if num_args == 6:
```
gitbuda, line 299: There has to be a better and more clean way 😄

tests/mgbench/benchmark.py
```diff
@@ -144,45 +104,427 @@ def filter_benchmarks(generators, patterns):
...
+        for query, funcname in queries[group]:
+            full_workload = []
+
+            log.info(
```
gitbuda, line 303: Aha, `log` is used here! Be consistent with print VS log 😄

tests/mgbench/benchmark.py
```diff
@@ -144,45 +104,427 @@ def filter_benchmarks(generators, patterns):
...
+            usage_workload = vendor.stop()
+
+            ret["database"] = usage_workload
+
+            results_key = [
+                dataset.NAME,
+                dataset.get_variant(),
+                group,
+                query + "_" + config_distribution,
+                WITHOUT_FINE_GRAINED_AUTHORIZATION,
```
gitbuda, line 370: The flag vs env var for authorization?

tests/mgbench/benchmark.py
```diff
@@ -196,22 +538,68 @@ def filter_benchmarks(generators, patterns):
...
+for dataset, queries in benchmarks:
+
+    # Save running configuration
+    workload_type = None
+    if len(args.mixed_workload) == 0:
+        workload_type = "Isolated"
+    elif len(args.mixed_workload) == 5:
+        workload_type = "Realistic"
+    else:
+        workload_type = "Mixed"
```
gitbuda, line 550: For the mixed_workload types, I think it would be better to use an enum with specific values, which then can be nicely named and checked in these ifs...

tests/mgbench/benchmark.py
```diff
@@ -362,9 +706,125 @@ def filter_benchmarks(generators, patterns):
                 log.success("Throughput: {:02f} QPS".format(ret["throughput"]))
 
                 # Save results.
-                results_key = [dataset.NAME, dataset.get_variant(), group, test, test_type]
+                results_key = [
+                    dataset.NAME,
+                    dataset.get_variant(),
+                    group,
+                    query,
+                    WITHOUT_FINE_GRAINED_AUTHORIZATION,
```
gitbuda, line 714: Flag?

tests/mgbench/benchmark.py
```diff
@@ -144,45 +104,427 @@ def filter_benchmarks(generators, patterns):
...
+            "num_workers": ret["num_workers"],
+            "memory": usage_workload["memory"],
+            "cpu": usage_workload["cpu"],
+        }
+        results_key = [
+            dataset.NAME,
+            dataset.get_variant(),
+            group,
+            config_distribution,
+            WITHOUT_FINE_GRAINED_AUTHORIZATION,
```
gitbuda, line 423: Flag?

tests/mgbench/benchmark.py
```diff
@@ -362,9 +706,125 @@ def filter_benchmarks(generators, patterns):
...
+                        "{}/{}/{}/{}".format(
+                            group, query, funcname, WITH_FINE_GRAINED_AUTHORIZATION
+                        ),
+                    )
+                    func = getattr(dataset, funcname)
+
+                    query_statistics = tail_latency(vendor, client, func)
+
+                    # Get number of queries to execute.
+                    # TODO: implement minimum number of queries, `max(10, num_workers)`
```
gitbuda, line 753: Feel free to implement 😄

tests/mgbench/benchmark.py
```diff
@@ -362,9 +706,125 @@ def filter_benchmarks(generators, patterns):
...
+            )
+            vendor.stop()
+
+            for group in sorted(queries.keys()):
+                for query, funcname in queries[group]:
+
+                    log.info(
+                        "Running query:",
+                        "{}/{}/{}/{}".format(
+                            group, query, funcname, WITH_FINE_GRAINED_AUTHORIZATION
```
gitbuda, line 745: Authorization flag

tests/mgbench/benchmark.py
```diff
@@ -144,45 +104,427 @@ def filter_benchmarks(generators, patterns):
...
+                elif t == "u":
+                    funcname = random.choices(update, k=1)[0]
+                    aditional_query = getattr(dataset, funcname)
+                    full_workload.append(aditional_query())
+                elif t == "a":
+                    funcname = random.choices(analytical, k=1)[0]
+                    aditional_query = getattr(dataset, funcname)
+                    full_workload.append(aditional_query())
+                elif t == "q":
+                    full_workload.append(base_query())
```
gitbuda, line 352: Also repetitive (and also the `if ... continue` block above), maybe there could be some sort of an abstract definition, e.g., the `Op` (read, write, ...) and `Workload` (a collection of queries), that would simplify/abstract all the logic about generating the queries. NOTE: if you go there, apply the change to all query generation code.

gitbuda (changes requested):
Overall not bad, a few things for sure can be improved, not all my comments are required except the ones that make the code more correct and consistent, but ofc let's comment one by one 😄

tests/mgbench/benchmark.py
```diff
@@ -144,45 +104,427 @@ def filter_benchmarks(generators, patterns):
...
+    Mixed workload can be run as a single configuration for all groups of queries,
+    Pass the positional arguments as values of what percentage of
+    write/read/update/analytical queries you want to have in your workload.
+    Running --mixed-workload 1000 20 70 10 0 will execute 1000 queries, 20% write, 
+    70% read, 10% update and 0% analytical. 
+
+    Mixed workload can also be run on each query under some defined load.
+    By passing one more positional argument, you are defining what percentage of that query
+    will be in mixed workload, and this is executed for each query. The rest of the queries will be 
+    selected from the apropriate groups
```
42jeremy, line 142:
```suggestion
    selected from the appropriate groups 
```

tests/mgbench/benchmark.py
```diff
@@ -196,22 +538,68 @@ def filter_benchmarks(generators, patterns):
...
-    memgraph = runners.Memgraph(
-        args.memgraph_binary,
-        args.temporary_directory,
-        not args.no_properties_on_edges,
-        args.bolt_port,
+    dataset.prepare(
+        cache.cache_directory("datasets", dataset.NAME, dataset.get_variant())
+    )
+
+    # TODO: Create some apstract class for vendors, that will hold this data
```
42jeremy, line 566:
```suggestion
    # TODO: Create some abstract class for vendors, that will hold this data
```

tests/mgbench/datasets.py
```diff
@@ -45,7 +45,9 @@ def __init__(self, variant=None):
             variant = self.DEFAULT_VARIANT
         if variant not in self.VARIANTS:
             raise ValueError("Invalid test variant!")
-        if (self.FILES and variant not in self.FILES) and (self.URLS and variant not in self.URLS):
+        if (self.FILES and variant not in self.FILES) and (
+            self.URLS and variant not in self.URLS
+        ):
```
42jeremy, line 50: I guess the formatting is automatic. I think we all need to use the same configuration for that, my automatic formatting seems to give another layout. I'll look at it

42jeremy:
I have reviewed and haven't found big issues. However I think I'm missing the big picture, perhaps we should jump on a call and discuss the big lines so I can perform a higher level review.

tests/mgbench/benchmark.py
```diff
@@ -20,63 +20,17 @@
...
+import statistics
 
 import datasets
 import log
 import helpers
 import runners
 
+
 WITH_FINE_GRAINED_AUTHORIZATION = "with_fine_grained_authorization"
 WITHOUT_FINE_GRAINED_AUTHORIZATION = "without_fine_grained_authorization"
```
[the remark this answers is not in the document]
antejavor: Regarding this flag, currently on the master running setup is not configurable you were running auth code and non auth code on each run.  Now, you can choose not to run auth code if you do not want. Probably running both should be default?  These strings `"with_fine_grained_authorization"` and `"without_fine_grained_authorization"` are needed because of the graphana and I need them in JSON structure. What I can change  to drop global vars or local vars is is to change a flag to something like this: 
```suggestion
parser.add_argument(
    "--authorization",
    choices=["with_fine_grained_authorization", "without_fine_grained_authorization"],
    default="without_fine_grained_authorization",
    help="Run each query with authorization",
)
```
But this means running just auth or not running auth, not both.

From my perspective, the real questions are:
1. Do we want to run auth as default? Now it pops to me that we should because of graphana. 
2.  Do we need tail latency on auth? Currently, we are not running it.  I think we probably should.

tests/mgbench/benchmark.py
```diff
@@ -144,28 +101,334 @@ def filter_benchmarks(generators, patterns):
...
+
+parser.add_argument(
+    "--mixed-workload",
+    nargs=5,
+    type=int,
+    default=[0, 0, 0, 0, 0],
+    help="Run combination that defines mixed workload;"
+    "Pass the positional arguments as values of what percentage of"
+    "write/read/update/analytical queries you want;"
+    "Running --mixed-workload 1000 20 70 10 0"
```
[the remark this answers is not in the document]
antejavor: The only change here is that you do not need to specify the query, because it will be automatically executed on every query from group. 
```
./benchmark.py --mixed-workload=[1000, 10, 10, 30, 50, 0] pokec/small/group/**query** 

./benchmark.py --mixed-workload=[1000, 30, 0, 0, 0, 70] pokec/small/group/**query**
```

tests/mgbench/benchmark.py
```diff
@@ -144,45 +104,427 @@ def filter_benchmarks(generators, patterns):
...
+    Running --mixed-workload 1000 30 0 0 0 70, will execute each query 700 times or 70%,
+    with the presence of 300 write queries from write type or 30%""",
+)
+
 args = parser.parse_args()
 
+
+def get_queries(gen, count):
+    # Make the generator deterministic.
+    random.seed(gen.__name__)
```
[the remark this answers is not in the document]
antejavor: I agree with you on this one 100%,  I was thinking about this at some point, but I prioritize other stuff for getting the results.  After merging this, I will open a separate PR for the logger, so we can log some extra stuff for the things you mention.

tests/mgbench/benchmark.py
```diff
@@ -144,45 +104,427 @@ def filter_benchmarks(generators, patterns):
...
+        num_workers=1,
+    )
+
+
+def tail_latency(vendor, client, func):
+    vendor.start_benchmark()
+    if args.warmup_run:
+        warmup(client)
+    latency = []
+    iteration = 100
```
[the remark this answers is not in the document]
antejavor: Will do in the next commit.

tests/mgbench/benchmark.py
antejavor: It is not the log you think it is. Take a look at what is log here: https://github.com/memgraph/memgraph/blob/master/tests/mgbench/log.py#L21
I will implement the right one in next PR.

tests/mgbench/benchmark.py
```diff
@@ -144,45 +104,427 @@ def filter_benchmarks(generators, patterns):
...
+    return query_stats
+
+
+def mixed_workload(vendor, client, dataset, group, queries):
+    num_args = len(args.mixed_workload)
+    if num_args > 6:
+        raise Exception(
+            "You have passed to many arguments for mixed workload",
+            "read description of --mixed-worload args",
+        )
```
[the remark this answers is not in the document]
antejavor: Ok, I fixed this by introducing simple Workload class, that now handles this info about number of arguments: https://github.com/memgraph/memgraph/pull/566/commits/d075b31040b6f67fa1d248f664b48899bed1baeb#diff-7bdbec771008d0634e8a5ecf53585bd3182efe304cf89a2809c1172d78ff9237R158

This defines the type of workload by name, so it is more readable.

tests/mgbench/benchmark.py
antejavor: Regarding the cleaning of repetitive code, I cleaned it up by using dictionaries.

tests/mgbench/benchmark.py
antejavor: I used the Workload class for that now.

tests/mgbench/benchmark.py
antejavor: I am using a Class for this since Enum is limited in construction logic. Regarding the naming, on the graph bench we are using names: 

- Isolated  - all pure queries 
- Mixed - all queries + some percentage of  most likely write 
- Realistic - groups of queries in different combinations write 30% read 30% update 10% etc. 

I am pointing this out, bcs we have single -mixed-workload flag for two similar but different workloads. We are calling different names on the graph bench platform.  Maybe we can separate this, but obviously not super important.

tests/mgbench/benchmark.py
antejavor: https://github.com/memgraph/memgraph/pull/566#discussion_r1016135559

tests/mgbench/benchmark.py
antejavor: https://github.com/memgraph/memgraph/pull/566#discussion_r1016135559

tests/mgbench/benchmark.py
antejavor: https://github.com/memgraph/memgraph/pull/566#discussion_r1016135559

tests/mgbench/benchmark.py
antejavor: This was rewritten, it is not repetitive anymore.

tests/mgbench/benchmark.py
antejavor: I need to balance this out bcs of that long-running expansion test, I will do this on the fly, today/tomorrow.

tests/mgbench/datasets.py
antejavor: I synced with @jbajic about this. Thanks for the tip.

antejavor:
@42jeremy Regarding your comment 😀, we can hop on a call tomorrow morning to sync about these changes.

tests/mgbench/benchmark.py
```diff
@@ -145,9 +105,355 @@ def filter_benchmarks(generators, patterns):
...
+    write/read/update/analytical queries you want to have in your workload.
+    Running --mixed-workload 1000 20 70 10 0 will execute 1000 queries, 20% write,
+    70% read, 10% update and 0% analytical.
+
+    Mixed workload can also be run on each query under some defined load.
+    By passing one more positional argument, you are defining what percentage of that query
+    will be in mixed workload, and this is executed for each query. The rest of the queries will be
+    selected from the appropriate groups
+    Running --mixed-workload 1000 30 0 0 0 70, will execute each query 700 times or 70%,
+    with the presence of 300 write queries from write type or 30%""",
```
42jeremy, line 142: If I understand correctly the first sentence, how the mixed-workload is interpreted depends on the other arguments (= single configuration VS defined load). If I am correct, would it make sense to provide a command/example for each? (feel free to disagree and resolve this comment yourself)

I am providing the following as an example, please correct and adapt if it does not make sense:

`./benchmark dataset/small --mixed-workload 1000 20 70 10 0:` run on dataset `small` from folder `dataset`, total of 1000 queries 20% write etc..
`./benchmark dataset/small/read_specific --mixed-workload 1000 25 0 20 10 45`:  run on dataset `small` from folder `dataset`, total of 1000 queries, 45% will be from read_specific, 25% write, 20% update, 10% analytical

tests/mgbench/datasets.py
```diff
@@ -103,8 +127,9 @@ class Pokec(Dataset):
     VARIANTS = ["small", "medium", "large"]
     DEFAULT_VARIANT = "small"
     FILES = None
+
     URLS = {
-        "small": "https://s3-eu-west-1.amazonaws.com/deps.memgraph.io/pokec_small.setup.cypher",
+        "small": "https://s3.eu-west-1.amazonaws.com/deps.memgraph.io/dataset/pokec/benchmark/pokec_small_import.cypher",
         "medium": "https://s3-eu-west-1.amazonaws.com/deps.memgraph.io/pokec_medium.setup.cypher",
         "large": "https://s3-eu-west-1.amazonaws.com/deps.memgraph.io/pokec_large.setup.cypher.gz",
```
42jeremy, line 134: Adresses of medium and large should change as well

42jeremy: https://s3.eu-west-1.amazonaws.com/deps.memgraph.io/dataset/pokec/pokec_small.setup.cypher
https://s3.eu-west-1.amazonaws.com/deps.memgraph.io/dataset/pokec/pokec_medium.setup.cypher
https://s3.eu-west-1.amazonaws.com/deps.memgraph.io/dataset/pokec/pokec_large.setup.cypher.gz

(from: https://github.com/memgraph/memgraph/blob/5c0e41ed44c283d172ecb9106bbc5809e8cb2906/tests/mgbench/datasets.py#L106-L110 )

42jeremy: (maybe that's something we'll tackle when merging your PR to project-pineapple, so you can ignore this comment)

tests/mgbench/benchmark.py
antejavor: I will add a bit more explanation in the next commit, and this  PR will have to follow up polish PR I will make this thing more straightforward.

antejavor:
Things that can be added to follow-up PR: 

- Polish mixed workload flags
- Add detailed logging
- Add more detailed resource tracking
- Polish `benchmark.py` script structure
- More vendors (future)
- Polish C++ wrapper (future)

tests/mgbench/benchmark.py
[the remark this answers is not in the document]
gitbuda: Both have to run on CI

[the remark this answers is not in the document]
antejavor: This is fixed now, both non-auth and auth options are running on CI.

gitbuda (approved):

PR #567, opened 2022-09-21
[master <] Add libcxx to the toolchain

gitbuda:
NOTE: Each library in the toolchain also has to be compiled with the `libc++`.

TODOs
- [x] Do we have 2 different toolchains or 2 versions of each library in the toolchain? Multiple toolchains seem much more compelling because in that case, Memgraph CMake config doesn't have to change. Maybe it makes sense to isolate libraries in a separate archive, but for now we just compile all the libraries with `libc++` if that's the configuration
- [x] Implement the stdlib switch
- [x] Make switches between architecture, OS and stdlib consistent (seems like it doesn't make much sense)
- [x] Add downloaded packages verification

The follow-up PR should be `toolchain-v5` with upgraded versions of GCC, Clang, and the latest libs versions.
* On Fedora 36, Python is making issues because `dnf` is using Python 10, which uses a not suitable glibc version (NOTE: I think Python is not a problem, the problem is too new system glibc version).

Screen Shot 2022-10-15 at 8 06 10 PM


jbajic (commented):
Why is this in v4, maybe move to v5 if we will create a new toolchain

gitbuda:
I wanted to put structure in-place (add libc++) and then after upgrade all libs in another PR just to make is smaller in size because these are well separated issues

jbajic (approved):

PR #578, opened 2022-09-30
[:pineapple: < T0918-MG] Implement filtering for scanall

gvolfing:

Implement filtering capabilites for scanall 

ScanVerticesrequest was not able to utilize filtering capabilites
before. With these modification it is now able to filter the scanned
verticies based on the filter_expressions field in the
ScanVerticesRequest message type.


src/storage/v3/shard_rsm.cpp
```diff
@@ -112,6 +124,248 @@ Value ConstructValueVertex(const memgraph::storage::v3::VertexAccessor &acc, mem
...
+  return visitor.visit(ast);
+}
+
+memgraph::expr::TypedValue ComputeExpression(memgraph::expr::DbAccessor &dba,
+                                             const std::optional<memgraph::storage::v3::VertexAccessor> &v_acc,
+                                             const std::optional<memgraph::storage::v3::EdgeAccessor> &e_acc,
+                                             const std::string &expression, std::string_view node_name,
+                                             std::string_view edge_name) {
+  memgraph::expr::AstStorage storage;
+  memgraph::expr::Frame<memgraph::expr::TypedValue> frame{static_cast<int64_t>(128)};
```
gitbuda, line 308: 128 here is extremely suspicious, shouldn't this be determined based on the actual expression?

gvolfing: @gitbuda 
Yes, the `128` is a leftover I forgot to remove.
We know 2 things at this stage, at no point, should the size of the `Frame` in this context be bigger than 2. At most it will be one `Vertex` and one `Edge` on the Frame. The way the signature is written, both the `VertexAccessor` and the `EdgeAccessor` is optional, based on that we can determine if the size of the frame should be 0, 1 or 2. I'm not sure if 0 is a common use-case though.
Based on that I can set the size of the frame, or just hard-code it to 2. (I'm not sure if the memory consumption would be significant).
What do you think?

src/storage/v3/shard_rsm.cpp
```diff
@@ -112,6 +124,248 @@ Value ConstructValueVertex(const memgraph::storage::v3::VertexAccessor &acc, mem
...
+  }
+  if (e_acc) {
+    edge_props = GetPropertiesFromAcessor(e_acc->Properties(View::OLD).GetValue());
+  }
+
+  std::vector<PropertyId> all_properties(vertex_props.size() + edge_props.size());
+
+  // Sorting might not needed, since the props are being converted from std::map
+  std::sort(vertex_props.begin(), vertex_props.end());
+  std::sort(edge_props.begin(), edge_props.end());
```
gvolfing, line 281: I am not sure about this one, please discuss.

src/storage/v3/shard_rsm.cpp
```diff
@@ -112,6 +124,248 @@ Value ConstructValueVertex(const memgraph::storage::v3::VertexAccessor &acc, mem
...
+  auto edge_identifier = memgraph::expr::Identifier(std::string(edge_name), false);
+  bool is_edge_identifier_present = false;
+
+  std::vector<memgraph::expr::Identifier *> identifiers;
+
+  if (v_acc && expression.find(node_name) != std::string::npos) {
+    is_node_identifier_present = true;
+    identifiers.push_back(&node_identifier);
+  }
+  if (e_acc && expression.find(node_name) != std::string::npos) {
```
42jeremy, line 327: Isn't it edge_name instead of node_name?

gvolfing: Yes, that should be edge_name. Good catch.

jbajic (approved):

src/storage/v3/shard_rsm.cpp
```diff
@@ -112,6 +124,248 @@ Value ConstructValueVertex(const memgraph::storage::v3::VertexAccessor &acc, mem
...
+
+  std::vector<PropertyId> all_properties(vertex_props.size() + edge_props.size());
+
+  // Sorting might not needed, since the props are being converted from std::map
+  std::sort(vertex_props.begin(), vertex_props.end());
+  std::sort(edge_props.begin(), edge_props.end());
+
+  auto it = std::set_symmetric_difference(vertex_props.begin(), vertex_props.end(), edge_props.begin(),
+                                          edge_props.end(), all_properties.begin());
+  all_properties.resize(it - all_properties.begin());
```
42jeremy, line 285: I don't quite understand, why are we doing this?

src/storage/v3/shard_rsm.cpp
```diff
@@ -112,6 +124,248 @@ Value ConstructValueVertex(const memgraph::storage::v3::VertexAccessor &acc, mem
...
+    vertex_props = GetPropertiesFromAcessor(v_acc->Properties(View::OLD).GetValue());
+  }
+  if (e_acc) {
+    edge_props = GetPropertiesFromAcessor(e_acc->Properties(View::OLD).GetValue());
+  }
+
+  std::vector<PropertyId> all_properties(vertex_props.size() + edge_props.size());
+
+  // Sorting might not needed, since the props are being converted from std::map
+  std::sort(vertex_props.begin(), vertex_props.end());
```
42jeremy, line 280: General question: do we need to specify a sorting lambda or the default is good enough?

src/storage/v3/shard_rsm.cpp
```diff
@@ -112,6 +124,248 @@ Value ConstructValueVertex(const memgraph::storage::v3::VertexAccessor &acc, mem
...
+  std::sort(edge_props.begin(), edge_props.end());
+
+  auto it = std::set_symmetric_difference(vertex_props.begin(), vertex_props.end(), edge_props.begin(),
+                                          edge_props.end(), all_properties.begin());
```
42jeremy, line 284: Here all_properties will be empty, see https://godbolt.org/z/P9GfbdcPo
`std::back_inserter(all_properties)` instead `of all_properties.begin()`

42jeremy (dismissed):
There is one blocking issue, rest is to be discussed

src/storage/v3/shard_rsm.cpp
gvolfing: Since they should be already in a sorted order at that point I will remove the explicit sorting.

src/storage/v3/shard_rsm.cpp
gvolfing: Yes, a lambda should have been there, I was not sure if the sorting was needed there at all or not, but I will just remove the sorting from there, since the elements should in a sorted order at that point anyways.

src/storage/v3/shard_rsm.cpp
gvolfing: Insteded of `reserve(SIZE)` I should have called the constructor with `SIZE`. But that part will change as well.

src/storage/v3/shard_rsm.cpp
```diff
@@ -112,6 +124,247 @@ Value ConstructValueVertex(const memgraph::storage::v3::VertexAccessor &acc, mem
...
+    labels.push_back(dba.NameToLabel(name));
+  }
+  return labels;
+}
+
+template <class TExpression>
+auto Eval(TExpression *expr, memgraph::expr::EvaluationContext ctx, memgraph::expr::AstStorage &storage,
+          memgraph::storage::v3::ExpressionEvaluator &eval, memgraph::expr::DbAccessor &dba) {
+  ctx.properties = NamesToProperties(storage.properties_, dba);
+  ctx.labels = NamesToLabels(storage.labels_, dba);
```
antaljanosbenjamin, line 237: Also the edge types are needed to be converted.

src/storage/v3/shard_rsm.cpp
gvolfing: This should have been std::set_union.

src/storage/v3/shard_rsm.cpp
antaljanosbenjamin: This is done in the [planner](https://github.com/memgraph/memgraph/blob/5347c06d7614dc13311819aa7e8e2c2c7a66a7ed/src/query/v2/plan/rule_based_planner.hpp#L322).

src/storage/v3/shard_rsm.cpp
```diff
@@ -101,31 +117,226 @@ std::optional<std::map<PropertyId, Value>> CollectAllPropertiesFromAccessor(
...
+
+  expr::SymbolGenerator symbol_generator(&symbol_table, identifiers);
+  (std::any_cast<Expression *>(expr))->Accept(symbol_generator);
+
+  if (is_node_identifier_present) {
+    frame[symbol_table.at(node_identifier)] = *v_acc;
+  }
+  if (is_edge_identifier_present) {
+    frame[symbol_table.at(edge_identifier)] = *e_acc;
+  }
```
antaljanosbenjamin, line 313: The position of these identifiers will be `-1` as nothing will fill their position correctly. I for now I think it is okay to search for these identifiers in the `SymbolTable::scopes_[0]::symbols` map as in my opinion we won't have expression that have multiple scopes, because only `FOR EACH` can introduce another scope.

src/storage/v3/shard_rsm.cpp
```diff
@@ -101,31 +118,239 @@ std::optional<std::map<PropertyId, Value>> CollectAllPropertiesFromAccessor(
...
+                                          });
+    MG_ASSERT(position_on_frame != symbol_table.table().end());
+    frame[symbol_table.at(node_identifier)] = *v_acc;
+  }
+  if (is_edge_identifier_present) {
+    // We look for the position of node_identifier in the frame.
+    auto position_on_frame = std::find_if(symbol_table.table().begin(), symbol_table.table().end(),
+                                          [&edge_identifier](const std::pair<int32_t, Symbol> &position_symbol_pair) {
+                                            return position_symbol_pair.second.name() == edge_identifier.name_;
+                                          });
```
antaljanosbenjamin, line 324: As you found a proper way to identify whether the edge/node identifier is present, maybe you can eliminate the checks based on strings (line 298 and 302) to make the logic really good. I think you have all the pieces here they just need a bit of restructuring.

src/storage/v3/shard_rsm.cpp
```diff
@@ -101,31 +118,239 @@ std::optional<std::map<PropertyId, Value>> CollectAllPropertiesFromAccessor(
...
+  if (v_acc && expression.find(node_name) != std::string::npos) {
+    is_node_identifier_present = true;
+    identifiers.push_back(&node_identifier);
+  }
+  if (e_acc && expression.find(edge_name) != std::string::npos) {
+    is_edge_identifier_present = true;
+    identifiers.push_back(&edge_identifier);
+  }
+
+  expr::SymbolGenerator symbol_generator(&symbol_table, identifiers);
```
antaljanosbenjamin, line 307:
```suggestion
  expr::SymbolGenerator symbol_generator(&symbol_table, identifiers);
```

The symbol generator receives the "predefined identifiers", which means they don't have to be matched in the cypher query, it is meant to used only with triggers, so I think they are not needed here, furthermore they are faulty, because those identifiers can only be used once in the expression, so `n.age + n.height` might not work.

src/storage/v3/shard_rsm.cpp
```diff
@@ -289,41 +289,30 @@ TypedValue ComputeExpression(DbAccessor &dba, const std::optional<memgraph::stor
...
+                   });
+  auto is_edge_identifier_present = edge_identifier_position_on_frame != symbol_table.table().end();
   if (is_edge_identifier_present) {
-    // We look for the position of node_identifier in the frame.
-    auto position_on_frame = std::find_if(symbol_table.table().begin(), symbol_table.table().end(),
-                                          [&edge_identifier](const std::pair<int32_t, Symbol> &position_symbol_pair) {
-                                            return position_symbol_pair.second.name() == edge_identifier.name_;
-                                          });
-    MG_ASSERT(position_on_frame != symbol_table.table().end());
     frame[symbol_table.at(edge_identifier)] = *e_acc;
```
antaljanosbenjamin, line 316: one more thing regarding to this: https://github.com/memgraph/memgraph/pull/578#discussion_r996815707

So you found the proper position/symbol for edge/node, but you are still using the "dummy" identifiers, so their position still will be wrong (-1). So I think edge_identifier / node_identifier should be gone, and edge_identifier_position_on_frame->second / node_identifier_position_on_frame->second should be used instead.

42jeremy: Oh indeed that's completely correct. Well caught!

antaljanosbenjamin (approved):

src/query/v2/bindings/bindings.hpp
```diff
@@ -11,5 +11,9 @@
 
 #pragma once
 
+#ifdef MG_AST_INCLUDE_PATH
+#error You are probably trying to include files from expr from both the storage and query engines! You will have a rought time kid!
```
gvolfing, line 15: Typo

src/storage/v3/bindings/bindings.hpp
```diff
@@ -11,5 +11,9 @@
 
 #pragma once
 
+#ifdef MG_AST_INCLUDE_PATH
+#error You are probably trying to include some files of mg-expr from both the storage and query engines! You will have a rought time kid!
```
gvolfing, line 15: rought time same typo as above

gitbuda (approved):

PR #580, opened 2022-10-03
[:pineapple: < MG-T1071] Implement create expand request in query

jbajic:
[master < Task] PR

__Commit Message__:
- Create ExpandOne request
- Refactor `CreateEdge` into `CreateExpand`


src/query/v2/plan/operator.cpp
```diff
@@ -414,6 +419,8 @@ ScanAll::ScanAll(const std::shared_ptr<LogicalOperator> &input, Symbol output_sy
 
 ACCEPT_WITH_INPUT(ScanAll)
 
+class DistributedScanAllCursor;
```
jbajic, line 422: Remove this

src/glue/v2/communication.cpp
```diff
@@ -204,9 +204,45 @@ Value ToBoltValue(msgs::Value value) {
     case msgs::Value::Type::Vertex:
     case msgs::Value::Type::Edge:
     case msgs::Value::Type::Path: {
-      throw utils::BasicException("Path, Vertex and Edge not supported!");
+      throw utils::BasicException("Path not supported!");
```
gvolfing, line 207: `msgs::Value::Type::Edge` and `msgs::Value::Type::Vertex:` will fallthrough here, but we only throw that "Path not supported!"

src/glue/v2/communication.cpp
```diff
@@ -204,9 +204,45 @@ Value ToBoltValue(msgs::Value value) {
...
+      for (const auto &kv : value.map_v) {
+        auto maybe_value = ToBoltValue(kv.second);
+        map.emplace(kv.first, std::move(maybe_value));
+      }
+      return Value{std::move(map)};
+    }
+    case msgs::Value::Type::Vertex:
+    case msgs::Value::Type::Edge:
+    case msgs::Value::Type::Path: {
+      throw utils::BasicException("Path not supported!");
```
gvolfing, line 244: Same as above.

gvolfing (approved):

src/query/v2/shard_request_manager.hpp
```diff
@@ -281,6 +283,42 @@ class ShardRequestManager : public ShardRequestManagerInterface {
     return responses;
   }
 
+  std::vector<CreateExpandResponse> Request(ExecutionState<CreateExpandRequest> &state,
+                                            std::vector<NewExpand> new_edges) override {
+    MG_ASSERT(!new_edges.empty());
+    MaybeInitializeExecutionState(state, new_edges);
+    std::vector<CreateExpandResponse> responses;
+    auto &shard_cache_ref = state.shard_cache;
+    size_t id{0};
```
42jeremy, line 292:
```suggestion
    int64_t id{0};
```

42jeremy (approved):

src/query/v2/shard_request_manager.hpp
antaljanosbenjamin: Why should this be `int64_t`? It is only used to index into arrays and increment it. For that a `size_t` is more appropriate in this case (yes, having the size type as unsigned in a big mistake in C++, but in this case a size_t conforms the API more than an `int64_t`).

src/coordinator/shard_map.hpp
```diff
@@ -115,7 +118,12 @@ struct ShardMap {
     bool mutated = false;
 
     for (auto &[label_id, label_space] : label_spaces) {
-      for (auto &[low_key, shard] : label_space.shards) {
+      for (auto it = label_space.shards.begin(); it != label_space.shards.end(); it++) {
+        auto &[low_key, shard] = *it;
+        std::optional<PrimaryKey> high_key;
+        if (const auto next_it = std::next(it); next_it != label_space.shards.end()) {
+          high_key = next_it->first;
+        }
```
antaljanosbenjamin, line 126: This is important, because without the max key the shards cannot determine properly whether a vertex should be stored on them or not. This functionality is necessary when creating edges.

src/query/v2/shard_request_manager.hpp
```diff
@@ -54,18 +54,20 @@ class RsmStorageClientManager {
...
-  TStorageClient &GetClient(const LabelId label_id, const Shard &key) { return cli_cache_[label_id].find(key)->second; }
+  TStorageClient &GetClient(const Shard &key) {
+    auto it = cli_cache_.find(key);
+    MG_ASSERT(it != cli_cache_.end(), "Non-existing shard client");
+    return it->second;
+  }
 
  private:
-  std::map<LabelId, std::map<Shard, TStorageClient>> cli_cache_;
+  std::map<Shard, TStorageClient> cli_cache_;
```
antaljanosbenjamin, line 70: I removed the top level indirection, because otherwise we always have to no which label belongs to a specific shard, but the `Shard` only knows about the addresses. As we store `Shard`s in the `Execution` state, I think this is a practical simplification, because we don't have to store the primary label in some other datastructure.

42jeremy (approved):

gvolfing (approved):

PR #581, opened 2022-10-05
[:pineapple: < T1074-MG] Implement filtering capabilities for expandone

gvolfing:
Add filtering capabilities to ExpandOne Request

                                                 ^
                                                 |
                                                 |
                                       I don't think there is mouch to elaborate on in this PR message...


gvolfing:
This branch was branched from https://github.com/memgraph/memgraph/pull/578
So there is a bunch of stuff here that was introduced in that PR, since that PR is not merged yet the changes show up here as well.

src/query/v2/requests.hpp
```diff
@@ -420,12 +420,13 @@ struct ExpandOneRequest {
...
-  //  List of expressions evaluated on edges
-  std::vector<Expression> expressions;
+
+  std::optional<std::vector<std::string>> vertex_expressions;
+  std::optional<std::vector<std::string>> edge_expressions;
+
   std::optional<std::vector<OrderBy>> order_by;
   std::optional<size_t> limit;
-  std::optional<Filter> filter;
+  std::optional<std::vector<std::string>> filters;
```
42jeremy, line 429: I've seen similar remarks so I'll give it as well: if there isn't difference between nullopt and an empty vector, we can just stick with the empty vector

42jeremy (approved):

src/storage/v3/shard_rsm.cpp
```diff
@@ -934,12 +935,115 @@ msgs::ReadResponses ShardRsm::HandleRead(msgs::ScanVerticesRequest &&req) {
...
+  }
+
   auto acc = shard_->Access(req.transaction_id);
   bool action_successful = true;
 
   std::vector<msgs::ExpandOneResultRow> results;
 
   for (auto &src_vertex : req.src_vertices) {
+    // Get Vertex acc
+    auto src_vertex_acc_opt = acc.FindVertex(ConvertPropertyVector((src_vertex.second)), View::OLD);
```
antaljanosbenjamin, line 1031:
```suggestion
    auto src_vertex_acc_opt = acc.FindVertex(ConvertPropertyVector((src_vertex.second)), View::NEW);
```

In general the request should contain the view (not necessary to add it right now), but let's go with the `NEW` view, that eliminates one possible failure during debugging.

antaljanosbenjamin, line 1031:
```suggestion
    auto src_vertex_acc_opt = acc.FindVertex(ConvertPropertyVector((src_vertex.second)), View::NEW);
```

In general the request should contain the view (not necessary to add it right now), but let's go with the `NEW` view, that eliminates one possible failure during debugging.

src/storage/v3/shard_rsm.cpp
```diff
@@ -934,12 +935,115 @@ msgs::ReadResponses ShardRsm::HandleRead(msgs::ScanVerticesRequest &&req) {
 }
 
 msgs::ReadResponses ShardRsm::HandleRead(msgs::ExpandOneRequest &&req) {
+  using EdgeProperties =
```
antaljanosbenjamin, line 938: I think neither `get_edge_properties` nor `maybe_filter_based_on_edge_uniquness` is used anywhere. In the expand one PR I changed how we filld the edges. Instead of using variants, we use a "filler function" (check `GetExpandOneResult` for details), which knows how to add an edge to the result. I found this way a bit more convinient, because it is basically the strategy pattern: the filler function is the strategy that knows how the response should be populated based on the request.

tests/simulation/shard_rsm.cpp
```diff
@@ -463,18 +463,18 @@ void AttemptToExpandOneWithWrongEdgeType(ShardClient &client, uint64_t src_verte
   // Edge properties to look for
   std::optional<std::vector<PropertyId>> edge_properties = {};
 
-  std::vector<msgs::Expression> expressions;
+  std::vector<std::string> expressions;
```
antaljanosbenjamin, line 466: Could you please add/extend a test where you use filters/expressions? Just to make sure they are working and nobody broke them in the future.

antaljanosbenjamin (changes requested):

src/storage/v3/shard_rsm.cpp
```diff
@@ -934,12 +935,115 @@ msgs::ReadResponses ShardRsm::HandleRead(msgs::ScanVerticesRequest &&req) {
...
+            auto [it, insertion_happened] = other_vertex_set.insert(&edge_acc.ToVertex());
+            return insertion_happened;
+          };
+          break;
+        }
+        case memgraph::msgs::EdgeDirection::IN: {
+          check_and_insert_unique_other_vertex = [](std::unordered_set<const storage::v3::VertexId *> &other_vertex_set,
+                                                    const memgraph::storage::v3::EdgeAccessor &edge_acc) {
+            auto [it, insertion_happened] = other_vertex_set.insert(&edge_acc.FromVertex());
+            return insertion_happened;
```
antaljanosbenjamin, line 999: One more thing: the set is based on pointer without a custom comparator, that means it will only differentiate between the vertex ids based on there address, which is different in each edge, even for the same vertices. So I think either a custom comparator or using value semantics is necessary.

Maybe also think about eliminating these functions (`std::function` is much more expensive than a lambda) by getting the vertex id conditionally:
```

const auto vertex_selector = [edge_direction](cosnt EdgeAccessor& edge_acc) {
  switch (edge_direction): {
    case msgs::EdgeDirection::IN:
      return edge_acc.FromVertex();
    case msgs::EdgeDirection::OUT:
      return edge_acc.ToVertex();
    case msgs::EdgeDirection::BOTH:
      LOG_FATAL("This is should never happen, memgraph::msgs::EdgeDirection::BOTH should not be passed here.");
  }
}
```

And then you can use it to get the correct vertex id. In my opinion this solution would have two advantages:
1. It is not using a `std::function` (`check_and_insert_unique_other_vertex`), it will be faster
2. It makes the logic a bit simpler in my opinion.

gvolfing: This part of the code was removed with `maybe_filter_based_on_edge_uniquness`.

src/storage/v3/shard_rsm.cpp
antaljanosbenjamin: You are :100:% right! Feel free to move the initialization of `edge_filler` into `HandleRead` and pass it as a parameter to `GetExpandOneResult`!

src/storage/v3/shard_rsm.cpp
```diff
@@ -51,6 +51,26 @@ using conversions::ToPropertyValue;
...
+
+using EdgeAccessors = std::vector<memgraph::storage::v3::EdgeAccessor>;
+
+using EdgeFiller = std::function<bool(const EdgeAccessor &edge, bool is_in_edge, msgs::ExpandOneResultRow &result_row)>;
+using EdgeUniqunessFunction = std::function<EdgeAccessors(EdgeAccessors &&, memgraph::msgs::EdgeDirection)>;
+
+struct VertexIdCmpr {
+  bool operator()(const storage::v3::VertexId *lhs, const storage::v3::VertexId *rhs) const {
+    return lhs->primary_label < rhs->primary_label && lhs->primary_key < rhs->primary_key;
+  }
```
antaljanosbenjamin, line 71: I think this is not a correct implementation. Let's say we have two vertices:
1. VertexA: label id is 1, primary key is [5]
2. VertexB: label id is 2, primary key is [4]

Then check the two cases:
1. VertexA < VertexB => `1 < 2 && 5 < 4` => false
2. VertexB < VertexA => `2 < 1 && 4 < 5` => false

I think the correct implementation would be `std::tie(lhs->primary_label, lhs->primary_key) < std::tie(rhs->primary_label, rhs->primary_key)`. See [`std::tie`](https://en.cppreference.com/w/cpp/utility/tuple/tie) and the [operators](https://en.cppreference.com/w/cpp/utility/tuple/operator_cmp) for `std::tuple`.

And the cherry on the cake would be to define this as `operator<` for `v3::VertexId` and use that function here as `*lhs < *rhs`.

src/storage/v3/shard_rsm.cpp
```diff
@@ -468,8 +480,99 @@ bool FillEdges(const std::vector<EdgeAccessor> &edges, const msgs::ExpandOneRequ
...
+EdgeUniqunessFunction InitializeEdgeUniqunessFunction(bool only_unique_neighbor_rows) {
+  // Functions to select connecting edges based on uniquness
+  EdgeUniqunessFunction maybe_filter_based_on_edge_uniquness;
+
+  if (only_unique_neighbor_rows) {
+    maybe_filter_based_on_edge_uniquness = [](EdgeAccessors &&edges,
+                                              memgraph::msgs::EdgeDirection edge_direction) -> EdgeAccessors {
+      std::function<bool(std::set<const storage::v3::VertexId *, VertexIdCmpr> &,
+                         const memgraph::storage::v3::EdgeAccessor &)>
+          vertex_selector;
```
antaljanosbenjamin, line 533: This is a really small thing, I just mention this because you have to modify this file.

I think you name this `vertex_selector` to my suggestion, but (as you pointed out) my suggestion would check the exact same thing for each vertices, so I like your approach more overall, however I think the name `vertex_selector` doesn't suit this logic. Something along the line of `is_edge_unique` or something similar.

src/storage/v3/shard_rsm.cpp
```diff
@@ -467,8 +480,99 @@ bool FillEdges(const std::vector<EdgeAccessor> &edges, const msgs::ExpandOneRequ
...
+        if (vertex_selector(other_vertex_set, edge)) {
+          ret.emplace_back(edge);
+        }
+      }
+
+      return ret;
+    };
+  } else {
+    maybe_filter_based_on_edge_uniquness =
+        [](EdgeAccessors &&edges, memgraph::msgs::EdgeDirection /*edge_direction*/) -> EdgeAccessors { return edges; };
```
antaljanosbenjamin, line 569:
```suggestion
        [](EdgeAccessors &&edges, memgraph::msgs::EdgeDirection /*edge_direction*/) -> EdgeAccessors { return std::move(edges); };
```

antaljanosbenjamin (changes requested):

antaljanosbenjamin (approved):

PR #584, opened 2022-10-06
[:pineapple: < MG] Model-based testing of simulated full cluster

spacejam:
This PR adds support for generating randomized workloads that will be executed against a simulated cluster, as well as against a correctness model. Initially this just generates ScanAll and CreateVertex requests, and anything that it creates, it also inserts into a `std::set`, and when we do a ScanAll, it asserts that we get the same number of requests back. This will become much more sophisticated over time, but it's already hitting pay-dirt.


tests/simulation/CMakeLists.txt
```diff
@@ -28,9 +28,35 @@ function(add_simulation_test test_cpp)
...
+  add_executable(${target_name} ${test_cpp})
+
+  # OUTPUT_NAME sets the real name of a target when it is built and can be
+  # used to help create two targets of the same name even though CMake
+  # requires unique logical target names
+  set_target_properties(${target_name} PROPERTIES OUTPUT_NAME ${exec_name})
+
+  # sanitize
+  target_compile_options(${target_name} PRIVATE -fsanitize=address)
+  target_link_options(${target_name} PRIVATE -fsanitize=address)
```
antaljanosbenjamin, line 51: I think using the address sanitizer this way is not correct, but at least misleading. We have a CMake configuration, called `ASAN` to enable the address sanitizer for the whole codebase. On top of linking `libasan`, it will also disable `jemalloc`. As both `jemalloc` and `libasan` defines their own `malloc`/`free`, I think they are not working well together. Probably we will use one of the `malloc`/`free` (I don't know which one), but the two things can easily conflict.

As a result, I think we shouldn't enable ASAN for some subset of libraries/executables, but only globally for everything.

spacejam:
Work on this has continued in a fork that is based off of T1082 here:
https://github.com/memgraph/memgraph/pull/598/files

tests/simulation/cluster_config.hpp
```diff
@@ -0,0 +1,51 @@
...
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#pragma once
+
+#include <rapidcheck.h>
+
+#include "testing_constants.hpp"
```
42jeremy, line 16: Full path

tests/simulation/test_cluster.hpp
```diff
@@ -0,0 +1,248 @@
...
+
+#include <chrono>
+#include <iostream>
+#include <limits>
+#include <set>
+#include <thread>
+
+#include <rapidcheck.h>
+
+#include "cluster_config.hpp"
```
42jeremy, line 20: full path

42jeremy (approved):

src/io/simulator/simulator_handle.cpp
```diff
@@ -50,6 +50,8 @@ void SimulatorHandle::IncrementServerCountAndWaitForQuiescentState(Address addre
   }
 }
 
+struct SimulationExceededConfiguredAbortTime : std::exception {};
```
antaljanosbenjamin, line 53: Please use `utils::BasicException` instead of `std::exception`.

tests/simulation/test_cluster.hpp
```diff
@@ -0,0 +1,251 @@
...
+  const auto key1 = memgraph::storage::v3::PropertyValue(create_vertex.key);
+  const auto key2 = memgraph::storage::v3::PropertyValue(create_vertex.value);
+
+  std::vector<msgs::Value> primary_key = {msgs::Value(int64_t(create_vertex.key)),
+                                          msgs::Value(int64_t(create_vertex.value))};
+
+  if (correctness_model.contains(std::make_pair(create_vertex.key, create_vertex.value))) {
+    // TODO(tyler) remove this early-return when we have properly handled setting non-unique vertexes
+    return;
+  }
```
antaljanosbenjamin, line 160: Here it is quite confusing for me why both `create_vertex.key` and `create_vertex.value` is used to create a key. Maybe try to rename the members of `CreateVertex` to explain their purpose better.

tests/simulation/test_cluster.hpp
```diff
@@ -0,0 +1,251 @@
...
+  nv.label_ids.push_back({label_id});
+
+  std::vector<msgs::NewVertex> new_vertices;
+  new_vertices.push_back(std::move(nv));
+
+  auto result = shard_request_manager.Request(state, std::move(new_vertices));
+  if (result.size() != 1) {
+    spdlog::error("did not get a result back when creating the vertex");
+    RC_ASSERT(result.size() == 1);
+  }
```
antaljanosbenjamin, line 176: Could you please put an here to make sure `result[0].success` is true?

I think both `MG_ASSERT` and `RC_ASSERT` is fine here, because this is an invariant that should be true always at the current stage.

tests/simulation/test_cluster.hpp
```diff
@@ -0,0 +1,251 @@
...
+               std::set<CompoundKey> &correctness_model, ScanAll scan_all) {
+  msgs::ExecutionState<msgs::ScanVerticesRequest> request{.label = "test_label"};
+
+  auto results = shard_request_manager.Request(request);
+
+  RC_ASSERT(results.size() == correctness_model.size());
+
+  for (const auto &vertex_accessor : results) {
+    const auto properties = vertex_accessor.Properties();
+    const CompoundKey key = std::make_pair(properties[0].second.int_v, properties[1].second.int_v);
```
antaljanosbenjamin, line 191: I think using `VertexAccessor::Id` or adding `VertexAccessor::PrimaryKey` is better than expecting the order of the properties given.

tests/simulation/testing_constants.hpp
```diff
@@ -0,0 +1,28 @@
...
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#pragma once
+
+namespace memgraph::tests::simulation {
+
+// TODO(tyler) increase this when we start standing up multiple machines in cluster tests
+static constexpr auto kMinimumShards = 1;
+static constexpr auto kMaximumShards = kMinimumShards + 1;
```
antaljanosbenjamin, line 18: I think it should be possible to increase the number of shards, because I don't see a reason why it shouldn't work and we definitely need to make it work for M2.

antaljanosbenjamin (changes requested):

src/io/simulator/simulator_handle.cpp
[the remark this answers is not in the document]
spacejam: done

tests/simulation/test_cluster.hpp
spacejam: VertexAccessor has no success field, not sure what you mean here

tests/simulation/testing_constants.hpp
spacejam: Not yet.

tests/simulation/test_cluster.hpp
spacejam: I've renamed the members to be a bit more clear

PR #587, opened 2022-10-09
Upgrade code to toolchain-v5 revert to c++20

gitbuda:
TLDR: Memgraph compiles with Clang 17 💪 (both v4 and v5 toolchain work under this PR)

### TODOs
- [x] Fix the unit tests
- [x] Apply other review comments
- [x] Revert changes in files where only copyright year has changed
- [x] NuRaft doesn't compile (undeclared identified `uint64_t`, `uint32_t`) -> probably swap to `master` -> consider doing this under another PR -> fixed under https://github.com/memgraph/memgraph/pull/1733 (has to be merged before this)
- [x] Use the skiplist improvement (https://github.com/memgraph/memgraph/pull/1635, merged from master)
- [x] Fix the preprocessor -> [preprocessor_patch.diff.zip](https://github.com/memgraph/memgraph/files/13902386/preprocessor_patch.diff.zip)
- [x] Break circular deps in `preprocessor.hpp/cpp` -> https://github.com/memgraph/memgraph/pull/961
- [x] Using toolchain-v5:
  - [x] Fix all `fmt` errors and improve the code - v10.1.0
    - [x] Probably the best is to create `utils/fmt.hpp` and put all global `fmt::formatter` definitions there -> tradeoff: if we change that file everything will have to be recompiled
    - [x] `path.string()` issue:
      - [x] Use `path.string()` everywhere -> NO because it's not required
      - [x] Use `<format/std.h>` (`.string()` works with both toolchains) -> __has_include to the rescue
    - [x] Implement formattable for complex nested types like `TypedValue` -> using https://fmt.dev/latest/api.html#ostream-api
    - [x] Fix `fmt` formattable error, keep in mind that toolchain-v4 works with an older version of fmt
    - [x] Make sure `utils::MesssageWithLink` works on v4 and v5
    - [x] `fmt` is probably redundant in some of the unit tests
  - [x] Upgrade jmalloc (probably move it under memgraph), latest 5.3.0, works/preferred 5.2.1 (other flags have to be set, take a look under #1250)
  - [x] Upgrade spdlog -> Added fmt compatible spdlog under toolchain (v1.12.0)
    - spdlog v1.11 works with fmt v9.1.0 (these 2 exactly should be used), but it doesn't work with the old toolchain. 1) move fmt to memgraph or 2) move spdlog to the toolchain (1 seems better because memgraph repo is heavily dependent on these to libraries + both are easy/fast to compile, on the other side, in 1), we can't easily compile with multiple toolchains because as soon as we pick a version of spdlog inside the repo, other code will break, e.g., spdlog v1.9.2 doesn't work with the new fmt version, BUT for this to work we also have to update toolchain-v4 + remove spdlog from the repo -> a new version of toolchain has to be provided 🤯) **TODO: add spdlog to toolchain-v5 + stay backward compatible by leaving the old code under `libs/CMakeLists.txt`**
  - [x] Upgrade rocksdb - v8.1.1
  - [x] Upgrade cppitertools - v2.1
  - [x] Upgrade rapidcheck @ 2023-08-15
  - [x] Not upgrade gtest (`mgclient` https://github.com/memgraph/mgclient/pull/52 `FetchContent`, not easy to upgrade above v1.8)
- [x] Check toolchain-v4 because if everything is ok this can be merged right away -> THIS PR CAN BE MERGED
- [x] Redeploy libs cache
- [x] Move fedora.sh setup away from this PR
- [x] Figure out the linker error failure `error: unable to find library -lfmt` (it's not only fmt) -> lld for some reason doesn't work -> try to compile clang with lld


src/storage/v2/disk/storage.cpp
```diff
@@ -1487,6 +1487,7 @@ DiskStorage::DiskAccessor::CheckVertexConstraintsBeforeCommit(
       }
 
       const auto &edge = edge_acc.find(gid);
+      // TODO(gitbuda): Doesn't work on the latest compiler, what's the best way to resolve?
```
gitbuda, line 1490: Let's figure what's the proper way to resolve this CC @as51340

as51340: edge_acc is just accessor to skip list. Nothing specific to disk, so no idea. Try to comment this line and see whether the same thing will fail somewhere else because it should, we have plenty of `find` commands throughout the codebase.

src/query/typed_value.cpp
```diff
@@ -315,20 +315,17 @@ TypedValue::operator storage::PropertyValue() const {
...
-                                                                                                 \
-  const type_param &TypedValue::Value##type_enum() const {                                       \
-    if (type_ != Type::type_enum)                                                                \
-      throw TypedValueException("TypedValue is of type '{}', not '{}'", type_, Type::type_enum); \
-    return field;                                                                                \
-  }                                                                                              \
-                                                                                                 \
+#define DEFINE_VALUE_AND_TYPE_GETTERS(type_param, type_enum, field)                                  \
+  type_param &TypedValue::Value##type_enum() {                                                       \
+    if (type_ != Type::type_enum) throw TypedValueException("TODO(gitbuda): Fix typed value type "); \
```
gitbuda, line 320: Change

src/query/typed_value.cpp
```diff
@@ -19,6 +19,7 @@
 #include <string_view>
 #include <utility>
 
+#include "query/fmt.hpp"
```
gitbuda, line 22: Probably add the formatter inside typed_value.hpp

tests/manual/query_hash.cpp
```diff
@@ -15,6 +15,7 @@
 #include <gflags/gflags.h>
 
 #include "query/frontend/stripped.hpp"
+#include "storage/v2/fmt.hpp"
```
gitbuda, line 18: Try in the case `PropertyValue` formatter is part of `property_value.hpp`

gitbuda:
@as51340 @andrejtonev @Ignition This PR is ready for review from the code perspective.
The only major thing to check/fix is the `lld` because for some reason it doesn't work (`ld.gold` works fine).

src/query/plan/preprocess.cpp
```diff
@@ -623,4 +633,19 @@ QueryParts CollectQueryParts(SymbolTable &symbol_table, AstStorage &storage, Cyp
...
 
+// NOTE: matchings is defined here because it can't be defined in hpp because of forward declare -> take care if you add
+// another constructor.
+FilterInfo::FilterInfo() : matchings{} {}
+FilterInfo::~FilterInfo() = default;
+
+FilterInfo::FilterInfo(const FilterInfo &) = default;
+FilterInfo &FilterInfo::operator=(const FilterInfo &) = default;
+FilterInfo::FilterInfo(FilterInfo &&) noexcept = default;
+FilterInfo &FilterInfo::operator=(FilterInfo &&) noexcept = default;
```
andrejtonev, line 644: Can this also be moved to hpp?

src/query/plan/preprocess.hpp
```diff
@@ -200,7 +205,7 @@ class PatternFilterVisitor : public ExpressionVisitor<void> {
   void Visit(NamedExpression &op) override{};
   void Visit(RegexMatch &op) override{};
 
-  std::vector<FilterMatching> getMatchings() { return matchings_; }
+  std::vector<FilterMatching> getMatchings();
```
andrejtonev, line 208: Note from Gareth. We could just move the declaration down in the hpp, instead of moving it to the cpp.

andrejtonev: This also applies to other constructor/desctuctors in this hpp

src/query/plan/preprocess.hpp
```diff
@@ -266,6 +271,17 @@ struct FilterInfo {
   /// elements.
   enum class Type { Generic, Label, Property, Id, Pattern };
 
+  FilterInfo();
+  ~FilterInfo();
+
+  FilterInfo(const FilterInfo &);
+  FilterInfo &operator=(const FilterInfo &);
+  FilterInfo(FilterInfo &&) noexcept;
+  FilterInfo &operator=(FilterInfo &&) noexcept;
```
andrejtonev, line 280: Since it is all default, we don't even need to define them.

src/query/procedure/fmt.hpp
```diff
@@ -0,0 +1,78 @@
...
+// licenses/APL.txt.
+
+#pragma once
+
+#include <fmt/ostream.h>
+
+#include "mg_procedure.h"
+
+template <>
+class fmt::formatter<mgp_log_level> {
```
andrejtonev, line 19: Is this duplicated from an operator<< or is this a completely new stringify?
Note from Gareth, we can create a ToString and use it in both palces

src/storage/v2/disk/storage.cpp
```diff
@@ -1368,7 +1368,7 @@ DiskStorage::DiskAccessor::CheckVertexConstraintsBeforeCommit(
 /// std::map<dst_vertex_gid, ...>
 /// Here we also do flushing of too many things, we don't need to serialize edges in read-only txn, check that...
 [[nodiscard]] utils::BasicResult<StorageManipulationError, void> DiskStorage::DiskAccessor::FlushModifiedEdges(
-    const auto &edge_acc) {
+    auto &edges_acc) {
```
andrejtonev, line 1371: Why did we lose `const`? Do we need to use `.cend()`?

src/utils/async_timer.cpp
```diff
@@ -55,13 +55,12 @@ uint64_t AddFlag(std::weak_ptr<std::atomic<bool>> flag) {
 void EraseFlag(uint64_t flag_id) { expiration_flags.access().remove(flag_id); }
 
 std::weak_ptr<std::atomic<bool>> GetFlag(uint64_t flag_id) {
-  const auto flag_accessor = expiration_flags.access();
-  const auto it = flag_accessor.find(flag_id);
-  if (it == flag_accessor.end()) {
+  auto flag_accessor = expiration_flags.access();
```
andrejtonev, line 58: Another `const` loss

src/utils/logging.hpp
```diff
@@ -23,6 +23,11 @@
 #include <optional>
 
 #include <fmt/format.h>
+// NOTE: fmt 9+ introduced fmt/std.h, it's important because of, e.g., std::path formatting. toolchain-v4 has fmt 8,
+// __has_include is here because of fmt 8 compatibility.
+#if __has_include("fmt/std.h")
```
andrejtonev, line 28: Shouldn't we check the version instead?

andrejtonev (commented):
Good! Just some comments

src/storage/v2/disk/storage.cpp
gitbuda: @andrejtonev, @as51340 and I played with this for a while, but couldn't figure out a nice solution, could you try that out and propose something that will work (I think we tried to force `cend` and it also didn't work...

CMakeLists.txt
```diff
@@ -215,10 +215,13 @@ set(CMAKE_CXX_FLAGS_RELWITHDEBINFO
...
-# TODO(gitbuda): Figure out why lld produces linker errors -> probably because clang inside toolchain is linked with gold -> check...
-# add_link_options(-fuse-ld=lld)
-# TODO: use mold linker -> the issue is licensing in case we build for Mac.
+# Use lld linker to speedup build and use less memory.
+add_link_options(-fuse-ld=lld)
+# NOTE: Moving to latest Clang (probably starting from 15), lld stopped to work
+# without explicit link_directories call.
+string(REPLACE ":" " " LD_LIBS $ENV{LD_LIBRARY_PATH})
+separate_arguments(LD_LIBS)
+link_directories(${LD_LIBS})
```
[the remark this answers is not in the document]
gitbuda: I've no idea. For the sake of time, I would move on here for now 🤔

src/integrations/kafka/consumer.cpp
```diff
@@ -58,7 +58,7 @@ utils::BasicResult<std::string, std::vector<Message>> GetBatch(RdKafka::KafkaCon
       default:
         auto error = msg->errstr();
         spdlog::warn("Unexpected error while consuming message in consumer {}, error: {} (code {})!",
-                     info.consumer_name, msg->errstr(), msg->err());
+                     info.consumer_name, msg->errstr(), RdKafka::err2str(msg->err()));
```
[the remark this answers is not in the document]
gitbuda: will see if that's easier compare to this

gitbuda:
This has to be continued soon, because toolchain upgrade is dependent on this, v2.14 release is the new target data

src/query/plan/preprocess.cpp
gitbuda: I'm pretty sure it's not possible because `FilterInfo` is not fully defined under hpp file (take a look at the comment under hpp).

src/query/typed_value.hpp
```diff
@@ -769,3 +769,10 @@ TypedValue operator%(const TypedValue &a, const TypedValue &b);
 std::ostream &operator<<(std::ostream &os, const TypedValue::Type &type);
 
 }  // namespace memgraph::query
+
+#if FMT_VERSION > 90000
```
gitbuda, line 773: Move out

src/storage/v2/property_value.hpp
```diff
@@ -502,3 +503,10 @@ inline void PropertyValue::DestroyValue() noexcept {
 }
 
 }  // namespace memgraph::storage
+
+#if FMT_VERSION > 90000
```
gitbuda, line 507: Move out

src/utils/stat.hpp
```diff
@@ -48,7 +49,7 @@ inline uint64_t GetDirDiskUsage(const std::filesystem::path &path) {
         spdlog::warn(
             "Skipping file path on collecting directory disk usage '{}' because it is not readable, check file "
             "ownership and read permissions!",
-            p);
+            p.path());
```
gitbuda, line 52: remove path

tests/macro_benchmark/clients/pokec_client.cpp
```diff
@@ -24,10 +24,12 @@
 #include <gflags/gflags.h>
 #include <json/json.hpp>
 
+#include "communication/bolt/v1/value.hpp"
```
gitbuda, line 27: This can probably be removed

src/utils/stat.hpp
gitbuda: not possible because p is directory entry, all good here

src/storage/v2/disk/storage.cpp
gitbuda: const added back here

src/utils/async_timer.cpp
gitbuda: also done

src/query/plan/preprocess.hpp
gitbuda: won't work, everything which is defaulted / custom defined -> moved under cpp file

tests/unit/query_plan_operator_to_string.cpp
```diff
@@ -214,23 +214,22 @@ TYPED_TEST(OperatorToStringTest, Filter) {
   auto node_ident = IDENT("person");
   auto property = this->dba.NameToProperty("name");
   auto property_ix = this->storage.GetPropertyIx("name");
-
-  FilterInfo generic_filter_info = {.type = FilterInfo::Type::Generic, .used_symbols = {node}};
+  auto generic_filter_info = FilterInfo(FilterInfo::Type::Generic, nullptr, {node});
 
   auto id_filter = IdFilter(this->symbol_table, node, LITERAL(42));
-  FilterInfo id_filter_info = {.type = FilterInfo::Type::Id, .id_filter = id_filter};
+  auto id_filter_info = FilterInfo(FilterInfo::Type::Id, nullptr, {}, {}, id_filter);
```
andrejtonev, line 220: Prefer {} over ()

andrejtonev (approved):
Looks good!
I added comments in random parts of the codebase. Github was very confused...

release/memgraph.service
```diff
@@ -1,12 +1,27 @@
 [Unit]
 Description=Memgraph: High performance, in-memory, transactional graph database
 
+# Auto-restart config
+#StartLimitIntervalSec=300
+#StartLimitBurst=5
```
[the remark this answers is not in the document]
gitbuda: I'm not sure how did you find this, this is from some other commit/branch 🤔

PR #588, opened 2022-10-10
[:pineapple: < T1055-MG] Futurize shard request manager

gvolfing:
[project-pineapples < T1055] PR

Make ShardRequestManager work with futures 

The communication between the ShardRequestManager and the RsmClient
used to be direct. In this PR this changes into a future-based
communication type. The RsmClient stores state about the currently
processed future (either read or write request) and exposes blocking
and non-blocking functionality to obtain the filled future. The
ShardRequestManager -for now- will send of the set of requests present
in the ExecutionState and block on each of them until the requests are
completed or the set of paginated responses(caused by, for example the
batch-limit in ScanAll) are ready for the next round.


src/io/rsm/rsm_client.hpp
```diff
@@ -131,6 +153,117 @@ class RsmClient {
...
+      ReadResponse<ReadResponseT> &&read_get_response = std::move(get_response_envelope.message);
+
+      PossiblyRedirectLeader(read_get_response);
+
+      if (read_get_response.success) {
+        async_read_before_ = std::nullopt;
+        return std::move(read_get_response.read_return);
+      }
+    } else if (result_has_error) {
+      SelectRandomLeader();
```
kostasrim, line 206: If there are 3 addresses there is 33% chance to select randomly the same recipient. Can this be an issue?

src/io/rsm/rsm_client.hpp
```diff
@@ -43,28 +45,48 @@ class RsmClient {
...
 
+  /// State for single async read/write operations. In the future this could become a map
+  /// of async operations that can be accessed via an ID etc...
+  std::optional<Time> async_read_before_;
+  std::optional<ResponseFuture<ReadResponse<ReadResponseT>>> async_read_;
+  ReadRequestT current_read_request_;
+
+  std::optional<Time> async_write_before_;
+  std::optional<ResponseFuture<WriteResponse<WriteResponseT>>> async_write_;
+  WriteRequestT current_write_request_;
```
kostasrim, line 56: I leave this comment for completeness, that is, I don't recommend any action for now. I am concerned for this `State` because once we use a map to allow concurrency we implicitly opt in to a very *synchronized* shared state. What I mean by that, is that the QE will be sending a lot of requests. Each associated `Future` of a request `X` must be removed from that shared state once its fulfilled (same when we insert a future, that is when we make a request). Therefore, in a concurrency setup, we will need to protect this Shared State by a mutex. However, this could be problematic in practice because this data structure is almost guaranteed to become *hot*, and when that happens we will have a lot "waiting time" between threads. That's something we should consider.

src/io/rsm/rsm_client.hpp
```diff
@@ -131,6 +153,117 @@ class RsmClient {
 
     return TimedOut{};
   }
+
+  /// AsyncRead methods
+  void SendAsyncReadRequest(ReadRequestT req) {
+    MG_ASSERT(!async_read_);
+
+    ReadRequest<ReadRequestT> read_req;
+    read_req.operation = req;
```
kostasrim, line 162: A `ReadRequestT` gets copied once this function is called and once here. This is expensive for large data structures (and all our `Messages` contain a vector, a map or some other dynamically allocated data structure).

src/io/rsm/rsm_client.hpp
```diff
@@ -131,6 +153,117 @@ class RsmClient {
...
+    }
+    if (!result_has_error) {
+      ResponseEnvelope<WriteResponse<WriteResponseT>> &&get_response_envelope =
+          std::move(get_response_result.GetValue());
+      WriteResponse<WriteResponseT> &&write_get_response = std::move(get_response_envelope.message);
+
+      PossiblyRedirectLeader(write_get_response);
+
+      if (write_get_response.success) {
+        async_read_before_ = std::nullopt;
```
kostasrim, line 258:
```suggestion
        async_write_before_ = std::nullopt;
```

src/io/rsm/rsm_client.hpp
```diff
@@ -131,6 +153,117 @@ class RsmClient {
...
+    async_write_.reset();
+
+    const Duration overall_timeout = io_.GetDefaultTimeout();
+    const bool past_time_out = io_.Now() < *async_write_before_ + overall_timeout;
+    const bool result_has_error = get_response_result.HasError();
+
+    if (result_has_error && past_time_out) {
+      // TODO static assert the exact type of error.
+      spdlog::debug("client timed out while trying to communicate with leader server {}", leader_.ToString());
+      async_read_before_ = std::nullopt;
```
kostasrim, line 247:
```suggestion
      async_write_before_ = std::nullopt;
```

kostasrim (changes requested):
Preliminary review, I got more comments. I also want to pull the branch and try it out locally because I think there are some more issues.

src/io/rsm/rsm_client.hpp
gvolfing: When this function is being called, it is possible that previous leader_ was in a `Candidate` state. This means that the Raft group is possibly in the process of leader election. So picking the same recipient than before might just be the correct guess.

src/io/rsm/rsm_client.hpp
```diff
@@ -131,6 +153,117 @@ class RsmClient {
...
+
+  std::optional<BasicResult<TimedOut, ReadResponseT>> AwaitAsyncReadRequest() {
+    ResponseResult<ReadResponse<ReadResponseT>> get_response_result = std::move(*async_read_).Wait();
+    async_read_.reset();
+
+    const Duration overall_timeout = io_.GetDefaultTimeout();
+    const bool past_time_out = io_.Now() < *async_read_before_ + overall_timeout;
+    const bool result_has_error = get_response_result.HasError();
+
+    if (result_has_error && past_time_out) {
```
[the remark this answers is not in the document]
gvolfing: We are not hiding the error. The only possible error type we can get from the `BasicResult<...>` is `TimeOut` error.
The `result_has_error` flag is indicating the `TimeOut` error we get from the future(if there is one).
The `past_time_out` flag is indicating if the time frame defined by the `RsmClient` is over or not.
That can mean that we are dealing with a redirection. And we can still retry at another address if we are within the time frame, defined by the `RsmClient`.
In short, if this condition is evaluating to true, then we propagate the presence of a `TimeOut` to the application code, otherwise retry at another address if we still have time to do so.

src/io/rsm/rsm_client.hpp
gvolfing: There were actually 3 copies, since we also cache the previous request.
We still need at least one deep copy because of the cached request IMO. But I will take it by const& and move where possible.
Same goes for `WriteRequestT`

src/io/rsm/rsm_client.hpp
kostasrim: Then the question would be what's the probability of this happening and if we can somehow mitigate this through some metadata. But I think this is beyond the current scope -- just thinking out loud here.

src/io/rsm/rsm_client.hpp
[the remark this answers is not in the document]
kostasrim: >The only possible error type we can get from the BasicResult<...> is TimeOut error. 
Yep and I think that's an issue in the long run, because a TimeOut is not the only error we can get... Anyway, leave this for now, it was the same as before.

src/query/v2/shard_request_manager.hpp
```diff
@@ -172,34 +177,34 @@ class ShardRequestManager : public ShardRequestManagerInterface {
   std::vector<VertexAccessor> Request(ExecutionState<ScanVerticesRequest> &state) override {
```
kostasrim, line 177: We can simplify a lot. Plz go over the suggested changes and what I will write here. All the functions I mention (including the suggestions) are for ScanVertices but the similar changes/logic should apply to the rest of the Requests.

First, observe that `PollOnRequests` is almost identical with `AwaitOnFirstRequest` with a few differences: 
1. `Poll` returns a `bool` to indicate at least one of the requests is ready. However it suffers from a small issue that it always returns false even if there are available futures because `paginated_response_tracker.at(*shard_it) == PaginatedResponseState::PartiallyFinished` will always be true if the first request in the request list is `PartiallyFinished`. This complicates the logic of the loop later on. 
2. `AwaitOnRequest` blocks on the next pending request within the `paginated_response_tracker`. 

Now if you look carefully on the `Request` body: 

```
   // Outer loop to make sure we are handling the paginated responses as well.
    while (!state.shard_cache.empty()) {
      // 1. Send the requests.
      SendAllRequests(state);
```
First, the `while (!state.shard_cache.empty())` is a bug, because it won't be true for continuations that are not sent yet (that is, the second, third etc batch from a shard request). Second, we already called `SendAllRequests` which effectively scattered the requests to their corresponding shards. That means, that we can iteratively block with a few modifications on the `RsmClient` and the body of the `Request` (plz see explanations for each suggested change).

src/query/v2/shard_request_manager.hpp
```diff
@@ -172,34 +177,34 @@ class ShardRequestManager : public ShardRequestManagerInterface {
...
+
+    // Outer loop to make sure we are handling the paginated responses as well.
+    while (!state.shard_cache.empty()) {
+      // 1. Send the requests.
+      SendAllRequests(state);
+
+      std::map<Shard, PaginatedResponseState> paginated_response_tracker;
+      for (auto shard_it = state.shard_cache.begin(); shard_it != state.shard_cache.end(); ++shard_it) {
+        paginated_response_tracker.insert(std::make_pair(*shard_it, PaginatedResponseState::Pending));
       }
```
kostasrim, line 189: remove this

src/query/v2/shard_request_manager.hpp
kostasrim, line 177:
```suggestion
    std::vector<VertexAccessor> Request(ExecutionState<ScanVerticesRequest> &state) override {
      MaybeInitializeExecutionState(state);
      std::vector<ScanVerticesResponse> responses;
  
      SendAllRequests(state);
  
      std::map<Shard, PaginatedResponseState> paginated_response_tracker;
      for (const auto &shard : state.shard_cache) {
        paginated_response_tracker.insert(std::make_pair(shard, PaginatedResponseState::Pending));
      }   
      do {
        AwaitOnPaginatedRequests(state, responses, paginated_response_tracker);
      } while (all_requests_gathered(paginated_response_tracker));
  
      MaybeCompleteState(state);
      // TODO(kostasrim) Before returning start prefetching the batch (this shall be done once we get MgFuture as return
      // result of storage_client.SendReadRequest()).
      return PostProcess(std::move(responses));
    }

```

kostasrim: That's it. We use the `PaginatedResponseTracker` to keep track of the pending requests. Look the changes in the `AwaitOnPaginatedRequests`.

src/query/v2/shard_request_manager.hpp
```diff
@@ -407,6 +399,257 @@ class ShardRequestManager : public ShardRequestManagerInterface {
...
+      } else {
+        state.requests[id].start_id.second = response.next_start_id->second;
+        paginated_response_tracker.at(*shard_it) = PaginatedResponseState::PartiallyFinished;
+        ++shard_it;
+      }
+      responses.push_back(std::move(response));
+    }
+
+    return at_least_one_result;
+  }
```
kostasrim, line 484:
```suggestion
```

kostasrim: No longer needed.

src/query/v2/shard_request_manager.hpp
```diff
@@ -407,6 +399,257 @@ class ShardRequestManager : public ShardRequestManagerInterface {
...
+      }
+
+      responses.push_back(std::move(response));
+      shard_it = shard_cache_ref.erase(shard_it);
+    }
+
+    return at_least_one_result;
+  }
+
+  void AwaitOnFirstRequest(ExecutionState<ScanVerticesRequest> &state, std::vector<ScanVerticesResponse> &responses,
```
kostasrim, line 558: Let's rename this and slightly alter its behaviour:
1. Instead of iterating over the next pending request only. Block on them one by one. 
2. If a pending request is ready (that is, there was no Redirection from the RsmClient), update its status on the Paginated tracker. 
3. If there was a redirection, immediately resent the request on the RsmClient (see there the changes).

src/query/v2/shard_request_manager.hpp
```diff
@@ -407,6 +399,257 @@ class ShardRequestManager : public ShardRequestManagerInterface {
...
+    }
+
+    if (!response.next_start_id) {
+      paginated_response_tracker.erase(*shard_it);
+      shard_cache_ref.erase(shard_it);
+    } else {
+      state.requests[request_idx].start_id.second = response.next_start_id->second;
+      paginated_response_tracker.at(*shard_it) = PaginatedResponseState::PartiallyFinished;
+    }
+    responses.push_back(std::move(response));
```
kostasrim, line 598:
```suggestion
      auto &shard_cache_ref = state.shard_cache;
  
      // Find the first request that is not holding a paginated response.
      size_t request_idx = 0;
      auto shard_it = shard_cache_ref.begin();
      for (;shard_it != shard_cache_ref.end(); ++shard_it, ++request_idx) {
        if (paginated_response_tracker.at(*shard_it) != PaginatedResponseState::Pending) {
          continue;
        }
  
        auto &storage_client = GetStorageClientForShard(
            *state.label, storage::conversions::ConvertPropertyVector(state.requests[request_idx].start_id.second));
        auto await_result = storage_client.AwaitAsyncReadRequest();
  
        if (!await_result) {
          // Redirection has occured.
          return;
        }
  
        if (await_result->HasError()) {
          throw std::runtime_error("ScanAll request timed out");
        }
  
        ReadResponses read_response_variant = await_result->GetValue();
        auto response = std::get<ScanVerticesResponse>(read_response_variant);
        if (!response.success) {
          throw std::runtime_error("ScanAll request did not succeed");
        }
  
        if (!response.next_start_id) {
          shard_cache_ref.erase(shard_it);
        } else {
          state.requests[request_idx].start_id.second = response.next_start_id->second;
        }
        paginated_response_tracker[*shard_it] = PaginatedResponseState::PartiallyFinished;
  
        responses.push_back(std::move(response));
      }

```

kostasrim: Rename this function to `AwaitOnPaginatedRequests`

src/io/rsm/rsm_client.hpp
```diff
@@ -131,6 +154,115 @@ class RsmClient {
...
+    if (!result_has_error) {
+      ResponseEnvelope<ReadResponse<ReadResponseT>> &&get_response_envelope = std::move(get_response_result.GetValue());
+      ReadResponse<ReadResponseT> &&read_get_response = std::move(get_response_envelope.message);
+
+      PossiblyRedirectLeader(read_get_response);
+
+      if (read_get_response.success) {
+        async_read_before_ = std::nullopt;
+        return std::move(read_get_response.read_return);
+      }
```
kostasrim, line 204:
```suggestion
      if (read_get_response.success) {
        async_read_before_ = std::nullopt;
        return std::move(read_get_response.read_return);
      }
      else {
          SendAsyncReadRequest(current_read_request_);
       }   
```

kostasrim: Don't waste time cause resenting the request (otherwise TimeOuts will be more likely + this enables the other suggestions to work)

kostasrim (changes requested):

42jeremy (approved):

42jeremy (approved):

PR #593, opened 2022-10-12
[:pineapple: < T1079] Add new dataset for mgbench

42jeremy:
[master < Task] PR
```
Add new dataset for mgbench
Add new script to generate dataset
```


tests/mgbench/dataset_creator.py
```diff
@@ -0,0 +1,94 @@
...
+
+# Create extra index: in distributed, this will be the schema
+f.write("CREATE INDEX ON :User(uuid);\n")
+f.write("CREATE INDEX ON :Permission(uuid);\n")
+f.write("CREATE INDEX ON :Identity(uuid);\n")
+
+platform_ids = [f"somePlatformId_{id}" for id in range(10)]
+
+# This is the number of clusters to change if you want a bigger dataset
+number_of_clusters = 3000000
```
gitbuda, line 73: I would make this somehow configurable, probably runtime flag, maybe an env variable

tests/mgbench/datasets.py
```diff
@@ -262,29 +261,124 @@ def benchmark__aggregation__min_max_avg(self):
...
-                {"id": self._get_random_vertex()})
+        return ("MATCH (n:User {id: $id}) RETURN n", {"id": self._get_random_vertex()})
 
     def benchmark__match__vertex_on_property(self):
-        return ("MATCH (n {id: $id}) RETURN n",
-                {"id": self._get_random_vertex()})
+        return ("MATCH (n {id: $id}) RETURN n", {"id": self._get_random_vertex()})
+
+
+class Distributed(Dataset):
```
gitbuda, line 285: I would call this `Dataset` a bit differently, e.g. `AccessControl` of something more related to what's actually inside 😄

42jeremy:
The dataset generation can be adapted, see customer folder.
We need more permissions and it should be organized per cluster only.

tests/mgbench/datasets.py
```diff
@@ -262,29 +261,124 @@ def benchmark__aggregation__min_max_avg(self):
...
+    # Explaination of datasets:
+    #   - empty_only_index: contains index; contains no data
+    #   - small/medium/large: contains index; contains data (respectively small/medium/large dataset)
+    #
+    # See dataset_creator.py to understand the datamodel and generate a dataset
+
+    NAME = "accesscontrol"
+    VARIANTS = ["empty_only_index", "small", "medium", "large"]
+    DEFAULT_VARIANT = "empty_only_index"
+    URLS = {
```
42jeremy, line 296: Have a way to keep it private?

42jeremy: Or have a way to specify local URLs for private tests

tests/mgbench/dataset_creator.py
```diff
@@ -0,0 +1,116 @@
+# Copyright 2021 Memgraph Ltd.
```
jbajic, line 1:
```suggestion
# Copyright 2022 Memgraph Ltd.
```

tests/mgbench/dataset_creator.py
```diff
@@ -0,0 +1,116 @@
...
+    f.write("CREATE INDEX ON :Identity(email);\n")
+
+    # Create extra index: in distributed, this will be the schema
+    f.write("CREATE INDEX ON :File(uuid);\n")
+    f.write("CREATE INDEX ON :Permission(uuid);\n")
+    f.write("CREATE INDEX ON :Identity(uuid);\n")
+
+    uuid = 1
+
+    files = []
```
jbajic, line 89: This is unused

tests/mgbench/dataset_creator.py
```diff
@@ -0,0 +1,116 @@
...
+#   - Permission: [Permission(uuid), Permission(name)]
+#   - Identity: [Identity(uuid), Identity(email)]
+#
+# Edges:
+#   - (:Permission)-[:IS_FOR_FILE]->(:File)
+#   - (:Permission)-[:IS_FOR_IDENTITYR]->(:Identity)
+#
+# AccessControl specific: uuid is the schema
+
+if __name__ == "__main__":
```
jbajic, line 50: Move all the staff to the main function, and define the main, because:
1. Everything is now global
2. Makes importing easier,
3. And testing (if we ever want to test it)

tests/mgbench/datasets.py
```diff
@@ -262,29 +261,124 @@ def benchmark__aggregation__min_max_avg(self):
...
+        assert type in ["File", "Permission", "Identity"]
+
+        first_uuid = Dataset.get_size(self)["uuid_ranges"][type]["first_uuid"]
+        last_uuid = Dataset.get_size(self)["uuid_ranges"][type]["last_uuid"]
+
+        random_value = random.randint(first_uuid, last_uuid)
+        return random_value
+
+    def __init__(self, variant=None):
+        Dataset.__init__(self, variant)
```
jbajic, line 351:
```suggestion
        super().__init__(self, variant)
```

tests/mgbench/datasets.py
```diff
@@ -262,29 +261,124 @@ def benchmark__aggregation__min_max_avg(self):
...
+                "Identity": {"first_uuid": 1000001, "last_uuid": 3000000},
+            },
+        },
+    }
+
+    def _get_random_uuid(self, type):
+        assert type in ["File", "Permission", "Identity"]
+
+        first_uuid = Dataset.get_size(self)["uuid_ranges"][type]["first_uuid"]
+        last_uuid = Dataset.get_size(self)["uuid_ranges"][type]["last_uuid"]
```
jbajic, line 345: Why specify `Dataset`, you should be able to just call get_size, since there is the inheritance
```suggestion
        first_uuid = self.get_size()["uuid_ranges"][type]["first_uuid"]
        last_uuid = self.get_size()["uuid_ranges"][type]["last_uuid"]
```

tests/mgbench/datasets.py
```diff
@@ -262,29 +261,124 @@ def benchmark__aggregation__min_max_avg(self):
...
+    def __init__(self, variant=None):
+        Dataset.__init__(self, variant)
+        self.next_value_idx = Dataset.get_size(self)["vertices"] + 1
+
+    def benchmark__create__vertex(self):
+        self.next_value_idx += 1
+        query = (f"CREATE (:File {{uuid: {self.next_value_idx}}});", {})
+        return query
+
+    def benchmark__create__edges(self):
```
jbajic, line 359: Why are there two underscores?
```suggestion
    def benchmark_create_edges(self):
```

42jeremy: I believe this is based on these double underscore that we know how to cluster macros to execute within a benchmark (the whole file is written like this).
Between __xxx__ you have a specific action "xxx" as: "match", "create", "aggregation" etc..

42jeremy: CF: https://github.com/memgraph/memgraph/blob/3daab6ce979c01c2b789936f2a1602c4b2209ef2/tests/mgbench/benchmark.py#L160-L163

jbajic: Ok maybe put this comment in the code somewhere, or in the README how to extend these benchmarks, because seeing it like this seems like a typo

42jeremy: I just found out there's already some documentation about it:
https://github.com/memgraph/memgraph/blob/3daab6ce979c01c2b789936f2a1602c4b2209ef2/tests/mgbench/datasets.py#L95-L98
I will resolve the comment

tests/mgbench/datasets.py
```diff
@@ -262,29 +261,124 @@ def benchmark__aggregation__min_max_avg(self):
...
+
+        first_uuid = self.get_size()["uuid_ranges"][type]["first_uuid"]
+        last_uuid = self.get_size()["uuid_ranges"][type]["last_uuid"]
+
+        random_value = random.randint(first_uuid, last_uuid)
+        return random_value
+
+    def __init__(self, variant=None):
+        super().__init__(variant)
+        self.next_value_idx = Dataset.get_size(self)["vertices"] + 1
```
jbajic, line 352:
```suggestion
        self.next_value_idx = self.get_size()["vertices"] + 1
```

tests/mgbench/benchmark.py
```diff
@@ -235,45 +255,45 @@ def filter_benchmarks(generators, patterns):
...
             print("Sample query:", get_queries(func, 1)[0][0])
-            print("Executing benchmark with", count, "queries that should "
-                  "yield a single-threaded runtime of",
-                  args.single_threaded_runtime_sec, "seconds.")
-            print("Queries are executed using", args.num_workers_for_benchmark,
-                  "concurrent clients.")
+            print(
+                "Executing benchmark with",
+                count,
+                "queries that should " "yield a single-threaded runtime of",
```
jbajic, line 274:
```suggestion
                "queries that should yield a single-threaded runtime of",
```

tests/mgbench/dataset_creator.py
```diff
@@ -0,0 +1,120 @@
...
+    f.write("MATCH (n) DETACH DELETE n;\n")
+
+    # Create the indexes
+    f.write("CREATE INDEX ON :File;\n")
+    f.write("CREATE INDEX ON :Permission;\n")
+    f.write("CREATE INDEX ON :Identity;\n")
+    f.write("CREATE INDEX ON :File(platformId);\n")
+    f.write("CREATE INDEX ON :File(name);\n")
+    f.write("CREATE INDEX ON :Permission(name);\n")
+    f.write("CREATE INDEX ON :Identity(email);\n")
```
jbajic, line 81: Maybe use with, not sure how it will look after so many indentations:
```suggestion
    with open(filename, "w") as f:

      f.write("MATCH (n) DETACH DELETE n;\n")

      # Create the indexes
      f.write("CREATE INDEX ON :File;\n")
      f.write("CREATE INDEX ON :Permission;\n")
      f.write("CREATE INDEX ON :Identity;\n")
      f.write("CREATE INDEX ON :File(platformId);\n")
      f.write("CREATE INDEX ON :File(name);\n")
      f.write("CREATE INDEX ON :Permission(name);\n")
      f.write("CREATE INDEX ON :Identity(email);\n")
```
and you can remove close

jbajic (approved):
Left a few small comments, but otherwise looks fine

tests/mgbench/benchmark.py
```diff
@@ -235,45 +255,45 @@ def filter_benchmarks(generators, patterns):
...
-                  "yield a single-threaded runtime of",
-                  args.single_threaded_runtime_sec, "seconds.")
-            print("Queries are executed using", args.num_workers_for_benchmark,
-                  "concurrent clients.")
+            print(
+                "Executing benchmark with",
+                count,
+                "queries that should yield a single-threaded runtime of",
+                args.single_threaded_runtime_sec,
+                "seconds.",
```
antaljanosbenjamin, line 276: is that extra comma is intentional?

tests/mgbench/benchmark.py
```diff
@@ -78,54 +78,62 @@ def filter_benchmarks(generators, patterns):
...
+parser.add_argument("--single-threaded-runtime-sec", type=int, default=10, help="single threaded duration of each test")
+parser.add_argument("--no-load-query-counts", action="store_true", help="disable loading of cached query counts")
+parser.add_argument("--no-save-query-counts", action="store_true", help="disable storing of cached query counts")
+parser.add_argument("--export-results", default="", help="file path into which results should be exported")
+parser.add_argument(
+    "--temporary-directory", default="/tmp", help="directory path where temporary data should " "be stored"
+)
+parser.add_argument("--no-properties-on-edges", action="store_true", help="disable properties on edges")
+parser.add_argument("--datasets", default="datasets", help="datasets to scan")
+parser.add_argument("--datasets-path", default=".", help="path to datasets to scan")
```
antaljanosbenjamin, line 123: Do we need two arguments? As far as I understood this requires two arguments to be used like: `--datasets-path=/home/me/Development/datasets --datasets=my_personal_dataset`. Why not just `--datasets-path=/home/me/Development/datasets/my_personal_dataset.py`? I think it would be easier to use without too much complexity in this file (basically `os.path.split` and `os.path.splitext` the functions that will solve the problem)

tests/mgbench/benchmark.py
42jeremy: It's put automatically by the formatter. It gets back when I remove it :(

tests/mgbench/benchmark.py
42jeremy: Done (there was no reason to do so)

antaljanosbenjamin (approved):

gitbuda (changes requested):
The script fails when the local dataset path is not provided

tests/mgbench/benchmark.py
jbajic: Black always adds a trailing comma when it does not change anything semantically since it makes consequent addition of stuff easier, plus if something is changed only one line will be changed and reduces changes of git conflicts with other files, so I would say this is a good practice to follow with formats that allow this

gitbuda:
@42jeremy is it possible to actually run this as an e2e test inside CI, in other words adding `accesscontrol/small` on each commit?

42jeremy:
One needs to remove index creation from cypher files

42jeremy:
> @42jeremy is it possible to actually run this as an e2e test inside CI, in other words adding `accesscontrol/small` on each commit?

I've added it to the git workflow, we should get a failure if a commit make the benchmark fails

42jeremy:
To run benchmark locally when dataset is on AWS (AccessControl):
Execute:
cd memgraph/tests/mgbench
./benchmark.py accesscontrol/medium --num-workers-for-import 1 --test-system-arg  "bolt-num-workers 1"

.github/workflows/diff.yaml
```diff
@@ -220,6 +230,16 @@ jobs:
...
 
+      - name: Run single benchmark test
+        run: |
+          # Activate toolchain.
+          source /opt/toolchain-v4/activate
+
+          # Run simulation tests.
+          cd tests/mgbench
+          rm -r .cache
+          ./benchmark.py accesscontrol/small --num-workers-for-import 1 --test-system-arg  "bolt-num-workers 1"
```
gitbuda, line 241: There is a bug here because the flag is called `--test-system-args` + the actual value should be `"--bolt-num-workers 1"`, leave the `--` so that we can really pass any database flag without some kind of string manipulation, ofc, it's a bit more error-prone but that's should be fine.

tests/mgbench/runners.py
```diff
@@ -79,8 +78,16 @@ def _get_args(self, **kwargs):
...
-                "Older versions of Memgraph can't disable properties on edges!"
+            assert self._properties_on_edges, "Older versions of Memgraph can't disable properties on edges!"
+
+        if self._extra_args != "":
+            args_list = self._extra_args.split(" ")
+            assert len(args_list) % 2 == 0
+            for i in range(0, len(args_list), 2):
+                kwargs[args_list[i]] = args_list[i + 1]
+
+        kwargs["split-file"] = self._split_file
```
gitbuda, line 89: This doesn't work with regular Memgraph (built on `master`), I would make the `mgbench` code totally independent of what "version" of Memgraph is run so that we can easily merge `mgbench` code from master or the other way around + we don't have to change the code later on. In other words, remove the split files logic from `tests/mgbench` and use extra flags to pass that info to Memgraph.

.github/workflows/diff.yaml
gitbuda: I think the split file can also be passed / should be passed as `--test-system-args` flag 🤔

42jeremy: This was the original approach I took, but it had other disadvantages. I will investigate that again!

tests/mgbench/benchmark.py
```diff
@@ -78,54 +79,64 @@ def filter_benchmarks(generators, patterns):
...
+parser.add_argument("--no-properties-on-edges", action="store_true", help="disable properties on edges")
+parser.add_argument("--datasets-path", default=".", help="path to datasets to scan")
+parser.add_argument("--test-system-args", default="")
 args = parser.parse_args()
 
+head_tail = os.path.split(args.datasets_path)
+path_without_dataset_name = head_tail[0]
+dataset_name = head_tail[1].split(".")[0]
+sys.path.append(path_without_dataset_name)
+dataset_to_use = importlib.import_module(dataset_name)
```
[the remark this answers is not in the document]
gitbuda: I think this is fixed so resolving this comment

.github/workflows/diff.yaml
gitbuda: that's fine (didn't know that), so leave the `--` 😄

42jeremy: -> Having "--bolt-num-workers 1 --split-file /tmp/config.txt" was making benchmark.py more confused about all the flags, that was the reason
(it is possible but will require more time)

gitbuda (approved):
Finally runs against memgraph v2!

42jeremy:
Let's merge this branch once https://github.com/memgraph/memgraph/pull/643 is merged into project-pineapples

PR #599, opened 2022-10-15
[master <] Add Fedora 36 as an OS

gitbuda:


gitbuda:
I've randomly tried to build Memgraph on Fedora because Fedora Server seems to gain traction recently. Ofc, there are issues, the toolchain complies with these dependencies, but the Memgraph compilation fails. The solution again seems to be a newer version of the toolchain.

environment/os/fedora-36.sh
```diff
@@ -0,0 +1,131 @@
...
+    wget # for downloading libs
+    libuuid-devel java-11-openjdk # required by antlr
+    readline-devel # for memgraph console
+    python3-devel # for query modules
+    openssl-devel
+    libseccomp-devel
+    python3 python3-pip python3-virtualenv nmap-ncat # for qa, macro_benchmark and stress tests
+    #
+    # IMPORTANT: python3-yaml does NOT exist on CentOS
+    # Install it manually using `pip3 install PyYAML`
```
jbajic, line 54: Remove comment, or update with Fedora

jbajic (approved):
Small comment, otherwise looks fine

environment/os/fedora-36.sh
gitbuda: done

PR #600, opened 2022-10-17
[:pineapple: < T1088-MG] Split file parsing

antaljanosbenjamin:
Final commit message:
```
Add temporary support for split files. This is only temporary solution until we
get the shard splitting implemented.
```


gvolfing (approved):

tests/unit/coordinator_shard_map.cpp
```diff
@@ -0,0 +1,104 @@
...
+  };
+
+  check_label("label_1",
+              {SchemaProperty{shard_map.properties.at("primary_property_name_1"), common::SchemaType::STRING}},
+              std::vector<PrimaryKey>{
+                  PrimaryKey{PropertyValue{""}},
+                  PrimaryKey{PropertyValue{"asdasd"}},
+                  PrimaryKey{PropertyValue{"qweqwe"}},
+                  PrimaryKey{PropertyValue{"bnm"}},
+                  PrimaryKey{PropertyValue{"tryuryturtyur"}},
```
jbajic, line 89: How come the first pk is empty?

tests/unit/coordinator_shard_map.cpp
```diff
@@ -0,0 +1,104 @@
...
+property_1
+string
+property_2
+int
+primary_property_name_2
+InT
+2
+[first,1 ,2]
+[     second   ,-1,  -9223372036854775808]
+)";
```
jbajic, line 54: From what I understand we don't have a way to put secondary labels here?

antaljanosbenjamin: Actually that is a bug, good catch!

jbajic (approved):

PR #605, opened 2022-10-19
[:pineapple: < T1067-MG] Use primary labels and indicies inside query execution

gvolfing:

Make label based Indexing work

Because of the lexicographical sharding, the labels themselves are
acting as indexes. If a label is specified in a MATCH query we can
safely narrow the range of Shards we have to scan through based on that
label. In this PR introduce the necessary changes in order to achieve
that.


src/query/v2/shard_request_manager.hpp
```diff
@@ -227,7 +228,7 @@ class ShardRequestManager : public ShardRequestManagerInterface {
     return str;
   }
   const std::string &LabelToName(memgraph::storage::v3::LabelId /*label*/) const override {
-    static std::string str{"dummy__label"};
+    static std::string str{"label"};
```
antaljanosbenjamin, line 231: You can use `shards_map_.GetLabelName` to return a proper value.

src/query/v2/shard_request_manager.hpp
```diff
@@ -244,6 +245,11 @@ class ShardRequestManager : public ShardRequestManagerInterface {
            }) != schema_it->second.end();
   }
 
+  bool IsPrimaryLabel(const LabelId name) const override {
```
antaljanosbenjamin, line 248:
```suggestion
bool IsPrimaryLabel(const LabelId label)
```

src/query/v2/shard_request_manager.hpp
```diff
@@ -129,6 +129,7 @@ class ShardRequestManagerInterface {
   virtual const std::string &PropertyToName(memgraph::storage::v3::PropertyId prop) const = 0;
   virtual const std::string &LabelToName(memgraph::storage::v3::LabelId label) const = 0;
   virtual const std::string &EdgeTypeToName(memgraph::storage::v3::EdgeTypeId type) const = 0;
+  virtual bool IsPrimaryLabel(LabelId name) const = 0;
```
antaljanosbenjamin, line 132:
```suggestion
virtual bool IsPrimaryLabel(LabelId label) const = 0;
```

src/query/v2/shard_request_manager.hpp
```diff
@@ -244,6 +245,11 @@ class ShardRequestManager : public ShardRequestManagerInterface {
            }) != schema_it->second.end();
   }
 
+  bool IsPrimaryLabel(const LabelId name) const override {
+    return std::find_if(shards_map_.label_spaces.begin(), shards_map_.label_spaces.end(),
+                        [name](auto &lb) { return lb.first == name; }) != shards_map_.label_spaces.end();
```
antaljanosbenjamin, line 250: Why not to use `shards_map_.label_spaces.contains(label)`?

src/query/v2/shard_request_manager.hpp
antaljanosbenjamin, line 231: You can use `shards_map_.GetLabelName` to return a proper value. Otherwise this function won't work for labels apart for "label". That would just push the hard coded value into a different place, but won't really help us with other labels.

src/query/v2/shard_request_manager.hpp
```diff
@@ -519,11 +536,22 @@ class ShardRequestManager : public ShardRequestManagerInterface {
...
-      storage_client.SendAsyncReadRequest(request);
+      if (!state.label) {
+        const auto &current_shard = state.shard_cache[shard_idx];
+
+        auto &storage_client = GetStorageClientForShard(current_shard);
+        ReadRequests req = request;
+        storage_client.SendAsyncReadRequest(request);
+      } else {
+        auto &storage_client = GetStorageClientForShard(
+            *state.label, storage::conversions::ConvertPropertyVector(request.start_id.second));
```
antaljanosbenjamin, line 549: I propose to use the same way of getting the storage client because of multiple reasons:
1. It is faster
2. Relying on the `request.start_id` is really bad in my opinion as that field should become optional soon.
3. It makes the logic simple as it eliminates on if/else.

antaljanosbenjamin (changes requested):

antaljanosbenjamin (approved):

42jeremy (approved):

src/coordinator/shard_map.cpp
```diff
@@ -206,12 +206,19 @@ std::ostream &operator<<(std::ostream &in, const ShardMap &shard_map) {
...
-Shards ShardMap::GetShards(const LabelName &label) {
+Shards ShardMap::GetShards(const LabelName &label) const {
   const auto id = labels.at(label);
-  auto &shards = label_spaces.at(id).shards;
+  const auto &shards = label_spaces.at(id).shards;
   return shards;
 }
 
+std::vector<Shards> ShardMap::GetShards() const {
+  std::vector<Shards> all_shards;
```
kostasrim, line 216:
```suggestion
  std::vector<Shards> all_shards;
  all_shards.reserve(label_spaces.size());
```

kostasrim (approved):
Minor comment

PR #607, opened 2022-10-19
[:pineapple: < T1086-MG] Test distributed operators e2e

kostasrim:
[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR

### DIFF
--- a/src/coordinator/shard_map.hpp
+++ b/src/coordinator/shard_map.hpp
@@ -118,12 +118,12 @@ struct ShardMap {
   Hlc shard_map_version;
   uint64_t max_property_id{kNotExistingId};
   uint64_t max_edge_type_id{kNotExistingId};
-  storage::v3::NameIdMapper properties;
-  storage::v3::NameIdMapper edge_types;
-  storage::v3::NameIdMapper labels;
+  std::map<PropertyName, PropertyId> properties;
+  std::map<EdgeTypeName, EdgeTypeId> edge_types;
   uint64_t max_label_id{kNotExistingId};
   std::map<LabelId, LabelSpace> label_spaces;
   std::map<LabelId, std::vector<SchemaProperty>> schemas;
+  std::map<LabelName, LabelId> labels;
 
   [[nodiscard]] static ShardMap Parse(std::istream &input_stream);
   friend std::ostream &operator<<(std::ostream &in, const ShardMap &shard_map);
--- a/src/coordinator/shard_map.hpp
+++ b/src/coordinator/shard_map.hpp
@@ -25,6 +25,7 @@
 #include "io/address.hpp"
 #include "storage/v3/config.hpp"
 #include "storage/v3/id_types.hpp"
+#include "storage/v3/name_id_mapper.hpp"
 #include "storage/v3/property_value.hpp"
 #include "storage/v3/schemas.hpp"
 #include "storage/v3/temporal.hpp"
@@ -117,10 +118,10 @@ struct ShardMap {
   Hlc shard_map_version;
   uint64_t max_property_id{kNotExistingId};
   uint64_t max_edge_type_id{kNotExistingId};
-  std::map<PropertyName, PropertyId> properties;
-  std::map<EdgeTypeName, EdgeTypeId> edge_types;
+  storage::v3::NameIdMapper properties;
+  storage::v3::NameIdMapper edge_types;
+  storage::v3::NameIdMapper labels;
   uint64_t max_label_id{kNotExistingId};
-  std::map<LabelName, LabelId> labels;
   std::map<LabelId, LabelSpace> label_spaces;
   std::map<LabelId, std::vector<SchemaProperty>> schemas;
 
--- a/src/query/v2/accessors.hpp
+++ b/src/query/v2/accessors.hpp
@@ -80,7 +80,8 @@ class VertexAccessor final {
   VertexAccessor(Vertex v, std::vector<std::pair<PropertyId, Value>> props,
                  const msgs::ShardRequestManagerInterface *manager);
 
-  VertexAccessor(Vertex v, std::map<PropertyId, Value> props, const msgs::ShardRequestManagerInterface *manager);
+  VertexAccessor(Vertex v, std::map<PropertyId, Value> &&props, const msgs::ShardRequestManagerInterface *manager);
+  VertexAccessor(Vertex v, const std::map<PropertyId, Value> &props, const msgs::ShardRequestManagerInterface *manager);
 
   [[nodiscard]] Label PrimaryLabel() const;
 
--- a/src/query/v2/accessors.hpp
+++ b/src/query/v2/accessors.hpp
@@ -23,6 +23,10 @@
 #include "utils/memory.hpp"
 #include "utils/memory_tracker.hpp"
 
+namespace memgraph::msgs {
+class ShardRequestManagerInterface;
+} // namespace memgraph::msgs
+
 namespace memgraph::query::v2::accessors {
 
 using Value = memgraph::msgs::Value;
@@ -36,7 +40,7 @@ class VertexAccessor;
 
 class EdgeAccessor final {
  public:
-  explicit EdgeAccessor(Edge edge);
+  explicit EdgeAccessor(Edge edge, const msgs::ShardRequestManagerInterface *manager);
 
   [[nodiscard]] EdgeTypeId EdgeType() const;
 
@@ -64,6 +68,7 @@ class EdgeAccessor final {
 
  private:
   Edge edge;
+  const msgs::ShardRequestManagerInterface *manager_;
 };
 
 class VertexAccessor final {
@@ -71,7 +76,8 @@ class VertexAccessor final {
   using PropertyId = msgs::PropertyId;
   using Label = msgs::Label;
   using VertexId = msgs::VertexId;
-  VertexAccessor(Vertex v, std::vector<std::pair<PropertyId, Value>> props);
+  VertexAccessor(Vertex v, std::vector<std::pair<PropertyId, Value>> props,
+                 const msgs::ShardRequestManagerInterface *manager);
 
   [[nodiscard]] Label PrimaryLabel() const;
 
@@ -140,6 +146,7 @@ class VertexAccessor final {
  private:
   Vertex vertex;
   std::vector<std::pair<PropertyId, Value>> properties;
+  const msgs::ShardRequestManagerInterface *manager_;
 };
 
 // inline VertexAccessor EdgeAccessor::To() const { return VertexAccessor(impl_.ToVertex()); }
--- a/src/query/v2/accessors.hpp
+++ b/src/query/v2/accessors.hpp
@@ -11,6 +11,7 @@
 
 #pragma once
 
+#include <map>
 #include <optional>
 #include <utility>
 #include <vector>
@@ -25,7 +26,7 @@
 
 namespace memgraph::msgs {
 class ShardRequestManagerInterface;
-} // namespace memgraph::msgs
+}  // namespace memgraph::msgs
 
 namespace memgraph::query::v2::accessors {
 
@@ -78,6 +79,8 @@ class VertexAccessor final {
   using VertexId = msgs::VertexId;
   VertexAccessor(Vertex v, std::vector<std::pair<PropertyId, Value>> props,
                  const msgs::ShardRequestManagerInterface *manager);
+
+  VertexAccessor(Vertex v, std::map<PropertyId, Value> props, const msgs::ShardRequestManagerInterface *manager);
 
   [[nodiscard]] Label PrimaryLabel() const;
 
--- a/src/query/v2/interpret/awesome_memgraph_functions.hpp
+++ b/src/query/v2/interpret/awesome_memgraph_functions.hpp
@@ -16,12 +16,15 @@
 #include <unordered_map>
 
 #include "query/v2/bindings/typed_value.hpp"
+#include "query/v2/db_accessor.hpp"
 #include "storage/v3/view.hpp"
 #include "utils/memory.hpp"
 
-namespace memgraph::query::v2 {
+namespace memgraph::msgs {
+class ShardRequestManagerInterface;
+} // namespace memgraph::msgs
 
-class DbAccessor;
+namespace memgraph::query::v2 {
 
 namespace {
 const char kStartsWith[] = "STARTSWITH";
@@ -31,7 +34,9 @@ const char kId[] = "ID";
 }  // namespace
 
 struct FunctionContext {
-  DbAccessor *db_accessor;
+  // TODO(kostasrim) consider optional here. ShardRequestManager does not exist on the storage.
+  // DbAccessor *db_accessor;
+  msgs::ShardRequestManagerInterface *manager;
   utils::MemoryResource *memory;
   int64_t timestamp;
   std::unordered_map<std::string, int64_t> *counters;
--- a/src/query/v2/interpreter.cpp
+++ b/src/query/v2/interpreter.cpp
@@ -903,7 +903,6 @@ PreparedQuery PrepareCypherQuery(ParsedQuery parsed_query, std::map<std::string,
         "convert the parsed row values to the appropriate type. This can be done using the built-in "
         "conversion functions such as ToInteger, ToFloat, ToBoolean etc.");
   }
-  shard_request_manager->StartTransaction();
   auto plan = CypherQueryToPlan(
       parsed_query.stripped_query.hash(), std::move(parsed_query.ast_storage), cypher_query, parsed_query.parameters,
       parsed_query.is_cacheable ? &interpreter_context->plan_cache : nullptr, shard_request_manager);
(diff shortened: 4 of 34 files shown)


tests/setup.sh
```diff
@@ -36,12 +36,12 @@ PYTHON_MINOR=$(python3 -c 'import sys; print(sys.version_info[:][1])')
 # NOTE (2021-11-15): PyPi doesn't contain pulsar-client for Python 3.9 so we have to use
 # our manually built wheel file. When they update the repository, pulsar-client can be
 # added as a regular PIP dependancy
-if [ $PYTHON_MINOR -lt 9 ]; then
-  pip --timeout 1000 install "pulsar-client==2.8.1"
-else
-  pip --timeout 1000 install https://s3-eu-west-1.amazonaws.com/deps.memgraph.io/pulsar_client-2.8.1-cp39-cp39-manylinux_2_5_x86_64.manylinux1_x86_64.whl
-fi
-
+#if [ $PYTHON_MINOR -lt 9 ]; then
```
kostasrim, line 39: This doesn't work on my machine so I had to comment it out. Once we get back streams we should address this.

src/query/v2/plan/operator.cpp
```diff
@@ -2245,11 +2252,12 @@ class LoadCsvCursor : public Cursor {
...
  private:
-  csv::Reader MakeReader(EvaluationContext *eval_context) {
+  csv::Reader MakeReader(ExecutionContext &context) {
+    auto &eval_context = context.evaluation_context;
     Frame frame(0);
     SymbolTable symbol_table;
-    DbAccessor *dba = nullptr;
-    auto evaluator = ExpressionEvaluator(&frame, symbol_table, *eval_context, dba, storage::v3::View::OLD);
+    auto evaluator =
+        ExpressionEvaluator(&frame, symbol_table, eval_context, context.shard_request_manager, storage::v3::View::OLD);
```
antaljanosbenjamin, line 2260: Why is this change necessary?

src/query/v2/plan/operator.cpp
```diff
@@ -2457,15 +2465,38 @@ class DistributedExpandCursor : public Cursor {
...
+    msgs::ExpandOneRequest request;
+    // to not fetch any properties of the edges
+    request.edge_properties.emplace();
+    request.src_vertices.push_back(get_dst_vertex(self_.common_.direction));
+    msgs::ExecutionState<msgs::ExpandOneRequest> request_state;
+    auto result_rows = context.shard_request_manager->Request(request_state, std::move(request));
+    MG_ASSERT(result_rows.size() == 1);
+    auto &result_row = result_rows.front();
+    frame[self_.common_.node_symbol] = accessors::VertexAccessor(
+        msgs::Vertex{result_row.src_vertex}, result_row.src_vertex_properties, context.shard_request_manager);
```
antaljanosbenjamin, line 2497: Why not to make the `GetProperties` request work?

src/query/v2/plan/operator.cpp
```diff
@@ -2560,7 +2580,7 @@ class DistributedExpandCursor : public Cursor {
         auto &edge = *current_in_edge_it_;
         ++current_in_edge_it_;
         frame[self_.common_.edge_symbol] = edge;
-        pull_node(edge, EdgeAtom::Direction::IN);
+        PullDstVertex(frame, context);
```
antaljanosbenjamin, line 2583: Why not to pass in the direction? Accessing the edge just to get an information that we already know seems unnecessary.

src/query/v2/interpret/awesome_memgraph_functions.hpp
```diff
@@ -31,7 +34,9 @@ const char kId[] = "ID";
 }  // namespace
 
 struct FunctionContext {
-  DbAccessor *db_accessor;
+  // TODO(kostasrim) consider optional here. ShardRequestManager does not exist on the storage.
```
antaljanosbenjamin, line 37: But both storage and query engine can provide a `NameIdMapper` which is actually necessary for the accessors!

src/query/v2/interpreter.cpp
```diff
@@ -940,6 +942,7 @@ PreparedQuery PrepareExplainQuery(ParsedQuery parsed_query, std::map<std::string
                                   InterpreterContext *interpreter_context,
                                   msgs::ShardRequestManagerInterface *shard_request_manager,
                                   utils::MemoryResource *execution_memory) {
+  shard_request_manager->StartTransaction();
```
antaljanosbenjamin, line 945: I had to solve the same problem for the profile query. I think the proper way to fix this is done in https://github.com/memgraph/memgraph/pull/617/commits/5939fb2b0cadd14bfee3081f57625ab709e533ea

src/query/v2/interpreter.cpp
```diff
@@ -1179,14 +1183,15 @@ PreparedQuery PrepareIndexQuery(ParsedQuery parsed_query, bool in_explicit_trans
 
 PreparedQuery PrepareAuthQuery(ParsedQuery parsed_query, bool in_explicit_transaction,
                                std::map<std::string, TypedValue> *summary, InterpreterContext *interpreter_context,
-                               DbAccessor *dba, utils::MemoryResource *execution_memory) {
+                               DbAccessor *dba, utils::MemoryResource *execution_memory,
```
antaljanosbenjamin, line 1186: Please remove the `DbAccessor` parameter from the function where it is not used anymore.

src/storage/v3/shard_rsm.cpp
```diff
@@ -322,9 +331,14 @@ std::optional<msgs::ExpandOneResultRow> GetExpandOneResult(
...
   }
-
+  std::optional<std::map<PropertyId, Value>> src_vertex_properties;
   /// Fill up source vertex properties
-  auto src_vertex_properties = FillUpSourceVertexProperties(v_acc, req);
+  if (req.src_vertex_properties) {
+    src_vertex_properties = FillUpSourceVertexProperties(v_acc, req);
+  } else {
+    src_vertex_properties = CollectAllPropertiesFromAccessor(*v_acc, storage::v3::View::NEW, schema);
+  }
```
antaljanosbenjamin, line 340: Please rethink this logic, because the `FillUpSourceVertexProperties` does the same check. Maybe try to unify the logic in a better way.

tests/e2e/distributed_queries/workloads.yaml
```diff
@@ -11,3 +11,23 @@ workloads:
     binary: "tests/e2e/pytest_runner.sh"
     args: ["distributed_queries/distributed_queries.py"]
     <<: *template_cluster
+
+  - name: "Distributed unwind collect"
```
antaljanosbenjamin, line 15: I is possible to move all of these tests into the original `distributed_queries.py` file? I feel that creating different workloads for such small tests are unnecessary. I know creating multiple tests in the same file is not viable (because we cannot delete the vertices/edges), but I think these are tests that will be removed in the long term as the cypher test cases cover them much better so it is okay to have a single, maybe a bit longer test case to check all of these.

tests/e2e/distributed_queries/order_by_and_limit.py
```diff
@@ -0,0 +1,46 @@
...
+def test_order_by_and_limit(connection):
+    wait_for_shard_manager_to_initialize()
+    cursor = connection.cursor()
+
+    assert has_n_result_row(cursor, "CREATE (n :label {property:1})", 0)
+    assert has_n_result_row(cursor, "CREATE (n :label {property:2})", 0)
+    assert has_n_result_row(cursor, "CREATE (n :label {property:3})", 0)
+    assert has_n_result_row(cursor, "CREATE (n :label {property:4})", 0)
+
+    assert has_n_result_row(cursor, "MATCH (n) WITH collect(n) AS nd UNWIND nd AS result RETURN result", 4)
```
antaljanosbenjamin, line 29: This is a copy-pasta error.

antaljanosbenjamin (changes requested):

src/query/v2/plan/operator.cpp
kostasrim: Because the expression evaluator now uses the shard_request_manager for `NameId` mappings

src/query/v2/requests.hpp
```diff
@@ -387,7 +387,8 @@ struct GetPropertiesResponse {
 enum class EdgeDirection : uint8_t { OUT = 1, IN = 2, BOTH = 3 };
 
 struct ExpandOneRequest {
-  // TODO(antaljanosbenjamin): Filtering based on the id of the other end of the edge?
+  // To get Edges with with destination vertex edge_wist_st
+  std::optional<VertexId> edge_with_dst;
```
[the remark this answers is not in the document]
kostasrim: I think it suffices for now as we only expanding one tupple at a time. We can change this when we support batched expansions

src/query/v2/accessors.hpp
```diff
@@ -140,6 +149,7 @@ class VertexAccessor final {
  private:
   Vertex vertex;
   std::vector<std::pair<PropertyId, Value>> properties;
+  const msgs::ShardRequestManagerInterface *manager_;
```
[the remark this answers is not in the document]
kostasrim: I ended up using the `NameIdMapper` in ShardMap.

tests/e2e/distributed_queries/distinct.py
```diff
@@ -0,0 +1,37 @@
...
+from common import *
+
+
+def test_distinct(connection):
+    wait_for_shard_manager_to_initialize()
+    cursor = connection.cursor()
+
+    assert has_n_result_row(cursor, "CREATE (n :label {property:0})", 0)
+    assert has_n_result_row(cursor, "CREATE (n :label {property:1})", 0)
+    assert has_n_result_row(cursor, "MATCH (n), (m) CREATE (n)-[:TO]->(m)", 0)
```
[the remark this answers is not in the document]
kostasrim: yep this works and I added it as a test case

tests/e2e/distributed_queries/optional_match.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def test_optional_match(connection):
+    wait_for_shard_manager_to_initialize()
+    cursor = connection.cursor()
+
+    assert has_n_result_row(cursor, "CREATE (n :label {property:0})", 0)
+
+    results = execute_and_fetch_all(
+        cursor, "MATCH (n:label) OPTIONAL MATCH (n:label)-[:TO]->(parent:label) RETURN parent"
```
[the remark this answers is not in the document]
kostasrim: It fails, I will figure out why

tests/e2e/distributed_queries/workloads.yaml
kostasrim: For now I rather not as we don't have `detatch delete`. 
> but I think these are tests that will be removed in the long term
I will do it myself when this is the case.

src/storage/v3/shard_rsm.cpp
kostasrim: The handlers are (slightly) a spagheti mess. As m1 is the priority, I won't spend much on this side at all. I will see if I can remove 1-2 lines but I won't pay much attention to it.

src/query/v2/plan/operator.cpp
kostasrim: We don't care about the direction. We care about fetching the properties of the dst vertex  (I know this is done for us on paths but we still have to fetch the last dst vertex). This is a slight hack, once we polish, we will use `GetProperties`

src/query/v2/plan/operator.cpp
```diff
@@ -221,6 +224,10 @@ class DistributedCreateNodeCursor : public Cursor {
       rqst.label_ids.push_back(msgs::Label{.id = primary_label});
       requests.push_back(std::move(rqst));
     }
+    const auto &rqst = requests.front();
+    msgs::Vertex v{.id = std::make_pair(rqst.label_ids.front(), rqst.primary_key)};
+    frame[nodes_info_.front()->symbol] =
+        TypedValue(query::v2::accessors::VertexAccessor(std::move(v), rqst.properties, context.shard_request_manager));
```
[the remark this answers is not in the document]
kostasrim: Nope, there is no need. We already know ALL the information of the soon to be created vertex and therefore we put it  on the frame. If the request fails, an exception will be thrown and we are safe. Otherwise, by the time we get a response the Vertex will already exist on the frame

src/query/v2/interpreter.cpp
kostasrim: I slightly altered this. I move the `StartTransaction` at the top of the `Prepare` function. That way, the plan rewritter has access to `NameId` mappers.

src/query/v2/interpreter.cpp
kostasrim: Yes and no, it's used on the `PullPlan`. I would rather create a separate PR and clean up the `db_accessor` alltogether. We won't use this anymore on QE side.

src/query/v2/requests.hpp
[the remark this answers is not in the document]
antaljanosbenjamin: My concern here is the rest of the request and the shard rsm is capable of receiving multiple vertices to expand from. The limitation is only present in the query engine, but we have tests that uses the requests and shard rsm directly, thus this change introduce inconsistency in terms of `ExpandOneRequest`. I think the change is quite small (vector instead of optional plus the changes on how it is accessed), but makes the request consistent.

src/query/v2/plan/operator.cpp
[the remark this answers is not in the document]
antaljanosbenjamin: The function is called `NodeCreationInfoToRequest`, there I think it should only transfrom the node creation info to a request and nothing more. This is a very unexpected side effect of this function which is not wanted in my opinion. Furthermore, the two things (transforming and putting value on the frame) can be clearly separated.

src/query/v2/plan/operator.cpp
antaljanosbenjamin: My explanation is wrong, but I think the requested change is valid.

Inside the `PullDstVertex` function the `get_dst_vertex` is called with `self_.common_.direction`, but that can be `BOTH` not just `IN` or `OUT`. That's why the direction was passed in previously. The current implementation will fail if `self_.common_.direction` is `BOTH`.

src/query/v2/interpreter.cpp
antaljanosbenjamin: Why though? The planning happens in the `PrepareCypherQuery` function and even with my changes the planner/rewriter has proper access to the shard map. I think the difference between the current version and my changes (that works better in my opinion, because it only starts a transaction when it is necessary, not in case of `BEGIN`/`COMMIT`/`ROLLBACK` queries) is small, but has advantages.

src/coordinator/shard_map.hpp
```diff
@@ -117,10 +118,10 @@ struct ShardMap {
   Hlc shard_map_version;
   uint64_t max_property_id{kNotExistingId};
   uint64_t max_edge_type_id{kNotExistingId};
-  std::map<PropertyName, PropertyId> properties;
-  std::map<EdgeTypeName, EdgeTypeId> edge_types;
+  storage::v3::NameIdMapper properties;
+  storage::v3::NameIdMapper edge_types;
+  storage::v3::NameIdMapper labels;
```
antaljanosbenjamin, line 123: I think keeping the ShardMap is as simple as possible is a good thing, because this structure is used to communicate the information. However by introducing `NameIdMapper` here we are kind of mixing responsibilities of different layers. We already had some not-the-best deicisions (e.g.: the ShardMap should be used to communicate the information only, not to fetch shards/mappings/etc) and with this change we just deepened this mistake.

However it is possible this is only my opinion, so if it is okay for the others, then it is also okay for me.

src/query/v2/accessors.hpp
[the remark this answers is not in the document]
antaljanosbenjamin: I think keeping the ShardMap is as simple as possible is a good thing, because the ShardMap structure should be used to communicate the information. However by introducing `NameIdMapper` we are kind of mixing responsibilities of different layers. We already had some not-the-best implementation things (e.g.: the ShardMap should be used to communicate the information only, not to fetch shards/mappings/etc) and with this change we just deepened this mistake.

However it is possible this is only my opinion, so if it is okay for the others, then it is also okay for me.

src/query/v2/plan/operator.cpp
kostasrim: Yes, now we use the `ShardRequestManager` to map property names to ids :D

src/query/v2/interpreter.cpp
kostasrim: I was wrong, I moved it back

antaljanosbenjamin (approved):

kostasrim:
* Fix Explain queries
* Add Vertex/Edge accessor support for properties
* Fix projections
* Fix expansions to fetch destination vertex properties
* Fix improper use of ShardMap on bolt and replaced it with the ShardRequestManager 
* Add NameToId mappers on ShardRequestManager
* Add e2e tests for operators
* Fix OPTIONAL MATCH

PR #611, opened 2022-10-21
[:pineapple: < T1083-MG] Add limit and OrderBy to ExpandOne

42jeremy:
```
Add Limit and OrderBy to ExpandOne request
```


src/storage/v3/shard_rsm.cpp
```diff
@@ -352,6 +354,41 @@ std::optional<msgs::ExpandOneResultRow> GetExpandOneResult(
   return result_row;
 }
 
+std::optional<msgs::ExpandOneResultRow> GetExpandOneResult2(
```
42jeremy, line 357: rename

42jeremy:
Once https://github.com/memgraph/memgraph/pull/607/files is merged, one can add extra e2e test in order_by_and_limit.py?

src/storage/v3/expr.cpp
```diff
@@ -186,7 +186,7 @@ TypedValue ComputeExpression(DbAccessor &dba, const std::optional<memgraph::stor
   expr::SymbolGenerator symbol_generator(&symbol_table, identifiers);
   (std::any_cast<Expression *>(expr))->Accept(symbol_generator);
 
-  if (node_identifier.symbol_pos_ != -1) {
+  if (node_identifier.symbol_pos_ != -1 && v_acc.has_value()) {
```
antaljanosbenjamin, line 189: When can a vertex be nullopt? Based on the code I would say when we want to evaluate order by expresssions on edges. However I think that is wrong. When we are evaluating expressions on edges, we should always be possible to access the vertex we are using to iterate over its edges.

src/storage/v3/shard_rsm.cpp
```diff
@@ -457,6 +492,129 @@ EdgeFiller InitializeEdgeFillerFunction(const msgs::ExpandOneRequest &req) {
   return edge_filler;
 }
 
+msgs::ReadResponses HandleReadWithOrderBy(msgs::ExpandOneRequest &&req, Shard::Accessor &&acc) {
```
antaljanosbenjamin, line 495: I don't like the idea of handling the requests separately depending on whether it contains an order by or not. E.g.: this version doesn't take into account (at least I cannot spot where) the `src_vertices` field of the request which is essential for this request.

src/storage/v3/shard_rsm.cpp
```diff
@@ -457,6 +492,129 @@ EdgeFiller InitializeEdgeFillerFunction(const msgs::ExpandOneRequest &req) {
...
+  auto batch_limit = req.limit;
+  auto dba = DbAccessor{&acc};
+
+  auto maybe_filter_based_on_edge_uniquness = InitializeEdgeUniqunessFunction(req.only_unique_neighbor_rows);
+  auto edge_filler = InitializeEdgeFillerFunction(req);
+
+  auto vertex_iterable = acc.Vertices(View::OLD);
+  const auto ordered_vertices = OrderByElements<VertexAccessor>(acc, dba, vertex_iterable, req.order_by);
+
+  for (const auto &ordered_vertice : ordered_vertices) {
```
antaljanosbenjamin, line 508: `ordered_vertice` is a wierd name.

src/storage/v3/shard_rsm.cpp
```diff
@@ -457,6 +492,129 @@ EdgeFiller InitializeEdgeFillerFunction(const msgs::ExpandOneRequest &req) {
...
+    if (!src_vertex_acc_opt) {
+      action_successful = false;
+      spdlog::debug("Encountered an error while trying to obtain VertexAccessor. Transaction id: {}",
+                    req.transaction_id.logical_id);
+      break;
+    }
+
+    if (!req.filters.empty()) {
+      // NOTE - DbAccessor might get removed in the future.
+      auto dba = DbAccessor{&acc};
```
antaljanosbenjamin, line 590: We already have a `dba` variable on line 573.

src/storage/v3/request_helper.hpp
```diff
@@ -9,14 +9,21 @@
 // by the Apache License, Version 2.0, included in the file
 // licenses/APL.txt.
 
+#pragma once
+
 #include <vector>
 
 #include "ast/ast.hpp"
+#include "pretty_print_ast_to_original_expression.hpp"
```
antaljanosbenjamin, line 17: Use full path please.

tests/simulation/shard_rsm.cpp
```diff
@@ -680,6 +680,70 @@ void AttemptToExpandOneWithUniqueEdges(ShardClient &client, uint64_t src_vertex_
...
+    if (read_res.HasError()) {
+      continue;
+    }
+
+    auto write_response_result = read_res.GetValue();
+    auto write_response = std::get<msgs::ExpandOneResponse>(write_response_result);
+
+    // We check that we do not have more results than the limit. Based on the data in the graph, we know that we should
+    // receive exactly limit responses.
+    MG_ASSERT(write_response.result.size() == limit);
```
antaljanosbenjamin, line 729: We should have `std::min(expand_one_req.src_vertices.size(), limit)` number of result rows.

src/storage/v3/shard_rsm.cpp
```diff
@@ -457,6 +492,129 @@ EdgeFiller InitializeEdgeFillerFunction(const msgs::ExpandOneRequest &req) {
...
+      action_successful = false;
+      break;
+    }
+    auto primary_key = src_vertex_acc.PrimaryKey(View::NEW);
+    if (primary_key.HasError()) {
+      action_successful = false;
+      break;
+    }
+
+    msgs::VertexId src_vertice(msgs::Label{.id = *label_id}, conversions::ConvertValueVector(*primary_key));
```
antaljanosbenjamin, line 543: `src_vertice` is also a weird name.

antaljanosbenjamin (changes requested):
I didn't do a very thorough review as I assume the code will dramatically change.

src/storage/v3/request_helper.cpp
```diff
@@ -68,15 +31,49 @@ VerticesIterable::Iterator GetStartVertexIterator(VerticesIterable &vertex_itera
...
+      if (maybe_in_edges.HasValue()) {
+        in_edges = maybe_in_edges.GetValue();
+      }
+      if (maybe_out_edges.HasValue()) {
+        out_edges = maybe_out_edges.GetValue();
+      }
+    }
+  }
+
+  return std::array<std::vector<EdgeAccessor>, 2>{in_edges, out_edges};
```
jbajic, line 76:
```suggestion
  return std::array<std::vector<EdgeAccessor>, 2>{std::move(in_edges), std::move(out_edges)};
```

src/storage/v3/shard_rsm.cpp
```diff
@@ -457,6 +492,129 @@ EdgeFiller InitializeEdgeFillerFunction(const msgs::ExpandOneRequest &req) {
...
+
+  std::vector<msgs::ExpandOneResultRow> results;
+  auto batch_limit = req.limit;
+  auto dba = DbAccessor{&acc};
+
+  auto maybe_filter_based_on_edge_uniquness = InitializeEdgeUniqunessFunction(req.only_unique_neighbor_rows);
+  auto edge_filler = InitializeEdgeFillerFunction(req);
+
+  auto vertex_iterable = acc.Vertices(View::OLD);
+  const auto ordered_vertices = OrderByElements<VertexAccessor>(acc, dba, vertex_iterable, req.order_by);
```
jbajic, line 506: Ah yes, here we need to order by src vertices, and not all vertices (`vertex_iterable`)...

src/storage/v3/shard_rsm.cpp
```diff
@@ -457,6 +492,129 @@ EdgeFiller InitializeEdgeFillerFunction(const msgs::ExpandOneRequest &req) {
...
+      }
+    }
+
+    auto [in_edge_accessors, out_edge_accessors] = GetEdgesFromVertex(src_vertex_acc, req.direction);
+    const auto in_ordered_edges = OrderByElements<EdgeAccessor>(acc, dba, in_edge_accessors, req.order_by);
+    const auto out_ordered_edges = OrderByElements<EdgeAccessor>(acc, dba, out_edge_accessors, req.order_by);
+
+    std::vector<EdgeAccessor> in_edge_ordered_accessors;
+    std::transform(in_ordered_edges.begin(), in_ordered_edges.end(), std::back_inserter(in_edge_ordered_accessors),
+                   [](auto &edge_element) { return edge_element.object_acc; });
```
jbajic, line 526:
```suggestion
                   [](const auto &edge_element) { return edge_element.object_acc; });
```

src/storage/v3/shard_rsm.cpp
```diff
@@ -457,6 +492,129 @@ EdgeFiller InitializeEdgeFillerFunction(const msgs::ExpandOneRequest &req) {
...
+
+    std::vector<EdgeAccessor> out_edge_ordered_accessors;
+    std::transform(out_ordered_edges.begin(), out_ordered_edges.end(), std::back_inserter(out_edge_ordered_accessors),
+                   [](auto &edge_element) { return edge_element.object_acc; });
```
jbajic, line 530:
```suggestion
                   [](const auto &edge_element) { return edge_element.object_acc; });
```

src/storage/v3/shard_rsm.cpp
42jeremy: Agree, I would still keep it like this. Reason is that `ExpandOneRequest` has member `src_vertices` and I believe we need to remain consistent.
Alternatively, I can modify `ExpandOneRequest` as well and give a different name to both

src/storage/v3/expr.cpp
42jeremy: It crashes without it. I'll add more tests and try to understand afterwards

42jeremy: I have removed the check and made it an assert instead.
Crash was due to a nullopt vertex_accessor (which was possible within our workflow).
To summarize how we solved the root cause: we will always keep the vertex_accessor when sorting edges (that comes from a single node). An example where this is needed: an ordering based on a map on the node, where edges would be the key.

src/storage/v3/shard_rsm.cpp
```diff
@@ -902,24 +944,73 @@ msgs::ReadResponses ShardRsm::HandleRead(msgs::ExpandOneRequest &&req) {
...
-                                     shard_->GetSchema(shard_->PrimaryLabel()));
 
-    if (!result) {
+    vertex_accessors.emplace_back(src_vertex_acc_opt.value());
+  }
+
+  if (!req.order_by.empty()) {
+    // Can we do differently to avoid this? We need OrderByElements but currently it returns vector<Element>, so this
+    // workaround is here to avoid more duplication later
+    auto sorted_vertices = OrderByVertices(acc, dba, vertex_accessors, req.order_by);
```
jbajic, line 961: How can `sorted_vertices` be local variable, shouldn't you iterate over ordered src vertices, if there is order by?

src/storage/v3/expr.cpp
```diff
@@ -187,6 +187,7 @@ TypedValue ComputeExpression(DbAccessor &dba, const std::optional<memgraph::stor
   (std::any_cast<Expression *>(expr))->Accept(symbol_generator);
 
   if (node_identifier.symbol_pos_ != -1) {
+    MG_ASSERT(v_acc.has_value());
```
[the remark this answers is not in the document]
42jeremy: Good catch, we always need a VertexAccessor now

src/storage/v3/shard_rsm.cpp
```diff
@@ -363,6 +363,44 @@ std::optional<msgs::ExpandOneResultRow> GetExpandOneResult(
   return result_row;
 }
 
+std::optional<msgs::ExpandOneResultRow> GetExpandOneResult(
+    VertexAccessor v_acc, msgs::VertexId src_vertex, const msgs::ExpandOneRequest &req,
+    std::vector<EdgeAccessor> in_edge_accessors, std::vector<EdgeAccessor> out_edge_accessors,
+    const EdgeUniqunessFunction &maybe_filter_based_on_edge_uniquness, const EdgeFiller &edge_filler,
+    const Schemas::Schema *schema) {
```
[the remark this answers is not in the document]
42jeremy: (Done locally as part of another commit)

src/storage/v3/shard_rsm.cpp
42jeremy: This is indeed a local variable that I used to update `vertex_accessors`. The "trick" was done since the return type of `OrderByElements` (now `OrderByVertices`) differed from what we exactly need to keep the code of `HandleRead(msgs::ExpandOneRequest &)`relatively compact and identical for the most part regardless of whether we need ordering or not.
I will look at the return type and see if there is an easy way to avoid this extra transformation. If not, I will at least rename the variable to clarify it's only local and that this is not a mistake.

src/storage/v3/shard_rsm.cpp
```diff
@@ -363,6 +363,44 @@ std::optional<msgs::ExpandOneResultRow> GetExpandOneResult(
   return result_row;
 }
 
+std::optional<msgs::ExpandOneResultRow> GetExpandOneResult(
```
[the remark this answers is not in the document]
42jeremy: I moved function out of shard_rsm, merging is still to be investigated

src/storage/v3/expr.cpp
antaljanosbenjamin: You assert on the edge accessor, not on the vertex accessor.

antaljanosbenjamin (changes requested):

src/storage/v3/expr.cpp
42jeremy: To explain for future us: two asserts were added, one on the `std::optional<VertexAccessor>` and one on the `std::optional<EdgeAccessor>`
As part of an another remark, the `VertexAccessor` no longer need to be optional so we are left with a single assert on `EdgeAccessor`

src/query/v2/requests.hpp
```diff
@@ -403,7 +403,7 @@ struct ExpandOneRequest {
   std::vector<std::string> vertex_expressions;
   std::vector<std::string> edge_expressions;
 
-  std::optional<std::vector<OrderBy>> order_by;
+  std::vector<OrderBy> order_by;
```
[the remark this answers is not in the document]
jbajic: But that isn't necessary since the expression itself contains either node or edge evaluation (i.e. `MG_SYMBOL_NODE` and `MG_SYMBOL_NODE`), and when passed in evaluator it is clear if it is vertex or edge order by

src/storage/v3/expr.hpp
```diff
@@ -9,6 +9,8 @@
 // by the Apache License, Version 2.0, included in the file
 // licenses/APL.txt.
 
+#pragma once
```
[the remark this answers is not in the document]
42jeremy: I think we need to thank Jure for that, he's the one that fixed it if I recall (I only inherited the PR)

tests/simulation/shard_rsm.cpp
```diff
@@ -680,6 +680,77 @@ void AttemptToExpandOneWithUniqueEdges(ShardClient &client, uint64_t src_vertex_
...
+    auto read_res = client.SendReadRequest(expand_one_req);
+    if (read_res.HasError()) {
+      continue;
+    }
+
+    auto write_response_result = read_res.GetValue();
+    auto write_response = std::get<msgs::ExpandOneResponse>(write_response_result);
+
+    // We check that we do not have more results than the limit. Based on the data in the graph, we know that we should
+    // receive exactly limit responses.
```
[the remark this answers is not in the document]
42jeremy: Comment is wrong and we are expecting a single result row because of how the graph is built. I'll change the test to have meaningful results

src/storage/v3/value_conversions.hpp
```diff
@@ -129,4 +129,27 @@ inline std::vector<Value> ConvertValueVector(const std::vector<v3::PropertyValue
 inline msgs::VertexId ToMsgsVertexId(const v3::VertexId &vertex_id) {
   return {msgs::Label{vertex_id.primary_label}, ConvertValueVector(vertex_id.primary_key)};
 }
+
+inline std::vector<std::pair<v3::PropertyId, v3::PropertyValue>> ConvertPropertyMap(
+    std::vector<std::pair<v3::PropertyId, Value>> &&properties) {
```
jbajic, line 134: Why force r-value, why not by value, and the user can decide to move (e.g. if you want to reuse the vector of values for something else). I would just pass reference here

src/storage/v3/request_helper.hpp
```diff
@@ -9,14 +9,26 @@
 // by the Apache License, Version 2.0, included in the file
 // licenses/APL.txt.
 
+#pragma once
+
 #include <vector>
 
 #include "ast/ast.hpp"
```
jbajic, line 16: I've just noticed this this included ast should be from storage/v3/bindings/ast
```suggestion
#include "storage/v3/bindings/ast/ast.hpp"
```

src/storage/v3/request_helper.hpp
```diff
@@ -99,18 +111,95 @@ class TypedValueVectorCompare final {
...
-  VertexAccessor vertex_acc;
+  TObjectAccessor object_acc;
 };
 
-std::vector<Element> OrderByElements(Shard::Accessor &acc, DbAccessor &dba, VerticesIterable &vertices_iterable,
-                                     std::vector<msgs::OrderBy> &order_bys);
+template <typename T>
+concept VerticesIt = utils::SameAsAnyOf<T, VerticesIterable, std::vector<VertexAccessor>>;
+template <VerticesIt TIterable>
+std::vector<Element<VertexAccessor>> OrderByVertices(DbAccessor &dba, TIterable &iterable,
```
jbajic, line 123:
```suggestion
template <typename T>
concept VerticesIt = utils::SameAsAnyOf<T, VerticesIterable, std::vector<VertexAccessor>>;

template <VerticesIt TIterable>
std::vector<Element<VertexAccessor>> OrderByVertices(DbAccessor &dba, TIterable &iterable,
```

src/storage/v3/request_helper.hpp
```diff
@@ -99,18 +111,95 @@ class TypedValueVectorCompare final {
...
+  order_bys.erase(it_to_remove, order_bys.end());
+
+  std::vector<Ordering> ordering;
+  ordering.reserve(order_bys.size());
+  std::transform(order_bys.begin(), order_bys.end(), std::back_inserter(ordering), [](const auto &order_by) {
+    if (memgraph::msgs::OrderingDirection::ASCENDING == order_by.direction) {
+      return Ordering::ASC;
+    }
+    MG_ASSERT(memgraph::msgs::OrderingDirection::DESCENDING == order_by.direction);
+    return Ordering::DESC;
```
jbajic, line 139: Replace with switch

src/storage/v3/shard_rsm.cpp
```diff
@@ -42,433 +42,21 @@
...
+using conversions::ConvertPropertyMap;
 using conversions::ConvertPropertyVector;
 using conversions::ConvertValueVector;
+using conversions::FromMap;
 using conversions::FromPropertyValueToValue;
 using conversions::ToMsgsVertexId;
 using conversions::ToPropertyValue;
 
-namespace {
-namespace msgs = msgs;
```
jbajic, line 58: I :heart: the move of all these functions

src/storage/v3/request_helper.cpp
```diff
@@ -13,70 +13,532 @@
...
+  auto &pk = maybe_pk.GetValue();
+  MG_ASSERT(schema.second.size() == pk.size(), "PrimaryKey size does not match schema!");
+  for (size_t i{0}; i < schema.second.size(); ++i) {
+    ret.emplace(schema.second[i].property_id, FromPropertyValueToValue(std::move(pk[i])));
+  }
+
+  return ret;
+}
+
+struct LocalError {};
```
jbajic, line 63: Unused

src/storage/v3/request_helper.cpp
```diff
@@ -13,70 +13,532 @@
...
+using SpecificEdgeProperties = std::tuple<msgs::VertexId, msgs::Gid, SpecificEdgePropertyDataSructure>;
 
-    for (const auto &order_by : order_bys) {
-      const auto val =
-          ComputeExpression(dba, *it, std::nullopt, order_by.expression.expression, expr::identifier_node_symbol, "");
-      properties_order_by.push_back(val);
+using SpecificEdgePropertiesVector = std::vector<SpecificEdgeProperties>;
+using AllEdgePropertiesVector = std::vector<AllEdgeProperties>;
+
+using EdgeFiller = std::function<bool(const EdgeAccessor &edge, bool is_in_edge, msgs::ExpandOneResultRow &result_row)>;
```
jbajic, line 188: I think these using are duplicated in both cpp and hpp request helper

src/storage/v3/request_helper.hpp
```diff
@@ -99,18 +111,99 @@ class TypedValueVectorCompare final {
...
+template <typename T>
+concept VerticesIt = utils::SameAsAnyOf<T, VerticesIterable, std::vector<VertexAccessor>>;
+
+template <VerticesIt TIterable>
+std::vector<Element<VertexAccessor>> OrderByVertices(DbAccessor &dba, TIterable &iterable,
+                                                     std::vector<msgs::OrderBy> &original_order_bys) {
+  auto order_bys = original_order_bys;
+  auto it_to_remove = std::remove_if(order_bys.begin(), order_bys.end(), [](const auto &order_by) {
+    // We only want to keep OrderBys not impliying edges-ordering
+    return std::string::npos != order_by.expression.expression.find(expr::identifier_edge_symbol);
```
antaljanosbenjamin, line 129: Just leaving a comment to not to forget to draw a conclusion on unified vs separate order bys.

src/storage/v3/request_helper.cpp
```diff
@@ -13,70 +13,518 @@
...
+                                                std::vector<msgs::OrderBy> &order_by_edges,
+                                                const VertexAccessor &vertex_acc) {
+  std::vector<Ordering> ordering;
+  ordering.reserve(order_by_edges.size());
+  std::transform(order_by_edges.begin(), order_by_edges.end(), std::back_inserter(ordering), [](const auto &order_by) {
+    if (memgraph::msgs::OrderingDirection::ASCENDING == order_by.direction) {
+      return Ordering::ASC;
+    }
+    MG_ASSERT(memgraph::msgs::OrderingDirection::DESCENDING == order_by.direction);
+    return Ordering::DESC;
```
antaljanosbenjamin, line 507: Could you please use a `switch` similarly to `OrderByVertices`? Or extract the logic into a function, then we will have the same logic in both places.

antaljanosbenjamin (changes requested):

antaljanosbenjamin (approved):

PR #618, opened 2022-10-25
[master < T1108-MG] Add multiple license support

jbajic:
[master < Task] PR

__Commit message__:
- Make license info available through `LicenseChecker`
- Add `LicenseInfoSender`
- Move license library from utils
- Rename `telemetry_lib` to `mg-telemetry`

### DIFF
--- a/src/license/license.cpp
+++ b/src/license/license.cpp
@@ -177,9 +177,10 @@ void LicenseChecker::RevalidateLicense(const std::string &license_key, const std
   }
 }
 
-void LicenseChecker::EnableTesting() {
+void LicenseChecker::EnableTesting(const LicenseType license_type) {
   enterprise_enabled_ = true;
   is_valid_.store(true, std::memory_order_relaxed);
+  license_type_ = license_type;
   spdlog::info("All Enterprise features are activated for testing.");
 }
 
--- a/src/license/license.cpp
+++ b/src/license/license.cpp
@@ -70,6 +70,17 @@ LicenseCheckResult IsValidLicenseInternal(const License &license, const std::str
 }
 }  // namespace
 
+std::string LicenseTypeToString(const LicenseType license_type) {
+  switch (license_type) {
+    case LicenseType::ENTERPRISE: {
+      return "enterprise";
+    }
+    case LicenseType::OEM: {
+      return "oem";
+    }
+  }
+}
+
 void RegisterLicenseSettings(LicenseChecker &license_checker, utils::Settings &settings) {
   settings.RegisterSetting(std::string{kEnterpriseLicenseSettingKey}, "",
                            [&] { license_checker.RevalidateLicense(settings); });
@@ -154,9 +165,9 @@ void LicenseChecker::RevalidateLicense(const std::string &license_key, const std
   if (!same_license_info) {
     license_type_ = maybe_license->type;
     if (license_type_ == LicenseType::ENTERPRISE) {
-      spdlog::info("All Enterprise features are active.");
+      spdlog::info("Enterprise license is active.");
     } else {
-      spdlog::info("OEM License is active.");
+      spdlog::info("OEM license is active.");
     }
     is_valid_.store(true, std::memory_order_relaxed);
     locked_previous_license_info->is_valid = true;
--- a/src/license/license.cpp
+++ b/src/license/license.cpp
@@ -21,6 +21,7 @@
 
 #include "slk/serialization.hpp"
 #include "utils/base64.hpp"
+#include "utils/cast.hpp"
 #include "utils/exceptions.hpp"
 #include "utils/logging.hpp"
 #include "utils/memory_tracker.hpp"
@@ -224,13 +225,13 @@ std::string LicenseCheckErrorToString(LicenseCheckError error, const std::string
           "following query:\n"
           "SET DATABASE SETTING \"enterprise.license\" TO \"your-license-key\"",
           feature);
-    case LicenseCheckError::LICENSE_INVALID_TYPE:
+    case LicenseCheckError::NOT_ENTERPRISE_LICENSE:
       return fmt::format("Your license has an invalid type. To use {} you need to have an enterprise license. \n",
                          feature);
   }
 }
 
-LicenseCheckResult LicenseChecker::IsEnterpriseEnabled(const utils::Settings &settings) const {
+LicenseCheckResult LicenseChecker::IsEnterpriseValid(const utils::Settings &settings) const {
   if (enterprise_enabled_) [[unlikely]] {
     return {};
   }
@@ -242,7 +243,7 @@ LicenseCheckResult LicenseChecker::IsEnterpriseEnabled(const utils::Settings &se
     return LicenseCheckError::INVALID_LICENSE_KEY_STRING;
   }
   if (maybe_license->type != LicenseType::ENTERPRISE) {
-    return LicenseCheckError::LICENSE_INVALID_TYPE;
+    return LicenseCheckError::NOT_ENTERPRISE_LICENSE;
   }
 
   return IsValidLicenseInternal(*maybe_license, license_info.second);
@@ -257,7 +258,7 @@ utils::Synchronized<std::optional<LicenseInfo>, utils::SpinLock> &LicenseChecker
   return previous_license_info_;
 }
 
-bool LicenseChecker::IsEnterpriseEnabledFast() const {
+bool LicenseChecker::IsEnterpriseValidFast() const {
   return license_type_ == LicenseType::ENTERPRISE && is_valid_.load(std::memory_order_relaxed);
 }
 
@@ -272,7 +273,7 @@ std::string Encode(const License &license) {
   slk::Save(license.organization_name, &builder);
   slk::Save(license.valid_until, &builder);
   slk::Save(license.memory_limit, &builder);
-  slk::Save(static_cast<std::underlying_type_t<LicenseType>>(license.type), &builder);
+  slk::Save(utils::UnderlyingCast(license.type), &builder);
   builder.Finalize();
 
   return std::string{license_key_prefix} + utils::base64_encode(buffer.data(), buffer.size());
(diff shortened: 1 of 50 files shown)


src/memgraph.cpp
```diff
@@ -676,6 +677,19 @@ void InitSignalHandlers(const std::function<void()> &shutdown_fun) {
...
+  return MG_TELEMETRY_ID_OVERRIDE;
+#else
+  // We assume we're on linux and we need to read the machine id from /etc/machine-id
+  const auto machine_id_lines = memgraph::utils::ReadLines("/etc/machine-id");
+  if (machine_id_lines.size() != 1) {
+    return "UNKNOWN";
+  }
+  return machine_id_lines[0];
+#endif
+}
```
gitbuda, line 691: I was thinking about moving this under utils but it has the telemetry-related `ifdef` + it's composed of simple units of work (`Readlines`) so it's fine, but do we add an e2e test that verifies `machine_id` is present (if it's not already there)?

jbajic: I am adding e2e right now, to have verification that everything works.

src/license/license_sender.cpp
```diff
@@ -0,0 +1,57 @@
...
+              {"swap", memory_info.swap},
+              {"memory_limit", license_info->license.memory_limit},
+              {"timestamp", utils::Timestamp::Now().SecWithNsecSinceTheEpoch()}};
+    }
+  });
+
+  if (data.empty()) {
+    return;
+  }
+  requests::RequestPostJson(url_, data,
```
gitbuda, line 53: What is going to happen in the timeout case?

src/license/license.cpp
```diff
@@ -140,7 +141,7 @@ void LicenseChecker::RevalidateLicense(const std::string &license_key, const std
 
   locked_previous_license_info.emplace(license_key, organization_name);
 
-  const auto maybe_license = GetLicense(locked_previous_license_info->license_key);
+  auto maybe_license = GetLicense(locked_previous_license_info->license_key);
```
antaljanosbenjamin, line 144: Why is this non-const anymore?

src/license/license_sender.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include "license/license_sender.hpp"
+
+#include "requests/requests.hpp"
+#include "telemetry/system_info.hpp"
+#include "utils/synchronized.hpp"
+#include "utils/timestamp.hpp"
+
+namespace memgraph::license {
+
+LicenseInfoSender::LicenseInfoSender(std::string url, std::string uuid, std::string machine_id,
```
gitbuda, line 21: Maybe rename the `uuid` to `run_id`, that's more clear to me?

src/license/license.cpp
jbajic: It is being moved in line 175

src/license/license_sender.cpp
jbajic: Nothing for now, I suggest logging that information

src/memgraph.cpp
```diff
@@ -927,6 +944,9 @@ int main(int argc, char **argv) {
       return memgraph::query::plan::CallProcedure::GetAndResetCounters();
     });
   }
+  memgraph::license::LicenseInfoSender license_info_sender(telemetry_server, run_id, machine_id,
+                                                           memgraph::license::global_license_checker.GetLicenseInfo(),
+                                                           std::chrono::seconds(5));
```
gitbuda, line 949: The question is what to put here, we mentioned 24h, but something like 8h seems better to me?

src/memgraph.cpp
```diff
@@ -906,12 +920,15 @@ int main(int argc, char **argv) {
   ServerT server(server_endpoint, &session_data, &context, FLAGS_bolt_session_inactivity_timeout, service_name,
                  FLAGS_bolt_num_workers);
 
+  const auto run_id = memgraph::utils::GenerateUUID();
```
antaljanosbenjamin, line 923: Please remove `Telemetry::GetRunId` as it is not used anymore.

src/license/license.hpp
```diff
@@ -12,26 +12,57 @@
...
 #include "utils/result.hpp"
 #include "utils/scheduler.hpp"
 #include "utils/settings.hpp"
+#include "utils/spin_lock.hpp"
+#include "utils/synchronized.hpp"
 
-namespace memgraph::utils::license {
+namespace memgraph::license {
+
+enum class LicenseType : uint8_t { ENTERPRISE, OEM };
```
antaljanosbenjamin, line 26: Maybe also pass a `LicenseType` parameter to `EnableTesting` to be able specify different license types for tests. It can be defaulted to `LicenseType::ENTERPRISE` to make our life easier.

tests/unit/utils_license.cpp
```diff
@@ -139,3 +137,20 @@ TEST_F(LicenseTest, LicenseInfoOverride) {
...
+  {
+    memgraph::license::License license_entr{organization_name, 0, 0, memgraph::license::LicenseType::ENTERPRISE};
+    const std::string license_key = memgraph::license::Encode(license_entr);
+    license_checker->SetLicenseInfoOverride(license_key, organization_name);
+    CheckLicenseValidity(true);
+  }
+  {
+    memgraph::license::License license_oem{organization_name, 0, 0, memgraph::license::LicenseType::OEM};
+    const std::string license_key = memgraph::license::Encode(license_oem);
+    license_checker->SetLicenseInfoOverride(license_key, organization_name);
```
antaljanosbenjamin, line 153: Maybe use "random numbers" instead of `0`, because it might interfere with the license type in a strange way (faulty serializer, deserializer) and it might not be caught.

antaljanosbenjamin (changes requested):

src/license/license.cpp
```diff
@@ -252,9 +272,10 @@ std::string Encode(const License &license) {
   slk::Save(license.organization_name, &builder);
   slk::Save(license.valid_until, &builder);
   slk::Save(license.memory_limit, &builder);
+  slk::Save(static_cast<std::underlying_type_t<LicenseType>>(license.type), &builder);
```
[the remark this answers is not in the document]
jbajic: Nice

src/license/license.hpp
```diff
@@ -50,18 +81,24 @@ struct LicenseChecker {
   void CheckEnvLicense();
   void SetLicenseInfoOverride(std::string license_key, std::string organization_name);
   void EnableTesting();
-  LicenseCheckResult IsValidLicense(const utils::Settings &settings) const;
-  bool IsValidLicenseFast() const;
+  // Checks if license is valid and if enterprise is enabled
+  LicenseCheckResult IsEnterpriseEnabled(const utils::Settings &settings) const;
+  bool IsEnterpriseEnabledFast() const;
```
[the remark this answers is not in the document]
jbajic: I agree that naming is more intuitive

src/license/license_sender.hpp
```diff
@@ -0,0 +1,48 @@
...
+#include "utils/scheduler.hpp"
+#include "utils/timer.hpp"
+
+namespace memgraph::license {
+
+class LicenseInfoSender final {
+ public:
+  explicit LicenseInfoSender(std::string url, std::string uuid, std::string machine_id,
+                             utils::Synchronized<std::optional<LicenseInfo>, utils::SpinLock> &license_info,
+                             std::chrono::duration<int64_t> request_frequency = std::chrono::hours(12));
```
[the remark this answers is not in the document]
jbajic: I will switch to hours since that is the magnitude of frequency we want

src/license/license_sender.hpp
```diff
@@ -0,0 +1,48 @@
...
+#include "license/license.hpp"
+#include "utils/scheduler.hpp"
+#include "utils/timer.hpp"
+
+namespace memgraph::license {
+
+class LicenseInfoSender final {
+ public:
+  explicit LicenseInfoSender(std::string url, std::string uuid, std::string machine_id,
+                             utils::Synchronized<std::optional<LicenseInfo>, utils::SpinLock> &license_info,
```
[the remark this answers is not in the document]
jbajic: Unfortunately, our `SpinLock` does not support shared lock, do you think it makes sense to extend it now?

src/memgraph.cpp
jbajic: I have moved it to the utils, Not sure if it makes sense to add tests for this system info stuff, what do you think, not sure how to confirm correct results or mock this stuff.

gitbuda: Yea it's platform-specific, I was thinking about that, maybe we can add a simple unit test that can be run on each platform we are building for (because even on different Linux machines this can be different + once we'll add support for Win/Mac, we'll also have to make this code multi-platform.

I think for now it's not worth investing much time into testing / cross-platform support, we'll get there.

src/license/license_sender.cpp
gitbuda: In this case, I think nothing is ok, to not distract the user, maybe put the log under the `TRACE` level but I would say nothing...

src/license/license_sender.hpp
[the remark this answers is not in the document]
antaljanosbenjamin: Yes, you are right and no, it doesn't make sense in my opinion.

src/license/license.cpp
```diff
@@ -177,9 +177,10 @@ void LicenseChecker::RevalidateLicense(const std::string &license_key, const std
   }
 }
 
-void LicenseChecker::EnableTesting() {
+void LicenseChecker::EnableTesting(const LicenseType license_type) {
   enterprise_enabled_ = true;
   is_valid_.store(true, std::memory_order_relaxed);
+  license_type_ = license_type;
```
antaljanosbenjamin, line 183: You will hate me, but the log on the next line is super misleading if the license type is OEM. Could you please rephrase that? Nothing fancy is necessary, just generalize it: "The specified license type is set for testing", or something something that is true for  both cases.

antaljanosbenjamin (approved):
Just one very small comment and the PR is good to go!

gitbuda (approved):
Seems robust 😄

PR #636, opened 2022-11-02
[< T298-MAGE] allShortest bugfix

brunos252:
All shortest paths wasn't working when using nodes called by their internal id, because in that case Operator Filter isn't called and all shortest paths to target node were received, instead of just the ones starting in the source node. Example of query not working:
```
MATCH path=(source)-[rel *ALLSHORTEST (r,n | 1)]-(target)
WHERE id(source)=5 AND id(target)=6
RETURN path;
```
while this worked:
```
MATCH path=(source)-[rel *ALLSHORTEST (r,n | 1)]-(target)
WHERE source.name="n5" AND target.name="n6"
RETURN path;
```
Which has now been fixed by addressing the `self_.common_.existing_node` flag.

The test was put inside `interpreter.cpp` unit tests, in order to call the query by referencing internal ids.

[master < Task] PR


src/query/plan/operator.cpp
```diff
@@ -1986,6 +1986,14 @@ class ExpandAllShortestPathsCursor : public query::plan::Cursor {
         }
 
         if ((current_weight > visited_cost_.at(next_vertex)).ValueBool()) continue;
+
+        // Place destination node on the frame, handle existence flag
+        if (self_.common_.existing_node) {
+          const auto &node = frame[self_.common_.node_symbol];
+          if ((node != TypedValue(next_vertex, pull_memory)).ValueBool()) continue;
```
antaljanosbenjamin, line 1993: It is a slight change, but you can avoid constructing a `TypedValue` by:
```suggestion
          const auto &node = frame[self_.common_.node_symbol];
          ExpectType(self_.common_.node_symbol, existing_node, TypedValue::Type::Vertex);
          if (node.ValueVertex() != next_vertex) continue;
```
Unless I am missing something, in that case please let me know. Also, please format the code before committing, I made the changes on github, so the formatting might be slightly off.

antaljanosbenjamin: And if my solution works, then please also remove the `pull_memory` variable.

antaljanosbenjamin (changes requested):
Could you please also add a test? Check `tests/unit/interpreter.cpp` to get to know how you can create unit tests easily.

jbajic (approved):

kostasrim (approved):

antaljanosbenjamin (approved):

PR #638, opened 2022-11-03
[:pineapple: < MG-T1148] Fix expand one source vertex pk setting

jbajic:
[master < Task] PR


src/storage/v3/shard_rsm.cpp
```diff
@@ -325,7 +337,7 @@ std::optional<msgs::ExpandOneResultRow> GetExpandOneResult(
   const auto primary_key = ConvertPropertyVector(std::move(src_vertex.second));
   auto v_acc = acc.FindVertex(primary_key, View::NEW);
 
-  auto source_vertex = FillUpSourceVertex(v_acc, req, src_vertex);
+  auto source_vertex = FillUpSourceVertex(v_acc, req);
```
antaljanosbenjamin, line 340: What if instead of doing a `PropertyValue` -> `Value` conversion in `FillUpSourceVertex`, we just copy the primary key of the vertex on line 337 instead of moving it? Then it should be possible to use the passed in value to populate the result?

`FillUpSourceVertex` is already a bit tricky, because it only populates the vertex id and secondary labels. Instead we could populate the vertex id from the value passed in, and then the `FillUpSourceVertex` function could be transformed to `FillUpSourceVertexSecondaryLabels`. This way we don't have to pass the vertex id into a function and the function also return something that is complete on its own level.

antaljanosbenjamin:
And the most important thing: please extend the existing e2e tests to make sure nobody breaks this.

gvolfing (approved):

src/storage/v3/shard_rsm.cpp
```diff
@@ -334,13 +321,14 @@ std::optional<msgs::ExpandOneResultRow> GetExpandOneResult(
     const EdgeUniqunessFunction &maybe_filter_based_on_edge_uniquness, const EdgeFiller &edge_filler,
     const Schemas::Schema *schema) {
   /// Fill up source vertex
-  const auto primary_key = ConvertPropertyVector(std::move(src_vertex.second));
+  const auto primary_key = ConvertPropertyVector(src_vertex.second);
   auto v_acc = acc.FindVertex(primary_key, View::NEW);
 
-  auto source_vertex = FillUpSourceVertex(v_acc, req);
+  auto source_vertex = FillUpSourceVertexSecondaryLabels(v_acc, req);
```
antaljanosbenjamin, line 327: I wasn't clear enough, sorry for that. I meant that `FillUpSourceVertexSecondaryLabels` could return a `std::optional<std::vector<msgs::Label>>`, so it will be complete in sense of it doesn't return a partially populated `msgs::Vertex`, but a full list of secondary labels. Of course for that we have to adjust how we call it, similarly to `FillUpSourceVertexProperties`:
```
  msgs::Vertex source_vertex{.id = std::move(src_vertex)};
  if (auto maybe_secondary_labels = FillUpSourceVertexSecondaryLabels(v_acc, req); maybe_secondary_labels.has_value()) {
    source_vertex.labels = std::move(*maybe_secondary_labels);
  } else {
    return std::nullopt;
  }
```

antaljanosbenjamin (approved):

PR #639, opened 2022-11-04
[:pineapple:  < T1145MG] Handle insertions of vertices that already exist in a better way

gvolfing:
[master < Task] PR
- [ ] Provide final commit message


src/storage/v3/shard_operation_result.hpp
```diff
@@ -0,0 +1,28 @@
...
+#include <variant>
+
+#include "storage/v3/result.hpp"
+#include "storage/v3/schema_validator.hpp"
+
+namespace memgraph::storage::v3 {
+
+struct AlreadyInsertedElement {};
+
+using ResultErrorType = std::variant<SchemaViolation, AlreadyInsertedElement, Error>;
```
antaljanosbenjamin, line 23: Why not just to extend the `Error` type?

antaljanosbenjamin (approved):

jbajic (approved):

PR #643, opened 2022-11-08
[master < MG-python-check] Add python check

jbajic:
[master < Task] PR


.github/workflows/diff.yaml
```diff
@@ -75,14 +75,30 @@ jobs:
...
+      - name: Changed Files
+        uses: tj-actions/changed-files@v34.3.2
+
+      - name: Python code analysis
+        run: |
+          for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
+            if [[ $file == *.py ]]; then
+              black $file --check --diff
+            fi
+          done
```
gitbuda, line 95: Move this into a script, so that it can also be run locally? Maybe include options to analyze and fix the code in-place?

jbajic: I am against this since it is really trivial to run black all one needs to do is `black .`, or have a `pre-commit` hook installed, this is just a CI verification all is correct. This part is only for checking the format of all changed files, locally `steps.changed-files.outputs.all_changed_files` does not exist, and --check and --diff flags are useless locally since they only inform what is wrong. What do you say?

.github/workflows/diff.yaml
```diff
@@ -75,14 +75,30 @@ jobs:
...
         run: |
           # Activate toolchain.
           source /opt/toolchain-v4/activate
 
           # Initialize dependencies.
           ./init
 
+      - name: Python code analysis
+        run: |
+          CHANGED_FILES=$(git diff -U0 HEAD~1... --name-only)
```
antaljanosbenjamin, line 88: Wouldn't this only contain the diffs from the latest commit? I think we should use something similar to how we run `clang-tidy`: `git diff -U0 ${{ env.BASE_BRANCH }}...`. Check below how it is used.

42jeremy (approved):

antaljanosbenjamin (approved):

PR #644, opened 2022-11-08
[:pineapple: < T1159-MG] Add memgraph functions lib

kostasrim:
[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR


gitbuda:
@kostasrim there should be some unit tests for these functions in the `tests/unit`, could you check them, test the `mg-fucntions`, and port these tests to `project-pineapples`?

tests/unit/query_v2_expression_evaluator.cpp
```diff
@@ -0,0 +1,1786 @@
...
+using memgraph::msgs::ScanVerticesResponse;
+using ShardMap = memgraph::coordinator::ShardMap;
+using LabelId = memgraph::storage::v3::LabelId;
+using PropertyId = memgraph::storage::v3::PropertyId;
+using memgraph::coordinator::Shards;
+using CompoundKey = memgraph::coordinator::PrimaryKey;
+using memgraph::expr::ExpressionRuntimeException;
+using memgraph::functions::FunctionRuntimeException;
+
+namespace {
```
jbajic, line 70: And you can drop all memgraph::query::v2 namespacing below
```suggestion
namespace memgraph::query::v2::tests {
```

src/functions/awesome_memgraph_functions.hpp
```diff
@@ -0,0 +1,1465 @@
...
+  FType<TypedValueT, Or<Null, Vertex, Edge>>("id", args, nargs);
+  const auto &arg = args[0];
+  if (arg.IsNull()) {
+    return TypedValueT(ctx.memory);
+  }
+  if (arg.IsVertex()) {
+    return TypedValueT(static_cast<int64_t>(arg.ValueVertex().CypherId()), ctx.memory);
+  }
+  return TypedValueT(static_cast<int64_t>(arg.ValueEdge().CypherId()), ctx.memory);
+}
```
antaljanosbenjamin, line 1036: I think the current solution is not good, because it adds a dummy function to both accessors. I think for both accessors there are better solutions.
For vertices:
1. Remove it completely. This was necessary to make `ScanById` possible, so targeting a single vertex. In the distributed case the same can be achieved with a `MATCH`. The users has to "unfold" the properties, but that is something we can live for now.
2. Make it return a map that is in a format of `{primary_label: "PrimaryLabelName", key: ["pk_value1", "pk_value2", 4, 5]}`. This way the same semantic can be kept as before.

For edges I think we should use the `Gid` as before.

src/query/v2/conversions.hpp
```diff
@@ -56,6 +56,10 @@ inline TypedValue ValueToTypedValue(const msgs::Value &value, msgs::ShardRequest
   throw std::runtime_error("Incorrect type in conversion");
 }
 
+inline const auto ValueToTypedValueFunctor = [](const msgs::Value &value, msgs::ShardRequestManagerInterface *manager) {
+  return ValueToTypedValue(value, manager);
+};
```
antaljanosbenjamin, line 61: Isn't `detail::Callable` does the same in `src/query/v2/bindings/eval.hpp`? If so, then please don't duplicate the same functionality.

src/storage/v3/conversions.hpp
```diff
@@ -69,6 +69,10 @@ TTypedValue PropertyToTypedValue(const PropertyValue &value) {
   LOG_FATAL("Unsupported type");
 }
 
+template <typename TypedValueT>
+inline const auto PropertyToTypedValueFunctor =
+    [](const PropertyValue &value) { return PropertyToTypedValue<TypedValueT>(value); };
+
```
antaljanosbenjamin, line 75: Isn't `PropertyToTypedValueConverter` does the same?

src/storage/v3/edge_accessor.cpp
```diff
@@ -179,4 +179,7 @@ Result<std::map<PropertyId, PropertyValue>> EdgeAccessor::Properties(View view)
   return std::move(properties);
 }
 
+// NOLINTNEXTLINE(readability-convert-member-functions-to-static)
+size_t EdgeAccessor::CypherId() const { return 10; }
```
antaljanosbenjamin, line 183: Return the global id of the edge.

tests/e2e/distributed_queries/awesome_memgraph_functions.py
```diff
@@ -0,0 +1,90 @@
...
+import pytest
+
+from common import *
+
+
+def test_vertex_creation_and_scanall(connection):
+    wait_for_shard_manager_to_initialize()
+    cursor = connection.cursor()
+
+    assert has_n_result_row(cursor, "CREATE (n :label {property:1})", 0)
```
gitbuda, line 26: @kostasrim since e.g. `VertexAccessor` returns 0 for `In/OutDegree`, I think it makes sense in this test to add real test cases for each function but somehow disable them. As we move on we'll implement some of these functionalities + when we reach the point where the release will be close, we just enable all the tests and observe/fix what is not working 🤔 

What do you think about that idea?

src/storage/v3/vertex_accessor.hpp
```diff
@@ -118,6 +118,9 @@ class VertexAccessor final {
   }
   bool operator!=(const VertexAccessor &other) const noexcept { return !(*this == other); }
 
+  // Dummy function
+  size_t CypherId() const;
```
antaljanosbenjamin, line 122: My issue with this is we are not planning to add a global id to vertices, so this is problematic in my opinion. I think it would be better not adding this dummy function at all.

tests/unit/query_v2_expression_evaluator.cpp
```diff
@@ -0,0 +1,1786 @@
...
+#include "query/v2/frontend/ast/ast.hpp"
+#include "query/v2/requests.hpp"
+#include "query/v2/shard_request_manager.hpp"
+#include "storage/v3/property_value.hpp"
+#include "storage/v3/storage.hpp"
+#include "utils/exceptions.hpp"
+#include "utils/string.hpp"
+
+#include "query_v2_query_common.hpp"
+//#include "utils/temporal.hpp"
```
antaljanosbenjamin, line 39:
```suggestion
#include "utils/string.hpp"
#include "query_v2_query_common.hpp"
```

Of course, please format it before committing.

tests/unit/query_v2_expression_evaluator.cpp
```diff
@@ -0,0 +1,1786 @@
...
+    auto *op1 = storage.Create<SubscriptOperator>(vertex_id, storage.Create<PrimitiveLiteral>("prop"));
+    auto value1 = Eval(op1);
+    EXPECT_EQ(value1.ValueInt(), 42);
+
+    auto *op2 = storage.Create<SubscriptOperator>(edge_id, storage.Create<PrimitiveLiteral>("prop"));
+    auto value2 = Eval(op2);
+    EXPECT_EQ(value2.ValueInt(), 43);
+  }
+  // TODO(kostasrim) Investigate
+  //  Shall we return null on missing properties? Or shall we throw bad optional access as we do now?
```
antaljanosbenjamin, line 579: I think this is obsolete.

antaljanosbenjamin (changes requested):
I will finish the review of the unit tests in the afternoon.

tests/e2e/distributed_queries/awesome_memgraph_functions.py
gitbuda: No, no, in this case, it wouldn't be commented code, it would be legit test cases just not executed in the CI (runtime config for these tests). I understand these two are very very similar, but I would like to avoid the situation to miss some of the required functionality, I think these e2e tests are a perfect spot to ensure we don't miss functionality. I agree QE work is more important, but since we are now covering these functions, we/you have the whole context, later the whole thing we'll have to be planned again, it's very likely that we'll miss something, etc...

tests/unit/query_v2_expression_evaluator.cpp
```diff
@@ -0,0 +1,1786 @@
...
+                                                            const memgraph::msgs::ShardRequestManagerInterface *manager,
+                                                            Label label = {}) {
+  static size_t id = 0;
+  return {Vertex{VertexId{label, id++}, {label}}, std::move(props), manager};
+}
+
+memgraph::query::v2::accessors::EdgeAccessor CreateEdge(std::vector<std::pair<PropertyId, Value>> props,
+                                                        const memgraph::msgs::ShardRequestManagerInterface *manager) {
+  auto edge = Edge{.src = VertexId{{}, 0},
+                   .dst = VertexId{{}, 0},
```
antaljanosbenjamin, line 555: If I am not mistaken, then `0` initializes the `PrimaryKey` part of the `VertexId`. It is a bit confusing to me, usually we use `{}` to initialize a vector.

tests/unit/query_v2_expression_evaluator.cpp
```diff
@@ -0,0 +1,1786 @@
...
+using EdgeType = memgraph::msgs::EdgeType;
+using Value = memgraph::msgs::Value;
+using VertexId = memgraph::msgs::VertexId;
+using Label = memgraph::msgs::Label;
+
+memgraph::query::v2::accessors::VertexAccessor CreateVertex(std::vector<std::pair<PropertyId, Value>> props,
+                                                            const memgraph::msgs::ShardRequestManagerInterface *manager,
+                                                            Label label = {}) {
+  static size_t id = 0;
+  return {Vertex{VertexId{label, id++}, {label}}, std::move(props), manager};
```
antaljanosbenjamin, line 549: Is this really does what you expect? I think `id++` will trigger [the 3rd ctor of vector](https://en.cppreference.com/w/cpp/container/vector/vector), which will fill up the primary key with `id++` number of `Null` values.

tests/unit/query_v2_expression_evaluator.cpp
```diff
@@ -0,0 +1,1786 @@
...
+
+// TODO(kostasrim) These will fail because of memory resource not propagating correctly. This should be done as part of
+// polishing the allocators.
+// TEST_F(ExpressionEvaluatorPropertyLookup, Vertex) {
+//   auto v1 = CreateVertex({{prop_age.second, memgraph::msgs::Value(static_cast<int64_t>(32))}}, shard_manager.get());
+//   frame[symbol] = TypedValue(v1);
+//   EXPECT_EQ(Value(prop_age).ValueInt(), 10);
+//   EXPECT_TRUE(Value(prop_height).IsNull());
+// }
+//  TEST_F(ExpressionEvaluatorPropertyLookup, Edge) {}
```
antaljanosbenjamin, line 1146: Could you please create a task from this just to make sure we handle the allocators properly?

tests/unit/query_v2_expression_evaluator.cpp
```diff
@@ -0,0 +1,1786 @@
...
+  //  ASSERT_EQ(EvaluateFunction("VALUETYPE", p).ValueString(), "PATH");
+}
+
+TEST_F(FunctionTest, Labels) {
+  ASSERT_THROW(EvaluateFunction("LABELS"), FunctionRuntimeException);
+  ASSERT_TRUE(EvaluateFunction("LABELS", TypedValue()).IsNull());
+  Label label{shard_manager->NameToLabel("label1")};
+  auto v = CreateVertex({}, shard_manager.get(), std::move(label));
+  std::vector<std::string> labels;
+  auto _labels = EvaluateFunction("LABELS", v).ValueList();
```
antaljanosbenjamin, line 1371: Please don't use underscore as prefix because that can be a [reserved name](https://stackoverflow.com/a/228797/6639989) in the global namespace.

tests/unit/query_v2_expression_evaluator.cpp
```diff
@@ -0,0 +1,1786 @@
...
+  labels.reserve(_labels.size());
+  for (auto label : _labels) {
+    labels.emplace_back(label.ValueString());
+  }
+  ASSERT_THAT(labels, UnorderedElementsAre("label1"));
+  ASSERT_THROW(EvaluateFunction("LABELS", 2), FunctionRuntimeException);
+}
+
+// TODO(kostasrim) Enable this once we fix accessors Path
+// TEST_F(FunctionTest, NodesRelationships) {}
```
antaljanosbenjamin, line 1381: Please remove that commented out function body. Also, if we cannot test a function, isn't it better to remove the whole function?

tests/unit/query_v2_expression_evaluator.cpp
```diff
@@ -0,0 +1,1786 @@
...
+// TEST_F(FunctionTest, Id) {
+//  auto v = CreateVertex({}, shard_manager.get());
+//  auto e = CreateEdge({}, shard_manager.get());
+//  EXPECT_TRUE(EvaluateFunction("ID", TypedValue()).IsNull());
+//  EXPECT_EQ(EvaluateFunction("ID", v).ValueInt(), 0);
+//  EXPECT_EQ(EvaluateFunction("ID", e).ValueInt(), 0);
+//  EXPECT_THROW(EvaluateFunction("ID"), FunctionRuntimeException);
+//  EXPECT_THROW(EvaluateFunction("ID", 0), FunctionRuntimeException);
+//  EXPECT_THROW(EvaluateFunction("ID", v, e), FunctionRuntimeException);
+//}
```
antaljanosbenjamin, line 1619: Please remove commented out code.

tests/unit/storage_v3_edge.cpp
```diff
@@ -2450,8 +2450,8 @@ TEST_P(StorageEdgeTest, EdgeDeleteFromSmallerCommit) {
 //       ASSERT_EQ(*vertex_to->InDegree(View::OLD), 1);
 //       auto e = edges[0];
 //       ASSERT_EQ(e.EdgeType(), et);
-//       ASSERT_EQ(e.FromVertex(), *vertex_from);
-//       ASSERT_EQ(e.ToVertex(), *vertex_to);
+//       ASSERT_EQ(e.From(), *vertex_from);
+//       ASSERT_EQ(e.To(), *vertex_to);
```
antaljanosbenjamin, line 2454: Feel free to delete these commented out tests.

antaljanosbenjamin (changes requested):

src/expr/interpret/eval.hpp
```diff
@@ -479,7 +480,7 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
   }
 
   TypedValue Visit(Function &function) override {
-    FunctionContext function_ctx{dba_, ctx_->memory, ctx_->timestamp, &ctx_->counters, view_};
+    functions::FunctionContext<DbAccessor> function_ctx{dba_, ctx_->memory, ctx_->timestamp, &ctx_->counters, view_};
```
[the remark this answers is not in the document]
kostasrim: I am not a big fun of `CTAD` so usually I avoid it. If you have a strong preference I am happy to use it, although I think it would fail in this particular case.

src/functions/awesome_memgraph_functions.hpp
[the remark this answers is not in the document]
kostasrim: for vertices I removed it

src/functions/awesome_memgraph_functions.hpp
```diff
@@ -0,0 +1,1465 @@
...
+    case TypedValueT::Type::LocalDateTime:
+      return TypedValueT("LOCAL_DATE_TIME", ctx.memory);
+    case TypedValueT::Type::Duration:
+      return TypedValueT("DURATION", ctx.memory);
+  }
+}
+
+// TODO: How is Keys different from Properties function?
+template <typename TypedValueT, typename FunctionContextT, typename Tag>
+TypedValueT Keys(const TypedValueT *args, int64_t nargs, const FunctionContextT &ctx) {
```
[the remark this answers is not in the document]
kostasrim: removed

src/functions/awesome_memgraph_functions.hpp
```diff
@@ -0,0 +1,1465 @@
...
+  if (function_name == "TIMESTAMP") return functions::impl::Timestamp<TypedValueT, FunctionContextT>;
+  if (function_name == "TOBOOLEAN") return functions::impl::ToBoolean<TypedValueT, FunctionContextT>;
+  if (function_name == "TOFLOAT") return functions::impl::ToFloat<TypedValueT, FunctionContextT>;
+  if (function_name == "TOINTEGER") return functions::impl::ToInteger<TypedValueT, FunctionContextT>;
+  if (function_name == "TYPE") return functions::impl::Type<TypedValueT, FunctionContextT>;
+  if (function_name == "VALUETYPE") return functions::impl::ValueType<TypedValueT, FunctionContextT>;
+
+  // List functions
+  if (function_name == "KEYS") return functions::impl::Keys<TypedValueT, FunctionContextT, Tag>;
+  if (function_name == "LABELS") return functions::impl::Labels<TypedValueT, FunctionContextT, Tag>;
```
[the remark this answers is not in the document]
kostasrim: I think they were grouped by return type. (The comments in the code are not mine) but the list is on par with the one mentioned in our docs https://memgraph.com/docs/cypher-manual/functions

src/query/v2/conversions.hpp
kostasrim: Yes but unfortunately that file depends on the generated ast which will blow up if you include it. I will figure out some other way but not in this PR. If you have a better idea let me know!

src/expr/interpret/eval.hpp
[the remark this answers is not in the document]
jbajic: I don't

src/query/v2/accessors.hpp
```diff
@@ -136,9 +136,11 @@ class VertexAccessor final {
...
   //  }
 
-  //  storage::Result<size_t> InDegree(storage::View view) const { return impl_.InDegree(view); }
-  //
-  //  storage::Result<size_t> OutDegree(storage::View view) const { return impl_.OutDegree(view); }
+  // NOLINTNEXTLINE(readability-convert-member-functions-to-static)
+  [[nodiscard]] size_t InDegree() const { return 0; }
+
+  // NOLINTNEXTLINE(readability-convert-member-functions-to-static)
+  [[nodiscard]] size_t OutDegree() const { return 0; }
```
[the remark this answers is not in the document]
antaljanosbenjamin: Then throw a `NotYetImplemented` error, but definitely do not return dummy values.

src/query/v2/shard_request_manager.hpp
```diff
@@ -129,6 +129,9 @@ class ShardRequestManagerInterface {
   virtual const std::string &PropertyToName(memgraph::storage::v3::PropertyId prop) const = 0;
   virtual const std::string &LabelToName(memgraph::storage::v3::LabelId label) const = 0;
   virtual const std::string &EdgeTypeToName(memgraph::storage::v3::EdgeTypeId type) const = 0;
+  virtual std::optional<storage::v3::PropertyId> MaybeNameToProperty(const std::string &name) const = 0;
+  virtual std::optional<storage::v3::EdgeTypeId> MaybeNameToEdge(const std::string &name) const = 0;
```
antaljanosbenjamin, line 133:
```suggestion
  virtual std::optional<storage::v3::EdgeTypeId> MaybeNameToEdgeType(const std::string &name) const = 0;
```

src/functions/awesome_memgraph_functions.hpp
```diff
@@ -1026,12 +966,9 @@ template <typename TypedValueT, typename FunctionContextT>
...
   const auto &arg = args[0];
-  if (arg.IsNull()) {
+  if (arg.IsNull() || arg.IsVertex()) {
     return TypedValueT(ctx.memory);
   }
-  if (arg.IsVertex()) {
-    return TypedValueT(static_cast<int64_t>(arg.ValueVertex().CypherId()), ctx.memory);
-  }
   return TypedValueT(static_cast<int64_t>(arg.ValueEdge().CypherId()), ctx.memory);
 }
```
antaljanosbenjamin, line 967: Why not to modify the `FType` on line 967 to simply "exclude" the vertices from this function?

src/functions/awesome_memgraph_functions.hpp
```diff
@@ -1449,16 +1388,26 @@ NameToFunction(const std::string &function_name) {
 
   // Memgraph specific functions
   if (function_name == "ASSERT") return functions::impl::Assert<TypedValueT, FunctionContextT>;
-  if (function_name == "COUNTER") return functions::impl::Counter<TypedValueT, FunctionContextT>;
   if (function_name == "TOBYTESTRING") return functions::impl::ToByteString<TypedValueT, FunctionContextT>;
   if (function_name == "FROMBYTESTRING") return functions::impl::FromByteString<TypedValueT, FunctionContextT>;
+  // Only on QE
+  if constexpr (std::is_same_v<Tag, QueryEngineTag>) {
+    if (function_name == "COUNTER") return functions::impl::Counter<TypedValueT, FunctionContextT>;
+  }
```
antaljanosbenjamin, line 1390: `COUNTER` is added twice, please remove this.

antaljanosbenjamin (changes requested):

src/query/v2/conversions.hpp
antaljanosbenjamin: `detail::Callable` doesn't depend on any generated thing, so it can be extracted to a file from which you can include in both places.

src/storage/v3/conversions.hpp
antaljanosbenjamin: This is not addresssed.

tests/unit/query_v2_expression_evaluator.cpp
```diff
@@ -0,0 +1,1726 @@
...
+#include <iterator>
+#include <memory>
+#include <unordered_map>
+#include <vector>
+
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+#include "coordinator/shard_map.hpp"
+#include "exceptions.hpp"
```
antaljanosbenjamin, line 23: Please use full path.

tests/unit/query_v2_expression_evaluator.cpp
```diff
@@ -0,0 +1,1726 @@
...
+accessors::VertexAccessor CreateVertex(std::vector<std::pair<PropertyId, Value>> props,
+                                       const memgraph::msgs::ShardRequestManagerInterface *manager, Label label = {}) {
+  static int64_t id = 0;
+  return {Vertex{VertexId{label, {memgraph::msgs::Value(id++)}}, {label}}, std::move(props), manager};
+}
+
+accessors::EdgeAccessor CreateEdge(std::vector<std::pair<PropertyId, Value>> props,
+                                   const memgraph::msgs::ShardRequestManagerInterface *manager) {
+  auto edge = Edge{.src = VertexId{{}, {}},
+                   .dst = VertexId{{}, {}},
```
antaljanosbenjamin, line 557: As you have tests that expecting the value of src/dst vertices, please differentiate between them somehow.

tests/unit/query_v2_expression_evaluator.cpp
```diff
@@ -0,0 +1,1726 @@
...
+            3);
+  ASSERT_THROW(EvaluateFunction("SIZE", 5), FunctionRuntimeException);
+}
+
+TEST_F(FunctionTest, StartNode) {
+  ASSERT_THROW(EvaluateFunction("STARTNODE"), FunctionRuntimeException);
+  ASSERT_TRUE(EvaluateFunction("STARTNODE", TypedValue()).IsNull());
+  auto v1 = CreateVertex({}, shard_manager.get());
+  auto e = CreateEdge({}, shard_manager.get());
+  const auto expected = VertexId{{}, 0};
```
antaljanosbenjamin, line 1266: Please also use `{}` because somebody might tricked to think `0` is some sort of id or something.

tests/unit/query_v2_expression_evaluator.cpp
```diff
@@ -0,0 +1,1726 @@
...
+static TypedValue MakeTypedValueList(TArgs &&...args) {
+  return TypedValue(std::vector<TypedValue>{TypedValue(args)...});
+}
+
+TEST_F(FunctionTest, EndNode) {
+  ASSERT_THROW(EvaluateFunction("ENDNODE"), FunctionRuntimeException);
+  ASSERT_TRUE(EvaluateFunction("ENDNODE", TypedValue()).IsNull());
+  auto v1 = CreateVertex({}, shard_manager.get());
+  auto e = CreateEdge({}, shard_manager.get());
+  const auto expected = VertexId{{}, 0};
```
antaljanosbenjamin, line 1196: Same here.

src/storage/v3/conversions.hpp
antaljanosbenjamin: This is not addressed.

kostasrim: As discussed these two functions are not identical (the one in the bindings contain an extra overload) so we decided to keep it as is.

tests/e2e/distributed_queries/awesome_memgraph_functions.py
kostasrim: I decided not to add more tests because:
1. I think the current functionality is tested well.
2. I would not like to have disabled tests as any potential change might require modifications to the disabled tests which requires the reviewer or the PR implemented to recall that such disabled tests exist.

tests/e2e/distributed_queries/awesome_memgraph_functions.py
```diff
@@ -0,0 +1,90 @@
...
+# License, and you may not use this file except in compliance with the Business Source License.
+#
+# As of the Change Date specified in that file, in accordance with
+# the Business Source License, use of this software will be governed
+# by the Apache License, Version 2.0, included in the file
+# licenses/APL.txt.
+
+import sys
+import time
+import typing
```
jbajic, line 14: Unused

tests/e2e/distributed_queries/awesome_memgraph_functions.py
```diff
@@ -0,0 +1,90 @@
...
+
+import mgclient
+import pytest
+
+from common import *
```
jbajic, line 19: Could you please import explicitly

jbajic (approved):
Just a couple of small comments!

src/query/v2/conversions.hpp
antaljanosbenjamin: `detail::Callable` does more thing that this function, so it is okay to not to unify them.

antaljanosbenjamin (approved):
I think this PR didn't make it worse, but we have to ensure that our PMR works properly. What does it mean is in the query engine we should always require a memory resource for `TypedValue` construction.

PR #649, opened 2022-11-14
[:pineapple: < MG-T0995] Implement top level error handling in storage

jbajic:
[master < Task] PR


src/storage/v3/result.hpp
```diff
@@ -19,16 +22,73 @@ namespace memgraph::storage::v3 {
...
+  // Schema Violations
+  SCHEMA_NO_SCHEMA_DEFINED_FOR_LABEL,
+  SCHEMA_VERTEX_PROPERTY_WRONG_TYPE,
+  SCHEMA_VERTEX_UPDATE_PRIMARY_KEY,
+  SCHEMA_VERTEX_UPDATE_PRIMARY_LABEL,
+  SCHEMA_VERTEX_SECONDARY_LABEL_IS_PRIMARY,
+  SCHEMA_VERTEX_PRIMARY_PROPERTIES_UNDEFINED,
+
+  OBJECT_NOT_FOUND,
+};
```
jbajic, line 41: Move this to common

src/expr/interpret/eval.hpp
```diff
@@ -385,7 +407,7 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
             typename TReturnType = std::enable_if_t<std::is_same_v<TTag, StorageTag>, bool>>
   TReturnType HasLabelImpl(const VertexAccessor &vertex, const LabelIx &label, StorageTag /*tag*/) {
     auto has_label = vertex.HasLabel(view_, GetLabel(label));
-    if (has_label.HasError() && has_label.GetError() == Error::NONEXISTENT_OBJECT) {
+    if (has_label.HasError() && has_label.GetError().code == Error::NONEXISTENT_OBJECT) {
```
antaljanosbenjamin, line 410: Maybe you can define the `operator==` for a `ShardError` and `ErrorCode`, so the `.code` part doesn't have to be written out. Of course that means the `ShardError` should be also compared only based on `code` and not the rest. That is questionable move, but it make sense in my opinion, as comparing `ShardError`s is a weird thing to do.

src/expr/interpret/eval.hpp
```diff
@@ -100,6 +100,28 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
 #undef BINARY_OPERATOR_VISITOR
 #undef UNARY_OPERATOR_VISITOR
 
+  void HandleShardError(Error &shard_error, const std::string_view accessed_object) {
```
antaljanosbenjamin, line 103: I think this name is very generic. Maybe consider making it more specific to read operations.

src/storage/v3/result.hpp
```diff
@@ -11,24 +11,36 @@
...
+  common::ErrorCode code;
+  // TODO Maybe add category
+  std::string message;
+  std::string source;
+
+  inline friend bool operator==(const ShardError &lhs, const ShardError &rhs) { return lhs.code == rhs.code; }
 };
 
+// NOLINTNEXTLINE(cppcoreguidelines-macro-usage)
+#define SHARD_ERROR(...) memgraph::storage::v3::ShardError(__VA_ARGS__, fmt::format("{}:{}", __FILE__, __LINE__))
```
antaljanosbenjamin, line 41: Maybe use (`std::source_location`)[https://en.cppreference.com/w/cpp/experimental/source_location]? We can store that also in the error. I know I mentioned the `std::string` source, but I just realized that now `source_location` is a superior option. Sorry for that.

src/glue/v2/communication.hpp
```diff
@@ -35,7 +35,7 @@ namespace memgraph::glue::v2 {
 /// @param storage::v3::View for deciding which vertex attributes are visible.
 ///
 /// @throw std::bad_alloc
-storage::v3::Result<communication::bolt::Vertex> ToBoltVertex(
+storage::v3::ShardResult<communication::bolt::Vertex> ToBoltVertex(
```
antaljanosbenjamin, line 38: Does it make sense to still use shard result here? The data pumped into these functions are already resigning on the query engine, thus the shard result doesn't make sense in my opinion here. I know it is something that you haven't modified technically, but I vote for removing things instead of carrying over a flaw.

src/query/v2/plan/operator.cpp
```diff
@@ -566,17 +567,24 @@ UniqueCursorPtr ScanAllById::MakeCursor(utils::MemoryResource *mem) const {
...
+      case common::ErrorCode::SERIALIZATION_ERROR:
+      case common::ErrorCode::PROPERTIES_DISABLED:
         throw QueryRuntimeException("Unexpected error when accessing relationships.");
+      case common::ErrorCode::SCHEMA_NO_SCHEMA_DEFINED_FOR_LABEL:
+      case common::ErrorCode::SCHEMA_VERTEX_PROPERTY_WRONG_TYPE:
+      case common::ErrorCode::SCHEMA_VERTEX_UPDATE_PRIMARY_KEY:
+      case common::ErrorCode::SCHEMA_VERTEX_UPDATE_PRIMARY_LABEL:
+      case common::ErrorCode::SCHEMA_VERTEX_SECONDARY_LABEL_IS_PRIMARY:
+      case common::ErrorCode::SCHEMA_VERTEX_PRIMARY_PROPERTIES_UNDEFINED:
+        throw QueryRuntimeException("SchemaViolation occurred when accessing relationships.");
```
antaljanosbenjamin, line 587: I think this function can be completely removed.

src/query/v2/plan/operator.cpp
```diff
@@ -843,7 +851,7 @@ concept AccessorWithProperties = requires(T value, storage::v3::PropertyId prope
                                           storage::v3::PropertyValue property_value) {
   {
     value.ClearProperties()
-    } -> std::same_as<storage::v3::Result<std::map<storage::v3::PropertyId, storage::v3::PropertyValue>>>;
+    } -> std::same_as<storage::v3::ShardResult<std::map<storage::v3::PropertyId, storage::v3::PropertyValue>>>;
```
antaljanosbenjamin, line 854: This concept is also not used, please remove them.

src/query/v2/common.hpp
```diff
@@ -85,36 +84,16 @@ inline void ExpectType(const Symbol &symbol, const TypedValue &value, TypedValue
 template <typename T>
 concept AccessorWithSetProperty = requires(T accessor, const storage::v3::PropertyId key,
                                            const storage::v3::PropertyValue new_value) {
-  { accessor.SetProperty(key, new_value) } -> std::same_as<storage::v3::Result<storage::v3::PropertyValue>>;
+  { accessor.SetProperty(key, new_value) } -> std::same_as<storage::v3::ShardResult<storage::v3::PropertyValue>>;
```
antaljanosbenjamin, line 87: Are these necessary?

src/query/v2/db_accessor.hpp
```diff
@@ -114,53 +113,49 @@ class VertexAccessor final {
 
   auto PrimaryKey(storage::v3::View view) const { return impl_.PrimaryKey(view); }
 
-  storage::v3::ShardOperationResult<bool> AddLabel(storage::v3::LabelId label) {
-    return impl_.AddLabelAndValidate(label);
-  }
+  storage::v3::ShardResult<bool> AddLabel(storage::v3::LabelId label) { return impl_.AddLabelAndValidate(label); }
```
antaljanosbenjamin, line 116: Maybe also remove these functions if it is easy. If it is too much hassle, please let's discuss what is the issue.

src/query/v2/requests.hpp
```diff
@@ -382,6 +389,7 @@ struct GetPropertiesRequest {
 
 struct GetPropertiesResponse {
   bool success;
+  std::optional<ShardError> error;
```
antaljanosbenjamin, line 392: Isn't the empty `error` mean the request was successful? If so, could you please remove the bool flags?

src/query/v2/bindings/eval.hpp
```diff
@@ -44,6 +45,6 @@ class Callable {
 }  // namespace detail
 using ExpressionEvaluator = memgraph::expr::ExpressionEvaluator<
     TypedValue, memgraph::query::v2::EvaluationContext, memgraph::msgs::ShardRequestManagerInterface, storage::v3::View,
-    storage::v3::LabelId, msgs::Value, detail::Callable, memgraph::storage::v3::Error, memgraph::expr::QueryEngineTag>;
+    storage::v3::LabelId, msgs::Value, detail::Callable, common::ErrorCode, memgraph::expr::QueryEngineTag>;
```
antaljanosbenjamin, line 48: Comment to me: check out where the error code is used in `mg-expr`.

src/storage/v3/edge_accessor.cpp
```diff
@@ -128,12 +129,12 @@ Result<PropertyValue> EdgeAccessor::GetProperty(PropertyId property, View view)
         break;
     }
   });
-  if (!exists) return Error::NONEXISTENT_OBJECT;
-  if (!for_deleted_ && deleted) return Error::DELETED_OBJECT;
+  if (!exists) return SHARD_ERROR(common::ErrorCode::NONEXISTENT_OBJECT);
+  if (!for_deleted_ && deleted) return SHARD_ERROR(common::ErrorCode::DELETED_OBJECT);
```
antaljanosbenjamin, line 133: What about including some of the prefix (`common::` or `ErrorCode::`) into the `SHARD_ERROR` macro or declaring it with `using ErrorCode = common::ErrorCode;` just to keep it short and thus easier to read?

src/storage/v3/name_id_mapper.hpp
```diff
@@ -47,6 +48,10 @@ class NameIdMapper final {
     return kUnmappedId;
   }
 
+  const std::string &IdToName(const LabelId label_id) const { return IdToName(label_id.AsInt()); }
+
+  const std::string &IdToName(const PropertyId property_id) const { return IdToName(property_id.AsInt()); }
+
```
antaljanosbenjamin, line 54: I think the `NameIdMapper` shouldn't now about label/property/type ids. It should remain a single `std::string<->uint64_t` bimap.

antaljanosbenjamin, line 54: I think the `NameIdMapper` shouldn't now about label/property/type ids. It should remain a single `std::string<->uint64_t` bimap. I feel that it is questionable, but I am feeling that keeping it simple with simple responsibilities is better for the long term.

antaljanosbenjamin, line 54: I think the `NameIdMapper` shouldn't now about label/property/type ids. It should remain a single `std::string<->uint64_t` bimap. I feel that it is questionable, but I am feeling that keeping it simple with simple responsibilities is better for the long term. If we need this functionality (I haven't seen it yet where do you use it), then a lightweight wrapper should solve the issue.

tests/unit/storage_v3.cpp
```diff
@@ -711,20 +711,20 @@ TEST_P(StorageV3, VertexDeleteLabel) {
 
     // Check whether label 5 exists
     ASSERT_FALSE(vertex->HasLabel(label5, View::OLD).GetValue());
-    ASSERT_EQ(vertex->HasLabel(label5, View::NEW).GetError(), Error::DELETED_OBJECT);
+    ASSERT_EQ(vertex->HasLabel(label5, View::NEW).GetError(), SHARD_ERROR(common::ErrorCode::DELETED_OBJECT));
```
antaljanosbenjamin, line 714: I think implementing an `operator==` for `ShardError` and `ErrorCode` would also make these tests more readable, because with the shard error we shouldn't depend on the source/message.

antaljanosbenjamin (changes requested):

src/expr/interpret/eval.hpp
jbajic: ah I already did that, just forgot to use it everywhere

spacejam (approved):
LGTM - I think the things Bendzo mentioned are more than what jumped out to me

src/query/v2/common.hpp
jbajic: Nope they are not neccessary

src/storage/v3/shard_rsm.cpp
```diff
@@ -461,18 +469,26 @@ EdgeFiller InitializeEdgeFillerFunction(const msgs::ExpandOneRequest &req) {
...
-      return true;
+      return {};
     };
   }
 
   return edge_filler;
 }
 
+auto CreateErrorResponse(const ShardError &shard_error, const auto transaction_id, const std::string_view action) {
+  msgs::ShardError message_shard_error{shard_error.code, shard_error.message};
```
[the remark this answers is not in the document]
jbajic: The reason for that is to differentiate between the error that we send and the error that we propagate in `Storage`. Does it make sense then to move the ShardError to common then and use it in messages?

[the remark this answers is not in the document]
jbajic: I am not sure if I follow why propagate the source to the caller's side. I thought that was only for us to log.

src/storage/v3/result.hpp
antaljanosbenjamin: Try `<source_location>`, maybe it works because we have stdlibc++ 11 which officially supports it.

src/query/v2/db_accessor.hpp
antaljanosbenjamin: These functions will be removed once we start implementing the query engine with multiframes.

src/storage/v3/result.hpp
jbajic: `https://en.cppreference.com/w/cpp/compiler_support/20` we have a lib support but not a compiler, it is supported by clang 15 and greater

antaljanosbenjamin (approved):

PR #654, opened 2022-11-17
[E202-MG < T1164-MG] Add aggregation distinct

niko4299:


src/query/plan/operator.cpp
```diff
@@ -3333,6 +3339,10 @@ class AggregateCursor : public Cursor {
       // Aggregations skip Null input values.
       if (input_value.IsNull()) continue;
       const auto &agg_op = agg_elem_it->op;
+      if (agg_elem_it->distinct) {
+        if (unique_values_it->contains(input_value)) break;
+        unique_values_it->emplace(input_value);
```
antaljanosbenjamin, line 3344: I think this is a good approach and this should work for all aggregations.

src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -2222,43 +2222,45 @@ antlrcpp::Any CypherMainVisitor::visitNumberLiteral(MemgraphCypher::NumberLitera
...
   auto function_name = std::any_cast<std::string>(ctx->functionName()->accept(this));
   std::vector<Expression *> expressions;
   for (auto *expression : ctx->expression()) {
     expressions.push_back(std::any_cast<Expression *>(expression->accept(this)));
   }
   if (expressions.size() == 1U) {
     if (function_name == Aggregation::kCount) {
-      return static_cast<Expression *>(storage_->Create<Aggregation>(expressions[0], nullptr, Aggregation::Op::COUNT));
+      return static_cast<Expression *>(
+          storage_->Create<Aggregation>(expressions[0], nullptr, Aggregation::Op::COUNT, ctx->DISTINCT()));
```
antaljanosbenjamin, line 2233: It is just a small detail, but maybe try to save this value instead of querying in in every case. `const auto is_distinct = ctx->DISTINCT() != nullptr;` should work in my opinion.

antaljanosbenjamin, line 2233: It is just a small detail, but maybe try to save this value instead of querying in in every case. `const auto is_distinct = ctx->DISTINCT() != nullptr;` should work in my opinion.

src/query/plan/operator.cpp
antaljanosbenjamin:
```suggestion
        auto insert_result = unique_values_it->insert(input_value);
        if (insert_result.second) {
          break;
        }
```
This way we don't do unnecessary lookups on the hot-path.

src/query/typed_value.hpp
```diff
@@ -96,6 +97,7 @@ class TypedValue {
   using TString = utils::pmr::string;
   using TVector = utils::pmr::vector<TypedValue>;
   using TMap = utils::pmr::map<utils::pmr::string, TypedValue>;
+  using TSet = utils::pmr::unordered_set<TypedValue, TypedValue::Hash, TypedValue::BoolEqual>;
```
antaljanosbenjamin, line 100: I think this type alias is not in a good place here. All of the above containers type are defined here because can be contained by a `TypedValue`, but this `TSet` is not, thus it is confusing. I think it would be better to define this alias in `AggregateCursor::AggregationValue`.

tests/benchmark/query/execution.cpp
```diff
@@ -416,6 +416,61 @@ BENCHMARK_TEMPLATE(Aggregate, MonotonicBufferResource)
 
 BENCHMARK_TEMPLATE(Aggregate, PoolResource)->Ranges({{4, 1U << 7U}, {512, 1U << 13U}})->Unit(benchmark::kMicrosecond);
 
+template <class TMemory>
```
antaljanosbenjamin, line 419: Do you know where the results of these benchmarks can be found?

gitbuda:
Seems good overall (haven't checked all the details), one thing on my mind, since `count(DISTINECT)` is a part of the openCypher spec, @niko4299 please check with queries are covered under the [TCK tests](https://github.com/memgraph/memgraph/tree/master/tests/gql_behave):
* How many more queries/features (under TCK there is a feature concept) from the TCK Memgraph now support?
* Do we have to add some Memgraph specific BDD test there?

tests/unit/query_plan.cpp
```diff
@@ -426,7 +426,54 @@ TYPED_TEST(TestPlanner, CreateWithSum) {
   AstStorage storage;
   auto ident_n = IDENT("n");
   auto n_prop = PROPERTY_LOOKUP(ident_n, prop);
-  auto sum = SUM(n_prop);
+  auto sum = SUM(n_prop, false);
+  auto query = QUERY(SINGLE_QUERY(CREATE(PATTERN(NODE("n"))), WITH(sum, AS("sum"))));
+  auto symbol_table = memgraph::query::MakeSymbolTable(query);
+  auto acc = ExpectAccumulate({symbol_table.at(*ident_n)});
+  auto aggr = ExpectAggregate({sum}, {});
```
antaljanosbenjamin, line 433: `ExpectAggregate` doesn't check the value of `distinct`, please update that. You can find it in `memgraph/memgraph/tests/unit/query_plan_checker.hpp`.

tests/unit/query_plan.cpp
```diff
@@ -426,7 +426,54 @@ TYPED_TEST(TestPlanner, CreateWithSum) {
...
+  auto planner = MakePlanner<TypeParam>(&dba, storage, symbol_table, query);
+  // We expect both the accumulation and aggregation because the part before
+  // WITH updates the database.
+  CheckPlan(planner.plan(), symbol_table, ExpectCreateNode(), acc, aggr, ExpectProduce());
+}
+
+TYPED_TEST(TestPlanner, MatchWithSumWithDistinctWhereReturn) {
+  // Test MATCH (n) WITH SUM(n.prop) + 42 AS sum WHERE sum < 42
```
antaljanosbenjamin, line 441: The comment is not in sync with the test in terms of `DISTINCT`: in the test we create a `SUM` operator with `true` as `distinct`, but in the query I don't see `DISTINCT`. Please add the `DISTINCT` keyword to the query in all test cases it is applicable and you added.

tests/unit/query_plan_accumulate_aggregate.cpp
```diff
@@ -609,3 +617,396 @@ TEST(QueryPlan, Unwind) {
...
+  group_by_vals.emplace_back(7.3);
+  group_by_vals.emplace_back(7.2);
+  group_by_vals.emplace_back("Johhny");
+  group_by_vals.emplace_back("Jane");
+  group_by_vals.emplace_back("1");
+  group_by_vals.emplace_back(true);
+  group_by_vals.emplace_back(false);
+  group_by_vals.emplace_back(std::vector<memgraph::storage::PropertyValue>{memgraph::storage::PropertyValue(1)});
+  group_by_vals.emplace_back(std::vector<memgraph::storage::PropertyValue>{memgraph::storage::PropertyValue(1),
+                                                                           memgraph::storage::PropertyValue(2)});
```
antaljanosbenjamin, line 744: Why is this so different compared to the test that is without distinct?

tests/unit/query_plan_accumulate_aggregate.cpp
```diff
@@ -609,3 +617,396 @@ TEST(QueryPlan, Unwind) {
     expected_y_it++;
   }
 }
+
+TEST_F(QueryPlanAggregateOps, WithDataDistinct) {
```
antaljanosbenjamin, line 621: I think it might make sense to create a parametrized tests. A good example can be found in `tests/unit/storage_v2_edge.cpp`, but please let sync about how to do this in order to avoid too much extra work. My idea is similar to what is done in the tests of edges: have two test cases: one that is run with `true`/`false` parameters and another one that is independent from the value of distinct.

antaljanosbenjamin (changes requested):
In general it is a nice work @niko4299, you fixed the following 11 openCyper Technology Compatibility Kit (TCK) scenarios:
```
tests/openCypher_M09/features/AggregationAcceptance.feature:195  Distinct on unbound node
tests/openCypher_M09/features/AggregationAcceptance.feature:207  Distinct on null
tests/openCypher_M09/features/AggregationAcceptance.feature:223  Collect distinct nulls
tests/openCypher_M09/features/AggregationAcceptance.feature:235  Collect distinct values mixed with nulls
tests/openCypher_M09/features/ColumnNameAcceptance.feature:48  Keeping used expression 3
tests/openCypher_M09/features/MatchingSelfRelationships.feature:197  Counting distinct undirected self-relationships in self-relationship graph
tests/openCypher_M09/features/OptionalMatchAcceptance.feature:309  OPTIONAL MATCH and `collect()`
tests/openCypher_M09/features/ReturnAcceptance2.feature:200  Arithmetic expressions inside aggregation
tests/openCypher_M09/features/ReturnAcceptance2.feature:560  DISTINCT inside aggregation should work with lists in maps
tests/openCypher_M09/features/ReturnAcceptance2.feature:593  DISTINCT inside aggregation should work with nested lists in maps
tests/openCypher_M09/features/ReturnAcceptance2.feature:609  DISTINCT inside aggregation should work with nested lists of maps in maps
```
And these scenarios made 3 more features complete:
- `ColumnNameAcceptance`
- `MatchingSelfRelationships`
- `OptionalMatchAcceptance`

I think you will make a lot of our users happy once this gets merged!

Edit:
PS.: only one failing scenario is remaining regarding to `DISTINCT` (`Fail when sorting on variable removed by DISTINCT`), but that is somewhat different and this PR is not supposed to fix that. That is more about how the `RETURN DISTINCT n.age` clause is handled.

tests/benchmark/query/execution.cpp
niko4299: Honestly, I thought they were run with benchmarks but now I can't find them.

src/query/plan/operator.cpp
```diff
@@ -25,6 +25,7 @@
 
 #include <cppitertools/chain.hpp>
 #include <cppitertools/imap.hpp>
+#include "query/typed_value.hpp"
```
antaljanosbenjamin, line 28: This is in the wrong block, please put it in the next include block.

src/query/plan/operator.cpp
```diff
@@ -3299,6 +3305,7 @@ class AggregateCursor : public Cursor {
     for (const auto &agg_elem : self_.aggregations_) {
       auto *mem = agg_value->values_.get_allocator().GetMemoryResource();
       agg_value->values_.emplace_back(DefaultAggregationOpValue(agg_elem, mem));
+      agg_value->unique_values_.emplace_back(AggregateCursor::AggregationValue::TSet(mem));
```
antaljanosbenjamin, line 3308:
```suggestion
      agg_value->unique_values_.emplace_back(AggregationValue::TSet(mem));
```

tests/unit/query_plan_accumulate_aggregate.cpp
antaljanosbenjamin: I don't know what did I see here. Checking it now every change makes sense.

src/query/plan/operator.cpp
kostasrim: The container is an `unordered_set` and the lookup is done in O(1). I don't think it would be much difference in this particular case

antaljanosbenjamin (approved):

PR #657, opened 2022-11-17
[🍍  <-] Improve simulator determinism

spacejam:
This makes the simulated tests deterministic for basic_request, multithreaded raft tests, and the cluster property tests. 

Determinism is tested explicitly by running tests twice and asserting that the simulator stats and latency histograms are identical, so if we accidentally add non-deterministic behavior, the tests will hopefully flag it quickly. There are cases where threads will behave slightly differently in-between simulator ticks, but this is fine, because the actual cross-thread communication points are fully controlled now, which is the extent of the partial order that we must impose to achieve the goal of replayable tests.

### DIFF
--- a/src/coordinator/shard_map.cpp
+++ b/src/coordinator/shard_map.cpp
@@ -9,6 +9,7 @@
 // by the Apache License, Version 2.0, included in the file
 // licenses/APL.txt.
 
+#include <cstdint>
 #include <optional>
 #include <unordered_map>
 #include <vector>
@@ -257,13 +258,13 @@ boost::uuids::uuid NewShardUuid(uint64_t shard_id) {
                             0,
                             0,
                             0,
-                            static_cast<unsigned char>(shard_id >> 56),
-                            static_cast<unsigned char>(shard_id >> 48),
-                            static_cast<unsigned char>(shard_id >> 40),
-                            static_cast<unsigned char>(shard_id >> 32),
-                            static_cast<unsigned char>(shard_id >> 24),
-                            static_cast<unsigned char>(shard_id >> 16),
-                            static_cast<unsigned char>(shard_id >> 8),
+                            static_cast<unsigned char>(shard_id >> UINT8_C(56)),
+                            static_cast<unsigned char>(shard_id >> UINT8_C(48)),
+                            static_cast<unsigned char>(shard_id >> UINT8_C(40)),
+                            static_cast<unsigned char>(shard_id >> UINT8_C(32)),
+                            static_cast<unsigned char>(shard_id >> UINT8_C(24)),
+                            static_cast<unsigned char>(shard_id >> UINT8_C(16)),
+                            static_cast<unsigned char>(shard_id >> UINT8_C(8)),
                             static_cast<unsigned char>(shard_id)};
 }
 
--- a/src/coordinator/shard_map.cpp
+++ b/src/coordinator/shard_map.cpp
@@ -9,7 +9,6 @@
 // by the Apache License, Version 2.0, included in the file
 // licenses/APL.txt.
 
-#include <cstdint>
 #include <optional>
 #include <unordered_map>
 #include <vector>
@@ -258,13 +257,13 @@ boost::uuids::uuid NewShardUuid(uint64_t shard_id) {
                             0,
                             0,
                             0,
-                            static_cast<unsigned char>(shard_id >> UINT8_C(56)),
-                            static_cast<unsigned char>(shard_id >> UINT8_C(48)),
-                            static_cast<unsigned char>(shard_id >> UINT8_C(40)),
-                            static_cast<unsigned char>(shard_id >> UINT8_C(32)),
-                            static_cast<unsigned char>(shard_id >> UINT8_C(24)),
-                            static_cast<unsigned char>(shard_id >> UINT8_C(16)),
-                            static_cast<unsigned char>(shard_id >> UINT8_C(8)),
+                            static_cast<unsigned char>(shard_id >> 56u),
+                            static_cast<unsigned char>(shard_id >> 48u),
+                            static_cast<unsigned char>(shard_id >> 40u),
+                            static_cast<unsigned char>(shard_id >> 32u),
+                            static_cast<unsigned char>(shard_id >> 24u),
+                            static_cast<unsigned char>(shard_id >> 16u),
+                            static_cast<unsigned char>(shard_id >> 8u),
                             static_cast<unsigned char>(shard_id)};
 }
 
--- a/src/coordinator/shard_map.cpp
+++ b/src/coordinator/shard_map.cpp
@@ -228,7 +228,7 @@ Hlc ShardMap::IncrementShardMapVersion() noexcept {
   return shard_map_version;
 }
 
-// TODO(antaljanosbenjamin) use a single map for all name id 
+// TODO(antaljanosbenjamin) use a single map for all name id
 // mapping and a single counter to maintain the next id
 std::unordered_map<uint64_t, std::string> ShardMap::IdToNames() {
   std::unordered_map<uint64_t, std::string> id_to_names;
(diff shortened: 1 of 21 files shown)


src/io/address.hpp
```diff
@@ -59,7 +59,7 @@ struct Address {
 
   static Address TestAddress(uint16_t port) {
     return Address{
-        .unique_id = boost::uuids::uuid{boost::uuids::random_generator()()},
+        .unique_id = boost::uuids::uuid{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, static_cast<unsigned char>(port)},
```
antaljanosbenjamin, line 62: Please put an MG_ASSERT to make sure `port` is never bigger than 255, because that would clash with lower port numbers. Or maybe populate the last to bytes with port.

src/io/message_histogram_collector.hpp
```diff
@@ -39,6 +39,8 @@ struct LatencyHistogramSummary {
   Duration p100;
   Duration sum;
 
+  friend bool operator==(const LatencyHistogramSummary &lhs, const LatencyHistogramSummary &rhs) = default;
```
antaljanosbenjamin, line 42: You need to include the `<compare>` header to reliably make this work.

antaljanosbenjamin (changes requested):

gvolfing (approved):

src/io/address.hpp
spacejam: applied

src/io/message_histogram_collector.hpp
spacejam: applied

tests/simulation/raft.cpp
```diff
@@ -229,6 +225,10 @@ void RunSimulation() {
 
   simulator.ShutDown();
 
+  srv_thread_1.join();
+  srv_thread_2.join();
+  srv_thread_3.join();
```
gvolfing, line 230: `std::jthread`s are RAII objects, they join when their destructor is called.

spacejam: we want to join them before retrieving the simulator stats because we want to make sure that they have finished interacting with the simulator before getting the final message counts etc...

antaljanosbenjamin (changes requested):

src/io/simulator/simulator_stats.hpp
```diff
@@ -21,5 +23,19 @@ struct SimulatorStats {
   uint64_t total_requests = 0;
   uint64_t total_responses = 0;
   uint64_t simulator_ticks = 0;
+
```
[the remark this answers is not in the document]
spacejam: yeah, that will be useful to get a sense of time calibration. I've added `elapsed_time` to the SimulatorStats

src/io/rsm/raft.hpp
```diff
@@ -182,7 +181,7 @@ struct PendingClientRequest {
 
 struct Leader {
   std::map<Address, FollowerTracker> followers;
-  std::unordered_map<LogIndex, PendingClientRequest> pending_client_requests;
+  std::map<LogIndex, PendingClientRequest> pending_client_requests;
```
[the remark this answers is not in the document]
spacejam: with the current sketch code it is not necessary, but soon we will iterate over the `pending_client_requests` to handle timeouts, and at that point, it will need to be a `std::map` for determinism. 

The benefits of a hashmap over a tree mostly start becoming significant when the tree structure involves more random memory for nodes (compared to the single allocation of the hashmap) than will fit in L1 cache, and then even more different at the L2 cache boundary. It's possible that we will eventually want to conditionally compile this to use a hashmap for production and a tree for simulation, but in general I estimate that the measurable difference is not going to impact us much in the near term.

src/coordinator/shard_map.cpp
```diff
@@ -268,6 +287,7 @@ std::vector<ShardToInitialize> ShardMap::AssignShards(Address storage_manager,
         if (initialized.contains(aas.address.unique_id)) {
           machine_contains_shard = true;
           if (aas.status != Status::CONSENSUS_PARTICIPANT) {
+            mutated = true;
```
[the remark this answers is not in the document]
spacejam: I noticed it while reading the code, but this did not express itself as a bug with the current tests as far as I saw.

src/io/simulator/simulator_stats.hpp
```diff
@@ -21,5 +25,22 @@ struct SimulatorStats {
   uint64_t total_requests = 0;
   uint64_t total_responses = 0;
   uint64_t simulator_ticks = 0;
+  Duration elapsed_time;
+
+  friend bool operator==(const SimulatorStats & /* lhs */, const SimulatorStats & /* rhs */) = default;
+
+  friend std::ostream &operator<<(std::ostream &in, const SimulatorStats &stats) {
+    auto elapsed_ms = stats.elapsed_time.count() / 1000;
```
antaljanosbenjamin, line 33: As the type of `Duration` can be easily changed from microseconds, please use `std::chrono::duration_cast<std::chrono::milliseconds>` to make sure we don't break this.

src/io/rsm/raft.hpp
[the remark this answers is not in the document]
antaljanosbenjamin: In the C++ community (maybe because of implementation details, I am not sure) a very well known rule of thumb is prefer unordered containers (sets and maps) over ordered ones unless you want to capitalize on the ordered or "node-based" behavior of the container. Therefore on the hot-path (this is probably on the hot-path), I would always prefer the unordered containers if possible.

I think the determinism doesn't depend on the ordered behavior of the map, only on the deterministic behavior, which is true for both `unordered_map` and `map`, or at least I don't see how the `unordered_map` here would affect the deterministic behavior of the simulation. As we totally control the communication between threads (by wait for all threads to block) and delivering the messages one-by-one, I think the order of how the messages are handled on a single thread shouldn't affect the deterministic behavior of the simulation.

Could you please explain why an `unordered_map` would hurt the simulation once we start to iterate over the pending requests?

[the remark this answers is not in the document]
spacejam: yeah, [the performance advantages of hashmaps are clear](https://twitter.com/sadisticsystems/status/1594299358303977472), but many industrial hashmap implementations incorporate some amount of runtime entropy to prevent collision attacks. I don't know about the specifics of our specific version of libstd++, but it's standard practice in the systems engineering community to avoid hashmaps in deterministic contexts.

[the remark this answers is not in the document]
spacejam: it should also be noted that when iterating in lexicographic order over tree, the point operation performance is often better than a hashmap, because you're likely to already have most of the nodes in the access path in L1 cache, so for this use case maps may be better anyway

src/io/simulator/simulator_stats.hpp
spacejam: thanks, applied

tests/simulation/raft.cpp
```diff
@@ -240,21 +240,51 @@ void RunSimulation() {
...
+  std::mt19937 generator(random_device());
+  std::uniform_int_distribution<> distribution;
+
   int n_tests = 50;
 
   for (int i = 0; i < n_tests; i++) {
-    spdlog::info("========================== NEW SIMULATION {} ==========================", i);
-    spdlog::info("\tTime\t\tTerm\tPort\tRole\t\tMessage\n");
-    RunSimulation();
+    uint64_t seed = distribution(generator);
```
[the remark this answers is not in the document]
spacejam: I believe I've fixed this by only advancing the clock about half of the time (not actually half, more based on prime number modulus to avoid some amount of harmonic distortion). The problem was that for certain requests, if there needed to be a certain number of network messages triggered in order to fulfill that request, if the simulator clock moved forward by an average of 15ms with every tick, it basically guaranteed that any request that happened to require more requests than this in a single simulation would time out repeatedly.

antaljanosbenjamin (approved):
Really nice! :rocket:

PR #659, opened 2022-11-20
[master <] Fix mgp::Value copy constructor

gitbuda:
`mgp_value_make_type` is taking ownership of the received pointer, if `mgp::Value(const Type&)` does not make a new copy, double free occurs.

### DIFF
--- a/include/_mgp.hpp
+++ b/include/_mgp.hpp
@@ -290,6 +290,8 @@ mgp_list *list_make_empty(size_t capacity, mgp_memory *memory) {
 
 mgp_list *list_copy(mgp_list *list, mgp_memory *memory) { return MgInvoke<mgp_list *>(mgp_list_copy, list, memory); }
 
+mgp_value *list_value_copy(mgp_list *val, mgp_memory *memory) { return mgp::value_make_list(list_copy(val, memory)); }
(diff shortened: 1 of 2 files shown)


include/_mgp.hpp
```diff
@@ -290,6 +290,8 @@ mgp_list *list_make_empty(size_t capacity, mgp_memory *memory) {
 
 mgp_list *list_copy(mgp_list *list, mgp_memory *memory) { return MgInvoke<mgp_list *>(mgp_list_copy, list, memory); }
 
+mgp_value *list_value_copy(mgp_list *val, mgp_memory *memory) { return mgp::value_make_list(list_copy(val, memory)); }
```
gitbuda, line 293: The `type_value_copy` functions are actually not required... 🤔

antepusic: Do you mean that the `type_value_copy` funcions should be removed in the next iteration of the PR?

antepusic: Do you mean that the `type_value_copy` funcions should be removed in the next iteration of the PR?

gitbuda:
This should be closed because @antepusic did the improvement in the #688

PR #663, opened 2022-11-23
[:pineapple:  < T1167-MG] Create ScanAllByPrimaryKey operator

gvolfing:
[master < Task] PR


src/query/v2/plan/operator.cpp
```diff
@@ -462,6 +463,202 @@ class DistributedScanAllAndFilterCursor : public Cursor {
...
+  }
+
+  enum class State : int8_t { INITIALIZING, COMPLETED };
+
+  using VertexAccessor = accessors::VertexAccessor;
+
+  bool MakeRequest(RequestRouterInterface &request_router, ExecutionContext &context) {
+    {
+      SCOPED_REQUEST_WAIT_PROFILE;
+      std::optional<std::string> request_label = std::nullopt;
```
jbajic, line 490:
```suggestion
      std::optional<std::string> request_label;
```

src/query/v2/plan/rewrite/index_lookup.hpp
```diff
@@ -559,31 +568,43 @@ class IndexLookupRewriter final : public HierarchicalLogicalOperatorVisitor {
...
-        if (filter.id_filter->is_symbol_in_value_ || !are_bound(filter.used_symbols)) continue;
-        auto *value = filter.id_filter->value_;
-        filter_exprs_for_removal_.insert(filter.expression);
-        filters_.EraseFilter(filter);
-        return std::make_unique<ScanAllById>(input, node_symbol, value, view);
-      }
-    }
-    // Now try to see if we can use label+property index. If not, try to use
+
+    // Try to see if we can use label+property index. If not, try to use
```
antaljanosbenjamin, line 572: Maybe mention primary keys explicitly? Even though they are some kind of implicit label+property indices, I think they are significant enough to be mentioned here.

src/query/v2/plan/operator.cpp
```diff
@@ -462,6 +463,202 @@ class DistributedScanAllAndFilterCursor : public Cursor {
...
+        filter_expressions_(filter_expressions),
+        primary_key_(primary_key) {
+    ResetExecutionState();
+  }
+
+  enum class State : int8_t { INITIALIZING, COMPLETED };
+
+  using VertexAccessor = accessors::VertexAccessor;
+
+  bool MakeRequest(RequestRouterInterface &request_router, ExecutionContext &context) {
```
jbajic, line 487: Why do we have `MakeRequest` and `MakeRequestSingleFrame`, when is one called and when the other?

src/query/v2/plan/rewrite/index_lookup.hpp
antaljanosbenjamin, line 572: Maybe mention primary keys explicitly? Even though they are some kind of implicit label+property indices, I think they are significant enough to be mentioned here.

src/query/v2/plan/rewrite/index_lookup.hpp
```diff
@@ -559,31 +568,43 @@ class IndexLookupRewriter final : public HierarchicalLogicalOperatorVisitor {
...
+
+    // First, try to see if we can find a vertex based on the possibly
+    // supplied primary key.
+    auto property_filters = filters_.PropertyFilters(node_symbol);
+    query::v2::LabelIx prim_label;
+    std::vector<std::pair<query::v2::Expression *, query::v2::plan::FilterInfo>> primary_key;
+
+    if (!property_filters.empty()) {
+      for (const auto &label : labels) {
+        if (db_->LabelIndexExists(GetLabel(label))) {
```
antaljanosbenjamin, line 588: Maybe add a new function to check if a label is a primary label. Right now it will be the same as `LabelIndexExists`, but as soon as we introduce label indices on secondary labels, they will diverge.

src/query/v2/plan/rewrite/index_lookup.hpp
```diff
@@ -559,31 +568,43 @@ class IndexLookupRewriter final : public HierarchicalLogicalOperatorVisitor {
...
+          prim_label = label;
+          primary_key = db_->ExtractPrimaryKey(GetLabel(prim_label), property_filters);
```
antaljanosbenjamin, line 590: tMaybe the `VertexCountCache` could return the schema. I am almost sure that it will come handy in the future. Furthermore this logic (extraxting the primary properties) doesn't fit the `VertexCountCache` in my opinion, would be better in here.

src/query/v2/plan/operator.cpp
```diff
@@ -462,6 +463,202 @@ class DistributedScanAllAndFilterCursor : public Cursor {
...
+
+ private:
+  const Symbol output_symbol_;
+  const UniqueCursorPtr input_cursor_;
+  const char *op_name_;
+  std::vector<VertexAccessor> current_batch_;
+  std::vector<VertexAccessor>::iterator current_vertex_it_;
+  State request_state_ = State::INITIALIZING;
+  std::optional<storage::v3::LabelId> label_;
+  std::optional<std::pair<storage::v3::PropertyId, Expression *>> property_expression_pair_;
```
antaljanosbenjamin, line 653: Is this used?

src/query/v2/plan/operator.cpp
```diff
@@ -462,6 +463,202 @@ class DistributedScanAllAndFilterCursor : public Cursor {
...
+  using VertexAccessor = accessors::VertexAccessor;
+
+  bool MakeRequest(RequestRouterInterface &request_router, ExecutionContext &context) {
+    {
+      SCOPED_REQUEST_WAIT_PROFILE;
+      std::optional<std::string> request_label = std::nullopt;
+      if (label_.has_value()) {
+        request_label = request_router.LabelToName(*label_);
+      }
+      current_batch_ = request_router.ScanVertices(request_label);
```
antaljanosbenjamin, line 494: Why do we do a `ScanVertices`? Is this present just because of `PullMultiple`? If so, then please remove `PullMultiple` and all the other functions that are used only from `PullMultiple`. As this PR is not intended to implement the `MultiFrame` approach, it is better not to have any implementation. In that case the execution of `PullMultiple` will fail with an assert, which is fine.

src/query/v2/plan/operator.cpp
```diff
@@ -462,6 +463,202 @@ class DistributedScanAllAndFilterCursor : public Cursor {
...
+        request_label = request_router.LabelToName(*label_);
+      }
+
+      // Evaluate the expressions that hold the PrimaryKey.
+      ExpressionEvaluator evaluator(&frame, context.symbol_table, context.evaluation_context, context.request_router,
+                                    storage::v3::View::NEW);
+
+      std::vector<msgs::Value> pk;
+      MG_ASSERT(primary_key_);
+      for (auto *primary_key : *primary_key_) {
```
antaljanosbenjamin, line 515:
```suggestion
      for (auto *primary_property : *primary_key_) {
```

src/query/v2/plan/operator.cpp
```diff
@@ -462,6 +463,202 @@ class DistributedScanAllAndFilterCursor : public Cursor {
...
+    return !current_batch_.empty();
+  }
+
+  bool MakeRequestSingleFrame(Frame &frame, RequestRouterInterface &request_router, ExecutionContext &context) {
+    {
+      SCOPED_REQUEST_WAIT_PROFILE;
+      std::optional<std::string> request_label = std::nullopt;
+      if (label_.has_value()) {
+        request_label = request_router.LabelToName(*label_);
+      }
```
antaljanosbenjamin, line 507:
```suggestion
```

I think `request_label` is not used anywhere.

src/query/v2/plan/operator.cpp
```diff
@@ -462,6 +463,202 @@ class DistributedScanAllAndFilterCursor : public Cursor {
...
+      current_batch_ = request_router.ScanVertices(request_label);
+    }
+    current_vertex_it_ = current_batch_.begin();
+    request_state_ = State::COMPLETED;
+    return !current_batch_.empty();
+  }
+
+  bool MakeRequestSingleFrame(Frame &frame, RequestRouterInterface &request_router, ExecutionContext &context) {
+    {
+      SCOPED_REQUEST_WAIT_PROFILE;
```
antaljanosbenjamin, line 503: First of all, I think this `SCOPED_REQUEST_WAIT_PROFILE` is not the best solution, because it is hard to put it on a separate scope and return something from the request. However for now we can use something like this:

```
      auto get_prop_result = std::invoke([&context, &req]() mutable {
        SCOPED_REQUEST_WAIT_PROFILE;
        return request_router.GetProperties(req);
      });
```

src/query/v2/plan/operator.cpp
```diff
@@ -462,6 +463,202 @@ class DistributedScanAllAndFilterCursor : public Cursor {
...
+      MG_ASSERT(get_prop_result.size() <= 1);
+
+      if (get_prop_result.empty()) {
+        current_batch_ = std::vector<VertexAccessor>{};
+      } else {
+        auto properties = get_prop_result[0].props;
+        // TODO (gvolfing) figure out labels when relevant.
+        msgs::Vertex vertex = {.id = get_prop_result[0].vertex, .labels = {}};
+
+        current_batch_ = {VertexAccessor(vertex, properties, &request_router)};
```
antaljanosbenjamin, line 532: I think this is more complex then it should be. Here we don't want to deal with a batch, only a single vertex. So basically we either found the vertex or not, but we will never find more than 1 vertex. Therefore using a vector to store the result is kind of confusing.

src/query/v2/plan/operator.cpp
```diff
@@ -462,6 +463,202 @@ class DistributedScanAllAndFilterCursor : public Cursor {
...
+    ResetExecutionState();
+  }
+
+ private:
+  const Symbol output_symbol_;
+  const UniqueCursorPtr input_cursor_;
+  const char *op_name_;
+  std::vector<VertexAccessor> current_batch_;
+  std::vector<VertexAccessor>::iterator current_vertex_it_;
+  State request_state_ = State::INITIALIZING;
```
antaljanosbenjamin, line 651: The request state can be removed, it was only used when we had pagination. Since then I think it doesn't make sense to use it, therefore I removed it from the other distributed scanall cursor on the other pr.

src/query/v2/plan/operator.cpp
```diff
@@ -462,6 +463,202 @@ class DistributedScanAllAndFilterCursor : public Cursor {
...
+    while (true) {
+      if (MustAbort(context)) {
+        throw HintedAbortError();
+      }
+
+      if (request_state_ == State::INITIALIZING) {
+        if (!input_cursor_->Pull(frame, context)) {
+          return false;
+        }
+      }
```
antaljanosbenjamin, line 553: As we produce at most 1 row for each row from our input, I would say we always have to pull our input. That means this logic seems unnecessary. See my other comments about states and batching.

src/query/v2/plan/pretty_print.cpp
```diff
@@ -86,10 +86,10 @@ bool PlanPrinter::PreVisit(query::v2::plan::ScanAllByLabelProperty &op) {
   return true;
 }
 
-bool PlanPrinter::PreVisit(ScanAllById &op) {
+bool PlanPrinter::PreVisit(query::v2::plan::ScanAllByPrimaryKey &op) {
   WithPrintLn([&](auto &out) {
-    out << "* ScanAllById"
-        << " (" << op.output_symbol_.name() << ")";
+    out << "* ScanAllByPrimaryKey"
+        << " (" << op.output_symbol_.name() << " :" << request_router_->LabelToName(op.label_) << ")";
```
antaljanosbenjamin, line 92: If it is easy to print the property name also, please consider printing it.

src/query/v2/plan/read_write_type_checker.hpp
```diff
@@ -59,7 +59,6 @@ class ReadWriteTypeChecker : public virtual HierarchicalLogicalOperatorVisitor {
   bool PreVisit(ScanAllByLabelPropertyValue &) override;
   bool PreVisit(ScanAllByLabelPropertyRange &) override;
   bool PreVisit(ScanAllByLabelProperty &) override;
```
antaljanosbenjamin, line 61: You haven't added `ScanAllByPrimaryKey` here.

src/query/v2/interpreter.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright 2022 Memgraph Ltd.
+// Copyright 2023 Memgraph Ltd.
```
antaljanosbenjamin, line 1: Please revert this if you haven't done any other changes.

src/query/v2/context.hpp
```diff
@@ -1,4 +1,4 @@
-// Copyright 2022 Memgraph Ltd.
+// Copyright 2023 Memgraph Ltd.
```
antaljanosbenjamin, line 1: Please revert this if you haven't done any other changes.

src/query/v2/multiframe.hpp
```diff
@@ -200,10 +200,10 @@ class ValidFramesConsumer {
...
 
   ~ValidFramesConsumer() noexcept;
-  ValidFramesConsumer(const ValidFramesConsumer &other) = delete;
-  ValidFramesConsumer(ValidFramesConsumer &&other) noexcept = delete;
-  ValidFramesConsumer &operator=(const ValidFramesConsumer &other) = delete;
-  ValidFramesConsumer &operator=(ValidFramesConsumer &&other) noexcept = delete;
+  ValidFramesConsumer(const ValidFramesConsumer &other) = default;
+  ValidFramesConsumer(ValidFramesConsumer &&other) noexcept = default;
+  ValidFramesConsumer &operator=(const ValidFramesConsumer &other) = default;
+  ValidFramesConsumer &operator=(ValidFramesConsumer &&other) noexcept = default;
```
antaljanosbenjamin, line 206: Are you sure we want to copy them? That could lead into issues because of the simultaneous access.

src/storage/v3/shard_rsm.cpp
```diff
@@ -535,13 +535,17 @@ msgs::ReadResponses ShardRsm::HandleRead(msgs::GetPropertiesRequest &&req) {
...
 
-  auto collect_props = [&req](const VertexAccessor &v_acc,
-                              const std::optional<EdgeAccessor> &e_acc) -> ShardResult<std::map<PropertyId, Value>> {
+  auto collect_props = [this, &req](
+                           const VertexAccessor &v_acc,
+                           const std::optional<EdgeAccessor> &e_acc) -> ShardResult<std::map<PropertyId, Value>> {
     if (!req.property_ids) {
+      const auto *schema = shard_->GetSchema(shard_->PrimaryLabel());
+      MG_ASSERT(schema);
+
```
antaljanosbenjamin, line 544: First of all, I had to do the similar changes on the `ExpandOne` PR. However I would put this after the condition as for edges we don't need the schema. Pushing the instantiation of a local variable as late as ~possible~ reasonable is always a good thing.

tests/mgbench/client.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright 2022 Memgraph Ltd.
+// Copyright 2023 Memgraph Ltd.
```
antaljanosbenjamin, line 1: Please revert this if you haven't done any other changes. If you cannot because of the pre-commit hook, then it is fine as is.

The same applies to other files also.

tests/unit/query_common.hpp
```diff
@@ -486,6 +486,7 @@ auto GetForeach(AstStorage &storage, NamedExpression *named_expr, const std::vec
 #define EDGE(...) memgraph::query::test_common::GetEdge(storage, __VA_ARGS__)
 #define EDGE_VARIABLE(...) memgraph::query::test_common::GetEdgeVariable(storage, __VA_ARGS__)
 #define PATTERN(...) memgraph::query::test_common::GetPattern(storage, {__VA_ARGS__})
+#define PATTERN(...) memgraph::query::test_common::GetPattern(storage, {__VA_ARGS__})
```
antaljanosbenjamin, line 489: IS this a duplication?

tests/unit/query_plan_checker_v2.hpp
```diff
@@ -0,0 +1,435 @@
...
+    MG_ASSERT(!schemas_.contains(primary_label), "You already created the schema for this label!");
+    schemas_.emplace(primary_label, schemas_types);
+  }
+
+ private:
+  std::unordered_map<std::string, memgraph::storage::v3::LabelId> primary_labels_;
+  std::unordered_map<std::string, memgraph::storage::v3::LabelId> secondary_labels_;
+  std::unordered_map<std::string, memgraph::storage::v3::EdgeTypeId> edge_types_;
+  std::unordered_map<std::string, memgraph::storage::v3::PropertyId> primary_properties_;
+  std::unordered_map<std::string, memgraph::storage::v3::PropertyId> secondary_properties_;
```
antaljanosbenjamin, line 426: Is there a reason to separate primary and secondary properties? I think they can go into the same container, as their mapping should be the same. If I am missing something, please let me know!

antaljanosbenjamin (changes requested):
I think you successfully tackled the hardest part. Most of my comments I think are small, easy to solve issues.

src/query/v2/plan/pretty_print.hpp
```diff
@@ -194,7 +194,7 @@ class PlanToJsonVisitor : public virtual HierarchicalLogicalOperatorVisitor {
   bool PreVisit(ScanAllByLabelPropertyRange &) override;
   bool PreVisit(ScanAllByLabelPropertyValue &) override;
   bool PreVisit(ScanAllByLabelProperty &) override;
-  bool PreVisit(ScanAllById &) override;
+  bool PreVisit(ScanAllByPrimaryKey & /*unused*/) override;
```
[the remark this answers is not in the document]
gvolfing: Yeah, I think clang-tidy had some issues with not marking any newly declaration with /*unused*/ if the identifier is not specified, I will remove and check if that's still a problem.

src/query/v2/plan/operator.cpp
gvolfing: Atm, we have to set a flag depending on what frame-type we want to use in the pull-chain, however the multi-frame version is not implemented in this PR, only the underlying logic that makes it possible to use and pick the new operator when appropriate and the single-frame version of its execution.
In order to get the `PrimaryKey` we have to evaluate the `Expression*`s that are members of the cursor object. For that we need to pass the `Frame&` to the `MakeRequestSingleFrame` function. Once the multi-frame version is implemented, it will probably take `MultiFrame&` to evaluate the same stuff and once we settle the `MultiFrame` vs `Frame` discussion one of these functions will disappear.
Since the currently implemented `MultiFrame` operators use the `MakeRequest` name, I did not wanted to use that in the case of the `SingleFrame` solution.
So,
single-frame solution -> `MakeRequestSingleFrame`
multi-frame solution -> `MakeRequest`

As for the multi-frame solution I will mark it not implemented appropriately, I forgot to do that.

src/query/v2/plan/operator.cpp
gvolfing: I will remove everything that is `Multiframe` related.

src/query/v2/plan/operator.cpp
gvolfing: Removed.

src/query/v2/multiframe.hpp
gvolfing: I will remove, these changes, since multi-frame is not used in this PR.

tests/unit/query_plan_checker_v2.hpp
gvolfing: This is part of the `FakeDistributedDbAccessor` mock class.
Yes, the mapping is the same and they could easily go into the same container, however, when setting up the tests and the logic of the mocked methods become a lot more simpler, when they are separated into different datastructures.
If you insist I can change that, but I would like the implementation of the mocked methods to be as plain and simple as possible.

src/query/v2/plan/operator.cpp
jbajic: Aha ok thanks for explanation :+1:

src/query/v2/plan/pretty_print.cpp
gvolfing: It seems to be a bit convoluted, if you insist I can spend some more time with it.

src/query/v2/plan/vertex_count_cache.hpp
```diff
@@ -52,11 +54,33 @@ class VertexCountCache {
...
+
+    std::transform(schema.begin(), schema.end(), std::back_inserter(schema_properties),
+                   [](const auto &schema_elem) { return schema_elem.property_id; });
+
+    for (const auto &property_filter : property_filters) {
+      const auto &property_id = NameToProperty(property_filter.property_filter->property_.name);
+      if (std::find(schema_properties.begin(), schema_properties.end(), property_id) != schema_properties.end()) {
+        pk.emplace_back(std::make_pair(property_filter.expression, property_filter));
+      }
+    }
```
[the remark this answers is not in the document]
gvolfing: FYI, during refactoring this part has been moved to `index_lookup.hpp`.

Added an other transformation to make sure they are in the same order.

src/query/v2/plan/vertex_count_cache.hpp
```diff
@@ -52,11 +55,16 @@ class VertexCountCache {
     return 1;
   }
 
-  // For now return true if label is primary label
-  bool LabelIndexExists(storage::v3::LabelId label) { return request_router_->IsPrimaryLabel(label); }
+  bool LabelIndexExists(storage::v3::LabelId label) { return PrimaryLabelExists(label); }
```
gvolfing, line 58: This used to be the original functionality of this function, this should not throw, since every time an scan operator of the `ScanAllByLabel` would be initiated, it would fail, instead of picking another operator this broke some tests previously. Returning false, breaks some of the e2e tests, so for now, until we have proper indexing based on labels, I would opt for this implementation.

src/query/v2/plan/vertex_count_cache.hpp
```diff
@@ -54,31 +55,14 @@ class VertexCountCache {
...
-      if (std::find(schema_properties.begin(), schema_properties.end(), property_id) != schema_properties.end()) {
-        pk.emplace_back(std::make_pair(property_filter.expression, property_filter));
-      }
-    }
+  bool LabelPropertyIndexExists(storage::v3::LabelId /*label*/, storage::v3::PropertyId /*property*/) { return false; }
 
-    return pk.size() == schema_properties.size()
-               ? pk
-               : std::vector<std::pair<query::v2::Expression *, query::v2::plan::FilterInfo>>{};
+  std::vector<memgraph::storage::v3::SchemaProperty> GetSchemaForLabel(storage::v3::LabelId label) {
```
antaljanosbenjamin, line 64: I didn't catch this before, but both this and the function in `RequestRouter` should return `const&` to not copy the schema always.

src/query/v2/plan/operator.cpp
```diff
@@ -518,6 +519,94 @@ class DistributedScanAllAndFilterCursor : public Cursor {
...
+  const Symbol output_symbol_;
+  const UniqueCursorPtr input_cursor_;
+  const char *op_name_;
+  storage::v3::LabelId label_;
+  std::optional<std::vector<Expression *>> filter_expressions_;
+  std::vector<Expression *> primary_key_;
+  std::optional<MultiFrame> own_multi_frames_;
+  std::optional<ValidFramesConsumer> valid_frames_consumer_;
+  ValidFramesConsumer::Iterator valid_frames_it_;
+  std::queue<FrameWithValidity> frames_buffer_;
```
antaljanosbenjamin, line 607: I have no issues of adding the members that will be required to implement the `MultiFrame` based approach, howevevr then please reset them properly in the `Reset` method, otherwise we will forget them. So please either remove the unused member variables or handle them in `Reset`.

src/query/v2/plan/operator.cpp
```diff
@@ -518,6 +519,94 @@ class DistributedScanAllAndFilterCursor : public Cursor {
...
+      return false;
+    }
+
+    auto &request_router = *context.request_router;
+    auto vertex = MakeRequestSingleFrame(frame, request_router, context);
+    if (vertex) {
+      frame[output_symbol_] = TypedValue(std::move(*vertex));
+      return true;
+    }
+    return false;
```
antaljanosbenjamin, line 586: I think there is an issue here. We shouldn't return false just because we couldn't fetch a vertex. We should only return false (which basically signals that the cursor cannot produce more results), when the input cursor is exhausted, thus we don't have more vertex ids to try to fetch. Therefore I think some kind of a loop is necessary here.

An example to imagine:
1. `CREATE (n:label {property:1});` 
2. `CREATE (n:label {property:10});`
3. `UNWIND [1, 42, 10] AS prop MATCH (n:label {property:prop}) RETURN n;` 

The 3rd query should return `(:label {property:1})` and `(:label {property:10})`, but because of line 586 the execution will be stopped when `label(property:42)` doesn't produce any results.

tests/unit/query_plan_checker_v2.hpp
```diff
@@ -0,0 +1,452 @@
...
+    auto find_in_prim_properties = primary_properties_.find(name);
+    if (find_in_prim_properties != primary_properties_.end()) {
+      return find_in_prim_properties->second;
+    }
+    auto find_in_secondary_properties = secondary_properties_.find(name);
+    if (find_in_secondary_properties != secondary_properties_.end()) {
+      return find_in_secondary_properties->second;
+    }
+
+    MG_ASSERT(false, "The property does not exist as a primary or a secondary property.");
```
antaljanosbenjamin, line 384: Please be consistent, either `LOG_FATAL` or `MG_ASSERT(false, "message")`.

antaljanosbenjamin (changes requested):

antaljanosbenjamin (approved):

jbajic (approved):

PR #667, opened 2022-11-25
[master < T507-FL] EmptyResult operator

### LINKED ISSUE #650: [BUG] MemGraph+neo4j-go-driver is stuck if cypher contains any path and does not have RETURN
**Memgraph version**
2.4.1 (mage)

**Environment**
docker on WSL2

**Describe the bug**
memgraph+neo4j-go-driver is stuck on `sess.Close` or `res.Consume` if cypher contains any path and does not have RETURN. neo4j works ok

**To Reproduce**
```go
package main

import (
	"context"
	"fmt"
	"time"

	"github.com/neo4j/neo4j-go-driver/v5/neo4j"
)

const url = "bolt://localhost:7687"

func query(driver neo4j.DriverWithContext, cypher string) error {
	ctx := context.Background()
	ctx, cancel := context.WithTimeout(ctx, 2*time.Second)
	defer cancel()
	sess := driver.NewSession(ctx, neo4j.SessionConfig{})
	defer sess.Close(ctx) // stuck here is no res.Consume

	res, err := sess.Run(ctx, cypher, nil)
	if err != nil {
		return err
	}

	for res.Next(ctx) {
		fmt.Println(res.Record().Values...)
	}

	err = res.Err()
	if err != nil {
		return err
	}

	if res.IsOpen() {
		fmt.Println("trying to consume")
		_, err := res.Consume(ctx)
		if err != nil {
			return err
		}
		fmt.Println("ok")
	}

	return nil
}

func example() error {
	driver, err := neo4j.NewDriverWithContext(url, neo4j.BasicAuth("", "", ""))
	if err != nil {
		return err
	}

	if err := query(driver, "RETURN 1+1"); err != nil {
		return err
	}
	if err := query(driver, "MATCH ()-[r]->() SET r.test = 123 RETURN DISTINCT 1"); err != nil {
		fmt.Println(err)
	}
	if err := query(driver, "MATCH ()-[r]->() SET r.test = 123"); err != nil {
		fmt.Println(err)
	}
	if err := query(driver, "MATCH (r)-[]->() SET r.test = 123"); err != nil {
		fmt.Println(err)
	}

	return nil
}

func main() {
	if err := example(); err != nil {
		panic(err)
	}
}
```
**Expected behavior**
No timeout errors

**Logs**
```
2
1
trying to consume
ConnectivityError: Timeout while reading from connection [server-side timeout hint: -1ns, user-provided context deadline: 2022-11-15 16:09:01.720096798 +0300 MSK m=+2.011375250]: context deadline exceeded
trying to consume
ConnectivityError: Timeout while reading from connection [server-side timeout hint: -1ns, user-provided context deadline: 2022-11-15 16:09:03.722075187 +0300 MSK m=+4.013353634]: context deadline exceeded
```

as51340:
[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR

Closes #650

### DIFF
--- a/src/query/plan/operator.cpp
+++ b/src/query/plan/operator.cpp
@@ -3058,12 +3058,19 @@ void EdgeUniquenessFilter::EdgeUniquenessFilterCursor::Shutdown() { input_cursor
 
 void EdgeUniquenessFilter::EdgeUniquenessFilterCursor::Reset() { input_cursor_->Reset(); }
 
-EmptyResult::EmptyResult(const std::shared_ptr<LogicalOperator> &input, const std::vector<Symbol> &symbols)
-    : input_(input ? input : std::make_shared<Once>()), symbols_(symbols) {}
+EmptyResult::EmptyResult(const std::shared_ptr<LogicalOperator> &input)
+    : input_(input ? input : std::make_shared<Once>()) {}
 
 ACCEPT_WITH_INPUT(EmptyResult)
 
-std::vector<Symbol> EmptyResult::ModifiedSymbols(const SymbolTable &) const { return symbols_; }
+std::vector<Symbol> EmptyResult::OutputSymbols(const SymbolTable &symbol_table) const {
+  // Propagate this to potential Produce.
+  return input_->OutputSymbols(symbol_table);
+}
+
+std::vector<Symbol> EmptyResult::ModifiedSymbols(const SymbolTable &table) const {
+  return input_->ModifiedSymbols(table);
+}
 
 class EmptyResultCursor : public Cursor {
  public:
--- a/src/query/plan/operator.cpp
+++ b/src/query/plan/operator.cpp
@@ -3059,7 +3059,7 @@ void EdgeUniquenessFilter::EdgeUniquenessFilterCursor::Shutdown() { input_cursor
 void EdgeUniquenessFilter::EdgeUniquenessFilterCursor::Reset() { input_cursor_->Reset(); }
 
 EmptyResult::EmptyResult(const std::shared_ptr<LogicalOperator> &input, const std::vector<Symbol> &symbols)
-    : input_(input), symbols_(symbols) {}
+    : input_(input ? input : std::make_shared<Once>()), symbols_(symbols) {}
 
 ACCEPT_WITH_INPUT(EmptyResult)
 
@@ -3073,12 +3073,14 @@ class EmptyResultCursor : public Cursor {
   bool Pull(Frame &frame, ExecutionContext &context) override {
     SCOPED_PROFILE_OP("EmptyResult");
 
-    // cache all the input
+    // spdlog::info("Executing EmptyResult operator");
     if (!pulled_all_input_) {
-      while (input_cursor_->Pull(frame, context)) pulled_all_input_ = true;
+      while (input_cursor_->Pull(frame, context)) {
+      }
+      pulled_all_input_ = true;
     }
     if (MustAbort(context)) throw HintedAbortError();
-    return true;
+    return false;
   }
 
   void Shutdown() override { input_cursor_->Shutdown(); }
--- a/src/query/plan/operator.cpp
+++ b/src/query/plan/operator.cpp
@@ -3064,7 +3064,6 @@ EmptyResult::EmptyResult(const std::shared_ptr<LogicalOperator> &input)
 ACCEPT_WITH_INPUT(EmptyResult)
 
 std::vector<Symbol> EmptyResult::OutputSymbols(const SymbolTable &symbol_table) const {
-  // Propagate this to potential Produce.
   return input_->OutputSymbols(symbol_table);
 }
 
@@ -3080,7 +3079,6 @@ class EmptyResultCursor : public Cursor {
   bool Pull(Frame &frame, ExecutionContext &context) override {
     SCOPED_PROFILE_OP("EmptyResult");
 
-    // spdlog::info("Executing EmptyResult operator");
     if (!pulled_all_input_) {
       while (input_cursor_->Pull(frame, context)) {
       }
--- a/src/query/plan/operator.cpp
+++ b/src/query/plan/operator.cpp
@@ -113,6 +113,7 @@ extern const Event UnionOperator;
 extern const Event CartesianOperator;
 extern const Event CallProcedureOperator;
 extern const Event ForeachOperator;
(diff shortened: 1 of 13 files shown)


src/query/plan/operator.cpp
```diff
@@ -3057,6 +3058,47 @@ void EdgeUniquenessFilter::EdgeUniquenessFilterCursor::Shutdown() { input_cursor
 
 void EdgeUniquenessFilter::EdgeUniquenessFilterCursor::Reset() { input_cursor_->Reset(); }
 
+EmptyResult::EmptyResult(const std::shared_ptr<LogicalOperator> &input, const std::vector<Symbol> &symbols)
```
kostasrim, line 3061:
```suggestion
EmptyResult::EmptyResult(std::shared_ptr<LogicalOperator> input, const std::vector<Symbol> &symbols)
```

src/query/plan/operator.cpp
```diff
@@ -3057,6 +3058,47 @@ void EdgeUniquenessFilter::EdgeUniquenessFilterCursor::Shutdown() { input_cursor
...
+    : input_(input), symbols_(symbols) {}
```
kostasrim, line 3062:
```suggestion
    : input_(std::move(input)), symbols_(symbols) {}
```

src/query/plan/operator.cpp
```diff
@@ -3057,6 +3058,47 @@ void EdgeUniquenessFilter::EdgeUniquenessFilterCursor::Shutdown() { input_cursor
...
+
+class EmptyResultCursor : public Cursor {
+ public:
+  EmptyResultCursor(const EmptyResult &self, utils::MemoryResource *mem)
+      : input_cursor_(self.input_->MakeCursor(mem)) {}
+
+  bool Pull(Frame &frame, ExecutionContext &context) override {
+    SCOPED_PROFILE_OP("EmptyResult");
+
+    // cache all the input
```
kostasrim, line 3076: This is not yet implemented and plz add a unit. Are you also planning to change the planner in this PR?

kostasrim (commented):
Preliminary review as asked

src/query/plan/operator.cpp
as51340: Yes, planner will be changed in this PR.

src/query/plan/operator.cpp
```diff
@@ -3057,6 +3058,56 @@ void EdgeUniquenessFilter::EdgeUniquenessFilterCursor::Shutdown() { input_cursor
...
+
+class EmptyResultCursor : public Cursor {
+ public:
+  EmptyResultCursor(const EmptyResult &self, utils::MemoryResource *mem)
+      : input_cursor_(self.input_->MakeCursor(mem)) {}
+
+  bool Pull(Frame &frame, ExecutionContext &context) override {
+    SCOPED_PROFILE_OP("EmptyResult");
+
+    // spdlog::info("Executing EmptyResult operator");
```
gitbuda, line 3083: commented code, remove it + check for the similar in the rest of the PR

gitbuda (changes requested):
There is some commented code + some form of testing is missing.

I would add a new test here -> https://github.com/memgraph/memgraph/tree/master/tests/gql_behave/tests/memgraph_V1, but probably there are other viable options 😄

as51340:
> There is some commented code + some form of testing is missing.
> 
> I would add a new test here -> https://github.com/memgraph/memgraph/tree/master/tests/gql_behave/tests/memgraph_V1, but probably there are other viable options smile

I changed all unit tests that make use of this new operator, doesn't this suffice?

src/query/plan/rule_based_planner.hpp
```diff
@@ -232,6 +252,10 @@ class RuleBasedPlanner {
           throw utils::NotYetImplemented("clause '{}' conversion to operator(s)", clause->GetTypeInfo().name);
         }
       }
+      // Is this the only situation that should be covered
+      if (!handle_empty_result && !query_part.remaining_clauses.empty()) {
+        input_op = std::make_unique<EmptyResult>(std::move(input_op));
+      }
```
jbajic, line 258: It is a bit strange that if `handle_empty_result` is false that we want to create `EmptyResult`, maybe rename the bool, or something just to make it more clear

src/query/plan/operator.cpp
as51340: I followed what other logical operators' constructors look like, it is with const?

src/query/plan/rule_based_planner.hpp
```diff
@@ -180,11 +180,16 @@ class RuleBasedPlanner {
         }
       }
       uint64_t merge_id = 0;
-      for (auto *clause : query_part.remaining_clauses) {
+      bool handle_empty_result = true;
```
antaljanosbenjamin, line 183: Can you describe the logic you implemented here? I don't see why for example a `MATCH` as a last clause should prevent the empty result. Also, shouldn't this be only relevant for the last clause of the last query part?

antaljanosbenjamin (commented):
The operator looks good, but on the planner side I am not sure if the logic is correct.

src/query/plan/rule_based_planner.hpp
as51340: Yes so it isn't relevant only to the last clause of the last query part because of the with, unwind and load csv clause. In this way we avoid having `EmptyResult` operator in the middle of the query plan.

as51340:
Clang-tidy is failing because of MACRO which was defined long time ago and I think it cannot be suppressed with any `hicpp` flags from what I've seen.

gitbuda (approved):
The e2e test I've mentioned has been added so I'll approve this PR

src/query/plan/operator.cpp
```diff
@@ -3057,6 +3058,54 @@ void EdgeUniquenessFilter::EdgeUniquenessFilterCursor::Shutdown() { input_cursor
...
+
+  bool Pull(Frame &frame, ExecutionContext &context) override {
+    SCOPED_PROFILE_OP("EmptyResult");
+
+    if (!pulled_all_input_) {
+      while (input_cursor_->Pull(frame, context)) {
+      }
+      pulled_all_input_ = true;
+    }
+    if (MustAbort(context)) throw HintedAbortError();
```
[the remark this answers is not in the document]
as51340: Done

jbajic (approved):

PR #668, opened 2022-11-28
[:pineapple: < MG-T1173] Benchmark datastructures

jbajic:
[master < Task] PR

__Commit message:__
- Add benchmarks comparing insert, remove, find and contains for std::map, std::set and utils::SkipList
- Create python plot script
- Create bash helper script for executing and plotting bemchmarks


jbajic:
I have removed b++ tree from benchmarks, since it is not compilable, if anyone thinks it is worthwhile to keep to implementation for the sake of doing benchmarks again, please say so

tests/benchmark/data_structures_insert.cpp
```diff
@@ -0,0 +1,92 @@
...
+  coordinator::Hlc start_timestamp;
+  storage::v3::IsolationLevel isolation_level{storage::v3::IsolationLevel::SNAPSHOT_ISOLATION};
+  storage::v3::Transaction transaction{start_timestamp, isolation_level};
+  auto *delta = storage::v3::CreateDeleteObjectDelta(&transaction);
+
+  for (auto _ : state) {
+    for (auto i{0}; i < state.range(0); ++i) {
+      std_map.insert({storage::v3::PrimaryKey{storage::v3::PropertyValue{i}},
+                      storage::v3::LexicographicallyOrderedVertex{storage::v3::Vertex{
+                          delta, std::vector<storage::v3::PropertyValue>{storage::v3::PropertyValue{i}}}}});
```
spacejam, line 65: I might change the value type to something as close to a zero-sized type as possible, like a bool etc...

spacejam (approved):
only a minor point about using a smaller value for std::map

tests/benchmark/data_structures_insert.cpp
jbajic: You mean of the properties, and not the primary key right?

tests/benchmark/data_structures_common.hpp
```diff
@@ -0,0 +1,61 @@
...
+    auto acc = skip_list.access();
+    acc.insert({storage::v3::PrimaryKey{storage::v3::PropertyValue{i}}});
+  }
+}
+
+template <typename TKey, typename TValue>
+inline void PrepareData(std::map<TKey, TValue> &std_map, const int64_t num_elements) {
+  coordinator::Hlc start_timestamp;
+  storage::v3::IsolationLevel isolation_level{storage::v3::IsolationLevel::SNAPSHOT_ISOLATION};
+  storage::v3::Transaction transaction{start_timestamp, isolation_level};
```
kostasrim, line 42: I am confused, why are these needed here?

tests/benchmark/data_structures_insert.cpp
```diff
@@ -0,0 +1,92 @@
...
+  }
+}
+
+static void BM_BenchmarkInsertStdSet(::benchmark::State &state) {
+  std::set<storage::v3::PrimaryKey> std_set;
+  coordinator::Hlc start_timestamp;
+  storage::v3::IsolationLevel isolation_level{storage::v3::IsolationLevel::SNAPSHOT_ISOLATION};
+  storage::v3::Transaction transaction{start_timestamp, isolation_level};
+  auto *delta = storage::v3::CreateDeleteObjectDelta(&transaction);
+
```
kostasrim, line 76: are these needed here?

tests/benchmark/data_structures_remove.cpp
```diff
@@ -0,0 +1,106 @@
...
+namespace memgraph::benchmark {
+
+///////////////////////////////////////////////////////////////////////////////
+// Testing Remove Operation
+///////////////////////////////////////////////////////////////////////////////
+static void BM_BenchmarkRemoveSkipList(::benchmark::State &state) {
+  utils::SkipList<storage::v3::PrimaryKey> skip_list;
+  PrepareData(skip_list, state.range(0));
+
+  // So we can also have elements that does don't exist
```
kostasrim, line 44:
```suggestion
  // So we can also have elements that don't exist
```

tests/benchmark/skip_list_common.hpp
```diff
@@ -81,3 +81,27 @@ inline void RunConcurrentTest(std::function<void(std::atomic<bool> *, Stats *)>
...
+  }
+
+  std::cout << "    Operations: " << stats.total << std::endl;
+  std::cout << "    Successful insert: " << stats.succ[0] << std::endl;
+  std::cout << "    Successful contains: " << stats.succ[1] << std::endl;
+  std::cout << "    Successful remove: " << stats.succ[2] << std::endl;
+  std::cout << "    Successful find: " << stats.succ[3] << std::endl;
+  std::cout << std::endl;
+
+  const auto tot = stats.succ[0] + stats.succ[1] + stats.succ[2] + stats.succ[3];
```
kostasrim, line 102:
```suggestion
  const auto tot = std::accumulate(stats.succ.begin(),  + stats.succ.begin() + 3, 0);
```

tests/benchmark/data_structures_insert.cpp
jbajic: You are right it is not needed for deltas

tests/benchmark/data_structures_insert.cpp
jbajic: After short private exchange we concluded it is a better to use a smallest possible type as value to simulate similar memory structure as set in map

kostasrim (approved):

PR #676, opened 2022-11-28
[:pineapple: < MG-T1163] Impl of Multiframe and iterators

42jeremy:
[master < Task] PR
- [ ] Check, and update documentation if necessary: no documentation needed
```Implementation of Multiframe```

### CODE REVIEW

File: src/query/v2/plan/operator.lcp [hunk 1]
```diff
@@ -332,12 +335,13 @@ and false on every following Pull.")
...
     public:
      OnceCursor() {}
+     void PullMultiple(MultiFrame &, ExecutionContext &) override;
      bool Pull(Frame &, ExecutionContext &) override;
      void Shutdown() override;
      void Reset() override;
 
     private:
-     bool did_pull_{false};
+     size_t pull_count_{false};
```
42jeremy, line 344:
```suggestion
     size_t pull_count_{0};
```

src/query/v2/multiframe.cpp
```diff
@@ -0,0 +1,120 @@
...
+// licenses/APL.txt.
+
+#include "query/v2/multiframe.hpp"
+
+#include "query/v2/bindings/frame.hpp"
+#include "utils/pmr/vector.hpp"
+
+namespace memgraph::query::v2 {
+
+MultiFrame::MultiFrame(FrameWithValidity default_frame, size_t number_of_frames,
```
antaljanosbenjamin, line 19: Maybe instead of `default_frame` we can use only the width of the `Frame` and construct a frame to ourselves. Basically the only information we need to construct an empty frame is how big should it be.

src/query/v2/multiframe.cpp
```diff
@@ -0,0 +1,120 @@
...
+      frames_(utils::pmr::vector<FrameWithValidity>(number_of_frames, default_frame, execution_memory)) {
+  MG_ASSERT(number_of_frames > 0);
+  MG_ASSERT(!default_frame.IsValid());
+}
+
+MultiFrame::MultiFrame(const MultiFrame &other) : default_frame_(other.default_frame_) {
+  /*
+  TODO
+  Do we just copy all frames or do we make distinctions between valid and not valid frames? Does it make any
+  difference?
```
antaljanosbenjamin, line 31: I think we should just copy and don't care about the valid/invalid state.

src/query/v2/multiframe.cpp
```diff
@@ -0,0 +1,120 @@
...
+ValidFramesModifier::Iterator ValidFramesModifier::end() {
+  return Iterator{&multiframe_.frames_[multiframe_.frames_.size()], *this};
+}
+
+ValidFramesConsumer::ValidFramesConsumer(MultiFrame &multiframe) : multiframe_(multiframe) {}
+
+// NOLINTNEXTLINE (bugprone-exception-escape)
+ValidFramesConsumer::~ValidFramesConsumer() noexcept {
+  // TODO Possible optimisation: only DefragmentValidFrames if one frame has been invalidated? Only if does not
+  // cost too much to store it
```
antaljanosbenjamin, line 94: That is a valid optimization. If it takes only one bool, then you can easily do it. However we can push that without it and improve on this later if it is too complicated.

src/query/v2/multiframe.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+ValidFramesConsumer MultiFrame::GetValidFramesConsumer() { return ValidFramesConsumer{*this}; }
+
+InvalidFramesPopulator MultiFrame::GetInvalidFramesPopulator() { return InvalidFramesPopulator{*this}; }
+
+ValidFramesReader::ValidFramesReader(MultiFrame &multiframe) : multiframe_(multiframe) {}
+
+ValidFramesReader::Iterator ValidFramesReader::begin() { return Iterator{&multiframe_.frames_[0], *this}; }
+ValidFramesReader::Iterator ValidFramesReader::end() {
+  return Iterator{&multiframe_.frames_[multiframe_.frames_.size()], *this};
```
antaljanosbenjamin, line 79: Here you basically get the address of an object referenced by an invalid reference (because you index into the vector with `size()`). I would instead do `multiframe_.frames_.data() + multiframe_.frames_.size()`. This way it is totally fine until the pointer is not dereferenced. Of course, if you apply my suggestion about how to construct `ValidFramesReader` this comment might become obsolete for this iterator, but it remains true for others.

src/query/v2/multiframe.cpp
```diff
@@ -0,0 +1,120 @@
...
+}
+
+bool MultiFrame::HasValidFrame() const noexcept {
+  return std::any_of(frames_.begin(), frames_.end(), [](auto &frame) { return frame.IsValid(); });
+}
+
+// NOLINTNEXTLINE (bugprone-exception-escape)
+void MultiFrame::DefragmentValidFrames() noexcept {
+  /*
+  from: https://en.cppreference.com/w/cpp/algorithm/remove
```
antaljanosbenjamin, line 58: It is a perfect solution, nicely done! :rocket:

src/query/v2/multiframe.cpp
```diff
@@ -0,0 +1,120 @@
...
+}
+
+InvalidFramesPopulator::InvalidFramesPopulator(MultiFrame &multiframe) : multiframe_(multiframe) {}
+
+InvalidFramesPopulator::Iterator InvalidFramesPopulator::begin() {
+  for (auto idx = 0UL; idx < multiframe_.frames_.size(); ++idx) {
+    if (!multiframe_.frames_[idx].IsValid()) {
+      return Iterator{&multiframe_.frames_[idx]};
+    }
+  }
```
antaljanosbenjamin, line 111: Why not a foreach? It deals with the ranges and you can get the address of the object if you (and you should) iterate over the `Frame`s using references.

src/query/v2/plan/operator.lcp
```diff
@@ -332,12 +335,13 @@ and false on every following Pull.")
...
     public:
      OnceCursor() {}
+     void PullMultiple(MultiFrame &, ExecutionContext &) override;
      bool Pull(Frame &, ExecutionContext &) override;
      void Shutdown() override;
      void Reset() override;
 
     private:
-     bool did_pull_{false};
+     size_t pull_count_{0};
```
antaljanosbenjamin, line 344: Why did you change this from a single bool to a counter?

src/query/v2/plan/operator.cpp
```diff
@@ -257,13 +257,30 @@ class DistributedCreateNodeCursor : public Cursor {
...
+  auto iterator_for_valid_frame_only = multi_frame.GetValidFramesConsumer();
+  auto first_it = iterator_for_valid_frame_only.begin();
+  MG_ASSERT(first_it != iterator_for_valid_frame_only.end());
+  if (pull_count_ < 1) {
+    auto *memory_resource = multi_frame.GetMemoryResource();
+    auto &frame = *first_it;
+    frame.MakeValid();
+    for (auto &value : frame.elems()) {
+      value = TypedValue{memory_resource};
+    }
```
antaljanosbenjamin, line 279: I assume you get this from my experiment branch. I did a mistake there. The `OnceCursor` shouldn't modify the `Frame` in the "old approach" because of the optional match. I think we will have other issues with the optional match, but this is for sure will be an issue. So maybe just call `MakeValid` on the first `Frame` of the `MultiFrame` (without checkout if it is valid or not, we shouldn't care in my opinion)

antaljanosbenjamin (changes requested):

src/query/v2/multiframe.hpp
```diff
@@ -0,0 +1,276 @@
...
+  MultiFrame(MultiFrame &&other) noexcept;  // move constructor
+  MultiFrame &operator=(const MultiFrame &other) = delete;
+  MultiFrame &operator=(MultiFrame &&other) noexcept = delete;
+
+  /*!
+  Returns a object on which one can iterate in a for-loop. By doing so, you will only get frames that are in a valid
+  state in the multiframe.
+  Iteration goes in a deterministic order.
+  One can't modify the validity of the frame nor its content with this implementation.
+  */
```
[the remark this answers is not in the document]
42jeremy: My mistake, old habit kicked in.

src/query/v2/multiframe.cpp
```diff
@@ -0,0 +1,120 @@
...
+MultiFrame::MultiFrame(MultiFrame &&other) noexcept
+    : default_frame_(std::move(other.default_frame_)), frames_(std::move(other.frames_)) {}
+
+void MultiFrame::ResetAllFramesInvalid() noexcept {
+  std::for_each(frames_.begin(), frames_.end(), [](auto &frame) { frame.MakeInvalid(); });
+}
+
+bool MultiFrame::HasValidFrame() const noexcept {
+  return std::any_of(frames_.begin(), frames_.end(), [](auto &frame) { return frame.IsValid(); });
+}
```
jbajic, line 53: I wouldn't be so sure these are `noexcept` https://en.cppreference.com/w/cpp/algorithm/for_each#Exceptions

src/query/v2/multiframe.hpp
```diff
@@ -0,0 +1,276 @@
...
+    Iterator(internal_ptr ptr, ValidFramesReader &iterator_wrapper) : ptr_(ptr), iterator_wrapper_(iterator_wrapper) {}
+
+    reference operator*() const { return *ptr_; }
+    pointer operator->() { return ptr_; }
+
+    // Prefix increment
+    Iterator &operator++() {
+      do {
+        ptr_++;
+      } while (!this->ptr_->IsValid() && *this != iterator_wrapper_.end());
```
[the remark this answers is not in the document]
42jeremy: > The order of the conditions should be reversed. This way it is possible that we dereference and end iterator which can point after the last element, so it might result in an invalid memory access.

This is true, I have updated all of them (regardless of the second remark)

src/query/v2/multiframe.cpp
42jeremy: I have also removed default_frame, if we don't care about valid/invalid state there's no need to keep it

src/query/v2/multiframe.cpp
42jeremy: (I think you mean `for(auto &frame: multiframe_.frames_)` and not `std::for_each(begin, end, UnaryFunction)`)
-> I can't remember the reason why I use idx based for loop instead of simple range for loop, I will replace it

src/query/v2/plan/operator.lcp
42jeremy: Unsure, I'll replace it

src/query/v2/multiframe.cpp
42jeremy: I would tend to think it's fine as we are not allocating any memory here and isValid/MakeInvalid do not throw exception
(Ps: i have just marked MakeInvalid and IsValid as noexpect)

src/query/v2/multiframe.cpp
```diff
@@ -0,0 +1,116 @@
...
+// licenses/APL.txt.
+
+#include "query/v2/multiframe.hpp"
+
+#include "query/v2/bindings/frame.hpp"
+#include "utils/pmr/vector.hpp"
+
+namespace memgraph::query::v2 {
+
+// #NoCommit uncomment https://github.com/memgraph/memgraph/pull/676#discussion_r1035704661
```
42jeremy, line 19: Uncomment

src/expr/interpret/frame.hpp
```diff
@@ -42,4 +42,19 @@ class Frame {
   utils::pmr::vector<TypedValue> elems_;
 };
 
+template <typename TypedValue>
```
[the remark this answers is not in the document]
antaljanosbenjamin: You are right, it is not necessary. The `TypedValue` itself is templatized to fit both the query engine and storage engine. However for this PR (as we would like to merge this as soon as possible) I might not necessary go into it. Once we merged this PR I am happy to create a small PR that only removes the template parameter from `Frame` and `FrameWithValidity`.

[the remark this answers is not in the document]
jbajic: I don't see this as anything critical, so I will mark this as resolved

kostasrim (commented):
Goood work, I really like what you did with the iterators!

src/query/v2/multiframe.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+// NOLINTNEXTLINE (bugprone-exception-escape)
+void MultiFrame::DefragmentValidFrames() noexcept {
+  /*
+  from: https://en.cppreference.com/w/cpp/algorithm/remove
+  "Removing is done by shifting (by means of copy assignment (until C++11)move assignment (since C++11)) the elements
+  in the range in such a way that the elements that are not to be removed appear in the beginning of the range.
+  Relative order of the elements that remain is preserved and the physical size of the container is unchanged."
+  */
+  [[maybe_unused]] const auto it =
```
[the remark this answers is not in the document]
42jeremy: This was to avoid the warning about ignore returned variable. I'll add   // NOLINTNEXTLINE (bugprone-unused-return-value) instead

src/query/v2/multiframe.hpp
```diff
@@ -0,0 +1,272 @@
...
+  InvalidFramesPopulator GetInvalidFramesPopulator();
+
+  void MakeAllFramesInvalid() noexcept;
+
+  bool HasValidFrame() const noexcept;
+
+  inline utils::MemoryResource *GetMemoryResource() { return frames_[0].GetMemoryResource(); }
+
+ private:
+  // NOLINTNEXTLINE (bugprone-exception-escape)
```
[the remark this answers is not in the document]
42jeremy: We need it, otherwise we get the following clang-warning:
```
memgraph/src/query/v2/multiframe.cpp:93:22: warning: an exception may be thrown in function '~ValidFramesConsumer' which should not throw exceptions [bugprone-exception-escape]
ValidFramesConsumer::~ValidFramesConsumer() noexcept {
```
Which is valid from his point of view (we use erase_if which might throw an exception). However in our case we woudn't allocate any memory

src/query/v2/multiframe.hpp
```diff
@@ -0,0 +1,272 @@
...
+  void DefragmentValidFrames() noexcept;
```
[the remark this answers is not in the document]
42jeremy: Same than here(https://github.com/memgraph/memgraph/pull/676#discussion_r1037128827)

src/query/v2/multiframe.cpp
```diff
@@ -0,0 +1,116 @@
...
+  MG_ASSERT(number_of_frames > 0);
+}
+
+MultiFrame::MultiFrame(const MultiFrame &other) {
+  frames_.reserve(other.frames_.size());
+  std::transform(other.frames_.begin(), other.frames_.end(), std::back_inserter(frames_),
+                 [](const auto &other_frame) { return other_frame; });
+}
+
+// NOLINTNEXTLINE (bugprone-exception-escape)
```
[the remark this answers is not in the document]
42jeremy: Same than here(https://github.com/memgraph/memgraph/pull/676#discussion_r1037128827)

src/query/v2/multiframe.cpp
```diff
@@ -0,0 +1,116 @@
...
+MultiFrame::MultiFrame(MultiFrame &&other) noexcept : frames_(std::move(other.frames_)) {}
```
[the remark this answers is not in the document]
42jeremy: Without it, we get: 
```
memgraph/src/query/v2/multiframe.cpp:41:13: warning: move constructors should be marked noexcept [hicpp-noexcept-move,performance-noexcept-move-constructor]
MultiFrame::MultiFrame(MultiFrame &&other) : frames_(std::move(other.frames_)) {}
```

src/query/v2/multiframe.hpp
```diff
@@ -0,0 +1,272 @@
...
+
+  utils::pmr::vector<FrameWithValidity> frames_ =
+      utils::pmr::vector<FrameWithValidity>(0, FrameWithValidity{1}, utils::NewDeleteResource());
+};
```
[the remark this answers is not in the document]
42jeremy: I think that gives a compilation error, i'll investigate.

src/query/v2/multiframe.cpp
```diff
@@ -0,0 +1,115 @@
...
+    : frames_(utils::pmr::vector<FrameWithValidity>(
+          number_of_frames, FrameWithValidity(size_of_frame, execution_memory), execution_memory)) {
+  MG_ASSERT(number_of_frames > 0);
+}
+
+MultiFrame::MultiFrame(const MultiFrame &other) {
+  frames_.reserve(other.frames_.size());
+  std::transform(other.frames_.begin(), other.frames_.end(), std::back_inserter(frames_),
+                 [](const auto &other_frame) { return other_frame; });
+}
```
antaljanosbenjamin, line 39:
```suggestion
MultiFrame::MultiFrame(const MultiFrame &other) :frames_{other.frames_} {}
```

Please format to be sure it is fine.

src/query/v2/multiframe.hpp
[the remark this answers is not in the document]
42jeremy: Reason was due to the move constructor, already solved by: https://github.com/memgraph/memgraph/pull/676#discussion_r1037170671

src/query/v2/multiframe.hpp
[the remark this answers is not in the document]
42jeremy: Heads up: I've implemented a version where we only need to know about the "after last valid frame"
Given that valid frames are at the beginning of the vector, we don't need to keep the first valid frame

src/query/v2/multiframe.hpp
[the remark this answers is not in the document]
kostasrim: uh makes sense

kostasrim (approved):

jbajic (approved):

src/query/v2/multiframe.hpp
```diff
@@ -0,0 +1,276 @@
...
+  explicit ValidFramesReader(MultiFrame &multiframe);
+
+  ~ValidFramesReader() = default;
+  ValidFramesReader(const ValidFramesReader &other) = delete;                 // copy constructor
+  ValidFramesReader(ValidFramesReader &&other) noexcept = delete;             // move constructor
+  ValidFramesReader &operator=(const ValidFramesReader &other) = delete;      // copy assignment
+  ValidFramesReader &operator=(ValidFramesReader &&other) noexcept = delete;  // move assignment
+
+  struct Iterator {
+    using iterator_category = std::forward_iterator_tag;
```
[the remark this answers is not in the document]
42jeremy: Done (Kostas gave me a hand there)

src/query/v2/multiframe.hpp
```diff
@@ -0,0 +1,311 @@
...
+  ValidFramesReader &operator=(ValidFramesReader &&other) noexcept = delete;
+
+  struct Iterator {
+    using iterator_category = std::forward_iterator_tag;
+    using difference_type = std::ptrdiff_t;
+    using value_type = const Frame;
+    using pointer = value_type *;
+    using reference = const Frame &;
+
+    Iterator() {}
```
antaljanosbenjamin, line 110:
```suggestion
    Iterator() = default;
```

This might be a breaker for being [trivial class type](https://en.cppreference.com/w/cpp/language/classes#Trivial_class) or not, because the default constructor has to be trivial, which means it cannot be user defined (even if it is dead simple like `{}`). As this is not a correctness issue, we can merge this PR like this, but please take care of this as soon as possible.

src/query/v2/multiframe.hpp
```diff
@@ -107,22 +107,31 @@ class ValidFramesReader {
...
       ptr_++;
       return *this;
     }
 
-    friend bool operator==(const Iterator &a, const Iterator &b) { return a.ptr_ == b.ptr_; };
-    friend bool operator!=(const Iterator &a, const Iterator &b) { return a.ptr_ != b.ptr_; };
+    // clang-tidy warning is wrong here, because we & qualify the function, meaning that you can't post increment
+    // temporaries, e.g, (it++)++
+    // NOLINTNEXTLINE (cert-dcl21-cpp)
+    Iterator operator++(int) & {
```
antaljanosbenjamin, line 124: I think the `&` qualifier is questionable because of two reasons:
1. In `src/io/network/addrinfo.hppsrc/io/network/addrinfo.hpp` we don't do that, just suppress the warning.
2. It is not done either in [`std` iterators](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:1,endLineNumber:8,positionColumn:1,positionLineNumber:8,selectionStartColumn:1,selectionStartLineNumber:8,startColumn:1,startLineNumber:8),source:'%23include+%3Cvector%3E%0A%0Aint+main()+%7B%0A++++std::vector%3Cint%3E+vec%3B%0A++++auto+it+%3D+(vec.begin()%2B%2B)%2B%2B%3B%0A++++return+*it%3B%0A%7D%0A'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),header:(),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:compiler,i:(compiler:clang1301,deviceViewOpen:'1',filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'-std%3Dc%2B%2B20+-Wall+-Weverything+-Wno-c%2B%2B98-compat+-Wno-c%2B%2B98-compat-pedantic',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+clang+13.0.1+(Editor+%231)',t:'0')),header:(),k:50,l:'4',m:50,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+clang+13.0.1',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+clang+13.0.1+(Compiler+%231)',t:'0')),header:(),l:'4',m:50,n:'0',o:'',s:0,t:'0')),k:50,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4), so somebody might be surprised by that.

As this is also not a concern from the correctness point of view, I think the PR can be merged as is and this should be taken care of right after the merge.

src/query/v2/plan/operator.cpp
antaljanosbenjamin: Here you still modify the values on the `Frame` which doesn't work well with the optional cursor. I think lines 277-279 should be deleted:
```suggestion
    auto &frame = *first_it;
    frame.MakeValid();
```

antaljanosbenjamin (changes requested):

src/query/v2/multiframe.hpp
kostasrim: suppressing the warning can lead to ub. `(it++)++` boom you used a temporary.

kostasrim: makes sense, as long as we keep the return type to copy by value

PR #680, opened 2022-11-30
[:pineapple: < MG-T1185] Replace skip list

jbajic:
[master < Task] PR

__Commit message:__
- Replace `SkipList` with `std::map` as vertex container
- Replace `SkipList` with `std::map` as edge container
- Replace `SkipList` with `std::set` as `LabelPropertyIndex` container
- Replace `SkipList` with `std::set` as `LabelIndex` container
- Remove `KeyStore` and `LexicographiOrderedVertex`


src/storage/v3/vertex.hpp
```diff
@@ -31,14 +31,14 @@ namespace memgraph::storage::v3 {
...
-  Vertex(Delta *delta, const std::vector<PropertyValue> &primary_properties) : keys{primary_properties}, delta{delta} {
+  Vertex(Delta *delta, PrimaryKey primary_properties) : keys{std::move(primary_properties)}, delta{delta} {
     MG_ASSERT(delta == nullptr || delta->action == Delta::Action::DELETE_OBJECT,
               "Vertex must be created with an initial DELETE_OBJECT delta!");
   }
 
   friend bool operator==(const Vertex &vertex, const PrimaryKey &primary_key) { return vertex.keys == primary_key; }
 
-  KeyStore keys;
+  PrimaryKey keys;
```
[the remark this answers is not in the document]
jbajic: The duplication of PrimaryKeys was removed, without having a pointer to the Primary Key, but instead renaming the current `Vertex` to `VertexData` and relying use Primary Key as key from `std::map` as key

src/storage/v3/indices.hpp
```diff
@@ -59,7 +60,7 @@ class LabelIndex {
   void UpdateOnAddLabel(LabelId label, Vertex *vertex, const Transaction &tx);
 
   /// @throw std::bad_alloc
-  bool CreateIndex(LabelId label, VerticesSkipList::Accessor vertices);
+  bool CreateIndex(LabelId label, VertexContainer &vertices);
```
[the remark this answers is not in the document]
jbajic: The issue with `LabelIndex` is that there is no order, and using any other structures will slow it down since lookup is important on update/delete. I would just keep `SkipList` for now

src/storage/v3/property_value.hpp
```diff
@@ -330,6 +330,36 @@ inline bool operator<(const PropertyValue &first, const PropertyValue &second) {
   }
 }
 
+inline bool operator>=(const PropertyValue &first, const PropertyValue &second) {
+  if (!PropertyValue::AreComparableTypes(first.type(), second.type())) return first.type() >= second.type();
+  switch (first.type()) {
```
jbajic, line 335: This is needed for comparisons in indices

src/storage/v3/shard.hpp
```diff
@@ -227,36 +225,31 @@ class Shard final {
                               const std::optional<utils::Bound<PropertyValue>> &lower_bound,
                               const std::optional<utils::Bound<PropertyValue>> &upper_bound, View view);
 
-    /// Return approximate number of all vertices in the database.
-    /// Note that this is always an over-estimate and never an under-estimate.
-    int64_t ApproximateVertexCount() const { return static_cast<int64_t>(shard_->vertices_.size()); }
+    /// Return number of all vertices in the database.
+    int64_t VertexCount() const { return static_cast<int64_t>(shard_->vertices_.size()); }
```
jbajic, line 229: I think I made a mistake here, it is still approximate, since it does not verify if the vertex is deleted or not, I am right @antaljanosbenjamin ?

antaljanosbenjamin: Yes, that is true. Furthermore be suspicious about errors regarding deleted vertices. I think this change shouldn't cause any issues, because we only removed the vertices from the `SkipList` when we were sure there is no active transaction that should access the vertex, but it is possible that the lack of `SkipList` (concretely that accessors of the `SkipList` will provide access to items that are removed by a future accessor) will cause some issues.

src/storage/v3/indices.cpp
```diff
@@ -278,9 +284,8 @@ bool LabelIndex::CreateIndex(LabelId label, VerticesSkipList::Accessor vertices)
   }
   try {
     auto acc = it->second.access();
-    for (auto &lgo_vertex : vertices) {
-      auto &vertex = lgo_vertex.vertex;
-      if (vertex.deleted || !utils::Contains(vertex.labels, label)) {
+    for ([[maybe_unused]] auto &vertex : vertices) {
```
antaljanosbenjamin, line 287: Why `[[maybe_unused]]`?

src/storage/v3/indices.cpp
```diff
@@ -425,17 +422,15 @@ bool LabelPropertyIndex::CreateIndex(LabelId label, PropertyId property, Vertice
     return false;
   }
   try {
-    auto acc = it->second.access();
-    for (auto &lgo_vertex : vertices) {
-      auto &vertex = lgo_vertex.vertex;
-      if (vertex.deleted || !utils::Contains(vertex.labels, label)) {
+    for ([[maybe_unused]] auto &vertex : vertices) {
```
antaljanosbenjamin, line 425: Same as for above, just a reminder that this happens here also.

src/storage/v3/indices.cpp
```diff
@@ -537,13 +531,12 @@ const PropertyValue kSmallestMap = PropertyValue(std::map<std::string, PropertyV
 const PropertyValue kSmallestTemporalData =
     PropertyValue(TemporalData{static_cast<TemporalType>(0), std::numeric_limits<int64_t>::min()});
 
-LabelPropertyIndex::Iterable::Iterable(utils::SkipList<Entry>::Accessor index_accessor, LabelId label,
-                                       PropertyId property,
+LabelPropertyIndex::Iterable::Iterable(LabelPropertyIndexContainer &index_accessor, LabelId label, PropertyId property,
```
antaljanosbenjamin, line 534: I would rename both this variable and the member variable from `index_accessor`.

src/storage/v3/indices.cpp
```diff
@@ -651,49 +644,63 @@ LabelPropertyIndex::Iterable::Iterator LabelPropertyIndex::Iterable::begin() {
...
-    return {this, index_accessor_.end()};
+    return {this, index_accessor_->end()};
   }
-  auto index_iterator = index_accessor_.begin();
+  auto index_iterator = index_accessor_->begin();
   if (lower_bound_) {
-    index_iterator = index_accessor_.find_equal_or_greater(lower_bound_->value());
+    index_iterator = std::ranges::find_if(*index_accessor_, [lower_bound = lower_bound_->value()](const auto &pair) {
+      return pair.value >= lower_bound;
+    });
```
antaljanosbenjamin, line 653: Check `std::ranges::lower_bound` [here](https://en.cppreference.com/w/cpp/algorithm/ranges/lower_bound), it might be more efficient (lg vs linear complexity.)

src/storage/v3/indices.cpp
```diff
@@ -651,49 +644,63 @@ LabelPropertyIndex::Iterable::Iterator LabelPropertyIndex::Iterable::begin() {
...
-  // The value `Null` won't ever appear in the index because it indicates that
-  // the property shouldn't exist. Instead, this value is used as an indicator
-  // to estimate the average number of equal elements in the list (for any
-  // given value).
-  return static_cast<int64_t>(acc.estimate_average_number_of_equals(
-      [](const auto &first, const auto &second) { return first.value == second.value; },
-      static_cast<int>(utils::SkipListLayerForAverageEqualsEstimation(acc.size()))));
+  // TODO Do check this
+  return static_cast<int64_t>(
+      std::ranges::count_if(it->second, [&value](const auto &elem) { return elem.value == value; }));
```
antaljanosbenjamin, line 669: Yeah, this doesn't work. Maybe for now just leave an `MG_ASSERT` for this.

src/storage/v3/indices.cpp
```diff
@@ -651,49 +644,63 @@ LabelPropertyIndex::Iterable::Iterator LabelPropertyIndex::Iterable::begin() {
...
-int64_t LabelPropertyIndex::ApproximateVertexCount(LabelId label, PropertyId property,
-                                                   const PropertyValue &value) const {
+int64_t LabelPropertyIndex::VertexCount(LabelId label, PropertyId property, const PropertyValue &value) const {
   auto it = index_.find({label, property});
   MG_ASSERT(it != index_.end(), "Index for label {} and property {} doesn't exist", label.AsUint(), property.AsUint());
-  auto acc = it->second.access();
   if (!value.IsNull()) {
     return static_cast<int64_t>(
-        acc.estimate_count(value, static_cast<int>(utils::SkipListLayerForCountEstimation(acc.size()))));
+        std::ranges::count_if(it->second, [&value](const auto &elem) { return elem.value == value; }));
```
antaljanosbenjamin, line 665: I would instead do a `std::ranges::lower_bound` and then march the items until the value matches. That way (assuming good distribution of values) the complexity would go from linear to close to logarithmic.

src/storage/v3/mvcc.hpp
```diff
@@ -11,9 +11,13 @@
 
 #pragma once
 
+#include <type_traits>
+#include "storage/v3/edge.hpp"
```
antaljanosbenjamin, line 15:
```suggestion
#include <type_traits>

#include "storage/v3/edge.hpp"
```

src/storage/v3/mvcc.hpp
```diff
@@ -77,10 +81,18 @@ inline void ApplyDeltasForRead(Transaction *transaction, const Delta *delta, Vie
...
+requires utils::SameAsAnyOf<TObj, Edge, Vertex>
 inline bool PrepareForWrite(Transaction *transaction, TObj *object) {
-  if (object->delta == nullptr) return true;
+  auto *delta_holder = std::invoke([object]() -> auto * {
+    if constexpr (std::is_same_v<TObj, Vertex>) {
+      return &object->second;
+    } else {
+      return object;
+    }
+  });
```
antaljanosbenjamin, line 92: Maybe make a function from this as this is used throughout this file. You might even use overloading:
```
VertexData* DeltaHolder(Vertex *vertex) {
    return vertex->second;
}

Edge* DeltaHolder(Edge* edge) {
    return edge;
}
```

Whichever is more suitable in your mind.

src/storage/v3/shard.cpp
```diff
@@ -1022,15 +1017,13 @@ void Shard::CollectGarbage(const io::Time current_time) {
     RemoveObsoleteEntries(&indices_, clean_up_before_timestamp);
   }
 
-  auto vertex_acc = vertices_.access();
   for (const auto &vertex : deleted_vertices_) {
-    MG_ASSERT(vertex_acc.remove(vertex), "Invalid database state!");
+    MG_ASSERT(vertices_.erase(*vertex), "Invalid database state!");
   }
```
antaljanosbenjamin, line 1022:
```suggestion
  for (const auto *vertex : deleted_vertices_) {
    MG_ASSERT(vertices_.erase(*vertex), "Invalid database state!");
  }
```

Do not take a reference to a pointer, but copy it.

antaljanosbenjamin (changes requested):

src/storage/v3/property_value.hpp
antaljanosbenjamin: Maybe exploit the fact that `(a >= b) <=> !(a < b)` to reduce code duplication and complexity.

src/storage/v3/indices.cpp
jbajic: I forgot to remove it, before it was destructured into pk and vertex_data

src/storage/v3/indices.cpp
jbajic: Hmm I did not know that, interesting

src/storage/v3/indices.cpp
```diff
@@ -651,49 +639,54 @@ LabelPropertyIndex::Iterable::Iterator LabelPropertyIndex::Iterable::begin() {
...
-int64_t LabelPropertyIndex::ApproximateVertexCount(LabelId label, PropertyId property,
-                                                   const std::optional<utils::Bound<PropertyValue>> &lower,
-                                                   const std::optional<utils::Bound<PropertyValue>> &upper) const {
-  auto it = index_.find({label, property});
-  MG_ASSERT(it != index_.end(), "Index for label {} and property {} doesn't exist", label.AsUint(), property.AsUint());
-  auto acc = it->second.access();
+  auto start_it = std::ranges::lower_bound(it->second, value, std::less{}, &Entry::value);
   return static_cast<int64_t>(
-      acc.estimate_range_count(lower, upper, static_cast<int>(utils::SkipListLayerForCountEstimation(acc.size()))));
+      std::ranges::count_if(start_it, it->second.end(), [&value](const auto &elem) { return elem.value == value; }));
```
antaljanosbenjamin, line 659: Just a note for the future, no need to fix it: maybe we can do something better with a different algorithm/hand-written for loop to "early exit", but that is totally not necessary right now.

antaljanosbenjamin (approved):

spacejam (approved):

src/storage/v3/indices.cpp
jbajic: I will leave a TODO

PR #681, opened 2022-11-30
[E206-MG < T1183-MG] Create user from env variables

niko4299:


src/memgraph.cpp
```diff
@@ -52,6 +53,7 @@
 #include "query/procedure/module.hpp"
 #include "query/procedure/py_module.hpp"
 #include "requests/requests.hpp"
+#include "spdlog/spdlog.h"
```
jbajic, line 56: Remove this it is already included

src/memgraph.cpp
```diff
@@ -424,6 +430,45 @@ void InitializeLogger() {
   CreateLoggerFromSink(sinks, ParseLogLevel());
 }
 
+std::pair<std::string, std::string> LoadUsernameAndPassword(std::string pass_file) {
```
jbajic, line 433: I would move this somewhere in utils

src/memgraph.cpp
```diff
@@ -424,6 +430,45 @@ void InitializeLogger() {
...
+  std::ifstream file(pass_file);
+  if (file.fail()) {
+    spdlog::warn("Problem with opening MG_PASSFILE, memgraph server will start without user");
+    return {"", ""};
```
jbajic, line 437:
```suggestion
    return {};
```

src/memgraph.cpp
jbajic:
```suggestion
std::pair<std::string, std::string> LoadUsernameAndPassword(const std::string& pass_file) {
```

src/memgraph.cpp
```diff
@@ -424,6 +430,45 @@ void InitializeLogger() {
...
+  }
+  std::vector<std::string> result;
+
+  std::string line;
+  std::getline(file, line);
+  size_t pos = 0;
+  std::string token;
+  std::string delimiter = ":";
```
jbajic, line 445:
```suggestion
  static constexpr std::string_view delimiter{":"};
```

src/memgraph.cpp
```diff
@@ -424,6 +430,45 @@ void InitializeLogger() {
...
+      line.erase(pos - 1, 1);
+      token += line.substr(0, pos);
+      line.erase(0, pos);
+
+    } else {
+      token += line.substr(0, pos);
+      result.push_back(token);
+      line.erase(0, pos + delimiter.length());
+      token = "";
+    }
```
jbajic, line 457: I am not sure I understand why are we doing this

src/memgraph.cpp
```diff
@@ -424,6 +430,45 @@ void InitializeLogger() {
...
+  }
+  result.push_back(line);
+  file.close();
+
+  if (result.size() != 2) {
+    spdlog::warn(
+        "Wrong data format. Data should be store in format: username:password, memgraph server will start without "
+        "user");
+    return {"", ""};
```
jbajic, line 466:
```suggestion
    return {};
```

src/memgraph.cpp
```diff
@@ -880,6 +925,18 @@ int main(int argc, char **argv) {
   interpreter_context.auth = &auth_handler;
   interpreter_context.auth_checker = &auth_checker;
 
+  auto *maybe_username = std::getenv(MG_USER);
+  auto *maybe_password = std::getenv(MG_PASSWORD);
+  auto *maybe_pass_file = std::getenv(MG_PASSFILE);
+  if (maybe_username && maybe_password) {
+    auth_handler.CreateUser(maybe_username, maybe_password);
+  } else if (maybe_pass_file) {
+    auto [username, password] = LoadUsernameAndPassword(maybe_pass_file);
```
jbajic, line 934:
```suggestion
    const auto [username, password] = LoadUsernameAndPassword(maybe_pass_file);
```

tests/integration/env_variable_check/runner.py
```diff
@@ -0,0 +1,133 @@
...
+import atexit
+import os
+import subprocess
+import sys
+import tempfile
+import time
+from typing import List
+
+SCRIPT_DIR = os.path.dirname(os.path.realpath(__file__))
+PROJECT_DIR = os.path.normpath(os.path.join(SCRIPT_DIR, "..", "..", ".."))
```
jbajic, line 24: Please use `pathlib` for any paths, it  si much easier to read and to work with
```suggestion
SCRIPT_DIR = Path(__file__).absolute()
PROJECT_DIR = SCRIPT_DIR.parents[3]
```

tests/integration/env_variable_check/runner.py
```diff
@@ -0,0 +1,133 @@
...
+def start_memgraph(memgraph_args):
+    memgraph = subprocess.Popen(list(map(str, memgraph_args)))
+    time.sleep(0.1)
+    assert memgraph.poll() is None, "Memgraph process died prematurely!"
+    wait_for_server(7687)
+
+    return memgraph
+
+
+def execute_test(memgraph_binary: str, tester_binary: str) -> None:
```
jbajic, line 57: either use typing everywhere or don't use at all
```suggestion
def execute_test(memgraph_binary: str, tester_binary: str):
```

tests/integration/env_variable_check/runner.py
```diff
@@ -0,0 +1,133 @@
...
+    del os.environ["MG_PASSFILE"]
+    print("\033[1;36m~~ Ended env variable check test ~~\033[0m")
+    cleanup()
+
+    if "MG_USER" in return_to_prev_state:
+        os.environ["MG_USER"] = return_to_prev_state["MG_USER"]
+    if "MG_PASSWORD" in return_to_prev_state:
+        os.environ["MG_PASSWORD"] = return_to_prev_state["MG_PASSWORD"]
+    if "MG_PASSFILE" in return_to_prev_state:
+        os.environ["MG_PASSFILE"] = return_to_prev_state["MG_PASSFILE"]
```
jbajic, line 118: Move all of the separate test cases into separate functions, so the overview of test cases is easier

src/memgraph.cpp
niko4299: We are doing this because someone can have `:` in their username or password for example ni:ko:krvavica which then would lead to wrong parsing but with this if you put ni\:ko:krvavica you will get right username and password.

kostasrim: I think it's because of the `pass file`

kostasrim, line 457: Can you share the format of the pass file?

kostasrim (changes requested):
A password as a non encrypted env variable isn't a security hazard?

src/memgraph.cpp
niko4299: The pass file has the format `username:password`.

niko4299:
> Overall good work. Question, a password as a non encrypted env variable isn't a security hazard?

That is a valid question and I was going to let does who have it on the server take care of security but @gitbuda what do you think?

src/memgraph.cpp
jbajic: Aha ok maybe just remove that part somewhere else and possibly rewrite it something like this:
```suggestion
std::size_t ReplaceAllExceptLast(std::string& inout, std::string_view what, std::string_view with)
{
    std::size_t count{};
    auto last_occurance = inout.find_last_of(what);
    for (std::string::size_type pos{};
         inout.npos != (pos = inout.find(what.data(), pos, what.length())) && inout.find_last_of(what) != pos;
         pos += with.length(), ++count) {
        inout.replace(pos, what.length(), with.data(), with.length());
    }
    return count;
}
```
you have a similar example here https://en.cppreference.com/w/cpp/string/basic_string/replace, but honestly, I would just forbid `:` in the username, I think that character makes only sense in the password, but I leave that decision to you

gitbuda:
> Overall good work. Question, a password as a non encrypted env variable isn't a security hazard?

https://www.postgresql.org/docs/current/libpq-envars.html -> Postgres has it but it's not recommended. We also provide the file -> I would keep the current approach 😄

src/memgraph.cpp
```diff
@@ -98,6 +98,10 @@
 #include "audit/log.hpp"
 #endif
 
+const char *const MG_USER = "MG_USER";
+const char *const MG_PASSWORD = "MG_PASSWORD";
+const char *const MG_PASSFILE = "MG_PASSFILE";
```
jbajic, line 103:
```suggestion
constexpr char *const kMgUser = "MG_USER";
constexpr char *const kMgPassword = "MG_PASSWORD";
constexpr char *const kMgPassfile = "MG_PASSFILE";
```

niko4299: I had to change constexpr to const because of clang-tidy.

jbajic: What is the error from clang-tidy?

kostasrim: no no it was my mistake and maybe I did not communicate it very well on our private chat. The error was about the `pointer itself` being global and `non const`. I think it's safe to use `constexpr` here and clang-tidy won't complain as long as the `pointer` itself is `const` (not only a pointer to const but a const pointer to a const object)

jbajic: Ah yes, something like this `constexpr const char *kMgUser`

gitbuda: seems like this is resolved, resolving the conversation

jbajic (approved):

kostasrim (approved):

PR #682, opened 2022-11-30
[master <- ] Add support for -h to show help in addition to --help

spacejam:
before:
```
λ ./memgraph -h
ERROR: unknown command line flag 'h'
```

after is same as --help

### DIFF
--- a/src/memgraph.cpp
+++ b/src/memgraph.cpp
@@ -135,6 +135,7 @@ std::optional<Enum> StringToEnum(const auto &value, const auto &mappings) {
 }  // namespace
 
 // Short help flag.
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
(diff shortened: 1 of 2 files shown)


src/memgraph.cpp
```diff
@@ -134,6 +134,9 @@ std::optional<Enum> StringToEnum(const auto &value, const auto &mappings) {
 }
 }  // namespace
 
+// Short help flag.
+DEFINE_bool(h, false, "Print usage and exit.");
```
jbajic, line 138: Since the macro creates a global variable we need this just to avoid the clang-tidy warning
```suggestion
// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
DEFINE_bool(h, false, "Print usage and exit.");
```

kostasrim (approved):

PR #683, opened 2022-11-30
[E206-MG < T1184-MG] Init file before server starts and init data file after servers starts

niko4299:


tests/integration/flag_check/runner.py
```diff
@@ -0,0 +1,138 @@
...
+        storage_directory.name + "/init_file.txt",
+        "--init-data-file",
+        storage_directory.name + "/init_data_file.txt",
+    ]
+    memgraph = start_memgraph(memgraph_args_with_init_file_and_init_data_file)
+    execute_with_user(["MATCH (n) RETURN n"])
+    execute_without_user(["MATCH (n) RETURN n"], True, "Handshake with the server failed!", True)
+    execute_flag_check(flag_check_binary, ["MATCH (n) RETURN n"], 2, "user", "user")
+    print("\033[1;36m~~ Ended env variable check test ~~\033[0m")
+    cleanup()
```
jbajic, line 122: Separate test cases

src/memgraph.cpp
```diff
@@ -472,6 +478,22 @@ struct SessionData {
...
               "Set to the regular expression that each user or role name must fulfill.");
 
+void InitFromCypherlFile(memgraph::query::InterpreterContext &ctx, std::string cypherl_file_path) {
+  memgraph::query::Interpreter interpreter(&ctx);
+  std::ifstream file(cypherl_file_path);
+  if (file.is_open()) {
+    std::string line;
+    while (std::getline(file, line)) {
+      if (!line.empty()) {
+        auto results = interpreter.Prepare(line, {}, {});
```
jbajic, line 488: Does this work with auth stuff, since you are not passing the username in the query?

tests/integration/flag_check/tester.cpp
```diff
@@ -0,0 +1,116 @@
...
+            FLAGS_failure_message, e.what());
+      }
+      return 0;
+    } else {
+      LOG_FATAL(
+          "The query shoudn't have failed but it failed with an "
+          "error message '{}'",
+          e.what());
+    }
+  }
```
kostasrim, line 75:
```suggestion
  } catch (const utils::BasicException &e) {
    if (FLAGS_should_fail) {
      if (!FLAGS_failure_message.empty() && e.what() != FLAGS_failure_message) {
        LOG_FATAL(
            "The query should have failed with an error message of '{}'' but "
            "instead it failed with '{}'",
            FLAGS_failure_message, e.what());
      }
      return 0;
    } else {
      LOG_FATAL(
          "The query shoudn't have failed but it failed with an "
          "error message '{}'",
          e.what());
    }
```

tests/integration/flag_check/tester.cpp
```diff
@@ -0,0 +1,116 @@
...
+            e.what());
+      }
+    }
+    if (!FLAGS_check_failure) continue;
+    if (FLAGS_should_fail) {
+      LOG_FATAL(
+          "The query should have failed but instead it executed "
+          "successfully!");
+    }
+  }
```
kostasrim, line 113: second time I see this, extract the common logic into a function and call it here and above.

src/memgraph.cpp
niko4299: While you don't log in you can do whatever you want, so with this, every user created will have all privileges you then manage by either denying some or revoking all and then granting some.

jbajic: But what about your second PR, if you create a user first, how can you load all the data inside without authenticating, or will the load always happen before user creation?

niko4299: You can do whatever you want since I'm already in the database and  I'm not connecting I don't need to authenticate. If you set init_file it will delete all other users and then start filling database with users from file.

jbajic: I don't think that init_file should delete any users at all, I think it makes sense to run init_file first and then create a user from env, variables or file if one is specified, otherwise, you cannot pass the username to the Interpreter.
But besides that, there is another issue Imagine the following use case.
Memgraph was started with `init_file` and it was populated in the start and was running for a while, and then it crashed. When you start it a second time would you expect that the `init_file` removes all previous data and auth settings, or that we can detect that initialization already happened and we don't want another import? I think the `init_file` is meant to be run only once, even if the flag persists, we should not run the init file twice. What do you think?

niko4299: I discussed with team and I agree with you.

gitbuda (approved):
This looks good to me!

tests/integration/flag_check/runner.py
```diff
@@ -0,0 +1,170 @@
...
+    subprocess.run(args).check_returncode()
+
+
+def execute_flag_check(binary: str, queries: List[str], expected: int, username: str = "", password: str = "") -> None:
+    args = [binary, "--username", username, "--password", password]
+
+    args.extend(queries)
+    args.append(str(expected))
+
+    subprocess.run(args).check_returncode()
```
antaljanosbenjamin, line 60: Why not implement the same thing in python in this file? Then the whole test could be self contained in this single python file.

tests/integration/flag_check/runner.py
```diff
@@ -0,0 +1,170 @@
...
+
+if __name__ == "__main__":
+    memgraph_binary = os.path.join(PROJECT_DIR, "build", "memgraph")
+    tester_binary = os.path.join(PROJECT_DIR, "build", "tests", "integration", "flag_check", "tester")
+    flag_checker_binary = os.path.join(PROJECT_DIR, "build", "tests", "integration", "flag_check", "flag_check")
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--memgraph", default=memgraph_binary)
+    parser.add_argument("--tester", default=tester_binary)
+    parser.add_argument("--flag_checker", default=flag_checker_binary)
```
antaljanosbenjamin, line 165: Are these arguments passed to the tests? If not, then please either remove them or pass them to the tests.

antaljanosbenjamin:
One more thing just popped into my mind: audit log! Before this change all queries run were kind of logged in audit log (except stream transformation queries), so I think it makes sense to include these queries also in my opinion.

tests/integration/flag_check/runner.py
niko4299: I followed the structure of other tests.

jbajic (approved):

antaljanosbenjamin (approved):

PR #699, opened 2022-12-07
[:pineapple: < T1191-MG] Implement CreateExpand with multiframe

kostasrim:
[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR


tests/unit/query_v2_create_expand_multiframe.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+  const auto &src = symbol_table.CreateSymbol("n", true);
+  node.symbol = symbol_table.CreateSymbol("u", true);
+
+  auto create_expand = query::v2::plan::CreateExpand(node, edge, nullptr, src, true);
+  auto cursor = create_expand.MakeCursor(utils::NewDeleteResource());
+
+  EXPECT_CALL(router, CreateExpand(testing::_))
+      .Times(1)
+      .WillOnce(::testing::Return(std::vector<msgs::CreateExpandResponse>{}));
```
antaljanosbenjamin, line 83: I think there is a way to avoid `testing::_`, please check big project how do they use these mocks.

In general I think this is okay.

tests/unit/query_v2_create_expand_multiframe.cpp
```diff
@@ -0,0 +1,88 @@
...
+  query::v2::plan::NodeCreationInfo node;
+  query::v2::plan::EdgeCreationInfo edge;
+  edge.edge_type = msgs::EdgeTypeId::FromUint(1);
+  edge.direction = EdgeAtom::Direction::IN;
+  auto id_alloc = IdAllocator(0, 100);
+
+  const auto &src = symbol_table.CreateSymbol("n", true);
+  node.symbol = symbol_table.CreateSymbol("u", true);
+
+  auto create_expand = query::v2::plan::CreateExpand(node, edge, nullptr, src, true);
```
antaljanosbenjamin, line 78: Don't you need a mocked operator?

42jeremy (approved):

src/query/v2/plan/operator.cpp
```diff
@@ -2450,6 +2466,64 @@ class DistributedCreateExpandCursor : public Cursor {
...
+      TypedValue &v1_value = frame[self_.input_symbol_];
+      const auto &v1 = v1_value.ValueVertex();
+      const auto &v2 = OtherVertex(frame);
+
+      // Set src and dest vertices
+      // TODO(jbajic) Currently we are only handling scenario where vertices
+      // are matched
+      const auto set_vertex = [](const auto &vertex, auto &vertex_id) {
+        vertex_id.first = vertex.PrimaryLabel();
+        vertex_id.second = vertex.GetVertex().id.second;
```
antaljanosbenjamin, line 2502: `GetVertex()` creates a copy, that I don't like. As I haven't noticed that before (if I was a reviewer, sorry for it), but it might cause issues, please update that function to return a `const &`.

But even then, I think we should use `vertex_id = vertex.Id()` which totally eliminates the need for this lambda.

tests/unit/mock_helpers.hpp
```diff
@@ -0,0 +1,82 @@
...
+  context.symbol_table = symbol_table;
+  context.evaluation_context.properties = NamesToProperties(storage.properties_, router);
+  context.evaluation_context.labels = NamesToLabels(storage.labels_, router);
+  context.edge_ids_alloc = id_alloc;
+  context.request_router = router;
+  return context;
+}
+
+inline MockedLogicalOperator &BaseToMock(plan::LogicalOperator *op) {
+  return *static_cast<MockedLogicalOperator *>(op);
```
antaljanosbenjamin, line 77: Please use either `utils::DownCast` or `dynamic_cast`. Probaby `dynamic_cast<MockedLogicalOperator&>(*op) is the best as it will throw in case of invalid cast.

tests/unit/mock_helpers.hpp
```diff
@@ -0,0 +1,82 @@
...
+}
+
+inline MockedCursor &BaseToMock(plan::Cursor *cursor) { return *static_cast<MockedCursor *>(cursor); }
```
antaljanosbenjamin, line 80: Same as above.

tests/unit/query_v2_create_expand_multiframe.cpp
```diff
@@ -0,0 +1,79 @@
...
+  EXPECT_CALL(BaseToMock(once_op.get()), MakeCursor(_)).Times(1).WillOnce(Return(std::move(once_cur)));
+
+  auto create_expand = plan::CreateExpand(node, edge, once_op, src, true);
+  auto cursor = create_expand.MakeCursor(utils::NewDeleteResource());
+
+  MockedRequestRouter router;
+  EXPECT_CALL(router, CreateExpand(_)).Times(1).WillOnce(Return(std::vector<msgs::CreateExpandResponse>{}));
+  auto context = MakeContext(ast, symbol_table, &router, &id_alloc);
+  auto multi_frame = CreateMultiFrame(context.symbol_table.max_position(), src, node.symbol, &router);
+  cursor->PullMultiple(multi_frame, context);
```
antaljanosbenjamin, line 76: What are the expectations about the `MultiFrame` after the pull? Check that every value is untouched and the edge is placed on the frame.

antaljanosbenjamin (changes requested):

tests/unit/mock_helpers.hpp
kostasrim: I did consider `DownCast` but opted in for `static_cast` because I assumed that the name `BaseToMock` is enough to avoid accidental misuse (and UB). As there is no benefit of having the static_cast here other than a potential misuse I replaced it with dynamic_cast (to avoid the kType).

src/query/v2/plan/operator.cpp
```diff
@@ -2450,6 +2466,64 @@ class DistributedCreateExpandCursor : public Cursor {
...
+          }
+          case EdgeAtom::Direction::OUT: {
+            set_vertex(v1, request.src_vertex);
+            set_vertex(v2, request.dest_vertex);
+            break;
+          }
+          case EdgeAtom::Direction::BOTH:
+            LOG_FATAL("Must indicate exact expansion direction here");
+        }
+      });
```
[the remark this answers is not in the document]
kostasrim: That's a good question. After looking at it appears that it was assumed that there is an ordering of the `from/to` vertices of the request. However, that's not true, because the `shard_rsm` handler does not care if for two vertices v1, v2 (from = v1 or from = v2 are both valid). I removed this, good catch, as I was mostly focused on moving staff around I assumed that logic was still valid (I did not implement this cursor and my assumptions felt short).

tests/unit/query_v2_create_expand_multiframe.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+  auto create_expand = plan::CreateExpand(node, edge, once_op, src, true);
+  auto cursor = create_expand.MakeCursor(utils::NewDeleteResource());
+
+  MockedRequestRouter router;
+  EXPECT_CALL(router, CreateExpand(_))
+      .Times(1)
+      .WillOnce(Return(std::vector<msgs::CreateExpandResponse>{msgs::CreateExpandResponse{}}));
+  auto context = MakeContext(ast, symbol_table, &router, &id_alloc);
+  auto multi_frame = CreateMultiFrame(context.symbol_table.max_position(), src, node.symbol, &router);
```
antaljanosbenjamin, line 78: I don't like that the `MultiFrame` is not populated in the `MockedCursor`. This way the cursor might be completely wrong and still managing to satisfy this test.

tests/unit/query_v2_create_expand_multiframe.cpp
```diff
@@ -0,0 +1,89 @@
...
+  cursor->PullMultiple(multi_frame, context);
+
+  auto frames = multi_frame.GetValidFramesReader();
+  for (auto &frame : frames) {
+    EXPECT_EQ(frame[edge.symbol].IsEdge(), true);
+    const auto &e = frame[edge.symbol].ValueEdge();
+    EXPECT_EQ(e.EdgeType(), edge.edge_type);
+  }
```
antaljanosbenjamin, line 86: Two things here:
1. The number of valid frames are not checked.
2. maybe it worth to populate some part of the multiframe, not the whole to make sure the cursor works well in that case also. Of course, the invalid frames than also have to be verified.

antaljanosbenjamin (changes requested):

tests/unit/query_v2_create_expand_multiframe.cpp
42jeremy: Now we are using the real Once

tests/unit/query_v2_create_expand_multiframe.cpp
42jeremy: This should be fixed now

src/query/v2/plan/operator.cpp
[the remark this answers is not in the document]
antaljanosbenjamin: There is an ordering between from/to, it is purposefully called source and destination vertex: even though both are valid, but the direction of the edge is `src -> dst`, so ordering them is important.

tests/unit/query_v2_create_expand_multiframe.cpp
antaljanosbenjamin: I will leave this as is for now. It is enough to check specific cases in case of somebody needs to debug something, it checks some very basic things and the more complex tests will be done on higher level.

src/query/v2/plan/operator.cpp
```diff
@@ -2376,6 +2376,22 @@ class DistributedCreateExpandCursor : public Cursor {
     return true;
   }
 
+  void PullMultiple(MultiFrame &multi_frame, ExecutionContext &context) override {
```
jbajic, line 2379: How does this work without sending bool?

antaljanosbenjamin: The bool that is returned previously can be "got" from the multi_frame itself for the single cases: if it has any valid frames, then it the bool is true, otherwise it is false.

The not so okay case if we push down a multi-frame that already has data (e.g.: `OPTIONAL MATCH` will get a multi-frame from its left side branch, then push down the same multiframe to the right side and we expect it to conserve the information. However the solution heavily depends on how we implement the optional operator, thus I wouldn't like to focus on how this logic should work. Our primary goal now is to make vertex/edge creation and retrieval work and see how does it perform. After that comes the aggregation and optional match probably.

libs/setup.sh
```diff
@@ -171,7 +171,7 @@ benchmark_tag="v1.6.0"
 repo_clone_try_double "${primary_urls[gbenchmark]}" "${secondary_urls[gbenchmark]}" "benchmark" "$benchmark_tag" true
 
 # google test
-googletest_tag="release-1.8.0"
+googletest_tag="release-1.12.1"
```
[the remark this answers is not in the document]
antaljanosbenjamin: I will update it soon!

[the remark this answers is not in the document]
antaljanosbenjamin: https://github.com/memgraph/memgraph/pull/733

gvolfing (approved):

tests/unit/mock_helpers.hpp
```diff
@@ -0,0 +1,82 @@
...
+#pragma once
+
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+#include "query/v2/common.hpp"
+#include "query/v2/context.hpp"
+#include "query/v2/plan/operator.hpp"
+#include "query/v2/request_router.hpp"
+
+namespace memgraph::query::v2 {
```
jbajic, line 21:
```suggestion
namespace memgraph::query::v2::tests {
```

tests/unit/query_v2_create_expand_multiframe.cpp
```diff
@@ -0,0 +1,94 @@
...
+#include "query/v2/common.hpp"
+#include "query/v2/context.hpp"
+#include "query/v2/plan/operator.hpp"
+#include "query/v2/requests.hpp"
+#include "storage/v3/property_value.hpp"
+#include "storage/v3/shard.hpp"
+#include "utils/logging.hpp"
+#include "utils/memory.hpp"
+
+namespace memgraph::query::v2 {
```
jbajic, line 26: namespace memgraph::query::v2::tests {

tests/unit/mock_helpers.hpp
```diff
@@ -0,0 +1,82 @@
...
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#pragma once
+
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+#include "query/v2/common.hpp"
+#include "query/v2/context.hpp"
```
jbajic, line 17:
```suggestion
#include <gmock/gmock.h>
#include <gtest/gtest.h>

#include "query/v2/common.hpp"
#include "query/v2/context.hpp"
```

jbajic (approved):
Small esthetic comments otherwise looks fine

antaljanosbenjamin (approved):

PR #706, opened 2022-12-12
[master < T499-FL] python-query-modules-reloading

as51340:
[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR


src/query/procedure/module.cpp
```diff
@@ -1027,37 +1022,33 @@ bool PythonModule::Close() {
...
-      if (dir_entry_stem.find(stem) != std::string_view::npos &&
-          (submodules.empty() || std::string_view(submodules.stem().c_str()).length() > dir_entry_stem.length())) {
-        submodules = dir_entry.path();
-      }
+  for (auto const &dir_entry : std::filesystem::recursive_directory_iterator(file_path_.parent_path())) {
+    std::string dir_entry_stem = dir_entry.path().stem().string();
+    if (dir_entry.is_regular_file() || dir_entry_stem == "__pycache__") continue;
+    if (dir_entry_stem.find(stem) != std::string_view::npos &&
+        (submodules.empty() || submodules.stem().string().size() > dir_entry_stem.length())) {
+      submodules = dir_entry.path();
```
antaljanosbenjamin: I think in case of `test_module.py` and `test_module_utils/test_module_internal` directories. This logic will choose `test_module_internal` as `submodules`.

I think a correct solution would be recursively iterating over the directory entries, and once we find a directory that contains `stem`, then
1. [disable the recursion](https://en.cppreference.com/w/cpp/filesystem/recursive_directory_iterator/disable_recursion_pending) on the iterator
2. Purge all `__pycache__` folders in this folder
3. Move on to the next one

Of course by introducing the recursion here we march into directories that are totally unnecessarily, but this shouldn't be called on the hot-path. Furthermore, it has the advantage that it works with mage without specifying any runtime flag.

antaljanosbenjamin (commented):
One comment only. Maybe the tests can be modified to cover the edge case `procedures/mage/test_module/test_module_internal` case also.

antaljanosbenjamin (approved):

Issue #684, opened 2022-11-30
Import big CSV files give out of memory error


qbadx:
**Memgraph version**
 2.4.2

**Environment**
docker 20.10.12
memgraph official docker image
Ubuntu 22.04.1 LTS
64 GB RAM

**Describe the bug**

Following the procedure to migrate data from neo4j to memgraph (https://memgraph.com/blog/how-to-migrate-from-neo4j-to-memgraph)

I have a 4.2 GB csv file with 36494229 lines and  56 columns out of neo4j but when I try to import it using the LOAD command as specified in the procedure, the used RAM sky rockets and exceed the 64 GB of available RAM.

I tried to split the file in smaller files and somehow managed to make it work with files < 500 MB but it is not scalable

**To Reproduce**

Load a big csv file with the LOAD command

LOAD CSV FROM "/tmp/test.csv" WITH HEADER as row WITH row WHERE ....

**Expected behavior**

Query failed: Memory limit exceeded! Atempting to allocate a chunk of 128B which would put the current use to 65.47GiB, while the maximum allowed size for allocation is set to 62.42GiB.

**Logs**
In memgraph UI
Query failed: Memory limit exceeded! Atempting to allocate a chunk of 128B which would put the current use to 65.47GiB, while the maximum allowed size for allocation is set to 62.42GiB.

In memgraph CLI
mg_raw_transport_recv: connection closed by server
Client received exception: failed to receive chunk size
Trying to reconnect
Connection failure: couldn't connect to host: Connection refused
Connection failure: couldn't connect to host: Connection refused
Connection failure: couldn't connect to host: Connection refused
Couldn't connect to: 'memgraph://127.0.0.1:7687'

**Additional context**

qbadx:
Hello @gitbuda and thx for you reply, 

I can't share the data unfortunately :/

Here are the stats : 

13 node labels and 3.367.565 node IDs
56 property keys and 39.412.325 property IDs
18 relationship types and 33.148.468 relationship IDs

The csv has  3.494.229 lines and 56 columns

Do you need more information ?

qbadx:
update:

I split the big CSV file into several smaller ones based on the label of the nodes and I end up with the biggest file being 500 MB.

This file in particular makes the memory goes very very high (almost 50/64 GB!)

As I said I can't share the data but it contains 2560553 lines with 3 fields : 

- dates YYYY-MM-DD
[...]

PR #712, opened 2022-12-13
[master<T1215-MG] Fix LOAD CSV large memory usage

antoniofilipovic:
[master < Task] PR
- [x] Update [changelog](https://docs.memgraph.com/memgraph/changelog) -> here [link](https://github.com/memgraph/docs/pull/647)
- [x] Provide the full content or a guide for the final git message:
Fix LOAD CSV bug for large memory usage

https://github.com/memgraph/memgraph/issues/684

### DIFF
--- a/src/utils/memory.hpp
+++ b/src/utils/memory.hpp
@@ -469,7 +469,8 @@ class PoolResource final : public MemoryResource {
   /// impl::Pool::MaxBlocksInChunk()) as the real maximum number of blocks per
   /// chunk. Allocation requests exceeding max_block_size are simply forwarded
   /// to upstream memory.
-  PoolResource(size_t max_blocks_per_chunk, size_t max_block_size, MemoryResource *memory = NewDeleteResource());
+  PoolResource(size_t max_blocks_per_chunk, size_t max_block_size, MemoryResource *memory_pools = NewDeleteResource(),
+               MemoryResource *memory_unpooled = NewDeleteResource());
 
   PoolResource(const PoolResource &) = delete;
   PoolResource &operator=(const PoolResource &) = delete;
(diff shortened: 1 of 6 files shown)


tests/unit/utils_memory.cpp
```diff
@@ -254,7 +254,7 @@ TEST(PoolResource, BigBlockAllocations) {
   TestMemory test_mem;
   const size_t max_blocks_per_chunk = 3U;
   const size_t max_block_size = 64U;
-  memgraph::utils::PoolResource mem(max_blocks_per_chunk, max_block_size, &test_mem);
+  memgraph::utils::PoolResource mem(max_blocks_per_chunk, max_block_size, &test_mem, &test_mem);
```
antaljanosbenjamin, line 257: I would use a different test memory resource to verify that big block are allocated with  the right resource.

src/utils/memory.hpp
```diff
@@ -469,7 +469,8 @@ class PoolResource final : public MemoryResource {
   /// impl::Pool::MaxBlocksInChunk()) as the real maximum number of blocks per
   /// chunk. Allocation requests exceeding max_block_size are simply forwarded
   /// to upstream memory.
-  PoolResource(size_t max_blocks_per_chunk, size_t max_block_size, MemoryResource *memory = NewDeleteResource());
+  PoolResource(size_t max_blocks_per_chunk, size_t max_block_size, MemoryResource *memory_pools = NewDeleteResource(),
+               MemoryResource *memory_unpooled = NewDeleteResource());
```
antaljanosbenjamin, line 473: Having a default value is okay, however in the `PullPlan::Pull` I would explicitly pass in a `NewDeleteResource` to express the intent: we want all of the big blocks are allocated with `NewDeleteResource`.

antaljanosbenjamin (changes requested):

src/utils/memory.hpp
antoniofilipovic: Makes sense! Added that part.

antaljanosbenjamin (approved):

antoniofilipovic:
@antaljanosbenjamin :ship: :it: when you can :)

PR #717, opened 2022-12-15
[master < T517] On delete triggers invalid edge reference

as51340:
[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR

### DIFF
--- a/src/query/trigger_context.hpp
+++ b/src/query/trigger_context.hpp
@@ -65,11 +65,6 @@ struct DeletedObject {
   TAccessor object;
 };
 
-template <>
-inline bool DeletedObject<EdgeAccessor>::IsValid() const {
-  return object.StorePropertiesOnEdges() && object.IsVisible(storage::View::OLD);
-}
-
 template <ObjectAccessor TAccessor>
 struct SetObjectProperty {
   explicit SetObjectProperty(const TAccessor &object, storage::PropertyId key, TypedValue old_value,
--- a/src/query/trigger_context.hpp
+++ b/src/query/trigger_context.hpp
@@ -65,6 +65,11 @@ struct DeletedObject {
   TAccessor object;
 };
 
+template <>
+inline bool DeletedObject<EdgeAccessor>::IsValid() const {
+  return object.StorePropertiesOnEdges() && object.IsVisible(storage::View::OLD);
+}
+
 template <ObjectAccessor TAccessor>
 struct SetObjectProperty {
   explicit SetObjectProperty(const TAccessor &object, storage::PropertyId key, TypedValue old_value,
--- a/src/storage/v2/edge_accessor.cpp
+++ b/src/storage/v2/edge_accessor.cpp
@@ -21,6 +21,33 @@
 namespace memgraph::storage {
 
 bool EdgeAccessor::IsVisible(const View view) const {
+  bool visible = true;
+  // When edges don't have properties, their isolation level is still dictated by MVCC ->
+  // iterate over the deltas of the from_vertex_ and see which deltas can be applied on edges.
+  if (!config_.properties_on_edges) {
+    Delta *delta = nullptr;
+    {
+      std::lock_guard<utils::SpinLock> guard(from_vertex_->lock);
+      visible = !from_vertex_->deleted;
+      delta = from_vertex_->delta;
+    }
+    ApplyDeltasForRead(transaction_, delta, view, [&](const Delta &delta) {
+      switch (delta.action) {
+        case Delta::Action::ADD_LABEL:
+        case Delta::Action::REMOVE_LABEL:
+        case Delta::Action::SET_PROPERTY:
+        case Delta::Action::REMOVE_IN_EDGE:
+        case Delta::Action::ADD_IN_EDGE:
+        case Delta::Action::REMOVE_OUT_EDGE:  // relevant for the from_vertex_
+        case Delta::Action::RECREATE_OBJECT:
+        case Delta::Action::DELETE_OBJECT:
+          break;
+        case Delta::Action::ADD_OUT_EDGE:  // also relevant for the from_vertex_
+          visible = false;
+      }
+    });
+    return visible;
+  }
   bool deleted = true;
   bool exists = true;
   Delta *delta = nullptr;
@@ -49,7 +76,6 @@ bool EdgeAccessor::IsVisible(const View view) const {
       }
     }
   });
-
   return exists && (for_deleted_ || !deleted);
 }
 
--- a/src/storage/v2/edge_accessor.cpp
+++ b/src/storage/v2/edge_accessor.cpp
@@ -21,16 +21,19 @@
 namespace memgraph::storage {
 
 bool EdgeAccessor::IsVisible(const View view) const {
-  bool visible = true;
+  bool exists = true;
+  bool deleted = true;
   // When edges don't have properties, their isolation level is still dictated by MVCC ->
   // iterate over the deltas of the from_vertex_ and see which deltas can be applied on edges.
   if (!config_.properties_on_edges) {
     Delta *delta = nullptr;
     {
       std::lock_guard<utils::SpinLock> guard(from_vertex_->lock);
-      visible = !from_vertex_->deleted;
+      deleted = from_vertex_->deleted;
       delta = from_vertex_->delta;
     }
+    // don't apply delta to some other edge
+    if (delta->vertex_edge.edge != edge_) return !deleted;
     ApplyDeltasForRead(transaction_, delta, view, [&](const Delta &delta) {
       switch (delta.action) {
         case Delta::Action::ADD_LABEL:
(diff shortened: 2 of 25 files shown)


src/query/trigger_context.hpp
```diff
@@ -65,6 +65,11 @@ struct DeletedObject {
   TAccessor object;
 };
 
+template <>
+inline bool DeletedObject<EdgeAccessor>::IsValid() const {
+  return object.StorePropertiesOnEdges() && object.IsVisible(storage::View::OLD);
+}
+
```
jbajic, line 72: This solves the crash but introduces a more subtle bug with isolation levels. The issue that I think it introduces is with visibility of objects in triggers. As you can see in `IsVisible` we are iterating through deltas when edges have properties since they have deltas as well then. However, even when edges don't have properties they still have deltas, their isolation is still dictated by mvcc. The issue is that their deltas are within start and end nodes, basically, you would have to iterate through start or end nodes (It does not matter which one, since deltas are stored in both) and look for the following deltas:
- `ADD_IN_EDGE`
- `ADD_OUT_EDGE`
- `REMOVE_IN_EDGE`
- `REMOVE_OUT_EDGE`
which can be applied on edges

src/storage/v2/edge_accessor.cpp
```diff
@@ -21,6 +21,33 @@
...
+  // When edges don't have properties, their isolation level is still dictated by MVCC ->
+  // iterate over the deltas of the from_vertex_ and see which deltas can be applied on edges.
+  if (!config_.properties_on_edges) {
+    Delta *delta = nullptr;
+    {
+      std::lock_guard<utils::SpinLock> guard(from_vertex_->lock);
+      visible = !from_vertex_->deleted;
+      delta = from_vertex_->delta;
+    }
+    ApplyDeltasForRead(transaction_, delta, view, [&](const Delta &delta) {
```
antaljanosbenjamin, line 34: I like this approach, but we should extend this.
1. We shouldn't just use a single bool variable, because there is a difference between whether the object doesn't exists for the transaction because it is not yet created or it is already deleted
2. The gid of the edge is important, otherwise we might consider deltas that are corresponding to other edges. Check how `delta.vertex_edge.edge` is used.

tests/e2e/triggers/workloads.yaml
```diff
@@ -2,10 +2,17 @@ bolt_port: &bolt_port "7687"
...
-      args: ["--bolt-port", *bolt_port, "--log-level=TRACE"]
+      args: ["--bolt-port", *bolt_port, "--log-level=TRACE", "--storage-properties-on-edges=True"]
       log_file: "triggers-e2e.log"
       setup_queries: []
       validation_queries: []
+storage_properties_edges_false: &storage_properties_edges_false
+  cluster:
+    main:
+      args: ["--bolt-port", "7687", "--log-level=TRACE", "--also-log-to-stderr", "--storage-properties-on-edges=False"]
+      log_file: "py-query-modules-reloading-e2e.log"
```
antaljanosbenjamin, line 13: Wrong log file name.

tests/e2e/triggers/workloads.yaml
```diff
@@ -2,10 +2,17 @@ bolt_port: &bolt_port "7687"
...
     main:
-      args: ["--bolt-port", *bolt_port, "--log-level=TRACE"]
+      args: ["--bolt-port", *bolt_port, "--log-level=TRACE", "--storage-properties-on-edges=True"]
       log_file: "triggers-e2e.log"
       setup_queries: []
       validation_queries: []
+storage_properties_edges_false: &storage_properties_edges_false
+  cluster:
+    main:
+      args: ["--bolt-port", "7687", "--log-level=TRACE", "--also-log-to-stderr", "--storage-properties-on-edges=False"]
```
antaljanosbenjamin, line 12: Please add a comment to explain why the bolt port is hard-coded (I guess because `pytest_runner.sh` cannot handle different bolt ports).

tests/e2e/triggers/workloads.yaml
```diff
@@ -18,14 +25,17 @@ workloads:
...
-  - name: "Triggers privilege check"
-    binary: "tests/e2e/triggers/memgraph__e2e__triggers__privileges"
-    args: ["--bolt-port", *bolt_port]
-    <<: *template_cluster
-
-
+  # - name: "Triggers privilege check"
+  #   binary: "tests/e2e/triggers/memgraph__e2e__triggers__privileges"
+  #   args: ["--bolt-port", *bolt_port]
+  #   <<: *template_cluster
```
antaljanosbenjamin, line 36: Please uncomment or remove.

tests/e2e/triggers/triggers_properties_false.py
```diff
@@ -0,0 +1,98 @@
...
+
+@pytest.mark.parametrize("ba_commit", ["BEFORE COMMIT", "AFTER COMMIT"])
+def test_create_on_create(ba_commit):
+    """
+    Args:
+        ba_commit (str): BEFORE OR AFTER commit
+    """
+    cursor = connect().cursor()
+    QUERY_TRIGGER_CREATE = f"""
+        CREATE TRIGGER DeleteTriggerEdgesCount
```
antaljanosbenjamin, line 26:
```suggestion
        CREATE TRIGGER CreateTriggerEdgesCount
```

tests/e2e/triggers/triggers_properties_false.py
```diff
@@ -0,0 +1,98 @@
...
+        CREATE (n)-[r:TYPE]->(m);
+    """
+    execute_and_fetch_all(cursor, QUERY_CREATE_EDGE)
+    # See if trigger was triggered
+    nodes = execute_and_fetch_all(cursor, "MATCH (n:Node) RETURN n")
+    assert len(nodes) == 2
+    created_edges = execute_and_fetch_all(cursor, "MATCH (n:CreatedEdge) RETURN n")
+    assert len(created_edges) == 1
+    execute_and_fetch_all(cursor, "MATCH (n) DETACH DELETE n;")
+    execute_and_fetch_all(cursor, "DROP TRIGGER DeleteTriggerEdgesCount")
```
antaljanosbenjamin, line 51:
```suggestion
    execute_and_fetch_all(cursor, "DROP TRIGGER CreateTriggerEdgesCount")
```

tests/e2e/triggers/triggers_properties_false.py
```diff
@@ -0,0 +1,98 @@
...
+    execute_and_fetch_all(cursor, "CREATE (n:Node {id: 2})")
+    execute_and_fetch_all(cursor, "CREATE (n:Node {id: 3})")
+    execute_and_fetch_all(cursor, "CREATE (n:Node {id: 4})")
+    res = execute_and_fetch_all(cursor, "MATCH (n:Node) RETURN n")
+    assert len(res) == 4
+    res2 = execute_and_fetch_all(cursor, "MATCH (n:DeletedEdge) RETURN n")
+    assert len(res2) == 0
+    QUERY_CREATE_EDGE = """
+        MATCH (n:Node {id: 1}), (m:Node {id: 2})
+        CREATE (n)-[r:TYPE]->(m);
```
antaljanosbenjamin, line 80: To trigger the bug that I mentioned please consider extending this test:
1. have one edge that is not deleted
2. have one edge that is deleted
3. have one edge that is created and deleted in the same transaction

antaljanosbenjamin (changes requested):
I haven't checked the tests, because

as51340:
Two more questions from my side: 
1. Is this the correct logic `if (delta->vertex_edge.edge != edge_) return !deleted;` 
2. Is `test_create_on_delete_explicit_transaction` valid for all three isolation levels? Workflow:
- create trigger
- create and delete edge in the explicit transaction

tests/e2e/triggers/workloads.yaml
as51340: Will do on the last commit because I cannot run this test locally.

src/storage/v2/edge_accessor.cpp
```diff
@@ -21,8 +21,40 @@
...
+      switch (delta.action) {
+        case Delta::Action::ADD_LABEL:
+        case Delta::Action::REMOVE_LABEL:
+        case Delta::Action::SET_PROPERTY:
+        case Delta::Action::REMOVE_IN_EDGE:
+        case Delta::Action::ADD_IN_EDGE:
+        case Delta::Action::ADD_OUT_EDGE:  // relevant for the from_vertex_
+          break;
+        case Delta::Action::RECREATE_OBJECT:
+        case Delta::Action::REMOVE_OUT_EDGE:  // also relevant for the from_vertex_
```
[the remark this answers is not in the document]
as51340: Added exists false so now it makes more sense.

tests/e2e/triggers/common.py
```diff
@@ -0,0 +1,25 @@
...
+
+def execute_and_fetch_all(cursor: mgclient.Cursor, query: str, params: dict = {}) -> typing.List[tuple]:
+    cursor.execute(query, params)
+    return cursor.fetchall()
+
+
+def connect(**kwargs) -> mgclient.Connection:
+    connection = mgclient.connect(host="localhost", port=7687, **kwargs)
+    connection.autocommit = True
+    return connection
```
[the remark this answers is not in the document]
as51340: Why would connection be a generator or is there some other reason why yield would be used?

antaljanosbenjamin (approved):
To answer your questions:
> Is this the correct logic if (delta->vertex_edge.edge != edge_) return !deleted;

No, but as I see you already fixed it. Accessing the fields of a delta without knowing what type of delta is that is bad (in most of the cases you can end up with undefined behavior).

> Is test_create_on_delete_explicit_transaction valid for all three isolation levels?

I would say it should be okay. The isolation levels are important when you work with multiple transaction simultaneously. In the test you don't have multiple transactions, so I would say the isolation levels should have no affect on the test. However there is always a possibility that I missed something or just plain wrong, so if you noticed something strange, feel free to bring it up, it might lead to interesting conversation on top of helping you.

tests/e2e/triggers/common.py
```diff
@@ -0,0 +1,25 @@
...
+
+def execute_and_fetch_all(cursor: mgclient.Cursor, query: str, params: dict = {}) -> typing.List[tuple]:
+    cursor.execute(query, params)
+    return cursor.fetchall()
+
+
+def connect(autocommit=True, **kwargs) -> mgclient.Connection:
+    connection = mgclient.connect(host="localhost", port=7687, **kwargs)
+    connection.autocommit = autocommit
+    yield connection
```
jbajic, line 25: Sorry if I was unclear, but the idea of yield is to continue the fixture clean up, so basically what would you do now, after yield is:
```python
    triggers_list = execute_and_fetch("SHOW TRIGGERS;")
    for trigger in triggers_list:
      execute_and_fetch_all(cursor, f"DROP TRIGGER {trigger[name]}")
    execute_and_fetch_all(cursor, "MATCH (n) DETACH DELETE n")
```
and remove all cleanup logic from test cases.

gitbuda:
There are still some comments left here. Can you resolve them all + uncomment the e2e test config?

jbajic (approved):

PR #719, opened 2022-12-16
[:pineapple:  < MG-T1149] Implement shard split on storage side

jbajic:
[master < Task] PR


src/storage/v3/shard.cpp
```diff
@@ -1048,6 +1099,20 @@ void Shard::StoreMapping(std::unordered_map<uint64_t, std::string> id_to_name) {
...
+  if (vertices_.size() > config_.split.max_shard_vertex_size) {
+    auto mid_elem = vertices_.begin();
+    std::ranges::advance(mid_elem, static_cast<VertexContainer::difference_type>(vertices_.size() / 2));
+    return SplitInfo{shard_version_, mid_elem->first};
+  }
+  return std::nullopt;
+}
+
+SplitData Shard::PerformSplit(const PrimaryKey &split_key) {
+  ++shard_version_;
```
jbajic, line 1112: Should be managed by coordinator

jbajic: Both sides of split should have same version

src/storage/v3/delta.hpp
```diff
@@ -129,6 +130,11 @@ inline bool operator==(const PreviousPtr::Pointer &a, const PreviousPtr::Pointer
 
 inline bool operator!=(const PreviousPtr::Pointer &a, const PreviousPtr::Pointer &b) { return !(a == b); }
 
+inline uint64_t GetNextDeltaUUID() noexcept {
```
antaljanosbenjamin, line 133: Is this really a `UUID`?

src/storage/v3/splitter.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+    const auto *old_vertex_ptr = &*split_key_it;
+    auto next_it = std::next(split_key_it);
+
+    const auto &[splitted_vertex_it, inserted, node] = splitted_data.insert(vertices_.extract(split_key_it->first));
+    MG_ASSERT(inserted, "Failed to extract vertex!");
+
+    // Update indices
+    update_indices(label_index_vertex_entry_map, data.label_indices, old_vertex_ptr, splitted_vertex_it);
+    update_indices(label_property_vertex_entry_map, data.label_property_indices, old_vertex_ptr, splitted_vertex_it);
```
antaljanosbenjamin, line 113: Does the pointer to vertex changes? I think extract + insert shouldn't move the underlying pair in the memory at all. However I might misunderstood why is this necessary.

src/storage/v3/splitter.cpp
```diff
@@ -0,0 +1,240 @@
...
+    // We need all transaction whose deltas need to be resolved for any of the
+    // entities
+    if (collected_transactions_.contains(transaction->commit_info->start_or_commit_timestamp.logical_id)) {
+      transactions.insert({commit_start, start_logical_id_to_transaction_[commit_start]->Clone()});
+    }
+  }
+
+  // It is necessary to clone all the transactions first so we have new addresses
+  // for deltas, before doing alignment of deltas and prev_ptr
+  AdjustClonedTransactions(transactions, cloned_vertices, cloned_edges);
```
antaljanosbenjamin, line 165: I think `Delta::vertex_edge::edge` should be adjusted somewhere because of the copied edges, but I couldn't find where do you modify them. Do I miss something here?

tests/benchmark/storage_v3_split.cpp
```diff
@@ -0,0 +1,171 @@
...
+  std::default_random_engine e1(r());
+  std::uniform_int_distribution<int> uniform_dist(0, state.range(0));
+
+  for (int64_t i{0}; i < state.range(0); ++i) {
+    auto acc = storage->Access(GetNextHlc());
+    MG_ASSERT(acc.CreateVertexAndValidate({secondary_label}, PrimaryKey{PropertyValue(i)},
+                                          {{secondary_property, PropertyValue(i)}})
+                  .HasValue(),
+              "Failed creating with pk {}", i);
+    if (i > 1) {
```
antaljanosbenjamin, line 132: Maybe parametrize the number of created edges also.

tests/unit/storage_v3_shard_split.cpp
```diff
@@ -0,0 +1,362 @@
...
+#include "storage/v3/shard.hpp"
+#include "storage/v3/vertex.hpp"
+#include "storage/v3/vertex_id.hpp"
+
+using testing::Pair;
+using testing::UnorderedElementsAre;
+
+namespace memgraph::storage::v3::tests {
+
+class ShardSplitTest : public testing::Test {
```
antaljanosbenjamin, line 34: Just a few ideas for more tests:
1. Split a storage, then call GC on the original storage, destroy it, then call GC on the splitted storage.
2. Split a storage, then call GC on the second storage, destroy it, then call GC on the original storage.

antaljanosbenjamin (changes requested):

src/storage/v3/delta.hpp
```diff
@@ -27,6 +28,11 @@ struct Edge;
 struct Delta;
 struct CommitInfo;
 
+inline uint64_t GetNextDeltaUUID() noexcept {
+  static uint64_t uuid{0};
+  return ++uuid;
+}
```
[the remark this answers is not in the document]
jbajic: I agree I wanted to make this a shard concern, but it is difficult to propagate the id generation since it would be needed in Vertex and Edge accessors.

src/storage/v3/splitter.cpp
jbajic: You are right the address does not change, I assumed it would. Yes I thought this is needed to adjust the vertex ptr in entries, but it seems redundant, and I will remove it and test

src/storage/v3/delta.hpp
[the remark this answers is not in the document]
antaljanosbenjamin: I think that is fine for now, but tagging @spacejam to verify.

src/storage/v3/delta.hpp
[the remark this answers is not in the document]
antaljanosbenjamin: I saw you made this into a `SpinLock`-ed `Synchronized`. It is a good question whether this or an atomic variable is more suitable here.

It makes testing definitely harder (no way to inject values we want to check), but feel free to keep it as is if you think it is too much hassle.

src/storage/v3/shard.cpp
```diff
@@ -1048,6 +1102,21 @@ void Shard::StoreMapping(std::unordered_map<uint64_t, std::string> id_to_name) {
...
 }
 
+std::optional<SplitInfo> Shard::ShouldSplit() const noexcept {
+  if (vertices_.size() > config_.split.max_shard_vertex_size) {
+    auto mid_elem = vertices_.begin();
+    std::ranges::advance(mid_elem, static_cast<VertexContainer::difference_type>(vertices_.size() / 2));
+    return SplitInfo{mid_elem->first, shard_version_};
+  }
+  return std::nullopt;
+}
```
spacejam, line 1112: in the near future, we will want to only calculate the split point once, because this is run every time `Cron` runs, and if we iterate over half of the entire index every ~200ish milliseconds it will kill the shard before the split is completed, because it will take the overall cluster a few round trips to accomplish that. I wouldn't fix that now but I'll add a TODO code suggestion related to it.

src/storage/v3/shard.cpp
```diff
@@ -1048,6 +1102,21 @@ void Shard::StoreMapping(std::unordered_map<uint64_t, std::string> id_to_name) {
   name_id_mapper_.StoreMapping(std::move(id_to_name));
 }
 
+std::optional<SplitInfo> Shard::ShouldSplit() const noexcept {
+  if (vertices_.size() > config_.split.max_shard_vertex_size) {
+    auto mid_elem = vertices_.begin();
+    std::ranges::advance(mid_elem, static_cast<VertexContainer::difference_type>(vertices_.size() / 2));
```
spacejam, line 1108:
```suggestion
    // TODO(tyler) the first time we calculate the split point, we should store it so that we don't have to
    // iterate over half of the entire index each time Cron is run until the split succeeds.
    std::ranges::advance(mid_elem, static_cast<VertexContainer::difference_type>(vertices_.size() / 2));
```

src/storage/v3/splitter.cpp
jbajic: Ah you are right we still need to set `edge`, in case when we have cloned the edge and have a new address, and then we need to adjust the delta `edge` part of `vertex_edge` to the cloned one edge.

src/storage/v3/shard.hpp
```diff
@@ -368,7 +391,7 @@ class Shard final {
   // Main object storage
   NameIdMapper name_id_mapper_;
   LabelId primary_label_;
-  // The shard's range is [min, max)
+  // The shard's range is [min, max>
```
spacejam, line 394:
```suggestion
  // The shard's range is [min, max)
```

This should remain unchanged, as it is the correct notation for a half-open interval: https://en.wikipedia.org/wiki/Interval_%28mathematics%29#Terminology

spacejam (approved):

src/storage/v3/indices.hpp
```diff
@@ -114,8 +122,35 @@ class LabelIndex {
 
   void Clear() { index_.clear(); }
 
+  std::map<IndexType, IndexContainer> SplitIndexEntries(const PrimaryKey &split_key) {
+    if (index_.empty()) {
+      return {};
+    }
```
gvolfing, line 128: Is this check actually needed?
If `index_` is empty, the loop will not be entered and the empty `cloned_indices` will be returned.

gvolfing (approved):

src/coordinator/hybrid_logical_clock.hpp
```diff
@@ -23,17 +23,17 @@ namespace memgraph::coordinator {
...
-  bool operator==(const Hlc &other) const = default;
-  bool operator<(const Hlc &other) const = default;
-  bool operator==(const uint64_t other) const { return logical_id == other; }
-  bool operator<(const uint64_t other) const { return logical_id < other; }
-  bool operator>=(const uint64_t other) const { return logical_id >= other; }
+  bool operator==(const Hlc &other) const noexcept = default;
+  bool operator<(const Hlc &other) const noexcept = default;
+  bool operator==(const uint64_t other) const noexcept { return logical_id == other; }
+  bool operator<(const uint64_t other) const noexcept { return logical_id < other; }
+  bool operator>=(const uint64_t other) const noexcept { return logical_id >= other; }
```
antaljanosbenjamin, line 36: Don't forget to mark the `auto operator<=>` also noexcept as defaulted operators are calling that function.

src/storage/v3/delta.hpp
```diff
@@ -27,6 +29,11 @@ struct Edge;
 struct Delta;
 struct CommitInfo;
 
+inline uint64_t GetNextDeltaId() noexcept {
```
antaljanosbenjamin, line 32:
```suggestion
inline uint64_t GetNextDeltaId() {
```

`WithLock` is not `noexcept`.

src/storage/v3/indices.hpp
antaljanosbenjamin: The same stands for `LabelPropertyIndex::SplitIndexEntries`

src/storage/v3/splitter.hpp
```diff
@@ -0,0 +1,107 @@
+// Copyright 2023 Memgraph Ltd.
```
antaljanosbenjamin, line 1: Two things popped into my mind: the `Shard` has two members:
1. `deleted_vertices_`
2. `deleted_edges_`

I think they should be taken care of in some way. I see two ways:
1. make a note somewhere about they are not being handled. This is fine as currently we cannot delete edges or vertices.
2. Handle them.

src/storage/v3/shard.hpp
```diff
@@ -396,41 +420,10 @@ class Shard final {
...
-
-  utils::FileRetainer file_retainer_;
-
-  // Global locker that is used for clients file locking
-  utils::FileRetainer::FileLocker global_locker_;
-
   // Holds all of the (in progress, committed and aborted) transactions that are read or write to this shard, but
   // haven't been cleaned up yet
   std::map<uint64_t, std::unique_ptr<Transaction>> start_logical_id_to_transaction_{};
+  Splitter shard_splitter_;
```
antaljanosbenjamin, line 426: Do we need to keep it as a member variable? It's construction seems super cheap (basically just references), so I would only create a splitter when it is necessary. The reason not primarily about memory footprint, but I instantly started to think what this splitter does that requires this to be alive during the whole life of a shard.

src/storage/v3/indices.hpp
```diff
@@ -114,8 +122,35 @@ class LabelIndex {
...
+
+    // Cloned index entries will contain new index entry iterators, but old
+    // vertices address which need to be adjusted after extracting vertices
```
antaljanosbenjamin, line 131: This comment is kind of obsolete: the address of vertices doesn't change after extraction if I am not mistaken.

src/storage/v3/indices.hpp
```diff
@@ -114,8 +122,35 @@ class LabelIndex {
...
+    std::map<IndexType, IndexContainer> cloned_indices;
+    for (auto &[index_type_val, index] : index_) {
+      auto entry_it = index.begin();
+      auto &cloned_indices_container = cloned_indices[index_type_val];
+      while (entry_it != index.end()) {
+        // We need to save the next pointer since the current one will be
```
antaljanosbenjamin, line 137:
```suggestion
        // We need to save the next iterator since the current one will be
```

src/storage/v3/transaction.hpp
```diff
@@ -54,6 +64,54 @@ struct Transaction {
...
+                                     delta.id, command_id);
+          break;
+      }
+    }
+    return copied_deltas;
+  }
+
+  // This does not solve the whole problem of copying deltas
+  std::unique_ptr<Transaction> Clone() const {
+    return std::make_unique<Transaction>(start_timestamp, *commit_info, CopyDeltas(commit_info.get()), command_id,
```
antaljanosbenjamin, line 111: Aren't you using the "old" commit info to construct the new deltas?

antaljanosbenjamin (commented):
I still have to go over the splitter source and tests, but the rest is reviewed from my perspective.

src/storage/v3/splitter.cpp
```diff
@@ -0,0 +1,351 @@
...
+}
+
+void Splitter::AdjustDeltaNext(const Delta &original, Delta &cloned,
+                               std::map<uint64_t, std::unique_ptr<Transaction>> &cloned_transactions) {
+  // Get cloned_delta->next transaction, using delta->next original transaction
+  auto cloned_transaction_it = std::ranges::find_if(cloned_transactions, [&original](const auto &elem) {
+    return elem.second->start_timestamp == original.next->commit_info->start_or_commit_timestamp ||
+           elem.second->commit_info->start_or_commit_timestamp == original.next->commit_info->start_or_commit_timestamp;
+  });
+  // TODO(jbajic) What if next in delta chain does not belong to cloned transaction?
```
antaljanosbenjamin, line 301: the "next pointer" for the delta points to another delta that is older than the current one and belongs to the same object. How is it possible that the next delta does not belong to the cloned transaction, but the current one does?

src/storage/v3/splitter.cpp
```diff
@@ -0,0 +1,351 @@
...
+  // MG_ASSERT(cloned_transaction_it != cloned_transactions.end(), "Cloned transaction not found");
+  if (cloned_transaction_it == cloned_transactions.end()) return;
+  // Find cloned delta in delta list of cloned transaction
+  auto found_cloned_delta_it = std::ranges::find_if(
+      cloned_transaction_it->second->deltas, [&original](const auto &elem) { return elem.id == original.next->id; });
+  MG_ASSERT(found_cloned_delta_it != cloned_transaction_it->second->deltas.end(), "Delta with given uuid must exist!");
+  cloned.next = &*found_cloned_delta_it;
```
antaljanosbenjamin, line 308: Why not to set the `found_cloned_delta_it->prev.Set(&cloned)`?

src/storage/v3/splitter.cpp
```diff
@@ -0,0 +1,351 @@
...
+}
+
+void Splitter::AdjustDeltaPrevPtr(const Delta &original, Delta &cloned,
+                                  std::map<uint64_t, std::unique_ptr<Transaction>> &cloned_transactions,
+                                  VertexContainer & /*cloned_vertices*/, EdgeContainer &cloned_edges) {
+  auto ptr = original.prev.Get();
+  switch (ptr.type) {
+    case PreviousPtr::Type::NULLPTR: {
+      // noop
+      break;
```
antaljanosbenjamin, line 318: Can this happen? That would mean the delta doesn't belong to any object.

src/storage/v3/splitter.cpp
```diff
@@ -0,0 +1,351 @@
...
+      MG_ASSERT(cloned_transaction_it != cloned_transactions.end(), "Cloned transaction not found");
+      // Find cloned delta in delta list of cloned transaction
+      auto found_cloned_delta_it =
+          std::ranges::find_if(cloned_transaction_it->second->deltas,
+                               [delta = ptr.delta](const auto &elem) { return elem.id == delta->id; });
+      MG_ASSERT(found_cloned_delta_it != cloned_transaction_it->second->deltas.end(),
+                "Delta with given id must exist!");
+
+      cloned.prev.Set(&*found_cloned_delta_it);
+      break;
```
antaljanosbenjamin, line 335: I think this could be spared with if you set the prev pointers in `AdjustDeltaNext`. Of course this requires some renaming, but I think we can spare a lot of work this way.

src/storage/v3/shard.hpp
jbajic: That is true, we don't need it there

src/storage/v3/transaction.hpp
jbajic: That is true, will fix it

src/storage/v3/splitter.cpp
```diff
@@ -0,0 +1,351 @@
...
+                [&current_next_delta = *current_next_delta](const auto &delta) { return delta == current_next_delta; });
+
+            current_next_delta = next_delta;
+          }
+        } else {
+          ++cloned_delta_it;
+        }
+        break;
+      }
+      case PreviousPtr::Type::EDGE:
```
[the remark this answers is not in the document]
jbajic: The only way that this can happen is if during collecting of vertices and edges we have collected a transaction that also includes delta chains on vertices and edges that remain on non splitted shard. So yeah you are correct I need to check for that as well.

src/storage/v3/splitter.cpp
```diff
@@ -0,0 +1,351 @@
...
+void Splitter::AdjustEdgeRef(Delta &cloned_delta, EdgeContainer &cloned_edges) const {
+  switch (cloned_delta.action) {
+    case Delta::Action::ADD_IN_EDGE:
+    case Delta::Action::ADD_OUT_EDGE:
+    case Delta::Action::REMOVE_IN_EDGE:
+    case Delta::Action::REMOVE_OUT_EDGE: {
+      // Find edge
+      if (config_.items.properties_on_edges) {
+        // Only case when not finding is when the edge is not on splitted shard
+        // TODO Do this after prune an move condition into assert
```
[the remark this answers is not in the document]
jbajic: The reason for pruning after adjusting is that we cannot prune without having a new delta chain. 
The comment is not related to the pruning or adjusting process but to the:
The reason for not finding an edge here is that when we split we take all edges that are connected to the splitted vertices. But when splitting transactions we take all transactions that have done any changes on splitted vertices or edges, but that transaction might also contain deltas on edges that we did not include in the split, and that's why we can encounter deltas that we do not care about. Does this make sense?

src/storage/v3/splitter.cpp
```diff
@@ -0,0 +1,351 @@
...
+        if (const auto cloned_edge_it =
+                std::ranges::find_if(cloned_edges, [edge_ptr = cloned_delta.vertex_edge.edge.ptr](
```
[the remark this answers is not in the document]
jbajic: You are correct, wow I missed that, thanks

src/storage/v3/splitter.cpp
```diff
@@ -0,0 +1,351 @@
...
+      // noop
+      break;
+    }
+  }
+}
+
+void Splitter::AdjustDeltaNext(const Delta &original, Delta &cloned,
+                               std::map<uint64_t, std::unique_ptr<Transaction>> &cloned_transactions) {
+  // Get cloned_delta->next transaction, using delta->next original transaction
+  auto cloned_transaction_it = std::ranges::find_if(cloned_transactions, [&original](const auto &elem) {
```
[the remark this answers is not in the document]
jbajic: I will separate `start_timestamp` search and commit info search, since `start_timestamp` is key of `cloned_transactions` map

src/storage/v3/splitter.cpp
jbajic: Because we follow a delta chain here and the explanation is similar to the above https://github.com/memgraph/memgraph/pull/719#discussion_r1114477610

src/storage/v3/splitter.cpp
jbajic: It should never happen. I will add assert

src/storage/v3/splitter.cpp
jbajic: That is true

antaljanosbenjamin (approved):

PR #720, opened 2022-12-18
[master < T534-FL-torch-mg-load-fix] torch and igraph removing from sys cache

as51340:
[master < Task] PR

- docs-changelog-message: Disable removing `torch` and `igraph` from `sys.modules` cache because of the issue occuring when reloading.
Connected issue on the PyTorch: https://github.com/pytorch/pytorch/issues/91039

### DIFF
--- a/src/query/procedure/module.cpp
+++ b/src/query/procedure/module.cpp
@@ -35,7 +35,7 @@ namespace memgraph::query::procedure {
 
 constexpr const char *func_code =
     "import ast\n\n"
-    "no_removals = ['collections', 'abc', 'sys']\n"
+    "no_removals = ['collections', 'abc', 'sys', 'torch', 'igraph']\n"
     "modules = set()\n\n"
     "def visit_Import(node):\n"
     "  for name in node.names:\n"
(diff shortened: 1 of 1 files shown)


gitbuda:
This is unusual because 2 third-party Mage modules are ignored...

Did you consider the ability to inject the blacklist as an argument and then inject it on the Mage side since the modules are introduced there?

as51340:
> This is unusual because 2 third-party Mage modules are ignored...
Mage modules aren't ignored, only Python packages are.
 
> Did you consider the ability to inject the blacklist as an argument and then inject it on the Mage side since the modules are introduced there?
I did but what would we get when all (re)loading is done on the core side?

gitbuda:
Having this runtime configurable makes a lot of sense to me because Memgraph doesn't have to be recompiled if another third-party module has to be included. The only real question to me is what is the default value?

What does happen with Memgraph without the changes here? Does it crash, or what exactly happens?

antoniofilipovic (approved):

vpavicic:
RN: `torch` and `igraph` can no longer be removed from the `svs.modules` cache to avoid issues after reload.

PR #725, opened 2022-12-21
[master < T521-FL] Fix nested FOREACH variable shadowing bug

brunos252:
[master < Task] PR

### DIFF
--- a/src/query/frontend/semantic/symbol_generator.cpp
+++ b/src/query/frontend/semantic/symbol_generator.cpp
@@ -334,7 +334,11 @@ SymbolGenerator::ReturnType SymbolGenerator::Visit(Identifier &ident) {
       }
       type = scope.visiting_edge->IsVariable() ? Symbol::Type::EDGE_LIST : Symbol::Type::EDGE;
     }
-    symbol = GetOrCreateSymbolLocalScope(ident.name_, ident.user_declared_, type);
+    if (scope.in_foreach) {
+      symbol = GetOrCreateSymbol(ident.name_, ident.user_declared_, type);
+    } else {
+      symbol = GetOrCreateSymbolLocalScope(ident.name_, ident.user_declared_, type);
+    }
   } else if (scope.in_pattern && !scope.in_pattern_atom_identifier && scope.in_match) {
     if (scope.in_edge_range && scope.visiting_edge->identifier_->name_ == ident.name_) {
       // Prevent variable path bounds to reference the identifier which is bound
--- a/src/query/frontend/semantic/symbol_generator.cpp
+++ b/src/query/frontend/semantic/symbol_generator.cpp
@@ -1,4 +1,4 @@
-// Copyright 2022 Memgraph Ltd.
+// Copyright 2023 Memgraph Ltd.
 //
 // Use of this software is governed by the Business Source License
 // included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
@@ -335,6 +335,10 @@ SymbolGenerator::ReturnType SymbolGenerator::Visit(Identifier &ident) {
       type = scope.visiting_edge->IsVariable() ? Symbol::Type::EDGE_LIST : Symbol::Type::EDGE;
     }
     if (scope.in_foreach) {
+      // Since foreach looks at symbols in a global scope it should also check it
+      if ((scope.in_create_node || scope.in_create_edge) && HasSymbol(ident.name_)) {
+        throw RedeclareVariableError(ident.name_);
+      }
       symbol = GetOrCreateSymbol(ident.name_, ident.user_declared_, type);
     } else {
       symbol = GetOrCreateSymbolLocalScope(ident.name_, ident.user_declared_, type);
--- a/src/query/frontend/semantic/symbol_generator.cpp
+++ b/src/query/frontend/semantic/symbol_generator.cpp
@@ -64,14 +64,6 @@ auto SymbolGenerator::CreateSymbol(const std::string &name, bool user_declared, 
   return symbol;
 }
 
-auto SymbolGenerator::GetOrCreateSymbolLocalScope(const std::string &name, bool user_declared, Symbol::Type type) {
-  auto &scope = scopes_.back();
-  if (auto maybe_symbol = FindSymbolInScope(name, scope, type); maybe_symbol) {
-    return *maybe_symbol;
-  }
-  return CreateSymbol(name, user_declared, type);
-}
-
 auto SymbolGenerator::GetOrCreateSymbol(const std::string &name, bool user_declared, Symbol::Type type) {
   // NOLINTNEXTLINE
   for (auto scope = scopes_.rbegin(); scope != scopes_.rend(); ++scope) {
@@ -206,7 +198,7 @@ bool SymbolGenerator::PreVisit(CallProcedure &call_proc) {
 
 bool SymbolGenerator::PostVisit(CallProcedure &call_proc) {
   for (auto *ident : call_proc.result_identifiers_) {
-    if (HasSymbolLocalScope(ident->name_)) {
+    if (HasSymbol(ident->name_)) {
       throw RedeclareVariableError(ident->name_);
     }
     ident->MapTo(CreateSymbol(ident->name_, true));
@@ -217,7 +209,7 @@ bool SymbolGenerator::PostVisit(CallProcedure &call_proc) {
 bool SymbolGenerator::PreVisit(LoadCsv &load_csv) { return false; }
 
 bool SymbolGenerator::PostVisit(LoadCsv &load_csv) {
-  if (HasSymbolLocalScope(load_csv.row_var_->name_)) {
+  if (HasSymbol(load_csv.row_var_->name_)) {
     throw RedeclareVariableError(load_csv.row_var_->name_);
   }
   load_csv.row_var_->MapTo(CreateSymbol(load_csv.row_var_->name_, true));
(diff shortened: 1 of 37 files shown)


brunos252:
As mentioned in Slack, this fix causes the following test to fail:
`FOREACH (i IN [1] | FOREACH (j IN [3,4] | CREATE (i {prop: j})));`
Expected output is to make 2 nodes, actual output is that nothing happens.
Neo4j has this functionality which also makes 2 nodes.

while this fix solves the issue linked, which is to be able to run:
`foreach (i in range(0,100)| create (n:NODE {i: i}));
match (n)
with n
order by n.id
with collect(n) as n_
FOREACH (i in range(0, size(n_) - 2) |
 FOREACH (node1 in [n_[i]] |
  FOREACH (node2 in [n_[i+1]] |
    FOREACH (j in [1] |
     MERGE (node1)-[:foo]->(node2)))));`
     
Personally I think the first foreach query doesn't make sense and should throw `RedeclareVariableError`, similar to the query `MATCH (n) CREATE (n);`
But aside from that, this fix allows us to use foreach with variables outside its scope, so for the second query we see `node1`. But to make the first query work, we would need to limit FOREACH's scope at the same time.

antaljanosbenjamin:
I think you are on the good path, however I think there is one thing that can be confusing: the difference between the following two queries:
- `FOREACH (i IN [1] | FOREACH (j IN [3,4] | CREATE (i {prop: j})));`
- `FOREACH (i IN [1] | FOREACH (j IN [3,4] | CREATE (o {prop: j})));`

The first one won't create any nodes, but the second one will create 2 nodes. How could we phrase that properly for the users? In my opinion we cannot explain this behavior in a way that users can understand. As I see there are two ways to make this easier to understand for the users:
1. Make shadowing a actual error. This is basically removing the "multi-scope" functionality we introduced with FOREACH.
2. We should clearly check whether we are introducing a new identifier (e.g.: `CREATE (i {prop:1})` introduces `i`) or want to reference and existing one (e.g.: `CREATE (n {prop:j})` references `j`, not introducing it). Every time we want to create an identifier, we only operate on the local scope, but when we want to reference a variable, then we should consider the parent scopes also. After Wednesday I can have a brief look and help you if you need help.

Even though I am against of releasing this in the current state (it will create a lot of confusion and angry users in my opinion), @gitbuda what do you think?

antaljanosbenjamin (commented):
I think you are on the good path, however I think there is one thing that can be confusing: the difference between the following two queries:
- `FOREACH (i IN [1] | FOREACH (j IN [3,4] | CREATE (i {prop: j})));`
- `FOREACH (i IN [1] | FOREACH (j IN [3,4] | CREATE (o {prop: j})));`

The first one won't create any nodes, but the second one will create 2 nodes. How could we phrase that properly for the users? In my opinion we cannot explain this behavior in a way that users can understand. As I see there are two ways to make this easier to understand for the users:
1. Make shadowing a actual error. This is basically removing the "multi-scope" functionality we introduced with FOREACH.
2. We should clearly check whether we are introducing a new identifier (e.g.: `CREATE (i {prop:1})` introduces `i`) or want to reference and existing one (e.g.: `CREATE (n {prop:j})` references `j`, not introducing it). Every time we want to create an identifier, we only operate on the local scope, but when we want to reference a variable, then we should consider the parent scopes also. After Wednesday I can have a brief look and help you if you need help.

Even though I am against of releasing this in the current state (it will create a lot of confusion and angry users in my opinion), @gitbuda what do you think?

brunos252:
In the current state, shadowing does throw an error, that we are trying to redeclare a variable. It is intuitive to me that you wouldn't use a variable that you use for iteration inside the foreach, same as using the iterator in a for loop.

src/query/frontend/semantic/symbol_generator.cpp
```diff
@@ -334,7 +334,15 @@ SymbolGenerator::ReturnType SymbolGenerator::Visit(Identifier &ident) {
...
-    symbol = GetOrCreateSymbolLocalScope(ident.name_, ident.user_declared_, type);
+    if (scope.in_foreach) {
+      // Since foreach looks at symbols in a global scope it should also check it
+      if ((scope.in_create_node || scope.in_create_edge) && HasSymbol(ident.name_)) {
+        throw RedeclareVariableError(ident.name_);
+      }
+      symbol = GetOrCreateSymbol(ident.name_, ident.user_declared_, type);
+    } else {
+      symbol = GetOrCreateSymbolLocalScope(ident.name_, ident.user_declared_, type);
+    }
```
antaljanosbenjamin, line 345:
```suggestion
    symbol = GetOrCreateSymbol(ident.name_, ident.user_declared_, type);
```

Plus changing the similar condition above `if ((scope.in_create_node || scope.in_create_edge) && HasSymbolLocalScope(ident.name_))` to `if ((scope.in_create_node || scope.in_create_edge) && HasSymbol(ident.name_))` should have the same behavior, isn't it?

antaljanosbenjamin (commented):
Sorry, I was checking this PR on my local machine and somehow I messed up the pull and I was checking the wrong state of the PR.

gitbuda:
Since like we are back on track here, I think we have time here 😄

antaljanosbenjamin:
Two more comment that popped into my mind: the functions `HasSymbolLocalScope` and `GetOrCreateSymbolLocalScope` and the whole "multiple scope" approach. My arguments:
1. As we want to prevent shadowing, we shouldn't care whether the identifier is declared in our local scope or not.
2. Furthermore we shouldn't care about scopes anymore. We introduced multi scopes just to support shadowing, but now we want to prevent users to shadow identifiers => Remove them, the code will be easier to understand/reason about.
3. As we have a bug here, we might have other bugs that we just cannot came up with yet, therefore eliminating the possibility of bugs reduces the chance of keeping a bug alive.

These are only my thoughts, not a decision from team/somebody, so please feel free to comment/argue them.

src/query/frontend/semantic/symbol_generator.cpp
brunos252: It will have the same behavior for FOREACH, I'm not sure why it is OK to do this for every case. before changes, where you propose putting `GetOrCreateSymbol`, there stood `GetOrCreateSymbolLocalScope`. I don't see why this can be replaced, since the LocalScope variant only looks in the local scope and creates a new symbol if it is not there. This is the scope other clauses are working with. For FOREACH I decided to use global scope to solve shadowing.

brunos252: I now see the comment below regarding scopes, I will try it out a bit and see if it is OK to eliminate local scopes, but then we also need to remove those LocalScope function calls elsewhere

brunos252:
I would agree that we don't need shadowing then, but we would need to see about edge cases, what are the arguments for having it. Do you maybe have some examples where shadowing is used?

brunos252:
I removed a total of 3 tests that were failing after removing local scope, coincidentally all related to FOREACH.
1. From unit test `query_semantic`:
```
FOREACH (i in [1] | CREATE (n)) RETURN n;
FOREACH (i in [1] | CREATE (n)) RETURN i;
```
Note: I assumed `[1]` because I didn't find what `i` iterates over, but it is irrelevant.
Since we deleted local scope, we can now return symbols from within `FOREACH`.

2. From unit test `query_plan`:
`FOREACH(i in [1, 2] | CREATE (n)) FOREACH (j in [1, 2] | CREATE (n))`

@antaljanosbenjamin

brunos252:
One inconvenience is, for:
```
FOREACH (i in [1, 2, 3] | CREATE (n)) RETURN n;
```
query returns only one node, the last one created, while all three are created.

antaljanosbenjamin (approved):

antoniofilipovic:
> [master < Task] PR
> 
> * [ ]  Check, and update documentation if necessary
> * [ ]  Update [changelog](https://docs.memgraph.com/memgraph/changelog)
> * [ ]  Provide the full content or a guide for the final git message

@brunos252 don't forget to link to provide a link to docs PR if there is need for one and changelog PR so I can merge this.

antoniofilipovic (changes requested):
Check what is with removed test, otherwise seems like changes Benjo requested are here and implemented.

tests/gql_behave/tests/memgraph_V1/features/foreach.feature
```diff
@@ -66,29 +66,13 @@ Feature: Foreach
...
-    When executing query:
-      """
-      MATCH (n) RETURN n.prop
-      """
-    Then the result should be:
-      | n.prop |
-      | 3 |
-      | 4 |
-    And no side effects
-
```
antoniofilipovic, line 84: This test should be returned. I am not sure if this is expected, but currently it throws error: "Cannot create node 'i' with labels or properties, because it is already declared"

On the other hand, following works: FOREACH (i IN [1] | FOREACH (j IN [3,4] | CREATE ( {prop: i+j})));

brunos252: Since we have specific scopes but use them globally, the query:
`FOREACH (i IN [1] | FOREACH (j IN [3,4] | CREATE ( {prop: i+j})));` 
works because `CREATE` can see the `i` from outer foreach.
For the same reason, the query:
`FOREACH (i IN [1] | FOREACH (j IN [3,4] | CREATE (i {prop: j})));`
Doesn't work because we are not using the `i` from the outer `FOREACH`, but we are redeclaring it as a node, which is not allowed. 
Hope this clarifies the confusion

antoniofilipovic: Okay, makes sense.

PR #729, opened 2022-12-30
[:pineapple: < T1216MG] T1216 implement aggregate

42jeremy:
[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR


src/query/v2/plan/operator.cpp
```diff
@@ -1233,6 +1233,55 @@ class AggregateCursor : public Cursor {
...
+
+    if (!pulled_all_input_) {
+      while (!pulled_all_input_) {
+        ProcessAll(multi_frame, &context);
+      }
+      multi_frame.MakeAllFramesInvalid();
+      aggregation_it_ = aggregation_.begin();
+
+      // in case there is no input and no group_bys we need to return true
+      // just this once
```
jbajic, line 1247: This comment seems false

src/query/v2/plan/operator.cpp
```diff
@@ -1233,6 +1233,55 @@ class AggregateCursor : public Cursor {
...
+      frame[aggregation_elem.output_sym] = *aggregation_values_it++;
+    }
+
+    // place remember values on the frame
+    auto remember_values_it = aggregation_it_->second.remember_.begin();
+    for (const Symbol &remember_sym : self_.remember_) {
+      frame[remember_sym] = *remember_values_it++;
+    }
+
+    aggregation_it_++;
```
jbajic, line 1282:
```suggestion
    ++aggregation_it_;
```

src/query/v2/plan/operator.cpp
```diff
@@ -1311,6 +1360,36 @@ class AggregateCursor : public Cursor {
...
+
+    // calculate AVG aggregations (so far they have only been summed)
+    for (size_t pos = 0; pos < self_.aggregations_.size(); ++pos) {
+      if (self_.aggregations_[pos].op != Aggregation::Op::AVG) continue;
+      for (auto &kv : aggregation_) {
+        AggregationValue &agg_value = kv.second;
+        auto count = agg_value.counts_[pos];
+        auto *pull_memory = context->evaluation_context.memory;
+        if (count > 0) {
+          agg_value.values_[pos] = agg_value.values_[pos] / TypedValue(static_cast<double>(count), pull_memory);
```
jbajic, line 1387: If there are no elements we should still be able to calculate avg, but this seems as temp code, so feel free to do what you think make sense

src/query/v2/plan/operator.cpp
antaljanosbenjamin: If you check the first condition (`if (!pulled_all_input_)`), that will be true only on the first run, so here we only return true only one.

src/query/v2/plan/operator.cpp
antaljanosbenjamin: After discussing it on slack we agreed that setting the aggregation value to null is fine and the default value is null (check `EnsureInitialized`), so this is fine.

jbajic (approved):

PR #730, opened 2023-01-02
[master < T1201] Run mgbench diff under 30 min

antejavor:
[master < Task] PR

The bench graph basic queries are ignored in standard CI/CD call `pokec/medium/*/*.` Tail latency has been turned off by default. Since latency is not present at the grafana at the moment.

### DIFF
--- a/tests/mgbench/benchmark.py
+++ b/tests/mgbench/benchmark.py
@@ -142,7 +142,7 @@ parser.add_argument(
     with the presence of 300 write queries from write type or 30%""",
 )
 
-parser.add_argument("--tail-latency", type=int, default=100, help="Number of queries for the tail latency statistics")
+parser.add_argument("--tail-latency", type=int, default=30, help="Number of queries for the tail latency statistics")
 
 parser.add_argument(
     "--performance-tracking",
--- a/tests/mgbench/benchmark.py
+++ b/tests/mgbench/benchmark.py
@@ -223,13 +223,14 @@ def filter_benchmarks(generators, patterns):
                         patterns,
                     ):
                         current[group].append((query_name, query_func))
-            if len(current) > 0:
+            if len(current) == 0:
+              continue
                 # Ignore benchgraph "basic" queries in standard CI/CD run
                 for pattern in patterns:
                     res = pattern.count("*")
                     key = "basic"
                     if res >= 2 and key in current.keys():
-                        del current[key]
+                        current.pop(key)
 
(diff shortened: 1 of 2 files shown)


antejavor:
Executing 10 iterations of each query for tail latency increases execution time approximately from 22 to 35 minutes. We can go with a minimum of 10 iterations in the future once I add them to grafana, as previously agreed to have latency and memory usage views in grafana (obviously not a priority at the moment). Further optimizations require a reduction of queries in the workload.

tests/mgbench/benchmark.py
```diff
@@ -224,6 +224,13 @@ def filter_benchmarks(generators, patterns):
                     ):
                         current[group].append((query_name, query_func))
             if len(current) > 0:
+                # Ignore benchgraph "basic" queries in standard CI/CD run
+                for pattern in patterns:
+                    res = pattern.count("*")
+                    key = "basic"
+                    if res >= 2 and key in current.keys():
+                        del current[key]
```
jbajic, line 232: Avoid usage of del
```suggestion
                        current.pop(key)
```

tests/mgbench/benchmark.py
```diff
@@ -224,6 +224,13 @@ def filter_benchmarks(generators, patterns):
                     ):
                         current[group].append((query_name, query_func))
             if len(current) > 0:
```
jbajic, line 226: This way you can reduce indentation
```suggestion
            if len(current) == 0:
              continue
```

tests/mgbench/benchmark.py
```diff
@@ -241,30 +248,34 @@ def warmup(client):
...
-        "p99": latency[math.floor(iteration * 0.99) - 1],
-        "p95": latency[math.floor(iteration * 0.95) - 1],
-        "p90": latency[math.floor(iteration * 0.90) - 1],
-        "p75": latency[math.floor(iteration * 0.75) - 1],
-        "p50": latency[math.floor(iteration * 0.50) - 1],
-    }
-    print("Query statistics for tail latency: ")
-    print(query_stats)
-    vendor.stop("tail_latency")
+    if iteration != 0 and iteration >= 10:
```
jbajic, line 252: When can the second condition be satisfied without satisfying the first?

antejavor: Epic fail :D

antejavor, line 252:
```suggestion
    if iteration >= 10:
```

jbajic (approved):

PR #732, opened 2023-01-09
[master < T1220-MG] Fix bug on properties in c++ api

antoniofilipovic:
This PR fixes few bugs:

- there was a problem with getting Properties - since `std::string_view` was used as a key to store Properties in `std::map`, due to the nature of how properties are retrieved from `mgp_properties_get_next` in combination with `std::string_view`, the new property would overwrite the old one and the user wouldn't get the correct state of properties (some would miss, std::map would be filled with same property)
- there was a missing set_property function in both `mgp.hpp` and `_mgp.hpp`

[master < Task] PR
- [x] Check, and update documentation if necessary -> [yes](https://github.com/memgraph/docs/pull/655)
- [x] Update [changelog](https://docs.memgraph.com/memgraph/changelog) -> still no
- [x] Provide the full content or a guide for the final git message: Fix bug on (vertex|edge) properties in C++ API

### DIFF
--- a/include/mgp.hpp
+++ b/include/mgp.hpp
@@ -564,7 +564,7 @@ class Node {
   /// @brief Returns an iterable & indexable structure of the node’s properties.
   std::map<std::string, Value> Properties() const;
 
-  void SetProperty(std::string property, Value value) const;
+  void SetProperty(std::string property, Value value);
 
   Value GetProperty(const std::string &property) const;
 
@@ -619,7 +619,7 @@ class Relationship {
   /// @brief Returns an std::map of the relationship’s properties.
   std::map<std::string, Value> Properties() const;
 
-  void SetProperty(std::string property, Value value) const;
+  void SetProperty(std::string property, Value value);
 
   Value GetProperty(const std::string &property) const;
 
@@ -2324,17 +2324,16 @@ inline void Node::AddLabel(const std::string_view label) {
 
 inline std::map<std::string, Value> Node::Properties() const {
   mgp_properties_iterator *properties_iterator = mgp::vertex_iter_properties(ptr_, memory);
-  std::map<std::string, Value> property_map_;
+  std::map<std::string, Value> property_map;
   for (auto *property = mgp::properties_iterator_get(properties_iterator); property;
        property = mgp::properties_iterator_next(properties_iterator)) {
-    auto value = Value(property->value);
-    property_map_.emplace(std::string(property->name), value);
+    property_map.emplace(std::string(property->name), Value(property->value));
   }
   mgp::properties_iterator_destroy(properties_iterator);
-  return property_map_;
+  return property_map;
 }
 
-inline void Node::SetProperty(std::string property, Value value) const {
+inline void Node::SetProperty(std::string property, Value value) {
   mgp::vertex_set_property(ptr_, property.data(), value.ptr());
 }
 
@@ -2391,22 +2390,22 @@ inline std::string_view Relationship::Type() const { return mgp::edge_get_type(p
 
 inline std::map<std::string, Value> Relationship::Properties() const {
   mgp_properties_iterator *properties_iterator = mgp::edge_iter_properties(ptr_, memory);
-  std::map<std::string, Value> property_map_;
+  std::map<std::string, Value> property_map;
   for (mgp_property *property = mgp::properties_iterator_get(properties_iterator); property;
        property = mgp::properties_iterator_next(properties_iterator)) {
-    Value value = Value(property->value);
-    property_map_.emplace(property->name, value);
+    property_map.emplace(property->name, Value(property->value));
   }
   mgp::properties_iterator_destroy(properties_iterator);
-  return property_map_;
+  return property_map;
 }
 
-inline void Relationship::SetProperty(std::string property, Value value) const {
+inline void Relationship::SetProperty(std::string property, Value value) {
   mgp::edge_set_property(ptr_, property.data(), value.ptr());
 }
 
 inline Value Relationship::GetProperty(const std::string &property) const {
-  return Value(mgp::edge_get_property(ptr_, property.data(), memory));
+  mgp_value *edge_prop = mgp::edge_get_property(ptr_, property.data(), memory);
+  return Value(&edge_prop);
 }
 
 inline Node Relationship::From() const { return Node(mgp::edge_get_from(ptr_)); }
--- a/include/mgp.hpp
+++ b/include/mgp.hpp
@@ -2331,7 +2331,7 @@ inline std::map<std::string, Value> Node::Properties() const {
     property_map_.emplace(std::string(property->name), value);
   }
   mgp::properties_iterator_destroy(properties_iterator);
-  return std::move(property_map_);
+  return property_map_;
 }
 
 inline void Node::SetProperty(std::string property, Value value) const {
@@ -2398,7 +2398,7 @@ inline std::map<std::string, Value> Relationship::Properties() const {
     property_map_.emplace(property->name, value);
   }
   mgp::properties_iterator_destroy(properties_iterator);
-  return std::move(property_map_);
+  return property_map_;
 }
 
 inline void Relationship::SetProperty(std::string property, Value value) const {
--- a/include/mgp.hpp
+++ b/include/mgp.hpp
@@ -66,6 +66,9 @@ struct MapItem;
 class Duration;
 class Value;
 
+struct StealType {};
+inline constexpr StealType steal{};
+
 inline mgp_memory *memory{nullptr};
 
 /* #region Graph (Id, Graph, Nodes, GraphRelationships, Relationships & Labels) */
@@ -954,7 +957,6 @@ class Value {
 
   explicit Value(mgp_value *ptr);
 
-  struct StealType {};
   explicit Value(StealType /*steal*/, mgp_value *ptr);
 
   // Null constructor:
@@ -1544,8 +1546,6 @@ inline Type ToAPIType(mgp_value_type type) {
 /* #region Graph (Id, Graph, Nodes, GraphRelationships, Relationships, Properties & Labels) */
 
 // Id:
-
-inline constexpr Value::StealType steal{};
 
 inline Id Id::FromUint(uint64_t id) { return Id(util::MemcpyCast<int64_t>(id)); }
 
@@ -2343,7 +2343,7 @@ inline void Node::SetProperty(std::string property, Value value) {
 
 inline Value Node::GetProperty(const std::string &property) const {
   mgp_value *vertex_prop = mgp::vertex_get_property(ptr_, property.data(), memory);
-  return Value(mgp::steal, vertex_prop);
+  return Value(steal, vertex_prop);
 }
 
 inline bool Node::operator<(const Node &other) const { return Id() < other.Id(); }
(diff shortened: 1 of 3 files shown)


include/mgp.hpp
```diff
@@ -653,7 +657,7 @@ class Relationship {
   class Properties Properties() const;
 
   /// @brief Returns the value of the relationship’s `property_name` property.
-  Value operator[](const std::string_view property_name) const;
+  Value operator[](const std::string &property_name) const;
```
antepusic, line 660: Is `Value &operator[](const std::string &property_name)` missing in class Relationship?

antepusic (changes requested):
Good work! Would it make sense to replace `string_view` with `string` elsewhere in the API as well?

include/mgp.hpp
antoniofilipovic: Done

antoniofilipovic:
> Good work! Would it make sense to replace `string_view` with `string` elsewhere in the API as well?

Since there are no problems, I would leave it as is.

include/mgp.hpp
```diff
@@ -596,10 +562,11 @@ class Node {
   bool HasLabel(std::string_view label) const;
 
   /// @brief Returns an iterable & indexable structure of the node’s properties.
-  class Properties Properties() const;
+  std::map<std::string, Value> Properties() const;
 
-  /// @brief Returns the value of the node’s `property_name` property.
-  Value operator[](const std::string_view property_name) const;
+  void SetProperty(std::string property, Value value) const;
```
antaljanosbenjamin, line 567:
```suggestion
  void SetProperty(std::string property, Value value);
```

Even though the underlying implementation can be `const` because of the operation doesn't change this object, logically this shouldn't be const. This objects represents a vertex which is changed by modifying one of its properties.

include/mgp.hpp
```diff
@@ -649,11 +616,12 @@ class Relationship {
   /// @brief Returns the relationship’s type.
   std::string_view Type() const;
 
-  /// @brief Returns an iterable & indexable structure of the relationship’s properties.
-  class Properties Properties() const;
+  /// @brief Returns an std::map of the relationship’s properties.
+  std::map<std::string, Value> Properties() const;
+
+  void SetProperty(std::string property, Value value) const;
```
antaljanosbenjamin, line 622:
```suggestion
  void SetProperty(std::string property, Value value);
```

include/mgp.hpp
```diff
@@ -985,6 +953,7 @@ class Value {
   friend class Result;
 
   explicit Value(mgp_value *ptr);
+  explicit Value(mgp_value **ptr);
```
antaljanosbenjamin, line 956: I would remove this and force the caller to dereference the pointer.

include/mgp.hpp
```diff
@@ -2390,6 +2322,27 @@ inline void Node::AddLabel(const std::string_view label) {
...
+inline std::map<std::string, Value> Node::Properties() const {
+  mgp_properties_iterator *properties_iterator = mgp::vertex_iter_properties(ptr_, memory);
+  std::map<std::string, Value> property_map_;
+  for (auto *property = mgp::properties_iterator_get(properties_iterator); property;
+       property = mgp::properties_iterator_next(properties_iterator)) {
+    auto value = Value(property->value);
+    property_map_.emplace(std::string(property->name), value);
+  }
+  mgp::properties_iterator_destroy(properties_iterator);
+  return property_map_;
```
antaljanosbenjamin, line 2334:
```suggestion
  std::map<std::string, Value> property_map;
  for (auto *property = mgp::properties_iterator_get(properties_iterator); property;
       property = mgp::properties_iterator_next(properties_iterator)) {
    auto value = Value(property->value);
    property_map.emplace(std::string(property->name), value);
  }
  mgp::properties_iterator_destroy(properties_iterator);
  return property_map;
```

I would also either move `value` into `property_map` or even better: let the compiler do that: `property_map_.emplace(std::string(property->name), Value(property->value));`. This eliminates an unnecessary copy of `Value`.

include/mgp.hpp
```diff
@@ -2390,6 +2322,27 @@ inline void Node::AddLabel(const std::string_view label) {
...
+}
+
+inline void Node::SetProperty(std::string property, Value value) const {
+  mgp::vertex_set_property(ptr_, property.data(), value.ptr());
+}
+
+inline Value Node::GetProperty(const std::string &property) const {
+  mgp_value *vertex_prop = mgp::vertex_get_property(ptr_, property.data(), memory);
+  return Value(&vertex_prop);
```
antaljanosbenjamin, line 2343: Why do you need to get the address of a pointer? Isn't `Value(vertex_prop)` would suffice?

include/mgp.hpp
```diff
@@ -2436,12 +2389,24 @@ inline mgp::Id Relationship::Id() const { return Id::FromInt(mgp::edge_get_id(pt
...
+inline std::map<std::string, Value> Relationship::Properties() const {
+  mgp_properties_iterator *properties_iterator = mgp::edge_iter_properties(ptr_, memory);
+  std::map<std::string, Value> property_map_;
+  for (mgp_property *property = mgp::properties_iterator_get(properties_iterator); property;
+       property = mgp::properties_iterator_next(properties_iterator)) {
+    Value value = Value(property->value);
+    property_map_.emplace(property->name, value);
+  }
+  mgp::properties_iterator_destroy(properties_iterator);
+  return property_map_;
```
antaljanosbenjamin, line 2401: Same comments as for `Node::Properties`.

include/mgp.hpp
```diff
@@ -2436,12 +2389,24 @@ inline mgp::Id Relationship::Id() const { return Id::FromInt(mgp::edge_get_id(pt
...
+}
+
+inline void Relationship::SetProperty(std::string property, Value value) const {
+  mgp::edge_set_property(ptr_, property.data(), value.ptr());
 }
 
-inline Value Relationship::operator[](const std::string_view property_name) const {
-  return Properties()[property_name];
+inline Value Relationship::GetProperty(const std::string &property) const {
+  return Value(mgp::edge_get_property(ptr_, property.data(), memory));
```
antaljanosbenjamin, line 2409: I think the same approach can be used in `Node::GetProperty` also.

include/mgp.hpp
```diff
@@ -2917,6 +2882,7 @@ inline bool Duration::operator<(const Duration &other) const {
 /* #region Value */
 
 inline Value::Value(mgp_value *ptr) : ptr_(mgp::value_copy(ptr, memory)) {}
+inline Value::Value(mgp_value **ptr) : ptr_(*ptr) {}
```
antaljanosbenjamin, line 2885: I think you don't need this.

tests/unit/cpp_api.cpp
```diff
@@ -419,3 +419,19 @@ TEST_F(CppApiTestFixture, TestDuration) {
...
+  auto node_1 = graph.CreateNode();
+
+  ASSERT_EQ(node_1.Properties().size(), 0);
+
+  std::map<std::string, mgp::Value> node1_prop = node_1.Properties();
+  node_1.SetProperty("b", mgp::Value(2.0));
+
+  ASSERT_EQ(node_1.Properties().size(), 1);
+  ASSERT_EQ(node_1.Properties()["b"].ValueDouble(), 2.0);
+  ASSERT_EQ(node_1.GetProperty("b").ValueDouble(), 2.0);
```
antaljanosbenjamin, line 436: If the type of the property doesn't matter, could you please use something that is more predictable than double? Maybe on some architectures the equality check might not fail for some precision related reasons.

antaljanosbenjamin (changes requested):

include/mgp.hpp
antoniofilipovic: Makese, sense. I will change it and use compailer to move value int map.

include/mgp.hpp
antoniofilipovic: Reason I added this part was because function Value::Value(mgp_value *ptr) : ptr_(mgp::value_copy(ptr, memory)) {} makes copy of mgp_value. In case of mgp::vertex_get_property, it returns copy of value. This is description of function:
"/// Get a copy of a vertex property mapped to a given name. Resulting value must be freed with mgp_value_destroy."

In case I make another copy, there will be one copy which is not deleted. And in that case there was memory leak. So with this function I am making a difference between case when we want copy, and when we don't want one.

include/mgp.hpp
antoniofilipovic: Problem is with function get_property which makes already copy of mgp_value. If we make another copy, then there will be 2 copies and one won't get deleted because we didn't use mgp_value_destory and one of them. I forgot to apply that here, so I am adding it now.

include/mgp.hpp
antoniofilipovic: Can you check inline Value::Value(mgp_value **ptr) : ptr_(*ptr) {} for more extensive comment? Also check Relationship::GetProperty. If that doesn't make sense, I will change it.

include/mgp.hpp
antoniofilipovic: Can you check `inline Value::Value(mgp_value **ptr) : ptr_(*ptr) {}` for more extensive comment why I used this? If that doens't make sense I can change it.

include/mgp.hpp
antoniofilipovic: Makes sense

include/mgp.hpp
antoniofilipovic: I agree, changed it!

tests/unit/cpp_api.cpp
antoniofilipovic: Of course :)

include/mgp.hpp
antaljanosbenjamin: Ah I see it. Then the overloaded constructor is fine, however it is not clear for the interface. I would rather use something like:
```
struct StealType {}:

inline constexpr StealType steal{};

constexpr auto 

Value(Steal /*steal*/, mgp_value* ptr) : ptr_{ptr}{};
```

If you consider how do you call them:
```
mgp_value *ptr = ...;

auto value1 = Value(&ptr);
auto value2 = Value(steal, ptr);
```
I think `value2` is more clear. For more info/example, please checkout `std::pair`'s [ctor](https://en.cppreference.com/w/cpp/utility/pair/pair) (8th form) uses [`std::piecewise_construct_t`](https://en.cppreference.com/w/cpp/utility/piecewise_construct).

antoniofilipovic: Wow, really cool! Will add this. Thanks for the suggestion :smile:

antoniofilipovic: Check if it is okay now :)

include/mgp.hpp
```diff
@@ -1543,6 +1545,8 @@ inline Type ToAPIType(mgp_value_type type) {
 
 // Id:
 
+inline constexpr Value::StealType steal{};
```
antoniofilipovic, line 1548: I am not sure if this was an idea of how to add a `steal` object?

antaljanosbenjamin: I would say just put it into the `mgp` namespace as it might be useful for other types also, not just for `Value`.

include/mgp.hpp
```diff
@@ -2917,6 +2885,7 @@ inline bool Duration::operator<(const Duration &other) const {
 /* #region Value */
 
 inline Value::Value(mgp_value *ptr) : ptr_(mgp::value_copy(ptr, memory)) {}
+inline Value::Value(StealType /*steal*/, mgp_value *ptr) : ptr_{ptr} {};
```
antaljanosbenjamin, line 2888:
```suggestion
inline Value::Value(StealType /*steal*/, mgp_value *ptr) : ptr_{ptr} {}
```

antaljanosbenjamin (approved):

include/mgp.hpp
antoniofilipovic: done

antepusic (approved):
Looks good to go :rocket:

PR #736, opened 2023-01-17
[master < T565-FL] Use indexes in optional match and foreach

Josipmrden:
Indexes were not applied when using optional match with label property index.The actual problem was that in the additional branch, optional match would not see the bounded symbol when filtering by that symbol's property. Index could not be fetched accordingly

FOREACH. Index was not applied correctly since the plan was forbidden to alter the inner children of the foreach logical operator. Merge can be one of the children, hence the index lookup rewriter can create the index.

[master < Task] PR

### DIFF
--- a/src/query/plan/rewrite/index_lookup.hpp
+++ b/src/query/plan/rewrite/index_lookup.hpp
@@ -1,4 +1,4 @@
-// Copyright 2022 Memgraph Ltd.
+// Copyright 2023 Memgraph Ltd.
 //
 // Use of this software is governed by the Business Source License
 // included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
@@ -444,7 +444,7 @@ class IndexLookupRewriter final : public HierarchicalLogicalOperatorVisitor {
 
   bool PreVisit(Foreach &op) override {
     prev_ops_.push_back(&op);
-    return false;
+    return true;
   }
 
   bool PostVisit(Foreach &) override {
--- a/src/query/plan/rewrite/index_lookup.hpp
+++ b/src/query/plan/rewrite/index_lookup.hpp
@@ -444,7 +444,9 @@ class IndexLookupRewriter final : public HierarchicalLogicalOperatorVisitor {
 
   bool PreVisit(Foreach &op) override {
     prev_ops_.push_back(&op);
-    return true;
+    op.input()->Accept(*this);
+    RewriteBranch(&op.update_clauses_);
+    return false;
   }
 
   bool PostVisit(Foreach &) override {
--- a/tests/unit/query_plan.cpp
+++ b/tests/unit/query_plan.cpp
@@ -1729,6 +1729,9 @@ TYPED_TEST(TestPlanner, Foreach) {
     std::list<BaseOpChecker *> updates{&create};
     std::list<BaseOpChecker *> input;
     CheckPlan(planner.plan(), symbol_table, ExpectForeach(input, updates), ExpectEmptyResult());
+
+    DeleteListContent(&on_match);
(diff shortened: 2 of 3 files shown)


src/query/plan/rewrite/index_lookup.hpp
```diff
@@ -444,7 +444,7 @@ class IndexLookupRewriter final : public HierarchicalLogicalOperatorVisitor {
 
   bool PreVisit(Foreach &op) override {
     prev_ops_.push_back(&op);
-    return false;
+    return true;
```
antaljanosbenjamin, line 447: This won't visit the internal brancch of the `FOREACH` operator, right?

Josipmrden: it will, my mistake. will correct to rewrite the branch with the new root.

antoniofilipovic (commented):
I checked the optional match and it is fixed, and for `foreach` clause I am not sure what was bug, so we can discuss and I will then check that part.

tests/unit/query_plan.cpp
```diff
@@ -662,6 +662,30 @@ TYPED_TEST(TestPlanner, MatchOptionalMatchWhereReturn) {
   DeleteListContent(&optional);
 }
 
+TYPED_TEST(TestPlanner, MatchOptionalMatchNodePropertyWithIndex) {
+  // Test MATCH (n) OPTIONAL MATCH (m) WHERE n.prop = m.prop RETURN n
```
antaljanosbenjamin, line 666: Both `n` and `m` should have label in the query, aren't they? You implemented the right thing, but this comment is a bit incomplete.

tests/unit/query_plan.cpp
```diff
@@ -1685,5 +1709,29 @@ TYPED_TEST(TestPlanner, Foreach) {
         QUERY(SINGLE_QUERY(FOREACH(i, {CREATE(PATTERN(NODE("n")))}), FOREACH(j, {CREATE(PATTERN(NODE("n")))})));
     CheckPlan<TypeParam>(query, storage, ExpectForeach(input, updates), ExpectEmptyResult());
   }
+
+  {
+    // FOREACH with index
```
antaljanosbenjamin, line 1714: Here we just verify that the planner does visit the inside of the `FOREACH` clause right? And I assume (based on the changes I saw) that didn't work before, right?

antaljanosbenjamin (approved):
I tihnk the changes are fine, only one comment should be extended. Please make a note somewhere about what is the expected behavior of `FOREACH` after the changes. Nothing serious, just something like:
> The inside branch of `FOREACH` will be planned prpoerly using the information inside it, thus it means proper index usage is expected among other changes from the plan rewriter.

A question that I don't know how to answer: does the clauses before the `FOREACH` can affect the planning of the inside branch of `FOREACH`? The reasons I am not sure how to answer it:
1. I don't know exactly how the `RewriteBranch` function works, whether it handles the informations properly. Based on a short look I had I think the filters are not transfered, so I think they won't have affect on the planning of the inside branch.
2. Actually I am also not sure whether we can write something that actually takes into account any filter, because we only allow "write" clauses, such as `FOREACH`, `CREATE`, `DELETE`, `DETACH`, `MERGE`, `REMOVE`, `SET`.

tests/unit/query_plan.cpp
Josipmrden: yes that's true, my mistake

tests/unit/query_plan.cpp
Josipmrden: yes. It used the ScanAll index rather than indexed scan.

antoniofilipovic (approved):

PR #688, opened 2022-12-01
Fix MAGE C++ API bugs

antepusic:
[master < Task] PR
- [x] Check, and update documentation if necessary
- [ ] Update [changelog](https://docs.memgraph.com/memgraph/changelog)
- [ ] Provide the full content or a guide for the final git message

PR #743, opened 2023-01-22
[master < T323-MAGE] Fix C++ API memory leak on Relationships()

antepusic:
NB: The first 19 commits are the same as the commits in #688 (already merged).

[master < Epic] PR
- [x] Write E2E tests
- [x] Compare the [benchmarking results](https://bench-graph.memgraph.com/) between the master branch and the Epic branch

### DIFF
--- a/include/mgp.hpp
+++ b/include/mgp.hpp
@@ -581,7 +581,7 @@ class List {
   List(List &&other) noexcept : ptr_(other.ptr_) { other.ptr_ = nullptr; }
 
   List &operator=(const List &other) = delete;
-  List &operator=(List &&other) = delete;
+  List &operator=(List &&other) noexcept;
 
   ~List();
 
@@ -676,7 +676,7 @@ class Map {
   Map(Map &&other) noexcept : ptr_(other.ptr_) { other.ptr_ = nullptr; }
 
   Map &operator=(const Map &other) = delete;
-  Map &operator=(Map &&other) = delete;
+  Map &operator=(Map &&other) noexcept;
 
   ~Map();
 
@@ -799,7 +799,7 @@ class Node {
   Node(Node &&other) noexcept : ptr_(other.ptr_) { other.ptr_ = nullptr; }
 
   Node &operator=(const Node &other) = delete;
-  Node &operator=(Node &&other) = delete;
+  Node &operator=(Node &&other) noexcept;
 
   ~Node();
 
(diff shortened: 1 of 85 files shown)


include/mgp.hpp
```diff
@@ -1814,10 +1815,12 @@ inline GraphRelationships::Iterator &GraphRelationships::Iterator::operator++()
...
 
-    if (next == nullptr) {
-      mgp::edges_iterator_destroy(out_relationships_iterator_);
-      out_relationships_iterator_ = nullptr;
+    if (next != nullptr) {
+      return *this;
     }
+
+    mgp::edges_iterator_destroy(out_relationships_iterator_);
+    out_relationships_iterator_ = nullptr;
```
antaljanosbenjamin, line 1823: You don't return anything here.

include/mgp.hpp
```diff
@@ -650,7 +651,7 @@ class Relationship {
   std::string_view Type() const;
 
   /// @brief Returns an iterable & indexable structure of the relationship’s properties.
-  class Properties Properties() const;
+  mgp::Properties Properties() const;
```
antoniofilipovic, line 654: From now on, we don't use the Properties class, so merge the master first into this PR.

antoniofilipovic (changes requested):
Adress Benjo's comment about not returning and check my comment about Properties.

include/mgp.hpp
antepusic: That’s the intended behavior!
If the iterator hasn’t found any more relationships on the current node, it doesn’t return anything. Instead, it moves on to go over the remaining nodes and return the first valid [outbound] relationship.

antepusic: That’s the intended behavior!
If the iterator hasn’t found any more relationships on the current node, it doesn’t return anything. Instead, it moves on to go over the remaining nodes and return the first valid [outbound] relationship.

include/mgp.hpp
antepusic: Done

include/mgp.hpp
antaljanosbenjamin: But that is a wrong behavior. Every non-void function **must** return a value. How would you check if something is returned or not? There is no way to achieve that. Therefore this function **must always** return a reference.

antaljanosbenjamin: Ahh sorry, I missed the scope of the function and thought you are never returrning from that function. My bad!

antaljanosbenjamin (approved):

jbajic (approved):

antoniofilipovic (approved):
Good to go.

PR #746, opened 2023-01-23
Try to add simplest mgbench dataset

gitbuda:


gitbuda:
@antejavor, if this is useful, feel free to squash-merge this to the base branch or add a comment about how the "test|example" dataset should look like 😄

We should provide the simplest possible example that can be copy-pasted when someone wants to add a new dataset on top of the `dataset.py|workload.py|common.py`.

PR #753, opened 2023-01-30
[master < T580-FL] Improve LOAD CSV speed

antoniofilipovic:
The idea of this top comment is the following:
- give a quick overview of what was improved (`1. What's improved`)
- give a more detailed overview of benchmarking (`2. Benchmarking`)
- explain steps towards speed improvements found during profiling (`3. Profiling summary`)

## 1. What's improved

After profiling analysis available in section `3. Profiling summary` there were 2 Pull requests to branch `master` of Memgraph:
 - #788 
 - #774

These two PRs proved to **increase** the performance of LOAD CSV import on rich graphs with lots of properties which was shown to be a critical point for improvement.

Another **important** thing to notice is that when doing the import, the **user** can **drastically enhance the performance** of import by using the `READ UNCOMMITED` isolation level on the transaction, due to the fact that it is probably the only transaction doing changes, and such transaction in most cases should be able to see its changes (changes that transaction made).

Description of test: Import X nodes with Y properties each from 1 file (`nodes.csv`) and then `Z` edges with `W` properties from another file( `edges.csv`).

| X | Y | Z| W |Throughput time (OLD)  [node/seconds] | Thorughput time (NEW) nodes [seconds] | Throughput (OLD) [edges/s] | Throughput (NEW) [edges/s] |
| -- | -- | -- |  -- | -- |  -- | -- | -- |
|100000 |10 | 200000| 10 |  120 K/s ( 0.83 s)  | **165 K/s (0.6 s)**  |  70 K/s (2.85 s) | **85 K/s (2.36 s)**  |
|100000 |20 | 200000| 20 |  48 K/s ( 2.06 s)  | **102 K/s (0.96 s)**  |  38 K/s (5.25 s) | **62.5 K/s (3.24 s)**  |
|100000 |30 | 200000| 30 |  28 K/s ( 3.5 s)  | **74 K/s (1.37 s)**  |  15 K/s (7.8 s) | **50 K/s (4.05s)**  |

## 2. Benchmarking

The idea of benchmarking was to check how much we improved with new changes introduced from **Profiling summary** down below, and how different ISOLATION levels on transaction impact throughput

 1. Import nodes and edges from the same file (in the same file there are 2 nodes and 1 edge per line)
- Description: Import 1 000 000 lines, and from each line create 2 nodes and 1 new edge. Each line will have X properties, of 10 string length. Each node has X/2 features, where X is even. No header.

| Number of properties per line | Import time (OLD) [seconds] | Import time (NEW) [seconds] | Throughput (OLD) [lines/s] | Throughput (NEW) [lines/s] |
| -- | -- | -- |  -- | -- |
|2 | 3.6s | 3.917 s| **277K** | 256K |
|10 | 9.6s | 8.4 s|  104 K | **119 K** |
|20 | 22s | 12.9 s| 45.5 K | **77.5 K**  |

2. Import nodes from one file, and then edges from another file
- Description: Import X nodes with Y properties nodes from 1 file and then Z edges with W properties from another file.

| X | Y | Z| W |Throughput time (OLD)  [node/seconds] | Thorughput time (NEW) nodes [seconds] | Throughput (OLD) [edges/s] | Throughput (NEW) [edges/s] |
| -- | -- | -- |  -- | -- |  -- | -- | -- |
|1M |10 | 5M| 5 |  117K/s (total 8.8seconds)  |  | 51K/s (total: 98 seconds)  |   |

3. Import nodes from one file, and edges from another file using `SNAPSHOT` or `READ UNCOMMITTED` isolation level
- Description: Import X nodes with few properties [1,3] and Z edges with few properties [1,3]

| X | Z |Throughput time (`SNAPSHOT`)  [node/seconds] | Throughput time (`READ UNCOMMITTED`) nodes [seconds] | Throughput (`SNAPSHOT`) [edges/s] | Throughput (`READ UNCOMMITTED`) [edges/s] | ratio (edges/nodes)|
| -- | -- | -- |  -- | -- | -- | -- |
|3500 | 50K|  50K/s (0.07 seconds)  |  50K/s (0.07 seconds)  | 100K/s (0.5 seconds) | 100K/s (0.5 seconds) | 14 |
| 3500 | 500K|  50K/s (0.07 seconds)  |  50K/s (0.07 seconds)  | 16K/s (31 seconds) | 106 K/s (4.7 seconds) :heavy_check_mark:  |  **142** |
|3.5M | 5M|  397 K/s (8.8 seconds)  |  397 K/s (8.8 seconds) | 90 K/s (55.1 seconds) |  92 K/s (54 seconds) | 1.5|

## 3. Profiling Summary

**The Problem and Goal**
LOAD CSV speed: 5000 lines/second at graphs with a lot of properties -> get at least 10 000 lines/second :heavy_check_mark: 

**Context**
 Dataset + Queries = Workload -> Read **Profiling Summary** below

| Experiment Name | Description | Throughput (lines/s) |  Throughput (edges/s) |
| -- | -- | -- | -- |
| Check baseline import speed on 100 000 lines, with 100 string properties per line, 10 char each string | Read more description [here](https://github.com/memgraph/memgraph/pull/753#issuecomment-1411921127).  **The test was done on [1]** | 4350 | 4350 |
| Implement `ReferenceExpressionEvaluator` to avoid copying every `row` when evaluating the expression | Visitor was the real bottleneck and I confirmed this with `gperftools` and by [measuring time spent in that function](https://github.com/memgraph/memgraph/pull/753#issuecomment-1411921127), it made sense to avoid copying the whole list just to get one object from it. **The test was done on [1]** | 8550 | 8550 |
|  Lookup of parameters [takes time](https://github.com/memgraph/memgraph/pull/753#issuecomment-1413677072) as it needs to iterate through `std::vector` each time to find a specific index | After profiling, it seemed it is better with `ReferenceExpressionEvaluator`, but `AtTokenPostion` was still taking a lot of time. I changed std::vector with std::map for faster lookup, in this case. **The test was done on [1]** | 9520 | 9520 |
| Test batch insertion of properties when creating vertices | Test batch insert of properties due to the fact that [flamegraph](https://user-images.githubusercontent.com/61245998/216326788-a423acc3-f4b4-473f-ba44-4378e0756529.svg) showed (read more [here](https://github.com/memgraph/memgraph/pull/753#issuecomment-1415948417)) it took a lot of time to insert property 1 by one. Due to guarantees on creating vertex that no two of the same properties can exist inside `{}`, we can batch insert properties without checking if it exists or not. **The test was done on [1]** | 10.9K | 10.9K |
| Implement `SetProperties` in `PropertyStore`| From profiling in flamegraph, it seems a lot of time is wasted on the DecodeExpectedProperty function which shows that probably problem is that for each new property when there is the setting of a new property, we need to check whether this property exists already, and copy memory block of properties. To avoid that, we can set all properties all at once. **The test was done on [1]** | 18.1K | 18.1K |
| Test using std::vector instead of `std::map` to process properties in batch| From flamegraph it seems like we can squeeze out a little bit more time out of execution since std::map takes a bit of time. **The test was done on [1]** |**>20K/s** | **>20K/s** |

**Test Summary**
| Test ID | Test query | Test description| Test file|
|--| --| --|--|
| 1 | `LOAD CSV FROM ".../generated_text.csv" NO HEADER AS row CREATE (a: Label {arg0:row[0],...,arg49:row[49]}), (b: Label {arg50:row[50],...,arg99:row[99]}), (a)-[:CONNECTED_TO]->(b);`  | 100 000 lines file with 100 properties per line, 10 char string each property. Create 2 nodes and 1 edge between them for each line (200 000 nodes created and 100 000 edges)| |
|TODO | TODO | TODO | TODO |

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -52,7 +52,7 @@ target_link_libraries(memgraph ${mg_single_node_v2_libs})
 # NOTE: `include/mg_procedure.syms` describes a pattern match for symbols which
 # should be dynamically exported, so that `dlopen` can correctly link the
 # symbols in custom procedure module libraries.
-target_link_libraries(memgraph "-Wl,--dynamic-list=${CMAKE_SOURCE_DIR}/include/mg_procedure.syms")
+target_link_libraries(memgraph "-Wl,--dynamic-list=${CMAKE_SOURCE_DIR}/include/mg_procedure.syms" "-lprofiler")
 set_target_properties(memgraph PROPERTIES
 
         # Set the executable output name to include version information.
--- a/src/query/db_accessor.hpp
+++ b/src/query/db_accessor.hpp
@@ -1,4 +1,4 @@
-// Copyright 2022 Memgraph Ltd.
+// Copyright 2023 Memgraph Ltd.
 //
 // Use of this software is governed by the Business Source License
 // included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
@@ -123,6 +123,11 @@ class VertexAccessor final {
 
   storage::Result<storage::PropertyValue> SetProperty(storage::PropertyId key, const storage::PropertyValue &value) {
     return impl_.SetProperty(key, value);
+  }
+
+  storage::Result<std::vector<storage::PropertyValue>> SetProperties(
+      std::map<storage::PropertyId, storage::PropertyValue> &properties) {
+    return impl_.SetBatchProperties(properties);
   }
 
   storage::Result<storage::PropertyValue> RemoveProperty(storage::PropertyId key) {
--- a/src/query/db_accessor.hpp
+++ b/src/query/db_accessor.hpp
@@ -126,8 +126,8 @@ class VertexAccessor final {
   }
 
   storage::Result<std::vector<storage::PropertyValue>> SetProperties(
-      std::map<storage::PropertyId, storage::PropertyValue> &properties) {
-    return impl_.SetBatchProperties(properties);
+      std::vector<std::pair<storage::PropertyId, storage::PropertyValue>> &properties) {
+    return impl_.SetProperties(properties);
   }
 
   storage::Result<storage::PropertyValue> RemoveProperty(storage::PropertyId key) {
--- a/src/query/interpret/eval.hpp
+++ b/src/query/interpret/eval.hpp
@@ -231,8 +231,6 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
   }
 
   TypedValue Visit(SubscriptOperator &list_indexing) override {
-    // auto start = std::chrono::steady_clock::now();
-
     ReferenceExpressionEvaluator referenceExpressionEvaluator(frame_, symbol_table_, ctx_, dba_, view_);
 
     auto *lhs = list_indexing.expression1_->Accept(referenceExpressionEvaluator);
@@ -255,14 +253,6 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
       if (index_int >= static_cast<int64_t>(list.size()) || index_int < 0) return TypedValue(ctx_->memory);
       // NOTE: Explicit move is needed, so that we return the move constructed
       // value and preserve the correct MemoryResource.
-      // auto end = std::chrono::steady_clock::now();
-      // std::chrono::duration<double> dif = end - start;
-      // total2 += dif;
-
-      // if (total2 > new_print2) {
-      //   std::cout << "Time difference visit = " << total2.count() << "[s]" << std::endl;
-      //   new_print2 += static_cast<std::chrono::duration<double>>(1.0);
-      // }
       return TypedValue(list[index_int]);
     }
 
--- a/src/query/interpret/eval.hpp
+++ b/src/query/interpret/eval.hpp
@@ -1,4 +1,4 @@
-// Copyright 2022 Memgraph Ltd.
+// Copyright 2023 Memgraph Ltd.
 //
 // Use of this software is governed by the Business Source License
 // included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
@@ -30,6 +30,78 @@
 #include "utils/exceptions.hpp"
 
 namespace memgraph::query {
+
+static std::chrono::duration<double> total2;
+static std::chrono::duration<double> new_print2;
+
+class ReferenceExpressionEvaluator : public ExpressionVisitor<TypedValue *> {
+ public:
+  ReferenceExpressionEvaluator(Frame *frame, const SymbolTable *symbol_table, const EvaluationContext *ctx,
+                               DbAccessor *dba, storage::View view)
+      : frame_(frame), symbol_table_(symbol_table), ctx_(ctx), dba_(dba), view_(view) {}
+
+  using ExpressionVisitor<TypedValue *>::Visit;
+
+  utils::MemoryResource *GetMemoryResource() const { return ctx_->memory; }
+
+#define UNSUCCESFUL_VISIT(expr_name) \
+  TypedValue *Visit(expr_name &expr) override { return nullptr; }
+
+  TypedValue *Visit(Identifier &ident) override { return &frame_->at(symbol_table_->at(ident)); }
+
+  UNSUCCESFUL_VISIT(NamedExpression);
+  UNSUCCESFUL_VISIT(OrOperator);
+  UNSUCCESFUL_VISIT(XorOperator);
+  UNSUCCESFUL_VISIT(AdditionOperator);
+  UNSUCCESFUL_VISIT(SubtractionOperator);
+  UNSUCCESFUL_VISIT(MultiplicationOperator);
+  UNSUCCESFUL_VISIT(DivisionOperator);
+  UNSUCCESFUL_VISIT(ModOperator);
+  UNSUCCESFUL_VISIT(NotEqualOperator);
+  UNSUCCESFUL_VISIT(EqualOperator);
+  UNSUCCESFUL_VISIT(LessOperator);
+  UNSUCCESFUL_VISIT(GreaterOperator);
+  UNSUCCESFUL_VISIT(LessEqualOperator);
+  UNSUCCESFUL_VISIT(GreaterEqualOperator);
+
+  UNSUCCESFUL_VISIT(NotOperator);
+  UNSUCCESFUL_VISIT(UnaryPlusOperator);
+  UNSUCCESFUL_VISIT(UnaryMinusOperator);
+
+  UNSUCCESFUL_VISIT(AndOperator);
+  UNSUCCESFUL_VISIT(IfOperator);
+  UNSUCCESFUL_VISIT(InListOperator);
+
+  UNSUCCESFUL_VISIT(SubscriptOperator);
+
+  UNSUCCESFUL_VISIT(ListSlicingOperator);
+  UNSUCCESFUL_VISIT(IsNullOperator);
+  UNSUCCESFUL_VISIT(PropertyLookup);
+  UNSUCCESFUL_VISIT(LabelsTest);
+
+  UNSUCCESFUL_VISIT(PrimitiveLiteral);
+  UNSUCCESFUL_VISIT(ListLiteral);
+  UNSUCCESFUL_VISIT(MapLiteral);
+  UNSUCCESFUL_VISIT(Aggregation);
+  UNSUCCESFUL_VISIT(Coalesce);
+  UNSUCCESFUL_VISIT(Function);
+  UNSUCCESFUL_VISIT(Reduce);
+  UNSUCCESFUL_VISIT(Extract);
+  UNSUCCESFUL_VISIT(All);
+  UNSUCCESFUL_VISIT(Single);
+  UNSUCCESFUL_VISIT(Any);
+  UNSUCCESFUL_VISIT(None);
+  UNSUCCESFUL_VISIT(ParameterLookup);
+  UNSUCCESFUL_VISIT(RegexMatch);
+
+ private:
+  Frame *frame_;
+  const SymbolTable *symbol_table_;
+  const EvaluationContext *ctx_;
+  DbAccessor *dba_;
+  // which switching approach should be used when evaluating
+  storage::View view_;
+};
 
 class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
  public:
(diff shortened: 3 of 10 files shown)


antoniofilipovic:
- [ ] Test with 100 000 lines with 50 props per node and without Lab, create 2 nodes and 1 edge per line
```
LOAD CSV FROM "/home/antonio/work/memgraph/github-repos/core/generated_text.csv" NO HEADER AS row
-> CREATE (a: Label {arg0:row[0],...,arg49:row[49]}),
-> (b: Label {arg50:row[50],...,arg99:row[99]}),
-> (a)-[:CONNECTED_TO]->(b);
```
Result:
dump-csv-100000-lines-node-props-50-string-15-edge-props-0-no-lab

antoniofilipovic:
- [ ] Test reading 100 000 lines with 50 props per node, create 2 nodes and 1 edge only for the last line:-> Empty set (14.435 sec)

dump-csv-100000-lines-node-props-50-string-15-edge-props-0-no-lab-last-line-only

antoniofilipovic:
Seems like problem is node creation, since reading through file is extremely fast, will test reading in Release mode, as all these two tests were done in Deubg mode

gitbuda:
> reading the whole file and creating 2 nodes and 1 edge only for the last line is really fast. Release mode: **(1.383 sec)**

Yes, reading is fast, the problem is Memgraph single thread speed -> 2 threads one for reading and one for execution (one more thread for reading) won't be very beneficial because parsing is ~2% of the execution?

antoniofilipovic:
> > reading the whole file and creating 2 nodes and 1 edge only for the last line is really fast. Release mode: **(1.383 sec)**
> 
> Yes, reading is fast, the problem is Memgraph single thread speed -> 2 threads one for reading and one for execution (one more thread for reading) won't be very beneficial because parsing is ~2% of the execution?

Yes, it seems like that. My best guess now is the problem is with properties and storing them.

antoniofilipovic:
- [ ] Test with 100 000 lines, 50 props per node without Lab, only create nodes and edges, no properties

16.9 seconds to create 200 000 nodes and 100 000 edges and read 100 000 lines but create 0 properties in Debug mode

Link [here](https://user-images.githubusercontent.com/61245998/215533727-89fccd3d-266d-41bd-8b73-1a9583c15fc6.svg)

antoniofilipovic:
Measuring time of function:
`Result<PropertyValue> VertexAccessor::SetProperty(PropertyId property, const PropertyValue &value)`

Out of 299 seconds to execute, 120 seconds are spent in above function:


```
Result<PropertyValue> VertexAccessor::SetProperty(PropertyId property, const PropertyValue &value) {
  auto start = std::chrono::steady_clock::now();

  utils::MemoryTracker::OutOfMemoryExceptionEnabler oom_exception;
  std::lock_guard<utils::SpinLock> guard(vertex_->lock);

  if (!PrepareForWrite(transaction_, vertex_)) return Error::SERIALIZATION_ERROR;

  if (vertex_->deleted) return Error::DELETED_OBJECT;

  auto current_value = vertex_->properties.GetProperty(property);
  CreateAndLinkDelta(transaction_, vertex_, Delta::SetPropertyTag(), property, current_value);
  vertex_->properties.SetProperty(property, value);

  UpdateOnSetProperty(indices_, property, value, vertex_, *transaction_);

  auto end = std::chrono::steady_clock::now();
  std::chrono::duration<double> dif = end-start;
  total+=dif;

  if (total>new_print){
    std::cout << "Time difference = " <<  total.count() << "[s]" << std::endl;
    new_print+=static_cast<std::chrono::duration<double>>(10.0);
  }

  return std::move(current_value);
}
```

antoniofilipovic:
Next step is to measure where is around 140 seconds spent if it is not in `SetProperty`

antoniofilipovic:
The next hypothesis is that there is additional time spent on `visitor`. To test that, I set a timer in function:
`TypedValue Visit(SubscriptOperator &list_indexing) override`

I run the following test:
- [ ]  Test on CSV file with 100 000 lines with 50 props per node and without Lab, create 2 nodes and 1 edge per line: (270 sec)

In this test, the time spent in the function `Visit(SubscriptOperator &list_indexing)` is 100 seconds.

Summary:

- out of 270 seconds in total, 100 is spent in `Visit(SubscriptOperator &list_indexing)` 
- out of 270 seconds in total, 100 is spent in `SetProperty` function
- out of 270 seconds, 240 seconds is by calling `PropsSetChecked` function

antoniofilipovic:
The next step is to confirm if this Visit is a real bottleneck and debug in `RelWithDebInfo` mode.

antoniofilipovic:
The latest hypothesis is that there is additional time spent on `visitor`. To test that, I set a timer in function:
`TypedValue Visit(SubscriptOperator &list_indexing) override ` and built Memgraph in `RelWithDebInfo`.

Testing hypothesis:
- [ ] Test on CSV file with 100 000 lines with 50 props per node and without Lab, create 2 nodes and 1 edge per line: (23.3 sec)

Out of 23.3 seconds in execution, 7 seconds were spent in `visitor`. That accounts for 30% of the time.

antoniofilipovic:
The next step is to implement reference-based expression eval to avoid copying of the list.

antoniofilipovic:
Another idea is to serialize also all properties at once, so we don't lose much time there. This is not the immediate next step, just one of the ideas to test afterward.

antoniofilipovic:
In the latest commit, I implemented `ReferenceExpressionEvaluator` to get from frame pointer to list object, without copying it.

I tested now hypothesis that this will speed up going through CSV:

- [ ] Test on CSV file with 100 000 lines with 50 props per node and without Lab, create 2 nodes and 1 edge per line: (11.7 sec)

Total time was reduced by **half**. It now "reads" 10k/s. Instead of 5k/s.

antoniofilipovic:
Next step is to profile again if it seems better. I profiled this time with `flamegraph` and here is the result:

flamegraph

It seems like `Vistor` is still very slow and the property is taking an equal amount of time. I will try to profile again with gperftools current state, and also profile with `flamegraph` before `ReferenceExpressionEvaluator` was introduced.

antoniofilipovic:
Hypothesis: lookup of parameters takes some time as it needs to iterate through std::vector each time to find a specific id.

- [ ] Test and profile introducing std::unordered_map instead of vector on  CSV file with 100 000 lines with 50 props per node and without Lab, create 2 nodes and 1 edge per line

flamegraph


Result: reduced time of import by **at least 1 second**. Now time is 10.5 seconds in most cases instead of 11.7 seconds.

antoniofilipovic:
The next step is to reduce time by setting properties as a batch instead of one by one.

antoniofilipovic:
One thing I tested.

What if two same properties appear inside `{}` like `arg0` in case below:

```
LOAD CSV FROM ".../generated_text.csv" NO HEADER AS row
CREATE (a: Label {arg0:row[0],arg1:row[1],arg0:row[2]}),
(b: Label {arg50:row[50],arg51:row[51],arg50:row[52]}),
(a)-[:CONNECTED_TO]->(b);
```

It throws an error:
`Same key can't appear twice in a map literal.`

This is why it  should be possible to optimize to batch process as we don't need to check for each property if it appeared already, only set new property.

On the other hand following situation is not possible currently to optimize:
```
LOAD CSV FROM "...generated_text.csv" NO HEADER AS row
CREATE (a: Label),
(b: Label),
(a)-[:CONNECTED_TO]->(b)
SET a.arg0=row[0], 
		a.arg1=row[1],
		a.arg0=row[2];
```
Because it generates multiple `SetProperty` cursors, and we can't batch process in that case.

antoniofilipovic:
The next thing to test was the batch insert of properties in the node. The hypothesis is that the insertion of properties takes a lot of time since for every property it needs to check whether that property already exists or not.

I implemented the `SetBatchProperties` method which accepts a map of PropertyValue and PropertyId and creates new properties.

- [ ] Test on CSV file with 100 000 lines with 50 props per node and without Lab, create 2 nodes and 1 edge per line: Execution time: 9.229 sec

With this implementation, I succeeded to reduce 1.5 more seconds in this test case.

 I profiled again with flame graph, and it showed that bottleneck is setting deltas:

flamegraph_setbatch

antoniofilipovic:
It seems that still setting properties is not optimized due to the fact that inside the `PropertyStore::SetProperty` function there is still a check to `FindSpecificPropertyAndBufferInfo` which takes a lot of time, according to `flamegraph`. In this case, when creating vertices we can have a guarantee that properties don't exist already and we could optimize that part not to search for properties in the property store.


Another thing that is interesting is that deltas are also taking 11% of the time. I made a small change to remove the setting of deltas just for the fun of testing out, and time was reduced even more.

antoniofilipovic:
As per the last post, I tested how it works without setting deltas.

- [ ] Test on CSV file with 100 000 lines with 50 props per node and without Lab, create 2 nodes and 1 edge per line: Execution time (8.363 sec)


From this test, we can see that it is faster without deltas, and one more thing that happens is that memory consumption was lower for 1.3 GB during import.


flamegraph

gitbuda:
"Turning deltas OFF" is tricky because of transactions, with a single `LOAD CSV` being trivial because there is only one transaction involved, but in general, it's more tricky. Maybe we can introduce something like the `IMPORT MODE ON|OFF;` query. I'm all for trying out, but that should be a separate PR 😄

antoniofilipovic:
> "Turning deltas OFF" is tricky because of transactions, with a single `LOAD CSV` being trivial because there is only one transaction involved, but in general, it's more tricky. Maybe we can introduce something like the `IMPORT MODE ON|OFF;` query. I'm all for trying out, but that should be a separate PR smile

I agree :smile: I will add that part in separate PR to test further memory implications and how can we speed up more data import.

antoniofilipovic:
Without deltas, it seems like we can get down more `import` time by not checking in `PropertyStore::SetProperty` function each time whether the property exists. One possibility was to not search for the property when we are sure it doesn't exist (Creating local vertex scenario), but due to the implementation detail of PropertyStore, we can't be sure whether metadata of `Type::Empty`  will or will not exist. A better solution would be to leave `SetProperty` as it is and implement `BatchSetProperty` in `PropertyStore`.

flamegraph

gitbuda:
I would just call the accessor method `SetProperties`. That method makes total sense to me. We can go even further and add everything (labels + props) in one go 😄

antoniofilipovic:
Implemented the `SetProperties` method in `PropertyStore` and it reduced time even more. Now for execution, it takes **5.5 seconds**.

- [ ] Test on CSV file with 100 000 lines with 50 props per node and without Lab, create 2 nodes and 1 edge per line: Execution time (5.5 sec)


Afterward, I profiled with `flamegraph` to test if we can go even faster. From the current output, seems like a reasonable amount of time it took in each of the methods.

flamegraph

antoniofilipovic:
I tested one hypothesis that creating of TypedValue in ExpressionEvaluator is taking some time, but when I returned a pointer to the `TypedValue` instead of the `TypedValue` object, I saw no gains at all. Another thing I still want to test is to use `std::vector` instead of std::map to process properties in batch.

antoniofilipovic:
Test using `std::vector` instead of `std::map` to process properties in batch.

- [ ] Test on CSV file with 100 000 lines with 50 props per node, create 2 nodes and 1 edge per line: Execution time (4.7 sec)


This made execution even faster, so we are now over **20K/s**.

flamegraph

From `flamegraph` it seems it can't be optimized much more.

antoniofilipovic:
Plan for the next few days:

- [ ] Test how we perform when index for label+property is created before insert and there is workload to create nodes, and in separate file workload to MATCH node and CREATE edge
- [ ] Test how we perform when index for label+property is created before insert and there is workload to create nodes from one file, and workload to MERGE nodes in other and CREATE edge
- [ ] Test how we perform when index for label+propery is created before insert, and there is workload to MERGE nodes and CREATE edges from same file (there can be duplicate nodes in the file with similar/same properties)

antoniofilipovic:
Today I benchmarked new changes compared to the old state. The test was the following:

- [ ] Import 1 000 000 lines, from each line create 2 nodes and 1 new edge. Each line will have X properties, of 10 string length.


| Number of properties per line | Import time old | Import time new |
| -- | -- |
|TODO | TODO | TODO|

antoniofilipovic:
From benchmarking today it seems that PropertyLookup visitor needs to be implemented in ReferenceBasedEvaluator as when the header in `LOAD CSV` is used, we don't gain an advantage from not copying the list.

antoniofilipovic:
> From benchmarking today it seems that PropertyLookup visitor needs to be implemented in ReferenceBasedEvaluator as when the header in `LOAD CSV` is used, we don't gain an advantage from not copying the list.
> 
> When not using a header it takes 2.5 seconds to import 200 000 lines with 20 features per line, each of 10 string length, and when using a header it takes 7.09 seconds.

With the new fix, with the header, it takes **1.59s** to load the same amount of data.

antoniofilipovic:
1. Import X nodes with Y properties nodes from 1 file and then Z edges with W properties from another file.

| X | Y | Z| W |Throughput time (OLD)  [node/seconds] | Thorughput time (NEW) nodes [seconds] | Throughput (OLD) [edges/s] | Throughput (NEW) [edges/s] |
| -- | -- | -- |  -- | -- |  -- | -- | -- |
|1M |10 | 5M| 5 |  117K/s (total 8.7seconds)  | TODO | 61.5 K/s (total: 81.3 seconds)  | TODO  |

Benchmarked again today, and it seems performance was much better as there was no need to activate the swap.

antoniofilipovic:
1. [flamegraph-nodes](https://user-images.githubusercontent.com/61245998/220083026-13107a6a-042d-4bb2-9098-2b638cb55079.svg)
2. [flamegraph-edges-index](https://user-images.githubusercontent.com/61245998/220083039-db3f7150-ee19-4b19-8c29-d15f4fecbdd2.svg)

I profiled the creation of 1M nodes and with this profiling we are familiar. From this `flamegraph,` we can see that most of the time goes on setting properties and getting values from rows. 

The second flamegraph is new and from there it seems that most of the time on edge creating goes on finding nodes. Setting properties on edges seems to take also time, and so does the creating edge.

antoniofilipovic:
The next step is to benchmark how fast would edge creation be if we use `ToInteger` function and index on a number instead of the string property of the node.

antoniofilipovic:
Now I created 1M nodes where each node has integer as id. After that I created 5M edges with similar query to above to test how fast it is to match two node when there is index on integer property.

Here are the queries used:
- node creation:
 ```
LOAD CSV FROM ".../generated_text_nodes_ints.csv" NO HEADER AS row
CREATE (:Label {arg0:toInteger(row[0]),arg1:row[1],arg2:row[2],arg3:row[3],arg4:row[4],arg5:row[5],arg6:row[6],arg7:row[7],arg8:row[8],arg9:row[9],arg10:row[10]});
```
- edge creation:
```
LOAD CSV FROM ".../generated_text_edges_ints.csv" NO HEADER AS row
MATCH (a:Label {arg0:toInteger(row[0])}), (b:Label {arg0:toInteger(row[1])}) CREATE (a)-[:EDGE_TYPE {arg2:row[2],arg3:row[3],arg4:row[4],arg5:row[5],arg6:row[6]}]->(b);
```

Here are 2 `flamegraphs` created for this scenario:
1. [flamegraph-nodes](https://user-images.githubusercontent.com/61245998/220128853-60616019-116d-4c44-974e-aa11a2ae3abb.svg) - this is flamegraph for creating nodes without two PRs to enhance the performance of node creation ([Visitor](https://github.com/memgraph/memgraph/pull/774) and [Property store](https://github.com/memgraph/memgraph/pull/788)).
2. [flamegraph-edges-index]() - this flamegraph is for MATCH-ing two nodes and creating edge between.

antoniofilipovic:
Now I benchmarked the same setup - creating nodes from one file and edges from another file. But for the index, I used string prefixed index. This means that every index had the same prefix, in this case, 10 characters "A": "AAAAAAAAAA<different_str>".

Here are the results:

antoniofilipovic:
- [ ] Profile on Memgraph docker speed also. Check if it is aligned with native Memgraph

antoniofilipovic:
- [ ] Profile Memgraph on a high ratio of edges/nodes (dense graph)

Memgraph seems to perform poorly when there is high ratio of edges/nodes.  I have made following test cases to inspect Memgraph's performance:

- [ ] Create 35 000 nodes, create index on nodes and then create 2 500 000 edges between  these nodes
- [ ] Create 350 000 nodes, create index on nodes and create 2 500 000 edges between these nodes


In the first use case, ratio edges/nodes was ~70. On second use case, ratio is 7.

It took Memgraph 95 seconds to import edges in first case, and in second 40 seconds. Which would mean that throughput was around 25K/s on first use case and around 60K/s on second one although there were less nodes.

For the second use case here is [flamegraph](https://user-images.githubusercontent.com/61245998/220680512-52cf9d28-1efc-4bd5-9a64-be9b54562c26.svg)

Most time went on `AdvanceUntilValid` function.

antoniofilipovic:
After a lot of testing, the problem seems to be due to deltas. One node could have 50000 deltas, and there can be a check on them all. Disabling deltas improved performance. On 3500 nodes and 50 000 edges on 1 node, it took 0.5 seconds to import edges without deltas, and 15 seconds when deltas were on.

antoniofilipovic:
Deltas proved to be a problem when there are a lot of nodes. After setting the transaction isolation level to session, problem went away. 

The test was the following: import 3500 nodes, and afterward create 50 000 edges where all edges go from one node, to random other nodes. With the `SNAPSHOT` isolation level, the import speed was around **15 seconds**. With isolation level set to `READ UNCOMMITTED`, import time was **0.5 seconds** for 50000 edges, meaning throughput went above **100K**.

antoniofilipovic:
Another hypothesis is that speed is largely determined by the number of nodes in the graph. I tested that hypothesis, and after importing **3.5M nodes**, it took 1.1 seconds for **50K edges**, meaning throughput was around **50K e/s**.

antoniofilipovic:
@gitbuda Yes, I plan to close this PR now. I think that is it.

PR #755, opened 2023-01-31
[master < T582-FL] Fix label-based auth using OLD view instead of NEW when merging nodes

Josipmrden:
```
UNWIND 
[{id: '1', lat: 10, lng: 10}, {id: '2', lat: 10, lng: 10}, {id: '3', lat: 10, lng: 10}] AS row
MERGE (o:Location {id: row.id}) RETURN o;
```
did not work and threw an exception because newly created nodes were not seen and therefore prohibited.

**Changelog message**: Label based auth improved to see information about newly created nodes when merging nodes inside a query.

[master < Task] PR

### DIFF
--- a/src/query/plan/operator.cpp
+++ b/src/query/plan/operator.cpp
@@ -410,10 +410,11 @@ VertexAccessor &CreateExpand::CreateExpandCursor::OtherVertex(Frame &frame, Exec
 template <class TVerticesFun>
 class ScanAllCursor : public Cursor {
  public:
-  explicit ScanAllCursor(Symbol output_symbol, UniqueCursorPtr input_cursor, TVerticesFun get_vertices,
-                         const char *op_name)
+  explicit ScanAllCursor(Symbol output_symbol, UniqueCursorPtr input_cursor, storage::View view,
+                         TVerticesFun get_vertices, const char *op_name)
       : output_symbol_(output_symbol),
         input_cursor_(std::move(input_cursor)),
+        view_(view),
         get_vertices_(std::move(get_vertices)),
         op_name_(op_name) {}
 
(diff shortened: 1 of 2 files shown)


src/query/plan/operator.cpp
```diff
@@ -448,7 +448,7 @@ class ScanAllCursor : public Cursor {
 #ifdef MG_ENTERPRISE
   bool FindNextVertex(const ExecutionContext &context) {
     while (vertices_it_.value() != vertices_.value().end()) {
-      if (context.auth_checker->Has(*vertices_it_.value(), memgraph::storage::View::OLD,
+      if (context.auth_checker->Has(*vertices_it_.value(), memgraph::storage::View::NEW,
```
antaljanosbenjamin, line 451: I think the proper solution here is to push down the value of `ScanAll::view_` into the `ScanAll` cursor, otherwise we can run into problems with the view. E.g.: the current solution might issues with `MATCH (n), (m) DELETE n`, because the second `ScanAll` might access a node that is deleted by in the same query, thus getting its labels with `View::NEW` will result in a `DELETED_OBJECT` error. I am not sure that this query will trigger, but I am almost sure that something can trigger this error if we are not careful with it.

antaljanosbenjamin (approved):

src/query/plan/operator.cpp
antaljanosbenjamin, line 451: I think the proper solution here is to push down the value of `ScanAll::view_` into the `ScanAll` cursor, otherwise we can run into problems with the view. E.g.: the current solution might issues with `MATCH (n), (m) DELETE n`, because the second `ScanAll` might access a node that is deleted by in the same query, thus getting its labels with `View::NEW` will result in a `DELETED_OBJECT` error. I am not sure that this query will trigger, but I am almost sure that something can trigger this error if we are not careful with it.

antaljanosbenjamin (changes requested):

antaljanosbenjamin (approved):

antoniofilipovic (approved):

vpavicic:
RN: Newly created nodes now comply to the previously set label-based authorization rules.

PR #757, opened 2023-01-31
[master < E056-MAGE] Mock graph APIs

antepusic:
[master < E056-MAGE] Mock graph APIs
- [x] Write E2E tests


include/_mgp_mock.py
```diff
@@ -0,0 +1,183 @@
...
+class ImmutableObjectError(Exception):
+    pass
+
+
+class LogicErrorError(Exception):
+    pass
+
+
+class Vertex:
+    __slots__ = ("_id", "_graph")
```
as51340, line 15: Nice optimization!

include/_mgp_mock.py
```diff
@@ -0,0 +1,183 @@
...
+    def id(self) -> int:
+        return self._id
+
+    @property
+    def graph(self) -> nx.MultiDiGraph:
+        return self._graph
+
+    @property
+    def labels(self) -> typing.List[int]:
+        return self._graph.nodes[self._id]["label"].split(":")
```
as51340, line 40: Can we make "label" a constant? This is because nx doesn't have its own label store right?

include/_mgp_mock.py
```diff
@@ -0,0 +1,183 @@
...
+        if nx.is_frozen(self._graph):
+            raise ImmutableObjectError("Cannot modify immutable object.")
+
+        labels = self._graph.nodes[self._id]["label"]
+        if labels.startswith(f"{label}:"):
+            labels = "\n" + labels  # pseudo-string starter
+            self._graph.nodes[self._id]["label"] = labels.replace(f"\n{label}:", "")
+        elif labels.endswith(f":{label}"):
+            labels += "\n"  # pseudo-string terminator
+            self._graph.nodes[self._id]["label"] = labels.replace(f":{label}\n", "")
```
as51340, line 61:
```suggestion
        elif labels.endswith(f":{label}"):
            self._graph.nodes[self._id]["label"] = labels.replace(f":{label}", "")
```

as51340, line 61: Why do you need pseudo-string starter and pseudo-string terminator?

include/_mgp_mock.py
```diff
@@ -0,0 +1,183 @@
...
+
+    @property
+    def end_id(self) -> int:
+        return self._edge[Edge.I_END]
+
+    def is_valid(self) -> bool:
+        return self._graph.has_edge(*self._edge)
+
+    def get_type_name(self):
+        return self._graph.get_edge_data(*self._edge)["type"]
```
as51340, line 110: I would make "type" a constant

include/_mgp_mock.py
```diff
@@ -0,0 +1,183 @@
...
+        return Edge(self._edges[index], self._graph)
+
+    def size(self) -> int:
+        return len(self._edges)
+
+
+class Graph:
+    __slots__ = ("nx", "_highest_edge_id")
+
+    I_KEY = 2
```
as51340, line 162: Extract this constant since it has the same value in class Edge and in class Graph.

tests/e2e/mock_api/procedures/edge.py
```diff
@@ -0,0 +1,85 @@
...
+    results["__eq__"] = test_utils.all_equal(
+        target_edge_1 == target_edge_1,
+        target_mock_edge_1 == target_mock_edge_1,
+        True,
+    ) and test_utils.all_equal(
+        target_edge_1 != target_edge_1,
+        target_mock_edge_1 != target_mock_edge_1,
+        False,
+    )
+
```
as51340, line 74: is this a tipfeller that is being tested that a == a?

tests/e2e/mock_api/procedures/vertex.py
```diff
@@ -0,0 +1,101 @@
...
+
+    ID_2 = 2
+
+    target_vertex_2 = test_utils.get_vertex(ctx, permanent_id=ID_2)
+    target_mock_vertex_2 = mock_ctx.graph.get_vertex_by_id(ID_2)
+
+    results["__eq__"] = test_utils.all_equal(
+        target_vertex == target_vertex,
+        target_mock_vertex == target_mock_vertex,
+        True,
```
as51340, line 84: same as above

as51340 (changes requested):
nicely written, left some minor comments and questions

tests/e2e/mock_api/procedures/properties.py
```diff
@@ -0,0 +1,184 @@
...
+    results = dict()
+
+    TARGET_ID = 0
+    target_vertex = test_utils.get_vertex(ctx, permanent_id=TARGET_ID)
+    target_mock_vertex = mock_ctx.graph.get_vertex_by_id(TARGET_ID)
+
+    properties = target_vertex.properties
+    mock_properties = target_mock_vertex.properties
+
+    results["get"] = test_utils.all_equal(
```
Josipmrden, line 17: can you extract these above as constants

tests/e2e/mock_api/procedures/record.py
```diff
@@ -0,0 +1,18 @@
+import mgp
+import mgp_mock
+import test_utils
+
+
+@mgp.read_proc
+def compare_apis(ctx: mgp.ProcCtx) -> mgp.Record(results_dict=mgp.Map):
+    results = dict()
+
+    record = mgp.Record(a=1, b=2.0, c="3")
```
Josipmrden, line 10: bool?

Josipmrden: what about date types

Josipmrden (commented):
Some comments

include/mgp_mock.py
```diff
@@ -0,0 +1,1011 @@
...
+import sys
+import typing
+from collections import namedtuple
+from functools import wraps
+
+import _mgp_mock
+import kafka
+import networkx as nx
+import pulsar
+
```
antoniofilipovic, line 12: Add some high level docs similar to what is in mgp.py.

include/mgp_mock.py
```diff
@@ -0,0 +1,1011 @@
+import datetime
+import inspect
+import sys
+import typing
+from collections import namedtuple
+from functools import wraps
```
antoniofilipovic, line 6: apply also here comment about import from _mgp_mock.py

include/mgp_mock.py
```diff
@@ -0,0 +1,1011 @@
...
+
+class ValueConversionError(Exception):
+    """
+    Signals that conversion between python and cypher values failed.
+    """
+
+    pass
+
+
+class Label:
```
antoniofilipovic, line 96: Provide here docs, similar to mgp.py

include/mgp_mock.py
```diff
@@ -0,0 +1,1011 @@
...
+            raise InvalidContextError()
+
+        try:
+            return self[property_name]
+        except KeyError:
+            return default
+
+    def set(self, property_name: str, value: object) -> None:
+        if nx.is_frozen(self._vertex_or_edge.graph):
+            raise ImmutableObjectError("Cannot modify immutable object.")
```
antoniofilipovic, line 150: Can you push this part to `_mgp_mock.Vertex` so you have following functionality:
```
if self._vertex_or_edge.is_frozen():
    raise ... 
```

I think `nx` should be hidden in `_mgp_mock` as it is encapsulation for that part, right? Does that make sense?

include/mgp_mock.py
```diff
@@ -0,0 +1,1011 @@
...
+        )
+
+        return vertex_or_edge_props[property_name]
+
+    def __setitem__(self, property_name: str, value: object) -> None:
+        if not self._vertex_or_edge.is_valid():
+            raise InvalidContextError()
+
+        if nx.is_frozen(self._vertex_or_edge.graph):
+            raise ImmutableObjectError("Cannot modify immutable object.")
```
antoniofilipovic, line 217: Also, try to push this usage to _mgp_mock.py, applies to rest of nx usage

include/mgp_mock.py
```diff
@@ -0,0 +1,1011 @@
...
+        if not self._vertex_or_edge.is_valid():
+            raise InvalidContextError()
+
+        vertex_or_edge_props = (
+            self._vertex_or_edge.graph.nodes[self._vertex_or_edge.id]
+            if isinstance(self._vertex_or_edge, _mgp_mock.Vertex)
+            else self._vertex_or_edge.graph.edges[self._vertex_or_edge.edge]
+        )
+
+        return vertex_or_edge_props[property_name]
```
antoniofilipovic, line 210: What if there is no property here with that name, should we raise KeyError?

include/mgp_mock.py
```diff
@@ -0,0 +1,1011 @@
...
+            yield item.name
+
+    def __getitem__(self, property_name: str) -> object:
+        if not self._vertex_or_edge.is_valid():
+            raise InvalidContextError()
+
+        vertex_or_edge_props = (
+            self._vertex_or_edge.graph.nodes[self._vertex_or_edge.id]
+            if isinstance(self._vertex_or_edge, _mgp_mock.Vertex)
+            else self._vertex_or_edge.graph.edges[self._vertex_or_edge.edge]
```
antoniofilipovic, line 207: Can you implement here self._vertex_or_edge.get_property(property) like in mgp.py?

include/mgp_mock.py
```diff
@@ -0,0 +1,1011 @@
...
+        if not isinstance(edge, _mgp_mock.Edge):
+            raise TypeError(f"Expected '_mgp_mock.Edge', got '{type(edge)}'")
+
+        self._edge = edge
+
+    def __deepcopy__(self, memo):
+        # This is the same as the shallow copy, because we want to share the
+        # underlying C struct. Besides, it doesn't make much sense to actually
+        # copy _mgp_mock.Edge as that is actually a reference to a graph element
+        # and not a proper value.
```
antoniofilipovic, line 277: Is this comment here necessary, we are using mock here?

include/mgp_mock.py
```diff
@@ -0,0 +1,1011 @@
...
+
+        return Properties(self._vertex)
+
+    @property
+    def in_edges(self) -> typing.Iterable[Edge]:
+        if not self.is_valid():
+            raise InvalidContextError()
+
+        for edge in self._vertex.graph.in_edges(self.id, keys=True):
+            yield Edge(_mgp_mock.Edge(edge, self._vertex.graph))
```
antoniofilipovic, line 407: Instead of referencing graph, can you implement here self._vertex.in_edges() method?

include/mgp_mock.py
```diff
@@ -0,0 +1,1011 @@
...
+
+    @property
+    def out_edges(self) -> typing.Iterable[Edge]:
+        if not self.is_valid():
+            raise InvalidContextError()
+
+        for edge in self._vertex.graph.out_edges(self.id, keys=True):
+            yield Edge(_mgp_mock.Edge(edge, self._vertex.graph))
```
antoniofilipovic, line 415: same here for out edges

include/mgp_mock.py
```diff
@@ -0,0 +1,1011 @@
...
+        return Vertices(self._graph)
+
+    def is_valid(self) -> bool:
+        return True
+
+    def __iter__(self) -> typing.Iterable[Vertex]:
+        if not self.is_valid():
+            raise InvalidContextError()
+
+        for id in self._graph.nodes:
```
antoniofilipovic, line 538: can we also call self._graph.nodes()?

antoniofilipovic: And that it yields _mgp_mock.Vertex?

include/mgp_mock.py
```diff
@@ -0,0 +1,1011 @@
...
+
+class Vertices:
+    __slots__ = ("_graph", "_len")
+
+    def __init__(self, graph):
+        if not isinstance(graph, nx.MultiDiGraph):
+            raise TypeError(f"Expected 'networkx.classes.multidigraph.MultiDiGraph', got '{type(graph)}'")
+
+        self._graph = graph
+        self._len = None
```
antoniofilipovic, line 523: Why does Vertices not accept _mgp_mock.Graph? I think there is no need for nx.MultiDiGraph.

include/mgp_mock.py
```diff
@@ -0,0 +1,1011 @@
...
+            raise InvalidContextError()
+
+        return Vertex(_mgp_mock.Vertex(vertex_id, self._graph.nx))
+
+    @property
+    def vertices(self) -> Vertices:
+        if not self.is_valid():
+            raise InvalidContextError()
+
+        return Vertices(self._graph.nx)
```
antoniofilipovic, line 587: You can pass here only self._graph

include/mgp_mock.py
```diff
@@ -0,0 +1,1011 @@
...
+        if not self.is_valid():
+            raise InvalidContextError()
+
+        if nx.is_frozen(self._graph.nx):
+            raise ImmutableObjectError("Cannot modify immutable object.")
+
+        if not nx.is_isolate(self._graph.nx, vertex.id):
+            raise LogicErrorError("Logic error.")
+
+        # TODO find out whether users can ever send this method a nonexistent vertex
```
antoniofilipovic, line 617: For when is this todo?

include/mgp_mock.py
```diff
@@ -0,0 +1,1011 @@
...
+    module = sys.modules[__name__]
+    for name, obj in inspect.getmembers(module):
+        if not defined_in_this_module(obj):
+            continue
+        if inspect.isclass(obj):
+            wrap_member_functions(obj)
+        elif inspect.isfunction(obj) and not name.startswith("_"):
+            setattr(module, name, wrap_function(obj))
+
+
```
antoniofilipovic, line 1010: I think Logger class is missing, we can also add that and provide api to logging

antoniofilipovic (changes requested):
Good work. I have made few comments to make code more consistent with mgp

include/_mgp_mock.py
```diff
@@ -0,0 +1,183 @@
...
+
+class ImmutableObjectError(Exception):
+    pass
+
+
+class LogicErrorError(Exception):
+    pass
+
+
+class Vertex:
```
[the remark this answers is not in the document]
antepusic: Absolutely! The documentation is the same as with mgp.py (excepting DB-specific errors) and I’ll add it to this file.

include/_mgp_mock.py
antepusic: That’s right, networkx does not have its own label store. I’ll rename the field.

include/_mgp_mock.py
```diff
@@ -0,0 +1,183 @@
...
+        return self._graph.nodes[self._id]["label"].split(":")
+
+    def is_valid(self) -> bool:
+        return self._graph.has_node(self._id)
+
+    def add_label(self, label: str) -> None:
+        if nx.is_frozen(self._graph):
+            raise ImmutableObjectError("Cannot modify immutable object.")
+
+        self._graph.nodes[self._id]["label"] += f":{label}"
```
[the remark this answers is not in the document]
antepusic: @antoniofilipovic The set of a node’s labels is a vector in storage ([link](https://github.com/memgraph/memgraph/blob/f1fe77adfbd6d647047f2e1774b809025b0db5d3/src/storage/v2/vertex_accessor.cpp#L161-L208)). As the mock API aims to be consistent with Memgraph, the labels have to be stored in an order-preserving structure.

Whereas storing labels in a set would make checking more computationally efficient, it’s not like graph datasets have nodes with e.g. thousands of labels.

include/_mgp_mock.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def remove_label(self, label: str) -> None:
+        if nx.is_frozen(self._graph):
+            raise ImmutableObjectError("Cannot modify immutable object.")
```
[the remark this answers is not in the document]
antepusic: Initially I had a similar descriptive message here, but that wasn’t consistent with the [error messages](https://github.com/memgraph/memgraph/blob/898c894a482f4090e9303fc9a39ae58482d2269f/src/query/procedure/py_module.cpp#L56-L110) used by the Python API.

include/_mgp_mock.py
antepusic: @antoniofilipovic I’m using the same Memgraph-to-nx graph “translation” as @as51340 in his [PR](https://github.com/memgraph/gqlalchemy/pull/215/files#diff-c88326fe380799b9832efa8b85885c70032853e96e73ab709ec8b84267519daf) on gqlalchemy, which stores labels as a string like this: `"Label1:Label2:Label3"`.
@Josipmrden There are three cases here wrt removal:
* first label: remove `{label}:`
* last label: remove `:{label}`
* default case: replace `:{label}:` with `:`.

@as51340 The pseudo-starter and -terminator cover the edge case where the first/last label is a substring of another label. For example, if the labels are `"Graph:DiGraph:Label3"` and I want to delete `Graph`, blindly removing `Graph:` would change the labels to `"Di:Label3"`.

include/_mgp_mock.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def edge_at(self, index: int) -> Edge:
+        return Edge(self._edges[index], self._graph)
+
+    def size(self) -> int:
+        return len(self._edges)
+
+
+class Graph:
+    __slots__ = ("nx", "_highest_edge_id")
```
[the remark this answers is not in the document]
antepusic: It’s not intended to be a “private” attribute, hence no `_` prefix.

include/_mgp_mock.py
```diff
@@ -0,0 +1,183 @@
...
+            raise TypeError(f"Expected 'networkx.classes.multidigraph.MultiDiGraph', got '{type(graph)}'")
+
+        self.nx = graph
+        self._highest_edge_id = None
+
+    def _new_edge_id(self):
+        if self._highest_edge_id is None:
+            self._highest_edge_id = max(edge[Graph.I_KEY] for edge in self.nx.edges(keys=True))
+
+        return self._highest_edge_id + 1
```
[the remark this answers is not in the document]
antepusic: It is optimized - note that `_highest_edge_id` is only computed by iteration if its initial value is `None`.

tests/e2e/mock_api/procedures/edge.py
antepusic: That part tests if `__eq__` works properly.

tests/e2e/mock_api/procedures/edge.py
```diff
@@ -0,0 +1,85 @@
...
+    TARGET_EDGE_1_ID = 9
+    TARGET_EDGE_2_ID = 37
+
+    target_edge_1 = test_utils.get_edge(ctx, permanent_id=TARGET_EDGE_1_ID)
+    target_edge_2 = test_utils.get_edge(ctx, permanent_id=TARGET_EDGE_2_ID)
+    target_mock_edge_1 = test_utils.get_mock_edge(mock_ctx, id=TARGET_EDGE_1_ID)
+    target_mock_edge_2 = test_utils.get_mock_edge(mock_ctx, id=TARGET_EDGE_2_ID)
+
+    results["is_valid"] = test_utils.all_equal(
```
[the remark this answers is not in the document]
antepusic: At some places in the tests I have multiple actual values that I want to compare, so I used the `all_equal` itertools recipe as [described](https://docs.python.org/3/library/itertools.html#itertools-recipes) in the Python reference.

tests/e2e/mock_api/procedures/label.py
```diff
@@ -0,0 +1,52 @@
...
+        label_1 != "Startup",
+        mock_label_1 != mock_label_2,
+        mock_label_1 != "Startup",
+        True,
+    ) and test_utils.all_equal(
+        label_1 != label_1,
+        label_1 != "Company",
+        mock_label_1 != mock_label_1,
+        mock_label_1 != "Company",
+        False,
```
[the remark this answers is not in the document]
antepusic: Answered above

tests/e2e/mock_api/procedures/record.py
antepusic: The point of this test is not to verify that the types work (all of them are either Python standard library or Python API, not the mock API), but to verify that the record created by the mock has the `fields` attribute.

tests/e2e/mock_api/procedures/vertex.py
antepusic: Answered above

include/_mgp_mock.py
as51340: makes sense

as51340 (approved):

include/_mgp_mock.py
as51340: Just put "label" as a constant.

as51340 (approved):
Just change "label" to constant and you have approve

include/mgp_mock.py
antepusic: Good idea! I initially created `_mgp_mock` only so that `mgp_mock` classes have the exact same methods as `mgp`, but this makes more sense from an engineering perspective.

include/mgp_mock.py
antepusic: In this case, Python automatically raises a KeyError. I’m currently checking whether this is consistent with what the Python API would do

include/mgp_mock.py
antepusic: I don’t see why this needs a dedicated method in `_mgp_mock`. The behavior of this `mgp_mock` method is already consistent with `mgp` and the hidden part (`_mgp_mock`) doesn’t factor into that.

include/mgp_mock.py
antepusic: Well, if the Python API implements shallow copying in place of deep copying, I’ll add the same so that users don’t experience unexpected behavior upon switching from the mock to the real API.

include/mgp_mock.py
antepusic: Again, this `mgp_mock` method that doesn’t use `_mgp_mock` behaves consistently with the analogous `mgp` method that uses `_mgp`. Which this in mind, I don’t see why there needs to be a dedicated method in `_mgp_mock`.

include/mgp_mock.py
antepusic: Will do

include/mgp_mock.py
antepusic: Fixed in the meantime

include/mgp_mock.py
antepusic: I’ll remove this by the next review round

include/mgp_mock.py
antepusic: Assuming that the Logger is a Memgraph-specific logger, why should it feature in the mock API? The mock API is supposed to run independently of Memgraph.

include/mgp_mock.py
antepusic: Changed in the meantime.

include/mgp_mock.py
antepusic: The same answer goes as with in_edges.

include/_mgp_mock.py
```diff
@@ -0,0 +1,183 @@
+import typing
+
+import networkx as nx
```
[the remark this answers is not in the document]
antepusic: Third-party library imports fixed

include/mgp_mock.py
antepusic: Added

include/mgp_mock.py
antepusic: Done

include/mgp_mock.py
antepusic: Done

include/mgp_mock.py
antepusic: Done

include/mgp_mock.py
```diff
@@ -0,0 +1,1916 @@
...
+        Examples:
+            ```vertex.add_label(label)```
+        """
+        if not self.is_valid():
+            raise InvalidContextError()
+
+        if self._vertex.is_deleted():
+            raise DeletedObjectError("Accessing deleted object.")
+
+        return self._vertex.add_label(label)
```
as51340, line 736: Why return when underlying _mgp_mock doesn't return anything?

include/mgp_mock.py
```diff
@@ -0,0 +1,1916 @@
...
+        Examples:
+            ```vertex.remove_label(label)```
+        """
+        if not self.is_valid():
+            raise InvalidContextError()
+
+        if self._vertex.is_deleted():
+            raise DeletedObjectError("Accessing deleted object.")
+
+        return self._vertex.remove_label(label)
```
as51340, line 759: same as above

include/mgp_mock.py
```diff
@@ -0,0 +1,1916 @@
...
+        Examples:
+            ```graph.detach_delete_vertex(vertex)```
+        """
+        if not self.is_valid():
+            raise InvalidContextError()
+
+        if self._graph.is_immutable():
+            raise ImmutableObjectError("Cannot modify immutable object.")
+
+        self._graph.delete_vertex(vertex.id)
```
as51340, line 1253: How is this method then different than delete_vertex when edges aren't touched?

include/mgp_mock.py
```diff
@@ -0,0 +1,1916 @@
...
+    # TODO try to implement
+    # def must_abort(self) -> bool:
+    #     if not self.is_valid():
+    #         raise InvalidContextError()
+
+    #     return self._graph._graph.nx.must_abort()
+
+    # def check_must_abort(self):
+    #     if self.must_abort():
+    #         raise AbortError
```
as51340, line 1369: just comment that you are aware of this code

include/mgp_mock.py
antepusic: The `mgp_mock.add_label` and `_mgp_mock.add_label` methods both return `None`, so there’s no harm in doing this.
In fact, the Python API does the same thing in `mgp.py`.

include/mgp_mock.py
antepusic: The `mgp_mock.remove_label` and `_mgp_mock.remove_label` methods both return `None`, so there’s no harm in doing this.
In fact, the Python API does the same thing in `mgp.py`.

include/mgp_mock.py
```diff
@@ -0,0 +1,1916 @@
...
+        return self._graph
+
+
+def transformation(func: typing.Callable[..., Record]):
+    raise_if_does_not_meet_requirements(func)
+
+    sig = inspect.signature(func)
+
+    params = tuple(sig.parameters.values())
+    if not params or not params[0].annotation is Messages:
```
as51340, line 1751:
```suggestion
    if not params or not isinstance(params[0].annotation, Messages):
```

include/mgp_mock.py
```diff
@@ -0,0 +1,1916 @@
...
+        if not len(params) == 2 or not params[1].annotation is Messages:
```
as51340, line 1752: isinstance I think is the preferred way again

include/mgp_mock.py
antepusic: Both `delete_vertex` and this method call `_mgp_mock.delete_vertex`, which in turn deletes a vertex from the NetworkX graph together with all its adjacent edges if they exist.
The only difference between this method and `delete_vertex` is that `delete_vertex` only does that operation if the edge is isolated/has no adjacent edges.

as51340 (changes requested):
Check few comments and it is approve

include/mgp_mock.py
antepusic: Aware :raised_back_of_hand:

include/mgp_mock.py
as51340: yes, yes just a matter of syntax, I agree. Then decision is up to you :)

include/mgp_mock.py
antepusic: Done

as51340 (approved):

tests/e2e/mock_api/procedures/test_utils/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+# copy_mock_python_api_e2e_files(test_utils.py)
```
Josipmrden, line 1: do you delete this? (auto resolve yourself whatever you do)

Josipmrden (approved):
Approve + 1 comment with your resolving power, I don't need

include/_mgp_mock.py
antoniofilipovic: @antepusic I would still use here set, as it is easier to maintain code

include/_mgp_mock.py
```diff
@@ -0,0 +1,453 @@
...
+        labels = self._graph.nx.nodes[self._id][NX_LABEL_ATTR]
+        if labels.startswith(f"{label}:"):
+            labels = "\n" + labels  # pseudo-string starter
+            self._graph.nx.nodes[self._id][NX_LABEL_ATTR] = labels.replace(f"\n{label}:", "")
+        elif labels.endswith(f":{label}"):
+            labels += "\n"  # pseudo-string terminator
+            self._graph.nx.nodes[self._id][NX_LABEL_ATTR] = labels.replace(f":{label}\n", "")
+        else:
+            self._graph.nx.nodes[self._id][NX_LABEL_ATTR] = labels.replace(f":{label}:", ":")
+
```
antoniofilipovic, line 196: I would still use here `set()` to store labels, as it is O(1) and much easier to read and maintain code.

include/mgp_mock.py
antoniofilipovic: Since this is how it is in mgp API, we can continue this way here.

include/mgp.py
```diff
@@ -1764,8 +1785,10 @@ def timestamp(self) -> int:
     def offset(self) -> int:
         """
         Supported stream sources:
-          - Kafka
+          * Kafka
 
+        Raises:
+            InvalidMessageError: If the message is not in a valid context.
         Raise InvalidArgumentError if the message is from an unsupported stream source.
```
antoniofilipovic, line 1792: Should this be differently formatted, like:
Raises:
```suggestion
        InvalidArgumentError: if the message is from an unsupported stream source.
```

include/_mgp_mock.py
```diff
@@ -0,0 +1,453 @@
...
+
+class Edge:
+    """Represents a graph edge."""
+
+    __slots__ = ("_edge", "_graph")
+
+    I_START = 0
+    I_END = 1
+    I_KEY = 2
```
antoniofilipovic, line 231: This you could isolate inside `class Constants`

tests/e2e/mock_api/procedures/test_utils/CMakeLists.txt
antoniofilipovic, line 1: Should this be commented?

tests/e2e/mock_api/procedures/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+copy_mock_python_api_e2e_files(edge.py)
+copy_mock_python_api_e2e_files(graph.py)
+copy_mock_python_api_e2e_files(label.py)
+copy_mock_python_api_e2e_files(path.py)
+copy_mock_python_api_e2e_files(properties.py)
+copy_mock_python_api_e2e_files(record.py)
+copy_mock_python_api_e2e_files(vertex.py)
+copy_mock_python_api_e2e_files(vertices.py)
+
+# add_subdirectory(test_utils)
```
antoniofilipovic, line 12: Why is this commented

include/mgp_mock.py
antoniofilipovic: I was thinking to have idea how it would work with logging, but we don't have to add that.

include/_mgp_mock.py
antoniofilipovic: This was changed so I will resolve this

include/mgp_mock.py
antoniofilipovic: @antepusic should we remove this?

vpavicic:
What's the story here? What is the release note here? I see @antoniofilipovic proposing additional documentation... when and who will write it? :)

antoniofilipovic:
> What's the story here? What is the release note here? I see @antoniofilipovic proposing additional documentation... when and who will write it? :)

@vpavicic story is that Ante implemented mock API with which you don't need Memgraph to test how your algorithm will work. I think @antepusic will give more details, you can check with him about docs and next steps.

include/mgp.py
antepusic: Fixed

include/mgp_mock.py
```diff
@@ -0,0 +1,1916 @@
...
+    """
+    Signals faulty logic within the program such as violating logical
+    preconditions or class invariants and may be preventable.
+    """
+
+    pass
+
+
+# DONE check how to trigger and if applicable elsewhere
+# DONE check if vertex/edge in graph; if not in graph & id <= highest, then deleted
```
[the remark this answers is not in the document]
antepusic: Done

include/_mgp_mock.py
antepusic: Done

include/_mgp_mock.py
```diff
@@ -0,0 +1,453 @@
...
+class DeletedObjectError(Exception):
+    pass
+
+
+class Graph:
+    """Wrapper around a NetworkX MultiDiGraph instance."""
+
+    __slots__ = ("nx", "_highest_vertex_id", "_highest_edge_id", "_valid")
+
+    I_KEY = 2
```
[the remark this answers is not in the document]
antepusic: It does, done

antepusic:
> But to confirm, currently, the user can only create a graph by hand. Another possible solution would be to use `gqlalchemy` to import memgraph, transform from memgraph graph networkx and then create mgp_mock `ctx` and run the module with `mgp_mock`, right? And this would be beneficial for debugging purposes.
@antoniofilipovic 
Concerning this point, there’s already a [PR on gqlalchemy](https://github.com/memgraph/gqlalchemy/pull/215) that exports the stored graph to networkx format; doing this here would be redundant

include/mgp_mock.py
antepusic: Yes, I removed this because must_abort is a DB-specific feature and thus not a part of the mock API

tests/e2e/mock_api/procedures/test_utils/CMakeLists.txt
antepusic: Fixed

tests/e2e/mock_api/procedures/CMakeLists.txt
antepusic: Fixed

antepusic:
@antoniofilipovic Final git message:
> Add Python query module API mock

include/_mgp_mock.py
antepusic: Keeping this unchanged so that the mock API is compatible out-of-the-box with the [import/export strategies PR](https://github.com/memgraph/gqlalchemy/pull/215).

antoniofilipovic (approved):

PR #759, opened 2023-02-03
[master < T575-FL] Improve overall performance of the C/C++ query modules API

as51340:
## Experiments with `mgp_result_record_insert`:

The assumption that needs to be tested is that there is some problem with our allocators:

| Description | Time spent on `mgp_result_record_insert` (%) :arrow_down:  |
| ----- | ----- |
| Baseline | 29.34 |
| memgraph::utils::pmr::unordered_map with std::string | 28.02 |
| memgraph::utils::pmr::unordered_map with std::string, reserve memory up-front = 4 | 27.41 |
| memgraph::utils::pmr::vector<memgraph::utils::pmr::string, TypedValue> | 36.57 |
| translation table with string to int mapping | 29.54 |
| reorganizing storage map[string to vector] | 22.23 |

### DIFF
--- a/src/query/plan/operator.cpp
+++ b/src/query/plan/operator.cpp
@@ -4459,7 +4459,12 @@ class CallProcedureCursor : public Cursor {
     }
     for (size_t i = 0; i < self_->result_fields_.size(); ++i) {
       std::string_view field_name(self_->result_fields_[i]);
-      auto result_it = values.find(field_name);
+      auto result_it =
+          std::find_if(values.begin(), values.end(),
+                       [&field_name](const std::pair<memgraph::utils::pmr::string, memgraph::query::TypedValue> &elem) {
+                         return elem.first == field_name;
+                       });
+      // auto result_it = values.find(field_name);
       if (result_it == values.end()) {
         throw QueryRuntimeException("Procedure '{}' did not yield a record with '{}' field.", self_->procedure_name_,
                                     field_name);
--- a/src/query/procedure/mg_procedure_impl.cpp
+++ b/src/query/procedure/mg_procedure_impl.cpp
@@ -1,4 +1,4 @@
-// Copyright 2022 Memgraph Ltd.
+// Copyright 2023 Memgraph Ltd.
 //
 // Use of this software is governed by the Business Source License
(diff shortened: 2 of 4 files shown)


as51340:
This PR is used for tracking profiling results for MGP. Both C API and C++ API are being profiled. The goal is to find bottlenecks for the node similarity implementation. Node similarity is quite specific algorithm in the MAGE library because it requires returning result for every pair of nodes thus, resulting in O(V^2) pair returned.

as51340:
First query that is being profiles:
`CALL node.similarity.jaccard() YIELD * RETURN *;`

as51340:
The next thing is to test the C++ and C API, so without any returning results.

**C++**
flamegraph_return_distinct_1_cpp_api

**C**

flamegraph_return_distinct_1_c_api


It can be seen that `mgp_result_record_insert` takes a lot of time and that too much time is spent on the API processing instead of computational logic.

as51340:
## Experiments with `mgp_result_record_insert`:

The assumption that needs to be tested is that there is some problem with our allocators:

| Description | Time spent on `mgp_result_record_insert` (%) |
| ----- | ----- |
| Baseline | 30.40 |
| memgraph::utils::pmr::unordered_map with std::string | 35.10 |

as51340:
New idea for the optimization was to mitigate the cost of emplacing string and replace it with with hopefully smaller cost of creating an integer. The process would like this:
1. create translation table map<string, int> only once 
2. for every record, use this translation table and create integer

However, this approach fails because field names must be extracted from the result signature but result signature is set as many times as there are results in thje Pull method of the CallProcedureCursor and it cannot be moved outside because of the possible deadlock that could occur.

as51340:
## Conclusion

Reorganizing MGP storage in a way that removes the need for mgp_record proved to have some potential by giving us 8-9% performance hit. However, after the sync with the team, we concluded that the effort of changing all APIs is too big for gaining only 8-9%.

antoniofilipovic:
@as51340 can we close this PR?

antoniofilipovic:
I am closing it as it would require too much work ATM for questionable gain.

as51340:
> @as51340 can we close this PR?

yes yes close it, I thought it was closed already :smile:

PR #760, opened 2023-02-06
[master < ] Update mgbench c++ client

antejavor:
- We have moved latency measurment to the client side. Each executed workload/query will come with its standard latency measurements (min, max, mean, p99, p95, p90, p75, p50) 
- Logging has been fixed, client is now reporting everything to stderr and stdout properly. 
- Validation mode accepts single query, execute it, and returns results alongside metadata. For workload validation between different vendors. 

Tasks: 
- [x] Add latency measurement on the client
- [x] Fix logging on the client  
- [x] Add validation mode 

Optional task: 
- [ ] Add per thread execution mode

### DIFF
--- a/tests/mgbench/client.cpp
+++ b/tests/mgbench/client.cpp
@@ -54,7 +54,10 @@ DEFINE_bool(queries_json, false,
 
 DEFINE_string(input, "", "Input file. By default stdin is used.");
 DEFINE_string(output, "", "Output file. By default stdout is used.");
-DEFINE_bool(validation, false, "Set to true to run client in validation mode");
+DEFINE_bool(validation, false,
+            "Set to true to run client in validation mode."
+            "Validation mode works for singe query and returns results for validation"
+            "with metadata");
 
 std::pair<std::map<std::string, memgraph::communication::bolt::Value>, uint64_t> ExecuteNTimesTillSuccess(
     memgraph::communication::bolt::Client *client, const std::string &query,
@@ -351,8 +354,8 @@ int main(int argc, char **argv) {
   spdlog::info("Running a bolt client with following settings:");
   spdlog::info("Adress: {} ", FLAGS_address);
   spdlog::info("Port: {} ", FLAGS_port);
-  spdlog::info("Username: {} ", FLAGS_port);
-  spdlog::info("Password: {} ", FLAGS_port);
+  spdlog::info("Username: {} ", FLAGS_username);
+  spdlog::info("Password: {} ", FLAGS_password);
   spdlog::info("Usessl: {} ", FLAGS_use_ssl);
   spdlog::info("Num of worker: {}", FLAGS_num_workers);
   spdlog::info("Max retries: {}", FLAGS_max_retries);
--- a/tests/mgbench/client.cpp
+++ b/tests/mgbench/client.cpp
@@ -22,6 +22,7 @@
 #include <vector>
 
 #include <gflags/gflags.h>
+#include <math.h>
 #include <json/json.hpp>
 
 #include "communication/bolt/client.hpp"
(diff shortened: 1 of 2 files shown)


antejavor:
@gitbuda The PR is ready for review, but let's keep it open for a bit. I may add a few more logging lines and per thread execution mode. We can merge during the week. Will leave it out in draft mode.

tests/mgbench/client.cpp
```diff
@@ -158,6 +188,34 @@ class Metadata final {
...
+nlohmann::json LatencyStatistics(std::vector<std::vector<double>> &worker_query_latency) {
+  nlohmann::json statistics = nlohmann::json::object();
+  std::vector<double> query_latency;
+  for (int i = 0; i < FLAGS_num_workers; i++) {
+    for (auto &e : worker_query_latency[i]) {
+      query_latency.push_back(e);
+    }
+  }
+  auto iterations = query_latency.size();
+  if (iterations > 10) {
```
gitbuda, line 200: Instead of a new variable, just use `query_latency.size()` + `10` should be a constant or runtime flag value?

tests/mgbench/client.cpp
```diff
@@ -158,6 +188,34 @@ class Metadata final {
...
+    statistics["mean"] = std::accumulate(query_latency.begin(), query_latency.end(), 0.0) / iterations;
+    statistics["p99"] = query_latency[floor(iterations * 0.99)];
+    statistics["p95"] = query_latency[floor(iterations * 0.95)];
+    statistics["p90"] = query_latency[floor(iterations * 0.90)];
+    statistics["p75"] = query_latency[floor(iterations * 0.75)];
+    statistics["p50"] = query_latency[floor(iterations * 0.50)];
+
+  } else {
+    spdlog::info("To few iterations to calculate latency values!");
+    statistics["iterations"] = query_latency.size();
```
gitbuda, line 214: E.g. there is the `iterations` variable, but still `query_latency.size()` is use here (because it's more explicit and convenient)

tests/mgbench/client.cpp
```diff
@@ -226,12 +290,94 @@ void Execute(
...
+  std::ostringstream oss;
+  for (int i = 0; i < results.size(); i++) {
+    oss << results[i];
+    res[std::to_string(i)] = oss.str();
+  }
+  return res;
+}
+
+// Validation mode works on single thread with 1 query.
+void Execute_validation(
```
gitbuda, line 308: Rename to `ExecuteValidation`

tests/mgbench/client.cpp
```diff
@@ -226,12 +290,94 @@ void Execute(
...
+  nlohmann::json res = nlohmann::json::object();
+  std::ostringstream oss;
+  for (int i = 0; i < results.size(); i++) {
+    oss << results[i];
+    res[std::to_string(i)] = oss.str();
+  }
+  return res;
+}
+
+// Validation mode works on single thread with 1 query.
```
gitbuda, line 307: Put commends like this in `///` because that's aligned with Doxygen

tests/mgbench/client.cpp
```diff
@@ -226,12 +290,94 @@ void Execute(
...
+      metadata.Append(ret.first);
+      results = ret.second;
+      duration = timer.Elapsed().count();
+      client.Close();
+    } else {
+      spdlog::info("Validation works with single query, pass just one query!");
+    }
+  });
+
+  thread.join();
```
gitbuda, line 344: All this can happen on the main thread, in other words you don't need an additional thread 🤔

gitbuda:
I've added a few comments, the whole thing should be polished and adjusted a bit, but overall it's very clear what is happening.

tests/mgbench/client.cpp
antejavor: It sould not be a runtime flag, since number of iterations is defined by the input number of queries. This was different in benchmark.py where we where executing some query defined number of iterations. 
This just defines a lower bound after which there is no sense to calculate latency values. Will transform this to local const varabile since it scope should be local.

antejavor:
Addressed all the comments in the recent commit. We can merge this, in my view.

gitbuda (approved):

antejavor:
This closes https://github.com/memgraph/memgraph/pull/760.

PR #763, opened 2023-02-06
[:pineapple:  < T1235-MG] Implement edge uniqueness filter with pull multiple

gvolfing:
[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR

### DIFF
--- a/src/query/v2/plan/operator.cpp
+++ b/src/query/v2/plan/operator.cpp
@@ -761,7 +761,7 @@ class DistributedScanByPrimaryKeyCursor : public Cursor {
               output_frame[output_symbol_] = TypedValue(it->second);
               populated_any = true;
               ++output_frame_it;
-            }           
+            }
             own_frames_it_->MakeInvalid();
           }
           break;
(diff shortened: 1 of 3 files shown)


src/query/v2/plan/operator.cpp
```diff
@@ -1333,28 +1333,92 @@ bool ContainsSameEdge(const TypedValue &a, const TypedValue &b) {
...
-    return true;
-  };
-
   while (input_cursor_->Pull(frame, context))
-    if (expansion_ok()) return true;
+    if (IsExpansionOk(frame, self_.expand_symbol_, self_.previous_symbols_)) return true;
   return false;
 }
 
+bool EdgeUniquenessFilter::EdgeUniquenessFilterCursor::PullMultiple(MultiFrame &output_multi_frame,
```
antaljanosbenjamin, line 1358: Why is this more complex than `FilterCursor`? In my mind `EdgeUniquenessFilter` is just a special filter, where we don't have an explicit filter expression, but an a very special, implicit (two edges are not the same).

As I see that `FilterCursor::PullMultiple` is much simpler than `EdgeUniquenessFilterCursor::PullMultiple`, I am wondering whether it is necessary or not.

antaljanosbenjamin (approved):

src/query/v2/plan/operator.cpp
```diff
@@ -1333,28 +1333,48 @@ bool ContainsSameEdge(const TypedValue &a, const TypedValue &b) {
...
+bool IsExpansionOk(Frame &frame, const Symbol &expand_symbol, const std::vector<Symbol> &previous_symbols) {
+  const auto &expand_value = frame[expand_symbol];
+  for (const auto &previous_symbol : previous_symbols) {
+    const auto &previous_value = frame[previous_symbol];
+    // This shouldn't raise a TypedValueException, because the planner
+    // makes sure these are all of the expected type. In case they are not
+    // an error should be raised long before this code is executed.
+    if (ContainsSameEdge(previous_value, expand_value)) return false;
+  }
+  return true;
```
jbajic, line 1346:
```suggestion
    // This shouldn't raise a TypedValueException, because the planner
    // makes sure these are all of the expected type. In case they are not
    // an error should be raised long before this code is executed.
   return std::ranges::all_of(previous_symbols, [&expand_value = frame[expand_symbol]](const auto& previous_symbol) {
     return ContainsSameEdge(previous_value, expand_value);
   });
```

PR #764, opened 2023-02-06
[T1234 < T1236] Add filtering support to WHERE EXISTS

Josipmrden:
This PR concerns the filtering of exists and nesting possible pattern filters in the existing infrastructure.


src/query/plan/preprocess.cpp
```diff
@@ -530,6 +480,78 @@ void Filters::AnalyzeAndStoreFilter(Expression *expr, const SymbolTable &symbol_
...
 }
 
+// Fills the given Matching, by converting the Match patterns to normalized
+// representation as Expansions. Filters used in the Match are also collected,
+// as well as edge symbols which determine Cyphermorphism. Collecting filters
+// will lift them out of a pattern and generate new expressions (just like they
+// were in a Where clause).
+void AddMatching(const std::vector<Pattern *> &patterns, Where *where, SymbolTable &symbol_table, AstStorage &storage,
+                 Matching &matching) {
+  auto expansions = NormalizePatterns(symbol_table, patterns);
```
as51340, line 490:
```suggestion
  const auto &expansions = NormalizePatterns(symbol_table, patterns);
```

src/query/plan/preprocess.cpp
```diff
@@ -530,6 +480,78 @@ void Filters::AnalyzeAndStoreFilter(Expression *expr, const SymbolTable &symbol_
...
+      // Add edge and node2 to all symbols
+      matching.expansion_symbols.insert(edge_sym);
+      matching.expansion_symbols.insert(node2_sym);
+    }
+    matching.expansions.push_back(expansion);
+  }
+  if (!edge_symbols.empty()) {
+    matching.edge_symbols.emplace_back(edge_symbols);
+  }
+  for (auto *pattern : patterns) {
```
as51340, line 516:
```suggestion
  for (auto *const pattern : patterns) {
```

src/query/plan/preprocess.cpp
```diff
@@ -530,6 +480,78 @@ void Filters::AnalyzeAndStoreFilter(Expression *expr, const SymbolTable &symbol_
...
+    matching.filters.CollectPatternFilters(*pattern, symbol_table, storage);
+    if (pattern->identifier_->user_declared_) {
+      std::vector<Symbol> path_elements;
+      for (auto *pattern_atom : pattern->atoms_)
```
as51340, line 520:
```suggestion
      for (auto * const pattern_atom : pattern->atoms_)
```

src/query/plan/preprocess.cpp
```diff
@@ -530,6 +480,78 @@ void Filters::AnalyzeAndStoreFilter(Expression *expr, const SymbolTable &symbol_
...
+        path_elements.emplace_back(symbol_table.at(*pattern_atom->identifier_));
```
as51340, line 521:
```suggestion
        path_elements.push_back(symbol_table.at(*pattern_atom->identifier_));
```

src/query/plan/preprocess.cpp
antoniofilipovic: Actually, it is good how it is, because function returns std::vector<Expansion>, so there is no need for reference. RVO should kick in

src/query/plan/operator.cpp
```diff
@@ -2269,10 +2269,24 @@ UniqueCursorPtr Filter::MakeCursor(utils::MemoryResource *mem) const {
...
 std::vector<Symbol> Filter::ModifiedSymbols(const SymbolTable &table) const { return input_->ModifiedSymbols(table); }
 
+static std::vector<UniqueCursorPtr> MakeCursorVector(const std::vector<std::shared_ptr<LogicalOperator>> &ops,
+                                                     utils::MemoryResource *mem) {
+  std::vector<UniqueCursorPtr> cursors;
+  cursors.reserve(ops.size());
+
+  if (!ops.empty()) {
+    for (const auto &op : ops) {
+      cursors.push_back(op->MakeCursor(mem));
```
antepusic, line 2279:
```suggestion
      cursors.emplace_back(op->MakeCursor(mem));
```

tests/gql_behave/tests/memgraph_V1/features/memgraph_exists.feature
```diff
@@ -65,3 +80,163 @@ Feature: WHERE exists
...
+          CREATE (:One {prop:1})-[:TYPE {prop: 1}]->(:Two {prop: 2})
+          """
+      When executing query:
+          """
+          MATCH (n:One) WHERE exists((n)-[:TYPE {prop: 1}]->(:Two {prop: 2})) RETURN n.prop;
+          """
+      Then the result should be:
+          | n.prop |
+          | 1      |
+
```
antepusic, line 175: Would it make sense to add these tests?
* Test exists with wrong edge property and correct node label property
* Test exists with correct edge property and wrong node label property
* Test exists with wrong edge property and node label property

antepusic (changes requested):
I added a few quick suggestions

src/query/plan/operator.cpp
antaljanosbenjamin: Just to note: with a temporary `push_back` does the same thing as `emplace_back` (move construct), as it has an overload with rvalue reference.

src/query/plan/operator.cpp
```diff
@@ -2282,8 +2296,10 @@ bool Filter::FilterCursor::Pull(Frame &frame, ExecutionContext &context) {
   ExpressionEvaluator evaluator(&frame, context.symbol_table, context.evaluation_context, context.db_accessor,
                                 storage::View::OLD);
   while (input_cursor_->Pull(frame, context)) {
-    if (pattern_filter_cursor_) {
-      pattern_filter_cursor_->Pull(frame, context);
+    if (!pattern_filter_cursors_.empty()) {
+      for (const auto &pattern_filter_cursor : pattern_filter_cursors_) {
+        pattern_filter_cursor->Pull(frame, context);
+      }
     }
```
antaljanosbenjamin, line 2303:
```suggestion
    for (const auto &pattern_filter_cursor : pattern_filter_cursors_) {
      pattern_filter_cursor->Pull(frame, context);
    }
```

src/query/plan/preprocess.hpp
```diff
@@ -125,6 +125,88 @@ struct Expansion {
   NodeAtom *node2 = nullptr;
 };
 
+struct FilterMatching;
+
+enum class PatternFilterType { EXISTS };
```
antaljanosbenjamin, line 130: Does it make sense to have an enum class with a single value?

src/query/plan/preprocess.hpp
```diff
@@ -125,6 +125,88 @@ struct Expansion {
...
+  }
+  void Visit(AndOperator &op) override {
+    op.expression1_->Accept(*this);
+    op.expression2_->Accept(*this);
+  }
+
+  // Unary operators
+  void Visit(UnaryPlusOperator &op) override{};
+  void Visit(UnaryMinusOperator &op) override{};
+  void Visit(IsNullOperator &op) override{};
```
antaljanosbenjamin, line 160: I couldn't figure out, so that why I am asking: what happens with a query `MATCH (n) WHERE exists((n)-[]->()) IS NOT NULL RETURN n`?

src/query/plan/preprocess.hpp
```diff
@@ -125,6 +125,88 @@ struct Expansion {
...
+
+  // Binary operators
+  void Visit(AdditionOperator &op) override{};
+  void Visit(SubtractionOperator &op) override{};
+  void Visit(MultiplicationOperator &op) override{};
+  void Visit(DivisionOperator &op) override{};
+  void Visit(ModOperator &op) override{};
+  void Visit(NotEqualOperator &op) override{};
+  void Visit(EqualOperator &op) override{};
```
antaljanosbenjamin, line 169: Same as above but `MATCH (n) WHERE exists((n)-[]->()) = true RETURN n`

src/query/plan/preprocess.hpp
```diff
@@ -125,6 +125,88 @@ struct Expansion {
...
+  void Visit(IfOperator &op) override{};
+  void Visit(ListLiteral &op) override{};
+  void Visit(MapLiteral &op) override{};
+  void Visit(LabelsTest &op) override{};
+  void Visit(Aggregation &op) override{};
+  void Visit(Function &op) override{};
+  void Visit(Reduce &op) override{};
+  void Visit(Coalesce &op) override{};
+  void Visit(Extract &op) override{};
+  void Visit(Exists &op) override;
```
antaljanosbenjamin, line 188: Please somehow emphasize that this is implemented (e.g.: moving up to the other functions), because it is super easy to not notice it.

antaljanosbenjamin (commented):
I tried to deep dive into the rule based planner, but it is more complicated than I thought, so I did a shallow review on that, therefore I am not confident to approve this PR. I think it might make more sense for other people in team field to review it, so they can deep dive into the planner which is a way to spread the knowledge about theh planner.

tests/gql_behave/tests/memgraph_V1/features/memgraph_exists.feature
Josipmrden: will add

src/query/plan/preprocess.hpp
Josipmrden: I just added it for later extension, so we do no assumptions when creating pattern filters now. does it make sense? :')

Josipmrden: i use it just in a if-else followed by an assert(false

src/query/plan/preprocess.hpp
Josipmrden: i forgot to implement, it's coming. I'm going to see for other ones as well does it make sense

src/query/plan/preprocess.hpp
Josipmrden: i forgot to implement, it's coming. I'm going to see for other ones as well does it make sense

src/query/plan/preprocess.hpp
antaljanosbenjamin: If you use enums, then use `switch` without `default` branches: reason: if you extend the enum but don't handle it in an switch, then the compiler will scream about it and the code won't compile. With if+assert you are pushing the same check to runtime.

antaljanosbenjamin: Whether to keep the enum or not: what other types of pattern filter can we have?

src/query/plan/preprocess.cpp
```diff
@@ -530,6 +480,78 @@ void Filters::AnalyzeAndStoreFilter(Expression *expr, const SymbolTable &symbol_
...
+  std::vector<Pattern *> patterns;
+  patterns.push_back(op.pattern_);
+
+  FilterMatching filter_matching;
+  AddMatching(patterns, nullptr, this->symbol_table_, this->storage_, filter_matching);
+
+  filter_matching.type = PatternFilterType::EXISTS;
+  filter_matching.symbol = std::make_optional<Symbol>(this->symbol_table_.at(op));
+
+  this->matchings_.push_back(std::move(filter_matching));
```
antoniofilipovic, line 552: no need for `this->matchings_`, you can only use `matchings_`

src/query/plan/preprocess.hpp
antoniofilipovic: for me this is okay, in case we expand usage for later with something else

src/query/plan/preprocess.hpp
```diff
@@ -121,6 +121,93 @@ struct Expansion {
...
+  void Visit(PrimitiveLiteral &op) override{};
+  void Visit(PropertyLookup &op) override{};
+  void Visit(ParameterLookup &op) override{};
+  void Visit(NamedExpression &op) override{};
+  void Visit(RegexMatch &op) override{};
+
+  std::vector<FilterMatching> getMatchings() { return matchings_; }
+
+  SymbolTable &symbol_table_;
+  AstStorage &storage_;
```
antoniofilipovic, line 204: store this as private variable and use `const` around it

src/query/plan/preprocess.hpp
```diff
@@ -121,6 +121,93 @@ struct Expansion {
...
+
+ private:
+  /// Collection of matchings in the filter expression being analyzed.
+  std::vector<FilterMatching> matchings_;
```
antoniofilipovic, line 208: Maybe it makes sense to store matchings_ also as public?

src/query/plan/rule_based_planner.cpp
```diff
@@ -499,7 +499,8 @@ std::unique_ptr<LogicalOperator> GenReturnBody(std::unique_ptr<LogicalOperator>
   // Where may see new symbols so it comes after we generate Produce and in
   // general, comes after any OrderBy, Skip or Limit.
   if (body.where()) {
-    last_op = std::make_unique<Filter>(std::move(last_op), nullptr, body.where()->expression_);
+    last_op = std::make_unique<Filter>(std::move(last_op), std::vector<std::shared_ptr<LogicalOperator>>{},
```
antoniofilipovic, line 502: I am not sure, but `{}` should work too.

src/query/plan/rule_based_planner.hpp
```diff
@@ -569,62 +583,55 @@ class RuleBasedPlanner {
   std::unique_ptr<LogicalOperator> GenFilters(std::unique_ptr<LogicalOperator> last_op,
                                               const std::unordered_set<Symbol> &bound_symbols, Filters &filters,
                                               AstStorage &storage, const SymbolTable &symbol_table) {
-    std::unique_ptr<LogicalOperator> pattern_filter =
-        ExtractPatternFilters(filters, storage, symbol_table, bound_symbols);
+    auto pattern_filters = ExtractPatternFilters(filters, symbol_table, storage, bound_symbols);
```
antoniofilipovic, line 586: since you are using pattern filters only when filter expr exists, does it make sense to move it inside `if{}`

src/query/plan/rule_based_planner.hpp
```diff
@@ -569,62 +583,55 @@ class RuleBasedPlanner {
...
+  std::vector<std::shared_ptr<LogicalOperator>> ExtractPatternFilters(Filters &filters, const SymbolTable &symbol_table,
+                                                                      AstStorage &storage,
+                                                                      const std::unordered_set<Symbol> &bound_symbols) {
+    std::vector<std::shared_ptr<LogicalOperator>> operators;
 
-    last_op = std::make_unique<EvaluatePatternFilter>(std::move(last_op), symbol_table.at(exists));
+    for (const auto &filter : filters) {
+      for (const auto &matching : filter.matchings) {
+        switch (matching.type) {
+          case PatternFilterType::EXISTS: {
```
antoniofilipovic, line 626: I would add here default so compailer doesn't scream in case we add new value here. I think it should prevent it.

```default:
        break;
```

tests/gql_behave/tests/memgraph_V1/features/memgraph_exists.feature
```diff
@@ -14,6 +14,21 @@ Feature: WHERE exists
...
 
+  Scenario: Test exists with empty edge and node specifiers return 2 entries
+      Given an empty graph
+      And having executed:
+          """
+          CREATE (:One {prop:1})-[:TYPE]->(:Two), (:One {prop: 3})-[:TYPE]->(:Two)
+          """
+      When executing query:
+          """
+          MATCH (n:One) WHERE exists((n)-[]-()) RETURN n.prop;
```
antoniofilipovic, line 25: add order by here so it doesn't break later

tests/gql_behave/tests/memgraph_V1/features/memgraph_exists.feature
```diff
@@ -104,3 +315,81 @@ Feature: WHERE exists
...
+
+  Scenario: Test exists is null
+      Given an empty graph
+      And having executed:
+          """
+          CREATE (:One {prop:1})-[:TYPE]->(:Two);
+          """
+      When executing query:
+          """
+          MATCH (n:One) WHERE exists((n)-[]-()) is null RETURN n.prop;
```
antoniofilipovic, line 341: What do we test with exists is null? Shouldn't exist be true or false?

tests/unit/query_plan.cpp
```diff
@@ -1750,9 +1752,81 @@ TYPED_TEST(TestPlanner, Exists) {
...
+                                                            new ExpectEvaluatePatternFilter()};
+
+    CheckPlan(
+        planner.plan(), symbol_table, ExpectScanAll(),
+        ExpectFilter(std::vector<std::list<BaseOpChecker *>>{pattern_filter_with_types, pattern_filter_without_types}),
+        ExpectProduce());
+
+    DeleteListContent(&pattern_filter_with_types);
+    DeleteListContent(&pattern_filter_without_types);
+  }
```
antoniofilipovic, line 1830: Does it make sense to test more complex query like:
 MATCH (n) WHERE n.prop=1 and exists((n)-[:TYPE]-(:Two))

Now here we should expect n.prop=1 be part of where and not of exists.

That is how it is implemented now.

tests/unit/query_plan_read_write_typecheck.cpp
```diff
@@ -57,7 +57,8 @@ TEST_F(ReadWriteTypeCheckTest, CreateNode) {
 TEST_F(ReadWriteTypeCheckTest, Filter) {
   std::shared_ptr<LogicalOperator> scan_all = std::make_shared<ScanAll>(nullptr, GetSymbol("node1"));
   std::shared_ptr<LogicalOperator> filter =
-      std::make_shared<Filter>(scan_all, nullptr, EQ(PROPERTY_LOOKUP("node1", dba.NameToProperty("prop")), LITERAL(0)));
+      std::make_shared<Filter>(scan_all, std::vector<std::shared_ptr<LogicalOperator>>{},
```
antoniofilipovic, line 60: I think you can shorten all of these with `{}`

antoniofilipovic (changes requested):
From me, there are a few changes to do and that should be it. I would just like to note that it seems to me there are a few optimizations that we could make, and we are not doing them now. 

For example, evaluating firstly WHERE part of the query without evaluating `exists`. That could be possible through hackish behaviour of code where we return false if filtering is executed and we don't need to continue evaluating exists, and -1 for example if we need to continue and evaluate exists (that would be case when you have `MATCH (n:Label) WHERE n.prop=1 and exists((n)-[]->(:Label))`. For example here if n.prop is not 1 there is no need to expanding node. 

Another thing I am not quite sure about now is how to move in filters from WHERE part of query into exists part. For example, now we are using `ExpandCursor`, and there could be valuable information from node labels to edge types we could use. For example in the query  `MATCH (n:Label)-[e:HAS]->(m:Type) WHERE n.prop=1 and exists((n)-[]->(m {a=2}))` due to the order of query plan it will to following things:
- ScanAllByLabel (n:Label)
- Filter n.prop=1 and exists()
- Expand (n)-[e]->(m)
- Filter (Type)
- Produce

Exists will do the following:
expand (n)  and search for node prop `a=2`. But node `m` should also have `label:Type`

But we see here that we could use Type in exists and filter out earlier maybe?

Otherwise I would say good job here, I don't have anything to add more.

src/query/plan/rule_based_planner.hpp
antaljanosbenjamin: I would avoid adding default, because my reasoning for opt-in to switch is exactly to make the compiler scream at us if we don't handle an enum value somewhere. I think for long time now the `default` branches are prohibited in our codebase (we didn't updated it retrospectively, but we are trying to catch them in PRs).

antaljanosbenjamin: Basically without default the compiler enforces us to think how to handle each enum value in places where the enum is used.

antoniofilipovic: That makes sense, I didn't think about that from such a point of view. Thanks for clarifying :smile:

antoniofilipovic:
@Josipmrden there are unresolved conversations here, please check what I wrote beforehand

src/query/plan/preprocess.cpp
```diff
@@ -530,6 +480,78 @@ void Filters::AnalyzeAndStoreFilter(Expression *expr, const SymbolTable &symbol_
...
+  std::vector<Pattern *> patterns;
+  patterns.push_back(op.pattern_);
+
+  FilterMatching filter_matching;
+  AddMatching(patterns, nullptr, this->symbol_table_, this->storage_, filter_matching);
+
+  filter_matching.type = PatternFilterType::EXISTS;
+  filter_matching.symbol = std::make_optional<Symbol>(this->symbol_table_.at(op));
+
+  matchings_.push_back(std::move(filter_matching));
```
antoniofilipovic, line 552: You can remove here also `this`

src/query/plan/rule_based_planner.hpp
antoniofilipovic: I understand now that this doesn't work, as `impl::ExtractFilters(bound_symbols, filters, storage);` will delete your filter used.

antoniofilipovic: Just confirm this, and I will resolve comment

antoniofilipovic: Confirmed

src/query/plan/preprocess.hpp
antoniofilipovic: I would change it to the public as it is in other `structs` that way as well. @Josipmrden

src/query/plan/rule_based_planner.cpp
antoniofilipovic: Instead of `std::vector<std::shared_ptr...>> `
@Josipmrden

src/query/plan/preprocess.hpp
antoniofilipovic: @Josipmrden check if it makes sense, otherwise I will resolve this

src/query/plan/preprocess.hpp
antoniofilipovic: @antaljanosbenjamin where to move this, and @Josipmrden please check this comment :smile:

antoniofilipovic:
@Josipmrden there are a few unresolved comments, and once that is done, we can move on with this PR.

Josipmrden:
Checks passing, one final check and then we merge.

antoniofilipovic (approved):
Good to go

PR #766, opened 2023-02-06
[DRAFT][T1231 < T1233] Where exists generation of symbols and plan generation

Josipmrden:
This PR concerns handling symbols and plan generation for the WHERE EXISTS query

### DIFF
--- a/src/query/frontend/semantic/symbol_generator.cpp
+++ b/src/query/frontend/semantic/symbol_generator.cpp
@@ -443,15 +443,14 @@ bool SymbolGenerator::PreVisit(Extract &extract) {
 bool SymbolGenerator::PreVisit(Exists &exists) {
   auto &scope = scopes_.back();
   scope.in_exists = true;
+  scope.in_pattern = true;
 
   const auto &symbol = CreateAnonymousSymbol();
   exists.MapTo(symbol);
 
-  exists.node_identifier_->Accept(*this);
-
-  scope.in_pattern = true;
+  exists.to_node_info_->Accept(*this);
   exists.relationship_info_->Accept(*this);
-  exists.node_info_->Accept(*this);
+  exists.from_node_info_->Accept(*this);
 
   return false;
 }
--- a/src/query/frontend/semantic/symbol_generator.cpp
+++ b/src/query/frontend/semantic/symbol_generator.cpp
@@ -302,6 +302,11 @@ SymbolGenerator::ReturnType SymbolGenerator::Visit(Identifier &ident) {
   if (scope.in_skip || scope.in_limit) {
     throw SemanticException("Variables are not allowed in {}.", scope.in_skip ? "SKIP" : "LIMIT");
   }
+  if (scope.in_exists && (scope.visiting_edge || scope.in_node_atom)) {
+    if (HasSymbol(ident.name_) || ConsumePredefinedIdentifier(ident.name_)) {
+      throw SemanticException("Unbounded variables are not allowed in exists!");
+    }
+  }
   Symbol symbol;
   if (scope.in_pattern && !(scope.in_node_atom || scope.visiting_edge)) {
     // If we are in the pattern, and outside of a node or an edge, the
@@ -431,8 +436,23 @@ bool SymbolGenerator::PreVisit(Extract &extract) {
 }
 
 bool SymbolGenerator::PreVisit(Exists &exists) {
+  auto &scope = scopes_.back();
+  scope.in_exists = true;
+
   exists.node_identifier_->Accept(*this);
+
+  scope.in_pattern = true;
+  exists.relationship_info_->Accept(*this);
+  exists.node_info_->Accept(*this);
+
   return false;
+}
+
+bool SymbolGenerator::PostVisit(Exists &exists) {
+  auto &scope = scopes_.back();
+  scope.in_exists = false;
+  scope.in_pattern = false;
+  return true;
 }
 
 // Pattern and its subparts.
--- a/src/query/frontend/semantic/symbol_generator.cpp
+++ b/src/query/frontend/semantic/symbol_generator.cpp
@@ -1,4 +1,4 @@
-// Copyright 2022 Memgraph Ltd.
+// Copyright 2023 Memgraph Ltd.
 //
 // Use of this software is governed by the Business Source License
 // included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
@@ -435,6 +435,11 @@ bool SymbolGenerator::PreVisit(Reduce &reduce) {
 bool SymbolGenerator::PreVisit(Extract &extract) {
   extract.list_->Accept(*this);
   VisitWithIdentifiers(extract.expression_, {extract.identifier_});
+  return false;
+}
+
+bool SymbolGenerator::PreVisit(Exists &exists) {
+  exists.node_identifier_->Accept(*this);
   return false;
 }
 
--- a/src/query/frontend/semantic/symbol_generator.cpp
+++ b/src/query/frontend/semantic/symbol_generator.cpp
@@ -308,7 +308,7 @@ SymbolGenerator::ReturnType SymbolGenerator::Visit(Identifier &ident) {
     throw SemanticException("Variables are not allowed in {}.", scope.in_skip ? "SKIP" : "LIMIT");
   }
   if (scope.in_exists && (scope.visiting_edge || scope.in_node_atom)) {
-    if (HasSymbol(ident.name_) || ConsumePredefinedIdentifier(ident.name_)) {
+    if (!HasSymbol(ident.name_) && !ConsumePredefinedIdentifier(ident.name_) && ident.user_declared_) {
       throw SemanticException("Unbounded variables are not allowed in exists!");
     }
   }
--- a/src/query/frontend/semantic/symbol_generator.cpp
+++ b/src/query/frontend/semantic/symbol_generator.cpp
@@ -64,6 +64,11 @@ auto SymbolGenerator::CreateSymbol(const std::string &name, bool user_declared, 
   return symbol;
 }
 
+auto SymbolGenerator::CreateAnonymousSymbol(Symbol::Type type) {
+  auto symbol = symbol_table_->CreateAnonymousSymbol();
+  return symbol;
+}
+
 auto SymbolGenerator::GetOrCreateSymbol(const std::string &name, bool user_declared, Symbol::Type type) {
   // NOLINTNEXTLINE
   for (auto scope = scopes_.rbegin(); scope != scopes_.rend(); ++scope) {
@@ -438,6 +443,9 @@ bool SymbolGenerator::PreVisit(Extract &extract) {
 bool SymbolGenerator::PreVisit(Exists &exists) {
   auto &scope = scopes_.back();
   scope.in_exists = true;
+
+  const auto &symbol = CreateAnonymousSymbol();
+  exists.MapTo(symbol);
 
   exists.node_identifier_->Accept(*this);
 
--- a/src/query/frontend/semantic/symbol_generator.cpp
+++ b/src/query/frontend/semantic/symbol_generator.cpp
@@ -446,26 +446,22 @@ bool SymbolGenerator::PreVisit(Extract &extract) {
 
 bool SymbolGenerator::PreVisit(Exists &exists) {
   auto &scope = scopes_.back();
+
+  if (exists.pattern_->atoms_.size() != 3) {
+    throw SemanticException("Exists functionality only supports triplet patterns!");
+  }
+
   scope.in_exists = true;
-  scope.in_pattern = true;
 
   const auto &symbol = CreateAnonymousSymbol();
   exists.MapTo(symbol);
 
-  scope.in_exists_source_node = true;
-  exists.from_node_info_->Accept(*this);
-  scope.in_exists_source_node = false;
-
-  exists.relationship_info_->Accept(*this);
-  exists.to_node_info_->Accept(*this);
-
-  return false;
+  return true;
 }
 
-bool SymbolGenerator::PostVisit(Exists &exists) {
+bool SymbolGenerator::PostVisit(Exists &) {
   auto &scope = scopes_.back();
   scope.in_exists = false;
-  scope.in_pattern = false;
   return true;
 }
 
@@ -477,6 +473,20 @@ bool SymbolGenerator::PreVisit(Pattern &pattern) {
   if ((scope.in_create || scope.in_merge) && pattern.atoms_.size() == 1U) {
     MG_ASSERT(utils::IsSubtype(*pattern.atoms_[0], NodeAtom::kType), "Expected a single NodeAtom in Pattern");
     scope.in_create_node = true;
+  } else if (scope.in_exists) {
+    // Currently we're not dealing with variable start planner so this feels a bit hacky
+    // TODO: Configure variable start planner to work with exists() in order to delete this part
+    for (auto i = 0; i < pattern.atoms_.size(); i++) {
+      if (i == 0) {
+        scope.in_exists_source_node = true;
+        pattern.atoms_[i]->Accept(*this);
+        scope.in_exists_source_node = false;
+      } else {
+        pattern.atoms_[i]->Accept(*this);
+      }
+    }
+
+    return false;
   }
   return true;
 }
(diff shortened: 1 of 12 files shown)


antaljanosbenjamin:
Why do we need the `IntegerLiteral`?

src/query/plan/rule_based_planner.hpp
```diff
@@ -567,6 +565,66 @@ class RuleBasedPlanner {
...
   }
+
+  std::unique_ptr<LogicalOperator> GenFilters(std::unique_ptr<LogicalOperator> last_op,
+                                              const std::unordered_set<Symbol> &bound_symbols, Filters &filters,
+                                              AstStorage &storage, const SymbolTable &symbol_table) {
+    std::unique_ptr<LogicalOperator> pattern_filter =
+        ExtractPatternFilters(filters, storage, symbol_table, bound_symbols);
+    auto *filter_expr = impl::ExtractFilters(bound_symbols, filters, storage);
+    if (filter_expr) {
+      last_op = std::make_unique<Filter>(std::move(last_op), std::move(pattern_filter), filter_expr);
```
antoniofilipovic, line 576: Does this mean if user writes :

```
MATCH (n)
WHERE n.prop=1 and exists((n)-[]->(m))
RETURN n;
```

we will first try to expand and then use n.prop?

src/query/plan/rule_based_planner.hpp
```diff
@@ -567,6 +565,66 @@ class RuleBasedPlanner {
...
+  std::unique_ptr<LogicalOperator> ExtractPatternFilters(const Filters &filters, AstStorage &storage,
+                                                         const SymbolTable &symbol_table,
+                                                         const std::unordered_set<Symbol> &bound_symbols) {
+    for (const auto &filter : filters) {
+      if (filter.type != FilterInfo::Type::Pattern) {
+        continue;
+      }
+
+      if (const auto *exists = utils::Downcast<Exists>(filter.expression)) {
+        return MakeExistsFilter(*exists, storage, symbol_table, bound_symbols);
```
antoniofilipovic, line 590: Dumb question, but what if in filters we have more than one Exists? Shouldn't we process that too?

antoniofilipovic: This just takes the first `Exists`, right?

src/query/plan/rule_based_planner.hpp
```diff
@@ -567,6 +565,66 @@ class RuleBasedPlanner {
...
+      }
+
+      throw SemanticException("Pattern filter does not exist!");
```
antoniofilipovic, line 593: This is bug for us, this shouldn't occur, right? Exists is always Pattern filter?

src/query/plan/rule_based_planner.hpp
```diff
@@ -567,6 +565,66 @@ class RuleBasedPlanner {
...
+      edge_types.push_back(GetEdgeType(type));
+    }
+
+    std::vector<Symbol> once_symbols(bound_symbols.begin(), bound_symbols.end());
+    std::unique_ptr<LogicalOperator> last_op = std::make_unique<Once>(once_symbols);
+
+    last_op = std::make_unique<Expand>(std::move(last_op), node1_symbol, node2_symbol, edge_symbol, direction,
+                                       edge_types, false, storage::View::OLD);
+
+    last_op = std::make_unique<Limit>(std::move(last_op), storage.Create<IntegerLiteral>());
```
antoniofilipovic, line 622: I am somewhat confident we can use here Identifier which will read from frame and return 1.

antoniofilipovic: Here we have a definition of an Identifier and what it does.

https://memgraph.notion.site/Semantic-Analysis-and-Symbol-Generation-57b5942e8165476da91f499be71fe326

If we create an anonymous symbol on the frame and store in it TypedValue(1), we can get such a symbol here, maybe, and use Identifier as Expression.

src/query/frontend/semantic/symbol_table.hpp
```diff
@@ -51,6 +51,7 @@ class SymbolTable final {
   const Symbol &at(const Identifier &ident) const { return table_.at(ident.symbol_pos_); }
   const Symbol &at(const NamedExpression &nexpr) const { return table_.at(nexpr.symbol_pos_); }
   const Symbol &at(const Aggregation &aggr) const { return table_.at(aggr.symbol_pos_); }
+  const Symbol &at(const Exists &exists) const { return table_.at(exists.symbol_pos_); }
```
antoniofilipovic, line 54: Can we make Exists as NamedExpression? As I understood, NamedExpression stores value on Frame as Exists does.

I am pretty confident that we can, I am not sure as what was reasoning to do it as new class. Because if we now that filter is of type Pattern, than expression in filter is used as NamedExpression and functionality should be the same, right?

```
std::unique_ptr<LogicalOperator> MakeExistsFilter(const NamedExpression &exists, AstStorage &storage,
                                                    const SymbolTable &symbol_table,
                                                    const std::unordered_set<Symbol> &bound_symbols) {
```

I think that should be possible.

antoniofilipovic (commented):
I need still to check tests, but here I left few comments.

gitbuda:
Depends on #767

src/query/plan/rule_based_planner.hpp
Josipmrden: We always try to expand first (next PR that I'm going to open).
The exists is going to perform slowly if there is a supernode with little amount of hits. However, it stops whenever it has found an edge, worst case is if it does not have an edge, then indeed it needs to go through all the edges.
However, the queries which don't have exists are not affected.

Josipmrden: An optimization would be to check first the expressions and then run the pattern filter, however that needs a little more architectural work, it's a bit harder to do. I'd stick to functionality for now and optimize later.

src/query/plan/rule_based_planner.hpp
Josipmrden: Addressed in following PRs

src/query/plan/rule_based_planner.hpp
Josipmrden: Yes. I'm not sure if we need to assert or something, is it better ?

Josipmrden:
@antaljanosbenjamin  I didn't know how to make an expression which returns a constant value (to simulate LIMIT 1), so I made this, but if we have something that would be great

src/query/frontend/semantic/symbol_table.hpp
Josipmrden: I would not put it under named expression, since indeed it does not have a name and the expression has totally different logic of execution. NamedExpressions are used in returns. This one executes a sub-branch

antaljanosbenjamin:
@Josipmrden check out `PrimitiveLiteral` and how it is used in `CypherMainVisitor::visitLiteral`. In short a `PrimitiveLiteral` can hold a `TypedValue`, so can easily inject `1` to simulate `LIMIT 1`.

src/query/frontend/semantic/symbol_table.hpp
antoniofilipovic: As we talked about this and the problem with using `NamedExpression` is that it would require changing of logic inside its visitor and has different logic, I will resolve this comment

src/query/plan/rule_based_planner.hpp
antaljanosbenjamin: If this is a programmer error => assert.

src/query/plan/rule_based_planner.cpp
```diff
@@ -495,7 +495,7 @@ std::unique_ptr<LogicalOperator> GenReturnBody(std::unique_ptr<LogicalOperator>
   // Where may see new symbols so it comes after we generate Produce and in
   // general, comes after any OrderBy, Skip or Limit.
   if (body.where()) {
-    last_op = std::make_unique<Filter>(std::move(last_op), body.where()->expression_);
+    last_op = std::make_unique<Filter>(std::move(last_op), nullptr, body.where()->expression_);
```
antaljanosbenjamin, line 498: So `MATCH (n) RETURN n WHERE exists((n)-[]->())` is not a valid query, right? Is it enforced?

src/query/plan/preprocess.hpp
```diff
@@ -79,6 +99,12 @@ class UsedSymbolsCollector : public HierarchicalTreeVisitor {
 
   std::unordered_set<Symbol> symbols_;
   const SymbolTable &symbol_table_;
+
+ private:
+  struct Scope {
+    bool in_exists{false};
+  };
+  std::vector<Scope> scopes_;
```
antaljanosbenjamin, line 107: You never change the size of `scopes_` from 1, so maybe they are not needed. A single `bool in_exists_` shuld suffice in my opinion.

antaljanosbenjamin (commented):
I have to check the tests, but it seems good!

src/query/plan/rule_based_planner.hpp
antoniofilipovic: @Josipmrden

src/query/plan/preprocess.hpp
antoniofilipovic: Agree

src/query/plan/rule_based_planner.hpp
antoniofilipovic: We can measure how this will work, but in my opinion, if it will be slow this should be the first optimization point. Checking whether a node has property should be less expensive than checking for a certain pattern for a node and doing expansion.

src/query/frontend/semantic/symbol_generator.cpp
```diff
@@ -302,6 +307,17 @@ SymbolGenerator::ReturnType SymbolGenerator::Visit(Identifier &ident) {
   if (scope.in_skip || scope.in_limit) {
     throw SemanticException("Variables are not allowed in {}.", scope.in_skip ? "SKIP" : "LIMIT");
   }
+
+  if (scope.in_exists && (scope.visiting_edge || scope.in_node_atom)) {
+    auto has_symbol = HasSymbol(ident.name_);
```
antoniofilipovic, line 312: I think there might be a bug here.

I tried the following query in order to break this part:
```
MATCH (n),(m) 
WITH n 
WHERE exists((n)-[]->(m)) 
RETURN n;
```

And it returns `An unknown exception occurred, this is unexpected. The real message should be in the database logs`

The problem might be that you are checking all the scopes before if there is a symbol `m` and that shouldn't be the case. I am not 100% sure how these scopes work with `WITH` clause, but maybe it makes sense to only check recent scope. That seems like reason why it later explodes somewhere, haven't figured out where yet.

src/query/plan/rule_based_planner.cpp
antoniofilipovic: I am not sure what should happen here, and what this part of the code does exactly, but it throws the same error as the following query:
`MATCH (n) RETURN n WHERE n.prop=1;`

And this is the error:
`mismatched input 'WHERE' expecting {<EOF>, ';'`

src/query/plan/rule_based_planner.hpp
antoniofilipovic: As we will optimize later if needed, I will resolve this.

antoniofilipovic (changes requested):
I think this part with WITH query is not correct and this shouldn't happen. I took a look at the tests and they are fine from my perspective. Good job there

src/query/plan/rule_based_planner.cpp
```diff
@@ -235,22 +235,11 @@ class ReturnBodyContext : public HierarchicalTreeVisitor {
...
-  bool PreVisit(Exists &exists) override {
-    in_exists_ = true;
-    return true;
-  }
-
-  bool PostVisit(Exists &exists) override {
-    in_exists_ = false;
-    return true;
+  bool PreVisit(Exists & /*exists*/) override {
+    throw SemanticException("Exists functionality works only with matching clauses!");
```
antoniofilipovic, line 239: I see that you introduced error if you try to use exists with WITH clause. I think this shouldn't break on rule based planner but on symbol generator. Is there reason why not throw error earlier. Theoretically, this is "wrong" as you are still letting use unbounded symbols in WITH clause

antoniofilipovic: There is part of the code in VisitReturnBody which should trigger deleting variables from scope which are old:

It is probably somewhere here, I tried to research it today:
```
if ((!where && body.order_by.empty()) || scope.has_aggregation) {
    // WHERE and ORDER BY need to see both the old and new symbols, unless we
    // have an aggregation. Therefore, we can clear the symbols immediately if
    // there is neither ORDER BY nor WHERE, or we have an aggregation.
    scope.symbols.clear();
    removed_old_names = true;
  }
```

I think problem happens somewhere here

src/query/plan/rule_based_planner.cpp
antoniofilipovic: @antaljanosbenjamin could you check this part. Why should this be valid query?

src/query/frontend/semantic/symbol_generator.cpp
```diff
@@ -328,7 +344,8 @@ SymbolGenerator::ReturnType SymbolGenerator::Visit(Identifier &ident) {
     }
     symbol = GetOrCreateSymbol(ident.name_, ident.user_declared_, type);
   } else if (scope.in_pattern && !scope.in_pattern_atom_identifier && scope.in_match) {
-    if (scope.in_edge_range && scope.visiting_edge->identifier_->name_ == ident.name_) {
+    if (scope.in_edge_range && scope.visiting_edge && scope.visiting_edge->identifier_ &&
```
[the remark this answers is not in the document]
antoniofilipovic: @Josipmrden

src/query/plan/rule_based_planner.hpp
```diff
@@ -567,6 +565,66 @@ class RuleBasedPlanner {
...
+                                                         const SymbolTable &symbol_table,
+                                                         const std::unordered_set<Symbol> &bound_symbols) {
+    for (const auto &filter : filters) {
+      if (filter.type != FilterInfo::Type::Pattern) {
+        continue;
+      }
+
+      if (const auto *exists = utils::Downcast<Exists>(filter.expression)) {
+        return MakeExistsFilter(*exists, storage, symbol_table, bound_symbols);
+      }
```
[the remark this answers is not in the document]
antoniofilipovic: @Josipmrden

src/query/frontend/semantic/symbol_generator.cpp
antoniofilipovic: @Josipmrden I am sorry for giving wrong review here but important thing:
**I think this shouldn't throw and should be allowed.** More details below.

src/query/plan/rule_based_planner.cpp
antoniofilipovic: I think we are inconsistent with throwing unbounded variable errors. For example, this is allowed:
```
MATCH (n), (m) 
WITH n 
WHERE id(m)>0 
RETURN n;
```

I think this one should be to:
```
MATCH (n),(m) 
WITH n 
WHERE exists((n)-[]->(m)) 
RETURN n;
```

src/query/frontend/semantic/symbol_generator.cpp
antoniofilipovic: I am not sure if that is the reason why you decided to throw error in rule based planner, to make this MVP implementation.

src/query/plan/rule_based_planner.cpp
antoniofilipovic: I think this is resolved as we won't fix this.

src/query/frontend/semantic/symbol_generator.cpp
antoniofilipovic: Resolved as we won't address it now

antoniofilipovic (approved):
For me all good now. Good job

src/query/plan/rule_based_planner.cpp
antaljanosbenjamin: Sorry, `WHERE` cannot go after `RETURN`, I was wrong here. I phrased my comment in a wrong way. I should have asked that we are sure we cannot use `WHERE exists(...)` in "return bodies", right? `RETURN` cannot have `WHERE` after it, but `WITH` can: `MATCH (n) WITH n WHERE n.prop=1 RETURN n`. I think I remember something to protect against that, but @Josipmrden could you please confirm?

src/query/plan/rule_based_planner.hpp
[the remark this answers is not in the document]
antaljanosbenjamin: #764

antaljanosbenjamin:
@Josipmrden could you please answer my two questions?

PR #767, opened 2023-02-06
[E214 < T1231] Add WHERE EXISTS  lexer and AST

Josipmrden:
This PR concerns the grammar for the where exists functionality.

### DIFF
--- a/tests/unit/cypher_main_visitor.cpp
+++ b/tests/unit/cypher_main_visitor.cpp
@@ -4312,11 +4312,11 @@ TEST_P(CypherMainVisitorTest, ExistsThrow) {
   auto &ast_generator = *GetParam();
 
   TestInvalidQueryWithMessage<SyntaxException>("MATCH (n) WHERE exists(p=(n)-[]->()) RETURN n;", ast_generator,
-                                               "No pattern identifiers can be present in exists!");
+                                               "Identifiers are not supported in exists(...).");
   TestInvalidQueryWithMessage<SyntaxException>("MATCH (n) WHERE exists((n)-[]->()-[]->()) RETURN n;", ast_generator,
-                                               "Exists function supports only triplet patterns!");
+                                               "exists(...) supports only node-edge-node patterns.");
   TestInvalidQueryWithMessage<SyntaxException>("MATCH (n) WHERE exists((n)) RETURN n;", ast_generator,
-                                               "Exists function supports only triplet patterns!");
+                                               "exists(...) supports only node-edge-node patterns.");
 }
 
 TEST_P(CypherMainVisitorTest, Exists) {


src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -2204,6 +2207,21 @@ antlrcpp::Any CypherMainVisitor::visitLiteral(MemgraphCypher::LiteralContext *ct
   return visitChildren(ctx);
 }
 
+antlrcpp::Any CypherMainVisitor::visitExistsExpression(MemgraphCypher::ExistsExpressionContext *ctx) {
+  auto *exists = storage_->Create<Exists>();
+  exists->pattern_ = std::any_cast<Pattern *>(ctx->patternPart()->accept(this));
+
+  if (exists->pattern_->identifier_) {
+    throw SyntaxException("No pattern identifiers can be present in exists!");
```
antoniofilipovic, line 2215:
```suggestion
    throw SyntaxException("Identifiers are not supported in exists(...).");
```

src/query/frontend/opencypher/grammar/Cypher.g4
```diff
@@ -275,6 +276,8 @@ reduceExpression : accumulator=variable '=' initial=expression ',' idInColl '|'
 
 extractExpression : idInColl '|' expression ;
 
+existsExpression : patternPart ;
```
antoniofilipovic, line 279: @antaljanosbenjamin do you think we should use here `anonymousPatternPart` as we don't support `( variable '=' anonymousPatternPart ) ` in `exists`? From what I understood there would need to refactor code as there is functionality for PatternPart but not for `anonymousPatternPart`

antoniofilipovic: In my opinion, we don't need to do so as it doesn't complicate code for worse, but makes it easier now to deliver, but it is not best practice nevertheless.

src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -2204,6 +2207,21 @@ antlrcpp::Any CypherMainVisitor::visitLiteral(MemgraphCypher::LiteralContext *ct
...
+  }
+
+  if (exists->pattern_->atoms_.size() != 3) {
+    throw SyntaxException("Exists function supports only triplet patterns!");
```
antoniofilipovic, line 2219:
```suggestion
    throw SyntaxException("exists(...) supports only node-edge-node patterns.");
```

antoniofilipovic (approved):
From me there is approve, I made a comment about `anonymousPatternPart`, but it seems from my perspective there is no need to implement that now.

src/query/frontend/opencypher/grammar/Cypher.g4
antaljanosbenjamin: I would say use only what we need. In my opinion both `anonymousPatternPart` and `patternElement` is suitable here.

What is the reason to use `patternPart` instead? What extra costs does it have?

src/query/frontend/ast/cypher_main_visitor.cpp
antaljanosbenjamin: I just raise awarness here without expecting any action: There is a difference between syntax and semantic exceptions. Obviously they are for syntactical or semantical errors. However I am not sure which category do these errors fall into, so that's why I don't know whether they are correct or not.

Therefore the goal of this comment is spreading knowledge and providing information to you so you can make the best decision.

antaljanosbenjamin (approved):
Nicely done!

src/query/frontend/opencypher/grammar/Cypher.g4
antoniofilipovic: Actually exists doesn't support whole `patternPart`, only `anonymusPatternPart`. Because in `exists`, you can't write the following which is allowed pattern part: `... WHERE exists(p = (n)-[]->(m));`
and in `patternPart` this is allowed: `( variable '=' anonymousPatternPart )`

src/query/frontend/ast/cypher_main_visitor.cpp
Josipmrden: I think it's okay that we here make a syntax exception. However, if I make this an anonymous pattern part, there can not be any mistake that the user can not enter an identifier in the exists() part. However, I'd still need to check in the other parts of the code that identifier is not set since it's the same Pattern object which can contain an identifier (done when pathfinding is performed).

Josipmrden: @antaljanosbenjamin  @antoniofilipovic

src/query/frontend/ast/cypher_main_visitor.cpp
Josipmrden: updated error message, pls resolve

src/query/frontend/opencypher/grammar/Cypher.g4
Josipmrden: I decided in the end to stick to patternPart as it has better error handling capabilities if the user messes up something with patterns

Josipmrden: I.E. I can throw a detailed message if the user specifies p=()-[]-() telling him that there is no identifier for pattern in exists. Whereas if I use an anonymous pattern part, I don't handle that case but it's somewhere in the air (atom visiting)

src/query/frontend/ast/cypher_main_visitor.cpp
antoniofilipovic: For me this is okay. I will resolve the comment.

antoniofilipovic:
@Josipmrden after Release build passes we can merge this.

antepusic (approved):

PR #772, opened 2023-02-08
[master< T589-FL] Introduce analytics mode

antoniofilipovic:
Release note: You can now choose between two different storage modes:
  - Transactional mode - the default database mode that favors
    strongly-consistent ACID transactions using WAL files and snapshots,
    but requires more time and resources during data import and analysis.
  - Analytical mode - speeds up import and data analysis but offers no ACID
    guarantees besides manually created snapshots.
  Switch between modes within the session using the `STORAGE MODE
  IN_MEMORY_{TRANSACTIONAL|ANALYTICAL};` query. 

User story: TODO

[master < Task] PR


.pre-commit-config.yaml
```diff
@@ -10,7 +10,7 @@ repos:
     hooks:
       - id: black
   - repo: https://github.com/pycqa/isort
-    rev: 5.10.1
+    rev: 5.12.0
```
gitbuda, line 13: we should make a separated PR to fix this, I'll to the PR soon

CMakeLists.txt
```diff
@@ -271,7 +271,7 @@ if (MG_ENTERPRISE)
   add_definitions(-DMG_ENTERPRISE)
 endif()
 
-set(ENABLE_JEMALLOC ON)
+set(ENABLE_JEMALLOC OFF)
```
Josipmrden, line 274: why do we need to disable this

src/query/interpreter.cpp
```diff
@@ -1861,6 +1870,36 @@ PreparedQuery PrepareIsolationLevelQuery(ParsedQuery parsed_query, const bool in
...
+  const auto analytics_mode = ToStorageAnalyticsMode(analytics_mode_query->analytics_mode_);
+
+  auto callback = [analytics_mode_query, analytics_mode, interpreter_context, interpreter]() -> std::function<void()> {
+    switch (analytics_mode_query->analytics_mode_) {
+      case AnalyticsModeQuery::AnalyticsMode::ON:
+        return [interpreter_context, analytics_mode] { interpreter_context->db->SetAnalyticsMode(analytics_mode); };
+      case AnalyticsModeQuery::AnalyticsMode::OFF:
+        return [interpreter_context, analytics_mode] { interpreter_context->db->SetAnalyticsMode(analytics_mode); };
+    }
+  }();
```
Josipmrden, line 1891: but these 2 switch statements have the same behaviour? can't you make a normal callback?

src/query/interpreter.cpp
```diff
@@ -1044,7 +1044,7 @@ std::optional<plan::ProfilingStatsWithTotalTime> PullPlan::Pull(AnyStream *strea
   // Also, we want to throw only when the query engine requests more memory and not the storage
   // so we add the exception to the allocator.
   // TODO (mferencevic): Tune the parameters accordingly.
-  utils::PoolResource pool_memory(128, 1024, &monotonic_memory, utils::NewDeleteResource());
+  utils::PoolResource pool_memory(8, 1024, utils::NewDeleteResource(), utils::NewDeleteResource());
```
Josipmrden, line 1047: why change this

src/query/plan/operator.cpp
```diff
@@ -10,6 +10,7 @@
 // licenses/APL.txt.
 
 #include "query/plan/operator.hpp"
+#include <malloc.h>
```
Josipmrden, line 13: why add this, it is the only change

src/storage/v2/edge_accessor.cpp
```diff
@@ -117,6 +117,7 @@ Result<storage::PropertyValue> EdgeAccessor::SetProperty(PropertyId property, co
   // current code always follows the logical pattern of "create a delta" and
   // "modify in-place". Additionally, the created delta will make other
   // transactions get a SERIALIZATION_ERROR.
+
```
Josipmrden, line 120: delete?

Josipmrden (changes requested):
I think we should think more about the naming, since when the disk storage comes in, we could have 3 distinct possibilities (in memory transactional, in memory analytical, and disk storage transactional), while here now we have only 2 possibilities which only tell us are we in analytics mode or not.

src/auth/models.cpp
```diff
@@ -44,7 +44,9 @@ const std::vector<Permission> kPermissionsAll = {Permission::MATCH,       Permis
                                                  Permission::FREE_MEMORY, Permission::TRIGGER,
                                                  Permission::CONFIG,      Permission::STREAM,
                                                  Permission::MODULE_READ, Permission::MODULE_WRITE,
-                                                 Permission::WEBSOCKET,   Permission::TRANSACTION_MANAGEMENT};
+                                                 Permission::WEBSOCKET,   Permission::TRANSACTION_MANAGEMENT,
+                                                 Permission::STORAGE_MODE};
```
as51340, line 48: Do we want to handle changing permission modes as privileges? I think there is a bigger story behind it, with disk storage engine coming on and three Memgraph modes. For example an issue I see: we'll try to implement disk storage engine in a way that whenever Memgraph stays out of RAM it fallbacks on disk -> not sure if this and analytics mode should be handled as privilege. Maybe we can ask buda about authorization for storage modes, what do you think?

src/auth/models.hpp
```diff
@@ -41,7 +41,8 @@ enum class Permission : uint64_t {
   MODULE_READ  = 1U << 18U,
   MODULE_WRITE = 1U << 19U,
   WEBSOCKET    = 1U << 20U,
-  TRANSACTION_MANAGEMENT = 1U << 21U
+  TRANSACTION_MANAGEMENT = 1U << 21U,
+  STORAGE_MODE    = 1U << 22U
 };
 // clang-format on
```
as51340, line 47: clang format comment

src/query/frontend/stripped_lexer_constants.hpp
```diff
@@ -208,7 +208,10 @@ const trie::Trie kKeywords = {"union",
                               "websocket",
                               "foreach",
                               "labels",
-                              "edge_types"};
+                              "edge_types",
+                              "off",
+                              "in_memory_transactional",
+                              "in_memory_analytical"};
```
as51340, line 215: What does this do? Should I forgot to use it in transactional query maybe :smile: ?

src/query/interpreter.cpp
```diff
@@ -1871,6 +1879,40 @@ PreparedQuery PrepareIsolationLevelQuery(ParsedQuery parsed_query, bool in_expli
...
+PreparedQuery PrepareStorageModeQuery(ParsedQuery parsed_query, const bool in_explicit_transaction,
+                                      InterpreterContext *interpreter_context) {
+  if (in_explicit_transaction) {
+    throw StorageModeModificationInMulticommandTxException();
+  }
+
+  auto *storage_mode_query = utils::Downcast<StorageModeQuery>(parsed_query.query);
+  MG_ASSERT(storage_mode_query);
+
+  const auto storage_mode = ToStorageStorageMode(storage_mode_query->storage_mode_);
```
as51340, line 1891: Why not just ToStorageMode?

src/storage/v2/isolation_level.hpp
```diff
@@ -16,5 +16,6 @@
 namespace memgraph::storage {
 
 enum class IsolationLevel : std::uint8_t { SNAPSHOT_ISOLATION, READ_COMMITTED, READ_UNCOMMITTED };
+enum class StorageMode : std::uint8_t { IN_MEMORY_ANALYTICAL, IN_MEMORY_TRANSACTIONAL };
```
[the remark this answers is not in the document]
antoniofilipovic: Yes, I forgot to refactor it. I will store it in some other file

src/query/interpreter.cpp
```diff
@@ -1871,6 +1879,40 @@ PreparedQuery PrepareIsolationLevelQuery(ParsedQuery parsed_query, bool in_expli
...
+
+  auto callback = [storage_mode, interpreter_context]() -> std::function<void()> {
+    auto exists_active_transaction = interpreter_context->interpreters.WithLock([](const auto &interpreters_) {
+      return std::any_of(interpreters_.begin(), interpreters_.end(), [](const auto &interpreter) {
+        return interpreter->transaction_status_.load() == TransactionStatus::ACTIVE;
```
as51340, line 1896: It seems to me this is the wrong check: By discarding only interpreters that are active you are allowing changing storage mode when interpreter is in the started commit state, started rollback or verifying state and I think this can fail in many ways.

src/auth/models.cpp
antoniofilipovic: This is good thinking. I am not sure if this fits this PR, like thinking forward so much as we don't know how yet storage will be implemented. I implemented it this way consistency-wise, I think we can refactor when we implement different storage mode, as we will now what are we refactoring for. I am not sure if that makes sense.

src/query/exceptions.hpp
```diff
@@ -211,6 +211,18 @@ class IsolationLevelModificationInMulticommandTxException : public QueryExceptio
...
+class StorageModeModificationInMulticommandTxException : public QueryException {
+ public:
+  StorageModeModificationInMulticommandTxException()
+      : QueryException("Storage mode cannot be modified in multicommand transactions.") {}
+};
+
+class StorageModeModificationInMultiTxException : public QueryException {
+ public:
+  StorageModeModificationInMultiTxException()
+      : QueryException("Storage mode cannot be modified when multiple transactions are running.") {}
```
as51340, line 223: I would enrich this error message by saying what the user could do next so he/she doesn't end on Discord disturbing Kate :smile: E.g Terminate your transaction by first running show transactions and then ... or a variant of this, I'm sure you will have great ideas

src/storage/v2/transaction.hpp
```diff
@@ -31,12 +31,14 @@ const uint64_t kTimestampInitialId = 0;
 const uint64_t kTransactionInitialId = 1ULL << 63U;
 
 struct Transaction {
-  Transaction(uint64_t transaction_id, uint64_t start_timestamp, IsolationLevel isolation_level)
+  Transaction(uint64_t transaction_id, uint64_t start_timestamp, IsolationLevel isolation_level,
+              StorageMode storage_mode)
```
as51340, line 35: I understand why you need StorageMode in Transaction because of deltas but to me at least, it doesn't look like a perfect solution that the transaction knows about the StorageMode in which it operates. But I think Benjo is the perfect person to him about this so pls. double check :smile:

src/query/interpreter.cpp
antoniofilipovic: Yes, you are right about discarding other interpreters. It should be that there are no active transactions. I first put that, but then there was bug on transactions, and didn't change as I forgot.

antoniofilipovic: Global storage mode is inside storage, same way as isolation level global is in storage. But why would then mvcc need to know about storage mode or anything like that, it is not interacting with Storage object, but with transaction, right?

antoniofilipovic: > this assumes that the prior transacaction has finished because otherwise which storage mode are we actually in? 

yes, that is why we want to all transactions to be finished.

src/storage/v2/transaction.hpp
antoniofilipovic: @as51340 I think  I was wrong. It makes sense to not store in transactions storage mode, as it is only a "global" state. I can refactor this because I see what you are saying and it should be good.

src/auth/models.cpp
as51340: I think it makes sense to revise that later :+1:

src/storage/v2/transaction.hpp
as51340: Great, I think it would be best to hold storage mode just in one place because of simplicity, that's the main reason so if you can refactor this and you think it makes sense :+1:

tests/unit/storage_v2_analytics_mode.cpp
```diff
@@ -0,0 +1,71 @@
...
+}
+
+inline constexpr std::array storage_modes{
+    memgraph::storage::StorageMode::IN_MEMORY_ANALYTICAL,
+    memgraph::storage::StorageMode::IN_MEMORY_TRANSACTIONAL,
+};
+
+}  // namespace
+
+std::string_view StorageModeToString(memgraph::storage::StorageMode storage_mode) {
```
as51340, line 25: Completely unimportant but as a side note what I learnt yesterday, maybe you will find it interesting. string_view can be constexpr while string cannot in c++17. This will change in c++20

tests/unit/storage_v2_analytics_mode.cpp
```diff
@@ -0,0 +1,71 @@
...
+    }
+  };
+};
+
+// you should be able to see nodes if there is analytics mode
+TEST_P(StorageModeTest, Mode) {
+  const memgraph::storage::StorageMode storage_mode = GetParam();
+
+  memgraph::storage::Storage storage{
+      {.transaction{.isolation_level = memgraph::storage::IsolationLevel::SNAPSHOT_ISOLATION}}};
```
as51340, line 48: I think this test is good. What do you think about simple extension that you also test read_committed? I think logically is not important but maybe as an idea

tests/unit/storage_v2_wal_file.cpp
```diff
@@ -20,8 +20,10 @@
 #include "storage/v2/durability/exceptions.hpp"
 #include "storage/v2/durability/version.hpp"
 #include "storage/v2/durability/wal.hpp"
+#include "storage/v2/isolation_level.hpp"
```
as51340, line 23: I think it should be possible to compile without these include statements

tests/unit/storage_v2_wal_file.cpp
```diff
@@ -621,3 +635,80 @@ TEST_P(WalFileTest, PartialData) {
   ASSERT_EQ(pos, infos.size() - 2);
   AssertWalInfoEqual(infos[infos.size() - 1].second, memgraph::storage::durability::ReadWalInfo(current_file));
 }
+
+std::string_view StorageModeToString(memgraph::storage::StorageMode storage_mode) {
```
as51340, line 639: This code is duplication. It seems to me like a good idea to extract it and put it under storage_test_utils or some file similar to this.

src/query/interpreter.cpp
as51340: Good point, I think mvcc shouldn't know also about specific details of the storage mode but since you wrote in one other comment that you figured out this, I assume you found a way to refactor this?

tests/unit/storage_v2_wal_file.cpp
```diff
@@ -20,8 +20,10 @@
 #include "storage/v2/durability/exceptions.hpp"
 #include "storage/v2/durability/version.hpp"
 #include "storage/v2/durability/wal.hpp"
+#include "storage/v2/isolation_level.hpp"
 #include "storage/v2/mvcc.hpp"
 #include "storage/v2/name_id_mapper.hpp"
+#include "storage/v2/storage.hpp"
```
as51340, line 26: check comment above

src/query/frontend/stripped_lexer_constants.hpp
as51340: Okay, and when do we add words here?

as51340 (commented):
High level question: Analytics mode is not implemeted as a  READ_UNCOMMITTED isolation level because we completely want to avoid using memory for deltas? Because okay this makes sense to me but then I am not sure what is the purpose of READ_UNCOMMITTED in the Memgraph? The only thing I can think of is WAL?

as51340 (commented):
Regarding READ_UNCOMMITTED: Maybe when user set the analytics mode, it would be good to log a message like: hey you can know expect semantic similar to when using READ_UNCOMMITTED.

src/query/interpreter.cpp
as51340, line 1896: interpreter_context contains all interpreters, including the one executing storage mode. If I am not wrong, you should include a check for yourself because this interpreter will certainly be active.

as51340 (changes requested):
Regarding READ_UNCOMMITTED: Maybe when user set the analytics mode, it would be good to log a message like: hey you can know expect semantic similar to when using READ_UNCOMMITTED.

If I got it correctly, the only difference between READ_UNCOMMITTED and ANALYTICS mode is that in ANALYTICS mode we don't store deltas, hence no WAL, so the user can recover data only by doing explicit snapshot?

I would suggest adding e2e tests for:
- checking that isn't possible to change storage mode when some interpreter is active
- checking that the user can recover data by doing explicit snapshot when in analytics mode
- forbid changing isolation level when in analytics mode

src/query/interpreter.cpp
```diff
@@ -1838,7 +1837,16 @@ constexpr auto ToStorageIsolationLevel(const IsolationLevelQuery::IsolationLevel
...
+constexpr auto ToStorageStorageMode(const StorageModeQuery::StorageMode storage_mode) noexcept {
+  switch (storage_mode) {
+    case StorageModeQuery::StorageMode::IN_MEMORY_TRANSACTIONAL:
+      return storage::StorageMode::IN_MEMORY_TRANSACTIONAL;
+    case StorageModeQuery::StorageMode::IN_MEMORY_ANALYTICAL:
+      return storage::StorageMode::IN_MEMORY_ANALYTICAL;
+  }
+}
+
+PreparedQuery PrepareIsolationLevelQuery(ParsedQuery parsed_query, const bool in_explicit_transaction,
```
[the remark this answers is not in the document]
antoniofilipovic: My idea was that changing the isolation level is not an "illegal" thing to do, so with proper documentation in form of: "Hey, you can change the isolation level, but it won't have any effect", I think we would be good to go. I talked with Buda and Mrma, and @gitbuda  is fond of the option to throw error. I am not sure 100% that error is a good option, as it is not something we can recover from (this changing state). But for sake of user experience, I think it is the best option, so I will add throwing of error here.

src/query/interpreter.cpp
```diff
@@ -1871,6 +1879,40 @@ PreparedQuery PrepareIsolationLevelQuery(ParsedQuery parsed_query, bool in_expli
...
+    auto exists_active_transaction = interpreter_context->interpreters.WithLock([](const auto &interpreters_) {
+      return std::any_of(interpreters_.begin(), interpreters_.end(), [](const auto &interpreter) {
+        return interpreter->transaction_status_.load() == TransactionStatus::ACTIVE;
+      });
+    });
+    if (exists_active_transaction) {
+      throw StorageModeModificationInMultiTxException();
+    }
+
+    return [interpreter_context, storage_mode] { interpreter_context->db->SetStorageMode(storage_mode); };
```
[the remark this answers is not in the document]
antoniofilipovic: Yes, you are right, I didn't think about that. Good catch. I will move this inside Lock part, so that we set state when we have lock. 

I think we should have implemented a transactional queue with locks and not atomics :joy:

src/query/interpreter.cpp
antoniofilipovic: I don't know if you remember, but we had a discussion about this during your development of the transactional queue. Since SetStorageModeQuery is not a Cypher query or Transaction query, the interpreter won't be active. So it won't be active.

as51340: Yes, you are right I forgot about that :+1:

antoniofilipovic:
> Regarding READ_UNCOMMITTED: Maybe when user set the analytics mode, it would be good to log a message like: hey you can know expect semantic similar to when using READ_UNCOMMITTED.
> 
> If I got it correctly, the only difference between READ_UNCOMMITTED and ANALYTICS mode is that in ANALYTICS mode we don't store deltas, hence no WAL, so the user can recover data only by doing explicit snapshot?

Besides that you can read uncommitted, the difference is that you can also let multiple transactions update the state of the same entity (node or edge).



> I would suggest adding e2e tests for:
> 
> * checking that isn't possible to change storage mode when some interpreter is active
> * checking that the user can recover data by doing explicit snapshot when in analytics mode

Good idea as I  had these two in the plan.
> * forbid changing isolation level when in analytics mode

When I add throwing of error, I can add that too.

tests/unit/storage_v2_analytics_mode.cpp
antoniofilipovic: I think it is a bonus, as it doesn't have any impact. Because if it works on snapshot isolation, should also work on other states. I would leave it as it is now. What do you think?

src/query/interpreter.cpp
antoniofilipovic: I thought about it even more, and I again I am not sure, because then really the problem is back again that mvcc needs to get reference somehow to storage, and that is problem because it shouldn't have that reference. I would maybe leave it as it is. 

I can check if there is some workaround, but when you think about it, transactions are the ones carrying information about what should transactions see, and it shouldn't be that mvcc is also looking at storage.

src/storage/v2/transaction.hpp
antoniofilipovic: I think I was wrong, I left comment above, let me know what do you think.

src/query/exceptions.hpp
antoniofilipovic: Nice!

src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -470,6 +470,20 @@ antlrcpp::Any CypherMainVisitor::visitIsolationLevelQuery(MemgraphCypher::Isolat
...
 
+antlrcpp::Any CypherMainVisitor::visitStorageModeQuery(MemgraphCypher::StorageModeQueryContext *ctx) {
+  auto *storage_mode_query = storage_->Create<StorageModeQuery>();
+
+  storage_mode_query->storage_mode_ = [mode = ctx->storageMode()]() {
+    if (mode->IN_MEMORY_ANALYTICAL()) {
+      return StorageModeQuery::StorageMode::IN_MEMORY_ANALYTICAL;
+    }
+    return StorageModeQuery::StorageMode::IN_MEMORY_TRANSACTIONAL;
+  }();
```
antaljanosbenjamin, line 481:
```suggestion
  storage_mode_query->storage_mode_ = std::invoke([mode = ctx->storageMode()]() {
    if (mode->IN_MEMORY_ANALYTICAL()) {
      return StorageModeQuery::StorageMode::IN_MEMORY_ANALYTICAL;
    }
    return StorageModeQuery::StorageMode::IN_MEMORY_TRANSACTIONAL;
  });
```

Just to increase readability.

src/query/frontend/opencypher/grammar/MemgraphCypherLexer.g4
```diff
@@ -67,6 +67,8 @@ HEADER              : H E A D E R ;
 IDENTIFIED          : I D E N T I F I E D ;
 IGNORE              : I G N O R E ;
 ISOLATION           : I S O L A T I O N ;
+IN_MEMORY_ANALYTICAL    : I N UNDERSCORE M E M O R Y UNDERSCORE A N A L Y T I C A L ;
+IN_MEMORY_TRANSACTIONAL : I N UNDERSCORE M E M O R Y UNDERSCORE T R A N S A C T I O N A L ;
```
antaljanosbenjamin, line 71: Just the question: don't we want to kepp the straigth line with the colons?

src/query/frontend/stripped_lexer_constants.hpp
antaljanosbenjamin: @as51340 if you added a new keyword, then you have to also modify this container.

src/query/exceptions.hpp
```diff
@@ -211,6 +211,26 @@ class IsolationLevelModificationInMulticommandTxException : public QueryExceptio
       : QueryException("Isolation level cannot be modified in multicommand transactions.") {}
 };
 
+class IsolationLevelModificationInAnalyticsException : public QueryException {
+ public:
+  IsolationLevelModificationInAnalyticsException()
+      : QueryException("Isolation level cannot be modified when storage mode is set to IN_MEMORY_ANALYTICAL.") {}
```
gitbuda, line 217: Append: "... set to IN_MEMORY_ANALYTICAL. IN_MEMORY_ANALYTICAL mode doesn't provide any isolation guarantees, you can think about it as an equivalent to READ_UNCOMMITED."

src/query/exceptions.hpp
```diff
@@ -211,6 +211,26 @@ class IsolationLevelModificationInMulticommandTxException : public QueryExceptio
...
+  StorageModeModificationInMulticommandTxException()
+      : QueryException("Storage mode cannot be modified in multicommand transactions.") {}
+};
+
+class StorageModeModificationInMultiTxException : public QueryException {
+ public:
+  StorageModeModificationInMultiTxException()
+      : QueryException(
+            "Storage mode cannot be modified when multiple transactions are running. Check the status of the "
+            "transactions using 'SHOW TRANSACTIONS' query.") {}
```
gitbuda, line 231: Append: "and ensure no other transactions are active while you attempt to switch the storage mode."

src/storage/v2/storage.cpp
```diff
@@ -2119,6 +2136,10 @@ void Storage::SetIsolationLevel(IsolationLevel isolation_level) {
   isolation_level_ = isolation_level;
 }
 
+void Storage::SetStorageMode(StorageMode storage_mode) { storage_mode_ = storage_mode; }
```
antaljanosbenjamin, line 2139: I think this should acquire `main_lock_` similarly to `SetIsolationLevel`.

src/storage/v2/storage.cpp
```diff
@@ -1918,7 +1933,9 @@ utils::BasicResult<Storage::CreateSnapshotError> Storage::CreateSnapshot() {
   std::shared_lock<utils::RWLock> storage_guard(main_lock_);
 
   // Create the transaction used to create the snapshot.
-  auto transaction = CreateTransaction(IsolationLevel::SNAPSHOT_ISOLATION);
+  // We don't care about storage mode level here, since snapshot only reads data
+  // We can only add log about potential dirty reads when creating snapshot
+  auto transaction = CreateTransaction(IsolationLevel::SNAPSHOT_ISOLATION, storage_mode_);
 
   // Create snapshot.
```
antaljanosbenjamin, line 1940: This might cause some snapshots be inconsistent, e.g.: some dandling edges or not symmetric egdes (one end of the edge contains the edge, while the other doesn't). So I would say this can cause serious issues in my opinion.

src/query/interpreter.cpp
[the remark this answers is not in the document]
antaljanosbenjamin: Also, it might happen that an already checked interpreter starts to execute another transaction, so this is only a best effort thing to do. The storage should acquire the main lock, so it would ensure that no transactions are active in the database.

src/storage/v2/storage.cpp
antoniofilipovic: Because analytics mode is already on the polar side of isolations, I think this is not the place where we need to take care of that. The only thing we can do is disable creating snapshots while other transactions are active in the case of analytics mode.

src/query/frontend/opencypher/grammar/MemgraphCypherLexer.g4
antoniofilipovic: But to answer your question, we failed to do so with transactional management keyword, updated it

src/storage/v2/storage.cpp
antaljanosbenjamin: > Yes, it can cause issues. In my opinion, we can document this, and enable users to do how they like here.

I think it is not that simple: basically the users can screw up their dataset:
1. Start to ingest data
2. periodic snapshots starts to kick in and in creates a corrupted snapshot
3. Memgraph crashes
4. User restarts Memgraph and voilà, they have a dataset that is corrupted but they don't know about it.

> The only thing we can do is disable creating snapshots while other transactions are active in the case of analytics mode.

I think that is the easiest solution, however then how can our users use snapshots in analytics mode? Maybe triggering it manually and doing it in a blocking manner? I am not aware of the possible use-cases, so these are just my input for somebody, who can make a decision.

src/storage/v2/storage.cpp
antoniofilipovic: I added main lock on check here

src/query/interpreter.cpp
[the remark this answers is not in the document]
antoniofilipovic: You are right here @antaljanosbenjamin added this. @as51340 check Benjo's comment and I added this check.

src/storage/v2/storage.cpp
antoniofilipovic: I think WRITE LOCK when in IN_MEMORY_ANALYTICAL mode is one solution. I will add that and we can discuss whether it makes sense.

tests/unit/storage_v2_analytics_mode.cpp
as51340: Your call, I agree that it shouldn't create problems.

src/query/exceptions.hpp
```diff
@@ -211,6 +211,21 @@ class IsolationLevelModificationInMulticommandTxException : public QueryExceptio
       : QueryException("Isolation level cannot be modified in multicommand transactions.") {}
 };
 
+class IsolationLevelModificationInAnalyticsException : public QueryException {
+ public:
+  IsolationLevelModificationInAnalyticsException()
+      : QueryException(
+            "Isolation level cannot be modified when storage mode is set to IN_MEMORY_ANALYTICAL."
+            "IN_MEMORY_ANALYTICAL mode doesn't provide any isolation guarantees, "
+            "you can think about it as an equivalent to READ_UNCOMMITED.") {}
```
as51340, line 220: Sorry for being picky but I think this last sentence should be a bit changed because I find think it is a bit weird to have "you can think about" in the exception message. I think "IN_MEMORY_ANALYTICAL mode doesn't provide any isolation guarantees. " is enough. Or: ... provide any iso guarantees, it is an equivalent to READ_UNCOMMITTED, to avoid explicit message to user.

src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -470,6 +470,20 @@ antlrcpp::Any CypherMainVisitor::visitIsolationLevelQuery(MemgraphCypher::Isolat
   return isolation_level_query;
 }
 
+antlrcpp::Any CypherMainVisitor::visitStorageModeQuery(MemgraphCypher::StorageModeQueryContext *ctx) {
+  auto *storage_mode_query = storage_->Create<StorageModeQuery>();
+
+  storage_mode_query->storage_mode_ = std::invoke([mode = ctx->storageMode()]() {
```
as51340, line 476: Nice!

src/query/interpreter.cpp
```diff
@@ -1971,12 +1970,25 @@ constexpr auto ToStorageIsolationLevel(const IsolationLevelQuery::IsolationLevel
...
                                          InterpreterContext *interpreter_context, Interpreter *interpreter) {
   if (in_explicit_transaction) {
     throw IsolationLevelModificationInMulticommandTxException();
   }
 
+  if (interpreter_context->db->GetStorageMode() == storage::StorageMode::IN_MEMORY_ANALYTICAL) {
+    throw IsolationLevelModificationInAnalyticsException();
+  }
+
   auto *isolation_level_query = utils::Downcast<IsolationLevelQuery>(parsed_query.query);
```
Josipmrden, line 1992: what if after this if user concurrently modifies storage mode to analytical?

Josipmrden (commented):
one comment

src/query/interpreter.cpp
antoniofilipovic: Nothing will break, but user will succeed in attempt, and when changes back to in memory transactional, it will be in that state. This is something that @as51340 suggested. I am not sure if we should sync this part, but I can move this check inside changing isolation level, and then we are sure.

antoniofilipovic: Good catch.

as51340: I think it's good as it is currently since Storage object isn't used in this method. Which check would you move where @antoniofilipovic, I don't get it?

src/storage/v2/storage.cpp
```diff
@@ -1907,27 +1931,47 @@ bool Storage::AppendToWalDataDefinition(durability::StorageGlobalOperation opera
...
+                               config_.durability.snapshot_retention_count, &vertices_, &edges_, &name_id_mapper_,
+                               &indices_, &constraints_, config_.items, uuid_, epoch_id_, epoch_history_,
+                               &file_retainer_);
+    // Finalize snapshot transaction.
+    commit_log_->MarkFinished(transaction.start_timestamp);
+  };
 
-  // Create the transaction used to create the snapshot.
-  auto transaction = CreateTransaction(IsolationLevel::SNAPSHOT_ISOLATION);
+  if (storage_mode_ == memgraph::storage::StorageMode::IN_MEMORY_ANALYTICAL) {
```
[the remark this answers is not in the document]
antaljanosbenjamin: Atomics cannot guarantee that the value of the atomic is not changed for longer period in my opinion, therefore any read access to `storage_mode_` has to be protected with a lock if we want to assume that its value remains the same.

src/query/interpreter.cpp
antoniofilipovic: @as51340 I might be wrong but I think it is not good as it is, because I am still using storage_mode_ (GetStorageMode). This check should be moved to `SetIsolationLevel`

src/storage/v2/storage.cpp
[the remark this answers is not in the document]
antoniofilipovic: Yes, I agree on this. Changed it and added maximum try count, see if that is okay

src/query/interpreter.cpp
```diff
@@ -2074,6 +2074,11 @@ PreparedQuery PrepareCreateSnapshotQuery(ParsedQuery parsed_query, bool in_expli
               spdlog::warn(utils::MessageWithLink("Periodic snapshots are disabled for analytical mode.",
                                                   "https://memgr.ph/replication"));
               break;
+            case storage::Storage::CreateSnapshotError::ReachedMaxNumTries:
+              spdlog::warn(utils::MessageWithLink(
+                  "Failed to create snapshot. Reached max number of tries. Please contact support",
+                  "https://memgr.ph/replication"));
```
antaljanosbenjamin, line 2080: Is the link correct here?

antaljanosbenjamin (approved):

Josipmrden (approved):

as51340 (approved):

antoniofilipovic:
@vpavicic release notes, as you wrote:

You can now choose between two different storage modes:
  - Transactional mode - the default database mode that favors
    strongly-consistent ACID transactions using WAL files and snapshots,
    but requires more time and resources during data import and analysis.
  - Analytical mode - speeds up import and data analysis but offers no ACID
    guarantees besides manually created snapshots.
  Switch between modes within the session using the `STORAGE MODE
  IN_MEMORY_{TRANSACTIONAL|ANALYTICAL};` query.

src/query/interpreter.cpp
```diff
@@ -2043,11 +2094,18 @@ PreparedQuery PrepareCreateSnapshotQuery(ParsedQuery parsed_query, bool in_expli
...
-        if (auto maybe_error = interpreter_context->db->CreateSnapshot(); maybe_error.HasError()) {
+        if (auto maybe_error = interpreter_context->db->CreateSnapshot({}); maybe_error.HasError()) {
           switch (maybe_error.GetError()) {
             case storage::Storage::CreateSnapshotError::DisabledForReplica:
               throw utils::BasicException(
                   "Failed to create a snapshot. Replica instances are not allowed to create them.");
+            case storage::Storage::CreateSnapshotError::DisabledForAnalyticsPeriodicCommit:
+              spdlog::warn(utils::MessageWithLink("Periodic snapshots are disabled for analytical mode.",
+                                                  "https://memgr.ph/replication"));
+              break;
```
antoniofilipovic, line 2105: This

Issue #612, opened 2022-10-23
dateTime() function is missing for handling datetime values with a timezone


david-sitsky:
I wrote about this issue here: https://discourse.memgraph.com/t/support-for-datetime-function-in-cypher/327/2 but it was suggested I report it here.

I have been trying out memgraph and I have Kafka records written in JSON format. I have been able to import the data successfully and have a “transformation” that reads it and runs the Cypher query per Kafka record successfully to create my graph data.

I have JSON properties written in a common “ISO date-time” format, for example:

2003-01-16T06:59:32.000Z

I can see memgraph’s Cypher has a localDateTime() function, but not a dateTime() function like in Neo4J. Is there a reason for this?

katarinasupe:
Thank you @david-sitsky for opening the issue here 😄

gitbuda:
@david-sitsky when we designed the feature, all our use cases but also the current ones don't need the timezone part (I think that's the issue here?) + we actually have an optimization on the storage side which is possible because of that. I'm not saying it's not possible to add another data type, but would like to understand the actual need first 😄 

Is it possible from your perspective to use the `localDateTime()` as an alternative to the `dateTime()`?

Thanks for reporting and let's figure out what's the best here 😄

david-sitsky:
Hi Marko - worst case I can use localDateTime() for now.

Josipmrden:
Can I close this @gitbuda ?

PrivateGER:
The thing I'm facing here is that it's impossible to migrate Neo4j data that used DateTime() without an external python convert script.

This issue should *really* be mentioned in your https://memgraph.com/blog/how-to-migrate-from-neo4j-to-memgraph article.

katarinasupe:
[...]

PR #773, opened 2023-02-08
[master < ] Add cypherl transform script

gitbuda:
[master < Task] PR
- [x] Enhance the user experience, mostly colors!
- [x] Implement the test script and test the script 😂

--> Here we can merge

- [x] Test with `mgconsole` batch import mode


katarinasupe:
Consider adding a potential fix for the issue https://github.com/memgraph/memgraph/issues/612 also here.

gitbuda:
> Consider adding a potential fix for the issue #612 also here.

I think #612 is a different concern, so I will merge this as is. We'll close/resolve #612 issue at a different place.

katarinasupe:
@gitbuda do you remember if you used any [specific options](https://neo4j.com/labs/apoc/4.1/export/cypher/#export-cypher-config) for export to obtain the cypherl file?

katarinasupe:
`{format: "plain"}` works, without it mgconsole prints the following error:
`Client received query exception: Index manipulation not allowed in multicommand transactions.`

Neither works in Memgraph Lab (cypherl import). This is an error if no config is provided:
`Invalid automatic transaction context: There is a transaction handler command BEGIN; within the query list. If you wish to create a manual transaction context, start with a BEGIN and end the query list with a COMMIT or a ROLLBACK.`
This is the error if plain format config is provided: 
Query should either create or update something, or return results!

katarinasupe:
It is not working as fast as expected. The issue happens because of indexes. Besides indexes for every label/node (or instead), \`UNIQUE IMPORT LABEL\`(\`UNIQUE IMPORT ID\`) index should be created for faster import. Example query in exported cypherl:

```
CREATE (:Driver`:UNIQUE IMPORT LABEL` {id:"D_1035", name:"Driver_19", status:"off-duty", `UNIQUE IMPORT ID`:0});
````

When I added CREATE INDEX ON :\`UNIQUE IMPORT LABEL\`(\`UNIQUE IMPORT ID\`) line before relationship import, it speeds up.

katarinasupe:
> `{format: "plain"}` works, without it mgconsole prints the following error: `Client received query exception: Index manipulation not allowed in multicommand transactions.`
> 
> Neither works in Memgraph Lab (cypherl import). This is an error if no config is provided: `Invalid automatic transaction context: There is a transaction handler command BEGIN; within the query list. If you wish to create a manual transaction context, start with a BEGIN and end the query list with a COMMIT or a ROLLBACK.` This is the error if plain format config is provided: `Query should either create or update something, or return results! `

This happens because without `{format: "plain"}` option, everything is wrapped in begin and commit and in the script, these are just translated into Memgraph BEGIN and COMMIT which is not okay in index and/or Memgraph Lab case.

PR #790, opened 2023-02-17
[master < T577] Transactional queue

as51340:
[master < Epic] PR
- [ ] Write E2E tests


as51340:
Few comments so it is clear what is happening, for me and for reviewers. To expose transactions, we have to somehow save all interpreters. The first idea was to save `BoltSessions`  so in the future we could add killing by session id. However, we decided not to use this approach because of complications that are occurring in the implementation (mostly circular dependencies issue).

Interpreters are saved as a raw pointer in a vector synced with a `SpinLock`. This should suffice in securing that multiple threads cannot modify interpreter object. Storing a raw pointer is not great but I think there shouldn't be a problem if we are deregistering it from the vector of interpreters in the destructor of the `BoltSession` since interpreter is completely managed within the `BoltSession`. Streams are covered so that at the beginning of the `CreateConsumer` method, interpreter is registered and at exit, deregistered. Interpreter may be active without a running transaction so to cover this case, `is_transaction_active` from storage accessor is used.

src/query/interpreter.cpp
```diff
@@ -2425,11 +2564,25 @@ Interpreter::PrepareResult Interpreter::Prepare(const std::string &query_string,
   }
 }
 
+std::vector<std::string> Interpreter::GetQueriesSummary() {
+  if (in_explicit_transaction_) {
+    return {"EXPLICIT TRANSACTION"};
+  }
+  std::vector<std::string> queries;
+  for (const auto &query_exec_ptr : query_executions_) {
+    if (query_exec_ptr) queries.push_back(query_exec_ptr->query);
```
Josipmrden, line 2573: should this ever be null?

src/query/interpreter.cpp
```diff
@@ -2425,11 +2564,25 @@ Interpreter::PrepareResult Interpreter::Prepare(const std::string &query_string,
...
+}
+
+void Interpreter::AbortTransactionByUser() { db_accessor_->AbortTransactionByUser(); }
+
 void Interpreter::Abort() {
   expect_rollback_ = false;
   in_explicit_transaction_ = false;
   if (!db_accessor_) return;
   db_accessor_->Abort();
+  // db_accessor_->AbortTransaction();
```
Josipmrden, line 2585: delete if not used

src/query/interpreter.cpp
```diff
@@ -1909,6 +1918,119 @@ PreparedQuery PrepareSettingQuery(ParsedQuery parsed_query, const bool in_explic
...
+  }
+
+  return callback;
+}
+
+PreparedQuery PrepareTransactionQueueQuery(ParsedQuery parsed_query, const std::optional<std::string> &username,
+                                           const uint64_t current_transaction_id, const bool in_explicit_transaction,
+                                           InterpreterContext *interpreter_context, DbAccessor *dba) {
+  if (in_explicit_transaction) {
+    // throw SettingConfigInMulticommandTxException{};
```
Josipmrden, line 2009: what to do with this piece?

src/query/interpreter.cpp
```diff
@@ -1909,6 +1918,119 @@ PreparedQuery PrepareSettingQuery(ParsedQuery parsed_query, const bool in_explic
...
+  }
+
+  spdlog::info("Number of interpreters in prepare transactions: {}", interpreter_context->interpreters->size());
```
Josipmrden, line 2012: people do not know what an interpreter is, either delete or put an appropriate message, possibly in debug mode

src/query/interpreter.cpp
```diff
@@ -1909,6 +1918,119 @@ PreparedQuery PrepareSettingQuery(ParsedQuery parsed_query, const bool in_explic
...
+                         if (UNLIKELY(!pull_plan)) {
+                           pull_plan = std::make_shared<PullPlanVector>(callback_fn());
+                         }
+
+                         if (pull_plan->Pull(stream, n)) {
+                           return QueryHandlerResult::COMMIT;
+                         }
+                         return std::nullopt;
+                       },
+                       RWType::NONE};
```
Josipmrden, line 2031: please review this part with Benjo or someone more experienced, I'm not sure what's a PullPlanVector

Josipmrden: or UNLIKELY

src/query/interpreter.cpp
```diff
@@ -1909,6 +1918,119 @@ PreparedQuery PrepareSettingQuery(ParsedQuery parsed_query, const bool in_explic
   // NOLINTNEXTLINE(clang-analyzer-cplusplus.NewDeleteLeaks)
 }
 
+Callback HandleTransactionQueueQuery(TransactionQueueQuery *transaction_query,
```
Josipmrden, line 1921: can we create a TransactionQueueHandler to offload some of the lines from Callback to the object?

Josipmrden: see ReplQueryHandler

src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -883,6 +895,39 @@ antlrcpp::Any CypherMainVisitor::visitShowSettings(MemgraphCypher::ShowSettingsC
...
+}
+
+antlrcpp::Any CypherMainVisitor::visitTerminateTransactions(MemgraphCypher::TerminateTransactionsContext *ctx) {
+  auto *terminator = storage_->Create<TransactionQueueQuery>();
+  terminator->action_ = TransactionQueueQuery::Action::TERMINATE_TRANSACTIONS;
+  // terminator->transaction_id_ = std::any_cast<Expression *>(ctx->transactionId()->accept(this));
+  // std::vector<std::string> transaction_ids{};
+  // for (auto *transaction_id : ctx->transactionIds()->transactionId()) {
+  //   transaction_ids.push_back(std::any_cast<std::string>(transaction_id));
+  // }
```
antoniofilipovic, line 926: You have commented code here, so please remove it

src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -631,6 +634,15 @@ void GetTopicNames(auto &destination, MemgraphCypher::TopicNamesContext *topic_n
...
   }
 }
+
+// std::vector<std::string> GetTransactionIds(MemgraphCypher::TransactionIdsContext *context) {
+//   std::vector<std::string> transaction_ids{};
+//   for (auto *transaction_id : context->transactionId()) {
+//     transaction_ids.push_back(transaction_id->accept(conte));
+//   }
+//   return transaction_ids;
+// }
```
antoniofilipovic, line 644: And here too

src/query/frontend/ast/cypher_main_visitor.hpp
```diff
@@ -358,6 +358,23 @@ class CypherMainVisitor : public antlropencypher::MemgraphCypherBaseVisitor {
    */
   antlrcpp::Any visitShowSettings(MemgraphCypher::ShowSettingsContext *ctx) override;
 
+  antlrcpp::Any visitTransactionIdList(MemgraphCypher::TransactionIdListContext *ctx) override;
```
antoniofilipovic, line 361: Add return type comment above

antoniofilipovic:
Please merge the main into epic E578-MG and merge epic E578-MG into this branch T577-FL, because there is code from PR of ReferenceBasedEvaluator here.

src/query/interpreter.cpp
```diff
@@ -1909,6 +1918,119 @@ PreparedQuery PrepareSettingQuery(ParsedQuery parsed_query, const bool in_explic
...
+  Callback callback;
+  switch (transaction_query->action_) {
+    case TransactionQueueQuery::Action::SHOW_TRANSACTIONS: {
+      callback.header = {"username", "transaction_id", "queries summary"};
+      callback.fn = [interpreter_context, username, isAdmin]() mutable {
+        std::vector<std::vector<TypedValue>> results;
+        results.reserve(interpreter_context->interpreters->size());
+        interpreter_context->interpreters.WithLock([&results, username, isAdmin](const auto &interpreters) {
+          spdlog::info("Number of interpreters in callback: {}", interpreters.size());
+          for (Interpreter *interpreter : interpreters) {
```
Josipmrden, line 1943: const?

src/query/interpreter.cpp
```diff
@@ -1909,6 +1918,119 @@ PreparedQuery PrepareSettingQuery(ParsedQuery parsed_query, const bool in_explic
...
+  evaluation_context.timestamp = QueryTimestamp();
+  evaluation_context.parameters = parameters;
+  ExpressionEvaluator evaluator(&frame, symbol_table, evaluation_context, db_accessor, storage::View::OLD);
+
+  bool isAdmin = interpreter_context->auth_checker->IsUserAdmin(username);
+
+  Callback callback;
+  switch (transaction_query->action_) {
+    case TransactionQueueQuery::Action::SHOW_TRANSACTIONS: {
+      callback.header = {"username", "transaction_id", "queries summary"};
```
Josipmrden, line 1937: maybe better to have header like query_summary.
And shouldn't it be singular, because it applies for a row, not the whole data frame?

src/query/db_accessor.hpp
```diff
@@ -316,6 +316,10 @@ class DbAccessor final {
 
   void FinalizeTransaction() { accessor_->FinalizeTransaction(); }
 
+  std::atomic<bool> *TransactionAbortedByUser() { return accessor_->TransactionAbortedByUser(); }
+
+  bool IsTransactionActive() { return accessor_->IsTransactionActive(); }
```
antoniofilipovic, line 321: You need to extend also SubgraphDBAccessor with these two functions. SubgraphDBAccessor is used in case when we use `project` and in that case we will call from mg_procedure_impl IsTransactionActive

src/query/interpreter.cpp
Josipmrden: why is it called summary, maybe it's a bit misleading because we're passing the actual query string there?

src/query/interpreter.cpp
```diff
@@ -1909,6 +1918,119 @@ PreparedQuery PrepareSettingQuery(ParsedQuery parsed_query, const bool in_explic
...
+    }
+    case TransactionQueueQuery::Action::TERMINATE_TRANSACTIONS: {
+      std::vector<std::string> transaction_ids;
+      std::transform(transaction_query->transaction_id_list_.begin(), transaction_query->transaction_id_list_.end(),
+                     std::back_inserter(transaction_ids), [&evaluator](Expression *expression) {
+                       return std::string(expression->Accept(evaluator).ValueString());
+                     });
+      callback.header = {"transaction_id", "status"};
+      callback.fn = [interpreter_context, db_accessor, transaction_ids, username, isAdmin]() mutable {
+        spdlog::info("Number of transactions to kill in callback: {}", transaction_ids.size());
```
Josipmrden, line 1970: also a message for debugging, put to debug and/or rewrite meaning

src/query/interpreter.cpp
```diff
@@ -1909,6 +1918,119 @@ PreparedQuery PrepareSettingQuery(ParsedQuery parsed_query, const bool in_explic
...
+          std::string status{"Transaction not found"};
+          interpreter_context->interpreters.WithLock(
+              [db_accessor, &transaction_id, &status, username, isAdmin](const auto &interpreters) {
+                for (Interpreter *interpreter : interpreters) {
+                  std::string intr_trans_str = std::to_string(interpreter->GetTransactionId());
+                  if (interpreter->HasActiveTransaction() && transaction_id == intr_trans_str) {
+                    if (interpreter->username_ == username || isAdmin) {
+                      // if (interpreter->in_explicit_transaction_) {
+                      //   interpreter->expect_rollback_ = true;
+                      // } else {
```
Josipmrden, line 1982: what's thisss

src/query/interpreter.cpp
```diff
@@ -1909,6 +1918,119 @@ PreparedQuery PrepareSettingQuery(ParsedQuery parsed_query, const bool in_explic
...
+      };
+      break;
+    }
+    case TransactionQueueQuery::Action::TERMINATE_TRANSACTIONS: {
+      std::vector<std::string> transaction_ids;
+      std::transform(transaction_query->transaction_id_list_.begin(), transaction_query->transaction_id_list_.end(),
+                     std::back_inserter(transaction_ids), [&evaluator](Expression *expression) {
+                       return std::string(expression->Accept(evaluator).ValueString());
+                     });
+      callback.header = {"transaction_id", "status"};
```
Josipmrden, line 1968: maybe put also a bool, like success = true/false, so you can retrieve it from gqlalchemy and be sure that the process has started.

src/query/interpreter.cpp
```diff
@@ -1909,6 +1918,119 @@ PreparedQuery PrepareSettingQuery(ParsedQuery parsed_query, const bool in_explic
...
+                  std::string intr_trans_str = std::to_string(interpreter->GetTransactionId());
+                  if (interpreter->HasActiveTransaction() && transaction_id == intr_trans_str) {
+                    if (interpreter->username_ == username || isAdmin) {
+                      // if (interpreter->in_explicit_transaction_) {
+                      //   interpreter->expect_rollback_ = true;
+                      // } else {
+                      interpreter->AbortTransactionByUser();
+                      interpreter->Abort();
+                      // }
+                      status = "Transaction killed";
```
Josipmrden, line 1986: is it killed or the process of killing started?

src/storage/v2/storage.cpp
```diff
@@ -1144,6 +1144,16 @@ void Storage::Accessor::FinalizeTransaction() {
   }
 }
 
+uint64_t Storage::Accessor::GetTransactionId() const { return transaction_.transaction_id; }
+
+bool Storage::Accessor::IsTransactionActive() const { return is_transaction_active_; }
+
+std::atomic<bool> *Storage::Accessor::TransactionAbortedByUser() { return &transaction_.aborted_by_user; }
```
antoniofilipovic, line 1151: When function returns bool, usually we are naming them with Is...

In this case you can change naming to IsTransactionAbortedByUser

tests/mgbench/client.cpp
```diff
@@ -48,13 +54,18 @@ DEFINE_bool(queries_json, false,
 
 DEFINE_string(input, "", "Input file. By default stdin is used.");
 DEFINE_string(output, "", "Output file. By default stdout is used.");
+DEFINE_bool(validation, false,
```
Josipmrden, line 57: is this your file or from main? merge to epic

tests/mgbench/runners.py
```diff
@@ -398,7 +398,13 @@ def execute(self, queries=None, file_path=None, num_workers=1):
             password=self._password,
             port=self._bolt_port,
         )
+
```
Josipmrden, line 401: this also is not your file i think

tests/unit/query_expression_evaluator.cpp
```diff
@@ -426,6 +427,47 @@ TEST_F(ExpressionEvaluatorTest, VertexAndEdgeIndexing) {
   }
 }
 
+TEST_F(ExpressionEvaluatorTest, TypedValueListIndexing) {
```
Josipmrden, line 430: this also is not your file

Josipmrden (changes requested):
Changes needed, but overall very good, not too many changes on the architecture!

src/storage/v2/storage.hpp
```diff
@@ -324,6 +325,16 @@ class Storage final {
...
+    uint64_t GetTransactionId() const;
+
+    bool IsTransactionActive() const;
+
+    std::atomic<bool> *TransactionAbortedByUser();
+
+    void AbortTransactionByUser();
+
+    bool transaction_aborted_by_user = false;
+
```
antoniofilipovic, line 337: You can change it to 
```suggestion
bool transaction_aborted_by_user{false};
```

src/storage/v2/storage.hpp
```diff
@@ -324,6 +325,16 @@ class Storage final {
...
 
+    uint64_t GetTransactionId() const;
+
+    bool IsTransactionActive() const;
+
+    std::atomic<bool> *TransactionAbortedByUser();
+
+    void AbortTransactionByUser();
+
+    bool transaction_aborted_by_user = false;
```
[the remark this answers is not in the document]
antoniofilipovic: agree

src/storage/v2/storage.hpp
antoniofilipovic: You can change it to 
```suggestion
bool transaction_aborted_by_user{false};
```

antoniofilipovic: And is there reason why variable is not stored as `std::atomic<bool>` when it is used that way?

antoniofilipovic: Are you using this anywhere? If not, you can remove it.

antoniofilipovic: Because, storage accessor has access to transaction, and there is no need for this variable

src/storage/v2/transaction.hpp
```diff
@@ -36,6 +36,7 @@ struct Transaction {
         start_timestamp(start_timestamp),
         command_id(0),
         must_abort(false),
+        aborted_by_user(false),
```
antoniofilipovic, line 39: This variable is bool type, so you can change it to `is_aborted_by_user`.

antoniofilipovic: Check for example in transaction.hpp naming for is_transaction_active_.

src/glue/auth_checker.cpp
```diff
@@ -114,6 +132,26 @@ bool AuthChecker::IsUserAuthorized(const memgraph::auth::User &user,
...
+      memgraph::query::AuthQuery::Privilege::MATCH,       memgraph::query::AuthQuery::Privilege::MERGE,
+      memgraph::query::AuthQuery::Privilege::SET,         memgraph::query::AuthQuery::Privilege::REMOVE,
+      memgraph::query::AuthQuery::Privilege::INDEX,       memgraph::query::AuthQuery::Privilege::STATS,
+      memgraph::query::AuthQuery::Privilege::AUTH,        memgraph::query::AuthQuery::Privilege::CONSTRAINT,
+      memgraph::query::AuthQuery::Privilege::DUMP,        memgraph::query::AuthQuery::Privilege::REPLICATION,
+      memgraph::query::AuthQuery::Privilege::DURABILITY,  memgraph::query::AuthQuery::Privilege::READ_FILE,
+      memgraph::query::AuthQuery::Privilege::FREE_MEMORY, memgraph::query::AuthQuery::Privilege::TRIGGER,
+      memgraph::query::AuthQuery::Privilege::CONFIG,      memgraph::query::AuthQuery::Privilege::STREAM,
+      memgraph::query::AuthQuery::Privilege::MODULE_READ, memgraph::query::AuthQuery::Privilege::MODULE_WRITE,
+      memgraph::query::AuthQuery::Privilege::WEBSOCKET};
```
antoniofilipovic, line 148: I have two thoughts about this. First, this seems not maintainable because if we change enum class `Privilege` we need to check that this vector is also updated. Second thought is that we can make this static array that it is accessed from other parts of code.

I think we can do the following to make it less bug prone.

Create following array:

static constexpr Privilege privilege_array[] = {
  Privilege::CREATE,
  Privilege::DELETE,
  // ... and so on
};

static_assert(std::size(privilege_array) == static_cast<std::size_t>(Privilege::WEBSOCKET), "Missing Privilege values in privilege_array");

And afterwards, you can initialize this

antoniofilipovic: with following :
const std::vector<Privilege> privileges{
  std::begin(privilege_array),
  std::end(privilege_array)
};

src/glue/auth_checker.cpp
```diff
@@ -114,6 +132,26 @@ bool AuthChecker::IsUserAuthorized(const memgraph::auth::User &user,
...
+  return std::all_of(privileges.begin(), privileges.end(), [&user_permissions](const auto privilege) {
+    return user_permissions.Has(memgraph::glue::PrivilegeToPermission(privilege)) ==
+           memgraph::auth::PermissionLevel::GRANT;
```
antoniofilipovic, line 151: What if Has returns Neutral, isn't that also okay?

antoniofilipovic: I am not sure when it returns Neutral to be honest

src/query/interpret/eval.hpp
```diff
@@ -31,6 +31,71 @@
 
 namespace memgraph::query {
 
+class ReferenceExpressionEvaluator : public ExpressionVisitor<TypedValue *> {
```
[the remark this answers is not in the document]
antoniofilipovic: yes

src/query/context.hpp
```diff
@@ -66,6 +66,7 @@ struct ExecutionContext {
   SymbolTable symbol_table;
   EvaluationContext evaluation_context;
   std::atomic<bool> *is_shutting_down{nullptr};
+  std::atomic<bool> *aborted_by_user{nullptr};
```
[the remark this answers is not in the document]
as51340: Because we have to sync this flag between two threads. Example:
Thread A runs long-running query.
Thread B runs kill transaction and sets aborted_by_user to true.
In order that thread A sees change made from thread B, it needs to pick up the fresh value, hence atomic.

src/memgraph.cpp
```diff
@@ -522,13 +522,20 @@ class BoltSession final : public memgraph::communication::bolt::Session<memgraph
       : memgraph::communication::bolt::Session<memgraph::communication::v2::InputStream,
                                                memgraph::communication::v2::OutputStream>(input_stream, output_stream),
         db_(data->db),
+        interpreter_context_(data->interpreter_context),
```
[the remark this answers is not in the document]
as51340: Not sure I get it: there are two classes InterpreterContext and Interpreter. InterpreterContext contains pointers to interpreters.

src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -31,6 +33,7 @@
 #include "query/exceptions.hpp"
 #include "query/frontend/ast/ast.hpp"
 #include "query/frontend/ast/ast_visitor.hpp"
+#include "query/frontend/opencypher/generated/MemgraphCypher.h"
```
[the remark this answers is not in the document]
as51340: Good question, I think it came from VSCode, will delete it.

src/query/interpreter.cpp
```diff
@@ -1909,6 +1918,119 @@ PreparedQuery PrepareSettingQuery(ParsedQuery parsed_query, const bool in_explic
...
+      std::transform(transaction_query->transaction_id_list_.begin(), transaction_query->transaction_id_list_.end(),
+                     std::back_inserter(transaction_ids), [&evaluator](Expression *expression) {
+                       return std::string(expression->Accept(evaluator).ValueString());
+                     });
+      callback.header = {"transaction_id", "status"};
+      callback.fn = [interpreter_context, db_accessor, transaction_ids, username, isAdmin]() mutable {
+        spdlog::info("Number of transactions to kill in callback: {}", transaction_ids.size());
+        std::vector<std::vector<TypedValue>> results;
+        for (const std::string &transaction_id : transaction_ids) {
+          std::string status{"Transaction not found"};
```
antoniofilipovic, line 1973: Maybe change this messsage to: "Transaction with id:{} not found",

src/query/frontend/semantic/required_privileges.cpp
```diff
@@ -80,6 +80,11 @@ class PrivilegeExtractor : public QueryVisitor<void>, public HierarchicalTreeVis
 
   void Visit(SettingQuery & /*setting_query*/) override { AddPrivilege(AuthQuery::Privilege::CONFIG); }
 
+  void Visit(TransactionQueueQuery &transaction_queue_query) override {
+    // AddPrivilege(AuthQuery::Privilege::SET);
+    // TODO replace privilege
```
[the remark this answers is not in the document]
as51340: If we decide to stick with the current approach of admin having all privileges, then the method will stay empty. Should be discussed tomorrow on daily.

src/query/interpret/eval.hpp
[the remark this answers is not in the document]
as51340: si si

src/storage/v2/storage.hpp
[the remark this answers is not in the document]
as51340: This should be deleted actually, it stayed from some experiments

src/query/interpreter.cpp
```diff
@@ -1909,6 +1918,119 @@ PreparedQuery PrepareSettingQuery(ParsedQuery parsed_query, const bool in_explic
...
+          interpreter_context->interpreters.WithLock(
+              [db_accessor, &transaction_id, &status, username, isAdmin](const auto &interpreters) {
```
antoniofilipovic, line 1975: you are not using here db_accessor

src/query/interpreter.cpp
as51340: I think yes, but this part needs to be a little bit redesigned I think, stay tuned :smile:

src/query/interpreter.cpp
antoniofilipovic: As user, it would be beneficial to know specific `id`s which are going to be killed.

src/query/interpreter.cpp
as51340: Will check

src/query/interpreter.cpp
```diff
@@ -1909,6 +1918,119 @@ PreparedQuery PrepareSettingQuery(ParsedQuery parsed_query, const bool in_explic
...
+                for (Interpreter *interpreter : interpreters) {
+                  std::string intr_trans_str = std::to_string(interpreter->GetTransactionId());
+                  if (interpreter->HasActiveTransaction() && transaction_id == intr_trans_str) {
+                    if (interpreter->username_ == username || isAdmin) {
```
antoniofilipovic, line 1979: change order maybe:
isAdmin || ...

It seems more likely user will be admin (probably)

src/query/interpreter.cpp
as51340: yes

src/query/interpreter.cpp
as51340: shouldn't be a problem

src/query/interpreter.cpp
```diff
@@ -1909,6 +1918,119 @@ PreparedQuery PrepareSettingQuery(ParsedQuery parsed_query, const bool in_explic
...
+                                 TypedValue(queries_summaries_typed)});
+            }
+          }
+        });
+        return results;
+      };
+      break;
+    }
+    case TransactionQueueQuery::Action::TERMINATE_TRANSACTIONS: {
+      std::vector<std::string> transaction_ids;
```
antoniofilipovic, line 1963: I would rename this variable to something more descriptive. User would like to kill those transactions, if they are found and if he has rights, so maybe:

kill_transaction_ids
maybe_kill_transaction_ids

src/query/interpreter.cpp
as51340: Old things => deleted

src/query/interpreter.cpp
as51340: Yes I was thinking about that, but then we return duplicated information: "status" and "status message" - GQLAlchemy user can still fetch this based on the status message, what do you think?

src/query/interpreter.cpp
as51340: From user's perspective it is killed immediately, actual stopping will happen upon checking MustAbort for Cypher queries and mgp_must_abort. But user doesn't care about any of this, so I would stay with current description

src/storage/v2/storage.hpp
as51340: yes, yes this is old, needs to be deleted

src/storage/v2/transaction.hpp
as51340: yes agree! Can you check constructor `Transctions &&` ? I am not sure if this line:
```
is_aborted_by_user_(false),
```
can create some problems and:
```
is_aborted_by_user_(other.is_aborted_by_user_)
```
cannot be written because this constructor is deleted from atomic

src/query/interpreter.cpp
```diff
@@ -1909,6 +1918,119 @@ PreparedQuery PrepareSettingQuery(ParsedQuery parsed_query, const bool in_explic
...
+                     std::back_inserter(transaction_ids), [&evaluator](Expression *expression) {
+                       return std::string(expression->Accept(evaluator).ValueString());
+                     });
+      callback.header = {"transaction_id", "status"};
+      callback.fn = [interpreter_context, db_accessor, transaction_ids, username, isAdmin]() mutable {
+        spdlog::info("Number of transactions to kill in callback: {}", transaction_ids.size());
+        std::vector<std::vector<TypedValue>> results;
+        for (const std::string &transaction_id : transaction_ids) {
+          std::string status{"Transaction not found"};
+          interpreter_context->interpreters.WithLock(
```
antoniofilipovic, line 1974: I think we can make this more readable and I have two ideas, so you can choose if you think it will work:

1. Instead of going over all interpreters for each transaction id, I think you could do following.

On `interpreter_contex->interpreters` call  `Lock()`.

Like this:
```
auto my_map_ptr = my_important_map_.Lock();
my_map_ptr->emplace(key, value);
```

Then filter out interpreters which have active transaction and where transaction id is std::vector<std::string> (you can change this to std::unordered_set) and username is correct or isAdmin.

And afterwards just kill those transactions.

2. idea is following:
for (Interpreter *interpreter : interpreters) {
    if (!interpreter->HasActiveTransaction() ||  !transaction_id == std::to_string(interpreter->GetTransactionId())) {
        continue;
    }
    if (interpreter->username_ == username || isAdmin){
        status = "Not enough rights to kill the transaction";
        break;
    }
     
    
    interpreter->AbortTransactionByUser();
    interpreter->Abort();
    status = "Transaction killed";
    break;
}

src/glue/auth_checker.cpp
as51340: Yes, assert should help under assumption we don't assign specific values to enum.

as51340: AhI just realized we already have kPrivilegesAll in ast.hpp

src/glue/auth_checker.cpp
as51340: I think the logic is that permission level must be set to grant otherwise it means that no other user has actually given specific permissions to this admin user.

src/query/interpreter.cpp
```diff
@@ -2398,6 +2532,11 @@ Interpreter::PrepareResult Interpreter::Prepare(const std::string &query_string,
       prepared_query = PrepareSettingQuery(std::move(parsed_query), in_explicit_transaction_, &*execution_db_accessor_);
     } else if (utils::Downcast<VersionQuery>(parsed_query.query)) {
       prepared_query = PrepareVersionQuery(std::move(parsed_query), in_explicit_transaction_);
+    } else if (utils::Downcast<TransactionQueueQuery>(parsed_query.query)) {
+      spdlog::info("Number of interpreters in else if: {}", interpreter_context_->interpreters->size());
```
antoniofilipovic, line 2536: I think there is no need for this info here

src/query/interpreter.cpp
as51340: This is per row status so the user will get:
| transaction id | status |  
| ----- | ----- |
so I think it doesn't make sense that in status transaction id is mentioned again

src/query/interpreter.cpp
antoniofilipovic: Why would this be beneficial to user? Maybe I don't know something

src/query/interpreter.cpp
antoniofilipovic, line 2585: this comment is here leftover.

src/query/interpreter.cpp
as51340: then in order not to introduce semantic changes, the condition that active transaction exists and is equal has to be duplicated in else branch

src/query/frontend/opencypher/grammar/MemgraphCypherLexer.g4
```diff
@@ -53,6 +53,7 @@ DIRECTORY           : D I R E C T O R Y ;
 DROP                : D R O P ;
 DUMP                : D U M P ;
 DURABILITY          : D U R A B I L I T Y ;
+EDGE_TYPES          : E D G E UNDERSCORE T Y P E S ;
```
antoniofilipovic, line 56: Where is this used?

src/query/interpreter.cpp
as51340: fico linguista, I like it! will do the change

src/query/interpreter.cpp
as51340: was used for debugging, will be del

src/query/frontend/opencypher/grammar/MemgraphCypherLexer.g4
as51340: some old things, I really don't know something happened when ante or me merged some things

src/query/interpreter.cpp
as51340: yes but you still have to iterate over transaction_ids sent by user, otherwise which ones you kill? I used this approach of locking inside the loop of transaction_ids so that interpreters don't stay locked for longer time than neccessary

.github/pull_request_template.md
```diff
@@ -1,11 +1,14 @@
...
 - [ ] Provide the full content or a guide for the final git message
 
 [master < Task] PR
 - [ ] Check, and update documentation if necessary
-- [ ] Update [changelog](https://docs.memgraph.com/memgraph/changelog)
 - [ ] Provide the full content or a guide for the final git message
+
+
+To keep docs changelog up to date, one more thing to do:
+- [ ] Write a release note here
```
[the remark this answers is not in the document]
as51340: I don't know what this is, this came out from some merge of me or ante :confused:

antoniofilipovic (changes requested):
I suggested a few changes here. And I think I found a bug. When you try to call terminate the transaction with a random ID, the Memgraph gets shut down. I am not sure why

`TERMINATE TRANSACTIONS "9223372036854775810";`

Also, did you have time to test how PR works on various scenarios, like BFS, DFS, All shortest paths, and so on?

as51340:
> I suggested a few changes here. And I think I found a bug. When you try to call terminate the transaction with a random ID, the Memgraph gets shut down. I am not sure why
> 
> `TERMINATE TRANSACTIONS "9223372036854775810";`
> 
> Also, did you have time to test how PR works on various scenarios, like BFS, DFS, All shortest paths, and so on?

Very possible, will check and correct

src/memgraph.cpp
```diff
@@ -522,13 +522,19 @@ class BoltSession final : public memgraph::communication::bolt::Session<memgraph
...
         interpreter_(data->interpreter_context),
         auth_(data->auth),
 #if MG_ENTERPRISE
         audit_log_(data->audit_log),
 #endif
         endpoint_(endpoint),
         run_id_(data->run_id) {
+    interpreter_context_->interpreters.WithLock([this](auto &interpreters) { interpreters.insert(&interpreter_); });
```
gitbuda, line 533: The `insert` and `erase` might produce some issues on the `mgbench` workload, so we should definitely run `mgbench` and observe if there are some degradations 😄

tests/unit/transaction_queue.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+#include <gtest/gtest.h>
+#include "gmock/gmock.h"
+
+class TransactionQueueSimpleTest : public ::testing::Test {
+ protected:
+  memgraph::storage::Storage db_;
+  std::filesystem::path data_directory{std::filesystem::temp_directory_path() / "MG_tests_unit_transaction_queue_intr"};
+  memgraph::query::InterpreterContext interpreter_context{&db_, {}, data_directory};
+  //    InterpreterFaker running_interpreter{&db_, {}, data_directory_running}
```
gitbuda, line 22: To delete 😄

tests/e2e/transaction_queue/test_transaction_queue.py
```diff
@@ -0,0 +1,85 @@
...
+    admin_cursor = connect(username="admin", password="").cursor()
+    admin_connection = threading.Thread(target=show_transactions_test, args=(admin_cursor, 1))
+    admin_connection.start()
+    admin_connection.join()
+    execute_and_fetch_all(a_cursor, "DROP USER admin")
+
+
+def test_long_running_query():
+    main_cursor = connect().cursor()
+    # run_query_in_separate_thread("", "", "CALL infinite_query.long_query() RETURN 1")
```
gitbuda, line 70: To delete 😄 (delete all commented code)

src/query/context.hpp
```diff
@@ -82,7 +91,9 @@ static_assert(std::is_move_assignable_v<ExecutionContext>, "ExecutionContext mus
 static_assert(std::is_move_constructible_v<ExecutionContext>, "ExecutionContext must be move constructible!");
 
 inline bool MustAbort(const ExecutionContext &context) noexcept {
-  return (context.is_shutting_down != nullptr && context.is_shutting_down->load(std::memory_order_acquire)) ||
+  return (context.transaction_status != nullptr &&
```
gitbuda, line 94: This is probably more likely to cause perf problems -> one more reason to run mgbench 👀

src/query/interpreter.cpp
```diff
@@ -1909,6 +1923,138 @@ PreparedQuery PrepareSettingQuery(ParsedQuery parsed_query, const bool in_explic
...
+        // 1. somehow make sure that they are executed consecutively, not de-scheluded from the cpu, so all atomic
+        // I think this would correspond to having lock on the Interpreter
+        // 2. don't create new transaction while we are in the process of killing this one
+
+        // the assumption is that this whole block should be executed atomically
+        std::optional<uint64_t> intr_trans = interpreter->GetTransactionId();
+        if (intr_trans.has_value() && std::to_string(intr_trans.value()) == transaction_id) {  // but this can change
+          if (interpreter->username_ == username ||
+              isAdmin) {  // this cannot change between executions of two transactions in the same interpreter
+            TransactionStatus loaded_status = TransactionStatus::ALIVE;
```
gitbuda, line 1969: What's the semantic of `NO_TRANSACTION`?

src/query/interpreter.cpp
```diff
@@ -1909,6 +1923,138 @@ PreparedQuery PrepareSettingQuery(ParsedQuery parsed_query, const bool in_explic
...
+            // if the transaction status is not ALIVE, then we don't do anything because it means we are in the process
+            // of committing, aborting (rollback) or no transaction is active
+            if (!interpreter->transaction_status_.compare_exchange_strong(loaded_status,
+                                                                          TransactionStatus::STARTED_KILLING)) {
+              spdlog::debug("Transaction {} cannot be killed because it already started committing or rollback. ",
```
gitbuda, line 1974: Maybe this should go under the `warning` level because user wanted to kill something which is about to be stopped -> the user probably has the wrong assumption about that particular transaction?

src/query/interpreter.cpp
```diff
@@ -1909,6 +1923,138 @@ PreparedQuery PrepareSettingQuery(ParsedQuery parsed_query, const bool in_explic
...
+                            transaction_id);
+              continue;
+            }
+            killed = true;
+            spdlog::debug("Transaction with id: {} successfully killed", transaction_id);
+          } else {
+            spdlog::debug("Not enough rights to kill the transaction");
```
gitbuda, line 1981: This is similar to the above, maybe put under the `warning` level. I'm not saying that we should 100%, just try to think about this from the user perspective 😄

src/query/context.hpp
as51340: why if `is_shutting_down` doesn't create problems? Should be the same

src/query/interpreter.cpp
as51340: That interpreter is idle, doesn't execute anything.

src/query/interpreter.cpp
```diff
@@ -1909,6 +1922,142 @@ PreparedQuery PrepareSettingQuery(ParsedQuery parsed_query, const bool in_explic
...
+
+  auto *transaction_queue_query = utils::Downcast<TransactionQueueQuery>(parsed_query.query);
+  MG_ASSERT(transaction_queue_query);
+  auto callback =
+      HandleTransactionQueueQuery(transaction_queue_query, username, parsed_query.parameters, interpreter_context, dba);
+
+  return PreparedQuery{std::move(callback.header), std::move(parsed_query.required_privileges),
+                       [callback_fn = std::move(callback.fn), pull_plan = std::shared_ptr<PullPlanVector>{nullptr}](
+                           AnyStream *stream, std::optional<int> n) mutable -> std::optional<QueryHandlerResult> {
+                         if (UNLIKELY(!pull_plan)) {
```
antoniofilipovic, line 2049: instead of `UNLIKELY` you can use [[unlikely]] from c++ 20, as we use in `PrepareShowConfigQuery`

antoniofilipovic: @as51340 maybe you know, but why is this path unlikely to execute. If I understand correctly, UNLIKELY means that this statement is unlikely to execute, due to the fact that there will be pull_plan. But I am not sure where do we set pull plan to be something else than `nullptr`. Do you know that part maybe?

antoniofilipovic: @as51340 maybe you know, but why is this path unlikely to execute. If I understand correctly, UNLIKELY means that this statement is unlikely to execute, due to the fact that there will be pull_plan. But I am not sure where do we set pull plan to be something else than `nullptr`. Do you know that part maybe?

src/query/exceptions.hpp
```diff
@@ -118,11 +118,7 @@ class QueryRuntimeException : public QueryException {
 class HintedAbortError : public utils::BasicException {
  public:
   using utils::BasicException::BasicException;
-  HintedAbortError()
-      : utils::BasicException(
-            "Transaction was asked to abort, most likely because it was "
-            "executing longer than time specified by "
-            "--query-execution-timeout-sec flag.") {}
+  HintedAbortError() : utils::BasicException("Transaction was asked to abort") {}
```
antoniofilipovic, line 121: You can extend messsage to say: "... either was executing longer than time specified on ... or someone else asked to abort"

src/storage/v2/storage.cpp
```diff
@@ -1144,6 +1144,13 @@ void Storage::Accessor::FinalizeTransaction() {
   }
 }
 
+std::optional<uint64_t> Storage::Accessor::GetTransactionId() const {
+  if (is_transaction_active_) {
+    return transaction_.transaction_id;
+  }
+  return {};
```
antoniofilipovic, line 1151: I think this is not responsibility of this function to return id only if transaction is active. In my opinion we should return `id` but other side should check if transaction is active, and other stuff which we need to do. This seems like there is some hidden logic there.

tests/e2e/transaction_queue/test_transaction_queue.py
```diff
@@ -0,0 +1,256 @@
...
+from common import connect, execute_and_fetch_all
+
+# Utility functions
+# -------------------------
+
+
+def get_non_show_transaction_id(results):
+    """Returns transaction id of the first transaction that is not SHOW TRANSACTIONS;"""
+    for res in results:
+        if res[2] != ["SHOW TRANSACTIONS"]:
```
Josipmrden, line 30: extract SHOW TRANSACTIONS into constant above

Josipmrden (commented):
minor comments from my side

tests/unit/transaction_queue.cpp
```diff
@@ -0,0 +1,106 @@
...
+#include "gmock/gmock.h"
+
+/*
+Tests rely on the fact that interpreters are sequentially added to runninng_interpreters to get transaction_id of its
+corresponding interpreter/.
+*/
+class TransactionQueueSimpleTest : public ::testing::Test {
+ protected:
+  memgraph::storage::Storage db_;
+  std::filesystem::path data_directory{std::filesystem::temp_directory_path() / "MG_tests_unit_transaction_queue_intr"};
```
antoniofilipovic, line 28: Is there reason for this specific path?

src/query/interpreter.cpp
```diff
@@ -2425,19 +2594,35 @@ Interpreter::PrepareResult Interpreter::Prepare(const std::string &query_string,
   }
 }
 
+std::vector<std::string> Interpreter::GetQueries() const {
+  if (in_explicit_transaction_) {
+    return {"EXPLICIT TRANSACTION"};
```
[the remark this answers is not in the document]
antoniofilipovic: I also agree that this is not good behaviour, we should return all queries executed in transaction, or have some other way for user to distinguish between transactions. Maybe you want to abort all transactions and that should be fine.

tests/unit/transaction_queue.cpp
```diff
@@ -0,0 +1,106 @@
...
+      },
+      0);
+
+  {
+    main_interpreter.Interpret("CREATE (:Person {prop: 1})");
+    auto show_stream = main_interpreter.Interpret("SHOW TRANSACTIONS");
+    ASSERT_EQ(show_stream.GetResults().size(), 2U);
+    // superadmin executing the transaction
+    EXPECT_EQ(show_stream.GetResults()[0][0].ValueString(), "");
+    ASSERT_TRUE(show_stream.GetResults()[0][1].IsString());
```
antoniofilipovic, line 82: Is there a way to check this ID also?

tests/unit/transaction_queue.cpp
```diff
@@ -0,0 +1,106 @@
...
+    EXPECT_EQ(show_stream.GetResults()[0][2].ValueList().at(0).ValueString(), "SHOW TRANSACTIONS");
+    // Also anonymous user executing
+    EXPECT_EQ(show_stream.GetResults()[1][0].ValueString(), "");
+    ASSERT_TRUE(show_stream.GetResults()[1][1].IsString());
+    EXPECT_EQ(show_stream.GetResults()[1][2].ValueList().at(0).ValueString(), "EXPLICIT TRANSACTION");
```
antoniofilipovic, line 87: And do we need some kind of guarantee that transaction above is one if this id?

tests/unit/transaction_queue.cpp
```diff
@@ -0,0 +1,106 @@
...
+    main_interpreter.Interpret("CREATE (:Person {prop: 1})");
+    auto show_stream = main_interpreter.Interpret("SHOW TRANSACTIONS");
+    ASSERT_EQ(show_stream.GetResults().size(), 2U);
+    // admin executing the transaction
+    EXPECT_EQ(show_stream.GetResults()[0][0].ValueString(), "");
+    ASSERT_TRUE(show_stream.GetResults()[0][1].IsString());
+    EXPECT_EQ(show_stream.GetResults()[0][2].ValueList().at(0).ValueString(), "SHOW TRANSACTIONS");
+    // Also admin executing
+    EXPECT_EQ(show_stream.GetResults()[1][0].ValueString(), "");
+    ASSERT_TRUE(show_stream.GetResults()[1][1].IsString());
```
antoniofilipovic, line 56: Is there a way to check this ID also?

tests/unit/transaction_queue.cpp
```diff
@@ -0,0 +1,106 @@
...
+    EXPECT_EQ(show_stream.GetResults()[1][2].ValueList().at(0).ValueString(), "EXPLICIT TRANSACTION");
```
antoniofilipovic, line 57: Is there any smarter way to check these types and values without doing referencing of list

tests/unit/interpreter_faker.hpp
```diff
@@ -0,0 +1,49 @@
...
+   * Return the query stream.
+   */
+  auto Interpret(const std::string &query, const std::map<std::string, memgraph::storage::PropertyValue> &params = {}) {
+    auto prepare_result = Prepare(query, params);
+    auto &stream = prepare_result.first;
+    auto summary = interpreter.Pull(&stream, {}, prepare_result.second);
+    stream.Summary(summary);
+    return std::move(stream);
+  }
+  memgraph::query::AllowEverythingAuthChecker auth_checker;
```
antoniofilipovic, line 46: Do you need this initialized too?

src/query/interpreter.cpp
as51340: @antoniofilipovic I don't know I reused this part of the code. @antaljanosbenjamin please help.

src/query/interpreter.cpp
```diff
@@ -14,19 +14,25 @@
 
 #include <algorithm>
 #include <atomic>
+#include <boost/system/detail/interop_category.hpp>
```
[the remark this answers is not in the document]
as51340: Included accidently.

src/query/interpreter.cpp
```diff
@@ -1909,6 +1922,142 @@ PreparedQuery PrepareSettingQuery(ParsedQuery parsed_query, const bool in_explic
   // NOLINTNEXTLINE(clang-analyzer-cplusplus.NewDeleteLeaks)
 }
 
+std::vector<std::vector<TypedValue>> TransactionQueueQueryHandler::ShowTransactions(
+    InterpreterContext *interpreter_context, const std::optional<std::string> &username, bool isAdmin) {
+  std::vector<std::vector<TypedValue>> results;
+  results.reserve(interpreter_context->interpreters->size());
```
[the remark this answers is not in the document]
as51340: Will change, can you maybe explain when could locking happen twice?

src/query/interpreter.cpp
```diff
@@ -2503,6 +2689,16 @@ void Interpreter::Commit() {
   // We should document clearly that all results should be pulled to complete
   // a query.
   if (!db_accessor_) return;
+  /*
+  At this point we must only check that the transaction is alive because it cannot be in the ended committed state,
+  started rollback state or finished rollbacked state in the same thread. The only danger is that other thread requested
+  its killing and if it has, then this committing step must abort. Exception should suffice.
+  */
+  TransactionStatus alive_status = TransactionStatus::ALIVE;
+  if (!transaction_status_.compare_exchange_strong(alive_status, TransactionStatus::STARTED_COMMITTING)) {
```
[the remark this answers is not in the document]
as51340: Default value is memory_order_seq_cst which performs acquire and release so should be the same. https://en.cppreference.com/w/cpp/atomic/memory_order

tests/e2e/transaction_queue/test_transaction_queue.py
as51340: Why :joy: ? Then it makes sense to change all queries to constants and I get a mess. + I think that SHOW TRANSACTIONS is not often used to make it a constant but pls. correct me if I am wrong.

tests/unit/transaction_queue.cpp
as51340: That would test storage, I don't think this this has nothing to do with transaction_queue and if the test passes that means that the correct transaction with the correct transaction_id has been killed

tests/unit/transaction_queue.cpp
as51340: Can you please explain what you meant?

tests/unit/transaction_queue.cpp
as51340: answered above

tests/unit/transaction_queue.cpp
as51340: Return result is list of strings so I don't think there is. What is wrong with this? If something is wrong the test will fail

tests/unit/transaction_queue.cpp
```diff
@@ -0,0 +1,106 @@
...
+    std::string run_trans_id = show_stream.GetResults()[1][1].ValueString();
+    std::string esc_run_trans_id = "'" + run_trans_id + "'";
+    auto terminate_stream = main_interpreter.Interpret("TERMINATE TRANSACTIONS " + esc_run_trans_id);
+    // check result of killing
+    ASSERT_EQ(terminate_stream.GetResults().size(), 1U);
+    EXPECT_EQ(terminate_stream.GetResults()[0][0].ValueString(), run_trans_id);
+    ASSERT_TRUE(terminate_stream.GetResults()[0][1].ValueBool());  // that the transaction is actually killed
+    // check the number of transactions now
+    auto show_stream_after_killing = main_interpreter.Interpret("SHOW TRANSACTIONS");
+    ASSERT_EQ(show_stream_after_killing.GetResults().size(), 1U);
```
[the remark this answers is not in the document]
as51340: Please explain, what should I comment?

src/storage/v2/storage.cpp
as51340: Yes but if there is no active transaction, there is nothing to return just garbage and hence the program can crash with exception

src/query/interpreter.cpp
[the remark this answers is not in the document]
as51340: @antoniofilipovic why abort all transactions?? I agree that returning all queries when in explicit transaction should be the way to go + that sync between threads is needed. The reason I iterated over `query_executions` vector is because if interpreter runs let's say third query, there will be two nullptr elements in the vector. But yes, I think this can be solved in a nicer way with returning just back() element.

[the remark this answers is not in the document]
as51340: But don't you think that returning all queries that are part of explicit transaction can create a mess? E.g when running thousands of queries inside the explicit transaction.

src/storage/v2/storage.hpp
```diff
@@ -324,6 +325,8 @@ class Storage final {
 
     void FinalizeTransaction();
 
+    std::optional<uint64_t> GetTransactionId() const;
```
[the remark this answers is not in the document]
as51340: Yes this was left intentionally because of that comment about not very probable but possible data race. I think I have an idea how to solve it.

src/query/interpreter.cpp
```diff
@@ -1909,6 +1922,142 @@ PreparedQuery PrepareSettingQuery(ParsedQuery parsed_query, const bool in_explic
...
+  for (const std::string &transaction_id : maybe_kill_transaction_ids) {
+    bool killed = false;
+    bool transaction_found = false;
+    interpreter_context->interpreters.WithLock([&transaction_id, &killed, &transaction_found, username,
+                                                isAdmin](const auto &interpreters) {
+      for (Interpreter *interpreter : interpreters) {
+        // the problem can happen e.g: you see that transaction is alive you check transaction_id and username but
+        // we kill the new transaction. This can happen even if the new transaction hasn't started because
+        // as soon as it starts it will be killed. However, this situation is extremely rare + random tests are
+        // passing so this is for now not handled.
```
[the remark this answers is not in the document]
as51340: Yes, I got the idea how solve it.

src/query/interpreter.cpp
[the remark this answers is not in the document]
antoniofilipovic: I am sorry, I expressed my thoughts incorrectly. I thought about aborting that specific transaction with all the other queries inside.

src/storage/v2/storage.cpp
antoniofilipovic: Okay, why do we delete the transaction id? I didn't know that happens.

tests/unit/transaction_queue.cpp
antoniofilipovic: Nothing is wrong but seems kind of messy with checking this way.

antoniofilipovic: But I don't have any idea how to do it better.

include/mgp.hpp
```diff
@@ -1572,6 +1585,14 @@ inline Id::Id(int64_t id) : id_(id) {}
 
 inline Graph::Graph(mgp_graph *graph) : graph_(graph) {}
 
+inline bool Graph::MustAbort() const { return must_abort(graph_); }
+
+inline void Graph::CheckMustAbort() const {
+  if (MustAbort()) {
+    throw MustAbortException("Query was asked to abort.");
+  }
```
antoniofilipovic, line 1593: Python API has `CheckMustAbort` and `MustAbort`, where later doesn't return error, but 0 or 1. I think we can add that too. It does make sense to have one which doesn't throw.

antoniofilipovic: You can check in `py_module.cpp` 

```
int mgp_must_abort(mgp_graph *graph) {
  MG_ASSERT(graph->ctx);

  static_assert(noexcept(memgraph::query::MustAbort(*graph->ctx)));
  return memgraph::query::MustAbort(*graph->ctx) ? 1 : 0;
}
```

src/storage/v2/storage.cpp
as51340: If in the meantime, transaction ended and new wasn't started.

include/mgp.hpp
as51340: There is already one, few lines above:

```
inline bool Graph::MustAbort() const { return must_abort(graph_); }
```

as51340: line 1588

tests/unit/transaction_queue.cpp
antoniofilipovic: agree

src/query/interpreter.cpp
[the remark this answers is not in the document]
antaljanosbenjamin: If you check `Synchronized::operator->` you can see that it implicitly performs the locking and that's how you can access the wrapped object. So basically `interpreters->size()` does the same as `interpreters.Lock()->size()`.

src/query/interpreter.hpp
```diff
@@ -11,8 +11,9 @@
 
 #pragma once
 
-#include <gflags/gflags.h>
+#include <unordered_set>
 
+#include <gflags/gflags.h>
 #include "query/auth_checker.hpp"
```
antaljanosbenjamin, line 17:
```suggestion
#include <unordered_set>

#include <gflags/gflags.h>

#include "query/auth_checker.hpp"
```

src/query/stream/streams.cpp
```diff
@@ -490,13 +490,17 @@ Streams::StreamsMap::iterator Streams::CreateConsumer(StreamsMap &map, const std
...
+
     EventCounter::IncrementCounter(EventCounter::MessagesConsumed, messages.size());
     CallCustomTransformation(transformation_name, messages, result, accessor, *memory_resource, stream_name);
 
     DiscardValueResultStream stream;
 
     spdlog::trace("Start transaction in stream '{}'", stream_name);
-    utils::OnScopeExit cleanup{[&interpreter, &result]() {
+    utils::OnScopeExit cleanup{[interpreter_context, &interpreter, &result]() {
+      interpreter_context->interpreters->erase(interpreter.get());
```
antaljanosbenjamin, line 503: The `CallCustomTransformation` I think can throw an exception. That means this `OnScopeExit` won't be on the stack when we exit the lambda, that means the interpreter won't be deregistered.

antaljanosbenjamin (approved):

src/query/interpreter.hpp
```diff
@@ -290,24 +298,35 @@ class Interpreter final {
...
 
+  std::atomic<TransactionStatus> transaction_status_{TransactionStatus::NO_TRANSACTION};
+
  private:
   struct QueryExecution {
     std::optional<PreparedQuery> prepared_query;
     utils::MonotonicBufferResource execution_memory{kExecutionMemoryBlockSize};
     utils::ResourceWithOutOfMemoryException execution_memory_with_exception{&execution_memory};
 
+    std::string query;
```
antoniofilipovic, line 328: Andi, I think you have this query info inside prepared_query, right?

src/query/context.hpp
```diff
@@ -24,6 +24,15 @@
 
 namespace memgraph::query {
 
+enum class TransactionStatus {
+  NO_TRANSACTION,
+  ALIVE,
+  CHECKING_STATUS,
+  KILLED,
+  STARTED_COMMITTING,
+  STARTED_ROLLBACK,
```
antoniofilipovic, line 33: I think we can come up here with better naming. This has to do that we use naming convention that combines adjectives and nouns to describe the state of the object:

NO_TRANSACTION -> INACTIVE
ALIVE -> ACTIVE
CHECKING_STATUS -> VERIFYING
KILLED -> TERMINATED
STARTED_COMMITING -> COMITED
STARTED_ROLLBACK -> ROLLED_BACK

src/query/interpreter.cpp
```diff
@@ -2504,6 +2704,20 @@ void Interpreter::Commit() {
   // a query.
   if (!db_accessor_) return;
 
+  while (transaction_status_.load(std::memory_order_acquire) == TransactionStatus::CHECKING_STATUS)
+    ;
+
+  /*
+  At this point we must only check that the transaction is alive because it cannot be in the ended committed state,
+  started rollback state or finished rollbacked state in the same thread. The only danger is that other thread killed
+  the transactions and if it did, then this committing step must abort. Exception should suffice.
```
antoniofilipovic, line 2713: Please update this comment with new states.

src/query/interpreter.hpp
```diff
@@ -338,6 +357,8 @@ class Interpreter final {
   // and deletion of a single query execution, i.e. when a query finishes,
   // we reset the corresponding unique_ptr.
   std::vector<std::unique_ptr<QueryExecution>> query_executions_;
+  // all queries that are run as part of the current transaction
+  utils::Synchronized<std::vector<std::string>> transaction_queries_;
```
antoniofilipovic, line 361: Does it make sense to add here utils::SpinLock after std::vector<std::string>

src/query/interpreter.hpp
antoniofilipovic, line 328: Maybe rename this as active query or something like that. Or last query, I think that is better naming.

antoniofilipovic, line 328: Does it make sense to make this query part of prepared_query?

as51340: Nice catch, this was left accidentally and needs to be deleted!

PR #791, opened 2023-02-19
[master < T597] Add restoring of replication roles upon database startup

Josipmrden:
[master < Task] PR

Release note: Replication role is now restored upon the startup of database. Replicas are no longer restored as main instances with manual effort needed to configure the replica again.


Josipmrden:
@vpavicic tagging for release note

vpavicic:
Can we just say this:

Replica instances no longer restart as main instances, but restore their original role.

Josipmrden:
yes @vpavicic

Josipmrden:
@gitbuda  and @antoniofilipovic  I'd need reviews

src/storage/v2/replication/replication_persistence_helper.hpp
```diff
@@ -23,18 +23,23 @@
 
 namespace memgraph::storage::replication {
 
-struct ReplicaStatus {
+static const std::string kReservedReplicationRoleName{"__replication_role"};
```
antoniofilipovic, line 26: Is there some reason not to use inline constexpr?

src/storage/v2/replication/replication_persistence_helper.cpp
```diff
@@ -70,6 +75,13 @@ std::optional<ReplicaStatus> JSONToReplicaStatus(nlohmann::json &&data) {
       data.at(kSSLKeyFile).get_to(replica_status.ssl->key_file);
       data.at(kSSLCertFile).get_to(replica_status.ssl->cert_file);
     }
+
+    if (data.find(kReplicationRole) != data.end()) {
+      auto role = replication::ReplicationRole::MAIN;  // initializing it to dummy value because of clang-tidy
+      data.at(kReplicationRole).get_to(role);          // here happens the assigning of the role
```
antoniofilipovic, line 81: replica_status.role = replication::ReplicationRole::MAIN;
data.at(kReplicationRole).get_to(replica_status.role.value());   

https://en.cppreference.com/w/cpp/utility/optional/value

I think this should work. `value()` returns reference if there is value inside, and get_to will store value on that reference, so all good, right?

src/storage/v2/storage.cpp
```diff
@@ -323,13 +323,6 @@ Storage::Storage(Config config)
       uuid_(utils::GenerateUUID()),
       epoch_id_(utils::GenerateUUID()),
       global_locker_(file_retainer_.AddLocker()) {
```
antoniofilipovic, line 325: Why did you delete this `warn` comment?

src/storage/v2/storage.cpp
```diff
@@ -418,14 +411,29 @@ Storage::Storage(Config config)
...
-  if (config_.durability.restore_replicas_on_startup) {
-    spdlog::info("Replica's configuration will be stored and will be automatically restored in case of a crash.");
+  if (config_.durability.restore_replication_state_on_startup) {
+    spdlog::info("Replication configuration will be stored and will be automatically restored in case of a crash.");
     utils::EnsureDirOrDie(config_.durability.storage_directory / durability::kReplicationDirectory);
     storage_ =
         std::make_unique<kvstore::KVStore>(config_.durability.storage_directory / durability::kReplicationDirectory);
-    RestoreReplicas();
+
+    RestoreReplicationRole();
```
antoniofilipovic, line 420: I see here we are restoring replication role on startup, but why are we restoring replica only for MAIN?

src/storage/v2/replication/replication_persistence_helper.hpp
Josipmrden: changed to string_view and inline

src/storage/v2/storage.cpp
Josipmrden: it is actually further down

Josipmrden: after the restoring of the replica role

src/storage/v2/storage.cpp
Josipmrden: we are restoring replication role, and restoring corresponding replicas if it's the main

antoniofilipovic (approved):
Looks good to me

Josipmrden:
@vpavicic

PR #792, opened 2023-02-19
Add notification warning when commit to SYNC replica fails

Josipmrden:
Up to now, an exception would be thrown if there was a write query on MAIN instance with at least one replica down. It's now upgraded to only show notification that main has committed data but not synced with all replicas.

[master < Task] PR

Release note: [Replication] When committing new changes on MAIN instance, a notification is displayed instead of exception if data has not been synced with all replicas (e.g. in case a replica is down).

### DIFF
--- a/src/query/interpreter.cpp
+++ b/src/query/interpreter.cpp
@@ -1,4 +1,4 @@
-// Copyright 2022 Memgraph Ltd.
+// Copyright 2023 Memgraph Ltd.
 //
 // Use of this software is governed by the Business Source License
 // included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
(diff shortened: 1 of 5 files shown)


Josipmrden:
@vpavicic tagging for release note

vpavicic:
@Josipmrden Like this?

From this release, when changes are commited on MAIN instance but cannot be synced with all REPLICA instances (for example, a replica is unavailable), a notification is displayed instead of an exepction.

Josipmrden:
> @Josipmrden Like this?
> 
> From this release, when changes are commited on MAIN instance but cannot be synced with all REPLICA instances (for example, a replica is unavailable), a notification is displayed instead of an exepction.

Yes, I like this one

src/query/interpreter.cpp
```diff
@@ -2590,8 +2590,15 @@ void Interpreter::Commit() {
 
   SPDLOG_DEBUG("Finished committing the transaction");
   if (!commit_confirmed_by_all_sync_repplicas) {
-    throw ReplicationException("At least one SYNC replica has not confirmed committing last transaction.");
+    std::vector<Notification> notifications;
+    notifications.reserve(1);
+    notifications.emplace_back(SeverityLevel::WARNING, NotificationCode::REPLICA_PORT_WARNING,
+                               "At least one SYNC replica has not confirmed committing last transaction.");
```
gitbuda, line 2596: All this is possible as a part of the constructor line.

gitbuda, line 2596: Expand the notification with the part in which you are trying to suggest user what it should be done.

gitbuda (changes requested):
Some minor suggestions in the inline comments.

antoniofilipovic:
@Josipmrden what is status with this PR

Josipmrden:
> @Josipmrden what is status with this PR

I'll convert this to Draft and we push it to v2.10

andrejtonev:
@Josipmrden should this be merged?

PR #803, opened 2023-02-21
[T1236 < T1238] Make where exists insensitive to variable positions and multihop

Josipmrden:
The PRs concern is to make exists take an arbitrary pattern (not triplets anymore), as well to put symbols not only on first place but on others as well.

[master < Task] PR


antoniofilipovic (changes requested):
I left initial review and suggested few changes, but need to go over this one more time as I am not sure how does `VaryQueryPartMatching` works exactly.

src/query/plan/variable_start_planner.cpp
```diff
@@ -209,19 +209,43 @@ CartesianProduct<VaryMatchingStart> VaryMultiMatchingStarts(const std::vector<Ma
...
 }
 
+CartesianProduct<VaryMatchingStart> VaryFilterMatchingStarts(const Matching &matching,
+                                                             const SymbolTable &symbol_table) {
+  auto i = 0;
+  for (const auto &filter : matching.filters) {
+    for ([[maybe_unused]] const auto &filter_matchings : filter.matchings) {
+      i += 1;
+    }
+  }
```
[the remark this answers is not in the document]
Josipmrden: ok

src/query/plan/variable_start_planner.cpp
```diff
@@ -209,19 +209,43 @@ CartesianProduct<VaryMatchingStart> VaryMultiMatchingStarts(const std::vector<Ma
...
+
+  std::vector<VaryMatchingStart> variants;
+  variants.reserve(i);
+
+  for (const auto &filter : matching.filters) {
+    for (const auto &filter_matching : filter.matchings) {
+      variants.emplace_back(filter_matching, symbol_table);
+    }
+  }
+
```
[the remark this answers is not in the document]
Josipmrden: Yes it's a premature optimization :')

src/query/plan/variable_start_planner.cpp
```diff
@@ -209,19 +209,43 @@ CartesianProduct<VaryMatchingStart> VaryMultiMatchingStarts(const std::vector<Ma
...
+  return MakeCartesianProduct(std::move(variants));
```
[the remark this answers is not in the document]
Josipmrden: I just copied the code which handled multi matchings, I think it's not such a big deal but we can check.

src/query/plan/variable_start_planner.cpp
```diff
@@ -248,20 +275,23 @@ VaryQueryPartMatching::iterator &VaryQueryPartMatching::iterator::operator++() {
   //    * (m2), (o1), (g1)
```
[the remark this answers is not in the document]
Josipmrden: added

src/query/plan/variable_start_planner.cpp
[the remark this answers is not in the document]
Josipmrden: I will not modify this as you solution also is complained by clang tidy.

src/query/plan/variable_start_planner.cpp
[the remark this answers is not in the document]
antoniofilipovic: I think we should not use `reserve` like this. It makes sense to reserve memory upfront if we are sure about the number of elements. Otherwise, we can just use `emplace_back`. What do you think?

src/query/plan/variable_start_planner.cpp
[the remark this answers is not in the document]
antoniofilipovic: We can leave it like this, it is no problem, but just wanted to note this

Josipmrden:
As for multiple exists, only one exists is generated

src/query/plan/pretty_print.cpp
```diff
@@ -589,6 +599,16 @@ bool PlanToJsonVisitor::PreVisit(Filter &op) {
...
 
+  auto i = 1;
+  for (const auto &pattern_filter : op.pattern_filters_) {
+    auto s = std::to_string(i);
+
+    pattern_filter->Accept(*this);
+    self["pattern_filter" + s] = PopOutput();
+
+    i++;
+  }
```
antoniofilipovic, line 610: The idea to make this more understandable:

```
for (const auto& pattern_filter : op.pattern_filters_) {
  const auto index = std::distance(op.pattern_filters_.begin(), std::find(op.pattern_filters_.begin(), op.pattern_filters_.end(), pattern_filter)) + 1;
  const auto key = "pattern_filter" + std::to_string(index);

  pattern_filter->Accept(*this);
  self[key] = PopOutput();
}
```

If not that, I would isolate "pattern_filter" + s as `key` variable

src/query/plan/variable_start_planner.cpp
```diff
@@ -155,6 +155,18 @@ auto ExpansionNodes(const std::vector<Expansion> &expansions, const SymbolTable
...
 
+FilterMatching ToFilterMatching(Matching &matching) {
+  FilterMatching filter_matching;
+  filter_matching.expansions = matching.expansions;
+  filter_matching.edge_symbols = matching.edge_symbols;
+  filter_matching.filters = matching.filters;
+  filter_matching.node_symbol_to_expansions = matching.node_symbol_to_expansions;
+  filter_matching.named_paths = matching.named_paths;
+  filter_matching.expansion_symbols = matching.expansion_symbols;
+
```
antoniofilipovic, line 166: Do you need to set up here `type`

antoniofilipovic: I saw now that this is done in code before

src/query/plan/variable_start_planner.cpp
```diff
@@ -283,6 +331,28 @@ void VaryQueryPartMatching::iterator::SetCurrentQueryPart() {
...
+  auto filter_matchings = *filter_it_;
+  auto iterator_cnt = 0;
+  for (auto &filter : current_query_part_.matching.filters) {
+    auto matchings_size = filter.matchings.size();
+
+    std::vector<FilterMatching> new_matchings;
+    new_matchings.reserve(matchings_size);
+
+    for (auto i = 0; i < matchings_size; i++) {
+      new_matchings.push_back(ToFilterMatching(filter_matchings[iterator_cnt]));
```
antoniofilipovic, line 347: One more question, why is `iterator_cnt` moving differently from `i`

antoniofilipovic: Okay, I think I understand, but maybe naming could be better.

Filter_matchings should be all the matchings inside exists filter across all exists filters, so maybe, all_filter_matchings

And iterator_cnt could be named - all_filter_matchings_idx

src/query/plan/variable_start_planner.cpp
```diff
@@ -209,19 +221,43 @@ CartesianProduct<VaryMatchingStart> VaryMultiMatchingStarts(const std::vector<Ma
   return MakeCartesianProduct(std::move(variants));
 }
 
+CartesianProduct<VaryMatchingStart> VaryFilterMatchingStarts(const Matching &matching,
+                                                             const SymbolTable &symbol_table) {
+  auto i = 0;
+  for (const auto &filter : matching.filters) {
+    for ([[maybe_unused]] const auto &filter_matchings : filter.matchings) {
```
antoniofilipovic, line 228: why is this for loop?

can't we change that to:

if (filter.matching){
    i+=1;
}

src/query/plan/variable_start_planner.cpp
```diff
@@ -209,19 +221,43 @@ CartesianProduct<VaryMatchingStart> VaryMultiMatchingStarts(const std::vector<Ma
   return MakeCartesianProduct(std::move(variants));
 }
 
+CartesianProduct<VaryMatchingStart> VaryFilterMatchingStarts(const Matching &matching,
+                                                             const SymbolTable &symbol_table) {
+  auto i = 0;
```
antoniofilipovic, line 226: I would change this to something more meaningful, since it reprsents all filter matchings counter, then to:
`all_filter_matchings_cnt`

src/query/plan/variable_start_planner.cpp
```diff
@@ -283,6 +331,28 @@ void VaryQueryPartMatching::iterator::SetCurrentQueryPart() {
...
     current_query_part_.merge_matching = *merge_it_;
   }
+  DMG_ASSERT(filter_it_ != filter_end_ || filter_begin_ == filter_end_,
+             "Either there are no filter matchings or we can always generate"
+             "a variation");
+
+  auto filter_matchings = *filter_it_;
+  auto iterator_cnt = 0;
+  for (auto &filter : current_query_part_.matching.filters) {
+    auto matchings_size = filter.matchings.size();
```
antoniofilipovic, line 341: Again here, does filter.matchings have any time more then 1 matching?

src/query/plan/variable_start_planner.cpp
```diff
@@ -242,26 +281,35 @@ VaryQueryPartMatching::iterator &VaryQueryPartMatching::iterator::operator++() {
...
+  //    * (m1), (o1), (g1), (f1)
+  //    * (m1), (o1), (g1), (f2)
+  //    * (m1), (o2), (g1), (f1)
+  //    * (m1), (o2), (g1), (f2)
+  //    * (m2), (o1), (g1), (f1)
+  //    * (m2), (o1), (g1), (f2)
+  //    * (m2), (o2), (g1), (f1)
+  //    * (m2), (o2), (g1), (f2)
+
   // Create variations by changing the merge part first.
```
antoniofilipovic, line 295: Filter part first, we need to change this comment

antoniofilipovic: It says " Create variations by changing the **merge**"...
Should be **filter**

antoniofilipovic (commented):
Last round of comments, and that should be it

src/query/plan/pretty_print.cpp
Josipmrden:
```
  for (auto i = 0; i < op.pattern_filters_.size(); i++) {
    auto pattern_filter_key = "pattern_filter" + std::to_string(i);

    op.pattern_filters_[i]->Accept(*this);
    self[pattern_filter_key] = PopOutput();
  }

```

I went with this

src/query/plan/rule_based_planner.cpp
```diff
@@ -509,6 +500,12 @@ std::unique_ptr<LogicalOperator> GenReturnBody(std::unique_ptr<LogicalOperator>
 
 namespace impl {
 
+bool HasBoundFilterSymbols(const std::unordered_set<Symbol> &bound_symbols, const FilterInfo &filter) {
+  return std::ranges::all_of(filter.used_symbols.begin(), filter.used_symbols.end(), [bound_symbols](auto symbol) {
+    return bound_symbols.find(symbol) != bound_symbols.end();
+  });
+}
```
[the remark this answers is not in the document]
Josipmrden: sure

src/query/plan/variable_start_planner.cpp
Josipmrden: but it can not be set because Matching is a superclass. it does not have these types. It is done after

src/query/plan/variable_start_planner.cpp
Josipmrden: there can be multiple. best i can do is size

src/query/plan/variable_start_planner.cpp
Josipmrden: makes sense

src/query/plan/variable_start_planner.cpp
Josipmrden: makes sense

src/query/plan/variable_start_planner.cpp
Josipmrden: makes sense

src/query/plan/variable_start_planner.cpp
Josipmrden: it can, if it's composite for example exists OR exists is 1 filter, while exists AND exists are 2 filters

Josipmrden:
@antoniofilipovic please see the changes

src/query/plan/variable_start_planner.cpp
antoniofilipovic: yes, I was wrong here

antoniofilipovic (approved):
All seems good to go to me

Josipmrden:
We let the checks pass and then merge.

PR #805, opened 2023-02-23
Time depended execution mgbench client 

antejavor:
Time depended on execution is introduced to the mgbench client. By passing the argument `time_dependent_execution`, the workload is executed x number of seconds. 

- If workload(specified queries) are all executed, it reruns from the beginning until the x time mark is not hit. This means rerunning the same queries all over again. 

- Since killing individual threads is not an option. The total duration can take longer than specified until the last started query is not completed. If this duration is higher than 1%, the final throughput is adjusted and decreased for the percentage of extra time execution has taken compared to specified. 

Hipotetical situation:  

- target duration = 30 seconds 
- realistic duration = 33 seconds
- executed queries = 100
- throughput =  executed queries / realistic duration = 100/33 = 3,03 Q/s
- adjusted throughput = throughput * 30/33  = 2,75 Q/s


tests/mgbench/client.cpp
```diff
@@ -58,6 +58,10 @@ DEFINE_bool(validation, false,
             "Set to true to run client in validation mode."
             "Validation mode works for singe query and returns results for validation"
             "with metadata");
+DEFINE_int32(time_dependent_execution, -1,
+             "Time dependend executins executes the queries for a specified number of minutes."
```
gitbuda, line 62: Isn't there a better choice in seconds or milliseconds? Maybe even double in seconds, something that maps to `std::chrono::duration<double>` (seconds expressed as `double`)

Pure milliseconds look best to me at the moment 🤔

antejavor: @gitbuda There are two things. First, it is much easier to think in minutes than in seconds and definitely milliseconds, so running some workload for benchmark.py for a constrained time in minutes seems most straightforward from that side. 
On the other hand, using minutes loses granularity, .... ok moving to seconds. they seems to be best for intuition and granurality we need. 😄

tests/mgbench/client.cpp
```diff
@@ -220,7 +224,111 @@ nlohmann::json LatencyStatistics(std::vector<std::vector<double>> &worker_query_
   return statistics;
 }
 
-void Execute(
+void ExecuteTimeDependentWorkload(
```
antejavor, line 227: Added a new function for time depended execution. Most of the code is shared, but results and exiting differ from normal execution. If you whan't, we can stitch everything into same function.

tests/mgbench/client.cpp
```diff
@@ -220,7 +224,111 @@ nlohmann::json LatencyStatistics(std::vector<std::vector<double>> &worker_query_
...
+  std::vector<double> worker_duration(FLAGS_num_workers, 0.0);
+  std::vector<std::vector<double>> worker_query_durations(FLAGS_num_workers);
+
+  // Start workers and execute queries.
+  auto size = queries.size();
+  std::atomic<bool> run(false);
+  std::atomic<uint64_t> ready(0);
+  std::atomic<uint64_t> position(0);
+
+  memgraph::utils::Timer workload_timer;
```
gitbuda, line 244: Adjust the time measurement

antejavor: Inital time measurment was not in sync with worker execution duration, now workload timer is based on total sum of durations of each workers. Hence the execution time is always greater by the set time limit.

tests/mgbench/client.cpp
```diff
@@ -220,7 +224,106 @@ nlohmann::json LatencyStatistics(std::vector<std::vector<double>> &worker_query_
...
+
+      auto &retries = worker_retries[worker];
+      auto &metadata = worker_metadata[worker];
+      auto &duration = worker_duration[worker];
+      auto &query_duration = worker_query_durations[worker];
+
+      memgraph::utils::Timer worker_timer;
+      while (true) {
+        double total_workload_duration =
+            std::accumulate(worker_duration.begin(), worker_duration.end(), 0.0) / FLAGS_num_workers;
```
gitbuda, line 264: Why accumulating her, isn't the `total_time` 1-1 mapped to the `worker_timer`? + Iterating over an array in the middle of a hot loop seems expansive. Where is `worker_duration` updated?

In general, this code seems more complex than it should be.

antejavor: Rewrite this to avoid accumulating.

tests/mgbench/client.cpp
```diff
@@ -220,7 +224,106 @@ nlohmann::json LatencyStatistics(std::vector<std::vector<double>> &worker_query_
...
+  double throughput = (total_iterations / final_duration) * execution_delta;
+  double raw_throughput = total_iterations / final_duration;
+
+  nlohmann::json summary = nlohmann::json::object();
+  summary["count"] = queries.size();
+  summary["duration"] = final_duration;
+  summary["time_limit"] = FLAGS_time_dependent_execution;
+  summary["queries_executed"] = total_iterations;
+  summary["throughput"] = throughput;
+  summary["raw_throughput"] = raw_throughput;
```
antejavor, line 317: Add a comment about this.

gitbuda (approved):
Seems ok now 😄

gitbuda:
The Jepsen test repeatedly fails for some reason. I'll merge because this change has nothing to do with Jepsen.

PR #811, opened 2023-02-27
[T1238 < T1247] Add exceptions for unimplemented EXISTS query parts

Josipmrden:
The PR namely concerns SET and WITH clauses where exists() functionality does not work yet. A not implemented exception is thrown.

### DIFF
--- a/src/query/frontend/semantic/symbol_generator.cpp
+++ b/src/query/frontend/semantic/symbol_generator.cpp
@@ -448,11 +448,11 @@ bool SymbolGenerator::PreVisit(Exists &exists) {
   auto &scope = scopes_.back();
 
   if (scope.in_set_property) {
-    throw SemanticException("Set property can not be used with exists, but only during matching!");
+    throw utils::NotYetImplemented("Set property can not be used with exists, but only during matching!");
   }
 
   if (scope.in_with) {
-    throw SemanticException("WITH can not be used with exists, but only during matching!");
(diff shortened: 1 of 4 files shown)


src/query/frontend/semantic/symbol_generator.cpp
```diff
@@ -447,6 +447,14 @@ bool SymbolGenerator::PreVisit(Extract &extract) {
 bool SymbolGenerator::PreVisit(Exists &exists) {
   auto &scope = scopes_.back();
 
+  if (scope.in_set_property) {
+    throw utils::NotYetImplemented("Set property can not be used with exists, but only during matching!");
+  }
+
+  if (scope.in_with) {
+    throw utils::NotYetImplemented("WITH can not be used with exists, but only during matching!");
+  }
```
antoniofilipovic, line 456: Is this the best option for a symbol generator, shouldn't we throw in a different part of the code? Rule base planer seems too far, maybe preprocess.hpp? I am not sure

Josipmrden: but it is currently a semantic error, because semantically currently it does not make sense to put exists with WITH clause?

antoniofilipovic: yes, true

antoniofilipovic (approved):
All seems good to go to me

PR #812, opened 2023-02-28
[master < T1246] Add index statistics for better query planning

Josipmrden:
The PR concerns providing more information to indexes which then can be used optimally in query planning, to make better decisions when to use one index over the other. It is only scoped to label property indexes.


src/query/interpreter.cpp
```diff
@@ -1399,6 +1400,141 @@ PreparedQuery PrepareDumpQuery(ParsedQuery parsed_query, std::map<std::string, T
...
+      vertices.begin(), vertices.end(), [&indices_info, &counter, &labels, execution_db_accessor](const auto &vertex) {
+        std::for_each(
+            indices_info.label_property.begin(), indices_info.label_property.end(),
+            [&labels, &counter, execution_db_accessor, &vertex](const auto &info) {
+              if (const auto &has_label = vertex.HasLabel(storage::View::OLD, info.first);
+                  has_label.HasError() || !*has_label ||
+                  (labels[0] != "*" && std::find(labels.begin(), labels.end(),
+                                                 execution_db_accessor->LabelToName(info.first)) == labels.end()))
+                return;
+              if (const auto &property = vertex.GetProperty(storage::View::NEW, info.second);
```
antoniofilipovic, line 1424: I think this storage::View::NEW can be kind of misleading, as if we will take into account changes which happened in this transaction, but actually we won't because we don't give possiblity to user to use this as part of transaction, and also, if there is some other transaction taking place which is changing this properties, we still won't see those changes with snapshot isolation level, and only thing that can influence this is either ANALYTICS MODE or isolation level read uncommited. On the other hand, view only has influence if we are inside same transaciton, which if I am not wrong is forbidden (and should be)

src/query/interpreter.cpp
```diff
@@ -1399,6 +1400,141 @@ PreparedQuery PrepareDumpQuery(ParsedQuery parsed_query, std::map<std::string, T
...
+    const std::span<std::string> labels, DbAccessor *execution_db_accessor, storage::Storage::Accessor *dba) {
+  auto results = std::vector<std::vector<TypedValue>>();
+  using VertexAccessorIterable = decltype(std::declval<query::DbAccessor>().Vertices(storage::View::OLD));
+  using LPIndex = std::pair<storage::LabelId, storage::PropertyId>;
+  VertexAccessorIterable vertices = execution_db_accessor->Vertices(storage::View::OLD);
+
+  std::map<LPIndex, std::map<storage::PropertyValue, int64_t>> counter;
+
+  // initialize index
+  const auto &indices_info = execution_db_accessor->ListAllIndices();
```
antoniofilipovic, line 1413: If I am not wrong, why do we iterate over labels of indices only, and in case of * analyze only such labels? Because I think there is better option to iterate over indices already created (iterate over such skiplists) and then there is no need to check for label. It could happen that some nodes have multiple labels, but most will probably be in different camps, right?

I am looking for a way to remove completely checks whether label exists.

src/query/interpreter.cpp
```diff
@@ -1399,6 +1400,141 @@ PreparedQuery PrepareDumpQuery(ParsedQuery parsed_query, std::map<std::string, T
...
+  std::for_each(
+      vertices.begin(), vertices.end(), [&indices_info, &counter, &labels, execution_db_accessor](const auto &vertex) {
+        std::for_each(
+            indices_info.label_property.begin(), indices_info.label_property.end(),
+            [&labels, &counter, execution_db_accessor, &vertex](const auto &info) {
+              if (const auto &has_label = vertex.HasLabel(storage::View::OLD, info.first);
+                  has_label.HasError() || !*has_label ||
+                  (labels[0] != "*" && std::find(labels.begin(), labels.end(),
```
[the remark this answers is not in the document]
antoniofilipovic: Is there some other way to write this, so that for example if user wants to analyze on labels `*` that we don't need to check every time HasLabel, but for example do something like this:

```
// preprocess varaible labels so it uses ids already instead of doing LabelToName for every vertex
// let's say we did that and it is stored inside
// labels_set_wanted variable which in case user sent `*` represents all labels inside indices_info in label_property
auto &all_vertex_labels_set = vertex.LabelsSet()

```



std::set_intersection(v1.begin(), v1.end(), v2.begin(), v2.end(),
                          std::back_inserter(v_intersection));

//and afterwards just go over these labels in v_insertion and increment counter

```

You can check here example code:
https://en.cppreference.com/w/cpp/algorithm/set_intersection

**Review by antoniofilipovic — CHANGES_REQUESTED** [review]
I need to take look at tests still, let me know when you check comments, we can discuss what makes sense to change.

File: src/query/interpreter.cpp [hunk 4]
```diff
@@ -23,6 +23,7 @@
 #include <unordered_map>
 #include <variant>
 
+#include "_mgp.hpp"
```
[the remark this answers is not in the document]
as51340: Bloody vs code :exploding_head:

src/utils/math.hpp
```diff
@@ -64,4 +65,29 @@ constexpr std::optional<uint64_t> RoundUint64ToMultiple(uint64_t val, uint64_t m
...
+typename std::enable_if<!std::numeric_limits<T>::is_integer, bool>::type LessThanDecimal(T a, T b) {
+  return (b - a) >
+         ((std::fabs(a) < std::fabs(b) ? std::fabs(b) : std::fabs(a)) * std::numeric_limits<double>::epsilon());
+}
+
+template <class T>
+typename std::enable_if<!std::numeric_limits<T>::is_integer, bool>::type GreaterThanDecimal(T a, T b) {
+  return (a - b) >
+         ((std::fabs(a) < std::fabs(b) ? std::fabs(b) : std::fabs(a)) * std::numeric_limits<double>::epsilon());
+}
```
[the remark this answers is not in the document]
as51340: I agree

src/query/plan/rewrite/index_lookup.hpp
```diff
@@ -557,17 +563,25 @@ class IndexLookupRewriter final : public HierarchicalLogicalOperatorVisitor {
...
+        std::optional<storage::IndexStats> new_stats = db_->GetIndexStats(GetLabel(label), GetProperty(property));
+
+        if (!found || (vertex_count * 10 < found->vertex_count) ||
+            (new_stats.has_value() && found->stats.has_value() && (vertex_count / 10 <= found->vertex_count) &&
+             (utils::LessThanDecimal(new_stats->avg_group_size, found->stats->avg_group_size) ||
+              (utils::ApproxEqualDecimal(found->stats->avg_group_size, new_stats->avg_group_size) &&
+               utils::GreaterThanDecimal(found->stats->stat_value, new_stats->stat_value)))) ||
+            found->vertex_count > vertex_count ||
+            (found->vertex_count == vertex_count && is_better_type(filter.property_filter->type_))) {
+          found = LabelPropertyIndex{label, filter, vertex_count, new_stats};
```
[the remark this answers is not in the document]
as51340: You mean like sequence of ifs? I am not sure what is better

src/query/interpreter.cpp
```diff
@@ -1399,6 +1400,141 @@ PreparedQuery PrepareDumpQuery(ParsedQuery parsed_query, std::map<std::string, T
...
+
+  std::map<LPIndex, std::map<storage::PropertyValue, int64_t>> counter;
+
+  // initialize index
+  const auto &indices_info = execution_db_accessor->ListAllIndices();
+  std::for_each(
+      vertices.begin(), vertices.end(), [&indices_info, &counter, &labels, execution_db_accessor](const auto &vertex) {
+        std::for_each(
+            indices_info.label_property.begin(), indices_info.label_property.end(),
+            [&labels, &counter, execution_db_accessor, &vertex](const auto &info) {
```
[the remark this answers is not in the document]
as51340: I tried that before, but structured binding doesn't work with foreach

src/query/interpreter.cpp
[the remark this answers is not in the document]
as51340: yep I cleaned it a bit

src/query/CMakeLists.txt
```diff
@@ -43,18 +43,20 @@ add_library(mg-query STATIC ${mg_query_sources})
 target_include_directories(mg-query PUBLIC ${CMAKE_SOURCE_DIR}/include)
 target_link_libraries(mg-query dl cppitertools Boost::headers)
 target_link_libraries(mg-query mg-integrations-pulsar mg-integrations-kafka mg-storage-v2 mg-license mg-utils mg-kvstore mg-memory)
+
```
jbajic, line 46: I would just remove all the changes from this file, since they are only formatting

src/query/interpreter.cpp
```diff
@@ -1399,6 +1399,141 @@ PreparedQuery PrepareDumpQuery(ParsedQuery parsed_query, std::map<std::string, T
                        RWType::R};
 }
 
+std::vector<std::vector<TypedValue>> AnalyzeGraphQueryHandler::AnalyzeGraphCreateStatistics(
+    const std::span<std::string> labels, DbAccessor *execution_db_accessor, storage::Storage::Accessor *dba) {
+  using VertexAccessorIterable = decltype(std::declval<query::DbAccessor>().Vertices(storage::View::OLD));
```
jbajic, line 1404: Remove `using` and use `auto later on`

src/query/frontend/opencypher/grammar/MemgraphCypher.g4
```diff
@@ -293,6 +296,8 @@ showUsersForRole : SHOW USERS FOR role=userOrRoleName ;
 
 dumpQuery: DUMP DATABASE ;
 
+analyzeGraphQuery: ANALYZE GRAPH ( ON LABELS ( listOfColonSymbolicNames | ASTERISK ) ) ? ( DELETE STATISTICS ) ? ;
```
Josipmrden, line 299: what happens if I specify

ANALYZE GRAPH ON LABELS * DELETE STATISTICS? 

I think it's a mutually exclusive case, should be separated with a |

src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -239,6 +239,23 @@ antlrcpp::Any CypherMainVisitor::visitDumpQuery(MemgraphCypher::DumpQueryContext
...
+    analyze_graph_query->labels_ =
+        std::any_cast<std::vector<std::string>>(ctx->listOfColonSymbolicNames()->accept(this));
+  } else {
+    analyze_graph_query->labels_.emplace_back("*");
+  }
+  if (ctx->DELETE()) {
+    analyze_graph_query->action_ = AnalyzeGraphQuery::Action::DELETE;
+  } else {
+    analyze_graph_query->action_ = AnalyzeGraphQuery::Action::ANALYZE;
+  }
```
Josipmrden, line 254: do handle analyze and throw if both not present

Josipmrden: we want a good exception if an error happens with syntax here

src/query/plan/rewrite/index_lookup.hpp
```diff
@@ -557,17 +561,25 @@ class IndexLookupRewriter final : public HierarchicalLogicalOperatorVisitor {
...
+
+        if (!found || (vertex_count * 10 < found->vertex_count) ||
+            (new_stats.has_value() && found->index_stats.has_value() && (vertex_count / 10 <= found->vertex_count) &&
+             (utils::LessThanDecimal(new_stats->avg_group_size, found->index_stats->avg_group_size) ||
+              (utils::ApproxEqualDecimal(found->index_stats->avg_group_size, new_stats->avg_group_size) &&
+               utils::LessThanDecimal(new_stats->stat_value, found->index_stats->stat_value)))) ||
+            found->vertex_count > vertex_count ||
+            (found->vertex_count == vertex_count && is_better_type(filter.property_filter->type_))) {
+          found = LabelPropertyIndex{label, filter, vertex_count, new_stats};
         }
```
Josipmrden, line 577: extract this into another private method

src/storage/v2/indices.hpp
```diff
@@ -131,6 +131,10 @@ class LabelIndex {
   Config::Items config_;
 };
 
+struct IndexStats {
+  double stat_value, avg_group_size;
```
Josipmrden, line 135: what is stat_value?
maybe better naming

src/query/frontend/ast/cypher_main_visitor.cpp
Josipmrden: or maybe I'm wrong

src/storage/v2/indices.cpp
```diff
@@ -688,6 +688,46 @@ int64_t LabelPropertyIndex::ApproximateVertexCount(LabelId label, PropertyId pro
...
+  for (auto it = stats_.cbegin(); it != stats_.cend();) {
+    if (it->first.first == label) {
+      deleted_indexes.push_back(it->first);
+      it = stats_.erase(it);
+    } else {
+      ++it;
+    }
+  }
+  return deleted_indexes;
+}
```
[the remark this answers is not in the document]
Josipmrden: I think he is doing it per label and deletion also happens per label (for every label property index present)

src/query/interpreter.cpp
```diff
@@ -1399,6 +1399,141 @@ PreparedQuery PrepareDumpQuery(ParsedQuery parsed_query, std::map<std::string, T
...
+    result.emplace_back(chi_squared_stat);
+    results.push_back(std::move(result));
+  });
+  return results;
+}
+
+std::vector<std::vector<TypedValue>> AnalyzeGraphQueryHandler::AnalyzeGraphDeleteStatistics(
+    const std::span<std::string> labels, storage::Storage::Accessor *dba) {
+  std::vector<std::pair<storage::LabelId, storage::PropertyId>> loc_results;
+  if (labels[0] == "*") {
```
Josipmrden, line 1462: kAsterisk constant put somewhere

src/query/interpreter.cpp
```diff
@@ -1399,6 +1399,141 @@ PreparedQuery PrepareDumpQuery(ParsedQuery parsed_query, std::map<std::string, T
...
+    const std::span<std::string> labels, DbAccessor *execution_db_accessor, storage::Storage::Accessor *dba) {
+  using VertexAccessorIterable = decltype(std::declval<query::DbAccessor>().Vertices(storage::View::OLD));
+  using LPIndex = std::pair<storage::LabelId, storage::PropertyId>;
+
+  auto results = std::vector<std::vector<TypedValue>>();
+  std::map<LPIndex, std::map<storage::PropertyValue, int64_t>> counter;
+
+  // Preprocess labels to avoid later checks
+  std::vector<LPIndex> indices_info = execution_db_accessor->ListAllIndices().label_property;
+  if (labels[0] != "*") {
```
Josipmrden, line 1412: kAsterisk put constant somewhere

src/query/interpreter.cpp
```diff
@@ -2358,6 +2493,9 @@ Interpreter::PrepareResult Interpreter::Prepare(const std::string &query_string,
     } else if (utils::Downcast<IndexQuery>(parsed_query.query)) {
       prepared_query = PrepareIndexQuery(std::move(parsed_query), in_explicit_transaction_,
                                          &query_execution->notifications, interpreter_context_);
+    } else if (utils::Downcast<AnalyzeGraphQuery>(parsed_query.query)) {
```
Josipmrden, line 2496: do we need both the db accessor and the storage accessor?

tests/e2e/analyze_graph/optimize_indexes.py
```diff
@@ -0,0 +1,344 @@
...
+    assert analyze_graph_results[first_index][3] == 100  # num groups
+    assert analyze_graph_results[first_index][4] == 1  # average group size
+    assert analyze_graph_results[first_index][5] == 0  # stat value
+    assert analyze_graph_results[1 - first_index][0] == "Label"
+    assert analyze_graph_results[1 - first_index][1] == "id2"
+    assert analyze_graph_results[1 - first_index][2] == 50  # num estimation nodes
+    assert analyze_graph_results[1 - first_index][3] == 50  # num groups
+    assert analyze_graph_results[1 - first_index][4] == 1  # average group size
+    assert analyze_graph_results[1 - first_index][5] == 0  # stat value
+    expected_explain_after_analysis = [
```
Josipmrden, line 198: change stat value comment to chi squared value

Josipmrden (commented):
Needs change but I'm not in the list of reviewers

src/storage/v2/indices.cpp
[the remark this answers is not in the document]
as51340: Yes, everything is done per label.

src/utils/math.hpp
```diff
@@ -64,4 +65,29 @@ constexpr std::optional<uint64_t> RoundUint64ToMultiple(uint64_t val, uint64_t m
...
+
+constexpr double ChiSquaredValue(double observed, double expected) {
+  if (utils::ApproxEqualDecimal(expected, 0.0)) {
+    return std::numeric_limits<double>::max();
+  }
+  return (observed - expected) * (observed - expected) / expected;
+}
```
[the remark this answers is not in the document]
as51340: Didn't know about this, tnxx!

src/storage/v2/storage.hpp
```diff
@@ -264,6 +265,30 @@ class Storage final {
       return storage_->indices_.label_property_index.ApproximateVertexCount(label, property, lower, upper);
     }
 
+    std::optional<storage::IndexStats> GetIndexStats(const storage::LabelId &label,
+                                                     const storage::PropertyId &property) const {
+      return storage_->indices_.label_property_index.GetIndexStats(label, property);
+    }
+
+    std::vector<std::pair<LabelId, PropertyId>> ClearIndexStats() {
+      return storage_->indices_.label_property_index.ClearIndexStats();
```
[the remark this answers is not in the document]
as51340: I would say yes because it will be documented that all deleted indices will be returned + it will be in the documentation that the command should be used after indices were created.

src/query/frontend/ast/ast.lcp
```diff
@@ -2403,6 +2403,26 @@ cpp<#
...
+
+    DEFVISITABLE(QueryVisitor<void>);
+    cpp<#)
+  (:private
+    #>cpp
+    friend class AstStorage;
+    cpp<#)
+  (:serialize (:slk))
+  (:clone))
+
```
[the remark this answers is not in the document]
as51340: It is in the master still so it should be ok. But in the future no need to keep it up to date, that's that Jure said

src/query/frontend/opencypher/grammar/MemgraphCypher.g4
as51340: > what happens if I specify:
It works okay.

Why is exclusive? The user can say which labels to delete

src/query/frontend/ast/cypher_main_visitor.cpp
as51340: I understand what you mean but wouldn't agree because there is no way to handle analyze except that is not delete + the exception will be thrown any way.

src/storage/v2/indices.hpp
as51340: > what is stat_value?

Statistical value. I will change it to the statistic

src/query/interpreter.cpp
as51340: True we don't, I'll refactor

src/query/frontend/opencypher/grammar/MemgraphCypher.g4
antoniofilipovic: I think this is correct how it is.

src/query/frontend/ast/cypher_main_visitor.cpp
antoniofilipovic: Exception will be thrown if there is something missing, i.e. you write `analyze graph on labels * delete ` without statistics, so I think all good here.

src/query/plan/rewrite/index_lookup.hpp
```diff
@@ -534,6 +539,23 @@ class IndexLookupRewriter final : public HierarchicalLogicalOperatorVisitor {
...
+     * @param new_stats: Label-property index candidate.
+     * @param vertex_count: New index's number of vertices.
+     */
+    auto compare_indices = [](std::optional<LabelPropertyIndex> &found, std::optional<storage::IndexStats> &new_stats,
+                              int vertex_count) {
+      return new_stats.has_value() && found->index_stats.has_value() && (vertex_count / 10 <= found->vertex_count) &&
+             (utils::LessThanDecimal(new_stats->avg_group_size, found->index_stats->avg_group_size) ||
+              (utils::ApproxEqualDecimal(found->index_stats->avg_group_size, new_stats->avg_group_size) &&
+               utils::LessThanDecimal(new_stats->statistic, found->index_stats->statistic)));
+    };
```
antoniofilipovic, line 557: You did a good job as it is much cleaner now.

Few things that I noticed as I was reading description.

First, I thought this function does all the comparing due to description. But what it actually does is it compares part, and other part is below. I would say this is good approach, and to make it better we have few options.
One is to add all comparisons from below here. From below meaning:

```
if (!found || (vertex_count * 10 < found->vertex_count) || compare_indices(found, new_stats, vertex_count) ||
            found->vertex_count > vertex_count ||
            (found->vertex_count == vertex_count && is_better_type(filter.property_filter->type_))) {
          found = LabelPropertyIndex{label, filter, vertex_count, new_stats};
}
```

Reason why is because function is named `compare_indices` and other parts of `if` in my opinion also compare indices. For example, why indices comparison is not `(vertex_count * 10 < found->vertex_count)` as it compares num vertices?

Then again there will be problem of multiple conditions and unreadability, and I would again opt for splitting accross multiple ifs:

```
if (!found)
    found = LabelPropertyIndex{label, filter, vertex_count, new_stats};
else if (vertex_count * 10 < found->vertex_count) 
    found = LabelPropertyIndex{label, filter, vertex_count, new_stats};
else if(...)
    found = LabelPropertyIndex{label, filter, vertex_count, new_stats};
```

But this lambda seems to me like good approach, and I would just group it here.

Other option is to keep it without lambda and split it with multiple ifs.

src/query/interpreter.cpp
```diff
@@ -1399,6 +1399,139 @@ PreparedQuery PrepareDumpQuery(ParsedQuery parsed_query, std::map<std::string, T
...
+      if (std::find(labels.begin(), labels.end(), execution_db_accessor->LabelToName(it->first)) == labels.end()) {
+        indices_info.erase(it);
+      } else {
+        ++it;
+      }
+    }
+  }
+  // Iterate over all indexed vertices
+  std::for_each(indices_info.begin(), indices_info.end(), [execution_db_accessor, &counter](const LPIndex &index_info) {
+    auto vertices = execution_db_accessor->Vertices(storage::View::OLD, index_info.first, index_info.second);
```
antoniofilipovic, line 1422: Nice!

tests/e2e/analyze_graph/optimize_indexes.py
```diff
@@ -0,0 +1,344 @@
...
+    assert len(delete_stats_results) == 2
+    if delete_stats_results[0][1] == "id1":
+        first_index = 0
+    else:
+        first_index = 1
+    assert delete_stats_results[first_index][0] == "Label"
+    assert delete_stats_results[first_index][1] == "id1"
+    assert delete_stats_results[1 - first_index][0] == "Label"
+    assert delete_stats_results[1 - first_index][1] == "id2"
+    # After deleting statistics, id2 should be chosen because it has less vertices
```
antoniofilipovic, line 49: My thinking is that it might be chosen due to bug in code. It makes sense to me to run explain query before you run delete and add check that `id1` is properly chosen.

tests/e2e/analyze_graph/optimize_indexes.py
```diff
@@ -0,0 +1,344 @@
...
+# the Business Source License, use of this software will be governed
+# by the Apache License, Version 2.0, included in the file
+# licenses/APL.txt.
+
+import sys
+
+import pytest
+from common import connect, execute_and_fetch_all
+
+# E2E tests for checking query semantic
```
antoniofilipovic, line 17: One top level comment. It seems to me there is no test when we have nodes with different labels here. I would say we need such test also, to cover for such case.

src/query/interpreter.cpp
```diff
@@ -1399,6 +1399,139 @@ PreparedQuery PrepareDumpQuery(ParsedQuery parsed_query, std::map<std::string, T
...
+
+  // Preprocess labels to avoid later checks
+  std::vector<LPIndex> indices_info = execution_db_accessor->ListAllIndices().label_property;
+  if (labels[0] != kAsterisk) {
+    for (auto it = indices_info.cbegin(); it != indices_info.cend();) {
+      if (std::find(labels.begin(), labels.end(), execution_db_accessor->LabelToName(it->first)) == labels.end()) {
+        indices_info.erase(it);
+      } else {
+        ++it;
+      }
```
jbajic, line 1417: How does this work? Should you not assign the new iterator after erasing, since the iterator you are holding is invalidated and you cannot increment on it anymore, e.g.
```suggestion
      if (std::find(labels.begin(), labels.end(), execution_db_accessor->LabelToName(it->first)) == labels.end()) {
        it = indices_info.erase(it);
      } else {
        ++it;
      }
```

src/query/interpreter.hpp
```diff
@@ -49,6 +49,8 @@ namespace memgraph::query {
 
 inline constexpr size_t kExecutionMemoryBlockSize = 1UL * 1024UL * 1024UL;
 
+constexpr std::string_view kAsterisk = "*";
```
jbajic, line 52: There already exists `kAsterisk` in `src/auth/models.hpp` I would think about reusing it, but to do that, it makes sense to move it someplace else. I don't have a strong opinion on this, feel free to ignore it.

jbajic (approved):

src/query/interpreter.cpp
as51340: I deleted this it accidentally, it should be like you wrote...

tests/e2e/analyze_graph/optimize_indexes.py
```diff
@@ -0,0 +1,344 @@
...
+        first_index = 0
+    else:
+        first_index = 1
+    # Check results
+    assert analyze_graph_results[first_index][0] == "Label"
+    assert analyze_graph_results[first_index][1] == "id1"
+    assert analyze_graph_results[first_index][2] == 100  # num estimation nodes
+    assert analyze_graph_results[first_index][3] == 100  # num groups
+    assert analyze_graph_results[first_index][4] == 1  # average group size
+    assert analyze_graph_results[first_index][5] == 0  # chi-squared value
```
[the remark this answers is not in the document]
as51340: yep just needs to be == (...) not []

src/query/interpreter.cpp
```diff
@@ -1399,6 +1399,139 @@ PreparedQuery PrepareDumpQuery(ParsedQuery parsed_query, std::map<std::string, T
...
+    std::for_each(vertices.begin(), vertices.end(), [&index_info, &counter](const auto &vertex) {
+      counter[index_info][*vertex.GetProperty(storage::View::OLD, index_info.second)]++;
+    });
+  });
+
+  results.reserve(counter.size());
+  std::for_each(counter.begin(), counter.end(), [&results, execution_db_accessor](const auto &counter_entry) {
+    const auto &[label_property, values_map] = counter_entry;
+    std::vector<TypedValue> result;
+    result.reserve(6);
```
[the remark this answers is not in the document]
as51340: this is purely optimization so I will leave it, otherwise I have to change semantic

src/utils/math.hpp
```diff
@@ -64,4 +68,28 @@ constexpr std::optional<uint64_t> RoundUint64ToMultiple(uint64_t val, uint64_t m
   return (numerator / multiple) * multiple;
 }
 
+template <typename T>
+concept FloatingPoint = std::is_floating_point_v<T>;
+
+template <FloatingPoint T>
+bool ApproxEqualDecimal(T a, T b) {
+  // return std::fabs(a - b) <=
+  //  ((std::fabs(a) < std::fabs(b) ? std::fabs(b) : std::fabs(a)) * s);
```
antoniofilipovic, line 77: This is commented, I think you can remove it.

src/utils/math.hpp
```diff
@@ -64,4 +68,28 @@ constexpr std::optional<uint64_t> RoundUint64ToMultiple(uint64_t val, uint64_t m
...
+  return boost::math::relative_difference(a, b) < std::numeric_limits<T>::epsilon();
+}
+
+template <FloatingPoint T>
+bool LessThanDecimal(T a, T b) {
+  // return (b - a) >
+  //  ((std::fabs(a) < std::fabs(b) ? std::fabs(b) : std::fabs(a)) * std::numeric_limits<double>::epsilon());
```
antoniofilipovic, line 84: And this too

antoniofilipovic (approved):
There are some commented things you left, remove that and it should be ready to ship it

PR #814, opened 2023-03-02
[master < MG-support-amzn2] Add support for Amazon Linux 2 and stop generating C++ using Lisp/LCP

jbajic:
[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR


src/query/plan/operator_type_info.cpp
```diff
@@ -0,0 +1,144 @@
...
+// License, and you may not use this file except in compliance with the Business Source License.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#include "query/plan/operator.hpp"
+
+const memgraph::utils::TypeInfo memgraph::query::plan::LogicalOperator::kType{0x120A36D770B69648ULL, "LogicalOperator",
```
gitbuda, line 14: How is this actually generated? We should probably document here how to add a new `kType` here because it's not 100% clear in the first place?

jbajic: Not sure I will check it out and see what is the idea behind this

jbajic: For now I would just like to get it working in Amazon Linux 2, and then see if all of this makes sense

jbajic:
Last successful execution https://github.com/memgraph/memgraph/actions/runs/4346884916

src/query/plan/operator_type_info.cpp
jbajic: I've checkout out and talked with Benjo, the hashing of the operator classes served a purpose to differentiate between operators, and when sending the same over RPC so it made it easier to confirm that the operators have not changed their compatibility (signature, internal functions, etc.). SO it is safe to replace this with an enum since we are not using the old style RPC.

gitbuda (approved):

antaljanosbenjamin (approved):

PR #825, opened 2023-03-10
Change PullPlan to use specific Memory Resource for LOAD CSV

antoniofilipovic:
This commit changes following:
- introduces std::variant to struct QueryExecution due to the fact that inside such memory can be long csv lines, so such memory can be reused
- changes PullPlan to use PoolResource with 1 chunk per block and 16384 characters of maximal length
- such change will enable to read long csv lines and reuse such memory as on every read of line you can always reuse old memory

[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR

### DIFF
--- a/src/query/interpreter.cpp
+++ b/src/query/interpreter.cpp
@@ -19,6 +19,7 @@
 #include <cstdint>
 #include <functional>
 #include <limits>
+#include <memory>
 #include <optional>
 #include <unordered_map>
 #include <variant>
@@ -63,6 +64,7 @@
 #include "utils/settings.hpp"
 #include "utils/string.hpp"
 #include "utils/tsc.hpp"
+#include "utils/typeinfo.hpp"
 #include "utils/variant_helpers.hpp"
 
 namespace EventCounter {
@@ -979,7 +981,8 @@ struct PullPlan {
                     std::optional<size_t> memory_limit = {});
   std::optional<plan::ProfilingStatsWithTotalTime> Pull(AnyStream *stream, std::optional<int> n,
                                                         const std::vector<Symbol> &output_symbols,
-                                                        std::map<std::string, TypedValue> *summary);
+                                                        std::map<std::string, TypedValue> *summary,
+                                                        std::optional<std::vector<Clause *>> maybe_clauses);
 
  private:
   std::shared_ptr<CachedPlan> plan_ = nullptr;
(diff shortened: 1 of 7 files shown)


antoniofilipovic:
I can add unit tests if necessary.

src/query/interpreter.cpp
```diff
@@ -976,7 +978,7 @@ struct PullPlan {
   explicit PullPlan(std::shared_ptr<CachedPlan> plan, const Parameters &parameters, bool is_profile_query,
                     DbAccessor *dba, InterpreterContext *interpreter_context, utils::MemoryResource *execution_memory,
                     std::optional<std::string> username, TriggerContextCollector *trigger_context_collector = nullptr,
-                    std::optional<size_t> memory_limit = {});
+                    std::optional<size_t> memory_limit = {}, bool contains_csv = false);
```
antaljanosbenjamin, line 981: I would phrase this differently: from the `PullPlan` perspective it doesn't matter if it is a CSV or not: the only information it needs is whether to use the monotonic buffer or not. I would focus on that: `use_monotonic_memory`.

src/query/interpreter.cpp
```diff
@@ -1037,21 +1042,28 @@ std::optional<plan::ProfilingStatsWithTotalTime> PullPlan::Pull(AnyStream *strea
...
-  // the stack allocated buffer.
-  // Also, we want to throw only when the query engine requests more memory and not the storage
-  // so we add the exception to the allocator.
-  // TODO (mferencevic): Tune the parameters accordingly.
-  utils::PoolResource pool_memory(128, 1024, &monotonic_memory, utils::NewDeleteResource());
-  std::optional<utils::LimitedMemoryResource> maybe_limited_resource;
+  utils::MonotonicBufferResource monotonic_memory{&stack_data[0], stack_size, &resource_with_exception};
+  std::optional<utils::PoolResource> pool_memory;
+
+  if (use_monotonic_memory_) {
```
antaljanosbenjamin, line 1050:
```suggestion
  if (!use_monotonic_memory_) {
```

antaljanosbenjamin (changes requested):

src/query/interpreter.cpp
```diff
@@ -1037,21 +1042,28 @@ std::optional<plan::ProfilingStatsWithTotalTime> PullPlan::Pull(AnyStream *strea
...
+    pool_memory.emplace(1, kExecutionPoolMaxBlockSize, utils::NewDeleteResource(), utils::NewDeleteResource());
```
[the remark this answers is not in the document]
antoniofilipovic: The idea for `LOAD CSV` is to reuse memory as much as possible, and not to hold or take more than 1 row needs. When the row is processed and the next row comes into place, it would be ideal if we only reuse memory that was already allocated.

kExecutionPoolMaxBlockSize is the size that should guarantee that bigger chunks (bigger lines) should fit inside memory (up to 8000 characters if I am not wrong). And in such cases, it will reuse `Chunk` objects inside `Pool`. Otherwise, it will use unpooled memory. 

To answer the question, setting `blocks_per_chunk`=1 should help to reuse memory more frequently on LOAD CSV and not to allocate more than we need. Idea is that columns inside LOAD  CSV will be of similar size and that we can on for each row reuse memory that we allocated for the row before. If I am not wrong when block_per_chunk is set to 1, for each chunk we have only 1 block inside Chunk, and it is the power of 2 of size, it will only allocate 1 block, as that is how much we will need for that row.

 Don't know if it makes sense to you.

src/query/interpreter.cpp
```diff
@@ -976,7 +978,7 @@ struct PullPlan {
   explicit PullPlan(std::shared_ptr<CachedPlan> plan, const Parameters &parameters, bool is_profile_query,
                     DbAccessor *dba, InterpreterContext *interpreter_context, utils::MemoryResource *execution_memory,
                     std::optional<std::string> username, TriggerContextCollector *trigger_context_collector = nullptr,
-                    std::optional<size_t> memory_limit = {});
+                    std::optional<size_t> memory_limit = {}, bool use_monotonic_memory = true);
```
Josipmrden, line 981: do we set any time this to false?

src/query/interpreter.cpp
```diff
@@ -1000,16 +1002,19 @@ struct PullPlan {
   // we have to keep track of any unsent results from previous `PullPlan::Pull`
   // manually by using this flag.
   bool has_unsent_results_ = false;
+
+  bool use_monotonic_memory_;
```
Josipmrden, line 1006: document this why is it important, and when is it set to true and when to false

src/query/interpreter.cpp
```diff
@@ -1037,21 +1042,28 @@ std::optional<plan::ProfilingStatsWithTotalTime> PullPlan::Pull(AnyStream *strea
...
+    // Also, we want to throw only when the query engine requests more memory and not the storage
+    // so we add the exception to the allocator.
+    // TODO (mferencevic): Tune the parameters accordingly.
+    pool_memory.emplace(128, 1024, &monotonic_memory, utils::NewDeleteResource());
+  }
 
+  std::optional<utils::LimitedMemoryResource> maybe_limited_resource;
   if (memory_limit_) {
-    maybe_limited_resource.emplace(&pool_memory, *memory_limit_);
+    maybe_limited_resource.emplace(&*pool_memory, *memory_limit_);
```
Josipmrden, line 1063: what does this do

src/query/interpreter.hpp
```diff
@@ -48,6 +48,7 @@ extern const Event FailedQuery;
 namespace memgraph::query {
 
 inline constexpr size_t kExecutionMemoryBlockSize = 1UL * 1024UL * 1024UL;
+inline constexpr size_t kExecutionPoolMaxBlockSize = 256UL;  // 2 ^ 15
```
Josipmrden, line 51: this is not 2^15?

src/query/interpreter.cpp
```diff
@@ -2310,18 +2326,41 @@ Interpreter::PrepareResult Interpreter::Prepare(const std::string &query_string,
...
+    if (utils::Downcast<CypherQuery>(parsed_query.query)) {
+      auto *cypher_query = utils::Downcast<CypherQuery>(parsed_query.query);
+      if (const auto &clauses = cypher_query->single_query_->clauses_;
+          std::any_of(clauses.begin(), clauses.end(),
+                      [](const auto *clause) { return clause->GetTypeInfo() == LoadCsv::kType; })) {
+        query_executions_[query_executions_.size() - 1] = std::make_unique<QueryExecution>(
+            utils::PoolResource(1, kExecutionPoolMaxBlockSize, utils::NewDeleteResource(), utils::NewDeleteResource()));
+        query_execution_ptr = &query_executions_.back();
+      }
+    }
```
Josipmrden, line 2352: comment this part of code why we are doing this

Josipmrden (commented):
minor comments

src/query/interpreter.cpp
antoniofilipovic: it just sends address to `optional` (which is used)

src/query/interpreter.hpp
antoniofilipovic: nice catch

src/query/interpreter.cpp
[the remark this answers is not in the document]
antoniofilipovic: @antaljanosbenjamin I made a mistake and my math was a bit of. `kExecutionPoolMaxBlockSize` was intended to be set to 2^15 which is 32 768 bytes as it would allow us to store 4K characters. I checked and for 100 characters we need 800 bytes for the TypedValue vector. 

I agree it is a bit aggressive. My idea and this point is to allocate only what we need. I tested with different sizes 1, 4, and 8 and for LOAD CSV there was no difference in the performance of the query. It seems like a cool idea to me because if there is a need to allocate more, it will happen, but once it does, MemoryResource can reuse it later.

This memory resource is used in EvalContext and since we are dealing with LOAD CSV it should be able to reuse memory, and not suffer performance issues on allocations.

vpavicic:
Import using the LOAD CSV clause has been further improved by ...

Finish the sentence... but something very simple :D

antaljanosbenjamin (approved):

antoniofilipovic:
> g the LOAD CSV clause has been further improved by ...

...has been further improved by using a memory allocator which will reuse memory blocks allocated during processing the `LOAD CSV` query @vpavicic

PR #826, opened 2023-03-12
[master < ] Extend memgraph::slk

gitbuda:
because SLK is simple, convenient and efficient

- [ ] Add loopback for simple serialization
- [ ] Revise `Builder` and `Reader` types
- [ ] Add compression as an option on the serialization side


gitbuda:
Non-relevant any more

PR #827, opened 2023-03-13
[🍍 <] Add Path and Graph

gitbuda:
- [ ] Implement full graph fetching
- [ ] Implement partial graph fetching
- [ ] Implement create full graph creation + partial graph fetching + global algorithm run benchmark
- [ ] Experiment with virtual `Message`

### DIFF
--- a/tests/simulation/common.hpp
+++ b/tests/simulation/common.hpp
@@ -118,7 +118,10 @@ class MockedShardRsm {
     return resp;
   }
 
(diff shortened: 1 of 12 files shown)


tests/simulation/common.hpp
```diff
@@ -118,6 +118,8 @@ class MockedShardRsm {
     return resp;
   }
 
+  msgs::GraphResponse ReadImpl(msgs::GraphRequest rqst) { LOG_FATAL("Implement Simulator ReadImpl GraphRequest"); }
```
gitbuda, line 121: This seems like not the right function signature (vector vs response)

PR #837, opened 2023-03-23
[master < T613] Ommit label based access control for users with global graph visibility

Josipmrden:
Here is the fix for improving performance on users who have global visibility on the graph, it is the same if label based access control is or is not performed as all checks will grant the ability for the user to perform an operation. Existing tests can prove that the fix is actually working as we have many label based auth methods. Since the PullPlan is in the anonymous namespace I would not change the codebase with an additional unit test.

[master < Task] PR
- [X] Check, and update documentation if necessary [not needed]
- Git message and release note: Improved performance on label based access control for users who have global visibility on the graph

- [X] Tag someone from docs team in the comments @vpavicic


src/query/interpreter.cpp
```diff
@@ -1018,7 +1018,14 @@ PullPlan::PullPlan(const std::shared_ptr<CachedPlan> plan, const Parameters &par
...
 #ifdef MG_ENTERPRISE
   if (license::global_license_checker.IsEnterpriseValidFast() && username.has_value() && dba) {
-    ctx_.auth_checker = interpreter_context->auth_checker->GetFineGrainedAuthChecker(*username, dba);
+    auto auth_checker = interpreter_context->auth_checker->GetFineGrainedAuthChecker(*username, dba);
+
+    // if the user has global privileges to read, edit and write anything, we don't need to perform authorization
+    // otherwise, we do assign the auth checker to check for label access control
+    if (!auth_checker->HasGlobalPrivilegeOnVertices(AuthQuery::FineGrainedPrivilege::CREATE_DELETE) ||
+        !auth_checker->HasGlobalPrivilegeOnEdges(AuthQuery::FineGrainedPrivilege::CREATE_DELETE)) {
+      ctx_.auth_checker = std::move(auth_checker);
```
antoniofilipovic, line 1027: As this is unique ptr, I think there is no need to move. On assignement, it will be "moved" (ownership) to the ctx.auth_checker.

https://en.cppreference.com/w/cpp/memory/unique_ptr

"the managing unique_ptr object is assigned another pointer via operator= or reset()."

antoniofilipovic (approved):

PR #839, opened 2023-03-24
[master < T614] Add SHA256 password encryption

Josipmrden:
The task includes adding SHA as part of password authorization, which is less secure but a faster way of executing authentication. BCRYPT is still going to be our main authentication method for passwords but an additional flag will be present to run Memgraph with SHA algorithm.

[master < Task] PR


src/auth/crypto.cpp
```diff
@@ -1,18 +1,31 @@
...
+inline constexpr std::string_view default_password_encryption = "bcrypt";
+inline constexpr std::string_view sha256_password_encryption = "sha256";
+inline constexpr std::string_view sha256_1024_iterations_password_encryption = "sha256-1024";
+
+inline constexpr uint64_t ONE_SHA_ITERATION = 1;
+inline constexpr uint64_t MULTIPLE_SHA_ITERATIONS = 1;
+
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
+DEFINE_string(password_encryption_algorithm, default_password_encryption.data(),
+              "The password encryption algorithm used for authentication");
```
gitbuda, line 25: We should put the options into the flag description, does GFlags has some way of specifying an enum? So, it's less error prone if we add more options?

gitbuda: Maybe somehow with the flag validation logic -> https://github.com/memgraph/memgraph/blob/master/src/utils/flag_validation.hpp

tests/mgbench/client.cpp
```diff
@@ -398,8 +400,12 @@ void ExecuteWorkload(
     final_duration += worker_duration[i];
   }
 
+  auto end_workload = std::chrono::steady_clock::now();
+  auto workload_duration = std::chrono::duration_cast<std::chrono::duration<double>>(end_workload - start_workload);
+
   final_duration /= FLAGS_num_workers;
   nlohmann::json summary = nlohmann::json::object();
+  summary["workload_duration"] = workload_duration.count();
```
antejavor, line 408: @Josipmrden Not sure did you plan to use these measurements anywhere. I know you used them for debugging, but "workload duration" is a bit wrong since it includes thread init +  connection time.  And we have workload as an important reference so I would call this something like **total_duration**, later we can have we have "thread_init_duration" "connection_duration" + "workload_duration" . If you are not going to use it, feel free to kick it out.

Josipmrden: will rename, but it can come in handy since we don't know how much was the total time of execution, so will rename it

src/auth/crypto.cpp
gitbuda: 💪

src/auth/crypto.cpp
```diff
@@ -1,19 +1,50 @@
...
 
 #include "auth/exceptions.hpp"
+#include "utils/flag_validation.hpp"
+
+inline constexpr std::string_view default_password_encryption = "bcrypt";
+inline constexpr std::string_view sha256_password_encryption = "sha256";
+inline constexpr std::string_view sha256_1024_iterations_password_encryption = "sha256-1024";
+
+inline constexpr uint64_t ONE_SHA_ITERATION = 1;
+inline constexpr uint64_t MULTIPLE_SHA_ITERATIONS = 1;
```
antoniofilipovic, line 25: should this be 1024?

antoniofilipovic:
@Josipmrden minor comments and suggestions. I think we should solve issue with depricated SHA256, it will complain all the time otherwise.

src/auth/crypto.cpp
```diff
@@ -38,5 +69,59 @@ bool VerifyPassword(const std::string &password, const std::string &hash) {
...
   return ret == 0;
 }
+}  // namespace BCrypt
+
+namespace SHA {
+std::string EncryptPassword(const std::string &password, const uint64_t number_of_iterations) {
+  unsigned char hash[SHA256_DIGEST_LENGTH];
+
+  SHA256_CTX sha256;
+  SHA256_Init(&sha256);
```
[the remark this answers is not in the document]
Josipmrden: It's deprecated to ensure backwards compatibility with older linux distributions. I don't have that functionality since newer version of openssl is not supported

antoniofilipovic (approved):
Looks good!

PR #846, opened 2023-03-29
Add SASL librdkafka dependency

gitbuda:
- [ ] Ensure static lib is used so that users who install the Memgraph package don't have to install anything in addition
- [ ] Install SASL development library to all build machines


gitbuda:
closing this PR, because this will be done as a part of #862

PR #850, opened 2023-03-29
Add experimental/v1 ON_DISK_TRANSACTIONAL storage

### LINKED ISSUE #842: Add on-disk transactional storage mode (v1/experimental)
The intention of this feature is to provide a disk-based storage option under Memgraph v2. The main reasons are:
* being able to store data on disk because it's cheaper
* being able to store more data because usually the amount of RAM is limited

The goal is to make memgraph on-disk storage as fast as possible while being able to store as much data as possible.

In the end, it should be possible to switch at runtime (not necessarily online) between the following storage modes:
* `IN_MEMORY_TRANSACTIONAL`
* `IN_MEMORY_ANALYTICAL`
* `ON_DISK_TRANSACTIONAL`

Additional requirements:
* Out-of-order store of vertices and edges (an edge can come before SRC and DST vertices)

gitbuda:
Closes #842


andrejtonev:
This merge has dropped the performance on certain benchmarks by some 10% (check single_vertex_read, single_edge_read, etc.)
Double check if this is due to a fundamental memgraph change or something to do with the benchmark itself.

src/storage/v2/storage.cpp
```diff
@@ -498,1825 +60,45 @@ Storage::Accessor::Accessor(Storage *storage, IsolationLevel isolation_level, St
...
-    std::vector<Storage::ReplicaInfo> replica_info;
-    replica_info.reserve(clients.size());
-    std::transform(
-        clients.begin(), clients.end(), std::back_inserter(replica_info), [](const auto &client) -> ReplicaInfo {
-          return {client->Name(), client->Mode(), client->Endpoint(), client->State(), client->GetTimestampInfo()};
-        });
-    return replica_info;
-  });
-}
+StorageMode Storage::GetStorageMode() const { return storage_mode_; }
```
andrejtonev, line 101: Is `storage_mode_` protected?
Seems like we are not using the accessor when reading it. Can this cause a race condition?
*similar story for the `isolation_level_`

andrejtonev (commented):
Just a few comments from today

src/query/interpreter.cpp
```diff
@@ -3165,6 +3306,9 @@ void Interpreter::Abort() {
   if (!db_accessor_) return;
 
   db_accessor_->Abort();
+  for (auto &qe : query_executions_) {
```
andrejtonev, line 3309: Could we just call `query_executions_.clear()`?

src/query/interpreter.cpp
```diff
@@ -3260,6 +3406,14 @@ void Interpreter::Commit() {
   utils::OnScopeExit clean_status(
       [this]() { transaction_status_.store(TransactionStatus::IDLE, std::memory_order_release); });
 
+  auto current_storage_mode = interpreter_context_->db->GetStorageMode();
+  auto creation_mode = db_accessor_->GetCreationStorageMode();
+  if (creation_mode != storage::StorageMode::ON_DISK_TRANSACTIONAL &&
+      current_storage_mode == storage::StorageMode::ON_DISK_TRANSACTIONAL) {
+    throw QueryException(
```
andrejtonev, line 3413: Switching to on-disk during a transaction should be impossible.

src/query/interpreter.hpp
```diff
@@ -208,11 +211,15 @@ class Interpreter;
...
  *
  */
+/// TODO: andi decouple in a separate file why here?
 struct InterpreterContext {
-  explicit InterpreterContext(storage::Storage *db, InterpreterConfig config,
+  explicit InterpreterContext(storage::Config storage_config, InterpreterConfig interpreter_config,
                               const std::filesystem::path &data_directory);
 
-  storage::Storage *db;
+  InterpreterContext(std::unique_ptr<storage::Storage> db, InterpreterConfig interpreter_config,
```
andrejtonev, line 219: `std::unique_ptr<storage::Storage> &&db`

src/query/interpreter.hpp
```diff
@@ -378,6 +382,13 @@ class Interpreter final {
       prepared_query.reset();
       std::visit([](auto &memory_resource) { memory_resource.Release(); }, execution_memory);
     }
+
+    void CleanRuntimeData() {
```
andrejtonev, line 386: Why is this needed?

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -1950,9 +1952,9 @@ void NextPermittedEdge(mgp_edges_iterator &it, const bool for_in) {
     const auto *auth_checker = it.source_vertex.graph->ctx->auth_checker.get();
     const auto view = it.source_vertex.graph->view;
     while (*impl_it != end) {
-      if (auth_checker->Has(**impl_it, memgraph::query::AuthQuery::FineGrainedPrivilege::READ)) {
-        const auto &check_vertex =
-            it.source_vertex.getImpl() == (*impl_it)->From() ? (*impl_it)->To() : (*impl_it)->From();
+      auto edgeAcc = **impl_it;
```
andrejtonev, line 1955: Is this a copy?

andrejtonev (commented):
Got up to storage

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1703 @@
...
+      }
+      case Delta::Action::ADD_LABEL: {
+        if (delta.label == label) {
+          MG_ASSERT(!has_label, "Invalid database state!");
+          has_label = true;
+        }
+        break;
+      }
+      case Delta::Action::DELETE_DESERIALIZED_OBJECT:
+      case Delta::Action::DELETE_OBJECT:
```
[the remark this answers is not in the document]
as51340: agree

src/query/interpreter.cpp
```diff
@@ -167,25 +171,27 @@ class ReplQueryHandler final : public query::ReplicationQueryHandler {
...
     }
     if (replication_role == ReplicationQuery::ReplicationRole::REPLICA) {
       if (!port || *port < 0 || *port > std::numeric_limits<uint16_t>::max()) {
         throw QueryRuntimeException("Port number invalid!");
       }
-      if (!db_->SetReplicaRole(
-              io::network::Endpoint(storage::replication::kDefaultReplicationServerIp, static_cast<uint16_t>(*port)))) {
+      if (!mem_storage->SetReplicaRole(
+              io::network::Endpoint(storage::replication::kDefaultReplicationServerIp, static_cast<uint16_t>(*port)),
+              storage::replication::ReplicationServerConfig{})) {
```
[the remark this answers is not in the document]
as51340: don't know, this is from before

src/storage/v2/storage.cpp
as51340: for both operations you exclusive lock (main_lock_)

src/query/interpreter.cpp
```diff
@@ -167,25 +171,27 @@ class ReplQueryHandler final : public query::ReplicationQueryHandler {
 
   /// @throw QueryRuntimeException if an error ocurred.
   void SetReplicationRole(ReplicationQuery::ReplicationRole replication_role, std::optional<int64_t> port) override {
+    auto *mem_storage = static_cast<storage::InMemoryStorage *>(db_);
     if (replication_role == ReplicationQuery::ReplicationRole::MAIN) {
-      if (!db_->SetMainReplicationRole()) {
+      if (!mem_storage->SetMainReplicationRole()) {
         throw QueryRuntimeException("Couldn't set role to main!");
       }
     }
```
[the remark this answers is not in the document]
as51340: why not exception? Usually it's thrown through whole codebase in such situations

src/query/interpreter.cpp
```diff
@@ -1188,11 +1199,38 @@ std::optional<plan::ProfilingStatsWithTotalTime> PullPlan::Pull(AnyStream *strea
...
+      config(interpreter_config),
+      streams{this, data_directory / "streams"} {
+  if (utils::DirExists(storage_config.disk.main_storage_directory)) {
+    db = std::make_unique<storage::DiskStorage>(storage_config);
+  } else {
+    db = std::make_unique<storage::InMemoryStorage>(storage_config);
+  }
+}
+
+InterpreterContext::InterpreterContext(std::unique_ptr<storage::Storage> db, InterpreterConfig interpreter_config,
```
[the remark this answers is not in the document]
as51340: yes

src/query/interpreter.cpp
```diff
@@ -2284,6 +2352,59 @@ PreparedQuery PrepareIsolationLevelQuery(ParsedQuery parsed_query, const bool in
...
+    if (SwitchingFromDiskToInMemory(current_mode, requested_mode)) {
+      throw utils::BasicException(
+          "You cannot switch from on-disk storage to in-memory storage while the database is running. "
+          "Please delete your data directory and restart the database. Once restarted, the Memgraph will automatically "
+          "start to use in-memory storage.");
+    }
+    if (SwitchingFromInMemoryToDisk(current_mode, requested_mode)) {
+      std::unique_lock main_guard{interpreter_context->db->main_lock_};
+
+      if (auto vertex_cnt_approx = interpreter_context->db->GetInfo().vertex_count; vertex_cnt_approx > 0) {
```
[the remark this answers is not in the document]
as51340: no because you can delete vertex only if:
1. has no edges
2. has edges but then you have to use detach delete

src/query/procedure/mg_procedure_impl.cpp
as51340: not sure what was happening here, cc @Darych take a look please

src/query/interpreter.hpp
as51340: cc @Darych

src/query/interpreter.hpp
as51340: yes

src/query/interpreter.cpp
as51340: cc @Darych  but I think no because we support prepared queries or something like this so we have to use it or something. There is a good reason for that I am sure.

src/query/interpreter.cpp
as51340: but I have a check:
```cpp
    throw StorageModeModificationInMulticommandTxException();
}
```

src/query/interpreter.cpp
```diff
@@ -2292,22 +2413,25 @@ PreparedQuery PrepareStorageModeQuery(ParsedQuery parsed_query, const bool in_ex
...
+  std::function<void()> callback;
 
-  auto callback = [storage_mode, interpreter_context]() -> std::function<void()> {
-    return [interpreter_context, storage_mode] { interpreter_context->db->SetStorageMode(storage_mode); };
-  }();
+  if (current_mode == storage::StorageMode::ON_DISK_TRANSACTIONAL ||
+      requested_mode == storage::StorageMode::ON_DISK_TRANSACTIONAL) {
+    callback = SwitchMemoryDevice(current_mode, requested_mode, interpreter_context).fn;
+  } else {
+    if (ActiveTransactionsExist(interpreter_context)) {
```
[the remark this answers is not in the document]
as51340: yes but for on-disk there is even stronger check: whether there is only one interpreter. I will rework this at some point because of Lab

src/query/interpreter.cpp
```diff
@@ -2284,6 +2352,59 @@ PreparedQuery PrepareIsolationLevelQuery(ParsedQuery parsed_query, const bool in
...
+      std::unique_lock main_guard{interpreter_context->db->main_lock_};
+
+      if (auto vertex_cnt_approx = interpreter_context->db->GetInfo().vertex_count; vertex_cnt_approx > 0) {
+        throw utils::BasicException(
+            "You cannot switch from in-memory storage to on-disk storage when the database "
+            "contains data. Please delete all entries from the database, run FREE MEMORY and then repeat this "
+            "query. ");
+      }
+
+      main_guard.unlock();
```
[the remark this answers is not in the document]
as51340: lock must be released before new DiskStorage has been created. And yes you are right that some thread can access the DB while being switched. However, this is solved by having creation_storage_mode_ in the accessor.  Then upon commit, I just check whether the storage mode is the same as during the start

src/storage/v2/storage.cpp
andrejtonev: You take the `main_lock_` on write/set but on on read/get.
This is why I was asking about the `Accessor`. There we take a shared lock on `main_lock_`.
We shouldn't use the `Storage::GetStorageMode()` directly since it is not protected. We should access it through the `Accessor` or add a shared_lock before. 
After our discussion on slack, I think we should add `GetStorageMode` and `GetIsolationLevel` to the `Accessor` and `DbAccessor` and replace all `interpreter_context->db->GetStorageMode()` calls with a call through an accessor.

src/query/interpreter.cpp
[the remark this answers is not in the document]
andrejtonev: Sorry for the badly worded comment :D 

The throw is fine. I just meant that we could return after the check.
The only thing we do after is 
`if (replication_role == ReplicationQuery::ReplicationRole::REPLICA) {...}`
witch will always be false if the first check passed, no need to do it.

src/query/interpreter.cpp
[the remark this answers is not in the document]
andrejtonev: But what happens to the storage if someone else is using it while you are switching it?
The `interpreter_context->db = std::make_unique<memgraph::storage::DiskStorage>(db_config);` will destroy the old storage. We have to make sure no one else can possibly be using it while this is happening. 

Nothing is stopping someone from accessing the original storage between lines:
`if (interpreter_context->interpreters->size() > 1) {`
and
`interpreter_context->db = std::make_unique<memgraph::storage::DiskStorage>(db_config);`.
I imagine that the DiskStorage creation can take some time, so it's not impossible that this happens.
If the only check is on commit, then someone could be using the destroyed storage for a long time.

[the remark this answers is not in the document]
as51340: we don't know what happens, probably it would but the pure point of switching to disk is that you should it alone, at the beginning. So yes there is probably a moment in which something would crash but I wouldn't say the problem is someone is using destroyed storage before commit. But again I agree this should be somehow refactored but this is not easy part. Maybe when we tackle the issue of switching from Lab

src/query/interpreter.cpp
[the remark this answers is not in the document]
as51340: no problemo, I added else branch

src/query/interpreter.cpp
andrejtonev: The `void Interpreter::Commit()` is called on COMMIT, right?
The `STORAGE MODE ON_DISK_TRANSACTIONAL;` will fail because we are in a transaction. 
So when is this check in Commit() true?

src/storage/v2/storage.cpp
as51340: why not taking shared_lock in storage class then, why through accessor? Maybe I am missing something but I find it stupid to duplicate all functionalites to accessor level. Especially when main_lock_ is part of the storage

as51340:
> This merge has dropped the performance on certain benchmarks by some 10% (check single_vertex_read, single_edge_read, etc.) Double check if this is due to a fundamental memgraph change or something to do with the benchmark itself.

If it is 10% that is more-less ok I think. Buda expected that due to virtual calls.

src/query/interpreter.cpp
as51340: which check?

src/query/interpreter.cpp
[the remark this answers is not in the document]
andrejtonev: Maybe something like this could work.
Every new connection adds an interpreter to the list. So this way they can't use it until we moved the storage.
**NOTE:** No idea if this code is actually okay.
```
interpreter_context->interpreters.WithLock([&](const auto &interpreters_) {
  if (interpreters_.size() > 1) {
    throw utils::BasicException(
        "You cannot switch from an in-memory storage mode to the on-disk storage mode when there are "
        "multiple sessions active. Close all other sessions and try again. As Memgraph Lab uses "
        "multiple sessions to run queries in parallel,  "
        "it is currently impossible to switch to the on-disk storage mode within Lab. "
        "Close it, connect to the instance with mgconsole "
        "and change the storage mode to on-disk from there. Then, you can reconnect with the Lab "
        "and continue to use the instance as usual.");
  }
  main_guard.unlock();
  interpreter_context->db = std::make_unique<memgraph::storage::DiskStorage>(interpreter_context->db->config_);
});
```

src/storage/v2/storage.cpp
andrejtonev: We could take the shared_lock, but that is also currently missing from the get functions. I think...
Also, from the comments on slack, I think we are trying to limit access to the database from the queries to the appropriate accessors and not directly.

src/query/interpreter.cpp
andrejtonev:
```
  if (creation_mode != storage::StorageMode::ON_DISK_TRANSACTIONAL &&
      current_storage_mode == storage::StorageMode::ON_DISK_TRANSACTIONAL) {
```
That check is done in the Commit() and should never be true, right?
Even if it could be true in the current implementation, I think we should make changes to the code so that it never is true.
We don't allow mode changes in a transaction and we do not allow it if other users are actively using the database.

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1703 @@
...
+  auto it = std::unique_ptr<rocksdb::Iterator>(
+      disk_transaction_->GetIterator(read_opts, disk_storage->kvstore_->vertex_chandle));
+  for (it->SeekToFirst(); it->Valid(); it->Next()) {
+    const auto &key = it->key();
+    if (Gid::FromUint(std::stoull(utils::ExtractGidFromKey(key.ToString()))) == gid) {
+      return LoadVertexToMainMemoryCache(key, it->value());
+    }
+  }
+  return std::nullopt;
+}
```
antoniofilipovic, line 774: I think in this function we check whether vertex exists in vertices_ first, then we check whethere there is vertex in index, then we get it from storage and load it to cache. But in loding to cache we again check if vertex exists in skip list and I think this repetition is costly.

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1703 @@
...
+/// NOTE: This will create Delta object which will cause deletion of old key entry on the disk
+std::optional<storage::VertexAccessor> DiskStorage::DiskAccessor::LoadVertexToMainMemoryCache(
+    const rocksdb::Slice &key, const rocksdb::Slice &value) {
+  auto main_storage_accessor = vertices_.access();
+
+  const std::string key_str = key.ToString();
+  storage::Gid gid = Gid::FromUint(std::stoull(utils::ExtractGidFromKey(key_str)));
+  if (VertexExistsInCache(main_storage_accessor, gid)) {
+    return std::nullopt;
+  }
```
antoniofilipovic, line 316: This function is called by FindVertex. FindVertex already does check whether it is contained in skip list.

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1703 @@
...
+                      index_delta);
+}
+
+std::optional<storage::VertexAccessor> DiskStorage::DiskAccessor::LoadVertexToLabelPropertyIndexCache(
+    const rocksdb::Slice &key, const rocksdb::Slice &value, Delta *index_delta,
+    utils::SkipList<storage::Vertex>::Accessor index_accessor) {
+  storage::Gid gid = Gid::FromUint(std::stoull(utils::ExtractGidFromLabelPropertyIndexStorage(key.ToString())));
+  if (VertexExistsInCache(index_accessor, gid)) {
+    return std::nullopt;
+  }
```
antoniofilipovic, line 343: I feel like this function shouldn't do this check. This check should be done before and then we just here load.

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1703 @@
...
+  const std::string value_str{value.ToString()};
+  const auto labels{utils::DeserializeLabelsFromLabelPropertyIndexStorage(value_str)};
+  return CreateVertex(index_accessor, gid, labels,
+                      utils::DeserializePropertiesFromLabelPropertyIndexStorage(value.ToString()), index_delta);
+}
+
+std::optional<EdgeAccessor> DiskStorage::DiskAccessor::DeserializeEdge(const rocksdb::Slice &key,
+                                                                       const rocksdb::Slice &value) {
+  const auto edge_parts = utils::Split(key.ToStringView(), "|");
+  const Gid edge_gid = Gid::FromUint(std::stoull(edge_parts[4]));
```
antoniofilipovic, line 354: Can this be part of some util function?

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1703 @@
...
+
+  auto edge_acc = edges_.access();
+  auto res = edge_acc.find(edge_gid);
+  if (res != edge_acc.end()) {
+    return std::nullopt;
+  }
```
antoniofilipovic, line 360: Can we do that check before somewhere, so this function only does deserialization of edge from rocksdb

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1703 @@
...
+
+  const auto [from_gid, to_gid] = std::invoke(
+      [](const auto &edge_parts) {
+        if (edge_parts[2] == "0") {  // out edge
```
antoniofilipovic, line 364: Can this index and zero be some kind of constants?

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1703 @@
...
+  auto *disk_storage = static_cast<DiskStorage *>(storage_);
+  rocksdb::ReadOptions ro;
+  std::string strTs = utils::StringTimestamp(transaction_.start_timestamp);
+  rocksdb::Slice ts(strTs);
+  ro.timestamp = &ts;
+  auto it =
+      std::unique_ptr<rocksdb::Iterator>(disk_transaction_->GetIterator(ro, disk_storage->kvstore_->vertex_chandle));
+  for (it->SeekToFirst(); it->Valid(); it->Next()) {
+    LoadVertexToMainMemoryCache(it->key(), it->value());
+  }
```
antoniofilipovic, line 394: Can we return this as iterable, and then load vertices to main memory cache as something iterates over vertices?

src/storage/v2/disk/storage.cpp
antoniofilipovic: I feel like this function should only do the loading, and check should be done before

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1703 @@
...
+  return VerticesIterable(AllVerticesIterable(vertices_.access(), &transaction_, view, &storage_->indices_,
+                                              &storage_->constraints_, storage_->config_.items));
+}
+
+VerticesIterable DiskStorage::DiskAccessor::Vertices(LabelId label, View view) {
+  index_storage_.emplace_back(std::make_unique<utils::SkipList<storage::Vertex>>());
+  auto &indexed_vertices = index_storage_.back();
+  index_deltas_storage_.emplace_back(std::list<Delta>());
```
antoniofilipovic, line 402: You can just pass empty()

antoniofilipovic: But I think we resolved that in other PR

src/storage/v2/disk/storage.cpp
antoniofilipovic: I know it is all intertwined but idea I think of this review is to separate it if possible

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1703 @@
...
+          CreateDeleteDeserializedIndexObjectDelta(&transaction_, index_deltas, std::nullopt),
+          indexed_vertices->access());
+    }
+  }
+
+  for (index_it->SeekToFirst(); index_it->Valid(); index_it->Next()) {
+    std::string key_str = index_it->key().ToString();
+    std::string it_value_str = index_it->value().ToString();
+    Gid curr_gid = Gid::FromUint(std::stoull(utils::ExtractGidFromLabelPropertyIndexStorage(key_str)));
+    // TODO: andi this will be optimized bla bla
```
antoniofilipovic, line 583: I think we shouldn't have this comment here :) Just TODO(andi) optimize

antoniofilipovic (commented):
Passed through some stuff in disk storage

PR #863, opened 2023-04-04
Improve libstdc++ dependency

### LINKED ISSUE #864: unable to install memgraphdb rpm on rhel 7
Hi,

When trying to install on a rhel 7 serveur, I have the following message :
#  sudo rpm -ivh memgraph-2.5.2_1-1.x86_64.rpm
error: Failed dependencies:
        libstdc >= 6 is needed by memgraph-2.5.2_1-1.x86_64
        libstdc++.so.6(GLIBCXX_3.4.29)(64bit) is needed by memgraph-2.5.2_1-1.x86_64

libstdc++ is installed on the server , see below:
rpm -qa|grep libstdc++
libstdc++-devel-4.8.5-44.el7.x86_64
libstdc++-4.8.5-44.el7.i686
libstdc++-devel-4.8.5-44.el7.i686
compat-libstdc++-33-3.2.3-72.el7.x86_64
compat-libstdc++-33-3.2.3-72.el7.i686
libstdc++-4.8.5-44.el7.x86_64

gitbuda:
Multiple issues:
* RPM install deps
  * http://ftp.rpm.org/max-rpm/s1-rpm-depend-auto-depend.html
  * https://stackoverflow.com/questions/14658034/how-do-you-make-it-so-that-cpack-doesnt-add-required-libraries-to-an-rpm
* Runtime libstdc++ deps
  * https://developers.redhat.com/articles/2023/04/03/leaner-libstdc-gcc-13 -> might eliminate the need for the `SharedLibraryHandle` and remote the `libstdc++` as a runtime dependency 🤔 👀
  * https://stackoverflow.com/questions/27385995/shipping-libstdc-so-6-with-application

Notes:
* Just setting `set(CPACK_RPM_PACKAGE_AUTOREQ " no")` -> doesn't work -> core dump on start -> `gdb` also can't load memgraph binary -> the crash was because of the persmissions
* There might be a problem with stripping symbols (should happen only in the release) -> debug binary installed from the package is smaller compared to the one which is built -> NOPE, permissions

closes #864


release/CMakeLists.txt
```diff
@@ -48,7 +48,7 @@ set(CPACK_DEBIAN_PACKAGE_DESCRIPTION "${CPACK_PACKAGE_DESCRIPTION_SUMMARY}
 
 # Add `openssl` package to dependencies list. Used to generate SSL certificates.
 # We also depend on `python3` because we embed it in Memgraph.
-set(CPACK_DEBIAN_PACKAGE_DEPENDS "openssl (>= 1.1.0), python3 (>= 3.5.0), libstdc++6")
+set(CPACK_DEBIAN_PACKAGE_DEPENDS "openssl (>= 1.1.0), python3 (>= 3.5.0)")
```
gitbuda, line 51: Add `logrotate` (audit logs) + `curl` (telemetry)

src/query/procedure/module.hpp
```diff
@@ -131,7 +131,12 @@ class ModuleRegistry final {
  private:
   class SharedLibraryHandle {
    public:
-    SharedLibraryHandle(const std::string &shared_library, int mode) : handle_{dlopen(shared_library.c_str(), mode)} {}
+    SharedLibraryHandle(const std::string &shared_library, int mode, const std::string &hint_message = "")
+        : handle_{dlopen(shared_library.c_str(), mode)} {
+      if (!handle_) {
+        spdlog::warn("Unable to load {}. {}", shared_library, hint_message);
```
gitbuda, line 137: Add more descriptive message, explaining problem with `cout` + non-string variable

release/CMakeLists.txt
andrejtonev: Seems like they already get added.

gitbuda: `CPACK` black magic, `curl` is added to the list, but `logrotate` is not for some reason...

antoniofilipovic (approved):
Looks good to me after the explanation on the call!

release/CMakeLists.txt
gitbuda: I'll skip adding new stuff for now, maybe during the next release (v2.10)

PR #871, opened 2023-04-11
Accessors-interface

Darych:
- [x] Build storage module
- [x] Fix tests in storage module
- [x] Build query engine
- [x] Think about InterpreterContext and SessionData. They hold raw pointers to Storage. DECISION: let's hold raw pointers because outside we created unique_ptr which will not be discarded till the end of process.
- [ ] Think about DbAccessor and storage::Accessor ownerships. In `interpreter.hpp` we have both accessors and use them, so we can't move ownership from one to another. Or we need to refactor the interpreter.
- [ ] Fix tests in query engine
- [ ] Think about Runtime Polymorphism

[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR


src/memgraph.cpp
```diff
@@ -907,10 +908,10 @@ int main(int argc, char **argv) {
     }
     db_config.durability.snapshot_interval = std::chrono::seconds(FLAGS_storage_snapshot_interval_sec);
   }
-  memgraph::storage::Storage db(db_config);
+  auto db = std::unique_ptr<memgraph::storage::Storage>(new memgraph::storage::InMemoryStorage(db_config));
```
gitbuda, line 911: https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique should be a better choice here 😄

gitbuda (commented):
* In a few places `new Something` should be replaced with `make_unique`
* Because the whole in-memory storage moved to a new file, we have to be careful when merging this. We should port all new features when merging.

gitbuda:
All here is moved under #943

PR #876, opened 2023-04-12
[E218 < T1250] Add HTTP Server [for Prometheus metrics]

Josipmrden:
The HTTP Server gives another endpoint in the Memgraph platform to query and inspect various checks inside the database (metrics, health checks, etc). In the scope of this Epic and PR, metrics will be exposed for the Prometheus integration, but anyone can add its own HTTP handler on another endpoint to query other states inside the database.

[master < Epic] PR
- [ ] Write E2E tests


src/communication/http/session.hpp
```diff
@@ -0,0 +1,193 @@
...
+ public:
+  template <typename... Args>
+  static std::shared_ptr<Session> Create(Args &&...args) {
+    return std::shared_ptr<Session>{new Session{std::forward<Args>(args)...}};
+  }
+
+  void Run() {
+    if (auto *ssl = std::get_if<SSLSocket>(&stream_); ssl != nullptr) {
+      try {
+        boost::beast::get_lowest_layer(*ssl).expires_after(std::chrono::seconds(30));
```
antoniofilipovic, line 59: Maybe change this 30 to some kind of constant

antoniofilipovic:
I have no other comments for this PR, all seems good

src/http_handlers/metrics.hpp
```diff
@@ -0,0 +1,127 @@
...
+      return send(bad_request("Illegal request-target"));
+    }
+
+    http::string_body::value_type body;
+
+    auto service_response = service_.GetMetrics().AsJson();
+    body.append(service_response.dump());
+
+    // Cache the size since we need it after the move
+    auto const size = body.size();
```
antoniofilipovic, line 112: `const auto size` ?

src/communication/http/session.hpp
```diff
@@ -0,0 +1,195 @@
...
+      });
+    };
+
+    // handle request
+    handler_.HandleRequest(std::move(req_), async_write);
+  }
+
+  void DoClose() {
+    std::visit(utils::Overloaded{[this](SSLSocket &stream) {
+                                   boost::beast::get_lowest_layer(stream).expires_after(std::chrono::seconds(30));
```
as51340, line 155: What does expire after 30secs?

as51340: Maybe also put as a constant.

src/http_handlers/metrics.hpp
```diff
@@ -0,0 +1,127 @@
...
+    return MetricsResponse{.vertex_count = info.vertex_count,
+                           .edge_count = info.edge_count,
+                           .average_degree = info.average_degree,
+                           .memory_usage = info.memory_usage,
+                           .disk_usage = info.disk_usage};
+  }
+
+ private:
+  const storage::Storage *db_;
+  query::InterpreterContext *interpreter_context_;
```
as51340, line 62: Where are those variables used? Also, even if you need both I think having interpreter is enough because you can fetch context from it.

antoniofilipovic: You get this from SessionData and I think this is different struct and that is how such struct is constructed

as51340: Sorry can you explain, not sure what you meant with such struct is constructred

Josipmrden: db_ is used, but the data is the same one that is passed to session. I don't use the interpreter context though and the interpreter, I could remove it.

src/communication/http/session.hpp
Josipmrden: it gives it 30 secs to handshake, expires_after sets the timeout for the next logical operation

as51340:
It seems to me that you included changes from some other PR. Can you remove them?

PR #877, opened 2023-04-13
[master < T620-FL] Fix max block size bug

antoniofilipovic:
[master < Task] PR

### DIFF
--- a/src/query/interpreter.hpp
+++ b/src/query/interpreter.hpp
@@ -51,7 +51,7 @@ extern const Event FailedQuery;
 namespace memgraph::query {
 
 inline constexpr size_t kExecutionMemoryBlockSize = 1UL * 1024UL * 1024UL;
-inline constexpr size_t kExecutionPoolMaxBlockSize = 32768UL;  // 2 ^ 15
+inline constexpr size_t kExecutionPoolMaxBlockSize = 2048UL;  // 2 ^ 11
 
 class AuthQueryHandler {
  public:
--- a/src/query/interpreter.cpp
+++ b/src/query/interpreter.cpp
@@ -2729,7 +2729,7 @@ Interpreter::PrepareResult Interpreter::Prepare(const std::string &query_string,
         // Using PoolResource without MonotonicMemoryResouce for LOAD CSV reduces memory usage.
         // QueryExecution MemoryResource is mostly used for allocations done on Frame and storing `row`s
         query_executions_[query_executions_.size() - 1] = std::make_unique<QueryExecution>(
-            utils::PoolResource(1, kExecutionPoolMaxBlockSize, utils::NewDeleteResource(), utils::NewDeleteResource()));
+            utils::PoolResource(8, kExecutionPoolMaxBlockSize, utils::NewDeleteResource(), utils::NewDeleteResource()));
         query_execution_ptr = &query_executions_.back();
       }
     }


Josipmrden (commented):
Can you maybe just explain what you did here, for reference, why these 2 numbers impact LOAD CSV issues. I can approve after.

antoniofilipovic:
> Can you maybe just explain what you did here, for reference, why these 2 numbers impact LOAD CSV issues. I can approve after.
Number 1 determined how many Blocks are in a Chunk. The pool is operating with Chunks and PoolResource with Pools.

Changing `blocks_per_chunk` from 1 to 8 as it is in the rest of the code serves as a decision for how many Blocks will be allocated in one Chunk. This enables you to allocate more memory at once for more memory objects of the same size, and each time a new object wants a new piece of memory, you don't need to ask (I presume kernel) to give you memory, but you already have it stored for such occasion. You only need to allocate if you run out of memory, and then you will not need to allocate for some time.

Benjo and I had a discussion about this, and I changed it in the first part of the code but forgot here. 

You can read more about the discussion on this [link](https://github.com/memgraph/memgraph/pull/825#discussion_r1151765452) if interested in my theory behind this. As Benjo said, we should run some measurements.

Josipmrden (approved):

antaljanosbenjamin:
Just a small clarification about this:
> you don't need to ask (I presume kernel) to give you memory

In released Memgraphs versions for a long time we are using jemalloc. That means memory allocation from Memgraph point of view not necessarily involve kernel calls. It is totally possible and expected (that is the reason we use jemalloc) that jemalloc has some allocated, but not used memory buffers to serve allocations from Memgraph.

You can kind of think about jemalloc as an extra layer of allocator, but it is globally used in Memgraph.

src/query/interpreter.hpp
```diff
@@ -51,7 +51,7 @@ extern const Event FailedQuery;
 namespace memgraph::query {
 
 inline constexpr size_t kExecutionMemoryBlockSize = 1UL * 1024UL * 1024UL;
-inline constexpr size_t kExecutionPoolMaxBlockSize = 32768UL;  // 2 ^ 15
+inline constexpr size_t kExecutionPoolMaxBlockSize = 2048UL;  // 2 ^ 11
```
antaljanosbenjamin, line 54: Could you please write down the reasons behind this change also?

antoniofilipovic: Yes. I made some comparisons, and there were no memory spikes when I reduced `max_block_size` to be smaller than one `row` that wants to allocate memory. To be specific, I tried with rows that had 100 properties with 10 chars for each property, which resulted in 100*10*~8 = 8000 bytes which needed to be allocated for each `row`, and there were no memory spikes when unpooled memory was used in such cases. 

And it seems to me that having a max of 32kB * 8 allocated memory when we have really big rows, would result in some kind of issues, so I reduced `max_block_size` as it seems unnecessary for the threshold to be so big when  `unpooled_memory` can be used without suffering memory issues.

antaljanosbenjamin (approved):

antoniofilipovic:
@vpavicic this bug was manifested in some LOAD CSV queries

RN: Fix performance degradation on some LOAD CSV queries due to inoptimal memory allocation

vpavicic:
@antoniofilipovic Memory allocation in LOAD CSV queries has been optimized to avoid performance degradation.

PR #882, opened 2023-04-16
[master] < MG-parallelize-index-creation

gvolfing:
[master < MG-parallelize-index-creation] PR


antaljanosbenjamin:
One thing that @gitbuda mentioned to me: maybe it make sense to introduce a flag that can disable parallel index creation in case there are some problems with it that we don't see yet.

antaljanosbenjamin (commented):
Please consider extracting the common logic from the exact creation: handling OOM exceptions, batch handling, etc. Maybe something similar to `RecoverOnMultipleThreads` would result in a cleaner approach.

antaljanosbenjamin (changes requested):

gitbuda:
@gvolfing I think this should be disabled by default (an opinion). Another thing, I think the `memgraph/config/flags.yml` should be updated because of the package config file 😄

src/storage/v2/indices.cpp
```diff
@@ -263,6 +265,95 @@ bool CurrentVersionHasLabelProperty(const Vertex &vertex, LabelId label, Propert
...
+    std::vector<std::jthread> threads;
+    threads.reserve(thread_count);
+
+    for (auto i{0U}; i < thread_count; ++i) {
+      threads.emplace_back(
+          [&skiplist_iter, &func, &index, &vertex_batches, &maybe_error, &batch_counter, &key, &vertices]() {
+            while (!maybe_error.Lock()->has_value()) {
+              const auto batch_index = batch_counter++;
+              if (batch_index >= vertex_batches.size()) {
+                return;
```
antoniofilipovic, line 332: Instead of creating more threads then are batches, you can set thread count to `min(thread_count, batch_size)`. This way you don't need to return here.

antoniofilipovic (commented):
Do we add tests here to check parallel index creation?

src/storage/v2/indices.cpp
gvolfing: That is a good point, however that `return` statement is there because of the concurrency of the threads. The batch_index is used to calculate which batch the given thread will work on next. If you have 10 worker threads and 15 batches overall, lets say the 10 threads will take the first 10 batches in some order. Individual threads will finish the processing of their assigned batches after some arbitrary time. We do not have a way to know if the there are still work left to do (batches left to process), so when a given thread asks for another batch to process(gets the value of the atomic counter), we have to make sure that the batch represented by the integer value is still actually a batch that is needed to be processed. If our thread wakes up and batch_index it gets is 15 in this specific case we know that this integer is not representing another batch, since `batch_index >= vertex_batches.size()`.

There is an argument to be made in cases where `thread_count  > batches.size()`, however since the parallel index creation is a settable flag, which will be false by default, I would say lets give that heuristic decision to the user. This should be only used in cases where we try to recover a huge set of data, where chances of `thread_count  > batches.size()` are slim.

Let me know if you disagree or if I misunderstood something.

antoniofilipovic: Oh I was wrong, you still need return here :smile:

antoniofilipovic: I was thinking something else.

On these two lines, it is possible that vertex_batches.size() > thread_count, so we reserve and create more threads than needed.

```
std::vector<std::jthread> threads;
threads.reserve(thread_count);
```
We need a return statement, but I was thinking that we can set `thread_count = min(thread_count, vertex_batches.size())`. What do you think about that?

gvolfing:
> Do we add tests here to check parallel index creation?
@antoniofilipovic 
As we discussed on the daily the comprehensive testing of index creation will be done in a subsequent PR.

src/storage/v2/indices.cpp
gvolfing: Yep, I will do that.

antaljanosbenjamin (approved):

antoniofilipovic:
@vpavicic and @gvolfing I think you have to check about release note here

vpavicic:
@antoniofilipovic & @gvolfing I have a release note in the docs PR, but the PR is still a draft, so is there more info coming? 
https://github.com/memgraph/docs/pull/833

Seems like the index page should be updated as well
https://memgraph.com/docs/memgraph/reference-guide/indexing

PR #886, opened 2023-04-18
[master < ] Change memory resource used in WITH clause

antoniofilipovic:
[master < Task] PR

- [ ] Tag someone from the docs team in the comments

This is draft PR and something that helps execution a lot.  WITH clause is accumulating at a certain point all results. Using PoolResource in such a case is not a good option, since there needs to be a lot of memory allocations. Allocating memory using MonotonicBufferResource and using the such resource in WITH clause could be a potential solution since this memory should live and is used during the whole query.


antoniofilipovic:
@gitbuda @Josipmrden I am considering closing this PR as we haven't had any more problems with the `WITH` clause after Benjo implemented [the ordering of chunks in PoolResource](https://github.com/memgraph/memgraph/pull/898). If we have, we can reopen this and implement the solution. What do you think?

PR #888, opened 2023-04-19
[master < MG] `AccumulateCursor` with `utils::pmr::deque`

antaljanosbenjamin:
[master < Task] PR

I checked it with a CSV from @antoniofilipovic using 200k rows. With `utils::pmr::vector` it took 27 seconds to import, with `utils::pmr::deque` it took only 22 seconds. I also checked the performance with a query where the copy should be cheap, thus the different behavior of `deque` might decrease the performance, but I didn't notice any measurable difference (neither good, nor bad). I used the following query: `UNWIND range(12000000, 20000000) AS id CREATE (n{prop:id}) WITH id MATCH (n) WHERE id(n) = id RETURN n;`.

### DIFF
--- a/.github/workflows/diff.yaml
+++ b/.github/workflows/diff.yaml
@@ -437,7 +437,7 @@ jobs:
       - name: Run mgbench
         run: |
           cd tests/mgbench
-          ./benchmark.py --num-workers-for-benchmark 12 --export-results benchmark_result.json pokec/medium/*/*
+          ./benchmark.py vendor-native --num-workers-for-benchmark 12 --export-results benchmark_result.json pokec/medium/*/*
 
       - name: Upload mgbench results
         run: |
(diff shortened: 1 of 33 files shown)


antaljanosbenjamin:
We tried to reproduce the results with @antoniofilipovic , but we couldn't managed. However the performance for a specific query that was provided by Antonio we could measure slightly better performance (~3-4%) and we didn't notice any performance degragdation on our benchmarks, so we decided to merge it. The new measurements:
With `utils::pmr::deque`:
Empty set (round trip in 19.886 sec)
Empty set (round trip in 25.869 sec)
Empty set (round trip in 20.282 sec)
Empty set (round trip in 21.615 sec)
Empty set (round trip in 23.612 sec)
Avarage is 22.253 sec

With `utils::pmr::vector`:
Empty set (round trip in 21.763 sec)
Empty set (round trip in 22.220 sec)
Empty set (round trip in 24.506 sec)
Empty set (round trip in 22.732 sec)
Empty set (round trip in 24.511 sec)
Empty set (round trip in 22.586 sec)
Avarage is 23.053 sec

Issue #896, opened 2023-04-24
[BUG] The SHOW CONFIG; query returns some wrong defaults


vpavicic:
**Memgraph version**
Memgraph 2.7.0

**Environment**
Some information about the environment you are using Memgraph on: operating
system, how do you connect, with or without docker, which driver etc.

**Describe the bug**
SHOW CONFIG; return wrong defaults on:

storage_properties_on_edges = false
storage_recover_on_setup = false
storage_snapshot_interval_sec = 0 
storage_snapshot_on_exit = false
storage_wal_enabled = false

**To Reproduce**
1. Run the following query 'SHOW CONFIG;` on a new clean instance

**Expected behavior**
All of these are true by default and the default snapshot interval is 300 secs

imilinovic:
Docker default values are different than build from source default values defined in gflags but the ones in gflags are shown which is expected

gitbuda:
Reopened because it is still an issue. One option is to completely remove the default values from the `SHOW CONFIG;` 🤔

PR #897, opened 2023-04-24
[master<] Fix hardcoded info about default config values 

vpavicic:
Some of the default values that would show up on SHOW CONFIG; were wrong so I fixed them. 

Closes https://github.com/memgraph/memgraph/issues/896


vpavicic:
This is not what I was trying to do :)

PR #900, opened 2023-04-24
[master < MG] Use correct memory resource

antaljanosbenjamin:
[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR

The `PoolResource`s in the `Interpreter` were using wrong upstream memory resouorces, this PR fixes that.


antaljanosbenjamin:
If you check out `ResourceWithOutOfMemoryException`, you can see it is super simple:
```cpp
// Allocate memory with the OutOfMemoryException enabled if the requested size
// puts total allocated amount over the limit.
class ResourceWithOutOfMemoryException : public MemoryResource {
 public:
  explicit ResourceWithOutOfMemoryException(utils::MemoryResource *upstream = utils::NewDeleteResource())
      : upstream_{upstream} {}

  utils::MemoryResource *GetUpstream() noexcept { return upstream_; }

 private:
  void *DoAllocate(size_t bytes, size_t alignment) override {
    utils::MemoryTracker::OutOfMemoryExceptionEnabler exception_enabler;
    return upstream_->Allocate(bytes, alignment);
  }

  void DoDeallocate(void *p, size_t bytes, size_t alignment) override { upstream_->Deallocate(p, bytes, alignment); }

  bool DoIsEqual(const utils::MemoryResource &other) const noexcept override { return upstream_->IsEqual(other); }

  MemoryResource *upstream_{utils::NewDeleteResource()};
};
```

Basically what it does is enabling `OutOfMemoryException` to be thrown if the allocation would go over the memory limit. Apart from that it uses the upstream for everything. So that means it doesn't cause any issues as long as the upstream memory doesn't cause any issues. By default (thus in the modified part of the code) the upstream resource is the `NewDeleteResource`, which doesn't have any issues even on multi threaded usage, so I think we are safe here.

as51340 (approved):

antoniofilipovic (approved):
All looks good

antoniofilipovic:
> r everything. So that means it doesn't cause any iss

My concern was whether we should use the same resource, not for overwriting but for bugs we had before with using MonotonicBufferResource in LOAD CSV for both pooled and unpooled. The thing is that such a bug should not happen as with NewDeleteResource, memory should be reused and actually, NewDeleteResource is also just wrapper, so all good on that side.

PR #908, opened 2023-05-02
[T1250 < T1251] Expose metrics data in critical parts of the system

Josipmrden:
The goal of this PR is to add counters, gauges, and histograms, to all the necessary parts of the system.
That includes:
- queries
- sessions
- streams
- transactions
- triggers
- snapshots
- operators
- general metrics

All the metrics could be exposed with the http server addition in the branch that this code will be merged to.

### DIFF
--- a/src/communication/v2/listener.hpp
+++ b/src/communication/v2/listener.hpp
@@ -35,7 +35,7 @@
 #include "utils/synchronized.hpp"
 
 namespace Statistics {
-extern const Event ActiveConnections;
+extern const Event ActiveSessions;
 }  // namespace Statistics
 
 namespace memgraph::communication::v2 {
@@ -115,10 +115,9 @@ class Listener final : public std::enable_shared_from_this<Listener<TSession, TS
       return OnError(ec, "accept");
     }
 
-    Statistics::IncrementCounter(Statistics::ActiveConnections);
-
     auto session = SessionHandler::Create(std::move(socket), data_, *server_context_, endpoint_, inactivity_timeout_,
                                           service_name_);
+
     session->Start();
     DoAccept();
   }
--- a/src/communication/v2/listener.hpp
+++ b/src/communication/v2/listener.hpp
@@ -1,4 +1,4 @@
-// Copyright 2022 Memgraph Ltd.
+// Copyright 2023 Memgraph Ltd.
 //
 // Use of this software is governed by the Business Source License
 // included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
@@ -30,8 +30,13 @@
 #include "communication/context.hpp"
 #include "communication/v2/pool.hpp"
 #include "communication/v2/session.hpp"
+#include "utils/event_counter.hpp"
 #include "utils/spin_lock.hpp"
 #include "utils/synchronized.hpp"
+
+namespace Statistics {
+extern const Event ActiveConnections;
+}  // namespace Statistics
 
 namespace memgraph::communication::v2 {
 
@@ -109,6 +114,8 @@ class Listener final : public std::enable_shared_from_this<Listener<TSession, TS
     if (ec) {
       return OnError(ec, "accept");
     }
+
+    Statistics::IncrementCounter(Statistics::ActiveConnections);
 
     auto session = SessionHandler::Create(std::move(socket), data_, *server_context_, endpoint_, inactivity_timeout_,
                                           service_name_);
--- a/src/communication/v2/session.hpp
+++ b/src/communication/v2/session.hpp
@@ -1,4 +1,4 @@
-// Copyright 2022 Memgraph Ltd.
+// Copyright 2023 Memgraph Ltd.
 //
 // Use of this software is governed by the Business Source License
 // included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
@@ -43,8 +43,13 @@
 
 #include "communication/context.hpp"
 #include "communication/exceptions.hpp"
+#include "utils/event_counter.hpp"
 #include "utils/logging.hpp"
 #include "utils/variant_helpers.hpp"
+
+namespace Statistics {
+extern const Event ActiveConnections;
+}  // namespace Statistics
 
 namespace memgraph::communication::v2 {
 
@@ -429,6 +434,8 @@ class Session final : public std::enable_shared_from_this<Session<TSession, TSes
       }
       lowest_layer.close();
     });
+
+    Statistics::DecrementCounter(Statistics::ActiveConnections);
   }
 
   void DoHandshake() {
--- a/src/communication/v2/session.hpp
+++ b/src/communication/v2/session.hpp
@@ -48,7 +48,10 @@
 #include "utils/variant_helpers.hpp"
 
 namespace Statistics {
-extern const Event ActiveConnections;
+extern const Event ActiveSessions;
+extern const Event ActiveTCPSessions;
+extern const Event ActiveSSLSessions;
+extern const Event ActiveWebSocketSessions;
 }  // namespace Statistics
 
 namespace memgraph::communication::v2 {
@@ -104,6 +107,8 @@ class WebsocketSession : public std::enable_shared_from_this<WebsocketSession<TS
   // Start the asynchronous accept operation
   template <class Body, class Allocator>
   void DoAccept(boost::beast::http::request<Body, boost::beast::http::basic_fields<Allocator>> req) {
+    Statistics::IncrementCounter(Statistics::ActiveWebSocketSessions);
+
     execution_active_ = true;
     // Set suggested timeout settings for the websocket
     ws_.set_option(boost::beast::websocket::stream_base::timeout::suggested(boost::beast::role_type::server));
@@ -218,6 +223,10 @@ class WebsocketSession : public std::enable_shared_from_this<WebsocketSession<TS
     if (!IsConnected()) {
       return;
     }
+
+    Statistics::DecrementCounter(Statistics::ActiveSessions);
+    Statistics::DecrementCounter(Statistics::ActiveWebSocketSessions);
+
     if (ec) {
       return OnError(ec, "close");
     }
@@ -264,12 +273,17 @@ class Session final : public std::enable_shared_from_this<Session<TSession, TSes
     if (execution_active_) {
       return false;
     }
+
+    Statistics::IncrementCounter(Statistics::ActiveSessions);
+
     execution_active_ = true;
     timeout_timer_.async_wait(boost::asio::bind_executor(strand_, std::bind(&Session::OnTimeout, shared_from_this())));
 
     if (std::holds_alternative<SSLSocket>(socket_)) {
+      Statistics::IncrementCounter(Statistics::ActiveSSLSessions);
       boost::asio::dispatch(strand_, [shared_this = shared_from_this()] { shared_this->DoHandshake(); });
     } else {
+      Statistics::IncrementCounter(Statistics::ActiveTCPSessions);
       boost::asio::dispatch(strand_, [shared_this = shared_from_this()] { shared_this->DoRead(); });
     }
     return true;
@@ -435,7 +449,7 @@ class Session final : public std::enable_shared_from_this<Session<TSession, TSes
       lowest_layer.close();
     });
 
-    Statistics::DecrementCounter(Statistics::ActiveConnections);
+    Statistics::DecrementCounter(Statistics::ActiveSessions);
   }
 
   void DoHandshake() {
@@ -457,6 +471,12 @@ class Session final : public std::enable_shared_from_this<Session<TSession, TSes
   }
 
   void OnClose(const boost::system::error_code &ec) {
+    if (ssl_context_.has_value()) {
+      Statistics::DecrementCounter(Statistics::ActiveSSLSessions);
+    } else {
+      Statistics::DecrementCounter(Statistics::ActiveTCPSessions);
+    }
+
     if (ec) {
       return OnError(ec);
     }
--- a/src/http_handlers/metrics.hpp
+++ b/src/http_handlers/metrics.hpp
@@ -19,8 +19,10 @@
 #include <json/json.hpp>
 
 #include <utils/event_counter.hpp>
+#include <utils/event_gauge.hpp>
 #include "query/interpreter.hpp"
 #include "storage/v2/storage.hpp"
+#include "utils/event_gauge.hpp"
 
 namespace memgraph::http {
 
@@ -41,6 +43,10 @@ struct MetricsResponse {
       metrics_response[event_counter.first] = event_counter.second;
     }
 
+    for (const auto &event_gauge : event_gauges) {
+      metrics_response[event_gauge.first] = event_gauge.second;
+    }
+
     return metrics_response;
   }
 
@@ -50,6 +56,7 @@ struct MetricsResponse {
   uint64_t memory_usage;
   uint64_t disk_usage;
   std::vector<std::pair<std::string, uint64_t>> event_counters;
+  std::vector<std::pair<std::string, uint64_t>> event_gauges;
 };
 
 template <typename TSessionData>
@@ -64,7 +71,8 @@ class MetricsService {
                            .average_degree = info.average_degree,
                            .memory_usage = info.memory_usage,
                            .disk_usage = info.disk_usage,
-                           .event_counters = GetEventCounters()};
+                           .event_counters = GetEventCounters(),
+                           .event_gauges = GetEventGauges()};
   }
 
  private:
@@ -74,14 +82,26 @@ class MetricsService {
 
   std::vector<std::pair<std::string, uint64_t>> GetEventCounters() {
     std::vector<std::pair<std::string, uint64_t>> event_counters;
-    event_counters.reserve(Statistics::End());
+    event_counters.reserve(Statistics::CounterEnd());
 
-    for (auto i = 0; i < Statistics::End(); i++) {
-      event_counters.emplace_back(Statistics::GetName(i),
+    for (auto i = 0; i < Statistics::CounterEnd(); i++) {
+      event_counters.emplace_back(Statistics::GetCounterName(i),
                                   Statistics::global_counters[i].load(std::memory_order_relaxed));
     }
 
     return event_counters;
+  }
+
+  std::vector<std::pair<std::string, uint64_t>> GetEventGauges() {
+    std::vector<std::pair<std::string, uint64_t>> event_gauges;
+    event_gauges.reserve(Statistics::GaugeEnd());
+
+    for (auto i = 0; i < Statistics::GaugeEnd(); i++) {
+      event_gauges.emplace_back(Statistics::GetGaugeName(i),
+                                Statistics::global_gauges[i].load(std::memory_order_seq_cst));
+    }
+
+    return event_gauges;
   }
 };
 
--- a/src/http_handlers/metrics.hpp
+++ b/src/http_handlers/metrics.hpp
@@ -35,11 +35,13 @@ struct MetricsResponse {
  public:
   nlohmann::json AsJson() {
     auto metrics_response = nlohmann::json();
-    metrics_response["vertex_count"] = vertex_count;
-    metrics_response["edge_count"] = edge_count;
-    metrics_response["average_degree"] = average_degree;
-    metrics_response["memory_usage"] = memory_usage;
-    metrics_response["disk_usage"] = disk_usage;
+    const auto *general_type = "General";
+
+    metrics_response[general_type]["vertex_count"] = vertex_count;
+    metrics_response[general_type]["edge_count"] = edge_count;
+    metrics_response[general_type]["average_degree"] = average_degree;
+    metrics_response[general_type]["memory_usage"] = memory_usage;
+    metrics_response[general_type]["disk_usage"] = disk_usage;
 
     for (const auto &event_counter : event_counters) {
       auto type = std::get<1>(event_counter);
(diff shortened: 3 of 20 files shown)


src/query/interpreter.cpp
```diff
@@ -2605,6 +2616,10 @@ Interpreter::PrepareResult Interpreter::Prepare(const std::string &query_string,
   const auto trimmed_query = utils::Trim(upper_case_query);
 
   if (trimmed_query == "BEGIN" || trimmed_query == "COMMIT" || trimmed_query == "ROLLBACK") {
+    if (trimmed_query == "BEGIN") {
+      Statistics::IncrementCounter(Statistics::ActiveTransactions);
+    }
```
as51340, line 2621: PrepareTransactionQuery method already does this check whether the trimmed_query is BEGIN

src/query/interpreter.cpp
```diff
@@ -2616,11 +2631,13 @@ Interpreter::PrepareResult Interpreter::Prepare(const std::string &query_string,
   // an explicit transaction block.
   if (in_explicit_transaction_) {
     AdvanceCommand();
-  }
-  // If we're not in an explicit transaction block and we have an open
-  // transaction, abort it since we're about to prepare a new query.
-  else if (db_accessor_) {
-    AbortCommand(&query_execution);
+  } else {
+    Statistics::IncrementCounter(Statistics::ActiveTransactions);
```
as51340, line 2635: Do we want to count open transaction if we are not in explicit transaction?

src/utils/event_gauge.hpp
```diff
@@ -0,0 +1,48 @@
...
+using Value = uint64_t;
+using Gauge = std::atomic<Value>;
+
+class EventGauges {
+ public:
+  explicit EventGauges(Gauge *allocated_gauges) noexcept : gauges_(allocated_gauges) {}
+
+  auto &operator[](const Event event) { return gauges_[event]; }
+
+  const auto &operator[](const Event event) const { return gauges_[event]; }
```
as51340, line 29: Is it necessary to have both with and without const? I am really not sure.

tests/unit/utils_histogram.cpp
```diff
@@ -0,0 +1,46 @@
...
+  ASSERT_EQ(histo.Percentile(99.0), 25);
+  ASSERT_EQ(histo.Percentile(99.89), 32);
+  ASSERT_EQ(histo.Percentile(99.99), 46);
+  ASSERT_EQ(histo.Percentile(100.0), 500);
+
+  uint64_t max = std::numeric_limits<uint64_t>::max();
+  histo.Measure(max);
+  auto observed_max = static_cast<double>(histo.Percentile(100.0));
+  auto diff = (max - observed_max) / max;
+  ASSERT_THAT(diff, testing::Lt(0.01));
```
as51340, line 45: I think there is compare_double function

src/communication/v2/listener.hpp
```diff
@@ -30,9 +30,14 @@
...
 #include "communication/v2/pool.hpp"
 #include "communication/v2/session.hpp"
+#include "utils/event_counter.hpp"
 #include "utils/spin_lock.hpp"
 #include "utils/synchronized.hpp"
 
+namespace Statistics {
+extern const Event ActiveSessions;
+}  // namespace Statistics
+
```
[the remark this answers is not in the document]
antoniofilipovic: Agree, probably in `DoAccept` as it is then callback to `OnAccept`

src/http_handlers/metrics.hpp
```diff
@@ -41,6 +66,9 @@ struct MetricsResponse {
   double average_degree;
   uint64_t memory_usage;
   uint64_t disk_usage;
+  std::vector<std::tuple<std::string, std::string, uint64_t>> event_counters;
+  std::vector<std::tuple<std::string, std::string, uint64_t>> event_gauges;
+  std::vector<std::tuple<std::string, std::string, uint64_t>> event_histograms;
```
antoniofilipovic, line 71: Does it make sense that these are `utils::pmr::std::vector` as we should measure memory allocated for these counters, gauges and histograms

src/query/interpreter.cpp
```diff
@@ -1595,7 +1606,7 @@ PreparedQuery PrepareIndexQuery(ParsedQuery parsed_query, bool in_explicit_trans
               [&index_notification, &label_name, &properties_stringified]<typename T>(T &&) {
                 using ErrorType = std::remove_cvref_t<T>;
                 if constexpr (std::is_same_v<ErrorType, storage::ReplicationError>) {
-                  EventCounter::IncrementCounter(EventCounter::LabelIndexCreated);
+                  Statistics::IncrementCounter(Statistics::LabelIndexCreated);
```
antoniofilipovic, line 1609: Why do we increment counter on error?

src/query/interpreter.cpp
```diff
@@ -1897,7 +1908,7 @@ PreparedQuery PrepareTriggerQuery(ParsedQuery parsed_query, bool in_explicit_tra
       case TriggerQuery::Action::CREATE_TRIGGER:
         trigger_notification.emplace(SeverityLevel::INFO, NotificationCode::CREATE_TRIGGER,
                                      fmt::format("Created trigger {}.", trigger_query->trigger_name_));
-        EventCounter::IncrementCounter(EventCounter::TriggersCreated);
+        Statistics::IncrementCounter(Statistics::TriggersCreated);
```
antoniofilipovic, line 1911: I think this is not best place as trigger can exist and this is then incremented although trigger was not created

src/utils/event_gauge.hpp
antoniofilipovic: I think it is good as then when you have const variable, you can't call function to get counter

src/memgraph.cpp
```diff
@@ -160,7 +160,7 @@ DEFINE_VALIDATED_int32(monitoring_port, 7444,
                        "Port on which the websocket server for Memgraph monitoring should listen.",
                        FLAG_IN_RANGE(0, std::numeric_limits<uint16_t>::max()));
 // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
-DEFINE_VALIDATED_int32(metrics_port, 9092, "Port on which the Memgraph server for exposing metrics should listen.",
+DEFINE_VALIDATED_int32(metrics_port, 9091, "Port on which the Memgraph server for exposing metrics should listen.",
```
antoniofilipovic, line 163: What was reason for changing this port?

src/communication/v2/listener.hpp
```diff
@@ -30,6 +30,7 @@
 #include "communication/context.hpp"
 #include "communication/v2/pool.hpp"
 #include "communication/v2/session.hpp"
+#include "utils/event_counter.hpp"
```
antoniofilipovic, line 33: Do you need this include now here?

src/communication/v2/listener.hpp
```diff
@@ -112,6 +114,7 @@ class Listener final : public std::enable_shared_from_this<Listener<TSession, TS
 
     auto session = SessionHandler::Create(std::move(socket), data_, *server_context_, endpoint_, inactivity_timeout_,
                                           service_name_);
```
antoniofilipovic, line 116: empty line

src/communication/v2/session.hpp
```diff
@@ -213,6 +225,10 @@ class WebsocketSession : public std::enable_shared_from_this<WebsocketSession<TS
     if (!IsConnected()) {
       return;
     }
+
```
antoniofilipovic, line 228: I am not sure why do you `DecrementCounter` for `ActiveSessions` when you don't `IncrementCounter` for `ActiveSessions` in `DoAccept`

src/http_handlers/metrics.hpp
```diff
@@ -50,17 +54,85 @@ class MetricsService {
...
+    std::vector<std::tuple<std::string, std::string, uint64_t>> event_gauges(memgraph::metrics::GaugeEnd());
+
+    for (auto i = 0; i < memgraph::metrics::GaugeEnd(); i++) {
+      event_gauges.emplace_back(memgraph::metrics::GetGaugeName(i), memgraph::metrics::GetGaugeType(i),
+                                memgraph::metrics::global_gauges[i].load(std::memory_order_seq_cst));
+    }
+
+    return event_gauges;
+  }
+
```
antoniofilipovic, line 119: Since you are doing load here, I think you can use `memory_order_acquire` as we are not reading anyting and `memory_order_seq_cst` includes reading too. Also, in function GetEventCounters I would also use `memory_order_acquire` instead of `memory_order_relaxed`

src/http_handlers/metrics.hpp
```diff
@@ -77,45 +149,47 @@ class MetricsRequestHandler final {
...
     }
 
     // Request path must be absolute and not contain "..".
-    if (req.target().empty() || req.target()[0] != '/' || req.target().find("..") != beast::string_view::npos) {
+    if (req.target().empty() || req.target()[0] != '/' || req.target().find("..") != boost::beast::string_view::npos) {
       return send(bad_request("Illegal request-target"));
     }
 
-    http::string_body::value_type body;
+    boost::beast::http::string_body::value_type body{};
```
antoniofilipovic, line 180: Can you use here https://en.cppreference.com/w/cpp/filesystem/path/is_absrel

src/query/interpreter.cpp
```diff
@@ -721,7 +725,7 @@ Callback HandleStreamQuery(StreamQuery *stream_query, const Parameters &paramete
   Callback callback;
   switch (stream_query->action_) {
     case StreamQuery::Action::CREATE_STREAM: {
-      EventCounter::IncrementCounter(EventCounter::StreamsCreated);
+      memgraph::metrics::IncrementCounter(memgraph::metrics::StreamsCreated);
```
antoniofilipovic, line 728: I think this method GetKafkaCreateCallback can throw error. Since this is callback, maybe it makes sense to move this inside functions?

src/query/interpreter.cpp
```diff
@@ -2616,11 +2623,13 @@ Interpreter::PrepareResult Interpreter::Prepare(const std::string &query_string,
...
-  }
-  // If we're not in an explicit transaction block and we have an open
-  // transaction, abort it since we're about to prepare a new query.
-  else if (db_accessor_) {
-    AbortCommand(&query_execution);
+  } else {
+    memgraph::metrics::IncrementCounter(memgraph::metrics::ActiveTransactions);
+    // If we're not in an explicit transaction block and we have an open
+    // transaction, abort it since we're about to prepare a new query.
+    if (db_accessor_) {
```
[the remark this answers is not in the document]
antoniofilipovic: Ah, you covered that in AbortCommand.

src/memgraph.cpp
Josipmrden: This is the default port for Kafka usually. Prometheus is 9090, so I decided going with 9091

src/communication/v2/listener.hpp
Josipmrden: I will ommit it

src/http_handlers/metrics.hpp
```diff
@@ -77,45 +152,51 @@ class MetricsRequestHandler final {
...
   template <class Body, class Allocator>
-  void HandleRequest(http::request<Body, http::basic_fields<Allocator>> &&req,
-                     std::function<void(http::response<http::string_body>)> &&send) {
+  // NOLINTNEXTLINE(misc-unused-parameters)
+  void HandleRequest(boost::beast::http::request<Body, boost::beast::http::basic_fields<Allocator>> &&req,
+                     std::function<void(boost::beast::http::response<boost::beast::http::string_body>)> &&send) {
     auto response_json = nlohmann::json();
     // Returns a bad request response
-    auto const bad_request = [&req, &response_json](beast::string_view why) {
+    auto const bad_request = [&req, &response_json](boost::beast::string_view why) {
```
as51340, line 160: const auto?

src/http_handlers/metrics.hpp
```diff
@@ -77,45 +152,51 @@ class MetricsRequestHandler final {
...
+    if (req.target().empty() || req.target()[0] != '/' || req.target().find("..") != boost::beast::string_view::npos) {
       return send(bad_request("Illegal request-target"));
     }
 
-    http::string_body::value_type body;
+    // NOLINTNEXTLINE(cppcoreguidelines-init-variables)
+    boost::beast::http::string_body::value_type body{};
 
-    auto service_response = service_.GetMetrics().AsJson();
+    auto service_response = service_.AsJson(service_.GetMetrics());
```
as51340, line 187: If we don't have any other format than JSON, wouldn't it be better to have only one method GetMetricsJSON? It is a bit weird to call two methods from same object. If we will need to support other formats, then it makes sense to separate functionalities but not atm

as51340: Or at least AsJson method should be a static or some utility method since it doesn't relate in any way with object

src/query/interpreter.cpp
```diff
@@ -2734,14 +2742,17 @@ Interpreter::PrepareResult Interpreter::Prepare(const std::string &query_string,
   // an explicit transaction block.
   if (in_explicit_transaction_) {
     AdvanceCommand();
+  } else {
+    memgraph::metrics::IncrementCounter(memgraph::metrics::ActiveTransactions);
```
as51340, line 2746: How about counting active transactions as those that are explicit or those that require creating db_accessor? Then no need to modify this else if. You can put this increment in the if condition below line 2798

as51340: Because here can actually happen that we count active transaction but don't create one

src/query/interpreter.cpp
```diff
@@ -2923,7 +2934,12 @@ void Interpreter::Abort() {
 
   expect_rollback_ = false;
   in_explicit_transaction_ = false;
+
+  memgraph::metrics::IncrementCounter(memgraph::metrics::RollbackedTransactions);
```
as51340, line 2938: I think Abort() function is called from multiple places and doesn't necessarily mean rollback. How about moving this DecrementCounter to the Rollback function?

src/query/interpreter.cpp
```diff
@@ -3108,6 +3124,9 @@ void Interpreter::Commit() {
   if (!commit_confirmed_by_all_sync_repplicas) {
     throw ReplicationException("At least one SYNC replica has not confirmed committing last transaction.");
   }
+
+  memgraph::metrics::IncrementCounter(memgraph::metrics::CommitedTransactions);
```
as51340, line 3128: Commit function is not noexcept so I would definitely put those two under utils::OnScopeExit

src/query/interpreter.cpp
```diff
@@ -2892,7 +2903,7 @@ Interpreter::PrepareResult Interpreter::Prepare(const std::string &query_string,
 
     return {query_execution->prepared_query->header, query_execution->prepared_query->privileges, qid};
   } catch (const utils::BasicException &) {
-    EventCounter::IncrementCounter(EventCounter::FailedQuery);
+    memgraph::metrics::IncrementCounter(memgraph::metrics::FailedQuery);
```
as51340, line 2906: What is exactly semantic of failed query? Wrong query typed or

src/query/interpreter.hpp
```diff
@@ -515,7 +515,7 @@ std::map<std::string, TypedValue> Interpreter::Pull(TStream *result_stream, std:
     query_execution.reset(nullptr);
     throw;
   } catch (const utils::BasicException &) {
-    EventCounter::IncrementCounter(EventCounter::FailedQuery);
+    memgraph::metrics::IncrementCounter(memgraph::metrics::FailedQuery);
```
as51340, line 518: If the semantic of failed query is wrongly typed, then it seems to me this is wrong but please explain it to me :)

src/storage/v2/storage.cpp
```diff
@@ -1251,6 +1267,9 @@ utils::BasicResult<StorageIndexDefinitionError, void> Storage::DropIndex(
   commit_log_->MarkFinished(commit_timestamp);
   last_commit_timestamp_ = commit_timestamp;
 
+  // We don't care if there is a replication error because on main node the change will go through
+  memgraph::metrics::DecrementCounter(memgraph::metrics::LabelIndexCreated);
```
as51340, line 1271: If we increment and decrement counters, shouldn't the name be: ActiveLabelIndices and ActiveLabelPropetyIndices, not Created?

as51340: Same for other metrics, not sure if it occurs at other places.

src/storage/v2/storage.cpp
```diff
@@ -1943,13 +1965,18 @@ utils::BasicResult<Storage::CreateSnapshotError> Storage::CreateSnapshot(std::op
...
+
     auto transaction = CreateTransaction(IsolationLevel::SNAPSHOT_ISOLATION, storage_mode_);
     // Create snapshot.
     durability::CreateSnapshot(&transaction, snapshot_directory_, wal_directory_,
                                config_.durability.snapshot_retention_count, &vertices_, &edges_, &name_id_mapper_,
                                &indices_, &constraints_, config_, uuid_, epoch_id_, epoch_history_, &file_retainer_);
     // Finalize snapshot transaction.
     commit_log_->MarkFinished(transaction.start_timestamp);
+
+    memgraph::metrics::Measure(memgraph::metrics::SnapshotCreationLatency_us,
```
as51340, line 1978: Beautiful mrma:)

src/communication/v2/session.hpp
Josipmrden: This is a `WebsocketSession`. I increment the counter in `Session::Start() (line 279)`, and there we switch to `WebsocketSession` in `Session::OnRead`

src/utils/event_gauge.hpp
as51340: But you can const_cast it or something?

src/http_handlers/metrics.hpp
Josipmrden: Sorry what?

src/query/interpreter.cpp
Josipmrden: I think yes, the query has not started executing but while parsing or semantics. E.g. syntaxexception or semanticexception.

src/query/interpreter.hpp
Josipmrden: I'm not entirely sure at the moment as I just adopted this as part of the counters that were already present in the system, but let me check quickly.

src/storage/v2/storage.cpp
Josipmrden: Yes, I don't have actually counters for active indexes, just created, like the streams. But I can quickly create it as this were the legacy counters

Josipmrden: oh i actually have them, but the name is misleading

src/query/interpreter.cpp
Josipmrden: I can move it yes. Makes sense

src/query/interpreter.cpp
```diff
@@ -2799,6 +2816,9 @@ Interpreter::PrepareResult Interpreter::Prepare(const std::string &query_string,
       }
     }
 
+    utils::OnScopeExit active_transaction_update_metric(
+        []() { memgraph::metrics::IncrementCounter(memgraph::metrics::ActiveTransactions); });
```
as51340, line 2820: Sorry I probably expressed myself wrongly but I thought under the condition. And you don't need here OnScopeExit, just  put it before dbaccessor is created

src/query/interpreter.cpp
```diff
@@ -1212,6 +1226,10 @@ PreparedQuery Interpreter::PrepareTransactionQuery(std::string_view query_upper)
       if (!in_explicit_transaction_) {
         throw ExplicitTransactionUsageException("No current transaction to rollback.");
       }
+
+      utils::OnScopeExit rollback_metric_update(
+          []() { memgraph::metrics::IncrementCounter(memgraph::metrics::RollbackedTransactions); });
+
```
as51340, line 1232: Abort isn't noexcept but I think but if you put it before Abort() call that's also fine.

src/query/interpreter.cpp
as51340: That's fine but why is then the check in interpreter.hpp?

src/query/interpreter.cpp
```diff
@@ -2734,14 +2749,16 @@ Interpreter::PrepareResult Interpreter::Prepare(const std::string &query_string,
...
   if (in_explicit_transaction_) {
     AdvanceCommand();
+  } else {
+    // If we're not in an explicit transaction block and we have an open
+    // transaction, abort it since we're about to prepare a new query.
+    if (db_accessor_) {
+      query_executions_.emplace_back(
+          std::make_unique<QueryExecution>(utils::MonotonicBufferResource(kExecutionMemoryBlockSize)));
+      AbortCommand(&query_executions_.back());
+    }
```
as51340, line 2759: You should be able to remove changes so that you have again else if as before

src/query/interpreter.hpp
as51340: Did you check maybe?

as51340 (changes requested):

src/query/interpreter.hpp
Josipmrden: Yes I see, I was wrong about the semantics. In the EventCounter it says: "Number of times executing a query failed"

That's why it is in: `interpreter.cpp::Prepare` (because it can be a semantic or a syntax exception or any exception while preparing the query), and in `interpreter.hpp::Pull` for query runtime exceptions

src/query/interpreter.cpp
gitbuda: @as51340, what do you mean here by semantic? `FailedQuery` is used for both parsing and execution error

as51340: Yes then it makes sense, I just wanted to check which errors are included

src/query/interpreter.hpp
as51340: :+1:

as51340 (approved):

PR #914, opened 2023-05-03
[master < T914] Add command NULLIF for identifying nulls in LOAD CSV

Josipmrden:
The `NULLIF` command used in `LOAD CSV` enables for users to parse the `LOAD CSV` in a way which treats characters which have the identical character sequence as the `NULLIF` character as nulls.

**Usage:**
`LOAD CSV FROM './file.csv' WITH HEADER NULLIF 'N/A' AS row
`...

[master < Task] PR

- [x] Write a release note here
`Added NULLIF command in the LOAD CSV clause to specify a character sequence which will treat identical values in the csv file as nulls`


antoniofilipovic:
I think you need to add the keyword `nullif` on s`rc/query/frontend/stripped_lexer_constants.hpp`  as mentioned on this [Notion](https://www.notion.so/memgraph/Adding-a-new-Memgraph-specific-query-e588285275624b03ac941b778676fa40) page.

src/query/plan/operator.cpp
```diff
@@ -4674,22 +4675,31 @@ auto ToOptionalString(ExpressionEvaluator *evaluator, Expression *expression) ->
...
+TypedValue CsvRowToTypedList(csv::Reader::Row &row, std::optional<utils::pmr::string> &nullif) {
   auto *mem = row.get_allocator().GetMemoryResource();
   auto typed_columns = utils::pmr::vector<TypedValue>(mem);
   typed_columns.reserve(row.size());
   for (auto &column : row) {
-    typed_columns.emplace_back(std::move(column));
+    if (!nullif.has_value() || column != nullif.value()) {
+      typed_columns.emplace_back(std::move(column));
+    } else {
+      typed_columns.emplace_back(TypedValue(mem));
```
antoniofilipovic, line 4686:
```suggestion
      typed_columns.emplace_back(mem);
```

Since you are using `emplace_back` on vector, you can pass just `mem` and it should construct `TypedValue` by passing `mem` to TypedValue `ctor`. I am not sure if this does a necessary copy, but I think yes.

I checked it does 1 move and 1 copy unnecessary.
https://godbolt.org/z/sbf38xYhd

src/query/plan/operator.cpp
```diff
@@ -4674,22 +4675,31 @@ auto ToOptionalString(ExpressionEvaluator *evaluator, Expression *expression) ->
...
+                            std::optional<utils::pmr::string> &nullif) {
   // a valid row has the same number of elements as the header
   auto *mem = row.get_allocator().GetMemoryResource();
   utils::pmr::map<utils::pmr::string, TypedValue> m(mem);
   for (auto i = 0; i < row.size(); ++i) {
-    m.emplace(std::move(header[i]), std::move(row[i]));
+    if (!nullif.has_value() || row[i] != nullif.value()) {
+      m.emplace(std::move(header[i]), std::move(row[i]));
+    } else {
+      m.emplace(std::move(header[i]), TypedValue(mem));
```
antoniofilipovic, line 4701: Same comment here

antoniofilipovic (changes requested):
Small change, otherwise looks good

src/query/plan/operator.cpp
Josipmrden: I was getting some weird behaviour with explicit constructor on memory but I'm going to check it once again

src/query/plan/operator.cpp
Josipmrden: Fixed with `piecewise_construct`

src/query/plan/operator.cpp
Josipmrden: Fixed with no arguments in `emplace_back()`

antoniofilipovic (approved):

PR #916, opened 2023-05-03
[master < T916] Fix MATCH + LOAD CSV to load csv only once

Josipmrden:
MATCH + LOAD CSV in this PR will be modified in a way that it allows for MATCH or MERGE to yield one value. If there are more values yielded, it will cause a runtime error.

[master < Task] PR

`MATCH and MERGE clauses now can be put before LOAD CSV if they yield only one row. That way, entities can be matched only once prior to reading rows from a file, which speeds up execution.`


src/query/plan/operator.cpp
```diff
@@ -4724,12 +4722,14 @@ class LoadCsvCursor : public Cursor {
 
     bool input_pulled = input_cursor_->Pull(frame, context);
 
-    // If the input is Once, we have to keep going until we read all the rows,
-    // regardless of whether the pull on Once returned false.
-    // If we have e.g. MATCH(n) LOAD CSV ... AS x SET n.name = x.name, then we
-    // have to read at most cardinality(n) rows (but we can read less and stop
-    // pulling MATCH).
-    if (!input_is_once_ && !input_pulled) return false;
+    if (input_pulled) {
```
antoniofilipovic, line 4725: Slight nitpicking but I think you don't need `input_pulled` variable, you can just place `input_cursor_->Pull(frame, context);` in `if` statement

antoniofilipovic (approved):
One nitpick, otherwise looks good to go

Josipmrden:
@vpavicic

vpavicic:
@Josipmrden  - ok?

To speed up the execution of the LOAD CSV clause, you can now MATCH and MERGE entities prior to reading the rows from a CSV file. Until now, the LOAD CSV clause had to be the first clause in the query.

PR #892, opened 2023-04-23
[master < E219] Map projection

antepusic:
[master < Epic] PR
- [x] Check, and update documentation if necessary
- [x] Write E2E tests
- [x] Compare the [benchmarking results](https://bench-graph.memgraph.com/) between the master branch and the Epic branch
- [x] Provide the full content or a guide for the final git message
> Add the map projection feature

* NB: Benchmarks compared between master and #931 because that branch builds on all other task branches and thus has all the new code

To keep docs changelog up to date, one more thing to do:
- [x] Write a release note here
- [x] Tag someone from docs team in the comments

Release note:
> It’s now possible to use the map projection syntax to create maps. Map projections are convenient for building maps based on existing values and they are used by data access tools like GraphQL and Spring Data.

PR #931, opened 2023-05-08
[T1254 < T1255] All-properties lookup

antepusic:
[[Task](https://github.com/memgraph/memgraph/pull/922) < Task] PR
> Add execution logic for all-properties lookup and map projection tests

To keep the docs changelog up to date, one more thing to do:
- [x] Write a release note here ← only the note from #892 should go in


tests/unit/query_plan.cpp
```diff
@@ -958,6 +958,19 @@ TYPED_TEST(TestPlanner, MapLiteralAggregationReturn) {
...
+//   AstStorage storage;
+//   FakeDbAccessor dba;
+//   auto sum = SUM(LITERAL(2), false);
+//   auto group_by_literal = LITERAL(42);
+//   auto *query = QUERY(SINGLE_QUERY(WITH(MAP(), AS("map")), RETURN(MAP({storage.GetPropertyIx("sum"), sum}),
+//                                                                   AS("result"), group_by_literal, AS("group_by"))));
+//   auto aggr = ExpectAggregate({sum}, {group_by_literal});
+//   CheckPlan<TypeParam>(query, storage, aggr, ExpectProduce());
+// }
+
```
antoniofilipovic, line 973: This is commented out

tests/unit/query_plan.cpp
```diff
@@ -1020,6 +1033,9 @@ TYPED_TEST(TestPlanner, MapWithAggregationAndGroupBy) {
   CheckPlan<TypeParam>(query, storage, aggr, ExpectProduce());
 }
 
+// TODO ante
+// TYPED_TEST(TestPlanner, MapProjectionWithAggregationAndGroupBy) {}
```
antoniofilipovic, line 1037: TODO still left

src/query/interpret/eval.hpp
```diff
@@ -467,7 +468,102 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
...
+      }
+      case TypedValue::Type::Graph: {
+        const auto &graph = expression_result.ValueGraph();
+
+        utils::pmr::vector<TypedValue> vertices(ctx_->memory);
+        vertices.reserve(graph.vertices().size());
+        for (const auto &v : graph.vertices()) {
+          vertices.emplace_back(TypedValue(v, ctx_->memory));
+        }
+        result.emplace("nodes", TypedValue(vertices, ctx_->memory));
```
antoniofilipovic, line 553: I am not 100% sure, but I think we should use `std::move` on vertices to explicitly use `move ctor`


result.emplace("nodes", TypedValue(std::move(vertices), ctx_->memory));

src/query/interpret/eval.hpp
```diff
@@ -467,7 +468,102 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
...
+
+        utils::pmr::vector<TypedValue> edges(ctx_->memory);
+        edges.reserve(graph.edges().size());
+        for (const auto &e : graph.edges()) {
+          edges.emplace_back(TypedValue(e, ctx_->memory));
+        }
+        result.emplace("edges", TypedValue(edges, ctx_->memory));
```
antoniofilipovic, line 560: Same here

src/query/interpret/eval.hpp
```diff
@@ -534,7 +630,22 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
 
   TypedValue Visit(MapProjectionLiteral &literal) override {
     TypedValue::TMap result(ctx_->memory);
-    for (const auto &pair : literal.elements_) result.emplace(pair.first.name, pair.second->Accept(*this));
+    for (const auto &pair : literal.elements_) {
+      if (pair.first.name == "*") {  // AllPropertiesSelector
```
antoniofilipovic, line 634: Can we make this `*` a constant

antoniofilipovic (changes requested):
There are few changes to make, but it seems overall good to me

src/query/interpret/eval.hpp
as51340: yes you are right fico. There is move constructor so no reason not to use it

src/query/interpret/eval.hpp
```diff
@@ -534,7 +630,22 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
...
+        auto maybe_selector = pair.second->Accept(*this);
+
+        if (maybe_selector.type() != TypedValue::Type::Map) {
+          throw QueryRuntimeException("Expected a map from all properties selection, got {}.", maybe_selector.type());
+        }
+
+        for (const auto &pair : maybe_selector.ValueMap()) {
```
as51340, line 641: I would suggest unwrapping pair into two variable in both inner and outer loop because here you have variable overloading which makes it hard to read

as51340: Also, I think you can change this inner for loop with `map::merge`. It will perform moving of elements from source to target map

tests/benchmark/query/eval.cpp
```diff
@@ -60,6 +60,8 @@ BENCHMARK_TEMPLATE(MapLiteral, NewDeleteResource)->Range(512, 1U << 15U)->Unit(b
 
 BENCHMARK_TEMPLATE(MapLiteral, MonotonicBufferResource)->Range(512, 1U << 15U)->Unit(benchmark::kMicrosecond);
 
+// TODO ante benchmark template for MapProjectionLiteral
```
as51340, line 63: TODO comment just that you are aware

src/query/interpret/eval.hpp
antepusic: Done

src/query/interpret/eval.hpp
antepusic: Done

src/query/interpret/eval.hpp
antepusic: Done

tests/unit/cypher_main_visitor.cpp
```diff
@@ -1032,18 +1032,52 @@ TEST_P(CypherMainVisitorTest, MapLiteral) {
...
-  // ast_generator.CheckLiteral(map_projection_literal->elements_[ast_generator.Prop("name")], "Morgan");
-  // ast_generator.CheckLiteral(map_projection_literal->elements_[ast_generator.Prop("lastName")], "Freeman");
-  // ast_generator.CheckLiteral(map_projection_literal->elements_[ast_generator.Prop("age")], "85");
+  ASSERT_EQ(4, map_projection_literal->elements_.size());
+  ASSERT_EQ(map_projection_literal->elements_[ast_generator.Prop("name")]->GetTypeInfo().name, "PropertyLookup");
+  ASSERT_EQ(map_projection_literal->elements_[ast_generator.Prop("*")]->GetTypeInfo().name, "AllPropertiesLookup");
+  ASSERT_EQ(map_projection_literal->elements_[ast_generator.Prop("age")]->GetTypeInfo().name, "Identifier");
+  ASSERT_EQ(map_projection_literal->elements_[ast_generator.Prop("lastName")]->GetTypeInfo().name,
+            std::string(typeid(ast_generator).name()).ends_with("CachedAstGenerator") ? "ParameterLookup"
+                                                                                      : "PrimitiveLiteral");
```
antepusic, line 1049: NB for reviewers: `CachedAstGenerator` turns `PrimitiveLiteral` elements into `ParameterLookup` instances. This is not an issue with `MapProjectionLiteral` because it happens to such elements in other classes like `MapLiteral` as well.

tests/unit/query_plan.cpp
antepusic: Done

src/query/interpret/eval.hpp
```diff
@@ -467,7 +469,105 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
...
+    std::map<std::string, TypedValue> result_;
+
+    auto expression_result = all_properties_lookup.expression_->Accept(*this);
+    switch (expression_result.type()) {
+      case TypedValue::Type::Null:
+        return TypedValue(ctx_->memory);
+      case TypedValue::Type::Vertex: {
+        const auto vertex = expression_result.ValueVertex();
+        auto properties = *vertex.Properties(view_);
+        for (const auto &[property_id, value] : properties) {
```
as51340, line 488:
```suggestion
        for (auto properties = *vertex.Properties(view_); const auto &[property_id, value] : properties) {
```

as51340: it's picky and feel free not to add it but I like it since you get variable removal from the scope automatically

src/query/interpret/eval.hpp
```diff
@@ -467,7 +469,105 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
...
+          result.emplace(dba_->PropertyToName(property_id), value);
+        }
+        return TypedValue(result, ctx_->memory);
+      }
+      case TypedValue::Type::Edge: {
+        const auto &edge = expression_result.ValueEdge();
+        auto properties = *edge.Properties(view_);
+        for (const auto &[property_id, value] : properties) {
```
as51340, line 496: same

src/query/interpret/eval.hpp
as51340, line 488:
```suggestion
        for (const auto vertex = expression_result.ValueVertex(); const auto &[property_id, value] : *vertex.Properties(view_);) {
```

It's picky and feel free to ignore it but I like it because of automatic scope

src/query/interpret/eval.hpp
```diff
@@ -476,24 +476,27 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
...
 
     auto expression_result = all_properties_lookup.expression_->Accept(*this);
     switch (expression_result.type()) {
       case TypedValue::Type::Null:
         return TypedValue(ctx_->memory);
       case TypedValue::Type::Vertex: {
-        const auto &vertex = expression_result.ValueVertex();
-        for (const auto &[property_id, value] : *vertex.Properties(view_)) {
+        const auto vertex = expression_result.ValueVertex();
+        auto properties = *vertex.Properties(view_);
```
antoniofilipovic, line 487: I think these two will create copies, right? Why not using reference here?

as51340: Shouldn't RVO work here? But there is really no sense in using vertex as a local variable

src/query/interpret/eval.hpp
```diff
@@ -467,7 +469,105 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
...
+        for (const auto &[property_id, value] : properties) {
+          result.emplace(dba_->PropertyToName(property_id), value);
+        }
+        return TypedValue(result, ctx_->memory);
+      }
+      case TypedValue::Type::Edge: {
+        const auto &edge = expression_result.ValueEdge();
+        auto properties = *edge.Properties(view_);
```
as51340, line 495: Same

as51340 (approved):

src/query/interpret/eval.hpp
antepusic: Right, I’d too expect RVO to prevent that. Anyway, I’ve accepted your suggestions and cut the local variables - they were only there for readability

antoniofilipovic: I think it should not kick in. As far as I understood, RVO kicks in when you have created local variable and then return that local variable. In this case, we return `vertex_v` from function which exists in `TypedValue`. And it that case I think it doesn't kick in.

antoniofilipovic: I think this is okay example: https://godbolt.org/z/ex4qc41v7

antoniofilipovic (approved):
All seems good on my side

src/query/interpret/eval.hpp
```diff
@@ -467,7 +469,102 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
...
       default:
-        throw QueryRuntimeException("Only nodes, edges, maps and temporal types have properties to be looked-up.");
+        throw QueryRuntimeException(
+            "Only nodes, edges, maps, temporal types and graphs have properties to be looked up.");
+    }
+  }
+
+  TypedValue Visit(AllPropertiesLookup &all_properties_lookup) override {
+    TypedValue::TMap result(ctx_->memory);
+    std::map<std::string, TypedValue> result_;
```
antoniofilipovic, line 479: @antepusic can you actually make this utils::pmr::unordered_map:

`utils::pmr::unordered_map<std::string, TypedValue> results(context.evaluation_context.memory);`

This way we will track memory

antoniofilipovic: Actually you can use utils::pmr::map directly, as we have for it constructors in TypedValue.

`utils::pmr::map<utils::pmr::string, TypedValue>`;

PR #933, opened 2023-05-10
[master < T891] (UN)LOCK query enhancement

### LINKED ISSUE #891: Implement return feedback on LOCK/UNLOCK DATA DIRECTORY query
Using the (UN)LOCK DATA DIRECTORY query is hard at the moment as Memgraph returns no feedback about the lock status: 
I propose the following messages: 

1. LOCK DATA DIRECTORY; - upon success, should return: Data directory is now locked.
2. UNLOCK DATA DIRECOTRY; - upon success should return: Data directory is now unlocked.
3. When you try to run LOCK DATA DIRECTORY on a locked data directory, the following message should be returned: Data directory is already locked.
4. When you try to run UNLOCK DATA DIRECTORY on an unlocked data directory, the following message should be returned: Data directory is already unlocked. 
(ATM it says" Failed to unlock the directory" and i had a mini heart attack)

Also to provide more control and information, I propose a new query
DATA DIRECTORY LOCK STATUS: 
that will return the info if the data directory is locked or unlocked

andrejtonev:
[master < Task] PR

docs PR [#881](https://github.com/memgraph/docs/pull/881)
resolves #891


src/query/frontend/stripped_lexer_constants.hpp
```diff
@@ -145,6 +145,7 @@ const trie::Trie kKeywords = {"union",
                               "drop",
                               "show",
                               "stats",
+                              "status",
```
antoniofilipovic, line 148: I think we forgot before to add keywords for `LOCK`, `UNLOCK`, `DATA`, and `DIRECTORY`. I think we can add those too here.

src/query/interpreter.cpp
```diff
@@ -1777,23 +1777,43 @@ PreparedQuery PrepareLockPathQuery(ParsedQuery parsed_query, bool in_explicit_tr
...
                        [interpreter_context, action = lock_path_query->action_](
                            AnyStream *stream, std::optional<int> n) -> std::optional<QueryHandlerResult> {
+                         std::vector<std::vector<TypedValue>> status;
+                         const auto locked_status = interpreter_context->db->IsPathLocked();
+                         std::string res;
+
                          switch (action) {
                            case LockPathQuery::Action::LOCK_PATH:
-                             if (!interpreter_context->db->LockPath()) {
+                             // LockPath returns false only if the path doesn't exist (not our case here?)
```
antoniofilipovic, line 1790: If we change here return value from LockPath, then we can know whether there was error when we tried to lock path (path doesn't exist) or for example that path can't be locked.

src/storage/v2/storage.cpp
```diff
@@ -1980,6 +1980,11 @@ utils::BasicResult<Storage::CreateSnapshotError> Storage::CreateSnapshot(std::op
   return CreateSnapshotError::ReachedMaxNumTries;
 }
 
+bool Storage::IsPathLocked() {
+  auto locker_accessor = global_locker_.Access();
+  return locker_accessor.IsPathLocked(config_.durability.storage_directory);
+}
```
antoniofilipovic, line 1986: For isPathLocked, I am not sure, but maybe we can do similar, and return enum, PathLocked::True, PathLocked::False, PathLocked::NonExistingPath? What do you think?

src/storage/v2/storage.cpp
```diff
@@ -1980,6 +1980,11 @@ utils::BasicResult<Storage::CreateSnapshotError> Storage::CreateSnapshot(std::op
...
+
 bool Storage::LockPath() {
   auto locker_accessor = global_locker_.Access();
   return locker_accessor.AddPath(config_.durability.storage_directory);
```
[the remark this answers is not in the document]
andrejtonev: Agree, the API can be a lot better.
The LockPath basically just adds the path to a set. We can return 2 types of error:
1. Path doesn't exist
2. Failed to add (failed to add to set)

src/storage/v2/storage.cpp
andrejtonev: I would like to have all 3 calls have the same behavior. Could we just use the BasicResult from the LockPath?
Or is that more for error passing?

src/query/interpreter.cpp
andrejtonev: The Un/LockPath just remove/add the path to a set. Not sure when an existing path couldn't be locked.
Does something else happen under the hood?

src/query/interpreter.cpp
```diff
@@ -1777,23 +1777,43 @@ PreparedQuery PrepareLockPathQuery(ParsedQuery parsed_query, bool in_explicit_tr
...
+                             res = interpreter_context->db->UnlockPath() ? "Data directory is now unlocked."
+                                                                         : "Data directory is already unlocked.";
                              break;
+                           case LockPathQuery::Action::STATUS:
+                             res = locked_status ? "Data directory is locked." : "Data directory is unlocked.";
+                             break;
+                             // Needed?
+                             // default:
+                             //   throw...
+                             // break;
```
antoniofilipovic, line 1808: I think you can also remove this, it should fail on cypher query parser if query is not correct

src/storage/v2/storage.cpp
antoniofilipovic: As we talked about this, I will resolve comment

src/storage/v2/storage.cpp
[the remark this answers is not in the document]
antoniofilipovic: As we talked about this too, I will resolve conversation

src/query/interpreter.cpp
antoniofilipovic: Talked about this too, so I will resolve conversation

src/query/interpreter.cpp
andrejtonev: Removed

andrejtonev:
Pushed the new API.
I also updated the tests and replication client to handle the new API.

src/query/frontend/stripped_lexer_constants.hpp
antoniofilipovic: @andrejtonev what do you think on this one?

antoniofilipovic (approved):
Looks good now! Great effort on first-ever PR on Memgraph :partying_face: Check one more comment that is left, and if you think it is resolved, just ping me on Slack so I can :ship: :it:

src/query/frontend/stripped_lexer_constants.hpp
andrejtonev: Ah sorry, it slipped my mind. I'll add them now.

Josipmrden (approved):

PR #943, opened 2023-05-19
[E850 < T943] Add larger than memory local cache

as51340:
[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR


src/query/plan/operator.cpp
```diff
@@ -493,7 +494,8 @@ UniqueCursorPtr ScanAll::MakeCursor(utils::MemoryResource *mem) const {
 
   auto vertices = [this](Frame &, ExecutionContext &context) {
     auto *db = context.db_accessor;
-    return std::make_optional(db->Vertices(view_));
+    auto vertices = std::make_optional(db->Vertices(view_));
```
andrejtonev, line 497: Why not just return?

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -1950,9 +1952,9 @@ void NextPermittedEdge(mgp_edges_iterator &it, const bool for_in) {
     const auto *auth_checker = it.source_vertex.graph->ctx->auth_checker.get();
     const auto view = it.source_vertex.graph->view;
     while (*impl_it != end) {
-      if (auth_checker->Has(**impl_it, memgraph::query::AuthQuery::FineGrainedPrivilege::READ)) {
-        const auto &check_vertex =
-            it.source_vertex.getImpl() == (*impl_it)->From() ? (*impl_it)->To() : (*impl_it)->From();
+      auto edgeAcc = **impl_it;
```
andrejtonev, line 1955: Why copy here?

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -1968,6 +1970,14 @@ void NextPermittedEdge(mgp_edges_iterator &it, const bool for_in) {
 mgp_error mgp_vertex_iter_in_edges(mgp_vertex *v, mgp_memory *memory, mgp_edges_iterator **result) {
   return WrapExceptions(
       [v, memory] {
+        auto dbAccessor = v->graph->impl;
+        if (std::holds_alternative<memgraph::query::DbAccessor *>(dbAccessor)) {
+          std::get<memgraph::query::DbAccessor *>(dbAccessor)
+              ->PrefetchInEdges(std::get<memgraph::query::VertexAccessor>(v->impl));
+        } else {
```
andrejtonev, line 1977: else if (hold Subgraph...)

src/query/procedure/mg_procedure_impl.cpp
andrejtonev: Same for lines 2010 2015 2133 2140

src/utils/memory.hpp
```diff
@@ -134,7 +134,7 @@ class Allocator {
 
   void deallocate(T *p, size_t count_elements) { memory_->Deallocate(p, count_elements * sizeof(T), alignof(T)); }
 
-  /// Return default NewDeleteResource() allocator.
+  /// Return default NewDevertex_id_leteResource() allocator.
```
andrejtonev, line 137: This seems like a typo.

src/utils/rocksdb_serialization.hpp
```diff
@@ -0,0 +1,292 @@
...
+#include "storage/v2/id_types.hpp"
+#include "storage/v2/property_store.hpp"
+#include "storage/v2/vertex.hpp"
+#include "storage/v2/vertex_accessor.hpp"
+#include "utils/exceptions.hpp"
+#include "utils/string.hpp"
+
+namespace memgraph::utils {
+
+constexpr const char *outEdgeDirection = "0";
```
andrejtonev, line 32: static constexpr

src/utils/rocksdb_serialization.hpp
```diff
@@ -0,0 +1,292 @@
...
+constexpr const char *inEdgeDirection = "1";
```
andrejtonev, line 33: static constexpr

andrejtonev (commented):
Okay, this is my first pass. Didn't get into storage or tests.
This will have to be a multi part review :|

src/query/interpreter.cpp
```diff
@@ -2061,6 +2126,54 @@ PreparedQuery PrepareIsolationLevelQuery(ParsedQuery parsed_query, const bool in
...
+      if (auto vertex_cnt_approx = interpreter_context->db->GetInfo().vertex_count; vertex_cnt_approx > 0) {
+        main_guard.unlock();
+        throw utils::BasicException(
+            "You cannot switch from in-memory storage to on-disk storage when the database "
+            "contains data. Please delete all entries from the database, run FREE MEMORY and then repeat this "
+            "query. ");
+      }
+
+      main_guard.unlock();
+      if (interpreter_context->interpreters->size() > 1) {
```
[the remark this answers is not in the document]
andrejtonev: Just realized that the interpreters is Synchronized so ignore this :D

src/storage/v2/constraints/constraints.hpp
```diff
@@ -0,0 +1,43 @@
...
+namespace memgraph::storage {
+
+struct Constraints {
+  Constraints(const Config &config, StorageMode storage_mode) {
+    std::invoke([this, config, storage_mode]() {
+      existence_constraints_ = std::make_unique<ExistenceConstraints>();
+      if (storage_mode == StorageMode::IN_MEMORY_TRANSACTIONAL || storage_mode == StorageMode::IN_MEMORY_ANALYTICAL) {
+        unique_constraints_ = std::make_unique<InMemoryUniqueConstraints>();
+      } else {
+        unique_constraints_ = std::make_unique<DiskUniqueConstraints>(config);
```
antoniofilipovic, line 28: I think using `switch` statement is a better option as in that case when and if we add new storage mode, build will break as we must explicitly cover cases

src/storage/v2/disk/storage.hpp
```diff
@@ -0,0 +1,253 @@
...
+    /// Main storage
+    utils::SkipList<storage::Vertex> vertices_;
+    utils::SkipList<storage::Vertex> indexed_vertices_;
+
+    /// We need them because query context for indexed reading is cleared after the query is done not after the
+    /// transaction is done
+    std::list<Delta> index_deltas_;
+    utils::SkipList<storage::Edge> edges_;
+    Config::Items config_;
+    std::vector<std::string> edges_to_delete_;
```
andrejtonev, line 188: should the vectors be protected?

src/storage/v2/disk/storage.hpp
```diff
@@ -0,0 +1,253 @@
...
+
+  std::unique_ptr<Storage::Accessor> Access(std::optional<IsolationLevel> override_isolation_level) override {
+    auto isolation_level = override_isolation_level.value_or(isolation_level_);
+    if (isolation_level != IsolationLevel::SNAPSHOT_ISOLATION) {
+      throw utils::NotYetImplemented("Disk storage supports only SNAPSHOT isolation level.");
+    }
+    return std::unique_ptr<DiskAccessor>(new DiskAccessor{this, isolation_level, storage_mode_});
+  }
+
+  RocksDBStorage *GetRocksDBStorage() { return kvstore_.get(); }
```
andrejtonev, line 201: const

src/storage/v2/disk/storage.hpp
```diff
@@ -0,0 +1,253 @@
...
+    std::vector<std::pair<std::string, std::string>> vertices_to_delete_;
+    rocksdb::Transaction *disk_transaction_;
+  };
+
+  std::unique_ptr<Storage::Accessor> Access(std::optional<IsolationLevel> override_isolation_level) override {
+    auto isolation_level = override_isolation_level.value_or(isolation_level_);
+    if (isolation_level != IsolationLevel::SNAPSHOT_ISOLATION) {
+      throw utils::NotYetImplemented("Disk storage supports only SNAPSHOT isolation level.");
+    }
+    return std::unique_ptr<DiskAccessor>(new DiskAccessor{this, isolation_level, storage_mode_});
```
andrejtonev, line 198: use make_unique

andrejtonev (commented):
First batch of comments from storage dir

src/memgraph.cpp
```diff
@@ -866,11 +866,15 @@ int main(int argc, char **argv) {
                      .wal_file_size_kibibytes = FLAGS_storage_wal_file_size_kib,
                      .wal_file_flush_every_n_tx = FLAGS_storage_wal_file_flush_every_n_tx,
                      .snapshot_on_exit = FLAGS_storage_snapshot_on_exit,
-                     .restore_replicas_on_startup = true,
-                     .items_per_batch = FLAGS_storage_items_per_batch,
-                     .recovery_thread_count = FLAGS_storage_recovery_thread_count,
-                     .allow_parallel_index_creation = FLAGS_storage_parallel_index_recovery},
-      .transaction = {.isolation_level = ParseIsolationLevel()}};
```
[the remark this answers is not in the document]
as51340: No, it wasn't on purpose :+1:

src/query/db_accessor.hpp
```diff
@@ -494,6 +506,10 @@ class SubgraphDbAccessor final {
 
   const std::string &EdgeTypeToName(storage::EdgeTypeId type) const;
 
+  void PrefetchOutEdges(const SubgraphVertexAccessor &vertex) const { db_accessor_.PrefetchOutEdges(vertex.impl_); }
+
+  void PrefetchInEdges(const SubgraphVertexAccessor &vertex) const { db_accessor_.PrefetchOutEdges(vertex.impl_); }
```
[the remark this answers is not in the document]
as51340: nice :joy:

src/mg_import_csv.cpp
```diff
@@ -699,13 +700,13 @@ int main(int argc, char *argv[]) {
   }
 
   std::unordered_map<NodeId, memgraph::storage::Gid> node_id_map;
-  memgraph::storage::Storage store{{
+  std::unique_ptr<memgraph::storage::Storage> store{new memgraph::storage::InMemoryStorage{{
```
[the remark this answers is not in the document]
as51340: I would leave like this yes because fico said we don't use that at all anymore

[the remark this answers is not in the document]
as51340: we need to delete that from master I think

src/query/interpreter.cpp
```diff
@@ -145,28 +152,30 @@ class ReplQueryHandler final : public query::ReplicationQueryHandler {
 
   /// @throw QueryRuntimeException if an error ocurred.
   void SetReplicationRole(ReplicationQuery::ReplicationRole replication_role, std::optional<int64_t> port) override {
```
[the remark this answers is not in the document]
as51340: Yes so atm we are sure because everything regarding replication is disabled (replication handler is disabled). So currently in my view not but if you really think we should, I can move check here also, I just didn't see any benefit from it at the moment of implementing

src/query/interpreter.cpp
```diff
@@ -1153,11 +1167,37 @@ std::optional<plan::ProfilingStatsWithTotalTime> PullPlan::Pull(AnyStream *strea
...
 }  // namespace
 
-InterpreterContext::InterpreterContext(storage::Storage *db, const InterpreterConfig config,
+InterpreterContext::InterpreterContext(const storage::Config storage_config, const InterpreterConfig interpreter_config,
                                        const std::filesystem::path &data_directory)
-    : db(db), trigger_store(data_directory / "triggers"), config(config), streams{this, data_directory / "streams"} {}
+    : trigger_store(data_directory / "triggers"),
+      config(interpreter_config),
+      streams{this, data_directory / "streams"} {
+  if (utils::DirExists(storage_config.disk.main_storage_directory)) {
```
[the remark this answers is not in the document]
as51340: Yes it persists. The reason is that once you move from InMemory to OnDisk, there is no going back. At least for now, maybe at some later stages, we should think what to do about it. If you want InMemory, two options:
1. delete mg_data
2. open new memgraph with different data directory

src/query/interpreter.cpp
```diff
@@ -1153,11 +1167,37 @@ std::optional<plan::ProfilingStatsWithTotalTime> PullPlan::Pull(AnyStream *strea
 }
 
 using RWType = plan::ReadWriteTypeChecker::RWType;
+
+bool IsWriteQueryOnMainMemoryReplica(storage::Storage *storage,
+                                     const query::plan::ReadWriteTypeChecker::RWType query_type) {
+  if (storage->GetStorageMode() != storage::StorageMode::ON_DISK_TRANSACTIONAL) {
```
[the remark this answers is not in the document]
as51340: Yep you are right

src/query/interpreter.cpp
```diff
@@ -1799,16 +1847,21 @@ PreparedQuery PrepareLockPathQuery(ParsedQuery parsed_query, bool in_explicit_tr
 }
 
 PreparedQuery PrepareFreeMemoryQuery(ParsedQuery parsed_query, bool in_explicit_transaction,
-                                     InterpreterContext *interpreter_context) {
+                                     InterpreterContext *interpreter_context, DbAccessor *dba) {
```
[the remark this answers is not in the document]
as51340: You are right. It is better to use GeStorageMode on storage because we don't create accessor for type of query.

src/query/interpreter.cpp
```diff
@@ -2061,6 +2126,54 @@ PreparedQuery PrepareIsolationLevelQuery(ParsedQuery parsed_query, const bool in
...
+    if (SwitchingFromDiskToInMemory(current_mode, requested_mode)) {
+      throw utils::BasicException(
+          "You cannot switch from on-disk storage to in-memory storage while the database is running. "
+          "Please restart the database to switch to in-memory storage.");
+    }
+    if (SwitchingFromInMemoryToDisk(current_mode, requested_mode)) {
+      std::unique_lock main_guard{interpreter_context->db->main_lock_};
+
+      if (auto vertex_cnt_approx = interpreter_context->db->GetInfo().vertex_count; vertex_cnt_approx > 0) {
+        main_guard.unlock();
```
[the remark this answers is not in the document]
as51340: You are right about first unlock but not about the second. The reason is that the main_lock needs to be released before we destroy InMemoyStorage

src/query/interpreter.cpp
```diff
@@ -2061,6 +2126,54 @@ PreparedQuery PrepareIsolationLevelQuery(ParsedQuery parsed_query, const bool in
...
+      main_guard.unlock();
+      if (interpreter_context->interpreters->size() > 1) {
+        throw utils::BasicException(
+            fmt::format("You cannot switch from in-memory storage to on-disk storage when there are "
+                        "multiple sessions active. Please close all other sessions and try again. {}",
+                        interpreter_context->interpreters->size()));
+      }
+
+      auto db_config = interpreter_context->db->config_;
+      interpreter_context->db = std::make_unique<memgraph::storage::DiskStorage>(db_config);
```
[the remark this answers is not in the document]
as51340: Not sure exactly what you mean but when the `STORAGE MODE ON_DISK_TRANSACTIONAL` query is called, this doesn't create a transaction, just the new interpreter so the condition that we aren't in a transaction is fulfilled for sure

src/query/interpreter.cpp
```diff
@@ -2061,6 +2126,54 @@ PreparedQuery PrepareIsolationLevelQuery(ParsedQuery parsed_query, const bool in
...
+
+      main_guard.unlock();
+      if (interpreter_context->interpreters->size() > 1) {
+        throw utils::BasicException(
+            fmt::format("You cannot switch from in-memory storage to on-disk storage when there are "
+                        "multiple sessions active. Please close all other sessions and try again. {}",
+                        interpreter_context->interpreters->size()));
+      }
+
+      auto db_config = interpreter_context->db->config_;
```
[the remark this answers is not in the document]
as51340: What do you mean, do we need to pass it to the new storage? I mean config object shares both in-memory and disk settings so I would pass it to new storage definitely.

src/query/interpreter.cpp
```diff
@@ -2061,6 +2126,54 @@ PreparedQuery PrepareIsolationLevelQuery(ParsedQuery parsed_query, const bool in
...
+      interpreter_context->db = std::make_unique<memgraph::storage::DiskStorage>(db_config);
+    }
+    return std::vector<std::vector<TypedValue>>();
+  };
+  return callback;
```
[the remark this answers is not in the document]
as51340: So not 100% sure on which exactly issues you think can happen but I understand what you meant when the session is created in the middle. I handled that in a different way. Each accessor gets created with some storage mode and then upon commit we check whether this is still the active mode, if not abort. I think this is better than having interpreter_context lock because otherwise you would need to lock Access method from the storage class, at least this was my conclusion but am willing to discuss that in more details.

src/query/plan/operator.cpp
as51340: ah left from debugging, stupid

src/query/procedure/mg_procedure_impl.cpp
as51340: Aidar worked on that code, he will know. CC @Darych

src/query/procedure/mg_procedure_impl.cpp
as51340: cc @Darych

src/storage/v2/disk/helper_storage.hpp
```diff
@@ -0,0 +1,390 @@
...
+//   return status.ok();
+// }
+
+class RocksDBStorage {
+ public:
+  explicit RocksDBStorage() {
+    options_.create_if_missing = true;
+    // options_.OptimizeLevelStyleCompaction();
+    std::filesystem::path rocksdb_path = "./rocks_experiment_unit";
+    MG_ASSERT(utils::EnsureDir(rocksdb_path), "Unable to create storage folder on the disk.");
```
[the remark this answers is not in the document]
as51340: The file should be deleted :)

src/storage/v2/disk/label_index.cpp
```diff
@@ -0,0 +1,261 @@
...
+  kvstore_->options_.comparator = new ComparatorWithU64TsImpl();
+  logging::AssertRocksDBStatus(rocksdb::TransactionDB::Open(kvstore_->options_, rocksdb::TransactionDBOptions(),
+                                                            config.disk.label_index_directory, &kvstore_->db_));
+}
+
+bool DiskLabelIndex::CreateIndex(LabelId label, const std::vector<std::pair<std::string, std::string>> &vertices) {
+  index_.emplace(label);
+  /// How to remove duplication
+  /// TODO: how to deal with commit timestamp in a better way
+  auto disk_transaction = std::unique_ptr<rocksdb::Transaction>(
```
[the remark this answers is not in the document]
as51340: definitely yes

src/storage/v2/durability/durability.cpp
```diff
@@ -140,17 +143,27 @@ void RecoverIndicesAndConstraints(const RecoveredIndicesAndConstraints &indices_
...
-    auto ret = CreateExistenceConstraint(constraints, item.first, item.second, vertices->access());
-    if (ret.HasError() || !ret.GetValue()) throw RecoveryFailure("The existence constraint must be created here!");
+  for (const auto &[label, property] : indices_constraints.constraints.existence) {
+    if (constraints->existence_constraints_->ConstraintExists(label, property)) {
+      throw RecoveryFailure("The existence constraint must be created here!");
+    }
+
+    if (auto violation = ExistenceConstraints::ValidateVerticesOnConstraint(vertices->access(), label, property);
+        violation.has_value()) {
+      throw RecoveryFailure("The existence constraint must be created here!");
```
andrejtonev, line 153: Use different messages if the two failures are different.
const auto or don't capture the variable

src/storage/v2/durability/durability.cpp
```diff
@@ -140,17 +143,27 @@ void RecoverIndicesAndConstraints(const RecoveredIndicesAndConstraints &indices_
...
+
     spdlog::info("A existence constraint is recreated from metadata.");
   }
   spdlog::info("Existence constraints are recreated from metadata.");
 
   // Recover unique constraints.
   spdlog::info("Recreating {} unique constraints from metadata.", indices_constraints.constraints.unique.size());
   for (const auto &item : indices_constraints.constraints.unique) {
-    auto ret = constraints->unique_constraints.CreateConstraint(item.first, item.second, vertices->access());
+    auto *mem_unique_constraints = static_cast<InMemoryUniqueConstraints *>(constraints->unique_constraints_.get());
```
andrejtonev, line 165: We cast in a few places, is it guaranteed that we end up here only if InMemory?

src/storage/v2/disk/label_index.cpp
```diff
@@ -0,0 +1,261 @@
...
+                  utils::SerializeIdType(vertex.gid));
+    if (auto status = disk_transaction->Delete(maybe_old_disk_key.value()); !status.ok()) {
+      return false;
+    }
+  }
+
+  for (const LabelId index_label : index_) {
+    if (utils::Contains(vertex.labels, index_label)) {
+      if (!disk_transaction
+               ->Put(utils::SerializeVertexAsKeyForLabelIndex(index_label, vertex.gid),
```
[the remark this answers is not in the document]
as51340: It shouldn't fail because it is only string manipulation

src/storage/v2/delta.hpp
```diff
@@ -123,6 +124,7 @@ inline bool operator!=(const PreviousPtr::Pointer &a, const PreviousPtr::Pointer
 struct Delta {
   enum class Action {
     // Used for both Vertex and Edge
+    DELETE_DESERIALIZED_OBJECT,
```
antoniofilipovic, line 127: As this is only used for disk, I would make a comment which should explain usage od delta action enums, which are used for in memory and which for disk storage

src/storage/v2/disk/label_index.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+  rocksdb::ReadOptions ro;
+  std::string strTs = utils::StringTimestamp(std::numeric_limits<uint64_t>::max());
+  rocksdb::Slice ts(strTs);
+  ro.timestamp = &ts;
+  auto it = std::unique_ptr<rocksdb::Iterator>(disk_transaction->GetIterator(ro));
+  for (it->SeekToFirst(); it->Valid(); it->Next()) {
+    if (std::string key = it->key().ToString(); gid == utils::ExtractGidFromLabelIndexStorage(key)) {
+      if (!disk_transaction->Delete(key).ok()) {
+        return false;
```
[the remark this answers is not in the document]
as51340: You mean why not log an error message? My reasoning was that it is enough to catch it in the storage and then throw serialization error

src/storage/v2/replication/replication_client.cpp
```diff
@@ -531,65 +533,71 @@ Storage::TimestampInfo Storage::ReplicationClient::GetTimestampInfo() {
...
+                                                                             const size_t buffer_size) {
   replication::Encoder encoder(stream_.GetBuilder());
   encoder.WriteBuffer(buffer, buffer_size);
 }
 
-replication::CurrentWalRes Storage::ReplicationClient::CurrentWalHandler::Finalize() { return stream_.AwaitResponse(); }
+replication::CurrentWalRes InMemoryStorage::ReplicationClient::CurrentWalHandler::Finalize() {
+  return stream_.AwaitResponse();
+}
 }  // namespace memgraph::storage
```
andrejtonev, line 603: should we move replication to inmemory or will there be replication for ondisk in the future?

andrejtonev (commented):
okay, just a couple of additional comments

src/storage/v2/inmemory/label_index.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+#include "storage/v2/inmemory/label_index.hpp"
+#include "storage/v2/inmemory/indices_utils.hpp"
+
+namespace memgraph::storage {
+
+InMemoryLabelIndex::InMemoryLabelIndex(Indices *indices, Constraints *constraints, Config config)
+    : LabelIndex(indices, constraints, config) {}
+
+void InMemoryLabelIndex::UpdateOnAddLabel(LabelId added_label, Vertex *vertex_after_update, const Transaction &tx) {
```
antoniofilipovic, line 20: Clang is complaining to me here that this function has different param names

src/storage/v2/inmemory/storage.hpp
```diff
@@ -0,0 +1,469 @@
...
+
+    ~InMemoryAccessor() override;
+
+    /// @throw std::bad_alloc
+    VertexAccessor CreateVertex() override;
+
+    std::optional<VertexAccessor> FindVertex(Gid gid, View view) override;
+
+    VerticesIterable Vertices(View view) override {
+      auto *mem_storage = static_cast<InMemoryStorage *>(storage_);
```
antoniofilipovic, line 89: why do we need static_cast here? Should proper call be resolved by function call?

src/query/interpreter.cpp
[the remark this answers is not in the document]
andrejtonev: But here we are checking to see if we are the only active session, right?
We check if there is another interpreter and throw if we aren't the only one.
But there is a case were another session could be created between our check and the storage switch (I think).

src/query/interpreter.cpp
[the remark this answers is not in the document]
andrejtonev: When we configure memgraph we are always passing both configurations?

src/query/interpreter.cpp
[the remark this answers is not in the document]
andrejtonev: Should we switch storage modes while in an explicit transaction?

src/query/interpreter.cpp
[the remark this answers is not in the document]
andrejtonev: Yup

src/storage/v2/inmemory/storage.hpp
```diff
@@ -0,0 +1,469 @@
...
+    bool LabelIndexExists(LabelId label) const override {
+      return static_cast<InMemoryStorage *>(storage_)->indices_.label_index_->IndexExists(label);
+    }
+
+    bool LabelPropertyIndexExists(LabelId label, PropertyId property) const override {
+      return static_cast<InMemoryStorage *>(storage_)->indices_.label_property_index_->IndexExists(label, property);
+    }
+
+    IndicesInfo ListAllIndices() const override {
+      auto *mem_storage = static_cast<InMemoryStorage *>(storage_);
```
antoniofilipovic, line 199: const missing here

src/storage/v2/inmemory/storage.hpp
```diff
@@ -0,0 +1,469 @@
...
+      return mem_storage->ListAllIndices();
+    }
+
+    ConstraintsInfo ListAllConstraints() const override {
+      auto *mem_storage = static_cast<InMemoryStorage *>(storage_);
```
antoniofilipovic, line 204: and here

src/storage/v2/disk/label_index.cpp
```diff
@@ -0,0 +1,261 @@
...
+/// TODO: this can probably be optimized
+bool DiskLabelIndex::ClearDeletedVertex(std::string_view gid, uint64_t transaction_commit_timestamp) const {
+  auto disk_transaction = std::unique_ptr<rocksdb::Transaction>(
+      kvstore_->db_->BeginTransaction(rocksdb::WriteOptions(), rocksdb::TransactionOptions()));
+  disk_transaction->SetReadTimestampForValidation(std::numeric_limits<uint64_t>::max());
+
+  rocksdb::ReadOptions ro;
+  std::string strTs = utils::StringTimestamp(std::numeric_limits<uint64_t>::max());
+  rocksdb::Slice ts(strTs);
+  ro.timestamp = &ts;
```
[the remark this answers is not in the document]
as51340: yes but how when we are passing address of timestamp not value?

src/storage/v2/disk/label_property_index.hpp
```diff
@@ -0,0 +1,84 @@
...
+                                                   const storage::PropertyId &property) const override;
+
+  void Clear() override;
+
+  RocksDBStorage *GetRocksDBStorage() const;
+
+ private:
+  utils::Synchronized<std::map<uint64_t, std::map<Gid, std::vector<std::pair<LabelId, PropertyId>>>>>
+      entries_for_deletion;
+  std::set<std::pair<LabelId, PropertyId>> index_;
```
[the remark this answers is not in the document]
as51340: Same as above

src/storage/v2/disk/rocksdb_storage.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+int ComparatorWithU64TsImpl::Compare(const rocksdb::Slice &a, const rocksdb::Slice &b) const {
+  int ret = CompareWithoutTimestamp(a, b);
+  if (ret != 0) {
+    return ret;
+  }
+  // Compare timestamp.
+  // For the same user key with different timestamps, larger (newer) timestamp
+  // comes first.
+  return -CompareTimestamp(ExtractTimestampFromUserKey(a), ExtractTimestampFromUserKey(b));
```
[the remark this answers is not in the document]
as51340: Since it is used only at one place, I will just swap b and a.

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1448 @@
...
+#include "utils/stat.hpp"
+#include "utils/string.hpp"
+
+namespace memgraph::storage {
+
+using OOMExceptionEnabler = utils::MemoryTracker::OutOfMemoryExceptionEnabler;
+
+namespace {
+
+constexpr const char *vertexHandle = "vertex";
```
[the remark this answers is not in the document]
as51340: clang says static is not needed in anonymous namespace

src/storage/v2/disk/storage.hpp
as51340: Here not because this is data for single transaction

src/storage/v2/disk/storage.hpp
as51340: The problem is that make_unique then calls private constructor and compiler says I cannot do that:
In template: calling a private constructor of class 'memgraph::storage::DiskStorage::DiskAccessor'clang(access_ctor)

src/storage/v2/durability/durability.cpp
as51340: Here yes because this durability is used only for in-memory storage

src/storage/v2/replication/replication_client.cpp
as51340: good question, at some point I think buda wants to integrate it

src/storage/v2/disk/label_index.cpp
```diff
@@ -0,0 +1,209 @@
...
+      }
+    }
+  }
+  return true;
+}
+
+/// TODO: duplication with label_property_index.cpp
+bool CommitWithTimestamp(rocksdb::Transaction *disk_transaction, uint64_t commit_ts) {
+  disk_transaction->SetCommitTimestamp(commit_ts);
+  auto status = disk_transaction->Commit();
```
antoniofilipovic, line 39: const auto

src/storage/v2/disk/label_index.cpp
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+  for (const LabelId index_label : index_) {
+    if (utils::Contains(vertex.labels, index_label)) {
+      if (!disk_transaction
+               ->Put(utils::SerializeVertexAsKeyForLabelIndex(index_label, vertex.gid),
+                     utils::SerializeVertexAsValueForLabelIndex(index_label, vertex.labels, vertex.properties))
+               .ok()) {
+        return false;
+      }
```
antoniofilipovic, line 97:
```suggestion
    if (!utils::Contains(vertex.labels, index_label)) {
        continue
    }
    if (!disk_transaction
            ->Put(utils::SerializeVertexAsKeyForLabelIndex(index_label, vertex.gid),
                    utils::SerializeVertexAsValueForLabelIndex(index_label, vertex.labels, vertex.properties))
            .ok()) {
        return false;
    }
```

src/storage/v2/disk/label_index.cpp
```diff
@@ -0,0 +1,209 @@
...
+}
+
+/// TODO: this can probably be optimized
+bool DiskLabelIndex::ClearDeletedVertex(std::string_view gid, uint64_t transaction_commit_timestamp) const {
+  auto disk_transaction = CreateAllReadingRocksDBTransaction();
+
+  rocksdb::ReadOptions ro;
+  std::string strTs = utils::StringTimestamp(std::numeric_limits<uint64_t>::max());
+  rocksdb::Slice ts(strTs);
+  ro.timestamp = &ts;
```
antoniofilipovic, line 111: You are using this on few places, maybe put this in function? Also, can you explain logic, why do you set `strTs` to `std::numeric_limits<uint64_t>::max()`?

src/storage/v2/disk/label_index.cpp
```diff
@@ -0,0 +1,209 @@
...
+  bool deletion_success = true;
+  entries_for_deletion.WithLock([&deletion_success, transaction_start_timestamp,
+                                 disk_transaction_ptr = disk_transaction.get()](auto &tx_to_entries_for_deletion) {
+    if (auto tx_it = tx_to_entries_for_deletion.find(transaction_start_timestamp);
+        tx_it != tx_to_entries_for_deletion.end()) {
```
antoniofilipovic, line 136: you can use `contains` here I think

src/storage/v2/disk/label_index.cpp
```diff
@@ -0,0 +1,209 @@
...
+      deletion_success = ClearTransactionEntriesWithRemovedIndexingLabel(*disk_transaction_ptr, tx_it->second);
+      tx_to_entries_for_deletion.erase(tx_it);
+    }
+  });
```
antoniofilipovic, line 140: I think you don't need to capture deleteion_success in lambda

```
bool deletion_success = entries_for_deletion.WithLock([transaction_start_timestamp,
                                 disk_transaction_ptr = disk_transaction.get()](auto &tx_to_entries_for_deletion) {
    if (auto tx_it = tx_to_entries_for_deletion.find(transaction_start_timestamp);
        tx_it != tx_to_entries_for_deletion.end()) {
        tx_to_entries_for_deletion.erase(tx_it);
      return ClearTransactionEntriesWithRemovedIndexingLabel(*disk_transaction_ptr, tx_it->second);
    }
    return false;
  });
``` 

you can do something like this so that lambda returns

src/storage/v2/disk/label_index.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+void DiskLabelIndex::UpdateOnAddLabel(LabelId added_label, Vertex *vertex_before_update, const Transaction &tx) {
+  entries_for_deletion.WithLock([added_label, vertex_before_update, &tx](auto &tx_to_entries_for_deletion) {
+    if (auto tx_it = tx_to_entries_for_deletion.find(tx.start_timestamp); tx_it != tx_to_entries_for_deletion.end()) {
+      if (auto vertex_label_index_it = tx_it->second.find(vertex_before_update->gid);
+          vertex_label_index_it != tx_it->second.end()) {
+        std::erase_if(vertex_label_index_it->second,
+                      [added_label](const LabelId &indexed_label) { return indexed_label == added_label; });
+      }
+    }
```
antoniofilipovic, line 155:
```suggestion
  auto tx_it = tx_to_entries_for_deletion.find(tx.start_timestamp);
    if (tx_it == tx_to_entries_for_deletion.end()) {
      return;
    }
    auto vertex_label_index_it = tx_it->second.find(vertex_before_update->gid);
    if (vertex_label_index_it != tx_it->second.end()) {
      return;
    }
    std::erase_if(vertex_label_index_it->second,
                  [added_label](const LabelId &indexed_label) { return indexed_label == added_label; });
      
```

I think this is a bit cleaner, as it has early returns where possible, and thus slightly less indents

src/storage/v2/disk/label_index.cpp
```diff
@@ -0,0 +1,209 @@
...
+  });
+}
+
+void DiskLabelIndex::UpdateOnRemoveLabel(LabelId removed_label, Vertex *vertex_before_update, const Transaction &tx) {
+  if (IndexExists(removed_label)) {
```
antoniofilipovic, line 160: use early return here if index doesn't exist

src/storage/v2/disk/label_index.cpp
```diff
@@ -0,0 +1,209 @@
...
+bool DiskLabelIndex::DropIndex(LabelId label) {
+  if (!(index_.erase(label) > 0)) {
+    return false;
+  }
+  auto disk_transaction = CreateAllReadingRocksDBTransaction();
+
+  rocksdb::ReadOptions ro;
+  std::string strTs = utils::StringTimestamp(std::numeric_limits<uint64_t>::max());
+  rocksdb::Slice ts(strTs);
+  ro.timestamp = &ts;
```
antoniofilipovic, line 182: This would be nice to make a one function

src/storage/v2/disk/label_property_index.cpp
```diff
@@ -0,0 +1,272 @@
...
+bool IsVertexIndexedByLabelProperty(const Vertex &vertex, LabelId label, PropertyId property) {
+  return utils::Contains(vertex.labels, label) && vertex.properties.HasProperty(property);
+}
+
+[[nodiscard]] bool ClearTransactionEntriesWithRemovedIndexingLabel(
+    rocksdb::Transaction &disk_transaction,
+    const std::map<Gid, std::vector<std::pair<LabelId, PropertyId>>> &transaction_entries) {
+  for (const auto &[vertex_gid, index] : transaction_entries) {
+    for (const auto &[indexing_label, indexing_property] : index) {
+      if (auto status = disk_transaction.Delete(
```
antoniofilipovic, line 36: I think you can apply here similar changes as in label_index.cpp throughout the file

src/storage/v2/disk/label_property_index.cpp
```diff
@@ -0,0 +1,272 @@
...
+  return CommitWithTimestamp(disk_transaction.get(), commit_timestamp);
+}
+
+bool DiskLabelPropertyIndex::ClearDeletedVertex(std::string_view gid, uint64_t transaction_commit_timestamp) const {
+  auto disk_transaction = CreateAllReadingRocksDBTransaction();
+
+  rocksdb::ReadOptions ro;
+  std::string strTs = utils::StringTimestamp(std::numeric_limits<uint64_t>::max());
+  rocksdb::Slice ts(strTs);
+  ro.timestamp = &ts;
```
antoniofilipovic, line 120: similar to label_index.cpp comment

src/storage/v2/disk/label_property_index.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+bool DiskLabelPropertyIndex::DeleteVerticesWithRemovedIndexingLabel(uint64_t transaction_start_timestamp,
+                                                                    uint64_t transaction_commit_timestamp) {
+  auto disk_transaction = CreateAllReadingRocksDBTransaction();
+
+  rocksdb::ReadOptions ro;
+  std::string strTs = utils::StringTimestamp(std::numeric_limits<uint64_t>::max());
+  rocksdb::Slice ts(strTs);
+  ro.timestamp = &ts;
+  bool deletion_success = true;
```
antoniofilipovic, line 140: similar to label_index.cpp comment

src/storage/v2/disk/label_property_index.cpp
```diff
@@ -0,0 +1,272 @@
...
+}
+
+void DiskLabelPropertyIndex::UpdateOnAddLabel(LabelId added_label, Vertex *vertex_after_update, const Transaction &tx) {
+  entries_for_deletion.WithLock([added_label, vertex_after_update, &tx](auto &tx_to_entries_for_deletion) {
+    if (auto tx_it = tx_to_entries_for_deletion.find(tx.start_timestamp); tx_it != tx_to_entries_for_deletion.end()) {
+      if (auto vertex_label_index_it = tx_it->second.find(vertex_after_update->gid);
+          vertex_label_index_it != tx_it->second.end()) {
+        std::erase_if(vertex_label_index_it->second, [added_label](const std::pair<LabelId, PropertyId> &index) {
+          return index.first == added_label;
+        });
```
antoniofilipovic, line 162: similar to label_index.cpp comment

src/storage/v2/disk/label_property_index.cpp
```diff
@@ -0,0 +1,272 @@
...
+bool DiskLabelPropertyIndex::DropIndex(LabelId label, PropertyId property) {
+  return index_.erase({label, property}) > 0;
+}
+
+bool DiskLabelPropertyIndex::IndexExists(LabelId label, PropertyId property) const {
+  return utils::Contains(index_, std::make_pair(label, property));
+}
+
+std::vector<std::pair<LabelId, PropertyId>> DiskLabelPropertyIndex::ListIndices() const {
+  return std::vector<std::pair<LabelId, PropertyId>>(index_.begin(), index_.end());
```
antoniofilipovic, line 198:
```suggestion
  return {index_.begin(), index_.end()};
```

src/storage/v2/disk/label_property_index.cpp
```diff
@@ -0,0 +1,272 @@
...
+// contained in a `PropertyValue`. Note that numbers (integers and doubles) are
+// treated as the same "type" in `PropertyValue`.
+const PropertyValue kSmallestBool = PropertyValue(false);
+static_assert(-std::numeric_limits<double>::infinity() < std::numeric_limits<int64_t>::min());
+const PropertyValue kSmallestNumber = PropertyValue(-std::numeric_limits<double>::infinity());
+const PropertyValue kSmallestString = PropertyValue("");
+const PropertyValue kSmallestList = PropertyValue(std::vector<PropertyValue>());
+const PropertyValue kSmallestMap = PropertyValue(std::map<std::string, PropertyValue>());
+const PropertyValue kSmallestTemporalData =
+    PropertyValue(TemporalData{static_cast<TemporalType>(0), std::numeric_limits<int64_t>::min()});
```
antoniofilipovic, line 211: I would change this from copy ctor to regular ctor using braced init

```suggestion
const PropertyValue kSmallestNumber{-std::numeric_limits<double>::infinity()};
const PropertyValue kSmallestString{""};
const PropertyValue kSmallestList{std::vector<PropertyValue>()};
const PropertyValue kSmallestMap{std::map<std::string, PropertyValue>()};
const PropertyValue kSmallestTemporalData{TemporalData{static_cast<TemporalType>(0), std::numeric_limits<int64_t>::min()}};
```

src/storage/v2/disk/label_property_index.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+uint64_t DiskLabelPropertyIndex::ApproximateVertexCount(LabelId label, PropertyId property) const { return 10; }
```
antoniofilipovic, line 213: unused property label and property. Can you explain why 10 is here?

src/storage/v2/disk/label_property_index.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+uint64_t DiskLabelPropertyIndex::ApproximateVertexCount(LabelId label, PropertyId property,
+                                                        const PropertyValue &value) const {
+  return 10;
```
antoniofilipovic, line 217: Why is 10 here as returned type?

src/storage/v2/disk/label_property_index.cpp
```diff
@@ -0,0 +1,272 @@
...
+}
+
+uint64_t DiskLabelPropertyIndex::ApproximateVertexCount(LabelId label, PropertyId property,
+                                                        const std::optional<utils::Bound<PropertyValue>> &lower,
+                                                        const std::optional<utils::Bound<PropertyValue>> &upper) const {
+  return 10;
```
antoniofilipovic, line 223: And why is 10 here as return value

src/storage/v2/disk/label_property_index.cpp
```diff
@@ -0,0 +1,272 @@
...
+    return it->second;
+  }
+  return {};
+}
+
+void DiskLabelPropertyIndex::LoadIndexInfo(const std::vector<std::string> &keys) {
+  for (const auto &label_property : keys) {
+    std::vector<std::string> label_property_split = utils::Split(label_property, ",");
+    index_.emplace(std::make_pair(LabelId::FromUint(std::stoull(label_property_split[0])),
+                                  PropertyId::FromUint(std::stoull(label_property_split[1]))));
```
antoniofilipovic, line 266: Do we need to add here MG_ASSERT or something? How do we know that it is always list of 2 and that converting to long long will work?

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1601 @@
...
+}
+
+}  // namespace
+
+void DiskStorage::LoadTimestampIfExists() {
+  if (utils::DirExists(config_.disk.durability_directory)) {
+    if (auto last_timestamp_ = durability_kvstore_->Get(lastTransactionStartTimeStamp); last_timestamp_.has_value()) {
+      timestamp_ = std::stoull(last_timestamp_.value());
+    }
+  }
```
antoniofilipovic, line 184: you can use early return here too:
```
if (!condition){
    return;
}
...
```

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1601 @@
...
+    std::vector<std::string> labels = utils::Split(label_index.value(), "|");
+    disk_label_index->LoadIndexInfo(labels);
+  }
+}
+
+void DiskStorage::LoadLabelPropertyIndexInfoIfExists() const {
+  if (auto label_property_index = durability_kvstore_->Get(label_property_index_str);
+      label_property_index.has_value()) {
+    auto *disk_label_property_index = static_cast<DiskLabelPropertyIndex *>(indices_.label_property_index_.get());
+    std::vector<std::string> keys = utils::Split(label_property_index.value(), "|");
```
antoniofilipovic, line 206:
```suggestion
    std::vector<std::string> keys{utils::Split(label_property_index.value(), "|")};
```

I think this should be better as no copy assignment is included.

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1601 @@
...
+  if (utils::DirExists(config_.disk.durability_directory)) {
+    LoadLabelIndexInfoIfExists();
+    LoadLabelPropertyIndexInfoIfExists();
+  }
+}
+
+void DiskStorage::LoadLabelIndexInfoIfExists() const {
+  if (auto label_index = durability_kvstore_->Get(label_index_str); label_index.has_value()) {
+    auto *disk_label_index = static_cast<DiskLabelIndex *>(indices_.label_index_.get());
+    std::vector<std::string> labels = utils::Split(label_index.value(), "|");
```
antoniofilipovic, line 197:
```suggestion
    std::vector<std::string> labels{utils::Split(label_index.value(), "|")};
```
I think this is also better without copy assignment

src/storage/v2/disk/storage.cpp
antoniofilipovic:
```suggestion
    std::vector<std::string> keys{utils::Split(label_property_index.value(), "|")};
```

I think this should be better as no copy assignment is included.

antoniofilipovic: Also, add const.

src/storage/v2/disk/rocksdb_storage.hpp
```diff
@@ -0,0 +1,84 @@
...
+#include "storage/v2/property_store.hpp"
+#include "utils/logging.hpp"
+
+namespace memgraph::storage {
+
+/// TODO: this should be somehow more wrapped inside the storage class so from the software engineering perspective
+/// it isn't great to have this here. But for now it is ok.
+/// Wraps RocksDB objects inside a struct. Vertex_chandle and edge_chandle are column family handles that may be
+/// nullptr. In that case client should take care about them.
+struct RocksDBStorage {
```
antoniofilipovic, line 31: Define copy ctor, copy assignment, move ctor and move assignment

antoniofilipovic: as move ctor and move assignment are deleted, and others are defaulted, and I don't know what do we want to permit

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1601 @@
...
+    const rocksdb::Slice &key, const rocksdb::Slice &value, Delta *index_delta) {
+  auto index_accessor = indexed_vertices_.access();
+
+  storage::Gid gid = Gid::FromUint(std::stoull(utils::ExtractGidFromLabelIndexStorage(key.ToString())));
+  if (VertexExistsInCache(index_accessor, gid)) {
+    return std::nullopt;
+  }
+
+  const std::string value_str = value.ToString();
+  std::vector<LabelId> labels = utils::DeserializeLabelsFromLabelIndexStorage(value_str);
```
antoniofilipovic, line 331:
```suggestion
  std::vector<LabelId> labels{utils::DeserializeLabelsFromLabelIndexStorage(value_str)};
```

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1601 @@
...
+    const rocksdb::Slice &key, const rocksdb::Slice &value) {
+  auto index_accessor = indexed_vertices_.access();
+
+  storage::Gid gid = Gid::FromUint(std::stoull(utils::ExtractGidFromLabelPropertyIndexStorage(key.ToString())));
+  if (VertexExistsInCache(index_accessor, gid)) {
+    return std::nullopt;
+  }
+
+  const std::string value_str = value.ToString();
+  std::vector<LabelId> labels = utils::DeserializeLabelsFromLabelPropertyIndexStorage(value_str);
```
antoniofilipovic, line 346: Similar comment to above.

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1601 @@
...
+  const auto edge_parts = utils::Split(key.ToStringView(), "|");
+  const Gid edge_gid = Gid::FromUint(std::stoull(edge_parts[4]));
+
+  auto edge_acc = edges_.access();
+  auto res = edge_acc.find(edge_gid);
+  if (res != edge_acc.end()) {
+    return std::nullopt;
+  }
+
+  auto [from_gid, to_gid] = std::invoke(
```
antoniofilipovic, line 363: Instead of returning strings, maybe you can return index of out id, and index of in edge gid, and then just with these indices reference values in list

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1601 @@
...
+      [&](const auto &edge_parts) {
+        if (edge_parts[2] == "0") {  // out edge
+          return std::make_pair(edge_parts[0], edge_parts[1]);
+        }
+        // in edge
+        return std::make_pair(edge_parts[1], edge_parts[0]);
+      },
+      edge_parts);
+
+  auto from_acc = FindVertex(Gid::FromUint(std::stoull(from_gid)), View::OLD);
```
antoniofilipovic, line 373: const auto

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1601 @@
...
+  auto to_acc = FindVertex(Gid::FromUint(std::stoull(to_gid)), View::OLD);
```
antoniofilipovic, line 374: const auto

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1601 @@
...
+          // We already inserted this edge into the vertex's out_edges list.
+          return false;
+        }
+      }
+    }
+    return disk_vertex_out_edge_gid == utils::SerializeIdType(vertex_acc.Gid());
+  });
+}
+
+Result<EdgeAccessor> DiskStorage::DiskAccessor::CreateEdge(VertexAccessor *from, VertexAccessor *to,
```
antoniofilipovic, line 879: I think this function should accept `const VertexAccessor *from`. And similar to `const VertexAccessor *to`

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1601 @@
...
+
+  return *maybe_edge;
+}
+
+VerticesIterable DiskStorage::DiskAccessor::Vertices(View view) {
+  auto *disk_storage = static_cast<DiskStorage *>(storage_);
+  rocksdb::ReadOptions ro;
+  std::string strTs = utils::StringTimestamp(transaction_.start_timestamp);
+  rocksdb::Slice ts(strTs);
+  ro.timestamp = &ts;
```
antoniofilipovic, line 390: Again, I think we can isolate this in function

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1601 @@
...
+  }
+  ApplyDeltasForRead(transaction, delta, view, [&deleted, &has_label, label](const Delta &delta) {
+    switch (delta.action) {
+      case Delta::Action::REMOVE_LABEL: {
+        if (delta.label == label) {
+          MG_ASSERT(has_label, "Invalid database state!");
+          has_label = false;
+        }
+        break;
+      }
```
antoniofilipovic, line 90: Todo: we need to check this if it is correct behaviour: https://memgraph.slack.com/archives/CS5827FK3/p1687530832459949

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1601 @@
...
+                                                           const std::string &old_disk_key) {
+  OOMExceptionEnabler oom_exception;
+  MG_ASSERT(from->transaction_ == to->transaction_,
+            "VertexAccessors must be from the same transaction when creating "
+            "an edge!");
+  MG_ASSERT(from->transaction_ == &transaction_,
+            "VertexAccessors must be from the same transaction in when "
+            "creating an edge!");
+
+  auto from_vertex = from->vertex_;
```
antoniofilipovic, line 891: `const auto *`

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1601 @@
...
+  auto to_vertex = to->vertex_;
```
antoniofilipovic, line 892: `const auto *`

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1601 @@
...
+Result<EdgeAccessor> DiskStorage::DiskAccessor::CreateEdge(VertexAccessor *from, VertexAccessor *to,
+                                                           EdgeTypeId edge_type) {
+  MG_ASSERT(from->transaction_ == &transaction_,
+            "VertexAccessor must be from the same transaction as the storage "
+            "accessor when deleting a vertex!");
+  MG_ASSERT(to->transaction_ == &transaction_,
+            "VertexAccessor must be from the same transaction as the storage "
+            "accessor when deleting a vertex!");
+
+  auto from_vertex = from->vertex_;
```
antoniofilipovic, line 952: `const auto *`

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1601 @@
...
+  return EdgeAccessor(edge, edge_type, from_vertex, to_vertex, &transaction_, &storage_->indices_,
+                      &storage_->constraints_, config_);
+}
+
+Result<std::optional<EdgeAccessor>> DiskStorage::DiskAccessor::DeleteEdge(EdgeAccessor *edge) {
+  MG_ASSERT(edge->transaction_ == &transaction_,
+            "EdgeAccessor must be from the same transaction as the storage "
+            "accessor when deleting an edge!");
+  auto edge_ref = edge->edge_;
+  auto edge_type = edge->edge_type_;
```
antoniofilipovic, line 1008: add const

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1601 @@
...
+
+  std::unique_lock<utils::SpinLock> guard;
+  if (config_.properties_on_edges) {
+    auto *edge_ptr = edge_ref.ptr;
```
antoniofilipovic, line 1012: add const

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1601 @@
...
+  if (!PrepareForWrite(&transaction_, from_vertex)) return Error::SERIALIZATION_ERROR;
+  MG_ASSERT(!from_vertex->deleted, "Invalid database state!");
+
+  if (to_vertex != from_vertex) {
+    if (!PrepareForWrite(&transaction_, to_vertex)) return Error::SERIALIZATION_ERROR;
+    MG_ASSERT(!to_vertex->deleted, "Invalid database state!");
+  }
+
+  auto delete_edge_from_storage = [&edge_type, &edge_ref, this](auto *vertex, auto *edges) {
+    std::tuple<EdgeTypeId, Vertex *, EdgeRef> link(edge_type, vertex, edge_ref);
```
antoniofilipovic, line 1046: add const

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1601 @@
...
+    }
+    std::swap(*it, *edges->rbegin());
+    edges->pop_back();
+    return true;
+  };
+
+  auto op1 = delete_edge_from_storage(to_vertex, &from_vertex->out_edges);
+  auto op2 = delete_edge_from_storage(from_vertex, &to_vertex->in_edges);
+  auto src_dest_del_key =
+      utils::SerializeEdge(from_vertex->gid, to_vertex->gid, edge_type, edge_ref, config_.properties_on_edges);
```
antoniofilipovic, line 1061: add const

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1601 @@
...
+  if (status.ok()) {
+    spdlog::debug("rocksdb: Saved edge with key {} and ts {}", serializedEdgeKey, *commit_timestamp_);
+  } else if (status.IsBusy()) {
+    spdlog::error("rocksdb: Edge with key {} and ts {} was changed and committed in another transaction",
+                  serializedEdgeKey, *commit_timestamp_);
+    return false;
+  } else {
+    spdlog::error("rocksdb: Failed to save edge with key {} and ts {}", serializedEdgeKey, *commit_timestamp_);
+    return false;
+  }
```
antoniofilipovic, line 1128: I think you can wrap this in function

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1601 @@
...
+
+    if (!WriteVertexToDisk(vertex)) {
+      return StorageDataManipulationError{SerializationError{}};
+    }
+
+    /// TODO: andi don't ignore the return value
+    disk_unique_constraints->SyncVertexToUniqueConstraintsStorage(vertex, *commit_timestamp_);
+    disk_label_index->SyncVertexToLabelIndexStorage(vertex, *commit_timestamp_);
+    disk_label_property_index->SyncVertexToLabelPropertyIndexStorage(vertex, *commit_timestamp_);
+
```
antoniofilipovic, line 1213: comment just to not forget to fix nodiscard errors

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1601 @@
...
+  auto *disk_label_index = static_cast<DiskLabelIndex *>(storage_->indices_.label_index_.get());
+  auto *disk_label_property_index =
+      static_cast<DiskLabelPropertyIndex *>(storage_->indices_.label_property_index_.get());
+
+  /// TODO: andi I don't like that std::optional is used for checking errors but that's how it was before, refactor!
+  for (Vertex &vertex : vertex_acc) {
+    /// TODO: refactor this check, it is unreadable
+    if (auto existence_constraint_validation_result = storage_->constraints_.existence_constraints_->Validate(vertex);
+        existence_constraint_validation_result.has_value()) {
+      return StorageDataManipulationError{existence_constraint_validation_result.value()};
```
antoniofilipovic, line 1180: Maybe I don't understand something, but here we are checking vertices one by one, and if there are not constraint violations we are storing them to rocks db. But what if constraint violation happens later on some vertex and we stored already a lot of them?

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1601 @@
...
+  }
+  // if (auto status = disk_storage->kvstore_->db_->Flush(rocksdb::FlushOptions{},
+  // disk_storage->kvstore_->vertex_chandle); !status.ok()) {
+  //   spdlog::error("rocksdb: Flush failed with status {}", status.ToString());
+  //   return StorageDataManipulationError{SerializationError{}};
+  // }
+  // if (auto status = disk_storage->kvstore_->db_->SyncWAL(); !status.ok()) {
+  //   spdlog::error("rocksdb: SyncWAL failed with status {}", status.ToString());
+  //   return StorageDataManipulationError{SerializationError{}};
+  // }
```
antoniofilipovic, line 1356: reminder to remove this commented part

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1601 @@
...
+DiskStorage::CheckExistingVerticesBeforeCreatingUniqueConstraint(LabelId label,
+                                                                 const std::set<PropertyId> &properties) const {
+  std::set<std::vector<PropertyValue>> unique_storage;
+  std::vector<std::pair<std::string, std::string>> vertices_for_constraints;
+
+  rocksdb::ReadOptions ro;
+  std::string strTs = utils::StringTimestamp(std::numeric_limits<uint64_t>::max());
+  rocksdb::Slice ts(strTs);
+  ro.timestamp = &ts;
+  auto it = std::unique_ptr<rocksdb::Iterator>(kvstore_->db_->NewIterator(ro, kvstore_->vertex_chandle));
```
antoniofilipovic, line 1296: isolate this part in function if possible

src/storage/v2/inmemory/label_index.cpp
```diff
@@ -0,0 +1,158 @@
...
+  auto it = index_.find(added_label);
+  if (it == index_.end()) return;
+  auto acc = it->second.access();
+  acc.insert(Entry{vertex_after_update, tx.start_timestamp});
+}
+
+bool InMemoryLabelIndex::CreateIndex(LabelId label, utils::SkipList<Vertex>::Accessor vertices,
+                                     const std::optional<ParalellizedIndexCreationInfo> &paralell_exec_info) {
+  auto create_index_seq = [this](LabelId label, utils::SkipList<Vertex>::Accessor &vertices,
```
antoniofilipovic, line 29: `const auto`

src/storage/v2/inmemory/label_index.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+    CreateIndexOnSingleThread(vertices, it, index_, label,
+                              [](Vertex &vertex, LabelId label, IndexAccessor &index_accessor) {
+                                TryInsertLabelIndex(vertex, label, index_accessor);
+                              });
+
+    return true;
+  };
+
+  auto create_index_par = [this](LabelId label, utils::SkipList<Vertex>::Accessor &vertices,
```
antoniofilipovic, line 41: `const auto`

src/storage/v2/inmemory/label_index.cpp
```diff
@@ -0,0 +1,158 @@
...
+}
+
+void InMemoryLabelIndex::RunGC() {
+  for (auto &index_entry : index_) {
+    index_entry.second.run_gc();
+  }
+}
+
+InMemoryLabelIndex::Iterable InMemoryLabelIndex::Vertices(LabelId label, View view, Transaction *transaction) {
+  auto it = index_.find(label);
```
antoniofilipovic, line 153: const auto

src/storage/v2/inmemory/indices_utils.hpp
```diff
@@ -0,0 +1,357 @@
...
+    std::pair<std::vector<std::pair<Gid, uint64_t>> /*vertex_recovery_info*/, uint64_t /*thread_count*/>;
+
+/// Traverses deltas visible from transaction with start timestamp greater than
+/// the provided timestamp, and calls the provided callback function for each
+/// delta. If the callback ever returns true, traversal is stopped and the
+/// function returns true. Otherwise, the function returns false.
+template <typename TCallback>
+inline bool AnyVersionSatisfiesPredicate(uint64_t timestamp, const Delta *delta, const TCallback &predicate) {
+  while (delta != nullptr) {
+    auto ts = delta->timestamp->load(std::memory_order_acquire);
```
antoniofilipovic, line 32: `const auto`

src/storage/v2/inmemory/indices_utils.hpp
```diff
@@ -0,0 +1,357 @@
...
+    delta = delta->next.load(std::memory_order_acquire);
+  }
+  return false;
+}
+
+/// Helper function for label index garbage collection. Returns true if there's
+/// a reachable version of the vertex that has the given label.
+inline bool AnyVersionHasLabel(const Vertex &vertex, LabelId label, uint64_t timestamp) {
+  bool has_label = false;
+  bool deleted = false;
```
antoniofilipovic, line 50:
```suggestion
  bool has_label{false};
  bool deleted{false};
```

src/storage/v2/inmemory/indices_utils.hpp
```diff
@@ -0,0 +1,357 @@
...
+}
+
+/// Helper function for label-property index garbage collection. Returns true if
+/// there's a reachable version of the vertex that has the given label and
+/// property value.
+inline bool AnyVersionHasLabelProperty(const Vertex &vertex, LabelId label, PropertyId key, const PropertyValue &value,
+                                       uint64_t timestamp) {
+  bool has_label = false;
+  bool current_value_equal_to_value = value.IsNull();
+  bool deleted = false;
```
antoniofilipovic, line 104:
```suggestion
  bool has_label{false};
  bool current_value_equal_to_value{value.IsNull()};
  bool deleted{false};
```

src/storage/v2/property_store.cpp
```diff
@@ -984,6 +987,46 @@ bool PropertyStore::HasProperty(PropertyId property) const {
...
+bool PropertyStore::HasAllPropertyValues(const std::vector<PropertyValue> &property_values) const {
+  /// TODO: andi extract this into a private method
+  auto property_map = Properties();
+  std::stringstream ss;
+  ss << "{";
+  for (const auto &[property_id, property_value] : property_map) {
+    ss << property_id.AsUint() << ": " << property_value;
+  }
+  ss << "}";
+  spdlog::debug("PropertyStore::HasAllPropertyValues: {}", ss.str());
```
antoniofilipovic, line 1005: This was used for debugging but we don't need it I guess

src/storage/v2/property_store.cpp
```diff
@@ -984,6 +987,46 @@ bool PropertyStore::HasProperty(PropertyId property) const {
...
+  std::vector<PropertyValue> value_array;
+  value_array.reserve(properties.size());
+  for (const auto &prop : properties) {
+    auto value = GetProperty(prop);
+    if (value.IsNull()) {
+      return std::nullopt;
+    }
+    value_array.emplace_back(std::move(value));
+  }
+  return std::move(value_array);
```
antoniofilipovic, line 1027: rvo should kick in so I think no need for std::move

src/storage/v2/property_value.hpp
```diff
@@ -330,6 +330,8 @@ inline bool operator<(const PropertyValue &first, const PropertyValue &second) n
   }
 }
 
+inline bool operator>(const PropertyValue &first, const PropertyValue &second) noexcept { return second < first; }
```
antoniofilipovic, line 333: clang say this may throw exception and it shouldn't throw. What to do about that?

src/storage/v2/vertex.hpp
```diff
@@ -25,7 +25,8 @@ namespace memgraph::storage {
 
 struct Vertex {
   Vertex(Gid gid, Delta *delta) : gid(gid), deleted(false), delta(delta) {
-    MG_ASSERT(delta == nullptr || delta->action == Delta::Action::DELETE_OBJECT,
+    MG_ASSERT(delta == nullptr || delta->action == Delta::Action::DELETE_OBJECT ||
+                  delta->action == Delta::Action::DELETE_DESERIALIZED_OBJECT,
```
antoniofilipovic, line 29: update MG_ASSERT that it can be also DELETE_DESERIALIZED_OBJECT...

src/memgraph.cpp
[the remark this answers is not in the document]
antoniofilipovic: how this didn't fail any tests, interesting

src/storage/v2/disk/storage.cpp
antoniofilipovic: Reminder on this one

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1649 @@
...
+    /// TODO: couple this condition
+    PropertyStore properties = utils::DeserializePropertiesFromLabelPropertyIndexStorage(it_value_str);
+    auto prop_value = properties.GetProperty(property);
+    if (key_str.starts_with(utils::SerializeIdType(label) + "|" + utils::SerializeIdType(property)) &&
+        !utils::Contains(gids, curr_gid) && IsPropertyValueWithinInterval(prop_value, lower_bound, upper_bound)) {
+      if (prop_value.IsInt()) {
+        spdlog::debug("Added to prop index value from label-property storage: {}", prop_value.ValueInt());
+      } else if (prop_value.IsDouble()) {
+        spdlog::debug("Added to prop index value from label-property storage: {}", prop_value.ValueDouble());
+      }
```
antoniofilipovic, line 597: Do you need these two?

src/storage/v2/disk/storage.cpp
```diff
@@ -0,0 +1,1649 @@
...
+  if (delta) {
+    delta->prev.Set(&*it);
+  }
+  return {&*it, &transaction_, &storage_->indices_, &storage_->constraints_, config_};
+}
+
+std::optional<VertexAccessor> DiskStorage::DiskAccessor::FindVertex(storage::Gid gid, View view) {
+  auto acc = vertices_.access();
+  auto vertex_it = acc.find(gid);
+  if (vertex_it != acc.end()) {
```
antoniofilipovic, line 757: you can use contains here

src/query/plan/operator.cpp
```diff
@@ -828,10 +833,12 @@ bool Expand::ExpandCursor::InitEdges(Frame &frame, ExecutionContext &context) {
         // old_node_value may be Null when using optional matching
         if (!existing_node.IsNull()) {
           ExpectType(self_.common_.node_symbol, existing_node, TypedValue::Type::Vertex);
+          context.db_accessor->PrefetchOutEdges(vertex);
```
antoniofilipovic, line 836: do we need these functions? As it just returns empty?

src/query/plan/operator.cpp
```diff
@@ -1264,6 +1276,7 @@ class STShortestPathCursor : public query::plan::Cursor {
 
       for (const auto &vertex : source_frontier) {
         if (self_.common_.direction != EdgeAtom::Direction::IN) {
+          context.db_accessor->PrefetchOutEdges(vertex);
```
antoniofilipovic, line 1279: do we need this?

antoniofilipovic (approved):

Issue #934, opened 2023-05-14
Implement return of transaction id on RUN / BEGIN


gitbuda:
* With this, client apps, most notably Memgraph Lab, could easily terminate initiated transactions.
* Probably the base way is to add transaction id into the `fields` inside the `SUCCESS` message -> https://neo4j.com/docs/bolt/current/bolt/message/#messages-success
* The transaction info should be returned for both single and multi-query transactions

andrejtonev:
The Bolt protocol doesn't have a transaction id. One session can have a single transaction at a time (https://neo4j.com/docs/python-manual/current/transactions/).
Would the transaction id be a Memgraph specific addition?
Do we even need it, since the client can just keep track of the session and terminate that way?

In addition, we should review the Reset implementation to make sure it is compliant (this gets into the bigger picture of the state machine mirroring).

andrejtonev:
https://github.com/memgraph/memgraph/pull/945

gitbuda:
Figure out do we have to do something else here...

PR #945, opened 2023-05-22
[master < T945] Handle user-defined metadata and expose it with SHOW TRANSACTIONS

andrejtonev:
[master < Task] PR

This feature should allow users to pass user-defined metadata via their clients. This is useful when a user has many transactions and they would want to better differentiate between them. 
Once the metadata is passed to Memgraph, the user can query it by executing SHOW TRANSACTIONS, whose output has a new field "metadata".

### DIFF
--- a/src/communication/bolt/v1/states/handlers.hpp
+++ b/src/communication/bolt/v1/states/handlers.hpp
@@ -73,6 +73,23 @@ inline std::pair<std::string, std::string> ExceptionToErrorMessage(const std::ex
           "should be in database logs."};
 }
 
+namespace helpers {
+
+/** Extracts metadata from the extras field.
+ * NOTE: In order to avoid a copy, the metadata in moved.
+ * TODO: Update if extra field is used for anything else.
+ */
+inline std::map<std::string, Value> ConsumeMetadata(Value &extra) {
+  std::map<std::string, Value> md;
+  auto &md_tv = extra.ValueMap()["tx_metadata"];
+  if (md_tv.IsMap()) {
+    md = std::move(md_tv.ValueMap());
+  }
+  return md;
+}
+
+}  // namespace helpers
+
 namespace details {
 
 template <bool is_pull, typename TSession>
@@ -255,12 +272,6 @@ State HandleRunV4(TSession &session, const State state, const Marker marker) {
     return State::Close;
   }
 
-  std::map<std::string, Value> md{};
-  auto &md_tv = extra.ValueMap()["tx_metadata"];
-  if (md_tv.IsMap()) {
-    md = md_tv.ValueMap();
-  }
-
   if (state != State::Idle) {
     // Client could potentially recover if we move to error state, but there is
     // no legitimate situation in which well working client would end up in this
@@ -275,7 +286,8 @@ State HandleRunV4(TSession &session, const State state, const Marker marker) {
 
   try {
     // Interpret can throw.
-    const auto [header, qid] = session.Interpret(query.ValueString(), params.ValueMap(), md);
+    const auto [header, qid] =
+        session.Interpret(query.ValueString(), params.ValueMap(), helpers::ConsumeMetadata(extra));
     // Convert std::string to Value
     std::vector<Value> vec;
     std::map<std::string, Value> data;
@@ -356,12 +368,6 @@ State HandleBegin(TSession &session, const State state, const Marker marker) {
     return State::Close;
   }
 
-  std::map<std::string, Value> md{};
-  auto &md_tv = extra.ValueMap()["tx_metadata"];
-  if (md_tv.IsMap()) {
-    md = md_tv.ValueMap();
-  }
-
   if (state != State::Idle) {
     spdlog::trace("Unexpected BEGIN command!");
     return State::Close;
@@ -375,7 +381,7 @@ State HandleBegin(TSession &session, const State state, const Marker marker) {
   }
 
   try {
-    session.BeginTransaction(md);
+    session.BeginTransaction(helpers::ConsumeMetadata(extra));
   } catch (const std::exception &e) {
     return HandleFailure(session, e);
   }
--- a/src/communication/bolt/v1/states/handlers.hpp
+++ b/src/communication/bolt/v1/states/handlers.hpp
@@ -23,6 +23,7 @@
 #include "communication/bolt/v1/state.hpp"
 #include "communication/bolt/v1/value.hpp"
 #include "communication/exceptions.hpp"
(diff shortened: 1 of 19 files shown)


src/memgraph.cpp
```diff
@@ -533,16 +533,23 @@ class BoltSession final : public memgraph::communication::bolt::Session<memgraph
   using memgraph::communication::bolt::Session<memgraph::communication::v2::InputStream,
                                                memgraph::communication::v2::OutputStream>::TEncoder;
 
-  void BeginTransaction() override { interpreter_.BeginTransaction(); }
+  void BeginTransaction(const std::map<std::string, memgraph::communication::bolt::Value> &metadata) override {
+    std::map<std::string, memgraph::storage::PropertyValue> metadata_pv;
+    for (const auto &kv : metadata) metadata_pv.emplace(kv.first, memgraph::glue::ToPropertyValue(kv.second));
```
gvolfing, line 538: 1. Did you consider structured bindings instead of `kv`?
2. This is just nitpicking, but please use the curly brackets with loops, even if the compound statement is not required.

src/memgraph.cpp
```diff
@@ -533,16 +533,23 @@ class BoltSession final : public memgraph::communication::bolt::Session<memgraph
...
   void RollbackTransaction() override { interpreter_.RollbackTransaction(); }
 
   std::pair<std::vector<std::string>, std::optional<int>> Interpret(
-      const std::string &query, const std::map<std::string, memgraph::communication::bolt::Value> &params) override {
+      const std::string &query, const std::map<std::string, memgraph::communication::bolt::Value> &params,
+      const std::map<std::string, memgraph::communication::bolt::Value> &metadata) override {
     std::map<std::string, memgraph::storage::PropertyValue> params_pv;
+    std::map<std::string, memgraph::storage::PropertyValue> metadata_pv;
     for (const auto &kv : params) params_pv.emplace(kv.first, memgraph::glue::ToPropertyValue(kv.second));
+    for (const auto &kv : metadata) metadata_pv.emplace(kv.first, memgraph::glue::ToPropertyValue(kv.second));
```
gvolfing, line 552: I know this is not your change, but if you decide to go with the curly brackets + structured bindings please do that here as well to keep the style consistent.

src/query/interpreter.cpp
```diff
@@ -112,6 +112,16 @@ void UpdateTypeCount(const plan::ReadWriteTypeChecker::RWType type) {
   }
 }
 
+template <typename T>
+concept HasEmpty = requires(T t) {
+  { t.empty() } -> std::convertible_to<bool>;
```
gvolfing, line 117: Please include the concepts header.

src/query/interpreter.cpp
```diff
@@ -112,6 +112,16 @@ void UpdateTypeCount(const plan::ReadWriteTypeChecker::RWType type) {
...
+};
+
+template <HasEmpty T>
+inline std::optional<T> GenOptional(const T &in) {
+  return in.empty() ? std::nullopt : std::optional<T>{in};
```
gvolfing, line 122: Did you consider `std::make_optional<T>(in)` instead of the constructor call?

src/query/interpreter.cpp
```diff
@@ -1177,11 +1187,12 @@ Interpreter::Interpreter(InterpreterContext *interpreter_context) : interpreter_
...
 }
 
-PreparedQuery Interpreter::PrepareTransactionQuery(std::string_view query_upper) {
+PreparedQuery Interpreter::PrepareTransactionQuery(std::string_view query_upper,
+                                                   const std::map<std::string, storage::PropertyValue> &metadata) {
   std::function<void()> handler;
 
   if (query_upper == "BEGIN") {
-    handler = [this] {
+    handler = [this, metadata] {
```
gvolfing, line 1195: Did you consider capturing `metadata` by `&`?

src/query/interpreter.cpp
```diff
@@ -2761,15 +2786,15 @@ Interpreter::PrepareResult Interpreter::Prepare(const std::string &query_string,
...
+    query_execution->prepared_query.emplace(PrepareTransactionQuery(trimmed_query, metadata));
     return {query_execution->prepared_query->header, query_execution->prepared_query->privileges, qid};
   }
 
   // Don't save BEGIN, COMMIT or ROLLBACK
   transaction_queries_->push_back(query_string);
 
   // All queries other than transaction control queries advance the command in
-  // an explicit transaction block.
+  // an explicit transaction block.in_explicit_transaction_
```
gvolfing, line 2797: Not sure if this is a typo.

gvolfing (approved):
Left some stylistic changes, nothing major, great job overall!

src/query/interpreter.cpp
andrejtonev: The handler gets returned and called in the Pull. Unfortunately I think we have to copy so it doesn't segfault.

src/communication/bolt/v1/states/handlers.hpp
```diff
@@ -347,6 +356,12 @@ State HandleBegin(TSession &session, const State state, const Marker marker) {
     return State::Close;
   }
 
+  std::map<std::string, Value> md{};
+  auto &md_tv = extra.ValueMap()["tx_metadata"];
+  if (md_tv.IsMap()) {
+    md = md_tv.ValueMap();
+  }
```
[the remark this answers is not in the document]
andrejtonev: Moved to a helper function.

src/memgraph.cpp
andrejtonev: Sure

src/query/interpreter.cpp
gvolfing: This would break the current state of the code, but if you do a generalized lambda capture `[this, metadata = std::move(metadata)]` then you can move the `std::map<>` instead of a copy. For that you would need to remove the `const`-ness of the `&` from the call-chain leading up to this part.
Also, -this is not the case here but if we want to do that for some reason- if at any point you would move `metadata` out of the closure, you would need to make the lambda `mutable` as well.
Not sure how big this map can become, if copying it is not a problem it does not really matter.

gvolfing: That's fine, maybe leave a TODO comment somewhere around here.

vpavicic:
Release note
When you need to differentiate many different transactions, you can now define and pass metadata via the client. In Memgraph, the metadata is visible in the output of the `SHOW TRANSACTIONS;` query. 

@andrejtonev ok?

andrejtonev:
Perfect!

andrejtonev:
closed #934

PR #949, opened 2023-05-25
[master < ] Added a default server name

### LINKED ISSUE #474: Fix Bolt HELLO message response doesn't contains "server" (default value empty)
**Memgraph version**
2.3.1

**Environment**
Windows, Docker, PHP Bolt library

**Describe the bug**
After connecting to memgraph over bolt first message is "hello" to login into. Server response for this message should be map with keys "server" and "connection_id" (optional "hints"). Currently response contains only "connection_id". 

**To Reproduce**
1. Connect to database over bolt and run hello message
2. Analyze response

**Expected behavior**
Response has to contain key "server". Missing key "server" have to contain server agent string identifier like "memgraph/2.3.1".

andrejtonev:
### NOTE
We should expand this PR and include the following flags to be updateable during the instance lifetime:
* all timeout related flags -> `cd build && ./memgraph --help | grep timeout`
* `--log-level`
* `--also-log-to-stderr`

### [master < Task] PR

### To keep docs changelog up to date, one more thing to do:

closes #474 

Update:
This issue has spiraled a bit. We are now trying to enable the user to easily change the name and cache it for future start-ups.

### DIFF
--- a/src/memgraph.cpp
+++ b/src/memgraph.cpp
@@ -9,464 +9,35 @@
 // by the Apache License, Version 2.0, included in the file
 // licenses/APL.txt.
 
-#include <algorithm>
-#include <atomic>
-#include <chrono>
-#include <csignal>
-#include <cstdint>
-#include <exception>
-#include <filesystem>
-#include <functional>
-#include <limits>
-#include <map>
-#include <memory>
-#include <optional>
-#include <regex>
-#include <string>
-#include <string_view>
-#include <thread>
-
-#include <fmt/core.h>
-#include <fmt/format.h>
-#include <gflags/gflags.h>
-#include <spdlog/common.h>
-#include <spdlog/sinks/daily_file_sink.h>
-#include <spdlog/sinks/dist_sink.h>
-#include <spdlog/sinks/stdout_color_sinks.h>
+#ifndef MG_ENTERPRISE
+#include "dbms/session_context_handler.hpp"
+#endif
 
 #include "audit/log.hpp"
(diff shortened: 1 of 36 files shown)


gitbuda:
A note here. I think we should expand this PR to also include that's possible to change the timeout settings during the lifetime of an instance of Memraph because that's very very correlated. Let me know when we get back to this.

antoniofilipovic:
@andrejtonev can this be reviewed too?

andrejtonev:
> 

@antoniofilipovic the PR should be ready, but it only implements the default server name and adds the name to local settings.
@gitbuda what are we doing with this PR? If I remember correctly, the we expanded the issue, and I'm not really sure if the PR, as is now, should be merged.

src/memgraph.cpp
```diff
@@ -106,6 +106,10 @@ constexpr const char *kMgUser = "MEMGRAPH_USER";
 constexpr const char *kMgPassword = "MEMGRAPH_PASSWORD";
 constexpr const char *kMgPassfile = "MEMGRAPH_PASSFILE";
 
+constexpr const char *kServerNameSettingKey = "server.name";
+static const auto kMgDefaultName = std::string("Memgraph/") + version_string;
```
gvolfing, line 110:
```suggestion
static const auto kMgDefaultName = fmt::format("Memgraph/{}", version_string);
```

gvolfing (approved):

andrejtonev:
moved to #1183

PR #952, opened 2023-06-01
[master < E] Add multi-tenancy v1

### LINKED ISSUE #833: Support multi-tenancy v1
INPUT:
* [One thing we're missing and would love to understand if it's in the product pipeline is multi-tenancy support. Not like in Neo's composite DB's, but separating data within the same DB using something like namespaces.](https://discord.com/channels/842007348272169002/1085639469442551929/1085860460450029649)

andrejtonev:
Closes #833 

[master < Epic] PR
- [ ] Write E2E tests

I guess this will be an epic. Still need to specify a task.


src/dbms/handler.hpp
```diff
@@ -0,0 +1,140 @@
...
+   *
+   * @tparam T1 Variadic template of context constructor arguments
+   * @tparam T2 Variadic template of config constructor arguments
+   * @param name Name associated with the new context/config pair
+   * @param args1 Arguments passed (as a tuple) to the context constructor
+   * @param args2 Arguments passed (as a tuple) to the config constructor
+   * @return NewResult
+   */
+  template <typename... T1, typename... T2>
+  NewResult New(std::string name, std::tuple<T1...> args1, std::tuple<T2...> args2) {
```
Ignition, line 54: Is there a reason we are deferring the construction of TContext and TConfig? 
Also https://en.cppreference.com/w/cpp/utility/make_from_tuple should make this code cleaner

src/dbms/handler.hpp
```diff
@@ -0,0 +1,140 @@
...
+   * @tparam T2 Variadic template of config constructor arguments
+   * @tparam I1 List of indexes associated with the first tuple
+   * @tparam I2 List of indexes associated with the second tuple
+   */
+  template <typename... T1, typename... T2, std::size_t... I1, std::size_t... I2>
+  NewResult New_(std::string name, std::tuple<T1...> &args1, std::tuple<T2...> &args2,
+                 std::integer_sequence<std::size_t, I1...> /*not-used*/,
+                 std::integer_sequence<std::size_t, I2...> /*not-used*/) {
+    // Make sure the emplace will succeed, since we don't want to create temporary objects that could break something
+    if (!Has(name)) {
```
Ignition, line 127: OK so this is why we are deferring.

src/dbms/session_context_handler.hpp
```diff
@@ -0,0 +1,611 @@
...
+    std::shared_lock<LockT> rd(lock_);
+    return storage_handler_.All();
+  }
+
+  /**
+   * @brief Return the number of vertex across all databases.
+   *
+   * @return uint64_t
+   */
+  std::tuple<uint64_t, uint64_t, uint64_t> Info() const {
```
Ignition, line 355: Can this be a named type? `std::tuple<uint64_t, uint64_t, uint64_t>` is non descriptive.

Ignition (commented):
Made it to storage_handler.cpp ... continue tomorrow.

src/auth/auth.cpp
```diff
@@ -314,4 +314,57 @@ std::vector<auth::User> Auth::AllUsersForRole(const std::string &rolename_orig)
...
+    SaveUser(*user);
+    return true;
+  }
+  return false;
+}
+
+void Auth::DeleteDatabase(const std::string &db) {
+  for (auto it = storage_.begin(kUserPrefix); it != storage_.end(kUserPrefix); ++it) {
+    auto username = it->first.substr(kUserPrefix.size());
+    if (username != utils::ToLowerCase(username)) continue;
```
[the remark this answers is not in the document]
Darych: Could you please explain what is the meaning of this comparison? If we read from RocksDB to which only we can write, what is the case when the username is in mixed/upper case?

src/auth/models.cpp
```diff
@@ -464,6 +483,78 @@ bool operator==(const Role &first, const Role &second) {
...
+void Databases::Remove(const std::string &db) {
+  denies_dbs_.emplace(db);
+  grants_dbs_.erase(db);
+}
+
+void Databases::Delete(const std::string &db) {
+  denies_dbs_.erase(db);
+  if (!allow_all_) {
+    grants_dbs_.erase(db);
+  }
```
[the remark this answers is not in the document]
andrejtonev: Yeah, I think this logic needs to be updated.
Currently I default to allow_all(true) and default_db_(kDefaultDB). But a better solution would be to default to false and none.
Then it would be up to the admin to define exactly what they want.

src/communication/bolt/v1/session.hpp
```diff
@@ -48,14 +54,31 @@ class SessionException : public utils::BasicException {
  * @tparam TOutputStream type of output stream that will be used
  */
 template <typename TInputStream, typename TOutputStream>
-class Session {
+#ifdef MG_ENTERPRISE
+class Session : public dbms::SessionInterface
+#else
+class Session
+#endif
```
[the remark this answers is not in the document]
andrejtonev: Just have an empty interface if ENTERPRISE OFF?

src/communication/bolt/v1/states/handlers.hpp
```diff
@@ -224,7 +225,11 @@ State HandleRunV1(TSession &session, const State state, const Marker marker) {
 
   DMG_ASSERT(!session.encoder_buffer_.HasData(), "There should be no data to write in this state");
 
+#ifdef MG_ENTERPRISE
+  spdlog::debug("[Run - {}] '{}'", session.GetID(), query.ValueString());
+#else
   spdlog::debug("[Run] '{}'", query.ValueString());
+#endif
```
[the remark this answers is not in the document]
andrejtonev: The database logic is not present in ENTERPRISE OFF, so there isn't really an ID that we could put here.

src/dbms/handler.hpp
```diff
@@ -0,0 +1,140 @@
...
+  /**
+   * @brief Delete the context/config pair associated with the name.
+   *
+   * @param name Name associated with the context/config pair to delete
+   * @return true on success
+   */
+  bool Delete(const std::string &name) {
+    if (auto itr = items_.find(name); itr != items_.end()) {
+      itr->second.DestroyAndSync();
+      items_.erase(itr);
```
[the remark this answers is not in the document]
andrejtonev: The handler itself is not thread safe.
The SessionContextHandler (which uses the Handler is thread safe)

src/dbms/handler.hpp
andrejtonev: :D

src/dbms/session_context_handler.hpp
andrejtonev: Sure

src/utils/sync_ptr.hpp
```diff
@@ -0,0 +1,188 @@
...
+   * @brief Construct a new synched pointer.
+   *
+   * @tparam TArgs variable templates used by the TContext constructor
+   * @param config Additional metadata associated with context
+   * @param args Arguments to pass to TContext constructor
+   */
+  template <typename... TArgs>
+  explicit SyncPtr(TArgs &&...args)
+      : timeout_{1000},
+        ptr_{new TContext(std::forward<TArgs>(args)...), std::bind(&SyncPtr::OnDelete, this, std::placeholders::_1)} {}
```
Ignition, line 126: Use `[this](TContext* ptr){ this->OnDelete(ptr); }`

src/utils/sync_ptr.hpp
```diff
@@ -0,0 +1,188 @@
...
+  /**
+   * @brief Block until OnDelete gets called.
+   *
+   */
+  void SyncOnDelete() {
+    std::unique_lock<std::mutex> lock(in_use_mtx_);
+    if (!in_use_cv_.wait_for(lock, timeout_, [this] { return !in_use_; })) {
+      throw utils::BasicException("Syncronization timeout!");
+    }
+  }
```
Ignition, line 165: I feel uneasy about this, we should talk.

src/utils/sync_ptr.hpp
```diff
@@ -0,0 +1,188 @@
...
+   *
+   * @param p Pointer to the undelying object.
+   */
+  void OnDelete(TContext *p) {
+    delete p;
+    {
+      std::lock_guard<std::mutex> lock(in_use_mtx_);
+      in_use_ = false;
+    }
+    in_use_cv_.notify_one();
```
Ignition, line 178: notify_all?

src/auth/auth.cpp
```diff
@@ -314,4 +314,57 @@ std::vector<auth::User> Auth::AllUsersForRole(const std::string &rolename_orig)
...
+    auto user = GetUser(username);
+    if (user) {
+      user->db_access().Delete(db);
```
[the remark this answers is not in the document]
andrejtonev: Yes!

src/communication/bolt/v1/states/handlers.hpp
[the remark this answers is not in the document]
Darych: @Ignition It's a database name not session ID. I added comment in global.hpp about this naming.

src/auth/models.cpp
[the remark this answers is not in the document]
andrejtonev: This will be a bit different in the end. 
In order to have the same behavior as before, the default db will be set to kDefaultDB, but allow_all will be false and the only databases granted will be the kDefaultDB. 
The idea is to have everything work the with and without the multi-tenancy. I also default to this when a user saved with the previous version is read.

src/communication/v2/session.hpp
```diff
@@ -253,23 +272,26 @@ class WebsocketSession : public std::enable_shared_from_this<WebsocketSession<TS
...
 
+#ifdef MG_ENTERPRISE
+  ~Session() { session_context_->Delete(session_); }
+
   Session(const Session &) = delete;
   Session(Session &&) = delete;
   Session &operator=(const Session &) = delete;
   Session &operator=(Session &&) = delete;
-  ~Session() = default;
+#endif
```
Darych, line 294: Why do we put copy/move constructors/operators only for MG_ENTERPRISE?

src/communication/v2/session.hpp
```diff
@@ -334,18 +356,22 @@ class Session final : public std::enable_shared_from_this<Session<TSession, TSes
...
-        data_{data},
+        session_{*session_context, endpoint, input_buffer_.read_end(), &output_stream_},
+        session_context_{session_context},
         endpoint_{endpoint},
         remote_endpoint_{GetRemoteEndpoint()},
         service_name_{service_name},
         timeout_seconds_(inactivity_timeout_sec),
         timeout_timer_(GetExecutor()) {
+#ifdef MG_ENTERPRISE
+    session_context_->Register(session_);
```
Darych, line 373: As I understood now we have cyclic dependency between session_context and session.
Do we have plans in next release to get rid of it?

src/glue/auth_handler.cpp
```diff
@@ -463,6 +552,7 @@ std::vector<std::vector<memgraph::query::TypedValue>> AuthQueryHandler::GetPrivi
     std::vector<std::vector<memgraph::query::TypedValue>> grants;
 #ifdef MG_ENTERPRISE
     std::vector<std::vector<memgraph::query::TypedValue>> fine_grained_grants;
+    std::vector<std::vector<memgraph::query::TypedValue>> database_grants;
```
Darych, line 555: Do we need this vector?

src/query/frontend/ast/cypher_main_visitor.hpp
```diff
@@ -524,6 +524,11 @@ class CypherMainVisitor : public antlropencypher::MemgraphCypherBaseVisitor {
    */
   antlrcpp::Any visitEntitiesList(MemgraphCypher::EntitiesListContext *ctx) override;
 
+  /**
+   * @return std::vector<std::string>
```
Darych, line 528: Are we sure that `visitWildcardName` returns vector?

tests/e2e/fine_grained_access/edge_type_filtering_tests.py
```diff
@@ -1,91 +1,116 @@
...
+def test_all_edge_types_all_labels_granted(switch):
     admin_connection = common.connect(username="admin", password="test")
-    user_connnection = common.connect(username="user", password="test")
+    user_connection = common.connect(username="user", password="test")
     common.execute_and_fetch_all(admin_connection.cursor(), "GRANT READ ON LABELS * TO user;")
     common.execute_and_fetch_all(admin_connection.cursor(), "GRANT READ ON EDGE_TYPES * TO user;")
 
-    results = common.execute_and_fetch_all(user_connnection.cursor(), "MATCH (n)-[r]->(m) RETURN n,r,m;")
+    if switch:
+        common.switch_db(user_connection.cursor())
```
Darych, line 15: Are we sure that `clean` database exists? We don't call `common.create_multi_db(user_connection.cursor(), switch)` to create `clean` database.

Darych: Oh I saw that we create database in `workloads.yaml`. Comment is resolved now.

tests/integration/auth/runner.py
```diff
@@ -241,70 +200,126 @@ def cleanup():
 
     # Run the test with all combinations of permissions
     print("\033[1;36m~~ Starting query test ~~\033[0m")
+    for db in ["memgraph", "db1"]:
+        print("\033[1;36m~~ Running against db {} ~~\033[0m".format(db))
+        execute_user_queries(["USE DATABASE {}".format(db)], should_fail=True, failure_message=UNAUTHORIZED_ERROR)
+        execute_admin_queries(["GRANT MULTI_DATABASE_USE TO User"])
```
Darych, line 206: Is it documented query?

tests/integration/auth/runner.py
```diff
@@ -241,70 +200,126 @@ def cleanup():
 
     # Run the test with all combinations of permissions
     print("\033[1;36m~~ Starting query test ~~\033[0m")
+    for db in ["memgraph", "db1"]:
+        print("\033[1;36m~~ Running against db {} ~~\033[0m".format(db))
+        execute_user_queries(["USE DATABASE {}".format(db)], should_fail=True, failure_message=UNAUTHORIZED_ERROR)
```
Darych, line 205: Why `USE DATABASE memgraph` should fail? It's a default database

tests/unit/dbms_storage.cpp
```diff
@@ -0,0 +1,225 @@
...
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
+// License, and you may not use this file except in compliance with the Business Source License.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#if 0
```
Darych, line 12: I think this file can be also removed from PR

tests/unit/utils_sync_ptr.cpp
```diff
@@ -0,0 +1,296 @@
...
+  ASSERT_FALSE(alive);
+
+  memgraph::utils::SyncPtr<Test> sp(alive);
+
+  {
+    using namespace std::chrono_literals;
+    sp.timeout(10000ms);  // 10sec
+
+    ASSERT_TRUE(alive);
+    auto sp_copy1 = sp.get();
```
Darych, line 96: Do we need `sp_copy1` and `sp_copy2` if we don't check them?

src/kvstore/kvstore.cpp
```diff
@@ -9,6 +9,8 @@
 // by the Apache License, Version 2.0, included in the file
 // licenses/APL.txt.
 
+#include <iostream>
+
```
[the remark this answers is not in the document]
andrejtonev: no...

src/storage/v2/config.hpp
```diff
@@ -62,15 +68,52 @@ struct Config {
...
 
+#define UPDATE_PATH(update)                                                                                 \
+  {                                                                                                         \
+    const auto old_path = std::filesystem::weakly_canonical(std::filesystem::absolute(config.disk.update)); \
+    const auto contained_path = contained(old_path, old_base);                                              \
+    if (!contained_path) {                                                                                  \
+      throw StorageConfigException("On-disk directories not contained in root.");                           \
+    }                                                                                                       \
+    config.disk.update = config.durability.storage_directory / *contained_path;                             \
+  }
```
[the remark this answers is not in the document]
andrejtonev: Nice

src/dbms/global.hpp
```diff
@@ -0,0 +1,117 @@
...
+   * @return std::string
+   */
+  virtual std::string UUID() const = 0;
+
+  /**
+   * @brief Return the currently active database.
+   *
+   * @return std::string
+   */
+  virtual std::string GetID() const = 0;
```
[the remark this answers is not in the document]
andrejtonev: Changed to GetDatabaseName()

src/dbms/global.hpp
```diff
@@ -0,0 +1,117 @@
...
+
+  /**
+   * @brief Gets called on database change.
+   *
+   * @return true on success
```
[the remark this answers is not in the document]
andrejtonev: What do you mean?

tests/integration/auth/runner.py
andrejtonev: Because we revoked its access from the user.
"memgraph" is THE default database, but each user can have a different private default database.

tests/integration/auth/runner.py
andrejtonev: It is now :D

src/glue/auth_handler.cpp
```diff
@@ -122,6 +122,29 @@ std::vector<std::vector<memgraph::query::TypedValue>> ShowRolePrivileges(
...
+  if (!memgraph::license::global_license_checker.IsEnterpriseValidFast() || !user) {
+    return {};
+  }
+
+  const auto &db = user->db_access();
+  const auto &allows = db.GetAllowAll();
+  const auto &grants = db.GetGrants();
+  const auto &denies = db.GetDenies();
+
+  std::vector<memgraph::query::TypedValue> res;  // First element is a list of granted databases, second of revoked ones
```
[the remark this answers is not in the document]
andrejtonev: This is just the result that gets sent to the user.
It has to be a vector of TypedValue; the TypedValue here will be a vector of 2 elements.
I'll check in other queries for a better solution.

tests/unit/utils_sync_ptr.cpp
andrejtonev: We don't need to check them, we just what to copy and then destroy them.
SyncPtr will block until they are destroyed. We create a new thread and wait there.
This is the simplest test; there are timeout tests as well.

src/glue/auth_handler.cpp
andrejtonev: no :)

src/communication/v2/session.hpp
andrejtonev: I was modifying the ~Session and made a mistake.
They should be deleted in any case.

src/communication/v2/session.hpp
andrejtonev: Just to clarify a bit (the code is a bit of a mess).
session_context_ (in the enterprise build) is the SessionContextHandler.
session_ is the SessionHL which implements the bolt::Session interface.

SessionContextHandler actually depends on SessionInterface, an interface that bolt::Session inherits.

SessionContextHandler uses this Register to keep track of the active sessions and loop through them when deleting a database. 
SessionHL uses the SessoinContextHandler to do all the multi-tenency stuff.

Maybe we could remove the SessionInterface somehow. Got any suggestions?

Darych: Currently no :) Need to dive into this more

src/dbms/global.hpp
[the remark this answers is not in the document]
Darych: The return type now is not bool

src/communication/v2/session.hpp
andrejtonev: Actually the registration and looping through is a leftover from an older design. It still has a function but it may be easier to remove than I tough.
Everything is now contained in the InterpreterContext and the only thing OnDelete (the callback I loop trough) does is that is double checks that everything is fine.
I think if we change the synchronization a bit, we could have the sessions just increment/decrement a ref counter which we check in SessionContextHandler and delete if 0. 
There are still some problems with the defunct storage, but it could work.

I'll put a TODO here and resolve this comment.

src/glue/auth_handler.cpp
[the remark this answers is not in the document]
andrejtonev: Other Show*(user) and Get*(username) functions of the AuthHandler directly return the TypedValue vector. 
I think it would be better to follow the same template. This gets called from the interpreter and its result is directly used.
Let me know if that's okay with you.

src/communication/bolt/v1/session.hpp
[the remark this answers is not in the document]
andrejtonev: Yeah, that made the code a bit nicer :)

src/glue/auth_handler.cpp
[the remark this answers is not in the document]
Ignition: Sure, outside the scope of this PR.

Ignition (approved):

Darych (approved):

andrejtonev:
@vpavicic 

Documentation PR [989](https://github.com/memgraph/docs/pull/989)

PR #1017, opened 2023-06-15
[E952 < T] Sync on delete

andrejtonev:
Storage has become a shared_ptr. SessionContext now has a copy of that pointer. When we want to delete a database, we first make sure no one is using it and then go through the sessions and execute their callbacks (that should release any SessionContexts that are associated with the storage) and then we delete the original storage pointer. However, this isn't a guarantee that the storage will be deleted.
The problem is that bolt::Sessions have to switch the database they are using on the fly. This is done by creating a temporary SessionHL (just copying a shared_ptr) when executing the API, then switching the database during Pull and having the SessionHL live until the Pull is done. There is a brief period between switching the database and destroying the previous SessionHL, where we could want to delete the storage, while the SessionHL associated with it is still active. This isn't a huge problem, we will delete our pointer and the storage will get deleted once all temporary SessionHL objects are destroyed. The problem is that we are not guaranteeing that the database has been dropped when we return from the query (as there could be this temporary SessionHL still holding a copy of the storage).
The solution is to wait until the storage is actually destroyed. 
So far, this hasn't caused any sync problems, even when stress testing it. But should be tested further before merging.


src/dbms/utils.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+#include <condition_variable>
+#include <functional>
+#include <iostream>
+#include <memory>
+
+namespace memgraph::dbms {
+
+template <typename TContext, typename TConfig>
+struct SyncPtr {
```
gitbuda, line 22: This is very generic -> move it under `src/utils/`?

src/dbms/utils.hpp
```diff
@@ -0,0 +1,62 @@
...
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
+// License, and you may not use this file except in compliance with the Business Source License.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+// TODO: Check if comment above is ok
```
gitbuda, line 11: Seems ok 😄

PR #1018, opened 2023-06-16
[master < T0869] Implement GraphQL compatibility

### LINKED ISSUE #869: Implement GraphQL compatibility
Many people expose GraphQL API to the underlying database engine, so-called "data access layer". The point here is to provide an easy way for any Memgraph user to fetch data stored in Memgraph via GraphQL. Make it possible to integrate with https://github.com/neo4j/graphql

Epic PR for issue:
https://github.com/memgraph/memgraph/pull/1018

**TASKS**
- [x] Issue: #841 PR: #892

gvolfing:
Closes #869


src/query/graphql_checker.hpp
```diff
@@ -0,0 +1,48 @@
...
+#include <string>
+#include <string_view>
+
+namespace memgraph::query {
+
+// TODO -1- (gvolfing) - check what happens if we can not provide
+// the same signature for any given functionality
+// TODO -2- (gvolfing) - test this once we have some facilities.
+// TODO -3- (gvolfing) - check if we should read in the mapping from
+// some files instead.
```
gitbuda, line 22: We should, I would probably add a file Memgraph can read under `memgraph/config` folder, if something more advanced is needed, we can use `YAML` format and generate a specific "GraphQL config" file that you can easily parse here. The same pattern as it's used right now inside `glags.yaml` + `generate.py`. Ofc, in this case I think it doesn't have to be that advanced, it can be something simple, we just have to ensure the right config file ends up in the Memgraph packages (DEB & RPM).

src/memgraph.cpp
```diff
@@ -26,14 +33,6 @@
 #include <string_view>
 #include <thread>
 
-#include <fmt/core.h>
-#include <fmt/format.h>
-#include <gflags/gflags.h>
-#include <spdlog/common.h>
-#include <spdlog/sinks/daily_file_sink.h>
-#include <spdlog/sinks/dist_sink.h>
-#include <spdlog/sinks/stdout_color_sinks.h>
```
gitbuda, line 35: The idea/principle here is that 3rd party libs are below std libs (our headers should be in a separated block below the 3rd party libs), that's just the convention which is true for most of the codebase.

src/query/frontend/ast/ast.hpp
```diff
@@ -2253,6 +2253,7 @@ class CallProcedure : public memgraph::query::Clause {
   memgraph::query::Expression *memory_limit_{nullptr};
   size_t memory_scale_{1024U};
   bool is_write_;
+  bool is_util_validate_procedure_{false};
```
gitbuda, line 2256: Shouldn't this be more generic, meaning, does a procedure `YIELD` some values or no? I think the implementation can be more generic? 🤔

src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -1180,13 +1181,20 @@ antlrcpp::Any CypherMainVisitor::visitCallProcedure(MemgraphCypher::CallProcedur
   const auto &maybe_found =
       procedure::FindProcedure(procedure::gModuleRegistry, call_proc->procedure_name_, utils::NewDeleteResource());
   if (!maybe_found) {
-    throw SemanticException("There is no procedure named '{}'.", call_proc->procedure_name_);
+    const auto mg_specific_name = procedure::FindApocReplacement(call_proc->procedure_name_);
+    if (mg_specific_name) {
+      if (*mg_specific_name == "mgps.validate") {
```
gitbuda, line 1186: Should this come from the config file? If the implementation is more generic that this part here shouldn't exists in the cypher main visitor? All the logic should be done when the cursors are instantiated?

src/query/procedure/graphql_checker.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+    {"dbms.components", "mgps.components"}, {"apoc.util.validate", "mgps.validate"}
+
+}};
+
+}  // namespace
+
+namespace memgraph::query::procedure {
+
+std::optional<std::string_view> FindApocReplacement(std::string_view apoc_name) {
```
gitbuda, line 29: I would rename this to `AliasLookup` or something more generic

src/query/frontend/ast/ast.hpp
andrejtonev: Agree, better to just have a flag telling us that the CALL is without YIELD. Later on we can throw if the procedure called is con compatible with a CALL without YIELD.

src/query/procedure/callable_alias_mapper.cpp
```diff
@@ -0,0 +1,60 @@
...
+      if (is_regular_file && has_json_extension) {
+        std::ifstream mapping_file(path);
+        try {
+          json mapping_data = json::parse(mapping_file);
+          mapping_ = mapping_data.get<std::unordered_map<std::string, std::string>>();
+        } catch (const std::exception &e) {
+          spdlog::warn("Parsing callable mapping was unsuccesful, reason: {}", e.what());
+        }
+      } else {
+        spdlog::warn("Path to callable mappings is not a regular file or does not have .json extension.");
```
andrejtonev, line 43: Here there is a bunch of warning messages, shouldn't we just throw or abort?

src/query/procedure/callable_alias_mapper.cpp
```diff
@@ -0,0 +1,60 @@
...
+  if (path != "") {
+    if (std::filesystem::exists(path)) {
+      const bool is_regular_file = std::filesystem::is_regular_file(path);
+      auto asd = path.extension();
+      const bool has_json_extension = (path.extension() == ".json");
+      if (is_regular_file && has_json_extension) {
+        std::ifstream mapping_file(path);
+        try {
+          json mapping_data = json::parse(mapping_file);
+          mapping_ = mapping_data.get<std::unordered_map<std::string, std::string>>();
```
andrejtonev, line 38: How could we handle differences in the argument layout between the two mappings?

src/query/procedure/module.cpp
```diff
@@ -1326,10 +1327,25 @@ std::optional<std::pair<ModulePtr, const T *>> MakePairIfPropFound(const ModuleR
...
+    return std::nullopt;
+  }
   auto [module_name, prop_name] = *result;
   auto module = module_registry.GetModuleNamed(module_name);
-  if (!module) return std::nullopt;
+  if (!module) {
+    // Check for possible callable aliases.
+    const auto maybe_valid_alias = gCallableAliasMapper.FindAlias(std::string(fully_qualified_name));
+    if (maybe_valid_alias) {
+      result = FindModuleNameAndProp(module_registry, *maybe_valid_alias, memory);
```
andrejtonev, line 1339: Perhaps put this in a helper function and call it twice, instead of having duplicate logic.

src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -1188,13 +1189,20 @@ antlrcpp::Any CypherMainVisitor::visitCallProcedure(MemgraphCypher::CallProcedur
   const auto &maybe_found =
       procedure::FindProcedure(procedure::gModuleRegistry, call_proc->procedure_name_, utils::NewDeleteResource());
   if (!maybe_found) {
-    throw SemanticException("There is no procedure named '{}'.", call_proc->procedure_name_);
+    const auto mg_specific_name = procedure::gCallableAliasMapper.FindAlias(call_proc->procedure_name_);
```
andrejtonev, line 1192: Isn't the Aliasing logic already implemented in the FindProcedure?

src/query/frontend/ast/cypher_main_visitor.cpp
gvolfing: The config file is for the mapping only. "mgps.validate" is a very special case. The issue is that so far this is the only known case where we have the 'no YIELD' related semantics to deal with the procedure, in fact we are not able to create such procedures with the current implementation. You are right that the logic should be as close to the cursor instantiation as possible, however this is the place where have to verify the semantics of the query.
It is always the case that if we call a procedure without yielding that is a semantic error, except in the case of  "mgps.validate", which is not even a procedure, it is a hardcoded cursor.

I do not think we should push semantic checks further down from here.
Once we have void procedures, this will not be an issue at all, we will be able to handle void procedures as actual query procedures instead of hardcoded, hacky cursor implementations.

IMO with our current possibilities this is out best bet. Yes, we can make this more generic, but that would be just propagating already flawed design down the line.

Let me know what you think.

src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -1188,13 +1189,20 @@ antlrcpp::Any CypherMainVisitor::visitCallProcedure(MemgraphCypher::CallProcedur
...
+    if (mg_specific_name) {
+      if (*mg_specific_name == "mgps.validate") {
+        call_proc->is_util_validate_procedure_ = true;
```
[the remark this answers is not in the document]
gvolfing: That is the solution once we have proper support for void procedures.

src/query/plan/operator.cpp
```diff
@@ -4695,10 +4697,116 @@ class CallProcedureCursor : public Cursor {
...
+
+  void Shutdown() override {}
+};
+
 UniqueCursorPtr CallProcedure::MakeCursor(utils::MemoryResource *mem) const {
   memgraph::metrics::IncrementCounter(memgraph::metrics::CallProcedureOperator);
   CallProcedure::IncrementCounter(procedure_name_);
 
+  if (is_util_validate_procedure_) {
+    return MakeUniqueCursorPtr<CallValidateProcedureCursor>(mem, this, mem);
```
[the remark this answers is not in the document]
gvolfing: We could, but I would rather separate the already working, proven Cursor and the workaround implementation for mgps.validate. Generic version without YIELD will make sense once we add the support.

src/query/plan/operator.cpp
```diff
@@ -4695,10 +4697,116 @@ class CallProcedureCursor : public Cursor {
   }
 };
 
+class CallValidateProcedureCursor : public Cursor {
```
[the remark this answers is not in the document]
gvolfing: It should, but the rest of the cursors declare the inheritance in the `cpp` file as well, I wanted to keep it consistent.

src/query/procedure/callable_alias_mapper.cpp
```diff
@@ -0,0 +1,60 @@
...
+#include <json/json.hpp>
+
+namespace memgraph::query::procedure {
+
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
+CallableAliasMapper gCallableAliasMapper;
+
+void CallableAliasMapper::LoadMapping(const std::filesystem::path &path) {
+  using json = nlohmann::json;
+  if (path != "") {
```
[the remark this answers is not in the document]
gvolfing: I changed this it `!path.empty()` it is needed in the case if the user does not specify the flag and by default nothing is loaded into the mapper. The difference between this and the next check is that this one yields a info level log, the next one yields a warning, since that one means that the user tried to set a path, but it was faulty.

src/query/procedure/callable_alias_mapper.cpp
gvolfing: We do not.
As of now we do not know about any procedure that would have a different signature in our and in whose name shall not be spoken's procedures. If we need it we will change it, but I would rather keep it as simple as possible.

src/query/procedure/module.cpp
gvolfing: I am really not sure what are you referring to here, where is the logic duplicated?`
FindModuleNameAndProp` and `FindAlias` are not always used together, the FindAlias is only called the if the unmapped, original procedure name is not found in the module registry.

src/query/frontend/ast/cypher_main_visitor.cpp
gvolfing: It is used in two places. Here, it is used just specifically check for `mgps.validate` so we can instantiate the hardcoded cursor when creating the physical plan, the other is called from the traditional procedure cursor, to actually load a procedure from the module registry.

src/query/procedure/module.cpp
as51340: I am also not sure what you meant andreja but I would also separate this into smaller functions, having 3 nested if statements can always be improved

src/query/procedure/callable_alias_mapper.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include <json/json.hpp>
+
+namespace memgraph::query::procedure {
+
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
+CallableAliasMapper gCallableAliasMapper;
+
+void CallableAliasMapper::LoadMapping(const std::filesystem::path &path) {
+  using json = nlohmann::json;
+  if (!path.empty()) {
```
as51340, line 29: I would rather use early return statement
`if (path.empty())` return

src/query/procedure/callable_alias_mapper.cpp
```diff
@@ -0,0 +1,59 @@
...
+    if (std::filesystem::exists(path)) {
+      const bool is_regular_file = std::filesystem::is_regular_file(path);
+      const bool has_json_extension = (path.extension() == ".json");
+      if (is_regular_file && has_json_extension) {
+        std::ifstream mapping_file(path);
+        try {
+          json mapping_data = json::parse(mapping_file);
+          mapping_ = mapping_data.get<std::unordered_map<std::string, std::string>>();
+        } catch (const std::exception &e) {
+          spdlog::warn("Parsing callable mapping was unsuccesful, reason: {}", e.what());
```
as51340, line 39: I think e.what() will not be very helpful here, what do you think? How about simply stating: Parsing JSON callable mapping unsuccessful

tests/e2e/graphql/workloads.yaml
```diff
@@ -0,0 +1,14 @@
+template_cluster: &template_cluster
+  cluster:
+    main:
+      args: ["--bolt-port", "7687", "--log-level=TRACE", "--callable_mappings_path", "graphql/callable_alias_mapping.json"]
```
as51340, line 4: either use `=` or `,` for args

tests/e2e/graphql/graphql_crud.py
```diff
@@ -0,0 +1,70 @@
...
+
+def test_connect_or_create(query_server):
+    created_user_uuid = create_node_query(query_server)
+
+    connect_or_create_query = (
+        'mutation {updateUsers(update: {posts: {connectOrCreate: {where: { node: { id: "1234" } }onCreate: { node: { content: "Some content" } }}}},where: { id: "'
+        + created_user_uuid
+        + '" }) {info {nodesCreated}}}'
+    )
+    expected_respone = '{"data":{"updateUsers":{"info":{"nodesCreated":1}}}}\n'
```
as51340, line 63: response

tests/e2e/graphql/graphql_server.py
```diff
@@ -0,0 +1,140 @@
...
+
+def server_returned_expected(expected_string: str, server_response: requests.Response) -> bool:
+    expected_json = json.loads(expected_string)
+    server_response_json = json.loads(server_response.text)
+
+    expected = _flatten(_ordered(expected_json))
+    actual = _flatten(_ordered(server_response_json))
+
+    for expected_item, actual_item in zip(expected, actual):
+        if expected_item != actual_item:
```
as51340, line 71: you can concatenate these two conditions with and

tests/e2e/graphql/graphql_crud.py
```diff
@@ -0,0 +1,70 @@
+import sys
+
+import pytest
+from graphql_server import *
+
+
+def test_create_query(query_server):
+    create_node_query(query_server)
```
as51340, line 8: I think `create_node_query` should just create node and then this function should do assertions

tests/e2e/graphql/graphql_crud.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def test_nested_create_query(query_server):
+    create_related_nodes_query(query_server)
```
as51340, line 12: same here

tests/e2e/graphql/graphql_crud.py
```diff
@@ -0,0 +1,70 @@
...
+        + created_user_uuid
+        + '" }) {info {nodesCreated}}}'
+    )
+    expected_respone = '{"data":{"updateUsers":{"info":{"nodesCreated":1}}}}\n'
+
+    gotten = query_server.send_query(connect_or_create_query)
+    assert expected_respone == str(gotten.text)
+
+
+if __name__ == "__main__":
```
as51340, line 69: I think it would be good to include tests for `dbms.components` and `apoc.util.validate` that they are correctly translated to our procedures

tests/setup.sh
```diff
@@ -1,10 +1,10 @@
 #!/bin/bash
-
 # shellcheck disable=1091
 
 set -Eeuo pipefail
-
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd "$DIR"
```
as51340, line 6: why this change? Lines 6 and 22

src/query/frontend/ast/cypher_main_visitor.cpp
as51340: you can use and here instead of having multiple indentations, short circuit will solve problem

as51340 (changes requested):
Nice job gabor! I added some comments, let me know what you think

src/query/frontend/opencypher/grammar/MemgraphCypherLexer.g4
```diff
@@ -23,8 +23,6 @@ lexer grammar MemgraphCypherLexer ;
 
 import CypherLexer ;
 
-UNDERSCORE : '_' ;
```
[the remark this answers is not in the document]
gvolfing: That way we had access to the symbol from `Cypher.g4`.
The issue was that the transpiled query sometimes created '_' as a standalone variable name, which we were not able to handle before on the grammar level.

src/query/plan/operator.cpp
```diff
@@ -4695,10 +4697,120 @@ class CallProcedureCursor : public Cursor {
...
+    ExpressionEvaluator evaluator(&frame, context.symbol_table, context.evaluation_context, context.db_accessor,
+                                  storage::View::NEW);
+
+    const auto args = self_->arguments_;
+    MG_ASSERT(args.size() == 3);
+
+    const auto predicate = args[0]->Accept(evaluator);
+    const bool predicate_val = predicate.ValueBool();
+
+    if (predicate_val) [[unlikely]] {
```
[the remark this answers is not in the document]
gvolfing: The transpiled query, that is using this hardcoded operator can be thought of as an assertion. The way it works is that if the predicate that is being evaluated here is true, then the rest of the query is just ignored and the query throws. This should be an exceptional exception.

src/query/plan/operator.cpp
```diff
@@ -4695,10 +4697,120 @@ class CallProcedureCursor : public Cursor {
...
+        }
+
+        while (true) {
+          found = message_val.find('%', found);
+          if (found == std::string::npos) {
+            break;
+          }
+
+          // If someone finishes the message with '%' that could be error prone.
+          if (found == message_val.size() - 1U) {
```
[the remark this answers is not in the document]
gvolfing: I would rather not use a function here, but I will move the boolean logic into a `const bool` variable and name it appropriately. Good catch!

src/query/plan/operator.cpp
```diff
@@ -4695,10 +4697,120 @@ class CallProcedureCursor : public Cursor {
...
+            break;
+          }
+
+          std::string replacement_str;
+          const auto format_specifier = message_val.at(found + 1U);
+          if (!std::isalpha(format_specifier)) {
+            ++found;
+            continue;
+          }
+          const bool does_argument_list_overflow = (message_args_val.size() < arg_index - 1U) && (arg_index > 0U);
```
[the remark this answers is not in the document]
gvolfing: I think this is the correct order. Take this string as an example,
`"The test with test id:%d has been completed by student with student id:%d with 96% accuracy."[2,123]`
If we were to switch the order of the checks we would throw for this otherwise correct format string.

tests/e2e/graphql/graphql_crud.py
gvolfing: As you can see, this function is being used as a building block for the subsequent tests, this would mean that either I have to assert the correct response after every creation in every test, or I just do not assert in the other tests, which feels a bit odd. I would prefer asserting in the function.
What do you think?

tests/e2e/graphql/graphql_crud.py
gvolfing: Same as above.

src/query/plan/operator.cpp
```diff
@@ -4695,10 +4697,120 @@ class CallProcedureCursor : public Cursor {
...
+                "There are more format specifiers in the CALL procedure error message, then arguments provided.");
+          }
+          // If the number of arguments given exceed the number
+          // of format specifiers, than we ingore the excess ones.
+          if (arg_index > message_args_val.size() - 1U) {
+            break;
+          }
+
+          auto &current_arg = message_args_val.at(arg_index);
+          switch (format_specifier) {
```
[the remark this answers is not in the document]
gvolfing: As per the previous comments, I would prefer to keep it a lambda what I would rather not fragment.

src/query/procedure/callable_alias_mapper.cpp
gvolfing: Throwing is a bit tricky here, this would `throw` during startup, so not much interaction with the actual user, since this is not going through bolt protocol. For now I will change this to MG_ASSERT statements, which is the same way we assert the correctness of query module path correctness.
I am pretty sure the recovery process does throw though. So maybe I can do that as well.
What do you think would be better?

src/query/plan/operator.cpp
```diff
@@ -4695,10 +4697,115 @@ class CallProcedureCursor : public Cursor {
...
+    MG_ASSERT(args.size() == 3U);
+
+    const auto predicate = args[0]->Accept(evaluator);
+    const bool predicate_val = predicate.ValueBool();
+
+    if (predicate_val) [[unlikely]] {
+      const auto &message = args[1]->Accept(evaluator);
+      const auto &message_args = args[2]->Accept(evaluator);
+
+      auto format_error_message = [message_val = message.ValueString(),
```
andrejtonev, line 4729: This message formatting stuff looks like a generic functionality that we could move to utils.

andrejtonev (commented):
Just a couple of comments.

src/query/procedure/callable_alias_mapper.cpp
andrejtonev: I think MG_ASSERT is fine since this is at startup. We don't want to run if the mapping is faulty. This way the user will know 100% that something is not right :D

src/query/frontend/ast/cypher_main_visitor.cpp
andrejtonev: You could also use a constant for the name

src/query/procedure/callable_alias_mapper.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include "utils/logging.hpp"
+
+namespace memgraph::query::procedure {
+
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
+CallableAliasMapper gCallableAliasMapper;
+
+void CallableAliasMapper::LoadMapping(const std::filesystem::path &path) {
+  using json = nlohmann::json;
+  if (!path.empty()) {
```
as51340, line 31:
```suggestion
  if (path.empty()) {
```

And add a unit test which covers the scenario of correctly and incorrectly loading the mapping

src/query/procedure/callable_alias_mapper.cpp
```diff
@@ -0,0 +1,62 @@
...
+  if (std::filesystem::exists(path)) {
+    const bool is_regular_file = std::filesystem::is_regular_file(path);
+    const bool has_json_extension = (path.extension() == ".json");
+    if (is_regular_file && has_json_extension) {
+      std::ifstream mapping_file(path);
+      try {
+        json mapping_data = json::parse(mapping_file);
+        mapping_ = mapping_data.get<std::unordered_map<std::string, std::string>>();
+      } catch (...) {
+        MG_ASSERT("Parsing callable mapping was unsuccesful. Make sure it is in correct json format.");
```
as51340, line 45: What will `MG_ASSERT` do, will it also crash memgraph? How is that different than and why did you decide to use that instead of `spdlog::error`?

tests/e2e/graphql/graphql_crud.py
as51340: Yes currently yes. But I think that the logic is a bit wrong: because currently you have a server which creates example but the first extension of this would be to let clients send example they want (and I think this is how it should be even ATM, graphql_crud should send object which server needs to work with). So I think that current way doesn't really respect open-closed principle for extension.

as51340: Also, by testing things in graphql_server you don't have clear boundary between what are tests and what are just "helper" classes

tests/e2e/graphql/temporary_procedures/mgps.py
```diff
@@ -0,0 +1,8 @@
+import typing
+
+import mgp
+
+
+@mgp.read_proc
+def components(context: mgp.ProcCtx) -> mgp.Record(versions=list, edition=str):
```
as51340, line 7: This makes sense yes but how is that exactly called in these files, can you explain me maybe? I don't see any calls to dbms.components or anything I know I am missing something :smile:

as51340 (changes requested):
Comments about existing stuff and few smaller things

init
```diff
@@ -93,6 +93,10 @@ setup_virtualenv tests/stress
 setup_virtualenv tests/integration/ldap
 
 # Setup tests dependencies.
+# NOTE: This is commented out because of the build order (at the time of
```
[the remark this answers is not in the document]
gitbuda: Nope, it's not that easy. There is an issue with the build order, as explained here. These lines are called inside the GHA job when possible. The whole thing has to be improved, but it's not just uncommenting these lines -> this is one of the DevOps tasks. So, this conversation should just be resolved.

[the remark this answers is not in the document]
gitbuda: It should be called in a different place, I just don't know the right place at this time 😂

src/memgraph.cpp
```diff
@@ -355,6 +356,11 @@ DEFINE_VALIDATED_string(query_modules_directory, "",
                           return true;
                         });
 
+// NOLINTNEXTLINE (cppcoreguidelines-avoid-non-const-global-variables)
+DEFINE_string(callable_mappings_path, "",
+              "The path to mappings that describe aliases to callables in cypher queries in the form of key-value "
+              "pairs in a json file.");
```
gitbuda, line 362: I would explain here what the "callables" means, e.g., query module procedures and functions. In general, give a bit more context about this flag, but in additional 2 sentences max.

src/query/procedure/callable_alias_mapper.cpp
andrejtonev: spdlog::error will just print an error message, MG_ASSERT will close Memgraph because this is an unrecoverable error.
The user might not see the error log and think the everything is fine with the map, but it isn't.
MG_ASSERT is used in other places during startup when we cannot recover.

tests/e2e/graphql/temporary_procedures/mgps.py
gvolfing: That is something that is getting called when the GraphQL Library initiates connection with Memgraph. It is sent automatically.

src/query/plan/operator.cpp
gvolfing: Yep I started to restructure that.

src/query/procedure/callable_alias_mapper.cpp
gvolfing: Yes, forgot to add `false`. It should crash memgraph yes, if the user intended to specify the mapping file, but made an error with it, we should not wait just wait silently. Throwing would be another option, but other facilites that run upon startup tend to MG_ASSERT so I would just stick with that as well.

tests/e2e/graphql/graphql_crud.py
gvolfing: Modified these, I removed the checks from the helper functions and the bare functionality of creating vertices is done in the CREATE tests themselves.

init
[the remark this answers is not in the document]
as51340: No no, I just think that commented code doesn't have anything to do here. If we don't use it, let's delete it -> if we need it GitHub has version history. But currently pollutes code

tests/e2e/graphql/workloads.yaml
```diff
@@ -0,0 +1,14 @@
+template_cluster: &template_cluster
+  cluster:
+    main:
+      args: ["--bolt-port", "7687", "--log-level", "TRACE", "--callable_mappings_path", "graphql/callable_alias_mapping.json"]
+      log_file: "graphql-e2e.log"
+      setup_queries: []
+      validation_queries: []
+
+workloads:
```
as51340, line 9: Can you also add a workload for disk storage? You can check how is done [here](https://github.com/memgraph/memgraph/pull/1107). Just so we don't need to open later another PR

src/utils/java_string_formatter.hpp
```diff
@@ -0,0 +1,101 @@
...
+concept TTypedValueLike = requires(T t) {
+  { t.ValueInt() } -> std::convertible_to<int>;
+  { t.ValueDouble() } -> std::convertible_to<double>;
+  { t.ValueString() } -> std::convertible_to<pmr::string>;
+};
+
+template <typename TString, typename TTypedValueLike>
+class JStringFormatter final {
+ public:
+  [[nodiscard]] TString FormatString(TString str, const pmr::vector<TTypedValueLike> &format_args) const {
```
as51340, line 41: Can you catch string here by reference?

as51340 (approved):
I really like the refactoring job you made and have just 2-3 more questions.

katarinasupe:
Trying to run create mutation and get the following error message:
`"message": "There is no procedure named 'dbms.components'."`
The flag callable_mappings_path is set to `/home/ubuntu/memgraph/build/settings-graphql.json` and the content of that is:
```
{
"dbms.components": "mgps.components",
"apoc.util.validate": "mgps.validate"
}

```

What am I doing wrong here?

gvolfing:
@katarinasupe 
We need another query module to be able to communicate with the GraphQL Server.
Take the one that already exist on memgraph/tests/e2e/graphql/temporary_procedures/mgps.py copy that somewhere and specify the path to its containing directory with the --query-modules-directory flag, that should load the query module on startup, so the alias mapping should find it when the GraphQL Library sends it through bolt protocol.

katarinasupe:
By running delete query, I got `Error message: Query should either create or update something, or return results!` when wanting to run `MATCH (n)-[r]->(m) RETURN r, n, m` after deletion. I tried running that query two times, and I did not succeed. Then I ran `MATCH (n) RETURN n`, got weird results in Lab and then I managed to run `MATCH (n)-[r]->(m) RETURN r, n, m` successfully. After running the delete, and doing other stuff in db, then running `MATCH (n) RETURN n` again, I randomly reproduce the same error. 

Weird results in Lab (this is style related and style depends on Labels + properties so I would say it's that kind of issues, something doing with Node objects not being good):
Screenshot 2023-07-28 at 16 51 36

logs:
```
[2023-07-28 14:51:03.967] [memgraph_log] [debug] [Run] 'MATCH (this:`Post`)
WHERE this.content = $param0
DETACH DELETE this'
[2023-07-28 14:51:04.043] [memgraph_log] [debug] [Run] 'SHOW STORAGE INFO'
[2023-07-28 14:51:04.150] [memgraph_log] [debug] [Run] 'MATCH (APP_INTERNAL_EXEC_VAR) return COUNT(APP_INTERNAL_EXEC_VAR) as cnt'
[2023-07-28 14:51:07.042] [memgraph_log] [debug] [Run] 'SHOW STORAGE INFO'
[2023-07-28 14:51:07.146] [memgraph_log] [debug] [Run] 'MATCH (APP_INTERNAL_EXEC_VAR) return COUNT(APP_INTERNAL_EXEC_VAR) as cnt'
[2023-07-28 14:51:10.040] [memgraph_log] [debug] [Run] 'SHOW STORAGE INFO'
[2023-07-28 14:51:10.144] [memgraph_log] [debug] [Run] 'MATCH (APP_INTERNAL_EXEC_VAR) return COUNT(APP_INTERNAL_EXEC_VAR) as cnt'
[2023-07-28 14:51:12.254] [memgraph_log] [debug] [Run] 'MATCH (n)-[r]->(m)'
[2023-07-28 14:51:12.255] [memgraph_log] [trace] Error message: Query should either create or update something, or return results!
[2023-07-28 14:51:12.318] [memgraph_log] [trace] Reset received
[2023-07-28 14:51:13.044] [memgraph_log] [debug] [Run] 'SHOW STORAGE INFO'
[2023-07-28 14:51:13.159] [memgraph_log] [debug] [Run] 'MATCH (APP_INTERNAL_EXEC_VAR) return COUNT(APP_INTERNAL_EXEC_VAR) as cnt'
[2023-07-28 14:51:16.040] [memgraph_log] [debug] [Run] 'SHOW STORAGE INFO'
[2023-07-28 14:51:16.157] [memgraph_log] [debug] [Run] 'MATCH (APP_INTERNAL_EXEC_VAR) return COUNT(APP_INTERNAL_EXEC_VAR) as cnt'
[2023-07-28 14:51:17.110] [memgraph_log] [debug] [Run] 'MATCH (n)-[r]->(m)'
[2023-07-28 14:51:17.110] [memgraph_log] [trace] Error message: Query should either create or update something, or return results!
[2023-07-28 14:51:17.162] [memgraph_log] [trace] Reset received
[2023-07-28 14:51:19.076] [memgraph_log] [debug] [Run] 'SHOW STORAGE INFO'
[2023-07-28 14:51:19.191] [memgraph_log] [debug] [Run] 'MATCH (APP_INTERNAL_EXEC_VAR) return COUNT(APP_INTERNAL_EXEC_VAR) as cnt'
[2023-07-28 14:51:22.044] [memgraph_log] [debug] [Run] 'SHOW STORAGE INFO'
[2023-07-28 14:51:22.156] [memgraph_log] [debug] [Run] 'MATCH (APP_INTERNAL_EXEC_VAR) return COUNT(APP_INTERNAL_EXEC_VAR) as cnt'
[2023-07-28 14:51:24.618] [memgraph_log] [debug] [Run] 'MATCH (n)

RETURN n'
[2023-07-28 14:51:25.042] [memgraph_log] [debug] [Run] 'SHOW STORAGE INFO'
[2023-07-28 14:51:25.147] [memgraph_log] [debug] [Run] 'MATCH (APP_INTERNAL_EXEC_VAR) return COUNT(APP_INTERNAL_EXEC_VAR) as cnt'
[2023-07-28 14:51:28.041] [memgraph_log] [debug] [Run] 'SHOW STORAGE INFO'
[2023-07-28 14:51:28.146] [memgraph_log] [debug] [Run] 'MATCH (APP_INTERNAL_EXEC_VAR) return COUNT(APP_INTERNAL_EXEC_VAR) as cnt'
[2023-07-28 14:51:31.054] [memgraph_log] [debug] [Run] 'SHOW STORAGE INFO'
[2023-07-28 14:51:31.169] [memgraph_log] [debug] [Run] 'MATCH (APP_INTERNAL_EXEC_VAR) return COUNT(APP_INTERNAL_EXEC_VAR) as cnt'
[2023-07-28 14:51:34.049] [memgraph_log] [debug] [Run] 'SHOW STORAGE INFO'
[2023-07-28 14:51:34.155] [memgraph_log] [debug] [Run] 'MATCH (APP_INTERNAL_EXEC_VAR) return COUNT(APP_INTERNAL_EXEC_VAR) as cnt'
[2023-07-28 14:51:37.041] [memgraph_log] [debug] [Run] 'SHOW STORAGE INFO'
[2023-07-28 14:51:37.145] [memgraph_log] [debug] [Run] 'MATCH (APP_INTERNAL_EXEC_VAR) return COUNT(APP_INTERNAL_EXEC_VAR) as cnt'
[2023-07-28 14:51:40.044] [memgraph_log] [debug] [Run] 'SHOW STORAGE INFO'
[2023-07-28 14:51:40.181] [memgraph_log] [debug] [Run] 'MATCH (APP_INTERNAL_EXEC_VAR) return COUNT(APP_INTERNAL_EXEC_VAR) as cnt'
[2023-07-28 14:51:43.043] [memgraph_log] [debug] [Run] 'SHOW STORAGE INFO'
[2023-07-28 14:51:43.169] [memgraph_log] [debug] [Run] 'MATCH (APP_INTERNAL_EXEC_VAR) return COUNT(APP_INTERNAL_EXEC_VAR) as cnt'
[2023-07-28 14:51:45.331] [memgraph_log] [debug] [Run] 'MATCH (n)-[r]->(m)

RETURN n,r,m'
[2023-07-28 14:51:46.047] [memgraph_log] [debug] [Run] 'SHOW STORAGE INFO'

```


Here are the steps to reproduce:
1. Load the dataset. Here is cypherl:
```
CREATE INDEX ON :__mg_vertex__(__mg_id__);
CREATE (:__mg_vertex__:`User` {__mg_id__: 0, `name`: "katarina", `id`: "06daddb7-1657-4e6a-8a7d-b11ec6d2698f"});
CREATE (:__mg_vertex__:`Post` {__mg_id__: 2, `id`: "69626a62-ac56-4681-bc52-af6f28e97ffb", `content`: "new hello world!"});
CREATE (:__mg_vertex__:`User` {__mg_id__: 3, `name`: "karmen", `id`: "938ab1fd-f70a-4026-933c-9c83a11a8738"});
CREATE (:__mg_vertex__:`User` {__mg_id__: 4, `name`: "ante", `id`: "f4d847df-8c97-4c16-bd06-1141010e2961"});
CREATE (:__mg_vertex__:`User` {__mg_id__: 5, `name`: "buda", `id`: "e79e2279-df6f-4d52-9742-c588b26897c0"});
CREATE (:__mg_vertex__:`Post` {__mg_id__: 7, `id`: "c5c27035-c075-4c07-a96d-9e9f0e1b8ca9", `content`: "Another post!"});
CREATE (:__mg_vertex__:`Post` {__mg_id__: 8, `content`: "New post!"});
MATCH (u:__mg_vertex__), (v:__mg_vertex__) WHERE u.__mg_id__ = 0 AND v.__mg_id__ = 2 CREATE (u)-[:`HAS_POST`]->(v);
MATCH (u:__mg_vertex__), (v:__mg_vertex__) WHERE u.__mg_id__ = 3 AND v.__mg_id__ = 8 CREATE (u)-[:`HAS_POST`]->(v);
MATCH (u:__mg_vertex__), (v:__mg_vertex__) WHERE u.__mg_id__ = 4 AND v.__mg_id__ = 7 CREATE (u)-[:`HAS_POST`]->(v);
DROP INDEX ON :__mg_vertex__(__mg_id__);
MATCH (u) REMOVE u:__mg_vertex__, u.__mg_id__;
```
2. Run the following mutation:
```
mutation {
  deletePosts(where: {content: "New post!"}){
    nodesDeleted,
    relationshipsDeleted
  }
}
```
3. Try running `MATCH` queries in db (I reproduced it with  `MATCH (n)-[r]->(m) RETURN r, n, m` and `MATCH (n) RETURN n`

katarinasupe:
I tried out some basic queries and aggregation queries. I also tried to create, update and delete mutations. Let me know if something needs to be added. My test examples and dataset can be found in Notion.

src/memgraph.cpp
```diff
@@ -355,6 +356,12 @@ DEFINE_VALIDATED_string(query_modules_directory, "",
                           return true;
                         });
 
+// NOLINTNEXTLINE (cppcoreguidelines-avoid-non-const-global-variables)
+DEFINE_string(callable_mappings_path, "",
```
gitbuda, line 360: @gvolfing, we should rename this flag to `--query-callable-mappings-path` because we have these categories in the flags naming (e.g. storage related flags are prefixed with `storage`, auth-related flags are prefixed with `auth`)

src/utils/java_string_formatter.hpp
gvolfing: We could, but we do a bunch of in-place modifications on the string so a copy would be needed either way. We could catch by non-const lvalue& but that would change the semantics of the function. This function should only be used to do Java like string formatting, we will do it very rearly. I would like to keep the function signature more ergonomic rather then one-less-string-copy faster.

gitbuda (approved):

Issue #865, opened 2023-04-04
Memgraph freezes when processing transform, CPUs all @ 100


lucasprograms:
**Memgraph version**
2.6

**Environment**
docker (AWS EC2 ubuntu 20.04)

**Describe the bug**
memgraph will somewhat frequently get pegged at near 100% for all CPUs (see screenshot). This seems to happen when processing a batch of kafka messages for a transformation.

There is no indication in the logs of anything that has gone awry, except for the INFO log of Kafka consumer <consumer> is processing a batch, without the followup finished processing message.

If I set the kafka offsets back 100k or so and restart the docker container, it processes fine, leading me to believe that it is not an issue with the messages or the transformation.

**To Reproduce**
Sorry I can't provide more info/context here, e.g., in terms of reproducing the issue. Just hoping to get an idea of whether something like this has come up before, or how I could approach debugging the issue.

**Expected behavior**
CPU behaves normally, no freezing

**Logs**
Nothing in info logs. IIRC nothing in debug or trace logs when this happened before.

antejavor:
@lucasprograms, thanks for opening the issue. The initial discussion: https://discord.com/channels/842007348272169002/890969266600423504/1091190328943464558
I noticed that screenshots are missing, so copied them from discord: 

![image](https://user-images.githubusercontent.com/18125448/230038483-0c93efc9-3074-478f-a40f-e30155a36de1.png)

After restart of docker container: 
![image](https://user-images.githubusercontent.com/18125448/230038892-e2ce6042-1dd8-4a28-aea2-abf38bed9403.png)

antaljanosbenjamin:
@lucasprograms Thanks for the report! We haven't experienced anything like this before. Maybe could you please lower the log levels to `TRACE`? It can make memgraph run slower, so I advise you to not try this in production. Then you should see more information (like `Start transaction in stream <stream name>`, `Processing row in stream <stream name>`, `Executing query `<query>` in stream <stream name>` and similar log message.

As you value privacy, I would like to point that out these log messages can leak information about your dataset and configuration, so please be aware of that.

Also because of that I don't ask you to share your whole log file, but maybe could you please run Memgraph with `TRACE` log and share the relevant part of the log files with reducted information? We are not actually interested in the stream names or queries, more like which messages do you see when Memgraph starts to process a batch. That helps us narrow down the possible location of the issue.

chrischabot:
Ran into the same issue with bluej

[...]

PR #1019, opened 2023-06-16
Replicate Memgraph 100% CPU issue

antejavor:
The initial approach, try to do some concurrent writes, occasionally read and delete. 
Tried so far:

- [x] Connecting via mgconsole and running create, read, and detach delete queries in parallel. 
- [x] Add query module that reads and writes to DB
- [x] Run on docker memgraph - Release version

Next step: 

- [ ] Connect to awesome streams

Related to the: 
https://github.com/memgraph/memgraph/issues/865


gitbuda:
Closing because we figured out the issue in a completely different way + it's getting fixed under #1124

PR #1025, opened 2023-06-20
[master < T0957] Fix storage GC to handle IN_MEMORY_ANALYTICAL

### LINKED ISSUE #957: Fix `SHOW STORAGE INFO;` is showing wrong vertex count constantly (GC maybe has a bug)
After importing data from https://drive.google.com/file/d/1lnwPwQzdlMBlSiKItZuRC58GeD6dLPMT/view?usp=sharing
and running the following query
```
MATCH (p:Post)
WITH localDateTime() - p.indexedAt as duration, p
WITH duration.day as days, p
WHERE days > 2
DETATCH DELETE p;
```
Memory usage + the `SHOW STORAGE INFO;` doesn't show correct data.
Maybe that's only visible while other data is streaming in, the instructions on how to run the streaming importer follow:
```
install ts-node (and dependencies)
git clone https://github.com/memgraph/bluej
cd into dir
npm install
ts-node src/index.ts
```

Ignition:
This fix corrects the garbage collection of our in memory store to handle deletions made while in IN_MEMORY_ANALYTICAL mode. They will now be collected rather than leak.

[master < Task] PR

- [x] Write a release note here -> "Deletions made in IN_MEMORY_ANALYTICAL now handled correctly"

Closes #957

### DIFF
--- a/src/storage/v2/storage.cpp
+++ b/src/storage/v2/storage.cpp
@@ -585,6 +585,12 @@ Result<std::optional<VertexAccessor>> Storage::Accessor::DeleteVertex(VertexAcce
   CreateAndLinkDelta(&transaction_, vertex_ptr, Delta::RecreateObjectTag());
   vertex_ptr->deleted = true;
 
+  // Need to inform the next CollectGarbage call that there are some
+  // non-transactional deletions that need to be collected
+  if (transaction_.storage_mode == StorageMode::IN_MEMORY_ANALYTICAL) {
+    storage_->gc_full_scan_vertices_delete_ = true;
+  }
+
   return std::make_optional<VertexAccessor>(vertex_ptr, &transaction_, &storage_->indices_, &storage_->constraints_,
                                             config_, true);
 }
@@ -654,6 +660,12 @@ Result<std::optional<std::pair<VertexAccessor, std::vector<EdgeAccessor>>>> Stor
 
   CreateAndLinkDelta(&transaction_, vertex_ptr, Delta::RecreateObjectTag());
   vertex_ptr->deleted = true;
+
+  // Need to inform the next CollectGarbage call that there are some
+  // non-transactional deletions that need to be collected
+  if (transaction_.storage_mode == StorageMode::IN_MEMORY_ANALYTICAL) {
+    storage_->gc_full_scan_vertices_delete_ = true;
+  }
 
   return std::make_optional<ReturnType>(
       VertexAccessor{vertex_ptr, &transaction_, &storage_->indices_, &storage_->constraints_, config_, true},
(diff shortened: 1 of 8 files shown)


tests/e2e/analytical_mode/common.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+@pytest.fixture
+def connect(**kwargs) -> mgclient.Connection:
+    connection = mgclient.connect(host="localhost", port=7687, **kwargs)
+    connection.autocommit = True
+    yield connection
+    cursor = connection.cursor()
+    execute_and_fetch_all(cursor, "MATCH (n) DETACH DELETE n")
+    execute_and_fetch_all(cursor, "STORAGE MODE IN_MEMORY_TRANSACTIONAL;")
```
gvolfing, line 30: Shouldn't we create a couple of vertices/edges in the fixture? I might be missing something here, but I don't think we actually deleting anything since we are calling the `GC` on the empty skip-lists.
Am I missing something?

tests/e2e/analytical_mode/workloads.yaml
```diff
@@ -0,0 +1,14 @@
+analytical_mode_cluster: &analytical_mode_cluster
+  cluster:
+    main:
+      args: ["--bolt-port", "7687", "--log-level=TRACE"]
+      log_file: "analytical_mode.log"
+      setup_queries: []
```
gvolfing, line 6: If I was missing something in the previous comment, this could also be a place where we could create an initial graph in the database, but I think the fixture is a better place to do that in this case.

gvolfing (approved):
I have one concern about the e2e tests, other than that I think this is good to go. Great job!

tests/e2e/analytical_mode/common.py
Ignition: The test code has `MERGE (n) DELETE n;` which inserts a vertex and instantly deletes in.

src/storage/v2/storage.cpp
```diff
@@ -1709,11 +1747,35 @@ void Storage::CollectGarbage() {
       MG_ASSERT(edge_acc.remove(edge), "Invalid database state!");
     }
   }
+
+  // EXPENSIVE full scan, is only run if an IN_MEMORY_ANALYTICAL transaction involved any deletions
+  // TODO: implement a fast internal iteration inside the skip_list (to avoid unnecessary find_node calls),
+  //  accessor.remove_if([](auto const & item){ return item.delta == nullptr && item.deleted;});
```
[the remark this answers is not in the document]
Ignition: Added to the comment

src/storage/v2/storage.cpp
```diff
@@ -1718,11 +1756,37 @@ void Storage::CollectGarbage() {
       MG_ASSERT(edge_acc.remove(edge), "Invalid database state!");
     }
   }
+
+  // EXPENSIVE full scan, is only run if an IN_MEMORY_ANALYTICAL transaction involved any deletions
```
antoniofilipovic, line 1760: Did you benchmark how much time it takes to do a full scan? I have really no perspective on this how much time it takes

antoniofilipovic (commented):
The code seems fine, I am a little bit worried about the full scan for doing detach delete.

src/storage/v2/storage.cpp
Ignition: No, but it is a relative behaviour.
Unlike in transaction mode, where we have a precise collection of what transactions have completed + the deltas made on each of those transactions. A situations like 1'000'000's of vertices with only one deletion is fast.
Yet in analytical, it will require a full scan, 1'000'000's of vertices with only one deletion is slow compared to the transaction mode.

Quick testing, `UNWIND RANGE(1, 1000000) AS i CREATE ();` then `MERGE (n) DELETE n; FREE MEMORY;` under the two storage modes.

IN_MEMORY_TRANSACTIONAL : round trip in 0.010 sec
IN_MEMORY_ANALYTICAL : round trip in 0.560 sec

antoniofilipovic: Okay, this seems fine. If we people will complain we can iterate.

antoniofilipovic (approved):

PR #964, opened 2023-06-06
[master < ] Add python & cpp batching

antoniofilipovic:
[master < Task] PR
- [ ] Check, and update documentation if necessary
- [ ] Provide the full content or a guide for the final git message

To keep docs changelog up to date, one more thing to do:
- [ ] Write a release note here
- [ ] Tag someone from docs team in the comments

PR #1043, opened 2023-06-26
[master < T] Fix multiple include error

antoniofilipovic:
[master < Task] PR

Made an error in the last [PR](https://github.com/memgraph/memgraph/pull/964) as I didn't put inline on function in detail namespace. But when building MAGE there is a multiple definition error, inline is thus required.


andrejtonev (approved):
Looks good

PR #1046, opened 2023-06-27
[master <] Handle partial results on timeout

### LINKED ISSUE #848: Implement partial results on timeout
Query partial results on timeout or query streaming: when querying a particularly dense part of the graph for a single page of data (e.g. 10 results), sometimes a few results can be found easily, but the query times out looking for the remaining results. It would be beneficial to return the results that were found. For example, Elasticsearch supports this via the [allow_partial_search_results query parameter](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-search.html).

DRAFT FEATURE SPEC: https://www.notion.so/memgraph/Partial-Results-on-Timeout

Server-side implementation -> https://github.com/memgraph/memgraph/pull/1046
pymgclient -> https://github.com/memgraph/pymgclient/issues/56
mgconsole -> https://github.com/memgraph/mgconsole/issues/62

Ignition:
- [x] Honor `RUN`'s `tx_timeout`
  - [x] Auto transaction
  - [x] Explicit transaction
- [x] Differentiate why transaction was aborted
- [x] Communicate to client that timeout was hit (in a way that partial results can be displayed)
  - Note: no change on server required, checking clients to ensure they work as intended
- [x] Test & improve `pymgclient` -> moved under memgraph/pymgclient#56
- [x] JavaScript client works
- [x] Test & improve `mgconsole` -> moved under memgraph/mgconsole#62

[master < Task] PR
- [x] Provide the full content or a guide for the final git message --> "Support per transaction timeouts"

- [x] Write a release note here, including added/changed clauses --> "Support per transaction timeouts"

related #848


src/query/exceptions.hpp
```diff
@@ -112,16 +112,39 @@ class QueryRuntimeException : public QueryException {
   using QueryException::QueryException;
 };
 
+enum AbortReason : uint8_t {
+  NO_ABORT = 0,
+  TERMINATED = 1,
+  SHUTDOWN = 2,
+  TIMEOUT = 3,
+};
```
gitbuda, line 120: Could you give a short description what each `AbortReason` is + maybe mention when it happens in real life?

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -2921,7 +2921,10 @@ mgp_error mgp_proc_add_deprecated_result(mgp_proc *proc, const char *name, mgp_t
 int mgp_must_abort(mgp_graph *graph) {
   MG_ASSERT(graph->ctx);
   static_assert(noexcept(memgraph::query::MustAbort(*graph->ctx)));
-  return memgraph::query::MustAbort(*graph->ctx) ? 1 : 0;
+  auto const reason = memgraph::query::MustAbort(*graph->ctx);
+  // TODO: should we expose the reasons to the C API, ATM its a boolean response
```
gitbuda, line 2925: I think we should 🤔

src/query/interpreter.cpp
```diff
@@ -1252,7 +1282,10 @@ PreparedQuery Interpreter::PrepareTransactionQuery(std::string_view query_upper,
 
       in_explicit_transaction_ = true;
       expect_rollback_ = false;
-      metadata_ = GenOptional(metadata);
+      metadata_ = GenOptional(extras.metadata_pv);
+
+      auto const timeout = DetermineTxTimeout(extras.tx_timeout, interpreter_context_->config);
+      explicit_transaction_timer_ = timeout ? std::make_shared<utils::AsyncTimer>(timeout.value()) : nullptr;
```
Josipmrden, line 1288: so this works because of
`explicit operator bool() const { return value_.has_value(); }` ?

Josipmrden (commented):
MInor comments

include/_mgp.hpp
```diff
@@ -703,7 +703,7 @@ inline void proc_add_deprecated_result(mgp_proc *proc, const char *name, mgp_typ
   MgInvokeVoid(mgp_proc_add_deprecated_result, proc, name, type);
 }
 
-inline bool must_abort(mgp_graph *graph) { return mgp_must_abort(graph); }
+inline int must_abort(mgp_graph *graph) { return mgp_must_abort(graph); }
```
antoniofilipovic, line 706: This will be breaking change for API I think. It should be okay, but we need to note that somewhere.

include/mgp.hpp
```diff
@@ -1602,10 +1620,24 @@ inline Id::Id(int64_t id) : id_(id) {}
...
+    default:
+      break;
+  }
+  return AbortReason::NO_ABORT;
+}
 
 inline void Graph::CheckMustAbort() const {
-  if (MustAbort()) {
+  if (MustAbort() != AbortReason::NO_ABORT) {
+    // Can't add extra information to MustAbortException because of ABI stability
```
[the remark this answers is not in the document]
antoniofilipovic: Just to note, in MAGE we are not using on single place `CheckMustAbort`. I think there is no even test covering this in Memgraph. So in my opinion, we can break ABI to provide more information

src/query/interpreter.cpp
```diff
@@ -1018,10 +1018,26 @@ struct PullPlanVector {
...
+struct TxTimeout {
+  TxTimeout() = default;
+  explicit TxTimeout(double value) noexcept : value_{std::in_place, value} {
+    // validation
+    // - negative timeout makes no sense
+    // - zero timeout means no timeout
+    if (value_ <= 0.0) value_.reset();
+  };
+  explicit operator bool() const { return value_.has_value(); }
+  double value() const { return *value_; }
```
antoniofilipovic, line 1030: What if user sends value smaller than 0, then we do `reset` and on `value()` it will be garbage, right?

src/query/interpreter.cpp
```diff
@@ -1236,14 +1251,29 @@ Interpreter::Interpreter(InterpreterContext *interpreter_context) : interpreter_
   MG_ASSERT(interpreter_context_, "Interpreter context must not be NULL");
 }
 
-PreparedQuery Interpreter::PrepareTransactionQuery(std::string_view query_upper,
-                                                   const std::map<std::string, storage::PropertyValue> &metadata) {
+auto DetermineTxTimeout(std::optional<int64_t> tx_timeout, InterpreterConfig const &config) -> TxTimeout {
```
antoniofilipovic, line 1254: What is the reason for using `auto` + trailing return type? Shouldn't we use that only in some instances of lamdas usage?

src/query/exceptions.hpp
```diff
@@ -112,16 +112,45 @@ class QueryRuntimeException : public QueryException {
...
 // TransientError, i. e. client will be encouraged to retry execution because it
 // could succeed if executed again.
 class HintedAbortError : public utils::BasicException {
  public:
   using utils::BasicException::BasicException;
-  HintedAbortError()
-      : utils::BasicException(
-            "Transaction was asked to abort either because it was executing longer than time specified or another user "
-            "asked it to abort.") {}
+  HintedAbortError(AbortReason reason) : utils::BasicException(AsMsg(reason)), reason_{reason} {}
```
antoniofilipovic, line 134: Clang is complaining to me here: "Single-argument constructors must be marked explicit to avoid unintentional implicit conversions". I am not sure if we should add explicit, seems fine this way too.

tests/manual/js/transaction_timeout/auto-transaction.ts
```diff
@@ -0,0 +1,39 @@
...
+          undefined,
+          { timeout: 50 } // NOTE: with a short timeout
+        )
+        .records()
+        .pipe(finalize(() => {
+          session.close();
+          driver.close();
+        }))
+        .subscribe({
+          next: record => { },
```
antoniofilipovic, line 34: I would add here maybe assert that we now that something was piped back?

antoniofilipovic (changes requested):
Good job, I have just one comment that I think should be fixed, otherwise all good

include/mgp.hpp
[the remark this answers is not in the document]
Ignition: Sure, I can change the message. RE: ABI I was thinking about adding extra MustAbortException types. But thinking about it now I could just subclass it...that should be ABI safe.

src/memgraph.cpp
```diff
@@ -501,6 +501,25 @@ namespace memgraph::metrics {
 extern const Event ActiveBoltSessions;
 }  // namespace memgraph::metrics
 
+auto ToQueryExtras(memgraph::communication::bolt::Value const &extra) -> memgraph::query::QueryExtras {
```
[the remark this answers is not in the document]
Ignition: It is a habit of mine:
1) I like to use `auto` as it reduces the noise on the left hand side of the function declaration
2) I also like to add in the return type to make it easier on static analysis + compiler. (i.e. they do not need to inspect the function definition)

src/query/interpreter.cpp
```diff
@@ -1018,10 +1018,26 @@ struct PullPlanVector {
...
+    // validation
+    // - negative timeout makes no sense
+    // - zero timeout means no timeout
+    if (value_ <= 0.0) value_.reset();
+  };
+  explicit operator bool() const { return value_.has_value(); }
+  double value() const { return *value_; }
+
+ private:
+  std::optional<double> value_;
```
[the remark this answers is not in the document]
Ignition: I'll change to `std::chrono::milliseconds`

src/query/interpreter.cpp
Ignition: Yes. Even though it is explicit conversion, implicit bool conversion is allowed in `if`'s and ternary conditionals.

src/query/interpreter.cpp
Ignition: Yes...the contract is that you will test before fetching. Should I rename to `valueUnchecked`, `valueUnsafe` or something else? (+ document that need to test before fetch)

src/query/exceptions.hpp
Ignition: I'll make it explicit...see no reason not to.

Josipmrden (approved):
I have no more comments, approve from me. Only check with Antonio now

src/query/interpreter.cpp
antoniofilipovic: Maybe `valueUnsafe`? It sounds good to me

antoniofilipovic (approved):
After resolving the comment about valueUnsafe, it seems okay for me to merge! I will give approval so you don't need to ping me!

src/query/interpreter.cpp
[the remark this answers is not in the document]
Ignition: fixed in latest commit

src/query/interpreter.cpp
Ignition: fixed in latest commit

src/query/exceptions.hpp
Ignition: fixed in latest commit

PR #1052, opened 2023-06-28
Add error handling on py batched proc init

antoniofilipovic:
[master < Task] PR

I PR for batched procedures I was not handling errors properly from init of procedures. Errors can occur, for example on this [PR ](https://github.com/memgraph/mage/pull/209/) on MAGE errors occur on initialization if the file doesn't exist. I forgot to take that into account.

### DIFF
--- a/src/query/procedure/py_module.cpp
+++ b/src/query/procedure/py_module.cpp
@@ -22,6 +22,7 @@
 #include <string_view>
 
 #include "mg_procedure.h"
+#include "query/exceptions.hpp"
 #include "query/procedure/mg_procedure_helpers.hpp"
 #include "query/procedure/mg_procedure_impl.hpp"
 #include "utils/memory.hpp"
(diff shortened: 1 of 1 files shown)


Josipmrden (approved):
Seems good

src/query/procedure/py_module.cpp
```diff
@@ -1062,17 +1063,35 @@ void CallPythonCleanup(const py::Object &py_cleanup) {
...
   auto call = [&](py::Object py_graph) -> std::optional<py::ExceptionInfo> {
     py::Object py_args(MgpListToPyTuple(args, py_graph.Ptr()));
     if (!py_args) return py::FetchError();
     auto py_res = py_initializer.Call(py_graph, py_args);
-    return {};
+    if (!py_res) return py::FetchError();
+    return std::nullopt;
   };
 
+  std::optional<std::string> maybe_msg;
```
gitbuda, line 1083: Isn't a better name here `maybe_err`?

antoniofilipovic: Ah, I was consistent just with the rest of the file.

PR #1060, opened 2023-07-01
Add package/release/run.sh build v4 debian-10

gitbuda:
The motivation here is a need for Debian 10 builder image because Jepsen v0.3.0 still requires Debian 10 image (only node containers, e.g., `jepsen-n1`)

### DIFF
--- a/release/package/run.sh
+++ b/release/package/run.sh
@@ -159,6 +159,19 @@ case "$1" in
         fi
     ;;
 
+    build)
+      shift 1
+      if [[ "$#" -ne 2 ]]; then
+          print_help
+      fi
+      # in the vX format, e.g. v5
+      toolchain_version="$1"
+      os="$2"
+      cd "$SCRIPT_DIR/$os"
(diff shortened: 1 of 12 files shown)


release/package/run.sh
```diff
@@ -159,6 +159,19 @@ case "$1" in
...
+      shift 1
+      if [[ "$#" -ne 2 ]]; then
+          print_help
+      fi
+      # in the vX format, e.g. v5
+      toolchain_version="$1"
+      os="$2"
+      cd "$SCRIPT_DIR/$os"
+      # TODO(gitbuda): Just copy or install into the container relevant environment/os script.
+      docker build -f Dockerfile --build-arg TOOLCHAIN_VERSION="toolchain-$toolchain_version" -t "memgraph-builder:v4_$os" .
```
gitbuda, line 172: consider `-t "memgraph/memgraph-builder:tag"` because that can be pushed to DockerHub right away

PR #1061, opened 2023-07-01
Fix Jepsen replication test issue

gitbuda:
- [x] Memgraph seems to be crashing on `n1` (memgraph in the `main` replication mode) in both tests during the final cleanup (sometimes the final output just says "Analysis invalid!", but before in the log there is also "Test Crashed!")
- [x] There seems to be an issue with key checking (`-oStrictHostKeyChecking=no` required, not included inside the Jepsen code) -> `infrastructure` related, locally it's possible to fix the problem by manually connecting from the control other nodes (`ssh root@nX # WHERE X=n1, n2, n3, ...`) -> MOVED -> is going to be done on the infra side
- [x] Increase Jepsen timeout
- [x] `--replication-restore-state-on-startup=false` by default -> NOTE: breaking change because `main` was recovered while `replica`s were not recovered by default (in v2.8 release)
- [x] Fix single test run `lein run test --workload xyz`

Screenshot 2023-07-01 at 4 12 53 PM

### Leftovers and Additional Improvements
- [ ] Enable [on-disk transaction queue test](https://github.com/orgs/memgraph/projects/13?pane=issue&itemId=33418746)
- [x] Run tests individually
- [x] Make hard cleanup (from `run.sh`) after each test (replaced with `./run.sh cluster-cleanup`)
- [ ] Built Memgraph under Debian 10 build container
- [ ] Extend the `run.sh` to be able to run `memgraph` under `gdb` if required

### DIFF
--- a/.github/workflows/diff.yaml
+++ b/.github/workflows/diff.yaml
@@ -17,307 +17,307 @@ on:
       - "licenses/*"
 
 jobs:
-  community_build:
-    name: "Community build"
-    runs-on: [self-hosted, Linux, X64, Diff]
-    env:
-      THREADS: 24
-      MEMGRAPH_ENTERPRISE_LICENSE: ${{ secrets.MEMGRAPH_ENTERPRISE_LICENSE }}
-      MEMGRAPH_ORGANIZATION_NAME: ${{ secrets.MEMGRAPH_ORGANIZATION_NAME }}
+  # community_build:
+  #   name: "Community build"
+  #   runs-on: [self-hosted, Linux, X64, Diff]
+  #   env:
+  #     THREADS: 24
+  #     MEMGRAPH_ENTERPRISE_LICENSE: ${{ secrets.MEMGRAPH_ENTERPRISE_LICENSE }}
+  #     MEMGRAPH_ORGANIZATION_NAME: ${{ secrets.MEMGRAPH_ORGANIZATION_NAME }}
 
-    steps:
-      - name: Set up repository
-        uses: actions/checkout@v3
-        with:
-          # Number of commits to fetch. `0` indicates all history for all
-          # branches and tags. (default: 1)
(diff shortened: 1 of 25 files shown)


tests/jepsen/project.clj
```diff
@@ -5,7 +5,10 @@
             :url "https://github.com/memgraph/memgraph/blob/master/release/LICENSE_ENTERPRISE.md"}
   :main jepsen.memgraph.core
   :dependencies [[org.clojure/clojure "1.10.0"]
-                 [jepsen "0.2.1-SNAPSHOT"]
```
gitbuda, line 8: try back 0.2.1

gitbuda: seems to be the same

.github/workflows/diff.yaml
```diff
@@ -17,307 +17,306 @@ on:
       - "licenses/*"
 
 jobs:
-  community_build:
```
gitbuda, line 20: revert

tests/jepsen/resources/node-config.edn
```diff
@@ -1,6 +1,10 @@
 [{"n1" {:replication-role :main}
-  "n2" {:replication-role :replica :replication-mode :async :port 10000}
-  "n3" {:replication-role :replica :replication-mode :async :port 10000}}
```
gitbuda, line 3: I would restore this file back, because it's fine as is

tests/jepsen/src/jepsen/memgraph/support.clj
```diff
@@ -26,6 +26,32 @@
    :--storage-wal-enabled
    :--storage-snapshot-interval-sec 300
    :--storage-properties-on-edges))
+  ;; (if (= node "n1")
```
gitbuda, line 29: Delete all that

PR #1067, opened 2023-07-07
[master <] Bugfix: Throw if user was created without a license

andrejtonev:
[master < Task] PR

De-serializing the user/role will throw if the user/role was created without a valid license, but is being read with a valid license.
To fix this, we just default construct the fine_grained_access_handler and read from it if a valid one is found in the auth db.

@kgolubic


src/auth/models.cpp
```diff
@@ -594,10 +595,11 @@ User User::Deserialize(const nlohmann::json &data) {
...
   if (memgraph::license::global_license_checker.IsEnterpriseValidFast()) {
-    if (!data["fine_grained_access_handler"].is_object()) {
-      throw AuthException("Couldn't load user data!");
+    FineGrainedAccessHandler fine_grained_access_handler;
+    // We can have an empty fine_grained if the user was created without a valid license
+    if (data["fine_grained_access_handler"].is_object()) {
+      fine_grained_access_handler = FineGrainedAccessHandler::Deserialize(data["fine_grained_access_handler"]);
     }
-    auto fine_grained_access_handler = FineGrainedAccessHandler::Deserialize(data["fine_grained_access_handler"]);
     return {data["username"], data["password_hash"], permissions, fine_grained_access_handler};
```
as51340, line 603:
```suggestion
    return {data["username"], data["password_hash"], permissions, std::move(fine_grained_access_handler)};
```

as51340 (changes requested):

as51340 (approved):

vpavicic:
@andrejtonev - a release note please :)

andrejtonev:
@vpavicic 

Users created with the Community Edition caused an exception when read with the Enterprise Edition. This was due missing information that was not saved in the Community Edition, but was checked for in the Enterprise Edition.

vpavicic:
@andrejtonev can I use the release note i wrote above? The point is that now all the users you create in the community edition will remain when you upgrade to the  enterprise edition

andrejtonev:
@vpavicic 
Remain sounds like we used to delete them. They persisted, it's just that they didn't work.

Maybe:
Users created in the Community Edition remain valid after the instance is upgraded to an Enterprise Edition. The user's fine grained permission will default to none after the upgrade.

vpavicic:
I don't think fine-grained permissions can even be set in the community edition, you can only create a user with a password, that's it

andrejtonev:
Exactly, the fine grained permissions is the info missing that causes the failure after upgrading to Enterprise. Now, instead of throwing an error when the fine grained permissions are missing, we just default them to none and move on.

vpavicic:
Oh ok! I won't include that as I don't think users need that much info.. what's important here is that if you created users in the community, they will still be there when you upgrade :) thnx!

PR #1077, opened 2023-07-17
[master < T1077] Fix flaky tx-queue-multiple test

as51340:
[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR


antoniofilipovic (approved):
All seems good! Is the only issue here that we are not using memory ordering and then it presumably ends in some kind of deadlock?

as51340:
> All seems good! Is the only issue here that we are not using memory ordering and then it presumably ends in some kind of deadlock?

I think that deadlock doesn't happen but rather helper threads don't get scheduled because of main thread (which just waits)

PR #1080, opened 2023-07-19
[master < T1080] overload << operator for mgp::Type

mpintaric55334:
During writing collections query module, the need to be able to print the mgp::Type was found, so it was solved with this PR.

[master < Task] PR

You are now able to print the type of mgp::Type() using operator << . For example, if you have a mgp::List() list, cout<< list <<endl will output "list".

### DIFF
--- a/include/mgp.hpp
+++ b/include/mgp.hpp
@@ -395,8 +395,9 @@ class List {
 
   /// @brief Returns the value at the given `index`.
   const Value operator[](size_t index) const;
+
   /// @brief Returns type as string.
-  const std::string TypeString();
+  const std::string TypeString() const;
 
   class Iterator {
    private:
@@ -492,9 +493,6 @@ class Map {
   /// @brief Returns the value at the given `key`.
   Value const At(std::string_view key) const;
 
-  /// @brief Returns type as string.
-  const std::string TypeString();
-
   class Iterator {
    public:
     friend class Map;
@@ -544,6 +542,8 @@ class Map {
   bool operator==(const Map &other) const;
   /// @exception std::runtime_error Map contains value of unknown type.
   bool operator!=(const Map &other) const;
+  /// @brief Returns type as string.
+  const std::string TypeString() const;
 
  private:
   mgp_map *ptr_;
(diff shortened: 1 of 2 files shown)


include/mgp.hpp
```diff
@@ -3185,6 +3226,44 @@ inline bool Value::IsDuration() const { return mgp::value_is_duration(ptr_); }
...
+    case Type::Date:
+      return "Type: Date";
+    case Type::LocalTime:
+      return "Type: LocalTime";
+    case Type::LocalDateTime:
+      return "Type: LocalDateTime";
+    case Type::Duration:
+      return "Type: Duration";
+    default:
+      return "Type: Unknown";
```
antoniofilipovic, line 3263: can you throw an error here?

include/mgp.hpp
```diff
@@ -3185,6 +3226,44 @@ inline bool Value::IsDuration() const { return mgp::value_is_duration(ptr_); }
 inline bool Value::operator==(const Value &other) const { return util::ValuesEqual(ptr_, other.ptr_); }
 
 inline bool Value::operator!=(const Value &other) const { return !(*this == other); }
+
+inline const std::string Value::TypeString() const {
+  const mgp::Type &valueType = Type();
+
+  switch (valueType) {
+    case Type::Null:
+      return "Type: Null";
```
antoniofilipovic, line 3235: Maybe it makes sense to just return Null, Node, etc.. so user can decide on what kind of error message will be written after that

antoniofilipovic: So instead of Type::Null, return just Null, etc. Does that make more sense?

antoniofilipovic (changes requested):
I would change that we return "<type>" instead of "type: <type>", otherwise looks good

antoniofilipovic:
@mpintaric55334 there is one other thing I haven't thought about until now. We can support instead of TypeString, operator<< which serves for output: https://learn.microsoft.com/en-us/cpp/standard-library/overloading-the-output-operator-for-your-own-classes?view=msvc-170. This is one example, what do you think?

include/mgp.hpp
mpintaric55334: Yeah, will do

mpintaric55334:
@vpavicic will update the docs when I get the green light from Antonio.

antoniofilipovic (approved):
Looks good to go

vpavicic:
Is this also C++ ?

mpintaric55334:
Yes, it is

PR #1082, opened 2023-07-20
[master < T1082] Jepsen - pause nemesis work until replica registration is done

as51340:
The feature can be considered complete if:
- the first action nemesis does is indeed after the timeout specified.  :white_check_mark: 
- without proper timeout replica gets killed while it was being restored (only it the nemesis action was kill not network partitioning) :white_check_mark: 
- check the test fails when the action kill-node happens during the replica registration :white_check_mark: 

[master < Task] PR


tests/jepsen/src/jepsen/memgraph/nemesis.clj
```diff
@@ -26,17 +26,28 @@
...
-       (gen/phases (gen/sleep 60))))
+  ))
+
+
+(defn full-generator
+  "Construct nemesis generator."
+  [opts]
+  (gen/phases
+   (gen/log "Waiting replicas to get data from main.")
+   (gen/sleep 40)
```
Darych, line 46: Willl we wait 40 seconds on each phase or just before the 1st phase?

Darych (approved):

tests/jepsen/src/jepsen/memgraph/nemesis.clj
as51340: it should be just before the 1st phase

as51340:
I am quite sure that this is good setup of nemesis but then Jepsen tests occasionally fail. Not sure if we should merge then

vpavicic:
@as51340 do we need a RN?

gitbuda:
In this case no because it's all internal stuff + testing, I would skip this one from the changelog perspective

PR #1085, opened 2023-07-20
[master < T1085] Remove double scan with expand from the planner

Josipmrden:
Possibly it is not needed to do double ScanAll in the operator tree but expand from already existing node.

@vpavicic An optimization to the query engine was added in order to prefer at all times scanning + expansion of nodes instead of scanning both source and destination vertices and then expanding to the edge between them.
No need for documentation updates.

[master < Task] PR


Ignition:
@Josipmrden Looking at Grafana dashboard it does drop throughput on some of the mgbench workloads. I think we need to understand this some more.

The `expand.common_.existing_node` looks like it should prevents the change in the plan to the double scan. Maybe we should check why that isn't there in our case.

gitbuda:
Seems like a good update, but not good results on the mgbench -> investigation needed

Josipmrden:
@gitbuda   After some talk with @Ignition  we see there are no performance degradations on the benchmarks. We will proceed with the removal of the feature. Also, if the need occurs again, we will re-add the optimization with the appropriate benchmark tests that will point whether we have degraded any queries.

Ignition (approved):

PR #1090, opened 2023-07-21
[master < T1090] Add overload for operator< in mgp::Value

mpintaric55334:
Before, when writing C++ query moduels, directly comparing two mgp::Value variables was not possible. With this PR, it is.

[master < Task] PR

You are now able to compare two mgp::Value variables with < operator.

### DIFF
--- a/include/mgp.hpp
+++ b/include/mgp.hpp
@@ -3197,42 +3197,31 @@ inline bool Value::operator<(const Value &other) const {
   switch (type) {
     case Type::Null:
       throw ValueException("Cannot compare Null types");
-      break;
     case Type::Bool:
       return ValueBool() < other.ValueBool();
-      break;
     case Type::Int:
       return ValueNumeric() < other.ValueNumeric();
-      break;
     case Type::Double:
       return ValueNumeric() < other.ValueNumeric();
-      break;
     case Type::String:
       return ValueString() < other.ValueString();
-      break;
     case Type::Node:
       return ValueNode() < other.ValueNode();
-      break;
     case Type::Relationship:
       return ValueRelationship() < other.ValueRelationship();
-      break;
     case Type::Date:
       return ValueDate() < other.ValueDate();
-      break;
     case Type::LocalTime:
       return ValueLocalTime() < other.ValueLocalTime();
-      break;
     case Type::LocalDateTime:
       return ValueLocalDateTime() < other.ValueLocalDateTime();
-      break;
     case Type::Duration:
       return ValueDuration() < other.ValueDuration();
-      break;
     default:
       if (IsPath() || IsList() || IsMap()) {
         throw ValueException("Operator < is not defined for this data type");
       } else {
-        return false;
+        throw ValueException("Undefined behaviour");
       }
   }
 }
--- a/include/mgp.hpp
+++ b/include/mgp.hpp
@@ -3189,32 +3189,51 @@ inline bool Value::operator==(const Value &other) const { return util::ValuesEqu
 inline bool Value::operator!=(const Value &other) const { return !(*this == other); }
 
 inline bool Value::operator<(const Value &other) const {
-  if (Type() != other.Type()) {
+  const mgp::Type &type = Type();
+  if (type != other.Type() && !(IsNumeric() && other.IsNumeric())) {
     throw ValueException("Values have to be of the same type");
-  } else if (IsNull()) {
-    throw ValueException("Cannot compare Null types");
-  } else if (IsPath() || IsList() || IsMap()) {
-    throw ValueException(TypeString() + " doesn't have < operator defined");
-  } else if (IsBool()) {
-    return ValueBool() < other.ValueBool();
-  } else if (IsNumeric()) {
-    return ValueNumeric() < other.ValueNumeric();
-  } else if (IsString()) {
-    return ValueString() < other.ValueString();
-  } else if (IsNode()) {
-    return ValueNode() < other.ValueNode();
-  } else if (IsRelationship()) {
-    return ValueRelationship() < other.ValueRelationship();
-  } else if (IsDate()) {
-    return ValueDate() < other.ValueDate();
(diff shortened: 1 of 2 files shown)


include/mgp.hpp
```diff
@@ -3185,6 +3187,55 @@ inline bool Value::IsDuration() const { return mgp::value_is_duration(ptr_); }
...
+      break;
+    case Type::Duration:
+      return ValueDuration() < other.ValueDuration();
+      break;
+    default:
+      if (IsPath() || IsList() || IsMap()) {
+        throw ValueException("Operator < is not defined for this data type");
+      } else {
+        return false;
+      }
```
antoniofilipovic, line 3236: Why do you need this else?

tests/unit/cpp_api.cpp
```diff
@@ -471,3 +471,27 @@ TYPED_TEST(CppApiTestFixture, TestNodeProperties) {
   ASSERT_EQ(node_1.Properties()["b"].ValueString(), "b");
   ASSERT_EQ(node_1.GetProperty("b").ValueString(), "b");
 }
+
+TYPED_TEST(CppApiTestFixture, TestValueOperatorLessThan) {
+  const int64_t int1 = 3;
+  const int64_t int2 = 4;
+  const double double1 = 3.5;
+  const mgp::Value &int_test1 = mgp::Value(int1);
```
antoniofilipovic, line 479: You don't need & here, as you are doing assignment, and I am not sure to what is that reference binding, probably to temporary variable and that could result in undefined behaviour

tests/unit/cpp_api.cpp
```diff
@@ -471,3 +471,27 @@ TYPED_TEST(CppApiTestFixture, TestNodeProperties) {
...
+  const mgp::Value &int_test2 = mgp::Value(int2);
+  const mgp::Value &double_test1 = mgp::Value(double1);
```
antoniofilipovic, line 481: Also here you don't need references

tests/unit/cpp_api.cpp
```diff
@@ -471,3 +471,27 @@ TYPED_TEST(CppApiTestFixture, TestNodeProperties) {
...
+
+  ASSERT_TRUE(int_test1 < int_test2);
+  ASSERT_TRUE(double_test1 < int_test2);
+
+  const std::string_view string1 = "string";
```
antoniofilipovic, line 486: You can use std::string here, as we are constructing it

tests/unit/cpp_api.cpp
```diff
@@ -471,3 +471,27 @@ TYPED_TEST(CppApiTestFixture, TestNodeProperties) {
...
+  const mgp::Value &string_test1 = mgp::Value(string1);
```
antoniofilipovic, line 487: No need for reference here

tests/unit/cpp_api.cpp
```diff
@@ -471,3 +471,27 @@ TYPED_TEST(CppApiTestFixture, TestNodeProperties) {
...
+  bool exceptionThrown = false;
+
+  try {
+    bool Result = int_test1 < string_test1;
+  } catch (const std::exception &e) {
+    exceptionThrown = true;
+  }
+
+  ASSERT_TRUE(exceptionThrown);
+}
```
antoniofilipovic, line 497: For this we have actually assert_throws, check for that in code in memgraph, just search assert_throws in VsCode:

`ASSERT_THROW(user->UpdatePassword(), AuthException);
`

antoniofilipovic (changes requested):
Few stuff and it should be good to ship

include/mgp.hpp
mpintaric55334: I left it in case someone tries something I oversighted or is undefined in compariosn to return false, but will remove it for now, there shouldnt be undefined behaviour

mpintaric55334: And without it I get error, because its a non-void function that has to return something, so i will remove false and rasie exception

include/mgp.hpp
```diff
@@ -3185,6 +3187,44 @@ inline bool Value::IsDuration() const { return mgp::value_is_duration(ptr_); }
...
+    case Type::LocalDateTime:
+      return ValueLocalDateTime() < other.ValueLocalDateTime();
+    case Type::Duration:
+      return ValueDuration() < other.ValueDuration();
+    default:
+      if (IsPath() || IsList() || IsMap()) {
+        throw ValueException("Operator < is not defined for this data type");
+      } else {
+        throw ValueException("Undefined behaviour");
+      }
```
antoniofilipovic, line 3225:
```suggestion
    case Type::Path:
    case Type::List:
    case Type::Map:
        throw ValueException("Operator < is not defined for this Path, List or Map data type");
     default:
        throw ValueException("Undefined behaviour");
```

You can use this concept of fail through in switch case, and then default for this error message. This is only comment left from me

tests/unit/cpp_api.cpp
```diff
@@ -471,3 +471,20 @@ TYPED_TEST(CppApiTestFixture, TestNodeProperties) {
...
+  const mgp::Value int_test2 = mgp::Value(int2);
+  const mgp::Value double_test1 = mgp::Value(double1);
+
+  ASSERT_TRUE(int_test1 < int_test2);
+  ASSERT_TRUE(double_test1 < int_test2);
+
+  const std::string string1 = "string";
+  const mgp::Value string_test1 = mgp::Value(string1);
+
+  ASSERT_THROW(int_test1 < string_test1, mgp::ValueException);
```
antoniofilipovic, line 489: Can you also add that it throws an error if you compare `Map` and `Int`, or something like that?

antoniofilipovic (changes requested):
Few comments left and that should be it

mpintaric55334:
@vpavicic I will start writing docs after green light from Antonio.

antoniofilipovic (approved):
Looks good to go

PR #1092, opened 2023-07-21
[master < T1092] Run jepsen tests always on nightcrawler

as51340:
Recovering replicas takes variable amount of time on different machines. In Jepsen tests we rely on a fact that all replicas should be prepared before nemesis starts killing them. Hence, I think it makes sense to run distributed tests on a single machine.

TODO:
- here I need to add config that then all other tasks (except Jepsen) shouldn't run on nightcrawler. Otherwise, Jepsen tests wait until all other tasks are finished.

[master < Task] PR

### DIFF
--- a/.github/workflows/diff.yaml
+++ b/.github/workflows/diff.yaml
@@ -319,7 +319,7 @@ jobs:
 
   release_jepsen_test:
     name: "Release Jepsen Test"
-    runs-on: [self-hosted, Linux, X64, Debian10, JepsenControl]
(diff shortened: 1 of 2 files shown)


gitbuda:
@as51340 should we close this?

as51340:
> @as51340 should we close this?

Depends what you and others think.

as51340:
Timeouts are tricky and we have to improve that but it's not a solution to run jepsen on a single machine

PR #1093, opened 2023-07-21
[master < T570] Implement hash on mgp::Value

imilinovic:
[master < Epic] PR
- [x] Write E2E tests

[master < Task] PR


include/mgp.hpp
```diff
@@ -19,7 +19,9 @@
 #include <vector>
 
 #include "_mgp.hpp"
+#include "mg_exceptions.hpp"
 #include "mg_procedure.h"
+#include "utils/fnv.hpp"
```
imilinovic, line 24: There is an issue that when running the MAGE setup script this need to be changed to 
`#include "../src/utils/fnv.hpp"`
There should be a more elegant way to fix this but I am not well versed in Cmake

include/mgp.hpp
```diff
@@ -1467,6 +1469,20 @@ inline bool ValuesEqual(mgp_value *value1, mgp_value *value2) {
   if (value1 == value2) {
     return true;
   }
+  // Make int and double comparable, (ex. this is true -> 1.0 == 1)
+  if ((mgp::value_get_type(value1) == MGP_VALUE_TYPE_INT || mgp::value_get_type(value1) == MGP_VALUE_TYPE_DOUBLE) &&
```
antoniofilipovic, line 1473: Can you isolate this check if the value is numeric in some function? So if mgp::value_get_type returns double or int we have numeric, otherwise not

include/mgp.hpp
```diff
@@ -1467,6 +1469,20 @@ inline bool ValuesEqual(mgp_value *value1, mgp_value *value2) {
...
+      (mgp::value_get_type(value2) == MGP_VALUE_TYPE_INT || mgp::value_get_type(value2) == MGP_VALUE_TYPE_DOUBLE)) {
+    double val1{NAN};
+    if (mgp::value_get_type(value1) == MGP_VALUE_TYPE_INT)
```
antoniofilipovic, line 1476:
```
if () {

}
```

include/mgp.hpp
```diff
@@ -1467,6 +1469,20 @@ inline bool ValuesEqual(mgp_value *value1, mgp_value *value2) {
...
+      val1 = static_cast<double>(mgp::value_get_int(value1));
+    else
```
antoniofilipovic, line 1478: same here for `if`

include/mgp.hpp
```diff
@@ -3547,4 +3563,60 @@ template <>
...
+        const auto &path = value.ValuePath();
+        for (size_t i = 0; i < path.Length() - 1; ++i) {
+          hash *= fnv_prime;
+          hash ^= std::hash<mgp::Node>{}(path.GetNodeAt(i));
+          hash *= fnv_prime;
+          hash ^= std::hash<mgp::Relationship>{}(path.GetRelationshipAt(i));
+        }
+        hash *= fnv_prime;
+        hash ^= std::hash<mgp::Node>{}(path.GetNodeAt(path.Length()));
+        return hash;
```
antoniofilipovic, line 3608: Can you isolate this in function of some sort, or at least this part in for loop, so you have function hashPathValue which returns hash

antoniofilipovic (changes requested):
Few changes and should be good to go. Fix only this what we talked about that we will copy what we need from utils/fnv.hpp inside mgp namespace

include/mgp.hpp
```diff
@@ -3547,4 +3563,60 @@ template <>
...
+      }
```
antepusic, line 3609: Analogously to these hashes, I think there should also be specializations of `hash` for `mgp::List`, `mgp::Map` and `mgp::Path` (see above).

antepusic (changes requested):

include/mgp.hpp
imilinovic: You mean to write different hash functions for each of those instead of using FnvCollection hashing?

imilinovic: Done, but the order of hashes had to be changed a bit since List hash requires Value hash.

include/mgp.hpp
imilinovic: fnv.hpp was copy pasted inside mgp::util so MAGE doesn't have any dependacies.

antoniofilipovic (approved):
All seems good now to me

antepusic (approved):
I think it’s all good too!

vpavicic:
@imilinovic - a release note please! what's new here?

imilinovic:
@vpavicic Done

PR #982, opened 2023-06-14
[master < T982] Add mgp any type to C++ query module wrapper

Josipmrden:
This is a C++ query module enhancement that allows users to write C++ query modules with arguments of variable type. So far only C API was able to do this, and the wrapper did not account for the Any type. This is an extension that allows for this misalignment.

[master < Task] PR
- [ ] Check, and update documentation if necessary
- [x] Provide the full content or a guide for the final git message

To keep docs changelog up to date, one more thing to do:
- [X] Write a release note here 
You are now able to use `mgp::Type::Any` in C++ query modules to specify that the argument of the query modules can be of any type.

- [X] Tag someone from docs team in the comments

PR #1094, opened 2023-07-21
[master < T1094] Extend insert on record to accept mgp::Value

mpintaric55334:
This is an additional overload of Record::Insert function. While writing collections query module in  C++ API, the need for returning mgp::Any datatype in AddProcedure was discovered. The changes in this PR adress this problem, and by overloading Record::Insert to accept mgp::Value, mgp::Any datatype can be inserted into Record.

[master < Task] PR

You are now able to return mgp::Type::Any inside Record when writing query modules in C++ API.

### DIFF
--- a/include/mgp.hpp
+++ b/include/mgp.hpp
@@ -3281,44 +3281,31 @@ inline void Record::Insert(const char *field_name, const Duration &duration) {
 inline void Record::Insert(const char *field_name, const Value &value) {
   switch (value.Type()) {
     case Type::Bool:
-      Insert(field_name, value.ValueBool());
-      break;
+      return Insert(field_name, value.ValueBool());
     case Type::Int:
-      Insert(field_name, value.ValueInt());
-      break;
+      return Insert(field_name, value.ValueInt());
     case Type::Double:
-      Insert(field_name, value.ValueDouble());
-      break;
+      return Insert(field_name, value.ValueDouble());
     case Type::String:
-      Insert(field_name, value.ValueString());
-      break;
+      return Insert(field_name, value.ValueString());
     case Type::List:
-      Insert(field_name, value.ValueList());
-      break;
+      return Insert(field_name, value.ValueList());
     case Type::Map:
-      Insert(field_name, value.ValueMap());
-      break;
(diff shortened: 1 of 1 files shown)


antoniofilipovic (approved):
I think this should be fine :smile: @Josipmrden please check this one and check if we can test it somehow

include/mgp.hpp
```diff
@@ -3276,6 +3278,53 @@ inline void Record::Insert(const char *field_name, const Duration &duration) {
   mgp::value_destroy(mgp_val);
 }
 
+inline void Record::Insert(const char *field_name, const Value &value) {
+  switch (value.Type()) {
+    case Type::Bool:
+      Insert(field_name, value.ValueBool());
+      break;
```
antoniofilipovic, line 3285: Actually you can combine these two into return Insert, so please change that I gave approve too soon :joy:

antoniofilipovic (changes requested):
Actually needs change, small change

include/mgp.hpp
mpintaric55334: Done

antoniofilipovic (approved):
Good work!

Josipmrden:
@mpintaric55334 Please edit the PR message, and add an appropriate message why was this added (reasoning.)

mpintaric55334:
Implemented Record::Insert for mgp::Value, because it is necessary if you set mgp::Any as return type in AddProcedure

Josipmrden:
https://github.com/memgraph/memgraph/pull/982 @mpintaric55334  here is an example of how it should look like, You should only leave the master < task with the checkboxes and tick all that you have done + write a release note message so @vpavicic can reason why was this added

mpintaric55334:
@vpavicic I will start writing documentation when I get the green light and instructions how to do that from Antonio or Josip.

vpavicic:
> @vpavicic I will start writing documentation when I get the green light and instructions how to do that from Antonio or Josip.

OK... I can give you instructions how to write the docs as well, but whatever you prefer :)

Josipmrden (approved):

PR #1095, opened 2023-07-22
[E952 < ] Database authorization speedup

andrejtonev:
Currently for each query we check if the user has access to it. We should check only on db changes or if the query is ran against a different db.


andrejtonev:
I will close this, since the slowdown isn't directly related to the 952 epic.
The fix will be moved to a new PR and merged into master.

PR #1099, opened 2023-07-24
[master < T1099] Add non-const mgp::Value return type for C++ API

mpintaric55334:
Using functions defined in mgp::Value, that return the Type of the value, such as ValueList, that extracted mgp::List from mgp::Value, only returned const types. While writing collections query module, there was need to return non const variables so move semantics could be used, this PR does exactly that.
[master < Task] PR

You are now able to extract non const data from mgp::Value @antoniofilipovic could you help me phrase this better, I dont think this is good.

### DIFF
--- a/include/mgp.hpp
+++ b/include/mgp.hpp
@@ -396,6 +396,9 @@ class List {
   /// @brief Returns the value at the given `index`.
   const Value operator[](size_t index) const;
 
+  ///@brief Same as above, but non const value
+  Value operator[](size_t index);
+
   class Iterator {
    private:
     friend class List;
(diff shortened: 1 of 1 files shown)


mpintaric55334:
@antoniofilipovic can you help me phrase the description better. @vpavicic I will write the documentation when I get the green light and know how from Antonio.

antoniofilipovic (approved):
Looks good to go

antoniofilipovic:
@mpintaric55334 @vpavicic for release note: Extend mgp::Value to return subtypes which are modifieable (non-const)

vpavicic:
@mpintaric55334  C++ API?

PR #1111, opened 2023-07-26
[master < T1111] Fix local buffer serialization

as51340:
[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR


tests/unit/storage_rocks.cpp
```diff
@@ -9,111 +9,109 @@
...
-//   auto gid = edge->Gid();
-//   auto ser_result = memgraph::utils::SerializeEdge(vertex1.Gid(), vertex2.Gid(), edge->EdgeType(), edge->edge_.ptr);
-//   ASSERT_EQ(ser_result.first,
-//             memgraph::utils::SerializeIdType(vertex1.Gid()) + "|" + memgraph::utils::SerializeIdType(vertex2.Gid()) +
-//                 "|0|" + std::to_string(edge->EdgeType().AsInt()) + "|" + memgraph::utils::SerializeIdType(gid));
-//   ASSERT_EQ(ser_result.second,
-//             memgraph::utils::SerializeIdType(vertex2.Gid()) + "|" + memgraph::utils::SerializeIdType(vertex1.Gid()) +
-//                 "|1|" + std::to_string(edge->EdgeType().AsInt()) + "|" + memgraph::utils::SerializeIdType(gid));
-// }
+TEST_F(RocksDBStorageTest, SerializeVertexGIDLabels) {
```
Darych, line 63: Should we also add tests for `SerializeVertexGIDLabelsOneProperty` and `SerializeVertexGIDLabelsManyProperties` ?

src/storage/v2/property_store.cpp
```diff
@@ -1292,12 +1292,11 @@ void PropertyStore::SetBuffer(const std::string_view buffer) {
 
   uint64_t size = 0;
   uint8_t *data = nullptr;
+  size = buffer.size();
```
Darych, line 1295: Good catch! Does it mean that in loop on 1304 line we lost last byte value?

tests/unit/storage_rocks.cpp
```diff
@@ -9,111 +9,109 @@
...
+}
+
+TEST_F(RocksDBStorageTest, SerializePropertiesExternalBuffer) {
+  memgraph::storage::PropertyStore props;
+  auto id = memgraph::storage::PropertyId::FromInt(0);
+  auto id_value = memgraph::storage::PropertyValue(1);
+  auto completion_percentage = memgraph::storage::PropertyId::FromInt(1);
+  auto completion_percentage_value = memgraph::storage::PropertyValue(14);
+  auto gender = memgraph::storage::PropertyId::FromInt(2);
+  auto gender_value = memgraph::storage::PropertyValue("man167863816386826");
```
Darych, line 103: Maybe we can add a comment or empty lines to emphasize that because of this line we start using external buffer, what do you think? Because on the first glance it was hard to see the difference with previous test.

tests/unit/storage_rocks.cpp
as51340: I mean we can but I don't see what we would get more since properties are serialized as value and labels as key so I think that we wouldn't get nothing more than we have.

src/storage/v2/property_store.cpp
as51340: Yes and that was edge case :bug: :lady_beetle:

tests/unit/storage_rocks.cpp
as51340: Yes sure

tests/unit/storage_rocks.cpp
Darych: We will be sure that properties are listed through commas as we designed. If something will be changed in serialization function we will catch it.

as51340: but properties are serialized just as with in-memory, no commas there?

as51340: Can you explain what do you mean?

Darych: Oops my bad. I messed with properties and labels. I meant to add multiple labels, not properties. You are right, properties are put as a serialized buffer.

Darych (approved):

vpavicic:
@as51340 release note or is it an internal fix?

as51340:
@vpavicic users actually have something out of it.
 Fixed bug when (de)serializing vertex and edge properties to RocksDB when the size of the serialization buffer was exactly 15B.

PR #1113, opened 2023-07-27
[master < ] Added warning missing-field-initialization flag and fixed code

andrejtonev:
Now there should be a warning when using {...} and not all variables have been initialized.
*if the member variable doesn't have a defined default value or the user isn't explicitly defining a subset (with .var_name = ...)


gitbuda (approved):

vpavicic:
@andrejtonev can you help me with a release note here or is this something internal?

andrejtonev:
This is just something internal that will (hopefully) help us create fewer
bugs.

PR #1115, opened 2023-07-27
[master < T1115] Improve performance on set properties

antoniofilipovic:
This PR enables batch update of properties which proved to improve performance when lot of properties are set.

Queries executed:
```
1.  FOREACH (i in range(0, 1000000) | CREATE (n:Label {id:i}));                              

2. CREATE INDEX ON :Label(id);                                                              

3. FOREACH (i IN range(0, 1000000 , 3) | MERGE (n:Label {id:i}) SET n += {prop2:"nmrtRnZQzm", prop3:"bielFtyXca", prop4:"AoCYiwqjuu", prop5:"GqyKUebYyi", prop6:"spdWPDdyiY", prop7:"dzpACFXOal", prop8:"NchfJxgMHL", prop9:"JKbOEGOhkl", prop10:"ftmaTyKESy", prop11:"cIgjUrbHcF"});

```

Before optimizations:
**Query 3: Empty set (round trip in 4.934 sec)**

After optimizations:
**Query 3: Empty set (round trip in 4.111 sec)**

Gain is somewhere between **10 and 20%** on set queries query.

[master < Task] PR


src/query/plan/operator.cpp
```diff
@@ -2782,44 +2785,34 @@ void SetPropertiesOnRecord(TRecordAccessor *record, const TypedValue &rhs, SetPr
...
-      set_props(get_props(rhs.ValueVertex()));
+    }
+    case TypedValue::Type::Vertex: {
+      PropertiesMap new_properties = get_props(rhs.ValueVertex());
+      update_props(new_properties);
       break;
+    }
     case TypedValue::Type::Map: {
+      PropertiesMap new_properties;
       for (const auto &kv : rhs.ValueMap()) {
```
as51340, line 2811:
```suggestion
      for (const auto &[prop_id, prop_value] : rhs.ValueMap()) {
```

src/query/plan/operator.cpp
```diff
@@ -2782,44 +2785,34 @@ void SetPropertiesOnRecord(TRecordAccessor *record, const TypedValue &rhs, SetPr
...
-            throw QueryRuntimeException("Unexpected error when setting properties.");
-        }
-      }
+  auto update_props = [&, record](PropertiesMap &new_properties) {
+    auto updated_properties = UpdatePropertiesChecked(record, new_properties);
 
-      if (should_register_change) {
-        register_set_property(std::move(*maybe_error), kv.first, std::move(kv.second));
+    if (should_register_change) {
+      for (auto &[id, old_value, new_value] : updated_properties) {
```
as51340, line 2792:
```suggestion
      for (const auto &[id, old_value, new_value] : updated_properties) {
```

src/storage/v2/edge_accessor.cpp
```diff
@@ -145,6 +147,44 @@ Result<bool> EdgeAccessor::InitProperties(const std::map<storage::PropertyId, st
   return true;
 }
 
+Result<std::vector<std::tuple<PropertyId, PropertyValue, PropertyValue>>> EdgeAccessor::UpdateProperties(
```
as51340, line 150: This method looks to me the same as for the vertex_accessor. Maybe from a code perspective it would be good to push it down to a common layer - maybe PropertyStore. If that is infeasible, then global util method is fine, as long as we remove duplication

src/storage/v2/edge_accessor.hpp
```diff
@@ -63,6 +63,9 @@ class EdgeAccessor final {
   /// @throw std::bad_alloc
   Result<bool> InitProperties(const std::map<storage::PropertyId, storage::PropertyValue> &properties);
 
+  Result<std::vector<std::tuple<PropertyId, PropertyValue, PropertyValue>>> UpdateProperties(
+      std::map<storage::PropertyId, storage::PropertyValue> &properties) const;
```
as51340, line 67: Just a note, no action needed: I saw that Google preferred for a long time pointers when modifying arguments and they used references only when they could be const. But they changed that so don't bother please

tests/unit/query_plan_create_set_remove_delete.cpp
```diff
@@ -1678,6 +1678,41 @@ TYPED_TEST(QueryPlanTest, SetPropertiesOnNull) {
...
 }
 
+TYPED_TEST(QueryPlanTest, UpdateSetPropertiesFromMap) {
+  auto storage_dba = this->db->Access();
+  memgraph::query::DbAccessor dba(storage_dba.get());
+  // Add a single vertex. ( {property: 43})
+  auto vertex_accessor = dba.InsertVertex();
+  auto old_value = vertex_accessor.SetProperty(dba.NameToProperty("property"), memgraph::storage::PropertyValue{43});
+  EXPECT_EQ(old_value.HasError(), false);
+  EXPECT_EQ(*old_value, memgraph::storage::PropertyValue());
```
as51340, line 1688: Shouldn't this fail? Shouldn't old_value be equal to PropertyValue(43)

as51340 (changes requested):
All good, just some minor comments. Do you have some benchmarks results?

tests/unit/query_plan_create_set_remove_delete.cpp
antoniofilipovic: No no, old value is empty, it didn't exist

src/query/plan/operator.cpp
antoniofilipovic: But then I can't move from const?

antoniofilipovic: It should fallback to copy if underlying objects are not mutable if I am not wrong.

src/storage/v2/edge_accessor.hpp
antoniofilipovic: I heard about that too, forgot about it when coding, but remembered it now. Cool stuff will remember that for next time.

src/storage/v2/vertex_accessor.cpp
```diff
@@ -254,6 +257,43 @@ Result<bool> VertexAccessor::InitProperties(const std::map<storage::PropertyId,
...
+  for (auto &[old_key, old_value] : old_properties) {
+    auto it = properties.emplace(old_key, old_value);
+    if (!it.second) {
+      auto &new_value = it.first->second;
+      id_old_new_change.emplace_back(std::make_tuple(it.first->first, old_value, new_value));
+    }
+  }
+
+  if (!vertex_->properties.InitProperties(properties)) return Error::SERIALIZATION_ERROR;
+  for (auto &[id, old_value, new_value] : id_old_new_change) {
```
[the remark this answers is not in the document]
as51340: yes sorry my bad

as51340 (approved):

Issue #909, opened 2023-05-03
SetProperties working poorly when there is a lot of indexes


Josipmrden:
**Memgraph version**
v2.7

**Environment**
Docker

**Describe the bug**
If there is 400 indexes in the database, SetProperties operator behaves poorly because it needs to update all the indexes. Update of indexes in the system is done sequentially via a for loop currently. There should be a smarter way to set properties and check which values need to be updated.

**To Reproduce**
1. Create 350 indexes
2. `FOREACH (i in range (1, 10000) | CREATE (:A1 {id: i}));` -> for every index
3. Try setting properties in bulk for 10 000 items

Expected:
It should be quick

Actual:
It is slow.

gitbuda:
TODO(@Josipmrden): Link map literal stuff

PR #1116, opened 2023-07-27
[master < T1116] Optimize index and constraint update

antepusic:
If there’s a large number of active indices, setting properties is slow and that’s evident when setting a large number of them (#909). This is due to the fact that UpdateOnSetProperty iterates over all indices in the database; this PR eliminates that.

[master < Task] PR

### DIFF
--- a/src/storage/v2/inmemory/label_property_index.hpp
+++ b/src/storage/v2/inmemory/label_property_index.hpp
@@ -21,7 +21,7 @@ struct LabelPropertyIndexStats {
 };
 
 /// TODO: andi. Too many copies, extract at one place
-using ParalellizedIndexCreationInfo =
+using ParallelizedIndexCreationInfo =
     std::pair<std::vector<std::pair<Gid, uint64_t>> /*vertex_recovery_info*/, uint64_t /*thread_count*/>;
 
 class InMemoryLabelPropertyIndex : public storage::LabelPropertyIndex {
@@ -43,7 +43,7 @@ class InMemoryLabelPropertyIndex : public storage::LabelPropertyIndex {
 
   /// @throw std::bad_alloc
   bool CreateIndex(LabelId label, PropertyId property, utils::SkipList<Vertex>::Accessor vertices,
-                   const std::optional<ParalellizedIndexCreationInfo> &paralell_exec_info);
+                   const std::optional<ParallelizedIndexCreationInfo> &parallel_exec_info);
 
   /// @throw std::bad_alloc
   void UpdateOnAddLabel(LabelId added_label, Vertex *vertex_after_update, const Transaction &tx) override;
--- a/src/storage/v2/disk/label_property_index.hpp
+++ b/src/storage/v2/disk/label_property_index.hpp
@@ -15,7 +15,7 @@
 namespace memgraph::storage {
 
 /// TODO: andi. Too many copies, extract at one place
-using ParalellizedIndexCreationInfo =
+using ParallelizedIndexCreationInfo =
     std::pair<std::vector<std::pair<Gid, uint64_t>> /*vertex_recovery_info*/, uint64_t /*thread_count*/>;
 
 class DiskLabelPropertyIndex : public storage::LabelPropertyIndex {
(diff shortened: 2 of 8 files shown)


src/storage/v2/inmemory/label_property_index.hpp
```diff
@@ -136,6 +136,8 @@ class InMemoryLabelPropertyIndex : public storage::LabelPropertyIndex {
  private:
   std::map<std::pair<LabelId, PropertyId>, utils::SkipList<Entry>> index_;
   std::map<std::pair<LabelId, PropertyId>, storage::LabelPropertyIndexStats> stats_;
+
+  std::map<LabelId, std::set<PropertyId>> index_exists_;
```
Josipmrden, line 140: some more appropriate name, index_exists_ more sounds like a method than the description of what is it, e.g. property_indices_by_label or similar

antepusic: Done

src/storage/v2/inmemory/unique_constraints.cpp
```diff
@@ -302,6 +308,9 @@ InMemoryUniqueConstraints::CreateConstraint(LabelId label, const std::set<Proper
     // be removed.
     constraints_.erase(constraint);
     return ConstraintViolation{ConstraintViolation::Type::UNIQUE, label, properties};
+  } else {
+    // Add the new constraint to the optimized structure only if there are no violations.
```
antepusic, line 312: NB: This is simpler than if the new constraint were added to the structures before checking for violations. In that case, the function would have to remove the constraint from both `constraints_` and `constraints_by_label_` if a violation was found.

src/query/frontend/ast/ast.hpp
```diff
@@ -1199,6 +1200,16 @@ class PropertyLookup : public memgraph::query::Expression {
...
+  };
+
+  memgraph::query::PropertyLookup::EvaluationMode evaluation_mode_ = EvaluationMode::GET_OWN_PROPERTY;
+  std::map<storage::PropertyId, storage::PropertyValue> empty_cache{};
+  std::map<storage::PropertyId, storage::PropertyValue> &cache_ = empty_cache;
+  // TODO replace the above two lines with e.g. optional cache_ and std::nullopt
+
   PropertyLookup *Clone(AstStorage *storage) const override {
     PropertyLookup *object = storage->Create<PropertyLookup>();
     object->expression_ = expression_ ? expression_->Clone(storage) : nullptr;
```
Josipmrden, line 1215: also add the cloning of the evaluation mode

src/query/frontend/ast/ast.hpp
```diff
@@ -1071,6 +1071,7 @@ class MapLiteral : public memgraph::query::BaseLiteral {
   }
 
   std::unordered_map<memgraph::query::PropertyIx, memgraph::query::Expression *> elements_;
+  std::unordered_set<int32_t> cacheable_property_lookups_{};
```
Josipmrden, line 1074: I think there is an override of the int to be some property type

src/query/interpret/eval.hpp
```diff
@@ -479,7 +479,20 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
       case TypedValue::Type::Null:
         return TypedValue(ctx_->memory);
       case TypedValue::Type::Vertex:
-        return TypedValue(GetProperty(expression_result_ptr->ValueVertex(), property_lookup.property_), ctx_->memory);
+        if (property_lookup.evaluation_mode_ == PropertyLookup::EvaluationMode::GET_ALL_PROPERTIES) {
+          if (property_lookup.cache_.size() == 0) {
```
Josipmrden, line 483: .empty()

src/query/frontend/ast/ast.hpp
Josipmrden: also update clone method

src/query/interpret/eval.hpp
```diff
@@ -479,7 +479,20 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
...
+            auto props = GetAllProperties(expression_result_ptr->ValueVertex());
+            property_lookup.cache_ = props;
+          }
+
+          if (!property_lookup.cache_.contains(ctx_->properties[property_lookup.property_.ix])) {
+            return TypedValue(ctx_->memory);
+          }
+
+          return TypedValue(property_lookup.cache_.at(ctx_->properties[property_lookup.property_.ix]), ctx_->memory);
```
Josipmrden, line 492: are there multiple property lookups here inside the map literal? I can't see where is it ensured that we share the cache between individual property lookups, maybe I don't see something

antepusic: The `cache_` member of PropertyLookup is the reference to MapLiteral cache (more precisely, the part that stores the cache for the identifier that's in the PropertyLookup). `property_lookup.cache_ = props;` updates the PropertyLookup cache, and that should be visible from inside MapLiteral because of referencing.

Issue #1118, opened 2023-07-27
[BUG] Map projection from NULL values


ZelvaMan:
**Memgraph version**
Memgraph 2.8.0

**Environment**
Using latest docker image 
connected using memgraph lab

**Describe the bug**
According to [documentation](https://memgraph.com/docs/memgraph/reference-guide/data-types#map-projection) map values should be null, when map_variable is null. But when i use All-properties selector it return error
![image](https://github.com/memgraph/memgraph/assets/31016933/2b456b25-ca5b-45c7-8fd2-a2b6e620282c)

**To Reproduce**
You can just run this query
```
OPTIONAL MATCH (node:DOESNT_EXIST)
RETURN node {.*}
```

**Expected behavior**
It should return null

**Logs**
screenshot is above in description

**Additional context**
Thanks for help. If this is indeed intended behavior, I would suggest updating docs to make in obvious.

Josipmrden:
@lomnido the PR associated with the issue is https://github.com/memgraph/memgraph/pull/1119 so you can track it :)

PR #1119, opened 2023-07-28
[master < T1119] Return NULL on map projection from a null value

antepusic:
This PR addresses Issue #1118, which is about making map projections from NULL values. In that case, Memgraph should return a NULL value without raising any exceptions.

[master < Task] PR
> Fix map projection behavior in case of null projection variable

> We’ve fixed a bug with [map projection](https://memgraph.com/docs/memgraph/reference-guide/data-types#map-projection). If you project from a variable that happens to be null, the map projection will have a null value instead of displaying an error.


src/query/interpret/eval.hpp
```diff
@@ -700,12 +700,14 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
...
 
-        if (maybe_all_properties_lookup.type() != TypedValue::Type::Map) {
-          throw QueryRuntimeException("Expected a map from AllPropertiesLookup, got {}.",
+        switch (maybe_all_properties_lookup.type())
+        case TypedValue::Type::Map:
+          all_properties_lookup = std::move(maybe_all_properties_lookup.ValueMap());
+        case TypedValue::Type::Null:
+          all_properties_lookup = std::move(maybe_all_properties_lookup.ValueNull());
+        default:
+          throw QueryRuntimeException("Expected a map or NULL from AllPropertiesLookup, got {}.",
```
Josipmrden, line 709: QueryRuntimeException is the currently message to the user. If you think this should never happen, then use LOG_FATAL to crash Memgraph. If you still believe it should be a message caused by user behaviour, then modify the message because user does not know what is AllPropertiesLookup since that's an implementation detail.

Josipmrden (changes requested):

src/query/interpret/eval.hpp
```diff
@@ -696,20 +696,27 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
 
     TypedValue::TMap result(ctx_->memory);
     TypedValue::TMap all_properties_lookup(ctx_->memory);
+
+    auto map_variable = literal.map_variable_->Accept(*this);
+    if (map_variable.type() == TypedValue::Type::Null) {
+      return TypedValue(ctx_->memory);
+    }
```
Josipmrden, line 703: I think we have IsNull()

Josipmrden: because it's typed value and it can be of different type

src/query/interpret/eval.hpp
antepusic: Fixed, it’s `LOG_FATAL` now.

Josipmrden (approved):

antepusic:

PR #1124, opened 2023-07-31
[master < ] Fixes for 100% percent cpu issues

Ignition:
[master < Task] PR

relates to #865

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,3 +19,16 @@ repos:
     rev: v13.0.0
     hooks:
       - id: clang-format
+#  - repo: local
+#    hooks:
+#      - id: clang-tidy
+#        name: clang-tidy
+#        description: Runs clang-tidy and checks for errors
+#        entry: python ./tools/pre-commit/clang-tidy.py
+#        language: python
+#        files: ^src/
+#        types: [c++, text]
+#        fail_fast: true
+#        require_serial: true
+#        args: [--compile_commands_path=build]
+#        pass_filenames: false
--- a/src/storage/v2/inmemory/indices_utils.hpp
+++ b/src/storage/v2/inmemory/indices_utils.hpp
@@ -14,6 +14,7 @@
 #include "storage/v2/mvcc.hpp"
 #include "storage/v2/transaction.hpp"
 #include "storage/v2/vertex.hpp"
+#include "storage/v2/vertex_info_helpers.hpp"
 #include "utils/spin_lock.hpp"
 #include "utils/synchronized.hpp"
 
@@ -156,62 +157,12 @@ inline bool AnyVersionHasLabelProperty(const Vertex &vertex, LabelId label, Prop
       });
 }
 
-// Helper function for iterating through label index. Returns true if this
-// transaction can see the given vertex, and the visible version has the given
-// label.
-inline bool CurrentVersionHasLabel(const Vertex &vertex, LabelId label, Transaction *transaction, View view) {
-  bool deleted = false;
-  bool has_label = false;
-  const Delta *delta = nullptr;
-  {
-    std::lock_guard<utils::SpinLock> guard(vertex.lock);
-    deleted = vertex.deleted;
-    has_label = utils::Contains(vertex.labels, label);
-    delta = vertex.delta;
-  }
-  ApplyDeltasForRead(transaction, delta, view, [&deleted, &has_label, label](const Delta &delta) {
-    switch (delta.action) {
-      case Delta::Action::REMOVE_LABEL: {
-        if (delta.label == label) {
-          MG_ASSERT(has_label, "Invalid database state!");
-          has_label = false;
-        }
-        break;
-      }
-      case Delta::Action::ADD_LABEL: {
-        if (delta.label == label) {
-          MG_ASSERT(!has_label, "Invalid database state!");
-          has_label = true;
-        }
-        break;
-      }
-      case Delta::Action::DELETE_DESERIALIZED_OBJECT:
-      case Delta::Action::DELETE_OBJECT: {
-        MG_ASSERT(!deleted, "Invalid database state!");
-        deleted = true;
-        break;
-      }
-      case Delta::Action::RECREATE_OBJECT: {
-        MG_ASSERT(deleted, "Invalid database state!");
-        deleted = false;
-        break;
-      }
-      case Delta::Action::SET_PROPERTY:
-      case Delta::Action::ADD_IN_EDGE:
-      case Delta::Action::ADD_OUT_EDGE:
-      case Delta::Action::REMOVE_IN_EDGE:
-      case Delta::Action::REMOVE_OUT_EDGE:
-        break;
-    }
-  });
-  return !deleted && has_label;
-}
-
 // Helper function for iterating through label-property index. Returns true if
 // this transaction can see the given vertex, and the visible version has the
 // given label and property.
 inline bool CurrentVersionHasLabelProperty(const Vertex &vertex, LabelId label, PropertyId key,
                                            const PropertyValue &value, Transaction *transaction, View view) {
+  bool exists = true;
   bool deleted = false;
   bool has_label = false;
   bool current_value_equal_to_value = value.IsNull();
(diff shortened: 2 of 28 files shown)


src/storage/v2/vertex_info_cache.cpp
```diff
@@ -0,0 +1,211 @@
...
+namespace memgraph::storage {
+
+VertexInfoCache::VertexInfoCache() = default;
+VertexInfoCache::~VertexInfoCache() = default;
+VertexInfoCache::VertexInfoCache(VertexInfoCache &&) noexcept = default;
+VertexInfoCache &VertexInfoCache::operator=(VertexInfoCache &&) noexcept = default;
+
+auto VertexInfoCache::getExists(View view, Vertex const *vertex) const -> std::optional<bool> {
+  auto const &cache = (view == View::OLD) ? old_.existsCache_ : new_.existsCache_;
+  if (cache.empty()) return std::nullopt;
```
andrejtonev, line 25: I expect a find() on an empty map to have similar performance to empty().
Can we just do the find check?

src/storage/v2/vertex_info_cache.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+namespace memgraph::storage {
+
+VertexInfoCache::VertexInfoCache() = default;
+VertexInfoCache::~VertexInfoCache() = default;
+VertexInfoCache::VertexInfoCache(VertexInfoCache &&) noexcept = default;
+VertexInfoCache &VertexInfoCache::operator=(VertexInfoCache &&) noexcept = default;
+
+auto VertexInfoCache::getExists(View view, Vertex const *vertex) const -> std::optional<bool> {
+  auto const &cache = (view == View::OLD) ? old_.existsCache_ : new_.existsCache_;
```
andrejtonev, line 24: Could we use a global macro or function here to avoid copy/pasting the same thing in every function?

src/storage/v2/vertex_info_cache.cpp
```diff
@@ -0,0 +1,211 @@
...
+}
+void VertexInfoCache::storeDeleted(View view, Vertex const *vertex, bool res) {
+  auto &cache = (view == View::OLD) ? old_.deletedCache_ : new_.deletedCache_;
+  cache.emplace(vertex, res);
+}
+
+void VertexInfoCache::invalidate(Vertex const *vertex) {
+  new_.existsCache_.erase(vertex);
+  new_.deletedCache_.erase(vertex);
+  // Should we invalidate other caches proactively, should be guarded by existence/deletion?
```
andrejtonev: I would invalidate everything.

src/storage/v2/vertex_info_cache.cpp
```diff
@@ -0,0 +1,211 @@
...
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#include "storage/v2/vertex_info_cache.hpp"
+
+#include "storage/v2/property_value.hpp"
+
+namespace memgraph::storage {
+
+VertexInfoCache::VertexInfoCache() = default;
```
andrejtonev, line 18: General question regarding the cache...
Should it be thread safe or is it only being used inside a single-threaded execution session?

src/storage/v2/vertex_info_helpers.hpp
```diff
@@ -0,0 +1,215 @@
...
+    dispatch(DELETE_DESERIALIZED_OBJECT);
+    dispatch(DELETE_OBJECT);
+    dispatch(RECREATE_OBJECT);
+    dispatch(SET_PROPERTY);
+    dispatch(ADD_LABEL);
+    dispatch(REMOVE_LABEL);
+    dispatch(ADD_IN_EDGE);
+    dispatch(ADD_OUT_EDGE);
+    dispatch(REMOVE_IN_EDGE);
+    dispatch(REMOVE_OUT_EDGE);
```
andrejtonev, line 57: Could we add a default state here and assert?

src/storage/v2/vertex_info_helpers.hpp
```diff
@@ -0,0 +1,215 @@
...
+  return ActionMethod<RECREATE_OBJECT>([&](Delta const & /*unused*/) { deleted = false; });
+}
+
+inline auto HasLabel_ActionMethod(bool &has_label, LabelId label) {
+  using enum Delta::Action;
+  // clang-format off
+  return utils::Overloaded{
+      ActionMethod<REMOVE_LABEL>([&, label](Delta const &delta) {
+        if (delta.label == label) {
+          MG_ASSERT(has_label, "Invalid database state!");
```
andrejtonev, line 83: These MG_ASSERTS are concerning. `has_label` is an input, we can't control it here.

src/storage/v2/vertex_info_helpers.hpp
```diff
@@ -0,0 +1,215 @@
...
+  // clang-format off
+  return utils::Overloaded{
+      ActionMethod<REMOVE_LABEL>([&](Delta const &delta) {
+        auto it = std::find(labels.begin(), labels.end(), delta.label);
+        MG_ASSERT(it != labels.end(), "Invalid database state!");
+        *it = labels.back();
+        labels.pop_back();
+      }),
+      ActionMethod<ADD_LABEL>([&](Delta const &delta) {
+        auto it = std::find(labels.begin(), labels.end(), delta.label);
```
andrejtonev, line 108: Could we add std::execution::unseq to these std::finds?

andrejtonev (commented):
First pass

src/storage/v2/vertex_info_cache.cpp
Ignition: Yep testing proved that it needed to invalidate. Code has been changed.

src/storage/v2/vertex_info_cache.cpp
Ignition: `find` will
```cpp
template <class K = key_type>
const_iterator find(const key_arg<K>& key) const {
  prefetch_heap_block();
  return find(key, hash_ref()(key));
}
```
The `empty()` check is a cheap way to avoid what could be a relatively expensive hash.

Ignition: The expectation is the cache rarely has entries, I'm ok with having optimized the expected case.

src/storage/v2/vertex_info_helpers.hpp
Ignition: No need we get a compile time error is one of the enums are missing.
eg. `error: enumeration value 'SET_PROPERTY' not handled in switch [-Werror,-Wswitch]`

src/storage/v2/vertex_info_helpers.hpp
Ignition: Its an in + out reference. This is logic that remains from pre refactor.

src/storage/v2/vertex_info_helpers.hpp
Ignition: I did try that, need to make sure its removed. The expected number of labels is low which means cost of SIMD here makes perf worse.

Ignition: Oh so this was an interesting insight. I think the code gen for std::find std::execution::unseq is a poor choice in this situation. Its not SIMD, but lots of differences that probably also make it not ideal. https://godbolt.org/z/M58bxeqcM.

andrejtonev: Cool! Yeah, I am not sure how mature `std::execution::unseq` actually is. 
Changing the compile to clang 15 seems to do a good job.

Ignition:
@andrejtonev I'm going to delay delivering benchmark until #1152 is done.

andrejtonev (approved):

Josipmrden (commented):
Some comments, will do an additional review afterwards

.pre-commit-config.yaml
```diff
@@ -19,3 +19,16 @@ repos:
...
+#        name: clang-tidy
+#        description: Runs clang-tidy and checks for errors
+#        entry: python ./tools/pre-commit/clang-tidy.py
+#        language: python
+#        files: ^src/
+#        types: [c++, text]
+#        fail_fast: true
+#        require_serial: true
+#        args: [--compile_commands_path=build]
+#        pass_filenames: false
```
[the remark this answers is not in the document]
Ignition: The clang-tidy execution is SLOW. I want to add it to pre-commit for people to use, but need to openly talk to people before this is uncommented to impact every commit. 
The pre-commit stuff is thrown in here because IMO why not. If we are meant to keep PRs clean I deliver separately in future.

src/storage/v2/inmemory/indices_utils.hpp
```diff
@@ -223,46 +174,46 @@ inline bool CurrentVersionHasLabelProperty(const Vertex &vertex, LabelId label,
...
-                         case Delta::Action::REMOVE_IN_EDGE:
-                         case Delta::Action::REMOVE_OUT_EDGE:
-                           break;
-                       }
-                     });
-  return !deleted && has_label && current_value_equal_to_value;
+
+  // Checking cache has a cost, only do it if we have any deltas
+  // if we have no deltas then what we already have from the vertex is correct.
+  if (delta && transaction->isolation_level != IsolationLevel::READ_UNCOMMITTED) {
```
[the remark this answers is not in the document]
Ignition: Yes just replicated the IsolationLevel check earlier than before to avoid even using the cache.

src/storage/v2/inmemory/storage.cpp
```diff
@@ -435,6 +438,9 @@ Result<EdgeAccessor> InMemoryStorage::InMemoryAccessor::CreateEdge(VertexAccesso
   CreateAndLinkDelta(&transaction_, to_vertex, Delta::RemoveInEdgeTag(), edge_type, from_vertex, edge);
   to_vertex->in_edges.emplace_back(edge_type, from_vertex, edge);
 
+  transaction_.manyDeltasCache.Invalidate(from_vertex, edge_type, EdgeDirection::OUT);
+  transaction_.manyDeltasCache.Invalidate(to_vertex, edge_type, EdgeDirection::IN);
```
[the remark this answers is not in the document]
Ignition: Yes because the cached edge set and degree count will be wrong

src/storage/v2/vertex_accessor.cpp
```diff
@@ -34,31 +36,39 @@ std::pair<bool, bool> IsVisible(Vertex *vertex, Transaction *transaction, View v
...
+          Deleted_ActionMethod(deleted),
+          Exists_ActionMethod(exists)
+      });
+      // clang-format on
+    });
+
+    if (useCache && n_processed >= FLAGS_delta_chain_cache_threashold) {
+      auto &cache = transaction->manyDeltasCache;
+      cache.StoreExists(view, vertex, exists);
+      cache.StoreDeleted(view, vertex, deleted);
```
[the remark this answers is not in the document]
Josipmrden: maybe it does not make sense if I think of it because how would you know if you need to update something. perhaps use a `std::optional` for every time. idk, I'm just discussing what's better at this point although everything looks clear to me

Josipmrden:
Going to review the correctness by testing it myself now

tests/unit/storage_v2_durability_inmemory.cpp
```diff
@@ -744,21 +745,28 @@ TEST_P(DurabilityTest, SnapshotFallback) {
...
+  auto prev_number_of_snapshots = current_number_of_snapshots;
+  auto snapshots = GetSnapshotsList();
+  current_number_of_snapshots = snapshots.size();
+  ASSERT_GE(current_number_of_snapshots, prev_number_of_snapshots + 1);
   ASSERT_EQ(GetBackupSnapshotsList().size(), 0);
   ASSERT_EQ(GetWalsList().size(), 0);
   ASSERT_EQ(GetBackupWalsList().size(), 0);
 
-  // Destroy last snapshot.
+  // Destroy snapshota.
```
antoniofilipovic, line 762:
```suggestion
  // Destroy snapshots.
```

src/storage/v2/vertex_info_cache.cpp
Ignition: Maybe in the future with parallel queries...for now AFAICT not needed

PR #1128, opened 2023-08-01
[master < T1128] Implement RemoveProperty for node cpp api

mpintaric55334:
A need for RemoveProperty function was found when implementing custom C++ Query module. So it was implemented.
[master < Task] PR
- [ x] Provide the full content or a guide for the final git message

- [ x] Write a release note here, including added/changed clauses
Users can now use RemoveProperty() function for class Node when writing custom query modules in C++ API.
- [ x] Tag someone from docs team in the comments


mpintaric55334:
@vpavicic will start writing the docs for this when we know where to write for C++ API

include/mgp.hpp
```diff
@@ -636,6 +636,9 @@ class Node {
   /// @brief Sets the chosen property to the given value.
   void SetProperty(std::string property, Value value);
 
+  /// @brief Sets the chosen property to the given value.
```
antoniofilipovic, line 639: Removes the property, I think this is left from copy paste :smile:

tests/unit/cpp_api.cpp
```diff
@@ -560,3 +560,14 @@ TYPED_TEST(CppApiTestFixture, TestMapErase) {
...
   ASSERT_EQ(map.Size(), 0);
 }
+TYPED_TEST(CppApiTestFixture, TestNodeRemoveProperty) {
+  mgp_graph raw_graph = this->CreateGraph(memgraph::storage::View::NEW);
+  auto graph = mgp::Graph(&raw_graph);
+  auto node = graph.CreateNode();
+
+  int64_t int1 = 100;
+  mgp::Value value = mgp::Value(int1);
+  node.SetProperty("key", value);
```
antoniofilipovic, line 570: Can you add one assert in the middle between setting and removing that we have actually set the property?

antoniofilipovic: So, size==1, or something like that

tests/unit/cpp_api.cpp
```diff
@@ -560,3 +560,14 @@ TYPED_TEST(CppApiTestFixture, TestMapErase) {
   map.Erase("2");
   ASSERT_EQ(map.Size(), 0);
 }
+TYPED_TEST(CppApiTestFixture, TestNodeRemoveProperty) {
+  mgp_graph raw_graph = this->CreateGraph(memgraph::storage::View::NEW);
+  auto graph = mgp::Graph(&raw_graph);
+  auto node = graph.CreateNode();
+
+  int64_t int1 = 100;
+  mgp::Value value = mgp::Value(int1);
```
antoniofilipovic, line 569:
```suggestion
  mgp::Value value{int1};
```

antoniofilipovic (changes requested):
Few changes

antoniofilipovic (approved):

PR #1129, opened 2023-08-01
[master < T1129] Improve performance of delta creation

antoniofilipovic:
MOTIVATION: `setProperties` is too slow because many deltas are created. Custom allocators would speed up the execution.

This PR will most likely improve the situation `FREE MEMORY;` because memory fragmentation will be lower.

[master < Task] PR


src/utils/bond.hpp
```diff
@@ -0,0 +1,54 @@
...
+template <typename Container>
+struct Bond {
+  explicit Bond(std::size_t initial_size)
+      : res_(std::make_unique<resource>(initial_size)),
+        container_(memgraph::utils::Allocator<Container>(res_.get()).template new_object<Container>()){};
+
+  Bond(Bond &&other) noexcept : res_(std::move(other.res_)), container_(other.container_) {
+    other.container_ = nullptr;
+    [[maybe_unused]] auto _ = other.res_.release();
+  }
```
Ignition, line 27: useful utility std::exchange
```suggestion
  Bond(Bond &&other) noexcept : res_(std::exchange(other.res_, nullptr)), container_(other.container_) {
    other.container_ = nullptr;
  }
```

src/utils/bond.hpp
```diff
@@ -0,0 +1,54 @@
...
+#include <memory>
+#include "storage/v2/delta.hpp"
+#include "utils/memory.hpp"
+#include "utils/pmr/list.hpp"
+
+template <typename Container>
+struct Bond {
+  explicit Bond(std::size_t initial_size)
+      : res_(std::make_unique<resource>(initial_size)),
+        container_(memgraph::utils::Allocator<Container>(res_.get()).template new_object<Container>()){};
```
Ignition, line 22: Why allocate the container inside the resource and store a pointer? Why not just have the container be passed the polymorphic_allocator (implicitly constructed by `memory_resource*`). And store the container as a member of Bond, then the container wouldn't be behind a pointer (simplifying move).
```suggestion
  explicit Bond(std::size_t initial_size)
      : res_(std::make_unique<resource>(initial_size)),
        container_(res_.get()){};
```

src/utils/bond.hpp
antoniofilipovic: Nice!

tests/unit/storage_v2_wal_file.cpp
```diff
@@ -63,6 +64,8 @@ class DeltaGenerator final {
                        gen->storage_mode_) {}
 
    public:
+    ~Transaction() {}
```
Josipmrden, line 67: isn't this implied?

src/utils/bond.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+  Bond &operator=(const Bond &other) = delete;
+
+  Bond &operator=(Bond &&other) = delete;
+
+  auto use() -> Container & { return *container_; }
+
+  auto use() const -> const Container & { return *container_; }
+
+  auto res() -> memgraph::utils::MonotonicBufferResource * { return res_.get(); }
```
Josipmrden, line 38:
```suggestion
  auto res() -> resource * { return res_.get(); }
```

src/utils/bond.hpp
```diff
@@ -0,0 +1,53 @@
...
+// licenses/APL.txt.
+
+#include <cstddef>
+#include <memory>
+#include "storage/v2/delta.hpp"
+#include "utils/memory.hpp"
+#include "utils/pmr/list.hpp"
+
+template <typename Container>
+struct Bond {
```
Josipmrden, line 19: document on top what is Bond about

src/storage/v2/storage.cpp
```diff
@@ -117,7 +117,7 @@ StorageMode Storage::Accessor::GetCreationStorageMode() const { return creation_
 
 std::optional<uint64_t> Storage::Accessor::GetTransactionId() const {
   if (is_transaction_active_) {
-    return transaction_.transaction_id.load(std::memory_order_acquire);
+    return transaction_.transaction_id;
```
Josipmrden, line 120: why did we remove load again?

antoniofilipovic: Because this is not used in atomic way. transaction.id is set once and isn't changed or removed any other time

src/utils/bond.hpp
antoniofilipovic: Check if it is okay

src/storage/v2/transaction.hpp
```diff
@@ -26,24 +27,28 @@
...
 #include "storage/v2/vertex_info_cache.hpp"
 #include "storage/v2/view.hpp"
+#include "utils/bond.hpp"
+#include "utils/pmr/list.hpp"
 
 namespace memgraph::storage {
 
 const uint64_t kTimestampInitialId = 0;
 const uint64_t kTransactionInitialId = 1ULL << 63U;
+using PmrListDelta = utils::pmr::list<Delta>;
```
Ignition, line 37: Thinking about space saving (maybe in a later PR), would forward_list be worth using? Do we use reverse iteration?

Also thinking of further work on this theme, it would be nice to have a special kind of skip list for destruction. We only need to destruct the deltas which are not trivially destructible. ie. track Action::DELETE_DESERIALIZED_OBJECT and Action::SET_PROPERTY deltas for this purpose.

Ignition (approved):

src/storage/v2/transaction.hpp
antoniofilipovic: Yes, it makes sense. I will open other PR once I find time to investigate this.

antoniofilipovic:
@vpavicic  RN: In every transaction, deltas are now created using a custom allocator, which should give in some scenarios faster queries, and faster and more complete deallocation of nonusable objects (detlas) at the end of the query.

PR #1131, opened 2023-08-01
[master < T1131] Set properties C API extension

Josipmrden:
Based on the SetProperties bulk methods, extension of the C API has been made in order to enable bulk insert of properties inside query modules.

[master < Task] PR


include/mgp.hpp
```diff
@@ -2710,9 +2716,9 @@ inline void Node::RemoveLabel(const std::string_view label) {
   mgp::vertex_remove_label(this->ptr_, mgp_label{.name = label.data()});
 }
 
-inline std::map<std::string, Value> Node::Properties() const {
+inline std::unordered_map<std::string, Value> Node::Properties() const {
```
antoniofilipovic, line 2719: This is a bit of a breaking change for current procedures that are using std::map in MAGE or other procedures. Can we have two different functions? What was the reason to modify this so it returns unordered_map? I know it is faster but is there anything else?

include/mgp.hpp
```diff
@@ -2725,6 +2731,17 @@ inline void Node::SetProperty(std::string property, Value value) {
   mgp::vertex_set_property(ptr_, property.data(), value.ptr());
 }
 
+inline void Node::SetProperties(std::unordered_map<std::string_view, Value> properties) {
```
antoniofilipovic, line 2734: What if this function accepts `mgp::Map` since we are already transforming from unordered_map to mgp_map which is std::map? So that would be one less transformation and less memory?

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -1721,6 +1721,68 @@ mgp_error mgp_vertex_set_property(struct mgp_vertex *v, const char *property_nam
...
+    std::map<memgraph::storage::PropertyId, memgraph::storage::PropertyValue> props;
+    for (const auto &item : properties->items) {
+      props.insert(std::visit(
+          [&item](auto *impl) {
+            return std::make_pair(impl->NameToProperty(item.first), ToPropertyValue(item.second));
+          },
+          v->graph->impl));
+    }
+
+    const auto result = v->getImpl().UpdateProperties(props);
```
antoniofilipovic, line 1748: I think SubgraphDbAccessor is missing `UpdateProperties` function.

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -1721,6 +1721,68 @@ mgp_error mgp_vertex_set_property(struct mgp_vertex *v, const char *property_nam
...
+    }
+
+    for (const auto &res : *result) {
+      const auto property_key = std::get<0>(res);
+      const auto old_value = memgraph::query::TypedValue(std::get<1>(res));
+      const auto new_value = memgraph::query::TypedValue(std::get<2>(res));
+
+      if (new_value.IsNull()) {
+        trigger_ctx_collector->RegisterRemovedObjectProperty(v->getImpl(), property_key, old_value);
+        return;
```
antoniofilipovic, line 1778: Do you need to `continue` here?

src/query/procedure/py_module.cpp
```diff
@@ -1935,6 +1990,60 @@ PyObject *PyVertexSetProperty(PyVertex *self, PyObject *args) {
...
+  if (map_err != mgp_error::MGP_ERROR_NO_ERROR) {
+    throw std::runtime_error{"Unexpected error during creating mgp_map"};
+  }
+
+  PyObject *key{nullptr};
+  PyObject *value{nullptr};
+  Py_ssize_t pos{0};
+  while (PyDict_Next(props, &pos, &key, &value)) {
+    if (!PyUnicode_Check(key)) {
+      throw std::invalid_argument("Dictionary keys must be strings");
```
antoniofilipovic, line 2019: Is in the case of throw deleter called on unique ptr? Or only on scope exit when it exists regularly?

antoniofilipovic (changes requested):
Few comments, overall good work

include/mgp.hpp
Josipmrden: The reason is it is just faster.

src/query/procedure/mg_procedure_impl.cpp
Josipmrden: will implement

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -1721,6 +1721,68 @@ mgp_error mgp_vertex_set_property(struct mgp_vertex *v, const char *property_nam
...
+    }
+
+    const auto result = v->getImpl().UpdateProperties(props);
+    if (result.HasError()) {
+      switch (result.GetError()) {
+        case memgraph::storage::Error::DELETED_OBJECT:
+          throw DeletedObjectException{"Cannot set the properties of a deleted vertex!"};
+        case memgraph::storage::Error::NONEXISTENT_OBJECT:
+          LOG_FATAL("Query modules shouldn't have access to nonexistent objects when setting a property of a vertex!");
+        case memgraph::storage::Error::PROPERTIES_DISABLED:
```
antepusic, line 1755: Add error message

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -1721,6 +1721,68 @@ mgp_error mgp_vertex_set_property(struct mgp_vertex *v, const char *property_nam
...
+        case memgraph::storage::Error::VERTEX_HAS_EDGES:
+          LOG_FATAL("Unexpected error when setting a property of a vertex.");
+        case memgraph::storage::Error::SERIALIZATION_ERROR:
+          throw SerializationException{"Cannot serialize setting a property of a vertex."};
+      }
+    }
+
+    ctx->execution_stats[memgraph::query::ExecutionStats::Key::UPDATED_PROPERTIES] += properties->items.size();
```
antepusic, line 1763: Should this also include the vertex/-ices (and analogously for `mgp_vertex_set_properties`) whose properties were updated?

include/mgp.hpp
Josipmrden: either way you have to construct the mgp::Map which means it happens either in the API or in the query module. I'd rather want users to work with STL because we can't guess what is the input to the API (map as argument from query module or anything else)

src/query/procedure/mg_procedure_impl.cpp
Josipmrden: this falls upon the case below (vertex has edges) and the error message is 
`LOG_FATAL("Unexpected error when setting a property of a vertex.");`

Josipmrden: Properties can not be disabled on the vertex, only on edges

src/query/procedure/mg_procedure_impl.cpp
Josipmrden: yes!

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -2288,6 +2350,68 @@ mgp_error mgp_edge_set_property(struct mgp_edge *e, const char *property_name, m
...
+    }
+
+    for (const auto &res : *result) {
+      const auto property_key = std::get<0>(res);
+      const auto old_value = memgraph::query::TypedValue(std::get<1>(res));
+      const auto new_value = memgraph::query::TypedValue(std::get<2>(res));
+
+      if (new_value.IsNull()) {
+        trigger_ctx_collector->RegisterRemovedObjectProperty(e->impl, property_key, old_value);
+        return;
```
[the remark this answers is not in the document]
Josipmrden: yes!

src/query/procedure/mg_procedure_impl.cpp
Josipmrden: I think if the property hasn't change we still report that we changed the value (to the same one) so counting is okay I think

src/query/procedure/py_module.cpp
Josipmrden: https://godbolt.org/z/WGYo6r1hb yes, because of unique ptr it will call the deleter

antepusic (approved):
All’s good now to my eyes :+1:

antoniofilipovic (approved):
Looks good to me

Josipmrden:
> @Josipmrden please before merging check if everything passes on MAGE so that we don't break something on our side

Reference mage PR: https://github.com/memgraph/mage/pull/338

antepusic:
@kgolubic Documentation added at [documentation#160](https://github.com/memgraph/documentation/pull/160).

PR #1132, opened 2023-08-02
[master < T1132] Optimize import of edges on disk

as51340:
[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR

### DIFF
--- a/src/storage/v2/disk/storage.cpp
+++ b/src/storage/v2/disk/storage.cpp
@@ -25,6 +25,7 @@
 #include "kvstore/kvstore.hpp"
 #include "spdlog/spdlog.h"
 #include "storage/v2/constraints/unique_constraints.hpp"
+#include "storage/v2/delta.hpp"
 #include "storage/v2/disk/rocksdb_storage.hpp"
 #include "storage/v2/disk/storage.hpp"
 #include "storage/v2/disk/unique_constraints.hpp"
@@ -65,6 +66,19 @@ constexpr const char *label_index_str = "label_index";
 constexpr const char *label_property_index_str = "label_property_index";
 constexpr const char *existence_constraints_str = "existence_constraints";
 constexpr const char *unique_constraints_str = "unique_constraints";
+
+bool VertexNeedsToBeSerialized(const Vertex &vertex) {
+  Delta *head = vertex.delta;
+  while (head != nullptr) {
+    if (head->action == Delta::Action::ADD_LABEL || head->action == Delta::Action::REMOVE_LABEL ||
+        head->action == Delta::Action::DELETE_OBJECT || head->action == Delta::Action::RECREATE_OBJECT ||
+        head->action == Delta::Action::SET_PROPERTY) {
+      return true;
+    }
+    head = head->next;
+  }
+  return false;
+}
 
 bool VertexExistsInCache(const utils::SkipList<Vertex>::Accessor &accessor, Gid gid) {
   return accessor.find(gid) != accessor.end();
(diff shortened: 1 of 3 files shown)


src/storage/v2/disk/storage.cpp
```diff
@@ -1233,33 +1247,35 @@ DiskStorage::DiskAccessor::CheckVertexConstraintsBeforeCommit(
 
   /// TODO: andi I don't like that std::optional is used for checking errors but that's how it was before, refactor!
   for (Vertex &vertex : vertex_acc) {
-    if (auto check_result = CheckVertexConstraintsBeforeCommit(vertex, unique_storage); check_result.HasError()) {
-      return check_result.GetError();
-    }
+    if (VertexNeedsToBeSerialized(vertex)) {
```
Darych, line 1250: It is comment more about code style.
In my opinion it's harder to read code when only if statement is inside loop and all logic inside this if statement. Engineer will try to search for else clause or end of if to see if there is something we do outside of if statement.
```
for (Vertex &vertex : vertex_acc) {
  if (VertexNeedsToBeSerialized(vertex)) {
    ...
  }
}
```
Maybe we can rewrite this into:
```
for (Vertex &vertex : vertex_acc) {
  if (!VertexNeedsToBeSerialized(vertex)) {
    continue;
  }
  ...
}
```

as51340: I agree but the problem is that iteration over out_edges should happen anyway, that's why I cannot use `continue`

src/utils/rocksdb_serialization.hpp
```diff
@@ -182,9 +182,13 @@ inline std::string SerializeVertexAsValueForLabelIndex(storage::LabelId indexing
   return SerializeVertexAsValueForAuxiliaryStorages(indexing_label, vertex_labels, property_store);
 }
 
-inline std::vector<storage::LabelId> DeserializeLabelsFromLabelIndexStorage(const std::string &value) {
-  const auto value_splitted = value.substr(0, value.find('|'));
-  return TransformFromStringLabels(utils::Split(value_splitted, ","));
+inline storage::LabelId DeserializeLabelFromIndexStorage(const std::string &value) {
+  std::string label = value.substr(0, value.find('|'));
+  return storage::LabelId::FromUint(std::stoull(label));
```
Darych, line 187: Are we sure that in value we have just one label?

as51340: yes, you can check serialization methods for LabelIndex and LabelPropertyIndex

as51340: ah Monday bloody Monday... You are right this is for deserializing labels from value string

as51340: and then there was the bug that when we deserialized labels, we didn't include indexing label :exploding_head:

src/utils/rocksdb_serialization.hpp
```diff
@@ -93,7 +95,7 @@ inline std::string SerializeVertexAsValueForAuxiliaryStorages(storage::LabelId l
   std::vector<storage::LabelId> labels_without_target;
   std::copy_if(vertex_labels.begin(), vertex_labels.end(), std::back_inserter(labels_without_target),
                [&label_to_remove](const auto &label) { return label_to_remove != label; });
```
Darych, line 97: Isn't emplace_back with for loop will be better there also?

as51340: storage::LabelId is trivially copyable so I don't think so but will change

Darych (approved):

Darych (approved):

Darych (approved):

Darych (approved):

PR #1134, opened 2023-08-02
[master < T1134] Fix bug with on-disk triggers

Darych:
[master < T1134] PR
- [ ] Write E2E tests

[master < Task] PR


src/query/interpreter.cpp
```diff
@@ -3790,6 +3791,9 @@ void RunTriggersIndividually(const utils::SkipList<Trigger> &triggers, Interpret
     auto storage_acc = interpreter_context->db->Access();
     DbAccessor db_accessor{storage_acc.get()};
 
+    // On-disk storage removes all Vertex/Edge Accessors after DbAccessor is destroyed.
```
as51340, line 3794: But doesn't this happen also for InMemory storage? How is that different for OnDisk?

src/query/interpreter.cpp
```diff
@@ -3790,6 +3791,9 @@ void RunTriggersIndividually(const utils::SkipList<Trigger> &triggers, Interpret
...
+    // So we need to adapt TriggerContext based on user transaction which is still alive.
+    auto trigger_context = original_trigger_context;
```
as51340, line 3796: Can you explain me why this copying works?

src/storage/v2/disk/storage.cpp
```diff
@@ -376,9 +376,11 @@ std::optional<EdgeAccessor> DiskStorage::DiskAccessor::DeserializeEdge(const roc
...
   const auto edge_type_id = storage::EdgeTypeId::FromUint(std::stoull(edge_parts[3]));
   auto maybe_edge = CreateEdge(&*from_acc, &*to_acc, edge_type_id, edge_gid, value.ToStringView(), key.ToString());
-  MG_ASSERT(maybe_edge.HasValue());
-
-  return *maybe_edge;
+  if (maybe_edge.HasValue()) {
+    return *maybe_edge;
+  }
+  // One of vertex for edge could be deleted, so we don't create this edge.
+  return std::nullopt;
```
as51340, line 383: Can you also explain what is happening here? Because you were saying to me that `DeserializeEdge` is a problem because edge can be deleted. But `PrefetchEdges` shouldn't prefetch deleted edges! And if that is the test when properties on edges are false I think I know where the bug comes from:
```
      if (config_.properties_on_edges) {
        if (auto maybe_old_disk_key = utils::GetOldDiskKeyOrNull(edge.ptr->delta); maybe_old_disk_key.has_value()) {
          if (!DeleteEdgeFromDisk(maybe_old_disk_key.value())) {
            return StorageDataManipulationError{SerializationError{}};
          }
        }
      }
```
We only delete edge if properties on edges are true.

as51340 (changes requested):

src/query/interpreter.cpp
as51340: We talked and it makes sense! Do you think it would be good to add check if we are on disk storage since for in-memory it will make unnecessary copies?

src/storage/v2/disk/storage.cpp
```diff
@@ -915,6 +926,12 @@ void DiskStorage::DiskAccessor::PrefetchOutEdges(const VertexAccessor &vertex_ac
         }
       }
     }
+    for (auto &deleted_edge_key : edges_to_delete_) {
+      if (deleted_edge_key == diskEdgeKey) {
+        // We already deleted this edge.
+        return false;
+      }
+    }
```
as51340, line 934: Hmm here we have to iterate over all deleted edges which is expensive. I think a better solution would be to check whether edge_accessor is alrerady deleted.

as51340: while iterating over edge_acc. Same for InEdges

as51340 (changes requested):

as51340:
And maybe it would be good to have fixes for triggers and for edges in separate PRs since they are unrelated

src/storage/v2/disk/storage.cpp
Darych: Don't you think that iterating through all edge_acc would be more expensive?

Darych:
Reverted changes related to edge deletion. Will move to another PR.

as51340 (approved):

src/storage/v2/disk/storage.cpp
Darych: The main issue is that we deleted edge in the same transaction, so changes to disk were not written yet.
In prefetch edges we read edges from disk and the deleted edge is still presented on disk and can be prefetched. That's the reason why we need to filter out such edges during edge prefetch.

PR #1139, opened 2023-08-03
[master < T1139] Improve disk indices

as51340:
[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR

### DIFF
--- a/src/storage/v2/disk/storage.cpp
+++ b/src/storage/v2/disk/storage.cpp
@@ -440,7 +440,7 @@ void DiskStorage::DiskAccessor::LoadVerticesFromDiskLabelIndex(LabelId label, st
   ro.timestamp = &ts;
   auto index_it = std::unique_ptr<rocksdb::Iterator>(disk_index_transaction->GetIterator(ro));
 
-  std::string serialized_label = utils::SerializeIdType(label);
+  const std::string serialized_label = utils::SerializeIdType(label);
   for (index_it->SeekToFirst(); index_it->Valid(); index_it->Next()) {
     std::string key = index_it->key().ToString();
     Gid curr_gid = Gid::FromUint(std::stoull(utils::ExtractGidFromLabelIndexStorage(key)));
@@ -518,7 +518,7 @@ void DiskStorage::DiskAccessor::LoadVerticesFromDiskLabelPropertyIndex(LabelId l
   ro.timestamp = &ts;
   auto index_it = std::unique_ptr<rocksdb::Iterator>(disk_index_transaction->GetIterator(ro));
 
-  std::string label_property_prefix = utils::SerializeIdType(label) + "|" + utils::SerializeIdType(property);
+  const std::string label_property_prefix = utils::SerializeIdType(label) + "|" + utils::SerializeIdType(property);
   for (index_it->SeekToFirst(); index_it->Valid(); index_it->Next()) {
     std::string key = index_it->key().ToString();
     Gid curr_gid = Gid::FromUint(std::stoull(utils::ExtractGidFromLabelPropertyIndexStorage(key)));
@@ -567,7 +567,7 @@ void DiskStorage::DiskAccessor::LoadVerticesFromDiskLabelPropertyIndexWithPointV
   ro.timestamp = &ts;
   auto index_it = std::unique_ptr<rocksdb::Iterator>(disk_index_transaction->GetIterator(ro));
 
-  std::string label_property_prefix = utils::SerializeIdType(label) + "|" + utils::SerializeIdType(property);
+  const std::string label_property_prefix = utils::SerializeIdType(label) + "|" + utils::SerializeIdType(property);
   for (index_it->SeekToFirst(); index_it->Valid(); index_it->Next()) {
     std::string key = index_it->key().ToString();
     std::string it_value = index_it->value().ToString();
@@ -642,7 +642,7 @@ void DiskStorage::DiskAccessor::LoadVerticesFromDiskLabelPropertyIndexForInterva
   ro.timestamp = &ts;
   auto index_it = std::unique_ptr<rocksdb::Iterator>(disk_index_transaction->GetIterator(ro));
 
-  std::string label_property_prefix = utils::SerializeIdType(label) + "|" + utils::SerializeIdType(property);
+  const std::string label_property_prefix = utils::SerializeIdType(label) + "|" + utils::SerializeIdType(property);
   for (index_it->SeekToFirst(); index_it->Valid(); index_it->Next()) {
     std::string key_str = index_it->key().ToString();
     std::string it_value_str = index_it->value().ToString();
--- a/src/storage/v2/disk/storage.cpp
+++ b/src/storage/v2/disk/storage.cpp
@@ -409,10 +409,10 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(LabelId label, View view) {
 
 std::unordered_set<Gid> DiskStorage::DiskAccessor::MergeVerticesFromMainCacheWithLabelIndexCache(
     LabelId label, View view, std::list<Delta> &index_deltas, utils::SkipList<Vertex> *indexed_vertices) {
-  std::unordered_set<Gid> gids;
-  gids.reserve(vertices_.size());
-
   auto main_cache_acc = vertices_.access();
+  std::unordered_set<Gid> gids;
+  gids.reserve(main_cache_acc.size());
+
   for (const auto &vertex : main_cache_acc) {
     gids.insert(vertex.gid);
     if (VertexHasLabel(vertex, label, &transaction_, view)) {
@@ -592,7 +592,7 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(LabelId label, PropertyId p
   index_deltas_storage_.emplace_back(std::list<Delta>());
   auto &index_deltas = index_deltas_storage_.back();
 
-  auto gids = MergeVerticesFromMainCacheWithLabelPropertyIndexCacheForIntevalSearch(
+  auto gids = MergeVerticesFromMainCacheWithLabelPropertyIndexCacheForIntervalSearch(
       label, property, view, lower_bound, upper_bound, index_deltas, indexed_vertices.get());
 
   LoadVerticesFromDiskLabelPropertyIndexForIntervalSearch(label, property, gids, lower_bound, upper_bound, index_deltas,
@@ -603,7 +603,7 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(LabelId label, PropertyId p
 }
 
 std::unordered_set<Gid>
-DiskStorage::DiskAccessor::MergeVerticesFromMainCacheWithLabelPropertyIndexCacheForIntevalSearch(
+DiskStorage::DiskAccessor::MergeVerticesFromMainCacheWithLabelPropertyIndexCacheForIntervalSearch(
     LabelId label, PropertyId property, View view, const std::optional<utils::Bound<PropertyValue>> &lower_bound,
     const std::optional<utils::Bound<PropertyValue>> &upper_bound, std::list<Delta> &index_deltas,
     utils::SkipList<Vertex> *indexed_vertices) {
--- a/src/storage/v2/disk/storage.cpp
+++ b/src/storage/v2/disk/storage.cpp
@@ -16,6 +16,7 @@
 
 #include <rocksdb/comparator.h>
 #include <rocksdb/db.h>
+#include <rocksdb/iterator.h>
 #include <rocksdb/slice.h>
 
 #include <rocksdb/options.h>
@@ -321,17 +322,16 @@ std::optional<storage::VertexAccessor> DiskStorage::DiskAccessor::LoadVertexToMa
 }
 
 std::optional<storage::VertexAccessor> DiskStorage::DiskAccessor::LoadVertexToLabelIndexCache(
-    const rocksdb::Slice &key, const rocksdb::Slice &value, Delta *index_delta,
+    std::string &&key, std::string &&value, Delta *index_delta,
     utils::SkipList<storage::Vertex>::Accessor index_accessor) {
-  storage::Gid gid = Gid::FromUint(std::stoull(utils::ExtractGidFromLabelIndexStorage(key.ToString())));
+  storage::Gid gid = Gid::FromUint(std::stoull(utils::ExtractGidFromLabelIndexStorage(key)));
   if (VertexExistsInCache(index_accessor, gid)) {
     return std::nullopt;
   }
 
-  const std::string value_str{value.ToString()};
-  const auto labels{utils::DeserializeLabelsFromLabelIndexStorage(value_str)};
-  return CreateVertex(index_accessor, gid, labels, utils::DeserializePropertiesFromLabelIndexStorage(value_str),
-                      index_delta);
+  std::vector<LabelId> labels_id{utils::DeserializeLabelsFromLabelIndexStorage(value)};
+  PropertyStore properties{utils::DeserializePropertiesFromLabelIndexStorage(value)};
+  return CreateVertex(index_accessor, gid, labels_id, std::move(properties), index_delta);
 }
 
 std::optional<storage::VertexAccessor> DiskStorage::DiskAccessor::LoadVertexToLabelPropertyIndexCache(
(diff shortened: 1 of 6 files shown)


src/storage/v2/disk/storage.cpp
```diff
@@ -402,43 +400,57 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(LabelId label, View view) {
...
+  LoadVerticesFromDiskLabelIndex(label, gids, index_deltas, indexed_vertices.get());
+
+  return VerticesIterable(AllVerticesIterable(indexed_vertices->access(), &transaction_, view, &storage_->indices_,
+                                              &storage_->constraints_, storage_->config_.items));
+}
+
+std::unordered_set<Gid> DiskStorage::DiskAccessor::MergeVerticesFromMainCacheWithLabelIndexCache(
+    LabelId label, View view, std::list<Delta> &index_deltas, utils::SkipList<Vertex> *indexed_vertices) {
+  std::unordered_set<Gid> gids;
+  gids.reserve(vertices_.size());
```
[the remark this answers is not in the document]
as51340: it's the same but I will do it for consistency reasons

src/storage/v2/disk/storage.cpp
```diff
@@ -447,24 +459,38 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(LabelId label, PropertyId p
...
-  std::string strTs = utils::StringTimestamp(transaction_.start_timestamp);
-  rocksdb::Slice ts(strTs);
-  ro.timestamp = &ts;
-  auto index_it = std::unique_ptr<rocksdb::Iterator>(disk_index_transaction->GetIterator(ro));
+  auto gids = MergeVerticesFromMainCacheWithLabelPropertyIndexCache(label, property, view, index_deltas,
+                                                                    indexed_vertices.get(), label_property_filter);
+
+  const auto disk_label_property_filter = [](const std::string &key, const std::string &label_property_prefix,
+                                             const std::unordered_set<Gid> &gids, Gid curr_gid) -> bool {
+    return key.starts_with(label_property_prefix) && !utils::Contains(gids, curr_gid);
```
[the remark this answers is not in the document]
as51340: I would avoid mixing logic of LabelIndex and LabelPropertyIndex at least for some time. Prefix search should come here anyway and then I would see. But for now I really wouldn't do that if that is ok

src/storage/v2/disk/storage.cpp
```diff
@@ -402,43 +400,57 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(LabelId label, View view) {
...
+  rocksdb::ReadOptions ro;
+  std::string strTs = utils::StringTimestamp(transaction_.start_timestamp);
+  rocksdb::Slice ts(strTs);
+  ro.timestamp = &ts;
+  auto index_it = std::unique_ptr<rocksdb::Iterator>(disk_index_transaction->GetIterator(ro));
 
+  std::string serialized_label = utils::SerializeIdType(label);
   for (index_it->SeekToFirst(); index_it->Valid(); index_it->Next()) {
     std::string key = index_it->key().ToString();
     Gid curr_gid = Gid::FromUint(std::stoull(utils::ExtractGidFromLabelIndexStorage(key)));
```
antoniofilipovic, line 446: I have seen combining std::stoull with utils::ExtractGidFrom... a few times already, maybe it makes sense to combine it into function?

src/storage/v2/disk/storage.cpp
```diff
@@ -402,43 +400,57 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(LabelId label, View view) {
...
-    spdlog::debug("Loaded vertex with key: {} from label index storage", key);
-    /// TODO: optimize
-    if (key.starts_with(utils::SerializeIdType(label)) && !utils::Contains(gids, curr_gid)) {
-      LoadVertexToLabelIndexCache(index_it->key(), index_it->value(),
+    spdlog::trace("Loaded vertex with key: {} from label index storage", key);
+    if (key.starts_with(serialized_label) && !utils::Contains(gids, curr_gid)) {
```
antoniofilipovic, line 448: From the perspective of performance, maybe you will gain some if you compare first if curr_gid is inside of gids, as this should be fairly fast, then compare if the string starts with some label.

Checking should be as complex as number you have to hash, and that number is probably smaller then key, right?

src/storage/v2/disk/storage.cpp
as51340: yes I think that extract should be separate from std::stoull but I should add support for creating gid from string. And this is also for LabelId, PropertyId ... So I will do that at some point in other PR to tackle everything at once

src/storage/v2/disk/storage.cpp
```diff
@@ -402,43 +400,57 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(LabelId label, View view) {
...
+  auto disk_index_transaction = disk_label_index->CreateRocksDBTransaction();
+  disk_index_transaction->SetReadTimestampForValidation(transaction_.start_timestamp);
+
+  rocksdb::ReadOptions ro;
+  std::string strTs = utils::StringTimestamp(transaction_.start_timestamp);
+  rocksdb::Slice ts(strTs);
+  ro.timestamp = &ts;
+  auto index_it = std::unique_ptr<rocksdb::Iterator>(disk_index_transaction->GetIterator(ro));
 
+  std::string serialized_label = utils::SerializeIdType(label);
```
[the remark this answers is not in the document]
as51340: yes, const is good, will add :)

src/storage/v2/disk/storage.cpp
as51340: hah understand what you mean but not sure because gids can be large... So how to verify that? Plus, it also depends what do you do in your tx. If you are just loading data using index then gids check isn't even necessary. But if you are also writing in you tx and using view semantic then things start to become crazy and gids can be super large set.

src/storage/v2/disk/storage.cpp
```diff
@@ -321,31 +321,29 @@ std::optional<storage::VertexAccessor> DiskStorage::DiskAccessor::LoadVertexToMa
...
-  return CreateVertex(index_accessor, gid, labels, utils::DeserializePropertiesFromLabelIndexStorage(value_str),
-                      index_delta);
+  std::vector<LabelId> labels_id{utils::DeserializeLabelsFromLabelIndexStorage(value)};
+  PropertyStore properties{utils::DeserializePropertiesFromLabelIndexStorage(value)};
+  return CreateVertex(index_accessor, gid, labels_id, std::move(properties), index_delta);
 }
 
 std::optional<storage::VertexAccessor> DiskStorage::DiskAccessor::LoadVertexToLabelPropertyIndexCache(
-    const rocksdb::Slice &key, const rocksdb::Slice &value, Delta *index_delta,
+    std::string &&key, std::string &&value, Delta *index_delta,
```
antoniofilipovic, line 337: Why not use `const &` here on key?

antoniofilipovic: And `value` too

src/storage/v2/disk/storage.cpp
```diff
@@ -321,31 +321,29 @@ std::optional<storage::VertexAccessor> DiskStorage::DiskAccessor::LoadVertexToMa
...
+  storage::Gid gid = Gid::FromUint(std::stoull(utils::ExtractGidFromLabelPropertyIndexStorage(key)));
   if (VertexExistsInCache(index_accessor, gid)) {
     return std::nullopt;
   }
 
-  const std::string value_str{value.ToString()};
-  const auto labels{utils::DeserializeLabelsFromLabelPropertyIndexStorage(value_str)};
-  return CreateVertex(index_accessor, gid, labels,
-                      utils::DeserializePropertiesFromLabelPropertyIndexStorage(value.ToString()), index_delta);
+  std::vector<LabelId> labels_id{utils::DeserializeLabelsFromLabelPropertyIndexStorage(value)};
```
antoniofilipovic, line 344: I think you can use here `const auto`

src/storage/v2/disk/storage.cpp
```diff
@@ -321,31 +321,29 @@ std::optional<storage::VertexAccessor> DiskStorage::DiskAccessor::LoadVertexToMa
...
+  storage::Gid gid = Gid::FromUint(std::stoull(utils::ExtractGidFromLabelIndexStorage(key)));
   if (VertexExistsInCache(index_accessor, gid)) {
     return std::nullopt;
   }
 
-  const std::string value_str{value.ToString()};
-  const auto labels{utils::DeserializeLabelsFromLabelIndexStorage(value_str)};
-  return CreateVertex(index_accessor, gid, labels, utils::DeserializePropertiesFromLabelIndexStorage(value_str),
-                      index_delta);
+  std::vector<LabelId> labels_id{utils::DeserializeLabelsFromLabelIndexStorage(value)};
```
antoniofilipovic, line 331: Same comments here as for LoadVertexToLabelPropertyIndexCache

src/storage/v2/disk/storage.cpp
```diff
@@ -402,43 +400,57 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(LabelId label, View view) {
   index_deltas_storage_.emplace_back(std::list<Delta>());
```
antoniofilipovic, line 400: You can modify this too, no need to send std::list<Delta>(), you can just do:
`index_deltas_storage_.emplace_back();`

src/storage/v2/disk/storage.cpp
```diff
@@ -402,43 +400,57 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(LabelId label, View view) {
...
       LoadVertexToLabelIndexCache(utils::SerializeVertexAsKeyForLabelIndex(label, vertex.gid),
                                   utils::SerializeVertexAsValueForLabelIndex(label, vertex.labels, vertex.properties),
                                   CreateDeleteDeserializedIndexObjectDelta(&transaction_, index_deltas, std::nullopt),
                                   indexed_vertices->access());
     }
   }
+  return gids;
+}
 
+void DiskStorage::DiskAccessor::LoadVerticesFromDiskLabelIndex(LabelId label, std::unordered_set<storage::Gid> gids,
```
antoniofilipovic, line 430: Why are gids passed by value?

src/storage/v2/disk/storage.cpp
antoniofilipovic: I run some benchmarks, seems we are not any better with my suggestion, so feel free to discard it: https://quick-bench.com/q/Eci2t1BBNMMVsItR1RNADJq0Xys

src/storage/v2/disk/storage.cpp
```diff
@@ -496,9 +538,27 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(LabelId label, PropertyId p
   index_deltas_storage_.emplace_back(std::list<Delta>());
```
antoniofilipovic, line 538: I think you can pass here also empty to avoid creating temporary:
`index_deltas_storage_.emplace_back();`

src/storage/v2/disk/storage.cpp
```diff
@@ -496,9 +538,27 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(LabelId label, PropertyId p
...
   auto &index_deltas = index_deltas_storage_.back();
 
+  auto label_property_filter = [this, &value](const Vertex &vertex, LabelId label, PropertyId property,
+                                              View view) -> bool {
+    return VertexHasLabel(vertex, label, &transaction_, view) &&
+           VertexHasEqualPropertyValue(vertex, property, value, &transaction_, view);
+  };
+
+  auto gids = MergeVerticesFromMainCacheWithLabelPropertyIndexCache(label, property, view, index_deltas,
```
antoniofilipovic, line 547: I think you can do here `const auto gids`

src/storage/v2/disk/storage.cpp
as51340: because I think it is better like this, I don't need this value anymore in the caller so I just "steal" them

src/storage/v2/disk/storage.cpp
```diff
@@ -496,9 +538,27 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(LabelId label, PropertyId p
...
+
+  LoadVerticesFromDiskLabelPropertyIndexWithPointValueLookup(label, property, gids, value, index_deltas,
+                                                             indexed_vertices.get());
+
+  return VerticesIterable(AllVerticesIterable(indexed_vertices->access(), &transaction_, view, &storage_->indices_,
+                                              &storage_->constraints_, storage_->config_.items));
+}
+
+void DiskStorage::DiskAccessor::LoadVerticesFromDiskLabelPropertyIndexWithPointValueLookup(
+    LabelId label, PropertyId property, std::unordered_set<storage::Gid> gids, const PropertyValue &value,
```
antoniofilipovic, line 558: Why are you passsing here gids as value?

src/storage/v2/disk/storage.cpp
as51340: then I cannot move it

src/storage/v2/disk/storage.cpp
```diff
@@ -321,31 +321,29 @@ std::optional<storage::VertexAccessor> DiskStorage::DiskAccessor::LoadVertexToMa
...
     return std::nullopt;
   }
 
-  const std::string value_str{value.ToString()};
-  const auto labels{utils::DeserializeLabelsFromLabelPropertyIndexStorage(value_str)};
-  return CreateVertex(index_accessor, gid, labels,
-                      utils::DeserializePropertiesFromLabelPropertyIndexStorage(value.ToString()), index_delta);
+  std::vector<LabelId> labels_id{utils::DeserializeLabelsFromLabelPropertyIndexStorage(value)};
+  PropertyStore properties{utils::DeserializePropertiesFromLabelPropertyIndexStorage(value)};
+  return CreateVertex(index_accessor, gid, labels_id, std::move(properties), index_delta);
```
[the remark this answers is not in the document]
as51340: yes it was like this before but I find it here more "descriptive" and tidy to have it separated. I think both is the same

src/storage/v2/disk/storage.cpp
```diff
@@ -575,7 +623,26 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(LabelId label, PropertyId p
           indexed_vertices->access());
     }
   }
+  return gids;
+}
+
+void DiskStorage::DiskAccessor::LoadVerticesFromDiskLabelPropertyIndexForIntervalSearch(
+    LabelId label, PropertyId property, std::unordered_set<storage::Gid> gids,
```
antoniofilipovic, line 630: Same comment about gids, just I think we should use here `const &`

src/storage/v2/disk/storage.cpp
```diff
@@ -507,37 +567,20 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(LabelId label, PropertyId p
...
-        HasVertexEqualPropertyValue(vertex, property, value, &transaction_, view)) {
-      LoadVertexToLabelPropertyIndexCache(
-          utils::SerializeVertexAsKeyForLabelPropertyIndex(label, property, vertex.gid),
-          utils::SerializeVertexAsValueForLabelPropertyIndex(label, vertex.labels, vertex.properties),
-          CreateDeleteDeserializedIndexObjectDelta(&transaction_, index_deltas, std::nullopt),
-          indexed_vertices->access());
-    }
-  }
-
+  const std::string label_property_prefix = utils::SerializeIdType(label) + "|" + utils::SerializeIdType(property);
```
antoniofilipovic, line 570: `const auto`

src/storage/v2/disk/storage.cpp
as51340: ah I was experimenting something with lambdas and then forgot to change

src/storage/v2/disk/storage.cpp
```diff
@@ -575,7 +623,26 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(LabelId label, PropertyId p
...
+  auto *disk_label_property_index =
+      static_cast<DiskLabelPropertyIndex *>(storage_->indices_.label_property_index_.get());
+
+  auto disk_index_transaction = disk_label_property_index->CreateRocksDBTransaction();
+  disk_index_transaction->SetReadTimestampForValidation(transaction_.start_timestamp);
+  rocksdb::ReadOptions ro;
+  std::string strTs = utils::StringTimestamp(transaction_.start_timestamp);
+  rocksdb::Slice ts(strTs);
+  ro.timestamp = &ts;
+  auto index_it = std::unique_ptr<rocksdb::Iterator>(disk_index_transaction->GetIterator(ro));
```
antoniofilipovic, line 643: Is it possible to split this function into more functions, so 1 would be LoadVerticesFromDiskLabelPropIndex and 2. would be do interval search?

antoniofilipovic: Maybe up until here we can call it Load... or CreateIterator, and afterwards it is DoIntervalSearch...

src/storage/v2/disk/storage.cpp
as51340: agree

src/storage/v2/disk/storage.cpp
as51340: answered -> changed

src/storage/v2/disk/storage.cpp
```diff
@@ -584,16 +651,13 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(LabelId label, PropertyId p
     /// TODO: couple this condition
     PropertyStore properties = utils::DeserializePropertiesFromLabelPropertyIndexStorage(it_value_str);
     auto prop_value = properties.GetProperty(property);
-    if (key_str.starts_with(utils::SerializeIdType(label) + "|" + utils::SerializeIdType(property)) &&
-        !utils::Contains(gids, curr_gid) && IsPropertyValueWithinInterval(prop_value, lower_bound, upper_bound)) {
+    if (key_str.starts_with(label_property_prefix) && !utils::Contains(gids, curr_gid) &&
```
antoniofilipovic, line 654: Also, I think you can do early return here in form of `continue` and then have one less indentation.

src/storage/v2/disk/storage.cpp
as51340: why? std::string vs auto

src/storage/v2/disk/storage.cpp
```diff
@@ -1138,7 +1202,7 @@ bool DiskStorage::DiskAccessor::WriteVertexToDisk(const Vertex &vertex) {
   auto status = disk_transaction_->Put(disk_storage->kvstore_->vertex_chandle, utils::SerializeVertex(vertex),
                                        utils::SerializeProperties(vertex.properties));
   if (status.ok()) {
-    spdlog::debug("rocksdb: Saved vertex with key {} and ts {}", utils::SerializeVertex(vertex), *commit_timestamp_);
+    spdlog::trace("rocksdb: Saved vertex with key {} and ts {}", utils::SerializeVertex(vertex), *commit_timestamp_);
```
antoniofilipovic, line 1205: Do we need here check if commit timestamp is not optional value? Maybe do MG_ASSERT because it shouldn't be

src/storage/v2/disk/storage.cpp
```diff
@@ -1162,7 +1226,7 @@ bool DiskStorage::DiskAccessor::WriteEdgeToDisk(const EdgeRef edge, const std::s
     status = disk_transaction_->Put(disk_storage->kvstore_->edge_chandle, serializedEdgeKey, "");
   }
   if (status.ok()) {
-    spdlog::debug("rocksdb: Saved edge with key {} and ts {}", serializedEdgeKey, *commit_timestamp_);
+    spdlog::trace("rocksdb: Saved edge with key {} and ts {}", serializedEdgeKey, *commit_timestamp_);
```
antoniofilipovic, line 1229: Same comment about commit timestamp

src/storage/v2/disk/storage.cpp
antoniofilipovic: If I am not wrong it should work correctly even if you pass `const &` due to reference collapsing, so on that side I think we are good. To be honest I am not just sure if it is the correct design pattern, what are flows in approach if you only function only accepts && ref. I think "having it as it is now" can maybe shoot us in the leg afterward.

If I remember correctly from a book I read, I think && ref should be used only to signal that you plan to move it to a different variable and in case of perfect forwarding when you want to pass const & or &&.

src/storage/v2/disk/storage.cpp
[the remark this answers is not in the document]
antoniofilipovic: But you are creating here additional variable on stack so in that case it is not the same, one operation more?

src/storage/v2/disk/storage.cpp
antoniofilipovic: I think auto as type is obvious from concatenation, and const as you don't plan to change value. So no need for std::string it is more code

src/storage/v2/disk/storage.cpp
as51340: I mean we can but then I have to check every such function.. And the reason I didn't do it from the beginning is because ro.timestamp needs pointer so I cannot allocate timestamp on stack

src/storage/v2/disk/storage.cpp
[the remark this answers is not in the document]
as51340: yes but that's really ignorable cost. Move is just casting to right-value ref

src/storage/v2/disk/storage.cpp
as51340: yes true and in this case I don't want to move it but just use it in the function and destroy it. And about reference collapsing, it will work yes but this should be more efficient. Because you basically discard it

src/storage/v2/disk/storage.cpp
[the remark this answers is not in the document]
antoniofilipovic: Hmm, I don't see why to create var from which you will move, and the move might not even move it has just permission to move, so I am not sure. I would discard var, but if you think it is okay, leave it and resolve conversation :+1:

antoniofilipovic (approved):
Looks good to go

Darych (approved):
LGTM

PR #1140, opened 2023-08-03
[master < T1140] Add ToString for mgp::Value and all mgp types

mpintaric55334:
[master < Task] PR
For util.md5 function in c++ util query module, a need for ToString method arised, so it was implemented.

Users can now call ToString() method on mgp::Value and Memgraph's data types when writing C++ query modules.


mpintaric55334:
@vpavicic will write documentation when PR is reviewed.

include/mgp.hpp
```diff
@@ -2486,6 +2528,21 @@ inline bool Map::operator==(const Map &other) const { return util::MapsEqual(ptr
...
 
+inline const std::string Map::ToString() const {
+  std::string return_string = "{";
+  const size_t map_size = Size();
+  size_t i = 0;
+  for (auto item : *this) {
+    if (i < map_size - 1) {
+      return_string += std::string(item.key) + "=" + item.value.ToString() + ", ";
+      i++;
+      continue;
```
antoniofilipovic, line 2539:
```suggestion
    if (i == map_size - 1) {
      break;
    }
    return_string += std::string(item.key) + "=" + item.value.ToString() + ", ";
    ++i;
```

include/mgp.hpp
```diff
@@ -2662,6 +2721,14 @@ inline bool Relationship::operator==(const Relationship &other) const {
 
 inline bool Relationship::operator!=(const Relationship &other) const { return !(*this == other); }
 
+inline const std::string Relationship::ToString() const {
+  const std::string node_from{"(" + std::to_string(From().Id().AsInt()) + ")"};
+  const std::string type{Type()};
```
antoniofilipovic, line 2726: Doesn't this construct emtpy type?

include/mgp.hpp
```diff
@@ -2723,6 +2790,19 @@ inline bool Path::operator==(const Path &other) const { return util::PathsEqual(
 
 inline bool Path::operator!=(const Path &other) const { return !(*this == other); }
 
+inline const std::string Path::ToString() const {
+  const size_t length = Length();
```
antoniofilipovic, line 2794: You can use `const auto` here

tests/unit/cpp_api.cpp
```diff
@@ -598,3 +598,47 @@ TYPED_TEST(CppApiTestFixture, TestValuePrint) {
...
+  mgp::Date date_1{"2020-12-12"};
+  ASSERT_EQ(mgp::Value(date_1).ToString(), "2020-12-12");
+
+  /*relationship*/
+  mgp_graph raw_graph = this->CreateGraph(memgraph::storage::View::NEW);
+  auto graph = mgp::Graph(&raw_graph);
+  auto node1 = graph.CreateNode();
+  auto node2 = graph.CreateNode();
+  auto rel = graph.CreateRelationship(node1, node2, "loves");
+  ASSERT_EQ(mgp::Value(rel).ToString(), "(0)-[loves,0]->(1)");
```
antoniofilipovic, line 635: `(id:0)-[id:0 :TYPE]->(id:1)`

include/mgp.hpp
```diff
@@ -2662,6 +2721,14 @@ inline bool Relationship::operator==(const Relationship &other) const {
 
 inline bool Relationship::operator!=(const Relationship &other) const { return !(*this == other); }
 
+inline const std::string Relationship::ToString() const {
+  const std::string node_from{"(" + std::to_string(From().Id().AsInt()) + ")"};
+  const std::string type{Type()};
+  const std::string relationship{"[" + type + "," + std::to_string(Id().AsInt()) + "]"};
```
antoniofilipovic, line 2727: How come you didn't output relationship properties?

include/mgp.hpp
```diff
@@ -2486,6 +2528,21 @@ inline bool Map::operator==(const Map &other) const { return util::MapsEqual(ptr
 
 inline bool Map::operator!=(const Map &other) const { return !(*this == other); }
 
+inline const std::string Map::ToString() const {
+  std::string return_string = "{";
+  const size_t map_size = Size();
+  size_t i = 0;
+  for (auto item : *this) {
```
antoniofilipovic, line 2535: Can you change this with while loop same as in List?

antoniofilipovic (changes requested):
Few comments, but good work

include/mgp.hpp
mpintaric55334: No, it returns the type of relationship

include/mgp.hpp
mpintaric55334: Isnt it better to keep the for lopp, cause I need to iterate through to the semi last element for string formatting, and with iterators I would have to do additional checks if the element is semi last

include/mgp.hpp
```diff
@@ -2590,6 +2653,26 @@ inline bool Node::operator==(const Node &other) const { return util::NodesEqual(
...
+  const size_t map_size = properties_map.size();
+  size_t i = 0;
+  for (auto item : properties_map) {
+    if (i == map_size - 1) {
+      properties += std::string(item.first) + "=" + item.second.ToString();
+      break;
+    }
+    properties += std::string(item.first) + "=" + item.second.ToString() + ", ";
+    ++i;
+  }
```
antoniofilipovic, line 2672: Can you isolate this part into function, it seems like the same for `Relationship::ToString` part of function

include/mgp.hpp
```diff
@@ -2590,6 +2653,26 @@ inline bool Node::operator==(const Node &other) const { return util::NodesEqual(
...
+  std::string properties = "";
+  std::map<std::string, Value> properties_map{Properties()};
+  const size_t map_size = properties_map.size();
+  size_t i = 0;
+  for (auto item : properties_map) {
+    if (i == map_size - 1) {
+      properties += std::string(item.first) + "=" + item.second.ToString();
+      break;
+    }
+    properties += std::string(item.first) + "=" + item.second.ToString() + ", ";
```
antoniofilipovic, line 2670: For properties, we usually return `"key : "value"` instead of `"key" = "value"`. I think we should change that here too

include/mgp.hpp
```diff
@@ -2486,6 +2531,24 @@ inline bool Map::operator==(const Map &other) const { return util::MapsEqual(ptr
...
+inline const std::string Map::ToString() const {
+  std::string return_string{"{"};
+  const size_t map_size = Size();
+  if (map_size == 0) {
+    return "{}";
+  }
+  size_t i = 0;
+  for (auto item : *this) {
+    if (i == map_size - 1) {
+      return_string += std::string(item.key) + "=" + item.value.ToString() + "}";
```
antoniofilipovic, line 2543: Same comment about changing "=" with ":" as below

include/mgp.hpp
```diff
@@ -2590,6 +2653,26 @@ inline bool Node::operator==(const Node &other) const { return util::NodesEqual(
 
 inline bool Node::operator!=(const Node &other) const { return !(*this == other); }
 
+inline const std::string Node::ToString() const {
+  std::string labels{""};
+  for (auto label : Labels()) {
+    labels += ":" + std::string(label) + " ";
+  }
+  std::string properties = "";
```
antoniofilipovic, line 2661: Check that you are consistent in your PR:
`std::string properties{""};`

include/mgp.hpp
```diff
@@ -2590,6 +2653,26 @@ inline bool Node::operator==(const Node &other) const { return util::NodesEqual(
...
+  std::map<std::string, Value> properties_map{Properties()};
+  const size_t map_size = properties_map.size();
```
antoniofilipovic, line 2663: You can use `auto` for type deduction

include/mgp.hpp
```diff
@@ -2590,6 +2653,26 @@ inline bool Node::operator==(const Node &other) const { return util::NodesEqual(
...
+  size_t i = 0;
```
antoniofilipovic, line 2664: `decltype(map_size)` will give you type here

include/mgp.hpp
```diff
@@ -2723,6 +2838,43 @@ inline bool Path::operator==(const Path &other) const { return util::PathsEqual(
...
+  const auto length = Length();
+  size_t i = 0;
+  std::string return_string = "";
+  for (i = 0; i < length; i++) {
+    std::string labels{""};
+    const auto node = GetNodeAt(i);
+    for (auto label : node.Labels()) {
+      labels += ":" + std::string(label) + " ";
+    }
+    return_string += "(id: " + std::to_string(node.Id().AsInt()) + " " + labels + ")-";
```
antoniofilipovic, line 2851: Can you use `node.ToString()` here?

include/mgp.hpp
```diff
@@ -2590,6 +2653,26 @@ inline bool Node::operator==(const Node &other) const { return util::NodesEqual(
...
+  size_t i = 0;
+  for (auto item : properties_map) {
+    if (i == map_size - 1) {
+      properties += std::string(item.first) + "=" + item.second.ToString();
+      break;
+    }
+    properties += std::string(item.first) + "=" + item.second.ToString() + ", ";
+    ++i;
+  }
+  return "Node(id: " + std::to_string(Id().AsInt()) + " " + labels + "properties: {" + properties + "} )";
```
antoniofilipovic, line 2673: I think you don't need to return `Node` here, so you can just do the following:

`(n:Label1:Label2 {prop1:"key", prop2:"key"}`

include/mgp.hpp
antoniofilipovic: I think it is not that of a problem if you use it here, it should be okay that we get all the labels and all the properties of node

include/mgp.hpp
```diff
@@ -2723,6 +2838,43 @@ inline bool Path::operator==(const Path &other) const { return util::PathsEqual(
...
+      if (i == map_size - 1) {
+        properties += std::string(item.first) + "=" + item.second.ToString();
+        break;
+      }
+      properties += std::string(item.first) + "=" + item.second.ToString() + ", ";
+      ++i;
+    }
+    return_string += "[type: " + std::string(rel.Type()) + ", id: " + std::to_string(rel.Id().AsInt()) +
+                     ", properties: {" + properties + "}]->";
+  }
```
antoniofilipovic, line 2868: Here then you can use `relationship.ToString()` if I am not wrong

tests/unit/cpp_api.cpp
```diff
@@ -598,3 +598,51 @@ TYPED_TEST(CppApiTestFixture, TestValuePrint) {
...
+  mgp::Date date_1{"2020-12-12"};
+  ASSERT_EQ(mgp::Value(date_1).ToString(), "2020-12-12");
+
+  /*relationship*/
+  mgp_graph raw_graph = this->CreateGraph(memgraph::storage::View::NEW);
+  auto graph = mgp::Graph(&raw_graph);
+  auto node1 = graph.CreateNode();
+  node1.AddLabel("Label1");
+  auto node2 = graph.CreateNode();
+  auto rel = graph.CreateRelationship(node1, node2, "loves");
```
antoniofilipovic, line 635: Types have first letter as capitat letter: "Loves"

antoniofilipovic (changes requested):
Few changes but looks better

include/mgp.hpp
mpintaric55334: I dont think so, because relationship to string also outputs nodes included in the relationship, so the nodes would be duplicated in the string

include/mgp.hpp
mpintaric55334: It will give me a const type, so no bueno

include/mgp.hpp
```diff
@@ -2590,6 +2653,26 @@ inline bool Node::operator==(const Node &other) const { return util::NodesEqual(
...
+inline const std::string Node::ToString() const {
+  std::string labels{""};
+  for (auto label : Labels()) {
+    labels += ":" + std::string(label) + " ";
+  }
+  std::string properties = "";
+  std::map<std::string, Value> properties_map{Properties()};
+  const size_t map_size = properties_map.size();
+  size_t i = 0;
+  for (auto item : properties_map) {
```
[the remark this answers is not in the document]
mpintaric55334: This gives me errors

include/mgp.hpp
```diff
@@ -2317,6 +2347,21 @@ inline bool List::operator==(const List &other) const { return util::ListsEqual(
...
+  const size_t size = Size();
+  if (size == 0) {
+    return "[]";
+  }
+  size_t i = 0;
+  while (i < size - 1) {
+    return_str += (*this)[i].ToString() + ", ";
+    i++;
+  }
+  return_str += (*this)[i].ToString() + "]";
```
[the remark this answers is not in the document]
mpintaric55334: Done

tests/unit/cpp_api.cpp
```diff
@@ -598,3 +598,53 @@ TYPED_TEST(CppApiTestFixture, TestValuePrint) {
...
+  mgp_graph raw_graph = this->CreateGraph(memgraph::storage::View::NEW);
+  auto graph = mgp::Graph(&raw_graph);
+  auto node1 = graph.CreateNode();
+  node1.AddLabel("Label1");
+  auto node2 = graph.CreateNode();
+  auto rel = graph.CreateRelationship(node1, node2, "Loves");
+  rel.SetProperty("key", mgp::Value("property"));
+  ASSERT_EQ(
+      mgp::Value(rel).ToString(),
+      "(id: 0 :Label1 properties: {} )-[type: Loves, id: 0, properties: {key: property}]->(id: 1 properties: {} )");
```
antepusic, line 639: Separate the node id, labels and properties with `,`, and I also think the space after the `properties` map shouldn't be there

antepusic (changes requested):
Looks pretty good! Left a few comments related to formatting and testing.

tests/unit/cpp_api.cpp
```diff
@@ -598,3 +598,77 @@ TYPED_TEST(CppApiTestFixture, TestValuePrint) {
...
+            "properties: {key: node_property, key2: node_property2})");
+  /*path*/
+  mgp::Path path = mgp::Path(node1);
+  path.Expand(rel);
+  auto node3 = graph.CreateNode();
+  auto rel2 = graph.CreateRelationship(node2, node3, "Loves2");
+  path.Expand(rel2);
+  ASSERT_EQ(
+      mgp::Value(path).ToString(),
+      "(id: 2, :Label1 :Label2, properties: {})-[type: Loves, id: 0, properties: {key: property}]->(id: 3, properties: "
```
antepusic, line 672: This is minor, but labels are not separated by space elsewhere, e.g. in mgconsole.

antepusic (approved):
I think we're good to go here :rocket: 
Left a small suggestion, but the PR is approvable either way.

include/mgp.hpp
```diff
@@ -2486,6 +2532,24 @@ inline bool Map::operator==(const Map &other) const { return util::MapsEqual(ptr
...
+  size_t i = 0;
+  for (const auto &[key, value] : *this) {
+    if (i == map_size - 1) {
+      return_string.append(std::string(key) + ": " + value.ToString() + "}");
+      break;
+    }
+    return_string.append(std::string(key) + ": " + value.ToString() + ", ");
+    ++i;
+  }
+  return return_string;
```
Josipmrden, line 2550: is it possible to use `fmt` library here?

antoniofilipovic: I wouldn't introduce fmt library here if not necessary

antoniofilipovic (approved):
LGTG

PR #1145, opened 2023-08-06
[master < T1145] Filter deleted edges during edge prefetch

Darych:
[master < T1145] PR
- [ ] Write E2E tests

[master < Task] PR

### DIFF
--- a/src/storage/v2/disk/storage.cpp
+++ b/src/storage/v2/disk/storage.cpp
@@ -874,12 +874,7 @@ bool DiskStorage::DiskAccessor::PrefetchEdgeFilter(const std::string_view disk_e
   bool isOutEdge = (edge_direction == EdgeDirection::OUT);
   DiskEdgeKey disk_edge_key(disk_edge_key_str);
   auto edges_res = (isOutEdge ? vertex_acc.OutEdges(storage::View::NEW) : vertex_acc.InEdges(storage::View::NEW));
-  std::string disk_vertex_gid;
-  if (isOutEdge) {
-    disk_vertex_gid = disk_edge_key.GetVertexOutGid();
-  } else {
-    disk_vertex_gid = disk_edge_key.GetVertexInGid();
-  }
+  const std::string disk_vertex_gid = (isOutEdge ? disk_edge_key.GetVertexOutGid() : disk_edge_key.GetVertexInGid());
   auto edge_gid = disk_edge_key.GetEdgeGid();
 
   if (disk_vertex_gid != utils::SerializeIdType(vertex_acc.Gid())) {
@@ -1269,8 +1264,8 @@ DiskStorage::DiskAccessor::CheckVertexConstraintsBeforeCommit(
 
     for (auto &edge_entry : vertex.out_edges) {
       EdgeRef edge = std::get<2>(edge_entry);
-      DiskEdgeKey src_dest_key(vertex.gid, std::get<1>(edge_entry)->gid, std::get<0>(edge_entry), edge,
-                               config_.properties_on_edges);
+      const DiskEdgeKey src_dest_key(vertex.gid, std::get<1>(edge_entry)->gid, std::get<0>(edge_entry), edge,
+                                     config_.properties_on_edges);
 
       /// TODO: expose temporal coupling
       /// NOTE: this deletion has to come before writing, otherwise RocksDB thinks that all entries are deleted
--- a/src/storage/v2/disk/storage.cpp
+++ b/src/storage/v2/disk/storage.cpp
@@ -12,6 +12,7 @@
 #include <limits>
 #include <optional>
 #include <stdexcept>
+#include <string_view>
 #include <vector>
 
 #include <rocksdb/comparator.h>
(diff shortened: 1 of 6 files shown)


src/storage/v2/disk/storage.cpp
```diff
@@ -868,7 +869,40 @@ DiskStorage::DiskAccessor::DetachDeleteVertex(VertexAccessor *vertex) {
       std::move(deleted_edges));
 }
 
-void DiskStorage::DiskAccessor::PrefetchEdges(const auto &prefetch_edge_filter) {
+bool DiskStorage::DiskAccessor::PrefetchEdgeFilter(const std::string_view &diskEdgeKey,
```
as51340, line 872:
```suggestion
bool DiskStorage::DiskAccessor::PrefetchEdgeFilter(const std::string_view diskEdgeKey,
```

src/storage/v2/disk/storage.cpp
```diff
@@ -868,7 +869,40 @@ DiskStorage::DiskAccessor::DetachDeleteVertex(VertexAccessor *vertex) {
...
+                                                   const VertexAccessor &vertex_acc, bool isOutEdge) {
+  // diskEdgeKey format: vertex_out_gid | vertex_in_gid | direction | edge_type | GID
```
as51340, line 874: unnecessary comment, if we change the format then comment becomes obsolete

src/storage/v2/disk/storage.cpp
```diff
@@ -868,7 +869,40 @@ DiskStorage::DiskAccessor::DetachDeleteVertex(VertexAccessor *vertex) {
       std::move(deleted_edges));
 }
 
-void DiskStorage::DiskAccessor::PrefetchEdges(const auto &prefetch_edge_filter) {
+bool DiskStorage::DiskAccessor::PrefetchEdgeFilter(const std::string_view &diskEdgeKey,
+                                                   const VertexAccessor &vertex_acc, bool isOutEdge) {
```
as51340, line 873: In general when functions receive boolean as an argument, it already indicates that it does more than one thing. That's why I preferred before explicit lambda

src/storage/v2/disk/storage.cpp
```diff
@@ -868,7 +869,40 @@ DiskStorage::DiskAccessor::DetachDeleteVertex(VertexAccessor *vertex) {
...
+  auto edges_res = (isOutEdge ? vertex_acc.OutEdges(storage::View::NEW) : vertex_acc.InEdges(storage::View::NEW));
+  if (isOutEdge) {
+    disk_vertex_gid = diskEdgeKey.substr(0, diskEdgeKey.find('|'));
+  } else {
+    auto vertex_in_start = diskEdgeKey.find('|') + 1;
+    disk_vertex_gid = diskEdgeKey.substr(vertex_in_start, diskEdgeKey.find('|', vertex_in_start) - vertex_in_start);
+  }
+  auto edge_gid = diskEdgeKey.substr(diskEdgeKey.rfind('|') + 1);
+
+  if (edges_res.HasValue()) {
```
as51340, line 885: Here should be an assert in my opinion because edges must exist

src/storage/v2/disk/storage.cpp
```diff
@@ -868,7 +869,40 @@ DiskStorage::DiskAccessor::DetachDeleteVertex(VertexAccessor *vertex) {
...
+        // We already inserted this edge into the vertex's in_edges list.
+        return false;
+      }
+    }
+  }
+
+  // We need to iterate through edges_to_delete_ also because removed edges are not presented in edges_res
+  for (auto &deleted_edge_key : edges_to_delete_) {
+    if (deleted_edge_key == diskEdgeKey) {
+      // We already deleted this edge.
```
as51340, line 897: Unnecessary comment

src/storage/v2/disk/storage.cpp
```diff
@@ -868,7 +869,40 @@ DiskStorage::DiskAccessor::DetachDeleteVertex(VertexAccessor *vertex) {
...
+      return false;
+    }
+  }
+
+  return disk_vertex_gid == utils::SerializeIdType(vertex_acc.Gid());
```
as51340, line 902: but we can do this check as soon as we assert that edges exist. Because it doesn't make sense to do 2 for loops and then at the do the check with disk_vertex_gid.

src/storage/v2/disk/storage.cpp
```diff
@@ -868,7 +869,40 @@ DiskStorage::DiskAccessor::DetachDeleteVertex(VertexAccessor *vertex) {
...
+    for (const auto &edge_acc : edges_res.GetValue()) {
+      if (utils::SerializeIdType(edge_acc.Gid()) == edge_gid) {
+        // We already inserted this edge into the vertex's in_edges list.
+        return false;
+      }
+    }
+  }
+
+  // We need to iterate through edges_to_delete_ also because removed edges are not presented in edges_res
+  for (auto &deleted_edge_key : edges_to_delete_) {
```
as51340, line 895: We can at least use std::unordered_set instead of std::vector to reduce the complexity of this check.

tests/e2e/triggers/workloads.yaml
```diff
@@ -25,7 +25,7 @@ disk_storage_properties_edges_false: &disk_storage_properties_edges_false
     main:
       args: ["--bolt-port", *bolt_port, "--log-level=TRACE", "--also-log-to-stderr", "--storage-properties-on-edges=False"]
       log_file: "triggers-e2e.log"
-      setup_queries: []
```
as51340, line 28: You should add `STORAGE MODE ON_DISK_TRANSACTIONAL` to the disk_template_cluster workload too.

as51340 (changes requested):

tests/e2e/triggers/workloads.yaml
Darych: It's added in another PR: https://github.com/memgraph/memgraph/pull/1134

src/storage/v2/disk/storage.cpp
Darych: Discussed on a call. Decided to have enum instead of bool

src/storage/v2/disk/storage.cpp
Darych: Agree

src/storage/v2/disk/storage.cpp
```diff
@@ -868,7 +869,38 @@ DiskStorage::DiskAccessor::DetachDeleteVertex(VertexAccessor *vertex) {
       std::move(deleted_edges));
 }
 
-void DiskStorage::DiskAccessor::PrefetchEdges(const auto &prefetch_edge_filter) {
+bool DiskStorage::DiskAccessor::PrefetchEdgeFilter(const std::string_view disk_edge_key_str,
+                                                   const VertexAccessor &vertex_acc, EdgeDirection edge_direction) {
+  bool isOutEdge = (edge_direction == EdgeDirection::OUT);
+  DiskEdgeKey disk_edge_key(disk_edge_key_str);
+  auto edges_res = (isOutEdge ? vertex_acc.OutEdges(storage::View::NEW) : vertex_acc.InEdges(storage::View::NEW));
+  std::string disk_vertex_gid;
```
as51340, line 877:
```suggestion
  const std::string disk_vertex_gid;
```
and then use IIFE

src/storage/v2/disk/storage.cpp
```diff
@@ -1264,8 +1269,8 @@ DiskStorage::DiskAccessor::CheckVertexConstraintsBeforeCommit(
 
     for (auto &edge_entry : vertex.out_edges) {
       EdgeRef edge = std::get<2>(edge_entry);
-      auto src_dest_key = utils::SerializeEdge(vertex.gid, std::get<1>(edge_entry)->gid, std::get<0>(edge_entry), edge,
-                                               config_.properties_on_edges);
+      DiskEdgeKey src_dest_key(vertex.gid, std::get<1>(edge_entry)->gid, std::get<0>(edge_entry), edge,
```
as51340, line 1272:
```suggestion
      const DiskEdgeKey src_dest_key(vertex.gid, std::get<1>(edge_entry)->gid, std::get<0>(edge_entry), edge,
```

as51340 (approved):
Two small suggestions and if checks pass good

src/storage/v2/disk/storage.cpp
Darych: Done

src/storage/v2/disk/storage.cpp
Darych: Done

PR #1155, opened 2023-08-10
Optimize splitting keys from on-disk storage

Darych:
[master < T1155] PR
- [ ] Write E2E tests

[master < Task] PR


src/utils/string.hpp
```diff
@@ -301,6 +301,26 @@ inline std::vector<std::string> RSplit(const std::string_view src, const std::st
   return res;
 }
 
+/**
+ * Get a first part of a string by `delimiter` into a string.
+ */
```
as51340, line 306: I think the name of the function is self-explanatory

src/utils/string.hpp
```diff
@@ -301,6 +301,26 @@ inline std::vector<std::string> RSplit(const std::string_view src, const std::st
...
+inline std::string_view GetFirstPartOfSplit(const std::string_view src, const char delimiter) {
+  auto delimPos = src.find(delimiter);
+  return delimPos == std::string::npos ? src : src.substr(0, delimPos);
+}
+
+/**
+ * Get a second part of a string by `delimiter` into a string.
+ */
```
as51340, line 314: I think the name of the function is self-explanatory

src/utils/string.hpp
```diff
@@ -301,6 +301,26 @@ inline std::vector<std::string> RSplit(const std::string_view src, const std::st
   return res;
 }
 
+/**
+ * Get a first part of a string by `delimiter` into a string.
+ */
+inline std::string_view GetFirstPartOfSplit(const std::string_view src, const char delimiter) {
+  auto delimPos = src.find(delimiter);
```
as51340, line 308:
```suggestion
  size_t delimPos = src.find(delimiter);
```
Nothing against auto but increases compilation time and if it is such short type like size_t I think it is better to have it explicit

src/utils/string.hpp
```diff
@@ -301,6 +301,26 @@ inline std::vector<std::string> RSplit(const std::string_view src, const std::st
   return res;
 }
 
+/**
+ * Get a first part of a string by `delimiter` into a string.
+ */
+inline std::string_view GetFirstPartOfSplit(const std::string_view src, const char delimiter) {
```
as51340, line 307: I like inline, makes sense here. How about returning std::string instead of string_view? Because then clients will always have to convert to string and it's again the same

as51340 (changes requested):

src/utils/string.hpp
```diff
@@ -301,6 +301,26 @@ inline std::vector<std::string> RSplit(const std::string_view src, const std::st
...
+  auto delimPos = src.find(delimiter);
+  return delimPos == std::string::npos ? src : src.substr(0, delimPos);
```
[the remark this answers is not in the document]
Darych: Yes, I thought about this. Will check performance, because we have one place `DeserializePropertiesFromAuxiliaryStorages` we pass string_view to function. But there should not be degradation. Will check

src/utils/string.hpp
Darych: Do you have ideas how to name these functions better? I'm trying to find names, but currently can't create something better.

src/utils/string.hpp
Darych: Yes, I thought about this. Will check performance, because we have one place DeserializePropertiesFromAuxiliaryStorages we pass string_view to function. But there should not be degradation. Will check

src/utils/string.hpp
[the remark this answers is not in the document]
as51340: how, `If no such character was found, there are no tokens in str at all, and the function returns a null pointer`?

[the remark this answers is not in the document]
Darych: In this case that means that string contains only delimiter/s.

src/utils/rocksdb_serialization.hpp
```diff
@@ -30,6 +30,50 @@ namespace memgraph::utils {
 static constexpr const char *outEdgeDirection = "0";
 static constexpr const char *inEdgeDirection = "1";
 
+namespace {
+struct StartEndPositions {
+  size_t start;
+  size_t end;
+
+  size_t Size() const { return end - start; }
+  bool Valid() const { return start != std::string::npos; }
```
andrejtonev, line 39: && start <= end

src/utils/rocksdb_serialization.hpp
```diff
@@ -30,6 +30,50 @@ namespace memgraph::utils {
...
+    result.start = result.end;
+    result.end = str.find(delim, result.start);
+    if (result.end == std::string::npos) {
+      if (i < partNumber - 1) {
+        // We didn't find enough parts.
+        result.start = std::string::npos;
+      }
+      return result;
+    }
+    ++result.end;
```
andrejtonev, line 55: On calls where we are not exhausting the number of parts, this will cause the delimiter to be included in the final result.
Is that what we wanted?
FindPartStartEndPositions("A|B|C", '|', 1) -> "A|"

src/utils/rocksdb_serialization.hpp
```diff
@@ -30,6 +30,50 @@ namespace memgraph::utils {
...
+  }
+  return result;
+}
+
+inline std::string_view GetViewOfFirstPartOfSplit(const std::string_view src, const char delimiter) {
+  size_t delimPos = src.find(delimiter);
```
andrejtonev, line 61: Could we just use the FindPartStartEndPositions here as well, or is the penalty too high?

src/utils/rocksdb_serialization.hpp
```diff
@@ -30,6 +30,50 @@ namespace memgraph::utils {
...
+namespace {
+struct StartEndPositions {
+  size_t start;
+  size_t end;
+
+  size_t Size() const { return end - start; }
+  bool Valid() const { return start != std::string::npos; }
+};
+
+inline StartEndPositions FindPartStartEndPositions(const std::string_view str, const char delim,
```
andrejtonev, line 42: Since this is being called through other functions, could we make the `partNumber` a template?

src/utils/rocksdb_serialization.hpp
```diff
@@ -30,6 +30,50 @@ namespace memgraph::utils {
...
+}
+
+inline std::string_view GetViewOfFirstPartOfSplit(const std::string_view src, const char delimiter) {
+  size_t delimPos = src.find(delimiter);
+  return delimPos == std::string::npos ? src : src.substr(0, delimPos);
+}
+
+inline std::string_view GetViewOfSecondPartOfSplit(const std::string_view src, const char delimiter) {
+  StartEndPositions startEndPos = FindPartStartEndPositions(src, delimiter, 2);
+  return (startEndPos.Valid() ? src.substr(startEndPos.start, startEndPos.Size()) : src);
```
andrejtonev, line 67: Maybe make this a helper function or even return this directly from FindPartStartEndPositions.

src/utils/rocksdb_serialization.hpp
```diff
@@ -111,13 +155,11 @@ inline std::string SerializeVertexAsValueForAuxiliaryStorages(storage::LabelId l
...
 inline std::string ExtractGidFromKey(const std::string &key) {
-  std::vector<std::string> key_vector = utils::Split(key, "|");
-  return key_vector[1];
+  return std::string(GetViewOfSecondPartOfSplit(key, '|'));
 }
 
 inline storage::PropertyStore DeserializePropertiesFromAuxiliaryStorages(const std::string &value) {
-  std::vector<std::string> value_vector = utils::Split(value, "|");
-  std::string properties_str = value_vector[1];
+  std::string_view properties_str = GetViewOfSecondPartOfSplit(value, '|');
```
andrejtonev, line 162: const

src/utils/rocksdb_serialization.hpp
```diff
@@ -167,10 +207,10 @@ inline std::string SerializeVertexAsValueForUniqueConstraint(const storage::Labe
 }
 
 inline storage::LabelId DeserializeConstraintLabelFromUniqueConstraintStorage(const std::string &key) {
-  std::vector<std::string> key_vector = utils::Split(key, "|");
-  std::vector<std::string> constraint_key = utils::Split(key_vector[0], ",");
+  std::string_view firstPartKey = GetViewOfFirstPartOfSplit(key, '|');
+  std::string_view constraint_key = GetViewOfFirstPartOfSplit(firstPartKey, ',');
   /// TODO: andi Change this to deserialization method directly into the LabelId class
-  return storage::LabelId::FromUint(std::stoull(constraint_key[0]));
+  return storage::LabelId::FromUint(std::stoull(std::string(constraint_key)));
```
andrejtonev, line 213: Way to avoid creating a string by using std::from_chars: https://stackoverflow.com/a/63125672

src/utils/rocksdb_serialization.hpp
```diff
@@ -167,10 +207,10 @@ inline std::string SerializeVertexAsValueForUniqueConstraint(const storage::Labe
 }
 
 inline storage::LabelId DeserializeConstraintLabelFromUniqueConstraintStorage(const std::string &key) {
-  std::vector<std::string> key_vector = utils::Split(key, "|");
-  std::vector<std::string> constraint_key = utils::Split(key_vector[0], ",");
+  std::string_view firstPartKey = GetViewOfFirstPartOfSplit(key, '|');
```
andrejtonev, line 210: const

src/utils/rocksdb_serialization.hpp
```diff
@@ -167,10 +207,10 @@ inline std::string SerializeVertexAsValueForUniqueConstraint(const storage::Labe
...
+  std::string_view constraint_key = GetViewOfFirstPartOfSplit(firstPartKey, ',');
```
andrejtonev, line 211: const

src/utils/rocksdb_serialization.hpp
andrejtonev: Also, is this the behavior we want; return everything if part wasn't found?

src/utils/rocksdb_serialization.hpp
Darych: In this case result.end == std::string::npos, so it's end of a string + 1.
At the end of loop we increment result.end for the case when '|' is found in middle of string

src/utils/rocksdb_serialization.hpp
Darych: We discussed with Andi about this here: https://github.com/memgraph/memgraph/pull/1155#discussion_r1289962009
I think this is what we want now. Because some cases expect this.

src/utils/rocksdb_serialization.hpp
andrejtonev: The ++end is being called outside the `if (result.end == std::string::npos) {...}` block.
Here's a simple test I ran: https://godbolt.org/z/bz493W8Kf
I print the results using find and split.
P.S.
Make sure godbolt is executing the code, you should see the terminal and the results there.

Darych: Thank you, Andreja! You are right :) I'm adding more unit tests for extraction functions.

src/utils/rocksdb_serialization.hpp
andrejtonev: Sorry, I meant doing `template <int partNumber>` but after doing some benchmarks, I don't see any difference, so this could just stay `const int` I guess.

src/utils/rocksdb_serialization.hpp
```diff
@@ -30,6 +31,47 @@ namespace memgraph::utils {
...
+  return FindPartOfStringView(src, delimiter, 2);
+}
+
+inline std::string_view GetViewOfThirdPartOfSplit(const std::string_view src, const char delimiter) {
+  return FindPartOfStringView(src, delimiter, 3);
+}
+
+}  // namespace
+
 /// TODO: try to move this to hpp files so that we can follow jump on readings
```
as51340, line 75: Remove what?

as51340 (approved):

src/utils/rocksdb_serialization.hpp
Darych: I think this is one of your comments. Currently, it's unclear to me. Maybe we should update the TODO or remove it?

andrejtonev (approved):

PR #1158, opened 2023-08-10
[master < T0968] - Fix concurrent query module racecondition

gvolfing:
Concurrent access to the same query module had a racecondition on the pointer that was used to handle the costum memory management. With this commit, a mapping has been added that keeps information about what thread used what pointer to handle the memory resources. Which should be fine since the respected query executions are running on a dedicated thread. Access to the mapping itself is threadsafe. A simple spinlock has been implemented here, as we shouldn't include memgraph::utils from this header and a traditional mutex might be overkill. A simple RAII wrapper for the mapping container has been also added for simpler client-side use.

### DIFF
--- a/query_modules/example.cpp
+++ b/query_modules/example.cpp
@@ -1,4 +1,4 @@
-// Copyright 2022 Memgraph Ltd.
+// Copyright 2023 Memgraph Ltd.
 //
 // Use of this software is governed by the Business Source License
 // included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
@@ -18,7 +18,7 @@ void ProcImpl(std::vector<mgp::Value> arguments, mgp::Graph graph, mgp::RecordFa
 
 void SampleReadProc(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory) {
   try {
-    mgp::memory = memory;
+    mgp::MemoryResourceDistributorGuard guard(memory);
 
     std::vector<mgp::Value> arguments;
     for (size_t i = 0; i < mgp::list_size(args); i++) {
@@ -34,7 +34,7 @@ void SampleReadProc(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *resul
 }
 
 void AddXNodes(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory) {
-  mgp::memory = memory;
+  mgp::MemoryResourceDistributorGuard guard(memory);
   auto graph = mgp::Graph(memgraph_graph);
 
   std::vector<mgp::Value> arguments;
@@ -49,7 +49,7 @@ void AddXNodes(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mg
 }
 
 void Multiply(mgp_list *args, mgp_func_context *ctx, mgp_func_result *res, mgp_memory *memory) {
-  mgp::memory = memory;
+  mgp::MemoryResourceDistributorGuard guard(memory);
 
   std::vector<mgp::Value> arguments;
   for (size_t i = 0; i < mgp::list_size(args); i++) {
@@ -67,7 +67,7 @@ void Multiply(mgp_list *args, mgp_func_context *ctx, mgp_func_result *res, mgp_m
 
 extern "C" int mgp_init_module(struct mgp_module *module, struct mgp_memory *memory) {
   try {
-    mgp::memory = memory;
+    mgp::MemoryResourceDistributorGuard guard(memory);
 
     AddProcedure(SampleReadProc, "return_true", mgp::ProcedureType::Read,
                  {mgp::Parameter("param_1", mgp::Type::Int), mgp::Parameter("param_2", mgp::Type::Double, 2.3)},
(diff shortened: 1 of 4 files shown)


query_modules/example.cpp
```diff
@@ -18,7 +18,11 @@ void ProcImpl(std::vector<mgp::Value> arguments, mgp::Graph graph, mgp::RecordFa
 
 void SampleReadProc(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory) {
   try {
-    mgp::memory = memory;
+    // The outcommented way of assigning the memory pointer is still
+    // working, but it is deprecated because of certain concurrency
+    // issues. Please use the guard instead.
+    // mgp::memory = memory;
+    mgp::MemoryDispatcherGuard guard(memory);
```
Josipmrden, line 25: I think we have quite a few query modules now implemented in MAGE so I guess we also need to make an additional PR to make these concurrently executable as well?

Josipmrden: this seems cool though, no overhead at all

tests/e2e/concurrent_query_modules/client.py
```diff
@@ -0,0 +1,33 @@
...
+import pytest
+
+
+def inner(query, number_of_executions):
+    connection = mgclient.connect(host="127.0.0.1", port=7687)
+    connection.autocommit = True
+    cursor = connection.cursor()
+    for _ in range(number_of_executions):
+        cursor.execute(query)
+    cursor.fetchall()
```
Josipmrden, line 13: you also have GQLAlchemy available in the venv so you can use it as well

tests/e2e/concurrent_query_modules/con_query_modules.py
```diff
@@ -0,0 +1,19 @@
+import sys
+
+import pytest
+from client import *
+
+query = "MATCH (my_person:Person)-[:FOLLOW]->(follow_person:Person) MATCH(follow_person) - [: LIKE] -> (post:Post) WHERE post.indexedAt IS NOT NULL AND NOT exists((post) - [: ROOT] -> (: Post)) WITH localDateTime() - post.indexedAt as duration, post, follow_person WHERE duration.day < 5 WITH(duration.day * 24) + duration.hour as hour_age, post, follow_person ORDER BY post.indexedAt DESC LIMIT 500 MATCH(: Person) - [l: LIKE] -> (post) WITH count(l) as likes, hour_age, post, follow_person CALL libmodule_test.hacker_news(likes, 123, 4.1) YIELD score RETURN ID(post), post.uri, hour_age, likes, score, follow_person ORDER BY score DESC, hour_age ASC, post.indexedAt DESC LIMIT 100;"
```
Josipmrden, line 6:
```suggestion
query = "MATCH (my_person:Person)-[:FOLLOW]->(follow_person:Person) MATCH (follow_person)-[: LIKE]->(post:Post) WHERE post.indexedAt IS NOT NULL AND NOT exists((post)-[:ROOT]->(:Post)) WITH localDateTime() - post.indexedAt as duration, post, follow_person WHERE duration.day < 5 WITH (duration.day * 24) + duration.hour as hour_age, post, follow_person ORDER BY post.indexedAt DESC LIMIT 500 MATCH(: Person) - [l: LIKE] -> (post) WITH count(l) as likes, hour_age, post, follow_person CALL libmodule_test.hacker_news(likes, 123, 4.1) YIELD score RETURN ID(post), post.uri, hour_age, likes, score, follow_person ORDER BY score DESC, hour_age ASC, post.indexedAt DESC LIMIT 100;"
```

tests/e2e/concurrent_query_modules/test_query_modules/module_test.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include <thread>
+
+const char *kProcedureHackerNews = "hacker_news";
+const char *kArgumentHackerNewsVotes = "votes";
+const char *kArgumentHackerNewsItemHourAge = "item_hour_age";
+const char *kArgumentHackerNewsGravity = "gravity";
+const char *kReturnHackerNewsScore = "score";
+
+void HackerNews(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory) {
+  mgp::MemoryDispatcherGuard guard(memory);
```
Josipmrden, line 26: we will also need to update MAGE docs because the way we have the How To Guide in the docs, they are suggesting the old way of handling the memory pointeer.

tests/e2e/concurrent_query_modules/test_query_modules/module_test.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include <chrono>
+#include <cmath>
+#include <list>
+#include <thread>
+
+const char *kProcedureHackerNews = "hacker_news";
+const char *kArgumentHackerNewsVotes = "votes";
+const char *kArgumentHackerNewsItemHourAge = "item_hour_age";
+const char *kArgumentHackerNewsGravity = "gravity";
+const char *kReturnHackerNewsScore = "score";
```
Josipmrden, line 23:
```suggestion
constexpr char *kProcedureHackerNews = "hacker_news";
constexpr char *kArgumentHackerNewsVotes = "votes";
constexpr char *kArgumentHackerNewsItemHourAge = "item_hour_age";
constexpr char *kArgumentHackerNewsGravity = "gravity";
constexpr char *kReturnHackerNewsScore = "score";
```

Josipmrden (commented):
Some comments but overall good job!

query_modules/example.cpp
gvolfing: Yes, the MAGE modules will have to be modified as well, we discussed it with @gitbuda and @antoniofilipovic, we will get to it once Fico is back.

tests/e2e/concurrent_query_modules/client.py
gvolfing: I didn't know that, but I would prefer to keep it this way if that's fine by you, I am not sure if there is some loss of control with the added abstraction of GQLAlchemy, if we want to extend these tests in the future. Let me know what you think!

Josipmrden (approved):

PR #1172, opened 2023-08-17
[T1171 < T1172] Add high write set property workload

Josipmrden:
Added a new workload for performance benchmarking of high set property workloads and alerting if future fixes degrade system performance.

[master < Task] PR
- [X] Check, and update documentation if necessary (not needed)
- [X] Provide the full content or a guide for the final git message (not needed)


tests/mgbench/workloads/high_write_set_property.py
```diff
@@ -0,0 +1,103 @@
...
+from workloads.base import Workload
+
+
+class HighWriteSetProperty(Workload):
+    NAME = "high_write_set_property"
+    CARDINALITY = 100000
+
+    def indexes_generator(self):
+        indexes = []
+        if "neo4j" in self.benchmark_context.vendor_name:
```
antejavor, line 21: There is no need to add Neo4j stuff for things that will not run on Neo4j. It is ok to skip that and return the Memgraph stuff.

antejavor (approved):
Looks good. We need to see how we are going to name this stuff in grafana, since both workloads are single-threaded and the rest of the stuff was multithreaded. Just to keep thinks more organized.

tests/mgbench/workloads/high_write_set_property.py
```diff
@@ -0,0 +1,103 @@
...
+""",
+                    {"id": i},
+                )
+            )
+
+        return queries
+
+    def benchmark__test__write_20_attributes_set_property(self):
+        return (
+            "MATCH (n:Node) SET n.prop1 = n.prop2, n.prop2 = n.prop3, n.prop3 = n.prop4, n.prop4 = n.prop5, n.prop5 = n.prop6, n.prop6 = n.prop7, n.prop7 = n.prop8, n.prop8 = n.prop9, n.prop9 = n.prop10, n.prop10 = n.prop11, n.prop11 = n.prop12, n.prop12 = n.prop13, n.prop13 = n.prop14, n.prop14 = n.prop15, n.prop15 = n.prop16, n.prop16 = n.prop17, n.prop17 = n.prop18, n.prop18 = n.prop19, n.prop19 = n.prop20, n.prop20 = n.prop1;",
```
antejavor, line 95: Just an extra comment to keep you aware, so the test will run this query for 1000x times, and then calculate QPS, memory, etc. Should the Node be distinct in each of that queries? 

Not getting and setting each node in query, since then you are writing in just one query.

vpavicic:
@Josipmrden - Do I need to document anything here?

Issue #874, opened 2023-04-12
Query failed: Connection was closed by server


ljhhuxiaoba:
I executed a query on my memgraph database and the connection closed.

memgraph version:2.7.0
operating system: windows 10
API：Cypher version 9
query:SET (CASE WHEN true THEN null END).a = 1
error message:Query failed: Connection was closed by server

gitbuda:
@ljhhuxiaoba This is a very nice finding! There is a segmentation fault on the Memgraph side. We'll try to include the fix in the next release.

gitbuda:
take a look at the @Ignition comment inside #1175 and figure out the root cause to close this issue

ljhhuxiaoba:
Hello, has this problem been fixed?

gitbuda:
Hi! This particular issue has been fixed under https://github.com/memgraph/memgraph/pull/1175 but we suspect there are more cases like this, and we have to figure out a proper fix across the board 💪 We plan to start that work soon 😃

PR #1175, opened 2023-08-18
[master < ] Fix segfault based on issue #874

DavIvek:
This PR is about handling queries in which we want to set a property of a null label, for example, `SET null.a = 1`. This type of query would cause a segmentation fault and would crash the database. To prevent this, QueryRunTimeException will be thrown.

[master < Task] PR

@vpavicic


gitbuda:
* the error string can be a bit more condensed/packed (smaller)
* in the error message user won't understand what an operator is (if it's not known from before), we should like this part of memgraph docs -> https://memgraph.com/docs/memgraph/reference-guide/inspecting-queries#operators (you should use the shortened link)

Ignition (commented):
It would be nice to have better insight into the root cause.
This is a problem with operator creation.
`SetProperty` can handle null values at execution, for example `UNWIND [null] AS x SET x.a = 1` is fine. So what is it about construction of operators that has an issue with null literals for example `SET null.a = 1`?

Josipmrden:
No comments from my side on the code except:
- branch name could be better (e.g. fix-crash-on-null-values-for-switch-case)
- Update the info for PR describing what is the fix about and on what query it works.

DavIvek:
Even though throwing this exception isn't the solution to a problem with operator creation, it will prevent segmentation fault. We can leave issue #874 open and for mg-v2.12.0 find the root cause of the problem.

Ignition (approved):

PR #1178, opened 2023-08-21
[master < T1178] Sort SHOW INDEX INFO

antepusic:
Quick feature that sorts index information by type, label and property (in that order).

[master < Task] PR
* [docs#1022](https://github.com/memgraph/docs/pull/1022)
> Sort SHOW INDEX INFO results alphabetically

> [`SHOW INDEX INFO`](https://memgraph.com/docs/memgraph/reference-guide/indexing#display-available-indexes) now displays index information in alphabetic order for easy orientation.


src/query/interpreter.cpp
```diff
@@ -2922,6 +2922,31 @@ PreparedQuery PrepareInfoQuery(ParsedQuery parsed_query, bool in_explicit_transa
           results.push_back({TypedValue("label+property"), TypedValue(db->LabelToName(item.first)),
                              TypedValue(db->PropertyToName(item.second))});
         }
+
+        std::sort(results.begin(), results.end(), [](const auto record_1, const auto record_2) {
```
as51340, line 2926:
```suggestion
        std::sort(results.begin(), results.end(), [](const auto &record_1, const auto &record_2) {
```

src/query/interpreter.cpp
```diff
@@ -2922,6 +2922,31 @@ PreparedQuery PrepareInfoQuery(ParsedQuery parsed_query, bool in_explicit_transa
...
+          const auto type_1 = record_1[0].ValueString();
+          const auto type_2 = record_2[0].ValueString();
+
+          if (type_1 != type_2) {
+            return type_1 < type_2;
```
as51340, line 2931: This means that label entry will always come before label-property right?

src/query/interpreter.cpp
```diff
@@ -2922,6 +2922,31 @@ PreparedQuery PrepareInfoQuery(ParsedQuery parsed_query, bool in_explicit_transa
...
+          }
+
+          if (type_1 == "label") {
```
as51340, line 2934: "label" should be some constant, I think I added it somewhere in the disk storage, try to find it and reuse, if not add your own.

src/query/interpreter.cpp
```diff
@@ -2922,6 +2922,31 @@ PreparedQuery PrepareInfoQuery(ParsedQuery parsed_query, bool in_explicit_transa
...
+          }
+
+          if (type_1 == "label+property") {
+            const auto label_1 = record_1[1].ValueString();
+            const auto label_2 = record_2[1].ValueString();
+            if (label_1 == label_2) {
+              return record_1[2].ValueString() < record_2[2].ValueString();
+            }
+            return label_1 < label_2;
+          }
```
as51340, line 2945:
```suggestion
    const auto label_1 = record_1[1].ValueString();
    const auto label_2 = record_2[1].ValueString();
    if (label_1 != label_2 || type_1 == "label") {
        return label_1 < label_2;
     }
     record_1[2].ValueString() < record_2[2].ValueString(); /// compare properties only if labels are the same && you have label-property indices. I think this is shorter than your code but check the logic please maybe I am wrong.
```

tests/e2e/show_index_info/test_show_index_info.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def test_show_index_info(cursor):
+    index_info = set(execute_and_fetch_all(cursor, "SHOW INDEX INFO;"))
+    expected_index_info = {
+        ("label", "Anatomy", None),
+        ("label", "Gene", None),
+        ("label+property", "Compound", "inchikey"),
+        ("label+property", "Gene", "id"),
+        ("label", "Compound", None),
```
as51340, line 25: shouldn't label index come before label-property index because of 
```
if (type_1 != type_2) {
    return type_1 < type_2;
}
```

as51340 (changes requested):

src/query/interpreter.cpp
antepusic: Right

tests/e2e/show_index_info/test_show_index_info.py
antepusic: This test only makes sure that all indices are returned by `SHOW INDEX INFO`. The next test (`test_index_info_sorted`) is about output sorting.

src/query/interpreter.cpp
antepusic: This makes sense, thanks!

tests/e2e/show_index_info/test_show_index_info.py
antepusic: Clarified

PR #1179, opened 2023-08-21
Add timestamp to DELETE_DESERIALIZED_OBJECT delta at which this object was created.

Darych:
[master < T] PR
- [ ] Write E2E tests

[master < Task] PR


andrejtonev (commented):
Just a couple of small comments. Still need to dive a bit deeper tomorrow.

src/utils/disk_utils.hpp
```diff
@@ -25,4 +25,13 @@ inline std::optional<std::string> GetOldDiskKeyOrNull(storage::Delta *head) {
   return std::nullopt;
 }
 
+inline uint64_t GetEarliestTimestamp(storage::Delta *head) {
+  uint64_t ts = head->timestamp->load(std::memory_order_acquire);
+  while (head->next != nullptr) {
+    head = head->next;
```
andrejtonev, line 31: `head = head->next` must be done after reading `head->timestamp` since `next` could be null. Just set ts to max and reorded the lines in the loop.

src/storage/v2/delta.hpp
```diff
@@ -154,9 +154,14 @@ struct Delta {
...
 
-  Delta(DeleteDeserializedObjectTag /*tag*/, std::atomic<uint64_t> *timestamp,
-        const std::optional<std::string> &old_disk_key)
-      : action(Action::DELETE_DESERIALIZED_OBJECT), timestamp(timestamp), command_id(0), old_disk_key(old_disk_key) {}
+  // DELETE_DESERIALIZED_OBJECT is used to load data from disk committed by past txs.
+  // Because of this object was created in past txs, we create timestamp by ourselves inside instead of having it from
+  // current tx. This timestamp we got from RocksDB timestamp stored in key.
+  Delta(DeleteDeserializedObjectTag /*tag*/, uint64_t ts, const std::optional<std::string> &old_disk_key)
+      : action(Action::DELETE_DESERIALIZED_OBJECT),
+        timestamp(new std::atomic<uint64_t>(ts)),
```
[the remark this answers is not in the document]
Darych: We use pointer to std::atomic<uint64_t> because this is a pointer to transaction commit_timestamp. When Memgraph sets commit timestamp in tx, all deltas automatically got updated timestamp :) Tricky hack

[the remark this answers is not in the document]
andrejtonev: Ah okay, I get it. I was just looking at the changes and didn't understand why are we using an atomic here.

src/storage/v2/disk/storage.hpp
```diff
@@ -22,6 +22,8 @@
 #include "utils/rw_lock.hpp"
 
 #include <rocksdb/db.h>
+#include <rocksdb/slice.h>
+#include <cstdint>
```
as51340, line 26: Do you need these imports?

src/storage/v2/mvcc.hpp
```diff
@@ -12,11 +12,13 @@
 #pragma once
 
 #include <atomic>
+#include <cstdint>
 #include <optional>
 
 #include "storage/v2/property_value.hpp"
 #include "storage/v2/transaction.hpp"
 #include "storage/v2/view.hpp"
+#include "utils/rocksdb_serialization.hpp"
```
as51340, line 21: I think you don't need it

src/storage/v2/mvcc.hpp
```diff
@@ -12,11 +12,13 @@
 #pragma once
 
 #include <atomic>
+#include <cstdint>
```
as51340, line 15: I think you don't need it

src/utils/disk_utils.hpp
```diff
@@ -25,4 +25,13 @@ inline std::optional<std::string> GetOldDiskKeyOrNull(storage::Delta *head) {
...
+    ts = std::min(ts, head->timestamp->load(std::memory_order_acquire));
```
as51340, line 32: Why min, shouldn't the next delta always be older

as51340 (changes requested):

src/storage/v2/disk/storage.cpp
```diff
@@ -408,7 +413,9 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(View view) {
   auto it =
       std::unique_ptr<rocksdb::Iterator>(disk_transaction_->GetIterator(ro, disk_storage->kvstore_->vertex_chandle));
   for (it->SeekToFirst(); it->Valid(); it->Next()) {
-    LoadVertexToMainMemoryCache(it->key().ToString(), it->value().ToString());
+    // We should pass it->timestamp().ToString() instead of "0"
+    // This is hack until RocksDB will support timestamp() in WBWI iterator
+    LoadVertexToMainMemoryCache(it->key().ToString(), it->value().ToString(), "0");
```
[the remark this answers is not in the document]
andrejtonev: Or just make it the default for the argument.

src/utils/disk_utils.hpp
Darych: We are looking for the earliest timestamp.
But probably that's wrong and we should look for the latest.
Because if the object was changed in transaction we should work with the latest timestamp in mvcc functions.

andrejtonev: Not sure what exactly you are trying to do, but know that there is a possibility that the oldest object may be scheduled for garbage collection. In any case, the memory won't get released if the gc can't get an exclusive lock.
Can this impact you in any way?

as51340: there is no gc in disk storage, as long as the transaction exists, objects are alive

Darych: We found a bug: https://github.com/memgraph/memgraph/issues/1195
Should be addressed separately

andrejtonev (approved):

src/utils/disk_utils.hpp
```diff
@@ -25,4 +25,14 @@ inline std::optional<std::string> GetOldDiskKeyOrNull(storage::Delta *head) {
   return std::nullopt;
 }
 
+inline uint64_t GetEarliestTimestamp(storage::Delta *head) {
+  if (head == nullptr) return 0;
+  uint64_t ts = head->timestamp->load(std::memory_order_acquire);
+  while (head->next != nullptr) {
+    head = head->next;
+    ts = std::min(ts, head->timestamp->load(std::memory_order_acquire));
```
as51340, line 33: why min? why not just iterate until you found the last delta? Last delta should always be earliest

as51340 (changes requested):

src/utils/disk_utils.hpp
Darych: Agree

as51340 (approved):

PR #1187, opened 2023-08-22
[master <] Introduce a reader writer spin lock

### LINKED ISSUE #1189: Reduce the spinlock contention
https://gist.github.com/Ignition/7e4631b874fb75ffef09e776fa1e68ee

Ignition:
There are situations where multiple readonly queries in parallel are contended on the same vertex lock. Take for example a highly connected supernode which is repetitively used in result values and it needs encode it to a bolt value for sending to the client.

Since `SpinLock` can only be used exclusively, it causes unnecessary blocking. `RWSpinLock` would allow shared reading.

Before: you can see heavy use of `__pthread_spin_lock`
before

After: in the same part of the profile the locks don't even appear, this is because of the `RWSpinLock`
after

[master < Task] PR

closes #1189

### DIFF
--- a/src/utils/rw_spin_lock.hpp
+++ b/src/utils/rw_spin_lock.hpp
@@ -1,0 +1,124 @@
+// Copyright 2023 Memgraph Ltd.
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
+// License, and you may not use this file except in compliance with the Business Source License.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+#pragma once
+
+#include <time.h>
+#include <algorithm>
+#include <atomic>
+#include <cstdint>
+
+namespace memgraph::utils {
+namespace {
+/// A helper for RWSpinLock, allows a contended spin lock to yield to another thread.
+struct yeilder {
+  void operator()() noexcept {
+#if defined(__i386__) || defined(__x86_64__)
+    // TODO: make portable
+    __builtin_ia32_pause();
+#endif
+    ++count;
+    if (count > 8) [[unlikely]] {
+      count = 0;
+      nanosleep(&shortpause, nullptr);
+      // Increase the backoff
+      shortpause.tv_nsec = std::min<decltype(shortpause.tv_nsec)>(shortpause.tv_nsec << 1, 512);
+    }
+  }
+
+ private:
+  uint_fast32_t count{0};
+  timespec shortpause = {.tv_sec = 0, .tv_nsec = 1};
+};
+}  // namespace
+
+/**
+ * A reader/writer spin lock.
+ * Stores in a uint32_t,
+ * 0x0000'0001 - is the write bit
+ * rest of the bits hold the count for the number of current writers.
+ * The lock is friendly to writers.
+ * - writer lock() will wait for all readers to leave unlock_shared()
+ * - new reader lock_shared() will wait until writer has unlock()
+ **/
+struct RWSpinLock {
+  RWSpinLock() = default;
+
+  void lock() {
+    // spin: to grant the UNIQUE_LOCKED bit
+    while (true) {
+      // optimistic: assume we will be granted the lock
+      auto const phase1 = std::atomic_ref{lock_status_}.fetch_or(UNIQUE_LOCKED, std::memory_order_acq_rel);
+      // check: we were granted UNIQUE_LOCK and no current readers
+      if (phase1 == 0) [[likely]]
+        return;
+      // check: we were granted UNIQUE_LOCK, but need to wait for readers
+      if ((phase1 & UNIQUE_LOCKED) != UNIQUE_LOCKED) [[likely]]
+        break;
+
+      // spin: to wait for UNIQUE_LOCKED to be available
+      auto maybe_yield = yeilder{};
+      while (true) {
+        auto const phase2 = std::atomic_ref{lock_status_}.load(std::memory_order_relaxed);
+        // check: we are able to obtain UNIQUE_LOCK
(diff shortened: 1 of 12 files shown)


src/utils/rw_spin_lock.hpp
```diff
@@ -0,0 +1,114 @@
...
+ * - new reader lock_shared() will wait until writer has unlock()
+ **/
+struct RWSpinLock {
+  RWSpinLock() = default;
+
+  void lock() {
+    // spin: to grant the UNIQUE_LOCKED bit
+    while (true) {
+      // optimistic: assume we will be granted the lock
+      auto const phase1 = std::atomic_ref{lockStatus_}.fetch_or(UNIQUE_LOCKED, std::memory_order_acq_rel);
```
as51340, line 53: can we be explicit about the type of phase1 and phase2, I find it much easier to read the code?

src/utils/rw_spin_lock.hpp
```diff
@@ -0,0 +1,114 @@
...
+      // check: we were granted UNIQUE_LOCK and no current readers
+      if (phase1 == 0) [[likely]]
+        return;
+      // check: we were granted UNIQUE_LOCK, but need to wait for readers
+      if (!(phase1 & UNIQUE_LOCKED)) [[likely]]
```
as51340, line 58: If I understand correctly this is equivalent to:
```suggestion
      if ((phase1 & UNIQUE_LOCKED) == 0) [[likely]]
```
But this can only happen if phase1 == 0 right? But if this can happen only then, this code won't ever get executed

as51340: TODO: andi wait

as51340: And can we do explicit check == 0? Maybe it is easier to understand than with ! in the context of working with bits

src/utils/rw_spin_lock.hpp
```diff
@@ -0,0 +1,114 @@
...
+ private:
+  using status_t = uint32_t;
+  enum FLAGS : status_t {
+    UNIQUE_LOCKED = 1,
+    READER = 2,
+  };
+
+  // TODO: ATM not atomic, just used via atomic_ref, because the type needs to be movable into skip_list
+  //       fix the design flaw and then make RWSpinLock a non-copy/non-move type
+  status_t lockStatus_ = 0;
```
as51340, line 112:
```suggestion
  status_t lock_status_ = 0;
```

src/storage/v2/disk/storage.cpp
```diff
@@ -90,7 +90,7 @@ bool VertexHasLabel(const Vertex &vertex, LabelId label, Transaction *transactio
   bool has_label = false;
```
as51340, line 90: Just a note: in a separate PR taking locks on vertices and edges in disk/storage.cpp will be removed since it is not needed

as51340 (changes requested):
Great job, I like it very much! Only think which I don't like is the usage of [[likely]]. I think it is impossible to predict the workload and hence we could actually lose performance by using it.

as51340:
@Josipmrden I used [this](https://godbolt.org/z/3c3a33ss8) to play a bit with it to understand exactly what is going on. Maybe you will find it useful, if not ignore

src/utils/rw_spin_lock.hpp
Ignition: In my opinion all options require careful reading. They all generate the same assembly.
```cpp
!(phase1 & UNIQUE_LOCKED)
(~phase1) & UNIQUE_LOCKED
(phase1 & UNIQUE_LOCKED) == 0
(phase1 & UNIQUE_LOCKED) != UNIQUE_LOCKED
```

My take is maybe `(phase1 & UNIQUE_LOCKED) != UNIQUE_LOCKED` communicates the best.

src/utils/rw_spin_lock.hpp
```diff
@@ -0,0 +1,124 @@
...
+// included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
+// License, and you may not use this file except in compliance with the Business Source License.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+#pragma once
+
+#include <time.h>
```
Josipmrden, line 13: is there a c++ library for this?

Josipmrden: ctime?

src/utils/rw_spin_lock.hpp
```diff
@@ -0,0 +1,124 @@
...
+#if defined(__i386__) || defined(__x86_64__)
+    // TODO: make portable
+    __builtin_ia32_pause();
+#endif
+    ++count;
+    if (count > 8) [[unlikely]] {
+      count = 0;
+      nanosleep(&shortpause, nullptr);
+      // Increase the backoff
+      shortpause.tv_nsec = std::min<decltype(shortpause.tv_nsec)>(shortpause.tv_nsec << 1, 512);
```
Josipmrden, line 32: can this overflow?

Josipmrden (approved):

src/utils/rw_spin_lock.hpp
Ignition: nanosleep is not part of the C standard, using time.h as a deliberate indication of non-portability.

src/utils/rw_spin_lock.hpp
Ignition: > tv_nsec | nanoseconds (valid values are [0, 999999999])

Hence the storage of `tv_nsec` is large enough for the `<< 1` operation

`shortpause.tv_nsec` will be capped to 512 so will never overflow

src/storage/v2/disk/storage.cpp
Ignition: After rebase all changes to on disk have been removed

PR #1193, opened 2023-08-23
Return cartesian product on multi match clauses

Josipmrden:
In multi match clauses, the result of 2 matches equals a cartesian product of rows. If both match clauses have filters, then there is a quadratic amount of filtering operations in the system.

[master < Task] PR


src/query/plan/pretty_print.cpp
```diff
@@ -188,12 +187,28 @@ bool PlanPrinter::PreVisit(query::plan::Filter &op) {
...
 }
 
+bool PlanPrinter::PreVisit(query::plan::EdgeUniquenessFilter &op) {
+  WithPrintLn([&](auto &out) {
+    out << "* EdgeUniquenessFilter [";
+    utils::PrintIterable(out, op.previous_symbols_, ", ", [](auto &out, const auto &sym) { out << sym.name(); });
+    out << " != " << op.expand_symbol_.name() << "]";
+  });
+  return true;
+}
```
antepusic, line 197: I suggest implementing `PlanPrinter::Previsit` for this operator with a `ToString` method, consistently with other operators that have detailed info (see [PR](https://github.com/memgraph/memgraph/pull/1204)).

tests/unit/query_plan.cpp
```diff
@@ -280,8 +280,43 @@ TYPED_TEST(TestPlanner, MatchMultiPattern) {
...
+                                                    new ExpectExpand()};
+
+  auto symbol_table = memgraph::query::MakeSymbolTable(query);
+  auto planner = MakePlanner<TypeParam>(&dba, this->storage, symbol_table, query);
+  CheckPlan(planner.plan(), symbol_table, ExpectIndexedJoin(left_indexed_join_ops, right_indexed_join_ops),
+            ExpectEdgeUniquenessFilter(), ExpectProduce());
+
+  DeleteListContent(&left_indexed_join_ops);
+  DeleteListContent(&right_indexed_join_ops);
 }
```
antepusic, line 320: What’s the reason as to why only one branch of `IndexedJoin` contains `ScanAllByLabelPropertyValue`, whereas the other one does `ScanAllByLabel`?

src/query/plan/preprocess.cpp
```diff
@@ -55,35 +56,127 @@ void ForEachPattern(Pattern &pattern, std::function<void(NodeAtom *)> base,
...
+    matching.number_of_expansion_groups = matching.number_of_expansion_groups < expansion.expansion_group_id.AsUint()
+                                              ? expansion.expansion_group_id.AsUint()
+                                              : matching.number_of_expansion_groups;
+    next_expansion_group_id = ExpansionGroupId::FromUint(matching.number_of_expansion_groups + 1);
+  }
+
+  // By the time we finished assigning expansions, no expansion should have its expansion group ID unassigned
+  for (const auto &expansion : matching.expansions) {
+    MG_ASSERT(expansion.expansion_group_id.AsInt() != -1,
+              "Expansion expansion group ID is not assigned to the pattern!");
```
antepusic, line 129:
```suggestion
              "Expansion group ID is not assigned to the pattern!");
```

antepusic (changes requested):
Good job! I’ve left a few comments below:

antoniofilipovic (changes requested):
I checked PR and don't have many comments. Good work, left few suggestions otherwise it seems fine to me

antoniofilipovic (commented):
I checked PR and don't have many comments. Good work, left few suggestions otherwise it seems fine to me

antepusic (changes requested):
Having investigated, the checks are failing because `cartesian_product_enabled_` is turned off during unit test execution.

antepusic (commented):
This looks good and very close to mergeable, and I’ve resolved some of my earlier comments. Before I hit approve, please respond to the remaining two comments I had left :smile:

antepusic (approved):
Everything looks good to me now, and the checks are now passing :rocket:

antoniofilipovic (approved):
Looks good to me

PR #1199, opened 2023-08-27
[master < T1199] Enable importing edges and vertexes with id

imilinovic:
[master < Epic] PR
- [ ] Check, and update documentation if necessary
- [ ] Write E2E tests
- [ ] Compare the [benchmarking results](https://bench-graph.memgraph.com/) between the master branch and the Epic branch
- [ ] Provide the full content or a guide for the final git message

[master < Task] PR
- [ ] Check, and update documentation if necessary
- [ ] Provide the full content or a guide for the final git message

To keep docs changelog up to date, one more thing to do:
- [ ] Write a release note here, including added/changed clauses
- [ ] Tag someone from docs team in the comments

PR #1208, opened 2023-08-29
[master < T1205] Change failing object insertion with gid to throw exception

imilinovic:
There is an idea to create MAGE modules that could import vertices and edges with their respective IDs. With the current implementation attempting to create a vertex/edge with an ID that already existed/exists would crash the Memgraph instance which is not ideal. Would modifying this to throw an exception instead be a problem (in replication)?

[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR

### DIFF
--- a/src/storage/v2/inmemory/storage.cpp
+++ b/src/storage/v2/inmemory/storage.cpp
@@ -483,7 +483,6 @@ Result<EdgeAccessor> InMemoryStorage::InMemoryAccessor::CreateEdgeEx(VertexAcces
     if (!inserted) {
       throw utils::BasicException("The edge must be inserted here");
     }
-
     MG_ASSERT(it != acc.end(), "Invalid Edge accessor!");
     edge = EdgeRef(&*it);
     if (delta) {
--- a/src/storage/v2/inmemory/storage.cpp
+++ b/src/storage/v2/inmemory/storage.cpp
@@ -18,6 +18,7 @@
 #include "storage/v2/edge_direction.hpp"
 #include "storage/v2/storage_mode.hpp"
 #include "storage/v2/vertex_accessor.hpp"
+#include "utils/exceptions.hpp"
 #include "utils/stat.hpp"
 
 /// REPLICATION ///
@@ -234,7 +235,9 @@ VertexAccessor InMemoryStorage::InMemoryAccessor::CreateVertexEx(storage::Gid gi
 
   auto *delta = CreateDeleteObjectDelta(&transaction_);
   auto [it, inserted] = acc.insert(Vertex{gid, delta});
-  MG_ASSERT(inserted, "The vertex must be inserted here!");
+  if (!inserted) {
+    throw utils::BasicException("The edge must be inserted here");
+  }
   MG_ASSERT(it != acc.end(), "Invalid Vertex accessor!");
   if (delta) {
     delta->prev.Set(&*it);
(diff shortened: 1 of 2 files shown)


imilinovic:
I tagged the two of you because I saw you worked on replication

src/storage/v2/inmemory/storage.cpp
```diff
@@ -234,7 +235,9 @@ VertexAccessor InMemoryStorage::InMemoryAccessor::CreateVertexEx(storage::Gid gi
 
   auto *delta = CreateDeleteObjectDelta(&transaction_);
   auto [it, inserted] = acc.insert(Vertex{gid, delta});
-  MG_ASSERT(inserted, "The vertex must be inserted here!");
+  if (!inserted) {
```
andrejtonev, line 238: Why do we want to throw and not assert?

Ignition:
This is a bad idea, the `Create*Ex` methods were protected to ensure they can only be used under a restricted context.

ie. we only want to force the gid when we are a `replica` and we are keeping in sync with `main`. In this case we:
- trust `main` to send the right info
- we ensure `replica` can't do writes, hence not possible to accidentally have gid issued to two different objects.

Having arbitrary MAGE dictate the gid is impossible to be safe as you have no guarantee that the gid has not already been issued.

src/storage/v2/inmemory/storage.hpp
```diff
@@ -76,6 +76,10 @@ class InMemoryStorage final : public Storage {
     /// @throw std::bad_alloc
     VertexAccessor CreateVertex() override;
 
+    // TODO Better naming
+    /// @throw std::bad_alloc
+    VertexAccessor CreateVertexEx(storage::Gid gid);
```
andrejtonev, line 81: `CreateVertexEx` and `CreateEdgeEx` was protected for a reason.
It is very dangerous and should only be used by the replication logic.
There is a special `ReplicationAccessor` for use in replication that enables these functions.
Why do you need these functions?

src/storage/v2/inmemory/storage.cpp
```diff
@@ -234,7 +235,9 @@ VertexAccessor InMemoryStorage::InMemoryAccessor::CreateVertexEx(storage::Gid gi
...
+    throw utils::BasicException("The edge must be inserted here");
```
gitbuda, line 239: + if we throw, the error message has to be much more descriptive / saying what should user do about this + assist him

in addition, the convention on a storage layer is not to `throw` because of the issue related to exceptions, consider returning `optional/result<VertexAccessor>` instead of plain `VertexAccessor` to force the error handling on the caller side... this is also not ideal because we'll change the storage API (tricky)

CC @antoniofilipovic not sure how critical is this, but we probably need a design sync here 😃

antoniofilipovic:
@gitbuda @Ignition @andrejtonev This started as an idea not to crash memgraph completely when importing data with JSON until when IDs are sent as part of JSON file for vertex and edge together with properties and labels to be created for that given vertex. 

@imilinovic opened the following PR which would enable such behaviour: https://github.com/memgraph/memgraph/pull/1199 but if memgraph already had those IDs covered, it would crash completely. In my opinion, handling "wrong id" should be done on the replication side, not in this function as it is not aware of replica logic. Maybe I am wrong, that's why PR is here.

Backstory: This all started as an idea, because Neo4j enables import and creating vertices with the given ID, whereas we don't enable that: https://neo4j.com/labs/apoc/4.1/import/load-json/

src/storage/v2/inmemory/storage.cpp
antoniofilipovic: It is not critical, but as I described here https://github.com/memgraph/memgraph/pull/1208#issuecomment-1697233546 it started as an idea to support loading JSON with given IDs in file as Neo4j supports it. I agree that instead of error, optional/Result<VertexAccessor> is better than throwing error.

andrejtonev:
This seems like a replication workflow. 
I don't see another case where we would want to overwrite the IDs.
If it is replication, then we should enable this logic on the replication side and use the `ReplicationAccessor`. 
I think in general asserting on overwrites is the correct approach, since something very bad is happening.

gitbuda:
also closing because it was relate to the other closed PR + the storage API change is tricky

PR #1209, opened 2023-08-29
[master < T1209] Fix API bug on accessing deleted object

antoniofilipovic:
Issue: When doing detach delete on graph, and accessing deleted objects from C++ API, if you are accessing `From` and `To` vertex of deleted edge, error occurs that you are accessing deleted vertex although it shouldn't.

When a vertex is detach deleted, its VertexAccessor is marked `for_deleted_`, and  out|in EdgeAccessors are marked `for_deleted_`. VertexAccessors which you can access from EdgeAccessor are not marked for_deleted_.  This creates a slight problem when working with deleted objects in API, because only EdgeAccessor is marked `for_deleted_` not also its `VertexAccessors`.  Therefore some `VertexAccessors` have wrong information, marking `for_deleted_` as false instead of true.

[master < Task] PR


antepusic (commented):
Everything looks fine, but I’d prefer to overload `FromVertex` and `ToVertex` instead of creating new methods. The new ones’ names (`FromVertexWithDeleted` and `ToVertexWithDeleted`) really aren’t very clear, and tbh I can’t think of a name that’s both clear and descriptive.

src/query/db_accessor.hpp
```diff
@@ -94,6 +94,10 @@ class EdgeAccessor final {
 
   VertexAccessor From() const;
 
+  VertexAccessor ToWithDeleted() const;
```
Josipmrden, line 97: What does this mean

src/storage/v2/edge_accessor.cpp
```diff
@@ -105,6 +105,15 @@ VertexAccessor EdgeAccessor::ToVertex() const {
   return VertexAccessor{to_vertex_, transaction_, indices_, constraints_, config_};
 }
 
+VertexAccessor EdgeAccessor::FromVertexWithDeleted() const {
+  return VertexAccessor{from_vertex_, transaction_, indices_,
+                        constraints_, config_,      for_deleted_ && from_vertex_->deleted};
```
Josipmrden, line 110: I'm seeing a lot of spaces here

src/query/db_accessor.hpp
Josipmrden: I'd like this to be documented

Josipmrden (approved):

vpavicic:
@antoniofilipovic - can you write me a release note (RN)?

antoniofilipovic:
This PR enables to access deleted vertex from deleted edge when being processed as part of trigger function

PR #1210, opened 2023-08-30
[master < T367-MAGE] Add traits under list iterator

ind1xa:
Idea of this PR is to make mgp::Iterator compatible with standard library traits for iterators.

[master < Task] PR


antoniofilipovic (approved):
Looks good to me

vpavicic:
Is this a fix in the C++ API? :D 

So the fix is that:
Now you can construct std::set of values (find unique values) by iterating over mgp::List as expected, (i.e. without any errors) and successfully perform any other operations dependent on the STL container requirements.

(i understand very little :D so i hope it's technically correct)

PR #1217, opened 2023-09-01
[master < T1217] Implement InDegree and OutDegree in O(1)

mpintaric55334:
While implementing refactor functions, we figured it would be good to implement In and Out degrees to use already made functions in c-api, so the degree getting operation is O(1).

[master < Task] PR

Users can now get the In and Out degrees of a node in O(1) in Cpp API.


antoniofilipovic (changes requested):
All good but would modfiy tests

PR #1225, opened 2023-09-05
[master < T1225] Fix bug on mgp dispatcher guard

antoniofilipovic:
This PR aims to fix the issue with calling allocations using mgp::memory without that variable being initialized. Even though mgp::MemorDispatcher is initialized, and mgp::memory is not, it gets called.

[master < Task] PR


antepusic (approved):
I think this one is all right

vpavicic:
@antoniofilipovic - can you please write me a release note (RN)?

antoniofilipovic:
Yes, sorry @vpavicic 

Query modules can now use mgp::DispatcherGuard without bug occurring. Problem was that wrong memory resource was being used, standard mgp::memory where mgp::DispatcherGuard should have been used

PR #1226, opened 2023-09-06
Add HashJoin operator

Josipmrden:
This PR implements the `HashJoin` operator and extends the query rewriter to convert `Cartesian` followed by `Filter` into `HashJoin`. This speeds up queries where `Filter` discards the majority of `Cartesian`’s output.

NB (by @antepusic): As this PR has a lot in common with #1193, I’ve set the base branch accordingly to make code review easier. After review, this PR should in any case be merged to `master`, and not the #1193 branch.

[master < Task] PR
> Implement the HashJoin operator to optimize Filter + Cartesian sequences

> Queries that have multiple `MATCH` clauses and join their results in a `WHERE` clause should now run more efficiently owing to the new HashJoin operator.


src/query/plan/rewrite/join.hpp
```diff
@@ -0,0 +1,547 @@
...
+      if (filter.property_filter->type_ != PropertyFilter::Type::EQUAL) {
+        continue;
+      }
+
+      if (filter.property_filter->value_->GetTypeInfo() != PropertyLookup::kType) {
+        continue;
+      }
+      auto rhs_lookup = static_cast<PropertyLookup *>(filter.property_filter->value_);
+
+      auto join_condition = static_cast<EqualOperator *>(filter.expression);
```
antoniofilipovic, line 519: you can use here `auto *` to empasize pointer

src/query/plan/rewrite/join.hpp
```diff
@@ -0,0 +1,547 @@
...
+}  // namespace impl
+
+template <class TDbAccessor>
+std::unique_ptr<LogicalOperator> RewriteWithJoinRewriter(std::unique_ptr<LogicalOperator> root_op,
+                                                         SymbolTable *symbol_table, AstStorage *ast_storage,
+                                                         TDbAccessor *db) {
+  impl::JoinRewriter<TDbAccessor> rewriter(symbol_table, ast_storage, db);
+  root_op->Accept(rewriter);
+  if (rewriter.new_root_) {
+    throw utils::NotYetImplemented("optimizing join");
```
antoniofilipovic, line 542: I would be more descriptive with message here. Can you elaborate what is not yet implemented here

src/query/plan/operator.cpp
```diff
@@ -5222,4 +5223,133 @@ void IndexedJoin::IndexedJoinCursor::Reset() {
...
+  }
+
+  void Reset() override {
+    left_op_cursor_->Reset();
+    right_op_cursor_->Reset();
+    hashtable_.clear();
+    right_op_frame_.clear();
+    hash_join_initialized_ = false;
+    common_value_found_ = false;
+  }
```
antoniofilipovic, line 5333: `left_op_frame_it_` is missing here

src/query/plan/operator.cpp
```diff
@@ -5222,4 +5223,133 @@ void IndexedJoin::IndexedJoinCursor::Reset() {
...
+
+  bool Pull(Frame &frame, ExecutionContext &context) override {
+    SCOPED_PROFILE_OP("HashJoin");
+
+    if (!hash_join_initialized_) {
+      // Pull all left_op_ frames
+      while (left_op_cursor_->Pull(frame, context)) {
+        ExpressionEvaluator evaluator(&frame, context.symbol_table, context.evaluation_context, context.db_accessor,
+                                      storage::View::NEW);
```
antoniofilipovic, line 5263: Why NEW view, shouldn't this view be OLD as you are just matching? I am just not sure how this depends when you have WITH + MATCH, maybe in that case you need NEW view

src/query/plan/operator.cpp
```diff
@@ -5222,4 +5223,133 @@ void IndexedJoin::IndexedJoinCursor::Reset() {
...
+          left_op_frame_it_ = hashtable_[common_value].begin();
+          break;
+        }
+      }
+    } else {
+      // Restore the right frame ahead of restoring the left frame
+      restore_frame(self_.right_symbols_, right_op_frame_);
+    }
+
+    restore_frame(self_.left_symbols_, *left_op_frame_it_);
```
antoniofilipovic, line 5310: Why do you need restoring these two frames?

antoniofilipovic (changes requested):
All seems good, have couple of questions and suggestions

src/query/plan/operator.cpp
```diff
@@ -5222,4 +5223,133 @@ void IndexedJoin::IndexedJoinCursor::Reset() {
...
+      while (left_op_cursor_->Pull(frame, context)) {
+        ExpressionEvaluator evaluator(&frame, context.symbol_table, context.evaluation_context, context.db_accessor,
+                                      storage::View::NEW);
+        auto left_value = self_.hash_join_condition_->expression2_->Accept(evaluator);
+        if (left_value.type() != TypedValue::Type::Null) {
+          hashtable_[left_value].emplace_back(frame.elems().begin(), frame.elems().end());
+        }
+      }
+      hash_join_initialized_ = true;
+    }
```
Josipmrden, line 5270: definitely put this initialization of hash join into a private method as it is one unit of work

src/query/plan/operator.cpp
```diff
@@ -5222,4 +5223,133 @@ void IndexedJoin::IndexedJoinCursor::Reset() {
...
+
+    if (!common_value_found_) {
+      // Pull from the right_op until there’s a mergeable frame
+      while (true) {
+        auto pulled = right_op_cursor_->Pull(frame, context);
+        if (!pulled) return false;
+
+        // Check if the join value from the pulled frame is shared with any left frames
+        ExpressionEvaluator evaluator(&frame, context.symbol_table, context.evaluation_context, context.db_accessor,
+                                      storage::View::NEW);
```
Josipmrden, line 5294: also think of OLD

src/query/plan/operator.cpp
```diff
@@ -5222,4 +5223,133 @@ void IndexedJoin::IndexedJoinCursor::Reset() {
...
+
+  bool Pull(Frame &frame, ExecutionContext &context) override {
+    SCOPED_PROFILE_OP("HashJoin");
+
+    if (!hash_join_initialized_) {
+      // Pull all left_op_ frames
+      while (left_op_cursor_->Pull(frame, context)) {
+        ExpressionEvaluator evaluator(&frame, context.symbol_table, context.evaluation_context, context.db_accessor,
+                                      storage::View::NEW);
+        auto left_value = self_.hash_join_condition_->expression2_->Accept(evaluator);
```
Josipmrden, line 5264: why is left value with expression2 and not expression1?

Josipmrden: can you provide an example

src/query/plan/rewrite/join.hpp
```diff
@@ -0,0 +1,547 @@
...
+  }
+
+  std::unique_ptr<HashJoin> GenHashJoin(const Cartesian &cartesian) {
+    const auto &left_op = cartesian.left_op_;
+    const auto &left_symbols = cartesian.left_symbols_;
+    const auto &right_op = cartesian.right_op_;
+    const auto &right_symbols = cartesian.right_symbols_;
+
+    auto modified_symbols = cartesian.ModifiedSymbols(*symbol_table_);
+    modified_symbols.insert(modified_symbols.end(), cartesian.left_symbols_.begin(), cartesian.left_symbols_.end());
```
Josipmrden, line 489: you can also use `left_symbols` instead of `cartesian.left_symbols`

tests/gql_behave/tests/memgraph_V1/features/cartesian.feature
```diff
@@ -172,6 +172,111 @@ Feature: Cartesian
...
+            """
+            MATCH (a:A) MATCH (b:A) WHERE a.prop = b.prop RETURN a, b
+            """
+        Then the result should be:
+            | a                     | b                     |
+            | (:A {id: 1, prop: 1}) | (:A {id: 1, prop: 1}) |
+            | (:A {id: 1, prop: 1}) | (:A {id: 2, prop: 1}) |
+            | (:A {id: 2, prop: 2}) | (:A {id: 2, prop: 2}) |
+            | (:A {id: 2, prop: 1}) | (:A {id: 1, prop: 1}) |
+            | (:A {id: 2, prop: 1}) | (:A {id: 2, prop: 1}) |
```
Josipmrden, line 242: create a test with different props
e.g.
`MATCH (a:A) MATCH (b:A) WHERE a.prop = b.prop2 RETURN a, b`

Josipmrden: and reverse

tests/unit/query_plan.cpp
```diff
@@ -290,6 +290,31 @@ TYPED_TEST(TestPlanner, MatchMultiPattern) {
   DeleteListContent(&right_cartesian_ops);
 }
 
+TYPED_TEST(TestPlanner, MatchMultiPatternWithHashJoin) {
+  // Test MATCH (a:label)-[r1]->(b), (c:label)-[r2]->(d) WHERE c.id = a.id return a, b, c, d;
+  FakeDbAccessor dba;
+  const auto label_name = "label";
+  const auto property = PROPERTY_PAIR(dba, "id");
```
Josipmrden, line 297: also test with 2 different properties that need to match

src/query/plan/rewrite/join.hpp
antepusic: Added, thanks!

src/query/plan/rewrite/join.hpp
antepusic: Changed (@Josipmrden: maybe you need to do the same in #1193 in index_lookup.hpp)

src/query/plan/operator.cpp
antepusic: That’s intended – as `left_op_frame_it_` is an iterator over one value in `hashtable_`, it’s only ever used if `left_op_frame_it_ = hashtable_[common_value].begin();` has been run beforehand. Because of that, there’s no harm if I don’t reset it here.
Also, in `Reset`, if the `hashtable_` is empty to begin with, resetting `left_op_frame_it_` makes no sense because it has never been set.

src/query/plan/operator.cpp
antepusic: Makes sense (and fits with e.g. `FilterCursor`), thanks for the catch!

src/query/plan/operator.cpp
antepusic: Done

src/query/plan/operator.cpp
antepusic: Initially, the cursor pulls all left frames to build a hashtable that’s going to be used for joining. Next, in the join phase, the saved left frame needs to be restored so that it can appear in the results. Similarly, the right frames can be joined with multiple left frames so they need to be saved and restored when appropriate.

src/query/plan/operator.cpp
antepusic: Done

src/query/plan/rewrite/join.hpp
antepusic: Done

src/query/plan/operator.cpp
antepusic: I’ve fixed this, but it shouldn’t cause issues anyway because HashJoin only handles `=` filters.

The `expression1_`/`expression2_` swap is because I followed Cartesian’s pretty print method:
bool PlanPrinter::PreVisit(query::plan::Cartesian &op) {
  WithPrintLn([&op](auto &out) {
    out << "* Cartesian {";
    utils::PrintIterable(out, op.left_symbols_, ", ", [](auto &out, const auto &sym) { out << sym.name(); });
    out << " : ";
    utils::PrintIterable(out, op.right_symbols_, ", ", [](auto &out, const auto &sym) { out << sym.name(); });
    out << "}";
  });
  Branch(*op.right_op_);
  op.left_op_->Accept(*this);
  return false;
}
As you can see, it prints the left symbols first, but then it prints right_op in the main branch.

tests/gql_behave/tests/memgraph_V1/features/cartesian.feature
antepusic: Added

src/query/plan/operator.cpp
antepusic: The left values have to be taken from `expression2_` and the right ones from `expression_1`, otherwise queries that use HashJoin return an empty result (e.g. the "Multiple match with WHERE x = y" tests in [cartesian.feature](https://github.com/memgraph/memgraph/blob/7b08af958086b123f07aa0c2cb51ccb270ceb0d4/tests/gql_behave/tests/memgraph_V1/features/cartesian.feature))

tests/unit/query_plan.cpp
antepusic: Added

antepusic:

src/query/plan/operator.cpp
antoniofilipovic: I would reset it just for consistency. If someone later changes the code and wants to check if left_op_frame_it_ is still set, but it should be reset here, it would be a nasty bug.

src/query/plan/operator.cpp
```diff
@@ -5223,4 +5224,137 @@ void IndexedJoin::IndexedJoinCursor::Reset() {
...
+    right_op_cursor_->Reset();
+    hashtable_.clear();
+    right_op_frame_.clear();
+    hash_join_initialized_ = false;
+    common_value_found_ = false;
+  }
+
+ private:
+  void InitializeHashJoin(Frame &frame, ExecutionContext &context) {
+    // Pull all left_op_ frames
```
antoniofilipovic, line 5330: I think this is clear from code

src/query/plan/operator.hpp
```diff
@@ -2527,6 +2528,54 @@ class IndexedJoin : public memgraph::query::plan::LogicalOperator {
...
 };
 
+/// Operator for producing the hash join of two input branches
+class HashJoin : public memgraph::query::plan::LogicalOperator {
+ public:
+  static const utils::TypeInfo kType;
+  const utils::TypeInfo &GetTypeInfo() const override { return kType; }
+
+  HashJoin() {}
+  /** Construct the operator with left input branch and right input branch. */
```
antoniofilipovic, line 2538: I think this is also clear from code

antoniofilipovic (approved):
Just two more comments, I will approve so you don't need to ping me again

PR #1227, opened 2023-09-06
[master < T1227] Optimise iteration

Ignition:
For example following query:

```
UNWIND RANGE (1, 500) AS i CREATE ();
MATCH (),(),() RETURN COUNT(*);
```

went from `6.549` seconds to `3.625` seconds

- [ ] Write a release note here, including added/changed clauses --> "Optimizations to scanning vertices"

### DIFF
--- a/src/query/plan/operator.cpp
+++ b/src/query/plan/operator.cpp
@@ -438,7 +438,7 @@ class ScanAllCursor : public Cursor {
 
     AbortCheck(context);
 
-    while (!vertices_ || vertices_it_.value() == vertices_.value().end()) {
+    while (!vertices_ || vertices_it_.value() == vertices_end_it_.value()) {
       if (!input_cursor_->Pull(frame, context)) return false;
       // We need a getter function, because in case of exhausting a lazy
       // iterable, we cannot simply reset it by calling begin().
@@ -449,6 +449,7 @@ class ScanAllCursor : public Cursor {
       // vertices _ = get_vertices_(frame, context);
       vertices_.emplace(std::move(next_vertices.value()));
       vertices_it_.emplace(vertices_.value().begin());
+      vertices_end_it_.emplace(vertices_.value().end());
     }
 #ifdef MG_ENTERPRISE
     if (license::global_license_checker.IsEnterpriseValidFast() && context.auth_checker && !FindNextVertex(context)) {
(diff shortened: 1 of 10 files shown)


src/query/plan/operator.cpp
```diff
@@ -438,7 +438,7 @@ class ScanAllCursor : public Cursor {
 
     AbortCheck(context);
 
-    while (!vertices_ || vertices_it_.value() == vertices_.value().end()) {
+    while (!vertices_ || vertices_it_.value() == vertices_end_it_.value()) {
```
andrejtonev, line 441: Does this really make a difference?

src/storage/v2/all_vertices_iterable.cpp
```diff
@@ -32,7 +32,7 @@ AllVerticesIterable::Iterator::Iterator(AllVerticesIterable *self, utils::SkipLi
       it_(AdvanceToVisibleVertex(it, self->vertices_accessor_.end(), &self->vertex_, self->transaction_, self->view_,
                                  self->indices_, self_->constraints_, self->config_)) {}
 
-VertexAccessor AllVerticesIterable::Iterator::operator*() const { return *self_->vertex_; }
+VertexAccessor const &AllVerticesIterable::Iterator::operator*() const { return *self_->vertex_; }
```
andrejtonev, line 35: Is a const ref the only version we need?

andrejtonev (approved):
Cool!

src/query/plan/operator.cpp
Ignition: Yes `end()` does a bunch of work to build that iterator.
Iterators are meant to be cheap, but this one is expensive.

Josipmrden (approved):

gitbuda:
@vpavicic this probably requires a bit more attention in the changelog because it's a significant improvement (100%) in the case where there is a lot of iterations over graph entities, the query @Ignition put into the description is just one example, but I think this optimization applies in many cases 👀  @Ignition maybe you can add a bit more details from the user perspective (e.g. what's the real world query that will benefit out of this) 😃

vpavicic:
@Ignition Do you have any more queries to show the usage/benefits?

Ignition:
@vpavicic not really, the query I optimized around demonstrated ScanAll performance gain. But I expect real usecase would be more complex. Hence IMO other parts of the plan would dominate the execution profile.

PR #1228, opened 2023-09-06
Add option to test e2e directory

antoniofilipovic:
The idea of this PR is to make it easier to test the whole directory. Instead of testing one by one tests from the e2e folder, now it should be possible to test the whole folder, i.e "triggers"

[master < Task] PR


andrejtonev:
@antoniofilipovic can't we already do this?
This already works for me `python runner.py --workloads-root-directory ./triggers/`

antoniofilipovic:
@andrejtonev it works, I didn't try that. I tried with `./run.sh` and it doesn't work with that: `./run.sh ./triggers/`

antoniofilipovic:
Closing PR as this is not needed

PR #1237, opened 2023-09-07
[master <] Reduce flake SnapshotFallback test

Ignition:
Fixed the wait period, this should ensure at least one snapshot was made. Also cleaned up the checking around this. And also better corruption.

[master < Task] PR

### DIFF
--- a/tests/unit/storage_v2_durability_inmemory.cpp
+++ b/tests/unit/storage_v2_durability_inmemory.cpp
@@ -736,34 +736,49 @@ TEST_P(DurabilityTest, SnapshotPeriodic) {
 // NOLINTNEXTLINE(hicpp-special-member-functions)
 TEST_P(DurabilityTest, SnapshotFallback) {
   // Create snapshot.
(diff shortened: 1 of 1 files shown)


antoniofilipovic:
I have been thinking, this may slow down our tests. Maybe we can lower snapshot interval period, because on my machine it takes around 450ms to create BaseDataset and 60ms to create Extended one.

Ignition:
Snapshot creation time in CI logs looks like at most ~500ms.
```
[2023-09-07 17:14:50.083] [stderr] [info] Starting snapshot creation to "/tmp/MG_test_unit_storage_v2_durability/snapshots/20230907151450083169_timestamp_16546"
[2023-09-07 17:14:50.530] [stderr] [info] Snapshot creation successful!
```

I think what would be better it to disable the ASYNC snapshot in this test. Instead do explicit snapshot creation in between each phase.

antoniofilipovic (approved):
Looks good to me

Josipmrden (approved):

Issue #1222, opened 2023-09-04
Fix memory usage exceeds --memory-limit


steffiindrayani:
**Memgraph version**
Memgraph 2.8.0

**Environment**
- Using Docker
- OS: Ubuntu 22.04 64 bit
- 16vCPU(s) 32 GiB

**Describe the bug**
I have set up the --memory-limit in `etc/memgraph/memgraph.conf` to 17000 MiB and it has been reflected on the Memgraph Lab as shown below. However, whenever I run a query that exceeds the memory limit, the Memgraph does not stop the transaction automatically.
<img width="952" alt="Screenshot 2023-09-04 111644" src="https://github.com/memgraph/memgraph/assets/17632209/d3088140-d867-4022-bdf4-934899d95610">

Moreover, I have also tried to add `Query Memory Limit 10000 MB` at the end of my Cypher query. However, it still keeps exceeding the limit and does not stop automatically.

Would you mind guiding me how to correctly limit the Memgraph memory usage and any other configurations that I have missed? Really appreciate your help. Thanks in advance.

**To Reproduce**
1. Modify `memgraph.conf` file and set --memory-limit to 17000 MiB
2. Create 4M Nodes and 4M Relationships
3. Run a query with additional `Query Memory Limit 10000 MB` at the end

**Expected behavior**
Memgraph stops the query right after it reached the memory usage limit

gitbuda:
Thanks for reporting! The team has already noticed this issue. Since there are multiple reported issues with the memory tracker (both issues and PRs have `memory` tags), we have already prioritized the work on that side. Hopefully, we can improve quickly.

PR to track is https://github.com/memgraph/memgraph/pull/1064, but the whole work will probably be split into multiple PRs.
CC @antoniofilipovic please write more details as soon as possible 😃

antoniofilipovic:
@gitbuda This bug seems to be fixed with the new [PR](https://github.com/memgraph/memgraph/pull/1250). Tested with memory limit set on 2GiB and tried query which would overflow the limit, it stopped as soon as it reached the limit. 

For now, this will stay open for @steffiindrayani to test once this PR lands on the master.

gitbuda:
#1250 got merged -> this issue should be resolved / experience massively improved.

gitbuda:
TTL on this issue ~2 weeks (as soon as we release v2.12 it should be verified)

as51340:
Closing because it is verified, works with 2.12.

PR #1250, opened 2023-09-14
Implement jemalloc extent hooks memory tracker

antoniofilipovic:
**Idea:**
This PR wants to introduce a memory tracker on top of `jemalloc` instead of using a memory tracker around new and delete. One problem with using a memory tracker on new and delete is that they are not tracking memory allocations happening in the background of jemalloc.

Currently, we are confident that the `jemalloc` tracker is tracking every memory allocation. Test included setting bytes on every page `jemalloc` returned. In that case, RES size was 1:1 with the jemalloc tracker.

 
Problems with current `v4.sh`:
- percpu arena is not obeyed
- oversize threshold is not obeyed
- memory retain is making impossible to track allocations due to lot of virtual memory allocations

New setup script for `jemalloc`:
```
./autogen.sh
MALLOC_CONF="retain:false,percpu_arena:percpu,oversize_threshold:0,muzzy_decay_ms:5000,dirty_decay_ms:5000" \
./configure \
  --disable-cxx \
  --enable-shared=no --prefix=$working_dir \
  --with-malloc-conf="retain:false,percpu_arena:percpu,oversize_threshold:0,muzzy_decay_ms:5000,dirty_decay_ms:5000"

```

PR steps to make it work fully:
- [x] add our hooks as a wrapper around `jemalloc` default hooks
- [x] fix issue inside jemalloc with percpu arena not being obeyed
- [x] disable jemalloc oversize_threashold which is used for custom arena on which is not possible to set extant hooks
- [x] fix the issue with the commit hook not being called, and alloc hook only getting calls for virtual memory 
**Note**: this is not fixed in the end, this is jemalloc internal behavior, but commit hook is not called
- [x] add counters to track `alloc`, `dalloc`, `destroy`, `commit`, `decommit`, `purge_forced`, `purge_lazy`, `split`, `merge`
- [x] add check if allocations are multiple of page size 
- [x] fix issue on jemalloc memory tracker increasing memory more than expected:
     explanation: jemalloc calls alloc hook more than necessary, this can be fixed but at a cost. If tcache is disabled, then it means that everything will be backed by allocation and in that case resident memory (RES) will increase a lot on normal workflows
- [x] check if when committing we should reduce size from virtual_allocated, increase on allocated
- [x] check if alignment needs to be included in memory allocation
- [x] swap current memory tracker with `jemalloc` extent hook memory tracker
- [x] ULTIMATE TEST: add bigger size for deltas initial allocation (1MB) and check if benchmark will pass: it did :heavy_check_mark: The thing is now we are tracking memory correctly, and if you allocate that much, there is bug in code
- [x] Check if transparent huge pages in jemalloc have any influence
- [x] check if caching when has any influence
- [x] update libs to have new jemalloc version
- [x] add custom tests to check memory allocation without using the RSS memory tracker behind - the problem is that inside RSS we can have other stuff, like loaded modules, etc

TODO(@antoniofilipovic): Don't forget to check if we need to update `vm.max_map_count`

[master < Task] PR


antoniofilipovic:
There is an additional config that we should check if we want to change:
from `muzzy_decay_ms:5000,dirty_decay_ms:5000` to `muzzy_decay_ms:0,dirty_decay_ms:0`

This will force `jemalloc` to purge memory pages once they are ready to be unused. Otherwise, jemalloc purges pages lazily, once there is a need for them in the system.

src/memory/memory_control.cpp
```diff
@@ -10,18 +10,297 @@
...
+static constexpr extent_hooks_t custom_hooks = {
+    .alloc = &my_alloc,
+    .dalloc = &my_dalloc,
+    .destroy = &my_destroy,
+    .commit = &my_commit,
+    .decommit = &my_decommit,
+    .purge_lazy = &my_purge_lazy,
+    .purge_forced = &my_purge_forced,
+    .split = &my_split,
+    .merge = &my_merge,
```
Ignition, line 202: Why not just use old_hooks directly...why add runtime assert + indirection

src/memory/new_delete.cpp
```diff
@@ -41,9 +41,7 @@ void *newImpl(const std::size_t size, const std::align_val_t align) {
 }
 
 void *newNoExcept(const std::size_t size) noexcept { return malloc(size); }
-void *newNoExcept(const std::size_t size, const std::align_val_t align) noexcept {
-  return aligned_alloc(size, static_cast<std::size_t>(align));
-}
+void *newNoExcept(const std::size_t size, const std::align_val_t align) noexcept { return malloc(size); }
```
Ignition, line 44: DANGER: why `aligned_alloc` -> `malloc`

tests/e2e/memgraph.py
```diff
@@ -130,9 +130,14 @@ def is_running(self):
     def stop(self):
         if not self.is_running():
             return
-        self.proc_mg.terminate()
-        code = self.proc_mg.wait()
-        assert code == 0, "The Memgraph process exited with non-zero!"
+
+        pid = self.proc_mg.pid
+        try:
+            os.kill(pid, 15)  # 15 is the signal number for SIGTERM
```
Ignition, line 136: Why this change?
Popen.terminate() - Stop the child. On POSIX OSs the method sends SIGTERM to the child.

src/memory/new_delete.cpp
antoniofilipovic: yes yes, reverted that, it was just some mine testing

src/memory/memory_control.cpp
antoniofilipovic: Hmm what do you mean?

antoniofilipovic: I need to use different hooks, because if I use old hooks then jemalloc won't call my hooks and I can't count memory.

tests/e2e/memgraph.py
antoniofilipovic: Yes, but there was some issue with this method. It reported that the memgraph didn't properly close, although from logs and by checking manually everything was fine. I checked for different ways to close the memgraph, and this one works. I couldn't see any issue, so I presumed the method was the problem.

src/memory/memory_control.cpp
Ignition: But these hooks don't do any counting, they assert + pass to old hooks

antoniofilipovic: Hmm, they do. If you check alloc, purge_forced, dalloc, there are calls to memory tracker?

antoniofilipovic: Ah, you are thinking on only few of them, like merge and split, not purge_forced also. Yes, that makes sense

src/query/interpreter.cpp
```diff
@@ -3068,6 +3067,8 @@ PreparedQuery PrepareSystemInfoQuery(ParsedQuery parsed_query, bool in_explicit_
             {TypedValue("average_degree"), TypedValue(info.average_degree)},
             {TypedValue("memory_usage"), TypedValue(static_cast<int64_t>(info.memory_usage))},
             {TypedValue("disk_usage"), TypedValue(static_cast<int64_t>(info.disk_usage))},
+            // {TypedValue("jemalloc_memory_allocated"),
+            //  TypedValue(utils::GetReadableSize(static_cast<double>(utils::total_memory_tracker.Amount())))},
```
antoniofilipovic, line 3071: Will move this to other PR or maybe leave it here, as it seems useful for end user, what are opinions on this one?

gitbuda:
STATUS: Ready to be merged

Ignition (approved):

src/memory/memory_control.cpp
```diff
@@ -22,6 +24,181 @@ namespace memgraph::memory {
...
+
+  sz = sizeof(unsigned);
+  unsigned n_arenas{0};
+  int err = mallctl("opt.narenas", (void *)&n_arenas, &sz, nullptr, 0);
+
+  if (err) {
+    return;
+  }
+
+  spdlog::trace("n areanas {}", n_arenas);
```
Josipmrden, line 151: I think we don't need this in the logs

Josipmrden (commented):
minor comments, can merge without re-assigning me to review

antoniofilipovic:
@vpavicic RN: Fix problems with memory limit not being honored.

antoniofilipovic:
PR fixes following issue: https://github.com/memgraph/memgraph/issues/1222

Josipmrden (approved):

antoniofilipovic:
@vpavicic I need to add docs page for VM max map count, similar to this one: https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html#_set_vm_max_map_count_to_at_least_262144

PR #1252, opened 2023-09-14
Fix bug in alias mapping

gvolfing:
If the original and the mapped value differ in the actual function call name, then the logic behind looking for possible mappings is not working as intended. This fixes the issue.


antoniofilipovic (approved):
Looks good to me

antoniofilipovic (approved):

PR #1265, opened 2023-09-18
Add filtering details to EXPLAIN and PROFILE

antepusic:
This PR extends the `EXPLAIN` and `PROFILE` clauses with detailed information about the `Filter` operator.

[master < Task] PR
> Add detailed Filter operator info to EXPLAIN and PROFILE clauses

> We’ve improved query inspection using `EXPLAIN` and `PROFILE` clauses by adding information about the operands of the Filter operator.


src/query/plan/pretty_print.cpp
```diff
@@ -180,7 +180,7 @@ bool PlanPrinter::PreVisit(query::plan::Foreach &op) {
 }
 
 bool PlanPrinter::PreVisit(query::plan::Filter &op) {
-  WithPrintLn([](auto &out) { out << "* Filter"; });
+  WithPrintLn([&](auto &out) { out << "* " << op.ToString(); });
```
as51340, line 183:
```suggestion
  WithPrintLn([&op](auto &out) { out << "* " << op.ToString(); });
```
taking all as a reference is costly

src/query/plan/preprocess.hpp
```diff
@@ -287,6 +286,8 @@ struct FilterInfo {
 /// expressions that should be generated.
 class Filters final {
  public:
+  std::vector<FilterInfo> all_filters_;
```
as51340, line 289: This is not such a good practice... It is better than to create a getter method than to put field as a private

src/query/plan/preprocess.hpp
```diff
@@ -213,7 +212,7 @@ class PatternFilterVisitor : public ExpressionVisitor<void> {
 /// Stores the symbols and expression used to filter a property.
 class PropertyFilter {
  public:
-  using Bound = ScanAllByLabelPropertyRange::Bound;
+  using Bound = utils::Bound<Expression *>;
```
as51340, line 215: Could you explain me why is this needed?

src/query/plan/operator.hpp
```diff
@@ -1120,6 +1124,45 @@ class Filter : public memgraph::query::plan::LogicalOperator {
...
+  static std::string SingleFilterName(const query::plan::FilterInfo &single_filter) {
+    using Type = query::plan::FilterInfo::Type;
+    if (single_filter.type == Type::Generic) {
+      return fmt::format("Generic {{{}}}", utils::IterableToString(single_filter.used_symbols, ", ",
+                                                                   [](const auto &symbol) { return symbol.name(); }));
+    } else if (single_filter.type == Type::Id) {
+      return fmt::format("id({})", single_filter.id_filter->symbol_.name());
+    } else if (single_filter.type == Type::Label) {
+      if (single_filter.expression->GetTypeInfo() != LabelsTest::kType) {
+        LOG_FATAL("This type of label filter syntax is not supported for query inspection!");
```
as51340, line 1138: Which type? I would suggest a bit better message

src/query/plan/operator.cpp
```diff
@@ -2450,17 +2458,18 @@ Filter::FilterCursor::FilterCursor(const Filter &self, utils::MemoryResource *me
...
   // nodes and edges.
   ExpressionEvaluator evaluator(&frame, context.symbol_table, context.evaluation_context, context.db_accessor,
                                 storage::View::OLD, context.frame_change_collector);
-  while (input_cursor_->Pull(frame, context)) {
+  while (input_cursor_->Pull(frame, context) || UNLIKELY(context.is_profile_query)) {
     for (const auto &pattern_filter_cursor : pattern_filter_cursors_) {
       pattern_filter_cursor->Pull(frame, context);
     }
     if (EvaluateFilter(evaluator, self_.expression_)) return true;
+    if (UNLIKELY(context.is_profile_query)) return false;
```
as51340, line 2472: This logic is a bit obscure to me. You are in the while loop if context.is_profile_query is true but then at the end you return false?

src/query/plan/operator.hpp
```diff
@@ -1120,6 +1124,45 @@ class Filter : public memgraph::query::plan::LogicalOperator {
   std::shared_ptr<memgraph::query::plan::LogicalOperator> input_;
   std::vector<std::shared_ptr<memgraph::query::plan::LogicalOperator>> pattern_filters_;
   Expression *expression_;
+  const memgraph::query::plan::Filters all_filters_;
+
+  static std::string SingleFilterName(const query::plan::FilterInfo &single_filter) {
```
as51340, line 1129: This is good but I think even better would be that this is a method of FilterInfo struct to avoid static methods, seems like a bit better design

as51340 (changes requested):

as51340:
Good job Ante, I have few ideas how to improve it and good to go

src/query/plan/pretty_print.cpp
antepusic: Done

src/query/plan/preprocess.hpp
antepusic: Operator.hpp must include this header because it contains the definition of `Filter::all_filters_`. To prevent a circular dependency, I redefine `Bound` here without relying on operator.hpp’s `ScanAllByLabelPropertyRange`. `Bound` itself is unchanged (compare its definition in `ScanAllByLabelPropertyRange`).

src/query/plan/operator.cpp
antepusic: The `for` loop executes in two cases:
* `input_cursor_->Pull` returns `true`
* in profile queries (even if there’s no input, the query plan should contain the branch(es) that do filtering)
This line force-exits the outer `while` loop for profile queries (in their case, the `while` condition is always true).

src/query/plan/preprocess.hpp
antepusic: I needed it because of [this line](https://github.com/memgraph/memgraph/blob/dfef16b8bfd621ef363cd81d4ffe584a7c727732/tests/unit/query_plan_operator_to_string.cpp#L235-L236). Access to `all_filters_` is already possible via the public iterators, will add a setter for it.

src/query/plan/operator.hpp
antepusic: Fixed (NB: for label filters, an `expression` of different type than `LabelsTest` is not forbidden, but it never appears)

src/query/plan/operator.hpp
antepusic: That makes sense! I've considered it, but decided against doing so for the following reasons:
* `SingleFilterName` is used only by `Filter::ToString`
* `ExpandVariable` has an analogous (but non-static) `OperatorName` method

src/query/plan/operator.hpp
```diff
@@ -1120,6 +1124,45 @@ class Filter : public memgraph::query::plan::LogicalOperator {
...
+    using Type = query::plan::FilterInfo::Type;
+    if (single_filter.type == Type::Generic) {
+      return fmt::format("Generic {{{}}}", utils::IterableToString(single_filter.used_symbols, ", ",
+                                                                   [](const auto &symbol) { return symbol.name(); }));
+    } else if (single_filter.type == Type::Id) {
+      return fmt::format("id({})", single_filter.id_filter->symbol_.name());
+    } else if (single_filter.type == Type::Label) {
+      if (single_filter.expression->GetTypeInfo() != LabelsTest::kType) {
+        LOG_FATAL("Label filters not using LabelsTest are not supported for query inspection!");
+      }
```
antepusic, line 1139: NB: for label filters, an expression of different type than `LabelsTest` is not forbidden, but it never appears

src/query/plan/operator.cpp
as51340: Okay but if the context.is_profile_query is true what do you need to execute, for loop of filter evaluation? Because it is a bit not clear that you want to got to while loop just to execute it once. IMO it is better to extract the part you need, not sure if it is for loop or evaluate filter, something like this:

```
if (context.is_profile_query) {
  for loop and/or
  evaluate filter
} else {
  while (input_cursor->Pull...) ...

```

as51340 (approved):
Adopt the change in this comment or don't it is up to you, LGTM

src/query/plan/preprocess.hpp
```diff
@@ -302,6 +301,8 @@ class Filters final {
   auto erase(iterator first, iterator last) { return all_filters_.erase(first, last); }
   auto erase(const_iterator first, const_iterator last) { return all_filters_.erase(first, last); }
 
+  void SetFilters(std::vector<FilterInfo> all_filters) { all_filters_ = all_filters; }
```
antoniofilipovic, line 304: Why not pass `all_filters` by `const &`, or move them here?

antoniofilipovic: I think this can accept rvalue, and then you move to all_filters_:
```
void SetFilters(std::vector<FilterInfo> &&all_filters) { all_filters_ = std::move(all_filters); }
```

antoniofilipovic (changes requested):
Just one comment, good job!

src/query/plan/preprocess.hpp
antepusic: Added

antepusic:

antoniofilipovic (approved):

PR #1266, opened 2023-09-18
[master < T1266] Improve deserialization performance

as51340:
## Preparation query:
```cypher
UNWIND range(1, 100000) as u CREATE (n:Node {string: "test"}) RETURN n;
```
## Before optimization:
```cypher
MATCH (n) RETURN n;
```
**Ran 5 times, avg. is: 0.5718**

## After:
```cypher
MATCH (n) RETURN n;
```

**Ran 5 times, avg. is: 0.3094**

### Conclusion: Roughly this PR gives 2x faster data deserialization from disk. Loading all edges isn't improved because the cost comes from different place.

[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR


src/storage/v2/id_types.hpp
```diff
@@ -31,6 +32,11 @@ namespace memgraph::storage {
     static name FromInt(int64_t id) { return name{utils::MemcpyCast<uint64_t>(id)}; }                         \
     uint64_t AsUint() const { return id_; }                                                                   \
     int64_t AsInt() const { return utils::MemcpyCast<int64_t>(id_); }                                         \
+    static name FromString(std::string_view id) {                                                             \
+      uint64_t value;                                                                                         \
+      std::from_chars(id.data(), id.data() + id.size(), value);                                               \
```
Darych, line 37: What do you think about checking return values of `from_chars`?
I wrote checks in https://github.com/memgraph/memgraph/blob/master/src/utils/rocksdb_serialization.hpp#L214
I think there we also need to change to `FromString()` function call

src/storage/v2/mvcc.hpp
```diff
@@ -131,13 +131,17 @@ inline Delta *CreateDeleteDeserializedObjectDelta(Transaction *transaction, std:
                                                   std::string &&ts) {
   transaction->EnsureCommitTimestampExists();
   // Should use utils::DecodeFixed64(ts.c_str()) once we will move to RocksDB real timestamps
-  return &transaction->deltas.use().emplace_back(Delta::DeleteDeserializedObjectTag(), std::stoull(ts), old_disk_key);
+  uint64_t ts_id = 0;
+  std::from_chars(ts.data(), ts.data() + ts.size(), ts_id);
```
Darych, line 135: Maybe create utils function which will convert string_view to uint64_t?
We will be able to use it here and in LabelId also.

src/storage/v2/id_types.hpp
as51340: The reason I went without check is the performance. Check will obviously slow down things and I think that if something bad happens, std::from_chars will throw exception automatically.

src/storage/v2/mvcc.hpp
as51340: I can add this

src/storage/v2/disk/name_id_mapper.hpp
```diff
@@ -44,7 +44,8 @@ class DiskNameIdMapper final : public NameIdMapper {
     }
     uint64_t res_id = 0;
     if (auto maybe_id_from_disk = name_to_id_storage_->Get(std::string(name)); maybe_id_from_disk.has_value()) {
-      std::from_chars(maybe_id_from_disk->data(), maybe_id_from_disk->data() + maybe_id_from_disk->size(), res_id);
+      auto id_disk_value = maybe_id_from_disk.value();
+      std::from_chars(id_disk_value.data(), id_disk_value.data() + id_disk_value.size(), res_id);
```
Darych, line 48: Can we call here `utils::ParseStringToUint64`?

Darych (approved):
LGTM

as51340:
Release note: The reading performance of vertices has been improved 2x.

PR #1278, opened 2023-09-19
Fix SHOW CONFIG doesn't show the run-time flag status

### LINKED ISSUE #1257: [BUG] Show config with run-time settings
**Memgraph version**
2.11.0

**Environment**
Running Memgraph locally on Ubuntu 20.04.

**Describe the bug**
When executing `SHOW CONFIG;` the run-time configurable flags are not updated.

**To Reproduce**
1. Run `SHOW CONFIG` and check variable "log_level"
2. Run `SET DATABASE SETTING "log.level" TO "DEBUG"`
3. Rerun `SHOW CONFIG`

**Expected behavior**
"log_level" should say "debug".

**Additional context**
This bug is due to the way `SHOW CONFIG` generates the results. It simply reads the configuration gflags. The flags themselves are being used when reading run-time configurations anymore. Separate atomic values are read (or simple callbacks executed) instead of reading the flags.

andrejtonev:
[master < Task] PR

closes #1257


src/flags/run_time_configurable.cpp
```diff
@@ -9,103 +9,170 @@
 // by the Apache License, Version 2.0, included in the file
 // licenses/APL.txt.
 
-#include "flags/run_time_configurable.hpp"
```
Ignition, line 12: Good practice is for the first include to be a cpp's corresponding hpp

src/flags/bolt.hpp
```diff
@@ -26,4 +26,4 @@ DECLARE_string(bolt_cert_file);
 // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
 DECLARE_string(bolt_key_file);
 // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
-DECLARE_string(bolt_server_name_for_init);
+// DECLARE_string(bolt_server_name_for_init); Moved to run_time_configurable
```
Ignition, line 29: Why move any of these? they were already in files related to purpose. Does the runtime configurable mechanism require them to be elsewhere?

src/flags/run_time_configurable.cpp
```diff
@@ -9,103 +9,170 @@
...
-    const auto log_level = memgraph::utils::global_settings.GetValue(kLogLevelSettingKey);
-    MG_ASSERT(log_level, "Log level is missing from the settings");
-    const auto ll_enum = memgraph::flags::LogLevelToEnum(*log_level);
-    if (!ll_enum) {
-      throw utils::BasicException("Unsupported log level {}", *log_level);
+void Initialize() {
+  constexpr bool kRestore = true;
+  auto register_flag = [&](
+                           const std::string &flag, const std::string &key, bool restore,
+                           std::function<void(const std::string &)> post_update = [](auto) {},
```
Ignition, line 100: For avoiding allocations prefer `auto &&` to take the lambda, rather than `std::function`

```cpp
#include <concepts>

auto demo(std::invocable<int> auto && f)
{ 
    return f(42);
}

int main()
{
    return demo([](int x){ return x; });
}
```

src/flags/bolt.hpp
andrejtonev: I didn't want people using them directly, so I hid them in the run time cpp.

src/flags/run_time_configurable.cpp
andrejtonev: Unfortunately I am not sure you can have a default post_update this way.
Am I wrong?

Ignition: You are right you would need to split the overload set by introducing a new overload 
```cpp
#include <concepts>

auto demo(std::invocable<int> auto && f)
{ 
    return f(2);
}

inline auto demo()
{
    return demo(/*default =*/[](int){ return 40;});
}

int main()
{
    return demo() + demo([](int x){ return x; });
}
```

andrejtonev: I would leave it as is. 
This is an initialization of run time settings and I would rather have a single function implementation than better performance.

src/flags/run_time_configurable.cpp
```diff
@@ -10,102 +10,170 @@
...
+                           std::function<bool(std::string_view)> validator = [](std::string_view) { return true; }) {
+    // Get flag info
+    gflags::CommandLineFlagInfo info;
+    gflags::GetCommandLineFlagInfo(flag.c_str(), &info);
+    // Register setting
+    auto update = GenHandler(flag, key);
+    memgraph::utils::global_settings.RegisterSetting(
+        key, info.default_value,
+        [update, post_update] {
```
antoniofilipovic, line 110: I think you can capture by ref here
```suggestion
        [&update, &post_update] {
```

src/flags/run_time_configurable.cpp
```diff
@@ -10,102 +10,170 @@
...
+      // Override value if passed via command line argument
+      if (!info.is_default) {
+        memgraph::utils::global_settings.SetValue(key, info.current_value);
+      } else {
+        // Force read from settings (restore previously saved value)
+        update();
+      }
+    } else {
+      // No restore, just force the current value
+      memgraph::utils::global_settings.SetValue(key, info.current_value);
```
antoniofilipovic, line 125:
```suggestion
    if (!restore) {
      // No restore, just force the current value
      memgraph::utils::global_settings.SetValue(key, info.current_value);
      return;
    }
    // Override value if passed via command line argument
    if (!info.is_default) {
      memgraph::utils::global_settings.SetValue(key, info.current_value);
      return;

    // Force read from settings (restore previously saved value)
    update();


    
```

antoniofilipovic (approved):
Few suggestions, but as it looks good to me, I will approve and sorry for taking longer

andrejtonev:
@vpavicic SHOW CONFIG was not reporting all configuration flags, specifically the run-time enabled ones. This PR fixes that.

src/flags/bolt.hpp
andrejtonev: @Ignition ping

PR #1281, opened 2023-09-20
[master <] Refactor interpreter to support multiple distributed clocks

Ignition:
Aim of this and future PRs:
For replication of system information + database changes, we need to support vector clocks (term from distributed computing).

How:
- Interpreter transaction ID will be decoupled from the storage's transaction ID [THIS PR]
- This will be the user friendly ID (what you can refer to from cypher) [THIS PR]
- The interpreter will track
  - current DB transaction (eventually multiple for multi database queries)
  - system clock, for things which need sequencing and replicating
- Commit of interpreter will do ensure all subsystem commit (two phase commit)
  - Phase 1: Durability as intended (may fail)
  - Phase 2: Commit, finalize durability (must not fail), then start replication
- Storage upgrades needed for replication
  - Index, Constraint, Analyze (stats) - observable performance, hence should be replicated
- System upgrades needed for replication
  - Needs to add semantics for Abort...maybe 
  - ATM does not exist....make it exist

[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR


src/query/frontend/semantic/uses_database.cpp
```diff
@@ -0,0 +1,176 @@
...
+class BLA : public QueryVisitor<void>, public HierarchicalTreeVisitor {
+ public:
+  using HierarchicalTreeVisitor::PostVisit;
+  using HierarchicalTreeVisitor::PreVisit;
+  using HierarchicalTreeVisitor::Visit;
+  using QueryVisitor<void>::Visit;
+
+  bool state() { return state_; }
+
+  bool PreVisit(Create & /*unused*/) override {
```
as51340, line 28: Can you expkain this usage_kind a bit more?

src/query/interpreter.hpp
```diff
@@ -139,6 +139,37 @@ struct QueryExtras {
...
+    in_explicit_db_ = in_explicit_db;
+  }
+
+  // TODO: don't provide explicitly via constructor, instead have a lazy way of getting the current/default
+  // DatabaseAccess
+  //       hence, explict bolt "use DB" in metadata wouldn't necessarily get access unless query required it.
+  std::optional<memgraph::dbms::DatabaseAccess> db_acc_;  // Current db (TODO: expand to support multiple)
+  // This cannot be std::optional because we need to move this accessor later on into a lambda capture
+  // which is assigned to std::function. std::function requires every object to be copyable, so we
+  // move this unique_ptr into a shrared_ptr.
```
as51340, line 166: shared

src/query/interpreter.hpp
```diff
@@ -139,6 +139,37 @@ struct QueryExtras {
...
+  void SetCurrentDB(memgraph::dbms::DatabaseAccess new_db, bool in_explicit_db) {
+    // do we lock here?
+    db_acc_ = std::move(new_db);
+    in_explicit_db_ = in_explicit_db;
+  }
+
+  // TODO: don't provide explicitly via constructor, instead have a lazy way of getting the current/default
+  // DatabaseAccess
+  //       hence, explict bolt "use DB" in metadata wouldn't necessarily get access unless query required it.
+  std::optional<memgraph::dbms::DatabaseAccess> db_acc_;  // Current db (TODO: expand to support multiple)
```
as51340, line 163: Can you explain a bit for what we use Gatekeeper ?

src/storage/v2/disk/storage.cpp
```diff
@@ -1545,18 +1545,92 @@ DiskStorage::CheckExistingVerticesBeforeCreatingUniqueConstraint(LabelId label,
...
+          }
+        } break;
+        case MetadataDelta::Action::LABEL_PROPERTY_INDEX_DROP: {
+          const auto &info = md_delta.label_property;
+          if (!disk_storage->PersistLabelPropertyIndexAndExistenceConstraintDeletion(info.label, info.property,
+                                                                                     label_property_index_str)) {
+            return StorageManipulationError{PersistenceError{}};
+          }
+        } break;
+        case MetadataDelta::Action::LABEL_INDEX_STATS_SET: {
```
as51340, line 1589: ha smart, save metadata deltas and process them at the end,

src/distributed/include/distributed/lamport_clock.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+ private:
+  friend struct LamportClock<Tag>;
+
+  timestamp(uint64_t value) : value_{value} {}
+  uint64_t value_;
+};
+
+constexpr struct internal_t {
+} internal;
```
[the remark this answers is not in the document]
Ignition: These are a tag types, used for tag dispatch to correct `get_timestamp` method.
internal, send, and receive are to be deliberate and consistent with the literature. 
Intra process events that require a timestamp use internal.
Inter process events that require a timestamp use send/receive.

src/query/frontend/semantic/uses_database.cpp
Ignition: I think this needs cleaning up.
Not sure if we ended up depending on it or not.
Its a visitor which would say if what kind of query we have and hence what the interpreter should do.
ie. does it need DB access or not + does it need DB transactional access or not + will that DB transactional access perform writes (aka expected to do work on commit)

I don't think we got around to completing this/depending on it.

src/query/interpreter.hpp
Ignition: I think this comment is old and not sure if still relevant, what do you think @andrejtonev?

as51340 (approved):

src/query/interpreter.hpp
Ignition: TODO: we will need to make resources (present a talk to core devs) for all this + once close to final state (aka less cleanup) we will document within the code + add diagrams/UML

src/query/interpreter.hpp
andrejtonev: Yeah, just some old comments we made during the refactoring.

andrejtonev:
@vpavicic 
Changes that impact the user are:
- Introduced ResourceLock to fix pthread mutex locking/unlocking problems and updated Storage to use it
    - This is a bugfix
- Replicable and durable statistics
    - We are now saving the statistics we get from ANALYZE GRAPH and are replicating and recovering on startup
- Bumped WAL/Snapshot versions
    - New deltas introduced, so WAL/Snapshot versions had to be updated. Shouldn't effect users with older versions, since we still support those.

Ignition:
related to #1359

PR #1286, opened 2023-09-22
Add mgbench support for disk storage and analytical mode

as51340:
The goal of this PR is to add support for running benchmarks on disk storage and analytical storage mode in a same way as benchmarks are being run for in_memory_transactional. This PR doesn't modify CI/CD pipeline but enables us to run mgbench locally.

[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR


.github/workflows/diff.yaml
```diff
@@ -417,15 +417,17 @@ jobs:
...
                     --github-run-number "${{ github.run_number }}" \
                     --head-branch-name "${{ env.BRANCH_NAME }}"
 
+      # TODO (andi) No need for path flags and for --disk-storage and --in-memory-analytical
       - name: Run mgbench
         run: |
           cd tests/mgbench
-          ./benchmark.py vendor-native --num-workers-for-benchmark 12 --export-results benchmark_result.json pokec/medium/*/*
+          ./benchmark.py vendor-native --num-workers-for-benchmark 12 --export-results benchmark_result.json --export-results-in-memory-analytical results_in_memory_analytical.json \
+          --export-results-on-disk-txn results_on_disk_txn.json pokec*/medium/*/*
```
antejavor, line 430: What is the * before pokec?

tests/mgbench/benchmark.py
```diff
@@ -161,7 +119,19 @@ def parse_args():
...
         "--export-results",
         default=None,
-        help="file path into which results should be exported",
+        help="file path into which results for in_memory_transactional storage mode should be exported",
+    )
+
+    benchmark_parser.add_argument(
+        "--export-results-in-memory-analytical",
+        default=None,
+        help="file path into which results for in_memory_analytical storage mode should be exported",
```
antejavor, line 128: I can see that the definition of --export-results... defines if something is being run. Can you please update the help messages to describe this?

tests/mgbench/benchmark.py
```diff
@@ -136,7 +94,7 @@ def parse_args():
     benchmark_parser.add_argument(
         "--single-threaded-runtime-sec",
         type=int,
-        default=10,
+        default=1,
```
antejavor, line 97: You only changed the runtime in seconds but didn't change the `--query-count-lower-bound` it was 30 before?

tests/mgbench/workloads/disk_pokec.py
```diff
@@ -0,0 +1,449 @@
...
+        return ("MATCH (n:User {id: $id}) RETURN n", {"id": self._get_random_vertex()})
+
+    # OK
+    def benchmark__update__vertex_on_property(self):
+        return (
+            "MATCH (n {id: $id}) SET n.property = -1",
+            {"id": self._get_random_vertex()},
+        )
+
+    # Basic benchmark queries
```
antejavor, line 302: You can delete all the basic queries (`benchmark__basic`), they were c/p for the benchgraph, have the same group and have _type in the name. There were two reasons:

- not to change the json structure 
- and to have a query type for benchgraph. 

The first batch of queries contained everything that is on Grafana.

antejavor (changes requested):
Left some comments.

antejavor:
Just to have a note on this, README file should also be updated at some point: https://github.com/memgraph/memgraph/blob/master/tests/mgbench/README.md

.github/workflows/diff.yaml
as51340: I added pokec_disk dataset. Therefore if you set pokec* it will run both in_memory and disk dataset

tests/mgbench/benchmark.py
as51340: Yes and I don't know if you checked grafana for this PR but after running each query for 1s, almost all results changed for in memory transactional. So probably 1s is not enough. But I am trying to find some good point for which mgbench won't be run for to long while not losing precision. Any advices?

tests/mgbench/workloads/disk_pokec.py
as51340: basic queries are c/p of what? Can you explain a bit more what you meant, not sure I understood?

as51340:
> Just to have a note on this, README file should also be updated at some point: https://github.com/memgraph/memgraph/blob/master/tests/mgbench/README.md

What do I need to update can you speed me up here a bit?

.github/workflows/diff.yaml
antejavor: Aha, nice! 🦾

tests/mgbench/benchmark.py
antejavor: I have seen, but I have noticed that @gitbuda and you are having a discussion regarding that. 

Import is 10-12 minutes per workload = 30 minutes if you run all workloads.  We can export snapshot datasets and inject them via custom import from S3. 

Expansion 4 took a lot of time (3-4 min I guess) and provides low value since this is a social network. You are returning almost the whole graph, due to the density of the graph. But on the other hand, it is the only query that stresses memgraph memory usage and performance. 

Those are two options that come to my mind.

tests/mgbench/workloads/disk_pokec.py
antejavor: Basic queries are c/p from the pokec.py (arango, create, aggregate). They were named `basic`, to drop Arango group and added the `_type` extension, we needed this for benchgraph. We did that to avoid changing the existing queries JSON paths (as I mentioned) . If you take a look, you will notice that basic queries are identical to (arango + create + aggregate). Only the shortest path is missing due to not being able to run on Neo4j. 

So feel free to delete them.

tests/mgbench/benchmark.py
antejavor: One more option: maybe we could define a smaller group of queries that we run on the diff, for example, `pokec*/medium/diff`, and then have a chron job running the `pokec*/medium/daily` that would involve a much bigger set of queries since they should only grow in size.

antejavor:
> What do I need to update can you speed me up here a bit?

Can you add a note here: https://github.com/memgraph/memgraph/blob/master/tests/mgbench/README.md#changelog-benchgraph-public-benchmark
 Write that improvements have been made for the sake of Memgraph internal performance testing, put the title like internal upgrades, and link two PRs mentioning what has changed. 
 
 I will retest rest of the things from benchgraph perspective and fix if necessary.

.github/workflows/diff.yaml
```diff
@@ -417,15 +417,17 @@ jobs:
...
                     --github-run-number "${{ github.run_number }}" \
                     --head-branch-name "${{ env.BRANCH_NAME }}"
 
+      # TODO (andi) No need for path flags and for --disk-storage and --in-memory-analytical
       - name: Run mgbench
         run: |
           cd tests/mgbench
-          ./benchmark.py vendor-native --num-workers-for-benchmark 12 --export-results benchmark_result.json pokec/medium/*/*
+          ./benchmark.py vendor-native --num-workers-for-benchmark 12 --export-results benchmark_result.json \
+          --export-results-on-disk-txn results_on_disk_txn.json pokec*/medium/*/*
```
antejavor, line 430: @as51340 If we are not going to run it on CI/CD the `--export-results-on-disk-txn` is not needed? 
```suggestion
          pokec*/medium/*/*
```

antejavor (approved):
@as51340 this is good from my side. Just check the last comment.

PR #1315, opened 2023-09-29
Enhance on-disk index usage

gvolfing:


gitbuda:
Closed because all is done under #1332

PR #1318, opened 2023-09-29
Fix a crash caused when declaring a path with only one node in OPTIONAL MATCH clause

### LINKED ISSUE #1293: Crash encountered about a query
I executed a query on my empty memgraph database and get an error.

memgraph version:2.11.0
operating system: windows 10
API：Cypher version 9
query:MATCH (n1) OPTIONAL MATCH p=(n1) RETURN 1;
error message:Query failed: Connection was closed by server

Actual behavior:
The query failed and the memgraph container exited abnormally with the error message.

DavIvek:
A query that defines path as just a single node that is already declared in the previous MATCH or OPTIONAL MATCH clause, for example, MATCH (n1) OPTIONAL MATCH p=(n1) RETURN 1; will fail and will crash memgraph.

When the user defines a path, memgraph will generate a ConstructNamedPath operator. Memgraph uses bound_symbols to save symbols for which it needs to generate a ScanAll operator. If the path has a relationship, memgraph will generate an Expand logical operator.
A case in which the path consists of only a single node and that node is already in bound_symbols isn't covered, so memgraph will fail in the planning phase.

[master < Task] PR

closes #1293


DavIvek:
With these changes, memgraph will cover this case by generating Filter and ConstructNamedPath operators without ScanAll operator, since ScanAll is already applied because it's in bound_symbols.

Ignition (changes requested):
Can you add a testcase.

Plus, there is a whole class of problem that would be great to fix while you are in this code. For example:

EXPLAIN MATCH (n2) OPTIONAL MATCH p=(n1)-[]->(n2)-[]->(n3) RETURN p;
+----------------------------------+
| QUERY PLAN                       |
+----------------------------------+
| " * Produce {p}"                 |
| " * Optional"                    |
| " |\\ "                          |
| " | * ConstructNamedPath"        |
| " | * EdgeUniquenessFilter"      |
| " | * Expand (n2)-[anon3]->(n3)" |
| " | * Expand (n1)-[anon2]->(n2)" |
| " | * ScanAll (n1)"              |
| " | * Once"                      |
| " * ScanAll (n2)"                |
| " * Once"                        |
+----------------------------------+

Would be better as 

+----------------------------------+
| QUERY PLAN                       |
+----------------------------------+
| " * Produce {p}"                 |
| " * Optional"                    |
| " |\\ "                          |
| " | * ConstructNamedPath"        |
| " | * EdgeUniquenessFilter"      |
| " | * Expand (n2)-[anon3]->(n3)" |
| " | * Expand (n2)<-[anon2]-(n1)" |
| " | * Once"                      |
| " * ScanAll (n2)"                |
| " * Once"                        |
+----------------------------------+

DavIvek:
Added a testcase to cover this case and opened an issue for possible optimization with a planner.

Ignition (approved):

PR #1323, opened 2023-10-03
Add DNS support for cluster replica address

### LINKED ISSUE #815: Add DNS support for cluster replica address
When running a cluster in Kubernetes we sometimes need to reorganize or restart memgraph replicas and/or mains, due to the nature of how networking works it's annoying / impossible to ensure the "respawned" instances have the same IP.

I would like that the memgraph main supported DNS look ups so that a replica could be registered as such:

```
REGISTER REPLICA replica-0 ASYNC TO "memgraph-replica-0.components";
```

This would allow us to move, recreate and scale replicas without having to un-register / register each time.

DavIvek:
Currently, users can only register replicas by providing ip addresses (and optionally ports). This PR should enable the user to register a replica as such:
`REGISTER REPLICA replica0 ASYNC TO "memgraph-replica-0.components";`
This would allow users to move, recreate, and scale replicas without having to un-register / register each time.

[master < Task] PR

closes #815


src/io/network/endpoint.cpp
```diff
@@ -40,26 +41,24 @@ Endpoint::IpFamily Endpoint::GetIpFamily(const std::string &ip_address) {
...
+  std::string parsed_address;
 
   std::vector<std::string> parts = utils::Split(address, delimiter);
   if (parts.size() == 1) {
     if (default_port) {
-      if (GetIpFamily(address) == IpFamily::NONE) {
-        return std::nullopt;
-      }
-      return std::pair{address, *default_port};
+      return std::pair{parsed_address, *default_port};
```
DavIvek, line 57: I removed this check because the same check is performed later after creating the Endpoint structure in replication.cpp:160.

src/io/network/endpoint.cpp
```diff
@@ -109,4 +111,15 @@ std::ostream &operator<<(std::ostream &os, const Endpoint &endpoint) {
   return os << endpoint.address << ":" << endpoint.port;
 }
 
+bool Endpoint::ValidReplicaAddress() const {
```
Ignition, line 114: To specific, this should be replica agnostic code. Better name "IsResolvableAddress"

src/io/network/endpoint.cpp
```diff
@@ -40,26 +41,24 @@ Endpoint::IpFamily Endpoint::GetIpFamily(const std::string &ip_address) {
 std::optional<std::pair<std::string, uint16_t>> Endpoint::ParseSocketOrIpAddress(
     const std::string &address, const std::optional<uint16_t> default_port = {}) {
   /// expected address format:
-  ///   - "ip_address:port_number"
-  ///   - "ip_address"
+  ///   - "address:port_number"
```
Ignition, line 44: Do not change the behaviour of `ParseSocketOrIpAddress` 
Introduce a new method for different behaviour.

src/io/network/endpoint.hpp
```diff
@@ -33,7 +33,7 @@ struct Endpoint {
   Endpoint &operator=(Endpoint &&) noexcept = default;
   ~Endpoint() = default;
 
-  enum class IpFamily : std::uint8_t { NONE, IP4, IP6 };
+  enum class IpFamily : std::uint8_t { NONE, IP4, IP6, DNS_ALIAS };
```
Ignition, line 36: DNS_ALIAS does not make sense for IpFamily

Ignition (changes requested):
Can't approve, even if this does work. ATM it decreases the comprehensibility of the code.
IP and URL have different distinct meanings, hence our code should also be distinct about them.

src/io/network/endpoint.hpp
DavIvek: Then when we register a replica with url instead of ip address should we just skip this IpFamily check? I'm asking this because when Endpoint is created it will throw an error if the address isn't neither IP4 nor IP6.

src/io/network/endpoint.cpp
```diff
@@ -109,4 +145,78 @@ std::ostream &operator<<(std::ostream &os, const Endpoint &endpoint) {
...
+std::optional<std::pair<std::string, uint16_t>> Endpoint::ParseSocketOrAddress(
+    const std::string &address, const std::optional<uint16_t> default_port = {}) {
+  const std::string delimiter = ":";
+  std::vector<std::string> parts = utils::Split(address, delimiter);
+  if (parts.size() == 1) {
+    if (GetIpFamily(address) == IpFamily::NONE) {
+      if (!IsResolvableAddress(address, *default_port)) {
+        return std::nullopt;
+      }
+      return ParseHostname(address, default_port);
```
Ignition, line 168: We do not want to prematurely resolve to an ip address.
We only want to know that ATM it is resolvable, we only want to split off the address from the port. 
Reason for this is this address is persisted, which means at server restart we want the url to be recovered so we can re-resolve the ip address.
DNS entries may be different on restart.

Ignition (changes requested):

src/storage/v2/replication/replication_client.cpp
```diff
@@ -29,11 +29,18 @@ ReplicationClient::ReplicationClient(Storage *storage, const memgraph::replicati
                                      const memgraph::replication::ReplicationEpoch *epoch)
     : name_{config.name},
       rpc_context_{CreateClientContext(config)},
-      rpc_client_{io::network::Endpoint(config.ip_address, config.port), &rpc_context_},
+      rpc_client_{io::network::Endpoint(), &rpc_context_},
```
Ignition, line 32: Surely `io::network::Endpoint{io::network::Endpoint::needs_resolving, config.ip_address, config.port}` would be fine here.

DavIvek:
@vpavicic maybe we could update part of the documentation about REGISTER Replica instances by adding an example for registration of replica with a hostname.

PR #1324, opened 2023-10-03
Parametrize CI workflows to be able to change build type

Darych:
Community/BSL default is going to be `Release` -> the same as before -> we just added more optionality into the build system


release/package/run.sh
```diff
@@ -89,9 +89,9 @@ make_package () {
     docker exec "$build_container" bash -c "cd $container_build_dir && rm -rf ./*"
     # TODO(gitbuda): cmake fails locally if remote is clone via ssh because of the key -> FIX
     if [[ "$os" =~ "-arm" ]]; then
-        docker exec "$build_container" bash -c "cd $container_build_dir && $ACTIVATE_TOOLCHAIN && cmake -DCMAKE_BUILD_TYPE=release -DMG_ARCH="ARM64" $telemetry_id_override_flag .."
+        docker exec "$build_container" bash -c "cd $container_build_dir && $ACTIVATE_TOOLCHAIN && cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DMG_ARCH="ARM64" $telemetry_id_override_flag .."
```
gitbuda, line 92: Parametrize `RelWithDebInfo` and inject it from the `package_all` workflow, `RelWithDebInfo` should be a default, but we want to be able to inject e.g. `Release` if needed.

andrejtonev:
Do we really want RelWithDebInfo everywhere?
In my opinion we shouldn't ship with debug info. Are we going to strip the symbols later on?

release/package/run.sh
Darych: Done

.github/workflows/daily_benchmark.yaml
```diff
@@ -32,7 +32,7 @@ jobs:
 
           # Build only memgraph release binaries.
           cd build
-          cmake -DCMAKE_BUILD_TYPE=release ..
```
gitbuda, line 35: revert to release, no config

.github/workflows/diff.yaml
```diff
@@ -390,7 +390,7 @@ jobs:
 
           # Build only memgraph release binaries.
           cd build
-          cmake -DCMAKE_BUILD_TYPE=release ..
+          cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo ..
```
gitbuda, line 393: release, no config

.github/workflows/package_all.yaml
```diff
@@ -8,6 +8,10 @@ on:
       memgraph_version:
         description: "Memgraph version to upload as. If empty upload is skipped. Format: 'X.Y.Z'"
         required: false
+      build_type:
+        description: "Memgraph Build type. Default value is RelWithDebInfo."
+        default: "RelWithDebInfo"
```
gitbuda, line 13: Release

.github/workflows/release_debian10.yaml
```diff
@@ -33,7 +33,7 @@ jobs:
 
           # Build community binaries.
           cd build
-          cmake -DCMAKE_BUILD_TYPE=release -DMG_ENTERPRISE=OFF ..
+          cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DMG_ENTERPRISE=OFF ..
```
gitbuda, line 36: The same mode as we ship to the downloads -> Release at the moment, config

.github/workflows/release_centos8.yaml
```diff
@@ -190,7 +190,7 @@ jobs:
 
           # Build release binaries.
           cd build
-          cmake -DCMAKE_BUILD_TYPE=release ..
+          cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo ..
```
gitbuda, line 193: The same mode as we ship to the downloads -> Release at the moment, config

.github/workflows/release_ubuntu2004.yaml
```diff
@@ -33,7 +33,7 @@ jobs:
 
           # Build community binaries.
           cd build
-          cmake -DCMAKE_BUILD_TYPE=release -DMG_ENTERPRISE=OFF ..
+          cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DMG_ENTERPRISE=OFF ..
```
gitbuda, line 36: The same mode as we ship to the downloads -> Release at the moment, config

tests/mgbench/Dockerfile.mgbench_client
```diff
@@ -27,7 +27,7 @@ RUN source /opt/toolchain-v4/activate && \
     rm -r build && \
     mkdir build && \
     cd build && \
-    cmake -DCMAKE_BUILD_TYPE=release .. && \
+    cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo .. && \
```
gitbuda, line 30: release, no config

tools/setup
```diff
@@ -8,7 +8,7 @@ mkdir -p ${script_dir}/build
 cd ${script_dir}/build
 
 # Setup cmake
-cmake -DCMAKE_BUILD_TYPE=Release \
+cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo \
```
gitbuda, line 11: the same as default -> Release, check where this is called, and pass value if possible

.github/workflows/package_all.yaml
```diff
@@ -8,6 +8,10 @@ on:
       memgraph_version:
         description: "Memgraph version to upload as. If empty upload is skipped. Format: 'X.Y.Z'"
         required: false
+      build_type:
```
MarkoBarisic, line 11: Perhaps change this input field to be a choice from predefined build types so we eliminate errors due to misspelling, the code would go like this
build_type:
  type: choice
  description: "..."
  default: 'RelWithDebInfo'
  options:
  - Release
  - RelWithDebInfo

.github/workflows/release_centos8.yaml
Darych: As I understand, we can make configurable build here only if make build type as option as for package_all. But we can't sync package_all option with this one I believe.

gitbuda (changes requested):
@Darych I think you missed 2 packages 👀 

line 99 in `package_all`
```
./run.sh package debian-11 --for-docker
./run.sh docker
```
These are used to build plain Docker

.github/workflows/release_debian10.yaml
gitbuda: Yes because, Jepsen depends on Debian 10, a we build natively on the host os (we have to change that), but this is a good point when we move to a unified runners CC @MarkoBarisic

.github/workflows/package_all.yaml
```diff
@@ -89,7 +96,7 @@ jobs:
       - name: "Build package"
         run: |
           cd release/package
-          ./run.sh package debian-11 --for-docker
+          ./run.sh package debian-11 ${{ github.event.inputs.build_type }} --for-docker
           ./run.sh docker
```
Darych, line 100: @gitbuda 
I don't think that we need to add build type for `run.sh docker` because it will use package from debian-11 from line 99.
The only thing that inside, docker script as temporary directory uses release folder here: https://github.com/memgraph/memgraph/blob/build-release-bins-with-debug-info/release/package/run.sh#L138
I'm not sure that we need to change this.

gitbuda:
I've run `Package All` workflow, there are some bugs, all are fixed, but I have to push last commit, will do that as soon as the current actions run passed because we actually need the Docker image from there

.github/workflows/package_all.yaml
```diff
@@ -20,7 +27,7 @@ jobs:
           fetch-depth: 0 # Required because of release/get_version.py
       - name: "Build package"
         run: |
-          ./release/package/run.sh package centos-7
+          ./release/package/run.sh package centos-7 ${{ github.event.inputs.build_type }}
```
andrejtonev, line 30: When do we actually pass the `github.event.inputs.build_type`?

andrejtonev (commented):
Looks good

.github/workflows/package_all.yaml
```diff
@@ -8,6 +8,13 @@ on:
       memgraph_version:
         description: "Memgraph version to upload as. If empty upload is skipped. Format: 'X.Y.Z'"
         required: false
+      build_type:
+        type: choice
+        description: "Memgraph Build type. Default value is Release."
+        default: 'Release'
+        options:
```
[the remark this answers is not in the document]
gitbuda: not needed because we only package in Release of RelWithDebInfo

.github/workflows/package_all.yaml
gitbuda: github does based on the spec above

tools/setup
```diff
@@ -1,5 +1,30 @@
...
+      if [[ "$supported_build_type" == "${build_type}" ]]; then
+            is_build_type_ok=true
+      fi
+done
+
+if [[ "$is_build_type_ok" == false ]]; then
+      echo "Unsupported build type: $build_type"
+      exit 1
+fi
+
```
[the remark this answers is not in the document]
gitbuda: let's skip deduplication for now

PR #1325, opened 2023-10-03
Decouple pure replication state from storage

Ignition:
This PR is decoupling:
- epoch
- durability
- replication role 
  - MAIN + replica connection
  - REPLICA + address and port

- [x] Extract pure replication state from replication storage state
- [x] Introduce a common handler using replication state + storage 
- [x] Decouple interpreter calls from storage (go via handler where possible)
- [x] Move parts to own CMake target


src/query/interpreter.cpp
```diff
@@ -286,39 +305,29 @@ class ReplQueryHandler final : public query::ReplicationQueryHandler {
...
-      case ReplicationQuery::SyncMode::ASYNC: {
-        repl_mode = storage::replication::ReplicationMode::ASYNC;
-        break;
-      }
-      case ReplicationQuery::SyncMode::SYNC: {
-        repl_mode = storage::replication::ReplicationMode::SYNC;
-        break;
-      }
-    }
+    auto repl_mode = convertToReplicationMode(sync_mode);
```
as51340, line 317: IIFE would be better IMO since the lambda is used just once.

src/query/interpreter.cpp
```diff
@@ -327,20 +336,26 @@ class ReplQueryHandler final : public query::ReplicationQueryHandler {
...
-      throw QueryRuntimeException(fmt::format("Couldn't unregister the replica '{}'", replica_name));
+  /// @throw QueryRuntimeException if an error occurred.
+  void DropReplica(std::string_view replica_name) override {
+    auto const result = handler_.UnregisterReplica(replica_name);
+    switch (result) {
+      using enum memgraph::storage::UnregisterReplicaResult;
+      case NOT_MAIN:
+        throw QueryRuntimeException("Replica can't unregister a replica!");
+      case COULD_NOT_BE_PERSISTED:
+        [[fallthrough]];
```
as51340, line 347: Why not throw here separate QueryRuntimeException?

src/replication/state.cpp
```diff
@@ -0,0 +1,140 @@
...
+  role_persisted = memgraph::replication::RolePersisted::YES;
+
+  const auto replication_status = *maybe_replication_status;
+  auto role = replication_status.role.value_or(ReplicationRole::MAIN);
+  if (role == ReplicationRole::REPLICA) {
+    return {ReplicationServerConfig{
+        .ip_address = kDefaultReplicationServerIp,
+        .port = replication_status.port,
+    }};
+  } else {
```
as51340, line 91: Else after return

src/storage/v2/replication/enums.hpp
```diff
@@ -13,11 +13,7 @@
 #include <cstdint>
 
 namespace memgraph::storage::replication {
-enum class ReplicationRole : uint8_t { MAIN, REPLICA };
-
-enum class ReplicationMode : std::uint8_t { SYNC, ASYNC };
 
 enum class ReplicaState : std::uint8_t { READY, REPLICATING, RECOVERY, INVALID };
```
as51340, line 17: I think this makes sense to decouple in a separate file.

as51340:
Just a note that clang tidy is failing for you

src/query/interpreter.cpp
Ignition: One change I made is to make it `inline`. 

Even though it is only used once, I want to pull out these replication -> query, and query -> replication conversions into their own functions. In order to make the functions that use them easy to follow.

i.e. don't want the lambda to distract from the rest of the code.

src/query/interpreter.cpp
Ignition: Refactor for now, maybe in future we will add another QueryRuntimeException.

src/storage/v2/replication/enums.hpp
Ignition: In another refactor

as51340 (approved):

Ignition:
related to #1359

PR #1327, opened 2023-10-03
Fix CPU 100% usage

### LINKED ISSUE #1316: [BUG] Memgraph CPU is on 100% on one busy thread
**Memgraph version**
v2.11

**Environment**
Ubuntu, Docker

**Describe the bug**
One of the Memgraph threads seems that it will not stop any time soon. It causes one of the CPU cores to be running at 100% for quite a long time.

**To Reproduce**
Steps to reproduce are currently unavailable.

**Expected behavior**
An assumption is that this is an invalid behaviour and that the thread should not be busy like this. There are reports from clients that this behaviour is temporary and that the problem goes away, but only after a few hours.

**Additional context**

gitbuda:
- [x] Explore more the idea that there is a while true loop because OnRead is calling DoRead and other way around -> this is correct because at some point session ends in constant spinning between the OnRead DoRead calls

closes #1316


Ignition:
State machine logic:
Need to dispatch task periodically to check and do DoShutdown (or another mechanism to signal to the Session to shutdown)
Need to dispatch is not authenticated to loop until authenticated.

Ignition:
Maybe the issue is that `buffer_.consume(buffer_.size());` is needed even in the `IsAuthenticated() == true` path.
Not sure what is on the stream that is being sent to us, but if its not consumed then the `ws.async_read` would just immediately trigger.

Darych:
Added error handling for websocket read operation

PR #1332, opened 2023-10-05
Cache the indices upon reading them from disk

gvolfing:
Create and utilize caches for indices, one per index access type. This is needed so different access methods within one transaction will not override the otherwise would-be common cache. On top of the caches themselves a counter has also been added per access methods type to make sure that calls to `AdvanceCommand()` are handled correctly from the caching point of view.


src/storage/v2/disk/storage.cpp
```diff
@@ -540,61 +540,110 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(View view) {
 
 VerticesIterable DiskStorage::DiskAccessor::Vertices(LabelId label, View view) {
   auto *disk_storage = static_cast<DiskStorage *>(storage_);
+  const auto cache_key = label;
```
as51340, line 543: No need for that

src/storage/v2/disk/storage.cpp
```diff
@@ -540,61 +540,110 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(View view) {
...
+    auto gids = merge_with_main_cache(label_index_cache, index_delta_storage);
+    LoadVerticesFromDiskLabelIndex(gids, cache_key, *index_deltas, *indexed_vertices);
+  } else {
+    if (label_index_cache.contains(cache_key) && view == View::OLD) {
+      merge_with_main_cache(label_index_cache, index_delta_storage);
+    } else {
+      auto gids = merge_with_main_cache(label_index_cache, index_delta_storage);
+      LoadVerticesFromDiskLabelIndex(gids, cache_key, *index_deltas, *indexed_vertices);
+    }
+  }
```
as51340, line 579: I don't think this is correct logic but may be wrong. First I would suggest removing these 3 variables:
```
auto &label_index_cache = transaction_.label_index_cache_;
auto &label_index_cache_ci = transaction_.label_index_cache_ci_;
auto &index_delta_storage = transaction_.index_deltas_storage_;
```
I think it just makes code harder to read. Whether you write in the code transaction_.label_index_cache_ or label_index_cache no big difference.
Second thing, `merge_with_main_cache` can be extracted since you can call it in all if-else branches. 
Third thing and the most important:
I don't see at any place that you return cached vertices instead of going to disk. If condition
`if (label_index_cache.contains(cache_key) && view == View::OLD)` is true, you call just `merge_with_main_cache` but in that method you already allocated new SkipList and new deltas, you didn't reuse anything. So not sure how currently this PR improves logic.

as51340: This comment applies for all other methods too

as51340 (changes requested):

src/storage/v2/disk/storage.cpp
gvolfing: 1. It is not much more to write, but I do not agree with you that it is not easier to read. I changed the reference names to be more concise, please check it and let me know what you think.
2. `merge_with_main_cache` is not that convenient to extract, since the 4 index-overloaded `Vertices(...)` functions are using different function calls to do the merge itself, where the functions are members of the `Accessor` class, meaning they are not convenient to be abstracted into functors(you will have to `std::bind` them or wrap them into a lambda to be able to pass them because of `this`). Using std::function to type erase might be a solution but the signature of the used member functions are different, so it would be harder to read it either way when providing a generic abstraction in my opinion. Since this implementation is likely to change once we introduce the proper on-disk indexing mechanism, I would rather keep it in this form. 
3. We already discussed this so I am leaving this here as information for other people. The caching mechanism is changing in a way where data from disk is read into the main cache which is bound to the lifetime of any given transaction and the index caches are merged with the main cache where appropriate.

src/storage/v2/transaction.hpp
```diff
@@ -113,7 +113,21 @@ struct Transaction {
...
-  std::vector<std::unique_ptr<utils::SkipList<Vertex>>> index_storage_;
+
+  using LabelIndex = LabelId;
+  using LabelPropertyIndex = std::pair<LabelId, PropertyId>;
+  using LabelPropertyValIndex = std::tuple<LabelId, PropertyId, PropertyValue>;
+  using RangeBoundary = std::optional<utils::Bound<PropertyValue>>;
+  using LabelPropertyRangeIndex = std::tuple<LabelId, PropertyId, RangeBoundary, RangeBoundary>;
+
+  std::map<LabelId, utils::SkipList<Vertex>> label_index_cache_;
+  uint64_t label_index_cache_ci_{command_id};
```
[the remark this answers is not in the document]
gvolfing: You need to track `command_id,` because calls to `AdvanceCommand()` can not be tracked just by the View itself when it comes to caching.

src/utils/bound.hpp
```diff
@@ -33,6 +33,14 @@ class Bound {
   Bound &operator=(const Bound &other) = default;
   Bound &operator=(Bound &&other) = default;
 
+  bool operator<(const Bound &other) const {
+    return std::make_tuple(value_, type_) < std::make_tuple(other.value_, other.type_);
+  }
+
+  bool operator==(const Bound &other) const {
+    return std::make_tuple(value_, type_) == std::make_tuple(other.value_, other.type_);
+  }
```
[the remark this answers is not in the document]
gvolfing: In one of the index cache datastructures, `Bound` is used as part of a key to a `std::map`.

src/storage/v2/disk/storage.hpp
```diff
@@ -63,37 +63,47 @@ class DiskStorage final : public Storage {
...
+
+    template <typename TIndex, typename TIndexCacheKey, typename TMergeFunc>
+    void SyncDeletedVertices(TIndex &index, TIndexCacheKey &cache_key,
+                             std::vector<std::list<Delta>> &index_delta_storage, View view, TMergeFunc &merge_func) {
+      if (!transaction_.vertices_to_delete_.empty()) {
+        index[cache_key] = utils::SkipList<Vertex>();
+        if (view == View::OLD) {
+          merge_func(index, index_delta_storage);
+        }
+      }
```
as51340, line 105:
```suggestion
      if (transaction_.vertices_to_delete_.empty()) {
        continue;
      }
      index[cache_key] = utils::SkipList<Vertex>();
      if (view == View::OLD) {
        merge_func(index, index_delta_storage);
      }
```
It's just a bit nicer with less indentation

src/storage/v2/disk/storage.cpp
```diff
@@ -547,54 +547,126 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(LabelId label, View view) {
...
-  auto &index_deltas = transaction_.index_deltas_storage_.back();
+  utils::SkipList<memgraph::storage::Vertex> *indexed_vertices{nullptr};
+  std::list<storage::Delta> *index_deltas{nullptr};
 
-  auto gids = MergeVerticesFromMainCacheWithLabelIndexCache(label, view, index_deltas, indexed_vertices.get());
-  LoadVerticesFromDiskLabelIndex(label, gids, index_deltas, indexed_vertices.get());
+  auto merge_with_main_cache = [&](auto &index, auto &index_delta_storage) -> std::unordered_set<Gid> {
+    indexed_vertices = &index[label];
+    index_delta_storage.emplace_back();
+    index_deltas = &index_delta_storage.back();
```
as51340, line 556:
```suggestion
    index_deltas = &index_delta_storage.emplace_back();
```

src/storage/v2/disk/storage.cpp
```diff
@@ -547,54 +547,126 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(LabelId label, View view) {
...
+        return VerticesIterable(AllVerticesIterable(cache[label].access(), storage_, &transaction_, view));
+      }
+      cache[label] = utils::SkipList<Vertex>();
+      merge_with_main_cache(cache, index_delta_storage);
+      return VerticesIterable(AllVerticesIterable(cache[label].access(), storage_, &transaction_, view));
+    }
+
+    cache[label] = utils::SkipList<Vertex>();
+    index_delta_storage.emplace_back();
+    index_deltas = &index_delta_storage.back();
```
as51340, line 586:
```suggestion
    index_deltas = &index_delta_storage.emplace_back();
```

src/storage/v2/disk/storage.cpp
```diff
@@ -547,54 +547,126 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(LabelId label, View view) {
...
+    LoadVerticesFromDiskLabelIndex({}, label, *index_deltas, transaction_.vertices_);
+  }
+
+  merge_with_main_cache(cache, index_delta_storage);
   return VerticesIterable(AllVerticesIterable(indexed_vertices->access(), storage_, &transaction_, view));
```
as51340, line 591:
```suggestion
auto merge_with_main_cache = [&](auto &index, auto &index_delta_storage) -> std::unordered_set<Gid> {
    indexed_vertices = &index[label];
    index_deltas = &index_delta_storage.emplace_back();
    return MergeVerticesFromMainCacheWithLabelIndexCache(label, view, *index_deltas, *indexed_vertices);
  };

  auto &cache = transaction_.label_index_cache_;
  auto &cache_ci = transaction_.label_index_cache_ci_;
  auto &index_delta_storage = transaction_.index_deltas_storage_;

  SyncDeletedVertices(cache, label, index_delta_storage, view, merge_with_main_cache);

  if (transaction_.command_id > cache_ci) {
    cache_ci = transaction_.command_id;
    merge_with_main_cache(cache, index_delta_storage);
    return VerticesIterable(AllVerticesIterable(indexed_vertices->access(), storage_, &transaction_, view));
  }
  if (cache.contains(label)) {
    if (view == View::NEW) {
      cache[label] = utils::SkipList<Vertex>();      
    }
    // TODO
    // we do not need this merge if we can make sure that
    // the removing and re-adding the same label to a given
    // vertex within one transaction is not permitted.
    merge_with_main_cache(cache, index_delta_storage);
    return VerticesIterable(AllVerticesIterable(cache[label].access(), storage_, &transaction_, view));
  }
  cache[label] = utils::SkipList<Vertex>();
  index_deltas = &index_delta_storage.emplace_back();
  LoadVerticesFromDiskLabelIndex({}, label, *index_deltas, transaction_.vertices_);
  merge_with_main_cache(cache, index_delta_storage);
  return VerticesIterable(AllVerticesIterable(indexed_vertices->access(), storage_, &transaction_, view));
```
I am trying to find a way to refactor this to make it a bit better... I think this should work. Just put that comment (TODO) here somewhere

as51340: I did like this for several reasons:
1. you have explicit steps to see what is going on to analyze specific use case
2. I think it is better to have return in every if case then just one at the end...

Let me know if I misunderstood the logic here

src/storage/v2/disk/storage.cpp
```diff
@@ -547,54 +547,126 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(LabelId label, View view) {
...
-  auto &index_deltas = transaction_.index_deltas_storage_.back();
+  utils::SkipList<memgraph::storage::Vertex> *indexed_vertices{nullptr};
+  std::list<storage::Delta> *index_deltas{nullptr};
 
-  const auto label_property_filter = [this](const Vertex &vertex, LabelId label, PropertyId property,
-                                            View view) -> bool {
-    return VertexHasLabel(vertex, label, &transaction_, view) &&
-           HasVertexProperty(vertex, property, &transaction_, view);
-  };
+  auto merge_with_main_cache = [&](auto &index, auto &index_delta_storage) -> std::unordered_set<Gid> {
```
as51340, line 608: All comments from before work also here

src/storage/v2/disk/storage.cpp
```diff
@@ -603,23 +675,56 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(LabelId label, PropertyId p
...
+      if (view == View::OLD) {
+        // TODO
+        // we do not need this merge if we can make sure that
+        // the removing and re-adding the same label to a given
+        // vertex within one transaction is not permitted.
+        merge_with_main_cache(cache, index_delta_storage);
+        return VerticesIterable(AllVerticesIterable(cache[cache_key].access(), storage_, &transaction_, view));
+      }
+      cache[cache_key] = utils::SkipList<Vertex>();
+      merge_with_main_cache(cache, index_delta_storage);
```
as51340, line 716: I think it is kinda not expected to go to disk just if you query by different value. IMO this query (with the assumption that label-property index :User(id) exists:
```cypher
MATCH (n:User {id: 10}), (m:User {id: 11})
RETURN n, m;
```
This query should go just once to disk I think. But if you cache based on property value it will go twice. Essentially, this means that we don't get any benefit when doing import

src/storage/v2/disk/storage.cpp
```diff
@@ -628,30 +733,64 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(LabelId label, PropertyId p
                                                      const std::optional<utils::Bound<PropertyValue>> &upper_bound,
                                                      View view) {
   auto *disk_storage = static_cast<DiskStorage *>(storage_);
+  const auto cache_key = std::make_tuple(label, property, lower_bound, upper_bound);
```
as51340, line 736: I think same applies when doing interval search

src/storage/v2/transaction.hpp
```diff
@@ -113,7 +113,21 @@ struct Transaction {
   rocksdb::Transaction *disk_transaction_;
   /// Main storage
   utils::SkipList<Vertex> vertices_;
-  std::vector<std::unique_ptr<utils::SkipList<Vertex>>> index_storage_;
+
+  using LabelIndex = LabelId;
+  using LabelPropertyIndex = std::pair<LabelId, PropertyId>;
+  using LabelPropertyValIndex = std::tuple<LabelId, PropertyId, PropertyValue>;
```
as51340, line 119: Not sure if caching by property value makes sense -> connected to one of the comments above

as51340 (changes requested):

gvolfing:
We will need to think of a more ergonomic way of handling caching in an upcoming release.

PR #1336, opened 2023-10-06
Implement key_exists under map under mgp

mpintaric55334:
To fix a bug reported by user in map.merge function, we realized we need to check if a given key exists in map. Instead of improvising a solution, it is best to just add the functionality to Cpp API, so we did that.

[master < Task] PR

RELEASE NOTE: Users can now check if a key exists in Map datatype in C++ API.


src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -1089,6 +1089,18 @@ mgp_error mgp_map_at(mgp_map *map, const char *key, mgp_value **result) {
       result);
 }
 
+mgp_error mgp_key_exists(mgp_map *map, const char *key, int *result) {
+  return WrapExceptions(
+      [&map, &key]() -> int {
+        auto found_it = map->items.find(key);
+        if (found_it == map->items.end()) {
+          return 0;
```
antoniofilipovic, line 1097: Why not return `bool`?

mpintaric55334: Other functions in impl.cpp that return bool in c++ return int in impl, so to be consistent with them

antoniofilipovic (approved):
Looks good to me

PR #1339, opened 2023-10-09
Add relaxation of write-write queries for edge import

Josipmrden:
This relaxation enables parallel import of edges concurrently. The relaxation is based on the fact that edges in our vertex structures are not ordered, and therefore we don't care how they are inserted in the collection, as long as all of them are inserted.

[master < Task] PR

### DIFF
--- a/src/storage/v2/inmemory/storage.cpp
+++ b/src/storage/v2/inmemory/storage.cpp
@@ -18,6 +18,7 @@
 #include "storage/v2/inmemory/replication/replication_client.hpp"
 #include "storage/v2/inmemory/replication/replication_server.hpp"
 #include "storage/v2/inmemory/unique_constraints.hpp"
+#include "storage/v2/mvcc.hpp"
 #include "utils/resource_lock.hpp"
 
 namespace memgraph::storage {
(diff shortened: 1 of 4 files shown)


src/storage/v2/inmemory/storage.cpp
```diff
@@ -18,6 +18,7 @@
 #include "storage/v2/inmemory/replication/replication_client.hpp"
 #include "storage/v2/inmemory/replication/replication_server.hpp"
 #include "storage/v2/inmemory/unique_constraints.hpp"
+#include "storage/v2/mvcc.hpp"
```
as51340, line 21: Remove this import

src/storage/v2/mvcc.hpp
```diff
@@ -104,6 +104,41 @@ inline bool PrepareForWrite(Transaction *transaction, TObj *object) {
...
 }
 
+/// This function prepares the object for a write. It checks whether there are
+/// any serialization errors in the process (eg. the object can't be written to
+/// from this transaction because it is being written to from another
+/// transaction) and returns a `bool` value indicating whether the caller can
+/// proceed with a write operation.
+template <typename TObj>
+inline bool PrepareForWriteWithExcludedDelta(Transaction *transaction, TObj *object,
+                                             const std::vector<Delta::Action> &exclude_deltas) {
```
as51340, line 114:
```suggestion
                                             const std::array<Delta::Action> &exclude_deltas) {
```

src/storage/v2/mvcc.hpp
```diff
@@ -104,6 +104,41 @@ inline bool PrepareForWrite(Transaction *transaction, TObj *object) {
...
+  if (object->delta == nullptr) return true;
+  auto *delta = object->delta;
+  while (delta != nullptr) {
```
as51340, line 117: Yes I mean it works.. But I don't think this is a proper solution because you don't have time connection to already existing delta, the existing delta you find could be from any transaction since last garbage collection. So in my view this is a hack which works but I think we lose consistency guarantees in some edge cases but not 100% sure. Depends largely here what we want to achieve. If we want some hack just for import, we could maybe somehow make use of edge import mode to connect deltas through some time. But if we want general improvement for WW conflicts, I would opt for updating our clients. Not sure how hard is it but this is what databases often do, there are often benchmarks comparing how many restarts databases do in order to execute some txn in highly concurrent environment

Issue #977, opened 2023-06-12
Fix QUERY MEMORY LIMIT is not honored during execution


jyenduri-uptycs:
**Memgraph version**
2.8.0

**Environment**
Ubuntu 20.04.6 LTS (GNU/Linux 5.15.0-1031-aws x86_64)
memgraphdb running as docker

**Describe the bug**
Query memory limit is not honored. 

I have a complex query which does alot of filters, when we apply QUERY MEMORY LIMIT it not honored
```
MATCH (vpcCidr:AWSCidrBlock{tenantId: '1224445', customerId: '78787777'})<-[:BLOCK_ASSOCIATION]-(vpc:AWSVpc )<-[:PART_OF_VPC]-(n:EC2Instance {
  customerId: '78787777', tenantId: '1224445', exposedInternet: true
})-[:PART_OF_SUBNET]->(s:EC2Subnet)-[:PART_OF_ROUTE_TABLE]->(:RouteTable)<-[:PART_OF_ROUTE_TABLE]-(rtr:RouteTableRoute)-[:BLOCK_ASSOCIATION]->(target:AWSCidrBlock)
WHERE rtr.gatewayId='local' AND vpcCidr.cidrBlock = target.cidrBlock
WITH vpc, n
MATCH (vpc)<--(n2:EC2Instance{tenantId: '1224445', customerId: '78787777'})-[:PART_OF_SUBNET]->(:EC2Subnet)<-[:PART_OF_SUBNET]-(:AWSNetworkAcl)<-[:MEMBER_OF_NETWORK_ACL]-(nIbound:AWSNetworkAclInboundRule{action: 'allow'})<-[:MEMBER_OF_IP_NETWORK_INBOUND_RULE]-(naclBlock:AWSCidrBlock)
WITH vpc, n2, naclBlock, nIbound
MATCH (n2:EC2Instance)-[:MEMBER_OF_EC2_SECURITY_GROUP|NETWORK_INTERFACE*..2]->(group:EC2SecurityGroup)<-[:MEMBER_OF_EC2_SECURITY_GROUP]-(ipInbound:IpPermissionInbound)<-[:MEMBER_OF_IP_RULE]-(ipRange:IpRange)
WITH n2, naclBlock,group, nIbound, ipRange, ipInbound
[...]

PR #1340, opened 2023-10-09
Use extent hooks for per query memory limit

antoniofilipovic:
Each thread is mapped to one arena in jemalloc. Since we are working with a multi-thread environment where multiple threads can be mapped to the same arena, the thread ID is used as an identifier which is mapped to the transaction id. Each transaction can therefore have multiple threads that do allocation. Each thread is mapped internally to transaction id and has its own tracker. When a transaction with that id is done, the tracker is destroyed. If we have multiple queries inside an explicit transaction, each will have the same id, but they can't overwrite each other since queries are sequentially executed.

[master < Task] PR

What needs to be done

- [x] Add test for limit per query only one - CREATE, SET, DELETE, procedure with one thread
- [x] Decide what happens with per-arena tracking if a user forgets to remove the thread from tracking
- [x] Move the new API for query tracking to a new namespace and isolate it inside the class


gitbuda:
STATUS: Ready to be merged

src/memory/utils.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+// TODO (af) think if following implementation would make sense
+/*
+std::string get_thread_id() {
+  static thread_local std::thread::id current_thread_id = std::this_thread::get_id();
+  // figure out how to cache this part
+  std::ostringstream oss;
+  oss << current_thread_id;
+  return oss.str();
+}
```
antoniofilipovic, line 31: I was thinking here maybe instead of calling get thread id every time to call this function, but I am not sure if this would work properly when threads are changed

src/memory/query_memory_control.cpp
```diff
@@ -0,0 +1,142 @@
...
+  return thread_arena;
+}
+
+void QueriesMemoryControl::AddTrackingOnArena(unsigned arena_id) { arena_tracking[arena_id].fetch_add(1); }
+
+void QueriesMemoryControl::RemoveTrackingOnArena(unsigned arena_id) { arena_tracking[arena_id].fetch_sub(1); }
+
+void QueriesMemoryControl::UpdateThreadToTransactionId(const std::thread::id &thread_id, uint64_t transaction_id) {
+  std::unique_lock<utils::RWSpinLock> guard{lock};
+  thread_id_to_transaction_id[get_string_thread_id(thread_id)] = transaction_id;
```
Ignition, line 47: Why string?

src/memory/query_memory_control.cpp
```diff
@@ -0,0 +1,142 @@
...
+}
+
+bool QueriesMemoryControl::EraseTransactionIdTracker(uint64_t transaction_id) {
+  std::unique_lock<utils::RWSpinLock> guard{lock};
+  transaction_id_tracker.erase(transaction_id);
+  return true;
+}
+
+bool QueriesMemoryControl::IsArenaTracked(unsigned arena_ind) {
+  return arena_tracking[arena_ind].load(std::memory_order_relaxed) != 0;
```
Ignition, line 89: memory_order_acquire, you don't want this to be an issue when code is LTO

src/memory/query_memory_control.hpp
```diff
@@ -0,0 +1,124 @@
...
+  std::unordered_map<std::string, uint64_t> thread_id_to_transaction_id;
+  std::unordered_map<uint64_t, utils::MemoryTracker> transaction_id_tracker;
+};
+
+inline QueriesMemoryControl &GetQueriesMemoryControl() {
+  static QueriesMemoryControl queries_memory_control_;
+  return queries_memory_control_;
+}
+
+namespace query {}  // namespace query
```
Ignition, line 96: remove, or put something in it

tests/e2e/memory/procedures/query_memory_limit_proc_multi_thread.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+  return mgp_global_alloc(two_sixty_eight_mb, (void **)(&ptr));
+}
+
+// change communication between threads with feature and promise
+std::atomic<int> num_allocations{0};
+std::vector<void *> ptrs_;
+
+void AllocFunc(mgp_graph *graph) {
+  [[maybe_unused]] const enum mgp_error tracking_error = mgp_track_current_thread_allocations(graph);
```
Ignition, line 38: Test shows `current` but no test to show `mgp_track_thread_allocations` API usage.

Ignition (changes requested):

tests/e2e/memory/procedures/query_memory_limit_proc.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include <string>
+#include <thread>
+#include <utility>
+#include <vector>
+
+#include "mg_procedure.h"
+#include "utils/on_scope_exit.hpp"
+
+enum mgp_error Alloc(void *ptr) {
+  const size_t two_sixty_eight_mb = 1 << 28;
```
Josipmrden, line 28: rename to mb268 or mb_size_268?

src/memory/query_memory_control.cpp
antoniofilipovic: Fixed

antoniofilipovic:
PR fixes this issue: https://github.com/memgraph/memgraph/issues/977

antoniofilipovic:
One more thing left to solve here:
- [ ] detach tracker from transaction after all pulling is done

Josipmrden (approved):

PR #1343, opened 2023-10-10
Provide textual information for inefficient plans with notifications

### LINKED ISSUE #1321: Enable indexing hints during query runs
The feature request comes from the fact that choosing the appropriate index between e.g. 2 similar ones is not an easy task and can produce different and inefficient query times if chosen incorrectly. 

Memgraph has the `ANALYZE GRAPH` method that can come in handy. However, not all the people have the privilege to run it as it is a very expensive operation and is likely to be run only when data is static, imported, and distribution will not change in a long time. 

Therefore, explicit indexing hints can be a good temporary workaround when using multiple properties on node filtering.

Josipmrden:
Users are often not creating indices when using read queries. Therefore, textual information is added via notifications to provide users with info and remind them to create indices.

[master < Task] PR

@vpavicic : Provided textual hints in Memgraph notification system which will be available to read in mgconsole and Memgraph Lab

closes #1321


src/query/plan/hint_provider.hpp
```diff
@@ -0,0 +1,266 @@
...
+    if (filtered_labels.empty() && filtered_properties.empty()) {
+      return;
+    }
+
+    if (scan_type == ScanAll::kType) {
+      if (!filtered_labels.empty() && !filtered_properties.empty()) {
+        hints_.push_back(
+            fmt::format("Sequential scan will be used on symbol `{0}` although there is a filter on labels {1} and "
+                        "properties {2}. Consider "
+                        "creating a label property index.",
```
vpavicic, line 236:
```suggestion
                        "creating a label-property index.",
```

src/query/plan/hint_provider.hpp
```diff
@@ -0,0 +1,266 @@
...
+#include <range/v3/view.hpp>
+
+#include "query/plan/operator.hpp"
+#include "query/plan/preprocess.hpp"
+
+namespace memgraph::query::plan {
+
+std::vector<std::string> ProvidePlanHints(const LogicalOperator *plan_root, const SymbolTable &symbol_table);
+
+class PlanHintsProvider final : public HierarchicalLogicalOperatorVisitor {
```
Ignition, line 31: Why a 2nd visitor? Why not have `IndexLookupRewriter` provide the hints.

Josipmrden: I meant of it as a hint provider not tied directly to indices -> providing any hints possible that we might encounter in the future.

src/query/plan/hint_provider.hpp
```diff
@@ -0,0 +1,266 @@
...
+ public:
+  PlanHintsProvider(const SymbolTable &symbol_table) : symbol_table_(symbol_table) {}
+
+  std::vector<std::string> hints() { return hints_; }
```
as51340, line 35:
```suggestion
  std::vector<std::string> &hints() { return hints_; }
```

src/query/plan/hint_provider.hpp
```diff
@@ -0,0 +1,266 @@
...
+          "Label index will be used on symbol `{0}` although there is also a filter on properties {1}. Consider "
+          "creating a label property index.",
+          scan_symbol.name(), filtered_properties));
+      return;
+    }
+  }
+
+  std::string ExtractAndJoin(auto &&collection, auto &&projection) {
+    auto elements = collection | ranges::views::transform(projection);
+    return boost::algorithm::join(elements, ", ");
```
as51340, line 262: you have ours utils::Join

src/query/plan/preprocess.hpp
```diff
@@ -313,6 +313,17 @@ class Filters final {
     return labels;
   }
 
+  auto FilteredProperties(const Symbol &symbol) const {
```
as51340, line 316:
```suggestion
  auto FilteredProperties(const Symbol &symbol) -> std::unordered_set<PropertyIx> const {
```

src/query/plan/hint_provider.hpp
```diff
@@ -0,0 +1,266 @@
...
+  bool PostVisit(LoadCsv & /*op*/) override { return true; }
+
+ private:
+  const SymbolTable &symbol_table_;
+  std::vector<std::string> hints_;
+
+  bool DefaultPreVisit() override { throw utils::NotYetImplemented("providing plan hints"); }
+
+  void HintIndexUsage(Filter &op) {
+    if (auto *maybe_scan_operator = dynamic_cast<ScanAll *>(op.input().get()); !maybe_scan_operator) {
```
as51340, line 213: does it have to be always ScanAll before Filter?

src/query/plan/hint_provider.hpp
```diff
@@ -0,0 +1,266 @@
...
+
+  bool PreVisit(LoadCsv &op) override { return true; }
+
+  bool PostVisit(LoadCsv & /*op*/) override { return true; }
+
+ private:
+  const SymbolTable &symbol_table_;
+  std::vector<std::string> hints_;
+
+  bool DefaultPreVisit() override { throw utils::NotYetImplemented("providing plan hints"); }
```
as51340, line 210: I don't understand this error message

src/query/plan/hint_provider.hpp
```diff
@@ -0,0 +1,266 @@
...
+  bool PostVisit(Filter &op) override {
+    HintIndexUsage(op);
+    return true;
+  }
+
+  bool PreVisit(ScanAll &op) override { return true; }
+
+  // Replace ScanAll with ScanAllBy<Index> in PostVisit, because removal of
+  // ScanAll may remove the last reference and thus free the memory. PostVisit
+  // should be the last thing ScanAll::Accept does, so it should be safe.
```
as51340, line 57: I think not relevant for this file

src/query/plan/hint_provider.hpp
```diff
@@ -0,0 +1,266 @@
...
+  using HierarchicalLogicalOperatorVisitor::PreVisit;
+  using HierarchicalLogicalOperatorVisitor::Visit;
+
+  bool Visit(Once &) override { return true; }
+
+  bool PreVisit(Filter &op) override { return true; }
+
+  // Remove no longer needed Filter in PostVisit, this should be the last thing
+  // Filter::Accept does, so it should be safe to remove the last reference and
+  // free the memory.
```
as51340, line 47: I think not relevant for this file

src/query/plan/hint_provider.hpp
```diff
@@ -0,0 +1,266 @@
...
+  bool PostVisit(ScanAll &scan) override { return true; }
+
+  bool PreVisit(Expand &op) override { return true; }
+
+  bool PostVisit(Expand & /*expand*/) override { return true; }
+
+  bool PreVisit(ExpandVariable &op) override { return true; }
+
+  // See if it might be better to do ScanAllBy<Index> of the destination and
+  // then do ExpandVariable to existing.
```
as51340, line 67: I think not relevant for this file

src/query/plan/hint_provider.hpp
```diff
@@ -0,0 +1,266 @@
...
+  bool PostVisit(ExpandVariable &expand) override { return true; }
+
+  // The following operators may only use index lookup in filters inside of
+  // their own branches. So we handle them all the same.
+  //  * Input operator is visited with the current visitor.
+  //  * Custom operator branches are visited with a new visitor.
```
as51340, line 73: I think not relevant for this file

as51340 (changes requested):

src/query/plan/hint_provider.hpp
Josipmrden: i tried to switch but it seems the elements are not a plain vector of strings but a view. so I'll keep it because boost manages to work with it

src/query/plan/hint_provider.hpp
as51340: but if it can be and we do dynamic_cast that will fail?

as51340 (approved):

PR #1348, opened 2023-10-11
Fix cached plan not getting invalidated

### LINKED ISSUE #1251: Fix simple MERGE with variable causes a big memory spike
**Memgraph version**
2.10.1

**Environment**
Memgraph Docker

**Describe the bug**
During the batch execution of the `MERGE` clause from GQL Alchemy, the `MERGE` clause that contains variable for a node will cause a significant memory spike 30x times. . 

**To Reproduce**
Causes an issue: 
```
#Bulk query execution (stopped at 100k nodes). Memory usage notes: start 77MB, end 2.57 GB! After FREE MEMORY 2.55 GB (approximated 6GB for 200k nodes ), after restart 102 MB

memgraph_conn = Memgraph(host='localhost', port=7687, username="", password="")
batch_data = []
batch_size = 300
index = 1
address = 1000
memgraph_conn.execute("CREATE INDEX ON :MyNode(address)")
for i in range(200000):
    s = f"s{str(index)}"
    query = f"MERGE ({s}:MyNode {{address: {address}}})"
    batch_data.append(query)
    index += 1
    address += 1

    if len(batch_data) >= batch_size:
        memgraph_conn.execute("".join(batch_data))
        batch_data = []

```
```
#Bulk query execution without MERGE variable. Memory usage notes: start 77MB, end 212 MB! After restart 126 MB

memgraph_conn = Memgraph(host='localhost', port=7687, username="", password="")
batch_data = []
batch_size = 300
index = 1
address = 1000
memgraph_conn.execute("CREATE INDEX ON :MyNode(address)")
for i in range(200000):
    s = f"s{str(index)}"
    query = f"MERGE (:MyNode {{address: {address}}})"
    batch_data.append(query)
    index += 1
    address += 1

    if len(batch_data) >= batch_size:
        memgraph_conn.execute("".join(batch_data))
        batch_data = []

```

**Expected behavior**
These two workloads should have similar memory usage. 

**Logs**
Not applicable 

**Additional context**
Also, once Memgraph was out of Memroy, there was no warning.

DavIvek:
This PR  is about resolving memory spike that happens when we use variable in MERGE statement. The MERGE clause that contains variable for a node will cause a significant memory spike 30x times. 

[master < Task] PR

closes #1251


DavIvek:
If we avoid to use monotonic memory, the difference in memory consumption (when we use variable and when we don't) will decrease but still remain significant.

gitbuda:
The problem here seems to be the caching of query plans (probably query plan cache). The solution is to implement some cache eviction strategy, LRU or some background process (LRU is the best place to start).

DavIvek:
Since we are not invoking a garbage collector for plan cache, for testing purposes, I changed the `Interpreter::PrepareIndexQuery` part of the function where we remove cached plans from the skip list. Instead of calling remove for every cached plan, now we call the skip lists clear() function to force memory deallocation because remove() doesn't deallocate memory, it only marks skip list entry as deleted (and the garbage collector should do deallocation when it's safe to do it). 

With that minor change, I was able to release all the memory used by the plan cache:
Firstly, by executing query: 'CREATE INDEX ON: MyNode1(address)', we will call `plan_cache->clear()`, following should remove all elements from the skip list.
After that, memory still isn't released, but if we execute the 'FREE MEMORY' query, memory usage should drop drastically. 

I'm still confused about what actually happens there when we call the FREE MEMORY query.

DavIvek:
TODO:

- [ ] rename query-plan-cache-ttl flag and think of what would be good default value
- [ ] remove unnecessary parts of code

src/dbms/database.hpp
```diff
@@ -145,9 +147,11 @@ class Database {
   /**
    * @brief Returns the PlanCache vector raw pointer
    *
-   * @return utils::SkipList<query::PlanCacheEntry>*
+   * @return uutils::Synchronized<utils::LRUCache<uint64_t, std::shared_ptr<CachedPlan>>, utils::SpinLock>
    */
-  utils::SkipList<query::PlanCacheEntry> *plan_cache() { return &plan_cache_; }
+  utils::Synchronized<utils::LRUCache<uint64_t, std::shared_ptr<query::CachedPlan>>, utils::SpinLock> *plan_cache() {
+    return &plan_cache_;
```
antoniofilipovic, line 153: You can use here I think utils::RWSpinLock, I think it should be even better than utils::SpinLock

src/utils/lru_cache.hpp
```diff
@@ -0,0 +1,64 @@
...
+    if (it != item_map.end()) {
+      item_list.erase(it->second);
+      item_map.erase(it);
+    }
+    item_list.push_front(std::make_pair(key, val));
+    item_map.insert(std::make_pair(key, item_list.begin()));
+    clean();
+  };
+  bool exist(const TKey &key) { return (item_map.count(key) > 0); };
+  TVal get(const TKey &key) {
```
antoniofilipovic, line 36: Can you return here TVal &, so instead of making a copy, you return by ref, I am not sure if this is what you wanted

src/utils/lru_cache.hpp
```diff
@@ -0,0 +1,64 @@
...
+class LRUCache {
+ public:
+  LRUCache(int cache_size_) : cache_size(cache_size_){};
+
+  void put(const TKey &key, const TVal &val) {
+    auto it = item_map.find(key);
+    if (it != item_map.end()) {
+      item_list.erase(it->second);
+      item_map.erase(it);
+    }
```
antoniofilipovic, line 30: Why are you cleaning if there was an element found, and then push again?

src/utils/lru_cache.hpp
```diff
@@ -0,0 +1,64 @@
...
+    item_list.push_front(std::make_pair(key, val));
+    item_map.insert(std::make_pair(key, item_list.begin()));
+    clean();
+  };
+  bool exist(const TKey &key) { return (item_map.count(key) > 0); };
```
antoniofilipovic, line 35: When will count be greater than 1, is that even possible?

src/utils/lru_cache.hpp
```diff
@@ -0,0 +1,64 @@
...
+    return it->second->second;
+  };
+  void clear() {
+    item_list.clear();
+    item_map.clear();
+  };
+  size_t size() { return item_map.size(); };
+
+ private:
+  void clean() {
```
antoniofilipovic, line 51: consider naming this method `try_clean`

src/utils/lru_cache.hpp
```diff
@@ -0,0 +1,64 @@
...
+  bool exist(const TKey &key) { return (item_map.count(key) > 0); };
+  TVal get(const TKey &key) {
+    if (!exist(key)) {
+      return nullptr;
+    }
+    auto it = item_map.find(key);
+    item_list.splice(item_list.begin(), item_list, it->second);
+    return it->second->second;
+  };
+  void clear() {
```
antoniofilipovic, line 44: consider naming this method `reset`, to be more different than clear

src/utils/lru_cache.hpp
antoniofilipovic: Also, name this method `exists`

src/utils/lru_cache.hpp
DavIvek: That way we will insert element at the back of the list again, because we are not sorting elements explicitly, but appending LRU element at the back of the list. So that way we will "mark" that element as LRU.

src/utils/lru_cache.hpp
```diff
@@ -0,0 +1,64 @@
...
+#pragma once
+
+#include <list>
+#include <unordered_map>
+#include <utility>
+
+namespace memgraph::utils {
+
+template <class TKey, class TVal>
+class LRUCache {
```
antoniofilipovic, line 21: Please write tests for this class, unit tests

src/dbms/database.hpp
```diff
@@ -156,7 +160,8 @@ class Database {
   query::stream::Streams streams_;                  //!< Streams associated with the storage
 
   // TODO: Move to a better place
-  utils::SkipList<query::PlanCacheEntry> plan_cache_;  //!< Plan cache associated with the storage
+  utils::Synchronized<utils::LRUCache<uint64_t, std::shared_ptr<query::CachedPlan>>, utils::SpinLock>
+      plan_cache_;  //!< Plan cache associated with the storage
```
antoniofilipovic, line 164: Consider creating a type for this method, i.e.
`using PlanCacheLRU = utils::Synchronized<utils::LRUCache<uint64_t, std::shared_ptr<query::CachedPlan>>, utils::SpinLock>;`

antoniofilipovic (changes requested):
I need to check if we can remove PlanCacheEntry, what we talked on Slack. Overall good job, bench graph was not impacted, please add unit tests for LRUCache class

src/utils/lru_cache.hpp
DavIvek: > When will count be greater than 1, is that even possible?

No, std::map::count() returns 0 if there is not element in the map and 1 otherwise.

src/utils/lru_cache.hpp
DavIvek: I didn't realize at first that comment is about put method, not get. We will use remove part of code in put method if we want to change value for the key that is already in the cache.

src/query/cypher_query_interpreter.cpp
```diff
@@ -128,27 +129,21 @@ std::unique_ptr<LogicalPlan> MakeLogicalPlan(AstStorage ast_storage, CypherQuery
...
-    plan_cache_access.emplace(plan_cache->access());
-    auto it = plan_cache_access->find(hash);
-    if (it != plan_cache_access->end()) {
-      if (it->second->IsExpired()) {
-        plan_cache_access->remove(hash);
-      } else {
-        return it->second;
-      }
+    if (plan_cache->WithLock([&](auto &cache) { return cache.exists(hash); })) {
+      return plan_cache->WithLock([&](auto &cache) { return cache.get(hash); });
```
antoniofilipovic, line 137: Since exists is combined with get, we can maybe have function, try_get, which returns if exists, otherwise returns nullptr, so instead of taking lock twice, we take once

src/query/cypher_query_interpreter.hpp
```diff
@@ -54,14 +56,8 @@ class CachedPlan {
   const auto &symbol_table() const { return plan_->GetSymbolTable(); }
   const auto &ast_storage() const { return plan_->GetAstStorage(); }
 
-  bool IsExpired() const {
-    // NOLINTNEXTLINE (modernize-use-nullptr)
-    return cache_timer_.Elapsed() > std::chrono::seconds(FLAGS_query_plan_cache_ttl);
-  };
-
  private:
```
antoniofilipovic, line 59: Consider renaming this class to `PlanWrapper`

antoniofilipovic (changes requested):
Two comments left, and that is it

antoniofilipovic (approved):
Looks good to me, good job here :partying_face:

PR #1361, opened 2023-10-13
Add retry logic possible when conflicting transactions

Josipmrden:
Most of drivers that Memgraph is compatible with use retry logic. That was not possible so far because did not send the TransientError code because of a logical error. This fix makes it possible for drivers to re-execute the query again and succeed.

[master < Task] PR


src/glue/SessionHL.cpp
```diff
@@ -172,6 +172,11 @@ std::map<std::string, memgraph::communication::bolt::Value> SessionHL::Pull(Sess
     // Wrap QueryException into ClientError, because we want to allow the
     // client to fix their query.
     throw memgraph::communication::bolt::ClientError(e.what());
+  } catch (const utils::BasicException &) {
+    // Exceptions inheriting from BasicException will result in a TransientError
+    // i. e. client will be encouraged to retry execution because it
+    // could succeed if executed again.
+    throw;
```
gitbuda, line 179: Will there be a case where there will be a `BasicException` which doesn't require (even allows) retry to happen? Is it possible that here, anything that is not a `QueryException` will be treated as a transient error and will be retried? If that's the case, it seems wrong.

From the other perspective, does it make sense to introduce `RetryBasicException` and then check for that specifically (the "retry-compatible" exceptions will inherit that)?

Josipmrden: in the handlers.hpp class we always for the `BasicException` return `TransientError` which can be retried when running managed transactions.  

```
  if (dynamic_cast<const utils::BasicException *>(&e)) {
    // Exception not derived from QueryException was thrown which means that
    // database probably aborted transaction because of some timeout,
    // deadlock, serialization error or something similar. We return
    // TransientError since retry of same transaction could succeed.
    return {"Memgraph.TransientError.MemgraphError.MemgraphError", e.what()};
  }
```

IMHO this is a problem because `AuthException` is also directly inherited from `BasicException`, so that means if running managed transactions with that query, they will be retried which is bad.
If queries are run with normal transactions in the driver (there are multiple methods for running transactions), then exponential backoff will not be applied.

I have added a semantical class `RetryBasicException` between `BasicException` and `TransactionSerializationException` and now that exception will be retried. 

The example of auth exception and a possible bug on that side exists now and has existed before the fix, so I would not change the scope of the PR to look for all exceptions and see which ones need retrying and which don't -> we need additional tests for it.

Josipmrden:
@vpavicic Not sure what to put in the docs since now the drivers will automatically retry the query.
For changelog -> Fixed wrong exceptions thrown on conflicting transactions. Drivers are now able to retry the queries in order to complete it successfully.

penghs520:
I wonder if this is compatible with older drivers.
I am currently using the v4.4.9 driver to connect to the latest version of memgraph because my java version is still stuck in jdk8
@Josipmrden

antejavor:
@penghs520 It will work with all drivers that support transient errors, I think the v4 has those. The change is on the Memgraph side, if the driver supports transient errors it will work.  But testing it will provide the full picture.

Issue #1359, opened 2023-10-12
Integrate multi-tenancy, replication and authentication


gitbuda:
The Big Integration

gitbuda:
Done under v2.15 release. Ofc there are further improvements and extensions, feel free to leave any feedback here as well.

PR #1370, opened 2023-10-16
Fix subtle replication bug

Ignition:
When going from REPLICA to MAIN, need to ensure current WAL files are finalised.


gitbuda (approved):

Ignition:
related to #1359

PR #1378, opened 2023-10-18
Replication refactor (part 5)

andrejtonev:
Extract common replication state out of storage, to have single common replication state per instance.

Fix durability of replication state, introducing versioning and migration to accommodate a common MAIN epoch (previously determined via storage recovery).

Single replication server, registered handlers now send their db_name in the payload and is demultiplexed on the server side. These handler methods are now static, as they are not intrusive of the storage itself. 

RPC protocol now includes a version number, this is to ensure memgraph servers pre db_name included in RPC payload do not have issues with those now we have change the RPC payload.

DBMS handler is now been specialized for community build. This reduces maintenance issues going forward as we can depend on a common interface for "getting the database".

These changes are a partial step in making Replication + Multi-tenancy work together.
We can now :crossed_fingers: use replication + multi-tenancy in a restricted way (ie. multi-tenency on MAIN will not break the REPLICA, because only memgraph default database will be replicated). 
We also have a feature flag at build time `MG_EXPERIMENTAL_REPLICATION_MULTITENANCY` which lets us replicate other multi-tenency databases, this is not fully featured yet so should not be used.

[master < Task] PR


src/rpc/version.hpp
```diff
@@ -0,0 +1,29 @@
...
+namespace memgraph::rpc {
+
+using Version = uint64_t;
+
+// versioning of RPC was/will be introduced in 2.13
+// We start the versioning with a strange number, to radically reduce the
+// probability of accidental match/conformance with pre 2.13 versions
+constexpr auto v1 = Version{2023'10'30'0'2'13};
+
+constexpr auto bad = Version{9999'10'30'0'2'13};
```
Ignition, line 25: TODO: remove this was left in from testing

src/dbms/replication_handler.cpp
```diff
@@ -0,0 +1,233 @@
...
+}  // namespace
+
+bool ReplicationHandler::SetReplicationRoleMain() {
+  auto const main_handler = [](RoleMainData const &) {
+    // If we are already MAIN, we don't want to change anything
+    return false;
+  };
+  auto const replica_handler = [this](RoleReplicaData const &) {
+    dbms_handler_.ForEach([](Database *db) {
+      // STEP 1) bring down all REPLICA servers
```
gvolfing, line 51: We should keep the `STEP` comments on the same level as the others, can we move this above the `ForEach` call?

src/dbms/replication_handler.cpp
```diff
@@ -0,0 +1,233 @@
...
+      return memgraph::dbms::RegisterReplicaError::NAME_EXISTS;
+    case memgraph::replication::RegisterReplicaError::END_POINT_EXISTS:
+      return memgraph::dbms::RegisterReplicaError::END_POINT_EXISTS;
+    case memgraph::replication::RegisterReplicaError::COULD_NOT_BE_PERSISTED:
+      return memgraph::dbms::RegisterReplicaError::COULD_NOT_BE_PERSISTED;
+    case memgraph::replication::RegisterReplicaError::SUCCESS:
+      break;
+  }
+
+  bool ok = true;
```
gvolfing, line 131: Can we rename this to something like `IsRegistrationSuccesful` or something more descriptive?

src/replication/include/replication/state.hpp
```diff
@@ -21,48 +21,70 @@
...
-  auto FetchReplicationData() -> FetchReplicationResult;
+
+  using ReplicationData_t = std::variant<RoleMainData, RoleReplicaData>;
+  using FetchReplicationResult_t = utils::BasicResult<FetchReplicationError, ReplicationData_t>;
+  auto FetchReplicationData() -> FetchReplicationResult_t;
+
+  auto GetRole() const -> ReplicationRole {
+    return std::holds_alternative<RoleReplicaData>(replication_data_) ? ReplicationRole::REPLICA
+                                                                      : ReplicationRole::MAIN;
+  }
```
gvolfing, line 64: I don't see a case where this would be an issue, but if we add any other `ReplicationRole` e.g. further specifying replica types if that is ever needed, then this logic will not hold up. Should we static_assert the size of the variant?

src/replication/state.cpp
```diff
@@ -11,134 +11,258 @@
...
 #include "utils/file.hpp"
+#include "utils/variant_helpers.hpp"
 
 constexpr auto kReplicationDirectory = std::string_view{"replication"};
 
 namespace memgraph::replication {
 
+auto BuildReplicaKey(std::string_view name) -> std::string {
+  auto key = std::string{durability::kReplicationReplicaPrefix};
+  key.append(name.begin(), name.end());
```
gvolfing, line 25: I think there is an overload for `string_view const &`, no need to specify begin() and end(), if you pass the entire view anyways,

src/rpc/CMakeLists.txt
```diff
@@ -7,5 +7,6 @@ find_package(fmt REQUIRED)
 find_package(gflags REQUIRED)
 
 add_library(mg-rpc STATIC ${rpc_src_files})
+add_library(mg::rpc ALIAS mg-rpc)
```
gvolfing, line 10: I'm really happy to see that we started to use cmake `ALIAS`es.

src/rpc/exceptions.hpp
```diff
@@ -19,19 +19,30 @@ namespace memgraph::rpc {
...
-            "database administrator."),
-        endpoint_(endpoint) {}
+  RpcFailedException(std::string_view msg) : utils::BasicException(msg) {}
+  SPECIALIZE_GET_EXCEPTION_NAME(RpcFailedException);
+};
 
-  /// Returns the endpoint associated with the error.
-  const io::network::Endpoint &endpoint() const { return endpoint_; }
-  SPECIALIZE_GET_EXCEPTION_NAME(RpcFailedException)
+class VersionMismatchRpcFailedException : public RpcFailedException {
```
gvolfing, line 28: Don't we want to add `final` to this classes? Not sure if that was a convention or not.

src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -9,96 +9,147 @@
...
   slk::Load(&req, req_reader);
-  replication::HeartbeatRes res{true, storage_->repl_storage_state_.last_commit_timestamp_.load(),
-                                std::string{repl_epoch_->id()}};
+  auto const db_acc = GetDatabaseAccessor(dbms_handler, req.db_name);
+  if (!db_acc) return;
+
+  // TODO: this handler is agnostic of InMemory, move to be reused by on-disk
+  auto const *storage = db_acc->get()->storage();
+  storage::replication::HeartbeatRes res{storage->id(), true,
+                                         storage->repl_storage_state_.last_commit_timestamp_.load(),
```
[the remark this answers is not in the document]
Ignition: There are a number of aspects which need restricted access. Via friendship or Passkey idiom, for now they will remain public until we get to the final stages of locking down structure.

src/replication/include/replication/replication_server.hpp
```diff
@@ -0,0 +1,65 @@
...
+  virtual ~ReplicationServer();
+
+  bool Start();
+
+ protected:
+  //  static void FrequentHeartbeatHandler(slk::Reader *req_reader, slk::Builder *res_builder);
+
+  communication::ServerContext rpc_server_context_;
+
+ public:
```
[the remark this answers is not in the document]
Ignition: There are a number of aspects which need restricted access. Via friendship or Passkey idiom, for now they will remain public until we get to the final stages of locking down structure.

src/replication/state.cpp
```diff
@@ -11,134 +11,258 @@
...
-                                                        .role = ReplicationRole::REPLICA});
-  if (durability_->Put(config.name, data.dump())) return true;
+  auto data = durability::ReplicationReplicaEntry{.config = config};
+
+  auto key = BuildReplicaKey(config.name);
+  if (durability_->Put(key, nlohmann::json(data).dump())) return true;
   spdlog::error("Error when saving replica {} in settings.", config.name);
   return false;
 }
+bool ReplicationState::SetReplicationRoleMain() {
```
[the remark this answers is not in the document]
Ignition: added newlines.

TODO: figure out clang-format setting to do this.

src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -9,96 +9,147 @@
...
-  replication::Decoder decoder(req_reader);
+  storage::replication::Decoder decoder(req_reader);
 
   auto maybe_epoch_id = decoder.ReadString();
   MG_ASSERT(maybe_epoch_id, "Invalid replication message");
 
-  auto &repl_storage_state = storage_->repl_storage_state_;
-  if (*maybe_epoch_id != repl_epoch_->id()) {
-    auto prev_epoch = repl_epoch_->SetEpoch(*maybe_epoch_id);
+  auto *storage = reinterpret_cast<storage::InMemoryStorage *>(db_acc->get()->storage());
```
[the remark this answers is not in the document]
Ignition: Done once per static function, don't want to make a static function to do a cast. Have change to be static_cast though.

src/replication/include/replication/state.hpp
Ignition: I see what you are saying about future proofing. Going to not do this for now, mainly because I don't want to put the static_assert in a header.

src/replication/state.cpp
Ignition: My IDE had issue with that, but the compiler is fine. I've made the change.

src/rpc/CMakeLists.txt
Ignition: This is because in the past I've been burned by missing target and cmake just does `-ltargetname` can happened to find something called `targetname` that I didn't want to link against.

`ALIAS` ensure it must be a cmake target

src/rpc/exceptions.hpp
Ignition: IMO minimal benefit in this situation, not going to add it.

gvolfing (approved):

CMakeLists.txt
```diff
@@ -42,10 +42,6 @@ endif()
 
 project(memgraph LANGUAGES C CXX)
 
-# NOTE: once in a while this needs to be toggled to check headers are
-#       correct and PCH isn't masking any include issues
-set(CMAKE_DISABLE_PRECOMPILE_HEADERS OFF)
```
Darych, line 47: Is it intended change?

Darych (approved):

CMakeLists.txt
Ignition: Yes, we introduce PCH + unity builds, its caused multiple issues. Removing it for now. Putting it in this PR to reduce number of PRs

Ignition:
related to #1359

PR #1379, opened 2023-10-18
Add OOM enabler in operator tree

antoniofilipovic:
This diff will introduce the OOM exception enabler to the operator tree.

The reason why OOM enabler was not introduced inside disk storage directly is that exceptions can blow during storing inside `rocksdb` and the transaction never commits, nor is the exception returned. We need to be careful when to add an OOM Exception enabler, and when to use a blocker in disk storage. Currently, the OOM enabler in the operator tree enables exceptions for all operations.

OOM enabler needs to be introduced also in mg_procedure_impl and that is what new diff will do. Currently, if a user specifies multiple threads that will do allocations from the procedure, due to the thread local variable for the OOM enabler to check if the error can be thrown, call in procedure will pass (although it shouldn't if memory is too big) and error will be not thrown.

- [x] Fix
- [x] Write e2e tests

[master < Task] PR


Josipmrden (commented):
seems that everything is okay, check grafana pls

antoniofilipovic:
This diff will introduce the OOM exception enabler to `in memory storage` only, the next diff will introduce it to disk storage. The problem currently with introducing OOM to disk storage is that exceptions can blow during storing inside `rocksdb` and transaction never commits, nor is the exception returned. We need to be careful when to add OOM Exception enabler, and when to use blocker in disk storage

tests/e2e/memory/memory_limit_accumulation.cpp
```diff
@@ -0,0 +1,80 @@
...
+    return 0;
+  }
+
+  const auto *accumulate_query = "MATCH (n) WITH n.id as id, collect(n.string) as list RETURN id, list;";
+
+  try {
+    client->Execute(accumulate_query);
+    [[maybe_unused]] auto results = client->FetchAll();
+    if (results->empty()) {
+      assert(true);
```
Josipmrden, line 68: should this pass since we're returning some results from the query?

tests/e2e/memory/memory_limit_delete.cpp
```diff
@@ -0,0 +1,80 @@
...
+    assert(false);
+    return 0;
+  }
+
+  const auto *delete_query = "MATCH (n) DETACH DELETE n RETURN count(n);";
+
+  try {
+    client->Execute(delete_query);
+    [[maybe_unused]] auto results = client->FetchAll();
+    if (results->empty()) {
```
Josipmrden, line 67: should this pass since we're returning the count?

tests/e2e/memory/memory_limit_accumulation.cpp
antoniofilipovic: Yes, because memory limit should be hit, and this should be empty.

tests/e2e/memory/memory_limit_delete.cpp
antoniofilipovic: Yes, because memory limit should be hit, and this should be empty.

as51340 (changes requested):
Several things:
- why do you need exception enabler in the operator tree?
- I think it is not good to disable OOM for disk storage
- not sure you need exception enabler in all methods (why would create index need it)?

antoniofilipovic:
> Several things:
> 
> * why do you need exception enabler in the operator tree?
> * I think it is not good to disable OOM for disk storage
> * not sure you need exception enabler in all methods (why would create index need it)?

-> The operator tree goes through all storage methods, so it is common place for all allocations to happen. It seems less error-prone than adding it to each method.
-> OOM for disk storage is not disabled but just moved to operator tree.
-> Index is also allocating memory, I am not sure why not add exception handler to that also as it might crash database.

src/storage/v2/storage.cpp
```diff
@@ -529,6 +538,7 @@ Result<std::vector<VertexAccessor>> Storage::Accessor::TryDeleteVertices(const s
 }
 
 void Storage::Accessor::MarkEdgeAsDeleted(Edge *edge) {
+  OOMExceptionEnabler oom_exception;
```
as51340, line 541: I think exception enablers in this file aren't necessary if delete operation in the operator tree has exceptions enabled.

as51340: and I think you added to the DeleteCursor

as51340 (changes requested):

as51340 (approved):

antoniofilipovic:
RN: @vpavicic : Out of memory exception is now thrown whenever memgraph is out of memory.

PR #1380, opened 2023-10-18
Fix accessing a variable bound to a list within BFS function

### LINKED ISSUE #1375: Fix accessing a variable bound to a list within a BFS filter function
The following query works:
```
MATCH p = (:Label {id:""})<-[*BFS ..2 (r, n | r.date IN ["2023-03"])]-(:Label)
RETURN p;
```
However, the following similar query returns an error:
```
WITH ["2023-03"] AS date
MATCH p = (:Label {id:"1"})<-[*BFS ..2 (r, n | r.date IN date)]-(:Label)
RETURN p;
> Client received exception: TypedValue is of type 'null', not 'list'
```

DavIvek:
The following query would throw an exception: 

```
WITH ["2023-03"] AS date
MATCH p = (:Label {id:"1"})<-[*BFS ..2 (r, n | r.date IN date)]-(:Label)
RETURN p;
> Client received exception: TypedValue is of type 'null', not 'list'`
```
with changes made by this PR, it should return the same as: 

```
MATCH p = (:Label {id:""})<-[*BFS ..2 (r, n | r.date IN ["2023-03"])]-(:Label)
RETURN p;
```

[master < Task] PR

closes #1375


DavIvek:
I believe that this PR is ready for review, but since this isn't part of the v2.12 release I canceled tests. With the last commit, I only added a test for the type of query that was failing before. Before that, all checks passed(except the release build that had some issues at that moment).

andrejtonev (commented):
Simple change :D 
Could we clean up the code a bit.
I think it would be better to remove the pointer and return a reference from the `get_list_literal` lambda. This reference would be a dereferenced address (first case) or a reference to the temp list (second case). We could then pass the ref into the other lambdas/functions instead of relying on the pointer.

src/query/interpret/eval.hpp
```diff
@@ -270,30 +270,28 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
   }
 
   TypedValue Visit(InListOperator &in_list) override {
-    TypedValue *_list_ptr = nullptr;
     TypedValue _list;
```
andrejtonev, line 273: No reason to have `_list`. This was used as a local temporary object that held a copy in case pointer was null.

src/query/interpret/eval.hpp
```diff
@@ -270,30 +270,28 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
...
     auto literal = in_list.expression1_->Accept(*this);
 
-    auto get_list_literal = [this, &in_list, &_list, &_list_ptr]() -> void {
+    auto get_list_literal = [this, &in_list]() -> TypedValue {
```
andrejtonev, line 276: Doing a copy could be fine. Just think about the types of objects that we could be copying. 
If they are large it could be a good idea to do a reference. That way we avoid a copy in case `ReferenceExpressionEvaluator` is valid.

src/query/interpret/eval.hpp
```diff
@@ -270,30 +270,28 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
...
-        _list_ptr = &_list;
+        return in_list.expression2_->Accept(*this);
       }
+      return *_list_ptr;
     };
 
-    auto do_list_literal_checks = [this, &literal, &_list_ptr]() -> std::optional<TypedValue> {
-      MG_ASSERT(_list_ptr, "List literal should have been defined");
-      if (_list_ptr->IsNull()) {
+    auto do_list_literal_checks = [this, &literal, &_list]() -> std::optional<TypedValue> {
```
andrejtonev, line 285: Would rather pass the list as an argument. No real reason to have the `_list`.

src/query/interpret/eval.hpp
```diff
@@ -312,13 +310,13 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
...
         // cache there is no need to check again as we did check first time
-        get_list_literal();
+        _list = get_list_literal();
         auto preoperational_checks = do_list_literal_checks();
         if (preoperational_checks) {
           return std::move(*preoperational_checks);
         }
         auto &cached_value = frame_change_collector_->GetCachedValue(*cached_id);
-        cached_value.CacheValue(*_list_ptr);
+        cached_value.CacheValue(_list);
```
andrejtonev, line 319: Could the list be moved?

andrejtonev (commented):
Looks good. Just a couple of comments.

src/query/interpret/eval.hpp
DavIvek: Why would we want to move _list here?

andrejtonev: There is no reason to have the `_list`, this is now just a simple temporary object. Assuming you did `auto list = get_list_literal()` above, you could then move it into the CacheValue call, so we don't do another copy. 

If I am not missing something, the list is not needed after the CacheValue call, so could be moved.

andrejtonev: The CacheValue is currently taking only const ref, but could be expanded to take in &&

Ignition (approved):

PR #1382, opened 2023-10-19
Throw when exists() combined with CASE

### LINKED ISSUE #1376: Result return failed when using exists() in CASE WHEN
I executed a query on my memgraph database and get an error.

memgraph version:2.11.0
operating system: windows 10
API：Cypher version 9

To reproduce:
CREATE ()-[:T]->();
MATCH (a) WHERE CASE WHEN TRUE THEN exists(()-[]->()) END RETURN a;

I think the query should return 2 nodes as result.But it returns 0 row of results.

as51340:
Closes #1376 
Out current architecture doesn't allow combining exists with anything except filtering because we would have to combine rule based planner with evaluation. Hence, current best solution is to explicitly throw (better than returning wrong results).

[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR


Josipmrden (approved):

as51340:
Changelog: When using exists within CASE, `utils::NotYetImplemented` exception will be thrown.

PR #1383, opened 2023-10-19
Support deleting paths

### LINKED ISSUE #1365: Add support for path deletion.
Currently, memgraph supports only the deletion of vertices or edges but a path deletion is also a feature defined by openCypher_M09. 
The next test should check if path deletion is properly supported:
```
Scenario: Detach deleting paths
    Given an empty graph
    And having executed:
      """
      CREATE (x:X), (n1), (n2), (n3)
      CREATE (x)-[:R]->(n1)
      CREATE (n1)-[:R]->(n2)
      CREATE (n2)-[:R]->(n3)
      """
    When executing query:
      """
      MATCH p = (:X)-->()-->()-->()
      DETACH DELETE p
      """
    Then the result should be empty
    And the side effects should be:
      | -nodes         | 4 |
      | -relationships | 3 |
      | -labels        | 1 | 
```

as51340:
Closes #1365 

[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR


src/query/plan/operator.cpp
```diff
@@ -2564,39 +2564,76 @@ void Delete::DeleteCursor::UpdateDeleteBuffer(Frame &frame, ExecutionContext &co
...
-          throw QueryRuntimeException("Edge not deleted due to not having enough permission!");
-        }
+        edge_auth_checker(ea);
 #endif
         buffer_.edges.push_back(ea);
         break;
       }
+      case TypedValue::Type::Path: {
+        auto path = expression_result.ValuePath();
+        std::transform(path.edges().cbegin(), path.edges().cend(), std::back_inserter(buffer_.edges),
```
andrejtonev, line 2607: Did some quick benchmarking and it seems like doing a loop to check auth and then just copying the whole vector is faster than transform.

andrejtonev: We could just directly copy in case MG_ENTERPRISE OFF, since we support path deletion, but simply do not check authentication in community.

andrejtonev (commented):
Cool addition, but I think the path code could be made faster.
Also missing tests.

src/query/plan/operator.cpp
```diff
@@ -2564,39 +2564,76 @@ void Delete::DeleteCursor::UpdateDeleteBuffer(Frame &frame, ExecutionContext &co
     expression_results.emplace_back(expression->Accept(evaluator));
   }
 
+#ifdef MG_ENTERPRISE
+  auto vertex_auth_checker = [&context](const VertexAccessor &va) {
+    if (license::global_license_checker.IsEnterpriseValidFast() && context.auth_checker &&
+        !context.auth_checker->Has(va, storage::View::NEW, query::AuthQuery::FineGrainedPrivilege::CREATE_DELETE)) {
+      throw QueryRuntimeException("Vertex not deleted due to not having enough permission!");
+    }
```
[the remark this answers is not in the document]
as51340: but it is under ifdef?

src/query/plan/operator.cpp
as51340: can you show me that benchmark, not sure what you mean?

src/query/plan/operator.cpp
```diff
@@ -2564,39 +2564,63 @@ void Delete::DeleteCursor::UpdateDeleteBuffer(Frame &frame, ExecutionContext &co
...
         break;
       }
+      case TypedValue::Type::Path: {
+        auto path = expression_result.ValuePath();
+#ifdef MG_ENTERPRISE
+        std::transform(path.edges().cbegin(), path.edges().cend(), std::back_inserter(buffer_.edges),
+                       [&edge_auth_checker](const auto &ea) {
+                         edge_auth_checker(ea);
+                         return ea;
+                       });
```
andrejtonev, line 2609: Splitting in two should be faster and a better control flow.
Is there a reason why you prefer `std::transform`?

andrejtonev (commented):
Still missing the tests.

src/query/plan/operator.cpp
as51340: tests we have in openCypher gql tests. I will just move them to main gql file. 
This task is very "functional" and corresponds to filter operation in functional programming. Process collection and take only those which satisfy filter.

andrejtonev: But you are throwing if auth fails. I assume somewhere above we catch, but in any case, you will terminate the `std::transform` on first throw, you will not copy just the authorized ones.

Plus, I am not sure the correct approach would be to delete just the elements we are authorized to. I would rather fail to delete anything if are not authorized to delete any one of the path's elements.

as51340: oh yes good point

src/query/plan/operator.cpp
```diff
@@ -2588,39 +2588,71 @@ void Delete::DeleteCursor::UpdateDeleteBuffer(Frame &frame, ExecutionContext &co
...
+                                     [&edge_auth_checker](const auto &ea) { return !edge_auth_checker(ea); });
+        auto vertices_res = std::any_of(path.vertices().cbegin(), path.vertices().cend(),
+                                        [&vertex_auth_checker](const auto &va) { return !vertex_auth_checker(va); });
+
+        if (edges_res || vertices_res) {
+          throw QueryRuntimeException(
+              "Path not deleted due to not having enough permission on all edges and vertices on the path!");
+        }
+        buffer_.nodes.insert(buffer_.nodes.begin(), path.vertices().begin(), path.vertices().end());
+        buffer_.edges.insert(buffer_.edges.begin(), path.edges().begin(), path.edges().end());
```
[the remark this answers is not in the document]
as51340: yep agree

andrejtonev (approved):
:+1:

as51340:
Changelog: The user can now detach delete path which is equivalent to deleting all nodes and relationships on that path.

PR #1385, opened 2023-10-20
Delete corrupted snapshots

### LINKED ISSUE #1369: Corrupted WAL/Snapshot files
**Memgraph version**
2.11

**Environment**
Ubuntu 22.04

**Describe the bug**
There are a couple of things going on here.

Tried to close memgraph while at 100% CPU. This caused the last Snapshot to be invalid. It was half the size of previous ones.

We are creating a Snapshot every 5 minutes. We should be retaining 3. 
There are many invalid Snapshots that we keep around and that keep getting opened and checked every 5 minutes. This seems to cause a `too many open files` error and also could be causes us to use up the RAM.
NOTE: Memgraph is idle the whole time (other than creating the snapshots).

After one of the crashes, the snapshot/wal was invalid and kept causes a critical error. Log below.
```
[2023-10-10 18:17:28.687] [memgraph_log] [warning] Found a corrupt snapshot file "/var/lib/memgraph/snapshots/20230920140500065767_timestamp_425155742" becuase of: Invalid snapshot data!
[2023-10-10 18:18:08.146] [memgraph_log] [error] Listener failed on accept: Too many open files
[2023-10-10 18:18:37.982] [memgraph_log] [critical] 
Assertion failed in file /memgraph/src/utils/file.cpp at line 346.
	Expression: 'fd_ != -1'ion failed in file /memgraph/src/utils/file.cpp at line 346.
	Expression: 'fd_ != -1'
	Message: 'While trying to open "/var/lib/memgraph/snapshots/20231010181837982557_timestamp_786807465" for writing an error occured: Too many open files (24)'
[2023-10-10 18:18:50.111] [memgraph_log] [warning] Invalid license key string. To use Enterprise features please set it to a valid string using the following query:
SET DATABASE SETTING "enterprise.license" TO "your-license-key"
[2023-10-10 18:23:27.180] [memgraph_log] [critical] You must have at least one WAL file that contains at least one delta that was created before the snapshot file!
```

**To Reproduce**
Hard to reproduce locally.

**Expected behavior**
Invalid Snapshots should be deleted or skipped. Crashing while writing a WAL or Snapshot should not leave the system in an unrecoverable state. Creating Snapshots while idle should not consume all the ram/open files.

@chrischabot

as51340:
Closes #1369 

Many issues going on in the #1369.
1. Listener failed on accept: Couldn't infer that corrupted snapshots cause too many opened files.
2. Corrupted snapshot causes memory leaking as reported by @chrischabot. I couldn't reproduce that locally, just pure loading of corrupted snapshot didn't cause any problem. I was waiting for 3 hours but memory stayed the same. Even checked with heaptrack -> all good
3. Warning message for analytical mode
4. Why do even corrupted snapshots occur?

- moves SetStorageMode to InMemoyStorage
- deletes corrupted snapshot files and deleted wal files
- fixes warning message for snapshots
- stops snapshot thread when switching from IN_MEMORY_TRANSACTIONAL to IN_MEMORY_ANALYTICAL
- starts snapshot thread when switching from IN_MEMORY_ANALYTICAL to IN_MEMORY_TRANSACTIONAL
- add precise recovery failure exception messages to track more easily corrupted snapshots

This PR doesn't solve the issue of Too many opened files. For that further work is needed but I think this PR is enough chunk of work to be merged.

[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR


src/storage/v2/durability/durability.cpp
```diff
@@ -30,6 +30,7 @@
 #include "storage/v2/durability/wal.hpp"
 #include "storage/v2/inmemory/label_index.hpp"
 #include "storage/v2/inmemory/label_property_index.hpp"
+#include "storage/v2/inmemory/storage.hpp"
```
Ignition, line 33: Either:
durability includes inmemory storage OR
inmemory storage includes durability
not both

src/storage/v2/inmemory/storage.cpp
```diff
@@ -1220,6 +1210,37 @@ Transaction InMemoryStorage::CreateTransaction(IsolationLevel isolation_level, S
   return {transaction_id, start_timestamp, isolation_level, storage_mode, false};
 }
 
+/// Main lock is taken by the caller.
+void InMemoryStorage::SetStorageMode(StorageMode storage_mode) {
+  std::unique_lock main_guard{main_lock_};
```
Ignition, line 1215: Comment and code disagree.

src/storage/v2/inmemory/storage.cpp
```diff
@@ -1220,6 +1210,37 @@ Transaction InMemoryStorage::CreateTransaction(IsolationLevel isolation_level, S
...
+  MG_ASSERT(
+      (storage_mode_ == StorageMode::IN_MEMORY_ANALYTICAL || storage_mode_ == StorageMode::IN_MEMORY_TRANSACTIONAL) &&
+      (storage_mode == StorageMode::IN_MEMORY_ANALYTICAL || storage_mode == StorageMode::IN_MEMORY_TRANSACTIONAL));
```
Ignition, line 1218: Maybe `storage_mode_` and `storage_mode` should be renamed to be more distinctive

src/storage/v2/inmemory/storage.cpp
```diff
@@ -1220,6 +1210,37 @@ Transaction InMemoryStorage::CreateTransaction(IsolationLevel isolation_level, S
...
+  if (storage_mode_ != storage_mode) {
+    if (storage_mode == StorageMode::IN_MEMORY_ANALYTICAL) {
+      snapshot_runner_.Stop();
+    } else {
+      snapshot_runner_.Run("Snapshot", config_.durability.snapshot_interval, [this] {
+        auto const &repl_state = repl_state_;
+        if (auto maybe_error = this->CreateSnapshot(repl_state, {true}); maybe_error.HasError()) {
```
Ignition, line 1225: Maybe an issue soon, repl_state will not exist inside storage, needs to be passed in to the `SetStorageMode` call

as51340: Shouldn't that then be fixed in PR which will introduce this change? Otherwise I don't know how to access repl_state in the callee of the `SetStorageMode`?

src/storage/v2/durability/durability.cpp
as51340: But I think I cannot forward declare `InMemoryStorage` since I need to access its methods and variables?

as51340: How should I do it?

Ignition: You was already passing the necessary data members (de-structured before the call). You changed it to use `InMemoryStorage *`...you didn't need to do that change.

src/storage/v2/inmemory/storage.cpp
Ignition: Two PR in flight at the same time....one of us will need to fix it :)

src/storage/v2/durability/durability.cpp
as51340: I opted for reverting changes here. The main reason is that I could couple only few arguments but not all hence whether I send 9 or 6 arguments is almost the same.

Ignition (approved):

as51340:
Changelog: Encountered corrupted snapshot and WAL files during recovery are now deleted. Snapshot thread is now stopped when switching from IN_MEMORY_TRANSACTIONAL to IN_MEMORY_ANALYTICAL. Before, the snapshot thread was running whole the time in ANALYTICAL mode although it wasn't doing anything.

PR #1387, opened 2023-10-20
Storage GC changes

Ignition:
- Make Bond move assignable
- Make benchmark use jemalloc to be a better indicator of production use
- Make parts of Transaction optional, so not to impact IN_MEMORY's GC
- Reduce contention on committed_transactions_
  * No need to do empty check + clear, just do clear
  * For some time now, the order of the committed_transactions_ has been random, hence no need to lock to attempt an insert order
  * Move all committed_transactions_ to local variable, rather than locking once per transaction being processed.
- Fix subtle GC bug, the commited_transaction unique_ptr that a delta was pointing to was deleted while delta was still potentially accessed
- garbage_undo_buffers_ processing can happen earlier
- As committed_transactions order is random, check all of them to see which can be processed
- Don't preallocate anything for the deltas monotonic pool

### DIFF
--- a/src/storage/v2/disk/storage.cpp
+++ b/src/storage/v2/disk/storage.cpp
@@ -271,7 +271,6 @@ DiskStorage::DiskAccessor::~DiskAccessor() {
     Abort();
   }
   FinalizeTransaction();
-  transaction_.deltas.~Bond<PmrListDelta>();
 }
 
 void DiskStorage::LoadPersistingMetadataInfo() {
@@ -462,12 +461,12 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(View view) {
         AllVerticesIterable(disk_storage->edge_import_mode_cache_->AccessToVertices(), storage_, &transaction_, view));
   }
   if (transaction_.scanned_all_vertices_) {
-    return VerticesIterable(AllVerticesIterable(transaction_.vertices_.access(), storage_, &transaction_, view));
+    return VerticesIterable(AllVerticesIterable(transaction_.vertices_->access(), storage_, &transaction_, view));
   }
 
   disk_storage->LoadVerticesToMainMemoryCache(&transaction_);
   transaction_.scanned_all_vertices_ = true;
-  return VerticesIterable(AllVerticesIterable(transaction_.vertices_.access(), storage_, &transaction_, view));
+  return VerticesIterable(AllVerticesIterable(transaction_.vertices_->access(), storage_, &transaction_, view));
 }
 
 VerticesIterable DiskStorage::DiskAccessor::Vertices(LabelId label, View view) {
@@ -586,7 +585,7 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(LabelId label, PropertyId p
 std::unordered_set<Gid> DiskStorage::MergeVerticesFromMainCacheWithLabelIndexCache(
     Transaction *transaction, LabelId label, View view, std::list<Delta> &index_deltas,
     utils::SkipList<Vertex> *indexed_vertices) {
-  auto main_cache_acc = transaction->vertices_.access();
+  auto main_cache_acc = transaction->vertices_->access();
   std::unordered_set<Gid> gids;
   gids.reserve(main_cache_acc.size());
 
@@ -638,7 +637,7 @@ void DiskStorage::LoadVerticesFromDiskLabelIndex(Transaction *transaction, Label
 std::unordered_set<Gid> DiskStorage::MergeVerticesFromMainCacheWithLabelPropertyIndexCache(
     Transaction *transaction, LabelId label, PropertyId property, View view, std::list<Delta> &index_deltas,
     utils::SkipList<Vertex> *indexed_vertices, const auto &label_property_filter) {
-  auto main_cache_acc = transaction->vertices_.access();
+  auto main_cache_acc = transaction->vertices_->access();
   std::unordered_set<storage::Gid> gids;
   gids.reserve(main_cache_acc.size());
 
@@ -721,7 +720,7 @@ std::unordered_set<Gid> DiskStorage::MergeVerticesFromMainCacheWithLabelProperty
     const std::optional<utils::Bound<PropertyValue>> &lower_bound,
     const std::optional<utils::Bound<PropertyValue>> &upper_bound, std::list<Delta> &index_deltas,
     utils::SkipList<Vertex> *indexed_vertices) {
-  auto main_cache_acc = transaction->vertices_.access();
+  auto main_cache_acc = transaction->vertices_->access();
   std::unordered_set<storage::Gid> gids;
   gids.reserve(main_cache_acc.size());
 
@@ -852,7 +851,7 @@ VertexAccessor DiskStorage::DiskAccessor::CreateVertex() {
   OOMExceptionEnabler oom_exception;
   auto *disk_storage = static_cast<DiskStorage *>(storage_);
   auto gid = disk_storage->vertex_id_.fetch_add(1, std::memory_order_acq_rel);
-  auto acc = transaction_.vertices_.access();
+  auto acc = transaction_.vertices_->access();
 
   auto *delta = CreateDeleteObjectDelta(&transaction_);
   auto [it, inserted] = acc.insert(Vertex{storage::Gid::FromUint(gid), delta});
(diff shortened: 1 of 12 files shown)


src/storage/v2/disk/storage.cpp
```diff
@@ -462,12 +461,12 @@ VerticesIterable DiskStorage::DiskAccessor::Vertices(View view) {
         AllVerticesIterable(disk_storage->edge_import_mode_cache_->AccessToVertices(), storage_, &transaction_, view));
   }
   if (transaction_.scanned_all_vertices_) {
-    return VerticesIterable(AllVerticesIterable(transaction_.vertices_.access(), storage_, &transaction_, view));
+    return VerticesIterable(AllVerticesIterable(transaction_.vertices_->access(), storage_, &transaction_, view));
```
antoniofilipovic, line 464: I would add MG_ASSERT which vertices contain value. If someone changes Transaction so that vertices don't contain a skip list for disk storage, we are all the way down in undefined behavior.

antoniofilipovic (commented):
All looks good, have just one comment

src/storage/v2/inmemory/storage.cpp
```diff
@@ -670,25 +670,19 @@ utils::BasicResult<StorageManipulationError, void> InMemoryStorage::InMemoryAcce
     // Replica can log only the write transaction received from Main
```
as51340, line 670: How are we sure that committed transactions are sorted by the commit timestamp in the WAL files

src/storage/v2/inmemory/storage.cpp
```diff
@@ -670,25 +670,19 @@ utils::BasicResult<StorageManipulationError, void> InMemoryStorage::InMemoryAcce
...
-        engine_guard.unlock();
-      });
+      could_replicate_all_sync_replicas =
+          mem_storage->AppendToWalDataDefinition(transaction_, *commit_timestamp_);  // protected by engine_guard
+      // TODO: release lock, and update all deltas to have a local copy of the commit timestamp
+      MG_ASSERT(transaction_.commit_timestamp != nullptr, "Invalid database state!");
+      transaction_.commit_timestamp->store(*commit_timestamp_, std::memory_order_release);  // protected by engine_guard
+      // Replica can only update the last commit timestamp with
+      // the commits received from main.
+      if (replState.IsMain() || desired_commit_timestamp.has_value()) {
```
as51340, line 680: Same check as in the outside if scope (line 672)

src/storage/v2/inmemory/storage.cpp
```diff
@@ -769,34 +765,31 @@ utils::BasicResult<StorageManipulationError, void> InMemoryStorage::InMemoryAcce
         // so the Wal files are consistent
```
as51340, line 765: For checking `if(!unique_constrant_violation)` prefer early return for removing one indentation

src/storage/v2/inmemory/storage.cpp
```diff
@@ -769,34 +765,31 @@ utils::BasicResult<StorageManipulationError, void> InMemoryStorage::InMemoryAcce
...
+        // TODO: release lock, and update all deltas to have a local copy of the commit timestamp
+        MG_ASSERT(transaction_.commit_timestamp != nullptr, "Invalid database state!");
+        transaction_.commit_timestamp->store(*commit_timestamp_,
+                                             std::memory_order_release);  // protected by engine_guard
+        // Replica can only update the last commit timestamp with
+        // the commits received from main.
+        if (replState.IsMain() || desired_commit_timestamp.has_value()) {
+          // Update the last commit timestamp
+          mem_storage->repl_storage_state_.last_commit_timestamp_.store(
+              *commit_timestamp_);  // protected by engine_guard
```
as51340, line 780: could we do together with AppendingToWalDataManipulation at line 767. I mean in the same if block.

src/storage/v2/inmemory/storage.cpp
```diff
@@ -984,8 +978,15 @@ void InMemoryStorage::InMemoryAccessor::FinalizeTransaction() {
   if (commit_timestamp_) {
```
as51340, line 978:
```suggestion
  if (!commit_timestamp_) { return; }
```

as51340 (changes requested):

src/utils/scheduler.hpp
```diff
@@ -58,17 +58,19 @@ class Scheduler {
...
-        auto now = std::chrono::system_clock::now();
         start_time += pause;
-        if (start_time > now) {
-          condition_variable_.wait_for(lk, start_time - now, [&] { return is_working_.load() == false; });
-        } else {
-          start_time = now;
+        {
+          std::unique_lock<std::mutex> lk(mutex_);
+          auto const stopped =
+              condition_variable_.wait_until(lk, start_time, [&] { return is_working_.load() == false; });
```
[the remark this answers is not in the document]
Ignition: `wait_until` will return true if predicate is true. `is_working_.load() == false` aka `stopped`

src/storage/v2/inmemory/storage.cpp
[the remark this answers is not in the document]
Ignition: This is a replication thing...not being changed in this PR

src/storage/v2/inmemory/storage.cpp
```diff
@@ -769,34 +765,31 @@ utils::BasicResult<StorageManipulationError, void> InMemoryStorage::InMemoryAcce
...
-          // and emplace back could take a long time.
-          engine_guard.unlock();
-        });
+        // TODO: release lock, and update all deltas to have a local copy of the commit timestamp
+        MG_ASSERT(transaction_.commit_timestamp != nullptr, "Invalid database state!");
+        transaction_.commit_timestamp->store(*commit_timestamp_,
+                                             std::memory_order_release);  // protected by engine_guard
+        // Replica can only update the last commit timestamp with
+        // the commits received from main.
+        if (replState.IsMain() || desired_commit_timestamp.has_value()) {
```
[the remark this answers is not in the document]
Ignition: This is a replication thing...not being changed in this PR

src/storage/v2/inmemory/storage.cpp
```diff
@@ -670,25 +670,19 @@ utils::BasicResult<StorageManipulationError, void> InMemoryStorage::InMemoryAcce
...
-          // Update the last commit timestamp
-          mem_storage->repl_storage_state_.last_commit_timestamp_.store(*commit_timestamp_);
-        }
-        // Release engine lock because we don't have to hold it anymore
-        // and emplace back could take a long time.
-        engine_guard.unlock();
-      });
+      could_replicate_all_sync_replicas =
+          mem_storage->AppendToWalDataDefinition(transaction_, *commit_timestamp_);  // protected by engine_guard
+      // TODO: release lock, and update all deltas to have a local copy of the commit timestamp
```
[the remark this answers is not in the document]
Ignition: Refactored from before, outside scope of this PR

src/storage/v2/inmemory/storage.cpp
```diff
@@ -670,25 +670,19 @@ utils::BasicResult<StorageManipulationError, void> InMemoryStorage::InMemoryAcce
...
+      MG_ASSERT(transaction_.commit_timestamp != nullptr, "Invalid database state!");
+      transaction_.commit_timestamp->store(*commit_timestamp_, std::memory_order_release);  // protected by engine_guard
+      // Replica can only update the last commit timestamp with
```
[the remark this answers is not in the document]
Ignition: Copy paste refactor, outside scope of this PR

src/storage/v2/inmemory/storage.cpp
Ignition: Outside scope of this PR, maybe change as part of replication work

src/storage/v2/inmemory/storage.cpp
Ignition: Outside the scope of PR. Don't want to change ATM, maybe a refactor for another PR.

src/storage/v2/inmemory/storage.cpp
Ignition: Was this way from before refactor, outside scope of PR

antoniofilipovic (approved):

PR #1393, opened 2023-10-23
Add cartesian mgbench

Josipmrden:
Internal benchmark for cartesian product.

[master < Task] PR

@vpavicic  internal improvement, no docs needed.


vpavicic:
Please add a "docs unnecessary" label so I have a clear overview what doesn't need to be written and what somebody forgot to classify! Thank you!

as51340 (approved):
Does grafana need to be redoployed?

Josipmrden:
> Does grafana need to be redoployed?

I would say no because we are just storing new identifier values to construct new graphs. It should be without downtime

antejavor (approved):
@Josipmrden Since we started adding workload, we should make sure @MarkoBarisic is aware of Postgress storage/size requirements. Not sure where all results are hosted, but just to keep an eye on that we are adding stuff to that instance.

PR #1394, opened 2023-10-23
Fix stress test tracker too early failure

antoniofilipovic:
Although the test never failed locally, on one PR it did fail already. https://github.com/memgraph/memgraph/actions/runs/6613091049/job/17960246226?pr=1384

Seems like it was too harsh to break the test after 3 tries. I gave it more attempts and a 10% diff between RES and the memory tracker immediately. This was the only test I was worried about if it would be flaky. 

[master < Task] PR


tests/stress/memory_tracker.py
```diff
@@ -260,10 +260,11 @@ def run_monitor_cleanup(repetition_count: int, sleep_sec: float) -> None:
...
             # is that memory tracker gets updated immediately
             # whereas RES takes some time
-            cnt_again = 3
+            # Tries 10 times or fails
+            cnt_again = 10
             skip_failure = False
-            # 10% is maximum increment, afterwards is fail
-            multiplier = 1
+            # 10% is maximum diff for this test to pass
+            multiplier = 1.10
```
Ignition, line 267: I do not like this fix, this test is fragile. I'd like a rational why a non fragile test is impossible, or convincing argument that this [fudge factor](https://en.wikipedia.org/wiki/Fudge_factor) is enough.

antoniofilipovic: So, the rationale for why the non-fragile test is impossible is when we have simultaneous calls to FREE MEMORY as well as creating data, memory usage from Jemalloc will drop instantaneously, whereas the resident size of memory takes time to drop to the same amount (I guess it takes Linux some time to drop memory usage to the same amount).

I don't like this test either, and it seems due to the interaction of our memory tracker with jemalloc where we instantly notice a change in memory allocation unless I do some global locks on the test (stop execution) after FREE MEMORY is executed and pause system for few hundreds of milliseconds, we will have short delay here.

This constant was chosen as the number above which we have an issue with the memory tracker. In experiments, it never got above 10% after a few times of calling.

Ignition (approved):

PR #1395, opened 2023-10-23
Handle analytical mode in community_detection during parallel changes

Darych:
NB: PR taken over by @antepusic; connected with https://github.com/memgraph/mage/pull/400

Since there are no ACID guarantees in analytical storage mode, the graph may get modified by parallel transactions while the user’s query is running. Query procedures and functions that return graph elements (nodes and relationships) should be robust to the deletion of graph elements, and this PR aims to fix that.

This PR contains changes to function evaluation and query module APIs that ensure the intended behavior:
* If a procedure would return a record that contains a deleted value in at least one field, that record should be skipped.
* If a function would return a deleted value, the actual returned value should be `null` instead.

[master < Task] PR
> Add handling of deleted return values for query procedures and functions ran in analytical mode


include/mg_procedure.h
```diff
@@ -587,6 +587,10 @@ enum mgp_error mgp_result_set_error_msg(struct mgp_result *res, const char *erro
 /// Return mgp_error::MGP_ERROR_UNABLE_TO_ALLOCATE if unable to allocate a mgp_result_record.
 enum mgp_error mgp_result_new_record(struct mgp_result *res, struct mgp_result_record **result);
 
+/// Delete last recode from result.
```
antepusic, line 590:
```suggestion
/// Delete last record from result.
```

antepusic (commented):
Everything is OK except for the one comment I left

src/query/interpret/eval.hpp
```diff
@@ -837,6 +838,7 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
 
   TypedValue Visit(Function &function) override {
     FunctionContext function_ctx{dba_, ctx_->memory, ctx_->timestamp, &ctx_->counters, view_};
+    bool is_transactional = MgpGraphIsTransactional(mgp_graph::NonWritableGraph(*dba_, view_));
```
antoniofilipovic, line 841: I think we should actually call dba_ to get us info whether we are in transactional state or not. So, idea is to have function which dba_ can call to return isTransactional, without calling mg procedure impl

src/query/interpret/eval.hpp
```diff
@@ -837,6 +838,7 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
 
   TypedValue Visit(Function &function) override {
```
antoniofilipovic, line 839: Do you know whether Visit will enter here only for mgp_func or something else also

src/query/interpret/eval.hpp
```diff
@@ -847,6 +849,9 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
       }
       auto res = function.function_(arguments, function.arguments_.size(), function_ctx);
       MG_ASSERT(res.GetMemoryResource() == ctx_->memory);
+      if (!is_transactional && res.ContainsDeleted()) {
```
antoniofilipovic, line 852:
```suggestion
      if (!is_transactional && res.ContainsDeleted()) [[unlikely]] {
```

src/query/interpret/eval.hpp
```diff
@@ -847,6 +849,9 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
...
+        return TypedValue(ctx_->memory);
```
antoniofilipovic, line 853: One general question, where was exception thrown before if object was deleted, where is this check?

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -319,6 +320,63 @@ mgp_value_type FromTypedValueType(memgraph::query::TypedValue::Type type) {
...
 }  // namespace
 
+bool MgpGraphIsTransactional(const mgp_graph &graph) noexcept {
+  auto storageMode = memgraph::storage::StorageMode::IN_MEMORY_TRANSACTIONAL;
+  std::visit(memgraph::utils::Overloaded{
+                 [&storageMode](memgraph::query::DbAccessor *dbAcc) { storageMode = dbAcc->GetStorageMode(); },
+                 [&storageMode](memgraph::query::SubgraphDbAccessor *dbAcc) { storageMode = dbAcc->GetStorageMode(); }},
+             graph.impl);
+  return storageMode != memgraph::storage::StorageMode::IN_MEMORY_ANALYTICAL;
+}
```
antoniofilipovic, line 330: Due to complex visit logic and complex logic here overall, I am more prone to solution where we move this check inside Storage. Also, one thing that popped on my mind now is that we can cache this info in mgp_graph during mgp_graph object creation as this will not change during procedure work, and we have benefit of not checking for every object if it was deleted or not every time

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -319,6 +320,63 @@ mgp_value_type FromTypedValueType(memgraph::query::TypedValue::Type type) {
...
+
+bool IsDeleted(mgp_vertex *vertex) { return vertex->getImpl().impl_.vertex_->deleted; }
```
antoniofilipovic, line 332: Did you think about case when you call procedure on trigger, and node was deleted but for_deleted_ which is used only for triggers is marked as true, which means that node can actually be used here, and you can work with it, so this will be some kind of edge case.

antoniofilipovic: So to expand imagine this:

STORAGE MODE IN_MEMORY_ANALYTICAL;

CREATE TRIGGER my_trigger on () DELETE CALL my_proc.count_deleted()

MATCH (n) DETACH DELETE n;

I think this will create scenario where node will be marked as deleted, but for_deleted_ flag inside VertexAccessor will be true, meaning we can work with it now inside that procedure

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -319,6 +320,63 @@ mgp_value_type FromTypedValueType(memgraph::query::TypedValue::Type type) {
...
+
+bool IsDeleted(mgp_edge *edge) { return edge->impl.impl_.edge_.ptr->deleted; }
+
+bool ContainsDeleted(mgp_path *path) {
+  for (auto &vertex : path->vertices) {
+    if (IsDeleted(&vertex)) return true;
+  }
```
antoniofilipovic, line 339: maybe we can use here std::any_of?

antoniofilipovic: Also, const auto

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -319,6 +320,63 @@ mgp_value_type FromTypedValueType(memgraph::query::TypedValue::Type type) {
...
+      return IsDeleted(val->vertex_v);
+    case MGP_VALUE_TYPE_EDGE:
+      return IsDeleted(val->edge_v);
+    case MGP_VALUE_TYPE_PATH:
+      return ContainsDeleted(val->path_v);
+    case MGP_VALUE_TYPE_LIST:
+      return ContainsDeleted(val->list_v);
+    case MGP_VALUE_TYPE_MAP:
+      return ContainsDeleted(val->map_v);
+    // other types are primitives
```
antoniofilipovic, line 372: In case we add a new type which is not primitive, we can add here all types and exception will be thrown on compiler that case is not covered

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -1001,6 +1059,11 @@ mgp_error mgp_list_copy(mgp_list *list, mgp_memory *memory, mgp_list **result) {
 
 void mgp_list_destroy(mgp_list *list) { DeleteRawMgpObject(list); }
 
+mgp_error mgp_list_contains_deleted(mgp_list *list, int *result) {
+  *result = ContainsDeleted(list);
+  return mgp_error::MGP_ERROR_NO_ERROR;
```
antoniofilipovic, line 1064: Why not use WrapExceptions?

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -1052,6 +1115,11 @@ mgp_error mgp_map_copy(mgp_map *map, mgp_memory *memory, mgp_map **result) {
 
 void mgp_map_destroy(mgp_map *map) { DeleteRawMgpObject(map); }
 
+mgp_error mgp_map_contains_deleted(mgp_map *map, int *result) {
+  *result = ContainsDeleted(map);
```
antoniofilipovic, line 1119: I think we should use WrapExceptions, don't know what might happen here

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -1574,10 +1647,12 @@ mgp_error mgp_result_record_insert(mgp_result_record *record, const char *field_
     if (find_it == record->signature->end()) {
       throw std::out_of_range{fmt::format("The result doesn't have any field named '{}'.", field_name)};
     }
+
     const auto *type = find_it->second.first;
+
```
antoniofilipovic, line 1652: You can remove this empty lines here

src/query/procedure/py_module.cpp
```diff
@@ -905,30 +918,66 @@ std::optional<py::ExceptionInfo> AddRecordFromPython(mgp_result *result, py::Obj
...
-    if (mgp_result_record_insert(record, field_name, field_val) != mgp_error::MGP_ERROR_NO_ERROR) {
+
+    if (!is_transactional) {
+      if (ContainsDeleted(field_val)) {
+        mgp_value_destroy(field_val);
+        for (auto &cache_entry : current_record_cache) {
+          mgp_value_destroy(cache_entry.field_val);
+        }
+
+        return std::nullopt;
```
antoniofilipovic, line 938: I would make a comment here that if one is deleted, we are removing allo of them from record, because whole record gets scraped

src/query/procedure/py_module.cpp
```diff
@@ -905,30 +918,66 @@ std::optional<py::ExceptionInfo> AddRecordFromPython(mgp_result *result, py::Obj
       PyErr_SetString(PyExc_TypeError, msg.c_str());
       return py::FetchError();
```
antoniofilipovic, line 919: If we return here py::FetchError whole std::vector will be still cached and we have a leak. I think we need to add here onScopeExit

antoniofilipovic: For std::vector

src/query/procedure/py_module.cpp
```diff
@@ -905,30 +918,66 @@ std::optional<py::ExceptionInfo> AddRecordFromPython(mgp_result *result, py::Obj
...
+      ss << "Unable to insert field '" << py::Object::FromBorrow(cache_entry.key) << "' with value: '"
+         << py::Object::FromBorrow(cache_entry.val) << "'; did you set the correct field type?";
       const auto &msg = ss.str();
       PyErr_SetString(PyExc_ValueError, msg.c_str());
-      mgp_value_destroy(field_val);
+      mgp_value_destroy(cache_entry.field_val);
       return py::FetchError();
     }
-    mgp_value_destroy(field_val);
+    mgp_value_destroy(cache_entry.field_val);
```
antoniofilipovic, line 974: This is repeated code, I would isolate this in lambda

antoniofilipovic: If I am not wrong, we can do that

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -1574,10 +1647,12 @@ mgp_error mgp_result_record_insert(mgp_result_record *record, const char *field_
     if (find_it == record->signature->end()) {
       throw std::out_of_range{fmt::format("The result doesn't have any field named '{}'.", field_name)};
     }
+
```
antoniofilipovic, line 1650: Where is logic for this insertion of values if not deleted for regular procedure in c and c++?

src/query/procedure/py_module.cpp
```diff
@@ -1164,9 +1213,26 @@ void CallPythonFunction(const py::Object &py_cb, mgp_list *args, mgp_graph *grap
   auto call = [&](py::Object py_graph) -> utils::BasicResult<std::optional<py::ExceptionInfo>, mgp_value *> {
     py::Object py_args(MgpListToPyTuple(args, py_graph.Ptr()));
     if (!py_args) return {py::FetchError()};
+    auto is_transactional = MgpGraphIsTransactional(*graph);
     auto py_res = py_cb.Call(py_graph, py_args);
     if (!py_res) return {py::FetchError()};
     mgp_value *ret_val = PyObjectToMgpValueWithPythonExceptions(py_res.Ptr(), memory);
+    if (!is_transactional && ContainsDeleted(ret_val)) {
+      mgp_value *null_val{nullptr};
```
antoniofilipovic, line 1221: Don't we need to delete here ret_val?

antoniofilipovic (changes requested):
I need to take a look into tests and MAGE PR still

src/query/typed_value.cpp
```diff
@@ -362,6 +362,33 @@ bool TypedValue::IsGraph() const { return type_ == Type::Graph; }
...
+      return false;
+    case Type::Map:
+      for (const auto &[_, map_value] : map_v) {
+        if (map_value.ContainsDeleted()) return true;
+      }
+      return false;
+    case Type::Vertex:
+      return vertex_v.impl_.vertex_->deleted;
+    case Type::Edge:
+      return edge_v.impl_.edge_.ptr->deleted;
```
Darych, line 380: Should we here handle a case when Edge doesn't have properties? In this case ptr should be invalid.

src/query/typed_value.cpp
```diff
@@ -362,6 +362,33 @@ bool TypedValue::IsGraph() const { return type_ == Type::Graph; }
...
+    case Type::Path:
+      for (const auto &vertex_acc : path_v.vertices()) {
+        if (vertex_acc.impl_.vertex_->deleted) return true;
+      }
+      std::ranges::any_of(path_v.edges(), [](auto &edge_acc) { return edge_acc.impl_.edge_.ptr->deleted; });
```
Darych, line 385: I think we should add `return` before `std::ranges::any_of...`

src/query/typed_value.cpp
```diff
@@ -362,6 +362,33 @@ bool TypedValue::IsGraph() const { return type_ == Type::Graph; }
...
+      for (const auto &[_, map_value] : map_v) {
+        if (map_value.ContainsDeleted()) return true;
+      }
+      return false;
+    case Type::Vertex:
+      return vertex_v.impl_.vertex_->deleted;
+    case Type::Edge:
+      return edge_v.impl_.edge_.ptr->deleted;
+    case Type::Path:
+      for (const auto &vertex_acc : path_v.vertices()) {
```
Darych, line 382: Can we rewrite this using `any_of` to be consistent?

tests/e2e/query_modules_storage_modes/test_query_modules_storage_modes.py
```diff
@@ -0,0 +1,281 @@
...
+        CALL {api}_api.pass_node_with_id(n)
+        YIELD node, id
+        RETURN node, id;"""
+    )
+
+    deleter.join()
+
+    result = cursor.fetchall()
+
+    assert len(result) == 1 and len(result[0]) == 2 and result[0][0].properties["id"] == "7X8X0"
```
Darych, line 91: Why do we have `len(result[0]) == 2` and check just the first one - `result[0][0]`. What about `result[0][1]`?

tests/e2e/query_modules_storage_modes/test_query_modules_storage_modes.py
```diff
@@ -0,0 +1,281 @@
...
+        """MATCH (n)-[e]->()
+        WITH collect(n) + collect(e) as list
+        RETURN python_api.pass_list(list);"""
+    )
+
+    deleter.join()
+
+    result = cursor.fetchall()
+
+    assert len(result) == 1 and len(result[0]) == 1 and result[0][0] is None
```
Darych, line 217: Is deleted edge always goes first - `result[0][0]`? Based on code, I would expect to have nodes at the beginning of list

src/query/db_accessor.hpp
```diff
@@ -694,6 +694,8 @@ class SubgraphDbAccessor final {
   std::optional<VertexAccessor> FindVertex(storage::Gid gid, storage::View view);
 
   Graph *getGraph();
+
+  storage::StorageMode GetStorageMode() const noexcept;
```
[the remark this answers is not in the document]
antepusic: I’ll make it consistent, thanks!

src/query/interpret/eval.hpp
antepusic: Moving the implementaton to the accessor also makes sense to me :+1:

src/query/interpret/eval.hpp
antepusic: I think it was thrown by the client

src/query/procedure/mg_procedure_impl.cpp
antepusic: The sole purpose of the deletion checks is to avoid returning deleted objects as results. Triggers operating on deleted objects can still use them, as long as they don’t return them (this is identical to the present behavior).

antepusic: I’ll add a test case for this, thanks for the heads-up!

src/query/procedure/mg_procedure_impl.cpp
antepusic: This is good from the maintainability perspective, adding it

src/query/procedure/mg_procedure_impl.cpp
antepusic: I didn’t envision any case where this function would throw an exception, but I can add it no problem

src/query/procedure/mg_procedure_impl.cpp
antepusic: I didn’t envision any case where this function would throw an exception, but I can add it no problem

src/query/procedure/mg_procedure_impl.cpp
antepusic: The deletion check cannot be done inside `mgp_result_record_insert` because, upon trying to insert a deleted value, the entire record should be skipped. This means that people writing query modules using the C and C++ APIs need to do a check manually.

src/query/interpret/eval.hpp
antepusic: Built-in functions too

tests/e2e/query_modules_storage_modes/test_query_modules_storage_modes.py
antepusic: The second field (`result[0][1]`) is the internal ID of the node that’s passed. I don’t evaluate its value because internal node IDs can vary (e.g. if we add more tests that execute before this one, the ID will be higher).

tests/e2e/query_modules_storage_modes/test_query_modules_storage_modes.py
antepusic: Should a function return a deleted value, or an object containing a deleted value, it would return `None` instead. A list comprised of [node, deleted_edge] falls under the latter case.

src/query/db_accessor.hpp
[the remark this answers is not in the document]
antepusic: Thanks for the suggestion! In the end, I’ve opted not to do it because of consistency:
* all `DbAccessor` method implementations are in the .hpp file already
* all `SubgraphDbAccessor` method implementations are in the .cpp file already

src/query/procedure/mg_procedure_impl.cpp
antepusic: Done

src/query/procedure/mg_procedure_impl.cpp
antepusic: Done

src/query/procedure/py_module.cpp
antepusic: Thanks for the suggestion! I’ve isolated this in a function of its own (in order to make the parent function simpler)

src/query/procedure/py_module.cpp
antepusic: Thanks for the catch!

src/query/typed_value.cpp
antepusic: Added. In solving this, my original idea was to add a `deleted` property to `EdgeRef`, but it wasn’t possible to update that property in storage.cpp because an `EdgeRef` object isn’t unique to its edge.

Because of this, there’s no way to check whether an `EdgeRef` refers to an existing edge if `ptr` is invalid (short of major changes to storage).

src/query/interpret/eval.hpp
```diff
@@ -847,6 +850,9 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
       }
       auto res = function.function_(arguments, function.arguments_.size(), function_ctx);
       MG_ASSERT(res.GetMemoryResource() == ctx_->memory);
+      if (!is_transactional && res.ContainsDeleted()) [[unlikely]] {
+        return TypedValue(ctx_->memory);
+      }
```
antoniofilipovic, line 855: Can we refactor bit this Visit function, I think only arguments allocation and processing (for loop) needs to be in separate if-else blocks, otherwise we can isolate function call, mg_assert and if check for deleted to be of common execution  part

src/query/plan/operator.cpp
```diff
@@ -4861,6 +4868,9 @@ class CallProcedureCursor : public Cursor {
         throw QueryRuntimeException("{}: {}", self_->procedure_name_, *result_->error_msg);
       }
       result_row_it_ = result_->rows.begin();
+      if (!result_->is_transactional) {
+        skip_rows_with_deleted_values();
+      }
```
antoniofilipovic, line 4873: Just to double check here, we are not stuck here in infinite loop if everything in result->rows needs to be skipped, right? In next iter it will try to pull next record, and if that is empty it will return false, otherwise enter procedure execution again

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -1632,7 +1624,8 @@ mgp_error mgp_result_new_record(mgp_result *res, mgp_result_record **result) {
         MG_ASSERT(res->signature, "Expected to have a valid signature");
         res->rows.push_back(mgp_result_record{
             res->signature,
-            memgraph::utils::pmr::map<memgraph::utils::pmr::string, memgraph::query::TypedValue>(memory)});
+            memgraph::utils::pmr::map<memgraph::utils::pmr::string, memgraph::query::TypedValue>(memory),
+            !res->is_transactional});
```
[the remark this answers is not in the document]
antoniofilipovic: Field name is `ignore_deleted_values`.

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -1574,10 +1640,15 @@ mgp_error mgp_result_record_insert(mgp_result_record *record, const char *field_
     if (find_it == record->signature->end()) {
       throw std::out_of_range{fmt::format("The result doesn't have any field named '{}'.", field_name)};
     }
+    if (record->ignore_deleted_values && ContainsDeleted(val)) [[unlikely]] {
+      record->values.emplace(field_name, ToTypedValue(*NewRawMgpObject<mgp_value>(memory), memory));
```
antoniofilipovic, line 1644: Do you need to emplace anything here if it has deleted values?

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -1960,6 +2031,11 @@ mgp_error mgp_vertex_copy(mgp_vertex *v, mgp_memory *memory, mgp_vertex **result
 
 void mgp_vertex_destroy(mgp_vertex *v) { DeleteRawMgpObject(v); }
 
+mgp_error mgp_vertex_is_deleted(mgp_vertex *v, int *result) {
+  *result = IsDeleted(v);
```
antoniofilipovic, line 2035: It can happen that we have some bug here, and we are not aware of it, and something may throw an exception, so I would use here wrap exceptions function in any case

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -2317,6 +2393,11 @@ mgp_error mgp_edge_copy(mgp_edge *e, mgp_memory *memory, mgp_edge **result) {
 
 void mgp_edge_destroy(mgp_edge *e) { DeleteRawMgpObject(e); }
 
+mgp_error mgp_edge_is_deleted(mgp_edge *e, int *result) {
+  *result = IsDeleted(e);
```
antoniofilipovic, line 2397: Same comment about wrap exceptions here as in other parts

src/query/procedure/py_module.cpp
```diff
@@ -888,15 +919,27 @@ std::optional<py::ExceptionInfo> AddRecordFromPython(mgp_result *result, py::Obj
   py::Object items(PyDict_Items(fields.Ptr()));
   if (!items) return py::FetchError();
   mgp_result_record *record{nullptr};
-  if (RaiseExceptionFromErrorCode(mgp_result_new_record(result, &record))) {
-    return py::FetchError();
+  auto is_transactional = storage::IsTransactional(graph->storage_mode);
```
antoniofilipovic, line 922: you can use here `const auto`

src/query/db_accessor.cpp
```diff
@@ -139,6 +140,8 @@ std::optional<VertexAccessor> SubgraphDbAccessor::FindVertex(storage::Gid gid, s
 
 query::Graph *SubgraphDbAccessor::getGraph() { return graph_; }
 
+storage::StorageMode SubgraphDbAccessor::GetStorageMode() const noexcept { return db_accessor_.GetStorageMode(); }
```
antoniofilipovic, line 143: If you change this return type to  `auto`, then I think you don't need above include for storage mode

src/query/typed_value.cpp
```diff
@@ -364,27 +364,35 @@ bool TypedValue::IsGraph() const { return type_ == Type::Graph; }
...
     case Type::Map:
-      for (const auto &[_, map_value] : map_v) {
-        if (map_value.ContainsDeleted()) return true;
-      }
-      return false;
+      return std::ranges::any_of(map_v, [](const auto &item) { return item.second.ContainsDeleted(); });
     case Type::Vertex:
       return vertex_v.impl_.vertex_->deleted;
     case Type::Edge:
+      if (!edge_v.impl_.storage_->config_.items.properties_on_edges) {
```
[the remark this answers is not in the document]
antoniofilipovic: I agree

tests/e2e/query_modules_storage_modes/query_modules/c_api.cpp
```diff
@@ -0,0 +1,62 @@
...
+void PassRelationship(mgp_list *args, mgp_func_context *ctx, mgp_func_result *res, mgp_memory *memory) {
+  auto *relationship = mgp::list_at(args, 0);
+
+  std::this_thread::sleep_for(std::chrono::seconds(1));
+
+  mgp::func_result_set_value(res, relationship, memory);
+}
+
+void PassNodeWithId(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory) {
+  auto *node = mgp::value_get_vertex(mgp::list_at(args, 0));
```
antoniofilipovic, line 36: const auto here?

tests/e2e/query_modules_storage_modes/query_modules/c_api.cpp
```diff
@@ -0,0 +1,62 @@
...
+  auto node_id = mgp::vertex_get_id(node).as_int;
+
+  std::this_thread::sleep_for(std::chrono::seconds(1));
```
antoniofilipovic, line 39: why sleep here?

antoniofilipovic (changes requested):
Looking good, few last changes

tests/e2e/query_modules_storage_modes/query_modules/c_api.cpp
antepusic: The query procedures & functions used in testing this API all have a short sleep period, during which the other parallel transaction deletes the graph element that’s the procedure/function’s return value.

src/query/typed_value.cpp
```diff
@@ -364,27 +364,35 @@ bool TypedValue::IsGraph() const { return type_ == Type::Graph; }
...
+      }
       return edge_v.impl_.edge_.ptr->deleted;
     case Type::Path:
-      for (const auto &vertex_acc : path_v.vertices()) {
-        if (vertex_acc.impl_.vertex_->deleted) return true;
-      }
-      std::ranges::any_of(path_v.edges(), [](auto &edge_acc) { return edge_acc.impl_.edge_.ptr->deleted; });
+      return std::ranges::any_of(path_v.vertices(),
+                                 [](auto &vertex_acc) { return vertex_acc.impl_.vertex_->deleted; }) ||
+             std::ranges::any_of(path_v.edges(), [](auto &edge_acc) { return edge_acc.impl_.edge_.ptr->deleted; });
```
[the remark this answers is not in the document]
antepusic: Thanks for the catch!

src/query/plan/operator.cpp
antepusic: While I don’t think that this would result in an infinite loop, you made me check and I’ve noticed that, if `skip_rows_with_deleted_values` skips **all** rows, `Pull` should return `false`. I’ll add the handling of this case

tests/e2e/query_modules_storage_modes/query_modules/c_api.cpp
antoniofilipovic: Is this sleep period enough, did you do any empirical measurements?

src/query/procedure/py_module.cpp
antepusic: Done

src/query/procedure/py_module.cpp
antepusic: Added

tests/gql_behave/continuous_integration
```diff
@@ -104,7 +104,7 @@ class MemgraphRunner:
         memgraph_binary = os.path.join(self.build_directory, "memgraph")
         args_mg = [
             memgraph_binary,
-            "--storage-properties-on-edges",
+            "--storage-properties-on-edges=true",
```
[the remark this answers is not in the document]
antepusic: When the flag wasn’t set, the GQLBehave tests would fail with messages that this flag was turned off (and the default config value seems to be false: [link](https://github.com/memgraph/memgraph/blob/master/src/flags/general.cpp#L72)). I’ve encountered this both locally and on here.

tests/e2e/query_modules_storage_modes/query_modules/c_api.cpp
antepusic: In the next line, this node is passed to `mgp::vertex_get_id`, which takes only non-const `mgp_vertex *`.

src/query/procedure/py_module.cpp
antepusic: Added (for both cases).

src/query/db_accessor.cpp
antepusic: Thanks, done!

antepusic: While I like this suggestion, changing return type to `auto` causes these errors across mg_procedure_impl.hpp:
```
error: function 'GetStorageMode' with deduced return type cannot be used before it is defined
```
Given that, I think it’s more prudent to keep this as-is and avoid rearranging other #includes.

src/query/procedure/mg_procedure_impl.cpp
antepusic: Added

src/query/procedure/mg_procedure_impl.cpp
antepusic: Thanks, added

src/query/procedure/mg_procedure_impl.cpp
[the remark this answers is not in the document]
antepusic: That’s right @antoniofilipovic! I’ve switched to a designated initialized so that the field names are visible here.

tests/e2e/query_modules_storage_modes/query_modules/c_api.cpp
antepusic: Added a few clarifying comments

src/query/interpret/eval.hpp
antepusic: Done (with the caveat that the function call remains inside the if-else blocks)

src/query/procedure/mg_procedure_impl.cpp
antepusic: Good catch, emplacing isn’t necessary in case of a deleted value because the record will get skipped anyway

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -1054,6 +1106,10 @@ mgp_error mgp_map_copy(mgp_map *map, mgp_memory *memory, mgp_map **result) {
 
 void mgp_map_destroy(mgp_map *map) { DeleteRawMgpObject(map); }
 
+mgp_error mgp_map_contains_deleted(mgp_map *map, int *result) {
+  return WrapExceptions([map, result] { *result = ContainsDeleted(map); });
```
antoniofilipovic, line 1110:
```suggestion
  return WrapExceptions([map, result] { return ContainsDeleted(map); });
```
I think you can just do that since we have WrapExceptions which writes to `*result` object

antepusic: Didn’t work, reverting suggestion

src/query/typed_value.cpp
antepusic: Done

src/query/typed_value.cpp
antepusic: Done

Darych (commented):
LGTM. I can't approve the PR since I'm author. But no more comments.
Good job, @antepusic !

src/query/plan/operator.cpp
antepusic: How it works if everything in `result->rows` needs to be skipped:
1. mark `stream_exhausted` as true
2. try to pull another result
3. If there are no results left to be pulled, the program exits the loop. If there are more results to be pulled, the program stays in the loop to run the procedure on them, then back to 1.

Since the input cursor will be exhausted at some point and return false upon `Pull`, no infinite loop should result.

antepusic:
@vpavicic Release note in the original comment (docs PR: https://github.com/memgraph/documentation/pull/361)

PR #1397, opened 2023-10-24
Fix failing e2e tests in relwithdebinfo mode

gvolfing:
In and only in RelWithDebInfo mode the access of the maybe_unused variable results in segfaults, this change is making sure that does no happen ever if the maybe_unused variable is nullopt without changing the overall logic.


gitbuda:
@gvolfing added this to v2.13

So, this is interesting, obviously RelWithDebInfo is not the same as Release, I guess we should level the Diff Release workflow in the Release mode CC @Darych 🤔

gitbuda (approved):

PR #1401, opened 2023-10-25
Add OOM enabler for MG procedure

antoniofilipovic:
OOM enabler needs to be introduced also in mg_procedure_impl and that is what this diff will do. Currently, if a user specifies multiple threads that will do allocations from the procedure, due to the thread local variable for the OOM enabler to check if the error can be thrown, call in procedure will pass (although it shouldn't if memory is too big) and error will be not thrown.

[master < Task] PR


src/query/plan/operator.cpp
```diff
@@ -4657,23 +4655,12 @@ void CallCustomProcedure(const std::string_view fully_qualified_procedure_name,
   if (memory_limit) {
     SPDLOG_INFO("Running '{}' with memory limit of {}", fully_qualified_procedure_name,
                 utils::GetReadableSize(*memory_limit));
-    utils::LimitedMemoryResource limited_mem(memory, *memory_limit);
-    mgp_memory proc_memory{&limited_mem};
```
antoniofilipovic, line 4661: Since we are tracking memory via extent hooks, I removed that part on the limited memory resource

src/query/plan/operator.cpp
```diff
@@ -4657,23 +4655,12 @@ void CallCustomProcedure(const std::string_view fully_qualified_procedure_name,
...
-    MG_ASSERT(result->signature == &proc.results);
-    // TODO: What about cross library boundary exceptions? OMG C++?!
-    proc.cb(&proc_args, &graph, result, &proc_memory);
-    size_t leaked_bytes = limited_mem.GetAllocatedBytes();
-    if (leaked_bytes > 0U) {
-      spdlog::warn("Query procedure '{}' leaked {} *tracked* bytes", fully_qualified_procedure_name, leaked_bytes);
-    }
```
antoniofilipovic, line 4668: This will be added in different PR as part of memory procedure tracking, this one here: https://github.com/memgraph/memgraph/pull/1443

antoniofilipovic:
This PR needs to be merged first: https://github.com/memgraph/memgraph/pull/1443

tests/e2e/memory/procedures/global_memory_limit_multi_thread_create_proc.cpp
```diff
@@ -0,0 +1,96 @@
...
+  for (int i = 0; i < num_vertices_per_thread; i++) {
+    struct mgp_vertex *vertex{nullptr};
+    auto enum_error = mgp_graph_create_vertex(graph, memory, &vertex);
+    if (enum_error == mgp_error::MGP_ERROR_UNABLE_TO_ALLOCATE) {
+      break;
+    }
+    if (enum_error != mgp_error::MGP_ERROR_NO_ERROR) {
+      [[maybe_unused]] const enum mgp_error untracking_error = mgp_untrack_current_thread_allocations(graph);
+      return;
+    }
```
as51340, line 44:
```suggestion
    if (enum_error == mgp_error::MGP_ERROR_UNABLE_TO_ALLOCATE || enum_error != mgp_error::MGP_ERROR_NO_ERROR) {
      break;
    }
```

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -158,6 +159,7 @@ template <typename TFunc, typename... Args>
 [[nodiscard]] mgp_error WrapExceptions(TFunc &&func, Args &&...args) noexcept {
   static_assert(sizeof...(args) <= 1, "WrapExceptions should have only one or zero parameter!");
   try {
+    memgraph::utils::MemoryTracker::OutOfMemoryExceptionEnabler oom_enabler;
```
as51340, line 162: I suppose you've put it here because every method uses this as a wrapper. Is there something which doesn't use WrapExceptions?

as51340 (approved):

src/query/procedure/mg_procedure_impl.cpp
antoniofilipovic: Yes, that is true. All methods should use WrapExceptions, I will double check

PR #1408, opened 2023-10-25
Add notification for user on max map count

antoniofilipovic:
This diff should enable notification to the user on the max map count used by the system.

[master < Task] PR

### DIFF
--- a/src/utils/stat.hpp
+++ b/src/utils/stat.hpp
@@ -54,4 +54,17 @@ inline uint64_t GetMemoryUsage() {
   return memory;
 }
 
+/// Returns the size of vm.max_map_count
+inline uint64_t GetVmMaxMapCount() {
+  uint64_t max_map_count = 0;
+  auto statm_data = utils::ReadLines("sysctl vm.max_map_count");
+  if (!statm_data.empty()) {
+    auto split = utils::Split(statm_data[0]);
+    if (split.size() == 2) {
+      max_map_count = std::stoull(split[2]);
+    }
+  }
+  return max_map_count;
+}
+
 }  // namespace memgraph::utils


src/memgraph.cpp
```diff
@@ -204,6 +206,17 @@ int main(int argc, char **argv) {
...
   std::cout << "To get started with Memgraph, visit https://memgr.ph/start" << std::endl;
 
+  auto vm_max_map_count = memgraph::utils::GetVmMaxMapCount();
+  if (vm_max_map_count.has_value()) {
+    if (vm_max_map_count.value() < kMgVmMaxMapCount) {
+      std::cout << "Max virtual memory areas vm.max_map_count " << vm_max_map_count.value()
+                << " is too low, increase to at least " << kMgVmMaxMapCount << std::endl;
+    }
+  } else {
+    std::cout << "Can't get info on vm.max_map_count, check that is too low, vm.max_map_count is at least "
```
andrejtonev, line 216:
```suggestion
    std::cout << "Can't get info on vm.max_map_count, check whether it is too low, vm.max_map_count is at least "
```

src/telemetry/collectors.cpp
```diff
@@ -79,7 +79,12 @@ const nlohmann::json GetResourceUsage(std::filesystem::path root_directory) {
   auto cpu_total = GetCpuUsage(pid);
   cpu["usage"] = cpu_total.second;
 
-  return {{"cpu", cpu}, {"memory", utils::GetMemoryUsage()}, {"disk", utils::GetDirDiskUsage(root_directory)}};
+  nlohmann::json json;
+  json.emplace_back("cpu", cpu);
+  json.emplace_back("memory", utils::GetMemoryUsage());
+  json.emplace_back("disk", utils::GetDirDiskUsage(root_directory));
+  json.emplace_back("vm_max_map_count", utils::GetVmMaxMapCount());
```
andrejtonev, line 86: What gets written in case of a nullopt?

andrejtonev (commented):
Looks good

src/query/interpreter.cpp
```diff
@@ -3116,13 +3117,16 @@ PreparedQuery PrepareSystemInfoQuery(ParsedQuery parsed_query, bool in_explicit_
       handler = [storage = current_db.db_acc_->get()->storage(), interpreter_isolation_level,
                  next_transaction_isolation_level] {
         auto info = storage->GetBaseInfo();
+        const auto vm_max_map_count = utils::GetVmMaxMapCount();
```
andrejtonev, line 3120: Maybe it makes sense just to pass back an integer instead of the optional<int>, since everywhere where we use it we check if it has a value and set it to `-1` in case there is no value.

antoniofilipovic: I would leave it like this because in memgraph.cpp I am getting value, and I have different logic for communicating to user that I couldn't read `vm_max_map_count`. If I return -1, then instead of checking if it has value, I need to check for -1, which is special number in this case, and could be error prone, if someone later decides to return 0 as special number.

src/utils/stat.hpp
```diff
@@ -54,4 +58,20 @@ inline uint64_t GetMemoryUsage() {
   return memory;
 }
 
+/// Returns the size of vm.max_map_count
+inline std::optional<uint64_t> GetVmMaxMapCount() {
```
andrejtonev, line 62: I don't like mixing of the uin64_t and int64_t.
We are returning here the unsigned and then casting it to signed and also mixing with the default value which is a signed as well.
We should just return a signed value here in my opinion.

andrejtonev (commented):
Last comment a promise :D

src/utils/stat.hpp
antoniofilipovic: Fixed it

tests/e2e/configuration/storage_info.py
```diff
@@ -54,13 +54,12 @@ def test_does_default_config_match():
 
     # The default value of these is dependent on the given machine.
     machine_dependent_configurations = [
-        "memory_usage",
+        "memory_res",
```
andrejtonev, line 57: Did we sync with the server side about this change?

tests/e2e/configuration/storage_info.py
```diff
@@ -54,13 +54,12 @@ def test_does_default_config_match():
...
         "disk_usage",
         "memory_allocated",
         "allocation_limit",
-        "vm_max_map_count",
+        "",
```
andrejtonev, line 61: Why has this disappeared?

src/telemetry/collectors.cpp
```diff
@@ -80,7 +80,7 @@ const nlohmann::json GetResourceUsage(std::filesystem::path root_directory) {
   cpu["usage"] = cpu_total.second;
   const auto vm_max_map_count = utils::GetVmMaxMapCount();
   return {{"cpu", cpu},
-          {"memory", utils::GetMemoryUsage()},
+          {"memory", utils::GetMemoryRES()},
```
andrejtonev, line 83: When did this happen?

antoniofilipovic: Here: https://github.com/memgraph/memgraph/pull/1426

tests/e2e/configuration/storage_info.py
antoniofilipovic: This is only for our side, on server side we didn't change anything: https://github.com/memgraph/memgraph/pull/1426

tests/e2e/configuration/storage_info.py
antoniofilipovic: Don't know really, this one snuck into PR, will revert

andrejtonev (approved):
:muscle:

antoniofilipovic:
@vpavicic RN: From now on via show storage info query user can get info on vm.max_map_count reported by system kernel

PR #1418, opened 2023-10-26
Fix release workflows

MarkoBarisic:
Fix the default value BUILD_TYPE used in release workflows


.github/workflows/release_centos8.yaml
```diff
@@ -40,9 +40,13 @@ jobs:
           # Initialize dependencies.
           ./init
 
+          # Set default build_type to Release
+          INPUT_BUILD_TYPE=${{ github.event.inputs.build_type }}
+          BUILD_TYPE=${INPUT_BUILD_TYPE:-"Release"}
```
Darych, line 45: Default build type is set in inputs. What is the purpose of checking again here?

MarkoBarisic: The input variable in gh actions has a caveat, the deafult only works when using wokrflow_dispatch, in other words when running the workflow manually. When the workflow is run automatically the entire workflow_dispatch section is skipped and no variable is infered, not even the default

MarkoBarisic: So basically this check is redundant only when running manually, but is crucial when running automatically.

Ignition (commented):
A big question is why ```
on:
  workflow_dispatch:
    inputs:
      build_type:
        type: choice
        description: "Memgraph Build type. Default value is Release."
        default: 'Release'
        options:
          - Release
          - RelWithDebInfo
```
is not working as expected

File: .github/workflows/release_centos8.yaml (hunk 1 above)
Comment (Darych) [reply .github/workflows/release_centos8.yaml:45]: Thank you for explanation! 🙂

**Review by Darych — APPROVED** [review]

**Comment by MarkoBarisic** [comment]
Regarding this code section
```
on:
  workflow_dispatch:
    inputs:
      build_type:
        type: choice
        description: "Memgraph Build type. Default value is Release."
        default: 'Release'
        options:
          - Release
          - RelWithDebInfo
```
The top part of the section `workflow_dispatch:` means that this workflow can be run manually and when that is the case apply logic from the rest of this section.
The thing with github actions is that when the workflow is NOT run manually, this entire section is skipped and no logic from it is applied, basically meaning that ${{ github.event.inputs.build_type }} is empty.

gitbuda (approved):

PR #1448, opened 2023-11-02
Setting label as parameter

### LINKED ISSUE #1405: Setting label as a variable not possible in Memgraph
Query parameters are able to be inserted in Memgraph for literals. That includes property values and intermediate variables. 
However, labels are hardcoded in the query and no one is currently able to set an arbitrary label via the query parameters.

DavIvek:
Currently, we are able to use query parameters with property values and intermediate variables. However, labels are hardcoded in the query and no one is currently able to set an arbitrary label via the query parameters.
This PR should enable users to set an label via the query parameters.

[master < Task] PR

closes #1405


DavIvek:
After changing .g4 file, I have to change other files that are affected with that change. cypher_main_visitor.cpp has to be changed, ATM  `NodeAtom->labels_` is `std::vector<memgraph::query::LabelIx>` and I think it should be expanded to some kind of `std::variant<std::vector<memgraph::query::LabelIx>, memgraph::query::ParameterLookup>` .

DavIvek:
Currently, this PR covers creating labels via query parameters. Since ATM we are throwing exceptions when we use ParameterLookup in MATCH/MERGE statements there is still work to be done. 
CI doesn't pass because we can't build a memgraph with tests with this changes, so I need to adjust tests according to new changes.

DavIvek:
Currently, I have a problem with the index lookup phase. We are doing index lookup in the query planning phase and we are evaluating parameters into labels/properties in the evaluation phase. Based on the parameter string before evaluation, we can't do a proper index lookup. We have to discuss if it's okay to resolve parameters in the planning phase(or earlier) and we have to make a couple of changes in our parameter lookup code because ATM code isn't suited to do parameter resolving before the evaluation phase.

DavIvek:
This is a wrong and overcomplicated approach.

PR #1462, opened 2023-11-06
Add GC for query memory tracker

antoniofilipovic:
This PR adds GC for the query memory tracker.

[master < Task] PR

### DIFF
--- a/src/memory/query_memory_control.cpp
+++ b/src/memory/query_memory_control.cpp
@@ -155,6 +155,7 @@ inline int &Get_Thread_Tracker() {
 bool QueriesMemoryControl::IsThreadTracked() { return Get_Thread_Tracker() == 1; }
 
 void QueriesMemoryControl::ResetTrackings() {
+  Get_Thread_Tracker() = 0;
   transaction_id_to_tracker.clear();
   thread_id_to_transaction_id.clear();
 }
--- a/src/memory/query_memory_control.cpp
+++ b/src/memory/query_memory_control.cpp
@@ -202,11 +202,13 @@ void CleanTracker() {
 #endif
 }
 
-bool IsTransactionTracked(uint64_t transaction_id) {
 #if USE_JEMALLOC
+bool IsTransactionTracked(uint64_t transaction_id) {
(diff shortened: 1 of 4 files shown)


src/memory/query_memory_control.cpp
```diff
@@ -102,6 +102,17 @@ void QueriesMemoryControl::InitializeArenaCounter(unsigned arena_ind) {
...
 
+void QueriesMemoryControl::ResetTrackings() {
+  // First disable trackings on arenas
+  // so we don't start tracking on some query if
+  // there was something active left
+  for (auto &[arena_id, counter] : arena_tracking) {
+    counter.store(0, std::memory_order_acquire);
+  }
+  transaction_id_to_tracker.clear();
+  thread_id_to_transaction_id.clear();
```
Ignition, line 113: What if a `RemoveTrackingOnArena` happens making a counter negative.
I'm not confident the behavior of this will be correct if `FreeMemory` called at any arbitrary point during a query execution.

antoniofilipovic: So this is completely outdated logic, I removed that part completely on different diff, so I will resolve this

src/memory/query_memory_control.cpp
```diff
@@ -154,6 +154,12 @@ inline int &Get_Thread_Tracker() {
 
 bool QueriesMemoryControl::IsThreadTracked() { return Get_Thread_Tracker() == 1; }
 
+void QueriesMemoryControl::ResetTrackings() {
+  Get_Thread_Tracker() = 0;
+  transaction_id_to_tracker.clear();
+  thread_id_to_transaction_id.clear();
```
Ignition, line 160: I don't get it. If our GC run is exclusive (ie. no active transactions) we get to run this Reset. But in that situation why hasn't transaction end already unregistered itself. Ideally these collections should already be clear.

antoniofilipovic: Yes, they are clear, but this is just to have GC to be 100% sure.

antoniofilipovic:
So, decided to close this PR as this might hide some bugs we have in code. Better idea would be to not use skip lists at all, but to hold everything as part of class (is_thread_tracked, memory_tracker) which is thread local.

PR #1466, opened 2023-11-07
Implement constant time label and edge type retrieval

gvolfing:
Users have requested a way to be able to retrieve the node-labels and edge-types inside Memgraph in constant time. This PR provides a way of doing that.

Two queries has been added
 - SHOW NODE_LABELS INFO
 - SHOW EDGE_TYPES INFO

to retrieve the list of labels on nodes and types of edges respectively.

### DIFF
--- a/src/query/db_accessor.hpp
+++ b/src/query/db_accessor.hpp
@@ -597,6 +597,13 @@ class DbAccessor final {
     return accessor_->ApproximateVertexCount(label, property, lower, upper);
   }
 
+  std::vector<std::string> ListAllPossiblyPresentVertexLabels() const {
+    return accessor_->ListAllPossiblyPresentVertexLabels();
+  }
+  std::vector<std::string> ListAllPossiblyPresentEdgeTypes() const {
+    return accessor_->ListAllPossiblyPresentEdgeTypes();
+  }
+
   storage::IndicesInfo ListAllIndices() const { return accessor_->ListAllIndices(); }
 
   storage::ConstraintsInfo ListAllConstraints() const { return accessor_->ListAllConstraints(); }
--- a/src/query/interpreter.cpp
+++ b/src/query/interpreter.cpp
@@ -3076,6 +3076,38 @@ PreparedQuery PrepareDatabaseInfoQuery(ParsedQuery parsed_query, bool in_explici
       };
       break;
     }
+    case DatabaseInfoQuery::InfoType::EDGE_TYPES: {
+      header = {"edge types"};
+      handler = [storage = current_db.db_acc_->get()->storage(), dba] {
+        auto edge_types = dba->ListAllPossiblyPresentEdgeTypes();
+        std::vector<std::vector<TypedValue>> results;
+        results.reserve(edge_types.size());
+        for (auto &edge_type : edge_types) {
+          results.push_back({TypedValue(edge_type)});
+        }
+
+        return std::pair{results, QueryHandlerResult::COMMIT};
+      };
+
+      break;
+    }
+    case DatabaseInfoQuery::InfoType::NODE_LABELS: {
+      header = {"node labels"};
+      handler = [storage = current_db.db_acc_->get()->storage(), dba] {
+        auto node_labels = dba->ListAllPossiblyPresentVertexLabels();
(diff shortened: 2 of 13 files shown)


gitbuda:
@gvolfing does it make sense to have a feature flag?

gvolfing:
@gitbuda It does. I will add it shortly.
I was also thinking how to make this work with recovery. But I realized we do not have to, since during recovery we will call the `AddLabel()` function on the vertices either way, so we get recovery out of the box. If the flags are set ofc.

src/query/interpreter.cpp
```diff
@@ -3045,6 +3045,48 @@ PreparedQuery PrepareDatabaseInfoQuery(ParsedQuery parsed_query, bool in_explici
...
+          results.push_back({TypedValue(storage->LabelToName(node_label))});
+        }
+
+        return std::pair{results, QueryHandlerResult::COMMIT};
+      };
+
+      break;
+    }
+
+      // NODE_LABELS
```
Darych, line 3089: Do we still need this comment?

src/storage/v2/storage.hpp
```diff
@@ -394,6 +399,18 @@ class Storage {
   Indices indices_;
   Constraints constraints_;
 
+  // Datastructures to provide fast retrieval of node-label and
+  // edge-type related metadata.
+  // Currently we should not remove any node-labels or edge-types even
+  // if the set of given types are currently not present in the
```
Darych, line 405: Why do we need to keep labels/types which are not present now in graph?

Ignition: We do not reference count their usage, hence no way to deterministically remove their validity upon removal of last usage.

Ignition (changes requested):
`NameIdMapper` already has the names we are looking for. No need to store a 2nd time in `SynchronizedMetaDataStore`.

Expose something like `Names` from the `NameIdMapper`, which would be the keys the `name_to_id_` SkipList

gvolfing:
> `NameIdMapper` already has the names we are looking for. No need to store a 2nd time in `SynchronizedMetaDataStore`.
> 
> Expose something like `Names` from the `NameIdMapper`, which would be the keys from the `name_to_id_` SkipList

@Ignition 
Issue is that in the `NameIdMapper` we do not have knowledge about what the individual names are. All and any of them could be labels, edge types, or even property names.
If we want to use the `NameIdMapper`, to retrieve this data, then we need to update it with the capability of keeping track of what names have which types. We can do that but that would convolute a couple of things (recovery in particular).
I can change it if you think it is a better approach, please let me know what you think.

src/query/db_accessor.hpp
```diff
@@ -597,6 +597,13 @@ class DbAccessor final {
     return accessor_->ApproximateVertexCount(label, property, lower, upper);
   }
 
+  std::vector<storage::LabelId> ListAllPossiblyPresentVertexLabels() const {
```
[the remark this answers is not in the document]
gvolfing: Because it might be possible that the user adds some vertex with a specific label to the database, then deletes it. With these commands we want to keep track of what kind of data `could` be present in the database, hence we do not only check for labels/edge-types that are currently exists in the database.

src/storage/v2/storage.hpp
gvolfing: Just as @Ignition  said.
Providing a solution that keeps track of the count of each element is currently not possible, and I would rather not add an atomic RMW op to every single call that creates a label/edgetype.

Ignition:
> let me know what you think.

Your choice....trade off between complexity.
Is the recovery versioned to make this change safe?
If so then having multiple `NameIdMapper`, one per kind, maybe the best approach.

gvolfing:
> > let me know what you think.
> 
> Your choice....trade off between complexity. Is the recovery versioned to make this change safe? If so then having multiple `NameIdMapper`, one per kind, maybe the best approach.

@Ignition 
I would like to merge this with the dedicated `MetaDataStore`, but I agree that the `NameIdMapper` should be reworked and used instead. That is an overall better solution. I created a [issue](https://github.com/memgraph/memgraph/issues/1546) for that so we can revisit that in a separate PR.

PR #1468, opened 2023-11-07
Fix race condition and arena tracking bug

antoniofilipovic:
Fixes issue on query not being tracked due to issue with jemalloc mallctl call to thread.arena which can give different arena in different times of calling.

[master < Task] PR


antoniofilipovic:
Fixes issue: https://github.com/memgraph/memgraph/issues/1469

Ignition (approved):

antoniofilipovic:
@vpavicic RN: fix race condition on MEMORY QUERY LIMIT queries, where query memory limit was not sometimes respected

Issue #1220, opened 2023-09-02
Fix deleting nodes increases memory usage & eventually crashes server as it OOM's


chrischabot:
2.11.1 / EC2 / Docker memgraph-platform

When I delete nodes that are no longer needed, the total memory used by Memgraph goes **up**

I started out with 56M nodes, ran the query below to delete older \:Post nodes, reducing the node count to 40M. So a total of 16M nodes deleted

Starting out MG was using 12Gb. 
Right after the query had completed memory usage was at 18 Gb
Waiting some time (~10 min) for recovery / deallocation processes to work, it settled on 13 Gb used

I would expect the total memory usage to go down, and not slightly up, after removing 16m nodes. After restarting Memgraph to reset the memory usage, it went down to 8.3 Gb used

Due to this behavior BlueJ crashes after being untouched for a few days as Memgraph memory usage keeps going up as nodes & relationships are deleted and the memory for them not being freed 

Expected behavior: Not having to restart Memgraph to free memory, not having server crash due to out of memory issues 

Screenshots: show storage info
- right as the query completed
- after 10 min of waiting
- stopping & starting 

![CleanShot 2023-09-02 at 20 08 56@2x](https://github.com/memgraph/memgraph/assets/11455514/5ece7212-4a54-4e8a-b63e-87dbe79576a6)
![CleanShot 2023-09-02 at 20 03 47@2x](https://github.com/memgraph/memgraph/assets/11455514/fab00c4f-6fa0-491d-a902-a120addd872f)
![CleanShot 2023-09-02 at 20 12 48@2x](https://github.com/memgraph/memgraph/assets/11455514/5c20f602-9545-4f04-92e1-2736d383615a)

gitbuda:
Thanks for reporting! Keep calm with the issue related to the memory, we are working on a lot of improvements regarding memory (there are issues that have existed for a long time). The active work can be tracked under:
* https://github.com/memgraph/memgraph/pull/1129
* https://github.com/memgraph/memgraph/issues/745

We'll verify this independently, but as soon as some of the mentioned PRs get merged, we have to rerun what is specified here and figure out is the issue still there 😃

chrischabot:
To make reproduction easier, the latest (daily) BlueJ dump can be found on https://bluej.memgraph.com/dump/bluej-latest.cypherl.gz

The query used to delete old posts is:
```
MATCH (p:Post)
WITH localDateTime() - p.indexedAt as duration, p
WITH duration.day as days, p
[...]

PR #1471, opened 2023-11-07
Fix garbage collection not working properly and add periodic jemalloc purge

imilinovic:
Fix garbage collection not working properly, add periodic jemalloc memory purge

[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR

This fixes [Issue1220](https://github.com/memgraph/memgraph/issues/1220)


antepusic (commented):
Good work! I’ve left a few comments, answer appreciated :pray:

src/flags/general.cpp
```diff
@@ -65,7 +65,10 @@ DEFINE_bool(allow_load_csv, true, "Controls whether LOAD CSV clause is allowed i
 // Storage flags.
 // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
 DEFINE_VALIDATED_uint64(storage_gc_cycle_sec, 30, "Storage garbage collector interval (in seconds).",
-                        FLAG_IN_RANGE(1, 24 * 3600));
+                        FLAG_IN_RANGE(1, static_cast<unsigned long>(24) * 3600));
```
as51340, line 68:
```suggestion
                        FLAG_IN_RANGE(1, 24L * 3600));
```

as51340: or maybe UL

src/flags/general.cpp
```diff
@@ -65,7 +65,10 @@ DEFINE_bool(allow_load_csv, true, "Controls whether LOAD CSV clause is allowed i
...
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
+DEFINE_VALIDATED_uint64(storage_gc_jemalloc_cycle_sec, 30, "Storage jemalloc garbage collector interval (in seconds)",
+                        FLAG_IN_RANGE(1, static_cast<unsigned long>(24) * 3600));
```
as51340, line 71:
```suggestion
                        FLAG_IN_RANGE(1, 24UL * 3600));
```

src/storage/v2/config.hpp
```diff
@@ -31,12 +31,13 @@ class StorageConfigException : public utils::BasicException {
 /// the storage. This class also defines the default behavior.
 struct Config {
   struct Gc {
-    enum class Type { NONE, PERIODIC };
+    enum class Type { NONE, PERIODIC, PERIODIC_JEMALLOC };
```
as51340, line 34: I would remove PERIODIC_JEMALLOC type. If it is periodic in gc_jemalloc, we will know what is it for.

src/storage/v2/inmemory/storage.cpp
```diff
@@ -117,6 +127,9 @@ InMemoryStorage::~InMemoryStorage() {
   if (config_.gc.type == Config::Gc::Type::PERIODIC) {
     gc_runner_.Stop();
   }
+  if (config_.gc.type == Config::Gc::Type::PERIODIC_JEMALLOC) {
```
as51340, line 130:
```suggestion
  if (config_.gc_jemalloc.type == Config::Gc::Type::PERIODIC_JEMALLOC) {
```

as51340: This is a bug I think

as51340 (changes requested):

src/storage/v2/inmemory/storage.cpp
imilinovic: Good catch!

src/storage/v2/inmemory/storage.cpp
```diff
@@ -101,8 +103,16 @@ InMemoryStorage::InMemoryStorage(Config config, StorageMode storage_mode)
...
     }
   }
+
+  if (config_.gc_jemalloc.type == Config::Gc::Type::PERIODIC_JEMALLOC) {
+    gc_jemalloc_runner_.Run("GC jemalloc", config_.gc_jemalloc.interval, [] { memory::PurgeUnusedMemory(); });
+  }
+
   if (config_.gc.type == Config::Gc::Type::PERIODIC) {
-    gc_runner_.Run("Storage GC", config_.gc.interval, [this] { this->CollectGarbage<false>(); });
+    // TODO: move out of storage have one global gc_runner_
```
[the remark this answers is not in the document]
imilinovic: Yes

src/storage/v2/inmemory/storage.cpp
```diff
@@ -11,6 +11,8 @@
 
 #include "storage/v2/inmemory/storage.hpp"
 #include "dbms/constants.hpp"
+#include "memory/global_memory_control.hpp"
+#include "storage/v2/config.hpp"
```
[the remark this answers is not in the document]
imilinovic: global_memory control is needed but config.hpp is not!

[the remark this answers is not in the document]
as51340: yep I meant config

as51340 (commented):
I think this is good enough. Could we add some test which will check that memory is actually reclaimed after purging it?

antoniofilipovic (commented):
Seems fine to me. The only option for me that seems possible to add a test is a stress test where we do a bunch of creates and wait for the GC cycle to kick in, so we assert that memory after gc should kick in is lower than before. Check if that is possible, and put a small gc cycle of 5 seconds, and in the meantime, you can do one FOREACH transaction. That is my idea for the test, otherwise, I am not sure

Josipmrden:
I'm a bit more for having this under one flag only, and not displaying to the user the info about jemalloc. Couldn't hurt since now we have both cycle seconds set to 30 seconds. I'd also add @Ignition to review this before merging.

src/storage/v2/inmemory/storage.cpp
```diff
@@ -1210,7 +1226,7 @@ void InMemoryStorage::CollectGarbage(std::unique_lock<utils::ResourceLock> main_
       main_lock_.lock_shared();
     }
   } else {
-    MG_ASSERT(main_guard.mutex() == std::addressof(main_lock_), "main_guard should be only for the main_lock_");
+    DMG_ASSERT(main_guard.mutex() == std::addressof(main_lock_), "main_guard should be only for the main_lock_");
```
Josipmrden, line 1229: why DMG assert here

Josipmrden:
```
PreparedQuery PrepareFreeMemoryQuery(ParsedQuery parsed_query, bool in_explicit_transaction, CurrentDB &current_db) {
    throw FreeMemoryModificationInMulticommandTxException();
  }

  MG_ASSERT(current_db.db_acc_, "Free Memory query expects a current DB");
  storage::Storage *storage = current_db.db_acc_->get()->storage();

  if (storage->GetStorageMode() == storage::StorageMode::ON_DISK_TRANSACTIONAL) {
    throw FreeMemoryDisabledOnDiskStorage();
  }

  return PreparedQuery{{},
                       std::move(parsed_query.required_privileges),
                       [storage](AnyStream *stream, std::optional<int> n) -> std::optional<QueryHandlerResult> {
                         storage->FreeMemory();
                         memory::PurgeUnusedMemory();
                         return QueryHandlerResult::COMMIT;
                       },
                       RWType::NONE};
}
```

this is the code from the interpreter. After doing free memory, it automatically purges it so it's returned back to the process. I would then assume that we need to do both actions with one flag. User should not be concerned about what jemalloc is, and if we ever need to debug it separately, we can do so later

antoniofilipovic:
@imilinovic @Josipmrden I am more than okay to have two separate threads, one for GC for our internal stuff, one for jemalloc. I think users already are impacted by jemalloc via vm.max_map_count and they can read about that already there, so it is one more config that they can set up. Also if jemalloc purge blocks the whole thread, I think it will be too slow to do gc cycles and it might be an issue later on.

src/storage/v2/inmemory/storage.cpp
antoniofilipovic: Because it is dangerous to do these asserts in production and also slower, so this is a bit better solution as per @Ignition

Josipmrden:
> @imilinovic @Josipmrden I am more than okay to have two separate threads, one for GC for our internal stuff, one for jemalloc. I think users already are impacted by jemalloc via vm.max_map_count and they can read about that already there, so it is one more config that they can set up. Also if jemalloc purge blocks the whole thread, I think it will be too slow to do gc cycles and it might be an issue later on.

Please discuss with @gitbuda about the feature before we merge this

as51340 (approved):
:muscle:

antoniofilipovic (approved):
:muscle:

imilinovic:
More proactive GC deletion moved to https://github.com/memgraph/memgraph/pull/1508 because I found a bug in it.

src/flags/general.cpp
```diff
@@ -65,7 +65,10 @@ DEFINE_bool(allow_load_csv, true, "Controls whether LOAD CSV clause is allowed i
 // Storage flags.
 // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
 DEFINE_VALIDATED_uint64(storage_gc_cycle_sec, 30, "Storage garbage collector interval (in seconds).",
-                        FLAG_IN_RANGE(1, 24 * 3600));
+                        FLAG_IN_RANGE(1, 24UL * 3600));
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
+DEFINE_VALIDATED_uint64(storage_gc_jemalloc_cycle_sec, 30, "Storage jemalloc garbage collector interval (in seconds)",
+                        FLAG_IN_RANGE(1, 24UL * 3600));
```
gitbuda, line 71: Why not put this under `storage_ch_cycle_sec`? Assume users don't know anything about allocations. How would they know to tweak this? If we keep this flag, we need to add a very detailed description of what does this do and how to use it properly and in which cases.

How does the `gc_cycle_sec` relate to this flag? E.g. what happens if GC sec is small vs `jemalloc_cycle_sec` is big, and the other way around?

src/storage/v2/inmemory/storage.cpp
```diff
@@ -101,8 +102,14 @@ InMemoryStorage::InMemoryStorage(Config config, StorageMode storage_mode)
           "those files into a .backup directory inside the storage directory.");
     }
   }
+  if (config_.gc_jemalloc.type == Config::Gc::Type::PERIODIC) {
+    gc_jemalloc_runner_.Run("GC jemalloc", config_.gc_jemalloc.interval, [] { memory::PurgeUnusedMemory(); });
+  }
   if (config_.gc.type == Config::Gc::Type::PERIODIC) {
-    gc_runner_.Run("Storage GC", config_.gc.interval, [this] { this->CollectGarbage<false>(); });
+    // TODO: move out of storage have one global gc_runner_
```
gitbuda, line 109: Why not do this under this PR?

tests/e2e/garbage_collection/gc_periodic.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    This test checks that periodic gc works.
+    It does so by checking that the allocated memory is lowered by at least 1/4 of the memory allocated by creating nodes.
+    If we choose a number a high number the test will become flaky because the memory only gets fully cleared after a while
+    due to jemalloc holding some memory for a while. If we'd wait for jemalloc to fully release the memory the test would take too long.
+    """
+    cursor = connection.cursor()
+
+    memory_pre_creation = get_memory(cursor)
+    execute_and_fetch_all(cursor, "CALL garbage_collection.create_nodes();")
```
gitbuda, line 50: Can't this be done with a regular query? What's the reason to have the a new query module?

src/flags/general.cpp
antoniofilipovic: I am more than okay to have two separate flags, one for GC for our internal stuff, one for jemalloc. I think users already are impacted by jemalloc via vm.max_map_count and they can read about that already there, so it is one more config that they can set up. Also if jemalloc purge blocks the whole thread, I think it will be too slow to do gc cycles and it might be an issue later on.

src/storage/v2/inmemory/storage.cpp
antoniofilipovic: This is @Ignition comment I think

imilinovic: Yes

src/flags/general.cpp
imilinovic: @antoniofilipovic 
As far as Fico told me purging jemalloc memory is safe and doesn't require any lock, but he will know more for sure

gitbuda: Aha, so, on one hand, there is a concern that one scheduler/thread would block the GC, while on the other side there are two flags for the user. Isn't the optimal solution to have one flag `storage_gc_cycle_sec` but two schedulers (both configured with the same flag)?

PR #1479, opened 2023-11-09
Fix parallel index loading

Josipmrden:
Parallel index loading was not performing in all the cases.

[master < Task] PR

@vpavicic Changelog message: Allow parallel index loading on all paths of code.


gitbuda (approved):
I'll accept this, but the whole recovery code seems convoluted, seems like we could reduce the code size.

This seems a tech debt, next time we touch this code (parallel constraints recovery), we should improve on the tech dept side.

CC @gvolfing @hal-eisen-MG

PR #1481, opened 2023-11-09
Remove filter profile info

Josipmrden:
Filter profile info was not updating during rewriting of queries.

[master < Task] PR

@vpavicic : Changelog message: Filter information message fixed in EXPLAIN and PROFILE queries to display what's exactly filtering in the filter expression.

@vpavicic  I think I'll need to look about what currently our profile and explain information provides in the docs so we can update it


src/query/plan/rewrite/index_lookup.hpp
```diff
@@ -105,6 +105,11 @@ class IndexLookupRewriter final : public HierarchicalLogicalOperatorVisitor {
     prev_ops_.pop_back();
     ExpressionRemovalResult removal = RemoveExpressions(op.expression_, filter_exprs_for_removal_);
     op.expression_ = removal.trimmed_expression;
+    if (op.expression_) {
+      Filters leftover_filters;
+      leftover_filters.CollectFilterExpression(op.expression_, *symbol_table_);
+      op.all_filters_ = leftover_filters;
```
as51340, line 111:
```suggestion
      op.all_filters_.CollectFilterExpression(op.expression_, *symbol_table_);
```

src/query/plan/rewrite/join.hpp
```diff
@@ -59,6 +59,12 @@ class JoinRewriter final : public HierarchicalLogicalOperatorVisitor {
 
     ExpressionRemovalResult removal = RemoveExpressions(op.expression_, filter_exprs_for_removal_);
     op.expression_ = removal.trimmed_expression;
+    if (op.expression_) {
+      Filters leftover_filters;
+      leftover_filters.CollectFilterExpression(op.expression_, *symbol_table_);
+      op.all_filters_ = leftover_filters;
```
as51340, line 65:
```suggestion
      op.all_filters_.CollectFilterExpression(op.expression_, *symbol_table_);
```

src/query/plan/rule_based_planner.hpp
```diff
@@ -863,13 +863,13 @@ class RuleBasedPlanner {
...
-    auto all_filters = filters;
     auto pattern_filters = ExtractPatternFilters(filters, symbol_table, storage, bound_symbols);
     auto *filter_expr = impl::ExtractFilters(bound_symbols, filters, storage);
 
     if (filter_expr) {
-      last_op =
-          std::make_unique<Filter>(std::move(last_op), std::move(pattern_filters), filter_expr, std::move(all_filters));
+      Filters operator_filters;
+      operator_filters.CollectFilterExpression(filter_expr, symbol_table);
+      last_op = std::make_unique<Filter>(std::move(last_op), std::move(pattern_filters), filter_expr, operator_filters);
```
as51340, line 872:
```suggestion
      last_op = std::make_unique<Filter>(std::move(last_op), std::move(pattern_filters), filter_expr, std::move(operator_filters));
```

tests/e2e/filter_info/filter_info.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def test_label_index_hint(memgraph):
+    memgraph.execute("CREATE (n:Label1:Label2 {prop: 1});")
+    memgraph.execute("CREATE INDEX ON :Label1;")
+
+    # TODO: Fix this test since it should only filter on :Label2 and prop
+    expected_explain = [
+        " * Produce {n}",
+        " * Filter (n :Label1:Label2), {n.prop}",
```
as51340, line 25: What to do with this TODO? I also find it weird that it has to do filtering on both labels

tests/e2e/index_hints/index_hints.py
```diff
@@ -407,14 +409,15 @@ def test_multiple_match_query(memgraph):
     memgraph.execute("CREATE INDEX ON :Label2;")
     memgraph.execute("CREATE INDEX ON :Label3;")
 
+    # TODO: Fix this test since it has the filtering info wrong (filtering by label that's already indexed)
```
as51340, line 412: Shouldn't we fix that before merging?

tests/e2e/analyze_graph/optimize_indexes.py
```diff
@@ -62,7 +62,7 @@ def test_analyze_graph_delete_statistics(delete_query, multi_db):
     # After deleting statistics, id2 should be chosen because it has less vertices
     expected_explain_after_delete_analysis = [
         (f" * Produce {{n}}",),
-        (f" * Filter (n :Label), {{n.id1}}, {{n.id2}}",),
+        (f" * Filter {{n.id1}}",),
```
as51340, line 65: Did stuff change only when using hints or in both situations?

src/query/plan/rewrite/join.hpp
as51340: I think all_filters_ variable should now be renamed since it doesn't store all filters. Also, I think `CollectFilterExpression` will not necessarily always collect leftover filters?

tests/e2e/index_hints/index_hints.py
as51340: The correct Filter would be:
```
Filter(m:Label3) right? for the line changed
```

as51340 (changes requested):
I think the name of this PR is a bit misleading since you aren't actually remove filter profile info but rather changing the output of filter explain.

src/query/plan/rule_based_planner.hpp
antoniofilipovic, line 872:
```suggestion
      auto get_operator_filters = [&filter_expr, &symbol_table](){
         Filters operator_filters;
         operator_filters.CollectFilterExpression(filter_expr, symbol_table);
         return operator_filters;
      };
      last_op = std::make_unique<Filter>(std::move(last_op), std::move(pattern_filters), filter_expr, get_operator_filters());
```

src/query/plan/rewrite/join.hpp
antoniofilipovic: Do you even need `all_filters`?

antoniofilipovic (changes requested):
Few comments

src/query/plan/rewrite/join.hpp
Josipmrden: that's a good observation, I may not need the filters, I'll see

Josipmrden: I need the filters since we implemented that in a way to parse easily the information for the operator. At this point I'm just fixing the bug of wrong printing but I think it makes sense to remove all filters totally and parse the expression in a visitor rather than having something which needs invalidation if changed.

Josipmrden: Also a comment for @as51340 : I need to reassign the operator's all_filters_ since when we apply the indexing we don't need to just add new filters but to reassign that variable which has changed. That's why I go into the op.expression again and reassign it. I think what `CollectFilterExpression` does is append to the list of already present filters

src/query/plan/rule_based_planner.hpp
Josipmrden: why ?

tests/e2e/analyze_graph/optimize_indexes.py
Josipmrden: in every situation of explain and profile

Josipmrden: which has multiple filtering operations

tests/e2e/filter_info/filter_info.py
Josipmrden: yes in the expression with the subclass LabelsTest we have an optimization which has an array of Labels. We have always kept it that way but I encountered this only when I fixed this bug of printing the correct way. So the actual behaviour now is that we will use an indexed scan but we will still test against all the labels which is in my opinion a bug, but I think it's out of scope for now as I'm fixing the printing and not the expression handling.

tests/e2e/index_hints/index_hints.py
Josipmrden: I agree it should be fixed but I assessed the effort and it's not low effort, it may take a couple of days to go through the filtering logic so I'd first merge this as this printing is totally wrong.

src/query/plan/rewrite/index_lookup.hpp
as51340: This is still the same

src/query/plan/rewrite/join.hpp
as51340: Okay. But can we change the code as in the suggestion or not?

as51340 (changes requested):

as51340 (approved):

PR #1482, opened 2023-11-09
Fix empty aggregation returns null

### LINKED ISSUE #1474: Memgraph returns a Record with null values, Neo4j doesn't return a Record
**Memgraph version**
Memgraph 2.12

**Environment**
Memgraph in docker 

**Describe the bug**
When using collect Memgraph will return a result, Neo4j won't return the result. This causes users to change the code base, increased traction for migration. 

**To Reproduce**

No results: 
```
MATCH (subnet:Subnet)
WHERE FALSE
RETURN id(subnet) AS subnetId
```

Returns a record with null subnetId. 
```
MATCH (subnet:Subnet)
WHERE FALSE
WITH subnet, collect(subnet.ip) AS ips
RETURN id(subnet) AS subnetId
```

**Expected behavior**
Neo4j does not return results, Memgraph probably should not also.

as51340:
Closes #1474.

[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR

### DIFF
--- a/src/query/plan/operator.cpp
+++ b/src/query/plan/operator.cpp
@@ -3463,7 +3463,9 @@ class AggregateCursor : public Cursor {
     SCOPED_PROFILE_OP_BY_REF(self_);
 
     if (!pulled_all_input_) {
-      ProcessAll(&frame, &context);
+      if (!ProcessAll(&frame, &context) && self_.AllAggregationsForCollecting()) {
+        return false;
+      }
       pulled_all_input_ = true;
       aggregation_it_ = aggregation_.begin();
 
@@ -3567,12 +3569,16 @@ class AggregateCursor : public Cursor {
    * cache cardinality depends on number of
    * aggregation results, and not on the number of inputs.
    */
-  void ProcessAll(Frame *frame, ExecutionContext *context) {
+  bool ProcessAll(Frame *frame, ExecutionContext *context) {
     ExpressionEvaluator evaluator(frame, context->symbol_table, context->evaluation_context, context->db_accessor,
                                   storage::View::NEW);
+
+    bool pulled = false;
     while (input_cursor_->Pull(*frame, *context)) {
       ProcessOne(*frame, &evaluator);
+      pulled = true;
     }
+    if (!pulled) return false;
 
     // post processing
     for (size_t pos = 0; pos < self_.aggregations_.size(); ++pos) {
@@ -3606,6 +3612,7 @@ class AggregateCursor : public Cursor {
           break;
       }
     }
+    return true;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


src/query/plan/operator.cpp
```diff
@@ -3819,6 +3824,12 @@ UniqueCursorPtr Aggregate::MakeCursor(utils::MemoryResource *mem) const {
   return MakeUniqueCursorPtr<AggregateCursor>(mem, *this, mem);
 }
 
+auto Aggregate::AllAggregationsForCollecting() const -> bool {
```
Josipmrden, line 3827: Define the method name like a question to signify from the name that it returns a bool
e.g. `AreAllAggregationsForCollecting`

src/query/plan/operator.cpp
```diff
@@ -3664,6 +3668,7 @@ class AggregateCursor : public Cursor {
     auto unique_values_it = agg_value->unique_values_.begin();
     auto agg_elem_it = self_.aggregations_.begin();
     const auto counts_end = agg_value->counts_.end();
+    spdlog::info("Updating agg value");
```
Josipmrden, line 3671: delete this log, I think it's going to overflow the logs

tests/e2e/replication/show_while_creating_invalid_state.py
```diff
@@ -697,7 +697,7 @@ def test_sync_replication_when_main_is_killed():
         )
 
         # 2/
-        QUERY_TO_CHECK = "MATCH (n) RETURN COLLECT(n.name);"
+        QUERY_TO_CHECK = "MATCH (n) RETURN COUNT(n.name);"
```
Josipmrden, line 700: why did you put count here?

tests/unit/query_plan_accumulate_aggregate.cpp
```diff
@@ -277,13 +277,11 @@ TYPED_TEST(QueryPlanAggregateOps, WithoutDataWithGroupBy) {
...
     auto results = this->AggregationResults(true, false, {Aggregation::Op::COLLECT_LIST});
-    EXPECT_EQ(results.size(), 1);
-    EXPECT_EQ(results[0][0].type(), TypedValue::Type::List);
+    EXPECT_EQ(results.size(), 0);
   }
   {
     auto results = this->AggregationResults(true, false, {Aggregation::Op::COLLECT_MAP});
-    EXPECT_EQ(results.size(), 1);
-    EXPECT_EQ(results[0][0].type(), TypedValue::Type::Map);
+    EXPECT_EQ(results.size(), 0);
```
Josipmrden, line 284: can you maybe add an additional gql behave test from the issue so we document it from the e2e perspective

Josipmrden (commented):
Easy fix, some comments

tests/e2e/replication/show_while_creating_invalid_state.py
as51340: Because this test relies that aggregation will return some result. This PR basically does the change that collect(null) is not any more [] but rather nothing while count(null) is still 0

as51340:
@vpavicic If `collect` is called on empty list, nothing will be returned. Before, the null was returned.

vpavicic:
The `collect` function, when called on an empty list, now yields no results instead of returning `NULL`.

PR #1489, opened 2023-11-10
Update the issue template

gitbuda:


as51340 (commented):
I would add a section which asks about queries. Something like:
```
If possible, please provide us with queries you are using. Either here or privately on Discord...
```

antoniofilipovic (approved):

gitbuda:
Yes, because in the meantime, we went in a slightly different direction -> closing this PR.

PR #1491, opened 2023-11-10
Fix frame change collector pmr type error

antoniofilipovic:
If we use different allocation types then that means that the locality of these allocations is different which is an issue in performance.

[master < Task] PR


antoniofilipovic:
Need to add GQL behave test still and clean code

Josipmrden (approved):

PR #1494, opened 2023-11-10
Fix for in list segmentation fault

Josipmrden:
Segmentation fault was happening on the query:

```
CREATE (o:Node) SET o.Status = 'This is the status';

match (o:Node)
where o.Status IN ['This is not the status', 'This is the status']
return o;
```

@antoniofilipovic 

[master < Task] PR

@vpavicic : Changelog message: Fixed segmentation fault when doing IN LIST operations in queries.


gitbuda:
@antoniofilipovic your review required here 😃

gitbuda (approved):

PR #1508, opened 2023-11-14
GC more proactive deletion

imilinovic:
Add more proactive GC deletion.

[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR


imilinovic:
This was supposed to go in [PR1471](https://github.com/memgraph/memgraph/pull/1471) but I think there is a bug in the code:

If you create a lot of nodes and then do a match detach delete query, after gc kicks off the memory will be lowered but will not revert to the value of the empty database because jemalloc holds some memory - that is expected. One way to get jemalloc to release the memory is to do a few CREATE (n) RETURN n; queries with another thread. If you do that and wait a couple seconds the memory will revert close to the value of the empty database. With this commit that memory never gets cleared. I don't know why that happens. 

If you have problems reproducing the issue ping me on Slack.

PR #1536, opened 2023-11-21
Fixes for clang-tidy / sonar issues

Ignition:
Small changes to improve code base

Build the unit tests main.cpp as an object library to reduce re-compilation.
Various code cleanup to resolve a number of clang-tidy / sonar issues.

### DIFF
--- a/include/mgp.hpp
+++ b/include/mgp.hpp
@@ -34,7 +34,7 @@ namespace mgp {
 
 class IndexException : public std::exception {
  public:
-  explicit IndexException(const std::string &message) : message_(message) {}
+  explicit IndexException(std::string message) : message_(std::move(message)) {}
   const char *what() const noexcept override { return message_.c_str(); }
 
  private:
@@ -43,7 +43,7 @@ class IndexException : public std::exception {
 
 class ValueException : public std::exception {
  public:
-  explicit ValueException(const std::string &message) : message_(message) {}
+  explicit ValueException(std::string message) : message_(std::move(message)) {}
   const char *what() const noexcept override { return message_.c_str(); }
 
  private:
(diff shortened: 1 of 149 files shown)


tests/unit/slk_core.cpp
antepusic, line 1: Quick question – why should `std::less<>` be explicitly added to sets of `std::string` here (and elsewhere) when it’s already the default `std::set` comparator?

include/mgp.hpp
antepusic, line 1: This changes the signatures of a hefty number of C++ API methods. Since all changes should be reflected in the [C++ API docs](https://memgraph.com/docs/custom-query-modules/cpp/cpp-api), I’ve added the docs needed tag to the PR.

antepusic:
One more comment – SonarCloud’s reported three more issues about making constructors explicit. Since you’ve already fixed a lot of them so far, just make sure to address them.

tests/unit/slk_core.cpp
Ignition: By default it would be `std::set<std::string>` -> `std::set<std::string, std::less<std::string>>`
Which means it can only compare `std::string`
i.e. `my_set.find("foo");` would actually make a `std::string` temporary so it can do the lookup.
with `std::less<>` you can compare `std::string` inside the set with the `char const *` `"foo"` passed in by `find`

Ignition:
Documentation note: https://github.com/memgraph/memgraph/pull/1536#discussion_r1400609754

antepusic (commented):
All looks good, will approve once the checks pass :smile:

antepusic (approved):
Approving (see my last comment) :rocket:

vpavicic:
Is this a part of the 2.13 release? Should it be added to the milestone?

PR #1539, opened 2023-11-22
Implement short circuiting of exists evaluation

antoniofilipovic:
The idea of this PR is to short circuit `exists` evaluation if possible.

Example here:
```
MATCH (c)<-[r1]-(b)
WHERE 5 != 5 AND 3 != 4 AND (
  exists((:Label { id: 1 })<-[*]-(:Label))
)
RETURN *;
```

Exists was evaluated here although it doesn't have to be evaluated.  For that, I implemented 2 different functionalities:
- typed value which now supports std::function to defer evaluation of operator
- different concatenation of filters in and operator

[master < Task] PR


src/glue/communication.cpp
```diff
@@ -15,6 +15,7 @@
 #include <string>
 #include <vector>
 
+#include "query/typed_value.hpp"
```
as51340, line 18: I think you don't need this include

src/query/typed_value.hpp
```diff
@@ -420,6 +421,9 @@ class TypedValue {
     new (&graph_v) std::unique_ptr<Graph>(graph_ptr);
   }
 
+  explicit TypedValue(std::function<void(TypedValue *)> &&other)
```
as51340, line 424: I would pass it as value and then move it not necessarily as rvalue reference.

src/query/typed_value.cpp
```diff
@@ -215,6 +216,9 @@ TypedValue::TypedValue(const TypedValue &other, utils::MemoryResource *memory) :
     case Type::Duration:
       new (&duration_v) utils::Duration(other.duration_v);
       return;
+    case Type::Function:
+      LOG_FATAL("Unsupported TypedValue::Function copy ctor");
```
as51340, line 220: Why you couldn't copy `std::function`?

src/query/typed_value.cpp
```diff
@@ -569,6 +579,8 @@ TypedValue &TypedValue::operator=(const TypedValue &other) {
       case Type::Duration:
         new (&duration_v) utils::Duration(other.duration_v);
         return *this;
+      case Type::Function:
+        LOG_FATAL("Unsupported TypedValue::Type copy assignment operator");
```
as51340, line 583: Same I think we could copy assign it

src/query/interpret/awesome_memgraph_functions.cpp
```diff
@@ -593,6 +593,7 @@ TypedValue ValueType(const TypedValue *args, int64_t nargs, const FunctionContex
     case TypedValue::Type::Duration:
       return TypedValue("DURATION", ctx.memory);
     case TypedValue::Type::Graph:
+    case TypedValue::Type::Function:
```
as51340, line 596: Why here could not returned empty graph and empty std::function?

src/glue/communication.cpp
```diff
@@ -127,6 +128,8 @@ storage::Result<Value> ToBoltValue(const query::TypedValue &value, const storage
       return Value(value.ValueLocalDateTime());
     case query::TypedValue::Type::Duration:
       return Value(value.ValueDuration());
+    case query::TypedValue::Type::Function:
+      throw communication::bolt::ValueException("Unsupported conversion from TypedValue to Value");
```
as51340, line 132:
```suggestion
      throw communication::bolt::ValueException("Unsupported conversion from TypedValue::Function to Value");
```

src/query/interpret/eval.hpp
```diff
@@ -903,7 +903,14 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
     return TypedValue(std::move(result), ctx_->memory);
   }
 
-  TypedValue Visit(Exists &exists) override { return TypedValue{frame_->at(symbol_table_->at(exists)), ctx_->memory}; }
+  TypedValue Visit(Exists &exists) override {
+    if (frame_->at(symbol_table_->at(exists)).IsFunction()) [[likely]] {
```
as51340, line 907: I would cache accesses to symbol_table and frame_

src/query/plan/operator.cpp
```diff
@@ -66,6 +66,7 @@
 #include "utils/pmr/unordered_set.hpp"
 #include "utils/pmr/vector.hpp"
 #include "utils/readable_size.hpp"
+#include "utils/result.hpp"
```
as51340, line 69: Probably not necessary include

src/query/plan/rule_based_planner.cpp
```diff
@@ -516,14 +517,30 @@ bool HasBoundFilterSymbols(const std::unordered_set<Symbol> &bound_symbols, cons
 
 Expression *ExtractFilters(const std::unordered_set<Symbol> &bound_symbols, Filters &filters, AstStorage &storage) {
   Expression *filter_expr = nullptr;
+  std::vector<FilterInfo> and_joinable_filters{};
```
as51340, line 520: Not sure I understand the logic, why you need inner loop which iterates over and_joinable_filters? Shouldn't it be enough that you iterate over all filters and you just skip those with Type::Pattern?

as51340 (changes requested):
Like the work and impl!

src/query/typed_value.cpp
antoniofilipovic: I just disabled it not to copy it unnecessarily. I will enable it, makes sense

src/query/typed_value.hpp
antoniofilipovic: But this is move ctor, so I think that should be fine

src/query/interpret/awesome_memgraph_functions.cpp
antoniofilipovic: This is part of API towards functions, I think we don't need it, and if we do there is bug somewhere. Graph is only used for procedures and should be used that way, Function is only used on defering evaluation on operator, so it shouldn't yet be used here. If it will be, then someone can update it in my opinion.

src/query/interpret/eval.hpp
antoniofilipovic: Okay, will do

src/query/plan/rule_based_planner.cpp
antoniofilipovic: Idea here is to push Type::Pattern filters do the end of loop, so that is why I am iterating always over all of them, and if it is pattern filter, then I insert element just before  that Pattern filter element. 

So if you have
Current = Generic


                                                        1            2           3           4
and_joinable_filters:                    Generic Generic Generic Pattern

I want to insert Generic before PAttern filter, so if you have multiple ANDS, if one of the generic fails, I want to stop execution and not evaluate pattern

as51340: aha I think you can then use std::partition that should be simpler and faster :)

Josipmrden:
Execution seems to be correct, good job!

src/query/interpret/eval.hpp
```diff
@@ -903,7 +903,15 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
...
 
-  TypedValue Visit(Exists &exists) override { return TypedValue{frame_->at(symbol_table_->at(exists)), ctx_->memory}; }
+  TypedValue Visit(Exists &exists) override {
+    TypedValue &frame_exists_value = frame_->at(symbol_table_->at(exists));
+    if (frame_exists_value.IsFunction()) [[likely]] {
+      TypedValue result{ctx_->memory};
+      frame_exists_value.ValueFunction()(&result);
+      return result;
+    }
+    return TypedValue{frame_exists_value, ctx_->memory};
```
Josipmrden, line 913: I would rather throw here, since this should be a function at any point in time at the moment.

Josipmrden: or `MG_ASSERT`

src/query/plan/rule_based_planner.cpp
```diff
@@ -516,14 +517,30 @@ bool HasBoundFilterSymbols(const std::unordered_set<Symbol> &bound_symbols, cons
...
+  // evaluate pattern ( exists() ) filter
+  auto split =
+      std::partition(and_joinable_filters.begin(), and_joinable_filters.end(),
+                     [](const FilterInfo &filter_info) { return filter_info.type != FilterInfo::Type::Pattern; });
+  for (auto it = and_joinable_filters.begin(); it != split; ++it) {
+    filter_expr = impl::BoolJoin<AndOperator>(storage, filter_expr, it->expression);
+  }
+  for (; split != and_joinable_filters.end(); ++split) {
+    filter_expr = impl::BoolJoin<AndOperator>(storage, filter_expr, split->expression);
+  }
```
Josipmrden, line 542: Why looping here 2 times? Isn't `std::partition` operating on the `and_joinable_filters` and then you can just go over them and `impl::BoolJoin<AndOperator>`?

https://en.cppreference.com/w/cpp/algorithm/partition

Josipmrden (commented):
1 commend, otherwise good

src/query/plan/rule_based_planner.cpp
as51340: Yes I also think you can just do in one loop since there is no iterator which you want to avoid, you want to connect all of them

as51340 (changes requested):

src/query/plan/rule_based_planner.cpp
antoniofilipovic: My bad

Josipmrden (approved):

as51340 (approved):
:muscle:

PR #1540, opened 2023-11-22
Bugfix: correct replicaition handler

### LINKED ISSUE #1515: Memgraph nodes crash suddenly when replicating
**Memgraph version**
v2.10.0

**Environment**
Running on a kubernetes cluster, a memgraph cluster setup with 3 nodes, 1 main and 2 slaves.

**Describe the bug**
Running a cluster that seems healthy:

```
memgraph> show replicas;
+-----------------------------------+-----------------------------------+-----------------------------------+-----------------------------------+-----------------------------------+-----------------------------------+
| name                              | socket_address                    | sync_mode                         | current_timestamp_of_replica      | number_of_timestamp_behind_master | state                             |
+-----------------------------------+-----------------------------------+-----------------------------------+-----------------------------------+-----------------------------------+-----------------------------------+
| "slave0"                          | "10.0.172.179:10000"              | "async"                           | 844                               | 0                                 | "ready"                           |
| "slave1"                          | "10.0.252.19:10000"               | "async"                           | 844                               | 0                                 | "ready"                           |
+-----------------------------------+-----------------------------------+-----------------------------------+-----------------------------------+-----------------------------------+-----------------------------------+
2 rows in set (round trip in 0.001 sec)
```
After a few minutes any harmless query like the following one that return 2 simple records:
```
match (n:Account) return n;
```

Causes the slaves stop syncing and raise this error:
```
[2023-11-15 16:07:09.747] [memgraph_log] [critical] 
Assertion failed in file /memgraph/src/storage/v2/inmemory/storage.cpp at line 496.
        Expression: 'inserted'
        Message: 'The edge must be inserted here!'
```

The original query never ends on the master and if I cancel it, the state of the cluster is:

```
memgraph> show replicas;
+-----------------------------------+-----------------------------------+-----------------------------------+-----------------------------------+-----------------------------------+-----------------------------------+
| name                              | socket_address                    | sync_mode                         | current_timestamp_of_replica      | number_of_timestamp_behind_master | state                             |
+-----------------------------------+-----------------------------------+-----------------------------------+-----------------------------------+-----------------------------------+-----------------------------------+
| "slave0"                          | "10.0.172.179:10000"              | "async"                           | 0                                 | 0                                 | "invalid"                         |
| "slave1"                          | "10.0.252.19:10000"               | "async"                           | 0                                 | 0                                 | "invalid"                         |
+-----------------------------------+-----------------------------------+-----------------------------------+-----------------------------------+-----------------------------------+-----------------------------------+
```

It does not happen every time this query (or any similar) run, but quite often. The same workload seems to work well when memgraph is running in 1 node.

Ignition:
Fixes root cause of a cascade of failures in replication code:
- Replica handling of deleting an edge is now corrected. Now tolerant of multiple edges of the same relationship type.
- Improved robustness: correct exception handling around failed stream of current WAL file. This now means a REPLICA failure will no longer prevent transactions on MAIN from performing WAL writes.
- Slightly better diagnostic messages, not user friendly but helps get developer to correct root cause quicker.

fixes #1515


src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -472,16 +478,15 @@ uint64_t InMemoryReplicationHandlers::ReadAndApplyDelta(storage::InMemoryStorage
...
-                                           &*to_vertex);
-        if (edges.HasError()) throw utils::BasicException("Invalid transaction!");
-        if (edges->edges.size() != 1) throw utils::BasicException("Invalid transaction!");
-        auto &edge = (*edges).edges[0];
-        auto ret = transaction->DeleteEdge(&edge);
-        if (ret.HasError()) throw utils::BasicException("Invalid transaction!");
+        if (!to_vertex) throw utils::BasicException("Invalid transaction! Please raise an issue, line: {}", __LINE__);
+        auto edgeType = transaction->NameToEdgeType(delta.edge_create_delete.edge_type);
+        auto edge =
+            transaction->FindEdge(delta.edge_create_delete.gid, View::NEW, edgeType, &*from_vertex, &*to_vertex);
```
Josipmrden, line 486: I would assume this FindEdge is quite costly if we're deleting from a supernode?
Does it make sense to probe the source out edges as well as destination in edges in order to choose the one with less edges?

src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -415,40 +416,44 @@ uint64_t InMemoryReplicationHandlers::ReadAndApplyDelta(storage::InMemoryStorage
         spdlog::trace("       Delete vertex {}", delta.vertex_create_delete.gid.AsUint());
         auto *transaction = get_transaction(timestamp);
         auto vertex = transaction->FindVertex(delta.vertex_create_delete.gid, View::NEW);
-        if (!vertex) throw utils::BasicException("Invalid transaction!");
+        if (!vertex) throw utils::BasicException("Invalid transaction! Please raise an issue, line: {}", __LINE__);
```
Josipmrden, line 419: does it make sense to put this in a `constexpr` somewhere?

src/storage/v2/disk/storage.cpp
```diff
@@ -949,6 +949,20 @@ Result<EdgeAccessor> DiskStorage::DiskAccessor::CreateEdge(VertexAccessor *from,
   return EdgeAccessor(edge, edge_type, from_vertex, to_vertex, storage_, &transaction_);
 }
 
+std::optional<EdgeAccessor> DiskStorage::DiskAccessor::FindEdge(Gid gid, View view, EdgeTypeId edge_type,
+                                                                VertexAccessor *from_vertex,
+                                                                VertexAccessor *to_vertex) {
+  auto res = from_vertex->OutEdges(view, {edge_type}, to_vertex);
```
Josipmrden, line 955: yes here is the implementation that worries me if we replicate with a supernode in the system

Josipmrden: Since for every delta we can have a linear search -> `O(n^2)` again

src/storage/v2/indices/indices.cpp
```diff
@@ -50,4 +65,8 @@ Indices::Indices(const Config &config, StorageMode storage_mode) {
   });
 }
 
+Indices::Something Indices::Analysis() {
```
Josipmrden, line 68: what is `Indices::Something`?

src/storage/v2/inmemory/label_index.cpp
```diff
@@ -10,8 +10,10 @@
 // licenses/APL.txt.
 
 #include "storage/v2/inmemory/label_index.hpp"
+#include <span>
```
Josipmrden, line 13: empty line

src/storage/v2/inmemory/storage.cpp
```diff
@@ -315,6 +319,24 @@ Result<EdgeAccessor> InMemoryStorage::InMemoryAccessor::CreateEdge(VertexAccesso
   return EdgeAccessor(edge, edge_type, from_vertex, to_vertex, storage_, &transaction_);
 }
 
+std::optional<EdgeAccessor> InMemoryStorage::InMemoryAccessor::FindEdge(Gid gid, const View view, EdgeTypeId edge_type,
+                                                                        VertexAccessor *from_vertex,
+                                                                        VertexAccessor *to_vertex) {
+  auto res = from_vertex->OutEdges(view, {edge_type}, to_vertex);
```
Josipmrden, line 325: same here

src/storage/v2/inmemory/storage.cpp
```diff
@@ -975,10 +1033,41 @@ void InMemoryStorage::InMemoryAccessor::Abort() {
...
+      auto edges_acc = mem_storage->edges_.access();
+      for (auto gid : my_deleted_edges) {
+        edges_acc.remove(gid);
+      }
+    }
+
+    // mem_storage->deleted_vertices_.WithLock(
+    // [&](auto &deleted_vertices) { deleted_vertices.splice(deleted_vertices.begin(), my_deleted_vertices); });
+    // mem_storage->deleted_edges_.WithLock(
+    // [&](auto &deleted_edges) { deleted_edges.splice(deleted_edges.begin(), my_deleted_edges); });
```
Josipmrden, line 1070: what's this

Josipmrden (commented):
Some comments about performance

src/dbms/inmemory/replication_handlers.cpp
Ignition: WalDeltaData::Type::EDGE_DELETE was previously implemented in terms of OutEdges. The new FindEdge is implemented in terms of OutEdges. Don't expect a performance degradation as previously replication didn't allow multiple edges of same type.

Also filter of a `std::vector<std::tuple<EdgeTypeId, Vertex *, EdgeRef>>` I doubt is a major expense even for a supernode.

You are right, performance could be improved. TODO: `FindEdge` internals, check both to and from vertex.

src/storage/v2/disk/storage.cpp
Ignition: Our usage during replication handling is recreating the delta creation for REPLICA, the view is NEW so would do very limited processing of the delta chain even for supernodes.

Ignition: All written deltas will have the same commit_timestamp + command_id, hence only one delta would be read at most during FindEdge.
```    
if (view == View::NEW && ts == commit_timestamp && cid <= transaction->command_id) {
    break;
}
```

src/dbms/inmemory/replication_handlers.cpp
andrejtonev: Doing a quick check to see who has fever edges and then using that node for the search.

src/storage/v2/inmemory/storage.cpp
andrejtonev: That's deleted :smile:

src/storage/v2/indices/indices.cpp
andrejtonev: Renamed to IndexStats

gvolfing (approved):
I have a small concern about the name and return type of the function in the above mentioned comment, otherwise looks good to go. Good job!

src/storage/v2/disk/storage.cpp
```diff
@@ -69,6 +69,31 @@
 #include "utils/string.hpp"
 #include "utils/typeinfo.hpp"
 
+namespace {
+
+memgraph::storage::EdgeDirection ApproximateSmallerDegree(memgraph::storage::Vertex *from_vertex,
+                                                          memgraph::storage::Vertex *to_vertex) {
```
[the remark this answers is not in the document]
Ignition: > We are not approximating anything

Its `Approximate` because it uses the current edge count, it is completely timestamp and hence also delta agnostic. Its not perfect but a good heuristic approach to select the smaller side.

> If the function returns EdgeDirection::OUT that means that the from_vertex has less outgoing edges than the to_vertex has incoming ones right?

Yes

> Then, on the caller side we .... Why not just return the result of that VertexAccessor member (either InEdges or OutEdges) ...

Its only after refactoring the call site got that simple. You are right we can move that bit here. Just need to find a good name for the function.

src/storage/v2/disk/edge_import_mode_cache.cpp
```diff
@@ -28,7 +30,7 @@ EdgeImportModeCache::EdgeImportModeCache(const Config &config)
 InMemoryLabelIndex::Iterable EdgeImportModeCache::Vertices(LabelId label, View view, Storage *storage,
                                                            Transaction *transaction) const {
   auto *mem_label_index = static_cast<InMemoryLabelIndex *>(in_memory_indices_.label_index_.get());
-  return mem_label_index->Vertices(label, view, storage, transaction);
+  return mem_label_index->Vertices(label, vertices_.access(), view, storage, transaction);  // HERE!
```
Ignition, line 33: Comment not needed

src/storage/v2/inmemory/storage.cpp
```diff
@@ -975,10 +1071,39 @@ void InMemoryStorage::InMemoryAccessor::Abort() {
...
     });
-    mem_storage->deleted_vertices_.WithLock(
-        [&](auto &deleted_vertices) { deleted_vertices.splice(deleted_vertices.begin(), my_deleted_vertices); });
-    mem_storage->deleted_edges_.WithLock(
-        [&](auto &deleted_edges) { deleted_edges.splice(deleted_edges.begin(), my_deleted_edges); });
+
+    /// We MUST unlink (aka. remove) entries in indexes and constraints
+    /// before we unlink (aka. remove) vertices from storage
+    /// this is because they point into vertices skip_list
+    /// hence we need to grab access to vertices now to prevent deletion during any
```
Ignition, line 1078: Comment assumption not true. Remove this comment.

src/utils/skip_list.hpp
```diff
@@ -1226,6 +1226,7 @@ class SkipList final : detail::SkipListNode_base {
             !candidate->marked.load(std::memory_order_acquire));
   }
 
+  /// WHAT happens to prevent READ of just unlinked entries
```
Ignition, line 1229: Comment not needed

Josipmrden (approved):

Issue #662, opened 2022-11-23
Add indexing properties on relationships, a.k.a. edge indexes


jhb:
According to https://memgraph.com/docs/memgraph/reference-guide/indexing it is possible to create indexes on node properties, but no such functionality exists for edges.

IMHO there are a lot of use cases where import properties are put on edges - sometimes the edge properties are even more relevant than the node properties.  E.g. I would like to find all edges that carry a certain date, or are within a range (if modeled with a start and end property). Or edges that carry a certain source of origin, and so on.

While surely the graph could be remodeled with edges as "connection nodes", making this a necessity is not really desirable for graph modeling, and somewhat defeats the purpose of graph databases. Also, competitors have this feature, and migrating data towards memgraph would then require remodeling the whole graph.

IMPORTANT:
* an option is to add node id to each edge -> very memory inefficient at least 16B more for each edge
* another option is to embed vertex IDs inside the index itself -> no memory penalty if the index is not used

PRs:
* https://github.com/memgraph/memgraph/pull/1964

gitbuda:
Hi @jhb!

Definitely a helpful feature :) I/we would consider adding it in the short run, but since we are doing a significant rearchitecting of the Memgraph ([project-pineapples](https://github.com/memgraph/memgraph/tree/project-pineapples) branch), features like this are not a priority at the moment. Once we are done with the groundwork, we'll get back 😄

That being said, one alternative that doesn't require remodeling the whole graph is to store some form of node identifier on the edge (on the application side). If nodes are indexed based on the identifier, once you have an edge, it's possible to query the edge fast because you can access it via node. In other words, having the index on the app side. I know it might not be the perfect solution. Is that feasible in your case?

jhb:
Hi Marko, thanks a lot for your feedback, and of course for your work on memgraph in the first place.

I am looking at different use cases at the moment. For my personal usecase, a generic graph editor (for knowlege management etc.) workarounds in the form of extra nodes that are indexed on behalf of edges might be doable. Those nodes become a bit of a hassle though because they need filtering in the app in all kinds of places - I guess indexing edges outside memgraph would be easier in that case (using a repository layer in front of the graph).

For other use cases (graph analysis etc.) I would imagine the workarounds to become quite cumbersome, because there is not app to carry the extra logic most of the time, and those extra nodes might spoil the experience of interacting directly with the graph.

Having said all this - there is no urgency, and I am looking forward to seeing what changes the pineapples will bring. 

Cheers J

[...]

PR #1542, opened 2023-11-24
Implement edge type indices

gvolfing:
Currently there are two different kinds of indices in the database indices for label and indices for label+property based look-ups on
vertices. With the edge-type indexing capabilities it should be possible to create and deal with indices on edge types in the same way as with indices on vertex labels and properties. The case for edge-type indexes is, that in the case if the user would like to query all the edges based on a specific edge-type, the created query-plan has to traverse the entire graph, looking for the specific edge-type. The edge-type indexing feature solves this issue, however the flag, `--storage-properties-on-edges` has to be enabled.


katarinasupe:
Can you link https://github.com/memgraph/memgraph/issues/662 to development if this PR is fixing that issue?

gvolfing:
@katarinasupe 
These are two different indexing strategies.
This one is about indexing the edge-types themselves. Not properties on edges. However I can see this one being used to ease the use-case that the user is talking about. Once this is done, I will link this PR in the issue you tagged here, but again, this is not a direct solution to the issue the user presents in https://github.com/memgraph/memgraph/issues/662.

src/query/plan/planner.hpp
```diff
@@ -54,8 +55,11 @@ class PostProcessor final {
   std::unique_ptr<LogicalOperator> Rewrite(std::unique_ptr<LogicalOperator> plan, TPlanningContext *context) {
     auto index_lookup_plan =
         RewriteWithIndexLookup(std::move(plan), context->symbol_table, context->ast_storage, context->db, index_hints_);
-    return RewriteWithJoinRewriter(std::move(index_lookup_plan), context->symbol_table, context->ast_storage,
-                                   context->db);
+    auto join_plan =
+        RewriteWithJoinRewriter(std::move(index_lookup_plan), context->symbol_table, context->ast_storage, context->db);
+    auto edge_index_plan = RewriteWithEdgeTypeIndexRewriter(std::move(join_plan), context->symbol_table,
```
Darych, line 60: Shouldn't we rewrite with EdgeTypeIndex first and only then RewriteJoin?

src/query/plan/operator.cpp
```diff
@@ -517,6 +518,60 @@ class ScanAllCursor : public Cursor {
...
+    vertices_end_it_ = std::nullopt;
+  }
+
+ private:
+  const ScanAllByEdgeType &self_;
+  const Symbol output_symbol_;
+  const UniqueCursorPtr input_cursor_;
+  storage::View view_;
+  TVerticesFun get_edges_;
+  std::optional<typename std::result_of<TVerticesFun(Frame &, ExecutionContext &)>::type::value_type> vertices_;
```
Darych, line 569: Do we operate with vertices or edges? From storage files and purpose of EdgeTypeIndex, we should operate with edges. If that's true, can we rename please fields to not mess up?

antoniofilipovic (changes requested):

antoniofilipovic (changes requested):
Good work, I left few comments where I think we might have bug

src/query/plan/rewrite/edge_type_index_lookup.hpp
```diff
@@ -0,0 +1,550 @@
...
+  bool PreVisit(Filter &op) override {
+    prev_ops_.push_back(&op);
+    filters_.CollectFilterExpression(op.expression_, *symbol_table_);
+    return true;
+  }
+
+  // Remove no longer needed Filter in PostVisit, this should be the last thing
+  // Filter::Accept does, so it should be safe to remove the last reference and
+  // free the memory.
+  bool PostVisit(Filter &op) override {
```
antoniofilipovic, line 58: So here idea is after expand to remove filter? What if we have more than one filter will we remove all of them?

src/storage/v2/inmemory/edge_type_index.cpp
```diff
@@ -0,0 +1,302 @@
...
+void InMemoryEdgeTypeIndex::RunGC() {
+  for (auto &index_entry : index_) {
+    index_entry.second.run_gc();
+  }
+}
+
+InMemoryEdgeTypeIndex::Iterable InMemoryEdgeTypeIndex::Edges(EdgeTypeId edge_type, View view, Storage *storage,
+                                                             Transaction *transaction) {
+  const auto it = index_.find(edge_type);
+  MG_ASSERT(it != index_.end(), "Index for edge-type {} doesn't exist", edge_type.AsUint());
```
Darych, line 298: Should we assert? It's normal to not have an index by some type. Or did we required to check it upper in stack?

tests/unit/storage_v2_durability_inmemory.cpp
```diff
@@ -2972,3 +3006,43 @@ TEST_P(DurabilityTest, ConstraintsRecoveryFunctionSetting) {
           &variant_existence_constraint_creation_func);
   MG_ASSERT(pval_existence, "Chose wrong type of function for recovery of existence constraint data");
 }
+
+// NOLINTNEXTLINE(hicpp-special-member-functions)
+TEST_P(DurabilityTest, EdgeTypeIndexRecovered) {
+  if (GetParam() == true) {
```
Darych, line 3012: Instead of placing all function body under one if, maybe we can reverse condition of if and make early return?

src/query/frontend/semantic/symbol.hpp
```diff
@@ -53,6 +53,8 @@ class Symbol {
   bool user_declared() const { return user_declared_; }
   int token_position() const { return token_position_; }
 
+  bool IsSymbolAnonym() const { return name_.substr(0U, 4U) == "anon"; }
```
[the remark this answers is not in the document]
gvolfing: Yes it is.
We are looking for the ()-[r:EDGETYPE]->() pattern here.
When we create node patterns like so, they are always prefixed with "anon", and this is the only way we can identify them atm.

src/query/plan/planner.hpp
gvolfing: No. We are only looking for the ()-[r:EDGETYPE]->() pattern. It will always has to be picked as the query plan we would like to execute.

src/storage/v2/inmemory/edge_type_index.cpp
```diff
@@ -0,0 +1,302 @@
...
+      case Delta::Action::ADD_LABEL:
+      case Delta::Action::REMOVE_LABEL:
+      case Delta::Action::SET_PROPERTY:
+      case Delta::Action::ADD_IN_EDGE: {
+        link = {delta.vertex_edge.edge_type, delta.vertex_edge.vertex, delta.vertex_edge.edge};
+        auto it = std::find(vertex->in_edges.begin(), vertex->in_edges.end(), link);
+        MG_ASSERT(it == vertex->in_edges.end(), "Invalid database state!");
+        break;
+      }
+      case Delta::Action::ADD_OUT_EDGE: {
```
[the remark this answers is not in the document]
gvolfing: Faulty fallthrough, will fix.

src/storage/v2/inmemory/storage.cpp
```diff
@@ -350,6 +351,10 @@ Result<EdgeAccessor> InMemoryStorage::InMemoryAccessor::CreateEdge(VertexAccesso
         transaction_.manyDeltasCache.Invalidate(from_vertex, edge_type, EdgeDirection::OUT);
         transaction_.manyDeltasCache.Invalidate(to_vertex, edge_type, EdgeDirection::IN);
 
+        // Update indices if they exist.
+        // TODO enforce the presence of config_.properties_on_edges
+        storage_->indices_.UpdateOnEdgeCreation(from_vertex, to_vertex, edge, edge_type, transaction_);
```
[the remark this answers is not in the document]
gvolfing: No.
If you look at the CreateVertexEx function there, we do not update the Label + LabelProperty indices there either.

src/storage/v2/inmemory/storage.cpp
```diff
@@ -350,6 +351,10 @@ Result<EdgeAccessor> InMemoryStorage::InMemoryAccessor::CreateEdge(VertexAccesso
         transaction_.manyDeltasCache.Invalidate(from_vertex, edge_type, EdgeDirection::OUT);
         transaction_.manyDeltasCache.Invalidate(to_vertex, edge_type, EdgeDirection::IN);
 
+        // Update indices if they exist.
```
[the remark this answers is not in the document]
gvolfing: I don't think those functions rely on Edge constructs in any way. The EdgeAccessor that they operate, on only rely on the to and from vertices which we do not modify with edge-type indices.

src/storage/v2/durability/snapshot.cpp
```diff
@@ -1378,6 +1383,349 @@ RecoveredSnapshot LoadSnapshotVersion15(const std::filesystem::path &path, utils
...
+        if (!property) throw RecoveryFailure("Couldn't read property of existence constraints!");
+        AddRecoveredIndexConstraint(&indices_constraints.constraints.existence,
+                                    {get_label_from_id(*label), get_property_from_id(*property)},
+                                    "The existence constraint already exists!");
+        SPDLOG_TRACE("Recovered metadata of existence constraint for :{}({})",
+                     name_id_mapper->IdToName(snapshot_id_map.at(*label)),
+                     name_id_mapper->IdToName(snapshot_id_map.at(*property)));
+      }
+      spdlog::info("Metadata of existence constraints are recovered.");
+    }
```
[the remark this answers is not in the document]
gvolfing: I would rather not.
As we add newer recovery versions, the older versions of recovery will be eventually deprecated in a rolling window manner. It is easier to just delete a function rather than having to look for supporting functions that are becoming obsolete as well.

src/storage/v2/durability/snapshot.cpp
```diff
@@ -1378,6 +1383,349 @@ RecoveredSnapshot LoadSnapshotVersion15(const std::filesystem::path &path, utils
...
+        }
+        AddRecoveredIndexConstraint(&indices_constraints.constraints.unique, {get_label_from_id(*label), properties},
+                                    "The unique constraint already exists!");
+        SPDLOG_TRACE("Recovered metadata of unique constraints for :{}",
+                     name_id_mapper->IdToName(snapshot_id_map.at(*label)));
+      }
+      spdlog::info("Metadata of unique constraints are recovered.");
+    }
+    spdlog::info("Metadata of constraints are recovered.");
+  }
```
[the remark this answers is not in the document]
gvolfing: Same comment as above.

src/storage/v2/replication/replication_client.cpp
```diff
@@ -346,6 +346,7 @@ void ReplicaStream::AppendTransactionEnd(uint64_t final_commit_timestamp) {
   EncodeTransactionEnd(&encoder, final_commit_timestamp);
 }
 
+// TODO figure this out... We should encorporate edge-type here somehow...
```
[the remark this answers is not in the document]
gvolfing: It is. Good catch. I should pay more attention to removing TODOs.

src/storage/v2/vertices_iterable.hpp
```diff
@@ -12,11 +12,14 @@
 #pragma once
 
 #include "storage/v2/all_vertices_iterable.hpp"
+#include "storage/v2/inmemory/edge_type_index.hpp"
```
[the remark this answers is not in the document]
gvolfing: We don't, originally the iterable was in this source file and I forgot to remove the remnants after moving it into it's own file.

src/query/plan/rewrite/edge_type_index_lookup.hpp
gvolfing: Good catch.
Since we are looking only for the ()-[r:EDGETYPE]->() pattern this is not really an issue.
If we decide to expand the edge-type indexing capabilities in the future, this might become a relevant concern, but for now I will just remove this functionality.

src/query/plan/rewrite/edge_type_index_lookup.hpp
```diff
@@ -0,0 +1,550 @@
...
+    prev_ops_.pop_back();
+    ExpressionRemovalResult removal = RemoveExpressions(op.expression_, filter_exprs_for_removal_);
+    op.expression_ = removal.trimmed_expression;
+    if (op.expression_) {
+      Filters leftover_filters;
+      leftover_filters.CollectFilterExpression(op.expression_, *symbol_table_);
+      op.all_filters_ = std::move(leftover_filters);
+    }
+
+    if (!op.expression_ || utils::Contains(filter_exprs_for_removal_, op.expression_)) {
```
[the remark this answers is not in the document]
gvolfing: Yes we did. On top of that we do not need the functionality at all. We are only looking for the ()-[r:RELATIONSHIP]->() pattern so if the filter visitors should be noop. I will fix this.

src/storage/v2/inmemory/edge_type_index.cpp
```diff
@@ -0,0 +1,302 @@
...
+      continue;
+    }
+
+    const bool edge_was_deleted = index_iterator_->edge->deleted;
+    auto [edge_ref, edge_type, deleted_from_vertex, deleted_to_vertex] = GetEdgeInfo();
+    if (edge_ref == EdgeRef(nullptr)) {
+      MG_ASSERT(false, "Invalid database state!");
+    }
+
+    if (edge_ref == current_edge_) {
```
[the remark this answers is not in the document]
gvolfing: We don't anymore, I did a bit of refactoring around this part and this remained here. Will remove.

src/storage/v2/inmemory/edge_type_index.cpp
```diff
@@ -0,0 +1,302 @@
...
+        it = next_it;
+        continue;
+      }
+
+      const bool edge_deleted = !std::ranges::all_of(it->from_vertex->out_edges, [&](const auto &edge) {
+        auto *to_vertex = std::get<InMemoryEdgeTypeIndex::kVertexPos>(edge);
+        return to_vertex != it->to_vertex;
+      });
+
+      if ((next_it != edges_acc.end() && it->edge->gid == next_it->edge->gid) || it->from_vertex->deleted ||
```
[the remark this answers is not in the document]
gvolfing: To be honest, I am not sure.
The same logic is there with the other indices as well, I did test if it is needed, but forgot to remove it.

src/storage/v2/inmemory/edge_type_index.cpp
```diff
@@ -71,17 +71,22 @@ ReturnType VertexDeletedConnectedEdges(Vertex *vertex, const Transaction *transa
       case Delta::Action::ADD_LABEL:
       case Delta::Action::REMOVE_LABEL:
       case Delta::Action::SET_PROPERTY:
-      case Delta::Action::ADD_IN_EDGE: {
-        link = {delta.vertex_edge.edge_type, delta.vertex_edge.vertex, delta.vertex_edge.edge};
-        auto it = std::find(vertex->in_edges.begin(), vertex->in_edges.end(), link);
-        MG_ASSERT(it == vertex->in_edges.end(), "Invalid database state!");
         break;
+      case Delta::Action::ADD_IN_EDGE: {
+        if (edge == delta.vertex_edge.edge.ptr) {
```
antoniofilipovic, line 76: I think this is now good :+1:

gvolfing:
Docs
https://github.com/memgraph/documentation/pull/502

src/storage/v2/inmemory/edge_type_index.cpp
```diff
@@ -195,6 +199,18 @@ void InMemoryEdgeTypeIndex::UpdateOnEdgeCreation(Vertex *from, Vertex *to, EdgeR
...
 
+void InMemoryEdgeTypeIndex::UpdateOnEdgeModification(Vertex *old_from, Vertex *old_to, Vertex *new_from, Vertex *new_to,
+                                                     EdgeRef edge_ref, EdgeTypeId edge_type, const Transaction &tx) {
+  auto it = index_.find(edge_type);
+  if (it == index_.end()) {
+    return;
+  }
+  auto acc = it->second.access();
+
+  acc.remove(Entry{old_from, old_to, edge_ref.ptr, tx.start_timestamp});
```
Darych, line 209: Does old version of edge has the same `tx.start_timestamp`? What it it was created in previous txs?

src/storage/v2/inmemory/storage.cpp
```diff
@@ -558,6 +557,13 @@ Result<EdgeAccessor> InMemoryStorage::InMemoryAccessor::EdgeSetFrom(EdgeAccessor
         CreateAndLinkDelta(&transaction_, to_vertex, Delta::RemoveInEdgeTag(), edge_type, new_from_vertex, edge_ref);
         to_vertex->in_edges.emplace_back(edge_type, new_from_vertex, edge_ref);
 
+        if (storage_->storage_mode_ != StorageMode::ON_DISK_TRANSACTIONAL) {
```
Darych, line 560: Do we really need this check in if? If we are here it means we are in-memory.

gvolfing: We don't need this check good catch! Also removed it from a couple of lines below.

src/storage/v2/inmemory/edge_type_index.cpp
gvolfing: It is always created in a previous one. Good catch. Fixed.

Darych (approved):
LGTM

PR #1547, opened 2023-11-28
SHOW DATABASES query improvement

andrejtonev:
`SHOW DATABASES` query returns a matrix with 2 columns: "Name", "Current".
Each row represents one database, with the "Current" being a simple flag (`*`) to show which database is currently the active one.

This is somewhat convoluted, especially when we just want to find our current database among many other databases.

This PR simplifies this by introducing a new query `SHOW DATABASE` which returns a single result - the name of the current database. While the `SHOW DATABASES` (note the `S` at the end) returns a list of names, without the "Current" field.

[master < Task] PR

### DIFF
--- a/src/query/frontend/ast/ast.hpp
+++ b/src/query/frontend/ast/ast.hpp
@@ -3553,7 +3553,7 @@ class MultiDatabaseQuery : public memgraph::query::Query {
 
   DEFVISITABLE(QueryVisitor<void>);
 
-  enum class Action { CREATE, USE, DROP };
+  enum class Action { CREATE, USE, DROP, SHOW };
 
   memgraph::query::MultiDatabaseQuery::Action action_;
   std::string db_name_;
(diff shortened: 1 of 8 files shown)


andrejtonev:
@tonilastre this is something I wanted to introduce for a while. Didn't get a chance to do it sooner.
I know you are using the `SHOW DATABASES` query, I think this change should simplify the LABs flow.
P.S.
Not sure when this will get merged. Need to sync with @gitbuda and you about the timeline.

gitbuda:
NOTE: we should not merge this to v2.13 release because it's breaking on the Lab side, I'll tmp reject PR just to not merge it accidentally

gitbuda (changes requested):
We just wait a bit

andrejtonev:
This will probably be merged into replication refactor part 7 and merged to master that way.

andrejtonev:
PR merged into #1550

PR #1548, opened 2023-11-28
Support multiple RUNs in explicit tx

as51340:
[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR

### DIFF
--- a/tests/drivers/python/v5_8/explicit_tx_multiple_run.py
+++ b/tests/drivers/python/v5_8/explicit_tx_multiple_run.py
@@ -8,14 +8,12 @@ driver = neo4j.GraphDatabase.driver(MEMGRAPH_URL, auth=("", ""))
 
 
 def fill_db(tx):
-    res = tx.run("UNWIND range(1, 2000) AS i CREATE (n:Node {id: i}) RETURN n")
+    tx.run("UNWIND range(1, 2000) AS i CREATE (n:Node {id: i}) RETURN n")
 
 
 def run_queries(tx):
-    big_result = tx.run("match (n) return n;")  # A query that forces the result to have has_more=true
-    short_result = tx.run("match (n) return n limit 1;")  # Any query you can run
-    print(f"Big result values: {big_result.values()}")
-    print(f"Short result values: {short_result.values()}")
+    tx.run("match (n) return n;")  # A query that forces the result to have has_more=true
+    tx.run("match (n) return n limit 1;")  # Any query you can run
 
 
 with driver.session() as session:
--- a/tests/unit/bolt_session.cpp
+++ b/tests/unit/bolt_session.cpp
@@ -1272,39 +1272,6 @@ TEST(BoltSession, ExplicitTxBeginRunPullCommit) {
   }
 }
 
-TEST(BoltSession, ExplicitTxBeginMultipleOrderedRunPullCommit) {
-  {
-    INIT_VARS;
-
-    ExecuteHandshake(input_stream, session, output, v4_3::handshake_req, v4_3::handshake_resp);
-    ExecuteInit(input_stream, session, output, true);
-    ExecuteBeginTransaction(input_stream, session, output);
-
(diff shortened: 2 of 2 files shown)


tests/drivers/python/v5_8/explicit_tx_multiple_run.py
```diff
@@ -0,0 +1,23 @@
...
+def run_queries(tx):
+    tx.run("match (n) return n;")  # A query that forces the result to have has_more=true
+    tx.run("match (n) return n limit 1;")  # Any query you can run
+
+
+with driver.session() as session:
+    tx = session.begin_transaction()
+    fill_db(tx)
+    run_queries(tx)
+    tx.commit()
```
andrejtonev, line 23: I can break this by running the following python script:
```
import neo4j

MEMGRAPH_URL = "bolt://localhost:7687"

driver = neo4j.GraphDatabase.driver(MEMGRAPH_URL)

def run_queries(tx):
    big_result = tx.run("match (n:Node) set n.id = 9999 return n;") #A query that forces the result to have has_more=true
    short_result = tx.run("match (n:Node) set n.id = 8888 return n;") #Any query you can run
    #print(short_result.values())
    print("\n\n\nBREAK\n\n\n")
    print(big_result.values())
    print(short_result.values())

with driver.session() as session:
    session.run("UNWIND RANGE(1,3000) AS idx CREATE (:Node{id:idx});")
    tx = session.begin_transaction()
    run_queries(tx)
    tx.commit()
```

Memgrapg trace:
```
[2023-11-29 15:56:44.630] [memgraph_log] [debug] [Run - memgraph] 'UNWIND RANGE(1,3000) AS idx CREATE (:Node{id:idx});'
PULL(0) -1
[2023-11-29 15:56:44.636] [memgraph_log] [debug] [Run - memgraph] 'UNWIND RANGE(3000,3100) AS idx CREATE (:Bla{id:idx});'
PULL(1) -1
[2023-11-29 15:56:44.636] [memgraph_log] [debug] [Run - memgraph] 'match (n:Node) set n.id = 9999 return n;'
RUN 0
PULL(2) -1
[2023-11-29 15:56:44.640] [memgraph_log] [debug] [Run - memgraph] 'match (n:Node) set n.id = 8888 return n;'
RUN 1
PULL(3) -1
PULL(4) 0
PULL(5) 0
PULL(6) -1
PULL(7) -1
```
The PULL message has two numbers; in the brackets an index that ++ on each call, at the end the qid (if not specified -1).

-1 just means do the last query. But we have interleaved PULLs, so I get a result that is a mix of 9999 and 8888 ids.

andrejtonev: Note, this breaks neo4j too. But I don't think we should support this in a way that can break our snapshot isolation.

andrejtonev (commented):
Our interpreter should be able to handle multiple RUN and PULL commands. Even out of order PULLs should work; although without any write order guarantees.

However, the bolt state machine needs substantial changes. 

Moreover, the drivers used combine RUN and PULL into one command, which means that the multiple RUNs before a single PULL is not actually something that a user can easily do. I think that handling interleaved PULLs is not something we can consider at this point.

TLDR: not sure if this "issue" should even be repaired, since no driver I know allows user to send multiple RUN commands and the neo4j implementation is broken.

tests/drivers/python/v5_8/explicit_tx_multiple_run.py
```diff
@@ -0,0 +1,23 @@
...
+
+driver = neo4j.GraphDatabase.driver(MEMGRAPH_URL, auth=("", ""))
+
+
+def fill_db(tx):
+    tx.run("UNWIND range(1, 2000) AS i CREATE (n:Node {id: i}) RETURN n")
+
+
+def run_queries(tx):
+    tx.run("match (n) return n;")  # A query that forces the result to have has_more=true
```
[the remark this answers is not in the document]
andrejtonev: Check comment below where I break this.

PR #1550, opened 2023-11-29
Replication refactor part 7

andrejtonev:
Multi-tenant replication

Behind the CMake option `MG_EXPERIMENTAL_REPLICATION_MULTITENANCY` functionality for MT replication is protected from general usage.
New Github workflow has been added so it can remain tested until full delivered for general usage.
On disk folder layout for MT has changed to be uuid based rather than name based. This is to deal with the fact REPLICA may need database to stick around long enough for read query to complete.
Durability migration code has been changed to accommodate these changes.
DMBS lookup is now also uuid based, which impacted RPC payloads used, which means the RPC version changed.
Storage config has changed as a subset of the config was important to pass from MAIN to REPLICA, so that created databases on REPLICA have same behavior.
Interpreter can now handle when no active database exists.
Introduction of system transactions which are separate to the database transactions, it new timestamp is what supports correct replication an recovery of CREATE/DROP of MT databases.

Notes for documentation changes needed:
- Multitenancy queries changes
  - `SHOW DATABASE;` is new and shows the current used database, NULL if there is not one.
  - `SHOW DATABASES;` now shows just the set of MT databases that exist.


src/dbms/dbms_handler.cpp
```diff
@@ -9,58 +9,214 @@
...
+    throw UnknownVersionException();
+  };
+
+  static auto GenKey(std::string_view name) -> std::string { return fmt::format("{}{}", kDBPrefix, name); }
+
+  static auto GenVal(utils::UUID uuid, std::filesystem::path rel_dir) {
+    nlohmann::json json;
+    json["uuid"] = uuid;
+    json["rel_dir"] = rel_dir;
+    // TODO: Serialize the configuration
```
Ignition, line 70: Finish the TODO

src/dbms/dbms_handler.cpp
```diff
@@ -9,58 +9,214 @@
...
+      const auto uuid = json.at("uuid").get<utils::UUID>();
+      const auto rel_dir = json.at("rel_dir").get<std::filesystem::path>();
+      spdlog::info("Restoring database {} at {}.", name, rel_dir);
+      auto new_db = New_(name, uuid, rel_dir);
+      MG_ASSERT(!new_db.HasError(), "Failed while creating database {}.", name);
+      directories.emplace(rel_dir.filename());
       spdlog::info("Database {} restored.", name);
     }
+    // Read the last timestamp
+    auto lcst = durability_->Get(kLSCTSKey);
```
Ignition, line 158: LSCTS -> LCSTS to reduce confusion, but still acronym isn't meaningful.

src/dbms/dbms_handler.cpp
```diff
@@ -9,58 +9,214 @@
...
+  SetupDefault_();
+
+  /*
+   * REPLICATION RECOVERY AND STARTUP
+   */
   // Startup replication state (if recovered at startup)
   auto replica = [this](replication::RoleReplicaData const &data) {
     // Register handlers
     InMemoryReplicationHandlers::Register(this, *data.server);
+    RegisterSystemRPC(data, *this);
```
Ignition, line 207: Consistency, maybe `*ReplicationHandlers::Register(...)`
SystemReplicationHandlers and MultiTenancyReplicationHandlers?

Ignition: Also maybe also/alternatively a single method to hold these all `RegisterReplicationHandlers(...)`

src/dbms/dbms_handler.cpp
```diff
@@ -69,7 +225,117 @@ DbmsHandler::DbmsHandler(
...
+
+  // Check if db exists
+  try {
+    // Low level handlers
+    if (!db_handler_.TryDelete(db_name)) {
+      return DeleteError::USING;
+    }
+  } catch (utils::BasicException &) {
+    return DeleteError::NON_EXISTENT;
+  }
```
Ignition, line 271: Do not use exception control flow here

src/dbms/dbms_handler.cpp
```diff
@@ -69,7 +225,117 @@ DbmsHandler::DbmsHandler(
...
+    database.streams()->DropAll();
+    database.thread_pool()->Shutdown();
+  }
+
+  // Remove from durability list
+  if (durability_) durability_->Delete(Durability::GenKey(db_name));
+
+  // Check if db exists
+  // Low level handlers
+  db_handler_.DeferDelete(db_name, [storage_path = *storage_path, db_name]() {
```
Ignition, line 318: db_name is string_view, danger in lambda capture

src/dbms/dbms_handler.hpp
```diff
@@ -48,6 +47,8 @@
 
 namespace memgraph::dbms {
 
+constexpr std::string_view kLSCTSKey = "last_commited_system_ts";
```
Ignition, line 50: Can stuff be moved to cpp? This implementation detail isn't needed to be public in other translation units.

src/dbms/dbms_handler.hpp
```diff
@@ -131,8 +133,52 @@ class DbmsHandler {
...
+          spdlog::debug("Different UUIDs");
+
+          // TODO: Fix this hack
+          if (config.name == kDefaultDB) {
+            spdlog::debug("Update default db's UUID");
+            // Default db cannot be deleted and remade, have to just update the UUID
+            db->storage()->config_.salient.uuid = config.uuid;
+            UpdateDurability(db->storage()->config_, ".");
+            return db;
+          }
```
Ignition, line 167: What is writes already exist in this other default database. I would rather wipe the whole default database, giving it new storage.

src/dbms/dbms_handler.cpp
```diff
@@ -9,58 +9,214 @@
...
+    const auto &name = entry.path().filename().string();
+    if (entry.is_directory() && !name.empty() && name.front() != '.') {
+      auto itr = directories.find(name);
+      if (itr == directories.end()) {
+        std::filesystem::remove_all(entry);
+      } else {
+        directories.erase(itr);
+      }
     }
   }
```
Ignition, line 192: Is this safe? Some admin forgets to start with the right flag `recovery_on_startup == false` and production databases could be removed from filesystem.

src/dbms/dbms_handler.hpp
```diff
@@ -366,7 +406,10 @@ class DbmsHandler {
     std::shared_lock<LockT> rd(lock_);
     for (auto &[_, db_gk] : db_handler_) {
       auto db_acc = db_gk.access();
-      if (db_acc) {                   // This isn't an error, just a defunct db
+      if (!db_acc) continue;  // This isn't an error, just a defunct db
+      if constexpr (std::is_invocable_v<decltype(f), DatabaseAccess>) {
+        if (f(*db_acc)) break;  // Run until the first successful one
+      } else if constexpr (std::is_invocable_v<decltype(f), Database *>) {
```
Ignition, line 412: AFAICT `ForOne` is unused.

src/dbms/dbms_handler.hpp
```diff
@@ -379,6 +422,88 @@ class DbmsHandler {
...
+          // TODO: data race issue? registered_replicas_ access not protected
+          // This is sync in any case, as this is the startup
+          for (auto &client : main_data.registered_replicas_) {
+            try {
+              auto stream = client.rpc_client_.Stream<storage::replication::CreateDatabaseRpc>(
+                  std::string(main_data.epoch_.id()), system_transaction_->system_timestamp, delta.config);
+              const auto response = stream.AwaitResponse();
+              if (response.result == storage::replication::CreateDatabaseRes::Result::FAILURE) {
+                // This replica needs SYSTEM recovery
+                client.behind_ = true;
```
Ignition, line 452: TODO: change to actual state machine. BEHIND being one of the states.

src/query/interpreter.hpp
```diff
@@ -281,7 +243,34 @@ class Interpreter final {
 
   void SetUser(std::string_view username);
 
+  struct SystemTransaction {
+    explicit SystemTransaction(std::unique_lock<utils::ResourceLock> guard, dbms::DbmsHandler &dbms_handler)
+        : system_guard_(std::move(guard)), dbms_handler_{&dbms_handler} {
+      dbms_handler_->NewSystemTransaction();
```
Ignition, line 249: Strange that the:
- system lock is `interpreter_context_->system_lock`
- Held in guard that SystemTransaction holds, owned by Interpreter
- dbms_handler controls the actual timestamp values.

Is it possible to system_lock upto the dbms_handler (it will be more inline with where it will need to be for future replication groups)

src/dbms/dbms_handler.hpp
```diff
@@ -515,18 +628,35 @@ class DbmsHandler {
...
+  void RecoverReplication(DatabaseAccess db_acc) {
+    if (allow_mt_repl || db_acc->name() == dbms::kDefaultDB) {
+      // Handle global replication state
+      spdlog::info("Replication configuration will be stored and will be automatically restored in case of a crash.");
+      // RECOVER REPLICA CONNECTIONS
+      memgraph::dbms::RestoreReplication(repl_state_, std::move(db_acc));
+    } else if (const ::memgraph::replication::RoleMainData *data =
+                   std::get_if<::memgraph::replication::RoleMainData>(&repl_state_.ReplicationData());
+               data && !data->registered_replicas_.empty()) {
+      spdlog::warn("Multi-tenant replication is currently not supported!");
```
Ignition, line 642: TODO: we need to implement

src/dbms/dbms_handler.hpp
```diff
@@ -515,18 +628,35 @@ class DbmsHandler {
...
   mutable LockT lock_{utils::RWLock::Priority::READ};  //!< protective lock
   storage::Config default_config_;                     //!< Storage configuration used when creating new databases
   DatabaseHandler db_handler_;                         //!< multi-tenancy storage handler
   std::unique_ptr<kvstore::KVStore> durability_;       //!< list of active dbs (pointer so we can postpone its creation)
-  bool delete_on_drop_;                                //!< Flag defining if dropping storage also deletes its directory
-  std::set<std::string> defunct_dbs_;  //!< Databases that are in an unknown state due to various failures
 #endif
+  std::optional<SystemTransaction> system_transaction_;      //!< Current system transaction (only one at a time)
+  uint64_t system_timestamp_{storage::kTimestampInitialId};  //!< System timestamp
+  std::atomic_uint64_t last_commited_system_timestamp_{
```
Ignition, line 654: Is/will this used by CreateDatabaseHandler to check/ensure serialized create databases.

src/dbms/replication_client.cpp
```diff
@@ -10,25 +10,32 @@
...
 namespace memgraph::dbms {
 
 void StartReplicaClient(DbmsHandler &dbms_handler, replication::ReplicationClient &client) {
   // No client error, start instance level client
   auto const &endpoint = client.rpc_client_.Endpoint();
   spdlog::trace("Replication client started at: {}:{}", endpoint.address, endpoint.port);
-  client.StartFrequentCheck([&dbms_handler](std::string_view name) {
-    // Working connection, check if any database has been left behind
-    dbms_handler.ForEach([name](dbms::Database *db) {
+  client.StartFrequentCheck([&client, &dbms_handler](std::string_view name, bool reconnect) {
```
Ignition, line 21: `client.StartFrequentCheck([&client`
Can't we change the call arguments of the lambda to receive the client, rather than capture by reference the client.

src/dbms/replication_client.cpp
```diff
@@ -10,25 +10,32 @@
...
+    // Check if any database has been left behind
+    dbms_handler.ForEach([name, reconnect](dbms::DatabaseAccess db_acc) {
       // Specific database <-> replica client
-      db->storage()->repl_storage_state_.WithClient(name, [&](storage::ReplicationStorageClient *client) {
-        if (client->State() == storage::replication::ReplicaState::MAYBE_BEHIND) {
+      db_acc->storage()->repl_storage_state_.WithClient(name, [&](storage::ReplicationStorageClient *client) {
+        if (reconnect || client->State() == storage::replication::ReplicaState::MAYBE_BEHIND) {
           // Database <-> replica might be behind, check and recover
-          client->TryCheckReplicaStateAsync(db->storage());
+          client->TryCheckReplicaStateAsync(db_acc->storage(), db_acc);
```
Ignition, line 34: Why both `db_acc->storage(), db_acc`?

src/dbms/replication_client.cpp
```diff
@@ -10,25 +10,32 @@
...
+    if (reconnect) client.behind_ = true;
+#ifdef MG_ENTERPRISE
+    dbms_handler.SystemRestore(client);
+#endif
+    // Check if any database has been left behind
+    dbms_handler.ForEach([name, reconnect](dbms::DatabaseAccess db_acc) {
       // Specific database <-> replica client
-      db->storage()->repl_storage_state_.WithClient(name, [&](storage::ReplicationStorageClient *client) {
-        if (client->State() == storage::replication::ReplicaState::MAYBE_BEHIND) {
+      db_acc->storage()->repl_storage_state_.WithClient(name, [&](storage::ReplicationStorageClient *client) {
```
Ignition, line 31: If we have client, I don't think we need to refind it by name using WithClient

src/dbms/replication_client.cpp
Ignition: Ok I see, std::any

src/dbms/replication_handler.cpp
```diff
@@ -214,20 +226,26 @@ bool ReplicationHandler::IsReplica() const { return dbms_handler_.ReplicationSta
...
     // Each individual client has already been restored and started. Here we just go through each database and start its
     // client
     for (auto &instance_client : mainData.registered_replicas_) {
-      spdlog::info("Replica {} restoration started for {}.", instance_client.name_, storage.id());
+      spdlog::info("Replica {} restoration started for {}.", instance_client.name_, db_acc->name());
 
-      const auto &ret = storage.repl_storage_state_.replication_clients_.WithLock(
-          [&](auto &storage_clients) -> utils::BasicResult<RegisterReplicaError> {
+      // Phase 1: ensure DB exists
+      // auto s = instance_client.rpc_client_.Stream<memgraph::replication::????>();
```
Ignition, line 240: I'm unsure about the state machine at this point. I feel is client is system behind than that implies db is maybe behind

src/dbms/replication_handler.cpp
```diff
@@ -214,20 +226,26 @@ bool ReplicationHandler::IsReplica() const { return dbms_handler_.ReplicationSta
...
+
+      // Phase 2: ensure storage is in sync
+      const auto &ret = db_acc->storage()->repl_storage_state_.replication_clients_.WithLock(
+          [&, db_acc](auto &storage_clients) mutable -> utils::BasicResult<RegisterReplicaError> {
+            // ONLY VALID if storage.id() exists on REPLICA
```
Ignition, line 245: Out of date comment

src/dbms/dbms_handler.hpp
Ignition: Or some kind of check that the database is unused. ie. Check timestamp is 0.

src/dbms/dbms_handler.hpp
Ignition: I was incorrect, this will be cleaned up once we remove the Experimental build flag.

src/dbms/replication_handler.cpp
```diff
@@ -259,4 +277,101 @@ void RestoreReplication(replication::ReplicationState &repl_state, storage::Stor
...
+void CreateDatabaseHandler(DbmsHandler &dbms_handler, slk::Reader *req_reader, slk::Builder *res_builder) {
+  memgraph::storage::replication::CreateDatabaseReq req;
+  memgraph::slk::Load(&req, req_reader);
+
+  namespace sr = memgraph::storage::replication;
+  sr::CreateDatabaseRes res(sr::CreateDatabaseRes::Result::FAILURE);
+
+  // TODO
+  // Check epoch
+  // Check ts
```
Ignition, line 291: TODO

src/dbms/replication_handler.cpp
```diff
@@ -259,4 +277,101 @@ void RestoreReplication(replication::ReplicationState &repl_state, storage::Stor
...
+  try {
+    // Get all current dbs
+    auto old = dbms_handler.All();
+    // Check/create the incoming dbs
+    for (const auto &config : req.database_configs) {
+      try {
+        dbms_handler.Get(config.uuid);
+      } catch (const UnknownDatabaseException &) {
+        // Missing db
+        if (dbms_handler.New(config).HasError()) {
```
Ignition, line 332: Should be done using dbms_handler.GetOrNew() so that it happens under one lock

src/dbms/replication_handler.cpp
```diff
@@ -259,4 +277,101 @@ void RestoreReplication(replication::ReplicationState &repl_state, storage::Stor
...
+        }
+      }
+      const auto it = std::find(old.begin(), old.end(), config.name);
+      if (it != old.end()) old.erase(it);
+    }
+    // Delete all the leftover old dbs
+    for (const auto &remove_db : old) {
+      if (dbms_handler.Delete(remove_db).HasError()) {
+        spdlog::debug("Failed to drop database \"{}\".", remove_db);
+        throw;
```
Ignition, line 344: throw what? Also:
DeleteError::NON_EXISTENT is fine. Something else removed it.
DeleteError::DEFAULT_DB is logic error. We shouldn't ever ask to drop default.

src/dbms/replication_handler.cpp
```diff
@@ -259,4 +277,101 @@ void RestoreReplication(replication::ReplicationState &repl_state, storage::Stor
...
+      }
+    }
+    // Successfully recovered
+    res = sr::SystemRecoveryRes(sr::SystemRecoveryRes::Result::SUCCESS);
+  } catch (...) {
+    // Failure
```
Ignition, line 350: We can get into a partial application. Some configs applied making new DBs, or some deletions applied (but all required DBs exist)

src/dbms/transaction.hpp
```diff
@@ -0,0 +1,55 @@
...
+        : action(Action::CREATE_DATABASE), config(std::move(config)) {}
+
+    Delta(const Delta &) = delete;
+    Delta(Delta &&) = delete;
+    Delta &operator=(const Delta &) = delete;
+    Delta &operator=(Delta &&) = delete;
+
+    ~Delta() {
+      switch (action) {
+        case Action::CREATE_DATABASE:
```
Ignition, line 36: SalientConfig needs to be destroyed, it has a std::string

src/dbms/transaction.hpp
```diff
@@ -0,0 +1,55 @@
...
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#pragma once
+
+#include "storage/v2/config.hpp"
+
+namespace memgraph::dbms {
+struct SystemTransaction {
+  struct Delta {
```
Ignition, line 18: Unlike data Deltas, which represent undo action (so it can rewind to previous timestamp), this System change is just an operation deferred to commit time.

src/glue/SessionHL.cpp
```diff
@@ -167,10 +167,10 @@ std::map<std::string, memgraph::communication::bolt::Value> SessionHL::Discard(s
 std::map<std::string, memgraph::communication::bolt::Value> SessionHL::Pull(SessionHL::TEncoder *encoder,
                                                                             std::optional<int> n,
                                                                             std::optional<int> qid) {
-  // TODO: Update once interpreter can handle non-database queries (db_acc will be nullopt)
-  auto *db = interpreter_.current_db_.db_acc_->get();
   try {
-    TypedValueResultStream<TEncoder> stream(encoder, db->storage());
+    auto &db = interpreter_.current_db_.db_acc_;
+    auto *storage = db ? db->get()->storage() : nullptr;
```
Ignition, line 172: I've not followed why this was needed

Ignition: Can current_db_ have it accessor revoked proactively?

src/query/interpreter.cpp
```diff
@@ -786,8 +785,9 @@ Callback HandleReplicationQuery(ReplicationQuery *repl_query, const Parameters &
       callback.header = {
           "name", "socket_address", "sync_mode", "current_timestamp_of_replica", "number_of_timestamp_behind_master",
           "state"};
-      callback.fn = [handler = ReplQueryHandler{dbms_handler}, replica_nfields = callback.header.size()] {
-        const auto &replicas = handler.ShowReplicas();
+      callback.fn = [handler = ReplQueryHandler{dbms_handler}, replica_nfields = callback.header.size(),
+                     db_acc = current_db.db_acc_] {
+        const auto &replicas = handler.ShowReplicas(*db_acc->get());
```
Ignition, line 790: TODO: `SHOW REPLICAS;` should become DB agnostic. New query `SHOW REPLICAS STATUS;` would show the status of the replicas for the current DB.

src/query/interpreter.cpp
```diff
@@ -1414,6 +1421,7 @@ bool IsQueryWrite(const query::plan::ReadWriteTypeChecker::RWType query_type) {
 }  // namespace
 
 Interpreter::Interpreter(InterpreterContext *interpreter_context) : interpreter_context_(interpreter_context) {
+  // here
```
Ignition, line 1424: rogue comment

src/query/interpreter.cpp
```diff
@@ -1424,6 +1432,7 @@ Interpreter::Interpreter(InterpreterContext *interpreter_context) : interpreter_
 
 Interpreter::Interpreter(InterpreterContext *interpreter_context, memgraph::dbms::DatabaseAccess db)
     : current_db_{std::move(db)}, interpreter_context_(interpreter_context) {
+  // here????too late?
```
Ignition, line 1435: rogue comment

src/query/interpreter.cpp
```diff
@@ -3387,13 +3395,13 @@ PreparedQuery PrepareConstraintQuery(ParsedQuery parsed_query, bool in_explicit_
 
 PreparedQuery PrepareMultiDatabaseQuery(ParsedQuery parsed_query, CurrentDB &current_db,
                                         InterpreterContext *interpreter_context,
-                                        std::optional<std::function<void(std::string_view)>> on_change_cb) {
+                                        std::optional<std::function<void(std::string_view)>> on_change_cb
+                                        /*,
+                                        utils::ResourceLockGuard &system_guard*/) {
```
Ignition, line 3400: rogue comment

src/query/interpreter.cpp
```diff
@@ -3694,6 +3701,12 @@ Interpreter::PrepareResult Interpreter::Prepare(const std::string &query_string,
     SetupInterpreterTransaction(extras);
   }
 
+  // If this is the first query command within a transaction
```
Ignition, line 3704: rogue comment

src/query/interpreter.cpp
```diff
@@ -3760,9 +3797,37 @@ Interpreter::PrepareResult Interpreter::Prepare(const std::string &query_string,
...
+    //   // When we have a database and it is connected to the replica, everything should work as expected
+    //   // What happens when adding a database in SYNC vs ASYNC?
+    //   // What happens when deleting a database in SYNC vs ASYNC?
+    //   //   We cannot drop until all communication between the database and replica have stopped.
+    //
+    //   // Send an drop rpc req, the replica is using the db so it doesn't drop???
+    //   // Create a new db with same name. <- different ID?
+    //
+    // } else {
+    //   system_shared.lock();
```
Ignition, line 3830: Review behavior/remove comment

src/query/interpreter.cpp
```diff
@@ -4025,10 +4093,22 @@ void Interpreter::Commit() {
   // We should document clearly that all results should be pulled to complete
   // a query.
   current_transaction_.reset();
-  if (!current_db_.db_transactional_accessor_) return;
+  if (!current_db_.db_transactional_accessor_ || !current_db_.db_acc_) {
+    // No database nor db transaction; check for system transaction
+    if (!system_transaction) return;
+    // TODO auto expected = TransactionStatus::SYSTEM;
```
Ignition, line 4099: Do we need to remove the SYSTEM enum?

src/query/stream/streams.cpp
```diff
@@ -644,6 +644,25 @@ void Streams::Drop(const std::string &stream_name) {
...
+      // By acquiring the write lock here for the consumer, we make sure there is
+      // no running Test function for this consumer, therefore it can be erased.
+      std::visit([&](const auto &stream_data) { stream_data.stream_source->Lock(); }, stream);
+      if (!storage_.Delete(name)) {
+        durability_ok = false;
+      }
+    }
+
+    streams.clear();
+    return durability_ok;  // TODO: do we need special case for this cleanup if false
```
Ignition, line 662: return ATM is ignored

src/storage/v2/replication/replication_client.cpp
```diff
@@ -29,14 +28,27 @@ namespace memgraph::storage {
...
   auto &replStorageState = storage->repl_storage_state_;
-  auto stream{client_.rpc_client_.Stream<replication::HeartbeatRpc>(
-      storage->id(), replStorageState.last_commit_timestamp_, std::string{replStorageState.epoch_.id()})};
+  replication::HeartbeatRes replica;
+
+  std::optional<rpc::Client::StreamHandler<replication::HeartbeatRpc>> hb_stream{};
+
+  hb_stream.emplace(client_.rpc_client_.Stream<replication::HeartbeatRpc>(
+      storage->uuid(), replStorageState.last_commit_timestamp_, std::string{replStorageState.epoch_.id()}));
+  replica = hb_stream->AwaitResponse();
```
Ignition, line 41: What?! AFAICT no need for optional stream + HeartbeatRes assignment

src/storage/v2/replication/replication_storage_state.hpp
```diff
@@ -39,13 +39,13 @@ class ReplicationStorageClient;
...
   // Only MAIN can send
-  void InitializeTransaction(uint64_t seq_num, Storage *storage);
+  void InitializeTransaction(uint64_t seq_num, Storage *storage, std::any gk);
   void AppendDelta(const Delta &delta, const Vertex &vertex, uint64_t timestamp);
   void AppendDelta(const Delta &delta, const Edge &edge, uint64_t timestamp);
   void AppendOperation(durability::StorageMetadataOperation operation, LabelId label,
                        const std::set<PropertyId> &properties, const LabelIndexStats &stats,
                        const LabelPropertyIndexStats &property_stats, uint64_t final_commit_timestamp);
-  bool FinalizeTransaction(uint64_t timestamp, Storage *storage);
+  bool FinalizeTransaction(uint64_t timestamp, Storage *storage, std::any gk);
```
Ignition, line 48: TODO: rename `gk` to be more accurate, its the access not the gatekeeper, maybe also introduce a more precise type than std::any

src/storage/v2/replication/rpc.cpp
```diff
@@ -214,43 +241,110 @@ void Load(memgraph::storage::replication::HeartbeatRes *self, memgraph::slk::Rea
...
+  memgraph::slk::Save(self.items.properties_on_edges, builder);
+  memgraph::slk::Save(self.items.enable_schema_metadata, builder);
+}
+
+void Load(memgraph::storage::SalientConfig *self, memgraph::slk::Reader *reader) {
+  memgraph::slk::Load(&self->name, reader);
+  memgraph::slk::Load(&self->uuid, reader);
+  uint8_t sm = 0;
+  memgraph::slk::Load(&sm, reader);
+  self->storage_mode = static_cast<storage::StorageMode>(sm);
```
Ignition, line 296: Maybe we should add some utility assistance to help with enums.

src/storage/v2/storage.hpp
```diff
@@ -216,8 +233,7 @@ class Storage {
     virtual ConstraintsInfo ListAllConstraints() const = 0;
 
     // NOLINTNEXTLINE(google-default-arguments)
-    virtual utils::BasicResult<StorageManipulationError, void> Commit(
-        std::optional<uint64_t> desired_commit_timestamp = {}, bool is_main = true) = 0;
+    virtual utils::BasicResult<StorageManipulationError, void> Commit(CommitReplArgs reparg = {}, std::any gk = {}) = 0;
```
Ignition, line 236: `gk` should be non-defaulted first argument.

src/utils/gatekeeper.hpp
```diff
@@ -159,55 +179,62 @@ struct Gatekeeper {
...
+    }
 
     void reset() {
       if (owner_) {
         {
           auto guard = std::unique_lock{owner_->mutex_};
           --owner_->count_;
         }
-        owner_->cv_.notify_all();
+        owner_->cv_.notify_all();  // TODO Under a lock???
```
Ignition, line 209: comment not needed
> The notifying thread does not need to hold the lock on the same mutex as the one held by the waiting thread(s). Doing so may be a pessimization, since the notified thread would immediately block again, waiting for the notifying thread to release the lock, though some implementations recognize the pattern and do not attempt to wake up the thread that is notified under lock.

tests/e2e/run.sh
```diff
@@ -25,7 +25,7 @@ if [ "$#" -eq 0 ]; then
   # NOTE: If you want to run all tests under specific folder/section just
   # replace the dot (root directory below) with the folder name, e.g.
   # `--workloads-root-directory replication`.
-  python3 runner.py --workloads-root-directory .
+  python3 runner.py --workloads-root-directory ../../build
```
Ignition, line 28: Use relative path to `$SCRIPT_DIR`

src/dbms/dbms_handler.cpp
```diff
@@ -275,18 +261,50 @@ DbmsHandler::DeleteResult DbmsHandler::TryDelete(std::string_view db_name) {
...
+DbmsHandler::DeleteResult DbmsHandler::Delete(utils::UUID uuid) {
+  auto wr = std::lock_guard(lock_);
+  std::string db_name;
+  try {
+    const auto db = Get_(uuid);
+    db_name = db->name();
+  } catch (const UnknownDatabaseException &) {
+    return DeleteError::NON_EXISTENT;
+  }
+  return Delete_(db_name);
```
Ignition, line 291: TODO: new overload of Delete_ that works with db we already have (no need to re-Get by name)

src/dbms/dbms_handler.cpp
andrejtonev: Not sure how to handle differences between the saved config and incoming command line flags.
This is a new feature and we need to decide how to handle it.

src/dbms/dbms_handler.cpp
andrejtonev: Added a TODO

src/dbms/dbms_handler.hpp
andrejtonev: We now make sure the database is clean before updating UUID.
It will fail otherwise.

src/dbms/handler.hpp
```diff
@@ -95,6 +95,36 @@ class Handler {
...
+      // Delete the database now
+      db_acc->reset();
+      post_delete_func();
+    } else {
+      // Defer deletion
+      db_acc->reset();
+      std::thread([gk = std::move(itr->second), post_delete_func = std::forward<Func>(post_delete_func)] {
+        gk.~Gatekeeper<T>();
+        post_delete_func();
+      }).detach();
```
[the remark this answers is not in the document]
andrejtonev: Added a ThreadPool to handle it.

src/dbms/dbms_handler.cpp
andrejtonev: This is in line with how we handle other similar situations.

src/dbms/dbms_handler.hpp
andrejtonev: Removed

src/query/interpreter.hpp
andrejtonev: Moved to DBMS

src/dbms/dbms_handler.hpp
andrejtonev: At the moment not used by the replica, but main should be already setup.

src/dbms/replication_client.cpp
andrejtonev: Different clients.
Maybe need to change names.

src/dbms/replication_handler.cpp
andrejtonev: Not sure.
We could set all database replication state to behind. Perhaps not needed if the replica has recovery and no delta has been sent in the meantime.
However the show replicas state is misleading.

src/dbms/replication_handler.cpp
andrejtonev: Using the Update

src/dbms/replication_handler.cpp
andrejtonev: Moved the slk save to OnExit, so I can just return.

src/dbms/replication_handler.cpp
andrejtonev: There is a TODO to remove the exception.
For the partial success, I would rather just fail for now.

src/dbms/transaction.hpp
andrejtonev: We use the term Delta for both MVCC(undo action) and WAL(do action). Lets keep it as is for now.

src/glue/SessionHL.cpp
andrejtonev: In cases where we are execution a non-db query we must check before using the storage.

src/query/interpreter.cpp
andrejtonev: SHOW REPLICA refactor will be done in another PR.

src/query/stream/streams.cpp
andrejtonev: I'll leave the TODO as is

src/storage/v2/replication/replication_client.cpp
andrejtonev: Sorry, left from an older attempt at recovery

src/dbms/replication_handler.cpp
andrejtonev: Lets deliver this as part of the future `SHOW REPLICAS` PR.

src/query/interpreter.cpp
andrejtonev: I would like to have a way to distinguish between system and data transactions.
Removed for now.

src/dbms/replication_client.cpp
andrejtonev: Gonna skip this for now.

src/dbms/dbms_handler.hpp
andrejtonev: Will resolve this comment, since there is another comment pointing to the same issue.

src/dbms/replication_handler.cpp
```diff
@@ -137,29 +142,36 @@ auto ReplicationHandler::RegisterReplica(const memgraph::replication::Replicatio
       case memgraph::replication::RegisterReplicaError::SUCCESS:
         break;
     }
+  }
 
   if (!allow_mt_repl && dbms_handler_.All().size() > 1) {
     spdlog::warn("Multi-tenant replication is currently not supported!");
+  } else {
+#ifdef MG_ENTERPRISE
```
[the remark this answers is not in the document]
andrejtonev: The 2 flags are not identical.
One is a feature flag, the other one a compilation flag.

src/dbms/dbms_handler.cpp
Ignition: Decide to leave as is. Problem we will come back to in the future.

tests/e2e/replication_experimental/multitenancy.py
```diff
@@ -0,0 +1,1508 @@
...
+
+    # 3/
+    interactive_mg_runner.kill(MEMGRAPH_INSTANCES_DESCRIPTION_WITH_RECOVERY, "replica_1")
+    safe_execute(execute_and_fetch_all, main_cursor, "USE DATABASE A;")
+    safe_execute(execute_and_fetch_all, main_cursor, "CREATE (:Node{on:'A'});")
+    safe_execute(execute_and_fetch_all, main_cursor, "USE DATABASE B;")
+    safe_execute(execute_and_fetch_all, main_cursor, "CREATE (:Node{on:'B'});")
+    interactive_mg_runner.start(MEMGRAPH_INSTANCES_DESCRIPTION_WITH_RECOVERY, "replica_1")
+
+    time.sleep(3)  # Wait for the frequent check to get called
```
antoniofilipovic, line 922: Instead of time.sleep I would advise on using mg_sleep_and_assert, that way we are not dependent on magic number

tests/e2e/replication_experimental/multitenancy.py
```diff
@@ -0,0 +1,1508 @@
...
+    actual_data = execute_and_fetch_all(cursor_replica, "MATCH (n) RETURN count(*);")
+    assert actual_data[0][0] == 8  # seven node
+    actual_data = execute_and_fetch_all(cursor_replica, "MATCH ()-[r]->() RETURN count(*);")
+    assert actual_data[0][0] == 3  # three relationships
+
+    execute_and_fetch_all(cursor_replica, "USE DATABASE B;")
+    actual_data = execute_and_fetch_all(cursor_replica, "MATCH (n) RETURN count(*);")
+    assert actual_data[0][0] == 3  # two node
+    actual_data = execute_and_fetch_all(cursor_replica, "MATCH ()-[r]->() RETURN count(*);")
+    assert actual_data[0][0] == 0  # zero relationships
```
antoniofilipovic, line 936: Same comment as above, but here also comments are not in sync with numbers and I would remove comments here

tests/e2e/replication_experimental/multitenancy.py
```diff
@@ -0,0 +1,1508 @@
...
```
antoniofilipovic, line 992: I would parametrize test above on replica name

tests/e2e/replication_experimental/multitenancy.py
```diff
@@ -0,0 +1,1508 @@
...
+        failed = True
+    assert failed
+
+
+# start main and replica
+# add databases
+# kill main
+# start main with replication state recovery, no data recovery
+# note: just for future-proofing
+# ???
```
antoniofilipovic, line 1503: What is this comment here?

tests/e2e/replication_experimental/multitenancy.py
```diff
@@ -0,0 +1,1508 @@
...
+    # 2/
+    execute_and_fetch_all(main_cursor, "USE DATABASE A;")
+    execute_and_fetch_all(main_cursor, "CREATE (:Node{on:'A'});")
+    execute_and_fetch_all(main_cursor, "CREATE (:Node)-[:EDGE]->(:Node)")
+    execute_and_fetch_all(main_cursor, "CREATE (:Node)-[:EDGE]->(:Node)")
+    execute_and_fetch_all(main_cursor, "CREATE (:Node)-[:EDGE]->(:Node)")
+
+    execute_and_fetch_all(main_cursor, "USE DATABASE B;")
+    execute_and_fetch_all(main_cursor, "CREATE (:Node{on:'B'});")
+    execute_and_fetch_all(main_cursor, "CREATE (:Node{on:'B'});")
```
antoniofilipovic, line 1021: Since you are using mostly same queries maybe you can isolate this into function for setting up main

tests/e2e/replication_experimental/multitenancy.py
```diff
@@ -0,0 +1,1508 @@
...
+    def retrieve_data():
+        execute_and_fetch_all(main_cursor, "USE DATABASE B;")
+        return set(execute_and_fetch_all(main_cursor, "SHOW REPLICAS;"))
+
+    expected_data = {
+        ("replica_1", "127.0.0.1:10001", "sync", 3, 0, "ready"),
+        ("replica_2", "127.0.0.1:10002", "async", 3, 0, "ready"),
+    }
+    actual_data = mg_sleep_and_assert(expected_data, retrieve_data)
+    assert actual_data == expected_data
```
antoniofilipovic, line 1048: No need for assert here as this function does the checking

tests/e2e/replication_experimental/multitenancy.py
```diff
@@ -0,0 +1,1508 @@
...
+    actual_data = execute_and_fetch_all(cursor_replica, "MATCH (n) RETURN count(*);")
+    assert actual_data[0][0] == 7  # seven node
+    actual_data = execute_and_fetch_all(cursor_replica, "MATCH ()-[r]->() RETURN count(*);")
+    assert actual_data[0][0] == 3  # three relationships
+
+    execute_and_fetch_all(cursor_replica, "USE DATABASE B;")
+    actual_data = execute_and_fetch_all(cursor_replica, "MATCH (n) RETURN count(*);")
+    assert actual_data[0][0] == 2  # two node
+    actual_data = execute_and_fetch_all(cursor_replica, "MATCH ()-[r]->() RETURN count(*);")
+    assert actual_data[0][0] == 0  # zero relationships
```
antoniofilipovic, line 1130: This test seems same as above one, just name is different, maybe you can parametrize it?

tests/e2e/replication_experimental/multitenancy.py
```diff
@@ -0,0 +1,1508 @@
...
+    execute_and_fetch_all(main_cursor, "USE DATABASE A;")
+    execute_and_fetch_all(main_cursor, "CREATE (:Node{on:'A'});")
+    execute_and_fetch_all(main_cursor, "USE DATABASE B;")
+    execute_and_fetch_all(main_cursor, "CREATE (:Node{on:'B'});")
+
+    # 4/
+    cursor_replica = connection(7688, "replica_1").cursor()
+    execute_and_fetch_all(cursor_replica, "USE DATABASE A;")
+    actual_data = execute_and_fetch_all(cursor_replica, "MATCH (n) RETURN count(*);")
+    assert actual_data[0][0] == 8  # seven node
```
antoniofilipovic, line 1178: Comments are not following code here too

tests/e2e/replication_experimental/multitenancy.py
```diff
@@ -0,0 +1,1508 @@
...
+    actual_data = mg_sleep_and_assert(expected_data, retrieve_data)
+    assert actual_data == expected_data
+
+    # 4/
+    execute_and_fetch_all(main_cursor, "USE DATABASE memgraph;")
+    execute_and_fetch_all(main_cursor, "DROP DATABASE A;")
+    execute_and_fetch_all(main_cursor, "DROP DATABASE B;")
+
+    # 5/
+    expected_data = execute_and_fetch_all(main_cursor, "SHOW DATABASES;")
```
antoniofilipovic, line 1240: If there is bug here, it will be hard to find it, I would rather advise to have expected data as dictionary here

tests/e2e/replication_experimental/multitenancy.py
```diff
@@ -0,0 +1,1508 @@
...
+    # 4/
+    expected_data = execute_and_fetch_all(main_cursor, "SHOW DATABASES;")
+
+    def retrieve_data(cursor):
+        return execute_and_fetch_all(cursor, "SHOW DATABASES;")
+
+    actual_data = mg_sleep_and_assert(expected_data, partial(retrieve_data, connection(7688, "replica_1").cursor()))
+    assert actual_data == expected_data
+    actual_data = mg_sleep_and_assert(expected_data, partial(retrieve_data, connection(7689, "replica_2").cursor()))
+    assert actual_data == expected_data
```
antoniofilipovic, line 1340: This seems similar astest above, I would parametrize it on name of replica

tests/e2e/replication_experimental/multitenancy.py
```diff
@@ -0,0 +1,1508 @@
...
+    execute_and_fetch_all(replica1_cursor, "BEGIN")
+    execute_and_fetch_all(replica2_cursor, "USE DATABASE A;")
+
+    execute_and_fetch_all(main_cursor, "USE DATABASE memgraph;")
+    execute_and_fetch_all(main_cursor, "DROP DATABASE A;")
+
+    # 5/
+    # TODO Remove this once there is a replica state for the system
+    execute_and_fetch_all(main_cursor, "CREATE DATABASE B;")
+    execute_and_fetch_all(main_cursor, "USE DATABASE B;")
```
antoniofilipovic, line 1391: Do you need create database B query here? It seems it is not used

tests/e2e/replication_experimental/multitenancy.py
```diff
@@ -0,0 +1,1508 @@
...
+
+    expected_data = {
+        ("replica_1", "127.0.0.1:10001", "sync", 0, 0, "ready"),
+        ("replica_2", "127.0.0.1:10002", "async", 0, 0, "ready"),
+    }
+    actual_data = mg_sleep_and_assert(expected_data, retrieve_data)
+    assert actual_data == expected_data
+
+    # 6/
+    assert execute_and_fetch_all(replica1_cursor, "MATCH(n) RETURN count(*);")[0][0] == 1  # one node
```
antoniofilipovic, line 1404: Unnecessary comment here

tests/e2e/replication_experimental/multitenancy.py
```diff
@@ -0,0 +1,1508 @@
...
+    except mgclient.DatabaseError:
+        failed = True
+    assert failed
+
+    failed = False
+    try:
+        execute_and_fetch_all(replica2_cursor, "MATCH(n) RETURN n;")
+    except mgclient.DatabaseError:
+        failed = True
+    assert failed
```
antoniofilipovic, line 1495: Why this fails if database is still there, it was dropped and created, right?

tests/e2e/set_properties/CMakeLists.txt
```diff
@@ -6,3 +6,5 @@ copy_set_properties_e2e_python_files(common.py)
 copy_set_properties_e2e_python_files(set_properties.py)
 
 add_subdirectory(procedures)
+
+copy_e2e_files(set_properties workloads.yaml)
```
antoniofilipovic, line 10: Why are we copying workloads.yml suddenly?

tests/e2e/run.sh
```diff
@@ -25,7 +25,7 @@ if [ "$#" -eq 0 ]; then
   # NOTE: If you want to run all tests under specific folder/section just
   # replace the dot (root directory below) with the folder name, e.g.
   # `--workloads-root-directory replication`.
-  python3 runner.py --workloads-root-directory .
+  python3 runner.py --workloads-root-directory "$SCRIPT_DIR/../../build"
```
antoniofilipovic, line 28: Why has this changed?

tests/unit/multi_tenancy.cpp
```diff
@@ -0,0 +1,390 @@
...
+std::set<std::string> GetDirs(auto path) {
+  std::set<std::string> dirs;
+  // Clean the unused directories
+  for (const auto &entry : std::filesystem::directory_iterator(path)) {
+    const auto &name = entry.path().filename().string();
+    if (entry.is_directory() && !name.empty() && name.front() != '.') {
+      dirs.emplace(name);
+    }
+  }
+  return dirs;
```
antoniofilipovic, line 56: Can we isolate this function into header file, it seems duplicated as I have seen it on one more place

tests/unit/multi_tenancy.cpp
```diff
@@ -0,0 +1,390 @@
...
+  auto &dbms = DBMS();
+  ASSERT_FALSE(dbms.New("db1").HasError());
+  ASSERT_FALSE(dbms.New("db2").HasError());
+  ASSERT_FALSE(dbms.New("db3").HasError());
+  ASSERT_FALSE(dbms.New("db4").HasError());
+
+  // 3
+  auto use = [&](auto &interpreter, const std::string &name, std::string_view res) {
+    RunMtQuery(interpreter, "USE DATABASE " + name, res);
+  };
```
antoniofilipovic, line 197: This can be maybe part of MultiTenantTest function? It is used also in test above

antoniofilipovic (changes requested):
Sendinf first batch of comments, from tests only multi_tenancy is not fully reviewed yet, will review that again tomorrow with src files

tests/e2e/replication_experimental/multitenancy.py
Ignition: Important to note: They are semantically different databases.

If you have a session on REPLICA for database X, but database X is dropped from RPC given by MAIN. Then the session, once command has committed/rolledback, is now no longer associated with that database X. 
Hence any query requiring a database should DatabaseError.
If MAIN creates X it is, and should, be considered a different database. It would be problematic for that session to allow the user to continue on the new database as if it was the old one.

tests/e2e/set_properties/CMakeLists.txt
Ignition: Need to handle running e2e tests based on what is built.

Before we would run all tests in the source test files, but `MG_EXPERIMENTAL_REPLICATION_MULTITENANCY` is required for these tests to pass. Hence now only workloads which should pass are copied to build directory, and we run based on them.

tests/e2e/run.sh
Ignition: https://github.com/memgraph/memgraph/pull/1550#discussion_r1452583018

src/query/interpreter.cpp
```diff
@@ -4025,10 +4050,22 @@ void Interpreter::Commit() {
   // We should document clearly that all results should be pulled to complete
   // a query.
   current_transaction_.reset();
-  if (!current_db_.db_transactional_accessor_) return;
+  if (!current_db_.db_transactional_accessor_ || !current_db_.db_acc_) {
+    // No database nor db transaction; check for system transaction
+    if (!system_transaction) return;
+    // TODO Distinguish between data and system transaction states
+    auto expected = TransactionStatus::ACTIVE;
+    if (!transaction_status_.compare_exchange_weak(expected, TransactionStatus::STARTED_COMMITTING))
```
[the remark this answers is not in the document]
andrejtonev: yes! good catch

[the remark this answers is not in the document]
Ignition: Good observation. We are now carefully rethinking TransactionStatus transitions, SYSTEM txns can not be killable. As they work ATM they do their work immediately, including any durability they do. That means a system commit does only replication + system timestamp stuff.

Hence system transactions if requested to be killed should do nothing. None of the operation ATM checked if it must abort, and at system commit time regardless of txn status it must do replication + system timestamp stuff. We just have to ensure transaction_status_ correctly get back to IDLE.

src/dbms/replication_handler.cpp
```diff
@@ -137,29 +136,36 @@ auto ReplicationHandler::RegisterReplica(const memgraph::replication::Replicatio
...
+    all_clients_good &= storage->repl_storage_state_.replication_clients_.WithLock(
+        [storage, &instance_client, db_acc = std::move(db_acc)](auto &storage_clients) mutable {  // NOLINT
           auto client = std::make_unique<storage::ReplicationStorageClient>(*instance_client.GetValue());
-          client->Start(storage);
+          // All good, start replica client
+          client->Start(storage, std::move(db_acc));
           // After start the storage <-> replica state should be READY or RECOVERING (if correctly started)
           // MAYBE_BEHIND isn't a statement of the current state, this is the default value
-          // Failed to start due to branching of MAIN and REPLICA
+          // Failed to start due an error like branching of MAIN and REPLICA
```
antoniofilipovic, line 168: Isn't it that in the tests you allow branching? I saw test if I remember correctly that you can have replica which has different data than main because something was done prior to becoming replica and it is allowed to register that replica on main?

src/dbms/dbms_handler.cpp
```diff
@@ -11,56 +11,204 @@
...
+    auto end = durability_->end(std::string(kDBPrefix));
+    for (; it != end; ++it) {
+      const auto &[key, config_json] = *it;
+      const auto name = key.substr(kDBPrefix.size());
+      auto json = nlohmann::json::parse(config_json);
+      const auto uuid = json.at("uuid").get<utils::UUID>();
+      const auto rel_dir = json.at("rel_dir").get<std::filesystem::path>();
+      spdlog::info("Restoring database {} at {}.", name, rel_dir);
+      auto new_db = New_(name, uuid, rel_dir);
+      MG_ASSERT(!new_db.HasError(), "Failed while creating database {}.", name);
```
antoniofilipovic, line 154: Can we also output here the error

src/dbms/replication_handler.cpp
andrejtonev: This is tricky with multi-tenancy.
What this comment refers to is the storage replication client. Which handles only the replication and the replica state of one storage.
When registering we check if the storage on MAIN and REPLICA have branched. 
Now, where it gets complicated... When looking at tenants (list of all storage), we check their UUIDs and if they don't match (even if the names are the same) we handle them as different. 
From discussing on slack, we decided to force the REPLICA to follow MAIN's tenant list. Meaning if we have a tenant named "db" on MAIN and REPLICA, but the "db" on REPLICA was created separately from the current MAIN, we delete it and create a new one that completely follows the MAIN's history and so no branching.

TLDR;
System and storage replication differs in that the system is forcefully updated on REPLICAs (don't care about branching), while the storage replication cares about branching at registration time. A storage on MAIN and REPLICA are connected via the UUID, so a tenant with the same name but a different UUID is handled as a different tenant.

src/dbms/handler.hpp
```diff
@@ -120,6 +154,7 @@ class Handler {
  private:
   std::unordered_map<std::string, utils::Gatekeeper<T>, string_hash, std::equal_to<>>
       items_;  //!< map to all active items
```
antoniofilipovic, line 156: What do these items_ do, is there better naming for it?

tests/e2e/replication_experimental/conftest.py
```diff
@@ -0,0 +1,43 @@
...
+# The fixture here is more complex because the connection has to be
+# parameterized based on the test parameters (info has to be available on both
+# sides).
+#
+# https://docs.pytest.org/en/latest/example/parametrize.html#indirect-parametrization
+# is not an elegant/feasible solution here.
+#
+# The solution was independently developed and then I stumbled upon the same
+# approach here https://stackoverflow.com/a/68286553/4888809 which I think is
+# optimal.
```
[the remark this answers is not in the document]
andrejtonev: Copy-pasted from another e2e test.
Can remove if not needed.

tests/e2e/replication_experimental/multitenancy.py
```diff
@@ -0,0 +1,1508 @@
...
+    execute_and_fetch_all(cursor, "SET REPLICATION ROLE TO REPLICA WITH PORT 10002;")
+    # Setup main
+    cursor = connection(7687, "main").cursor()
+    execute_and_fetch_all(cursor, "REGISTER REPLICA replica_1 SYNC TO '127.0.0.1:10001';")
+    execute_and_fetch_all(cursor, "REGISTER REPLICA replica_2 ASYNC TO '127.0.0.1:10002';")
+
+
+def test_manual_databases_create_multitenancy_replication(connection):
+    # Goal: to show that replication can be established against REPLICA which already
+    # has the database we need (which was unused so far)
```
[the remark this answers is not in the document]
andrejtonev: I'll update the comment, but this is important. The db is clean on replica. This is a different situation from a db with the same name and data on it.

src/query/interpreter.cpp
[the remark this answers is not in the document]
andrejtonev: We now ignore the initial state and just update i to IDLE using the while compare exchange weak.
Will let you resolve the comment if you think it's okay this way.

src/dbms/handler.hpp
andrejtonev: This is a generic handler that stores anything. 
Can't think of a better name. `objects` seems as vague as `items` :shrug:

tests/e2e/replication_experimental/multitenancy.py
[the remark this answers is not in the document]
andrejtonev: Updated the comments

tests/e2e/replication_experimental/multitenancy.py
```diff
@@ -0,0 +1,1508 @@
...
+    try:
+        execute_and_fetch_all(cursor, "REGISTER REPLICA replica_2 ASYNC TO '127.0.0.1:10002';")
+    except mgclient.DatabaseError:
+        failed = True
+    assert not failed
+
+
+def test_manual_databases_create_multitenancy_replication_main_behind(connection):
+    # Goal: to show that replication can be established against REPLICA which already
+    # has the database we need (which was unused so far)
```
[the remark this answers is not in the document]
andrejtonev: Updated comment

tests/e2e/replication_experimental/multitenancy.py
```diff
@@ -0,0 +1,1508 @@
...
+    def retrieve_data():
+        execute_and_fetch_all(main_cursor, "USE DATABASE A;")
+        return set(execute_and_fetch_all(main_cursor, "SHOW REPLICAS;"))
+
+    expected_data = {
+        ("replica_1", "127.0.0.1:10001", "sync", 7, 0, "ready"),
+        ("replica_2", "127.0.0.1:10002", "async", 7, 0, "ready"),
+    }
+    actual_data = mg_sleep_and_assert(expected_data, retrieve_data)
+    assert actual_data == expected_data
```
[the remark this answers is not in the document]
andrejtonev: You're right :+1:

tests/e2e/replication_experimental/multitenancy.py
```diff
@@ -0,0 +1,1508 @@
...
+
+    cursor_replica = connection(7688, "replica_1").cursor()
+    execute_and_fetch_all(cursor_replica, "USE DATABASE A;")
+    actual_data = execute_and_fetch_all(cursor_replica, "MATCH (n) RETURN count(*);")
+    assert actual_data[0][0] == 1  # one node
+    actual_data = execute_and_fetch_all(cursor_replica, "MATCH ()-[r]->() RETURN count(*);")
+    assert actual_data[0][0] == 0  # zero relationships
+
+
+def test_automatic_databases_create_multitenancy_replication_dirty_main(connection):
```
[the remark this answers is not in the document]
andrejtonev: It's just a different setup and at time of registration we need to handle different situations.

tests/e2e/replication_experimental/multitenancy.py
```diff
@@ -0,0 +1,1508 @@
...
+
+    execute_and_fetch_all(main_cursor, "USE DATABASE B;")
+    execute_and_fetch_all(main_cursor, "CREATE (:Node{on:'B'});")
+    execute_and_fetch_all(main_cursor, "CREATE (:Node{on:'B'});")
+
+    # 3/
+    interactive_mg_runner.kill(MEMGRAPH_INSTANCES_DESCRIPTION, "replica_1")
+    time.sleep(3)  # In order for the frequent check to run
+    interactive_mg_runner.start(MEMGRAPH_INSTANCES_DESCRIPTION, "replica_1")
+    time.sleep(3)  # In order for the frequent check to run
```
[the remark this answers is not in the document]
andrejtonev: Sleep is here to make sure the fc thread is ran. I removed all other sleep that are not necessary.

tests/e2e/replication_experimental/multitenancy.py
```diff
@@ -0,0 +1,1508 @@
...
+    execute_and_fetch_all(main_cursor, "CREATE (:Node)-[:EDGE]->(:Node)")
+    execute_and_fetch_all(main_cursor, "CREATE (:Node)-[:EDGE]->(:Node)")
+
+    execute_and_fetch_all(main_cursor, "USE DATABASE B;")
+    execute_and_fetch_all(main_cursor, "CREATE (:Node{on:'B'});")
+    execute_and_fetch_all(main_cursor, "CREATE (:Node{on:'B'});")
+
+    # 3/
+    interactive_mg_runner.kill(MEMGRAPH_INSTANCES_DESCRIPTION, "replica_1")
+    interactive_mg_runner.start(MEMGRAPH_INSTANCES_DESCRIPTION, "replica_1")
```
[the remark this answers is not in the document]
andrejtonev: Here we want to do it so fast that the fc won't have a chance to detect the connection drop.
The state will be different and we need to handle a reconnect with and without drop detection.

tests/e2e/replication_experimental/multitenancy.py
```diff
@@ -0,0 +1,1508 @@
...
+    actual_data = execute_and_fetch_all(cursor_replica, "MATCH (n) RETURN count(*);")
+    assert actual_data[0][0] == 7  # seven node
+    actual_data = execute_and_fetch_all(cursor_replica, "MATCH ()-[r]->() RETURN count(*);")
+    assert actual_data[0][0] == 3  # three relationships
+
+    execute_and_fetch_all(cursor_replica, "USE DATABASE B;")
+    actual_data = execute_and_fetch_all(cursor_replica, "MATCH (n) RETURN count(*);")
+    assert actual_data[0][0] == 2  # two node
+    actual_data = execute_and_fetch_all(cursor_replica, "MATCH ()-[r]->() RETURN count(*);")
+    assert actual_data[0][0] == 0  # zero relationships
```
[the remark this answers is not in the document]
andrejtonev: I added a couple of helper functions, but I would leave the exact checks for the test.

tests/e2e/replication_experimental/multitenancy.py
andrejtonev: Same comment as for the other sleep

tests/e2e/replication_experimental/multitenancy.py
andrejtonev: Updated

tests/e2e/replication_experimental/multitenancy.py
andrejtonev: It's so we need to do `SHOW REPLICA` on a particular db

tests/unit/multi_tenancy.cpp
andrejtonev: I'll just leave them as is. 
They are used in two different tests and we don't have a tests common directory of library.

src/dbms/dbms_handler.cpp
andrejtonev: What do you mean?

src/dbms/dbms_handler.cpp
```diff
@@ -11,56 +11,204 @@
...
+  for (const auto &entry : std::filesystem::directory_iterator(db_dir)) {
+    const auto &name = entry.path().filename().string();
+    if (entry.is_directory() && !name.empty() && name.front() != '.') {
+      auto itr = directories.find(name);
+      if (itr == directories.end()) {
+        std::filesystem::remove_all(entry);
+      } else {
+        directories.erase(itr);
+      }
     }
```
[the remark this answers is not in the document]
andrejtonev: Using the error_code variant of the remove_all. Other things shouldn't throw.

src/dbms/replication_handler.cpp
```diff
@@ -259,4 +265,177 @@ void RestoreReplication(replication::ReplicationState &repl_state, storage::Stor
       },
       repl_state.ReplicationData());
 }
+
+namespace system_replication {
+#ifdef MG_ENTERPRISE
+void SystemHeartbeatHandler(const uint64_t ts, slk::Reader *req_reader, slk::Builder *res_builder) {
+  replication::SystemHeartbeatReq req;
+  replication::SystemHeartbeatReq::Load(&req, req_reader);
+  memgraph::slk::Load(&req, req_reader);
```
antoniofilipovic, line 274: Do you need both these?

antoniofilipovic: Shouldn't SystemHeartbeatReq call slk in case it needs to call it?

src/dbms/database_handler.hpp
```diff
@@ -51,7 +51,7 @@ class DatabaseHandler : public Handler<Database> {
    * @param config Storage configuration
    * @return HandlerT::NewResult
    */
-  HandlerT::NewResult New(std::string_view name, storage::Config config, replication::ReplicationState &repl_state) {
+  HandlerT::NewResult New(storage::Config config, replication::ReplicationState &repl_state) {
```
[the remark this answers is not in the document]
andrejtonev: `DatabaseHandler` is not thread-safe; `DbmsHandler` is thread-safe and is the only thing using it.

src/dbms/database_handler.hpp
```diff
@@ -61,8 +61,7 @@ class DatabaseHandler : public Handler<Database> {
       spdlog::info("Tried to generate new storage using a claimed directory.");
       return NewError::EXISTS;
     }
-    config.name = name;  // Set storage id via config
-    return HandlerT::New(std::piecewise_construct, name, config, repl_state);
+    return HandlerT::New(std::piecewise_construct, config.salient.name, config, repl_state);
```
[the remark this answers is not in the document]
andrejtonev: The `Handler` was created at a time when multiple things were being handled at the same time.
Now all of that is encompassed in the `Database` class.
I would still leave it as is, the generic `Handler` could still be useful.
Regarding inheritance vs using; I think it's a stile choice. Any reason to not inherit.

src/dbms/dbms_handler.cpp
```diff
@@ -11,56 +11,204 @@
...
+
+  struct MigrationException : public utils::BasicException {
+    MigrationException() : utils::BasicException("Failed to migrate to the current durability version!") {}
+  };
+
+  static DurabilityVersion StringToVersion(std::string_view str) {
+    if (str == "V0") return DurabilityVersion::V0;
+    if (str == "V1") return DurabilityVersion::V1;
+    throw VersionException();
+  };
```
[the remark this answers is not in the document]
andrejtonev: `StringToVersion` not used, removed.

src/dbms/dbms_handler.cpp
```diff
@@ -11,56 +11,204 @@
...
+        // Generate a UUID
+        auto const uuid = utils::UUID();
+        // New json values
+        auto new_key = GenKey(key);
+        auto path = root;
+        if (key != kDefaultDB) {  // Special case for non-default DBs
+          // Move directory to new UUID dir
+          path = root / kMultiTenantDir / std::string{uuid};
+          std::filesystem::path old_dir(root / kMultiTenantDir / key);
+          std::filesystem::rename(old_dir, path);
```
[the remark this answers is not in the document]
andrejtonev: This is on startup. I'm going to error check and then assert.

src/query/frontend/semantic/required_privileges.cpp
```diff
@@ -106,6 +106,7 @@ class PrivilegeExtractor : public QueryVisitor<void>, public HierarchicalTreeVis
         AddPrivilege(AuthQuery::Privilege::MULTI_DATABASE_EDIT);
         break;
       case MultiDatabaseQuery::Action::USE:
+      case MultiDatabaseQuery::Action::SHOW:
```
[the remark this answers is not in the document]
andrejtonev: `SHOW` limits the databases to the ones that the user is allowed to use.
I get what you mean, but when calling `USE` you are going to want to do `SHOW`.
At the end of the day, this is a UX thing. Will update when/if someone complains.

src/dbms/replication_handler.cpp
andrejtonev: Good catch!!

src/query/interpreter.cpp
```diff
@@ -786,8 +785,9 @@ Callback HandleReplicationQuery(ReplicationQuery *repl_query, const Parameters &
       callback.header = {
           "name", "socket_address", "sync_mode", "current_timestamp_of_replica", "number_of_timestamp_behind_master",
```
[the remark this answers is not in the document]
andrejtonev: Has been like this for a hear and a half.
This is a UX change, not sure if changing it will break some user scripts.

src/storage/v2/storage.hpp
andrejtonev: Will not do at this time.

src/query/interpreter.cpp
```diff
@@ -3524,48 +3529,56 @@ PreparedQuery PrepareMultiDatabaseQuery(ParsedQuery parsed_query, CurrentDB &cur
...
+          [db_acc = current_db.db_acc_, pull_plan = std::shared_ptr<PullPlanVector>(nullptr)](
+              AnyStream *stream, std::optional<int> n) mutable -> std::optional<QueryHandlerResult> {
+            if (!pull_plan) {
+              std::vector<std::vector<TypedValue>> results;
+              auto db_name = db_acc ? TypedValue{db_acc->get()->storage()->name()} : TypedValue{};
+              results.push_back({std::move(db_name)});
+              pull_plan = std::make_shared<PullPlanVector>(std::move(results));
+            }
+
+            if (pull_plan->Pull(stream, n)) {
```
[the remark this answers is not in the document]
andrejtonev: By default the true path is the likely path.
None of the `if(Pull)...` use `[[likely]]`, I don't think there is a need for it.

src/query/stream/streams.cpp
```diff
@@ -644,6 +644,25 @@ void Streams::Drop(const std::string &stream_name) {
...
+  });
```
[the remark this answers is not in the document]
andrejtonev: We are following the previous behavior. Stop/StopAll do Lock, Stop, Persist. Drop does only Lock, Delete.
We extended Drop to do DropAll, but did not change the meaning of Drop. 
So, you need to do StopAll and then DropAll.

src/query/interpreter.cpp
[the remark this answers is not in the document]
andrejtonev: @gitbuda this is a small change, but could have big consequences on some users.
@antoniofilipovic can you make a separate issue? I don't want to deliver this as part of this huge PR.

src/utils/thread_pool.hpp
```diff
@@ -23,6 +23,16 @@
...
 
+template <typename Func>
+struct CopyMovableFunctionWrapper {
+  CopyMovableFunctionWrapper(Func &&func) : func_{std::make_shared<Func>(std::move(func))} {}
+
+  void operator()() { (*func_)(); }
+
+ private:
+  std::shared_ptr<Func> func_;
+};
```
antoniofilipovic, line 34: How come you didn't end up using Gareth's movable function wrapper

antoniofilipovic (changes requested):
Last changes and comments, otherwise good to go from my side

src/utils/thread_pool.hpp
andrejtonev: Gareth's function wrapper is a move-only function.
This means we would need to update the ThreadPool to use it and that is out of scope of this PR.

antoniofilipovic (approved):
Looks good to me

src/query/interpreter.cpp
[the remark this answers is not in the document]
andrejtonev: I'm just going to resolve it :D

Ignition:
related to #1359

PR #1557, opened 2023-12-01
Add function for property sizes

Josipmrden:
`
MATCH (n) RETURN property_size(n, "prop");
`
can now return the number of bytes used for a specific property

[master < Task] PR


src/storage/v2/storage.hpp
```diff
@@ -305,6 +309,14 @@ class Storage {
     return PropertyId::FromUint(name_id_mapper_->NameToId(name));
   }
 
+  std::optional<PropertyId> NameToPropertyIfExists(const std::string_view name) const {
```
as51340, line 312: I would rename it to `GetNameToProperty`, makes more sense to me

src/query/interpret/awesome_memgraph_functions.cpp
```diff
@@ -442,6 +442,32 @@ TypedValue Size(const TypedValue *args, int64_t nargs, const FunctionContext &ct
...
+    return TypedValue(static_cast<int64_t>(property_size), ctx.memory);
+  }
+
+  if (graph_entity.IsEdge()) {
+    uint64_t property_size = graph_entity.ValueEdge().GetPropertySize(*maybe_property_id).GetValue();
+    return TypedValue(static_cast<int64_t>(property_size), ctx.memory);
+  }
+
+  // Graph entity is null
+  return TypedValue(0, ctx.memory);
```
as51340, line 468:
```suggestion
  const auto &property_name = args[1].ValueString();
  const auto maybe_property_id = dba->NameToPropertyIfExists(property_name);
  if (!maybe_property_id) {
    return TypedValue(0, ctx.memory);
  }
  uint64_t property_size = 0;
  const auto &graph_entity = args[0];
  if (graph_entity.IsVertex()) {
    property_size = graph_entity.ValueVertex().GetPropertySize(*maybe_property_id).GetValue();
  } else if (graph_entity.IsEdge()) {
    uint64_t property_size = graph_entity.ValueEdge().GetPropertySize(*maybe_property_id).GetValue();
  }

  // Graph entity is null
  return TypedValue(static_cast<int64_t>(property_size), ctx.memory);
```

src/storage/v2/name_id_mapper.hpp
```diff
@@ -83,6 +83,18 @@ class NameIdMapper {
     return id;
   }
 
+  /// This method unlike NameToId does not insert the new property id if not found
+  /// but just returns either std::nullopt or the value of the property id if it
+  /// finds it.
+  virtual std::optional<uint64_t> NameToIdIfExists(const std::string_view name) {
```
as51340, line 89: I think this method and all methods which are using this one can be const

src/storage/v2/property_store.cpp
```diff
@@ -1037,9 +1192,25 @@ PropertyValue PropertyStore::GetProperty(PropertyId property) const {
...
+uint64_t PropertyStore::PropertySize(PropertyId property) const {
+  uint64_t size = 0;
+  const uint8_t *data = nullptr;
+  std::tie(size, data) = GetSizeData(buffer_);
+  if (size % 8 != 0) {
+    // We are storing the data in the local buffer.
+    size = sizeof(buffer_) - 1;
+    data = &buffer_[1];
+  }
+  Reader reader(data, size);
```
as51340, line 1204: This part of code is duplicated at multiple places, can we extract it?

src/storage/v2/vertex_accessor.cpp
```diff
@@ -403,6 +403,10 @@ Result<PropertyValue> VertexAccessor::GetProperty(PropertyId property, View view
   return std::move(value);
 }
 
+Result<uint64_t> VertexAccessor::GetPropertySize(PropertyId property) const {
+  return vertex_->properties.PropertySize(property);
```
as51340, line 407: I think this is not correct. Since we don't have strict schema, the type of property can change and hence its size can change. I think this method needs to accept `View` and get the property value depending on the `View` similar to how `GetProperty` operates. Then you will probably have some duplicated code so please extract that also

src/storage/v2/property_store.cpp
```diff
@@ -1037,9 +1192,25 @@ PropertyValue PropertyStore::GetProperty(PropertyId property) const {
...
+
+  uint64_t property_size = 0;
+  if (FindSpecificPropertySize(&reader, property, property_size) != ExpectedPropertyStatus::EQUAL) return 0;
```
as51340, line 1207: How about throwing exception if the property doesn't exist?

src/storage/v2/property_store.cpp
```diff
@@ -875,6 +1016,20 @@ enum class ExpectedPropertyStatus {
...
+    auto ret = DecodeExpectedPropertySize(reader, property, size);
+    // Because the properties are sorted in the buffer, we only need to
+    // continue searching for the property while this function returns a
+    // `SMALLER` value indicating that the ID of the found property is smaller
+    // than the seeked ID. All other return values (`MISSING_DATA`, `EQUAL` and
+    // `GREATER`) terminate the search.
+    if (ret != ExpectedPropertyStatus::SMALLER) {
+      return ret;
+    }
+  }
```
as51340, line 1030:
```suggestion
    ExpectedPropertyStatus ret;
  while ((ret = DecodeExpectedPropertySize(reader, property, size)) == ExpectedPropertyStatus::SMALLER) {}
  return ret;

```

src/storage/v2/property_store.cpp
as51340: If you don't want exception then reimplement using std::optional. You don't need to know then what is the status. I would prefer exception with the specific status check however.

tests/e2e/awesome_functions/awesome_functions.py
```diff
@@ -0,0 +1,147 @@
...
+    nonexistent_bytes = get_bytes("nonexistent_prop")
+
+    # No property stored, no bytes allocated
+    assert null_bytes == 0
+
+    # 1 byte metadata, 1 byte prop id, but value is encoded in the metadata
+    assert bool_bytes == 2
+
+    # 1 byte metadata, 1 byte prop id + payload size 1 byte to store the int
+    assert s_int_bytes == 3
```
as51340, line 66: Is this because int stored is 4 so we can use int8?

tests/e2e/awesome_functions/awesome_functions.py
```diff
@@ -0,0 +1,147 @@
...
+    assert m_int_bytes == 4
+
+    # 1 byte metadata, 1 byte prop id + payload size 4 bytes to store the int
+    assert l_int_bytes == 6
+
+    # 1 byte metadata, 1 byte prop id + payload size 8 bytes to store the int
+    assert xl_int_bytes == 10
+
+    # 1 byte metadata, 1 byte prop id + payload size 8 bytes to store the float
+    assert float_bytes == 10
```
as51340, line 78: Why for float 8B payload?

tests/e2e/awesome_functions/awesome_functions.py
```diff
@@ -0,0 +1,147 @@
...
+
+def test_indexed_join_with_indices(memgraph):
+    def get_bytes(prop_name):
+        res = list(memgraph.execute_and_fetch(f"MATCH (n) RETURN property_size(n, '{prop_name}') AS size"))
+        return res[0]["size"]
+
+    memgraph.execute(
+        """
+        CREATE (n:Node)
+        SET n.null_prop = null,
```
as51340, line 26: Great description of what is going on. I think it would be better to have 1 test for each property type because it would be easier to change things in the future but ok

as51340 (changes requested):
Nice job, however I think we need to incorporate View

src/storage/v2/property_store.cpp
```diff
@@ -572,6 +606,92 @@ std::optional<TemporalData> DecodeTemporalData(Reader &reader) {
...
+      property_size += SizeToByteSize(payload_size);
+      return true;
+    }
+    case Type::STRING: {
+      auto size = reader->ReadUint(payload_size);
+      if (!size) return false;
+      property_size += SizeToByteSize(payload_size);
+
+      std::string str_v(*size, '\0');
+      if (!reader->ReadBytes(str_v.data(), *size)) return false;
```
Ignition, line 634: Don't need to read, just validate enough bytes in reader + skip those bytes.

src/storage/v2/property_store.cpp
```diff
@@ -1037,9 +1192,25 @@ PropertyValue PropertyStore::GetProperty(PropertyId property) const {
   return value;
 }
 
+uint64_t PropertyStore::PropertySize(PropertyId property) const {
+  uint64_t size = 0;
+  const uint8_t *data = nullptr;
+  std::tie(size, data) = GetSizeData(buffer_);
```
Ignition, line 1198: Just use bindings `auto [size, data] = ...`

src/storage/v2/vertex_accessor.cpp
Ignition: Yeah not thread safe, need to get shared lock to do anything on properties.
And as Andi said it should be consistent to the isolation level, hence you maybe need to copy + apply deltas.

tests/e2e/awesome_functions/awesome_functions.py
Ignition: because its actually `double`

src/storage/v2/storage.hpp
Josipmrden: I mean it's not a getter, because you are doing a transformation, name -> property

tests/e2e/awesome_functions/awesome_functions.py
Josipmrden: yes, if it fits in a shorter byte int then we encode it in less bytes

tests/e2e/awesome_functions/awesome_functions.py
Josipmrden: Yes I will separate each test into its own test so we can easily test which test failed.

Josipmrden:
@as51340 @Ignition re-requested, everything should be addressed now

src/storage/v2/property_store.cpp
```diff
@@ -983,6 +1131,31 @@ std::pair<uint64_t, uint8_t *> GetSizeData(const uint8_t *buffer) {
...
+  std::tie(size, data) = GetSizeData(buffer);
+  if (size % 8 != 0) {
+    // We are storing the data in the local buffer.
+    size = sizeof(buffer) - 1;
+    data = &buffer[1];
+    in_local_buffer = true;
+  }
+
+  // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
+  auto *non_const_data = const_cast<uint8_t *>(data);
```
as51340, line 1154: Why declare it const and then cast to non-const?

src/storage/v2/property_store.cpp
```diff
@@ -983,6 +1131,31 @@ std::pair<uint64_t, uint8_t *> GetSizeData(const uint8_t *buffer) {
...
+  uint8_t *data{nullptr};
+  bool in_local_buffer;
+};
+
+template <size_t N>
+BufferInfo GetBufferInfo(const uint8_t (&buffer)[N]) {
+  uint64_t size = 0;
+  const uint8_t *data = nullptr;
+  bool in_local_buffer = false;
+  std::tie(size, data) = GetSizeData(buffer);
```
as51340, line 1145:
```suggestion
  auto [size, data] = GetSizeData(buffer);
```

as51340 (approved):
:muscle:

src/query/interpret/awesome_memgraph_functions.cpp
```diff
@@ -442,6 +442,31 @@ TypedValue Size(const TypedValue *args, int64_t nargs, const FunctionContext &ct
...
+TypedValue PropertySize(const TypedValue *args, int64_t nargs, const FunctionContext &ctx) {
+  FType<Or<Null, Vertex, Edge>, Or<String>>("property_size", args, nargs);
+
+  auto *dba = ctx.db_accessor;
+
+  const auto &property_name = args[1].ValueString();
+  const auto maybe_property_id = dba->NameToPropertyIfExists(property_name);
+
+  if (!maybe_property_id) {
+    return TypedValue(0, ctx.memory);
```
Ignition, line 454: Would NULL response be better?

src/query/interpret/awesome_memgraph_functions.cpp
```diff
@@ -442,6 +442,31 @@ TypedValue Size(const TypedValue *args, int64_t nargs, const FunctionContext &ct
...
+  }
+
+  uint64_t property_size = 0;
+  const auto &graph_entity = args[0];
+  if (graph_entity.IsVertex()) {
+    property_size = graph_entity.ValueVertex().GetPropertySize(*maybe_property_id, ctx.view).GetValue();
+  }
+
+  if (graph_entity.IsEdge()) {
```
Ignition, line 463: else if

src/query/interpret/awesome_memgraph_functions.cpp
```diff
@@ -1325,6 +1350,7 @@ std::function<TypedValue(const TypedValue *, int64_t, const FunctionContext &ctx
   if (function_name == "PROPERTIES") return Properties;
   if (function_name == "RANDOMUUID") return RandomUuid;
   if (function_name == "SIZE") return Size;
+  if (function_name == "PROPERTY_SIZE") return PropertySize;
```
Ignition, line 1353: underscore in the function name, not consistent with previous naming e.g. `RANDOMUUID`

src/storage/v2/edge_accessor.cpp
```diff
@@ -264,6 +265,29 @@ Result<PropertyValue> EdgeAccessor::GetProperty(PropertyId property, View view)
...
+
+  {
+    auto guard = std::shared_lock{edge_.ptr->lock};
+    Delta *delta = edge_.ptr->delta;
+    if (!delta) {
+      return edge_.ptr->properties.PropertySize(property);
+    }
+  }
+
+  auto property_result = this->GetProperty(property, view);
```
Ignition, line 279: This still needs to be under the read lock

src/storage/v2/edge_accessor.cpp
```diff
@@ -264,6 +265,29 @@ Result<PropertyValue> EdgeAccessor::GetProperty(PropertyId property, View view)
...
+
+  if (property_result.HasError()) {
+    return property_result.GetError();
+  }
+
+  auto property_store = storage::PropertyStore();
+  property_store.SetProperty(property, *property_result);
```
Ignition, line 286: You have not applied deltas in this case, the property that was got, maybe inconsistent with isolation mode. Same applies for VertexAccessor

Issue #1360, opened 2023-10-13
Linking MAGE to Memgraph slows down query modules


as51340:
**Memgraph version**
2.11, 2.10, 2.9

**Environment**
Docker and local build

**Describe the bug**
When running Memgraph together with MAGE, calling of query modules is much slower than in the case when we just create this one query module from the Memgraph LAB.
With memgraph-mage and memgraph-platform the import of 2000 nodes using LOAD CSV takes 2-5 min. When using just memgraph image and we create query module separately, import of 2000 nodes takes 17s.

Importing query:
```cypher
LOAD CSV FROM "https://download.memgraph.com/datasets/music-genres-social-network/csv/nodes-user.csv.gz"
  WITH HEADER IGNORE BAD AS row
  CREATE (:`__mg_node__`:`USER` {
    `__mg_id__`: row.`__mg_id__:ID:INDEX:TEMPORARY`,
    `id`: toInteger(row.`id:INTEGER`),
    `genres`: convert.str2object(row.`genres:JSON`)
  });
```

Link to the MAGE function that is used: [convert.str2object](https://github.com/memgraph/mage/blob/main/python/convert.py).

**Expected behavior**
Memgraph with MAGE shouldn't be slower.

as51340:
I successfully reproduced the issue just with a bit different numbers but the logic is the same:
- import without MAGE = few seconds
- import with MAGE = 1 min

Perf wasn't very helpful but it obviously has to do something with linking Python.
![profile_2](https://github.com/memgraph/memgraph/assets/53269502/97d71c1a-9c83-4dbd-a90a-202c67de38b5)

cc @tonilastre to keep you updated

antejavor:
This may be related to that issue: https://github.com/memgraph/memgraph/issues/969 cc @tonilastre @as51340

gitbuda:
+1 report about `convert.str2object` being slow

as51340:
Failed to do some meaningful work here except that there is a big difference in profiling whether you link the module at the beginning (by setting query modules) or you just create a query module from lab.

This is result when whole mage is linked to Memgraph:


![Image](https://github.com/memgraph/memgraph/assets/53269502/8487f096-89c4-459a-a3d7-9d4c7bf54e80)




This is result when just one module is linked to Memgraph:


![Image](https://github.com/memgraph/memgraph/assets/53269502/2e2416af-fe95-43b5-a5a5-b8c2616d1202)

as51340:
[...]

Issue #969, opened 2023-06-07
Time taken to call a python query module is to darn high!


chrischabot:
When implementing a simple python query module function, something simple that returns "hello world", calling it takes ~ 800ms on my system

Now python does have a global interpreter lock and a bunch of overhead -- however 800ms is to damn high! Unreasonably so, gotta be something weird going on

![jimmy-mcmillan-too-damn-high](https://github.com/memgraph/memgraph/assets/11455514/41262a11-f565-4c09-9e1e-8290455909f8)

katarinasupe:
I see this issue has been put on pause. @as51340, why is that? 
We have various reports of this on our Discord + every Cloud instance runs Memgraph MAGE image, which, along with Memgraph Platform, has slower execution of custom query modules in Python. Enterprise Cloud users who are experiencing slow execution time with the custom Python procedure can't rewrite it in C++ because there is no possibility of uploading that module to a running Memgraph Cloud instance. 

Latest source: https://discord.com/channels/842007348272169002/890968933040009276/1179069780125298698
Related internal Slack discussion: https://memgraph.slack.com/archives/CSR01RS80/p1697092530088239

as51340:
@imilinovic will continue work when he gets back

imilinovic:
I've already identified the problem and the fix, it should make it in this release.

PR #1558, opened 2023-12-03
Fix slow python QM (Python GC changes)

imilinovic:
The Python GC was being called after execution of every query which caused the queries to be extremely slow. Python has automatic and sophisticated GC that gets called automatically after the number of allocations minus the number of deallocations exceeds the threshold (700 by default) so manually calling it every time was unnecessary. 

Relying only on the automatic GC can cause some objects to not get deleted if for example the user stops using QM for a while. To combat that Python GC cycle was added but by default set to high value (180 seconds) because it is not as critical and doesn't really need to be run very often. For that reason a new flag was also added.

[master < Epic] PR
- [ ] Write E2E tests

closes: #969 
closes: #1360 

[master < Task] PR

Python query modules are significantly faster now.


src/query/procedure/py_module.cpp
```diff
@@ -1095,7 +1097,7 @@ void CallPythonProcedure(const py::Object &py_cb, mgp_list *args, mgp_graph *gra
   std::optional<std::string> maybe_msg;
   {
     py::Object py_graph(MakePyGraph(graph, memory));
-    utils::OnScopeExit clean_up(PyObjectCleanup(py_graph, !is_batched));
```
as51340, line 1098: Why this !is_batched is removed?

as51340, line 1098: Why is `!is_batched` removed?

src/storage/v2/CMakeLists.txt
```diff
@@ -42,4 +42,4 @@ add_library(mg-storage-v2 STATIC
         inmemory/replication/recovery.cpp
 )
 
-target_link_libraries(mg-storage-v2 mg::replication Threads::Threads mg-utils gflags absl::flat_hash_map mg-rpc mg-slk mg-events mg-memory)
+target_link_libraries(mg-storage-v2 mg::replication Threads::Threads mg-utils gflags absl::flat_hash_map mg-rpc mg-slk mg-events mg-memory mg-query)
```
as51340, line 45: why that, this creates dependency between storage and query?

src/query/procedure/py_module.cpp
```diff
@@ -867,6 +867,24 @@ py::Object MgpListToPyTuple(mgp_list *list, PyObject *py_graph) {
   return MgpListToPyTuple(list, reinterpret_cast<PyGraph *>(py_graph));
 }
 
+void PyCollectGarbage() {
```
as51340, line 870: can this be const method?

src/storage/v2/inmemory/storage.cpp
```diff
@@ -139,6 +140,11 @@ InMemoryStorage::InMemoryStorage(Config config, StorageMode storage_mode)
     }
   }
 
+  if (config_.gc_python.type == Config::Gc::Type::PERIODIC) {
```
as51340, line 143: What if Memgraph is used without query modules? Then it doesn't make sense to use resources for doing PERIODIC gc

as51340 (changes requested):

src/query/procedure/py_module.cpp
imilinovic: Because PyObjectCleanup doesn't have start_gc parameter anymore since we don't run it after every query

src/storage/v2/CMakeLists.txt
imilinovic: To do periodic python GC we need PyCollectGarbage from PyModule, I don't know a way to avoid the dependancy

src/storage/v2/inmemory/storage.cpp
imilinovic: There are Python query modules that are there by default without MAGE and the timer is by default set to quite high (180 seconds) so it shouldn't affect perfomance noticeably. GC only affects the python modules and is run on separate thread (so it shouldn't affect the database at all)

src/query/procedure/py_module.cpp
imilinovic: This isn't a part of a class so no.

as51340: ah didn't see

src/storage/v2/CMakeLists.txt
as51340: ha but I think this could cause compilation slow-down... @Ignition advices?

Ignition: This highlights a design smell.

utils::Scheduler gc_python_runner_;
AND 
utils::Scheduler gc_jemalloc_runner_;

Should not be scoped to be the responsibility of storage. 
We need 1 runner for all of memgraph for jemalloc, hence should be in memgraph.cpp or some somewhere in glue code.
For python GC (not looked too deeply) either it needs to be 1 for all of memgraph (if we have a singleton python interpreter) or once per interpreter (which is would be query scoped).

imilinovic: Moved both the jemalloc and python schedulers to memgraph.cpp so no more dependancy

antoniofilipovic (commented):
Have you done any comparison in terms of performance between using python module with GC after procedure call vs periodic GC? What we will get in terms of memory usage also? One client heavily depends python migration.py module

antoniofilipovic:
From code perspective everything looks fine, I would love to see numbers on following:
 - migrate.py GC vs periodic GC in terms of memory
 - migrate.py GC vs periodic GC in terms of performance
 - any regular py module memory and performance in GC vs periodic GC

imilinovic:
For migrate.py I created a mysql database with a table with 1e6 entries and then ran:
`CALL migrate.mysql(credentials) YIELD row RETURN count(row);`
and
`CALL migrate.mysql(credentials) YIELD row RETURN row;`
For both of these performances were exactly the same both in memory and speed
First query: 
New GC: 21,69s max_memory_allocated: 55MiB
Old GC: 21,67s max_memory_allocated: 53MiB

Second query:
New GC: 51,16s max_memory_allocated 55MiB
Old GC: 50,97s max_memory_allocated 56MiB

Memory and speed vary every run so I would say its the same

The noticeable difference happens when the python QM is called a lot of times.
The following query will call the function convert.str2object 2000 times (this function is currently implemented in core and is no longer part of MAGE)
```
LOAD CSV FROM "https://download.memgraph.com/datasets/music-genres-social-network/csv/nodes-user.csv.gz"
  WITH HEADER IGNORE BAD AS row
  CREATE (:`__mg_node__`:`USER` {
    `__mg_id__`: row.`__mg_id__:ID:INDEX:TEMPORARY`,
    `id`: toInteger(row.`id:INTEGER`),
    `genres`: convert.str2object(row.`genres:JSON`)
  });
```

New GC: 518ms
Old GC: 17.02s

src/storage/v2/inmemory/storage.cpp
as51340: Separate thread still influences the machine's resources but understand what you mean.

as51340 (approved):
Great job @imilinovic !

PR #1560, opened 2023-12-04
Commit, Rollback and Discard were sending back TransientErrors

andrejtonev:
Bolt has specific exception types it expects. 
We need to translate between Bolt and Query worlds. This translation was missing from Commit, Rollback and Discard. 
This would lead to driver behaving wrongly. 

[master < Task] PR


src/glue/SessionHL.cpp
antepusic, line 1: Does `Abort` throw any exceptions that need to be handled?

antepusic (commented):
Looks good! I’ve left a comment about another case that might need handling

Ignition (approved):
:nut_and_bolt:

PR #1572, opened 2023-12-08
Performance tuning based on stress test

Ignition:
While looking at the large data stress test noticed a couple of things which shouldn't have been noticeable.

Some situations which were copying from a local optional, added move to reduce copy of a temporary. 

Also changed `EdgeAccessor::GetProperty` to construct inside of an optional rather than see the cost of `PropertyValue::operator=(PropertyValue &&)`

2024-01-19: Added more things. 
- Added [[unlikely]] to improve expected branching
- std::move or `const &` PropertyIx/LabelIx types
- set underlying type on enum class to make other types smaller
- Preferred vector construction from move iterators
- Check `properties_on_edges` once outside the hot loop
- InMemoryStorage now has a stop_source, which async sources and get a stop_token from to check if they should stop.
- Optimize the application of the predicate in `AnyVersionSatisfiesPredicate`. In heavy edge creation workloads most delta actions are not interesting, a cheaper check is preferred. 
- PropertyValue move constructor optimize, only one switch required.
- PropertyValue move assignment optimize, if types are the same, move union value, destroy is necessary, only use one switch. Else destroy `this` and delegate to move construction.
- skip_list more efficient putting leftovers back. Order in the stack does not matter.
- skip_list restrict the scope of the unique_lock guards, once `preds[layer]->nexts[layer]` have all been set the guards can be released.

### DIFF
--- a/src/query/frontend/ast/ast.hpp
+++ b/src/query/frontend/ast/ast.hpp
@@ -2217,7 +2217,7 @@ class IndexQuery : public memgraph::query::Query {
 
  protected:
   IndexQuery(Action action, LabelIx label, std::vector<PropertyIx> properties)
-      : action_(action), label_(std::move(label)), properties_(properties) {}
+      : action_(action), label_(std::move(label)), properties_(std::move(properties)) {}
 
  private:
   friend class AstStorage;
--- a/src/query/frontend/ast/ast.hpp
+++ b/src/query/frontend/ast/ast.hpp
@@ -1209,7 +1209,8 @@ class PropertyLookup : public memgraph::query::Expression {
   }
 
  protected:
-  PropertyLookup(Expression *expression, PropertyIx property) : expression_(expression), property_(property) {}
+  PropertyLookup(Expression *expression, PropertyIx property)
+      : expression_(expression), property_(std::move(property)) {}
 
  private:
   friend class AstStorage;
@@ -1805,9 +1806,9 @@ class EdgeAtom : public memgraph::query::PatternAtom {
   static const utils::TypeInfo kType;
   const utils::TypeInfo &GetTypeInfo() const override { return kType; }
 
-  enum class Type { SINGLE, DEPTH_FIRST, BREADTH_FIRST, WEIGHTED_SHORTEST_PATH, ALL_SHORTEST_PATHS };
+  enum class Type : uint8_t { SINGLE, DEPTH_FIRST, BREADTH_FIRST, WEIGHTED_SHORTEST_PATH, ALL_SHORTEST_PATHS };
 
-  enum class Direction { IN, OUT, BOTH };
+  enum class Direction : uint8_t { IN, OUT, BOTH };
 
   /// Lambda for use in filtering or weight calculation during variable expand.
   struct Lambda {
@@ -2216,7 +2217,7 @@ class IndexQuery : public memgraph::query::Query {
 
  protected:
   IndexQuery(Action action, LabelIx label, std::vector<PropertyIx> properties)
-      : action_(action), label_(label), properties_(properties) {}
+      : action_(action), label_(std::move(label)), properties_(properties) {}
 
  private:
   friend class AstStorage;
--- a/src/query/interpret/eval.hpp
+++ b/src/query/interpret/eval.hpp
@@ -1121,11 +1121,11 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
           throw QueryRuntimeException("Unexpected error when getting properties.");
       }
     }
-    return *maybe_props;
+    return *std::move(maybe_props);
   }
 
   template <class TRecordAccessor>
-  storage::PropertyValue GetProperty(const TRecordAccessor &record_accessor, PropertyIx prop) {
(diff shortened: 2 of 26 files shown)


src/storage/v2/edge_accessor.cpp
```diff
@@ -207,19 +207,20 @@ Result<PropertyValue> EdgeAccessor::GetProperty(PropertyId property, View view)
   if (!storage_->config_.items.properties_on_edges) return PropertyValue();
   bool exists = true;
   bool deleted = false;
-  PropertyValue value;
```
Josipmrden, line 210: RVO doesn't do anything here?

Ignition: Maybe. But regardless of that without LTO, the cost of `PropertyValue::operator=(PropertyValue &&)` is noticeable and hence we want to avoid it. Hence the `emplace` on line 216 to do move construction inside an optional.

Maybe PropertyValue should be header only? But that can be done in another PR if needed.

src/query/frontend/ast/ast.hpp
```diff
@@ -2216,7 +2217,7 @@ class IndexQuery : public memgraph::query::Query {
 
  protected:
   IndexQuery(Action action, LabelIx label, std::vector<PropertyIx> properties)
-      : action_(action), label_(label), properties_(properties) {}
+      : action_(action), label_(std::move(label)), properties_(properties) {}
```
as51340, line 2220: Move `properties`?

src/query/interpret/eval.hpp
```diff
@@ -1121,11 +1121,11 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
           throw QueryRuntimeException("Unexpected error when getting properties.");
       }
     }
-    return *maybe_props;
+    return *std::move(maybe_props);
```
as51340, line 1124: Isn't it better to leave so that RVO kicks in? Can it kick in here?

src/utils/skip_list.hpp
```diff
@@ -1028,40 +1029,45 @@ class SkipList final : detail::SkipListNode_base {
...
-      allocator.construct(new_node, top_layer, std::forward<TObjUniv>(object));
-
-      // The paper is also wrong here. It states that the loop should go up to
-      // `top_layer` which is wrong.
-      for (int layer = 0; layer < top_layer; ++layer) {
-        new_node->nexts[layer].store(succs[layer], std::memory_order_release);
-        preds[layer]->nexts[layer].store(new_node, std::memory_order_release);
+        size_t node_bytes = sizeof(TNode) + top_layer * sizeof(std::atomic<TNode *>);
+        MemoryResource *memoryResource = GetMemoryResource();
+        Allocator<TNode> allocator(memoryResource);
```
as51340, line 1056: Move it to closer to where is first used

src/storage/v2/edge_accessor.hpp
```diff
@@ -86,6 +86,8 @@ class EdgeAccessor final {
   Result<std::map<PropertyId, PropertyValue>> Properties(View view) const;
 
   Gid Gid() const noexcept;
+  auto GidPropertiesOnEdges() const { return edge_.ptr->gid; }
```
as51340, line 89:
```suggestion
  auto GidPropertiesOnEdges() const -> Gid { return edge_.ptr->gid; }
```

src/storage/v2/indices/indices.cpp
```diff
@@ -32,10 +32,10 @@ void Indices::AbortEntries(LabelId label, std::span<std::pair<PropertyValue, Ver
       ->AbortEntries(label, vertices, exact_start_timestamp);
 }
 
-void Indices::RemoveObsoleteEntries(uint64_t oldest_active_start_timestamp) const {
-  static_cast<InMemoryLabelIndex *>(label_index_.get())->RemoveObsoleteEntries(oldest_active_start_timestamp);
+void Indices::RemoveObsoleteEntries(uint64_t oldest_active_start_timestamp, std::stop_token token) const {
+  static_cast<InMemoryLabelIndex *>(label_index_.get())->RemoveObsoleteEntries(oldest_active_start_timestamp, token);
```
as51340, line 36:
```suggestion
  static_cast<InMemoryLabelIndex *>(label_index_.get())->RemoveObsoleteEntries(oldest_active_start_timestamp, std::move(token));
```

as51340 (approved):
Nice job! Few comments only

src/query/interpret/eval.hpp
Ignition: Nope, no RVO does not apply here

src/storage/v2/edge_accessor.hpp
Ignition: :+1: 
Had to move `Gid Gid() const noexcept;` down because it shadowed Gid the type.

src/storage/v2/indices/indices.cpp
Ignition: Nope because its not the last usage. On the next line, 37, it is moved.

Issue #1563, opened 2023-12-05
ToString is rounding float values to 6 decimals


Josipmrden:
**Memgraph version**
v2.12.1

**Environment**
Docker

**Describe the bug**
ToString is rounding float numbers to 6 decimal places which shouls not happen

**To Reproduce**
See the picture
![Screenshot from 2023-12-05 14-28-14](https://github.com/memgraph/memgraph/assets/22621791/fb71f0cf-03ad-4829-a4f7-7f4667920acd)

**Expected behavior**
The string should be parsed without rounding

PR #1576, opened 2023-12-12
ToString function now returns double values with precision 15

Darych:
[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR

#1563


src/utils/string.hpp
```diff
@@ -459,4 +459,30 @@ inline std::string_view Substr(const std::string_view string, size_t pos = 0, si
...
+ *
+ * @return The string representation of the double value.
+ *
+ * @throws None
+ */
+inline std::string DoubleToString(const double value) {
+  static const int PRECISION = 16;
+  static const int STR_LENGTH = PRECISION + 1;  // 1 for dot
+
+  char buf[32];
```
andrejtonev, line 477: Using `stringstream` would be more c++-like.
Plus you could avoid a copy between buf and string and at the end move the stringstream, so we make sure to avoid that copy as well. 
See `std::precision` and `std::fixed`. It's not super powerful, it almost does what you want, but there are some issue with rounding and scientific notation used in some cases. So, I think, at the end you will need to do the same manipulation you are currently doing.

Darych: I thought about `stringstream`. Yes, it's more C++ way. I will test this approach.
I used `sprintf` because it's under the hood of `std::to_string`

Ignition: `fmt::format("{:.16f}", d)` https://godbolt.org/z/ehGYcco5d
But note adding precision inn cases like 5.8 it makes the number returned worse.

Darych: For this reason I prefer to use `sprintf` or `stringstream`.

Ignition: My point was that that precision of 16, in all cases, will make make `5.8` -> `5.7999999999999998` no matter what. This is because the actual represented value is `5.79999999999999982236431605997495353221893310546875` and hence with 16 digits it rounds the boundary 82 down to 80 hence `5.7999999999999998`. If we used 15 digits then boundary 98 would round up to 100 giving us clean `5.800000000000000`.

Ignition: I'm not sure what is generally the best practice for printing floating numbers.
All examples https://godbolt.org/z/8sPbbenxW

src/utils/string.hpp
```diff
@@ -459,4 +460,30 @@ inline std::string_view Substr(const std::string_view string, size_t pos = 0, si
...
+ *
+ * @throws None
+ */
+inline std::string DoubleToString(const double value) {
+  static const int PRECISION = 16;
+  static const int STR_LENGTH = PRECISION + 1;  // 1 for dot
+
+  std::stringstream ss;
+  ss << std::setprecision(PRECISION) << std::fixed << value;
+  std::string res(ss.view(), 0, STR_LENGTH);
```
Ignition, line 480: `auto res = ss.str();`

src/utils/string.hpp
Ignition: Look into https://github.com/ulfjack/ryu 
> For example, consider the binary 32-bit floating point number 00111110100110011001100110011010. The stored value is exactly 0.300000011920928955078125. However, this floating point number is also the closest number to the decimal number 0.3, so that is what Ryu outputs.

Ignition (dismissed):
Arbitrary precision of 16 isn't a good UX. I believe user expectation would more in line with what Ryu would give us.

hal-eisen-MG:
Has anyone been looking at this recently? I'd like to get to a decision about whether or not we can include it ASAP. Thanks!

Darych:
> Arbitrary precision of 16 isn't a good UX. I believe user expectation would more in line with what Ryu would give us.

We decided to merge this PR with precision 15 for now. Later we will add Ryu library in #1590

Ignition (approved):

src/utils/string.hpp
```diff
@@ -459,4 +460,30 @@ inline std::string_view Substr(const std::string_view string, size_t pos = 0, si
...
+ */
+inline std::string DoubleToString(const double value) {
+  static const int PRECISION = 15;
+  static const int STR_LENGTH = PRECISION + 1;  // 1 for dot
+
+  std::stringstream ss;
+  ss << std::setprecision(PRECISION) << std::fixed << value;
+  std::string res(ss.view(), 0, STR_LENGTH);
+
+  res.erase(res.find_last_not_of('0') + 1);
```
andrejtonev, line 482: Can we put a comment here saying that we are guaranteed to have at least one char different from `0` (the `.`). Otherwise we could get `npos` and we shouldn't increment it.

src/utils/string.hpp
```diff
@@ -459,4 +460,30 @@ inline std::string_view Substr(const std::string_view string, size_t pos = 0, si
...
+ *
+ * @throws None
+ */
+inline std::string DoubleToString(const double value) {
+  static const int PRECISION = 15;
+  static const int STR_LENGTH = PRECISION + 1;  // 1 for dot
+
+  std::stringstream ss;
+  ss << std::setprecision(PRECISION) << std::fixed << value;
+  std::string res(ss.view(), 0, STR_LENGTH);
```
andrejtonev, line 480: We could just use the string_view and avoid a copy here.
```
auto sv = ss.view();
sv = sv.substr(0, sv.find_last_not_of('0') + 1);
if (sv.ends_with('.')) {
    sv = sv.substr(0, sv.size() - 1);
}
```

andrejtonev (commented):
Just a couple of comments.

src/utils/string.hpp
Darych: Thank you! Changed the code.

src/utils/string.hpp
```diff
@@ -459,4 +460,31 @@ inline std::string_view Substr(const std::string_view string, size_t pos = 0, si
...
+ *
+ * @param value The double value to be converted.
+ *
+ * @return The string representation of the double value.
+ *
+ * @throws None
+ */
+inline std::string DoubleToString(const double value) {
+  static const int PRECISION = 15;
+  static const int STR_LENGTH = PRECISION + 1;  // 1 for dot
```
andrejtonev, line 476: We shouldn't need to calculate this ourselves.
Using `std::precision` with `std::fixed` means that we are defining the number of decimal places, not the overall number of digits. So the length is actually different from `STR_PRECISON`.
We can just use the whole stringstream's string_view.

Darych: https://godbolt.org/z/T555Kr654
`setprecision` sets count of numbers after dot + dot symbol.
So this case will be broken.
`EXPECT_EQ(DoubleToString(1234567.01234567891234567), "1234567.01234567");`

Darych: Updated.

PR #1581, opened 2023-12-12
Cleanup filesystem after unittest

Ignition:
Reduce system dirtiness after unit tests.

A number of unit test left their temporary storage directories behind after test completion. Added deterministic folder deletion to those unit tests.


andrejtonev (approved):
Good!

PR #1584, opened 2023-12-13
Cleanup filesystem after e2e tests

Ignition:
When an e2e test passes then it data directory does not need to remain.


andrejtonev (approved):
As long as the tests don't complain, I don't see a problem with this. :muscle:

as51340 (approved):

PR #1589, opened 2023-12-15
Add atomic memory block around unsafe code blocks

antoniofilipovic:
Fixes https://github.com/memgraph/memgraph/actions/runs/7100898041/job/19359933556

Slack thread: https://memgraph.slack.com/archives/CS5827FK3/p1702652195334839?thread_ts=1702627200.164709&cid=CS5827FK3

If idea seems good, I will apply this to rest of operations in vertex_accessor and edge_accessor which from memory allocation should be "atomic".

[master < Task] PR


src/utils/oom_blocker_block.hpp
```diff
@@ -0,0 +1,44 @@
...
+      function_();
+    }
+    total_memory_tracker.DoCheck();
+  }
+
+ private:
+  std::function<void()> function_;
+};
+template <typename Callable>
+OOMBlockerBlock(Callable &&) -> OOMBlockerBlock<Callable>;
```
as51340, line 42: What is this for?

src/storage/v2/vertex_accessor.cpp
```diff
@@ -107,8 +108,11 @@ Result<bool> VertexAccessor::AddLabel(LabelId label) {
   if (vertex_->deleted) return Error::DELETED_OBJECT;
   if (std::find(vertex_->labels.begin(), vertex_->labels.end(), label) != vertex_->labels.end()) return false;
 
-  CreateAndLinkDelta(transaction_, vertex_, Delta::RemoveLabelTag(), label);
-  vertex_->labels.push_back(label);
+  utils::OOMBlockerBlock oom_blocker_block{[transaction = transaction_, vertex = vertex_, &label]() {
+    CreateAndLinkDelta(transaction, vertex, Delta::RemoveLabelTag(), label);
+    vertex->labels.push_back(label);
+  }};
+  std::invoke(oom_blocker_block);
```
as51340, line 115:
```suggestion
  std::invoke([transaction = transaction_, vertex = vertex_, &label]() {
    CreateAndLinkDelta(transaction, vertex, Delta::RemoveLabelTag(), label);
    vertex->labels.push_back(label);
  });

```

src/utils/memory_tracker.hpp
```diff
@@ -49,6 +49,7 @@ class MemoryTracker final {
 
   void Alloc(int64_t size);
   void Free(int64_t size);
+  void DoCheck();
```
as51340, line 52: But don't we do this check already in `MemoryTracker::Alloc` whenever `utils::MemoryTracker::OutOfMemoryExceptionEnabler` exists?

src/storage/v2/vertex_accessor.cpp
```diff
@@ -287,20 +291,28 @@ Result<bool> VertexAccessor::InitProperties(const std::map<storage::PropertyId,
...
-    if (!value.IsNull()) {
-      transaction_->constraint_verification_info.AddedProperty(vertex_);
-    } else {
-      transaction_->constraint_verification_info.RemovedProperty(vertex_);
-    }
-  }
-
-  return true;
+  bool result{false};
+  utils::OOMBlockerBlock oom_blocker_block{
```
as51340, line 295: same comm

as51340 (changes requested):
I don't understand why we need `DoCheck`?

tests/stress/memory_limit.py
```diff
@@ -246,7 +257,7 @@ def execution_handler() -> None:
 
     with multiprocessing.Pool(processes=worker_count) as p:
         for worker in p.map(execute_function, workers):
-            log.info(f"Worker {worker.type} finished!")
+            print(f"Worker {worker.type} finished!")
```
as51340, line 260: Why is `log.info` replaced with `print`?

src/utils/oom_blocker_block.hpp
antoniofilipovic: For explicit template deduction  on constructor, but I think we don't need it since implicit deduction kicks in

src/storage/v2/vertex_accessor.cpp
antoniofilipovic: I can't do that as this way it doesn't produce OOMBlockerBlock, but just calls lambda...

src/utils/memory_tracker.hpp
antoniofilipovic: That would require additional logic and passing some flags or something when doing tracking from global_memory_tracker, so I wouldn't go there unless it is necessary.

tests/stress/memory_limit.py
antoniofilipovic: This was for my debugging, will remove it

as51340 (approved):
:muscle:

Ignition (changes requested):
Rather than block one kind of exception, why not make the code exception safe? e.g. if SetProperty has memory exception then need to unlink and delete delta inside a catch + rethrow

antoniofilipovic:
> Rather than block one kind of exception, why not make the code exception safe? e.g. if SetProperty has memory exception then need to unlink and delete delta inside a catch + rethrow

@Ignition I would say it would make code even more dependent on OOM Exception checks. If I understood correctly, in that case I would need to have Try Catch blocks inside the SetProperty function in VertexAccessor, which means that an exception would be thrown, we will have stack unwinding there and afterward we need to process exception and do additional steps to revert changes. This way it is in my opinion more performant and "cleaner" to have Atomic memory block, which doesn't allow to throw exception, and it is more reusable also.

src/utils/atomic_memory_block.hpp
```diff
@@ -0,0 +1,44 @@
...
+  void operator()() {
+    {
+      utils::MemoryTracker::OutOfMemoryExceptionBlocker oom_blocker;
+      function_();
+    }
+    total_memory_tracker.DoCheck();
+  }
+
+ private:
+  std::function<void()> function_;
```
Ignition, line 41: Don't use `std::function`, use `Callable`

Ignition (approved):
For we need now, this works. 
One small change requested.

PR #1601, opened 2023-12-21
HA I: Coordinator communicating with main and replicas

as51340:
[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR


src/dbms/replication_handler.cpp
```diff
@@ -180,28 +244,83 @@ auto ReplicationHandler::RegisterReplica(const memgraph::replication::Replicatio
...
 
+#ifdef MG_ENTERPRISE
+  auto const coordinator_handler =
+      [this, name](replication::RoleCoordinatorData &coordinatorData) -> UnregisterReplicaResult {
+    if (!dbms_handler_.ReplicationState().TryPersistUnregisterReplicaOnMain(name)) {
+      return UnregisterReplicaResult::COULD_NOT_BE_PERSISTED;
+    }
+
+    // Remove database specific clients
+    dbms_handler_.ForEach([name](Database *db) {
```
andrejtonev, line 306: Why does the coordinator care about storage clients?
Shouldn't it handle just the high-level?

src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -316,6 +316,30 @@ antlrcpp::Any CypherMainVisitor::visitEdgeImportModeQuery(MemgraphCypher::EdgeIm
 antlrcpp::Any CypherMainVisitor::visitSetReplicationRole(MemgraphCypher::SetReplicationRoleContext *ctx) {
   auto *replication_query = storage_->Create<ReplicationQuery>();
   replication_query->action_ = ReplicationQuery::Action::SET_REPLICATION_ROLE;
+
+#ifdef MG_ENTERPRISE
+  // Licence check in the interpreter.
+  if (ctx->MAIN() || ctx->REPLICA()) {
+    if (!ctx->WITH() || !ctx->PORT()) {
+      throw SemanticException("Port must be specified when setting replication role to main or replica!");
```
andrejtonev, line 324: I think the port should be optional. We might be enterprise and still not use the coordinator. The port is superfluous in that case.

andrejtonev (commented):
I didn't get to go through the whole thing, but I have some overall comments.
1) The coordinator should not be part of the main/replica.
2) Regarding the MAIN registration, the coordinator shouldn't care if something is main/replica. There should be a mechanism that lets it figure that out via RPC.
3) The coordinator role should be set at startup, not through cypher.

We can discuss this in more detail next week. :christmas_tree:

as51340:
> I didn't get to go through the whole thing, but I have some overall comments.
> 
> 1. The coordinator should not be part of the main/replica.
> 2. Regarding the MAIN registration, the coordinator shouldn't care if something is main/replica. There should be a mechanism that lets it figure that out via RPC.
> 3. The coordinator role should be set at startup, not through cypher.
> 
> We can discuss this in more detail next week. 🎄

1. and 3. are basically the same point. I think both approaches are valid because in the end things will get changed. I mentioned that the end goal is to have coordinators as part of main/replica not as separate instances but at the moment I don't see why is this a problem as an intermediate step.
2. Again, this is an implementation detail, for now it shouldn't be a problem.

PR #1602, opened 2023-12-21
Add query params for labels

### LINKED ISSUE #1405: Setting label as a variable not possible in Memgraph

DavIvek:
Currently, we are able to use query parameters with property values and intermediate variables. However, labels are hardcoded in the query and no one is currently able to set an arbitrary label via the query parameters.
This PR should enable users to set an label via the query parameters.
Also, this PR came out of #1448 which is closed, because approach was probably wrong.

[master < Task] PR

Release note: Besides using parameters with properties, now it's possible to use them with labels.

closes #1405


DavIvek:
TODO:

- [ ] write tests

src/query/cypher_query_interpreter.cpp
```diff
@@ -10,9 +10,12 @@
 // licenses/APL.txt.
 
 #include "query/cypher_query_interpreter.hpp"
+#include <string_view>
+#include "query/frontend/ast/ast.hpp"
```
antoniofilipovic, line 14: Do you need these 2 includes

antoniofilipovic (commented):
All looks good, just check on these 2 includes :muscle:

src/query/cypher_query_interpreter.cpp
DavIvek: no, I'll remove them.

antoniofilipovic (approved):

PR #1606, opened 2023-12-27
Fix index hints

### LINKED ISSUE #1592: Memgraph doesn't respect to Index Hinting
**Memgraph version**
2.13.0

**Environment**
CentOS 7

**Describe the bug**
Memgraph doesn't consider using the Index that is instructed to be used with `USING INDEX`.

**To Reproduce**
I have about 14 million nodes with Device label (`id` property is indexed) and about 50 thousand nodes with Topic label (`id` property is indexed). When I run the following command, the planner decides to use `Device (id)` index instead of `Topic (id)` index:

```
USING INDEX :Topic(id) MATCH (d:Device)-[:SUBSCRIBE]->(t:Topic) WHERE t.id IN ['testDevices'] AND d.id > '0' RETURN d.token, d.platform LIMIT 10000;
```

Query Plan:
```
memgraph> EXPLAIN USING INDEX :Topic(id) MATCH (d:Device)-[:SUBSCRIBE]->(t:Topic) WHERE t.id IN ['testDevices'] AND d.id > '0' RETURN d.token, d.platform LIMIT 10000;
+---------------------------------------------------+
| QUERY PLAN                                        |
+---------------------------------------------------+
| " * Limit"                                        |
| " * Produce {d.token, d.platform}"                |
| " * Filter (t :Topic), {t.id}"                    |
| " * Expand (d)-[anon1:SUBSCRIBE]->(t)"            |
| " * ScanAllByLabelPropertyRange (d :Device {id})" |
| " * Once"                                         |
+---------------------------------------------------+
```

**Expected behavior**
I expect the planner to use `Topic (id)` index:

```
+--------------------------------------------------+
| QUERY PLAN                                       |
+--------------------------------------------------+
| " * Limit"                                       |
| " * Produce {d.token, d.platform}"               |
| " * Filter (d :Device)"                          |
| " * Expand (t)<-[anon1:SUBSCRIBE]-(d)"           |
| " * ScanAllByLabelPropertyValue (t :Topic {id})" |
| " * Unwind"                                      |
| " * Once"                                        |
+--------------------------------------------------+
```

DavIvek:
The issue is that even if the user gives a hint to memgraph's planner which index to use, memgraph will still generate multiple plans (with the user's hint included) and compare their costs based on its query cost estimator. 
To resolve the issue I added one more return value(from cost estimator) that tells us if the plan is generated by using index hints. Plans generated by using index hints will have bigger priority than ones that don't. 

[master < Task] PR

Release note: Index hints fixed to also work with queries that are using expanding from one node to another in their plans. 

closes #1592


src/query/plan/planner.hpp
```diff
@@ -59,9 +59,9 @@ class PostProcessor final {
   }
 
   template <class TVertexCounts>
-  double EstimatePlanCost(const std::unique_ptr<LogicalOperator> &plan, TVertexCounts *vertex_counts,
-                          const SymbolTable &table) {
-    return query::plan::EstimatePlanCost(vertex_counts, table, parameters_, *plan);
+  std::pair<double, bool> EstimatePlanCost(const std::unique_ptr<LogicalOperator> &plan, TVertexCounts *vertex_counts,
+                                           const SymbolTable &table) {
+    return query::plan::EstimatePlanCost(vertex_counts, table, parameters_, *plan, index_hints_);
   }
```
antepusic, line 65:
```
double cost = post_process.EstimatePlanCost(rewritten_plan, dba, symbol_table).first;
```
The `.first` in the above line is a bit arbitrary. I think the code will be more maintainable if `EstimatePlanCost` returned a struct. That way, you can take advantage of named fields.

tests/benchmark/query/planner.cpp
antepusic, line 1: Can you add a test that shows Memgraph will use the hinted plan instead of the equal-cost plan that doesn’t use index hints?

antepusic: This PR means that there are two factors in plan selection: total cost and the presence of index hints. I think it would be a good idea to add a test showing that total cost takes precedence over index hinting, i.e. that Memgraph will prefer the less costly plan with no index hints over a more costly one with index hints.

antepusic (changes requested):
Good work! Left a few comments for the sake of maintainability and thorough testing.

antepusic (approved):
Thanks for addressing my comments! I think this is approvable now.

Josipmrden (changes requested):

Josipmrden:
Please just alter the comments in planner.hpp to make the logic more readable

Josipmrden (approved):

PR #1607, opened 2023-12-27
Fix deleting replica gets main stuck

### LINKED ISSUE #1527: Deleting a SYNC replica gets MAIN stuck in kubernetes
**Memgraph version**
2.12.0

**Environment**
Local KIND kubernetes cluster on M1 macbook, running a slightly modified (to enable it to run multiple nodes) version of official helm chart (https://github.com/memgraph/helm-charts/pull/12)

**Describe the bug**
Creating a fresh cluster with 1 MAIN and 2 SYNC REPLICAs, and then restarting one of the REPLICAs gets the MAIN stuck.

When checking the `show replicas` on MAIN at the time of restarting a REPLICA, the REPLICA state goes from ready to invalid, shortly after which `show replicas` stops working on MAIN until MAIN is restarted. When MAIN is restarted, it actually picks up the new IP of REPLICA and continues working as expected (i.e. fully recovers).

**To Reproduce**
 
values file:
```
image:
  tag: 2.12.0

memgraphConfig:
  - "--also-log-to-stderr=true"
  - "--replication-restore-state-on-startup=true"

replicaCount: 3
```

in the helm chart from PR linked above run this:
`helm upgrade --install memgraph . -n memgraph -f values-file-from-above.yaml`

this creates 3 pods, `memgraph-0`, `memgraph-1`, `memgraph-2` - 0 will be MAIN, 1 and 2 will be replicas. Lets set that up:
```
# on both replicas run:
SET REPLICATION ROLE TO REPLICA WITH PORT 10000;

# on main run:
REGISTER REPLICA one SYNC TO "memgraph-1.memgraph.memgraph.svc.cluster.local:10000";
REGISTER REPLICA two SYNC TO "memgraph-2.memgraph.memgraph.svc.cluster.local:10000";
```

delete the pod memgraph-2. this causes the problem described above

**Expected behavior**
After pod memgraph-2 comes back, DNS resolution should be repeated e.g. with backoff:
- first time we try resolving that DNS record, it will likely fail because old pod is gone, and new one still doesn't exist
- a couple of seconds later the DNS resolution should yield a new ip and cluster should continue working as expected
right now this happens only after MAIN is restarted too

**Logs**
main logs show the following after delete of replica:
```
[2023-11-17 14:53:53.875] [memgraph_log] [error] Couldn't replicate data to two. For more details, visit https://memgr.ph/replication.
[2023-11-17 14:56:05.145] [memgraph_log] [error] [client.hpp:154] Couldn't connect to remote address 10.244.0.96:10000
[2023-11-17 14:56:05.145] [memgraph_log] [error] Failed to connect to replica two at the endpoint 10.244.0.96:10000. For more details, visit https://memgr.ph/replication.
```

**Additional context**

This ticket originated from this discord thread, more details can be found there: https://discord.com/channels/842007348272169002/890968055339958292/1175071185793450064

**Verification Environment**
Docker image, probably ARM

DavIvek:
[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR

closes #1527


src/io/network/endpoint.cpp
```diff
@@ -166,6 +165,7 @@ bool Endpoint::IsResolvableAddress(const std::string &address, uint16_t port) {
   };
   addrinfo *info = nullptr;
   auto status = getaddrinfo(address.c_str(), std::to_string(port).c_str(), &hints, &info);
+  if (status != 0) throw NetworkError(gai_strerror(status));
```
andrejtonev, line 168: This function didn't throw before. All usages are checking the return `bool`. Seems like it would be better to `return` false instead of throwing. 
If you throw, you would need to check that all execution paths are handling the throw correctly. First impression's that we are not try/catching and just relying on the bool.

src/io/network/socket.cpp
```diff
@@ -55,17 +56,21 @@ bool Socket::IsOpen() const { return socket_ != -1; }
...
+      }
+      // If the connect failed close the file descriptor to prevent file
+      // descriptors being leaked
+      close(sfd);
     }
-    // If the connect failed close the file descriptor to prevent file
-    // descriptors being leaked
-    close(sfd);
+  } catch (const NetworkError &e) {
+    return false;
```
andrejtonev, line 73: What throws above?
Could we have a file descriptor that needs closing?

PR #1609, opened 2024-01-02
Fix storage mode flag

### LINKED ISSUE #1057: Implement recovery of the STORAGE MODE
Once Memgraph ends up in the storage mode `IN_MEMORY_ANALYTICAL` it should stay in that mode after restart if the recovery part is set properly (data directory, recovery on startup, etc)

### LINKED ISSUE #1599: Storage mode flag not working as expected
**Memgraph version**
2.13

**Environment**
Docker, Mac M1

**Describe the bug**
When I set flag --storage-mode to IN_MEMORY_ANALYTICAL it doesn't actually change the storage mode, but only the value of the --storage-mode flag in memgraph.conf.

**To Reproduce**
1. Run Memgraph with: `docker run --name memgraph-test -p 7687:7687 -p 7444:7444 memgraph/memgraph:2.13.0 --storage-mode=IN_MEMORY_ANALYTICAL`

**Expected behavior**
Storage mode is changed to IN MEMORY ANALYTICAL. 

**Logs**
Screenshot 2023-12-20 at 13 36 20
Screenshot 2023-12-20 at 13 36 34

**Additional context**
I suspect that the issue is not only in the result of SHOW STORAGE MODE query, but that the storage mode isn't actually changed. The reason for that is because user tried it and import is not any faster.

imilinovic:
Storage mode configuration flag always set the mode to TRANSACTIONAL instead of the given mode.

[master < Epic] PR
- [ ] Write E2E tests

closes #1599 
closes #1057

[master < Task] PR

--storage-mode configuration flag is now working properly

### DIFF
--- a/src/storage/v2/inmemory/storage.cpp
+++ b/src/storage/v2/inmemory/storage.cpp
@@ -1,4 +1,4 @@
-// Copyright 2023 Memgraph Ltd.
+// Copyright 2024 Memgraph Ltd.
 //
 // Use of this software is governed by the Business Source License
 // included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
@@ -153,7 +153,7 @@ InMemoryStorage::InMemoryStorage(Config config, StorageMode storage_mode)
   }
 }
 
-InMemoryStorage::InMemoryStorage(Config config) : InMemoryStorage(config, StorageMode::IN_MEMORY_TRANSACTIONAL) {}
+InMemoryStorage::InMemoryStorage(Config config) : InMemoryStorage(config, config.storage_mode) {}
 
 InMemoryStorage::~InMemoryStorage() {
   if (config_.gc.type == Config::Gc::Type::PERIODIC) {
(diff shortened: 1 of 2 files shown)


src/storage/v2/inmemory/storage.cpp
```diff
@@ -153,8 +153,6 @@ InMemoryStorage::InMemoryStorage(Config config, StorageMode storage_mode)
   }
 }
 
-InMemoryStorage::InMemoryStorage(Config config) : InMemoryStorage(config, StorageMode::IN_MEMORY_TRANSACTIONAL) {}
-
```
imilinovic, line 157: IN_MEMORY_TRANSACTIONAL is the default value of Config so the behaviour is the same.

src/storage/v2/inmemory/storage.cpp
```diff
@@ -65,14 +65,14 @@ auto FindEdges(const View view, EdgeTypeId edge_type, const VertexAccessor *from
 
 using OOMExceptionEnabler = utils::MemoryTracker::OutOfMemoryExceptionEnabler;
 
-InMemoryStorage::InMemoryStorage(Config config, StorageMode storage_mode)
-    : Storage(config, storage_mode),
+InMemoryStorage::InMemoryStorage(Config config)
+    : Storage(config, config.storage_mode),
```
imilinovic, line 69: IN_MEMORY_TRANSACTIONAL is set as default value in Config::storage_mode so the default behaviour will be the same

as51340 (approved):
:muscle:

DavIvek (approved):

PR #1610, opened 2024-01-03
Add support for triggers in database dump

### LINKED ISSUE #1198: Add backup and recovery of triggers via DUMP DATABASE command
Related issue -> https://github.com/memgraph/mgconsole/issues/64

DavIvek:
[master < Task] PR

closes #1198


DavIvek:
TODO:

- [ ] configure existing tests
- [ ] write tests

src/query/dump.cpp
```diff
@@ -536,6 +552,25 @@ PullPlanDump::PullChunk PullPlanDump::CreateInternalIndexCleanupPullChunk() {
   };
 }
 
-void DumpDatabaseToCypherQueries(query::DbAccessor *dba, AnyStream *stream) { PullPlanDump(dba).Pull(stream, {}); }
+PullPlanDump::PullChunk PullPlanDump::CreateTriggersPullChunk() {
+  return [this](AnyStream *stream, std::optional<int>) {
+    if (trigger_store_->HasTriggers()) {
```
as51340, line 557: No need for check, if there are none `GetTriggerInfo` will return empty vector and you will iterate over empty vector

src/query/dump.cpp
```diff
@@ -536,6 +552,25 @@ PullPlanDump::PullChunk PullPlanDump::CreateInternalIndexCleanupPullChunk() {
...
+      auto triggers = trigger_store_->GetTriggerInfo();
+      for (auto &trigger : triggers) {
+        std::ostringstream os;
+        std::replace(trigger.statement.begin(), trigger.statement.end(), '\n', ' ');
```
as51340, line 561: I think this should be copied not changed in place since you then lose information about the original statement. Not a problem currently but could be for a future me or you who don't know this small caveat

src/query/interpreter.cpp
```diff
@@ -1818,7 +1818,7 @@ PreparedQuery PrepareProfileQuery(ParsedQuery parsed_query, bool in_explicit_tra
 PreparedQuery PrepareDumpQuery(ParsedQuery parsed_query, CurrentDB &current_db) {
   MG_ASSERT(current_db.execution_db_accessor_, "Dump query expects a current DB transaction");
   auto *dba = &*current_db.execution_db_accessor_;
-  auto plan = std::make_shared<PullPlanDump>(dba);
+  auto plan = std::make_shared<PullPlanDump>(dba, current_db.db_acc_->get()->trigger_store());
```
as51340, line 1821: Unchecked access to db_acc. AFAIK, db_acc should always have value but again, someone may change the code and we are in troubles. Ask @andrejtonev if db_acc can ever be without value. And if no, can we remove std::optional

as51340 (changes requested):
Good job but check some small things I suggested

src/query/interpreter.cpp
andrejtonev: Quick background:
The `memgraph::dbms::DatabaseAccess` is a mechanism used to make sure a database exists. The user can CREATE/DROP databases at any point. The access is a reference counter that protects a database from being dropped while used by someone/something else.

At this point the `db_acc_` has to have a value. The problem is that the plan will be executed in the future and we don't control when that is.
What you should do is copy the `DatabaseAccess` itself into the plan and call `db_acc->trigger_store()` from the plan. This way the database cannot be dropped until the plan has been executed.

src/query/dump.cpp
```diff
@@ -21,9 +22,13 @@
 
 #include <fmt/format.h>
 
+#include "dbms/database.hpp"
 #include "query/db_accessor.hpp"
 #include "query/exceptions.hpp"
+#include "query/path.hpp"
 #include "query/stream.hpp"
+#include "query/stream/streams.hpp"
```
as51340, line 30: Not sure this is necessary to include

src/query/dump.cpp
```diff
@@ -21,9 +22,13 @@
 
 #include <fmt/format.h>
 
+#include "dbms/database.hpp"
 #include "query/db_accessor.hpp"
 #include "query/exceptions.hpp"
+#include "query/path.hpp"
```
as51340, line 28: Also this

as51340, line 28: Is this necessary to include?

src/query/dump.cpp
as51340, line 30: Is this necessary?

src/query/dump.hpp
```diff
@@ -13,23 +13,26 @@
 
 #include <ostream>
 
+#include "dbms/database.hpp"
 #include "query/db_accessor.hpp"
 #include "query/stream.hpp"
+#include "query/trigger.hpp"
```
as51340, line 19: no need to include here

as51340 (approved):
Good, just remove extra includes and ready to go :muscle:

PR #1615, opened 2024-01-04
Fix REPLICA timestamps

Ignition:
Subtle bug where storage timestamp was being changed (for GetInfo and CreateSnapshot) as if it was MAIN when it was actually REPLICA.


gvolfing (approved):
Looks good to me.

We might have some other places where we do not put throwing sys/socket related functions into try blocks.

Ignition:
related to #1359

PR #1618, opened 2024-01-05
Allow clang compilers with different names

Ignition:
Rather than inspect for exactly binary name of "clang", check the compiler ID instead.


Ignition:
> Isn't this more confusing because even if you activate the toolchain, users must set `CC` and `CXX` explicitly (that's not obvious, especially for a new user)? 🤔

Yes, ATM it is. I didn't want to make a toolchain changes to add CC + CXX to the sourced environment. I could change v4/v5 toolchain, then rebuild CI builder image?

PR #1620, opened 2024-01-05
Extend ToBoolean function so it casts "t" to true and "f" to false.

### LINKED ISSUE #1538: ToBoolean casts the strings "t" and "f" to null
**Memgraph version**
2.11

**Describe the bug**
For csv like
```
id,is_a_fun_guy_at_parties
3,f
42,t
```

```
LOAD CSV FROM $file_path
WITH header NULLIF '' AS row
MERGE (a:Node { id: ToInteger(row.node_id) })
SET a.isAFunGuyAtParties = ToBoolean(row.is_a_fun_guy_at_parties)
```

All `Node` entities loaded like so will have a null value for `isAFunGuyAtParties`. Since, e.g., postgresql COPY to CSV will record true values as `t` and false values as `f`, it may be worth updating `ToBoolean` to accept the string 't' as true, and 'f' as false. Otherwise, updating the documentation to describe which values are treated as t/f could be helpful.

DavIvek:
User from community requested this extension for ToBoolean function. Since, e.g., postgresql COPY to CSV will record true values as "t" and false values as "f". 
We also might need to update documentation so it's clear what values ToBoolean casts to true or false.

[master < Task] PR

Release note: Strings "t" and "f" are now resolved into true and false.

closes #1538

### DIFF
--- a/src/query/interpret/awesome_memgraph_functions.cpp
+++ b/src/query/interpret/awesome_memgraph_functions.cpp
@@ -1,4 +1,4 @@
-// Copyright 2023 Memgraph Ltd.
+// Copyright 2024 Memgraph Ltd.
 //
 // Use of this software is governed by the Business Source License
 // included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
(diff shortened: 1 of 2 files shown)


DavIvek:
NOTE: `ToBoolean(t)` will still return an error since it's resolved as a symbol. It doesn't make sense to me to resolve it as a boolean value. 
The same thing is with boolean values not being resolved as symbols.
For example let's say we want to match a node from a database:
`MATCH (truE:Node) return truE;` 
After executing above query we will get n rows (n being a number of nodes with label Node) of true values. `truE` won't get resolved into symbol.

imilinovic (approved):
Looks good

as51340 (approved):

PR #1629, opened 2024-01-09
Fix using path in property value after CREATE

### LINKED ISSUE #1331: Crash caused when using path in property value
I executed a query on my empty memgraph database and get a crash.

memgraph version:2.11.0
operating system: windows 11
API：Cypher version 9
query:CREATE p0=()-[:T0]->() MERGE ({k:(size(p0))}) RETURN 1
error message:Query failed: Connection was closed by server

Actual behavior:
The query failed and the memgraph container exited abnormally with the error message.

DavIvek:
Next query would crash database:
`CREATE p0=()-[:T0]->() MERGE ({k:(size(p0))}) RETURN 1`
Problem is that Memgraph isn't updating bound symbols with path identifiers in CREATE. Mentioned problem will be resolved with this PR. 

[master < Task] PR

Release note: It's possible to use path identifier from CREATE clause in other parts of the query.

closes #1331

### DIFF
--- a/src/query/plan/rule_based_planner.hpp
+++ b/src/query/plan/rule_based_planner.hpp
@@ -1,4 +1,4 @@
-// Copyright 2023 Memgraph Ltd.
+// Copyright 2024 Memgraph Ltd.
 //
 // Use of this software is governed by the Business Source License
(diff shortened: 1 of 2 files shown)


antepusic (approved):
This should fix the issue :+1:

PR #1635, opened 2024-01-11
Fix skip list cpp

Ignition:
SkipList's iterators and find methods had some issues. 

Fixing them up so that they work as expected.

### DIFF
--- a/src/utils/skip_list.hpp
+++ b/src/utils/skip_list.hpp
@@ -1,4 +1,4 @@
-// Copyright 2023 Memgraph Ltd.
+// Copyright 2024 Memgraph Ltd.
 //
 // Use of this software is governed by the Business Source License
 // included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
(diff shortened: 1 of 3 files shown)


src/utils/skip_list.hpp
```diff
@@ -591,13 +591,20 @@ class SkipList final : detail::SkipListNode_base {
...
 
    public:
-    TObj &operator*() const { return node_->obj; }
+    using value_type = TObj;
+    using difference_type = std::ptrdiff_t;
 
-    TObj *operator->() const { return &node_->obj; }
+    Iterator() = default;
+    Iterator(Iterator const &) = default;
+    Iterator(Iterator &&) = default;
```
andrejtonev, line 599: noexcept

src/utils/skip_list.hpp
```diff
@@ -591,13 +591,20 @@ class SkipList final : detail::SkipListNode_base {
...
+    Iterator &operator=(Iterator const &) = default;
+    Iterator &operator=(Iterator &&) = default;
```
andrejtonev, line 601: noexcept

src/utils/skip_list.hpp
```diff
@@ -591,13 +591,20 @@ class SkipList final : detail::SkipListNode_base {
     Iterator(TNode *node) : node_(node) {}
 
    public:
-    TObj &operator*() const { return node_->obj; }
+    using value_type = TObj;
+    using difference_type = std::ptrdiff_t;
 
-    TObj *operator->() const { return &node_->obj; }
+    Iterator() = default;
```
andrejtonev, line 597: Shouldn't this be deleted or private?
We don't handle `node_ == nullptr` so lets make `Iterator(TNode *node)` the only constructor.

andrejtonev: Perhaps you actually wanted to do a destructor, since we are missing it.

src/utils/skip_list.hpp
```diff
@@ -610,8 +617,14 @@ class SkipList final : detail::SkipListNode_base {
       }
     }
 
+    Iterator operator++(int) {
+      Iterator old = *this;
+      ++(*this);
+      return old;
+    }
```
andrejtonev, line 624: What's the use case for this?

src/utils/skip_list.hpp
```diff
@@ -591,13 +591,20 @@ class SkipList final : detail::SkipListNode_base {
     Iterator(TNode *node) : node_(node) {}
 
    public:
-    TObj &operator*() const { return node_->obj; }
+    using value_type = TObj;
+    using difference_type = std::ptrdiff_t;
```
andrejtonev, line 595: Should define all 5 tags

src/utils/skip_list.hpp
```diff
@@ -610,8 +617,14 @@ class SkipList final : detail::SkipListNode_base {
...
+      ++(*this);
+      return old;
+    }
+
    private:
-    TNode *node_;
+    TNode *node_{nullptr};
   };
 
   class ConstIterator final {
```
andrejtonev, line 630: Same comments as for the `Iterator`

andrejtonev (commented):
Nice. Just a few comments.

andrejtonev (approved):

PR #1642, opened 2024-01-15
Fix accumulated path evaluation in builtin algorithms.

Darych:
Fix accumulated path evaluation in DFS algorithm.
Add checks for Path expansions.
Add shrink function to Path class.


Darych:
Fix accumulated path evaluation in BFS, AllShortestPaths and WeightedShortestPath algos too.

src/query/plan/operator.cpp
```diff
@@ -1558,12 +1564,14 @@ class SingleSourceShortestPathCursor : public query::plan::Cursor {
...
       frame[self_.filter_lambda_.inner_edge_symbol] = edge;
       frame[self_.filter_lambda_.inner_node_symbol] = vertex;
+      std::optional<Path> curr_acc_path = std::nullopt;
       if (self_.filter_lambda_.accumulated_path_symbol) {
         MG_ASSERT(frame[self_.filter_lambda_.accumulated_path_symbol.value()].IsPath(),
                   "Accumulated path must have Path type");
         Path &accumulated_path = frame[self_.filter_lambda_.accumulated_path_symbol.value()].ValuePath();
         accumulated_path.Expand(edge);
         accumulated_path.Expand(vertex);
+        curr_acc_path = accumulated_path;
```
antoniofilipovic, line 1574: Why are we making here a copy of the whole path, shouldn't this be a memory and performance issue?

src/query/plan/operator.cpp
```diff
@@ -1578,17 +1586,26 @@ class SingleSourceShortestPathCursor : public query::plan::Cursor {
...
       processed_.emplace(vertex, edge);
     };
 
     // populates the to_visit_next_ structure with expansions
     // from the given vertex. skips expansions that don't satisfy
     // the "where" condition.
-    auto expand_from_vertex = [this, &expand_pair](const auto &vertex) {
+    auto expand_from_vertex = [this, &pull_memory, &frame, &expand_pair](const auto &vertex) {
       if (self_.common_.direction != EdgeAtom::Direction::IN) {
+        Path curr_acc_path(pull_memory);
```
antoniofilipovic, line 1598: I think instead of using pull memory, you could use frame memory since each object on frame is instantiated with "frame" memory. Because when we are doing move with objects using different memory we will actually have copy

src/query/plan/operator.cpp
```diff
@@ -1657,11 +1674,17 @@ class SingleSourceShortestPathCursor : public query::plan::Cursor {
       }
 
       // expand only if what we've just expanded is less then max depth
-      if (static_cast<int64_t>(edge_list.size()) < upper_bound_) expand_from_vertex(expansion.second);
+      if (static_cast<int64_t>(edge_list.size()) < upper_bound_) {
+        if (self_.filter_lambda_.accumulated_path_symbol) {
+          MG_ASSERT(curr_acc_path.has_value(), "Expected non-null accumulated path");
+          frame[self_.filter_lambda_.accumulated_path_symbol.value()] = curr_acc_path.value();
```
antoniofilipovic, line 1680: Can you use move here on curr_acc_path

src/query/plan/operator.cpp
```diff
@@ -1558,12 +1564,14 @@ class SingleSourceShortestPathCursor : public query::plan::Cursor {
 #endif
       frame[self_.filter_lambda_.inner_edge_symbol] = edge;
       frame[self_.filter_lambda_.inner_node_symbol] = vertex;
+      std::optional<Path> curr_acc_path = std::nullopt;
```
antoniofilipovic, line 1567: One idea, can you make curr_acc_path a std::shared_ptr so in that case when you change value on frame, you create a new shared_ptr but in other case you just use same and don't make copy every time?

src/query/plan/operator.cpp
```diff
@@ -1578,17 +1586,26 @@ class SingleSourceShortestPathCursor : public query::plan::Cursor {
...
+        Path curr_acc_path(pull_memory);
+        if (self_.filter_lambda_.accumulated_path_symbol) {
+          curr_acc_path = frame[self_.filter_lambda_.accumulated_path_symbol.value()].ValuePath();
+        }
         auto out_edges = UnwrapEdgesResult(vertex.OutEdges(storage::View::OLD, self_.common_.edge_types)).edges;
-        for (const auto &edge : out_edges) expand_pair(edge, edge.To());
+        for (const auto &edge : out_edges) {
+          expand_pair(edge, edge.To());
+          if (self_.filter_lambda_.accumulated_path_symbol) {
+            frame[self_.filter_lambda_.accumulated_path_symbol.value()] = curr_acc_path;
```
antoniofilipovic, line 1606: Why are you getting path on line 1600, and then copying it here, it shouldn't change in the meantime right?

antoniofilipovic (changes requested):
This seems like good upgrade on our expansions, I am little worried about memory issues and performance, did you do any measures on bigger graphs (100K nodes, 1M edges) with expansions using path?

src/query/plan/operator.cpp
Darych: I need to put accumulated_path to queue to save the state.
I can put the value of accumulated_path from frame to the queue. But it would be also path copying.

src/query/plan/operator.cpp
Darych: It's changed inside expand_pair. I restore the state of Path. But I can call Shrink() to remove last added part instead of copying full path. This will be better

Josipmrden:
these queries on dataset in the file `graph_edges.cypher` crash my database:
`match p=(n)<-[*BFS (r, n, p | size(nodes(p)) > 0)]-(m {id:1}) return p;`
`match p=(n)<-[*BFS (r, n, p | r.id > 0)]-(m {id:1}) return p;`

these ones are okay
`match p=(n)<-[*BFS (r, n | r.id > 0)]-(m {id:1}) return p;`
`match p=(n)<-[*WSHORTEST (r, n | 1) total_weight (r, n | r.id > 0)]-(m {id:1}) return p;`
`match p=(n)<-[*WSHORTEST (r, n | 1) total_weight (r, n, p | r.id > 0)]-(m {id:1}) return p;`
`match p=(n)<-[*WSHORTEST (r, n | 1) total_weight (r, n, p | size(nodes(p)) > 0)]-(m {id:1}) return p;`

Josipmrden (changes requested):

src/query/plan/operator.cpp
antoniofilipovic: So let's say you need multiple same paths for expansion from one node, that could be shared pointer, right?

src/query/plan/operator.cpp
Darych: Yes, correct, Josip.

src/query/plan/operator.cpp
Darych: We discussed in the call about this.
Based on the latest changes, we will not need shared_ptr there.
Resolving conversation.

antoniofilipovic (approved):
LGTM

Josipmrden:
Can you just add one of these queries to gql behave which crashed the db and then we're good?

Josipmrden (commented):
Left comment outside the review

Josipmrden (approved):

PR #1646, opened 2024-01-18
HA: Automatic failover

as51340:
Here we introduce automatic failover. Coordinator is constantly pinging other instances (one instance-one thread) and when it realizes that the main instance is down for 5s, it promotes first alive replica to become main. If promoting replica to main failed, coordinator remains consistent and will try to promote replica to main on the next ping. Coordinator is pinging instances every second. Restoration of roles is not supported.

[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR


src/coordination/coordinator_data.cpp
```diff
@@ -0,0 +1,215 @@
...
+    auto instance = std::ranges::find_if(
+        coord_data->registered_instances_,
+        [instance_name](const CoordinatorInstance &instance) { return instance.InstanceName() == instance_name; });
+
+    MG_ASSERT(instance != coord_data->registered_instances_.end(), "Instance {} not found during callback!",
+              instance_name);
+    return *instance;
+  };
+
+  replica_succ_cb_ = [find_instance](CoordinatorData *coord_data, std::string_view instance_name) -> void {
```
andrejtonev, line 36: This comment goes for all callbacks here.
1) Why do we create these callbacks? Can't we just put them inside the frequent check?
2) I don't think we want to `MG_ASSERT` here. We need to be supper robust. Check and handle it.

src/coordination/coordinator_data.cpp
```diff
@@ -0,0 +1,215 @@
...
+          spdlog::warn("Failover aborted since promoting replica to main failed!");
+          break;
+        case SUCCESS:
+          break;
+      }
+    }
+  };
+}
+
+auto CoordinatorData::DoFailover() -> DoFailoverStatus {
```
andrejtonev, line 81: A general comment for Failover.
REPLICAs currently ingest any data coming in from "MAIN". I don't think failover will work until we make the REPLICA-MAIN logic more robust.
The coordinator needs to label each instance and tell REPLICAs who the true MAIN is.
There could be a new field in the RPC that IDs the originating instance and REPLICAs ignore anything coming in from a different ID.

If we do failover with the current setup, we can have a situation where MAIN loses the connection to COORDINATOR, but does no lose the connection to REPLICAs and will continue sending messages to them, which they will ingest no questions asked.

src/coordination/coordinator_data.cpp
andrejtonev: If we really need to reuse the logic in these callbacks, why not create member functions?
Uses less space and is faster.

src/coordination/coordinator_data.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+  chosen_replica_instance->PrepareForFailover();
+
+  std::vector<ReplicationClientInfo> repl_clients_info;
+  repl_clients_info.reserve(std::ranges::distance(replica_instances));
+
+  auto not_chosen_replica_instance = [&chosen_replica_instance](const CoordinatorInstance &instance) {
+    return instance != *chosen_replica_instance;
+  };
+  auto not_main = [](const CoordinatorInstance &instance) { return !instance.IsMain(); };
```
andrejtonev, line 100: std::move latter or just inline it in the filter

src/coordination/coordinator_data.cpp
```diff
@@ -0,0 +1,215 @@
...
+  // PROMOTE REPLICA TO MAIN
+  // THIS SHOULD FAIL HERE IF IT IS DOWN
+  if (auto result = registered_replica->client_.SendPromoteReplicaToMainRpc(std::move(repl_clients_info)); !result) {
+    registered_replica->client_.StartFrequentCheck();
+    return SetInstanceToMainCoordinatorStatus::COULD_NOT_PROMOTE_TO_MAIN;
+  }
+
+  registered_replica->replication_role_ = replication_coordination_glue::ReplicationRole::MAIN;
+  registered_replica->client_.SetSuccCallback(main_succ_cb_);
+  registered_replica->client_.SetFailCallback(main_fail_cb_);
```
andrejtonev, line 180: Seems dangerous to start frequent check and then set the callbacks.
Why not set then and then start?

andrejtonev (commented):
Just a initial overview and some comments.
Looks good. Maybe we should sync before going farther.

Ignition (commented):
Did a few comments, bigger discussion I sent over slack.
Will re-review once we sync.

src/coordination/coordinator_data.cpp
as51340: Yes 100% agree, will be handled, probably not immediately but yes

src/coordination/coordinator_data.cpp
as51340: explain please, really not sure what do you mean here

src/coordination/coordinator_data.cpp
as51340: Yes, that's wrong

src/coordination/coordinator_client.cpp
```diff
@@ -27,82 +27,78 @@ auto CreateClientContext(const memgraph::coordination::CoordinatorClientConfig &
...
-    thread_pool_.Shutdown();
-  });
-  const auto endpoint = rpc_client_.Endpoint();
-  // Logging can throw
-  spdlog::trace("Closing replication client on {}:{}", endpoint.address, endpoint.port);
-}
+      config_{std::move(config)},
+      coord_data_{coord_data},
+      succ_cb_{std::move(succ_cb)},
+      fail_cb_{std::move(fail_cb)} {}
```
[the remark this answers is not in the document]
as51340: Observer/listener design pattern, I don't want coordinator client to know about failover logic, the client should think about communication

src/coordination/coordinator_data.cpp
```diff
@@ -0,0 +1,215 @@
...
+    auto &instance = find_instance(coord_data, instance_name);
+    MG_ASSERT(instance.IsReplica(), "Instance {} is not a replica!", instance_name);
+    instance.UpdateLastResponseTime();
+  };
+
+  replica_fail_cb_ = [find_instance](CoordinatorData *coord_data, std::string_view instance_name) -> void {
+    spdlog::trace("Instance {} performing replica failure callback", instance_name);
+    auto &instance = find_instance(coord_data, instance_name);
+    MG_ASSERT(instance.IsReplica(), "Instance {} is not a replica!", instance_name);
+    instance.UpdateInstanceStatus();
```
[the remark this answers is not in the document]
as51340: atm, we don't support unregistering instances but once we do, you are 100% right, removing atomic and doing this under the unique lock

src/coordination/coordinator_data.cpp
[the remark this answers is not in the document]
as51340: in this way callbacks just need find_instance lambda, otherwise they need this ptr to coordinator data. I just find it like this a bit more cleaner than having to think how was something captured in lambdas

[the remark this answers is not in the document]
as51340: plus usually in classic observers, this is regularly provided as argument to avoid cyclic dependency

as51340: Check comment about callbacks down

as51340: > I don't think we want to MG_ASSERT here. We need to be supper robust. Check and handle it.

This situation should never ever happen. This is just defensive programming to assure I didn't make some big mistake. But instance executing replica callback must be an instance, no recovery here that's logic failure if it happens

src/coordination/coordinator_data.cpp
antoniofilipovic: We didn't start it at all, that part is missing

antoniofilipovic:
```
registered_replica->replication_role_ = replication_coordination_glue::ReplicationRole::MAIN;
  registered_replica->client_.SetSuccCallback(main_succ_cb_);
  registered_replica->client_.SetFailCallback(main_fail_cb_);
 // this is missing
  registered_replica->client_.StartFrequentCheck();
```

src/coordination/coordinator_data.cpp
antoniofilipovic: Isn't this also fine since this just creates filter variable with callback?

src/coordination/coordinator_data.cpp
antoniofilipovic: Yes, also agree, we talked about this but want to just get feedback on generally how all this works

src/coordination/coordinator_data.cpp
```diff
@@ -111,39 +112,43 @@ auto CoordinatorData::DoFailover() -> DoFailoverStatus {
...
+auto CoordinatorData::ShowInstances() const -> std::vector<CoordinatorInstanceStatus> {
+  std::vector<CoordinatorInstanceStatus> instances_status;
+  instances_status.reserve(registered_instances_.size());
 
-  return CoordinatorInstanceStatus{.instance_name = main_instance->InstanceName(),
-                                   .socket_address = main_instance->SocketAddress(),
-                                   .is_alive = main_instance->IsAlive()};
-};
+  auto const stringify_repl_role = [](const CoordinatorInstance &instance) -> std::string {
+    if (!instance.IsAlive()) return "";
```
andrejtonev, line 129: Maybe return something "na", "unknown" or similar.

src/coordination/coordinator_client.cpp
```diff
@@ -27,82 +27,78 @@ auto CreateClientContext(const memgraph::coordination::CoordinatorClientConfig &
...
-            last_response_time->store(std::chrono::system_clock::now(), std::memory_order_acq_rel);
-          }
+          spdlog::trace("Sending frequent heartbeat to machine {} on {}", instance_name,
+                        rpc_client_.Endpoint().SocketAddress());
+          auto stream{rpc_client_.Stream<memgraph::replication_coordination_glue::FrequentHeartbeatRpc>()};
+          stream.AwaitResponse();
+          succ_cb_(coord_data_, instance_name);
         } catch (const rpc::RpcFailedException &) {
-          // Nothing to do...wait for a reconnect
+          fail_cb_(coord_data_, instance_name);
```
andrejtonev, line 56: Maybe a todo:
1) Try to avoid searching for the instance by name.
2) Have a single thread that goes through all instances.

andrejtonev (approved):
Just a couple of small comments.

src/coordination/coordinator_data.cpp
as51340: Will do it here https://github.com/memgraph/memgraph/pull/1671

src/coordination/coordinator_client.cpp
as51340: 1. there will be at most few instances, that's why searching them is basically constant operation in time
2. Currently, only one thread is searching instances, the one which is executing the code, could you explain a bit more?

src/coordination/coordinator_data.cpp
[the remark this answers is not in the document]
antoniofilipovic: discussed on call

src/coordination/coordinator_data.cpp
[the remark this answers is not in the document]
antoniofilipovic: discussed on call

Ignition (approved):

as51340:
@kgolubic release note: Added automatic failover from a single coordinator.

PR #1653, opened 2024-01-19
Add support for query params in load csv

### LINKED ISSUE #1386: Use query parameter with LOAD CSV command
**Memgraph version**
2.11.0

**Environment**
Memgraph docker image: `memgraph/memgraph:2.11.0`.

**Describe the bug**
I have a file with CSV data and I want to use `LOAD CSV FROM $file ...` and send `file` as a query parameter, but I get the following error:

```
line 1:15 extraneous input '$' expecting {ACTIVE, AFTER, ALTER, ANALYZE, ASYNC, AUTH, BAD, BATCH_INTERVAL, 
BATCH_LIMIT, BATCH_SIZE, BEFORE, BUILD, BOOTSTRAP_SERVERS, CALL, CHECK, CLEAR, COMMIT, COMMITTED, 
CONFIG, CONFIGS, CONSUMER_GROUP, CREATE_DELETE, CREDENTIALS, CSV, DATA, DELIMITER, DATABASE, DENY, 
DROP, DUMP, EDGE, EDGE_TYPES, EXECUTE, FOR, FOREACH, FREE, FROM, GLOBAL, GRANT, GRAPH, HEADER, 
IDENTIFIED, IMPORT, INACTIVE, IN_MEMORY_ANALYTICAL, IN_MEMORY_TRANSACTIONAL, ISOLATION, KAFKA, 
LABELS, LEVEL, LOAD, LOCK, MAIN, MODE, NEXT, NO, NOTHING, NULLIF, PASSWORD, PORT, PRIVILEGES, PULSAR, 
READ, REGISTER, REPLICA, REPLICAS, REPLICATION, REVOKE, ROLE, ROLES, QUOTE, SESSION, SETTING, SETTINGS, 
SNAPSHOT, START, STATS, STATUS, STORAGE, STREAM, STREAMS, SYNC, TERMINATE, TIMEOUT, TO, TOPICS, 
TRANSACTION, TRANSACTIONS, TRANSFORM, TRIGGER, TRIGGERS, UNCOMMITTED, UNLOCK, UPDATE, USE, USER, 
USERS, VERSION, '[', '{', ALL, AND, ANY, AS, ASC, ASCENDING, ASSERT, BFS, BY, CASE, CONSTRAINT, CONTAINS, 
COUNT, CREATE, CYPHERNULL, DELETE, DESC, DESCENDING, DETACH, DISTINCT, ELSE, END, ENDS, EXISTS, EXPLAIN, 
EXTRACT, FALSE, FILTER, IN, INDEX, INFO, IS, KEY, LIMIT, L_SKIP, MATCH, MERGE, NODE, NONE, NOT, ON, OPTIONAL, 
OR, ORDER, PROCEDURE, PROFILE, QUERY, REDUCE, REMOVE, RETURN, SET, SHOW, SINGLE, STARTS, THEN, TRUE, 
UNION, UNIQUE, UNWIND, WHEN, WHERE, WITH, WSHORTEST, ALLSHORTEST, XOR, YIELD, StringLiteral, 
DecimalLiteral, OctalLiteral, HexadecimalLiteral, FloatingLiteral, UnescapedSymbolicName, EscapedSymbolicName}
```

**To Reproduce**
1. Have a CSV file
2. Call `LOAD CSV FROM $file AS row CREATE (:Node { column: row[0] });` with query param `"file": "file:///dir/file.csv"`

**Expected behavior**
I expect `LOAD CSV FROM $file ...` to work with the query parameter so I can provide a file name as a query param. This works ok in Neo4j.

DavIvek:
With this PR, the user is able to use the parameter for the file in the `LOAD CSV` clause. 
The following query is possible: 
`LOAD CSV FROM $file AS row CREATE (:Node { column: row[0] });` 
if the file is a valid path to CSV file.

[master < Task] PR

closes #1386


src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -392,8 +392,10 @@ antlrcpp::Any CypherMainVisitor::visitLoadCsv(MemgraphCypher::LoadCsvContext *ct
 
   auto *load_csv = storage_->Create<LoadCsv>();
   // handle file name
-  if (ctx->csvFile()->literal()->StringLiteral()) {
+  if (ctx->csvFile()->literal()) {
```
as51340, line 395: is it ok to remove `StringLiteral` check?

as51340 (approved):
:muscle:

src/query/frontend/ast/cypher_main_visitor.cpp
DavIvek: Yes, if we don't remove if we will get an error if we use parameter instead of `StringLiteral`because `literal` won't exist so we can't call `StringLiteral` from something that doesn't exist.

DavIvek: But I'll add another `if  `that checks if given `literal` is `StringLiteral` because now if we run the query for example with `integet` literal we won't get descriptive error message as before.

antoniofilipovic (approved):
LGTM

PR #1654, opened 2024-01-19
Add support for query params in return limit

### LINKED ISSUE #1537: Add support for using variables in the LIMIT clause.
Memgraph doesn't support using variables in the LIMIT clause. Ex. important for querying a small random part of the entire graph.

```
match (u:User)
with count(u) as counts
match (u:User)
with u limit s         //  Variables are not allowed in LIMIT
return user

```

DavIvek:
With this PR, the user is able to use the parameter for the limit in return body of the query.
Following query is possible:
`MATCH (n) RETURN n LIMIT $limit;`

[master < Task] PR

closes #1537


src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -1752,7 +1752,11 @@ antlrcpp::Any CypherMainVisitor::visitReturnBody(MemgraphCypher::ReturnBodyConte
     body.skip = static_cast<Expression *>(std::any_cast<Expression *>(ctx->skip()->accept(this)));
   }
   if (ctx->limit()) {
-    body.limit = static_cast<Expression *>(std::any_cast<Expression *>(ctx->limit()->accept(this)));
```
DavIvek, line 1755: I removed this static_cast because it looks to me that we don't need it.

PR #1655, opened 2024-01-19
Fix memory limit transaction id bug

antoniofilipovic:
Transaction id from interpreter which was used as key to which transaction are we tracking memory was not same as a transaction id for transaction started in storage. 

The problem is that the transaction id from the interpreter changes as new queries can be prepared in the same session and we got the reference to the wrong transaction id. 

Logs are still here for demonstration, PR needs additional cleaning

 
[master < Task] PR


as51340 (changes requested):
can you ask also @andrejtonev for review here? I know he was dealing with this interpreter quite a lot... it seems correct but really not sure

PR #1661, opened 2024-01-24
Refactor: Making stream interpreter agnostic

Ignition:
Part of a larger refactor for fixing where the system transaction is controlled from.

### DIFF
--- a/src/query/interpreter.cpp
+++ b/src/query/interpreter.cpp
@@ -106,6 +106,7 @@
 #include "dbms/replication_handler.hpp"
 #include "query/auth_query_handler.hpp"
 #include "query/interpreter_context.hpp"
+#include "query/stream/stream_consumer_factory.hpp"
 #include "replication/state.hpp"
 
 #ifdef MG_ENTERPRISE
@@ -1226,7 +1227,7 @@ Callback::CallbackFunction GetKafkaCreateCallback(StreamQuery *stream_query, Exp
           credentials = get_config_map(stream_query->credentials_, "Credentials"),
           default_server = interpreter_context->config.default_kafka_bootstrap_servers]() mutable {
     std::string bootstrap = bootstrap_servers ? std::move(*bootstrap_servers) : std::move(default_server);
-
+    auto factory = memgraph::query::stream::StreamConsumerFactory{interpreter_context};
     db_acc->streams()->Create<query::stream::KafkaStream>(stream_name,
                                                           {.common_info = std::move(common_stream_info),
                                                            .topics = std::move(topic_names),
@@ -1234,7 +1235,7 @@ Callback::CallbackFunction GetKafkaCreateCallback(StreamQuery *stream_query, Exp
                                                            .bootstrap_servers = std::move(bootstrap),
                                                            .configs = std::move(configs),
                                                            .credentials = std::move(credentials)},
-                                                          std::move(owner), db_acc, interpreter_context);
+                                                          std::move(owner), db_acc, factory);
 
     return std::vector<std::vector<TypedValue>>{};
   };
@@ -1256,10 +1257,11 @@ Callback::CallbackFunction GetPulsarCreateCallback(StreamQuery *stream_query, Ex
           common_stream_info = std::move(common_stream_info), service_url = std::move(service_url), owner = username,
           default_service = interpreter_context->config.default_pulsar_service_url]() mutable {
     std::string url = service_url ? std::move(*service_url) : std::move(default_service);
+    auto factory = memgraph::query::stream::StreamConsumerFactory{interpreter_context};
     db->streams()->Create<query::stream::PulsarStream>(
         stream_name,
         {.common_info = std::move(common_stream_info), .topics = std::move(topic_names), .service_url = std::move(url)},
-        std::move(owner), db, interpreter_context);
+        std::move(owner), db, factory);
 
     return std::vector<std::vector<TypedValue>>{};
   };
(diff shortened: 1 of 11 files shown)


src/query/interpreter.cpp
```diff
@@ -1226,15 +1227,15 @@ Callback::CallbackFunction GetKafkaCreateCallback(StreamQuery *stream_query, Exp
...
+    auto factory = memgraph::query::stream::StreamConsumerFactory{interpreter_context};
     db_acc->streams()->Create<query::stream::KafkaStream>(stream_name,
                                                           {.common_info = std::move(common_stream_info),
                                                            .topics = std::move(topic_names),
                                                            .consumer_group = std::move(consumer_group),
                                                            .bootstrap_servers = std::move(bootstrap),
                                                            .configs = std::move(configs),
                                                            .credentials = std::move(credentials)},
-                                                          std::move(owner), db_acc, interpreter_context);
+                                                          std::move(owner), db_acc, factory);
```
as51340, line 1238: You can also move factory. But it is just a pointer so irrelevant I think

src/query/stream/stream_consumer_factory.cpp
```diff
@@ -0,0 +1,237 @@
...
+std::pair<TypedValue /*query*/, TypedValue /*parameters*/> ExtractTransformationResult(
+    const utils::pmr::map<utils::pmr::string, TypedValue> &values, std::string_view transformation_name,
+    std::string_view stream_name) {
+  if (values.size() != kExpectedTransformationResultSize) {
+    throw StreamsException(
+        "Transformation '{}' in stream '{}' did not yield all fields (query, parameters) as required.",
+        transformation_name, stream_name);
+  }
+
+  auto get_value = [&](std::string_view field_name) mutable -> const TypedValue & {
```
as51340, line 47: You can capture just values instead of everything

src/query/stream/stream_consumer_factory.cpp
```diff
@@ -0,0 +1,237 @@
...
+  MG_ASSERT(params_value.IsNull() || params_value.IsMap());
+  return {query_value, params_value};
+}
+
+template <typename TMessage>
+void CallCustomTransformation(std::string_view transformation_name, const std::vector<TMessage> &messages,
+                              mgp_result &result, Storage::Accessor &storage_accessor, MemoryResource &memory_resource,
+                              std::string_view stream_name) {
+  DbAccessor db_accessor{&storage_accessor};
+  {
```
as51340, line 68: No need to add scope

src/query/stream/stream_consumer_factory.cpp
```diff
@@ -0,0 +1,237 @@
...
+    auto maybe_transformation =
```
as51340, line 69: if trans is const, this can also be const right?

src/query/stream/stream_consumer_factory.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+    spdlog::trace("Start transaction in stream '{}'", stream_name);
+    memgraph::utils::OnScopeExit cleanup{[&interpreter, &result]() {
+      result.rows.clear();
+      interpreter->Abort();
+    }};
+
+    uint32_t i = 0;
+    auto const total_retries = interpreter_context->config.stream_transaction_conflict_retries;
+    const auto retry_interval = interpreter_context->config.stream_transaction_retry_interval;
```
as51340, line 135: I know we don't care about east or west const but it is weird to read different things in 2 consecutive rows

src/query/stream/stream_consumer_factory.cpp
```diff
@@ -0,0 +1,237 @@
...
+        }
+
+        spdlog::trace("Commit transaction in stream '{}'", stream_name);
+        interpreter->CommitTransaction();
+        result.rows.clear();
+        break;
+      } catch (const memgraph::query::TransactionSerializationException &e) {
+        interpreter->Abort();
+
+        if (i == total_retries) {
```
as51340, line 166: You can use here ResettableCounter you added last week right?

src/query/stream/stream_consumer_factory.hpp
```diff
@@ -0,0 +1,40 @@
...
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+#pragma once
+
+#include "query/interpreter_context.hpp"
+#include "query/stream/istream_consumer_factory.hpp"
+
+namespace memgraph::query::stream {
+struct StreamConsumerFactory : IStreamConsumerFactory {
```
as51340, line 17: Why introduce virtual interface if we have just one impl? I don't think this will change soon in the future?

src/query/stream/stream_consumer_factory.cpp
```diff
@@ -0,0 +1,237 @@
...
+    const auto retry_interval = interpreter_context->config.stream_transaction_retry_interval;
+    while (true) {
+      try {
+        interpreter->BeginTransaction();
+        for (auto &row : result.rows) {
+          spdlog::trace("Processing row in stream '{}'", stream_name);
+          auto [query_value, params_value] = ExtractTransformationResult(row.values, transformation_name, stream_name);
+          memgraph::storage::PropertyValue params_prop{params_value};
+
+          std::string query{query_value.ValueString()};
```
as51340, line 144:
```suggestion
          const std::&string query{query_value.ValueString()};
```

I think no need to copy it here

as51340 (approved):
Refactoring looks good, have few cpp comments but :muscle:

src/query/stream/stream_consumer_factory.hpp
Ignition: Its the whole point of this, to decouple stream evaluation from query. The plan is to untangle, dbms/database/query/streams/triggers so that system transaction can exist independently.

src/query/stream/stream_consumer_factory.cpp
Ignition: Refactored by moving from elsewhere. So is at it was. But also it makes no difference. Values, transformation_name, and stream_name are all used inside the lambda. Also this is local use only, lambda would be completely inlined.

src/query/stream/stream_consumer_factory.cpp
Ignition: For this usage, it was moved code from elsewhere. I do not ATM want to use ResettableCounter just in case I have the retry count wrong and change behaviour to be off by one.

For now I'll leave as is.

src/query/stream/stream_consumer_factory.cpp
Ignition: Going from pmr::string to string

Ignition:
Decided this approach was unnecessary for now.

PR #1663, opened 2024-01-25
Simplify auth::Auth

andrejtonev:
Moved various flags under a single config. 
Moved regex logic under auth::Auth

[master < Task] PR

### DIFF
--- a/src/auth/auth.hpp
+++ b/src/auth/auth.hpp
@@ -1,4 +1,4 @@
-// Copyright 2023 Memgraph Ltd.
+// Copyright 2024 Memgraph Ltd.
 //
 // Licensed as a Memgraph Enterprise file under the Memgraph Enterprise
 // License (the "License"); by using this file, you agree to be bound by the terms of the License, and you may not use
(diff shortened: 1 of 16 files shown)


src/auth/auth.hpp
```diff
@@ -31,7 +32,36 @@ static const constexpr char *const kAllDatabases = "*";
...
+          custom_password_regex{password_regex_str != glue::kDefaultPasswordRegex} {}
+
+    std::string name_regex_str{glue::kDefaultUserRoleRegex};
+    std::string password_regex_str{glue::kDefaultPasswordRegex};
+    bool password_permit_null{true};
+
+   private:
+    friend class Auth;
+    bool custom_name_regex{false};
+    bool custom_password_regex{false};
```
Ignition, line 51: Can we prebuild the regex here, building regex is costly compared to using a regex.

Ignition (changes requested):

src/auth/auth.hpp
andrejtonev: You're right, but gonna keep the string as well, since there are some logs and error messages that rely on it.

Ignition (approved):

PR #1667, opened 2024-01-25
Add step to refresh jepsen cluster before test

MarkoBarisic:
This PR will reduce flakiness of Jepsen tests which is mainly due to issues with the local Jepsen cluster.

Every time a Jepsen test is run, the cluster will be refreshed, ie. brought down and back up again.

[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR

### DIFF
--- a/tests/jepsen/run.sh
+++ b/tests/jepsen/run.sh
@@ -187,7 +187,7 @@ PROCESS_RESULTS() {
 CLUSTER_UP() {
   PRINT_CONTEXT
   "$script_dir/jepsen/docker/bin/up" --daemon
-  sleep 1
+  sleep 10
   # Ensure all SSH connections between Jepsen containers work
   for node in $(docker ps --filter name=jepsen* --filter status=running --format "{{.Names}}"); do
       if [ "$node" == "jepsen-control" ]; then
--- a/tests/jepsen/run.sh
+++ b/tests/jepsen/run.sh
@@ -24,7 +24,7 @@ PRINT_CONTEXT() {
 
 HELP_EXIT() {
     echo ""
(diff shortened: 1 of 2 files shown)


andrejtonev (approved):
Makes sense!

tests/jepsen/run.sh
```diff
@@ -184,6 +184,37 @@ PROCESS_RESULTS() {
...
+  PRINT_CONTEXT
+  "$script_dir/jepsen/docker/bin/up" --daemon
+  sleep 1
+  # Ensure all SSH connections between Jepsen containers work
+  for node in $(docker ps --filter name=jepsen* --filter status=running --format "{{.Names}}"); do
+      if [ "$node" == "jepsen-control" ]; then
+          continue
+      fi
+      node_hostname="${node##jepsen-}"
+      docker exec jepsen-control bash -c "ssh -oStrictHostKeyChecking=no -t $node_hostname exit"
```
Ignition, line 197: Is 1 sec sleep deterministic enough to be confident that it is up?

I don't want the situation where we have a healthy cluster that we just brought down just to then bring it up, but didn't wait long enough for it to respond to the ssh check.

MarkoBarisic: Your consern makes sense, I should push it a bit longer, what do you think 5s or 10s?

andrejtonev: Could we make it time independent?
Perhaps a loop where we retry for a couple of times and fail after a certain number of tries?

MarkoBarisic: I understand, but I would like to merge ASAP because there is operational cost to making this PR wait for merge. This loop checking is a good idea, but it may be an overkill at the moment. Let's merge and if problems start popping up, which I don't think will, I will make the change. What do you think?

PR #1672, opened 2024-01-29
HA: Support restart of instances

as51340:
This PR adds functionalities which handle restart of instances. When using `--coordinator-server-port` flag, replica will always be restarted as replica. The main will get back as replica to cluster if in the time the instance was down, failover was done. If no failover was done, the old main will again become main. When performing failover, all other instances except the chosen new main are sent to the new main so it could register replication servers towards instances, even if some instance is down.

[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR


src/coordination/coordinator_client.cpp
```diff
@@ -59,20 +59,16 @@ void CoordinatorClient::StartFrequentCheck() {
...
-
 void CoordinatorClient::PauseFrequentCheck() { instance_checker_.Pause(); }
 void CoordinatorClient::ResumeFrequentCheck() { instance_checker_.Resume(); }
 
-auto CoordinatorClient::SetSuccCallback(HealthCheckCallback succ_cb) -> void { succ_cb_ = std::move(succ_cb); }
-auto CoordinatorClient::SetFailCallback(HealthCheckCallback fail_cb) -> void { fail_cb_ = std::move(fail_cb); }
-
-auto CoordinatorClient::ReplicationClientInfo() const -> const CoordinatorClientConfig::ReplicationClientInfo & {
-  return config_.replication_client_info;
+auto CoordinatorClient::SetCallbacks(HealthCheckCallback succ_cb, HealthCheckCallback fail_cb) -> void {
```
andrejtonev, line 65: When defining a void function just define `void f()` instead of `auto f() -> void`

src/coordination/coordinator_client.cpp
```diff
@@ -90,17 +86,18 @@ auto CoordinatorClient::SendPromoteReplicaToMainRpc(
   return false;
 }
 
-auto CoordinatorClient::SendSetToReplicaRpc(CoordinatorClient::ReplClientInfo replication_client_info) const -> bool {
+auto CoordinatorClient::DemoteToReplica() const -> bool {
+  const auto instance_name = config_.instance_name;
```
andrejtonev, line 90: const auto&

src/coordination/coordinator_data.cpp
```diff
@@ -35,33 +35,55 @@ CoordinatorData::CoordinatorData() {
...
     auto lock = std::lock_guard{coord_data->coord_data_lock_};
     spdlog::trace("Instance {} performing replica failure callback", instance_name);
     auto &instance = find_instance(coord_data, instance_name);
-    MG_ASSERT(instance.IsReplica(), "Instance {} is not a replica!", instance_name);
-    instance.UpdateInstanceStatus();
+    instance.OnFailPing();
   };
 
-  main_succ_cb_ = [find_instance](CoordinatorData *coord_data, std::string_view instance_name) -> void {
+  main_succ_cb_ = [this, find_instance](CoordinatorData *coord_data, std::string_view instance_name) -> void {
```
andrejtonev, line 48: Why capture `this` and pass `CoordinatorData*` as function argument?

src/coordination/coordinator_data.cpp
```diff
@@ -35,33 +35,55 @@ CoordinatorData::CoordinatorData() {
...
-    MG_ASSERT(instance.IsMain(), "Instance {} is not a main!", instance_name);
-    instance.UpdateLastResponseTime();
+
+    if (instance.IsAlive()) {
+      instance.OnSuccessPing();
+    } else {
+      auto const new_role = coord_data->ClusterHasAliveMain() ? replication_coordination_glue::ReplicationRole::REPLICA
+                                                              : replication_coordination_glue::ReplicationRole::MAIN;
+      if (new_role == replication_coordination_glue::ReplicationRole::REPLICA) {
+        thread_pool_.AddTask([&instance, coord_data, instance_name]() {
```
andrejtonev, line 60: Why do this via the tread pool instead of in the same thread?

src/coordination/coordinator_data.cpp
```diff
@@ -35,33 +35,55 @@ CoordinatorData::CoordinatorData() {
...
-  main_succ_cb_ = [find_instance](CoordinatorData *coord_data, std::string_view instance_name) -> void {
+  main_succ_cb_ = [this, find_instance](CoordinatorData *coord_data, std::string_view instance_name) -> void {
     auto lock = std::lock_guard{coord_data->coord_data_lock_};
     spdlog::trace("Instance {} performing main successful callback", instance_name);
+
     auto &instance = find_instance(coord_data, instance_name);
-    MG_ASSERT(instance.IsMain(), "Instance {} is not a main!", instance_name);
-    instance.UpdateLastResponseTime();
+
+    if (instance.IsAlive()) {
```
andrejtonev, line 54: I'm not really following.
We had a successful frequent check. But we are still checking if the instance is alive?
If not alive, we try to set it to replica?

andrejtonev (commented):
I did a quick overview. 
I think we should sync before going further.

src/coordination/coordinator_data.cpp
```diff
@@ -35,33 +35,55 @@ CoordinatorData::CoordinatorData() {
...
+      instance.OnSuccessPing();
+    } else {
+      auto const new_role = coord_data->ClusterHasAliveMain() ? replication_coordination_glue::ReplicationRole::REPLICA
```
andrejtonev, line 57: This works only if this was the last MAIN alive, otherwise, figure out a new MAIN.

src/coordination/coordinator_data.cpp
andrejtonev: You can release the stream and then do the callback.

src/coordination/coordinator_data.cpp
```diff
@@ -80,44 +102,36 @@ CoordinatorData::CoordinatorData() {
   };
 }
 
-auto CoordinatorData::DoFailover() -> DoFailoverStatus {
-  using ReplicationClientInfo = CoordinatorClientConfig::ReplicationClientInfo;
+auto CoordinatorData::ClusterHasAliveMain() const -> bool {
```
andrejtonev, line 105: For private api that are not thread-safe use a _ suffix naming convention.

src/coordination/coordinator_data.cpp
```diff
@@ -35,33 +35,55 @@ CoordinatorData::CoordinatorData() {
...
     spdlog::trace("Instance {} performing main failure callback", instance_name);
     auto &instance = find_instance(coord_data, instance_name);
-    MG_ASSERT(instance.IsMain(), "Instance {} is not a main!", instance_name);
-    if (bool main_alive = instance.UpdateInstanceStatus(); !main_alive) {
-      spdlog::info("Main instance {} is not alive, starting automatic failover", instance_name);
+    instance.OnFailPing();
+
+    if (!ClusterHasAliveMain()) {
+      spdlog::info("Cluster without main instance, starting automatic failover");
       switch (auto failover_status = DoFailover(); failover_status) {
```
andrejtonev, line 87: `MAIN_ALIVE` shouldn't be possible.

andrejtonev (commented):
Another round.
Thanks for the sync

src/coordination/coordinator_data.cpp
```diff
@@ -35,105 +35,94 @@ CoordinatorData::CoordinatorData() {
...
 
   main_succ_cb_ = [find_instance](CoordinatorData *coord_data, std::string_view instance_name) -> void {
     auto lock = std::lock_guard{coord_data->coord_data_lock_};
     spdlog::trace("Instance {} performing main successful callback", instance_name);
+
     auto &instance = find_instance(coord_data, instance_name);
-    MG_ASSERT(instance.IsMain(), "Instance {} is not a main!", instance_name);
-    instance.UpdateLastResponseTime();
+
+    if (instance.IsAlive() || !coord_data->ClusterHasAliveMain_()) {
```
andrejtonev, line 53: Why `|| !coord_data->ClusterHasAliveMain_()`?

andrejtonev: We need to know that the MAIN that's alive is not us.

src/coordination/coordinator_data.cpp
```diff
@@ -35,105 +35,94 @@ CoordinatorData::CoordinatorData() {
...
+      instance.OnSuccessPing();
+      return;
+    }
+
+    bool const demoted = instance.DemoteToReplica(coord_data->replica_succ_cb_, coord_data->replica_fail_cb_);
+    if (demoted) {
+      instance.OnSuccessPing();
+      spdlog::info("Instance {} demoted to replica", instance_name);
+    } else {
+      spdlog::error("Instance {} failed to become replica", instance_name);
```
andrejtonev, line 63: Do we need to do `OnFailedPing()`?

src/coordination/coordinator_data.cpp
```diff
@@ -34,85 +34,59 @@ CoordinatorData::CoordinatorData() {
   replica_succ_cb_ = [find_instance](CoordinatorData *coord_data, std::string_view instance_name) -> void {
     auto lock = std::lock_guard{coord_data->coord_data_lock_};
     spdlog::trace("Instance {} performing replica successful callback", instance_name);
-    auto &instance = find_instance(coord_data, instance_name);
-    instance.OnSuccessPing();
+    find_instance(coord_data, instance_name).OnSuccessPing();
```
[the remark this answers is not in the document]
as51340: No

src/coordination/coordinator_data.cpp
as51340: if the instance is alive then we are just updating last response time.
If not, we are dead and starting to become alive. If there is no alive main, no failover was performed so old main comes back again as main. If not, old main should become replica.

src/coordination/coordinator_data.cpp
as51340: No need, from our perspective instance stays dead

andrejtonev (approved):

as51340:
@kgolubic release note: Added support for restart of replication instances. Main can restart as main and as replica depending on the cluster state in the moment of restart. Replica always stays replica

Issue #1685, opened 2024-01-31
Memory consumption extremely high on in-memory transactional mode after import without freeing


Josipmrden:
**Memgraph version**
v2.14

**Environment**
Docker

**Describe the bug**
After ingesting 3 million nodes with one property, memory does not seem to free and memory usage shows 3GB consumed. It is never freed. If you restart the same container for the data to be restored from the snapshot, the memory usage shows around 700MB which is the estimated amount. Something seems not to free in the deltas or memory is not given back to the process, that's an assumption

**To Reproduce**
1. Start Memgraph
2. Run `FOREACH (i in range(1, 3000000) | CREATE (:Label1:Label2:Label3:Label4:Label5:Label6));`
3. Observe memory usage
4. Restart Memgraph
5. Observe memory usage
6. Stop Memgraph
7. Start Memgraph with fresh database
8. run `STORAGE MODE IN_MEMORY_ANALYTICAL;`
9. run `FOREACH (i in range(1, 3000000) | CREATE (:Label1:Label2:Label3:Label4:Label5:Label6));`
10. Observe much less memory being used in the analytical mode than the transactional mode

**Expected behavior**
Memory should be freed so users don't run into Out Of Memory Exceptions.

**Additional context**

**Verification Environment**
Once we fix it, what do you need to verify the fix?
Do you need:
* Plain memgraph package -> for which Linux?
* Plain memgraph Docker image?
* Which architecture do you use ARM | x86?
* Full Memgraph platform?

### NOTE: A bunch of work here has been done, but still TODOs:
- [x] Ensure deltas are allocated to separate pages than those of storage

Ignition:
5. Using a common allocation arena results in interleaved allocations.

This means vertices and delta allocation happen in that order and use a common page.

`| V D V D V D | V D V D . . |`

When Deltas are freed their space in the page can be reused.

`| V . V . V . | V . V . . . |`

But the number of pages does not go down because the vertices can't relocate. Hence the resident memory does not go down after the transaction has finished.

TODO: use separate allocation arenas, which will force the two sub systems to use different page sets.

`| V V V V V . |` + `| D D D D D . |`
will become 
`| V V V V V . |` + ` `

Ignition:
More delta reduction 80B -> 56B #1747

Ignition:
Status update, with all the changes done so far the effect is less severe. I have just retested:
```
IN_MEMORY_TRANSACTIONAL ~750MiB
[...]

PR #1688, opened 2024-02-02
Improve memory handling of Deltas

Ignition:
- Reduce delta from 104B to 80B
- Hold and pass them around as in a deque
- Detect and deleted deltas within commit if safe to do so

relates to #1685

## profile without jemalloc

| Before        | After           | 
| ------------- |-------------| 
| Before      | After |

### DIFF
--- a/src/storage/v2/delta.hpp
+++ b/src/storage/v2/delta.hpp
@@ -1,4 +1,4 @@
-// Copyright 2023 Memgraph Ltd.
+// Copyright 2024 Memgraph Ltd.
 //
 // Use of this software is governed by the Business Source License
 // included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
@@ -57,9 +57,11 @@ class PreviousPtr {
     explicit Pointer(Edge *edge) : type(Type::EDGE), edge(edge) {}
 
     Type type{Type::NULLPTR};
-    Delta *delta{nullptr};
-    Vertex *vertex{nullptr};
-    Edge *edge{nullptr};
+    union {
+      Delta *delta = nullptr;
+      Vertex *vertex;
+      Edge *edge;
+    };
   };
 
   PreviousPtr() : storage_(0) {}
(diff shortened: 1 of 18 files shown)


src/storage/v2/delta.hpp
```diff
@@ -228,32 +222,39 @@ struct Delta {
...
   union {
-    std::optional<std::string> old_disk_key;
-    LabelId label;
+    Action action;
+    struct {
+      Action action = Action::DELETE_DESERIALIZED_OBJECT;
+      std::optional<std::string> value;
+    } old_disk_key;
+    struct {
+      Action action;
```
Josipmrden, line 248: why do you add action everywhere if it's in the top level as well, is it because of padding?

src/storage/v2/id_types.hpp
```diff
@@ -50,10 +50,10 @@ namespace memgraph::storage {
   inline bool operator<=(const name &first, const name &second) { return first.AsUint() <= second.AsUint(); } \
   inline bool operator>=(const name &first, const name &second) { return first.AsUint() >= second.AsUint(); }
 
-STORAGE_DEFINE_ID_TYPE(Gid);
-STORAGE_DEFINE_ID_TYPE(LabelId);
-STORAGE_DEFINE_ID_TYPE(PropertyId);
-STORAGE_DEFINE_ID_TYPE(EdgeTypeId);
+STORAGE_DEFINE_ID_TYPE(Gid, uint64_t, int64_t, utils::ParseStringToUint64);
```
Josipmrden, line 53: can you check maybe name_id_mapper functionality ? do we have any risk of running that because every time someone makes a typo the name id mapper is still going to autoincrement the value?
also, do we want separate counters for every of them (separate for label, property and edge type) ?

Josipmrden (commented):
Looks good, just a few comments

src/storage/v2/id_types.hpp
Ignition: Ok sure `NameToId` is not const, so even in non-write based queries we can store and bump the id. 
That should be changed, a const version is needed, which would return std::nullopt.

We should keep separate counters.

Ignition: I would fix the NameToId issue in another PR, if and when it become a priority/likely issue.

It is possibly to saturate the mappings with typos, but as it stands right now maybe not a priority.

Josipmrden (approved):

PR #1693, opened 2024-02-06
Distributed coordinators

as51340:
Only leader performs callbacks => distributed failover logic.
Refactoring of coordinator_instance, replication_instance and raft setup


src/coordination/coordinator_instance.cpp
```diff
@@ -18,80 +18,271 @@
...
+    }
+  };
+}
+
+auto CoordinatorInstance::ClusterHasAliveMain_() const -> bool {
+  auto const alive_main = [](ReplicationInstance const &instance) { return instance.IsMain() && instance.IsAlive(); };
+  return std::ranges::any_of(repl_instances_, alive_main);
+}
+
+auto CoordinatorInstance::TryFailover() -> void {
```
andrejtonev, line 99: Seems to have very similar logic to `SetReplicationInstanceToMain` can we unify it in some way?

andrejtonev (commented):
Looks good.
Need to get more into the log stuff and before I approve.

src/coordination/coordinator_instance.cpp
```diff
@@ -18,80 +18,271 @@
...
+    auto instance = std::ranges::find_if(
+        coord_instance->repl_instances_,
+        [instance_name](ReplicationInstance const &instance) { return instance.InstanceName() == instance_name; });
+
+    MG_ASSERT(instance != coord_instance->repl_instances_.end(), "Instance {} not found during callback!",
+              instance_name);
+    return *instance;
+  };
+
+  replica_succ_cb_ = [find_instance](CoordinatorInstance *coord_instance, std::string_view instance_name) -> void {
```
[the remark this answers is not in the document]
as51340: Yes

src/coordination/coordinator_instance.cpp
```diff
@@ -18,80 +18,271 @@
...
+    spdlog::trace("Instance {} performing main successful callback", instance_name);
+
+    auto &instance = find_instance(coord_instance, instance_name);
+
+    if (instance.IsAlive()) {
+      instance.OnSuccessPing();
+      return;
+    }
+
+    bool const is_latest_main = !coord_instance->ClusterHasAliveMain_();
```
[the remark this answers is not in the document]
as51340: I messed up rebase, this was already done... Now should be good

src/coordination/coordinator_instance.cpp
```diff
@@ -18,80 +18,271 @@
...
+      spdlog::error("Instance {} failed to become replica", instance_name);
+    }
+  };
+
+  main_fail_cb_ = [find_instance](CoordinatorInstance *coord_instance, std::string_view instance_name) -> void {
+    auto lock = std::lock_guard{coord_instance->coord_instance_lock_};
+    spdlog::trace("Instance {} performing main failure callback", instance_name);
+    find_instance(coord_instance, instance_name).OnFailPing();
+
+    if (!coord_instance->ClusterHasAliveMain_()) {
```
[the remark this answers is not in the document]
as51340: exactly

src/coordination/coordinator_instance.cpp
as51340: Yes, nice catch I will see what can be done

src/coordination/coordinator_instance.cpp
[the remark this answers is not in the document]
as51340: rebase mistake..

src/coordination/coordinator_log_store.cpp
```diff
@@ -13,229 +13,164 @@
...
+  uint64_t const end_index = index + cnt;
+  for (uint64_t i = index; i < end_index; ++i) {
     ptr<log_entry> le = nullptr;
     {
-      std::lock_guard<std::mutex> l(logs_lock_);
-      le = logs_[ii];
+      auto lock = std::lock_guard{logs_lock_};
+      le = logs_[i];
     }
     assert(le.get());
```
antoniofilipovic, line 135: MG_ASSERT here?

src/coordination/coordinator_log_store.cpp
```diff
@@ -13,229 +13,164 @@
...
 
-  ptr<buffer> buf_out = buffer::alloc(sizeof(int32) + cnt * sizeof(int32) + size_total);
+  auto buf_out = buffer::alloc(sizeof(int32) + cnt * sizeof(int32) + size_total);
   buf_out->pos(0);
   buf_out->put((int32)cnt);
 
   for (auto &entry : logs) {
-    ptr<buffer> &bb = entry;
-    buf_out->put((int32)bb->size());
+    auto &bb = entry;  // TODO: (andi) This smells like not needed
```
antoniofilipovic, line 146: Can we remove this?

antoniofilipovic (commented):
Few comments, but looks good to me

src/coordination/coordinator_instance.cpp
```diff
@@ -18,81 +18,305 @@
...
+
+  if (std::ranges::any_of(repl_instances_, socket_address_matches)) {
+    return RegisterInstanceCoordinatorStatus::ENDPOINT_EXISTS;
+  }
+
+  if (!raft_state_.RequestLeadership()) {
+    return RegisterInstanceCoordinatorStatus::NOT_LEADER;
+  }
+
+  auto const res = raft_state_.AppendRegisterReplicationInstance(instance_name);
```
antoniofilipovic, line 285: What could be other reason for this?

src/coordination/coordinator_instance.cpp
```diff
@@ -18,81 +18,305 @@
...
+  try {
+    repl_instances_.emplace_back(this, std::move(config), replica_succ_cb_, replica_fail_cb_);
+  } catch (CoordinatorRegisterInstanceException const &) {
+    return RegisterInstanceCoordinatorStatus::RPC_FAILED;
+  }
+
+  if (res->get_result_code() != nuraft::cmd_result_code::OK) {
+    spdlog::error("Failed to register instance {} with error code {}", instance_name, res->get_result_code());
+    return RegisterInstanceCoordinatorStatus::RAFT_COULD_NOT_APPEND;
+  }
```
antoniofilipovic, line 305: Maybe we should wrap their result into our result, so that we don't deal with nuraft here, only in raft state

tests/e2e/high_availability_experimental/coord_cluster_registration.py
```diff
@@ -0,0 +1,284 @@
...
+        coordinator3_cursor, "REGISTER INSTANCE instance_1 ON '127.0.0.1:10011' WITH '127.0.0.1:10001'"
+    )
+    execute_and_fetch_all(
+        coordinator3_cursor, "REGISTER INSTANCE instance_2 ON '127.0.0.1:10012' WITH '127.0.0.1:10002'"
+    )
+    execute_and_fetch_all(
+        coordinator3_cursor, "REGISTER INSTANCE instance_3 ON '127.0.0.1:10013' WITH '127.0.0.1:10003'"
+    )
+    execute_and_fetch_all(coordinator3_cursor, "SET INSTANCE instance_3 TO MAIN")
+    assert add_coordinator(coordinator3_cursor, "ADD COORDINATOR 1 ON '127.0.0.1:10111'")
```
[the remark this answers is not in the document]
as51340: 1 is raft server id which is sent to raft api. Has to be int i think

src/coordination/raft_state.cpp
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  auto maybe_stop = utils::ResettableCounter<20>();
+  while (!raft_server_->is_initialized() && !maybe_stop()) {
+    std::this_thread::sleep_for(std::chrono::milliseconds(250));
+  }
+
+  if (!raft_server_->is_initialized()) {
+    throw RaftServerStartException("Failed to initialize raft server on {}:{}", raft_address_, raft_port_);
+  }
```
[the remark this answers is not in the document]
as51340: but then it shouldn't be is_initialized check inside while loop condition

src/coordination/coordinator_instance.cpp
as51340: what

src/coordination/coordinator_instance.cpp
as51340: how do you mean wrap? different methods are being called so not sure what do you mean

src/coordination/coordinator_log_store.cpp
as51340: yes

src/coordination/coordinator_log_store.cpp
as51340: I think so

src/coordination/coordinator_instance.cpp
```diff
@@ -27,66 +27,97 @@ using nuraft::ptr;
...
+    auto &repl_instance = find_repl_instance(self, repl_instance_name);
+    repl_instance.OnFailPing();
+    // We need to restart main uuid from instance since it was "down" at least a second
+    // There is slight delay, if we choose to use isAlive, instance can be down and back up in less than
+    // our isAlive time difference, which would lead to instance setting UUID to nullopt and stopping accepting any
+    // incoming RPCs from valid main
+    // TODO(antoniofilipovic) this needs here more complex logic
+    // We need to get id of main replica is listening to on successful ping
+    // and swap it to correct uuid if it failed
+    repl_instance.ResetMainUUID();
```
andrejtonev, line 71: It seems to me that REPLICA's connection has nothing to do with MAIN's UUID.
UUID is changed during Failover, so that should invalidate the REPLICA's state, not this.

src/coordination/coordinator_instance.cpp
```diff
@@ -27,66 +27,97 @@ using nuraft::ptr;
...
 
-  main_succ_cb_ = [find_instance](CoordinatorInstance *coord_instance, std::string_view instance_name) -> void {
-    auto lock = std::lock_guard{coord_instance->coord_instance_lock_};
-    spdlog::trace("Instance {} performing main successful callback", instance_name);
+  main_succ_cb_ = [find_repl_instance](CoordinatorInstance *self, std::string_view repl_instance_name) -> void {
+    auto lock = std::lock_guard{self->coord_instance_lock_};
+    spdlog::trace("Instance {} performing main successful callback", repl_instance_name);
 
-    auto &instance = find_instance(coord_instance, instance_name);
+    auto &repl_instance = find_repl_instance(self, repl_instance_name);
```
andrejtonev, line 78: Is it replica instance or main instance? Can we change the name?

src/coordination/coordinator_log_store.cpp
```diff
@@ -13,229 +13,164 @@
...
+ptr<log_entry> CoordinatorLogStore::last_entry() const {
+  auto lock = std::lock_guard{logs_lock_};
 
-  if (disk_emul_delay) {
-    uint64_t cur_time = timer_helper::get_timeofday_us();
-    disk_emul_logs_being_written_[cur_time + disk_emul_delay * 1000] = idx;
-    // disk_emul_ea_.invoke();
-  }
+  uint64_t const last_idx = start_idx_ + logs_.size() - 1;
+  auto const last_src = FindOrDefault_(last_idx - 1);
```
andrejtonev, line 58: `uint64_t const last_idx = start_idx_ + logs_.size() - 1;` has a `-1` and `FindOrDefault_(last_idx - 1);` has is also.
Is this correct?

src/coordination/coordinator_log_store.cpp
```diff
@@ -13,229 +13,164 @@
...
+        src = entry->second;
+      } else {
+        throw RaftCouldNotFindEntryException("Could not find entry at index {}", i);
       }
-      src = entry->second;
     }
-    ret->push_back(MakeClone(src));
-    accum_size += src->get_buf().size();
-    if (batch_size_hint_in_bytes && accum_size >= (ulong)batch_size_hint_in_bytes) break;
+    (*ret)[curr_index] = MakeClone(src);
```
andrejtonev, line 102: Is it safe to append element without holding the logs_lock_?

src/coordination/coordinator_log_store.cpp
```diff
@@ -13,229 +13,164 @@
...
   size_t size_total = 0;
-  for (ulong ii = index; ii < index + cnt; ++ii) {
+  uint64_t const end_index = index + cnt;
+  for (uint64_t i = index; i < end_index; ++i) {
     ptr<log_entry> le = nullptr;
     {
-      std::lock_guard<std::mutex> l(logs_lock_);
-      le = logs_[ii];
+      auto lock = std::lock_guard{logs_lock_};
+      le = logs_[i];
```
andrejtonev, line 134: Similar to above. Not sure how the ptr<> works, is it safe to handle it without holding the lock?

src/coordination/coordinator_log_store.cpp
```diff
@@ -13,229 +13,164 @@
...
-  ptr<buffer> buf_out = buffer::alloc(sizeof(int32) + cnt * sizeof(int32) + size_total);
+  auto buf_out = buffer::alloc(sizeof(int32) + cnt * sizeof(int32) + size_total);
   buf_out->pos(0);
   buf_out->put((int32)cnt);
 
   for (auto &entry : logs) {
-    ptr<buffer> &bb = entry;
-    buf_out->put((int32)bb->size());
-    buf_out->put(*bb);
+    buf_out->put(static_cast<int32>(entry->size()));
```
andrejtonev, line 147: Do we have to do static cast from unsigned to signed?

andrejtonev (approved):
Added a few comments.
Plus a general comment.
Seems like the state of a single instance is getting smeared over various places/callbacks. We have a general problem in our codebase with state machines that aren't defined in a single place, but use a various flags/enums to define a single state, which in turn gets managed in various places.
I think we should think about moving a way from this format towards something more cohesive.

src/coordination/coordinator_log_store.cpp
as51340: To avoid warning

src/coordination/coordinator_log_store.cpp
as51340: Should be, this part is used from the example. Whole log_store is, I just stripped parts we didn't need.

src/coordination/coordinator_log_store.cpp
as51340: we are appending to local var so yes

src/coordination/coordinator_log_store.cpp
as51340: yes

as51340: because of start_idx_

src/coordination/coordinator_instance.cpp
as51340: it's replication instance that's why repl

src/coordination/coordinator_instance.cpp
as51340: agree but there is some edge case which fico wanted to cover with this code. This code exists from before I just renamed function to more explicit

antoniofilipovic (approved):

src/coordination/coordinator_instance.cpp
antoniofilipovic: I work on that here: https://github.com/memgraph/memgraph/pull/1711

as51340:
@kgolubic release note: Added support for registering multiple coordinators. Only leader is sending health checks to replication instances.

PR #1699, opened 2024-02-07
Better auth user/role handling

andrejtonev:
Stop auth module from creating users
Check if there are users OR if an auth module is used
Role supports database access
Authenticate() returns user or role
If module is used, no data is locally saved
AuthChecker generates QueryUser (user or role)
QueryUser actually authorizes

[master < Task] PR


src/auth/auth.cpp
```diff
@@ -535,32 +497,66 @@ std::vector<auth::User> Auth::AllUsersForRole(const std::string &rolename_orig)
 }
 
 #ifdef MG_ENTERPRISE
-bool Auth::GrantDatabaseToUser(const std::string &db, const std::string &name, system::Transaction *system_tx) {
+bool Auth::GrantDatabase(const std::string &db, const std::string &name, system::Transaction *system_tx) {
```
Ignition, line 500: When auth module used we need to be more explicit about the failure reason.
Normally its just false -> no user or role with that name
With auth module false must be -> no role with that name

Change return type to enum + better diagnostic in interpreter

src/auth/auth.cpp
```diff
@@ -569,18 +565,40 @@ void Auth::DeleteDatabase(const std::string &db, system::Transaction *system_tx)
       SaveUser(*user, system_tx);
     }
   }
+  for (auto it = storage_.begin(kRolePrefix); it != storage_.end(kRolePrefix); ++it) {
+    auto rolename = it->first.substr(kRolePrefix.size());
+    if (auto role = GetRole(rolename)) {
```
Ignition, line 570: More efficient role deserialize using `it->second`

src/auth/auth.cpp
```diff
@@ -569,18 +565,40 @@ void Auth::DeleteDatabase(const std::string &db, system::Transaction *system_tx)
...
+  }
 }
 
 bool Auth::SetMainDatabase(std::string_view db, const std::string &name, system::Transaction *system_tx) {
   if (auto user = GetUser(name)) {
-    if (!user->db_access().SetDefault(db)) {
-      throw AuthException("Couldn't set default database '{}' for user '{}'!", db, name);
-    }
-    SaveUser(*user, system_tx);
+    SetMainDatabase(db, *user, system_tx);
```
Ignition, line 579: TODO rename to Default (not Main)

src/auth/auth.hpp
```diff
@@ -266,8 +292,17 @@ class Auth final {
    * @throw AuthException if unable to find or update the user
    */
   bool SetMainDatabase(std::string_view db, const std::string &name, system::Transaction *system_tx = nullptr);
+  void SetMainDatabase(std::string_view db, User &user, system::Transaction *system_tx = nullptr);
+  void SetMainDatabase(std::string_view db, Role &role, system::Transaction *system_tx = nullptr);
 #endif
 
+  Epoch UpdateEpoch() { return ++epoch_; }
```
Ignition, line 299: Every durability/Save in auth should be calling this

src/auth/models.cpp
```diff
@@ -471,12 +474,21 @@ Role Role::Deserialize(const nlohmann::json &data) {
...
 #ifdef MG_ENTERPRISE
   if (memgraph::license::global_license_checker.IsEnterpriseValidFast()) {
+    Databases db_access;
+    if (data[kDatabases].is_structured()) {
+      db_access = Databases::Deserialize(data[kDatabases]);
+    } else {
+      // Back-compatibility
+      spdlog::warn("Role without specified database access. Given access to the default database.");
+      db_access.Add(dbms::kDefaultDB);
+      db_access.SetDefault(dbms::kDefaultDB);
```
Ignition, line 484: Maybe do this as an auth migration step, rather than perpetual backwards compatibility

src/auth/models.hpp
```diff
@@ -388,6 +401,18 @@ class User final {
...
+  bool DeniesDB(std::string_view db_name) const {
+    bool denies = database_access_.Denies(db_name);
+    if (role_) denies |= role_->DeniesDB(db_name);
+    return denies;
+  }
+  bool GrantsDB(std::string_view db_name) const {
+    bool grants = database_access_.Grants(db_name);
+    if (role_) grants |= role_->GrantsDB(db_name);
+    return grants;
+  }
```
Ignition, line 414: Need a way to revoke/clear the grant/deny

src/communication/websocket/auth.cpp
```diff
@@ -12,19 +12,45 @@
...
+               *user_or_role_);
+    // Missing user/role; delete from cache
+    if (!success) user_or_role_.reset();
   }
+  // Check permissions
+  if (user_or_role_) {
+    return std::visit(
+        utils::Overloaded{
+            [&](auth::User &user) { return user.GetPermissions().Has(permission) == auth::PermissionLevel::GRANT; },
+            [&](auth::Role &role) { return role.permissions().Has(permission) == auth::PermissionLevel::GRANT; }},
```
Ignition, line 48: Can we unify GetPermissions/permissions then we can use one generic lambda, no need for Overloaded

src/query/query_user.hpp
```diff
@@ -0,0 +1,40 @@
...
+  QueryUser(std::optional<std::string> name) : name_{std::move(name)} {}
+  virtual ~QueryUser() = default;
+
+  virtual bool IsAuthorized(const std::vector<AuthQuery::Privilege> &privileges, const std::string &db_name) const = 0;
+
+#ifdef MG_ENTERPRISE
+  virtual std::string GetDefaultDB() const = 0;
+#endif
+
+  std::string name() const { return name_ ? *name_ : ""; }
```
Ignition, line 31: This should be the username for all diagnostic reasons but for permission modification will need to be the role name (when auth module usage)

Modify the `using UserOrRole = std::variant<User, Role>;` so that Role is changed with `RolledUser`? where it has the username string + the role. This can change how QueryUser is constructed. And name() becomes two methods `username()` and `permission_owner_name`?

Ignition (changes requested):

andrejtonev:
We will disable the cache invalidation. I'll leave the code and it should be working, I will just not update the epoch, so the cache user/role will always be "up to date".

src/auth/auth.cpp
andrejtonev: Renamed `Auth::Database::SetDefault` to `SetMain`
SET MAIN DATABASE is the query so I'll reflect that in the code

src/auth/models.cpp
andrejtonev: There are 3 scenarios that we need to cover and I'm not sure migration is the solution here.
1) community -> does not generate the database kv element
2) enterprise no license -> generates it, but leaves it empty
3) enterprise with license -> generates it and actually stores the data

Ignition (approved):

src/query/stream/streams.cpp
```diff
@@ -469,31 +481,39 @@ void Streams::Create(const std::string &stream_name, typename TStream::StreamInf
...
                             interpreter = std::make_shared<Interpreter>(interpreter_context, std::move(db_acc)),
                             result = mgp_result{nullptr, memory_resource},
                             total_retries = interpreter_context->config.stream_transaction_conflict_retries,
                             retry_interval = interpreter_context->config.stream_transaction_retry_interval](
                                const std::vector<typename TStream::Message> &messages) mutable {
+    // Set interpreter's user to the stream owner
+    // NOTE: We generate an empty user to avoid interpreter's access control and rely only on the one inside the stream
+    // itself
+    // TODO: Fix
+    interpreter->SetUser(interpreter_context->auth_checker->GenQueryUser(std::nullopt, std::nullopt));
```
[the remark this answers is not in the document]
andrejtonev: There are essentially 2 authorizers; the global one (AuthChecker) and  a local one (FineGrainedAuthChecker). The later is created at time of PullPlan creation and gets created only of a user/role is defined. This is how it's always been and works in the general flow. 
For some reason streams don't like it and have never defined a user at the interpreter level and so, never actually used it.
I now force a "user" to be defined. Her a user means an auth::user/role or am empty (userless) user. By connecting the userless user to the interpreter I don't generate the FineGrained AuthChecker and the test are happy.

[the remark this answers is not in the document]
andrejtonev: Updated comment

gitbuda:
related to https://github.com/memgraph/memgraph/issues/1359

kgolubic:
@andrejtonev I would use the following release note. Is everything in it?


- Improved the authentication module with better user and role handling mechanisms. Key enhancements include:
  - Elimination of automatic user creation by the auth module
    - Introduction of checks to determine the presence of users or the use of an auth module before proceeding
  - Support for roles with database access
  - The `authenticate()` function now returns a user or role
  - When an auth module is in use, no data is locally saved
  - AuthChecker now generates `QueryUser` (user or role), which is responsible for actual authorization checks

andrejtonev:
I would also specify that when using an auth module, the module needs to return a role and that role is the only way to define privileges for the connection.

PR #1703, opened 2024-02-08
Revert e2e tests and remove flaky ones

antoniofilipovic:
[master < Task] PR


MarkoBarisic (changes requested):
While overwriting `if: false` with `if: true ` is a valid solution, I think it should be removed instead of overwritten. The reason is that it is redundant code and we don't use that syntax anywhere else in the `diff.yaml`.

antoniofilipovic:
> `e ` is a valid solution, I think it should be removed instead of overwritten. The reason is that it is redundant code and we don't use that syntax anywhere else in the

Understood, I wanted to make it easy to revert back to not having e2e tests, will remove it

MarkoBarisic (approved):
LGTM

PR #1708, opened 2024-02-09
Fix go driver tests

DavIvek:
[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR


tests/drivers/go/v4/metadata.go
```diff
@@ -1,90 +1,95 @@
...
+	check_md(result, err)
 
-  // Explicit transaction
-  fmt.Println("Checking explicit transaction metadata...")
-  tx, err := session.BeginTransaction(neo4j.WithTxMetadata(map[string]interface{}{"ver":"transaction", "str":"oho", "num":456}))
-  if err != nil {
-    handle_error(err)
-  }
-  tx.Run("MATCH (n) RETURN n LIMIT 1", map[string]interface{}{})
-  go check_tx(driver)
```
DavIvek, line 86: I don't think we need to run this concurrently because transaction shouldn't finish until we didn't inspect it with SHOW TRANSACTION query.

DavIvek:
Most of the changes are code formatting changes made by VSC. I can revert that if needed.

tests/drivers/go/v4/metadata.go
andrejtonev: So the issue was that we would create the second thread and exit without any synchronization.
Now, I don't know neo4j go driver good enough to say if we can execute everything in a single thread.
Generally we do a single transaction per thread.
I was looking at this at the same time.
My solution would be to just sync correctly.
```
diff --git a/tests/drivers/go/v4/metadata.go b/tests/drivers/go/v4/metadata.go
index ab5f8837b..a3d2ad311 100644
--- a/tests/drivers/go/v4/metadata.go
+++ b/tests/drivers/go/v4/metadata.go
@@ -3,6 +3,7 @@ package main
 import "github.com/neo4j/neo4j-go-driver/neo4j"
 import "log"
 import "fmt"
+import "sync"
 
 func handle_error(err error) {
   log.Fatal("Error occured: %s", err)
@@ -82,9 +83,27 @@ func main() {
   if err != nil {
     handle_error(err)
   }
-  tx.Run("MATCH (n) RETURN n LIMIT 1", map[string]interface{}{})
-  go check_tx(driver)
-  tx.Commit()
+  tx_res, err := tx.Run("MATCH (n) RETURN n LIMIT 1", map[string]interface{}{})
+  if err != nil {
+    handle_error(err)
+  }
+
+  var wg sync.WaitGroup
+  wg.Add(1)
+  go func(wg *sync.WaitGroup) {
+    defer wg.Done()
+    check_tx(driver)
+  }(&wg)
+  wg.Wait()
+
+  _, err = tx_res.Consume()
+  if err != nil {
+    handle_error(err)
+  }
+  err = tx.Commit()
+  if err != nil {
+    handle_error(err)
+  }
 
   fmt.Println("All ok!")
 }
```

DavIvek: I saw that we execute a transaction per thread in other tests but I don't understand why is that needed since we still need to wait for that thread before commiting.

andrejtonev: You cannot have multiple transactions in a single ~thread~ session.
`neo4j.exceptions.TransactionError: Explicit transaction already open`

Here we actually open another session, which is allowed and we execute the tx in that session. So, I think you are right.
We can execute multiple transactions in a single thread, as long as we create another session.

I would leave other multi-threaded tests as they are, because sometimes we actually want to execute queries simultaneously. But, here, we don't actually care.

andrejtonev:
> Most of the changes are code formatting changes made by VSC. I can revert that if needed.

Revert back and just do the small change. No need to pollute the diff. :muscle:

tests/drivers/go/v4/metadata.go
```diff
@@ -1,90 +1,95 @@
...
+import (
+	"fmt"
+	"log"
+
+	"github.com/neo4j/neo4j-go-driver/neo4j"
+)
 
 func handle_error(err error) {
-  log.Fatal("Error occured: %s", err)
+	log.Fatal("Error occured: %s", err)
```
andrejtonev, line 11: Just redo without the formatting changes.

andrejtonev (approved):

andrejtonev:
By the way, great catch! Both the error and my overdoing it with threads. :D

antoniofilipovic (approved):
I am not skillful enough in go, so approve from me

PR #1710, opened 2024-02-11
Fix unbound variable causing crash inside subquery

### LINKED ISSUE #1636: Unbound variable crashes Memgraph and does not yield error
**Memgraph version**
Memgraph v2.13 

**Environment**
Docker 

**Describe the bug**
Here's a bug that crashes memgraph. Normally if you reference a variable which is not in scope anymore you get a Unbound variable error:
```
MATCH (user:User {id:$uid})
with true as _
MATCH(user2:User {id:$uid})
WHERE user.age > 0  // user is not defined
return collect(true) as state
```

If you wrap the internal match in a sub-query like:

```
MATCH (user:User {id:$uid})
CALL {
  //with user // uncommenting this line will yield the correct answer
  MATCH (user2:User {id:$uid})
  WHERE user.age > 0 // user is not defined
  return collect(true) as state
}
return true as passed
```

The result is: Connection was closed by server. This seems to occur only if the unbound variable is used in conjunction with WHERE. Returning the unbound variable does not crash memgraph (see below).

While testing this I also found the following query, which should fail with an unbound variable error but doesn't:

MATCH (user:User {id:$uid})
CALL {
  with true as _
  return user.id as id  // user shouldn't be defined here
}
return id

**Expected behavior**
Memgraph does not crash, and yields and proper error.

imilinovic:
Having an unbound variable inside a subquery causes a crash due to checking all scopes instead of only the local (last) one in case of a call subquery.

[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR

closes #1636 

Unbound variable inside a subquery no longer causes a crash.

### DIFF
--- a/src/query/frontend/semantic/symbol_generator.cpp
+++ b/src/query/frontend/semantic/symbol_generator.cpp
@@ -1,4 +1,4 @@
-// Copyright 2023 Memgraph Ltd.
+// Copyright 2024 Memgraph Ltd.
 //
 // Use of this software is governed by the Business Source License
 // included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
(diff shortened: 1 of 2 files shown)


DavIvek (approved):
💪

PR #1715, opened 2024-02-13
Fixup memory e2e tests

Ignition:
Fixup and reintroduce memory tracker e2e tests

- Remove the e2e that did concurrent `mgp_*` calls on the same transaction (ATM this is unsupported)
- Fix up the concurrent `mgp_global_alloc` test to be testing it more precisely
- Fix bug where QUERY LIMIT would cause an issue during logging
- Reduce the memory limit on detach delete test due to recent memory optimizations around deltas.
- Fixed deadlocks that could happen because of memory tracker

Changelog: Fix deadlocks in jemalloc caused by memory tracker


src/utils/query_memory_tracker.cpp
```diff
@@ -17,18 +17,20 @@
...
   auto *proc_tracker = GetActiveProc();
 
   if (proc_tracker == nullptr) {
-    return;
+    return true;
   }
 
-  proc_tracker->Alloc(static_cast<int64_t>(size));
+  bool ok = proc_tracker->Alloc(static_cast<int64_t>(size));
+  return ok;
```
antoniofilipovic, line 33:
```suggestion
  return proc_tracker->Alloc(static_cast<int64_t>(size));
```

src/memory/global_memory_control.cpp
```diff
@@ -118,10 +120,14 @@ static bool my_commit(extent_hooks_t *extent_hooks, void *addr, size_t size, siz
     return err;
   }
 
+  [[maybe_unused]] auto blocker = memgraph::utils::MemoryTracker::OutOfMemoryExceptionBlocker{};
   if (GetQueriesMemoryControl().IsThreadTracked()) [[unlikely]] {
-    GetQueriesMemoryControl().TrackAllocOnCurrentThread(length);
+    bool ok = GetQueriesMemoryControl().TrackAllocOnCurrentThread(length);
+    DMG_ASSERT(ok);
```
antoniofilipovic, line 126: So, the logic here is to assert we cannot fail on commit. I wouldn't agree with that logic, as we can fail, and then we will crash the database. I would rather remove DMG_ASSERT, leave OOMBlocker as is, and catch OOM on next allocation via alloc hook

src/memory/new_delete.cpp
```diff
@@ -28,6 +28,12 @@ void *newImpl(const std::size_t size) {
     return ptr;
   }
 
+  [[maybe_unused]] auto blocker = memgraph::utils::MemoryTracker::OutOfMemoryExceptionBlocker{};
+  auto maybe_msg = memgraph::utils::MemoryErrorStatus().msg();
+  if (maybe_msg) {
+    throw memgraph::utils::OutOfMemoryException{*maybe_msg};
```
antoniofilipovic, line 34: Does it make sense to do std::move on maybe_msg and update OOM class to accept rvalue ref and move to BasicException

tests/e2e/memory/procedures/query_memory_limit_proc_multi_thread.cpp
```diff
@@ -13,73 +13,123 @@
...
+  auto ptr = std::invoke([&] {
+    void *ptr;
+    [[maybe_unused]] auto alloc_err = mgp_global_alloc(1, &ptr);
+    return safe_ptr{ptr};
+  });
+
+  try {
+    auto futures = std::vector<std::future<result_t>>{};
+    futures.reserve(N_THREADS);
+    // change communication between threads with feature and promise
```
antoniofilipovic, line 97: This is done, so I would remove comment

antoniofilipovic (changes requested):
I would remove on commit hook DMG_ASSERT, as we don't know how to recover from that, but from alloc we have a mechanism. And exception can occur there

src/memory/global_memory_control.cpp
Ignition: 1. DMG_ASSERT is only for a debug build, so production release builds are never effected
2. As setup ATM, OOMBlocker should provide the guarantee that this will always be true.

antoniofilipovic: Makes sense

antoniofilipovic (approved):

PR #1717, opened 2024-02-13
Forbid writing to cluster-managed main on restart

as51340:
When managed by coordinator, main should't be writeable. Only after receiving RPC for accepting messages, main can start accepting write queries again. RPC will be sent if after main restart, main will stay main and not become replica.

[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR


andrejtonev (commented):
Good work!
Just 2 notes:
1) We should set the enable_write to false on startup/restart
2) This is not thread safe, right?

as51340:
By thread safe you mean, one thread can set it to true from rpc and other could just be initializing it or?

andrejtonev:
> By thread safe you mean, one thread can set it to true from rpc and other could just be initializing it or?

We have multiple threads that could be reading/writing this variable. The RPC is setting it and all query execution threads are reading it. My comment was me basically just checking that replication state is still not thread safe and that we need to fix that.

as51340:
> We have multiple threads that could be reading/writing this variable. The RPC is setting it and all query execution threads are reading it. My comment was me basically just checking that replication state is still not thread safe and that we need to fix that.


Yes, agree. I can put this under atomic but still whole state isn't thread safe. You want me to put writing_enabled under atomic before merging this?

andrejtonev:
> Yes, agree. I can put this under atomic but still whole state isn't thread safe. You want me to put writing_enabled under atomic before merging this?

I think it's fine for now. We need to fix the whole state. Just set it to false on startup and I think that's it.

as51340:
> I think it's fine for now. We need to fix the whole state. Just set it to false on startup and I think that's it.

It already is false on startup I think

src/coordination/coordinator_handlers.cpp
```diff
@@ -148,7 +154,7 @@ void CoordinatorHandlers::PromoteReplicaToMainHandler(replication::ReplicationHa
       }
     }
   }
-  spdlog::error(fmt::format("FICO : Promote replica to main was success {}", std::string(req.main_uuid_)));
+  spdlog::info("Promote replica to main was success {}", std::string(req.main_uuid_));
```
antoniofilipovic, line 157: Nice

src/utils/functional.hpp
```diff
@@ -0,0 +1,27 @@
...
+// licenses/APL.txt.
+
+#pragma once
+
+#include <algorithm>
+#include <vector>
+
+namespace memgraph::utils {
+
+template <class F, class T, class R = typename std::result_of<F(T)>::type, class V = std::vector<R>>
```
antoniofilipovic, line 19: Can you use here std::result_of_t?

antoniofilipovic (commented):
Only one comment, looks good

andrejtonev:
> It already is false on startup I think

Ah in Fetch it's false. Okay!

andrejtonev (approved):

antoniofilipovic (approved):

src/utils/functional.hpp
as51340: will change

as51340:
@kgolubic Release note: To avoid possible split brain issue and data inconsistencies, writing on main is by default forbidden after it gets restarted until the coordinator explicitly allows old main to be writeable. This will only happen if there was no new main in the cluster in the time between main failed and came back.

PR #1723, opened 2024-02-15
Drop graph functionality

Josipmrden:
Drop graph functionality will be added to delete everything in the database in an efficient manner

Documentation pr: https://github.com/memgraph/documentation/pull/616

[master < Task] PR

@kgolubic Added a new command, `DROP GRAPH`, which deletes data in the database in an efficient manner.


src/query/trigger.cpp
```diff
@@ -432,6 +432,21 @@ void TriggerStore::DropTrigger(const std::string &name) {
   storage_.Delete(name);
 }
 
+void TriggerStore::DropAll() {
+  std::unique_lock store_guard{store_lock_};
+  std::vector<std::string> trigger_names{};
+  for (auto const &[name, trigger_data] : storage_) {
+    trigger_names.push_back(name);
```
gvolfing, line 439: Since we know the size of `triger_names` in advance, it might be worth reserving the vector's capacity.

src/storage/v2/indices/indices.cpp
```diff
@@ -42,6 +42,11 @@ void Indices::RemoveObsoleteEntries(uint64_t oldest_active_start_timestamp, std:
       ->RemoveObsoleteEntries(oldest_active_start_timestamp, std::move(token));
 }
 
+void Indices::DropGraphClearIndices() const {
+  static_cast<InMemoryLabelIndex *>(label_index_.get())->DropGraphClearIndices();
+  static_cast<InMemoryLabelPropertyIndex *>(label_property_index_.get())->DropGraphClearIndices();
+}
```
gvolfing, line 48: Since we added `EdgeTypeIndices`, I think dropping them should also be here.

src/storage/v2/inmemory/label_index.cpp
```diff
@@ -235,4 +235,13 @@ std::vector<LabelId> InMemoryLabelIndex::Analysis() const {
   }
   return res;
 }
+
+void InMemoryLabelIndex::DropGraphClearIndices() {
+  for (auto &[label, storage] : index_) {
+    storage.clear();
+  }
+  index_.clear();
```
gvolfing, line 243: `std::map<>::clear()` will destruct all the elements it is holding, since the `SkipList` destructor cleans itself up completely, there is no need to explicitly call `storage.clear()` on every skip-list here.
The function call on line 243 should be enough.

src/storage/v2/inmemory/label_property_index.cpp
```diff
@@ -498,4 +498,14 @@ void InMemoryLabelPropertyIndex::AbortEntries(LabelId label,
     }
   }
 }
+
+void InMemoryLabelPropertyIndex::DropGraphClearIndices() {
+  for (auto &[label_prop, storage] : index_) {
+    storage.clear();
+  }
+  index_.clear();
```
gvolfing, line 506: Same as above.

src/storage/v2/inmemory/storage.cpp
```diff
@@ -2214,6 +2214,66 @@ void InMemoryStorage::FreeMemory(std::unique_lock<utils::ResourceLock> main_guar
...
+    unlink_remove_clear(gc_deltas.deltas_);
+  }
+  // 1.b.2) clear the list of deltas deques
+  linked_undo_buffers.clear();
+
+  indices_.DropGraphClearIndices();
+  constraints_.DropGraphClearConstraints();
+
+  vertices_.clear();
+  edges_.clear();
```
gvolfing, line 2272: Just a comment here, so I don't forget, the new metadata `SkipList` will have to be added here as well, once [that PR](https://github.com/memgraph/memgraph/pull/1808) is ready.

src/storage/v2/inmemory/unique_constraints.cpp
```diff
@@ -524,4 +524,11 @@ void InMemoryUniqueConstraints::Clear() {
 }
 bool InMemoryUniqueConstraints::empty() const { return constraints_.empty() && constraints_by_label_.empty(); }
 
+void InMemoryUniqueConstraints::DropGraphClearConstraints() {
+  for (auto &[label_props, storage] : constraints_) {
+    storage.clear();
+  };
+  constraints_.clear();
```
gvolfing, line 531: Same as above.

gvolfing (changes requested):

src/storage/v2/indices/indices.cpp
gvolfing: [InMemoryEdgeTypeIndex](https://github.com/memgraph/memgraph/blob/master/src/storage/v2/inmemory/edge_type_index.hpp)

src/query/trigger.cpp
Ignition: Why even have a temporary `triger_names`?
Why not `DeletePrefix("")`?

src/storage/v2/inmemory/storage.cpp
```diff
@@ -2398,8 +2346,20 @@ std::vector<std::pair<LabelId, PropertyId>> InMemoryStorage::InMemoryAccessor::D
 }
 
 void InMemoryStorage::InMemoryAccessor::DropGraph() {
+  // STEP 1 of GC Cleanup
+  std::list<Gid> current_deleted_vertices;
+  std::list<Gid> current_deleted_edges;
+  GCRapidDeltaCleanup(current_deleted_vertices, current_deleted_edges);
```
Ignition, line 2352: This now makes it clear that `DropGraph` needs different Delta cleanup. You know all vertexes and edges are being deleted. You don't need to collect a subset to remove, because all of them will be removed. With that in mind, you do not need to unlink any delta, you just have to prevent GC from navigating and cleaning those deltas by taking they away from GCs control.

gvolfing (approved):
Looks good to me.

Ignition (dismissed):
AFAICT looks good to me

src/storage/v2/inmemory/storage.cpp
```diff
@@ -2466,13 +2466,14 @@ std::vector<std::pair<LabelId, PropertyId>> InMemoryStorage::InMemoryAccessor::D
...
-  GCRapidDeltaCleanup(current_deleted_vertices, current_deleted_edges);
-
   auto *mem_storage = static_cast<InMemoryStorage *>(storage_);
 
+  // we take the control from the GC to clear any deltas
+  auto gc_guard = std::unique_lock{mem_storage->gc_lock_};
+  mem_storage->garbage_undo_buffers_.WithLock([&](auto &garbage_undo_buffers) { garbage_undo_buffers.clear(); });
+  mem_storage->committed_transactions_.WithLock([&](auto &committed_transactions) { committed_transactions.clear(); });
+
+  // also, we're the only transaction running, so we can safely remove the data as well
```
Ignition, line 2539: So now we remove deltas without unlinking.
This is done before the vertices/edges those deltas are linked to are remove.

Is it possible to concurrently somewhere else, like snapshot, to have a skip list accessor where the yet to be removed vertex/edge would try and follow the delta that wasn't unlinked.

IMO its safer to:
- lock gc
- mass clear the vertex/edge skip_lists
- clear the deltas

Ignition (changes requested):

src/storage/v2/inmemory/storage.cpp
Ignition: Chat with @Josipmrden, we don't think this scenario is possible, as snapshot would require txn accessor which is blocked by DROP GRAPH's txn accessor

Ignition (approved):

PR #1728, opened 2024-02-16
Changing replication roles backed-up by RAFT

as51340:
This PR deals with backing-up replication roles with raft log. 
We don't have a way to check if coordinator is alive -> unknown.
is_alive -> health
Follower doesn't have info about whether replication instance is alive or not.

[master < Epic] PR
- [ ] Write E2E tests

[master < Task] PR


src/coordination/coordinator_cluster_state.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+auto CoordinatorClusterState::MainExists() const -> bool {
+  return std::any_of(instance_roles.begin(), instance_roles.end(), [](auto const &entry) {
+    return entry.second == replication_coordination_glue::ReplicationRole::MAIN;
+  });
+}
+
+auto CoordinatorClusterState::IsMain(std::string const &instance_name) const -> bool {
+  auto const it = instance_roles.find(instance_name);
+  return it != instance_roles.end() && it->second == replication_coordination_glue::ReplicationRole::MAIN;
```
Ignition, line 26: This is a cpp file consider `using memgraph::replication_coordination_glue::ReplicationRole;` to reduce verbosity throughout the file.

src/coordination/coordinator_cluster_state.cpp
```diff
@@ -0,0 +1,124 @@
...
+    case RaftLogAction::REGISTER_REPLICATION_INSTANCE:
+      instance_roles[instance_name] = replication_coordination_glue::ReplicationRole::REPLICA;
+      break;
+    case RaftLogAction::UNREGISTER_REPLICATION_INSTANCE:
+      spdlog::info("Unregistering instance: {}", instance_name);
+      instance_roles.erase(instance_name);
+      spdlog::info("Instance {} unregistered", instance_name);
+      break;
+    case RaftLogAction::SET_INSTANCE_AS_MAIN:
+      MG_ASSERT(instance_roles.find(instance_name) != instance_roles.end(),
```
Ignition, line 50: MG_ASSERT...use only if this something that should never happen.
Also reduce the cost, `instance_roles.find(instance_name)` is called, use it.
instance_roles[instance_name] is basically a 2nd find.

src/coordination/coordinator_cluster_state.cpp
```diff
@@ -0,0 +1,124 @@
...
+                "Instance does not exist as part of raft state!");
+      instance_roles[instance_name] = replication_coordination_glue::ReplicationRole::REPLICA;
+      break;
+  }
+}
+
+auto CoordinatorClusterState::Serialize(ptr<buffer> &data) -> void {
+  auto const role_to_string = [](auto const &role) {
+    switch (role) {
+      case replication_coordination_glue::ReplicationRole::MAIN:
```
Ignition, line 65: consider `using enum` to make the cases shorter

src/coordination/coordinator_cluster_state.cpp
```diff
@@ -0,0 +1,124 @@
...
+        return "main";
+      case replication_coordination_glue::ReplicationRole::REPLICA:
+        return "replica";
```
Ignition, line 68: Prefer returning string_view

src/coordination/coordinator_cluster_state.cpp
```diff
@@ -0,0 +1,124 @@
...
+    }
+  };
+
+  auto const entry_to_string = [&role_to_string](auto const &entry) {
+    return entry.first + "_" + role_to_string(entry.second);
```
Ignition, line 73: Prefer using fmt format

src/coordination/coordinator_cluster_state.cpp
```diff
@@ -0,0 +1,124 @@
...
+  };
+
+  auto instances_str_view = instance_roles | ranges::views::transform(entry_to_string);
+  uint32_t size =
```
Ignition, line 77: 2 passes, this first pass does not need strings, you can calculate the would be size without doing `entry_to_string`

src/coordination/coordinator_cluster_state.cpp
```diff
@@ -0,0 +1,124 @@
...
+  };
+
+  CoordinatorClusterState cluster_state;
+  buffer_serializer bs(data);
+  while (bs.size() > 0) {
+    auto const entry = bs.get_str();
+    auto const first_dash = entry.find('_');
+    auto const instance_name = entry.substr(0, first_dash);
+    auto const role_str = entry.substr(first_dash + 1);
+    cluster_state.InsertInstance(instance_name, str_to_role(role_str));
```
Ignition, line 101: Do we need to transport human readable roles?
Why not just a number, or single character?

src/coordination/coordinator_cluster_state.cpp
```diff
@@ -0,0 +1,124 @@
...
+  }
+  return cluster_state;
+}
+
+auto CoordinatorClusterState::GetInstances() const -> std::vector<std::pair<std::string, std::string>> {
+  auto const role_to_string = [](auto const &role) -> std::string {
+    switch (role) {
+      case replication_coordination_glue::ReplicationRole::MAIN:
+        return "main";
```
Ignition, line 110: again...could be string_view

src/coordination/coordinator_state_machine.cpp
```diff
@@ -51,61 +78,93 @@ auto CoordinatorStateMachine::commit_config(ulong const log_idx, ptr<cluster_con
...
                                                    ptr<buffer> &data_out, bool &is_last_obj) -> int {
-  // Put dummy data.
-  data_out = buffer::alloc(sizeof(int32));
-  buffer_serializer bs(data_out);
-  bs.put_i32(0);
+  spdlog::info("read logical snapshot object, obj_id: {}", obj_id);
 
+  ptr<SnapshotCtx> ctx = nullptr;
+  {
+    std::lock_guard<std::mutex> ll(snapshots_lock_);
```
Ignition, line 90: start using CTAD, makes refactoring easier if lockable type changes.
auto guard = std::lock_guard{snapshots_lock_};

src/coordination/coordinator_state_machine.cpp
```diff
@@ -51,61 +78,93 @@ auto CoordinatorStateMachine::commit_config(ulong const log_idx, ptr<cluster_con
...
 
-auto CoordinatorStateMachine::apply_snapshot(snapshot &s) -> bool {
-  spdlog::info("apply snapshot {} term {}", s.get_last_log_idx(), s.get_last_log_term());
   {
-    auto lock = std::lock_guard{last_snapshot_lock_};
-    ptr<buffer> snp_buf = s.serialize();
-    last_snapshot_ = snapshot::deserialize(*snp_buf);
+    std::lock_guard<std::mutex> ll(snapshots_lock_);
+    auto entry = snapshots_.find(snapshot.get_last_log_idx());
+    assert(entry != snapshots_.end());
```
Ignition, line 115: DMG_ASSERT

src/coordination/coordinator_cluster_state.cpp
as51340: yes yes MG_ASSERT is here intentionally, that shouldn't happen
About the cost, agree

src/coordination/coordinator_cluster_state.cpp
as51340: how exactly? + ranges should just be lazy I think right? So at the end there should be just one pass for accumulate

src/coordination/coordinator_cluster_state.cpp
as51340: agree we don't but I will leave it as it is for now because it is easier to debug and there are million of things which I don't about state machine. Will write a task in the backlog for the future

src/coordination/coordinator_cluster_state.cpp
as51340: how, why? won't string_view bind to a temporary? how would the signature of the method look like then?

src/coordination/replication_instance.cpp
```diff
@@ -52,44 +45,51 @@ auto ReplicationInstance::IsReadyForUUIDPing() -> bool {
...
 }
 
-auto ReplicationInstance::DemoteToReplica(HealthCheckInstanceCallback replica_succ_cb,
-                                          HealthCheckInstanceCallback replica_fail_cb) -> bool {
+auto ReplicationInstance::PromoteToMainAsFollower(utils::UUID new_uuid, HealthCheckInstanceCallback main_succ_cb,
+                                                  HealthCheckInstanceCallback main_fail_cb) -> void {
+  main_uuid_ = new_uuid;
+  succ_cb_ = main_succ_cb;
+  fail_cb_ = main_fail_cb;
+}
```
antoniofilipovic, line 71: Maybe instead of two functions, we can have just one with template bool parameter, similar to TryRegister and Register in replication_handler.cpp

src/coordination/raft_state.cpp
```diff
@@ -100,8 +101,8 @@ auto RaftState::MakeRaftState(BecomeLeaderCb become_leader_cb, BecomeFollowerCb
     throw RaftCouldNotParseFlagsException("Failed to parse flags: {}", e.what());
   }
```
antoniofilipovic, line 102: We don't parse here anymore

src/coordination/coordinator_instance.cpp
```diff
@@ -61,46 +70,67 @@ auto CoordinatorInstance::FindReplicationInstance(std::string_view replication_i
...
+  } else {
+    auto process_repl_instance_as_follower = [](auto const &instance) -> InstanceStatus {
+      return {.instance_name = instance.first, .cluster_role = instance.second, .health = "unknown"};
+    };
+    {
+      // TODO: (andi) Not this lock should be taken.
+      // Will be fixed once I move socket addresses and other required info to the RAFT log.
+      auto lock = std::shared_lock{coord_instance_lock_};
+      std::ranges::transform(raft_state_.GetInstances(), std::back_inserter(instances_status),
+                             process_repl_instance_as_follower);
```
antoniofilipovic, line 113: Maybe I am missing something, but can't we go here over repl_instances and then lock is fine?

src/coordination/coordinator_instance.cpp
```diff
@@ -123,169 +153,53 @@ auto CoordinatorInstance::TryFailover() -> void {
...
-  if (std::ranges::any_of(repl_instances_, socket_address_matches)) {
-    return RegisterInstanceCoordinatorStatus::ENDPOINT_EXISTS;
-  }
+                new_main_instance_name, *latest_epoch, *latest_commit_timestamp);
 
   if (!raft_state_.RequestLeadership()) {
-    return RegisterInstanceCoordinatorStatus::NOT_LEADER;
+    spdlog::error("Failed to request leadership for promoting instance to main {}.", new_main_instance_name);
+    return;
   }
```
antoniofilipovic, line 165: I would do this much sooner. If it is not leader no need to do all RPCs before

src/coordination/coordinator_instance.cpp
```diff
@@ -362,35 +345,231 @@ auto CoordinatorInstance::UnregisterReplicationInstance(std::string instance_nam
...
 
-// TODO: (andi) Add to the RAFT log.
-auto CoordinatorInstance::SetMainUUID(utils::UUID new_uuid) -> void { main_uuid_ = new_uuid; }
+        if (!new_main.PromoteToMainAsLeader(new_main_uuid, std::move(repl_clients_info),
+                                            &CoordinatorInstance::MainSuccessCallback,
+                                            &CoordinatorInstance::MainFailCallback)) {
+          spdlog::error("Failed to promote instance {} to main", instance_name);
+          // TODO: (andi) What to do on network failure when appended to raft log?
+        }
+        // Must be executed under the lock.
```
antoniofilipovic, line 525: I would like to sync with you why you chose this approach. I think this can fail in following way:
- the main coordinator tries failover
- sends a commit to all other coordinators
- The coordinator who tried failover doesn't succeed because of some failure on one of the replica instances
- the main coordinator dies
- new coordinator becomes main, but has wrong state - thinks we succeded in failover but we didn't

src/coordination/coordinator_instance.cpp
```diff
@@ -340,16 +266,73 @@ void CoordinatorInstance::ReplicaSuccessCallback(std::string_view repl_instance_
...
+    return instance.ReplicationSocketAddress() == config.ReplicationSocketAddress();
+  };
+
+  if (std::ranges::any_of(repl_instances_, repl_socket_address_matches)) {
+    return RegisterInstanceCoordinatorStatus::REPL_ENDPOINT_EXISTS;
+  }
+
+  if (!raft_state_.RequestLeadership()) {
+    return RegisterInstanceCoordinatorStatus::NOT_LEADER;
+  }
```
antoniofilipovic, line 313: Is there a reason why is this not earlier in function call

src/coordination/coordinator_instance.cpp
```diff
@@ -362,35 +345,231 @@ auto CoordinatorInstance::UnregisterReplicationInstance(std::string instance_nam
...
+
+  auto const res = raft_state_.AppendSetInstanceAsMain(instance_name);
+  if (!res->get_accepted()) {
+    spdlog::error(
+        "Failed to accept request for promoting instance {}. Most likely the reason is that the instance is not "
+        "the leader.",
+        instance_name);
+    return SetInstanceToMainCoordinatorStatus::RAFT_COULD_NOT_ACCEPT;
+  }
```
antoniofilipovic, line 415: In general, we need to sync, I am not sure how this should be handled in case of coordinator failure combined with setting instance as main failure

antoniofilipovic (changes requested):
Good work overall :muscle: 

I need to sync with you, might be I am missing something but for now it seems we can end in incorrect state on different coordinator instances

src/coordination/coordinator_cluster_state.cpp
Ignition: You run the range twice...once to calculate size (which does unnecessary temporary strings). And a 2nd time to put into the buffer.

src/coordination/coordinator_cluster_state.cpp
Ignition: These string_views would be backed by string literals, which are essentially `const char *` which have a static storage duration.

src/coordination/coordinator_cluster_state.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+using replication_coordination_glue::ReplicationRole;
+
+auto CoordinatorClusterState::MainExists() const -> bool {
+  return std::ranges::any_of(instance_roles, [](auto const &entry) { return entry.second == ReplicationRole::MAIN; });
+}
+
+auto CoordinatorClusterState::IsMain(std::string_view instance_name) const -> bool {
+  auto const it = instance_roles.find(instance_name);
+  return it != instance_roles.end() && it->second == ReplicationRole::MAIN;
```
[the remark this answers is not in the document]
as51340: should be improved at some point

src/coordination/replication_instance.cpp
as51340: Yes

src/coordination/raft_state.cpp
as51340: Will be removed

src/coordination/coordinator_instance.cpp
as51340: Lock on coord cluster state preferred

src/coordination/coordinator_cluster_state.cpp
```diff
@@ -0,0 +1,181 @@
...
+auto CoordinatorClusterState::MainExists() const -> bool {
+  auto lock = std::shared_lock{log_lock_};
+  return std::ranges::any_of(instance_roles_,
+                             [](auto const &entry) { return entry.second.role == ReplicationRole::MAIN; });
+}
+
+auto CoordinatorClusterState::IsMain(std::string_view instance_name) const -> bool {
+  auto lock = std::shared_lock{log_lock_};
+  auto const it = instance_roles_.find(instance_name);
+  return it != instance_roles_.end() && it->second.role == ReplicationRole::MAIN;
```
antoniofilipovic, line 54: find_if here can be used here

src/coordination/coordinator_cluster_state.cpp
antoniofilipovic: :value_learn:

src/coordination/coordinator_state_machine.cpp
```diff
@@ -12,37 +12,85 @@
...
+auto CoordinatorStateMachine::pre_commit(ulong const /*log_idx*/, buffer & /*data*/) -> ptr<buffer> { return nullptr; }
+
 auto CoordinatorStateMachine::commit(ulong const log_idx, buffer &data) -> ptr<buffer> {
   buffer_serializer bs(data);
-  std::string str = bs.get_str();
 
-  spdlog::info("commit {} : {}", log_idx, str);
+  auto const [parsed_data, log_action] = DecodeLog(data);
+  cluster_state_.DoAction(parsed_data, log_action);
+  // std::invoke(raft_commit_cb_, parsed_data, log_action);
```
antoniofilipovic, line 90: I would remove this comment here

antoniofilipovic (commented):
Looks good now :muscle: I have a few comments and a general overview, I didn't see in `distributed_coordinators.py` tests a new test with distributed coordinators failover where one dies and the other one continues working, I think we should add that now also

src/coordination/coordinator_instance.cpp
```diff
@@ -465,5 +522,14 @@ auto CoordinatorInstance::ChooseMostUpToDateInstance(
 
   return new_main_res;
 }
+
+auto CoordinatorInstance::IsMain(std::string_view instance_name) const -> bool {
+  return raft_state_.IsMain(instance_name);
```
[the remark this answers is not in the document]
as51340: No we do the check atm with IsAlive. But this will change, I want to have in the raft log at any point just one main instance. Next PR will improve

antoniofilipovic (approved):
:ship: :it:

as51340:
@kgolubic Changing instance's replication role is now backed by Raft.

PR #1371, opened 2023-10-16
Update operating systems

gitbuda:

PR #1731, opened 2024-02-16
Fix zstd under text search

gitbuda:
NOTE: Under mgcxx -> zstd 1.5.5 (latest stable) is used `Compiling zstd-sys v2.0.9+zstd.1.5.5`.
NOTE: Under latest versions of `pulsar-client-cpp` zstd is a required dependency.

IDEA#1 -> FAIL: An option is to disable zstd for pulsar-client-cpp 2.8.1 because it seems to be not required, but that will bite us later because latest versions require zstd (we'll have to resolve the problem as soon as we decide to upgrade pulsar) -> try to align versions of zstd (at least experiment). -> for some reason doesn't help 🤔

IDEA#2 -> FAIL: Problem is also `mgcxx` build because it's a separate build process -> test what would happen if we build under toolchain-v4? -> Test by forcing clang as the only compiler. -> GCC was picked -> with clang it's the same... -> but this PR has a useful change (setting the right compiler for `mgcxx`).

IDEA#3: For some reason, `memgraph` binary gets linked properly -> why is that so? What's different in the test case?

IDEA#4: **Compiling everything with zstd v1.5.5** seems to be working! 🥳 -> done under https://github.com/memgraph/memgraph/pull/1371

TODO: Add the ability alter latest toolchain (extent environment/toolchain/v4/build.sh) -> because the idea is just to alted toolchain with zstd v1.5.5 to test would that be feasible.

### Resources

* https://stackoverflow.com/questions/6940384/how-to-deal-with-symbol-collisions-between-statically-linked-libraries/6940389
* https://discourse.cmake.org/t/avoiding-duplicate-symbols-when-linking-a-library-multiple-times/3816


gitbuda:
Done under https://github.com/memgraph/memgraph/pull/1371

PR #1738, opened 2024-02-19
Fix early exit for OR expression

Josipmrden:
Fix or expression not making an early exit if first branch is evaluated to true.

[master < Task] PR

Release note: "Added an early exit when doing OR filter expressions if the first expression evaluates to `true`"
@kgolubic


src/query/interpret/eval.hpp
```diff
@@ -261,6 +260,20 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
...
 
+  TypedValue Visit(OrOperator &op) override {
+    auto value1 = op.expression1_->Accept(*this);
+    if (value1.IsBool() && value1.ValueBool()) {
+      // If first expression is true, don't evaluate the second one.
+      return value1;
+    }
+    auto value2 = op.expression2_->Accept(*this);
+    try {
+      return value1 || value2;
```
as51340, line 271: if 
```suggestion
      if (value2.IsBool()) return value2;
      return false;
```

as51340: if we came to this point, value1 is not bool or false so I think correct thing is just to return value2 ?

as51340 (changes requested):

as51340 (approved):
:muscle:

src/query/interpret/eval.hpp
```diff
@@ -261,6 +260,20 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
     }
   }
 
+  TypedValue Visit(OrOperator &op) override {
+    auto value1 = op.expression1_->Accept(*this);
+    if (value1.IsBool() && value1.ValueBool()) {
+      // If first expression is true, don't evaluate the second one.
```
antoniofilipovic, line 266: No need for comment

antoniofilipovic (approved):
LGTM, left comment tiny

kgolubic:
Release note entry:

> Added an early exit when doing `OR` filter expressions if the first expression evaluates to `true`.

PR #1747, opened 2024-02-21
Compact Delta 80B -> 56B

Ignition:
Make special structure for `old_disk_key`. `std::optional<std::string>` was 40B, which is the largest member of out action union. Replaced with 8B, structure.

This makes largest member now `vertex_edge` at 24B, this means Delta is now only 56B.

🥳🎉 Now less than a cacheline 🎊✨

For reference you can compare with the last Delta improvement #1688 
image
Query used: `FOREACH (i in range(1, 3000000) | CREATE (:Label1:Label2:Label3:Label4:Label5:Label6));`

### DIFF
--- a/src/storage/v2/delta.hpp
+++ b/src/storage/v2/delta.hpp
@@ -123,6 +123,26 @@ inline bool operator==(const PreviousPtr::Pointer &a, const PreviousPtr::Pointer
 
 inline bool operator!=(const PreviousPtr::Pointer &a, const PreviousPtr::Pointer &b) { return !(a == b); }
 
+struct opt_str {
+  opt_str(std::optional<std::string> const &other) : str_{other ? new_cstr(*other) : nullptr} {}
+
+  ~opt_str() { delete[] str_; }
+
+  auto as_opt_str() const -> std::optional<std::string> {
+    if (!str_) return std::nullopt;
+    return std::optional<std::string>{std::in_place, str_};
+  }
+
+ private:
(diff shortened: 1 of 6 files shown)


src/storage/v2/delta.hpp
```diff
@@ -123,6 +123,30 @@ inline bool operator==(const PreviousPtr::Pointer &a, const PreviousPtr::Pointer
 
 inline bool operator!=(const PreviousPtr::Pointer &a, const PreviousPtr::Pointer &b) { return !(a == b); }
 
+struct opt_str {
+  opt_str(std::optional<std::string> const &other) : str_{other ? new_cstr(*other) : nullptr} {}
+
+  ~opt_str() {
+    if (!str_) return;
+    auto *to_del = std::exchange(str_, nullptr);
+    delete to_del;
```
andrejtonev, line 132: Any reason to not just `delete str_`?

Ignition: true, should be fine. I'll modify and push it up.

andrejtonev (approved):
:muscle:

Josipmrden (approved):

kgolubic:
@Ignition is there a related doc PR? To me, it seems that the 80B value needs to be updated to 56B on this page only https://memgraph.com/docs/fundamentals/storage-memory-usage (an the math needs to be done in formulas :) ).

Do you want me to take over the docs update?

For RN I would go with:
- `Reduced the seize of Delta from 80B to 56B`

PR #1753, opened 2024-02-24
Fix pre-commit on debian-12

gitbuda:
A quick fix, there is a TODO for proper fix across all operating systems.


hal-eisen-MG:
Can we close this?

hal-eisen-MG:
Moot

PR #1759, opened 2024-02-26
Main lock friendly GC + snapshot

Ignition:
GC and snapshot should not block new transactions unnecessarily.

- Snapshot only need unique lock if in analytical mode.
- GC may already be passed main lock when changing between storage modes and doing free memory accordingly. GC can use that information for optimization. 
- In all other GC cases, optimizations should not be achieved by getting unique lock.


Josipmrden (approved):

katarinasupe:
Is the 2.15.1 a final one? @Ignition @gitbuda

PR #1743, opened 2024-02-20
Ensure replication works on HA cluster in different scenarios

antoniofilipovic:
PR's idea is to cover different HA scenarios and ensure replication works in all of them.

Replication didn't work when the instance was a few epochs or commits away due to issues.

There were a few issues:
- [x] epochs were not recovered as they should on WAL recovery, which would cause MAIN to think REPLICA diverges, as MAIN didn't log in its history that such epoch did exist due to a bug in `durability.cpp `in WAL file recovery
- [x] our branching point logic was wrong, we would set the branching point if the instance is few commits away even

This PR doesn't still cover cases when MAIN is behind REPLICA. This will be done in the next PR as a force sync of the database by doing drop and recreate. This can happen in the MT case as well as in a single database case.

[master < Task] PR
- [ ] Check, and update documentation if necessary
- [ ] Provide the full content or a guide for the final git message

To keep docs changelog up to date, one more thing to do:
- [ ] Write a release note here, including added/changed clauses
- [ ] Tag someone from docs team in the comments

PR #1777, opened 2024-03-01
Add force sync REPLICA with MAIN

antoniofilipovic:
This PR enables the cluster to continue executing if the MAIN is behind REPLICA. It can happen in the scenario as in the test case, or when one of the tenants on other REPLICA is behind MAIN's tenant. In both cases, the result is the same -> force sync of tenant with MAIN's tenant.

This is as agreed in user story for 2.15/2.16 -> [Link here](https://www.notion.so/memgraph/Automatic-failover-to-enable-high-availability-phase-III-3d58a7d566f740f9b0a1a21006ec1df5?p=553251c9f1494954a34be4463d0dfcce&pm=s)

[master < Task] PR


src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -329,6 +334,78 @@ void InMemoryReplicationHandlers::SnapshotHandler(dbms::DbmsHandler *dbms_handle
...
+    storage::replication::SnapshotRes res{false, 0};
+    slk::Save(res, res_builder);
+    return;
+  }
+
+  storage::replication::Decoder decoder(req_reader);
+
+  auto *storage = static_cast<storage::InMemoryStorage *>(db_acc->get()->storage());
+
+  auto storage_guard = std::unique_lock{storage->main_lock_};
```
andrejtonev, line 359: I would rather use `db.try_exclusively([](auto &in) {...});`
That way we know we are absolutely the only one handling this database.

src/storage/v2/replication/replication_client.cpp
```diff
@@ -76,13 +76,37 @@ void ReplicationStorageClient::UpdateReplicaState(Storage *storage, DatabaseAcce
...
+        if (!this->ForceResetStorage(storage)) {
+          spdlog::error(
+              "You cannot register REPLICA {} to this MAIN because MAIN couldn't reset REPLICA's storage."
+              "Please resolve data conflicts and start the "
+              "replication on a clean instance.",
+              client_.name_);
+          return;
+        }
+        state = replication::ReplicaState::RECOVERY;
+        this->RecoverReplica(0, storage);
```
andrejtonev, line 105: This should be executed in via the client's thread pool
```
client_.thread_pool_.AddTask([storage, current_commit_timestamp, gk = std::move(db_acc), this] {
    this->RecoverReplica(current_commit_timestamp, storage);
});
```

src/utils/functional.hpp
```diff
@@ -19,9 +19,9 @@
 namespace memgraph::utils {
 
 template <template <typename, typename...> class Container, typename T, typename Allocator = std::allocator<T>,
-          typename F, typename R = std::invoke_result_t<F, T>>
+          typename F = std::identity, typename R = std::decay_t<std::invoke_result_t<F, T>>>
 requires ranges::range<Container<T, Allocator>> &&
-    (!std::same_as<Container<T, Allocator>, std::string>)auto fmap(F &&f, const Container<T, Allocator> &v)
+    (!std::same_as<Container<T, Allocator>, std::string>)auto fmap(const Container<T, Allocator> &v, F f = {})
         -> std::vector<R> {
   return v | ranges::views::transform(std::forward<F>(f)) | ranges::to<std::vector<R>>();
```
andrejtonev, line 26: Generally better to pass in callable objects via &&.
If you choose to copy it, do not forward it.

andrejtonev (commented):
First pass. A few comments.

as51340:
I will leave this one to Andreja to handle it, high-level looks good to me

src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -329,6 +334,78 @@ void InMemoryReplicationHandlers::SnapshotHandler(dbms::DbmsHandler *dbms_handle
...
 }
 
+void InMemoryReplicationHandlers::ForceResetStorageHandler(dbms::DbmsHandler *dbms_handler,
+                                                           const std::optional<utils::UUID> &current_main_uuid,
+                                                           slk::Reader *req_reader, slk::Builder *res_builder) {
+  storage::replication::ForceResetStorageReq req;
+  slk::Load(&req, req_reader);
+  auto db_acc = GetDatabaseAccessor(dbms_handler, req.db_uuid);
+  if (!db_acc) {
+    storage::replication::SnapshotRes res{false, 0};
```
antoniofilipovic, line 344: Wrong response

src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -329,6 +334,78 @@ void InMemoryReplicationHandlers::SnapshotHandler(dbms::DbmsHandler *dbms_handle
...
+    slk::Save(res, res_builder);
+    return;
+  }
+  if (!current_main_uuid.has_value() || req.main_uuid != current_main_uuid) [[unlikely]] {
+    LogWrongMain(current_main_uuid, req.main_uuid, storage::replication::SnapshotReq::kType.name);
+    storage::replication::SnapshotRes res{false, 0};
```
antoniofilipovic, line 350: Wrong res

src/storage/v2/replication/replication_client.cpp
```diff
@@ -76,13 +76,37 @@ void ReplicationStorageClient::UpdateReplicaState(Storage *storage, DatabaseAcce
...
+          "replication on a clean instance.",
+          client_name, client_name, client_name);
+    };
+#ifdef MG_ENTERPRISE
+    if (!FLAGS_coordinator_server_port) {
+      log_error();
+      return;
+    }
+    replica_state_.WithLock([&](auto &state) {
+      state = replication::ReplicaState::DIVERGED_FROM_MAIN;
```
antoniofilipovic, line 94: Duplicated from line 79

antoniofilipovic: Actually change to RECOVERY, double check

src/dbms/inmemory/replication_handlers.cpp
antoniofilipovic: I think there is no need as we are only transaction due to stream and we have `main_lock_`

src/storage/v2/replication/replication_client.cpp
antoniofilipovic: Not changed to recovery, but changed logic

src/storage/v2/replication/replication_client.cpp
antoniofilipovic: It is already

src/dbms/inmemory/replication_handlers.cpp
Ignition: Transactions and database access are different levels of concern.
I think storage/transaction level is correct here.
But its problematic that an RPC call can be held up forever by any other current transactions.

antoniofilipovic: Yes, you are right, makes sense

src/storage/v2/durability/durability.cpp
```diff
@@ -437,17 +436,13 @@ std::optional<RecoveryInfo> Recovery::RecoverData(std::string *uuid, Replication
...
-          epoch_history->emplace_back(std::string(wal_file.epoch_id), last_loaded_timestamp_value);
-        }
-
-        auto last_epoch_updated = !epoch_history_empty && epoch_history->back().first == wal_file.epoch_id &&
-                                  epoch_history->back().second < last_loaded_timestamp_value;
-        if (last_epoch_updated) {
+        if (epoch_history->empty() || epoch_history->back().first != wal_file.epoch_id) {
+          // no history or new epoch, add it
+          epoch_history->emplace_back(wal_file.epoch_id, last_loaded_timestamp_value);
+          repl_storage_state.epoch_.SetEpoch(wal_file.epoch_id);
```
Ignition, line 443: ??

src/storage/v2/replication/replication_client.cpp
Ignition: this is already in a task

src/storage/v2/durability/durability.cpp
antoniofilipovic: That is from last PR, your suggestion: https://github.com/memgraph/memgraph/pull/1743#discussion_r1508861787

src/dbms/inmemory/replication_handlers.cpp
antoniofilipovic: I think it is a general issue we have, with snapshots also, so we need to deal with both cases in different way. I would leave it as TODO in case it is the real issue

antoniofilipovic:
> General question about the scheme; why do we care about epochs? Shouldn't everyone have the same history? If the epoch do not align, then we have a branching point and the two instances cannot continue as part of the same cluster. I understand that we might decide to delete data in that case. But I still do not understand why we care about epoch even in this case. We care about who has the highest timestamp. In case we have a branching point we might care about minimizing the differences. Not sure about this either. We should prioritize deleting the least amount of data. Timestamps don't exactly correlate with "amount of data" but should be enough at this point.

We synced about this. So the issue was that the Notion page was out of sync. I don't deal with epochs here.

antoniofilipovic:
@kgolubic tagging you here:
RN: In case REPLICA diverges from MAIN, MAIN will force reset data on REPLICA instance

antoniofilipovic:
Docs here:
https://github.com/memgraph/documentation/pull/630

Issue #1544, opened 2023-11-26
Call without return fails


Josipmrden:
**Memgraph version**
v2.12

**Describe the bug**
Query fails although it passes with Neo4j. Possible an error with call and return -> return after call needs to be allowed
CREATE (z:Zone {customerId: 1, id: 1});
MATCH (z:Zone {customerId: 1, id: 1})
WITH z
UNWIND range(1,1000) as id
CALL {
    WITH id, z
    CREATE (r:AWSResource {customerId: 1, name: toString(id), category: toString(id % 10), hash: "someHash"+toString(id), type: "EC2"})
    CREATE (r)-[:IN]->(z)
};

PR #1784, opened 2024-03-04
Allow query with a write subquery to omit RETURN

andrejtonev:
We already allow a write query to omit RETURN.
However, queries like
```
CREATE (a:A {id: 1});
MATCH (a:A {id: 1})
WITH a
UNWIND range(1,1000) as id
CALL {
    WITH id, a
    CREATE (b:B {prop: 1})
    CREATE (b)-[:IN]->(z)
};
```
would be categorized as read, because the subquery's (query under CALL) read/write status would be ignored.
Now a query has updates if any one of the subqueries has updates.

closes https://github.com/memgraph/memgraph/issues/1544

[master < Task] PR

### DIFF
--- a/tests/gql_behave/tests/memgraph_V1/features/create.feature
+++ b/tests/gql_behave/tests/memgraph_V1/features/create.feature
@@ -175,3 +175,15 @@ Feature: Create
             CREATE (a:A) CREATE (a:B)
             """
         Then an error should be raised
+
+    Scenario: CREATE via CALL:
+        When executing query:
+            """
+            UNWIND range(1,10) as id
+            CALL {
+                WITH id
+                CREATE (b:B {prop: 1})
+                CREATE (b)-[:IN]->(z)
+            };
+            """
+        Then the result should be empty
--- a/src/query/frontend/ast/ast.hpp
+++ b/src/query/frontend/ast/ast.hpp
@@ -2026,6 +2026,7 @@ class SingleQuery : public memgraph::query::Tree, public utils::Visitable<Hierar
   }
 
   std::vector<memgraph::query::Clause *> clauses_;
+  bool has_update{};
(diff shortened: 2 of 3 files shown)


andrejtonev:
@kgolubic 
Changelog: Queries with sub-queries are correctly identified as READ/WRITE by examining the sub-query as well.

tests/gql_behave/tests/memgraph_V1/features/create.feature
```diff
@@ -175,3 +175,15 @@ Feature: Create
             CREATE (a:A) CREATE (a:B)
             """
         Then an error should be raised
+
+    Scenario: CREATE via CALL:
```
Josipmrden, line 179: just curious, what was the output before in this query?

Josipmrden (approved):

tests/gql_behave/tests/memgraph_V1/features/create.feature
Josipmrden: oh ok an exception is thrown that a query should return results.

andrejtonev: Yeah, we were qualifying this whole thing as READ
```
CREATE (a:A {id: 1});
MATCH (a:A {id: 1})
WITH a
UNWIND range(1,1000) as id
CALL {
    WITH id, a
    CREATE (b:B {prop: 1})
    CREATE (b)-[:IN]->(z)
};
```
even though the 
```
    WITH id, a
    CREATE (b:B {prop: 1})
    CREATE (b)-[:IN]->(z)
```
was correctly categories as a WRITE.
We were simply ignoring the subcall. Now we save the subcall's READ/WRITE status and apply to the higher level query.

Issue #1625, opened 2024-01-09
Split function causes OutOfMemory bug 


antejavor:
**Memgraph version**
Memgraph v2.13

**Environment**
Docker 

**Describe the bug**
Running a split function on a string with an empty argument should probably separate the string into characters. 
In current implementations, it eats all the memory. 

**To Reproduce**
RETURN split('abc','') as outOfMemoryBug

**Expected behavior**
List of a,b,c strings.  

**Logs**

Memory limit exceeded! Attempting to allocate a chunk of 5.00GiB which would put the current use to 7.55GiB, while the maximum allowed size for allocation is set to 7.34GiB.

katarinasupe:
@antejavor is this duplicated? https://github.com/memgraph/memgraph/issues/609

andrejtonev:
> @antejavor is this duplicated? https://github.com/memgraph/memgraph/issues/609

@katarinasupe, yes :smile:

andrejtonev:
This is easily fixed by changing the split logic. It doesn't support empty delimiter and just runs forever.
I fixed the issue, but I'm not 100% sure I like allowing `split("abc", "")` with an empty delimiter. 
I would rather have two functions; one `split("abd", "not empty") ` and another one `split("abc")` that just splits every character.

andrejtonev:
closes https://github.com/memgraph/memgraph/issues/609

Issue #609, opened 2022-10-20
[BUG] split() with an empty delimiter causes the server to crash


weelink:
**Memgraph version**
Which version did you use?

Docker image: memgraph/memgraph-platform:2.5.0-memgraph2.4.1-lab2.2.1-mage1.3.1

**Environment**
Some information about the environment you are using Memgraph on: operating
system, how do you connect, with or without docker, which driver etc.

- Docker compose
- Ubuntu 22.04.1 LTS
- Running Memgraph Lab in Firefox

**Describe the bug**
Trying to convert a string to a list containing the individual characters causes the server to crash (e.g. the string "abcd" should become ["a", "b", "c", "d"]). I used the `split()` function to do this, and the delimiter is an empty string.

Running it with a delimiter is fine: `RETURN split("a,b,c", ",")` produces the correct output.

**To Reproduce**
1. Run the following query: RETURN split("abc", "")
2. Click on 'Run Query'

**Expected behavior**
The result of  `[List] 3 items` to be outputted in the Data results tab.

**Logs**
[2022-10-20 07:42:50.693][Debug][Run] 'RETURN split("abc", "")'
[2022-10-20 07:42:52.940][Info]Accepted a connection from Bolt: 127.0.0.1:59870
[2022-10-20 07:42:52.940][Info]Using version 4.3 of protocol
[2022-10-20 07:42:52.940][Info]Client connected 'neo4j-javascript/4.4.6'
[2022-10-20 07:42:58.857][Warning]Running out of available RAM, only 131 MB left. For more details, visit https://memgr.ph/ram.
[2022-10-20 07:43:41.040] ~ Trying to reconnect to Memgraph logs for the first time at 07:43:41...
[2022-10-20 07:43:41.543] ~ Connecting to Memgraph logs...
[2022-10-20 07:43:41.544] ~ Trying to reconnect to Memgraph logs for the second time at 07:43:42...
[2022-10-20 07:43:42.545] ~ Connecting to Memgraph logs...
[2022-10-20 07:43:42.551] ~ Trying to reconnect to Memgraph logs for the third time at 07:43:44...
[2022-10-20 07:43:44.553] ~ Connecting to Memgraph logs...
[2022-10-20 07:43:44.555] ~ Successfully connected to Memgraph logs. If there are no new logs please check your --log-level settings.

gitbuda:
@weelink thanks very much for a nice bug report! We'll investigate and fix 😄

gitbuda:
[...]

PR #1785, opened 2024-03-04
Split now supports empty delimiter

andrejtonev:
Split was not checking the delimiter's size and would run forever when delimiter was empty.
Now, in case of an empty delimiter, we just split all characters.

closes https://github.com/memgraph/memgraph/issues/1625 
closes https://github.com/memgraph/memgraph/issues/609

[master < Task] PR


antepusic (approved):

andrejtonev:
@kgolubic 
Changelog: Updated split function to handle empty delimiter. In case we have an empty delimiter, we extract each individual character.

PR #1792, opened 2024-03-06
Fix bug prone add server to cluster behavior

antoniofilipovic:
nuRaft has a potential issue when adding a server to the cluster. The request may be accepted but not yet committed, so the next command when it wants to be executed will fail. 

This PR fixes behavior.

Check docs for more info: https://github.dev/eBay/NuRaft
[master < Task] PR

### DIFF
--- a/src/coordination/raft_state.cpp
+++ b/src/coordination/raft_state.cpp
@@ -11,10 +11,11 @@
 
 #ifdef MG_ENTERPRISE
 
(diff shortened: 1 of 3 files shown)


as51340 (approved):

antoniofilipovic:
@kgolubic RN:
- When adding adding coordinators in cluster, action will fully succeed or fail before coordinators returns result from query

PR #1801, opened 2024-03-08
Pmr allocator unify

Ignition:
Query allocator and evaluation allocator were different.
After analysis, was determined they should be the same, this will help future development reduce TypeValue copies during queries. 

Changes:
- Common allocator, PoolResource backed by MonotonicResource
- Optimized Pool, now O(1) alloc/dealloc as all chunks in Pool form a single free list
- 2nd PoolResource, using bin sizing, not as perfect for memory usage but O(1) bin selection
- Now have jemalloc's background thread to make sure decay and return to OS happens
- Less temporary memory allocations
  - CSV reader now maintains a common line buffer it reuses on line reads
  - Writing out bolt values, now reuses a values buffer
  - Evaluating an int no longer makes temporary strings for errors it most likely never throws
  - ExpandVariable will reuse existing edge list in frame it one existed
- Optimized ProperyValue to be faster at destruction/copy/move

related to #1685


antoniofilipovic:
@Ignition is this ready for review?

src/query/interpreter.cpp
```diff
@@ -1910,7 +1851,7 @@ PreparedQuery Interpreter::PrepareTransactionQuery(std::string_view query_upper,
   std::function<void()> handler;
 
   if (query_upper == "BEGIN") {
-    ResetInterpreter();
+    //    ResetInterpreter();
```
antoniofilipovic, line 1854: Why is this commented out?

antoniofilipovic, line 1854: Why is this commented out?

antoniofilipovic: I think it can be removed, as you have that part of code down below

src/query/interpreter.hpp
```diff
@@ -65,6 +65,54 @@ extern const Event SuccessfulQuery;
...
+  static auto upstream_resource() -> utils::MemoryResource * {
+    // singleton ResourceWithOutOfMemoryException
+    // explicitly backed by NewDeleteResource
+    static auto upstream = utils::ResourceWithOutOfMemoryException{utils::NewDeleteResource()};
+    return &upstream;
+  }
+
+#ifndef MG_MEMORY_PROFILE
+  memgraph::utils::MonotonicBufferResource monotonic{kMonotonicInitialSize, upstream_resource()};
+  memgraph::utils::PoolResource pool{kPoolBlockPerChunk, &monotonic, upstream_resource()};
```
antoniofilipovic, line 112: I am not sure about using `ResourceWithOutOfMemoryException`. Here is my understanding. So, in the operator tree and in other places where we need OOM to happen, we have already integrated OOMExceptionEnabler. Now, for every allocation we will have that enabled and I think we don't need that. I would remove `ResourceWithOutOfMemory` for `monotonic` and `pool` since we will use `pool` and `monotonic` in the operator tree and there we have enabled exceptions. Also for mg_procedure we have already enabled OOMException so we don't need even for that case ResourceWithOutOfMemoryException

src/query/interpreter.hpp
```diff
@@ -65,6 +65,54 @@ extern const Event SuccessfulQuery;
...
+  }
+  auto resource_without_pool_or_mono() -> utils::MemoryResource * { return upstream_resource(); }
+
+ private:
+  // At least one page to ensure not sharing page with other subsystems
+  static constexpr auto kMonotonicInitialSize = 4UL * 1024UL;
+  // TODO: need to profile to check for good defaults, also maybe PoolResource
+  //  needs to be smarter. We expect more reuse of smaller objects than larger
+  //  objects. 64*1024B is maybe wasteful, whereas 256*32B maybe sensible.
+  //  Depends on number of small objects expected.
```
antoniofilipovic, line 99: Did you already profile that?

src/query/interpreter.hpp
```diff
@@ -304,46 +352,30 @@ class Interpreter final {
...
     void CleanRuntimeData() {
-      if (prepared_query.has_value()) {
-        prepared_query.reset();
-      }
+      // Called from Commit/Abort once query has been fully used
+
+      prepared_query.reset();
       notifications.clear();
+      // TODO: double check is summary still needed here
+      //  can we dispose of it and also execution_memory at this point?
```
antoniofilipovic, line 378: Is this double checked?

src/query/interpreter.hpp
```diff
@@ -304,46 +352,30 @@ class Interpreter final {
...
-      prepared_query.reset();
-      std::visit([](auto &memory_resource) { memory_resource.Release(); }, execution_memory);
-    }
+    ~QueryExecution() = default;
 
     void CleanRuntimeData() {
-      if (prepared_query.has_value()) {
-        prepared_query.reset();
-      }
+      // Called from Commit/Abort once query has been fully used
```
antoniofilipovic, line 373: I would remove this comment, I am not sure we need it here?

src/storage/v2/property_value.hpp
```diff
@@ -403,38 +419,88 @@ inline PropertyValue::PropertyValue(PropertyValue &&other) noexcept : type_(std:
...
+        std::destroy_at(&list_v.val_);
+        break;
+      case Type::Map:
+        std::destroy_at(&map_v.val_);
+        break;
+      case Type::TemporalData:
+        break;
+    }
+    // construct
+    auto *new_this = std::launder(this);
```
antoniofilipovic, line 474: This is my lack of knowledge, but I am not sure I understand why you need std::launder here, if you can explain I would appreciate :smile:

antoniofilipovic (dismissed):
Great work, really! I have a few changes I think we need to align on, and I want to get into bit manipulation la ittle bit more later today.

tests/unit/utils_memory.cpp
```diff
@@ -194,134 +194,134 @@ TEST(MonotonicBufferResource, AllocationWithInitialBufferOnStack) {
...
-  EXPECT_EQ(ptr, mem.Allocate(max_block_size));
-  EXPECT_EQ(test_mem.new_count_, 0U);
-}
+//
+//// NOLINTNEXTLINE(hicpp-special-member-functions)
+// TEST(PoolResource, SingleSmallBlockAllocations) {
+//   TestMemory test_mem;
+//   const size_t max_blocks_per_chunk = 3U;
+//   const size_t max_block_size = 64U;
+//   memgraph::utils::PoolResource mem(max_blocks_per_chunk, max_block_size, &test_mem);
```
antoniofilipovic, line 203: I would rather delete these tests as we are not using them anymore. If we will restore them, we will find your commit

src/query/plan/rule_based_planner.cpp
```diff
@@ -629,7 +629,9 @@ std::unique_ptr<LogicalOperator> GenReturn(Return &ret, std::unique_ptr<LogicalO
   // SET n.prop = n.prop + 1 RETURN n.prop`. If we match same `n` multiple 'k'
   // times, we want to return 'k' results where the property value is the same,
   // final result of 'k' increments.
-  bool accumulate = is_write;
+  // Exception: if we know k == 1 then no need to accumulate
+  //            ie. If input_op is the CreateNode that produced it.
+  bool accumulate = is_write && !utils::Downcast<CreateNode>(&*input_op);
```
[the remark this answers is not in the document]
Ignition: I'll not deliver this change...not sure of correctness or not. But it makes almost no difference in mgbench so not worth risking introducing correctness bugs.

src/query/interpreter.hpp
Ignition: This was there to be consistent with how the Pull allocator worked. I think you are right. But I'll deliver as is in this PR and we can remove the `ResourceWithOutOfMemoryException` in a future PR.

src/query/interpreter.hpp
Ignition: No, but getting the right pool sizes is problematic anyhow...I don't think tuning it now makes sense.

Future work once we fix cursors and evaluation to have better allocation patterns, then we could profile the size usage and use that histogram to dictate the size of chunk per size category.

gitbuda (approved):
🚀

kgolubic:
@Ignition can you please provide me with a Release note entry? Is this a feature or bugfix?

Ignition:
> @Ignition can you please provide me with a Release note entry? Is this a feature or bugfix?

@kgolubic 
BUGFIX: Query allocations now use a single allocation subsystem rather than two. 
BUGFIX: Memory no longer horded by bolt workers.

PR #1804, opened 2024-03-09
Isolate cursor.hpp

gitbuda:
Not the whole `operator.hpp` is required if we want to introduce a new cursor.

### DIFF
--- a/src/query/plan/cursor.hpp
+++ b/src/query/plan/cursor.hpp
@@ -1,0 +1,80 @@
+// Copyright 2024 Memgraph Ltd.
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
+// License, and you may not use this file except in compliance with the Business Source License.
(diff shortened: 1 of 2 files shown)


Ignition:
ATM all the `*Operator::MakeCursor`'s are in operator.hpp
What is the motivation to split this part out? Are Operators doing to be split up as well soon?

andrejtonev:
Are there any files that should change the file included?

gitbuda:
@Ignition I'm experimenting under https://github.com/gitbuda/memgraph/pull/12, each cursor there has a standalone cpp file (ofc if possible) -> this was just the smallest chunk of work which seems to be useful 🤔 
@andrejtonev -> Including `cursor.hpp` is not needed since `operator.hpp` and everything else includes the operator. When and if we decouple more, additional includes might get required 👀

hal-eisen-MG:
Can we close this?

gitbuda:
I think useful, but got in conflict + it's easy to redo if needed -> closing this PR

PR #1811, opened 2024-03-12
Fix timestamps saving on-disk

### LINKED ISSUE #1799: Durability issue with on-disk storage mode
**Memgraph version**
2.15

**Describe the bug**
Wrong results when while matching all nodes in database, using on-disk storage mode.
Screenshot from 2024-03-08 14-39-00

NOTE: This data existed in database before restart

DavIvek:
The problem is that Memgraph saves the current timestamp only in the DiskStorage destructor, but if the process gets killed with the SIGKILL signal, for example, then the destructor won't be called and the timestamp won't be saved (in short, timestamp tells us which database state we can access).
Because of that, the timestamp will start from 1 and old keys will be used, hence user will have to query the database until he gets to the timestamp he used before the database was restarted.
To fix this, we can save timestamp on every commit.

[master < Task] PR

Release note: If the database crashes, after the restart it will be in the same state as before the crash (on-disk storage mode)

closes #1799


as51340 (changes requested):
Can you try to add test? I think with unit test could be quite easy

src/storage/v2/disk/storage.cpp
```diff
@@ -1768,7 +1768,8 @@ utils::BasicResult<StorageManipulationError, void> DiskStorage::DiskAccessor::Co
     return StorageManipulationError{SerializationError{}};
   }
   spdlog::trace("rocksdb: Commit successful");
-
+  disk_storage->durable_metadata_.SaveBeforeClosingDB(disk_storage->timestamp_, disk_storage->vertex_count_,
+                                                      disk_storage->edge_count_);  // maybe rename this function
```
Darych, line 1772: This comment will be lost. Let's rename it or live with it :)
But do we close DB inside? I don't think so, also not in this function, so maybe call it differently? What about `UpdateMetadata`?

as51340 (approved):

PR #1827, opened 2024-03-16
Implement execution part of one pattern comprehension in Return clause.

Darych:
### Description
Pattern comprehension with filters works in Return statement.

### Documentation checklist
    - **[Release note text]**
    - **[Documentation PR link]**


src/query/interpret/eval.hpp
```diff
@@ -1105,8 +1105,15 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
     }
   }
 
-  TypedValue Visit(PatternComprehension & /*pattern_comprehension*/) override {
-    throw utils::NotYetImplemented("Expression evaluator can not handle pattern comprehension.");
+  TypedValue Visit(PatternComprehension &pattern_comprehension) override {
+    TypedValue &frame_pattern_comprehension_value = frame_->at(symbol_table_->at(pattern_comprehension));
```
antoniofilipovic, line 1109: I think you can put const here, and it will then be more obvious later we are returning copy of that value
```suggestion
    TypedValue const &frame_pattern_comprehension_value = frame_->at(symbol_table_->at(pattern_comprehension));
```

src/query/plan/operator.cpp
```diff
@@ -5770,17 +5771,20 @@ UniqueCursorPtr HashJoin::MakeCursor(utils::MemoryResource *mem) const {
   return MakeUniqueCursorPtr<HashJoinCursor>(mem, *this, mem);
 }
 
-RollUpApply::RollUpApply(const std::shared_ptr<LogicalOperator> &input,
-                         std::shared_ptr<LogicalOperator> &&second_branch)
-    : input_(input), list_collection_branch_(second_branch) {}
+RollUpApply::RollUpApply(const std::shared_ptr<LogicalOperator> &&input,
```
antoniofilipovic, line 5774: I am pretty confident you need to remove here `const` to capture by rvalue reference to be able to move it afterwards, and then you can std::move value into intput_
```suggestion
RollUpApply::RollUpApply(std::shared_ptr<LogicalOperator> &&input,
```

src/query/plan/operator.cpp
```diff
@@ -5770,17 +5771,20 @@ UniqueCursorPtr HashJoin::MakeCursor(utils::MemoryResource *mem) const {
...
+                         std::shared_ptr<LogicalOperator> &&list_collection_branch,
+                         const std::vector<Symbol> &list_collection_symbols, Symbol result_symbol)
+    : input_(input),
+      list_collection_branch_(list_collection_branch),
+      list_collection_symbols_(list_collection_symbols),
+      result_symbol_(result_symbol) {}
```
antoniofilipovic, line 5780: Since you are making copy of `result_symbol` in input parameter of constructor, you can use std::move for move construction of object here

antoniofilipovic (commented):
Only a couple of comments. I won't request changes as I am OOO from tomorrow not to block your merge, but please respond/fix before merge

src/query/plan/operator.cpp
Darych: Good catch! Thank you!

src/query/plan/operator.cpp
```diff
@@ -5791,4 +5795,71 @@ bool RollUpApply::Accept(HierarchicalLogicalOperatorVisitor &visitor) {
...
+          context.frame_change_collector->ResetTrackingValue(symbol.name());
+        }
+      }
+    };
+
+    TypedValue result(std::vector<TypedValue>(), context.evaluation_context.memory);
+    if (input_cursor_->Pull(frame, context)) {
+      while (list_collection_cursor_->Pull(frame, context)) {
+        // collect values from the list collection branch
+        for (const auto &output_symbol : self_.list_collection_symbols_) {
```
Josipmrden, line 5827: not sure if list_collection_symbols_ should be a list or just one element?

Darych: It should be just one element. I also thought about this.
The reason why I put it as a vector because `ModifiedSymbols` function returns `vector<Symbol>`.
But inside the `RollupApply` operator I think we can validate size and store just one symbol to not confuse other developers.

Josipmrden: is it fine that we do an `MG_ASSERT` somewhere during query planning?

Josipmrden (approved):

antoniofilipovic (approved):
Good work :muscle:

kgolubic:
@Darych can you give me a Release note entry and docs PR link?

PR #1836, opened 2024-03-19
HA: Add force reset of cluster

antoniofilipovic:
### Description

This PR introduces a force reset of cluster state to state that was before the operation started.

_Implemented_
_Enable force reset to work on:_
- [x] coordinator becoming a leader in an unhealthy state
- [x] failover fail
- [x] register instance fail
- [x] unregister instance fail
- [x]  promote instance to main fail
- [x] force reset fail

_Additional work:_
- [x] Add test when we enter force reset
- [x] Fix bug on not demoting first all instances to replicas
- [x] Take care of instances which were down and couldn't be demoted to replica, but are in unknown state
- [x] Add test when force reset needs to happen
- [x] Add test when main dies before force reset needs to happen
- [x] Add test when replica dies before force reset needs to happen
- [x] Add test when few coordinators die when force reset is happening

[master < Task] PR
    Add force reset of cluster in HA
    - When the action doesn't succeed atomically cluster enters a force reset state. 
    - For instances that are down introduce demote to replica callback.
    - Use unique lock and unlock raft log to change state
    - Add action for instance which needs to be demoted to replica

### Documentation checklist
    - **[Release note text]**
    - **[Documentation PR link]**

### DIFF
--- a/src/coordination/coordinator_instance.cpp
+++ b/src/coordination/coordinator_instance.cpp
@@ -17,7 +17,6 @@
 #include "dbms/constants.hpp"
 #include "nuraft/coordinator_state_machine.hpp"
 #include "nuraft/coordinator_state_manager.hpp"
-#include "utils/counter.hpp"
 #include "utils/functional.hpp"
 #include "utils/resource_lock.hpp"
 
@@ -183,17 +182,19 @@ void CoordinatorInstance::ForceResetCluster() {
     return;
   }
 
-  utils::OnScopeExit do_another_reset{[this]() {
-    if (raft_state_.IsLockOpened()) {
-      thread_pool_.AddTask([this]() { this->ForceResetCluster(); });
-    }
-  }};
-
-  if (!raft_state_.AppendOpenLockForceReset()) {
-    MG_ASSERT(!raft_state_.IsLeader(), "Coordinator is not follower, something is wrong.");
+  if (!raft_state_.AppendOpenLock()) {
     spdlog::trace("Appending log force reset failed, aborting force reset");
     return;
   }
+
+  utils::OnScopeExit maybe_do_another_reset{[this]() {
+    if (raft_state_.IsLockOpened() && raft_state_.IsLeader()) {
+      spdlog::trace("Adding task to try force reset cluster again.");
+      thread_pool_.AddTask([this]() { this->ForceResetCluster(); });
+      return;
+    }
+    spdlog::trace("Lock is not opened anymore or coordinator is not leader, not doing force reset again.");
+  }};
 
   auto const instances = raft_state_.GetReplicationInstances();
 
@@ -209,44 +210,35 @@ void CoordinatorInstance::ForceResetCluster() {
   });
 
   auto const new_uuid = utils::UUID{};
-  auto const append_log_update_uuid_failed = [this, &new_uuid](ReplicationInstance &repl_instance) {
-    if (!repl_instance.SendSwapAndUpdateUUID(new_uuid)) {
-      // We assume this cannot fail as we filtered out instances which are not alive...
-      return true;
-    }
-    if (!raft_state_.AppendUpdateUUIDForInstanceLog(repl_instance.InstanceName(), new_uuid)) {
-      // error is only if we swap uuid but append log doesn't succeed
-      return true;
-    }
-    return false;
-  };
-
-  for (auto const &repl_instance : repl_instances_ | ranges::views::filter([&new_uuid](ReplicationInstance &instance) {
-                                     instance.OnSuccessPing();
-                                     return instance.SendSwapAndUpdateUUID(new_uuid);
-                                   })) {
-    if (!raft_state_.AppendUpdateUUIDForInstanceLog(repl_instance.InstanceName(), new_uuid)) {
+  auto const update_on_success_ping_on_swap_uuid = [&new_uuid](ReplicationInstance &instance) {
+    if (!instance.SendSwapAndUpdateUUID(new_uuid)) {
       return;
     }
-  }
+    instance.OnSuccessPing();
+  };
+  std::ranges::for_each(repl_instances_, update_on_success_ping_on_swap_uuid);
 
-  if (std::ranges::any_of(repl_instances_, append_log_update_uuid_failed)) {
-    spdlog::error("Force reset failed since update log or swap uuid failed, assuming coordinator is now follower.");
-    MG_ASSERT(!raft_state_.IsLeader(), "Coordinator is not follower");
-    spdlog::trace("Adding task to try force reset cluster again.");
-    thread_pool_.AddTask([this]() { this->ForceResetCluster(); });
+  auto const is_swap_uuid_success = [](ReplicationInstance &instance) { return instance.IsAlive(); };
+
+  auto instances_with_swapped_uuid = repl_instances_ | ranges::views::filter(is_swap_uuid_success);
+
+  auto update_uuid_failed = [&new_uuid, this](auto &repl_instance) {
+    return !raft_state_.AppendUpdateUUIDForInstanceLog(repl_instance.InstanceName(), new_uuid);
+  };
+  if (std::ranges::any_of(instances_with_swapped_uuid, update_uuid_failed)) {
+    spdlog::error("Force reset failed since update log swap uuid failed, assuming coordinator is now follower.");
     return;
   }
 
   if (!raft_state_.AppendUpdateUUIDForNewMainLog(new_uuid)) {
     spdlog::error("Update log for new MAIN failed, assuming coordinator is now follower");
-    MG_ASSERT(!raft_state_.IsLeader(), "Coordinator is not follower");
     return;
   }
-
+  // TODO(antoniofilipovic) Think how we can isolate this part in common function
   auto const get_ts = [](ReplicationInstance &replica) { return replica.GetClient().SendGetInstanceTimestampsRpc(); };
 
-  auto maybe_instance_db_histories = repl_instances_ | ranges::views::transform(get_ts) | ranges::to<std::vector>();
+  auto maybe_instance_db_histories =
+      instances_with_swapped_uuid | ranges::views::transform(get_ts) | ranges::to<std::vector>();
 
   auto const ts_has_value = [](auto const &zipped) -> bool {
     auto &[replica, res] = zipped;
@@ -258,7 +250,7 @@ void CoordinatorInstance::ForceResetCluster() {
     return std::make_pair(replica.InstanceName(), res.GetValue());
   });
 
-  auto instance_db_histories = ranges::views::zip(repl_instances_, maybe_instance_db_histories) |
+  auto instance_db_histories = ranges::views::zip(instances_with_swapped_uuid, maybe_instance_db_histories) |
                                ranges::views::filter(ts_has_value) | transform_to_pairs | ranges::to<std::vector>();
 
   auto [most_up_to_date_instance, latest_epoch, latest_commit_timestamp] =
@@ -269,28 +261,23 @@ void CoordinatorInstance::ForceResetCluster() {
   auto const is_not_new_main = [&new_main](ReplicationInstance const &repl_instance) {
     return repl_instance.InstanceName() != new_main.InstanceName();
   };
-  auto repl_clients_info = repl_instances_ | ranges::views::filter(is_not_new_main) |
+  auto repl_clients_info = instances_with_swapped_uuid | ranges::views::filter(is_not_new_main) |
                            ranges::views::transform(&ReplicationInstance::ReplicationClientInfo) |
                            ranges::to<ReplicationClientsInfo>();
 
   if (!new_main.PromoteToMain(new_uuid, std::move(repl_clients_info), &CoordinatorInstance::MainSuccessCallback,
                               &CoordinatorInstance::MainFailCallback)) {
-    spdlog::warn("Force reset failed since promoting replica to main failed, trying again!");
-    thread_pool_.AddTask([this]() { this->ForceResetCluster(); });
-    spdlog::trace("Adding task to try force reset cluster again.");
+    spdlog::warn("Force reset failed since promoting replica to main failed.");
     return;
   }
 
-  // TODO(antoniofilipovic) this will set it in healthy state -> check if we need something else
+  // This will set cluster in healthy state again
   if (!raft_state_.AppendSetInstanceAsMainLog(most_up_to_date_instance, new_uuid)) {
-    spdlog::error("Update log for new MAIN failed, assuming coordinator is now follower");
-    MG_ASSERT(!raft_state_.IsLeader(), "Coordinator is not follower");
-    thread_pool_.AddTask([this]() { this->ForceResetCluster(); });
-    spdlog::trace("Adding task to try force reset cluster again.");
+    spdlog::error("Update log for new MAIN failed");
     return;
   }
 
-  MG_ASSERT(raft_state_.IsHealthy(), "After force reset we need to be in healthy state.");
+  MG_ASSERT(!raft_state_.IsLockOpened(), "After force reset we need to be in healthy state.");
 }
 
 auto CoordinatorInstance::TryFailover() -> void {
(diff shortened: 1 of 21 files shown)


src/coordination/coordinator_instance.cpp
```diff
@@ -68,6 +72,7 @@ CoordinatorInstance::CoordinatorInstance()
                 repl_instance.StopFrequentCheck();
               }
               auto lock = std::unique_lock{coord_instance_lock_};
+              std::ranges::for_each(repl_instances_, [](auto &repl_instance) { repl_instance.StopFrequentCheck(); });
               repl_instances_.clear();
               spdlog::info("Stopped all replication instance frequent checks.");
```
as51340, line 77: I think we should put trace log-level logs

src/coordination/coordinator_instance.cpp
```diff
@@ -33,6 +32,11 @@ CoordinatorInstance::CoordinatorInstance()
     : thread_pool_{1},
       raft_state_(RaftState::MakeRaftState(
           [this]() {
+            if (raft_state_.IsLockOpened()) {
+              spdlog::error("Leader hasn't encountered healthy state, doing force reset of cluster.");
+              thread_pool_.AddTask([this]() { this->ForceResetCluster(); });
```
as51340, line 37: will thread from the thread pool start immediately?

src/coordination/coordinator_instance.cpp
```diff
@@ -199,10 +320,17 @@ auto CoordinatorInstance::TryFailover() -> void {
 
   auto *new_main = &FindReplicationInstance(most_up_to_date_instance);
 
-  if (!raft_state_.AppendOpenLockFailover(most_up_to_date_instance)) {
+  if (!raft_state_.AppendOpenLock()) {
     spdlog::error("Aborting failover as instance is not anymore leader.");
     return;
   }
+
+  utils::OnScopeExit do_reset{[this]() {
```
as51340, line 328: Merge with other OnScopeExit

src/coordination/coordinator_state_machine.cpp
```diff
@@ -38,21 +38,8 @@ auto CoordinatorStateMachine::CreateLog(nlohmann::json &&log) -> ptr<buffer> {
...
-}
-
-auto CoordinatorStateMachine::SerializeOpenLockFailover(std::string_view instance_name) -> ptr<buffer> {
-  return CreateLog({{"action", RaftLogAction::OPEN_LOCK_FAILOVER}, {"info", std::string(instance_name)}});
-}
-
-auto CoordinatorStateMachine::SerializeOpenLockSetInstanceAsMain(std::string_view instance_name) -> ptr<buffer> {
-  return CreateLog({{"action", RaftLogAction::OPEN_LOCK_SET_INSTANCE_AS_MAIN}, {"info", std::string(instance_name)}});
+auto CoordinatorStateMachine::SerializeOpenLock() -> ptr<buffer> {
+  return CreateLog({{"action", RaftLogAction::OPEN_LOCK}, {"info", nullptr}});
```
as51340, line 42: no need to send {"info", nullptr}

src/coordination/include/nuraft/coordinator_cluster_state.hpp
```diff
@@ -61,8 +61,8 @@ struct CoordinatorInstanceState {
 void to_json(nlohmann::json &j, ReplicationInstanceState const &instance_state);
 void from_json(nlohmann::json const &j, ReplicationInstanceState &instance_state);
 
-using TRaftLog = std::variant<CoordinatorToReplicaConfig, std::string, utils::UUID, CoordinatorToCoordinatorConfig,
-                              InstanceUUIDUpdate>;
+using TRaftLog = std::variant<std::monostate, CoordinatorToReplicaConfig, std::string, utils::UUID,
```
as51340, line 64: I don't think monostate is good here :joy: try to refactor this

tests/e2e/high_availability/distributed_coords.py
```diff
@@ -1420,5 +1420,135 @@ def get_vertex_count():
...
+    follower_data = [
+        ("coordinator_1", "127.0.0.1:10111", "", "unknown", "coordinator"),
+        ("coordinator_2", "127.0.0.1:10112", "", "unknown", "coordinator"),
+        ("coordinator_3", "127.0.0.1:10113", "", "unknown", "coordinator"),
+        ("instance_1", "", "", "unknown", "main"),
+        ("instance_2", "", "", "unknown", "replica"),
+        ("instance_3", "", "", "unknown", "unknown"),
+    ]
+
+    mg_sleep_and_assert(leader_data, show_instances_coord3)
```
as51340, line 1547: How do you know which one will be leader?

src/coordination/coordinator_instance.cpp
```diff
@@ -159,6 +164,122 @@ auto CoordinatorInstance::ShowInstances() const -> std::vector<InstanceStatus> {
...
+    if (!instance.SendSwapAndUpdateUUID(new_uuid)) {
+      return;
+    }
+    instance.OnSuccessPing();
+  };
+  std::ranges::for_each(repl_instances_, update_on_success_ping_on_swap_uuid);
+
+  auto const is_swap_uuid_success = [](ReplicationInstance &instance) { return instance.IsAlive(); };
+
+  auto instances_with_swapped_uuid = repl_instances_ | ranges::views::filter(is_swap_uuid_success);
```
as51340, line 223: Instead of swapping and then checking which was swapped you can apply general map principle with which you transform each replication instance into boolean based on the return value of SendSwapAndUpdateUUID

src/coordination/coordinator_instance.cpp
```diff
@@ -159,6 +164,122 @@ auto CoordinatorInstance::ShowInstances() const -> std::vector<InstanceStatus> {
...
+
+  auto update_uuid_failed = [&new_uuid, this](auto &repl_instance) {
+    return !raft_state_.AppendUpdateUUIDForInstanceLog(repl_instance.InstanceName(), new_uuid);
+  };
+  if (std::ranges::any_of(instances_with_swapped_uuid, update_uuid_failed)) {
+    spdlog::error("Force reset failed since update log swap uuid failed, assuming coordinator is now follower.");
+    return;
+  }
```
as51340, line 231: Why not do this in the same loop? Seems better to have this atomic.

src/coordination/coordinator_instance.cpp
```diff
@@ -159,6 +164,122 @@ auto CoordinatorInstance::ShowInstances() const -> std::vector<InstanceStatus> {
...
+  auto transform_to_pairs = ranges::views::transform([](auto const &zipped) {
+    auto &[replica, res] = zipped;
+    return std::make_pair(replica.InstanceName(), res.GetValue());
+  });
+
+  auto instance_db_histories = ranges::views::zip(instances_with_swapped_uuid, maybe_instance_db_histories) |
+                               ranges::views::filter(ts_has_value) | transform_to_pairs | ranges::to<std::vector>();
+
+  auto [most_up_to_date_instance, latest_epoch, latest_commit_timestamp] =
+      ChooseMostUpToDateInstance(instance_db_histories);
```
as51340, line 257: This part is repeated I think, abstract

as51340 (changes requested):
Nice work, some small comments :muscle:

src/coordination/coordinator_instance.cpp
antoniofilipovic: I think we don't need an on-scope exit to resume frequent checks anymore so I will delete it. Added comment to explain why I think we don't need it anymore

src/coordination/coordinator_instance.cpp
antoniofilipovic: My reasoning was to get which instance is alive with SwapUUID, but I missed I need to demote each instance to replica so I get join these two actions together, which makes sense absolutely

tests/e2e/high_availability/distributed_coords.py
antoniofilipovic: Since we are registering on coordinator 3 so it requests leadership

src/coordination/coordinator_instance.cpp
antoniofilipovic: Done

src/coordination/coordinator_instance.cpp
antoniofilipovic: Almost immediatelly, as soon as there is conditional variable notify on change

src/coordination/include/nuraft/coordinator_cluster_state.hpp
antoniofilipovic: I am not sure what is wrong with monostate, we can discuss it on Monday but seems to me from cpp ref it is used that way as an empty tag?

src/coordination/coordinator_state_machine.cpp
antoniofilipovic: You are right

src/coordination/coordinator_instance.cpp
antoniofilipovic: I did it now, check if that was what you thought

src/coordination/coordinator_instance.cpp
antoniofilipovic: It may not be primary use, but it can be used that way. I can refactor code, but let's sync first

src/coordination/coordinator_instance.cpp
```diff
@@ -46,16 +50,28 @@ CoordinatorInstance::CoordinatorInstance()
...
+            std::ranges::for_each(main_instances, [this](auto &main_instance) {
               spdlog::info("Started pinging main instance {}", main_instance.config.instance_name);
               repl_instances_.emplace_back(this, main_instance.config, client_succ_cb_, client_fail_cb_,
                                            &CoordinatorInstance::MainSuccessCallback,
                                            &CoordinatorInstance::MainFailCallback);
             });
 
+            // In case we got out of force reset but instances weren't still demoted
+            // we need to apply functions to these instances to demote them
+            std::ranges::for_each(instances, [this](ReplicationInstanceState const &replication_instance_state) {
```
as51340, line 66: instead of for_each + if, use views::filter

src/coordination/coordinator_instance.cpp
```diff
@@ -159,6 +175,171 @@ auto CoordinatorInstance::ShowInstances() const -> std::vector<InstanceStatus> {
...
+  repl_instances_.clear();
+
+  if (!raft_state_.IsLeader()) {
+    spdlog::trace("Exiting force reset as coordinator is not any more leader!");
+    return;
+  }
+
+  if (!raft_state_.AppendOpenLock()) {
+    spdlog::trace("Appending log force reset failed, aborting force reset");
+    MG_ASSERT(!raft_state_.IsLeader(), "Coordinator is leader but append open lock failed, encountered wrong state.");
```
as51340, line 210: I don't think we should assert here. Why it's a problem if appending fails and we are leader

src/coordination/coordinator_instance.cpp
```diff
@@ -159,6 +175,171 @@ auto CoordinatorInstance::ShowInstances() const -> std::vector<InstanceStatus> {
...
+  // If at any point later RPC fails for alive instance, we consider this failure
+
+  std::ranges::for_each(instances, [this](auto &replica) {
+    repl_instances_.emplace_back(this, replica.config, client_succ_cb_, client_fail_cb_,
+                                 &CoordinatorInstance::ReplicaSuccessCallback,
+                                 &CoordinatorInstance::ReplicaFailCallback);
+  });
+
+  auto instances_mapped_to_resp = repl_instances_ | ranges::views::transform([](auto &instance) {
+                                    return std::pair{instance.InstanceName(), instance.SendFrequentHeartbeat()};
```
as51340, line 237: We send SendFrequentHeartbeat to check if instance is alive?

src/coordination/coordinator_instance.cpp
```diff
@@ -159,6 +175,171 @@ auto CoordinatorInstance::ShowInstances() const -> std::vector<InstanceStatus> {
...
+  // If instance in raft log is MAIN, it can be REPLICA but raft append failed when we demoted it
+  // If instance in raft log is REPLICA, it can be MAIN but raft log failed when we promoted it
+  // CRUX of problem: We need universal callback which will demote instance to replica and only then change to
+  // REPLICA callbacks
+
+  auto needs_demote_setup_failed = [&instances_mapped_to_resp, this](ReplicationInstance &repl_instance) {
+    if (instances_mapped_to_resp[repl_instance.InstanceName()]) {
+      return false;
+    }
+    if (!raft_state_.AppendInstanceNeedsDemote(repl_instance.InstanceName())) {
```
as51340, line 314: How do we use this append to raft log? Is it necessary

src/coordination/coordinator_instance.cpp
```diff
@@ -549,6 +740,26 @@ void CoordinatorInstance::ReplicaFailCallback(std::string_view repl_instance_nam
...
+void CoordinatorInstance::DemoteSuccessCallback(std::string_view repl_instance_name) {
+  spdlog::trace("Instance {} performing demote to replica successful callback", repl_instance_name);
+
+  auto &repl_instance = FindReplicationInstance(repl_instance_name);
+
+  if (!repl_instance.SendDemoteToReplicaRpc()) {
+    return;
+  }
+
+  if (!raft_state_.AppendSetInstanceAsReplicaLog(repl_instance_name)) {
```
as51340, line 752: Do we need to request leadership? What if this operation fails?

src/coordination/include/coordination/coordinator_rpc.hpp
```diff
@@ -16,6 +16,7 @@
 
 #include "coordination/coordinator_communication_config.hpp"
 #include "replication_coordination_glue/common.hpp"
+#include "replication_coordination_glue/role.hpp"
```
as51340, line 19: no need

src/coordination/include/coordination/replication_instance.hpp
```diff
@@ -23,6 +23,7 @@
 
 #include <libnuraft/nuraft.hxx>
 
+#include <functional>
```
as51340, line 26: no need

as51340 (changes requested):

src/coordination/coordinator_instance.cpp
antoniofilipovic: If we are the leader and we can't append open lock, there is something broken with coordinator right, we should kill it?

src/coordination/coordinator_instance.cpp
antoniofilipovic: Yes

src/coordination/coordinator_instance.cpp
antoniofilipovic: Yes, because at this point force reset can succeed and coordinator can die, so that other coordinator can continue where we left off

src/coordination/coordinator_instance.cpp
antoniofilipovic: Then we are not the leader any more and someone else will continue this action, so I think we must not take leadership here

src/coordination/include/nuraft/coordinator_cluster_state.hpp
```diff
@@ -61,8 +64,8 @@ struct CoordinatorInstanceState {
 void to_json(nlohmann::json &j, ReplicationInstanceState const &instance_state);
 void from_json(nlohmann::json const &j, ReplicationInstanceState &instance_state);
 
-using TRaftLog = std::variant<CoordinatorToReplicaConfig, std::string, utils::UUID, CoordinatorToCoordinatorConfig,
-                              InstanceUUIDUpdate>;
+using TRaftLog = std::variant<std::string, utils::UUID, CoordinatorToReplicaConfig, CoordinatorToCoordinatorConfig,
```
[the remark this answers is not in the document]
antoniofilipovic: I am not sure, let's sync tomorrow on this one

src/coordination/coordinator_instance.cpp
antoniofilipovic: makes sense

src/coordination/include/nuraft/coordinator_cluster_state.hpp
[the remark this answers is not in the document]
antoniofilipovic: I will fix this on some other PR

as51340 (approved):

antoniofilipovic:
@kgolubic resolved on Slack chat, sorry for not pinging earlier

Issue #1768, opened 2024-02-28
Create a flag to automatically generate indices


gvolfing:
Implement a solution to automatically create indices in Memgraph. Since edge-type indices are a lot more rare than label and label+property indices we might want to create separate flags automatically create them.
Question is do we want flags for:
 - Label indices, Label+Property indices, Edge-type indices? (3 separate)
 - (Label indices, Label+Property indices), (Edge-type indices)? (2 separate "traditional indices" + "edge-type")
 - One flag for all
 The last one is probably a bad idea, because most use-cases would likely never require edge-type indices.

gvolfing:
One potential solution might be tied to this issue
https://github.com/memgraph/memgraph/issues/1546

This is not necessarily the way we have to go about this but if we want to detect what indices are currently existing in the database, we either have to check that here, or with the index data-structures.

gitbuda:
A note, only label / edge-type index creation by default is I think an option here, but will expand 😄

Josipmrden:
Done with the PRs linked

PR #1841, opened 2024-03-21
Implement automatic index creation

gvolfing:
### Description

Currently indices can be created by specifying them in form of queries, meaning that the user explicitly has to create them. With this feature 2 flags has been added to Memgraph's configuration options, one to automatically create indices on labels that have been encountered by the database automatically, and one for edge-type indices.

[master < Task] PR
    - **[FINAL GIT MESSAGE]**

### Documentation checklist
    - **[Release note text]**
    - **[Documentation PR link]**


gvolfing:
Closes:
https://github.com/memgraph/memgraph/issues/1768

gvolfing:
docs:
https://github.com/memgraph/documentation/pull/613

src/flags/general.cpp
```diff
@@ -131,6 +131,14 @@ DEFINE_uint64(storage_recovery_thread_count,
 DEFINE_bool(storage_enable_schema_metadata, false,
             "Controls whether metadata should be collected about the resident labels and edge types.");
 
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
+DEFINE_bool(storage_enable_automatic_label_index_creation, false,
+            "Controls whether label indexes on vertices should be created automatically.");
+
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
+DEFINE_bool(storage_enable_automatic_edge_type_index_creation, false,
+            "Controls whether edge-type indexes on relationships should be created automatically.");
```
gitbuda, line 140: let's change the format to be aligned with the rest of the flags (put `enabled` at the end), so e.g. `storage_automatic_label_index_creation_enabled` (the same for the other flag)

src/storage/v2/inmemory/storage.cpp
```diff
@@ -807,6 +812,25 @@ utils::BasicResult<StorageManipulationError, void> InMemoryStorage::InMemoryAcce
...
+
+      if (storage_->config_.salient.items.enable_edge_type_index_auto_creation) {
+        storage_->edge_types_to_auto_index_.WithLock([&](auto &edge_type_indices) {
+          for (const auto edge_type : edge_type_indices) {
+            CreateIndex(edge_type, false);
+          }
+          edge_type_indices.clear();
+        });
+      }
+
```
antoniofilipovic, line 833: I might not understand something here so let me phrase it as question.

What happens if there is other transaction which is currently updating creating more EDGE_TYPE edges? I might be missing something but from what I get, you first collect all edge types for certain transaction in system and then you create them at transaction end, but what if other transaction is now emplacing 10000 more new edges with that edge types, I think we will miss these edge types in index. I am only 60% sure I got  everything correctly, but that is my question

antoniofilipovic (commented):
Good work, I have important question so let me know if we need to sync

src/storage/v2/inmemory/storage.cpp
gvolfing: @antoniofilipovic 
Shouldn't be an issue. `labels_to_auto_index_` and `edge_types_to_auto_index_` are `memgraph::utils::synchronized` datastructures updating and reading from them are blocking concurrent threads to operate on them. In the case edge-types, once we get to line 826 we take a lock to the datastructure, and that will make the other thread(s) that are creating the edges with the same edge-type block and progress until they actually hold the lock. And once they actually acquire the lock using operator->() (since it is overloaded on `utils::synchronized)` they can continue execution without any issues.

tests/e2e/index_auto_creation/index_auto_creation.py
```diff
@@ -0,0 +1,318 @@
...
+    assert index_count_is(index_stats, label, 1)
+
+    execute_and_fetch_all(cursor, f"DROP INDEX ON :{label}")
+    index_stats = get_index_stats(cursor)
+    assert number_of_index_structures_are(index_stats, 0)
+
+    execute_and_fetch_all(cursor, f"MATCH (n) DETACH DELETE n")
+
+
+def test_auto_create_multiple_label_index(cursor):
```
Darych, line 67: Can we add test when we assign multiple labels to one vertex?

tests/e2e/index_auto_creation/index_auto_creation.py
```diff
@@ -0,0 +1,318 @@
...
+
+    execute_and_fetch_all(cursor, f"DROP INDEX ON :{label_from}")
+    index_stats = get_index_stats(cursor)
+    assert number_of_index_structures_are(index_stats, 4)
+
+    execute_and_fetch_all(cursor, f"DROP INDEX ON :{label_to}")
+    index_stats = get_index_stats(cursor)
+    assert number_of_index_structures_are(index_stats, 3)
+
+    execute_and_fetch_all(cursor, f"DROP EDGE INDEX ON :{edge_type1}")
```
Darych, line 302: Do you intentionally not use loop to drop edge indices?

tests/e2e/index_auto_creation/index_auto_creation.py
```diff
@@ -0,0 +1,318 @@
...
+#########################
+
+
+def test_auto_create_single_edge_type_index(cursor):
+    label_from = "LABEL_FROM"
+    label_to = "LABEL_TO"
+    edge_type = "SOMEEDGETYPE"
+    execute_and_fetch_all(cursor, f"CREATE (n:{label_from})")
+    execute_and_fetch_all(cursor, f"CREATE (n:{label_to})")
+    execute_and_fetch_all(cursor, f"MATCH (n:{label_from}), (m:{label_to}) CREATE (n)-[:{edge_type}]->(m)")
```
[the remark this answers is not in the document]
gvolfing: That way the indention looks very similar to the test cases where multiple edges are created between the same nodes. Just aesthetics, I can change it if you want to.

tests/e2e/index_auto_creation/index_auto_creation.py
gvolfing: IMO this is more readable for tests. I can change it if you insist.

tests/e2e/index_auto_creation/index_auto_creation.py
gvolfing: Added.

src/storage/v2/inmemory/storage.cpp
```diff
@@ -815,19 +822,31 @@ utils::BasicResult<StorageManipulationError, void> InMemoryStorage::InMemoryAcce
...
         storage_->labels_to_auto_index_.WithLock([&](auto &label_indices) {
-          for (const auto label : label_indices) {
-            CreateIndex(label, false);
+          for (auto &label : label_indices) {
+            --label.second;
+            // If there are multiple transactions that would like to create an
+            // auto-created index on a specific label, we only build the index
+            // when the last one commits.
+            if (label.second == 0) {
+              CreateIndex(label.first, false);
```
Darych, line 831: Why do we create only during last commit instead of the first commit?

tests/e2e/index_auto_creation/index_auto_creation.py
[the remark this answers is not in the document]
Darych: I just want to test the case, when we run a query which assigns labels to multiple vertices.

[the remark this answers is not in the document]
gvolfing: I see what you mean, will modify this test-case.

src/storage/v2/inmemory/storage.cpp
gvolfing: It is possible that 2 or more concurrent transactions are trying to insert a vertex, attempting to auto-create the same index. There are 3 steps in auto-creating a vertex
 1. Update the metadata data-structure that holds the indices we would like to create during commit.
 2. Call that function that would update the indexes with the current vertex.(This will not fire in either case since the index is not yet existing meaning that at this point there is only a set of vertices that may or may not be commited, nothing in the indices).
 3. Create the index by during commit.
 
 It is possible that `TXN1` is at step 3, while `TXN2` just finished with step 2, but not yet at step 3. If that is the case the index will only contain the vertex from `TXN1`, since `TXN2` just missed the opportunity the update the index, since it does not exist yet.
 
 If we create the index on the last active transaction that is introducing that specific label or edge-type, the index will be built correctly without missing any entries.

Darych: The problem I see here is following.
0. Auto label index creation is enabled.
1. Assume a streaming app which inserts data to Memgraph non-stop with the same label which previously hasn't been indexed.
2. In parallel somebody runs read transactions by this new label. This will always do ScanAll instead of ScanByLabelIndex.
3. Label index will not be created because there are infinite parallel write transactions.

Currently I don't see a proper solution without losing data entries. Maybe we should update indices during commit, but this will affect reads during the same transaction when in ScanByLabelIndex we will not see vertices added in the same txn.

So I would say let's keep it now as it is, because this will give correct results. And think later can we improve there or not.

Darych (approved):

PR #1860, opened 2024-03-26
Extend mgbench to improve memory tracking on TCO project

DavIvek:
### Description

We want to know how much memory Memgraph consumes in its idle state and after loading data, so that will allow us to check progress in the Lower TCO project.
Another change in this PR is the way that memory usage is calculated.
It would be more clear to see the execution memory usage, not the whole memory footprint.
```
Empty memgraph 45MB
Full memgraph 250MB → result we care about 205MB (250-45)
Peak query usage 560MB →result we care about 310MB (560-250)
```
 
Just an additional NOTE, with this changes memory usage on [bench-graph](https://bench-graph.memgraph.com/) in all cases will decrease. This reduction isn't due to a decrease in memory itself, but rather a shift in the calculation method for memory usage. After this is approved, we should inform Memgraph's team about this.

NOTE FOR REVIEWERS: To review, you can check new panels added to [bench-graph](https://bench-graph.memgraph.com/), since all of the introduced changes are shown there.

[master < Task] PR
    - **[FINAL GIT MESSAGE]**

### Documentation checklist
    - **[Release note text]**
    - **[Documentation PR link]**


andrejtonev:
@DavIvek @antejavor can we add a toggle in bench-graph that would switch between the old (total memory) and new (memory divided into base, data, execution)? In order to get to the old mode, we just need to sum the new results.

DavIvek:
@andrejtonev I will take a look.

antejavor:
@andrejtonev To me, that sounds fine, I am always for more granular measurements, you can always just keep both. cc @DavIvek

antejavor:
One more thing here that is an extension of my previous comment. I just discussed this on the call with @DavIvek 

My concern on this is with Grafana, I think we should keep the top-level, general performance panels (throughput and memory usage) as clean as possible that provide a summary of the current situation on PR. Thinking about folks who are not and won't be engaged in TCO but developing something for Memgraph. Providing a basic overall summary for glimpsing is a must. 

Having more things to visualize is a big win IMO, but more things to look at. I would keep the two general panels  as they were and maybe consider adding TCO board (**Memgraph - Memory view detailed**) 

cc @andrejtonev

andrejtonev (approved):
Gread job!

PR #1850, opened 2024-03-25
Add zoned datetime to utils::temporal

antepusic:
### Description

This PR adds a dedicated class for zoned datetimes to `utils::temporal` according to the [feature spec](https://www.notion.so/memgraph/Zoned-temporal-types-1d0de6c83b5a429180f28ec22d3e00fa?pvs=4#d22bf8ce99714a4991b1efe77517e0bc) (relevant section: **`ZonedDateTime` specification**).

The added `ZonedDateTime` class will be the base for a new Memgraph data type, to be implemented in my next PR.

[master < Task] [PR#1850](#1850)
- [x] Provide the full content or a guide for the final git message
    > Add zoned datetime to utils::temporal 

### Documentation checklist
- [x] Add the documentation label tag
- [x] Add the bug / feature label tag
- [x] Add the milestone for which this feature is intended
- [x] Write a release note, including added/changed clauses (N/A – docs unnecessary)
- [x] Link the documentation PR here (N/A – docs unnecessary)
- [x] Tag someone from docs team in the comments  (N/A – docs unnecessary)

PR #1866, opened 2024-03-27
Add the zoned datetime data type to Memgraph

antepusic:
### Description

| Specification: [Zoned temporal types](https://www.notion.so/memgraph/Zoned-temporal-types-1d0de6c83b5a429180f28ec22d3e00fa)

This PR integrates the `utils::ZonedDateTime` object developed in #1850 in order to support zoned datetimes in Memgraph.
Additions to the query module APIs will be handled in a separate PR.

[master < Epic] PR
- [x] Write E2E tests
- [x] Compare the [benchmarking results](https://bench-graph.memgraph.com/) between the master branch and the Epic branch
    > Add the zoned datetime data type to Memgraph

### Documentation checklist
    > Added timezone support to Memgraph with the new zoned datetime type, which represents points in time that have a defined timezone. You can now work with temporal data from different locales, as well as easily migrate datasets with zoned values.
    - https://github.com/memgraph/documentation/pull/723


gvolfing (approved):
LGTM

src/storage/v2/property_store.cpp
```diff
@@ -520,6 +591,69 @@ std::optional<uint64_t> DecodeTemporalDataSize(Reader &reader) {
...
+  }
+
+  auto type_value = reader.ReadUint(metadata->id_size);
+  if (!type_value) return std::nullopt;
+
+  auto microseconds_value = reader.ReadInt(metadata->payload_size);
+  if (!microseconds_value) return std::nullopt;
+
+  if (metadata->type == Type::ZONED_TEMPORAL_DATA) {
+    auto tz_str_length = reader.ReadUint(TZ_NAME_LENGTH_SIZE);
```
Darych, line 609: Maybe we can have a function in Reader class to read timezone in this format? The same as we have `ReadTimezoneOffset`

Darych: Or even better we can incapsulate logic of types in one function and return `utils::Timezone`.

tests/gql_behave/tests/memgraph_V1/features/match.feature
```diff
@@ -469,6 +469,26 @@ Feature: Match
...
             | P13DT12H |
 
+    Scenario: Test match with order by and datetime
+        Given an empty graph
+        And having executed:
+            """
+            CREATE({a: DATETIME('2024-01-22T08:11:31[Etc/UTC]')}),
+                  ({a: DATETIME('2024-01-22T08:12:31[Etc/UTC]')}),
+                  ({a: DATETIME('2024-01-22T09:12:31[Europe/Zurich]')}),
+                  ({a: DATETIME('2024-01-22T09:12:31[Europe/Warsaw]')})
```
Darych, line 479: Can we add edge cases like +18, -18 hours, not 00 minutes but [30, 45](https://www.timeanddate.com/time/time-zones-interesting.html).

tests/unit/query_expression_evaluator.cpp
```diff
@@ -2386,4 +2454,38 @@ TYPED_TEST(FunctionTest, Duration) {
...
+              today.SysMicrosecondsSinceEpoch().count(), one_sec_in_microseconds);
+  EXPECT_EQ(this->EvaluateFunction("DATETIME", TypedValue(std::map<std::string, TypedValue>{})).ValueZonedDateTime(),
+            memgraph::utils::ZonedDateTime({{}, {}, memgraph::utils::DefaultTimezone()}));
+
+  EXPECT_THROW(this->EvaluateFunction("DATETIME", "{}"), memgraph::utils::BasicException);
+  EXPECT_THROW(
+      this->EvaluateFunction("DATETIME", TypedValue(std::map<std::string, TypedValue>{{"hours", TypedValue(1970)}})),
+      QueryRuntimeException);
+  EXPECT_THROW(
+      this->EvaluateFunction("DATETIME", TypedValue(std::map<std::string, TypedValue>{{"seconds", TypedValue(1970)}})),
```
Darych, line 2488: What will happen if you will pass just "year" or "month" or "day" or "millisecond" or "microsecond"?

src/communication/bolt/v1/decoder/decoder.hpp
```diff
@@ -131,6 +131,18 @@ class Decoder {
             return ReadUnboundedEdge(data);
           case Signature::Path:
             return ReadPath(data);
+          case Signature::DateTime:
+            return ReadDateTime(data);
+          case Signature::DateTimeZoneId:
+            return ReadDateTimeZoneId(data);
+          case Signature::LegacyDateTime: {
+            if (major_v_ > 4) return false;
```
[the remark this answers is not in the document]
antepusic: 1. Per the [documentation](https://memgraph.com/docs/client-libraries) Memgraph does provide backward compatibility. In any case, we should support the pre-Bolt v5 structures because mgclient is v4.
2. The latest versions of mgconsole/Lab still open Bolt v4 connections, which means that Bolt uses the legacy structures for their values.
3. The legacy vs. latest distinction only applies to the Bolt protocol. In Memgraph, they are represented by the same type.
4. No (legacy datetime is a structure for zoned temporal values, and my PR is the one that will add support for them)

src/communication/bolt/v1/decoder/decoder.hpp
```diff
@@ -630,5 +642,95 @@ class Decoder {
...
+  }
+
+  bool ReadDateTime(Value *data) {
+    int64_t time_since_epoch;
+    if (!ComputeTimeSinceEpoch(&time_since_epoch)) {
+      return false;
+    }
+
+    Value tz_offset_secs;
+    if (!ReadValue(&tz_offset_secs, Value::Type::Int)) {
```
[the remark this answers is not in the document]
antepusic: Yes. With zoned temporal values, the Bolt protocol specifies that each structure (`DateTime`, `DateTimeZoneId`, and `Legacy DateTime` & `Legacy DateTimeZoneId`) has a field with timezone data.

tests/unit/query_expression_evaluator.cpp
antepusic: In that case, the created zoned datetime value will use the default values for all other properties. I’ll add a test to verify that.

src/query/interpret/eval.hpp
```diff
@@ -714,6 +752,9 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
         result.emplace("microsecond", TypedValue(lt.microsecond, ctx_->memory));
         return TypedValue(result, ctx_->memory);
       }
+      case TypedValue::Type::ZonedDateTime: {
+        throw QueryRuntimeException("Can't coerce `{}` to Map.", expression_result.ValueZonedDateTime().ToString());
```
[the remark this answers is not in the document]
antepusic: Either way is doable; in this case I chose this way for the following reasons:
* It’s in line with other graph databases.
* The year...microsecond components aren’t actual members (whereas they are in e.g. `LocalDateTime`); they’re rather computed properties
* More consistent for users - If in the future we add new ZonedDateTime components (e.g. week, epochMicroseconds or offset), existing queries with AllPropertyLookup (i.e. map projections) will change since the map projection should contain them as new fields). By forbidding ZonedDateTime in AllPropertyLookup, this is avoided.

src/storage/v2/property_value.hpp
```diff
@@ -290,8 +308,12 @@ inline std::ostream &operator<<(std::ostream &os, const PropertyValue &value) {
                            [](auto &stream, const auto &pair) { stream << pair.first << ": " << pair.second; });
       return os << "}";
     case PropertyValue::Type::TemporalData:
-      return os << fmt::format("type: {}, microseconds: {}", TemporalTypeTostring(value.ValueTemporalData().type),
+      return os << fmt::format("type: {}, microseconds: {}", TemporalTypeToString(value.ValueTemporalData().type),
                                value.ValueTemporalData().microseconds);
+    case PropertyValue::Type::ZonedTemporalData:
+      auto temp_value = value.ValueZonedTemporalData();
+      return os << fmt::format("type: {}, microseconds: {}, timezone: {}", ZonedTemporalTypeToString(temp_value.type),
+                               temp_value.IntMicroseconds(), temp_value.timezone.ToString());
```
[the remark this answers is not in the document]
antepusic: Either way works. I only didn’t make a dedicated function because this line would be the only place where it’s used.

src/storage/v2/property_store.cpp
antepusic: Agree that encapsulation is a good idea, will do.

Darych (approved):

src/utils/temporal.hpp
```diff
@@ -414,13 +464,33 @@ struct ZonedDateTime {
   friend ZonedDateTime operator-(const ZonedDateTime &zdt, const Duration &dur) { return zdt + (-dur); }
 
   friend Duration operator-(const ZonedDateTime &lhs, const ZonedDateTime &rhs) {
-    return Duration(lhs.MicrosecondsSinceEpoch()) - Duration(rhs.MicrosecondsSinceEpoch());
+    return Duration(lhs.SysMicrosecondsSinceEpoch().count()) - Duration(rhs.SysMicrosecondsSinceEpoch().count());
+  }
+
+  int64_t LocalYear() const { return int(LocalYMD().year()); }
+  uint64_t LocalMonth() const { return static_cast<unsigned>(LocalYMD().month()); }
+  uint64_t LocalDay() const { return static_cast<unsigned>(LocalYMD().day()); }
```
gvolfing, line 472: Can we use static_cast<> instead of c-style cast on line 470?
also can we just static cast into the numbered type? e.g. `...return static_cast<uint64_t>(...`?

antepusic:
@kgolubic Tagging you about the release note & docs

gvolfing (approved):

src/utils/temporal.hpp
antepusic: Fixed L470 to use `static_cast`. I use `static_cast<int>` and `static_cast<unsigned>` because other conversions aren’t available: `.year` returns `std::chrono::year`, which only has `operator int`. Analogously, `std::chrono::month` and `::day` only have `operator unsigned` available.
* https://en.cppreference.com/w/cpp/chrono/year
* https://en.cppreference.com/w/cpp/chrono/month
* https://en.cppreference.com/w/cpp/chrono/day

PR #1869, opened 2024-03-27
Update validation.py

DavIvek:
### Description

Validation.py which is part of the mgbench isn't up to date, with this PR it will be.

[master < Task] PR
    - **[FINAL GIT MESSAGE]**

### Documentation checklist
    - **[Release note text]**
    - **[Documentation PR link]**


tests/mgbench/validation.py
```diff
@@ -88,6 +88,13 @@ def pars_args():
         help="number of workers used to import the dataset",
     )
 
+    parser.add_argument(
```
antejavor, line 91: Just a comment here for understanding so this script was added so we can check if Neo4j and Memgraph return the same results from `LDBC` queries, since those queries are complicated. 

It is sufficient to run just a single query and get the results back from the query, there is no need for concurrency, hence multiple workers. 

I referred to `validated.py` as being out of date because it didn't fill the purpose we were discussing  (the persistent results between runs, etc). Also,  not run in our CI/CD environment for things you mentioned you need (validating results). 

There is a lot of duplication and hardcoding here in this script.

DavIvek: BenchmarkContext object was constructed without providing number of workers for benchmark, it will be overwritten later in client.cpp anyway, but I will remove this to make code more clear.

antejavor (approved):
LGTM

imilinovic (approved):

PR #1884, opened 2024-03-29
Allow user to connect without database

andrejtonev:
### Description

Allow users to connect without a database defined.
If a user has a default db, the user will connect to that database.
If a user does not have a default db, we will connect without a database defined. Only db-less queries should be allowed.
If a user defines the database via the client, we will use the same logic as above, but each query will be executed against the user-defined db. Multi-tenant auth will be executed at first query execution.

Q: Should we allow multi-tenant queries (ex: `USE DATABASE x`) even without db?

[master < Task] PR
    - **[FINAL GIT MESSAGE]**

### Documentation checklist
    - **[Release note text]**
    - **[Documentation PR link]**

### DIFF
--- a/src/glue/SessionHL.cpp
+++ b/src/glue/SessionHL.cpp
@@ -161,9 +161,18 @@ bool SessionHL::Authenticate(const std::string &username, const std::string &pas
   }
 #ifdef MG_ENTERPRISE
   // Start off with the default database
-  interpreter_.SetCurrentDB(GetDefaultDB(), false);
+  try {
+    interpreter_.SetCurrentDB(GetDefaultDB(), false);
+  } catch (auth::AuthException &) {
+    // Failed to get default db, connect without db
+    interpreter_.ResetDB();
+  }
 #endif
-  implicit_db_.emplace(GetCurrentDB());
+  const auto db = GetCurrentDB();
+  if (db.empty())
+    implicit_db_.reset();
+  else
+    implicit_db_.emplace(std::move(db));
   return res;
 }
 
@@ -254,7 +263,8 @@ using memgraph::communication::bolt::Value;
 #ifdef MG_ENTERPRISE
 auto SessionHL::Route(std::map<std::string, Value> const &routing,
                       std::vector<memgraph::communication::bolt::Value> const & /*bookmarks*/,
-                      std::map<std::string, Value> const & /*extra*/) -> std::map<std::string, Value> {
+                      std::map<std::string, Value> const &
+                      /*extra*/) -> std::map<std::string, Value> {
   auto routing_map = ranges::views::transform(
                          routing, [](auto const &pair) { return std::pair(pair.first, pair.second.ValueString()); }) |
                      ranges::to<std::map<std::string, std::string>>();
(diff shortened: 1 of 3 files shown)


src/glue/SessionHL.cpp
```diff
@@ -121,9 +122,14 @@ void MultiDatabaseAuth(memgraph::query::QueryUserOrRole *user, std::string_view
 namespace memgraph::glue {
 
 #ifdef MG_ENTERPRISE
-std::string SessionHL::GetDefaultDB() {
+std::optional<std::string> SessionHL::GetDefaultDB() {
```
Ignition, line 125: Future change: do we want stronger default DB? `name + uuid`

tests/integration/auth/runner.py
```diff
@@ -208,89 +208,89 @@ def cleanup():
...
-                            details.append("DENIED TO USER")
-                        if mapped:
-                            if role_perm == "GRANT":
-                                details.append("GRANTED TO ROLE")
-                            elif role_perm == "DENY":
-                                details.append("DENIED TO ROLE")
-                        expected.append(", ".join(details))
-                    execute_checker(checker_binary, expected)
-    print("\033[1;36m~~ Finished permissions test ~~\033[0m\n")
+    # # Run the test with all combinations of permissions
```
Ignition, line 211: Is this bring removed or updated?

src/glue/SessionHL.cpp
andrejtonev: That makes sense.
I will implement this now, since we missed the 2.17 release.

andrejtonev: The whole interface is name based. 
The user can define the default database and access to a database via its name.
Drivers that connect directly to a database use the name.
There is no auth logic that handles database UUIDs. It should be transparent to it.
If a database is deleted, it is removed from auth as well.

kgolubic:
@andrejtonev I'm missing the Release note entry and documentation PR.

tests/integration/auth/runner.py
andrejtonev: That was by mistake

kgolubic:
@andrejtonev I will use this for RN:

- Users can now connect to Memgraph without specifying a database. If a user has a default database, they will connect to that database. If not, they will connect without a database defined, allowing only database-less queries. Users can also define the database via the client, executing each query against the specified database. Multi-tenant authentication will be executed upon the first query execution.
[#1884](https://github.com/memgraph/memgraph/pull/1884)

andrejtonev:
### Description

Allow users to connect without a database defined.
If a user has a default db, the user will connect to that database.
If a user does not have a default db, we will connect without a database defined. Only db-less queries should be allowed.
If a user defines the database via the client, we will use the same logic as above, but each query will be executed against the user-defined db. Multi-tenant auth will be executed at first query execution.

Q: Should we allow multi-tenant queries (ex: `USE DATABASE x`) even without db? **YES**

Bugfix: Was not updating the explicit flag when user explicitly defines the same database as the current one. This would in turn allow db switching, which is prohibited if user connects via `mgclient.connect(database="db", ...)`

[master < Task] PR
    - **[FINAL GIT MESSAGE]**


### Documentation checklist
    - **Allow database-less connections**
    - [853](https://github.com/memgraph/documentation/pull/853)

Ignition (approved):

PR #1896, opened 2024-04-02
Add GitHub action for running benchmark on large dataset manually

DavIvek:
### Description

The purpose of this PR is to add GitHub action for running benchmarks on large datasets manually.
We already have a workflow file with a similar purpose but it looks to me like it isn't up to date. Maybe instead of creating a new action, we can just update the old one (I will sync with someone from infra about this).

[master < Task] PR
    - **[FINAL GIT MESSAGE]**

### Documentation checklist
    - **[Release note text]**
    - **[Documentation PR link]**


.github/workflows/benchmark_on_pokec_large.yaml
```diff
@@ -0,0 +1,101 @@
+name: Run benchmark on pokec large dataset
+
+on:
+  workflow_dispatch:
+  pull_request:
```
MarkoBarisic, line 5: We shouldn't run a workflow with timeout of 720 minutes on every pull request push.
On which events do we want to run this workflow, write your ideas in the reply.

release/package/mgbuild.sh
```diff
@@ -562,7 +563,7 @@ test_memgraph() {
       docker exec -u mg $build_container bash -c "$EXPORT_LICENSE && $EXPORT_ORG_NAME && $ACTIVATE_TOOLCHAIN && $ACTIVATE_CARGO && cd $MGBUILD_ROOT_DIR/build "'&& ulimit -s 262144 && ctest -R memgraph__benchmark -V'
     ;;
     mgbench)
-      docker exec -u mg $build_container bash -c "$EXPORT_LICENSE && $EXPORT_ORG_NAME && cd $MGBUILD_ROOT_DIR/tests/mgbench "'&& ./benchmark.py vendor-native --num-workers-for-benchmark 12 --export-results benchmark_result.json pokec/medium/*/*'
+      docker exec -u mg $build_container bash -c "$EXPORT_LICENSE && $EXPORT_ORG_NAME && cd $MGBUILD_ROOT_DIR/tests/mgbench && ./benchmark.py vendor-native --num-workers-for-benchmark 12 --export-results benchmark_result.json pokec/$POKEC_SIZE/*/*"
```
MarkoBarisic, line 566: The idea is good, but needs some modifications:
- move the `POKEC_SIZE` local variable to the `mgbench` case.  Run `shift 1` before the variable line, and take `$1` into variable value
- validate the content of the passed variable, make sure it's _small, medium or large_ (guessing what sizes we have here, check that). If the validation fails, print out the correct error message. Take a look at the `check_support` function in this file.

release/package/mgbuild.sh
```diff
@@ -502,6 +502,7 @@ test_memgraph() {
   local EXPORT_LICENSE="export MEMGRAPH_ENTERPRISE_LICENSE=$enterprise_license"
   local EXPORT_ORG_NAME="export MEMGRAPH_ORGANIZATION_NAME=$organization_name"
   local BUILD_DIR="$MGBUILD_ROOT_DIR/build"
+  local POKEC_SIZE=${2:-'medium'}
```
MarkoBarisic, line 505: Move to `mgbench` case

MarkoBarisic (changes requested):

.github/workflows/benchmark_on_pokec_large.yaml
DavIvek: This was only for testing, I will change this part.

MarkoBarisic (approved):
LGTM

PR #1897, opened 2024-04-03
Add durable coordinator cluster config

antoniofilipovic:
### Description

This diff fixes the bug where the coordinator on restart doesn't have info on the cluster server config. This results in an issue that the coordinator that has an empty state can become a leader (it votes for itself and there are as it doesn't see any other coordinator in a cluster), and another coordinator which has cluster info stays as a follower as it sees other coordinator has become a leader.

[master < Task] PR
    - Fix bug on the wrong coordinator becoming the leader 

### Documentation checklist
    - Fix bug on wrong coordinator becoming leader due to missing data on cluster state
    - *No need, only changelog


src/coordination/coordinator_instance.cpp
```diff
@@ -387,6 +387,7 @@ auto CoordinatorInstance::TryFailover() -> void {
     if (raft_state_.IsLockOpened() && raft_state_.IsLeader()) {
```
as51340, line 387: I would rename PR - the main thing is durable cluster config

src/coordination/coordinator_log_store.cpp
```diff
@@ -46,7 +46,8 @@ auto CoordinatorLogStore::FindOrDefault_(uint64_t index) const -> ptr<log_entry>
 
 uint64_t CoordinatorLogStore::next_slot() const {
   auto lock = std::lock_guard{logs_lock_};
-  return start_idx_ + logs_.size() - 1;
+  auto slot = start_idx_ + logs_.size() - 1;
+  return slot;
```
as51340, line 50: just return or log slot if you want to log

src/coordination/include/nuraft/coordinator_state_manager.hpp
```diff
@@ -60,6 +61,7 @@ class CoordinatorStateManager : public state_mgr {
   ptr<srv_config> my_srv_config_;
   ptr<cluster_config> cluster_config_;
   ptr<srv_state> saved_state_;
+  std::unique_ptr<kvstore::KVStore> kv_store_;
```
as51340, line 64: I think you don't need unique_ptr to kvstore? just use value semantic

src/coordination/coordinator_state_manager.cpp
```diff
@@ -31,37 +33,59 @@ CoordinatorStateManager::CoordinatorStateManager(CoordinatorInstanceInitConfig c
...
+  // TODO(antoniofilipovic): Implement storing of server state to disk. For now
+  // as server state is just term and voted_for, we don't have to store it
+  spdlog::trace("Saving server state in coordinator state manager.");
   ptr<buffer> buf = state.serialize();
   saved_state_ = srv_state::deserialize(*buf);
 }
 
 auto CoordinatorStateManager::read_state() -> ptr<srv_state> {
-  // Just return in-memory data in this example.
-  // May require reading from disk here, if it has been written to disk.
```
as51340, line 64: Is it true that `read_state` and `save_state` is for Raft logs and `save_config`, `load_config` for cluster membership?

src/coordination/coordinator_state_manager.cpp
```diff
@@ -31,37 +33,59 @@ CoordinatorStateManager::CoordinatorStateManager(CoordinatorInstanceInitConfig c
...
   auto servers = cluster_config_->get_servers();
+  std::vector<std::tuple<int, std::string, std::string>> servers_vec;
   for (auto const &server : servers) {
-    spdlog::trace("Server id: {}, endpoint: {}", server->get_id(), server->get_endpoint());
+    servers_vec.emplace_back(static_cast<int>(server->get_id()), server->get_endpoint(), server->get_aux());
+    spdlog::trace("Stored cluster config with id: {}, endpoint: {} and aux data: {} to disk.",
+                  static_cast<int>(server->get_id()), server->get_endpoint(), server->get_aux());
   }
+  nlohmann::json json(servers_vec);
+  kv_store_->Put("servers", json.dump());
```
as51340, line 76:
```suggestion
  kv_store_->Put("servers", nlohmann::json(servers_vec).dump());
```

src/coordination/coordinator_state_manager.cpp
```diff
@@ -31,37 +33,59 @@ CoordinatorStateManager::CoordinatorStateManager(CoordinatorInstanceInitConfig c
...
 auto CoordinatorStateManager::save_config(cluster_config const &config) -> void {
-  // Just keep in memory in this example.
-  // Need to write to disk here, if want to make it durable.
+  MG_ASSERT(kv_store_, "Disk folder should be created");
   ptr<buffer> buf = config.serialize();
   cluster_config_ = cluster_config::deserialize(*buf);
-  spdlog::info("Saving cluster config.");
+  spdlog::info("Saving cluster config to disk.");
   auto servers = cluster_config_->get_servers();
+  std::vector<std::tuple<int, std::string, std::string>> servers_vec;
```
as51340, line 69:
```suggestion
auto const servers_vec = ranges::views::transform(cluser_config_->get_servers(), [](auto const &server) {
  return std::make_tuple(...);
}) | to<std::vector>();
```

No need to change but with this you avoid some boring half bugs like missing const on servers or specifying whole std::tuple ...

src/coordination/coordinator_state_manager.cpp
```diff
@@ -31,37 +33,59 @@ CoordinatorStateManager::CoordinatorStateManager(CoordinatorInstanceInitConfig c
...
+  kv_store_ = std::make_unique<kvstore::KVStore>(config.durability_dir);
 }
 
 auto CoordinatorStateManager::load_config() -> ptr<cluster_config> {
-  // Just return in-memory data in this example.
-  // May require reading from disk here, if it has been written to disk.
-  spdlog::trace("Loading cluster config");
+  MG_ASSERT(kv_store_, "Durability folder should be created.");
+  spdlog::trace("Loading cluster config from RocksDb");
+  auto servers = kv_store_->Get("servers");
```
as51340, line 43: const

src/coordination/coordinator_state_manager.cpp
```diff
@@ -31,37 +33,59 @@ CoordinatorStateManager::CoordinatorStateManager(CoordinatorInstanceInitConfig c
...
+  }
+  spdlog::trace("Loading cluster config from disk.");
+  auto const json = nlohmann::json::parse(servers.value());
+  auto const real_servers = json.get<std::vector<std::tuple<int, std::string, std::string>>>();
+  cluster_config_->get_servers().clear();
+  for (auto const &real_server : real_servers) {
+    auto const &[coord_id, endpoint, aux] = real_server;
+    spdlog::trace("Recreating cluster config with id: {}, endpoint: {} and aux data: {} from disk.", coord_id, endpoint,
+                  aux);
+    auto one_server_config = cs_new<srv_config>(coord_id, 0, endpoint, aux, false);
```
as51340, line 56: You can move endpoint and aux. Aux is a big string (json) so that should improve perf

tests/unit/coordinator_instance.cpp
```diff
@@ -34,18 +34,23 @@ using memgraph::io::network::Endpoint;
...
-  void SetUp() override {}
+  void SetUp() override {
+    if (!std::filesystem::exists(main_data_directory)) return;
+    std::filesystem::remove_all(main_data_directory);
+  }
 
-  void TearDown() override {}
+  void TearDown() override {
+    if (!std::filesystem::exists(main_data_directory)) return;
+    std::filesystem::remove_all(main_data_directory);
```
as51340, line 44: are these all unit tests on which we need to delete data dir

as51340 (changes requested):
Looks great!

src/coordination/coordinator_instance.cpp
antoniofilipovic: Makes sense

src/coordination/include/nuraft/coordinator_state_manager.hpp
antoniofilipovic: The problem is you can't default construct it because that constructor is deleted. You can only construct it with path. But that way I can't check if the path is valid so I would like to leave it as it is

src/coordination/coordinator_state_manager.cpp
antoniofilipovic: It is true what you said for `save_config` and `load_config`, they are for cluster membership. And as far as I can tell `read_state` and `save_state` is for `server_state` which holds info on term and what has been `voted_for`

src/coordination/coordinator_state_manager.cpp
antoniofilipovic: Ah, ranges, nice!

src/coordination/coordinator_state_manager.cpp
antoniofilipovic: Yes, makes sense

tests/unit/coordinator_instance.cpp
antoniofilipovic: Yes, only four of them

src/coordination/include/nuraft/coordinator_state_manager.hpp
as51340: I think no problem with this, I see many places where check isn't done. Or do the check before, in memgraph.cpp. I think it doesn't make sense to unique_ptr for that

src/coordination/coordinator_state_manager.cpp
```diff
@@ -31,37 +34,60 @@ CoordinatorStateManager::CoordinatorStateManager(CoordinatorInstanceInitConfig c
...
+  auto const servers_vec =
+      ranges::views::transform(
+          cluster_config_->get_servers(),
+          [](auto const &server) {
+            spdlog::trace("Created cluster config with id: {}, endpoint: {} and aux data: {} to disk.",
+                          static_cast<int>(server->get_id()), server->get_endpoint(), server->get_aux());
+            return std::tuple{static_cast<int>(server->get_id()), server->get_endpoint(), server->get_aux()};
+          }) |
+      ranges::to<std::vector>();
+  kv_store_->Put("servers", nlohmann::json(servers_vec).dump());
```
as51340, line 78: "servers" as constexpr variable

as51340 (changes requested):

src/coordination/include/nuraft/coordinator_state_manager.hpp
antoniofilipovic: I will do check before

src/coordination/coordinator_state_manager.cpp
antoniofilipovic: Yes, forgot about that

as51340 (approved):
:muscle:

antoniofilipovic:
@kgolubic tagging you here

src/memgraph.cpp
```diff
@@ -444,9 +444,10 @@ int main(int argc, char **argv) {
   }
 
   if (FLAGS_coordinator_id && FLAGS_coordinator_port) {
-    coordinator_state.emplace(CoordinatorInstanceInitConfig{.coordinator_id = FLAGS_coordinator_id,
-                                                            .coordinator_port = FLAGS_coordinator_port,
-                                                            .bolt_port = FLAGS_bolt_port});
+    auto const high_availability_data_dir = FLAGS_data_directory + "/high_availability";
```
gitbuda, line 447: I would do the following (NOTE: adding one more folder level):
```
auto const high_availability_data_dir = std::path(FLAGS_data_directory) / std::path("high_availability") / std::path("coordinator");
```

PR #1914, opened 2024-04-08
Fix jepsen bank test

as51340:
### Description

- For bank test to run successfully we need to have 100% durable storage -> hence we need to `fsync` after every txn.
- Args will now be passed when running all Jepsen tests from run.sh to lein.
- Plotter added to the bank test for easier inspection of accounts
- edn files from resources can point only to a single configuration to avoid misuses
- test-all lein functionality removed for the simplicity
- Debian 12 instructions
- cluster-refresh correctly accepts args now

    - **Fix Jepsen bank test**

### Documentation checklist
    - **[Release note text]**
    - **[Documentation PR link]**


tests/jepsen/src/jepsen/memgraph/large.clj
```diff
@@ -61,11 +61,9 @@
                                       (when (= replication-role :main)
                                         (utils/with-session conn session
                                           (try
+                                            ; Can fail for various reasons, not important at this point.
                                             (client/detach-delete-all session)
-                                            (catch Exception e
-                                              (when-not (string/includes? (str e) "At least one SYNC replica has not confirmed committing last transaction.")
-                                                (throw (Exception. (str "Invalid exception when deleting all nodes: " e)))); Exception due to down sync replica is accepted/expected
-                                              )))))
+                                            (catch Exception _)))))
```
as51340, line 66: Fix should be committed independently from bank test

as51340:
@kgolubic No need for RN, internal improvement

.github/workflows/diff.yaml
```diff
@@ -15,496 +15,10 @@ on:
...
-          --toolchain $TOOLCHAIN \
-          --os $OS \
-          --arch $ARCH \
-          stop --remove
-
   release_jepsen_test:
     name: "Release Jepsen Test"
     runs-on: [self-hosted, Linux, X64, DockerMgBuild]
-    timeout-minutes: 80
+    timeout-minutes: 360
```
MarkoBarisic, line 21: Why 6 hours timeout? This is far too long for a diff workflow job.

.github/workflows/diff.yaml
```diff
@@ -15,496 +15,10 @@ on:
       - "licenses/*"
 
 jobs:
-  community_build:
```
MarkoBarisic, line 18: Instead of deleting everything but jepsen from `diff` for testing, you can just add
`if: false` in the same level as `name: `. The effect will be the same

MarkoBarisic: This is a tip for future work

MarkoBarisic (changes requested):

.github/workflows/diff.yaml
MarkoBarisic: Resolve this after reading :+1:

.github/workflows/diff.yaml
as51340: just for testing, will revert once it will be for merging

MarkoBarisic (approved):
LGTM, make sure to revert changes in diff before merge

antoniofilipovic (approved):

PR #1923, opened 2024-04-09
Add checks after request leadership

antoniofilipovic:
### Description

This diff checks if everything is properly updated before the follower coordinator becomes LEADER on user action. It is still possible for user to do action while failover is happening but now we have a way to detect that everything is properly set up.

[master < Task] PR
    - Add setup check on leadership change

### Documentation checklist
- [x] Add the bug/feature label tag
    - Add check to confirm correct setup after requesting leadership 
    - No docs only changelog

### DIFF
--- a/src/coordination/coordinator_instance.cpp
+++ b/src/coordination/coordinator_instance.cpp
@@ -17,6 +17,7 @@
 #include "dbms/constants.hpp"
 #include "nuraft/coordinator_state_machine.hpp"
 #include "nuraft/coordinator_state_manager.hpp"
+#include "utils/counter.hpp"
 #include "utils/functional.hpp"
 #include "utils/resource_lock.hpp"
 
@@ -79,7 +80,9 @@ CoordinatorInstance::CoordinatorInstance(CoordinatorInstanceInitConfig const &co
                                         &CoordinatorInstance::DemoteFailCallback);
                 });
             std::ranges::for_each(repl_instances_, [](auto &instance) { instance.StartFrequentCheck(); });
-            semaphore_leader.release();
+            auto lock = std::unique_lock{leader_mutex_};
+            leader_set_up_.store(true);
+            become_leader_.notify_all();
           },
           [this]() {
             thread_pool_.AddTask([this]() {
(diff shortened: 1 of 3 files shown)


antoniofilipovic:
@kgolubic tagging you here, but this is for 2.17 release

src/coordination/coordinator_instance.cpp
```diff
@@ -354,6 +358,9 @@ void CoordinatorInstance::ForceResetCluster() {
   std::ranges::for_each(repl_instances_, [](auto &instance) { instance.StartFrequentCheck(); });
 
   MG_ASSERT(!raft_state_.IsLockOpened(), "After force reset we need to be in healthy state.");
+  auto leader_lock = std::unique_lock{leader_mutex_};
+  leader_set_up_.store(true);
+  become_leader_.notify_all();
```
as51340, line 363: Why is this done in`ForceReset`? Also, what is the meaning of `leader_set_up_`, why it is saved in coordinator_instance?

src/coordination/coordinator_instance.cpp
```diff
@@ -539,9 +548,39 @@ auto CoordinatorInstance::SetReplicationInstanceToMain(std::string_view instance
...
+  auto maybe_stop = utils::ResettableCounter<5>();
+  do {
+    spdlog::trace("Waiting for coordinator to become leader.");
+    if (!become_leader_.wait_for(lock, waiting_period, [this]() { return leader_set_up_.load(); })) {
+      waiting_period *= expontential_backoff;
+    }
+  } while (!maybe_stop() && !leader_set_up_);
+
+  return leader_set_up_;
+}
```
as51340, line 570: I understand this, we are waiting for callback to be executed to make sure we are the leader, right?

src/coordination/coordinator_instance.cpp
as51340: Doesn't `ForceReset` only change status of replication instances?

as51340 (changes requested):

src/coordination/coordinator_instance.cpp
antoniofilipovic: Change as talked on call

antoniofilipovic:
This PR will get closed as we discussed and we won't introduce any checks on request leadership rather completely remove it

Issue #1648, opened 2024-01-18
Cartesian product + Index cause wrong result


Aurora-yzj:
**Memgraph version**
2.13.0

**Environment**
Operating system: Ubuntu 22.04
Architecture: AMD64
Connect method: any connect method
Driver: any driver
With docker

**Describe the bug**
Memgraph outputs an empty query result when using an index.

**To Reproduce**
1. Run the following query to insert the initial value and create the index:
```
CREATE INDEX ON :L6(k38);
CREATE (n0 :L0:L6{k0 : false, k1 : 890765719, k40 : true, k2 : false, k3 : false, k4 : false, k41 : true, k5 : -1311698154, k6 : "MCotH", k37 : false, k36 : true, k39 : false, id : 0});
CREATE (n0 :L0:L6{k0 : true, k1 : 1304545343, k40 : true, k2 : true, k3 : true, k4 : false, k41 : false, k5 : 1173141482, k6 : "kAIBO99", k37 : true, k36 : true, k39 : false, k38 : 269756692, id : 1});
CREATE (n0 :L0:L6:L11{k0 : false, k70 : "Ml7qmnNQ", k1 : 1098267178, k40 : true, k2 : true, k3 : false, k4 : true, k41 : false, k5 : -786706104, k66 : "QTakNN", k6 : "Wzw0uPB", k65 : 1803875538, k68 : "JeNb8E7V1", k67 : false, k37 : true, k36 : true, k69 : -109922962, k39 : false, k38 : -385421608, id : 2});
CREATE (n0 :L0{k1 : -1006597681, k2 : true, k3 : true, k4 : false, k5 : -1599968686, id : 3, k6 : "xikzo6"});
CREATE (n0 :L0:L6{k1 : 1588792135, k40 : true, k2 : false, k3 : false, k4 : true, k41 : true, k5 : 680249422, k6 : "XF7kSg", k37 : false, k36 : false, k39 : false, k38 : 1463404308, id : 4});
CREATE (n0 :L0:L6{k0 : true, k1 : -1731198578, k40 : true, k2 : true, k3 : false, k4 : true, k41 : false, k5 : -376075495, k6 : "wxjyq", k37 : false, k36 : true, k39 : false, k38 : -1643381344, id : 5});
CREATE (n0 :L0{k0 : true, k1 : -421862127, k2 : false, k3 : true, k4 : false, k5 : -804881356, id : 6, k6 : "5uWHrml5"});
[...]

PR #1924, opened 2024-04-09
Filter embedding after Cartesian operator

andrejtonev:
Bugfix: Planner wrongly assumes Filter will always be right after an EdgeUniquenessFilter.

This in turn cases some plans to miss a Cartesian to Join swap.

### DIFF
--- a/src/query/plan/rewrite/index_lookup.hpp
+++ b/src/query/plan/rewrite/index_lookup.hpp
@@ -149,29 +149,30 @@ class IndexLookupRewriter final : public HierarchicalLogicalOperatorVisitor {
         input = input->input().get();
       }
 
-      std::vector<Symbol> modified_symbols;
-      for (const auto &filter : op.all_filters_) {
-        if (filter.property_filter) {
-          modified_symbols.push_back(filter.property_filter->symbol_);
-        }
-      }
-
-      auto does_modify = [&]() {
-        for (const auto &sym_in : input->ModifiedSymbols(*symbol_table_)) {
-          if (std::find(modified_symbols.begin(), modified_symbols.end(), sym_in) != modified_symbols.end()) {
-            return true;
+      const bool is_child_cartesian = input->GetTypeInfo() == Cartesian::kType;
+      if (is_child_cartesian) {
+        std::vector<Symbol> modified_symbols;
+        for (const auto &filter : op.all_filters_) {
+          if (filter.property_filter) {
+            modified_symbols.push_back(filter.property_filter->symbol_);
           }
         }
-        return false;
-      };
-
-      const bool is_child_cartesian = input->GetTypeInfo() == Cartesian::kType;
-      if (is_child_cartesian && does_modify()) {
-        // if we removed something from filter in front of a Cartesian, then we are doing a join from
-        // 2 different branches
-        auto *cartesian = dynamic_cast<Cartesian *>(input);
-        auto indexed_join = std::make_shared<IndexedJoin>(cartesian->left_op_, cartesian->right_op_);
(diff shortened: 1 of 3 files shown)


tests/e2e/query_planning/query_planning_cartesian.py
```diff
@@ -38,5 +38,37 @@ def test_indexed_join_with_indices(memgraph):
...
+        " * Expand (n2)<-[r1]-(n3)",
+        " * Filter (n2 :Label1)",
+        " * ScanAll (n2)",
+        " * Once",
+    ]
+
+    results = list(
+        memgraph.execute_and_fetch(
+            "EXPLAIN MATCH (n0 :Label1)<-[r0]-(n1), (n2 :Label1)<-[r1]-(n3 :Label0) UNWIND [1] AS a0 MATCH (n4)<-[r2]-(n5 :Label0 :Label1) WHERE (((n2.Prop0) > (n0.Prop1)))  RETURN *"
+        )
```
Josipmrden, line 66: Where is this n2.Prop0 > n0.Prop1 evaluated, I can't seem to see it from the explain :eyes:

andrejtonev: It is inside the `ScanAllByLabelPropertyRange` filter. The printout isn't verbose...

Ignition: Why does the query need to be this complex?

andrejtonev: This is the query that broke our planner/optimizer.
It has to be so complex because otherwise our planner works :smile: 
What happened is that the `Unwind` got between the `Filter {n2.Prop0 > n0.Prop1}` being embedded and the `EdgeUniquenessFilter`.
Before if the `EdgeUniquenessFilter` was right before the `Filter`, we would know that a `Cartesian` was being used and would swap it with a `Join`. 
This logic broke once the `Unwind` was introduced.
This PR tries to make sure all queries that need to swap to a Join do so.

src/query/plan/rewrite/index_lookup.hpp
```diff
@@ -136,21 +136,43 @@ class IndexLookupRewriter final : public HierarchicalLogicalOperatorVisitor {
...
+        parent = input;
+        input = input->input().get();
+      }
+
+      std::vector<Symbol> modified_symbols;
+      for (const auto &filter : op.all_filters_) {
+        if (filter.property_filter) {
+          modified_symbols.push_back(filter.property_filter->symbol_);
+        }
+      }
```
Ignition, line 157: Building `modified_symbols` before we know we need it.
Prefer to be done only if `is_child_cartesian == true`
Atm `does_modify` is O(n^2), your call on if you change it the number of symbols should be small that it shouldn't ever matter.

Ignition (changes requested):

kgolubic:
@andrejtonev I'm missing the Release note entry.

andrejtonev:
### Description

Filters can get embedded into index scans. 
In cases where a filter in after a Cartesian, we need to substitute the Cartesian for a Join. 
This PR fixes a particular case where that substitution was not performed.

closes #1648 

[master < Task] PR
    - **[FINAL GIT MESSAGE]**


### Documentation checklist
    - **Bugfix: Converting a Cartesian to Join in all applicable cases**
    - **[Documentation PR link]**

Ignition (approved):

kgolubic:
For Release note I'll use:

- Fixed an issue where filters following a Cartesian operation were not always correctly embedded into index scans, causing some query plans to miss a Cartesian-to-Join swap.
#1924

PR #1926, opened 2024-04-10
Auto index creation fixes

gvolfing:
The auto index creation is experiencing issues with transactions that create more than one entities at a time. This fix is resolving that issue.


Darych (approved):

Josipmrden (approved):

kgolubic:
@gvolfing for RN I would use:

Resolved an issue in the automatic index creation process where transactions creating multiple entities simultaneously could lead to unexpected behaviors. This patch ensures reliable auto index creation when dealing with multiple entity creations in a single transaction.

Is that OK?

gvolfing:
@kgolubic Perfect.

PR #1930, opened 2024-04-11
install_deps.sh auto downloads and extracts toolchain archive

n0vakovic:
### Description

Extended install_deps.sh script to auto download and extract toolchain archive based on OS and architecture.
This is done by introducing "prepare TOOLCHAIN_RUN_DEPS" command. 

[master < Task] PR
    - **install_deps.sh auto downloads and extracts toolchain archive**


environment/os/install_deps.sh
```diff
@@ -18,6 +36,7 @@ Usage:
     ./install_deps.sh [command] [dependency_set]
 
 Commands:
+    prepare       Downloads and extracts specified dependencies.
```
MarkoBarisic, line 39: Add info that it can only be run with `TOOLCHAIN_RUN_DEPS`

MarkoBarisic (changes requested):

environment/os/install_deps.sh
n0vakovic: Done ✅
Added clarification to command help that only TOOLCHAIN_RUN_DEPS is supported for prepare

MarkoBarisic (approved):
LGTM

PR #1935, opened 2024-04-12
Count number of hops

DavIvek:
### Description
To be able to limit number of hops in query, firstly we need to count number of hops.
Firstly, we want to count number of hops in DFS, BFS and classic expands. Every unit of work is counted as hop, including filtering on edge types, edge properties or visiting already processed nodes.

[master < Task] PR
    - **[FINAL GIT MESSAGE]**

### Documentation checklist
    - **[Release note text]**
    - **[Documentation PR link]**


src/query/interpret/frame.hpp
```diff
@@ -1,4 +1,4 @@
-// Copyright 2022 Memgraph Ltd.
+// Copyright 2024 Memgraph Ltd.
```
gitbuda, line 1: This should be reverted because nothing else changed in the file 👀

src/query/plan/operator.cpp
```diff
@@ -366,6 +366,7 @@ EdgeAccessor CreateEdge(const EdgeCreationInfo &edge_info, DbAccessor *dba, Vert
                         Frame *frame, ExecutionContext &context, ExpressionEvaluator *evaluator) {
   auto maybe_edge = dba->InsertEdge(from, to, edge_info.edge_type);
   if (maybe_edge.HasValue()) {
+    context.number_of_hops++;
```
gitbuda, line 369: I would also skip this for now, only "read" expansions for now 🤔

gitbuda: But whatever we do here, it has to be document in the documentatio.

src/query/plan/operator.cpp
```diff
@@ -1168,14 +1173,18 @@ auto ExpandFromVertex(const VertexAccessor &vertex, EdgeAtom::Direction directio
       memory);
 
   if (direction != EdgeAtom::Direction::OUT) {
-    auto edges = UnwrapEdgesResult(vertex.InEdges(view, edge_types)).edges;
+    auto edges_result = UnwrapEdgesResult(vertex.InEdges(view, edge_types));
+    context->number_of_hops += edges_result.expanded_count;
+    auto edges = std::move(edges_result.edges);
```
gitbuda, line 1178: Redundant line, you can just `std::move(edges_result.edges)` two lines further, there are more cases like this in the current version of the code.

gitbuda (changes requested):
@DavIvek A few minor fixes required, but overall looks good (this is the right size of a PR and the right first step 💪)

src/query/context.hpp
```diff
@@ -11,6 +11,7 @@
 
 #pragma once
 
+#include <sys/types.h>
```
Josipmrden, line 14: do you need this one?

src/query/interpreter.cpp
```diff
@@ -2044,6 +2044,8 @@ std::optional<plan::ProfilingStatsWithTotalTime> PullPlan::Pull(AnyStream *strea
   }
 
   summary->insert_or_assign("plan_execution_time", execution_time_.count());
+  summary->insert_or_assign("number_of_hops", ctx_.number_of_hops);
+  spdlog::trace("Number of hops in the frame: {}", ctx_.number_of_hops);
```
Josipmrden, line 2048: i think this is a little bit aggressive. maybe to ommit this log

Josipmrden: we already have it in the summary

Josipmrden (commented):
some comments but LGTM

Josipmrden (approved):

gitbuda (approved):
all comments resolved -> 🚀

PR #1941, opened 2024-04-15
Improve COQ on HA

antoniofilipovic:
### Description

1. Separate replication flag from ha flag ✅
2. coordinator instance constructor ✅
3. unit test state manager ✅

[master < Task] PR
    - Improve COQ on HA

### Documentation checklist
    - Done
    - No RN
    - No docs


antoniofilipovic:
@kgolubic no docs here

src/coordination/coordinator_instance.cpp
```diff
@@ -115,6 +63,78 @@ CoordinatorInstance::CoordinatorInstance(CoordinatorInstanceInitConfig const &co
     auto &repl_instance = self->FindReplicationInstance(repl_instance_name);
     std::invoke(repl_instance.GetFailCallback(), self, repl_instance_name);
   };
+
+  raft_state_ = std::make_unique<RaftState>(config, GetBecomeLeaderCallback(), GetBecomeFollowerCallback());
+  raft_state_->InitRaftServer();
```
as51340, line 68: Why not Init call from the constructor instead of explicit call?

tests/unit/coordinator_state_manager.cpp
```diff
@@ -0,0 +1,100 @@
...
+    old_config->get_servers().push_back(temp_srv_config);
+    state_manager_->save_config(*old_config);
+    ASSERT_EQ(old_config->get_servers().size(), 2);
+  }
+
+  ptr<CoordinatorStateManager> state_manager_copy = cs_new<CoordinatorStateManager>(config);
+  auto loaded_config = state_manager_copy->load_config();
+
+  ASSERT_EQ(old_config->get_servers().size(), loaded_config->get_servers().size());
+  auto zipped_view = ranges::views::zip(old_config->get_servers(), loaded_config->get_servers());
```
as51340, line 94: Nice!

as51340 (approved):
Nice work! Just check the comment about init raft server

as51340:
@antoniofilipovic I realized that we haven't changed naming of show instances fields and we didn't add bolt socket address to the output. I thought we want to add this we this PR, we can do it in the another one no problem. What do you say?

src/coordination/coordinator_instance.cpp
antoniofilipovic: Because there is the problem of `InitServer` calling coordinator `BecomeLeader` callback while `raft_state_` unique_ptr is not constructed still due to how std::make_unique works. It first calls constructor and only then initializes pointer

antoniofilipovic:
> is we this PR, we can do it in the another one no problem. What do you sa

@as51340 I think we need to define that first, need to ping @gitbuda to define how show instances work and then we can change it. I think this is different user story for different PR, do you agree?

as51340:
I think we don't need to define anything, bolt server should be visible, otherwise users don't know how to connect. And fields have wrong names, that's all

antoniofilipovic:
> I think we don't need to define anything, bolt server should be visible, otherwise users don't know how to connect. And fields have wrong names, that's all

Works, I was referencing to this task so didn't add it here, we can push it in next PR: https://www.notion.so/memgraph/Code-of-Quality-COQ-2e9ba8803f13421db9cfb94ff6e620ff?pvs=4

PR #1951, opened 2024-04-19
Add env variables for HA k8 cluster setup

antoniofilipovic:
### Description

This diff adds an env option to start the coord instance/data instance with env variables.

TODO:
- [x] remove usage of FLAGS_management_port, FLAGS_coordinator_port, FLAGS_coordinator_id
- [x] add coordinator env test
- [x] add data instance env test
- [x] logs where we get value from
- [x] remove FLAGS_management_port checks from code
- [x] check bug on the empty start of registration when the instance is not available
- [x]  add parsing of env vars for spaces

[master < Task] PR
    - Add env variables for HA k8 cluster setup

### Documentation checklist
    - You can now set HA cluster with env variables
    - For docs: update env and also don't forget to update part of init script that we are now catching all exceptions


src/coordination/coordinator_state.cpp
```diff
@@ -36,6 +36,7 @@ CoordinatorState::CoordinatorState(ReplicationInstanceInitConfig const &config)
       .port = static_cast<uint16_t>(config.management_port),
   };
   data_ = CoordinatorMainReplicaData{.coordinator_server_ = std::make_unique<CoordinatorServer>(mgmt_config)};
+  spdlog::trace("Created coordinator server.");
```
as51340, line 39: Like it but expand so we can also find out on which coordinator server, bolt server etc. I would print whole config

src/coordination/include/coordination/coordinator_instance.hpp
```diff
@@ -76,6 +77,9 @@ class CoordinatorInstance {
  private:
   template <ranges::forward_range R>
   auto GetMostUpToDateInstanceFromHistories(R &&alive_instances) -> std::optional<std::string> {
+    if (ranges::empty(alive_instances)) {
```
as51340, line 80: Maybe the problem happens when we pass alive_instances in the function... Don't know, seems strange. This is nice solution so let's leave it like this

src/dbms/dbms_handler.cpp
```diff
@@ -46,7 +47,7 @@ std::string RegisterReplicaErrorToString(query::RegisterReplicaError error) {
 
 // Per storage
 // NOTE Storage will connect to all replicas. Future work might change this
-void RestoreReplication(replication::RoleMainData &mainData, DatabaseAccess db_acc) {
+void RestoreReplication(replication::RoleMainData &mainData, DatabaseAccess db_acc, bool is_coordinator_managed) {
```
as51340, line 50: I don't like this passing booleans into functions. This however is a bit different because we are just propagating it to the constructor of ReplicationStorageClient so I guess it is fine. But my question is why we wouldn't do this part in the ReplicationStorageClient constructor instead of complicating it here?

src/flags/coord_flag_env_handler.cpp
```diff
@@ -0,0 +1,59 @@
...
+namespace memgraph::flags {
+
+auto GetFinalCoordinationSetup() -> CoordinationSetup {
+  CoordinationSetup coordination_setup{};
+
+  auto *maybe_management_port = std::getenv(kMgManagementPort);
+  auto *maybe_coordinator_port = std::getenv(kMgCoordinatorPort);
+  auto *maybe_coordinator_id = std::getenv(kMgCoordinatorId);
+
+  bool are_envs_set = maybe_management_port || maybe_coordinator_port || maybe_coordinator_id;
```
as51340, line 30: const

src/flags/coord_flag_env_handler.cpp
```diff
@@ -0,0 +1,59 @@
...
+  bool are_flags_set = FLAGS_management_port || FLAGS_coordinator_port || FLAGS_coordinator_id;
```
as51340, line 31: const

src/flags/coord_flag_env_handler.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+  if (are_envs_set) {
+    coordination_setup.management_port = maybe_management_port ? std::stoi(maybe_management_port) : 0;
```
as51340, line 34: I think you can make yourself life easier by specifying constructor for CoordinatorSetup and then just passing one or other option

as51340: this is simple logic but the code looks more complex than necessary IMO

src/flags/coord_flag_env_handler.hpp
```diff
@@ -0,0 +1,27 @@
...
+namespace memgraph::flags {
+
+constexpr const char *kMgManagementPort = "MEMGRAPH_MANAGEMENT_PORT";
+constexpr const char *kMgCoordinatorPort = "MEMGRAPH_COORDINATOR_PORT";
+constexpr const char *kMgCoordinatorId = "MEMGRAPH_COORDINATOR_ID";
+
+struct CoordinationSetup {
+  int management_port;
+  int coordinator_port;
+  uint coordinator_id;
```
as51340, line 22: I think ports should be uint16_t? Can you double check with the rest of the code?

src/memgraph.cpp
```diff
@@ -84,11 +90,14 @@ void InitFromCypherlFile(memgraph::query::InterpreterContext &ctx, memgraph::dbm
...
+        spdlog::trace("Executing line: {}", line);
         auto results = interpreter.Prepare(line, {}, {});
         memgraph::query::DiscardValueResultStream stream;
         interpreter.Pull(&stream, {}, results.qid);
-      } catch (const memgraph::query::UserAlreadyExistsException &e) {
-        spdlog::warn("{} The rest of the init-file will be run.", e.what());
+      } catch (std::exception const &e) {
+        spdlog::warn("Exception occurred while executing one line. The rest of the init-file will be run. {}",
+                     e.what());
       }
```
as51340, line 101: This is suspicious. I think I did the same try-catch in one PR but @andrejtonev had some comment that we are catching things we shouldn't. Here if you catch exception, you will continue to execute other lines but is that expected behaviour? Not sue to me it would also seem logical to abort any other commands once the failure occurs

src/memgraph.cpp
```diff
@@ -437,34 +452,62 @@ int main(int argc, char **argv) {
 
   auto system = memgraph::system::System{db_config.durability.storage_directory, FLAGS_data_recovery_on_startup};
 
+  auto const coordination_setup = memgraph::flags::GetFinalCoordinationSetup();
   // singleton replication state
-  memgraph::replication::ReplicationState repl_state{ReplicationStateRootPath(db_config)};
+  memgraph::replication::ReplicationState repl_state{
+      ReplicationStateRootPath(db_config, coordination_setup.management_port != 0),
+      coordination_setup.management_port != 0};
```
as51340, line 459: this sending coordination_setup.mgmt_port != 0 at multiple places doesn't look best

src/memgraph.cpp
```diff
@@ -437,34 +452,62 @@ int main(int argc, char **argv) {
...
+
+  int extracted_bolt_port{0};
```
as51340, line 461: Use IILE

src/memgraph.cpp
```diff
@@ -437,34 +452,62 @@ int main(int argc, char **argv) {
...
-        "port for coordinator instance and management port for replication instance.");
-  }
 
-  if (FLAGS_coordinator_id && FLAGS_coordinator_port) {
-    try {
+  auto try_init_coord_state = [&coordinator_state,
+                               &extracted_bolt_port](memgraph::flags::CoordinationSetup const &coordination_setup) {
+    if (!(coordination_setup.management_port || coordination_setup.coordinator_port ||
+          coordination_setup.coordinator_id)) {
+      spdlog::trace("Skipping coordinator initialization.");
```
as51340, line 479: I think skipping is not the correct verb, some failure ?

src/mg_import_csv.cpp
```diff
@@ -717,7 +717,7 @@ int main(int argc, char *argv[]) {
                      .snapshot_wal_mode = memgraph::storage::Config::Durability::SnapshotWalMode::DISABLED,
                      .snapshot_on_exit = true},
       .salient = {.items = {.properties_on_edges = FLAGS_storage_properties_on_edges}}};
-  memgraph::replication::ReplicationState repl_state{memgraph::storage::ReplicationStateRootPath(config)};
+  memgraph::replication::ReplicationState repl_state{memgraph::storage::ReplicationStateRootPath(config), false};
```
as51340, line 720: when we pass here false it is impossible to know what it refers to. If you really need to have bool arg here, at least make it by default false

src/query/interpreter.cpp
```diff
@@ -2890,14 +2876,14 @@ PreparedQuery PrepareAuthQuery(ParsedQuery parsed_query, bool in_explicit_transa
 PreparedQuery PrepareReplicationQuery(ParsedQuery parsed_query, bool in_explicit_transaction,
                                       std::vector<Notification> *notifications,
                                       ReplicationQueryHandler &replication_query_handler, CurrentDB &current_db,
-                                      const InterpreterConfig &config) {
+                                      const InterpreterConfig &config, bool is_coordinator_managed) {
```
as51340, line 2879: Again how can we avoid this?

as51340 (changes requested):

as51340:
In general, the logic seems correct. I don't like few things which you will realize from the comments.

src/flags/coord_flag_env_handler.cpp
antoniofilipovic: I agree that it looks complex, I think I fixed it now

src/dbms/dbms_handler.cpp
antoniofilipovic: I don't like it too, I organized it differently now, we have singleton object which we can access to check if instance is coordinator managed depending on flags/env

src/memgraph.cpp
antoniofilipovic: Agree, resolved now

src/memgraph.cpp
antoniofilipovic: nice

src/mg_import_csv.cpp
antoniofilipovic: fixed

src/flags/coord_flag_env_handler.hpp
antoniofilipovic: it is uint32_t so I changed it to that

src/memgraph.cpp
antoniofilipovic: Try catch was here before if you check diff, but only for specific exceptions. Now we can throw some of the coordinator exceptions. I can make it specific for only coord exceptions. In my opinion it makes sense to abort completely but I didn't want to disrupt logic from before, don't know if some of the customers are relying on it

src/flags/coord_flag_env_handler.cpp
```diff
@@ -0,0 +1,59 @@
...
+      spdlog::trace("Read coordinator setup from env variables: {}.", CoordinationSetupInstance().ToString());
+      return CoordinationSetup(maybe_management_port ? std::stoi(maybe_management_port) : 0,
+                               maybe_coordinator_port ? std::stoi(maybe_coordinator_port) : 0,
+                               maybe_coordinator_id ? static_cast<uint32_t>(std::stoul(maybe_coordinator_id)) : 0);
+    }
+    spdlog::trace("Read coordinator setup from runtime flags {}.", CoordinationSetupInstance().ToString());
+    return CoordinationSetup{FLAGS_management_port, FLAGS_coordinator_port, FLAGS_coordinator_id};
+  }();  // iile
+
+  return CoordinationSetupInstance();
```
as51340, line 56: Why is this named same as function `CoordinationSetupInstance`?

src/memgraph.cpp
as51340: Check this PR https://github.com/memgraph/memgraph/pull/1465

as51340: comments from pr

src/memgraph.cpp
```diff
@@ -524,6 +566,12 @@ int main(int argc, char **argv) {
 #endif
   }
 
+  auto *maybe_ha_init_file = std::getenv(kMgHaClusterInitQueries);
+  if (maybe_ha_init_file) {
+    spdlog::trace("initializing coordinator from cypher file.");
```
as51340, line 571:
```suggestion
    spdlog::trace("Initializing coordinator from cypher file.");
```

src/memgraph.cpp
```diff
@@ -524,6 +566,12 @@ int main(int argc, char **argv) {
 #endif
   }
 
+  auto *maybe_ha_init_file = std::getenv(kMgHaClusterInitQueries);
```
as51340, line 569: Maybe just write comment that this must be done after creating `CoordinatorState` so someone doesn't make a mistake when refactoring

src/query/interpreter.cpp
```diff
@@ -1004,15 +1004,12 @@ Callback HandleReplicationQuery(ReplicationQuery *repl_query, const Parameters &
   evaluation_context.timestamp = QueryTimestamp();
   evaluation_context.parameters = parameters;
   auto evaluator = PrimitiveLiteralExpressionEvaluator{evaluation_context};
-
+  auto const is_coordinator_managed = flags::CoordinationSetupInstance().IsCoordinatorManaged();
```
as51340, line 1007: Can we rename IsCoordinatorManaged to IsManagedByCoordinator? 
I have a feeling that I am reading if coordinator is being managed instead of something else

src/query/interpreter.cpp
```diff
@@ -1004,15 +1004,12 @@ Callback HandleReplicationQuery(ReplicationQuery *repl_query, const Parameters &
...
   Callback callback;
   switch (repl_query->action_) {
     case ReplicationQuery::Action::SET_REPLICATION_ROLE: {
 #ifdef MG_ENTERPRISE
-      if (FLAGS_coordinator_id) {
-        throw QueryRuntimeException("Coordinator can't set roles!");
-      }
-      if (FLAGS_management_port) {
+      if (is_coordinator_managed) {
```
as51340, line 1012: Why are we using 2 checks:
1. is_coordinator_managed
2. IsCoordinator and IsDataInstances

Why not rely just on one, for example check IsCoordinator

tests/e2e/memgraph.py
```diff
@@ -124,7 +125,10 @@ def start(self, restart=False, args=None, setup_queries=None):
             "300",
             "--storage-properties-on-edges",
         ] + self.args
-        self.bolt_port = extract_bolt_port(args_mg)
+        if bolt_port:
+            self.bolt_port = bolt_port
+        else:
+            self.bolt_port = extract_bolt_port(args_mg)
         self.proc_mg = subprocess.Popen(args_mg)
```
as51340, line 132: Does it make sense to have both here if else which calls extract_bolt_port and in the caller? Maybe just choose between caller and callee (I would do in callee logic, I think is generally better)

as51340 (changes requested):
Much better :muscle:

src/flags/coord_flag_env_handler.cpp
antoniofilipovic: resolved on call

src/query/interpreter.cpp
antoniofilipovic: fix by sending coordinator state

as51340 (approved):
:muscle:

PR #1953, opened 2024-04-22
Remove misc-include-cleaner from Clang-Tidy

as51340:
### Description

Remove misc-include-cleaner from Clang-Tidy

[master < Task] PR
    - **Remove misc-include-cleaner from Clang-Tidy**

### CI Testing Labels

### Documentation checklist
    - **[Release note text]**
    - **[Documentation PR link]**


as51340:
@kgolubic nothing to be documented here, internal tool improvement.

antoniofilipovic (commented):
I am not sure if this is good practice to remove it, can we include @Ignition in this?

antoniofilipovic:
> I am not sure if this is good practice to remove it, can we include @Ignition in this?

In general this is what I found on internet about this:
```
The point of IWYU (include what you use - misc-include-cleaner) has nothing to do with decreasing compile times. It's meant to ensure that a .cpp file includes all of its dependencies directly rather than relying on one .h file to include another .h file, to include another .h file, etc. The only reason why IWYU could theoretically reduce compile times is if you had previously been including big .h files for no reason, other then for the indirect dependency.
```
I think it is good reason to keep it around

as51340:
Yes but our codebase isn't written with this in mind. Not every file includes everything else directly, check. Imagine that interpreter.cpp has to include directly every other dependency. I had the bug this morning on CI because interpreter.cpp didn't include directly coordination/register_main_replica_status.hpp which is stupid. 

We can ask @Ignition for opinion yes but I don't think this rule should be enforced

Ignition (approved):

Ignition:
For the record, I've changed my mind on this. The above comment by me is wrong.
`misc-include-cleaner` only came in because we updated our toolchain. Our clang-tidy is not stable, we currently do opt-out `*,-thing` rather than opt-in `-*,thing`. Which means on updating clang from 13 to 17 we implicitly gained many new clang-tidy checks. We should make a list of these and check if we agree/disagree with them for our codebase.

PR #1964, opened 2024-04-22
Implement EdgeType + Property indices

gvolfing:
Source issue -> https://github.com/memgraph/memgraph/issues/662

### Description

Implementation of EdgeType + Property indices.

[master < Task] PR
    - Add support for edge type property index

### CI Testing Labels

### Documentation checklist
    - Add support for edge type property index
    - https://github.com/memgraph/documentation/pull/855


tests/unit/query_plan_checker.hpp
```diff
@@ -165,6 +166,8 @@ class OpChecker : public BaseOpChecker {
  public:
   void CheckOp(LogicalOperator &op, const SymbolTable &symbol_table) override {
     auto *expected_op = dynamic_cast<TOp *>(&op);
+    auto debug_op_typeinfo = op.GetTypeInfo().name;
+    auto debug_expected_typeinfo = TOp::kType.name;
```
Darych, line 170: Do we still need these 2 variables? If yes, why we don't use them?

tests/unit/storage_v2_durability_inmemory.cpp
```diff
@@ -3046,6 +3099,89 @@ TEST_P(DurabilityTest, EdgeTypeIndexRecovered) {
...
+    auto acc = db.Access();
+    auto vertex = acc->CreateVertex();
+    auto edge = acc->CreateEdge(&vertex, &vertex, db.storage()->NameToEdgeType("et"));
+    ASSERT_TRUE(edge.HasValue());
+    ASSERT_FALSE(acc->Commit().HasError());
+  }
+}
+
+// // NOLINTNEXTLINE(hicpp-special-member-functions)
+// TEST_P(DurabilityTest, EdgeTypePropertyIndexRecoveredWithoutEdgeTypeIndices) {
```
Darych, line 3144: Do we plan to uncomment this?

tests/unit/storage_v2_indices.cpp
```diff
@@ -1673,3 +1677,405 @@ TYPED_TEST(IndexTest, EdgeTypeIndexRepeatingEdgeTypesBetweenSameVertices) {
...
+  for (int i = 0; i < 5; ++i) {
+    auto edge_acc = this->CreateEdge(&vertex_from, &vertex_to, this->edge_type_id1, acc.get());
+    edge_acc.SetProperty(this->edge_prop_id1, memgraph::storage::PropertyValue(i));
+  }
+
+  EXPECT_EQ(acc->ApproximateEdgeCount(this->edge_type_id1, this->edge_prop_id1), 5);
+
+  EXPECT_THAT(this->GetIds(acc->Edges(this->edge_type_id1, this->edge_prop_id1, View::NEW), View::NEW),
+              UnorderedElementsAre(0, 1, 2, 3, 4));
+}
```
Darych, line 2081: I think we also need to add changing property value on indexed edge type + property to test UpdateOnModifiedEdge

src/storage/v2/durability/wal.cpp
```diff
@@ -369,6 +377,20 @@ WalDeltaData ReadSkipWalDeltaData(BaseDecoder *decoder) {
...
       break;
     }
+    case WalDeltaData::Type::EDGE_PROPERTY_INDEX_CREATE:
+    case WalDeltaData::Type::EDGE_PROPERTY_INDEX_DROP: {
+      if constexpr (read_data) {
+        auto edge_type = decoder->ReadString();
+        if (!edge_type) throw RecoveryFailure("Invalid WAL data!");
+        delta.operation_edge_type_property.edge_type = std::move(*edge_type);
+        auto property = decoder->ReadString();
+        if (!property) throw RecoveryFailure("Invalid WAL data!");
```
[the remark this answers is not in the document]
Darych: Should be addressed in another PR for all types of delta.

kgolubic:
@Darych I'm missing the Release note entry and documentation PR.

src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -283,12 +283,20 @@ antlrcpp::Any CypherMainVisitor::visitCreateEdgeIndex(MemgraphCypher::CreateEdge
   auto *index_query = storage_->Create<EdgeIndexQuery>();
   index_query->action_ = EdgeIndexQuery::Action::CREATE;
   index_query->edge_type_ = AddEdgeType(std::any_cast<std::string>(ctx->labelName()->accept(this)));
+  if (ctx->propertyKeyName()) {
+    auto name_key = std::any_cast<PropertyIx>(ctx->propertyKeyName()->accept(this));
```
antoniofilipovic, line 287: name_key can be const

src/query/plan/rewrite/edge_type_index_lookup.hpp
```diff
@@ -555,16 +595,31 @@ class EdgeTypeIndexRewriter final : public HierarchicalLogicalOperatorVisitor {
...
     }
 
-    LOG_FATAL("Fatal error while rewriting query plan.");
+    if (EdgeTypePropertyIndexingPossible()) {
+      // This is currently the simplest way the get the most out of edge-type + proeprty indexing.
+      // We only remove the filter if we are looking for edges with a specific edge-type. If we
+      // have some other type of property filtering e.g. range or value lookup, we do not want to
+      // remove the filters to keep the semantics of the original query in tackt. We can optimize
+      // this by adding additional operators for the specific kinds of edge-type+property filtering.
+      // On top of this we only want to remove the filter if when?????
```
antoniofilipovic, line 608: What is with this part of comment?

src/query/plan/rewrite/edge_type_index_lookup.hpp
```diff
@@ -555,16 +595,31 @@ class EdgeTypeIndexRewriter final : public HierarchicalLogicalOperatorVisitor {
...
     if (maybe_id_lookup_value_) {
       return std::make_unique<ScanAllByEdgeId>(input, output_symbol, maybe_id_lookup_value_, view);
     }
 
-    LOG_FATAL("Fatal error while rewriting query plan.");
+    if (EdgeTypePropertyIndexingPossible()) {
+      // This is currently the simplest way the get the most out of edge-type + proeprty indexing.
+      // We only remove the filter if we are looking for edges with a specific edge-type. If we
+      // have some other type of property filtering e.g. range or value lookup, we do not want to
+      // remove the filters to keep the semantics of the original query in tackt. We can optimize
```
antoniofilipovic, line 606:
```suggestion
      // remove the filters to keep the semantics of the original query intact. We can optimize
```

src/query/plan/operator.cpp
```diff
@@ -677,6 +678,31 @@ std::vector<Symbol> ScanAllByEdgeType::ModifiedSymbols(const SymbolTable &table)
...
+                                                     Symbol output_symbol, storage::EdgeTypeId edge_type,
+                                                     storage::PropertyId property, storage::View view)
+    : ScanAll(input, output_symbol, view), edge_type_(edge_type), property_(property) {}
+
+ACCEPT_WITH_INPUT(ScanAllByEdgeTypeProperty)
+
+UniqueCursorPtr ScanAllByEdgeTypeProperty::MakeCursor(utils::MemoryResource *mem) const {
+  memgraph::metrics::IncrementCounter(memgraph::metrics::ScanAllByEdgeTypeOperator);
+
+  auto edges = [this](Frame &, ExecutionContext &context) {
```
antoniofilipovic, line 691: better name would be get_edges

antoniofilipovic: I think edges can be `const`

src/query/interpreter.cpp
```diff
@@ -2874,20 +2875,45 @@ PreparedQuery PrepareEdgeIndexQuery(ParsedQuery parsed_query, bool in_explicit_t
...
   auto edge_type = storage->NameToEdgeType(index_query->edge_type_.name);
 
+  std::vector<storage::PropertyId> properties;
+  std::vector<std::string> properties_string;
+  properties.reserve(index_query->properties_.size());
+  properties_string.reserve(index_query->properties_.size());
+  for (const auto &prop : index_query->properties_) {
+    properties.push_back(storage->NameToProperty(prop.name));
+    properties_string.push_back(prop.name);
+  }
```
antoniofilipovic, line 2885: Can we check these properties in cypher main visitor? I would say we should check for number of properties already there

src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -283,12 +283,20 @@ antlrcpp::Any CypherMainVisitor::visitCreateEdgeIndex(MemgraphCypher::CreateEdge
...
+    index_query->properties_ = {name_key};
```
antoniofilipovic, line 288: Why properties are std::vector instead of only single property?

src/query/interpreter.cpp
```diff
@@ -2874,20 +2875,45 @@ PreparedQuery PrepareEdgeIndexQuery(ParsedQuery parsed_query, bool in_explicit_t
...
+                                               index_query->edge_type_.name, ix_properties.front().name);
+      }
 
       handler = [dba, edge_type, label_name = index_query->edge_type_.name,
+                 properties_stringified = std::move(properties_stringified), properties = std::move(properties),
                  invalidate_plan_cache = std::move(invalidate_plan_cache)](Notification &index_notification) {
-        auto maybe_index_error = dba->CreateIndex(edge_type);
+        MG_ASSERT(properties.size() <= 1U);
+        auto maybe_index_error =
+            properties.empty() ? dba->CreateIndex(edge_type) : dba->CreateIndex(edge_type, properties[0]);
```
antoniofilipovic, line 2910: Here we can only have one property instead of vector

src/storage/v2/inmemory/edge_type_property_index.cpp
```diff
@@ -0,0 +1,332 @@
...
+        const auto type = std::get<kEdgeTypeIdPos>(edge);
+        if (type == edge_type) {
+          auto *to_vertex = std::get<kVertexPos>(edge);
+          if (to_vertex->deleted) {
+            continue;
+          }
+          auto *edge_ptr = std::get<kEdgeRefPos>(edge).ptr;
+          edge_acc.insert({edge_ptr->properties.GetProperty(property), &from_vertex, to_vertex, edge_ptr, 0});
+        }
+      }
```
antoniofilipovic, line 121:
```suggestion
        if (type != edge_type) {
            continue
        }
        auto *to_vertex = std::get<kVertexPos>(edge);
        if (to_vertex->deleted) {
            continue;
        }
        auto *edge_ptr = std::get<kEdgeRefPos>(edge).ptr;
        edge_acc.insert({edge_ptr->properties.GetProperty(property), &from_vertex, to_vertex, edge_ptr, 0});
    }
      }
```

src/storage/v2/inmemory/edge_type_property_index.cpp
```diff
@@ -0,0 +1,332 @@
...
+    return false;
+  }
+
+  utils::MemoryTracker::OutOfMemoryExceptionEnabler oom_exception;
+  try {
+    auto edge_acc = it->second.access();
+    for (auto &from_vertex : vertices) {
+      if (from_vertex.deleted) {
+        continue;
+      }
```
antoniofilipovic, line 109: During indexing there can be no other operation in database, that holds still, right?

src/query/interpreter.cpp
```diff
@@ -2874,20 +2875,45 @@ PreparedQuery PrepareEdgeIndexQuery(ParsedQuery parsed_query, bool in_explicit_t
...
         utils::OnScopeExit invalidator(invalidate_plan_cache);
 
         if (maybe_index_error.HasError()) {
           index_notification.code = NotificationCode::EXISTENT_INDEX;
-          index_notification.title = fmt::format("Index on edge-type {} already exists.", label_name);
+          index_notification.title =
```
antoniofilipovic, line 2915: I think it is safer to do switch-case because in that case if someone extends with other errors we will not miss them, or return them in wrong way to user

antoniofilipovic (commented):
I need to check few more stuff, this is first batch of comments

src/storage/v2/inmemory/edge_type_property_index.cpp
```diff
@@ -0,0 +1,332 @@
...
+
+namespace memgraph::storage {
+
+bool InMemoryEdgeTypePropertyIndex::Entry::operator<(const Entry &rhs) const {
+  if (edge->gid < rhs.edge->gid) {
+    return true;
+  }
+  if (rhs.edge->gid < edge->gid) {
+    return false;
+  }
```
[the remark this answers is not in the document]
Darych: We already checked `edge->gid < rhs.edge->gid` in previous comparison. But I think this can be rewritten as `edge->gid > rhs.edge->gid` to be more clear.
In the case of edge gids equality we will compare tuples.

src/query/frontend/ast/cypher_main_visitor.cpp
Darych: The same thing is done for label-property indices. It could be probably preparation for indices with 2 or more properties.

src/query/interpreter.cpp
Darych: This is done the same way as for label-property index. If we are not going to add more properties to indices, I think we can simplify our codebase for several indices.

src/query/interpreter.cpp
Darych: Do you mean switch case by `maybe_index_error.GetError()`?

src/query/interpreter.cpp
Darych: Agree. But this is done the same way as for label-property index. The same question about our plans can be applied.

src/storage/v2/inmemory/edge_type_property_index.cpp
Darych: Yes, we take a unique lock in https://github.com/memgraph/memgraph/blob/impl-edge-type-property-indices/src/storage/v2/inmemory/storage.cpp#L1467

src/query/interpreter.cpp
antoniofilipovic: yes

src/storage/v2/inmemory/edge_type_property_index.cpp
```diff
@@ -0,0 +1,333 @@
...
+    return;
+  }
+  auto acc = it->second.access();
+
+  auto entry_to_update = std::ranges::find_if(acc, [&](const auto &entry) {
+    return entry.from_vertex == old_from && entry.to_vertex == old_to && entry.edge == edge_ref.ptr;
+  });
+  const auto value = entry_to_update->value;
+  acc.remove(Entry{value, entry_to_update->from_vertex, entry_to_update->to_vertex, entry_to_update->edge,
+                   entry_to_update->timestamp});
```
antoniofilipovic, line 228: I think we shouldn't remove here this entry, what if some other transaction is using that index, and that edge is still visible for it?

src/storage/v2/inmemory/edge_type_property_index.cpp
```diff
@@ -0,0 +1,333 @@
...
+  const auto &from_edges = from_vertex->out_edges;
+  const auto &to_edges = to_vertex->in_edges;
+
+  auto it = std::find_if(from_edges.begin(), from_edges.end(), [&](const auto &from_entry) {
+    const auto &from_edge = std::get<kEdgeRefPos>(from_entry);
+    return std::any_of(to_edges.begin(), to_edges.end(), [&](const auto &to_entry) {
+      const auto &to_edge = std::get<kEdgeRefPos>(to_entry);
+      return index_iterator_->edge->gid == from_edge.ptr->gid && from_edge.ptr->gid == to_edge.ptr->gid;
+    });
+  });
```
antoniofilipovic, line 308: Why do we have here O(n^2) complexity?

Shouldn't we just iterate over all from_edges and over all to_edges and then just check they are same? This should be O(N) complexity in general

antoniofilipovic (changes requested):
Let me know if we need to sync tomorrow, I am generally not sure for couple of things

src/query/plan/rewrite/edge_type_index_lookup.hpp
```diff
@@ -555,16 +595,31 @@ class EdgeTypeIndexRewriter final : public HierarchicalLogicalOperatorVisitor {
...
+      }
+
+      return std::make_unique<ScanAllByEdgeTypeProperty>(input, output_symbol, edge_type_, *property_, view);
+    }
+
+    if (EdgeTypeIndexingPossible()) {
+      return std::make_unique<ScanAllByEdgeType>(input, output_symbol, edge_type_, view);
+    }
+
+    return nullptr;
```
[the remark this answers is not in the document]
Darych: I agree with you. I wanted to dive deeper into this on next PR. If you think this is crucial I can jump on this in this PR.

src/storage/v2/inmemory/edge_type_property_index.cpp
Darych: Agree. Since we are under guards, maybe we can update the value via find + updating fields using iterator?
I checked how do we do with label properties indices, it looks like we postpone cleaning operation and collect all garbage later.

src/storage/v2/inmemory/edge_type_property_index.cpp
Darych: I agree. Rewrote it

src/query/interpreter.cpp
Darych: I checked and decided to keep it as it is for now. Storage errors are empty structs and not enums, so you can't switch now on them. Need to change all storage errors to be enums.

src/query/plan/rewrite/edge_type_index_lookup.hpp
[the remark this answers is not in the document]
antoniofilipovic: You can jump on it in next PR, it is fine to me.

src/storage/v2/inmemory/edge_type_property_index.cpp
antoniofilipovic: I wouldn't update fields in the entry, as maybe some other transaction will look for that entry. Correct thing would be to insert new entry just

Darych: But in this case we need to collect garbage, right? Hm, will think how to improve there and check how do we do this for label-property indices.

antoniofilipovic: Not yet, what if this transaction that inserted index aborts? Then old entry is correct one, right

antoniofilipovic: resolved on call

antoniofilipovic (approved):

PR #1991, opened 2024-04-30
Replicate only durable commits to replicas during recovery

as51340:
### Description

Send only durable part of the buffer to replicas during recovery
Better checker for jepsen test for easier debugging.
Improved logs to more easily reason about applying deltas during recovery.

Before this PR, even with `fsync` set to 1, replicas could get non-durable data from main. If main would die and get back, replica could be in more advanced state than main = issue.

[master < Task] PR
    - **Replica only durable commits to replicas during recovery**

### Documentation checklist
    - **During recovery process, when replicas are behind main, they will receive only commits which are already fsynced on main.**
    - **[Documentation PR link]**

### DIFF
--- a/src/storage/v2/inmemory/replication/recovery.cpp
+++ b/src/storage/v2/inmemory/replication/recovery.cpp
@@ -98,10 +98,8 @@ uint64_t ReplicateCurrentWal(const utils::UUID &main_uuid, const InMemoryStorage
   stream.AppendFilename(wal_file.Path().filename());
   utils::InputFile file;
   MG_ASSERT(file.Open(wal_file.Path()), "Failed to open current WAL file at {}!", wal_file.Path());
-  const auto [buffer, buffer_size] = wal_file.CurrentFileBuffer();
-  stream.AppendSize(file.GetSize() + buffer_size);
+  stream.AppendSize(file.GetSize());
   stream.AppendFileData(&file);
-  stream.AppendBufferData(buffer, buffer_size);
   auto response = stream.Finalize();
   return response.current_commit_timestamp;
 }
--- a/src/storage/v2/replication/replication_client.cpp
+++ b/src/storage/v2/replication/replication_client.cpp
@@ -9,19 +9,20 @@
 // by the Apache License, Version 2.0, included in the file
 // licenses/APL.txt.
 
-#include <algorithm>
+#include "replication/replication_client.hpp"
 
 #include "flags/coord_flag_env_handler.hpp"
-#include "replication/replication_client.hpp"
 #include "storage/v2/inmemory/storage.hpp"
 #include "storage/v2/replication/enums.hpp"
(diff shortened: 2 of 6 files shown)


src/utils/file.cpp
```diff
@@ -544,7 +544,9 @@ void OutputFile::EnableFlushing() {
   TryFlushing();
 }
 
-std::pair<const uint8_t *, size_t> OutputFile::CurrentBuffer() const { return {buffer_, buffer_position_.load()}; }
+std::pair<const uint8_t *, size_t> OutputFile::DurableBuffer() const {
+  return {buffer_, buffer_position_.load() - written_since_last_sync_};
```
Ignition, line 548: Not safe, `buffer_position_` and `written_since_last_sync_` are not atomically written together/read together. Just maintain a new position which is updated only on sync.

src/storage/v2/inmemory/replication/recovery.cpp
```diff
@@ -98,7 +98,7 @@ uint64_t ReplicateCurrentWal(const utils::UUID &main_uuid, const InMemoryStorage
   stream.AppendFilename(wal_file.Path().filename());
   utils::InputFile file;
   MG_ASSERT(file.Open(wal_file.Path()), "Failed to open current WAL file at {}!", wal_file.Path());
-  const auto [buffer, buffer_size] = wal_file.CurrentFileBuffer();
+  const auto [buffer, buffer_size] = wal_file.DurableBuffer();  // Send only the fsynced part of the buffer.
```
Ignition, line 101: Why write the buffer at all? stream.AppendFileData(&file); would contain all that is durable

src/storage/v2/replication/replication_client.cpp
```diff
@@ -305,12 +305,12 @@ void ReplicationStorageClient::RecoverReplica(uint64_t replica_commit, memgraph:
         std::visit(
             utils::Overloaded{
                 [&replica_commit, mem_storage, &rpcClient, main_uuid = main_uuid_](RecoverySnapshot const &snapshot) {
-                  spdlog::debug("Sending the latest snapshot file: {}", snapshot);
+                  spdlog::debug("Sending the latest snapshot file: {} to {}", snapshot, rpcClient.Endpoint());
```
antoniofilipovic, line 308: I think we mentioned we would log here also the replica name, right? Maybe not here but in the line 293 where we write "Starting replica recovery"

antoniofilipovic (approved):
All good to me

antoniofilipovic:
Maybe also  comment about the PR name, we are sending only durable commits to replica in recovery, the buffer is not sent at all

src/storage/v2/replication/replication_client.cpp
as51340: I think any info is good. How can I fetch replica name?

as51340:
@kgolubic RN above, docs written

Issue #1995, opened 2024-05-02
Replica going OOM


katarinasupe:
[**Source on Discord**](https://discord.com/channels/842007348272169002/1234877974042382348/1235278868760367134)
 
**Memgraph version**
2.16

**Environment**
Some information about the environment you are using Memgraph on: operating
system, architecture (ARM, x86), how do you connect, with or without docker,
which driver etc.

**Describe the bug**
After the connection is established using the IP route, the replica is holding on to too many deltas or something else in memory. The memory footprint on the replica keeps increasing until it OOMs even though the main pod is relatively stable:

![image](https://github.com/memgraph/memgraph/assets/61758502/d72d2be0-a687-4b50-a790-464815b10b83)

The CPU creeps on the replica pod as well:
![image](https://github.com/memgraph/memgraph/assets/61758502/6c516f88-4a48-4f34-9523-bce2dbd8b953)

The main instance of memgraph is constantly updating / upserting data into the graph (<100k total ents/rels). We have a TTL process that runs in the background to clear stale data out of the graph.

PR #2003, opened 2024-05-03
Bugfix: Replica now handles skipped transactions

Ignition:
### Description

On REPLICA the commit timestamp is calculated in an interesting way. This is for reasons related to durability based on the timestamp. The problem is that some timestamps can be skipped, and hence never used on the replica side and hence never finalised. This would mean the commit log would provide incorrect information to GC about what the oldest active timestamp is. Causing GC not to clear away the Deltas being made.

Fix is to make all skipped timestamps to be marked as finalised.

Another memory related issue was the RPC session buffer retained the size of its high watermark. Post Session::Execute we now limit the buffer back to a max of 4MiB.

fixes: #1995

[master < Task] PR
    - **Bugfix: Replica now handles skipped transactions**

### Documentation checklist
    - **Replica now correctly handles skipped timestamps, resulting in the memory used for Deltas being correctly released**
    - **[Documentation PR link]**

### DIFF
--- a/src/rpc/protocol.cpp
+++ b/src/rpc/protocol.cpp
@@ -23,6 +23,8 @@
 
 namespace memgraph::rpc {
 
+constexpr auto kBufferRetainLimit = 4 * 1024 * 1024;  // 4MiB
+
 Session::Session(Server *server, io::network::Endpoint endpoint, communication::InputStream *input_stream,
                  communication::OutputStream *output_stream)
     : server_(server), endpoint_(std::move(endpoint)), input_stream_(input_stream), output_stream_(output_stream) {}
@@ -37,7 +39,10 @@ void Session::Execute() {
   }
 
   // Remove the data from the stream on scope exit.
-  utils::OnScopeExit shift_data([&, ret] { input_stream_->Shift(ret.stream_size); });
+  auto const shift_data = utils::OnScopeExit{[&, ret] {
+    input_stream_->Shift(ret.stream_size);
+    input_stream_->ShrinkBuffer(kBufferRetainLimit);
+  }};
 
   // Prepare SLK reader and builder.
   slk::Reader req_reader(input_stream_->data(), input_stream_->size());
--- a/src/storage/v2/inmemory/storage.cpp
+++ b/src/storage/v2/inmemory/storage.cpp
@@ -35,6 +35,7 @@
 #include "utils/stat.hpp"
 
 #include <mutex>
+#include <ranges>
 
 namespace memgraph::storage {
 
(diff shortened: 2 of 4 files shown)


Ignition:
@kgolubic tagging you here

kgolubic:
@Ignition I would use this for RN:

Addressed a flaw where skipped timestamps were not finalized on replicas,  leading to inaccuracies in the oldest active timestamp information provided to the GC. This fix ensures all skipped timestamps are now marked as finalized, allowing the GC to accurately manage memory and clear Deltas as needed. Additionally, the RPC session buffer now resets to a maximum of 4 MiB post-session execution, improving memory management.

src/rpc/protocol.cpp
```diff
@@ -37,7 +39,10 @@ void Session::Execute() {
   }
 
   // Remove the data from the stream on scope exit.
-  utils::OnScopeExit shift_data([&, ret] { input_stream_->Shift(ret.stream_size); });
+  auto const shift_data = utils::OnScopeExit{[&, ret] {
+    input_stream_->Shift(ret.stream_size);
+    input_stream_->ShrinkBuffer(kBufferRetainLimit);
+  }};
```
antoniofilipovic, line 45: I didn't go to nitty-gritty details here but ret.stream_size can be bigger than 4MB when getting data via stream?

src/storage/v2/inmemory/storage.cpp
```diff
@@ -2394,10 +2396,18 @@ void InMemoryStorage::FreeMemory(std::unique_lock<utils::ResourceLock> main_guar
...
+  // Normal logic, when txn is on MAIN
   if (!desired_commit_timestamp) {
     return timestamp_++;
   }
-  timestamp_ = std::max(timestamp_, *desired_commit_timestamp + 1);
+  // Special case logic, when txn is on REPLICA
+  auto normal_next_timestamp = timestamp_;
+  // any timestamps which are to be skipped need to be marked as finished
+  // otherwise GC would stop working correctly
+  for (auto used_timestamp : std::ranges::views::iota(normal_next_timestamp, *desired_commit_timestamp)) {
```
antoniofilipovic, line 2407:
```suggestion
  for (auto const used_timestamp : std::ranges::views::iota(normal_next_timestamp, *desired_commit_timestamp)) {
```

antoniofilipovic (approved):
Awesome!

gvolfing (approved):

src/rpc/protocol.cpp
Ignition: This is OnScopeExit straight after the Shift that has remove exactly `ret.stream_size` from the start of the buffer. Shift would AFAICT should have reset stream back to have 0. `ShrinkBuffer` will be to the max of kBufferRetainLimit and have, likely kBufferRetainLimit, and adjust accordingly. 
I have gone back and changed the implementation in order to only copy `have_` amount upon resize.

PR #2018, opened 2024-05-09
Fix: MVCC design on replica and replica stream isolation

as51340:
### Description

Jepsen found a few bugs in replication, this PR fixes two of them.

When the replica recovered completely, on the next txn replication, it expected `replica_stream_` to be already destroyed. There MG_ASSERT failed. Now, the replica stream is created as a local variable for transactions.

Another bug is related to the MVCC of the replica. Because REPLICA’s transaction timestamps are contingent on MAIN’s timestamps, resulting in the transaction timestamp counter (used for obtaining the start transaction ID, and commit timestamp ID) staying the same when the READ transaction started, MVCC guarantees were broken. Any time the WRITE transaction would commit before the READ transaction would finish, GC could incorrectly unlink deltas as the WRITE transaction would mark its start timestamp as finished, not knowing that the READ transaction is still in the system. This would result in the replica’s READ query reading two different data snapshots during a single transaction.

[master < Task] PR
    - **Fix: MVCC design on replica and replica stream isolation**

### CI Testing Labels

### Documentation checklist
    - **[Release note text]**
    - **[Documentation PR link]**

### DIFF
--- a/src/storage/v2/replication/replication_client.cpp
+++ b/src/storage/v2/replication/replication_client.cpp
@@ -329,7 +329,7 @@ void ReplicationStorageClient::RecoverReplica(uint64_t replica_commit, memgraph:
                   spdlog::debug("Sending the latest wal files to {}", client_.name_);
                   auto response = TransferWalFiles(main_uuid, mem_storage->uuid(), rpcClient, wals);
                   replica_commit = response.current_commit_timestamp;
-                  spdlog::debug("Wal files successfully transferred.");
+                  spdlog::debug("Wal files successfully transferred to {}.", client_.name_);
                 },
                 [this, &replica_commit, mem_storage, &rpcClient,
                  main_uuid = main_uuid_](RecoveryCurrentWal const &current_wal) {
@@ -342,7 +342,7 @@ void ReplicationStorageClient::RecoverReplica(uint64_t replica_commit, memgraph:
                     spdlog::debug("Sending current wal file to {}", client_.name_);
                     replica_commit = ReplicateCurrentWal(main_uuid, mem_storage, rpcClient, *mem_storage->wal_file_);
                   } else {
-                    spdlog::debug("Cannot recover using current wal file");
+                    spdlog::debug("Cannot recover using current wal file {}", client_.name_);
                   }
                 },
                 [](auto const &in) {
@@ -357,7 +357,6 @@ void ReplicationStorageClient::RecoverReplica(uint64_t replica_commit, memgraph:
       }
     }
 
-    spdlog::trace("Current timestamp on replica: {}", replica_commit);
     // To avoid the situation where we read a correct commit timestamp in
     // one thread, and after that another thread commits a different a
     // transaction and THEN we set the state to READY in the first thread,
@@ -368,10 +367,12 @@ void ReplicationStorageClient::RecoverReplica(uint64_t replica_commit, memgraph:
     // and we will go to recovery.
     // By adding this lock, we can avoid that, and go to RECOVERY immediately.
     const auto last_commit_timestamp = storage->repl_storage_state_.last_commit_timestamp_.load();
-    SPDLOG_INFO("Replica timestamp: {}", replica_commit);
-    SPDLOG_INFO("Last commit: {}", last_commit_timestamp);
+    SPDLOG_INFO("Replica {} timestamp: {}, Last commit: {}", client_.name_, replica_commit, last_commit_timestamp);
     if (last_commit_timestamp == replica_commit) {
-      replica_state_.WithLock([](auto &val) { val = replication::ReplicaState::READY; });
+      replica_state_.WithLock([name = client_.name_](auto &val) {
+        spdlog::trace("Replica {} set to ready", name);
+        val = replication::ReplicaState::READY;
+      });
       return;
     }
   }
--- a/src/storage/v2/replication/replication_client.cpp
+++ b/src/storage/v2/replication/replication_client.cpp
@@ -189,15 +189,17 @@ void ReplicationStorageClient::TryCheckReplicaStateSync(Storage *storage, Databa
   }
 }
 
-void ReplicationStorageClient::StartTransactionReplication(const uint64_t current_wal_seq_num, Storage *storage,
-                                                           DatabaseAccessProtector db_acc) {
+auto ReplicationStorageClient::StartTransactionReplication(const uint64_t current_wal_seq_num, Storage *storage,
+                                                           DatabaseAccessProtector db_acc)
+    -> std::optional<ReplicaStream> {
   auto locked_state = replica_state_.Lock();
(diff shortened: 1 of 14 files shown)


src/storage/v2/replication/replication_client.cpp
```diff
@@ -233,7 +239,12 @@ bool ReplicationStorageClient::FinalizeTransactionReplication(Storage *storage,
   // valid during a single transaction replication (if the assumption
   // that this and other transaction replication functions can only be
   // called from a one thread stands)
+  spdlog::trace("Finalizing transaction on replica {} in state {}", client_.name_, StateToString());
   if (State() != replication::ReplicaState::REPLICATING) {
+    // TODO(antoniofilipovic) This was missing as we could have been in RECOVERY STATE
+    // because we missed a transaction, and we need to finalize current transaction
+    spdlog::trace("Skipping finalizing transaction on replica {} because it's not replicating", client_.name_);
+    replica_stream_.reset();
```
antoniofilipovic, line 247: Let's say 1 transaction runs and the immediately after the other starts. Due to other starting we switch to the RECOVERY state. We then need to do RECOVERY, we do the recovery we are in a READY state. After starting transaction again, we haven't cleared replica_stream

antoniofilipovic: False assumption, removing this

src/storage/v2/replication/replication_client.cpp
```diff
@@ -205,68 +214,72 @@ void ReplicationStorageClient::StartTransactionReplication(const uint64_t curren
...
     case DIVERGED_FROM_MAIN:
       spdlog::error(utils::MessageWithLink("Couldn't replicate data to {} since replica has diverged from main.",
                                            client_.name_, "https://memgr.ph/replication"));
-      return;
+      return replica_stream;
     case READY:
-      MG_ASSERT(!replica_stream_);
       try {
-        replica_stream_.emplace(storage, client_.rpc_client_, current_wal_seq_num, main_uuid_);
+        replica_stream.emplace(storage, client_.rpc_client_, current_wal_seq_num, main_uuid_);
```
Ignition, line 229: Move `std::optional<ReplicaStream> replica_stream;` closer by adding a scope for this case ie.
```cpp
case READY: {
  auto replica_stream = std::optional<ReplicaStream>{};
  //...
  return replica_stream;
}
```

src/storage/v2/replication/replication_client.hpp
```diff
@@ -121,32 +147,33 @@ class ReplicationStorageClient {
...
+      replica_state_.WithLock([&replica_stream](auto &state) {
+        replica_stream.reset();
         state = replication::ReplicaState::MAYBE_BEHIND;
       });
       LogRpcFailure();
       return;
     }
     try {
-      callback(*replica_stream_);  // failure state what if not streaming (std::nullopt)
+      callback(*replica_stream);  // failure state what if not streaming (std::nullopt)
```
Ignition, line 176: Does the catch need to reset?

src/storage/v2/replication/replication_storage_state.cpp
```diff
@@ -14,72 +14,95 @@
...
                                               const LabelPropertyIndexStats &property_stats,
-                                              uint64_t final_commit_timestamp) {
+                                              uint64_t final_commit_timestamp,
+                                              std::span<std::optional<ReplicaStream>> replica_streams) {
   replication_clients_.WithLock([&](auto &clients) {
-    for (auto &client : clients) {
-      client->IfStreamingTransaction([&](auto &stream) {
-        stream.AppendOperation(operation, label, properties, stats, property_stats, final_commit_timestamp);
-      });
+    for (int i = 0; i < clients.size(); i++) {
```
Ignition, line 60: ranges-v3 enumerate or zip

Ignition: Also with raw loops, be intentional: `for (int i = 0; i != clients.size(); ++i)`

src/storage/v2/replication/replication_storage_state.cpp
```diff
@@ -14,72 +14,95 @@
...
-  });
+bool ReplicationStorageState::FinalizeTransaction(uint64_t timestamp, Storage *storage, DatabaseAccessProtector db_acc,
+                                                  std::vector<std::optional<ReplicaStream>> replica_streams) {
+  return replication_clients_.WithLock(
+      [=, db_acc = std::move(db_acc), replica_streams = std::move(replica_streams)](auto &clients) mutable {
+        bool finalized_on_all_replicas = true;
+        MG_ASSERT(clients.empty() || db_acc.has_value(),
+                  "Any clients assumes we are MAIN, we should have gatekeeper_access_wrapper so we can correctly "
+                  "handle ASYNC tasks");
+        for (int i = 0; i < clients.size(); i++) {
```
Ignition, line 93: again ranges-v3 (unfortunately zip/enumerate are C++23 https://en.cppreference.com/w/cpp/ranges/zip_view https://en.cppreference.com/w/cpp/ranges/enumerate_view)

Ignition (commented):
Few minor comments, otherwise very good

src/storage/v2/replication/replication_client.cpp
```diff
@@ -205,68 +214,72 @@ void ReplicationStorageClient::StartTransactionReplication(const uint64_t curren
...
         LogRpcFailure();
       }
-      return;
+      return replica_stream;
   }
 }
 
-bool ReplicationStorageClient::FinalizeTransactionReplication(Storage *storage, DatabaseAccessProtector db_acc) {
+bool ReplicationStorageClient::FinalizeTransactionReplication(Storage *storage, DatabaseAccessProtector db_acc,
+                                                              std::optional<ReplicaStream> replica_stream) {
```
andrejtonev, line 240: Why are we copying it here?
Make is a && so the transaction must move the stream.
This way we also handle ASYNC replicas; since the lifetime of the stream is not dependent on the transaction, but the background replication process.

andrejtonev (approved):
Makes sense to me.

src/storage/v2/replication/replication_client.cpp
antoniofilipovic: We are moving it here:
```
return repl_storage_state_.FinalizeTransaction(durability_commit_timestamp, this, std::move(db_acc),
                                                 std::move(streams));
```
 but we can be more explicit with rvalue ref, I agree.

src/storage/v2/replication/replication_client.hpp
antoniofilipovic: No need, we don't need to be so explicit with destruction anymore since it is not a private variable, will on RPC fail to change state to MAYBE_BEHIND, and destruction will happen in finalize transaction function, after the object gets out of scope

Ignition (approved):

kgolubic:
@as51340 I will use this as RN:

The replication stream is now handled as a local variable for transactions, preventing assertion failures during recovery.

PR #2055, opened 2024-05-20
Unify and embed multiple filters that operate on the same property

andrejtonev:
Introduced new operator RangeOperator

Rewriting the Cypher AST tree to combine multiple filters into a range filter
Inserting range filter into range scans


src/query/frontend/ast/pretty_print.cpp
```diff
@@ -251,6 +252,13 @@ BINARY_OPERATOR_VISIT(SubscriptOperator, "Subscript");
 
 #undef BINARY_OPERATOR_VISIT
 
+void ExpressionPrettyPrinter::Visit(RangeOperator &op) {
```
Ignition, line 255: How do we reach this code/see its effect?

src/query/plan/preprocess.cpp
```diff
@@ -184,7 +186,8 @@ void AddExpansionsToMatching(std::vector<Expansion> &expansions, Matching &match
   }
 }
 
-auto SplitExpressionOnAnd(Expression *expression) {
+template <typename F1, typename F2>
+void SplitExpressionOnAnd(Expression *expression, F1 &&if_cb, F2 &&else_cb) {
```
Ignition, line 190: Only two usages...inline into both places and refactor them to have simple implementations. The generic callbacks reduce readability.

src/query/plan/rule_based_planner.cpp
```diff
@@ -368,6 +369,12 @@ class ReturnBodyContext : public HierarchicalTreeVisitor {
 
 #undef VISIT_BINARY_OPERATOR
 
+  bool PostVisit(RangeOperator &op) override {
+    bool res = op.expr1_->Accept(*this);
+    res |= op.expr2_->Accept(*this);
```
Ignition, line 374: Does 2nd Accept need to be visited?
Can be do a short-cutting `||`

andrejtonev: We want to check both branches.

src/query/plan/preprocess.cpp
andrejtonev: Sure

src/query/frontend/ast/pretty_print.cpp
andrejtonev: This is used when outputting the plan as a JSON.
The JSON is sent as part of the summary when running an EXPLAIN query.

Query (with index):
```
EXPLAIN MATCH (n:l) WHERE 1 < n.p < 10 RETURN n;
```
Explain query result:
```
+---------------------------------------------+
| QUERY PLAN                                  |
+---------------------------------------------+
| " * Produce {n}"                            |
| " * ScanAllByLabelPropertyRange (n :l {p})" |
| " * Once"                                   |
+---------------------------------------------+
```
JSON:
```
"explain": {
  "input":{
    "input":{
      "name":"Once"
    },
    "label":"l",
    "lower_bound":{
      "type":"exclusive",
      "value":"(ParameterLookup 7)"
    },
    "name":"ScanAllByLabelPropertyRange",
    "output_symbol":"n",
    "property":"p",
    "upper_bound":{
      "type":"exclusive",
      "value":"(ParameterLookup 13)"
    }
  }
}
```

kgolubic:
@andrejtonev I'm missing the Release note entry.

andrejtonev:
### Description

A query like 
`MATCH(n) WHERE n.p > 10 AND n.p < 20 RETURN n;`
will generate a tree with 2 independent filters (`n.p > 10` and `n.p < 20`).
This becomes a problem when we come to our optimizer. 
The optimizer tries to embed filters into range scans. The optimizer will embed one filter, but not the other. It limits the scans to a single filter, so only the upper or lower bound gets embedded.

This PR tries to introduce a new operator RangeOperator. 
It should fix this problem and generate a single filter with a lower and upper bound, which gets correctly inserted into a ranged scan.
The operator is not generated by the `CypherMainVisitor` but later as part of the tree rewrite. This is so other optimizations that move filters around can get picked up and potentially combine filters from different parts of the query into a single range.

[master < Task] PR
    - **[FINAL GIT MESSAGE]**


### Documentation checklist
    - **Better range filter optimizations**
    - **[Documentation PR link]**

Ignition (approved):

kgolubic:
For release note I'll use:

- Improved query optimization by unifying multiple property-based filters into a single filter that applies both upper and lower bounds. 
[#2055](https://github.com/memgraph/memgraph/pull/2055)

PR #2060, opened 2024-05-22
Lazy evaluation of replication DNS

as51340:
### Description

- `ReplicationServerConfig` is now using `Endpoint` struct.
- `ReplicationClientConfig` is now using `Endpoint` struct.
- Replication can now use DNS properly by lazily postponing resolution until connecting sockets.
- Bumped replication durability version to V4, added corresponding migration function.
- Fixed the bug where migration from V2 to V3 wasn't made durable.
- V4 replication durability is saving Endpoint to `replica_server` key, `replica_ip_address` isn't used anymore.
- Removed V1Main and V2Main tests because `to_json` is now serializing only the latest versions.
- `ReplicationReplicaEntry` is also using `DurabilityVersion` but implicitly through `ReplicationRoleEntry`.

[master < Task] PR
    - **Lazy evaluation of replication DNS**

### CI Testing Labels

### Documentation checklist
    - **When using DNS in replication, its resolution is postponed until streaming data using RPC connections**
    - **Replication durability version is bumped to V4.**
    - @andrejtonev anything more for RN, what do you think?
    - **[Documentation PR link]**

### DIFF
--- a/src/replication/include/replication/status.hpp
+++ b/src/replication/include/replication/status.hpp
@@ -61,8 +61,8 @@ struct ReplicationRoleEntry {
 
 // from key: "__replication_replica:"
 struct ReplicationReplicaEntry {
-  DurabilityVersion version =
-      DurabilityVersion::V4;  // if not V4, no version was specified. No changes between [V1, V3]
+  // NOTE: There are multiple versions of ReplicationReplicaEntry, but we distinguish them in terms of
+  // ReplicationRoleEntry.
   ReplicationClientConfig config;
   friend bool operator==(ReplicationReplicaEntry const &, ReplicationReplicaEntry const &) = default;
 };
--- a/src/replication/include/replication/status.hpp
+++ b/src/replication/include/replication/status.hpp
@@ -33,6 +33,7 @@ enum class DurabilityVersion : uint8_t {
   V1,  // no distinct key for replicas
   V2,  // epoch, replica prefix introduced
   V3,  // this version, main uuid introduced
+  V4   // addresses as provided by users are saved to disk instead of eager evaluation
 };
 
 // fragment of key: "__replication_role"
@@ -52,7 +53,7 @@ struct ReplicaRole {
 // from key: "__replication_role"
 struct ReplicationRoleEntry {
   DurabilityVersion version =
-      DurabilityVersion::V3;  // if not latest then migration required for kReplicationReplicaPrefix
+      DurabilityVersion::V4;  // if not latest has been read then migration required to the latest
   std::variant<MainRole, ReplicaRole> role;
 
   friend bool operator==(ReplicationRoleEntry const &, ReplicationRoleEntry const &) = default;
@@ -60,6 +61,8 @@ struct ReplicationRoleEntry {
 
 // from key: "__replication_replica:"
 struct ReplicationReplicaEntry {
+  DurabilityVersion version =
+      DurabilityVersion::V4;  // if not V4, no version was specified. No changes between [V1, V3]
   ReplicationClientConfig config;
   friend bool operator==(ReplicationReplicaEntry const &, ReplicationReplicaEntry const &) = default;
 };
--- a/src/replication/state.cpp
+++ b/src/replication/state.cpp
@@ -158,6 +158,7 @@ auto ReplicationState::FetchReplicationData() -> FetchReplicationResult_t {
                 if (json.is_discarded()) return FetchReplicationError::PARSE_ERROR;
                 try {
                   durability::ReplicationReplicaEntry data = json.get<durability::ReplicationReplicaEntry>();
+                  // TODO: (andi) Here should be HandleVersionMigration code
                   auto key_name = std::string_view{replica_name}.substr(strlen(durability::kReplicationReplicaPrefix));
                   if (key_name != data.config.name) {
                     return FetchReplicationError::PARSE_ERROR;
@@ -222,10 +223,15 @@ bool ReplicationState::HandleVersionMigration(durability::ReplicationRoleEntry &
         main.main_uuid = utils::UUID{};
       }
       data.version = durability::DurabilityVersion::V3;
-      break;
+      [[fallthrough]];
     }
     case durability::DurabilityVersion::V3: {
-      // do nothing - add code if V4 ever happens
+      // Real migration is happening in JSON functions, here just bump version to V4.
+      data.version = durability::DurabilityVersion::V4;
+      [[fallthrough]];
+    }
+    case durability::DurabilityVersion::V4: {
+      // do nothing - add code if V5 ever happens
       break;
     }
   }
@@ -295,7 +301,7 @@ utils::BasicResult<RegisterReplicaError, ReplicationClient *> ReplicationState::
     auto endpoint_check = [&](auto const &replicas) {
       auto endpoint_matches = [&config](auto const &replica) {
         const auto &ep = replica.rpc_client_.Endpoint();
-        return ep.GetResolvedIPAddress() == config.ip_address && ep.GetPort() == config.port;
+        return ep.GetAddress() == config.address && ep.GetPort() == config.port;
       };
       return std::any_of(replicas.begin(), replicas.end(), endpoint_matches);
     };
--- a/src/replication/state.cpp
+++ b/src/replication/state.cpp
@@ -158,9 +158,6 @@ auto ReplicationState::FetchReplicationData() -> FetchReplicationResult_t {
                 if (json.is_discarded()) return FetchReplicationError::PARSE_ERROR;
                 try {
                   durability::ReplicationReplicaEntry data = json.get<durability::ReplicationReplicaEntry>();
-                  if (!HandleVersionMigration(data)) {
-                    return FetchReplicationError::PARSE_ERROR;
-                  }
 
                   auto key_name = std::string_view{replica_name}.substr(strlen(durability::kReplicationReplicaPrefix));
                   if (key_name != data.config.name) {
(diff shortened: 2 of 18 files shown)


src/replication/state.cpp
```diff
@@ -222,10 +223,15 @@ bool ReplicationState::HandleVersionMigration(durability::ReplicationRoleEntry &
         main.main_uuid = utils::UUID{};
       }
       data.version = durability::DurabilityVersion::V3;
-      break;
+      [[fallthrough]];
     }
     case durability::DurabilityVersion::V3: {
-      // do nothing - add code if V4 ever happens
+      // Real migration is happening in JSON functions, here just bump version to V4.
```
andrejtonev, line 229: There is no real data change to be made;  but that is not the only thing we need to do here. We are actually updating the durability. So we need to update the durable version as well.

src/replication/status.cpp
```diff
@@ -72,31 +79,42 @@ void from_json(const nlohmann::json &j, ReplicationRoleEntry &p) {
...
+      std::string address;
       // NOLINTNEXTLINE(cppcoreguidelines-init-variables)
       uint16_t port;
-      j.at(kIpAddress).get_to(ip_address);
+      // In V4 we renamed key `replica_ip_address` to `replica_address`
+      if (version == DurabilityVersion::V4) {
+        j.at(kAddress).get_to(address);
+      } else {
+        j.at(kIpAddress).get_to(address);
+      }
```
andrejtonev, line 90: We shouldn't care about the versions here.
Check what's present in the JSON instead.
```
if (j.contains(kIpAddress)) {
        address = j.at(kIpAddress);
      }
```

src/replication/status.cpp
```diff
@@ -42,8 +43,14 @@ void to_json(nlohmann::json &j, const ReplicationRoleEntry &p) {
...
     auto common = nlohmann::json{{kVersion, p.version},
                                  {kReplicationRole, replication_coordination_glue::ReplicationRole::REPLICA},
-                                 {kIpAddress, replica.config.ip_address},
                                  {kPort, replica.config.port}};
+
+    if (p.version == DurabilityVersion::V4) {
+      common[kAddress] = replica.config.address;
+    } else {
+      common[kIpAddress] = replica.config.address;
+    }
```
andrejtonev, line 52: I know it's much more verbose, but let's be explicit
```
if (p.version == DurabilityVersion::V1 || p.version == DurabilityVersion::V2 || p.version == DurabilityVersion::V3)
```

src/replication/status.cpp
```diff
@@ -72,31 +79,42 @@ void from_json(const nlohmann::json &j, ReplicationRoleEntry &p) {
...
+
       j.at(kPort).get_to(port);
-      auto config = ReplicationServerConfig{.ip_address = std::move(ip_address), .port = port};
+      auto config = ReplicationServerConfig{.address = std::move(address), .port = port};
       auto replica_role = ReplicaRole{.config = std::move(config)};
       if (j.contains(kMainUUID)) {
         replica_role.main_uuid = j.at(kMainUUID);
       }
-
+      // TODO: (andi) Should we bump here to V4? After reading old version, modify and bump to new?
```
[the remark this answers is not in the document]
andrejtonev: That should be done in the migration function. Not here.

src/replication/status.cpp
```diff
@@ -72,31 +79,42 @@ void from_json(const nlohmann::json &j, ReplicationRoleEntry &p) {
...
 void to_json(nlohmann::json &j, const ReplicationReplicaEntry &p) {
-  auto common = nlohmann::json{{kReplicaName, p.config.name},
-                               {kIpAddress, p.config.ip_address},
+  auto common = nlohmann::json{{kVersion, p.version},
+                               {kReplicaName, p.config.name},
                                {kPort, p.config.port},
                                {kSyncMode, p.config.mode},
                                {kCheckFrequency, p.config.replica_check_frequency.count()}};
 
+  if (p.version == DurabilityVersion::V4) {
```
andrejtonev, line 112: Again, lets be explicit about the versions.

src/replication/status.cpp
```diff
@@ -116,15 +137,22 @@ void from_json(const nlohmann::json &j, ReplicationReplicaEntry &p) {
...
-      .ip_address = j.at(kIpAddress).get<std::string>(),
       .port = j.at(kPort).get<uint16_t>(),
       .replica_check_frequency = std::chrono::seconds{seconds},
   };
+
+  if (version == DurabilityVersion::V4) {
+    config.address = j.at(kAddress).get<std::string>();
+  } else {
+    config.address = j.at(kIpAddress).get<std::string>();
+  }
```
andrejtonev, line 148: Just check what's inside the json.

src/replication/status.cpp
```diff
@@ -116,15 +137,22 @@ void from_json(const nlohmann::json &j, ReplicationReplicaEntry &p) {
...
+
   if (!key_file.is_null()) {
     config.ssl = ReplicationClientConfig::SSL{};
     key_file.get_to(config.ssl->key_file);
     cert_file.get_to(config.ssl->cert_file);
   }
-  p = ReplicationReplicaEntry{.config = std::move(config)};
+  // TODO: (andi) Should we bump here to V4? After reading old version, modify and bump to new?
```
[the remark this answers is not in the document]
andrejtonev: We should add a `HandleVersionMigration` for this type as well.
We need to update the durable version as well.

src/replication/status.cpp
as51340: ReplicationReplicaEntry didn't have durability version before. So p.version can never be DurabilityVersion::v1, v2. v3.

That's why I wrote in that way. What do you think?

src/replication/state.cpp
```diff
@@ -226,8 +229,70 @@ bool ReplicationState::HandleVersionMigration(durability::ReplicationRoleEntry &
...
+      break;
+    }
+  }
+  return true;
+}
+
+bool ReplicationState::HandleVersionMigration(durability::ReplicationReplicaEntry &data) const {
+  // Versions V1, V2 and V3 are all actually V1 because no changes were made.
+  // If version wasn't specified when reading from json we set it to V3(= V1 = V2)
+  switch (data.version) {
```
andrejtonev, line 267: Similar comments as the other migration.

src/replication/status.cpp
```diff
@@ -42,13 +45,14 @@ void to_json(nlohmann::json &j, const ReplicationRoleEntry &p) {
...
 
-    if (p.version == DurabilityVersion::V4) {
-      common[kAddress] = replica.config.address;
+    if (p.version == DurabilityVersion::V1 || p.version == DurabilityVersion::V2 ||
+        p.version == DurabilityVersion::V3) {
+      common[kIpAddress] = replica.config.repl_server.GetAddress();  // non-resolved
+      common[kPort] = replica.config.repl_server.GetPort();
     } else {
-      common[kIpAddress] = replica.config.address;
+      common[kReplicaServer] = replica.config.repl_server;  // non-resolved
```
andrejtonev, line 55: Again, I think this type of logic is a mistake. The to_json should only care about the current version.

as51340: I agree about this one

src/replication/state.cpp
```diff
@@ -226,8 +229,70 @@ bool ReplicationState::HandleVersionMigration(durability::ReplicationRoleEntry &
       [[fallthrough]];
     }
     case durability::DurabilityVersion::V3: {
-      // Real migration is happening in JSON functions, here just bump version to V4.
+      std::map<std::string, std::string> to_put;
```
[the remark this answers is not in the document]
as51340: > Seems like the KVStore first does the Puts and then Deletes, so some keys might get deleted by accident.
I think this is mistake, we should first Delete and then Put because RocksDB applies operations in order

src/replication/state.cpp
```diff
@@ -226,8 +229,70 @@ bool ReplicationState::HandleVersionMigration(durability::ReplicationRoleEntry &
...
+        if (old_key == durability::kReplicationRoleName) {
+          continue;
+        }
+        // Turn old data to new data
+        auto old_json = nlohmann::json::parse(old_data, nullptr, false);
+        if (old_json.is_discarded()) return false;
+        try {
+          durability::ReplicationReplicaEntry new_data = old_json.get<durability::ReplicationReplicaEntry>();
+          new_data.version = durability::DurabilityVersion::V4;
+          to_put.emplace(old_key, nlohmann::json(new_data).dump());
```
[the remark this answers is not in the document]
as51340: then I have modification while in the loop, not sure this is best way to interact with RocksDB. I will use `PutMultiple`

src/replication/state.cpp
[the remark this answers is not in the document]
as51340: I will just use PutMultiple

src/replication/state.cpp
as51340: Which comments?

as51340: solved on call

src/replication/include/replication/status.hpp
```diff
@@ -52,14 +53,16 @@ struct ReplicaRole {
...
+      DurabilityVersion::V4;  // if not latest has been read then migration required to the latest
   std::variant<MainRole, ReplicaRole> role;
 
   friend bool operator==(ReplicationRoleEntry const &, ReplicationRoleEntry const &) = default;
 };
 
 // from key: "__replication_replica:"
 struct ReplicationReplicaEntry {
+  DurabilityVersion version =
+      DurabilityVersion::V4;  // if not V4, no version was specified. No changes between [V1, V3]
```
andrejtonev, line 65: Don't need this.

src/replication/status.cpp
```diff
@@ -107,6 +112,10 @@ void to_json(nlohmann::json &j, const ReplicationReplicaEntry &p) {
   j = std::move(common);
 }
 void from_json(const nlohmann::json &j, ReplicationReplicaEntry &p) {
+  using io::network::Endpoint;
+  // This value did not exist until DurabilityVersion::V4, hence default DurabilityVersion::V3
+  DurabilityVersion version = j.value(kVersion, DurabilityVersion::V3);
```
andrejtonev, line 117: Don't need the version.

andrejtonev (approved):
Just remove the version from the ReplicaEntry and that's it.
:muscle:

Issue #1582, opened 2023-12-12
Query behavior changed: `count(*)>0` on an empty set


jchvz:
**Memgraph version**
v2.13.0

**Environment**
Docker: an image with the base image `memgraph/memgraph-mage:1.13-memgraph-2.13` + some query modules

**Describe the bug**
The operation `count(*)>0` on an empty set used to return `false`, but now it returns an empty set. 

Here's a side-by-side of the behavior on `v2.12` vs `v2.13`:

![image](https://github.com/memgraph/memgraph/assets/39979644/8faa13ed-b32c-44d6-b877-b83e4922dab4)

**To Reproduce**
- Ensure the label `DoesNotExist` really does not exist :)
- Run this query on both memgraph `v2.12` and `v2.13` (as shown above):
```cypher
MATCH (n:DoesNotExist) RETURN count(*)>0 AS exists;
```
- Observe the different results

**Expected behavior**
We expect that the example query would return `true` if the nodetype exists and `false` if it does not. Forgive me if this is intended / not a bug, but it was an unexpected change for us. We use this to check if a nodetype exists in the graph.

**Logs**
n/a

**Additional context**
None

**Verification Environment**
Would appreciate a docker image for AMD to verify the fix. Perhaps with MAGE but that is not a requirement.

antepusic:
Hello @jchvz! I’ve opened #2095 in order to resolve the issue you’re having. The PR contains tests which verify that your query should now work as expected.

jchvz:
Thanks @antepusic !

PR #2095, opened 2024-06-03
Exclude constants from grouping keys

antepusic:
Exclude constants from grouping keys


tests/unit/query_plan.cpp
antepusic, line 1: This test suite needed updating now that aggregations are no longer grouped by constant values.

antoniofilipovic:
@antepusic move description to comment section

antepusic:
### Description

#### Context

If the following two conditions are met, the result of an aggregation is an empty set (cf. #1531):
1. there’s no input to aggregate over
2. results are returned grouped

Issue #1582 (the query `MATCH (n:DoesNotExist) RETURN count(*) > 0 AS exists;` returning an empty set) was caused by Memgraph mistakenly identifying the `0` constant as a grouping key. 

#### Changes

This PR fixes the issue by excluding constants from the grouping keys, in line with Cypher:

> For better query readability, Cypher only recognizes a sub-expression in aggregating expressions as a grouping key if the grouping key is either:
> * A variable - e.g. the `p` in `RETURN p, p.age - max(f.age)`.
> * A property access - e.g. the `p.age` in `RETURN p.age, p.age - max(f.age)`.
> * A map access - e.g. the `p.age` in `WITH {name:'Keanu Reeves', age:58} AS p RETURN p.age, p.age - max(p.age)`.
(from [Neo4j docs](https://neo4j.com/docs/cypher-manual/current/functions/aggregating/#grouping-keys))

[master < Task] PR
    > Exclude constants from grouping keys

### Documentation checklist
    > Constant values aren’t used anymore to group the values going into aggregation functions. #2095
- [x] ~Link the documentation PR here~ (N/A  – changelog only)

src/query/plan/rule_based_planner.cpp
```diff
@@ -341,14 +345,17 @@ class ReturnBodyContext : public HierarchicalTreeVisitor {
...
     if (has_aggr && !(aggr1 && aggr2)) {                                                     \
       /* Group by the expression which does not contain aggregation. */                      \
-      /* Possible optimization is to ignore constant value expressions */                    \
-      group_by_.emplace_back(aggr1 ? op.expression2_ : op.expression1_);                     \
+      if (aggr1 && !IsConstantLiteral(op.expression2_)) {                                    \
+        group_by_.emplace_back(op.expression2_);                                             \
+      }                                                                                      \
+      if (aggr2 && !IsConstantLiteral(op.expression1_)) {                                    \
+        group_by_.emplace_back(op.expression1_);                                             \
+      }                                                                                      \
```
antoniofilipovic, line 353: Is it possible that happens that aggr1 is not aggregation, but aggr2 is, but op.expression_1 is IsConstantLiteral and then we don't add group by, but we emplace in has_aggregation_ that there is aggregation?

For example 

```
MATCH (n) 
RETURN 1 + sum(n.prop) AS prop_sum;
```

antoniofilipovic (commented):
Just one comment

src/query/plan/rule_based_planner.cpp
antepusic: Good catch, fixed

antepusic: Yes, it’s possible. In that case we have an aggregation on one side. Normally this means the planner has to group aggregations by the other side’s values, but since the other side is a constant (i.e. just one possibility), there’s no need to group by. This is as intended.

antoniofilipovic (approved):

kgolubic:
For release note I'll use:

- Resolved an issue where constants were incorrectly used as grouping keys in aggregation queries.
[#2095](https://github.com/memgraph/memgraph/pull/2095)

PR #2102, opened 2024-06-09
Add property compression

DavIvek:
This PR handles compression of whole `PropertyStore` object with well-known compression algorithm `zlib`. For that purpose new config flags are introduced:
`--storage-property-store-compression-enabled`
`--storage-property-store-compression-level`


kgolubic:
@DavIvek I'm missing the Release note entry and documentation PR.

src/storage/v2/property_store.cpp
```diff
@@ -27,9 +29,13 @@
 #include "storage/v2/property_value.hpp"
 #include "storage/v2/temporal.hpp"
 #include "utils/cast.hpp"
+#include "utils/compressor.hpp"
 #include "utils/logging.hpp"
 #include "utils/temporal.hpp"
 
+// NOLINTNEXTLINE (cppcoreguidelines-avoid-non-const-global-variables)
+DEFINE_bool(property_store_compression, false, "Enable property store compression.");
```
gitbuda, line 37: @DavIvek to be consistent with the rest of the flags this flags should be prefixed with the `storage` because it's part of the storage + because it's bool the `enabled` has to be there, so `--storage-property-store-compression-enabled` is the right flag name 😄

DavIvek: @gitbuda Thanks for the reminder!

DavIvek:
TODO:
- [ ] add compression rate flag
- [ ] test replication and durability
- [ ] handle wal files writes better (cache uncompressed data before writing to wal file

src/utils/compressor.cpp
```diff
@@ -0,0 +1,125 @@
...
+  const auto compressed_bound_copy = compress_bound;
+
+  const int result = compress2(
+      compressed_data.get(), &compress_bound, input, original_size,
+      memgraph::flags::StoragePropertyStoreCompressionLevelToInt(FLAGS_storage_property_store_compression_level));
+
+  if (result == Z_OK) {
+    std::unique_ptr<uint8_t[]> compressed_buffer_data;
+    const uint32_t compressed_size = compress_bound;
+    if (compress_bound < compressed_bound_copy) {
```
Darych, line 92: Why do you need this check?

Ignition: it was because `compressed_bound_copy` was a guess of the compressed size, the value of `compress_bound` was changed by the call to `compress2`. If they are one and the same then we can use the allocated buffer as is. If not then we need to allocate a smaller buffer to shrink to only what was needed.

Code has been refactored so this is more obvious without need for a comment.

Ignition (approved):
I did some code cleanup from my point of view this is all good. @Darych can you recheck before adding to merge queue?

src/storage/v2/property_store.cpp
```diff
@@ -1398,48 +1408,208 @@ uint32_t ToPowerOf8(uint32_t size) {
...
+  memcpy(compressed_data.get(), &orig_size, sizeof(uint32_t));
+
+  // next byte is the mod before multiple of 8
+  const uint8_t mod = size_needed % 8;
+  compressed_data[sizeof(uint32_t)] = mod;
+
+  // the rest of the buffer is the compressed data
+  memcpy(compressed_data.get() + metadata_size, compressed_view.data(), compressed_view.size_bytes());
+
+  SetSizeData(buffer, compressed_size_to_multiple_of_8 + kUseCompressedBuffer, compressed_data.release());
```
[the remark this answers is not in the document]
DavIvek: I think using `kUseCompressedBuffer` is correct, since we are using `metadata_size` in the calculation of `compressed_size_to_multiple_of_8`. We have to use `kUseCompressedBuffer` when updating the size since we are storing metadata that indicates that data is compressed in the size field with `kUseCompressedBuffer`.
@Ignition can you confirm if this is right?

DavIvek:
Please delete either the [master < EPIC] or [master < Task] part, depending on what are your needs.

[master < Epic] PR
- [ ] Write E2E tests
    - [FINAL GIT MESSAGE]

### CI Testing Labels


### Documentation checklist
    - **[Release note text]**
    - **[Documentation PR link]**

src/storage/v2/property_store.cpp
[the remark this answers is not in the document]
Ignition: Yeah `compressed_size_to_multiple_of_8` is the actual allocated size. The `kUseCompressedBuffer` is just the lower 3 bits being set. Maybe would have been more obvious as `compressed_size_to_multiple_of_8 | kUseCompressedBuffer` to show we are bitwise inserting some information.

PR #2104, opened 2024-06-14
Fix libs/setup.sh destroying uncommitted changes

Ignition:
`libs/setup.sh` had a bug that could destroy uncommitted changes. This was at least 
annoying, at worst remove hours/days of uncommitted work. Also other minor shell 
script issues


Ignition:
Still not sure what root cause was...closing PR as I'm not sure this change actually fixes this problem

PR #2084, opened 2024-05-31
Add NuRaft logs

as51340:
### Description

- Added logger object for NuRaft logs which serves as adapter between spdlog and NuRaft logger :heavy_check_mark: 
- Logger object is shared between state machine, state manager and raft server :heavy_check_mark: 
- Existing logs from state machine and state manager are moved under this new logger :heavy_check_mark: 
- `CoordinatorInstanceInitConfig` extended with `nuraft_log_file` :heavy_check_mark: 
   - if arg not provided, logger will still be created but without any sinks, effectively creating no-op logger :heavy_check_mark: 
- Added `nuraft_log_file` flag :heavy_check_mark: 

### Logs examples
```
[2024-05-31 13:43:06.386] [NuRaft] [info] [asio_service.cxx:771] Raft ASIO listener initiated, UNSECURED
[2024-05-31 13:43:06.386] [NuRaft] [trace] [coordinator_state_machine.cpp:216] Getting last committed index from state machine.
[2024-05-31 13:43:06.386] [NuRaft] [trace] [coordinator_state_machine.cpp:216] Getting last committed index from state machine.
[2024-05-31 13:43:06.386] [NuRaft] [trace] [coordinator_state_machine.cpp:216] Getting last committed index from state machine.
[2024-05-31 13:43:06.386] [NuRaft] [trace] [coordinator_state_manager.cpp:100] Reading server state in coordinator state manager.
[2024-05-31 13:43:06.386] [NuRaft] [trace] [coordinator_state_manager.cpp:49] Loading cluster config from RocksDb
[2024-05-31 13:43:06.386] [NuRaft] [trace] [coordinator_state_manager.cpp:52] Didn't find anything stored on disk for cluster config.
[2024-05-31 13:43:06.386] [NuRaft] [trace] [coordinator_state_machine.cpp:204] Getting last snapshot from state machine.
[2024-05-31 13:43:06.386] [NuRaft] [trace] [coordinator_state_machine.cpp:207] There is no snapshot.
[...]

PR #2113, opened 2024-06-19
Add NuRaft log file env variable

as51340:
### Description

Add NuRaft log file env variables
Refactored CoordinationSetup flag vs. env.
Added test

[master < Task] PR
    - **Add NuRaft log file env variable**

### CI Testing Labels

### Documentation checklist
    - @kgolubic NuRaft Logger is added in 2.18 so this is not yet released. This PR contains some improvements to that so I think no need for new release note, I think we should just extend release note from https://github.com/memgraph/memgraph/pull/2084 by adding: You can also use environment variable `MEMGRAPH_NURAFT_LOG_FILE` for setting path to NuRaft logs.
    - **https://github.com/memgraph/documentation/pull/848**

### DIFF
--- a/src/flags/coord_flag_env_handler.cpp
+++ b/src/flags/coord_flag_env_handler.cpp
@@ -9,6 +9,7 @@
 // by the Apache License, Version 2.0, included in the file
 // licenses/APL.txt.
 
+#include <algorithm>
 #include <cstdint>
 #include <cstdlib>
 #include <string>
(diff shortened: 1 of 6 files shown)


kgolubic:
@as51340 although this is a small change on the docs side, I would still go with a separate Release note entry. I'll use:

In addition to specifying log file paths via the `--nuraft-log-file flag`, users can now set the `MEMGRAPH_NURAFT_LOG_FILE` environment variable to control the location of NuRaft logs. 
#2113

src/flags/coord_flag_env_handler.cpp
```diff
@@ -22,61 +24,91 @@
...
-        "input in environment variables");
-  }
+  bool const any_envs_set = std::ranges::any_of(maybe_coord_envs, [](char const *env) { return env != nullptr; });
+
+  std::array<std::any, 4> maybe_coord_flags{FLAGS_management_port, FLAGS_coordinator_port, FLAGS_coordinator_id,
+                                            FLAGS_nuraft_log_file};
 
-  auto const canonicalize_string = [](auto &&rng) {
-    auto const is_space = [](auto c) { return c == ' '; };
+  auto const is_flag_set = [](auto const &flag) {
```
Ignition, line 62: Don't use std::any, you can just compare with default initialized values.

```cpp
auto const is_flag_set = []<typename T>(T const &flag) { return flag != T{}; };
```

as51340: But I cannot use then std::array? Or which collection can I use then, can you expand?

as51340: solved on the call

Ignition (approved):

PR #2142, opened 2024-06-26
Fix test instance_1 or instance_2 main

antoniofilipovic:
This diff fixes an issue where instance_1 or instance_2 in the test could be main, depending on the election


antoniofilipovic:
@kgolubic no docs needed

as51340 (approved):

PR #2176, opened 2024-07-04
Query module improvements

imilinovic:
Performance enhancements:
- C++ API `MemoryDispatcherGuard` had an expensive registration and fetch mechanism, replaced with thread_local state
- Argument construction made temporary buffers unnecessarily
- Hot-loading required module lookup on every call, we modified implementation to keep module lifetime for the whole function usage


include/mgp.hpp
```diff
@@ -120,33 +120,16 @@ class MemoryDispatcher final {
...
-    const auto this_id = std::this_thread::get_id();
-    std::shared_lock lock(mut_);
-    return map_.contains(this_id);
-  }
+  bool IsThisThreadRegistered() noexcept { return current_memory.has_value(); }
 
  private:
-  std::unordered_map<std::thread::id, mgp_memory *> map_;
-  std::shared_mutex mut_;
+  static thread_local std::optional<mgp_memory *> current_memory;
```
antoniofilipovic, line 132: What if a user creates multiple threads in QM? In that case, user would need to GetMemoryResource from main thread and register MemoryResource on each thread created in QM, right? But at that point, we have I guess different concurrency issue on that one mgp_memory? I think that issue is expected as in memory.cpp we don't have any guarantees for multiple threads accessing the same memory resource. So we should just advise users not to do such things in QMs?

Ignition:
### Tracking


### Standard development
- [ ] Update unit/E2E tests


### CI Testing Labels


### Documentation checklist
    - **[ Release note text ]**

include/mgp.hpp
Ignition: ATM any usage of mgp_memory is not thread safe. We will handle that when a user requires concurrent usage of mgp_memory

antoniofilipovic: @Ignition I would then update the docs to let users know this will not work, maybe here: https://memgraph.com/docs/custom-query-modules/manage-query-modules

src/query/plan/operator.cpp
```diff
@@ -5176,26 +5178,28 @@ class CallProcedureCursor : public Cursor {
...
+        proc = maybe_found->second;
+
+        if (proc->info.is_write != self_->is_write_) {
+          auto get_proc_type_str = [](bool is_write) { return is_write ? "write" : "read"; };
+          throw QueryRuntimeException("The procedure named '{}' was a {} procedure, but changed to be a {} procedure.",
+                                      self_->procedure_name_, get_proc_type_str(self_->is_write_),
+                                      get_proc_type_str(proc->info.is_write));
+        }
       }
+
```
antoniofilipovic, line 5202: I think you don't need this check, since the module will be nullptr always on the next pull
```suggestion
      
      auto maybe_found = procedure::FindProcedure(procedure::gModuleRegistry, self_->procedure_name_,
                                                  context.evaluation_context.memory);
      if (!maybe_found) {
        throw QueryRuntimeException("There is no procedure named '{}'.", self_->procedure_name_);
      }

      module = std::move(maybe_found->first);
      proc = maybe_found->second;

      if (proc->info.is_write != self_->is_write_) {
        auto get_proc_type_str = [](bool is_write) { return is_write ? "write" : "read"; };
        throw QueryRuntimeException("The procedure named '{}' was a {} procedure, but changed to be a {} procedure.",
                                    self_->procedure_name_, get_proc_type_str(self_->is_write_),
                                    get_proc_type_str(proc->info.is_write));
      }
      

```

antoniofilipovic (changes requested):
Few comments, but very nice

src/query/procedure/callable_alias_mapper.hpp
```diff
@@ -32,10 +32,10 @@ class CallableAliasMapper final {
   ~CallableAliasMapper() = default;
 
   void LoadMapping(const std::filesystem::path &);
-  [[nodiscard]] std::optional<std::string_view> FindAlias(const std::string &) const noexcept;
+  [[nodiscard]] std::optional<std::string_view> FindAlias(std::string_view name) const noexcept;
 
  private:
-  std::unordered_map<std::string, std::string> mapping_;
+  std::map<std::string, std::string, std::less<>> mapping_;
```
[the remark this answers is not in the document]
Ignition: To avoid needing to allocate `std::string` for lookups. I prefer to pass around `std::string_view`. We could use a heterogeneous hash function for strings if needed, but for now I am happy to move from unordered_map to map, until a performance related issue highlights that it needs to be a unordered_map.

src/query/plan/operator.cpp
```diff
@@ -5176,26 +5178,28 @@ class CallProcedureCursor : public Cursor {
...
+    auto proc = static_cast<const mgp_proc *>(nullptr);
+
     while (result_row_it_ == result_->rows.end()) {
-      // It might be a good idea to resolve the procedure name once, at the
-      // start. Unfortunately, this could deadlock if we tried to invoke a
-      // procedure from a module (read lock) and reload a module (write lock)
-      // inside the same execution thread. Also, our RWLock is set up so that
-      // it's not possible for a single thread to request multiple read locks.
-      // Builtin module registration in query/procedure/module.cpp depends on
-      // this locking scheme.
```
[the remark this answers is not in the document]
Ignition: This no longer applies. The ModulePtr in the past use to lock the registry (this IMO was too strong), hence this comment to avoid deadlock. 

We now prevent replacing query module while it is in use. The shared_ptr's use_count() != 1 means at least one query is using it.

src/query/plan/operator.cpp
Ignition: It is possible to go around this loop multiple times.
- First call to procedure could give 0 rows.
- hence we go around again, no need to FindProcedure again

src/query/plan/operator.cpp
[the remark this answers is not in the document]
antoniofilipovic: Aha yes, removing is prevented due to check

src/query/plan/operator.cpp
antoniofilipovic: You are right, in that case it will try to pull again and do the processing, makes sense

antoniofilipovic:
@Ignition re-request review when it is ready to be reviewed again, I resolved older comments

src/query/plan/operator.cpp
```diff
@@ -5148,6 +5148,8 @@ class CallProcedureCursor : public Cursor {
   bool stream_exhausted{true};
   bool call_initializer{false};
   std::optional<std::function<void()>> cleanup_{std::nullopt};
+  std::shared_ptr<procedure::Module> module_{};
+  const mgp_proc *proc_{nullptr};
```
antoniofilipovic, line 5152: This would be a breaking change in my opinion. If we cache, this means the user can't reload the module. It also means that such queries that reload a module will block the the memgraph's other queries that want to take a shared lock and execute queries with procedures. I am against this change honestly

antoniofilipovic (approved):
Good work!

MarkoBarisic:
Don't merge before I approve, there have been changes in `mgbuild.sh`, `jepsen/run.sh` and `diff_jepsen.yaml`

MarkoBarisic (approved):
LGTM

PR #2138, opened 2024-06-25
Introduce management servers on coordinators to get health state

antoniofilipovic:
This diff adds health output to follower coordinators by contacting the leader coordinator.
The following things are changed/introduced
- each coordinator now has a coordinator management server and coordinator client connector
- coordinator client connection exists from each coordinator to all others 
- management port is serialized on disk in the state machine durability folder under `/high_availability/nuraft/network`
- since migration from v1 to v2 for `/high_availability/nuraft/network` RocksDB instance is not possible this introduces breaking change

This allows followers to correctly answer the SHOW INSTANCES command in Cypher

PR #2186, opened 2024-07-08
Separate coordinator client creation from ShowInstances

antoniofilipovic:
This diff moves client creation on config change

### DIFF
--- a/src/coordination/coordinator_instance.cpp
+++ b/src/coordination/coordinator_instance.cpp
@@ -89,6 +89,7 @@ CoordinatorInstance::CoordinatorInstance(CoordinatorInstanceInitConfig const &co
   // If something is not yet constructed in coordinator instance, we get UB
   raft_state_ = std::make_unique<RaftState>(config, GetBecomeLeaderCallback(), GetBecomeFollowerCallback(),
                                             CoordinationClusterChangeObserver{this});
+  AddOrUpdateClientConnectors(raft_state_->GetCoordinatorToCoordinatorConfigs());
   raft_state_->InitRaftServer();
 }
 
@@ -281,7 +282,7 @@ auto CoordinatorInstance::GetCoordinatorsInstanceStatus() const -> std::vector<I
   return utils::fmap(raft_state_->GetCoordinatorInstances(), coord_instance_to_status);
 }
 
-auto CoordinatorInstance::ShowInstancesStatusAsFollower() -> std::vector<InstanceStatus> {
+auto CoordinatorInstance::ShowInstancesStatusAsFollower() const -> std::vector<InstanceStatus> {
   spdlog::trace("Processing show instances as follower");
   auto instances_status = GetCoordinatorsInstanceStatus();
   auto const stringify_inst_status = [raft_state_ptr =
@@ -310,7 +311,7 @@ auto CoordinatorInstance::ShowInstancesStatusAsFollower() -> std::vector<Instanc
   return instances_status;
 }
 
-auto CoordinatorInstance::ShowInstancesAsLeader() -> std::vector<InstanceStatus> {
+auto CoordinatorInstance::ShowInstancesAsLeader() const -> std::vector<InstanceStatus> {
   spdlog::trace("Processing show instances as leader");
   auto instances_status = GetCoordinatorsInstanceStatus();
 
@@ -342,7 +343,7 @@ auto CoordinatorInstance::ShowInstancesAsLeader() -> std::vector<InstanceStatus>
   return instances_status;
 }
 
-auto CoordinatorInstance::ShowInstances() -> std::vector<InstanceStatus> {
+auto CoordinatorInstance::ShowInstances() const -> std::vector<InstanceStatus> {
   if (is_leader_ready_) {
     return ShowInstancesAsLeader();
   }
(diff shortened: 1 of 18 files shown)


antoniofilipovic:
**Important**: This PR is part of https://github.com/memgraph/memgraph/pull/2138 and docs will be added for that PR

### Tracking




### Standard development
- [x] Update unit/E2E tests


### CI Testing Labels


### Documentation checklist
    - Part of same PR which is https://github.com/memgraph/memgraph/pull/2138

src/coordination/coordinator_instance.cpp
```diff
@@ -339,24 +360,19 @@ auto CoordinatorInstance::ShowInstances() -> std::vector<InstanceStatus> {
     return ShowInstancesStatusAsFollower();
   }
   CoordinatorInstanceConnector *follower{nullptr};
```
as51340, line 362:
```suggestion
  CoordinatorInstanceConnector *leader{nullptr};
```

src/coordination/coordinator_instance.cpp
```diff
@@ -198,6 +201,24 @@ auto CoordinatorInstance::GetLeaderCoordinatorData() const -> std::optional<Coor
...
 
+void CoordinatorInstance::AddOrUpdateClientConnectors(std::vector<CoordinatorToCoordinatorConfig> const &configs) {
+  auto connectors = coordinator_connectors_.Lock();
+
+  for (auto const &config : configs) {
+    if (config.coordinator_id == raft_state_->GetCoordinatorId()) {
+      continue;
+    }
+    auto const connector = std::ranges::find_if(
+        *connectors, [&config](auto &&connector) { return connector.first == config.coordinator_id; });
```
as51340, line 212: you already have coordinator_id in CoordinatorToCoordinatorConfig saved, no need to have std::list of pairs

src/coordination/coordinator_state_manager.cpp
```diff
@@ -93,35 +94,57 @@ auto CoordinatorStateManager::HandleVersionMigration() -> void {
...
       logger_(logger),
-      durability_(config.state_manager_durability_dir_) {
+      durability_(config.state_manager_durability_dir_),
+      observer_(observer) {
   auto const c2c = CoordinatorToCoordinatorConfig{
+      // TODO(antonio) sync with Andi on this one
       config.coordinator_id_, io::network::Endpoint("0.0.0.0", config.bolt_port_),
       io::network::Endpoint{"0.0.0.0", static_cast<uint16_t>(config.coordinator_port_)},
-      io::network::Endpoint{"0.0.0.0", static_cast<uint16_t>(config.management_port_)}, config.coordinator_hostname};
+      io::network::Endpoint{config.coordinator_hostname, static_cast<uint16_t>(config.management_port_)},
```
as51340, line 108: not 100% sure, check how coordinators are returned in show instances and then see if this change makes sense. With this change yo for sure need to change that part of the code so just like this isn't good

as51340: GetCoordinatorsInstanceStatus this method

src/coordination/include/coordination/coordination_observer.hpp
```diff
@@ -0,0 +1,33 @@
...
+#include "coordination/coordinator_communication_config.hpp"
+
+namespace memgraph::coordination {
+
+class CoordinatorInstance;
+
+class CoordinationClusterChangeObserver {
+ public:
+  explicit CoordinationClusterChangeObserver(CoordinatorInstance *instance);
+  void Update(std::vector<CoordinatorToCoordinatorConfig> const &configs);
```
as51340, line 25: this could be just lambda function for less-instrusive Observer pattern. No need for whole class, just lambda function which captures pointer to CoordinatorInstance, then you don't need new files etc. you just pass functions

as51340 (changes requested):

src/coordination/include/coordination/coordination_observer.hpp
antoniofilipovic: I understand. Having class means better readability and separation. Usually lambas are used for strategy pattern. I would leave it like this rather if you don't mind

src/coordination/coordinator_state_manager.cpp
antoniofilipovic: resolved on call

src/coordination/include/coordination/coordination_observer.hpp
as51340: Well the question is why is this more readable and why separation? There is some new object (Observer) but what do we get by having it as a class, not much IMO just few more files. Lambdas are often used in such situations also in Observer not just in Strategy.

I will leave this one to you to decide then

src/coordination/coordinator_state_manager.cpp
```diff
@@ -93,35 +94,57 @@ auto CoordinatorStateManager::HandleVersionMigration() -> void {
...
-auto CoordinatorStateManager::load_config() -> ptr<cluster_config> {
+auto CoordinatorStateManager::GetCoordinatorToCoordinatorConfigs() const
+    -> std::vector<CoordinatorToCoordinatorConfig> {
+  std::vector<CoordinatorToCoordinatorConfig> coordinator_to_coordinator_mappings;
+  auto const &cluster_config_servers = cluster_config_->get_servers();
+  coordinator_to_coordinator_mappings.reserve(cluster_config_servers.size());
+
+  std::ranges::transform(
+      cluster_config_servers, std::back_inserter(coordinator_to_coordinator_mappings),
+      [](auto const &server) -> CoordinatorToCoordinatorConfig {
```
as51340, line 128:
```suggestion
      [](auto  &&server) -> CoordinatorToCoordinatorConfig {
```

as51340 (approved):
Small comment, but looks good :muscle:

src/coordination/include/coordination/coordination_observer.hpp
antoniofilipovic: Readable as we don't create a lambda function in the constructor of the coordinator instance, so we don't pollute the constructor and the coordinator instance is in that way the owner of the function with lambda. On the other hand this way, the coordinator instance via some other class is connected to the state manager, which sounds a lot like separation of concerns.
 So I would leave it like this.

src/coordination/coordinator_state_manager.cpp
antoniofilipovic: I will apply it in merged PR :muscle:

PR #2190, opened 2024-07-09
Fixed edge type property index lookup scan.

Darych:
Fixed edge type property index lookup scan.
Fixed disabling of edge type property index scan in case of more complex queries.
Why: queries which tries to construct a path based on edge type property index are returned as nulls.
Root cause: Edge type property index scan doesn't fill vertices in frame.


Darych:
---
___


### Tracking


### Standard development
- [ ] Update unit/E2E tests


### CI Testing Labels


### Documentation checklist
    - **[ Release note text ]**

gitbuda:
@Darych thinking out loud here, is it true that the user will always get a plan that can be executed? In other words, it shouldn't be the case that if there is an index, we generate a plan that will throw an error preventing the user from executing anything. In other words, the suboptimal and correct plan is much better than the optimal plan throwing an exception.

src/query/plan/rewrite/edge_type_index_lookup.hpp
```diff
@@ -599,19 +633,18 @@ class EdgeTypeIndexRewriter final : public HierarchicalLogicalOperatorVisitor {
...
+      switch (prop_filter_type) {
+        case PropertyFilter::Type::IS_NOT_NULL:
+        case PropertyFilter::Type::EQUAL: {
+          filter_exprs_for_removal_.insert(property_filter_->expression);
+          filters_.EraseFilter(*property_filter_);
+          return std::make_unique<ScanAllByEdgeTypeProperty>(input, output_symbol, edge_type_, *property_, view);
+        }
+        default: {
+          spdlog::warn("Filter is not supported by edge type property index.");
+        }
```
antoniofilipovic, line 646: I don't think `spdlog::warn` is necessary here. The user can't do anything about it, so I would rather remove it.

antoniofilipovic (changes requested):
Small comment

Darych:
> @Darych thinking out loud here, is it true that the user will always get a plan that can be executed? In other words, it shouldn't be the case that if there is an index, we generate a plan that will throw an error preventing the user from executing anything. In other words, the suboptimal and correct plan is much better than the optimal plan throwing an exception.

Yes, in this implementation we should fallback to not optimal ScanAll and return the result

src/query/plan/rewrite/edge_type_index_lookup.hpp
Darych: Thank you!

kgolubic:
For release note I'll use:

- Fixed edge type property index lookup scan. The edge type property index scan
now correctly fills vertices in the frame, preventing queries that construct a
path based on edge type property index from returning nulls. The edge type
property index scan is disabled for more complex queries, defaulting to
`ScanAll` instead. [#2190](https://github.com/memgraph/memgraph/pull/2190)

PR #2191, opened 2024-07-10
Handle bad_alloc being thrown by 3rd party code

Ignition:
Also covers the case of if we accidentally use it. We should be checking MemoryErrorStatus upon nullptr, and then throwing typically OutOfMemoryException not std::bad_alloc.


antoniofilipovic (approved):

kgolubic:
For release note I'll use:

- Correctly handle memory limits when triggered by allocations in third-party code.
[#2191](https://github.com/memgraph/memgraph/pull/2191)

PR #2195, opened 2024-07-11
Fix issue in flaky HA test for SHOW REPLICAS output

antoniofilipovic:
This diff fixes the issue that the `SHOW REPLICAS` query didn't cover, where `instance_2` becomes main instead of `instance_1`, as the leader coordinators starts pinging  instance_2 earlier than instance_1. Show instances query did have such case covered, but show replicas didn't.


antoniofilipovic:
@kgolubic no need for docs or RN

PR #2197, opened 2024-07-11
Periodic commit delete planning

Josipmrden:
Query planning for periodic delete. Delete and periodic commit can now be combined in this query. Delete now has a buffer after which it knows when to flush and detach delete. Queries are still being blocked on the storage level.


Josipmrden:
### Tracking


### Standard development
- [ ] Update unit/E2E tests


### CI Testing Labels


### Documentation checklist
    - **[ Release note text ]**

src/query/plan/operator.cpp
```diff
@@ -3007,58 +3007,71 @@ bool Delete::DeleteCursor::Pull(Frame &frame, ExecutionContext &context) {
   OOMExceptionEnabler oom_exception;
   SCOPED_PROFILE_OP("Delete");
 
-  if (delete_executed_) {
-    return false;
+  if (self_.buffer_size_ && buffer_size_ == -1) {
```
antoniofilipovic, line 3010:
```suggestion
  if (self_.buffer_size_ && buffer_size_ == -1) [[unlikely]] {
```

src/query/plan/operator.cpp
```diff
@@ -3007,58 +3007,71 @@ bool Delete::DeleteCursor::Pull(Frame &frame, ExecutionContext &context) {
...
+    // Limit expression doesn't contain identifiers so graph view is not
+    // important.
+    ExpressionEvaluator evaluator(&frame, context.symbol_table, context.evaluation_context, context.db_accessor,
+                                  storage::View::OLD);
+    TypedValue buffer_size = self_.buffer_size_->Accept(evaluator);
+    if (buffer_size.type() != TypedValue::Type::Int)
+      throw QueryRuntimeException("Number of periodically deleted elements must be an integer.");
+
```
antoniofilipovic, line 3018: Can we cover this part in planning so we don't have to worry about that here? I would put here DMG_ASSERT that this is int type

src/query/plan/operator.cpp
```diff
@@ -3007,58 +3007,71 @@ bool Delete::DeleteCursor::Pull(Frame &frame, ExecutionContext &context) {
...
-    switch (res.GetError()) {
-      case storage::Error::SERIALIZATION_ERROR:
-        throw TransactionSerializationException();
-      case storage::Error::VERTEX_HAS_EDGES:
-        throw RemoveAttachedVertexException();
-      case storage::Error::DELETED_OBJECT:
-      case storage::Error::PROPERTIES_DISABLED:
-      case storage::Error::NONEXISTENT_OBJECT:
-        throw QueryRuntimeException("Unexpected error when deleting a node.");
+  if (!has_more || (buffer_size_ > -1 && pulled_ >= buffer_size_)) {
```
antoniofilipovic, line 3030: It would be nicer if buffer_size is optional here, so you can have something like
```
if (!has_more || (buffer_size.has_value() && ...)
...
```

src/query/plan/operator.hpp
```diff
@@ -1392,6 +1392,8 @@ class Delete : public memgraph::query::plan::LogicalOperator {
   /// Whether the vertex should be detached before deletion. If not detached,
   ///            and has connections, an error is raised when deleting edges.
   bool detach_;
+  // when buffer size is reached, delete will be triggered
+  Expression *buffer_size_;
```
antoniofilipovic, line 1396: I would call it  periodic_commit_limit, as it is from periodic commit, so it makes sense to be tied together here

src/query/plan/rewrite/periodic_delete.hpp
```diff
@@ -0,0 +1,79 @@
...
+    periodic_subquery_seen_ = true;
+    return true;
+  }
+
+  bool PreVisit(Delete &op) override {
+    // since periodic commit needs to flush deltas over time, buffer size needs to be put in delete
+    // so it can know when to trigger delete of nodes and/or relationships
+    if (commit_frequency_) {
+      op.buffer_size_ = commit_frequency_;
+    }
```
antoniofilipovic, line 50: I had to check query planning for this part but let me know if I got something wrong. I am not sure why will `prerevisit` Delete after `previsiting` PeriodicCommit but since we are having periodic commit always at the end of our PullPlan (because we generate it at GenReturnBody function, which is always after HandleWriteClause in rule_based_planner.hpp), this means we will previsit in our visitor first periodic commit and then Delete, which means this should always work as it should (meaning if we actually have periodic commit + delete, we will notice it)

antoniofilipovic (changes requested):
Good work, couple of comments

src/query/plan/rewrite/periodic_delete.hpp
```diff
@@ -0,0 +1,79 @@
...
+    if (periodic_subquery_seen_) {
+      throw utils::NotYetImplemented("Delete can not be used with periodic subquery, please use USING PERIODIC COMMIT");
+    }
+    return true;
+  }
+
+  // TODO: there are more places where an expression could be holding onto an EnumValueAccess
```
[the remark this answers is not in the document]
Josipmrden: sorry, I was copying this file, will remove the comment.

src/query/plan/operator.cpp
```diff
@@ -3007,58 +3007,71 @@ bool Delete::DeleteCursor::Pull(Frame &frame, ExecutionContext &context) {
   OOMExceptionEnabler oom_exception;
   SCOPED_PROFILE_OP("Delete");
 
-  if (delete_executed_) {
-    return false;
+  if (self_.buffer_size_ && buffer_size_ == -1) {
+    // Limit expression doesn't contain identifiers so graph view is not
```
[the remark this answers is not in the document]
Josipmrden: again yeah, due to copying, sorry again

src/query/plan/operator.cpp
Josipmrden: covered in separate function. Unfortunately I can't use the expression evaluator here, as that is some other scope (making sure expression eval works with planning and not only execution)

src/query/plan/operator.hpp
Josipmrden: I was more thinking like, maybe something else in the future can trigger the buffer size to be different?

src/query/plan/operator.cpp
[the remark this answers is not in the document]
Josipmrden: also abstracted this into its own function

src/query/plan/rewrite/periodic_delete.hpp
Josipmrden: Plan's first operator is always Produce or Empty result (the last part of the query). Meaning, PeriodicCommit's previsit needs to be before the Delete, in order for periodic commit to influence delete and transfer the commit frequency, i.e. buffer size to delete.

src/query/plan/rewrite/periodic_delete.hpp
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  bool PreVisit(Delete &op) override {
+    // since periodic commit needs to flush deltas over time, buffer size needs to be put in delete
+    // so it can know when to trigger delete of nodes and/or relationships
+    if (commit_frequency_) {
+      op.buffer_size_ = commit_frequency_;
+    }
+    if (periodic_subquery_seen_) {
+      throw utils::NotYetImplemented("Delete can not be used with periodic subquery, please use USING PERIODIC COMMIT");
```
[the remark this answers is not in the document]
Josipmrden: Consider these 2 plans
This one we can do now
```
memgraph> explain using periodic commit 2 match (n) detach delete n;
+---------------------+
| QUERY PLAN          |
+---------------------+
| " * EmptyResult"    |
| " * PeriodicCommit" |
| " * Delete"         |
| " * ScanAll (n)"    |
| " * Once"           |
+---------------------+
```
We can do this plan because periodic commit is in the same branch as delete, and we can transfer easily the information about what is going to be the commit frequency, so periodic commit and delete can be in sync by how often we flush and periodically commit (after 2 rows, both delete will perform the exact delete and the periodic commit will commit).

We need this because Delete is currently a pipeline breaker, it will not delete until it has aggregated everything it can delete, so delete operations can be efficient. If it does not delete in that manner, deletes will take n^2 time if nodes are detach deleted one by one.

The second plan is:
```
memgraph> explain match (n) CALL { WITH n detach delete n } in transactions of 2 rows;
+-----------------------+
| QUERY PLAN            |
+-----------------------+
| " * EmptyResult"      |
| " * Accumulate"       |
| " * PeriodicSubquery" |
| " |\\ "               |
| " | * EmptyResult"    |
| " | * Delete"         |
| " | * Produce {n}"    |
| " | * Once"           |
| " * ScanAll (n)"      |
| " * Once"             |
+-----------------------+
```
We can't execute this because currently, how we execute subqueries, is that we exhaust the subquery branch when we do an input pull. Since the first operator in the subquery branch (right one) is `Once`, we will pull `Once` and the second time it will signal the `Delete` to execute, since once returned false. Therefore, deletes will take `O(n^2)` complexity.
Idea therefore is to introduce some other kind of operator because `Delete` behaves different from other operators, but since we have a way to periodically execute deletes, I would postpone that work.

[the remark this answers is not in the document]
Darych: Thank you for explanation! I would suggest to add more obvious error message what user should do in case of this error message. Because, if me as user got this message, I still won't know what to do and will need to go to the documentation and try to understand what developers wanted from me :D

Josipmrden:
@antoniofilipovic  @Darych  I need to add `query_plan.cpp` delete queries additionally.

src/query/plan/operator.cpp
```diff
@@ -3007,17 +3006,10 @@ bool Delete::DeleteCursor::Pull(Frame &frame, ExecutionContext &context) {
   OOMExceptionEnabler oom_exception;
   SCOPED_PROFILE_OP("Delete");
 
-  if (self_.buffer_size_ && buffer_size_ == -1) {
-    // Limit expression doesn't contain identifiers so graph view is not
-    // important.
+  if (self_.buffer_size_ && buffer_size_ == -1) [[unlikely]] {
```
Darych, line 3009: Are we comparing here optional and -1?

Darych (commented):
All good just one comment :)

src/query/interpret/eval.cpp
```diff
@@ -13,28 +13,42 @@
 
 namespace memgraph::query {
 
-int64_t EvaluateInt(ExpressionEvaluator *evaluator, Expression *expr, std::string_view what) {
-  TypedValue value = expr->Accept(*evaluator);
+int64_t EvaluateInt(ExpressionVisitor<TypedValue> &eval, Expression *expr, std::string_view what) {
```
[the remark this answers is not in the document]
Josipmrden: So I was a bit confused by the input types of both of these functions, so I changed them to be consistent. However, changing the return type is a bit tricky because I will be editing the code that's not in this PR. I can make a brief change in a different PR if that's okay so we separate concerns? EvaluateInt is currently not used in this PR's logic, but was used some time ago when evaluating bounds during path traversing (not part of this query)

src/query/plan/operator.cpp
Josipmrden: reworked this to work with optional only

src/query/interpret/eval.cpp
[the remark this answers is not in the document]
Darych: I agree with you that changes to `EvaluateInt` don't make sense in this PR, so let's do them in another one.

Darych (approved):

src/query/plan/operator.hpp
antoniofilipovic: It sounds to me like premature optimization, but I will leave it up to you

src/query/plan/rewrite/periodic_delete.hpp
antoniofilipovic: Yes yes, that is what I wanted to confirm, exactly. So `PeriodicCommit` will be visited in `previsit` function before `Delete` every time, so this should work

antoniofilipovic (approved):

kgolubic:
For release notes I'll use:

- Added query planning for periodic commit with delete
[#2197](https://github.com/memgraph/memgraph/pull/2197)

PR #2208, opened 2024-07-17
Add spatial datatype support

imilinovic:
Add spatial datatype support
- datatype (property value, typed value, and bolt value)
- functions (construction, distance, withinbbox)


imilinovic:
---
___


### Tracking


### Standard development
- [ ] Update unit/E2E tests


### CI Testing Labels


### Documentation checklist
    - **[ Release note text ]**

src/query/interpret/awesome_memgraph_functions.cpp
```diff
@@ -638,6 +649,9 @@ TypedValue ValueType(const TypedValue *args, int64_t nargs, const FunctionContex
       return TypedValue("DURATION", ctx.memory);
     case TypedValue::Type::Enum:
       return TypedValue("ENUM", ctx.memory);
+    case TypedValue::Type::Point2d:
+    case TypedValue::Type::Point3d:
+      return TypedValue("POINT", ctx.memory);
```
antoniofilipovic, line 654: Since opencypher9 doesn't define point, did you maybe check what neo4j does here?

src/query/interpret/awesome_memgraph_functions.cpp
```diff
@@ -1383,6 +1397,281 @@ TypedValue DateTime(const TypedValue *args, int64_t nargs, const FunctionContext
...
+
+  using CRS = storage::CoordinateReferenceSystem;
+  if (!mg_crs.has_value()) {
+    auto is_wgs = (from_longitude || from_latitude);
+    if (!z_opt.has_value()) {
+      mg_crs = is_wgs ? CRS::WGS84_2d : CRS::Cartesian_2d;
+    } else {
+      mg_crs = is_wgs ? CRS::WGS84_3d : CRS::Cartesian_3d;
+    }
+  }
```
antoniofilipovic, line 1518: Should `RETURN point({longitude:1, latitude:2, srid:7205}).srid;` be a valid query? In this case we combine longitude, latitude and `srid` value of Cartesian, and from lines 1512 to 1518 it seems that if you define longitude/latitude you should use crs which is `wgs84_2d` or  `wgs84_3d`?

src/query/interpret/awesome_memgraph_functions.cpp
```diff
@@ -1383,6 +1397,281 @@ TypedValue DateTime(const TypedValue *args, int64_t nargs, const FunctionContext
...
+    if (enum_val.HasError()) throw QueryRuntimeException("Invalid enum '{}'", s1);
+    return TypedValue(*enum_val, ctx.memory);
+  }
+  auto const &s2 = args[1].ValueString();
+  auto enum_val = ctx.db_accessor->GetEnumValue(s1, s2);
+  if (enum_val.HasError()) throw QueryRuntimeException("Invalid enum '{}::{}'", s1, s2);
+  return TypedValue(*enum_val, ctx.memory);
+}
+
+TypedValue Point(const TypedValue *args, int64_t nargs, const FunctionContext &ctx) {
```
antoniofilipovic, line 1415: There are multiple combinations in this code and seems like easy to introduce bug. I am not sure for one question below, but what I would rather do is rewrite function that if you start with defining x: you need to also find y: and throw error otherwise. Instead of finding every combination, you are narrowing it down. Seems like it would be less bug prone

antoniofilipovic: It is really nice code btw, but seems hard to catch every combination

antoniofilipovic: From reading this code I was not able to figure out if we should allow following combination or not:
`RETURN point({longitude:1, latitude:2, srid:7205}).srid;`
This is just my objection

src/storage/v2/durability/version.hpp
```diff
@@ -20,14 +20,15 @@ namespace memgraph::storage::durability {
...
-const uint64_t kVersion{18};
+const uint64_t kVersion{19};
 
 const uint64_t kOldestSupportedVersion{14};
 const uint64_t kUniqueConstraintVersion{13};
 // Edge-type index version is 17. Edge-type property index version is 18.
 // But they are written in the same section.
 const uint64_t kEdgeIndicesVersion{17};
 const uint64_t kEnumsVersion{18};
+const uint64_t kPointsVersion{19};
```
antoniofilipovic, line 31: Seems like we don't need this constant?

src/storage/v2/replication/slk.cpp
```diff
@@ -39,6 +39,48 @@ void Load(storage::Enum *enum_val, slk::Reader *reader) {
   *enum_val = storage::Enum{storage::EnumTypeId{etype}, storage::EnumValueId{evalue}};
 }
 
+void Save(const storage::Point2d &point2d_val, slk::Builder *builder) {
+  slk::Save(point2d_val.crs(), builder);
+  slk::Save(point2d_val.x(), builder);
+  slk::Save(point2d_val.y(), builder);
+}
+
+void Load(storage::Point2d *point2d_val, slk::Reader *reader) {
```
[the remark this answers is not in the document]
Ignition: My preference is to change slk approach to return values and not to do modification of an existing object. Not doing any slk pattern changes for this PR.

src/query/interpret/awesome_memgraph_functions.cpp
Ignition: It says `POINT NOT NULL`

src/query/interpret/awesome_memgraph_functions.cpp
Ignition: 7205 we do `Invalid SRID`
7203 we accept, which is a 🐛 should be rejected.

src/storage/v2/replication/slk.cpp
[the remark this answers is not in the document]
Ignition: Also cost of copy assignment here is nothing as these are simple trivial values.

src/storage/v2/point_functions.cpp
```diff
@@ -0,0 +1,120 @@
...
+  auto ub_y = upper_bound.y();
+
+  auto p_x = point.x();
+  auto p_y = point.y();
+  auto crs = point.crs();
+
+  if (lb_x <= ub_x || crs == Cartesian_2d) [[likely]] {
+    return lb_x <= p_x && p_x <= ub_x && lb_y <= p_y && p_y <= ub_y;
+  } else {
+    return lb_x <= p_x && p_x <= ub_x + 360.0 && lb_y <= p_y && p_y <= ub_y;
```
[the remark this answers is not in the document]
Ignition: Good catch. Correct logic is `(lb_x <= p_x || p_x <= ub_x)` no need for any modification of the bounds.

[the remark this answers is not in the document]
Ignition: `WithinBBox` -> "Within bounding box"
Just with WGS-84 that you can wrap around the sphere, hence this wrap around case.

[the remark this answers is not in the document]
Ignition: @imilinovic

tests/unit/property_value_v2.cpp
```diff
@@ -18,6 +18,21 @@
...
 
 using namespace memgraph::storage;
+using enum CoordinateReferenceSystem;
+
+///!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+///!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+///!!!!!!!!!!!THIS FILE IS TOO LONG!!!!!!!!!!!!
+///!!!!!!!!!!!!! TODO: REFACTOR !!!!!!!!!!!!!!!
+///!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+///!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```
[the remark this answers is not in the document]
Ignition: No, ideally we will eventually have alternative generic variant type which is high performance. Then most of what this test does is redundant. We only have the test ATM because we have to hand code a variant, which is error prone.

antoniofilipovic (approved):
Looks good to me

Josipmrden:
```
RETURN
  point({x: 3, y: 0}) AS cartesian_2d,
  point({x: 0, y: 4, z: 1}) AS cartesian_3d,
  point({latitude: 12, longitude: 56}) AS geo_2d,
  point({latitude: 12, longitude: 56, height: 1000}) AS geo_3d
```

in mgconsole it is printed as `unknown_value`, all 4 of them. What's the best way to inspect the created values?

Josipmrden:
```
RETURN
  point.withinBBox(
    null,
    point({longitude: 56.7, latitude: 12.78}),
    point({longitude: 57.0, latitude: 13.0})
  ) AS in
```

`If null is provided as any of the arguments, null is returned.
`

Up to you for the decision, for now we have an exception which is also okay.

Josipmrden:
Tested import, setting, copying of values, returning, durability, functions, property size, and invalid values. Seems legit, with some comments that I left.

Josipmrden (commented):
Commented in the comment above

Ignition:
> I believe they are the same CRS, right?
2d cartesian crs is different from 3d cartesian crs

> in mgconsole it is printed as unknown_value
mgconsole is outside scope for 2.19

> If null is provided as any of the arguments, null is returned.
Good catch will add null to the valid types of withinBBox (+ double check distance)

PR #2217, opened 2024-07-19
Periodic commit transaction logic

Josipmrden:
Periodic commit transaction logic.


src/query/cypher_query_interpreter.cpp
```diff
@@ -125,19 +125,20 @@ ParsedQuery ParseQuery(const std::string &query_string, UserParameters const &us
 }
 
 std::unique_ptr<LogicalPlan> MakeLogicalPlan(AstStorage ast_storage, CypherQuery *query, const Parameters &parameters,
-                                             DbAccessor *db_accessor,
+                                             DbAccessor *db_accessor, DatabaseAccessProtector db_acc,
```
andrejtonev, line 128: Can we pass in the protector at run-time instead of at plan creation time?
If we cache the plan with a protector, we need to make sure the cache is destroyed before attempting to delete the storage. Otherwise we will be blocked.

src/storage/v2/inmemory/edge_type_index.cpp
```diff
@@ -208,6 +208,12 @@ void InMemoryEdgeTypeIndex::Iterable::Iterator::AdvanceUntilValid() {
     auto *from_vertex = index_iterator_->from_vertex;
     auto *to_vertex = index_iterator_->to_vertex;
 
+    if (self_->transaction_->original_start_timestamp.has_value()) {
+      if (index_iterator_->timestamp >= self_->transaction_->original_start_timestamp.value()) {
+        continue;
+      }
+    }
+
```
andrejtonev, line 216: Can we merge this with the `IsEdgeVisible`?

src/storage/v2/inmemory/label_property_index.cpp
```diff
@@ -202,6 +202,12 @@ void InMemoryLabelPropertyIndex::Iterable::Iterator::AdvanceUntilValid() {
       continue;
     }
 
+    if (self_->transaction_->original_start_timestamp.has_value()) {
+      if (index_iterator_->timestamp >= self_->transaction_->original_start_timestamp.value()) {
+        continue;
+      }
+    }
+
```
andrejtonev, line 210: This logic gets repeated a lot, can we define it in a single place?

src/storage/v2/inmemory/storage.cpp
```diff
@@ -1039,8 +1039,30 @@ utils::BasicResult<StorageManipulationError, void> InMemoryStorage::InMemoryAcce
...
+
+  if (result.HasError()) {
+    return result;
+  }
+
+  FinalizeTransaction();
+
+  auto original_start_timestamp = !transaction_.original_start_timestamp.has_value()
+                                      ? transaction_.start_timestamp
+                                      : transaction_.original_start_timestamp.value();
```
andrejtonev, line 1053: It's more readable if you reverse the order (check for a positive result)

src/storage/v2/inmemory/storage.cpp
```diff
@@ -1039,8 +1039,30 @@ utils::BasicResult<StorageManipulationError, void> InMemoryStorage::InMemoryAcce
...
+
+  auto *mem_storage = static_cast<InMemoryStorage *>(storage_);
+
+  auto new_transaction = mem_storage->CreateTransaction(transaction_.isolation_level, transaction_.storage_mode);
```
andrejtonev, line 1057: Might need a way to kill a periodic commit transaction using a global uuid.

src/query/cypher_query_interpreter.cpp
```diff
@@ -146,8 +147,8 @@ std::shared_ptr<PlanWrapper> CypherQueryToPlan(uint64_t hash, AstStorage ast_sto
     }
   }
 
-  auto plan = std::make_shared<PlanWrapper>(
-      MakeLogicalPlan(std::move(ast_storage), query, parameters, db_accessor, predefined_identifiers));
+  auto plan = std::make_shared<PlanWrapper>(MakeLogicalPlan(std::move(ast_storage), query, parameters, db_accessor,
+                                                            std::move(db_acc), predefined_identifiers));
```
Ignition, line 151: Any plan which holds onto DatabaseAccessProtector should not be cached. Need to do analysis to say this query is not cachable.

src/query/plan/operator.cpp
```diff
@@ -5986,8 +5986,9 @@ UniqueCursorPtr RollUpApply::MakeCursor(utils::MemoryResource *mem) const {
   return MakeUniqueCursorPtr<RollUpApplyCursor>(mem, *this, mem);
 }
 
-PeriodicCommit::PeriodicCommit(std::shared_ptr<LogicalOperator> &&input, Expression *commit_frequency)
-    : input_(std::move(input)), commit_frequency_(commit_frequency) {}
+PeriodicCommit::PeriodicCommit(std::shared_ptr<LogicalOperator> &&input, Expression *commit_frequency,
+                               DatabaseAccessProtector db_acc)
```
Ignition, line 5990: Don't put `DatabaseAccessProtector` in the Operator, pass it at execution time using the `ExecutionContext`

src/query/plan/read_write_type_checker.hpp
```diff
@@ -87,6 +87,13 @@ struct ReadWriteTypeChecker : public virtual HierarchicalLogicalOperatorVisitor
   bool PreVisit(CallProcedure &) override;
   bool PreVisit(Foreach &) override;
 
+  bool PreVisit(Apply &) override;
+  bool PreVisit(IndexedJoin &) override;
+  bool PreVisit(HashJoin &) override;
+  bool PreVisit(RollUpApply &) override;
```
Ignition, line 93: Why did these get introduced?

src/query/trigger.cpp
```diff
@@ -158,13 +158,13 @@ Trigger::Trigger(std::string name, const std::string &query, const storage::Prop
       event_type_{event_type},
       owner_{std::move(owner)} {
   // We check immediately if the query is valid by trying to create a plan.
-  GetPlan(db_accessor);
+  GetPlan(db_accessor, 0);
```
Ignition, line 161: `DatabaseAccessProtector` is `std::any` which is why this compiles...but it doesn't do any protection

Ignition (changes requested):

src/query/cypher_query_interpreter.cpp
Ignition: This likely isn't even needed if we just use ExecutionContext instead.

src/storage/v2/inmemory/storage.cpp
Ignition: @andrejtonev this should be fine, we already decoupled interpreter txn id from the storage txn id.

src/query/cypher_query_interpreter.cpp
Josipmrden: yes, I will pass it inside ExecutionContext, I didn't think of this beforehand

src/query/plan/read_write_type_checker.hpp
Josipmrden: it was a random bug that I encountered when trying to do a write query on a replica. Apparently since these didn't get written, I was able to perform a write query on the replica for the PeriodicSubquery. This is the leftover from other PR's since, for example CALL didn't get inspected if inside there was some write operation.

src/query/trigger.cpp
Josipmrden: removed the db_acc so this argument is also removed.

Josipmrden:
TODO: Profile needs to be disabled on periodic commit since deltas will get deleted.

Josipmrden:
// TODO: Disable periodic commit in explicit transactions -> I'm not sure how this is even possible to achieve, since at this point the transactions are having different timestamps

src/storage/v2/inmemory/edge_type_index.cpp
Josipmrden: It's a bit tricky since this is not tied to the mvcc logic, but to keeping the initial iterator from indices in place. I have made the logic inside one place in `indices_utils.hpp`.
The `IsEdgeVisible` is again called during edge creation so I'm not even sure what to pass to the edge at that point, which is a correct timestamp or not. And even if I introduce an optional, it seems like there is a lot going on in that method.

src/storage/v2/inmemory/label_property_index.cpp
Josipmrden: added it under `indices_utils.hpp`

src/storage/v2/inmemory/storage.cpp
Ignition: https://en.cppreference.com/w/cpp/utility/optional/value_or

src/storage/v2/inmemory/storage.cpp
Josipmrden: I have tested this. Interpreter still holds the initial transaction ID and that one is used to kill the periodic transaction successfully.
The transaction ID's that are subsequent in committing don't affect this flow.

andrejtonev (approved):
Good luck :D

kgolubic:
For release notes I'll use:

- Added periodic commit transaction logic.
[#2217](https://github.com/memgraph/memgraph/pull/2217)

PR #2252, opened 2024-07-30
Better error messages in OIDC SSO

imilinovic:
Add better error messages in OIDC SSO and make sure to always throw an useful message on failure.


imilinovic:
### Tracking


### Standard development
- [ ] Update unit/E2E tests


### CI Testing Labels


### Documentation checklist
    - **[ Release note text ]**

src/auth/reference_modules/oidc.py
```diff
@@ -57,7 +57,7 @@ def _parse_response(response):
 def _load_role_mappings(raw_role_mappings: str) -> dict:
     if raw_role_mappings:
         role_mapping = {}
-        raw_role_mappings = raw_role_mappings.split(";")
+        raw_role_mappings = raw_role_mappings.strip(" ;").split(";")
```
andrejtonev, line 60: Can you just give me an example here?

andrejtonev (approved):

src/auth/reference_modules/oidc.py
imilinovic: This is just so when you have trailing ; it is ignored instead of throwing an error

Usually it would be `sso-role1:mg_role1; sso-role2:mg_role2`
Now it can also be `sso-role1:mg_role1; sso-role2:mg_role2;`

Docs already specify how it should be but one user mistakenly added the `;` at the end and I thought it wouldn't hurt to add this.

andrejtonev: I would rather split and then control the array

andrejtonev: Just to clarify why.
Your fix is very specific. You will have another user inputting "sso-role1:mg_role1;;sso-role2:mg_role2" and then we are back to the same problem. 
Split and then check the result. If empty remove, and throw if not compliant in a way we can't handle.

src/auth/reference_modules/oidc.py
```diff
@@ -57,12 +56,18 @@ def _parse_response(response):
...
+            if len(mapping_list) == 0:
+                continue
+            if len(mapping_list) != 2:
+                raise ValueError(f"Invalid role mapping: {mapping}")
+            idp_role, mg_role = mapping_list
             role_mapping[idp_role.strip()] = mg_role.strip()
         return role_mapping
-    return {}
+
+    raise ValueError("Missing role mappings")
```
andrejtonev, line 70: Is this an error?
Do we have some default behavior?

imilinovic: You cannot use SSO without role mappings.
This would otherwise fail later and throw `Cannot map role {role} to Memgraph role`.
So this is unecessary in a way but I would say missing the whole env variable deserves a separate error.

PR #2277, opened 2024-08-13
Use query memory for the call procedure

Ignition:
Call procedure used its own dedicated memory. This should not be needed, using query's memory instead.


Ignition:
Work merged as part of #2561

PR #2295, opened 2024-08-28
Remove system replication out of experimental features

### LINKED ISSUE #2227: Remove system-replication out of experimental features
"Just" remove the value/handling from `--experimental-enabled`, make the feature part of `replication` feature.

User quote: "There should be just two features, replication and HA. Replication is classic database replication where you have another copy of the database; primary and replica. HA automatic failover is a mechanism of making sure a given database is available if one node has an issue or breaking in any way."

DavIvek:
This PR removes system replication out of experimental features.

closes #2227


src/replication_handler/include/replication_handler/replication_handler.hpp
```diff
@@ -51,8 +51,7 @@ void SystemRestore(replication::ReplicationClient &client, system::System &syste
       [](auto &state) { return state != memgraph::replication::ReplicationClient::State::RECOVERY; });
   {
     using enum memgraph::flags::Experiments;
-    bool full_system_replication =
-        flags::AreExperimentsEnabled(SYSTEM_REPLICATION) && license::global_license_checker.IsEnterpriseValidFast();
+    bool full_system_replication = license::global_license_checker.IsEnterpriseValidFast();
```
gitbuda, line 54: This is now inconsistent because there is no such a thing as `fully_system_replication`, there is only `replication`, just remove the variable and replace it with `license::global_license_checker.IsEnterpriseValidFast()`.

DavIvek: I was thinking of doing a cleanup, I just want to see if all the tests pass

src/replication_handler/include/replication_handler/replication_handler.hpp
```diff
@@ -51,8 +51,7 @@ void SystemRestore(replication::ReplicationClient &client, system::System &syste
       [](auto &state) { return state != memgraph::replication::ReplicationClient::State::RECOVERY; });
   {
     using enum memgraph::flags::Experiments;
```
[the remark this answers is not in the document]
DavIvek: yes

tests/e2e/replication_experimental/multitenancy.py
```diff
@@ -109,7 +106,7 @@ def do_manual_setting_up(connection):
 MEMGRAPH_INSTANCES_DESCRIPTION_WITH_RECOVERY = {
     "replica_1": {
         "args": [
-            "--experimental-enabled=system-replication",
+            "--replication-system-replication-enabled=true",
```
gitbuda, line 109: I guess this is just not removed yet (it's ok because the PR is still in draft)? But the problem is if it's actually working 😂 In this config, the test should crash and not pass 👀

src/flags/experimental.cpp
```diff
@@ -48,8 +47,7 @@ auto const canonicalize_string = [](auto &&rng) {
 
 namespace memgraph::flags {
 
-auto const mapping = std::map{std::pair{"system-replication"sv, Experiments::SYSTEM_REPLICATION},
-                              std::pair{"text-search"sv, Experiments::TEXT_SEARCH},
+auto const mapping = std::map{std::pair{"text-search"sv, Experiments::TEXT_SEARCH},
```
andrejtonev, line 50: This change will make this a breaking change. 
Users who didn't remove "system-replication" from the list will get an error.
I don't think many people used this feature, so just add the `breaking` label to the PR.

src/replication_handler/include/replication_handler/replication_handler.hpp
```diff
@@ -66,7 +64,7 @@ void SystemRestore(replication::ReplicationClient &client, system::System &syste
         return std::nullopt;
       });
 
-      if (full_system_replication) {
+      if (is_enterprise) {
```
andrejtonev, line 67: Did we decide which system replication is enterprise? 
Are both the multi-tenant and auth enterprise?
MT has to be, but AUTH could be community.

DavIvek: Did behavior change with the introduced line changes?

DavIvek: As I understand, purpose of this work is just to remove system replication out of experimental features and leave everything else as it was.

DavIvek: Both are enterprise as I know.

andrejtonev (approved):

PR #2332, opened 2024-09-16
Add create ha test

as51340:
Added `hacreate` test meant to test high availability capabilities of Memgraph. Test starts 6 instances (3 data instances and 3 coordinators) and uses 2 generators to generate operations. Client generator starts one thread per instance and every 2 seconds chooses between `add-nodes` operation (run on data instances) and `show-instances` operations (run on coordinators). Operation `add-nodes` creates a batch of nodes with monotonically increasing IDs and operation `show-instances` runs `SHOW INSTANCES` query. Nemesis generator turns on every ~30'' and then randomly chooses between killing current cluster's leader and cluster's main instance. The test is considered valid if all coordinators have the same view of the cluster (`show-instances` successful), if there are no `show-instances` results where less than 3 coordinators are seen, if there is never more than one main in the cluster (as seen by coordinators), if there are never less than 6 instances in the result of `SHOW INSTANCES`, if there are no failed `setup-cluster` and `show-instances` operations and if several consistency guarantees hold. At the end of the test we retrieve all nodes with `MATCH (n:Node) RETURN n.id as id`. Ideally, we get on MAIN and all REPLICAS exactly the same result with monotonically increasing ids `e.g [1, 2, 3, .... 1000000]`. However, many things can go wrong, hence we check: 
1. There are no duplicated IDs returned because that would mean some vertices got replicated multiple times
2. There are no missing IDs because that would mean some vertices didn't get replicated at all.
3. We have 2 measures of consistency, 1st is based on Jaccard similarity and the 2nd on Hamming similarity.
  a) Jaccard similarity tests that all IDs that should exist (based on client's request for creation) actually do exist
  b) Hamming similarity tests additionally to Jaccard similarity also that the order in which IDs get replicated is correct.

Test will be run every night at 12:00 AM as part of new workflow `stress_jepsen.yaml`.

### DIFF
--- a/.github/workflows/check_flakiness.yaml
+++ b/.github/workflows/check_flakiness.yaml
@@ -11,27 +11,35 @@ on:
       community_core:
         type: boolean
         default: true
+        description: ""
       coverage_core:
         type: boolean
         default: true
+        description: ""
       debug_core:
         type: boolean
         default: true
+        description: ""
       debug_integration:
         type: boolean
         default: true
+        description: ""
       jepsen_core:
         type: boolean
         default: true
+        description: ""
       release_core:
         type: boolean
         default: true
+        description: ""
       release_e2e:
         type: boolean
         default: true
+        description: ""
       release_stress:
         type: boolean
         default: true
+        description: ""
       flakiness_runs:
         type: number
         default: 3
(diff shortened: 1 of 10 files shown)


as51340:
---
___


### Tracking


### Standard development
- [ ] Update unit/E2E tests


### CI Testing Labels


### Documentation checklist
    - **[ Release note text ]**

Ignition (approved):

PR #2336, opened 2024-09-17
Add correctness to indices for SetEdgeFrom and SetEdgeTo

Josipmrden:
Correctness modifications to edge indices when there is a change in setting the edge's from and to vertices.

### DIFF
--- a/src/storage/v2/edge_accessor.cpp
+++ b/src/storage/v2/edge_accessor.cpp
@@ -45,11 +45,9 @@ bool EdgeAccessor::IsDeleted() const {
 }
 
 bool EdgeAccessor::IsVisible(const View view) const {
-  bool exists = true;
-  bool deleted = true;
-  // When edges don't have properties, their isolation level is still dictated by MVCC ->
-  // iterate over the deltas of the from_vertex_ and see which deltas can be applied on edges.
-  if (!storage_->config_.salient.items.properties_on_edges) {
+  auto check_from_vertex_integrity = [&view, this]() -> bool {
+    bool exists = true;
+    bool deleted = true;
     Delta *delta = nullptr;
     {
       auto guard = std::shared_lock{from_vertex_->lock};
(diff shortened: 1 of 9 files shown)


src/storage/v2/inmemory/edge_type_index.hpp
```diff
@@ -35,8 +35,22 @@ class InMemoryEdgeTypeIndex : public storage::EdgeTypeIndex {
...
-    bool operator<(const Entry &rhs) const { return edge->gid < rhs.edge->gid; }
-    bool operator==(const Entry &rhs) const { return edge->gid == rhs.edge->gid; }
+    bool operator<(const Entry &rhs) const {
+      if (edge->gid < rhs.edge->gid) {
+        return true;
+      }
+
+      if (edge->gid > rhs.edge->gid) {
+        return false;
+      }
```
Ignition, line 45: No need, the lexical ordering is done by the tuple below.

src/storage/v2/inmemory/edge_type_index.hpp
```diff
@@ -35,8 +35,22 @@ class InMemoryEdgeTypeIndex : public storage::EdgeTypeIndex {
...
+
+      return std::make_tuple(edge, from_vertex, to_vertex, timestamp) <
+             std::make_tuple(rhs.edge, rhs.from_vertex, rhs.to_vertex, rhs.timestamp);
```
Ignition, line 48: prefer std::tie

src/storage/v2/edge_accessor.cpp
```diff
@@ -70,51 +68,112 @@ bool EdgeAccessor::IsVisible(const View view) const {
...
         case Delta::Action::DELETE_OBJECT:
           break;
-        case Delta::Action::ADD_OUT_EDGE: {  // relevant for the from_vertex_ -> we just deleted the edge
+        case Delta::Action::ADD_OUT_EDGE: {
+          if (delta.vertex_edge.edge == edge_ && delta.vertex_edge.edge_type == edge_type_) {
+            attached = true;
+          }
+          break;
+        }
+        case Delta::Action::REMOVE_OUT_EDGE: {
```
andrejtonev, line 77: Check for edge type here as well?

andrejtonev: Do we have to traverse the whole chain?
Is there a case where an edge has been deleted, but could reaper afterwards?
I guess if you are changing the type back and forth?

src/storage/v2/edge_accessor.cpp
```diff
@@ -70,51 +68,112 @@ bool EdgeAccessor::IsVisible(const View view) const {
...
+        case Delta::Action::DELETE_OBJECT:
+          break;
+        case Delta::Action::ADD_IN_EDGE: {
+          if (delta.vertex_edge.edge == edge_ && delta.vertex_edge.edge_type == edge_type_) {
+            attached = true;
           }
           break;
         }
-        case Delta::Action::REMOVE_OUT_EDGE: {  // also relevant for the from_vertex_ -> we just added the edge
+        case Delta::Action::REMOVE_IN_EDGE: {
```
andrejtonev, line 115: Type check?

src/storage/v2/edge_accessor.cpp
```diff
@@ -70,51 +68,112 @@ bool EdgeAccessor::IsVisible(const View view) const {
...
           }
           break;
         }
       }
     });
+    return attached;
+  };
+  auto check_presence_of_edge = [&view, this]() -> bool {
+    bool deleted = true;
+    bool exists = true;
```
andrejtonev, line 127: Do we need two bools to define this?
Is one not enough?

src/storage/v2/inmemory/edge_type_index.cpp
```diff
@@ -185,14 +185,8 @@ void InMemoryEdgeTypeIndex::UpdateOnEdgeModification(Vertex *old_from, Vertex *o
...
     return;
   }
-  auto acc = it->second.access();
-
-  auto entry_to_update = std::ranges::find_if(acc, [&](const auto &entry) {
-    return entry.from_vertex == old_from && entry.to_vertex == old_to && entry.edge == edge_ref.ptr;
-  });
 
-  acc.remove(Entry{entry_to_update->from_vertex, entry_to_update->to_vertex, entry_to_update->edge,
-                   entry_to_update->timestamp});
```
andrejtonev, line 195: Is the timestamp enough here, so we don't need to remove the old entry?

andrejtonev: Are we handling the obsolete entries via GC?

src/storage/v2/edge_accessor.cpp
Josipmrden: why do we want to check edge here? it has been detached from the vertex. it's not part of the vertex

andrejtonev: Maybe I misunderstood how EdgeTypeChange works, but, in that case, wouldn't we have a remove edge and than add edge with the same gid, just different types. So an edge could still be attached even if there's a REMOVE delta.

src/storage/v2/edge_accessor.cpp
Josipmrden: I just left it from previous code, they were using 2 bools and use 2 bools in the return check

src/storage/v2/edge_accessor.cpp
Josipmrden: `wouldn't we have a remove edge and than add edge with the same gid, just different types.`

yes, that's the situation now

andrejtonev: So why don't we need to check the type?

src/storage/v2/edge_accessor.cpp
Josipmrden: pushed one bool instead of 2

src/storage/v2/edge_accessor.cpp
andrejtonev: Fine as is. The type check would be redundant.

andrejtonev (approved):

PR #2342, opened 2024-09-20
Fix: ScanAllById regardless of function case

Josipmrden:
Bug was introduced when the function name was not passed as uppercase. It would result of ScanAllById operator not being utilized.

### DIFF
--- a/tests/e2e/awesome_functions/id.py
+++ b/tests/e2e/awesome_functions/id.py
@@ -1,0 +1,57 @@
+# Copyright 2023 Memgraph Ltd.
+#
+# Use of this software is governed by the Business Source License
+# included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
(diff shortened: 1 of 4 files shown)


tests/e2e/awesome_functions/id.py
```diff
@@ -0,0 +1,57 @@
+# Copyright 2023 Memgraph Ltd.
+#
+# Use of this software is governed by the Business Source License
+# included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
+# License, and you may not use this file except in compliance with the Business Source License.
+#
+# As of the Change Date specified in that file, in accordance with
+# the Business Source License, use of this software will be governed
+# by the Apache License, Version 2.0, included in the file
+# licenses/APL.txt.
```
imilinovic, line 10: Somehow year is wrong

imilinovic (approved):

andrejtonev (approved):

PR #2344, opened 2024-09-23
Improvements on e2e testing infrastructure

as51340:
Added check loop after stopping and killing each instance to check that the process actually got terminated with SIGTERM and SIGKILL. All replication and HA tests now use SIGKILL signal since it causes immediate termination compared to SIGTERM which could take some time. Improved test_replication_correct_replica_chosen_up_to_date_data from single_coordinator.py since it could result in non-deterministic execution. Added .flake8 file to tests/ folder to avoid unnecessary warnings. Refactored coordinator.py to stop using fixtures since same thing could be tested without unnecessary deletions of data directories and instance's restarts. In interactive_mg_runner.py, is_port_in_use is replaced by connectable_port function, now used both before starting to check if no other instance runs on that port and after start to allow clients to use bolt server immediately when is available, not before. We use python's socket package to check when the socket is ready compared to previous way of constantly creating mgclient connections. These also reduced tests' running times and now we are back under an hour.  Cleanup function was deleted since cleaning is done on memgraph.py level and on test level. ignore_auth_failure flag is removed since it wasn't necessary. Method execute_setup_queries will not execute only if there are some queries to be run and it is responsible for creating its connection. cleanup function in the runner.py runs only if cluster is defined in workloads.yaml otherwise we assume that developer is responsible for managing cleanings. Index replication removed because of flakiness.


as51340:
### Tracking


### Standard development
- [ ] Update unit/E2E tests


### CI Testing Labels


### Documentation checklist
- [x] Add the bug / feature / tests label
    - **[ Release note text ]**
- [x] **[ Tag someone from docs team ]** @hal-eisen-MG I think this can go with 2.21

Ignition (approved):

PR #2421, opened 2024-10-15
Bugfix: Unify instance callbacks in HA

as51340:
Running time for Jepsen test increased to 1h. Cleaned API of the state machine and cluster state. Added StateCheckRPC sent by coordinators to data instances to check in which state are they currently. Upon receiving PromoteToMainRPC, writing will be automatically enabled. Merged all callbacks of data instances into one type to simplify the logic. This lead to simplification in ReplicationInstanceClient and ReplicationInstanceConnector. Failover method modifies only raft state. All RPCs are sent from callbacks to reconcile data instance's state.


tests/CMakeLists.txt
```diff
@@ -1,28 +1,28 @@
...
-add_subdirectory(unit)
-
-# property based test binaries
-add_subdirectory(property_based)
-
-# integration test binaries
-add_subdirectory(integration)
-
-# e2e test binaries
+# add_subdirectory(benchmark)
```
andrejtonev, line 1: Remove before commit

tests/e2e/CMakeLists.txt
```diff
@@ -37,67 +37,67 @@ function(copy_e2e_files TARGET_PREFIX FILE_NAME)
...
-add_subdirectory(query_modules)
-add_subdirectory(constraints)
-add_subdirectory(inspect_query)
-add_subdirectory(filter_info)
-add_subdirectory(queries)
-add_subdirectory(query_modules_storage_modes)
-add_subdirectory(garbage_collection)
-add_subdirectory(query_planning)
-add_subdirectory(awesome_functions)
+# add_subdirectory(fine_grained_access)
```
andrejtonev, line 40: Remove before commit

src/coordination/coordinator_instance.cpp
```diff
@@ -413,393 +397,137 @@ auto CoordinatorInstance::ReconcileClusterState_() -> ReconcileClusterStateStatu
     if (raft_state_->IsLockOpened()) {
```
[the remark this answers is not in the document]
as51340: it means raft state isn't completely consistent.

src/coordination/coordinator_instance.cpp
```diff
@@ -413,393 +397,137 @@ auto CoordinatorInstance::ReconcileClusterState_() -> ReconcileClusterStateStatu
...
       spdlog::error(
           "Don't execute actions while coordinator is not connected to other coordinators, it could lead to "
-          "diverging of cluster log.");
+          "divergence of cluster logs.");
     } else {
       return ReconcileClusterStateStatus::SUCCESS;
     }
   }
```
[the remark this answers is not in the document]
as51340: comment above

src/coordination/coordinator_instance.cpp
```diff
@@ -323,6 +334,8 @@ auto CoordinatorInstance::ShowInstances() const -> std::vector<InstanceStatus> {
 auto CoordinatorInstance::ReconcileClusterState() -> ReconcileClusterStateStatus {
   utils::ExponentialBackoff backoff{std::chrono::milliseconds(1000), std::chrono::seconds(60)};
 
+  is_leader_ready_.store(false, std::memory_order_release);
```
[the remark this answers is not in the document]
as51340: we talked about it yesterday.. 
it's fifo,
multiple reconciliations shouldn't happen at the same time

tests/e2e/high_availability/CMakeLists.txt
```diff
@@ -6,7 +6,6 @@ copy_e2e_python_files(high_availability coord_cluster_registration.py)
 copy_e2e_python_files(high_availability distributed_coords.py)
 copy_e2e_python_files(high_availability disable_writing_on_main_after_restart.py)
 copy_e2e_python_files(high_availability manual_setting_replicas.py)
-copy_e2e_python_files(high_availability not_replicate_from_old_main.py)
```
as51340, line 9: removed because this test was testing what happens with replica when registered in two clusters. Before, when the replica got registered on the 2nd cluster coordinator from the 1st cluster would never try to swap uuid. Now it does

src/coordination/coordinator_instance.cpp
```diff
@@ -813,9 +541,7 @@ auto CoordinatorInstance::SetReplicationInstanceToMain(std::string_view instance
     return SetInstanceToMainCoordinatorStatus::MAIN_ALREADY_EXISTS;
   }
 
-  auto const is_new_main = [&instance_name](ReplicationInstanceConnector const &instance) {
-    return instance.InstanceName() == instance_name;
-  };
+  auto const is_new_main = [&instance_name](auto &&instance) { return instance.InstanceName() == instance_name; };
```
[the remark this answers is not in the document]
as51340: changed the logic of these user parts

src/coordination/coordinator_instance.cpp
```diff
@@ -1056,63 +679,38 @@ auto CoordinatorInstance::UnregisterReplicationInstance(std::string_view instanc
...
     return raft_state_->IsCurrentMain(instance.InstanceName()) && instance.IsAlive();
   };
 
   if (is_current_main(*inst_to_remove)) {
     return UnregisterInstanceCoordinatorStatus::IS_MAIN;
   }
 
-  if (!raft_state_->AppendOpenLock()) {
-    return UnregisterInstanceCoordinatorStatus::FAILED_TO_OPEN_LOCK;
+  if (!raft_state_->AppendUnregisterReplicationInstanceLog(instance_name)) {
```
andrejtonev, line 697: How can we append without opening/closing the lock?

src/coordination/coordinator_instance.cpp
```diff
@@ -1144,304 +742,241 @@ auto CoordinatorInstance::AddCoordinatorInstance(CoordinatorToCoordinatorConfig
...
-  // enabling writing.
-  if (raft_state_->IsCurrentMain(repl_instance.InstanceName())) {
-    spdlog::trace(
-        "Instance {} wasn't considered alive but the health ping passed now. Instance considered current main, trying "
-        "to enable writing.",
-        repl_instance_name);
-    if (!repl_instance.EnableWritingOnMain()) {
-      spdlog::error("Failed to enable writing on instance {}.", repl_instance_name);
-      return;
+  auto lock = std::unique_lock{coord_instance_lock_};
```
andrejtonev, line 757: Shouldn't this lock be before the ready and locked checks?

src/coordination/coordinator_instance.cpp
```diff
@@ -1144,304 +742,241 @@ auto CoordinatorInstance::AddCoordinatorInstance(CoordinatorToCoordinatorConfig
...
+            break;
+          };
+          case FailoverStatus::FAILURE_LOCK_CLOSED: {
+            spdlog::trace("Opening the lock failed during failover.");
+            break;
+          };
+          case FailoverStatus::FAILURE_LOCK_OPENED: {
+            spdlog::trace(
+                "Writing to Raft log partially succeeded during failover, reconciliation task will be scheduled.");
+            thread_pool_.AddTask([this]() { this->ReconcileClusterState(); });
```
andrejtonev, line 798: I would put this in a function TryReconcile where we first set the ready to false and then put it in the queue.

src/coordination/coordinator_instance.cpp
```diff
@@ -1144,304 +742,241 @@ auto CoordinatorInstance::AddCoordinatorInstance(CoordinatorToCoordinatorConfig
...
   }
 
   if (!is_leader_ready_) {
-    spdlog::trace("Returning from ReplicaSuccessCallback since the leader is not ready.");
+    spdlog::trace("Leader is not ready, not executing instance fail callback.");
     return;
   }
 
-  auto &repl_instance = FindReplicationInstance(repl_instance_name);
+  auto lock = std::unique_lock{coord_instance_lock_};
```
andrejtonev, line 845: check locks position

as51340:
https://github.com/memgraph/memgraph/pull/2421#discussion_r1806011764

I don't think multiple reconcile requests could be in fifo. We add such task if failover fails (the lock remains opened) or when becoming leader. When lock is opened, other callbacks won't get executed

src/coordination/coordinator_instance.cpp
```diff
@@ -413,393 +364,149 @@ auto CoordinatorInstance::ReconcileClusterState_() -> ReconcileClusterStateStatu
...
-    instance.SetCallbacks(&CoordinatorInstance::DemoteSuccessCallback, &CoordinatorInstance::DemoteFailCallback);
-  });
-  spdlog::trace("Set demote callbacks to all dead instances.");
-
-  std::ranges::for_each(repl_instances_, [](auto &instance) {
-    MG_ASSERT(instance.GetSuccessCallback() != nullptr && instance.GetFailCallback() != nullptr,
-              "Callbacks are not properly set");
-    instance.StartFrequentCheck();
-  });
+  std::ranges::for_each(repl_instances_, [](auto &&instance) { instance.StartStateCheck(); });
```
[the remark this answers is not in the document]
as51340: we always add task ReconcileClusterState and that functions runs in while loop and ends only when success is returned.

src/coordination/coordinator_instance.cpp
```diff
@@ -413,393 +364,149 @@ auto CoordinatorInstance::ReconcileClusterState_() -> ReconcileClusterStateStatu
...
-  MG_ASSERT(!raft_state_->IsLockOpened(), "After failover we need to be in healthy state.");
-
-  // TODO: (andi) What happens in this case? How to come out of this situation?
-  if (!new_main.EnableWritingOnMain()) {
-    spdlog::error("Failover successful but couldn't enable writing on instance.");
-  } else {
-    spdlog::info("Failover successful! Instance {} promoted to main.", new_main.InstanceName());
+    return FailoverStatus::FAILURE_LOCK_OPENED;
   }
+  return FailoverStatus::SUCCESS;
```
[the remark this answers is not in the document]
as51340: we do it in the callback same as other actions

src/coordination/coordinator_instance.cpp
```diff
@@ -824,136 +529,115 @@ auto CoordinatorInstance::SetReplicationInstanceToMain(std::string_view instance
...
   if (raft_state_->IsLockOpened()) {
     return RegisterInstanceCoordinatorStatus::LOCK_OPENED;
   }
 
-  // TODO: (andi) Change that this is being asked from raft state
-  if (std::ranges::any_of(repl_instances_,
-                          [instance_name = config.instance_name](ReplicationInstanceConnector const &instance) {
-                            return instance.InstanceName() == instance_name;
-                          })) {
+  if (!is_leader_ready_) {
```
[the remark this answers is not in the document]
as51340: no it's not. Ideally we remove IsLockOpen and I plan to do that soon

src/coordination/coordinator_instance.cpp
as51340: because we only do one action. Locks are important when having multiple appends

src/coordination/coordinator_instance.cpp
as51340: is_leader_ready is atomic and raft_state is completely unconnected with coord_instance_lock_

src/coordination/coordinator_instance.cpp
as51340: what's locks position?

src/coordination/coordinator_instance.cpp
as51340: makes sense

src/coordination/coordinator_instance.cpp
andrejtonev: Depends on how the lock is serializing the commands.
is_leader_ready is not protected under the lock beacuse we need to change it quickly inside the reaft logs.
IsLockOpen should be protected under the lock, since otherwise you could check it, then wait for the unique lock; some other thread is holding the unique lock and changes the IsLockOpen. When you finally get the unique_lock, the IsLockOpen has changed.

src/coordination/coordinator_instance.cpp
andrejtonev: Fine while logic is protected by the mtx

src/coordination/coordinator_instance.cpp
as51340: will be changed when I remove opening the lock and is_leader_ready. Now this is protected with IsLockOpened

hal-eisen-MG:
@as51340 I'm not understanding the description. Would it be possible to rephrase from the user's point of view? How does the change to the callbacks impact them?

PR #2424, opened 2024-10-15
Compilation error if module tries to set mgp::memory

andrejtonev:
Will cause compilation errors when compiling modules that set `mgp::memory`.
Module implementations must use `MemoryDispatcherGuard`, otherwise instance will segfault.

### DIFF
--- a/include/mgp.hpp
+++ b/include/mgp.hpp
@@ -126,11 +126,12 @@ inline bool IsThisThreadRegistered() noexcept { return current_memory.has_value(
 
 // Leaving "memory" variable here in order to generate a better module compilation error.
 // It should never be used; instead use MemoryDispatcherGuard.
-inline struct UnsupportedMgpMemory {
+struct UnsupportedMgpMemory {
   UnsupportedMgpMemory() = default;
   void operator=(mgp_memory *)
       __attribute__((error("mgp::memory must not be used as of v2.18.1. Please use MemoryDispatcherGuard instead.")));
-} memory;
+};
+inline UnsupportedMgpMemory memory;  // NOSONAR
 
 class MemoryDispatcherGuard final {
  public:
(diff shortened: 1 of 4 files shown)


andrejtonev:
### Tracking


### Standard development
- [ ] Update unit/E2E tests


### CI Testing Labels


### Documentation checklist
    - **[ Release note text ]**

include/mg_exceptions.hpp
```diff
@@ -105,4 +105,10 @@ struct SerializationException : public std::exception {
   const char *what() const noexcept override { return "Error in serialization!"; }
 };
 
+struct UnregisteredMemoryUsage : public std::exception {
+  const char *what() const noexcept override {
+    return "Thread not registered for memory usage. Please use MemoryDispatcher.";
```
imilinovic, line 110: Maybe its better to say MemoryDispatcherGuard instead of MemoryDispatcher since thats what the user sees/uses

include/mgp.hpp
```diff
@@ -128,10 +128,21 @@ inline bool IsThisThreadRegistered() noexcept { return current_memory.has_value(
...
+  template <typename T>
+  struct unsupported_tag {
+    static const bool value = false;
+  };
+
+  UnsupportedMgpMemory() {}
+  template <typename T>
+  UnsupportedMgpMemory(T *) {
+    static_assert(unsupported_tag<T>::value,
+                  "mgp::memory must not be used as of v2.18.1. Please use MemoryDispatcher instead.");
```
imilinovic, line 143: Same comment as above

andrejtonev:
One of the reasons we had mgp::memory is so that MemHandlerCallback always succeeded. 
Need to think about the implications of having it throw.

andrejtonev:
> One of the reasons we had mgp::memory is so that MemHandlerCallback always succeeded.
> Need to think about the implications of having it throw.
I think I was overthinking this. The mgp::memory is protected now. 
If a user doesn't use the guard, the memory won't be set and we will segfault. Just like if you forgot to set the mgp::memory before.

Ignition (approved):

andrejtonev:
Issue #2426 is fixed only by updating the procedures to use MemoryDispatcherGuard. 
This PR allows us to generate a compilation error if someone is compiling modules that use mpg::memory directly.

PR #2449, opened 2024-10-29
Spatial rewrite point.withinbbox to use point index and make point index usable without WITH clause

imilinovic:
Make planner utilize point index for point.withinbbox queries and make point.distance queries usable without WITH statement.

Example
`MATCH (m:L1) WHERE NOT point.withinbbox(m.prop, point({x:-1, y:-1, crs:"cartesian"}), point({x:1, y:1, crs:"cartesian"})) RETURN m;`

Point.distance queries are now usable as:
`MATCH (m:L1) WHERE point.distance(m.prop, point({x:0, y:0, z:0, crs:"wgs-84-3d"})) < 111320 RETURN m;`
instead of always having to use WITH clause like this:
`WITH point({x:0, y:0, z:0, crs:"wgs-84-3d"}) AS a MATCH (m:L1) WHERE point.distance(m.prop, a) >= 111320 RETURN m;`

### DIFF
--- a/tests/gql_behave/tests/memgraph_V1/features/spatial.feature
+++ b/tests/gql_behave/tests/memgraph_V1/features/spatial.feature
@@ -850,6 +850,17 @@ Feature: Spatial related features
 
         When executing query:
             """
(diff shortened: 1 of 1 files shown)


imilinovic:
Comment

Ignition (approved):
:+1: Just add a few more tests please

PR #2477, opened 2024-11-13
Point exact match leveraging spatial index

imilinovic:
…ways replaces the operator, sometimes when it shouldn't


imilinovic:
Comment

gitbuda:
The spatial feature set is not a priority in the short run; this can stay open until we get back.

PR #2497, opened 2024-11-22
Small optimisations

Ignition:


Ignition:
### Tracking


### Standard development
- [ ] Update unit/E2E tests


### CI Testing Labels


### Documentation checklist
    - **[ Release note text ]**

src/utils/skip_list.hpp
```diff
@@ -1036,17 +1036,27 @@ class SkipList final : detail::SkipListNode_base {
...
-        std::unique_lock<SpinLock> guards[kSkipListMaxHeight];
+
+        auto locked_count = 0;
+        TNode *locked[kSkipListMaxHeight];
+        auto guard = OnScopeExit{[&] {
+          for (auto i = 0; i != locked_count; ++i) {
+            locked[i]->lock.unlock();
+          }
+        }};
+
```
andrejtonev, line 1049: Does this give enough of a performance benefit to justify not using the std::unique_lock?

src/utils/skip_list.hpp
```diff
@@ -1060,8 +1070,6 @@ class SkipList final : detail::SkipListNode_base {
 
         MemoryResource *memoryResource = GetMemoryResource();
         void *ptr = memoryResource->Allocate(node_bytes, SkipListNodeAlign<TObj>());
-        // `calloc` would be faster, but the API has no such call.
-        memset(ptr, 0, node_bytes);
```
andrejtonev, line 1064: Just a nitpick. This and the 2 lines 
```
        Allocator<TNode> allocator(memoryResource);
        allocator.construct(new_node, top_layer, std::forward<TObjUniv>(object));
```
are divided by an empty line and comment. Feels like they are not connected, but in reality the reason why we don't need the memset is because we are constructing before any usage.
Can you just combine the lines into a single section?

andrejtonev: What do you think about extending the Allocator<>.allocate API to support user-defined alignment?
Then you don't need to allocate using the memory resources and construct using the allocator. Instead all would be handed by the same object

andrejtonev (approved):
lgtm

src/utils/skip_list.hpp
Ignition: Yeah, the expected number of locks is much less than 32. The constructor + destruction of unique_ptr was surprisingly noticable even when it was locking nothing.

src/utils/skip_list.hpp
Ignition: TNode is dynamic in size as there are a random number of next pointers. You need to allocate not using the allocator API and use the memory resource directly. 
We then construct the TNode header and populate the next pointers.

PR #2535, opened 2024-12-11
Prevent server crash

Ignition:
Don't ASSERT, throw query exception. It is possible that an ill formed query can lead to problems during planning. Analysis and good error reporting is difficult for this case, for now we will in general just throw. In future we will improve planner.


as51340 (approved):
:muscle:

Issue #2491, opened 2024-11-21
Query crashes Memgraph because it can't generate a valid plan


Josipmrden:
**Memgraph version**
v2.21.0

**Environment**
Docker

**Describe the bug**
The following query crashes Memgraph
```
MATCH (d_1)-[l_4]->(hyc_1)
WITH *
MATCH (d_4)-[l_5]->(hyc_1)
WITH *
MATCH (d_2)-[l_6]->(h_2)
WITH *
MATCH (d_5)-[l_7]->(h_2)
CALL {
  MATCH (d_1)-[]->(i)-[]-()<-[]-(hyc_1)
  RETURN 1 as i_1
}
CALL {
  WITH d_4, hyc_1
  MATCH (d_4)-[]->(i)-[]-()<-[]-(hyc_1)
  RETURN 1 as i_2
}
RETURN 1;
```

**To Reproduce**
Just run Memgraph with the query above

**Expected behavior**
The query should generate a valid plan at least once

**Logs**
```
[2024-11-21 09:10:25.289] [memgraph_log] [critical] 
Assertion failed in file  at line 155.
        Expression: 'valid_plan_found'
        Message: 'Could not create a valid query plan!'
```

Ignition:
Some investigation the problem is around:
```
CALL {
// BUG around (anonA)<-[anonB]-(hyc_1)
// the code detects anonA isn't bound so opts for edge scan for anonB
MATCH (d_1)-[]->(i)-[]-()<-[]-(hyc_1)
RETURN 1 as i_1
}
```
With all the WITH * there are no pattern variations, just one deterministic variation which makes a ScanByEdgeType which is invalid (no edge index it replace it with).

I feel the variable start code is wrong and don't account for the alternative permutations of `(d_1)-[]->(i), (i)-[]-(anon), (anon)<-[]-(hyc_1)`
FYI a `WITH d_1, hyc_1` inside the call, it avoids the issue...I assume that fixes the variable start issue....but I'm not sure why.

This kind of issue would also be fixed with an equality saturation approach. Issue here is that we do one substitution which leads to a dead end plan, alternative substitution would have yielded a valid plan.

gitbuda:
Yea, this is a deeper issue, @Ignition proposed that the proper way of detecting this should be the the plan analysis and detecting there is a problem. The best quick fix is to throw instead of crashing Memgraph.

PR #2543, opened 2024-12-12
Fix variable start planner not able to generate valid plans

Josipmrden:
Variable start planner fixed so it can always generate at least one valid plan. Valid plans are currently all the ones that start expanding from a node. There was an edge case where no valid plan could be constructed, as all constructed plans would start expanding from an edge without edge indices in place.

### DIFF
--- a/tests/e2e/edge_indices/edge_indices.py
+++ b/tests/e2e/edge_indices/edge_indices.py
@@ -39,7 +39,7 @@ def test_scan_all_by_edge_type_index_plan(memgraph):
 
     expected_explain = [
         " * Produce {prop}",
-        " * ScanAllByEdgeType (m)<-[r:TYPE]-(n)",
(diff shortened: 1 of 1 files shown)


Josipmrden:
### Tracking
- [X] **https://github.com/memgraph/memgraph/issues/2491**


### Documentation checklist
- [X] Add the documentation label
- [X] Add the bug / feature label
- [X] Add the milestone for which this feature is intended
- [X] Write a release note, including added/changed clauses
    - **Planner fixed so it can always generate at least one valid plan. Valid plans are currently all the ones that start expanding from a node. There was an edge case where no valid plan could be constructed, as all constructed plans would start expanding from an edge without edge indices in place.**
- [X] @katarinasupe

Ignition (approved):
Fix the remaining test, then LGTM

PR #2555, opened 2024-12-17
Add SHOW LICENSE INFO query

Josipmrden:
`SHOW LICENSE INFO` query was added to display a more granular view of what's the status with the license.


gitbuda (changes requested):
Let's add the e2e test for the `SHOW LICENSE INFO;` output 🙏

gitbuda (approved):

Issue #1848, opened 2024-03-22
No support for or expressions for labels


arminus:
**Memgraph version**
2.15.0

**Environment**
Windows/Docker

**Describe the bug**
Or expressions for labels are apparently not supported in Memgraph.

**To Reproduce**
A query like this
```
MATCH (n:Movie|Person)
RETURN n.name AS name, n.title AS title
```
(see https://neo4j.com/docs/cypher-manual/current/clauses/match/#label-expression-match-or-expression)

throws an error.

**Expected behavior**
This can be worked around by using multiple matches together with union, but this gets very verbose and hard to read quickly with multiple labels.

The same should be possible for [multiple relation types](https://neo4j.com/docs/cypher-manual/current/clauses/match/#match-on-rel-type)

**Logs**
Error: mismatched input '|' expecting {')', '{', '$'}" 

**Additional context**

**Verification Environment**
N/A

antejavor:
Hi @arminus, thanks for opening the issue. 

 The syntax is a bit different: 

```
CREATE (n:T1:T2{id:1});
MATCH (n:T1:T2) RETURN n;
MATCH (n:T1) RETURN n;
```

Let me know if this helps and we can close the issue.

arminus:
This seems to work like an AND operator, I "need" an OR operator, i.e. what I mean is this:

```
CREATE (n:T1{id:1});
CREATE (n:T2{id:2});
MATCH (n:T1:T2) RETURN n;
```
=> in this case, Neo4J returns 2 nodes for `MATCH (n:T1|T2) RETURN n;` memgraph doesn't return anything.

antejavor:
Hi @arminus, thanks for clarification, we are talking about this feature: https://neo4j.com/docs/cypher-manual/current/clauses/match/#label-expression-match-or-expression 

I have given this a big priority since it is a pretty basic feature but useful to a lot of folks.

antejavor:
+ 1 from community, we need this.

jroith:
These expressions are also important for us. It should be possible to rewrite the expressions to a where clause for a simple implementation.

icfly2:
Any progress on this? Not a deal breaker but as we are very seriously considering swapping our large neo4j to memgraph, not having to rewrite a lot of queries that are like:

```cypher
MATCH (s0:a {number: "IBAN123"})-[r0]-(b0:a {number:"IBAN345"})
match (s0)-[rs1]-(s1)
[...]

PR #2558, opened 2024-12-18
Add dynamic relationship creation in Cypher

Josipmrden:
The edge type can now be a symbolic name, parameter or property lookup. The
implication of this is that it's easier to create more edge types under a
single graph.

Use-cases:
  * import of many edge types, with this feature, you need only a single query
    to create many edge types
  * if you encode info as edge types, not as properties (if you only want a
    small number of distinct CREATE queries), your query performance can
    increase (in general) because the processing of edge types is faster
    compared to dealing with properties, but that might not always be the case
    AKA you need a deep understanding of your workload to get the performance
    benefits.


Josipmrden:
### Tracking


### Standard development
- [ ] Update unit/E2E tests


### CI Testing Labels


### Documentation checklist
    - **[ Release note text ]**

src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -2523,9 +2523,27 @@ antlrcpp::Any CypherMainVisitor::visitRelationshipLambda(MemgraphCypher::Relatio
...
   for (auto *edge_type : ctx->relTypeName()) {
-    types.push_back(AddEdgeType(std::any_cast<std::string>(edge_type->accept(this))));
+    if (edge_type->symbolicName()) {
+      types.emplace_back(AddEdgeType(std::any_cast<std::string>(edge_type->accept(this))));
+    } else if (edge_type->parameter()) {
+      // If we have a parameter, we have to resolve it.
+      const auto *param_lookup = std::any_cast<ParameterLookup *>(edge_type->accept(this));
+      const auto edge_type_name = parameters_->AtTokenPosition(param_lookup->token_position_).ValueString();
+      types.emplace_back(storage_->GetEdgeTypeIx(edge_type_name));
+      query_info_.is_cacheable = false;  // We can't cache queries with edge type parameters.
```
gitbuda, line 2535: Add an additional comment explaining why we can't cache queries with edge type parameters?

src/query/plan/operator.cpp
```diff
@@ -366,9 +375,11 @@ std::vector<Symbol> CreateExpand::ModifiedSymbols(const SymbolTable &table) cons
 }
 
 std::string CreateExpand::ToString() const {
-  return fmt::format("CreateExpand ({}){}[{}:{}]{}({})", input_symbol_.name(),
+  const auto *maybe_edge_type_id = std::get_if<storage::EdgeTypeId>(&edge_info_.edge_type);
+  const bool is_expansion_static = maybe_edge_type_id != nullptr;
+  return fmt::format("{} ({}){}[{}:{}]{}({})", "CreateExpand", input_symbol_.name(),
                      edge_info_.direction == query::EdgeAtom::Direction::IN ? "<-" : "-", edge_info_.symbol.name(),
-                     dba_->EdgeTypeToName(edge_info_.edge_type),
+                     is_expansion_static ? dba_->EdgeTypeToName(*maybe_edge_type_id) : "<DYNAMIC>",
```
gitbuda, line 382: Why not unpack the whole expression and output something like `<DYNAMIC variable>`?

src/query/plan/operator.cpp
```diff
@@ -201,6 +201,15 @@ std::vector<storage::LabelId> EvaluateLabels(const std::vector<StorageLabelType>
   return result;
 }
 
+storage::EdgeTypeId EvaluateEdgeType(const StorageEdgeType &edge_type, ExpressionEvaluator &evaluator,
```
gitbuda, line 204: Aha, one implication of this whole PR is users can more easily create a bunch of EdgeTypeIDs (I mean, it's not that hard even with the "static" EdgeType). Should we update documentation or put some kind of a limit on how many edge types is expected to be present in a given graph? 🤔

src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -2523,9 +2523,27 @@ antlrcpp::Any CypherMainVisitor::visitRelationshipLambda(MemgraphCypher::Relatio
...
+    } else {
+      auto variable = std::any_cast<std::string>(edge_type->variable()->accept(this));
+      users_identifiers.insert(variable);
+      auto *expression = static_cast<Expression *>(storage_->Create<Identifier>(variable));
+      for (auto *lookup : edge_type->propertyLookup()) {
+        auto key = std::any_cast<PropertyIx>(lookup->accept(this));
+        auto *property_lookup = storage_->Create<PropertyLookup>(expression, key);
+        expression = property_lookup;
+      }
+      types.emplace_back(expression);
```
gitbuda, line 2545: What about caching of this? Because if the above is not cacheable, this one should also be not cacheable? 🤔

src/query/plan/operator.cpp
gitbuda: Move this also to the PR description.

src/query/plan/operator.cpp
gitbuda: 2 options:
* if expression evaluation is not simple here, rewrite the string into `<DYNAMIC parameter|property_lookup>` (something more descriptive, but not huge)
* if expression evaluation is simple here -> add the exact name from the query

src/query/plan/rule_based_planner.hpp
```diff
@@ -321,11 +321,18 @@ class RuleBasedPlanner {
...
     transformed_edge_types.reserve(edge_types.size());
+
     for (const auto &type : edge_types) {
-      transformed_edge_types.push_back(GetEdgeType(type));
+      if (const auto *edge_type_atom = std::get_if<EdgeTypeIx>(&type)) {
+        transformed_edge_types.push_back(GetEdgeType(*edge_type_atom));
+      } else {
+        throw QueryException(
+            "Failed to work with dynamic edge types! Please contact Memgraph support as this scenario should not "
+            "happen!");
```
gitbuda, line 334: `"Failed to work with dynamic edge types! Avoid using parameters or variables when creating a new edge. Also, please contact Memgraph support, as this scenario should not happen (this is an unexpected code path, but we didn't want to just crash Memgraph, help us improve the query planner)!"`

@Ana-Marija-design this is an interesting error message to craft 😃

gitbuda (changes requested):

src/query/frontend/ast/cypher_main_visitor.cpp
Josipmrden: @DavIvek can you please guide me through this one, why we can't cache, I copied from node dynamic creation

DavIvek: Queries with edge types and node labels set as parameters shouldn't be cached because these parameters are resolved during the parsing stage. The same parameter could be resolved to different values if the user changes its value.

src/query/frontend/ast/cypher_main_visitor.cpp
DavIvek: These queries can be cached because expressions are not evaluated during the parsing stage. Queries with dynamic relationship types and labels can be cached, but their plans shouldn't be once support for dynamic relationship types and labels is added to the `MERGE` and `MATCH` clauses. In that case, the planner wouldn't be able to select the correct plan for these queries since expression evaluation happens in a later phase. However, since this feature is not currently supported in `MERGE` and `MATCH`, it's not an issue, and these queries can still be cached.

Related issue: https://github.com/memgraph/memgraph/issues/1848

src/query/plan/operator.cpp
Josipmrden: out of scope for this PR, approved by @colinbarry

Josipmrden: we don't have the `ExpressionEvaluator` at this point so unable to do more fine grained outputs.

src/query/plan/operator.cpp
Josipmrden: In the PR description. I wouldn't add any limit, let the users use it, I don't think they'll go over it (same with dynamic node creation that we already have).

src/query/frontend/ast/ast.hpp
```diff
@@ -2011,7 +2013,11 @@ class EdgeAtom : public memgraph::query::PatternAtom {
     object->direction_ = direction_;
     object->edge_types_.resize(edge_types_.size());
     for (auto i = 0; i < object->edge_types_.size(); ++i) {
-      object->edge_types_[i] = storage->GetEdgeTypeIx(edge_types_[i].name);
+      if (const auto *edge_type = std::get_if<EdgeTypeIx>(&edge_types_[i])) {
+        object->edge_types_[i] = storage->GetEdgeTypeIx(edge_type->name);
+      } else {
+        object->edge_types_[i] = std::get<Expression *>(edge_types_[i])->Clone(storage);
+      }
```
colinbarry, line 2020: Something to consider: using `std::visit` rather than `get_if`. That way, if we expand expand the variant's alternatives, we'll get compile-time errors.

```cpp
std::visit([&](auto &&arg) -> QueryEdgeType {
    using T = std::decay_t<decltype(arg)>;
    if constexpr (std::is_same_v<T, EdgeTypeIx>) {
      return storage->GetEdgeTypeIx(arg.name);
    } else if constexpr (std::is_same_v<T, Expression *>) {
      return arg->Clone(storage);
    }
  }, edge);
```

Ignition: or 
```
auto const do_thing = Overloaded{
  [&](EdgeTypeIx const &) { ... },
  [&](Expression const *) { ... },
};
std::visit(do_thing, edge);
```

gitbuda (approved):

PR #2561, opened 2024-12-18
Change query modules allocator

imilinovic:
- Change query modules to use multipool instead of monotonic allocator because monotonic would sometimes cause OOM on query modules that create and delete a lot of objects due to not being able to reclaim memory.
- Add reference type to C++ API and change MGP::List and MGP::Map get operations to return a reference.
- Removed noexcept from C++ API calls that didn't make sense.

BREAKING: Incorrect usage of mgp_value (C - API) now causes a crash - you must be very careful not to do a double deallocation. `mgp_value` has a destructor and therefore `mgp_value_destroy` must only be called on a pointer. Also after running `mgp_value_destroy` you MUST not call mgp_`type`_destroy on the underlying data because that will cause double deallocation and a crash.

### DIFF
--- a/include/mgp.hpp
+++ b/include/mgp.hpp
@@ -1266,7 +1266,7 @@ class Value {
 
   ~Value();
 
-  /// @brief Returns the pointer to the stored value.
+  /// @brief Returns the pointer to the stored value with the least significat bit set to 0.
   mgp_value *ptr() const;
 
   /// @brief Returns the type of the value.
@@ -1360,6 +1360,7 @@ class Value {
   std::string ToString() const;
 
  private:
+  /// least signifact bit signifies whether the value is a reference type
   mgp_value *ptr_;
 };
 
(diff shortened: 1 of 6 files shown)


include/mgp.hpp
```diff
@@ -3654,236 +3661,236 @@ inline Value &Value::operator=(Value &&other) noexcept {
...
 
 inline Value::~Value() {
-  if (ptr_ != nullptr) {
+  if (ptr_ != nullptr && (reinterpret_cast<uintptr_t>(ptr_) & 1) == 0) {
     mgp::value_destroy(ptr_);
   }
 }
 
-inline mgp_value *Value::ptr() const { return ptr_; }
+inline mgp_value *Value::ptr() const { return reinterpret_cast<mgp_value *>(reinterpret_cast<uintptr_t>(ptr_) & ~1); }
```
Ignition, line 3669: you need to make sure you compliment the right type `~uintptr_t{1}`

Ignition (approved):

PR #2568, opened 2024-12-21
Fix client-side routing in K8s HA deployment scenario

as51340:
Up to this point, the 1st leader couldn't find out information about its bolt server. In order to support that, users will run `ADD COORDINATOR 1..` additionally to `ADD COORDINATOR 2..` and `ADD COORDINATOR 3...`. Informations about coordinators in NuRaft cluster are saved at two places. Information about coordinator server and management server are saved inside `raft_server::aux` while information about bolt server inside application logs. The process of adding a coordinator (not self) then consists of two steps. In the 1st step we add coordinator to the networking part of NuRaft. In the 2nd step, we create an application log which stores information about the current state of the cluster. If the 1st step passed but the 2nd step didn't pass, we crash the database since this is an invalid situations. This situation should never occur at the startup but it serves us as a protection mechanism to remain consistent and safe. Data structures `CoordinatorInstanceAux` and `CoordinatorInstanceContext` are used to represent coordinator state in application logs and as part of NuRaft networking configuration. 

Files from nuraft/ and coordination/ are all moved to coordination/ since there is no more clear boundary between files in each folder. 

Flag `--instance-uuid-frequency-sec` is removed since it's not used anymore (it should have been deprecated past release). Flags `--instance-down-timeout-sec` and `--instance-health-check-frequency-sec` are still used but moved under ownership of coordinator instance. 

When run on empty cluster (cluster on which neither of `ADD COORDINATOR` queries were run), queries `SHOW INSTANCES` and `SHOW INSTANCE` will return the coordinator on which that query is run. Correct bolt server will be returned if `ADD COORDINATOR` is run on that coordinator and if not, empty string will be returned.

SSL is removed from all configurations since we don't support it yet. 

Coordinator IDs are represent using int32_t since this is used by NuRaft. Until now we were using uint32_t and doing unnecessary casts.


as51340:
### Tracking


### Standard development
- [ ] Update unit/E2E tests


### CI Testing Labels


### Documentation checklist
    - **[ Release note text ]**

src/communication/bolt/v1/states/handlers.hpp
```diff
@@ -488,6 +488,7 @@ State HandleGoodbye() {
 
 template <typename TSession>
 State HandleRoute(TSession &session, const Marker marker) {
+  spdlog::trace("Received ROUTE message");
```
as51340, line 491: test comment

src/coordination/coordinator_instance.cpp
```diff
@@ -710,43 +729,126 @@ auto CoordinatorInstance::UnregisterReplicationInstance(std::string_view instanc
...
-    return AddCoordinatorInstanceStatus::BOLT_ENDPOINT_ALREADY_EXISTS;
+  coordinator_instances_context.emplace_back(
+      CoordinatorInstanceContext{.id = config.coordinator_id, .bolt_server = bolt_server_to_add});
+
+  auto data_instances = raft_state_->GetDataInstancesContext();
+  auto uuid = raft_state_->GetCurrentMainUUID();
+
+  // If we managed to add it to the NuRaft configuration but not to our app logs.
+  if (!raft_state_->AppendClusterUpdate(std::move(data_instances), std::move(coordinator_instances_context), uuid)) {
+    LOG_FATAL("Couldn't append application log when adding coordinator {} to the cluster.", config.coordinator_id);
```
andrejtonev, line 849: better comment explaining to the user what they need to do

andrejtonev (approved):

as51340:
The change is breaking only on coordinators. Data on data instances (snapshots, wals...) can be used, this is breaking only for HA.

PR #2594, opened 2025-01-10
wip: Replace VERIFYING transaction status with an orthogonal flag

colinbarry:


colinbarry:

---
___


### Tracking


### Standard development
- [ ] Update unit/E2E tests


### CI Testing Labels


### Documentation checklist
    - **[ Release note text ]**

src/query/interpreter.cpp
```diff
@@ -6095,35 +6127,47 @@ void Interpreter::Abort() {
...
-      [this]() { transaction_status_.store(TransactionStatus::IDLE, std::memory_order_release); });
+  utils::OnScopeExit const clean_status([this]() {
+    for (TransactionStatus expected = TransactionStatus::STARTED_ROLLBACK;
+         !transaction_status_.compare_exchange_weak(expected, TransactionStatus::UPDATING_TRANSACTION_INFO);
+         expected = TransactionStatus::STARTED_ROLLBACK) {
+    }
+
+    current_transaction_.reset();
+    metadata_ = std::nullopt;
+    transaction_status_.store(TransactionStatus::IDLE, std::memory_order_release);
```
as51340, line 6163: this assumes that the state doesn't get changed from the moment you updated transaction info to the moment you are going into IDLE state which is basically a lock mechanism

as51340 (changes requested):
I believe you did all state transitions right. My concern is that this gets too complex. When I first did this feature, the engineer who reviewed said it may be better to just add locking mechanisms because there are a lot of "assumed transitions" which you need to take into account. At that point, we decided it's OK to use it as it is but I have a hard time following these transitions. 

Let's wait for what will @andrejtonev say but I would consider adding locks

src/query/interpreter.cpp
```diff
@@ -2327,7 +2327,6 @@ PreparedQuery Interpreter::PrepareTransactionQuery(std::string_view query_upper,
 
       expect_rollback_ = false;
       in_explicit_transaction_ = false;
-      metadata_ = std::nullopt;
```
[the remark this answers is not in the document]
as51340: I assume because you now have Updating and reading txn info

src/query/interpreter.cpp
```diff
@@ -2341,7 +2340,6 @@ PreparedQuery Interpreter::PrepareTransactionQuery(std::string_view query_upper,
       Abort();
       expect_rollback_ = false;
       in_explicit_transaction_ = false;
-      metadata_ = std::nullopt;
```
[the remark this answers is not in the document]
as51340: I assume because you now have Updating and reading txn info

colinbarry:
Fair point: there are a lot of quite obscure state transitions which could occur, and having explicit locking rather than atomic state changes and waits would simplify the code considerably.

src/query/interpreter.cpp
```diff
@@ -4340,8 +4359,7 @@ auto ShowTransactions(const std::unordered_set<Interpreter *> &interpreters, Que
       return false;
     };
 
-    if (transaction_id.has_value() && (same_user(interpreter->user_or_role_, user_or_role) ||
-                                       privilege_checker(user_or_role, get_interpreter_db_name()))) {
```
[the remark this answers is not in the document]
colinbarry: Accidental deletion. Will put these checks back in.

src/query/interpreter.hpp
```diff
@@ -366,7 +365,11 @@ class Interpreter final {
   void Abort();
 
   std::atomic<TransactionStatus> transaction_status_{TransactionStatus::IDLE};  // Tie to current_transaction_
-  std::optional<uint64_t> current_transaction_;
+  struct {
+    utils::SpinLock transaction_info_lock_;
```
andrejtonev, line 369: Does it make sense to switch to RW lock?

src/query/interpreter_context.cpp
```diff
@@ -49,21 +49,13 @@ std::vector<std::vector<TypedValue>> InterpreterContext::TerminateTransactions(
...
                          privilege_checker = std::move(privilege_checker)](const auto &interpreters) {
     for (Interpreter *interpreter : interpreters) {
-      TransactionStatus alive_status = TransactionStatus::ACTIVE;
-      // if it is just checking kill, commit and abort should wait for the end of the check
-      // The only way to start checking if the transaction will get killed is if the transaction_status is
-      // active
-      if (!interpreter->transaction_status_.compare_exchange_strong(alive_status, TransactionStatus::VERIFYING)) {
+      // Quick check to skip any transactions which are already flagged for
+      // termination, or which have began to commit or rollback.
+      if (interpreter->transaction_status_.load(std::memory_order_acquire) != TransactionStatus::RUNNING) {
```
andrejtonev, line 54: Since we are only checking that the transaction is RUNNING, seems like we can try to terminate the same transaction twice.
There is no re-check once the lock is acquired and we blindly set the state to TERMINATING.
To me it seems like we can call terminate twice, both pass the first check and wait on the lock. Then one sets it to TERMINATING, which triggers an ABORT. The second terminate then takes the lock and puts the state into TERMINATING (even tough it was in ABORT).

src/query/interpreter.cpp
```diff
@@ -6283,23 +6288,14 @@ void Interpreter::Commit() {
...
-  auto expected = TransactionStatus::ACTIVE;
-  while (!transaction_status_.compare_exchange_weak(expected, TransactionStatus::STARTED_COMMITTING)) {
-    if (expected == TransactionStatus::TERMINATED) {
-      throw memgraph::utils::BasicException(
-          "Aborting transaction commit because the transaction was requested to stop from other session. ");
-    }
-    expected = TransactionStatus::ACTIVE;
-    std::this_thread::sleep_for(std::chrono::milliseconds(1));
+  TransactionStatus const prior_status{transaction_status_.exchange(TransactionStatus::COMMITTING)};
+  if (prior_status == TransactionStatus::TERMINATING) {
```
andrejtonev, line 6294: This will leave the status in COMMITTING. Is this handled correctly on the aborting side?

andrejtonev (commented):
I see that Andi already went through the PR.
Didn't dive super deep because of that, just added a few comments.

colinbarry:
Closing: implemented by #3833

PR #2598, opened 2025-01-13
Bugfix: PeriodicCommit crashes if commit generates replication error

andrejtonev:
PeriodicCommit now handles commit errors.
On error the last commit is aborted, while everything committed previously is still permanent.
Ignoring the replication error, since it does not influence the data in any way (unlike other commit errors).

### DIFF
--- a/src/query/plan/operator.cpp
+++ b/src/query/plan/operator.cpp
@@ -47,6 +47,7 @@
 #include "storage/v2/id_types.hpp"
 #include "storage/v2/indices/point_iterator.hpp"
 #include "storage/v2/property_value.hpp"
+#include "storage/v2/storage_error.hpp"
 #include "storage/v2/view.hpp"
 #include "utils/algorithm.hpp"
 #include "utils/event_counter.hpp"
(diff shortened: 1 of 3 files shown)


as51340 (commented):
I think that handling replication errors specifically shows appropriately what we do in SYNC replication. I also didn't see the point in throwing an exception and I think this had some consequences on drivers.
GetTimestampInfo -> that's a good change, side effects are hard to reason about and unnecessary.

andrejtonev:
Gonna split this in two 
Fix periodic commit and merge in 3.0
Commit notifications for 3.1

src/query/plan/operator.cpp
```diff
@@ -6272,13 +6297,18 @@ class PeriodicCommitCursor : public Cursor {
...
+      commit_result = context.db_accessor->PeriodicCommit({}, context.db_acc);
       pulled_ = 0;
     } else if (!pull_value && pulled_ > 0) {
       // do periodic commit for the rest of pulled items
-      [[maybe_unused]] auto commit_result = context.db_accessor->PeriodicCommit({}, context.db_acc);
+      commit_result = context.db_accessor->PeriodicCommit({}, context.db_acc);
+    }
+
+    if (commit_result.HasError()) {
+      HandlePeriodicCommitError(commit_result.GetError());
```
as51340, line 6311: Isn't this logic with if, else-if a bit strange?

as51340 (approved):

PR #2629, opened 2025-01-21
Port dynamic algorithms and make them enterprise

imilinovic:
INCLUDED: pagerank, community detection, katz, node2vec
EXCLUDED: betweeness centrality


imilinovic:
### Tracking


### Standard development
- [ ] Update unit/E2E tests


### CI Testing Labels


### Documentation checklist
    - **[ Release note text ]**

release/docker/v6_deb.dockerfile
```diff
@@ -12,7 +12,7 @@ RUN apt-get update && apt-get install -y \
 
 # NOTE: The following are required to run built-in Python modules. For the full
 # list, please visit query_modules/CMakeLists.txt.
-RUN pip3 install --break-system-packages  networkx==3.4.2 numpy==2.2.1 scipy==1.15.1
+RUN pip3 install --break-system-packages  networkx==3.4.2 numpy==2.2.1 scipy==1.15.1 gensim==4.3.3
```
gitbuda, line 15: This is super questionable, node2vec_online depends on gensim:
* how much bigger is the base docker image
* what's the risk of the gensim package breaking our release process?

query_modules/mg_utility/mg_graph.hpp
```diff
@@ -0,0 +1,445 @@
...
+#include "data_structures/graph_view.hpp"
+#include "mg_exceptions.hpp"
+
+namespace mg_graph {
+
+enum class GraphType : std::uint8_t { kDirectedGraph, kUndirectedGraph };
+
+/// Graph representation.
+template <typename TSize = std::uint64_t>
+class Graph : public GraphView<TSize> {
```
DavIvek, line 34: We already have a similar implementation of these data structures in `include/` and `src/query/procedure/` (e.g., `include/mgp.hpp`), which are added in `query_modules/mg_utility/`. We should consider reusing them, and if any required functionalities are missing, we should expand the existing implementations instead. Adding redundant code like this will only increase maintenance efforts in the future.

DavIvek (changes requested):
As mentioned in the comment, I think we should aim to reuse existing functions and data structure implementations in `memgraph`, as we currently have two implementations serving the same purpose.

release/docker/v6_deb.dockerfile
DavIvek: Can we move the `networkx` and `gensim` packages into the init script since some query modules depend on them and they are now being packaged with `memgraph`.

DavIvek:
@gitbuda @imilinovic I noticed that we have different versions of python packages for various distributions and build options. Is there a reason why they are not consistent and should we change this?

release/docker/v6_deb.dockerfile
gitbuda: ~500MB increase. The size is ok for now -> we have to figure out a better way how to resolve deps and install/run time.

DavIvek (approved):

DavIvek:
> As mentioned in the comment, I think we should aim to reuse existing functions and data structure implementations in `memgraph`, as we currently have two implementations serving the same purpose.

As discussed, this will be addressed in the upcoming release.

PR #2636, opened 2025-01-27
Fix parallel recovery without snapshot

as51340:


as51340:
### Tracking


### Standard development
- [ ] Update unit/E2E tests


### CI Testing Labels


### Documentation checklist
    - **[ Release note text ]**

andrejtonev (commented):
We have an edge case that this doesn't fix.
Since the recovery info is generated from the snapshot, it will:
- be empty if no snapshot is present <- this is handled by this pr
- not be empty if a snapshot is present
If the snapshot is present the parallel recovery info will be defined by the contents of the snapshot. This means that if we have a snapshot and wal files, when recovering indices/constraints in parallel, we will only recover the part defined by the snapshot.

This is easy to verify. Run `./memgraph-storage-wal-enabled --storage-snapshot-interval-sec 10000 --storage-parallel-schema-recovery`
```
CREATE INDEX ON :L;
UNWIND RANGE(1,1000) AS i CREATE (:L{p:i});
CREATE SNAPSHOT
UNWIND RANGE(1001,2000) AS i CREATE (:L{p:i});
**restart**
MATCH(n) RETURN count(*); #returns 2000
MATCH(n:L) RETURN count(*); #returns 1000
```

as51340:
> We have an edge case that this doesn't fix. Since the recovery info is generated from the snapshot, it will:
> 
> * be empty if no snapshot is present <- this is handled by this pr
> * not be empty if a snapshot is present
>   If the snapshot is present the parallel recovery info will be defined by the contents of the snapshot. This means that if we have a snapshot and wal files, when recovering indices/constraints in parallel, we will only recover the part defined by the snapshot.
> 
> This is easy to verify. Run `./memgraph-storage-wal-enabled --storage-snapshot-interval-sec 10000 --storage-parallel-schema-recovery`
> 
> ```
> CREATE INDEX ON :L;
> UNWIND RANGE(1,1000) AS i CREATE (:L{p:i});
> CREATE SNAPSHOT
> UNWIND RANGE(1001,2000) AS i CREATE (:L{p:i});
> **restart**
> MATCH(n) RETURN count(*); #returns 2000
> MATCH(n:L) RETURN count(*); #returns 1000
> ```

To reproduce, also need to set --storage-snapshot-on-exit = false

as51340:
Only creating index without data with the following flags also crashes database:
```
                "--log-level=TRACE",
                "--storage-snapshot-interval-sec=10000",
                "--storage-snapshot-on-exit=false",
                "--storage-parallel-schema-recovery=true",
                "--storage-wal-enabled=true",
```

as51340:
Another issue @andrejtonev :
If recovery is done solely using snapshot, we don't measure the time it took to finish recovery. The metric is only sent if recovery is done using snapshot+wals

andrejtonev:
superseded by #2658

PR #2639, opened 2025-01-28
Avoid crash when replica recovery fails

as51340:
Improved usage of const semantics. Remove unnecessary includes. Simplify RoleMainData's constructor.
Removed unnecessary namespaces.
Use std::tie where appropriate.
Remove coordinator_instance unit test because all things from there are already tested on e2e level. The test wasn't running before.
Enable raft_state test to test bolt+routing.

GetRecoverySteps won't crash the database if they fail. Instead, the recovery task will simply leave the replica in state maybe_behind if some issue occured.


src/replication/state.cpp
```diff
@@ -111,7 +111,7 @@ bool ReplicationState::TryPersistRoleMain(std::string new_epoch, utils::UUID mai
   return false;
 }
 
-bool ReplicationState::TryPersistUnregisterReplica(std::string_view name) {
+bool ReplicationState::TryPersistUnregisterReplica(std::string_view name) const {
```
Ignition, line 114: Same thing `durability_->Delete(key)` is modifying a resource. Should not be considered `const`

as51340: done

Ignition (approved):

PR #2648, opened 2025-01-30
Add timeouts for force reset and system heartbeat RPC messages

as51340:
The AppendDeltas response message doesn't return current timestamp anymore since it wasn't used before. If a replica cannot read epoch id, the database won't crash, rather it will return negative response. System heartbeat RPC will now also use timeout set by default to 10''. Force reset RPC now also uses a pre-defined timeout of 10''. If resetting storage fails, the replica will be considered MAYBE_BEHIND and on the next ping, branching should be observed again and a new request for resetting storage should be created.

### DIFF
--- a/.github/workflows/stress_jepsen.yaml
+++ b/.github/workflows/stress_jepsen.yaml
@@ -78,7 +78,7 @@ jobs:
           cd tests/jepsen
           ./run.sh test \
           --binary ../../build/memgraph \
-          --run-args "--workload hacreate --nodes-config resources/cluster.edn --time-limit 36000 --concurrency 6" \
+          --run-args "--workload hacreate --nodes-config resources/cluster.edn --time-limit 32400 --concurrency 6" \
           --ignore-run-stdout-logs \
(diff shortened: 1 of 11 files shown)


as51340:
### Tracking


### Standard development
- [ ] Update unit/E2E tests


### CI Testing Labels


### Documentation checklist
    - **[ Release note text ]**

andrejtonev (commented):
The only reservation I have is that ForceResetStorage and SystemReocvery might take a very long time to finish.
These functions are not cancellable, so what happens what this times out because it just took to long to execute?

as51340:
> The only reservation I have is that ForceResetStorage and SystemReocvery might take a very long time to finish. These functions are not cancellable, so what happens what this times out because it just took to long to execute?

Discussed on the call:
If force reset fails, the replica will be maybe_behind, on the next ping we try again.
For system recovery, I will test it with a lot of tenants to see how much time it takes and decide on the proper value timeout. For now, ignoring it.

andrejtonev (approved):

PR #2658, opened 2025-01-31
Bugfix: Parallel recovery doesn't work with WAL files

andrejtonev:
Recalculating vertex batches if WALs are present <- fixes indices/constraints recovery
Using parallel recovery only is vertex batches > 1 <- fixes recovery without snapshots
Replication and RECOVER SNAPSHOT now recover indices/constraints in parallel as well
A bit simpler RecoverData logic

### DIFF
--- a/src/storage/v2/durability/durability.cpp
+++ b/src/storage/v2/durability/durability.cpp
@@ -366,22 +366,14 @@ void RecoverIndicesStatsAndConstraints(utils::SkipList<Vertex> *vertices, NameId
   }
 
   RecoverIndicesAndStats(indices_constraints.indices, indices, vertices, name_id_mapper, properties_on_edges,
-                         GetParallelExecInfoIndices(recovery_info, config), storage_dir);
+                         GetParallelExecInfo(recovery_info, config), storage_dir);
   RecoverConstraints(indices_constraints.constraints, constraints, vertices, name_id_mapper,
                      GetParallelExecInfo(recovery_info, config));
 }
 
 std::optional<ParallelizedSchemaCreationInfo> GetParallelExecInfo(const RecoveryInfo &recovery_info,
                                                                   const Config &config) {
-  return config.durability.allow_parallel_schema_creation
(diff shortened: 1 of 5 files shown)


src/storage/v2/durability/durability.cpp
```diff
@@ -597,6 +568,25 @@ std::optional<RecoveryInfo> Recovery::RecoverData(
     *wal_seq_num = *previous_seq_num + 1;
 
     spdlog::info("All necessary WAL files are loaded successfully.");
+
+    // Regenerate the vertex batches
+    // TODO edges?
```
as51340, line 573: I think we need to regenerate edges too since we are recovering them in parallel too

src/storage/v2/durability/durability.cpp
```diff
@@ -597,6 +568,25 @@ std::optional<RecoveryInfo> Recovery::RecoverData(
...
+      if (pos == batched) {
+        const auto left = size - pos;
+        if (left <= config.durability.items_per_batch) {
+          recovery_info.vertex_batches.emplace_back(v_itr->gid, left);
+          break;
+        }
+        recovery_info.vertex_batches.emplace_back(v_itr->gid, config.durability.items_per_batch);
+        batched += config.durability.items_per_batch;
+      }
+    }
```
as51340, line 589: Did you measure maybe? How slow is it to regenerate batches?

as51340 (changes requested):

src/storage/v2/durability/durability.cpp
andrejtonev: Edge indices are currently not recovered in parallel.
The snapshot has edge batches, but are only used internally.

src/storage/v2/durability/durability.cpp
andrejtonev: For a dataset that recovers in 30s, the batch recalculation takes 0.1s and index recovery 3s.
For me that is acceptable

as51340 (approved):

PR #2661, opened 2025-02-02
Rollout toolchain-v6 by Github Actions update

gitbuda:


gitbuda:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - **[ Release note text ]**

.github/workflows/reusable_package.yaml
```diff
@@ -5,23 +5,23 @@ on:
...
+        description: "Toolchain version. Default value is v6."
+        default: v6
       build_type:
         type: string
         description: "Memgraph Build type. Default value is Release."
         default: Release
       additional_build_args:
         type: string
-        description: "Additional build flags (--for-platform, --for-docker ...). Default value is empty."
+        description: "Additional build flags (--for-docker ...). Default value is empty."
```
gitbuda, line 24: Delete platform also from the `release/package/mgbuild.sh`

environment/toolchain/v6/build.sh
```diff
@@ -155,8 +155,8 @@ pushd build
 
 #modify CFLAGS for older CPU support
 if [[ "$for_arm" = false ]]; then
-    export CFLAGS="$CFLAGS -mtune=x64-64-v2 -mavx"
-    export CXXFLAGS="$CXXFLAGS -mtune=x64-64-v2 -mavx"
+    export CFLAGS="$CFLAGS -mtune=generic -mavx"
```
gitbuda, line 158: Does it make sense to parametrize this? 🤔 I would say yes 👀 

@DavIvek, what do you think will be the impact on vector search performance? `v2` vs `generic` what's the best way to measure?

mattkjames7: It looks like I made a typo there originally, it should have been `x86-64-v2`

mattkjames7: Even after fixing the type here, GCC cannot be compiled with the `-mtune=x86-64-v2` flag:
```
checking for x86_64-linux-gnu-gcc... x86_64-linux-gnu-gcc
checking whether the C compiler works... no
configure: error: in `/home/matt/memgraph/environment/toolchain/v6/build/gcc-14.2.0/build':
configure: error: C compiler cannot create executables
```
As far as I can work out `-mtune=generic` is the default option, so if we do end up going with that, then this flag will make 0 difference and can be removed.

DavIvek: Maybe adding proper benchmarks for the vector index and then comparing them against the current version would be a good way to verify this, but my guess is that it won’t have a significant impact.

mattkjames7 (approved):

libs/setup.sh
```diff
@@ -34,6 +34,7 @@ clone () {
   local dir_name=$2
   local checkout_id=$3
   local shallow=$4
+  git config --global http.version HTTP/1.1
```
gitbuda, line 37: Put a comment what is this doing and why the version has to be fixed to `1` 🤔

mattkjames7: removed :)

PR #2679, opened 2025-02-06
SHOW REPLICAS uses local timestamp

andrejtonev:
FrequentCheck sets state to BEHIND after 3 failed pings.
GetTimestampInfo uses the local ts, does not send an RPC.

### DIFF
--- a/src/replication/include/replication/replication_client.hpp
+++ b/src/replication/include/replication/replication_client.hpp
@@ -25,7 +25,7 @@ namespace memgraph::replication {
 struct ReplicationClient;
 
 template <typename F>
-concept FrequentCheckCB = std::invocable<F, bool, ReplicationClient &>;
+concept FrequentCheckCB = std::invocable<F, ReplicationClient &>;
 
 struct ReplicationClient {
   explicit ReplicationClient(const ReplicationClientConfig &config);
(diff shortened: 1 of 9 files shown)


andrejtonev:

---
___


### Tracking


### Standard development
- [ ] Update unit/E2E tests


### CI Testing Labels


### Documentation checklist
    - **[ Release note text ]**

src/storage/v2/replication/replication_client.hpp
```diff
@@ -108,6 +108,9 @@ class ReplicationStorageClient {
   auto Name() const -> std::string const & { return client_.name_; }
   auto Endpoint() const -> io::network::Endpoint const & { return client_.rpc_client_.Endpoint(); }
 
+  void SetBehind() {
```
as51340, line 111: SetMaybeBehind maybe better

src/replication/include/replication/replication_client.hpp
```diff
@@ -40,27 +40,32 @@ struct ReplicationClient {
...
+      replica_checker_.Run("Replica Checker",
+                           [this, cb = std::forward<FS>(callback), fail_cb = std::forward<FF>(fail_callback),
+                            failed_attempts = 0UL]() mutable {
+                             constexpr auto kFailureAfterN = 3UL;
+                             try {
+                               {
+                                 auto stream{rpc_client_.Stream<replication_coordination_glue::FrequentHeartbeatRpc>()};
+                                 stream.AwaitResponse();
+                               }
+                               cb(*this);
```
as51340, line 57: rename to succ_cb just to be consistent succ vs fail

src/storage/v2/replication/replication_client.cpp
```diff
@@ -448,20 +442,23 @@ void ReplicationStorageClient::RecoverReplica(uint64_t replica_commit, Storage *
...
       spdlog::debug("One of recovery steps failed, setting replica state to MAYBE_BEHIND");
+      last_known_ts_.store(replica_commit, std::memory_order_release);
       replica_state_.WithLock([](auto &val) { val = ReplicaState::MAYBE_BEHIND; });
       return;
     }
   }
 
   const auto last_durable_timestamp = storage->repl_storage_state_.last_durable_timestamp_.load();
   spdlog::info("Replica {} timestamp: {}, Last commit: {}", client_.name_, replica_commit, last_durable_timestamp);
+  last_known_ts_.store(replica_commit, std::memory_order_release);
```
as51340, line 461: I think we shouldn't update last_known_ts_ when in recovery. This could actually set the timestamp to a lower value than it was before the recovery task started because commits could be executed while recovering.

as51340 (changes requested):
My general comment is that maybe it would be best to rely just on Frequent checker to update timestamp. In that case no need to think about side effects e.g ForceReset. Also, shouldn't tests stay the same? Isn't it that we should get the same timestamp just after some time (1s = on the next ping).

src/storage/v2/replication/replication_client.cpp
andrejtonev: `replica_commit` should be updated as the recovery process goes on. §
In theory it should be the latest applied ts.
Will double check

andrejtonev:
@as51340 
Can't rely on FrequentCheck as it just pings the replica. I though about using Heartbeat in the check, but that is a db specific call. This means we would need to handle multi-tenancy inside the frequent check and that just seems cumbersome. Don't want to mix replication and multi-tenancy on this level if possible.

Needed to update the tests, as now we have additional info, the replica's timestamp. This was always 0 when replica was down. Now it is the last known ts.
Additionally we used to set the state to behind after GetTimestamp failed. We don't do that anymore.
Also the frequent check now sets the state to behind on a connection loss, before it would set it after a re-connection (meaning the connection was lost, but the state would be updated only after the connection was reestablished).

src/storage/v2/replication/replication_client.cpp
andrejtonev: The `replica_commit` is first read from the heartbeat. Then any successful recovery step updates it.
This means that it should be the best guess at what the replica's ts is.
In any case, it is not a problem even if the ts was wrong. This is used only in the show info query and is a best guess.
No functional logic relies on it.

as51340 (approved):
Discussed on the call, seems good

PR #2686, opened 2025-02-10
Fix data race in skip list usage

Ignition:
Issue was initially found with unique constraint. But root cause was with our `find_equal_or_greater`.


src/storage/v2/inmemory/storage.cpp
```diff
@@ -969,16 +973,6 @@ void InMemoryStorage::InMemoryAccessor::Abort() {
   // if we have no deltas then no need to do any undo work during Abort
   // note: this check also saves on unnecessary contention on `engine_lock_`
   if (!transaction_.deltas.empty()) {
-    // CONSTRAINTS
```
andrejtonev, line 972: Why are we not doing this in Abort anymore?

andrejtonev (commented):
Nice work!
Just wondering why the constraints abort has been moved to Commit.
What if I do a ROLLBACK, wouldn't the constraints be left in a bad state?

src/storage/v2/inmemory/storage.cpp
Ignition: Only unique constraint tracks anything. They are only inserted in Commit while under the lock. They are now removed before releasing the lock.

andrejtonev (approved):

PR #2691, opened 2025-02-11
Toolchain v7

mattkjames7:
New toolchain for building memgraph based on v6. 
- Added `openmp` to projects
- Updated stable version of packages
(more to be added here)


mattkjames7:
### TODOs

- [x] Use `CMAKE_TOOLCHAIN_FILE`
- [x] Add `ninja` and `conan` to toolchain or environment/os
- [x] caching conan builds
- [x] fix multiple versions of zlib
- [x] make sure RocksDB compiles correctly
- [x] install cargo within os scripts
- [x] split `init` script into `init` and `init-test` scripts to build a couple of minutes sooner
- [x] switch to `uv` for installing test dependencies 
- [x] merge #3074 
- [x] merge #2899 
- [x] update supported OS list (Fedora 42, Rocky 9/10, Debian 13?)
- [x] clean up old OS scripts in environment
- [x] figure out how to include `ccache` in conan build
- [x] script to build memgraph with a single command
- [x] share `.conan2` directory within mgbuild container
- [x] ~~share `.cache/pip` directory within mgbuild container~~
- [x] Toolchain builds:
  - [x] CentOS 9
  - [x] CentOS 10
  - [x] Debian 12
  - [x] Debian 12 arm
  - [x] Debian 13
  - [x] Debian 13 arm
  - [x] ~~Fedora 41~~
  - [x] Fedora 42
  - [x] Fedora 42 arm
  - [x] Rocky 10
  - [x] Ubuntu 22.04
  - [x] Ubuntu 24.04
  - [x] Ubuntu 24.04 arm
- [x] mgbuild images:
  - [x] CentOS 9
  - [x] CentOS 10
  - [x] Debian 12
  - [x] Debian 12 arm
  - [x] Debian 13
  - [x] Debian 13 arm
  - [x] ~~Fedora 41~~
  - [x] Fedora 42
  - [x] Fedora 42 arm
  - [x] Rocky 10
  - [x] Ubuntu 22.04
  - [x] Ubuntu 24.04
  - [x] Ubuntu 24.04 arm
- [x] use separate conan cache for each OS in CI
- [x] Don't build tests during packaging
- [ ] Re-enable IPO, perhaps add option to disable
- [x] Combine templates
- [ ] Decide whether to undo override of `layout` in the conanfile
- [x] Clean up `gflags`-related hacks in CmakeLists.txt
- [x] Check if `conandeps_legacy` actually needs to be in `tests/unit/CMakeLists.txt`

### Tracking


### Standard development
- [ ] Update unit/E2E tests


### CI Testing Labels


### Documentation checklist
    - New toolchain (v7) with OpenMP, updated `clang`, `cmake`, `boost` and more. Updated Linux distributions support: added Fedora 42 (x86_64 and aarch64), Debian 13 (x86_64 and aarch64) and Rocky 10 (x86_64 only); dropped support for Debian 11. [#2691](https://github.com/memgraph/memgraph/pull/2691)
- [x] **[ [Documentation PR link memgraph/documentation](https://github.com/memgraph/documentation/pull/1429) ]**

environment/toolchain/v7/clean.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash -ex
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+PREFIX=/opt/toolchain-v6
```
gitbuda, line 3: change to v7

mattkjames7: well spotted!

src/query/CMakeLists.txt
```diff
@@ -117,7 +117,7 @@ if(NOT "${MG_PYTHON_PATH}" STREQUAL "")
 endif()
 
 if("${MG_PYTHON_VERSION}" STREQUAL "")
-    find_package(Python3 3.5 REQUIRED COMPONENTS Development)
+    find_package(Python3 3.13 REQUIRED COMPONENTS Development)
```
gitbuda, line 120: I think this should stay 3.5 because that's minimum version

mattkjames7: This was definitely a mistake on my part!

mattkjames7:
Timing using `uv` to manage Python virtual environment in `init-test` vs just `pip`:

|         |   `pip`  |   `uv`   |
|:--------|:--------:|:--------:|
|cached   |  2m 30s  |  1m 16s  |
|uncached |  3m 37s  |  2m 32s  |

config/memgraph.conf.template
```diff
@@ -0,0 +1,389 @@
+# Memgraph Configuration
```
gitbuda: Matt: "We don't need this -> delete"

environment/toolchain/v7/rocksdb-v9.10.0.patch
```diff
@@ -0,0 +1,95 @@
+diff -Naur a/CMakeLists.txt b/CMakeLists.txt
```
gitbuda: We'll use 8.1.1 -> delete

environment/toolchain/v7/rocksdb-v10.5.1.patch
```diff
@@ -0,0 +1,98 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
```
gitbuda: We'll use 8.1.1 -> delete

repo-JyEI41bu/src/coordination/coordinator_state_machine.cpp
```diff
@@ -0,0 +1,392 @@
+// Copyright 2025 Memgraph Ltd.
```
gitbuda: remove that as well

repo-JyEI41bu/src/coordination/coordinator_state_manager.cpp
```diff
@@ -0,0 +1,235 @@
+// Copyright 2025 Memgraph Ltd.
```
gitbuda: remove that as well

repo-JyEI41bu/src/coordination/data_instance_management_server_handlers.cpp
```diff
@@ -0,0 +1,373 @@
+// Copyright 2025 Memgraph Ltd.
```
gitbuda: remove that as well

.github/workflows/build_rc.yml
```diff
@@ -100,7 +100,7 @@ jobs:
     uses: ./.github/workflows/reusable_release_tests.yaml
     with:
       os: ubuntu-24.04
-      toolchain: v6
+      toolchain: v7
```
Ignition: If we are updating `default: 'v7'` or hardcoding environment `TOOLCHAIN: v7` then do we still need to be explicit with toolchain we pass in?

CMakeLists.txt
```diff
@@ -60,12 +51,30 @@ message(STATUS "Set toolchain tools using MG_TOOLCHAIN_ROOT: ${MG_TOOLCHAIN_ROOT
...
+endif()
+
+if (NOT MG_TOOLCHAIN_ROOT)
+    # This is used to so that 3rd party libs can find the
+    # correct libraries from our toolchain
+    message(FATAL "Need toolchain")
+endif ()
+
+# Add toolchain to CMAKE_PREFIX_PATH so that find_package can locate libraries
+list(APPEND CMAKE_PREFIX_PATH "${MG_TOOLCHAIN_ROOT}")
```
Ignition: is this still needed?

CMakeLists.txt
```diff
@@ -60,12 +51,30 @@ message(STATUS "Set toolchain tools using MG_TOOLCHAIN_ROOT: ${MG_TOOLCHAIN_ROOT
...
+
 project(memgraph LANGUAGES C CXX)
 
 # this reduces issues in build folders from inside containers
 # which are different locations than the host
 set(CMAKE_BUILD_RPATH_USE_ORIGIN ON)
-
+set(CMAKE_BUILD_RPATH "${MG_TOOLCHAIN_ROOT}/lib64")
+set(CMAKE_FIND_ROOT_PATH "${CMAKE_CURRENT_BINARY_DIR}/generators")
```
Ignition, line 76: Is this still needed?

CMakeLists.txt
```diff
@@ -60,12 +51,30 @@ message(STATUS "Set toolchain tools using MG_TOOLCHAIN_ROOT: ${MG_TOOLCHAIN_ROOT
...
+# Add toolchain to CMAKE_PREFIX_PATH so that find_package can locate libraries
+list(APPEND CMAKE_PREFIX_PATH "${MG_TOOLCHAIN_ROOT}")
+
 project(memgraph LANGUAGES C CXX)
 
 # this reduces issues in build folders from inside containers
 # which are different locations than the host
 set(CMAKE_BUILD_RPATH_USE_ORIGIN ON)
-
+set(CMAKE_BUILD_RPATH "${MG_TOOLCHAIN_ROOT}/lib64")
```
Ignition, line 75: Why was this needed?

CMakeLists.txt
```diff
@@ -185,7 +194,16 @@ endif()
 
 # -----------------------------------------------------------------------------
 
+# This is used to so that 3rd party libs can find the
+# correct libraries from our toolchain
+# Check MG_TOOLCHAIN_ROOT so it's available for custom libstdc++ configuration
 # Custom libstdc++ configuration - REQUIRED
+if(NOT DEFINED MG_TOOLCHAIN_ROOT OR "${MG_TOOLCHAIN_ROOT}" STREQUAL "" OR NOT EXISTS "${MG_TOOLCHAIN_ROOT}" OR NOT IS_DIRECTORY "${MG_TOOLCHAIN_ROOT}")
```
Ignition: We can simplify since we know MG_TOOLCHAIN_ROOT is not falsy. `NOT EXISTS "${MG_TOOLCHAIN_ROOT}" OR NOT IS_DIRECTORY "${MG_TOOLCHAIN_ROOT}"`

CMakeLists.txt
```diff
@@ -218,14 +236,38 @@ set_target_properties(libstdc++_custom PROPERTIES
...
+find_package(ctre)
+find_package(absl)
+find_package(croncpp)
+find_package(range-v3)
+find_package(asio)
+find_package(ZLIB)
+
+# Now load mgclient which depends on OpenSSL
+find_package(OpenSSL REQUIRED)
+find_package(mgclient)
```
Ignition: Maybe we should add REQUIRED to all these

CMakeLists.txt
```diff
@@ -270,17 +312,21 @@ set(CMAKE_CXX_FLAGS_RELWITHDEBINFO
...
 # Use lld linker to speedup build and use less memory.
 add_link_options(-fuse-ld=lld)
-if (DEFINED ENV{LD_LIBRARY_PATH})
-    string(REPLACE ":" " " LD_LIBS $ENV{LD_LIBRARY_PATH})
-    separate_arguments(LD_LIBS)
-    link_directories(${LD_LIBS})
-endif ()
 add_link_options(LINKER:--build-id)
+link_directories("${MG_TOOLCHAIN_ROOT}/lib")
+link_directories("${MG_TOOLCHAIN_ROOT}/lib64")
```
Ignition: Is this still needed?

CMakeLists.txt
```diff
@@ -270,17 +312,21 @@ set(CMAKE_CXX_FLAGS_RELWITHDEBINFO
...
 # After linker is set, check we can use link-time optimization
 cmake_policy(SET CMP0138 NEW)
 include(CheckIPOSupported)
-check_ipo_supported() # fatal error if IPO is not supported
+message(STATUS "CMAKE_CXX_COMPILER_CLANG_SCAN_DEPS: ${CMAKE_CXX_COMPILER_CLANG_SCAN_DEPS}")
+check_ipo_supported(RESULT ipo_supported OUTPUT output) # fatal error if IPO is not supported
+if (NOT ipo_supported)
+    message(WARNING "IPO test failed")
+    message(WARNING "${output}")
+endif()
```
Ignition, line 327: Comment is wrong. Release performance, there will be places where IPO/LTO makes a difference.

build.sh
```diff
@@ -0,0 +1,133 @@
+#!/bin/bash
```
Ignition, line 1: harden the script with `set -euo pipefail`

build.sh
```diff
@@ -0,0 +1,133 @@
...
+    echo "Deleting existing build directory"
+    rm -rf build
+fi
+
+# run check for operating system dependencies
+./environment/os/install_deps.sh check TOOLCHAIN_RUN_DEPS
+./environment/os/install_deps.sh check MEMGRAPH_BUILD_DEPS
+
+echo "Creating virtual environment and installing conan"
+python3 -m venv env
```
Ignition: better to use idiomatic+overridable locations
 
```sh
VENV_DIR="${VENV_DIR:-.venv}"
python3 -m venv "$VENV_DIR"
source "$VENV_DIR/bin/activate"
```

build.sh
```diff
@@ -0,0 +1,133 @@
...
+if [[ -n "$MG_SANITIZERS" ]]; then
+    echo "Sanitizers enabled: $MG_SANITIZERS"
+    export MG_SANITIZERS="$MG_SANITIZERS"
+else
+    echo "No sanitizers enabled"
+fi
+
+# install conan dependencies
+export MG_TOOLCHAIN_ROOT=/opt/toolchain-v7
+conan install . --build=missing -pr $PROFILE_TEMPLATE -s build_type=$BUILD_TYPE
```
Ignition: Unquoted variables "$PROFILE_TEMPLATE"

build.sh
```diff
@@ -0,0 +1,133 @@
...
+fi
+
+# delete existing build directory
+if [ -d "build" ]; then
+    echo "Deleting existing build directory"
+    rm -rf build
+fi
+
+# run check for operating system dependencies
+./environment/os/install_deps.sh check TOOLCHAIN_RUN_DEPS
```
Ignition:
```sh
if ! ./environment/os/install_deps.sh check TOOLCHAIN_RUN_DEPS; then
    echo "Error: Dependency check failed for TOOLCHAIN_RUN_DEPS"
    exit 1
fi
```

build.sh
```diff
@@ -0,0 +1,133 @@
...
+    cmake --preset $PRESET
+fi
+
+if [[ "$config_only" = true ]]; then
+    exit 0
+fi
+
+# Build command with optional target
+if [[ -n "$TARGET" ]]; then
+    cmake --build build --target $TARGET -j $(nproc)
```
Ignition: We can use `cmake --build --preset $PRESET ...`

conanfile.py
```diff
@@ -0,0 +1,77 @@
...
+    description = "Conan file for Memgraph"
+    settings = "os", "compiler", "build_type", "arch"
+
+    exports_sources = "CMakeLists.txt", "src/*"
+
+    def layout(self):
+        # Custom layout to put everything in build/ instead of build/Release or build/RelWithDebInfo
+        self.folders.build = "build"
+        self.folders.generators = "build/generators"
+        self.folders.package = "build/package"
```
Ignition, line 26: Future we should fix this, once we make tests dependant on a build location, then we can use standard CMake layout (better for developer).

build.sh
```diff
@@ -0,0 +1,133 @@
...
+else
+    cmake --build build -j $(nproc)
+fi
+
+deactivate
```
Ignition: `trap 'deactivate 2>/dev/null' EXIT ERR`

environment/toolchain/v7/build.sh
```diff
@@ -0,0 +1,1536 @@
...
+    SCRIPT_SOURCE="\${BASH_SOURCE[0]}"
+elif [ -n "\$ZSH_VERSION" ]; then
+    current_shell="zsh"
+    SCRIPT_SOURCE="\${(%):-%N}"
+else
+    echo "Unsupported shell. Only bash and zsh are supported." >&2
+    return 1 2>/dev/null || exit 1
+fi
+
+SCRIPT_DIR=\$( cd -- "\$( dirname -- "\${SCRIPT_SOURCE}" )" &> /dev/null && pwd )
```
Ignition: readonly

environment/toolchain/v7/build.sh
```diff
@@ -0,0 +1,1536 @@
+#!/bin/bash -e
```
Ignition: Harden to `#!/bin/bash -euo pipefail`?

environment/toolchain/v7/build.sh
```diff
@@ -0,0 +1,1536 @@
...
+mkdir -p .gnupg
+chmod 700 .gnupg
+
+popd
+# create build directory
+mkdir -p build
+pushd build
+
+log_tool_name "GCC $GCC_VERSION"
+if [ ! -f $PREFIX/bin/gcc ]; then
```
Ignition: We should be quoting "$PREFIX/bin/gcc"

environment/toolchain/v7/build.sh
```diff
@@ -0,0 +1,1536 @@
...
+# verify libaio
+if [ ! -f libaio-CHECKSUMS ]; then
+    wget https://releases.pagure.org/libaio/CHECKSUMS -O libaio-CHECKSUMS
+fi
+cat libaio-CHECKSUMS | grep "SHA256 (libaio-$LIBAIO_VERSION.tar.gz)" | sha256sum -c
+# verify libevent
+if [ ! -f libevent-$LIBEVENT_VERSION.tar.gz.asc ]; then
+    wget https://github.com/libevent/libevent/releases/download/release-$LIBEVENT_VERSION/libevent-$LIBEVENT_VERSION.tar.gz.asc
+fi
+if false; then
```
Ignition: Why is this disabled? `if false`

environment/toolchain/v7/build.sh
```diff
@@ -0,0 +1,1536 @@
...
+    $GPG --verify zstd-$ZSTD_VERSION.tar.gz.sig zstd-$ZSTD_VERSION.tar.gz
+fi
+# verify wangle
+echo "$WANGLE_SHA256 wangle-$FBLIBS_VERSION.tar.gz" | sha256sum -c
+popd
+
+pushd build
+source $PREFIX/activate
+export PIP_BREAK_SYSTEM_PACKAGES=1
+pip install conan==2.17.1
```
Ignition: Don't make this global, make it local to the exact pip command you intend.
```sh
PIP_BREAK_SYSTEM_PACKAGES=1 pip install conan==2.17.1
```
Also this should why is conan here? We should have it in our own separate build venv right?

environment/toolchain/v7/build.sh
```diff
@@ -0,0 +1,1536 @@
...
+    fi
+    make -j$CPUS
+    # make -k check # run test suite
+    make install
+    popd && popd
+fi
+
+log_tool_name "GDB $GDB_VERSION"
+if [[ ! -f "$PREFIX/bin/gdb" && "$DISTRO" != "amzn-2" ]]; then
+    if [ -d gdb-$GDB_VERSION ]; then
```
Ignition: Inconsistent, use bash `[[` + `]]`

environment/toolchain/v7/build.sh
```diff
@@ -0,0 +1,1536 @@
...
+        fi
+    done
+fi
+
+# create archives directory
+mkdir -p archives && pushd archives
+# download all archives
+if [ ! -f gcc-$GCC_VERSION.tar.gz ]; then
+    # wget https://ftp.gnu.org/gnu/gcc/gcc-$GCC_VERSION/gcc-$GCC_VERSION.tar.gz
+    wget https://mirrorservice.org/sites/sourceware.org/pub/gcc/releases/gcc-$GCC_VERSION/gcc-$GCC_VERSION.tar.gz
```
Ignition, line 123: We should be checksum validating

environment/toolchain/v7/clean.sh
```diff
@@ -0,0 +1,58 @@
...
+# rm -rf "$PREFIX/include/python3.13/Python.h"
+# rm -rf "$PREFIX/include/sodium.h"
+# rm -rf "$PREFIX/include/libunwind.h"
+# rm -rf "$PREFIX/include/lz4.h"
+# rm -rf "$PREFIX/include/lzma.h"
+# rm -rf "$PREFIX/include/zlib.h"
+# rm -rf "$PREFIX/include/zstd.h"
+# rm -rf "$PREFIX/include/snappy.h"
+
+# rm -rf "$PREFIX"
```
Ignition, line 58: We clean nothing?

environment/toolchain/v7/toolchain.cmake
```diff
@@ -0,0 +1,39 @@
+set(CMAKE_SYSTEM_NAME Linux)
+execute_process(
+     COMMAND uname -m
+     OUTPUT_VARIABLE uname_result
+     OUTPUT_STRIP_TRAILING_WHITESPACE
+)
+set(CMAKE_SYSTEM_PROCESSOR "${uname_result}")
```
Ignition, line 9: Work when we are using native compilers. This will break if we move to use cross compilers.

init
```diff
@@ -47,7 +47,7 @@ else
     done
 fi
 
-if [[ "$check_system_packages" == "true" ]]; then
+if [[ $check_system_packages ]]; then
```
Ignition: Will always pass?

libs/CMakeLists.txt
```diff
@@ -84,21 +84,29 @@ macro(add_external_project_impl name)
 
   # Add MG_TOOLCHAIN_ROOT to KW_CMAKE_PREFIX_PATH
   if(KW_CMAKE_PREFIX_PATH)
-    set(KW_CMAKE_PREFIX_PATH "${MG_TOOLCHAIN_ROOT}${LIST_SEP}${KW_CMAKE_PREFIX_PATH}")
+    set(KW_CMAKE_PREFIX_PATH "${MG_TOOLCHAIN_ROOT};${KW_CMAKE_PREFIX_PATH}")
```
Ignition: https://cmake.org/cmake/help/latest/module/ExternalProject.html#miscellaneous-options
This is to be consistent for LIST_SEPARATOR in ExternalProject_Add we use `set(LIST_SEP "-*-")` so that we ensure it gets pass CMAKE_ARGS without an issue.

libs/setup.sh
```diff
@@ -293,76 +240,36 @@ else
...
 popd
 
-# croncpp
-croncpp_ref="v2023.03.30"
-repo_clone_try_double "${primary_urls[croncpp]}" "${secondary_urls[croncpp]}" "croncpp" "$croncpp_ref"
+
+if [[ "$deactivate_toolchain" == "true" ]]; then
+  echo "Deactivating toolchain"
+  deactivate
+fi
```
Ignition: prefer trap

CMakeLists.txt
mattkjames7: Without this, it doesn't find `libstdc++.so.6`

CMakeLists.txt
mattkjames7: the generators path doesn't seem to be needed any more

environment/toolchain/v7/clean.sh
mattkjames7: this script is just used for cleaning and rebuilding specific parts of the toolchain - if we comment out one of the `rm` commands, it will remove the file that is checked for in the `build.sh` script

.github/workflows/build_rc.yml
mattkjames7: It might come in handy to use this in future. The reason that older toolchain support is being removed is because the build process is now quite different, as it relies on `conan`. Building in future will probably still work with toolchain v7 when we move to v8, v9 etc. and migrate more parts of the build to `conan`. That being said - it would be a lot simpler if we only support a single toolchain at any one time.

CMakeLists.txt
mattkjames7: nope, it doesn't seem to be!

init
mattkjames7: fixed

libs/CMakeLists.txt
mattkjames7: not sure why this changed, reverted

CMakeLists.txt
mattkjames7: done

CMakeLists.txt
mattkjames7: Something weird is going on here that I still need to figure out. I placed the warning in as a temporary measure - it will be reverted back to a fatal error once the issue has been sorted.

One thing I am not sure on yet, is whether the build should fail or not if this check fails. We set IPO to be used in the subsequent lines and therefore link with `-flto` successfully. My assumption is that the build will actually fail at linking time if the toolchain truly doesn't support IPO, but that is just an assumption.

CMakeLists.txt
mattkjames7: It looks like line 316 _is_ needed, but 317 isn't

build.sh
mattkjames7: good shout, done

build.sh
mattkjames7: good idea, done

build.sh
mattkjames7: also a good idea, done

build.sh
mattkjames7: Is this in place of running the configure then the build? i.e.?
```bash
cmake --build build --preset $PRESET -j $(nproc)
```
instead of 
```bash
cmake --preset $PRESET
cmake --build build -j $(nproc)
```
I assume that this would be redundant?
```bash
cmake --preset $PRESET
cmake --build build --preset $PRESET -j $(nproc)
```

conanfile.py
mattkjames7: I have added a TODO for me here. I plan to overhaul the testing in a lot of places soon, so fixing this, when I do that should come quite naturally.

build.sh
mattkjames7: nice, done

CMakeLists.txt
mattkjames7: done

environment/toolchain/v7/build.sh
mattkjames7: done

environment/toolchain/v7/build.sh
mattkjames7: done

environment/toolchain/v7/build.sh
mattkjames7: done (and a bunch of others too)

environment/toolchain/v7/build.sh
mattkjames7: good question, this seems to have been the case since toolchain v5

environment/toolchain/v7/build.sh
mattkjames7: I have removed it, we don't need this any more, since we use the virtual env

environment/toolchain/v7/build.sh
mattkjames7: done

environment/toolchain/v7/build.sh
mattkjames7: I have no idea why these were not being verified, but I have just done so and fixed this code.

environment/toolchain/v7/build.sh
mattkjames7: done

libs/setup.sh
mattkjames7: done

build.sh
mattkjames7: fixed

.github/workflows/build_rc.yml
mattkjames7: I have actually removed these anyway now :grin:

environment/toolchain/v7/toolchain.cmake
mattkjames7: If we move to cross compiling, would need to do more work on the rest of the toolchain, I think, so I'll add a note here.

libs/CMakeLists.txt
```diff
@@ -84,7 +84,7 @@ macro(add_external_project_impl name)
 
   # Add MG_TOOLCHAIN_ROOT to KW_CMAKE_PREFIX_PATH
   if(KW_CMAKE_PREFIX_PATH)
-    set(KW_CMAKE_PREFIX_PATH "${MG_TOOLCHAIN_ROOT};${KW_CMAKE_PREFIX_PATH}")
+    set(KW_CMAKE_PREFIX_PATH "${MG_TOOLCHAIN_ROOT}${LIST_SEP}${KW_CMAKE_PREFIX_PATH}")
```
Ignition, line 87: `ExternalProject_Add` will need `LIST_SEPARATOR "${LIST_SEP}"` adding back in

CMakeLists.txt
Ignition: Document that with a comment...not a blocker, but something to explore at some point (likely by me)

CMakeLists.txt
Ignition: Curious, again document with a comment

CMakeLists.txt
Ignition: there is a toolchain setup issue. check_ipo_supported is to check that the chosen complier + linker work for IPO/LTO. If it fails here it is surprising that it would work continue to build and link.

One thing I don't like is this:
```
# This is used to so that 3rd party libs can find the
# correct libraries from our toolchain
set(MG_TOOLCHAIN_ROOT "$ENV{MG_TOOLCHAIN_ROOT}")

if(NOT DEFINED ENV{MG_TOOLCHAIN_ROOT} OR "$ENV{MG_TOOLCHAIN_ROOT}" STREQUAL "")
    message(FATAL_ERROR "MG_TOOLCHAIN_ROOT environment variable is required but not set. Please set it to the toolchain directory.")
endif()

message(STATUS "MG_TOOLCHAIN_ROOT: ${MG_TOOLCHAIN_ROOT}")

# This needs to appear before project() in order for the IPO check to pass
set(CMAKE_CXX_COMPILER_AR "${MG_TOOLCHAIN_ROOT}/bin/llvm-ar" CACHE FILEPATH "C++ compiler archiver" FORCE)
set(CMAKE_CXX_COMPILER_RANLIB "${MG_TOOLCHAIN_ROOT}/bin/llvm-ranlib" CACHE FILEPATH "C++ compiler ranlib" FORCE)
set(CMAKE_C_COMPILER_AR "${MG_TOOLCHAIN_ROOT}/bin/llvm-ar" CACHE FILEPATH "C compiler archiver" FORCE)
set(CMAKE_C_COMPILER_RANLIB "${MG_TOOLCHAIN_ROOT}/bin/llvm-ranlib" CACHE FILEPATH "C compiler ranlib" FORCE)
set(CMAKE_AR "${MG_TOOLCHAIN_ROOT}/bin/llvm-ar" CACHE FILEPATH "Archiver" FORCE)
set(CMAKE_RANLIB "${MG_TOOLCHAIN_ROOT}/bin/llvm-ranlib" CACHE FILEPATH "Ranlib" FORCE)
message(STATUS "Set toolchain tools using MG_TOOLCHAIN_ROOT: ${MG_TOOLCHAIN_ROOT}")

# -----------------------------------------------------------------------------

# Capture MG_TOOLCHAIN_ROOT from environment variable if not already set
if (NOT MG_TOOLCHAIN_ROOT)
    if (DEFINED ENV{MG_TOOLCHAIN_ROOT})
        set(MG_TOOLCHAIN_ROOT $ENV{MG_TOOLCHAIN_ROOT})
    endif()
endif()

if (NOT MG_TOOLCHAIN_ROOT)
    # This is used to so that 3rd party libs can find the
    # correct libraries from our toolchain
    message(FATAL "Need toolchain")
endif ()
```

We still have unconditional read from an environment variable `set(MG_TOOLCHAIN_ROOT "$ENV{MG_TOOLCHAIN_ROOT}")` and we also set up a bunch of tools that should have been set in the toolchain file.

Ignition: toolchain should have something like:
```
# Specify the cross compiler
set(CMAKE_C_COMPILER /path/to/gcc)
set(CMAKE_CXX_COMPILER /path/to/g++)

# Archiver
set(CMAKE_AR /path/to/ar)

# Linker
set(CMAKE_LINKER /path/to/ld)  #### <-- this should be lld

# Ranlib (for static libraries)
set(CMAKE_RANLIB /path/to/ranlib)

# NM (symbol listing)
set(CMAKE_NM /path/to/nm)

# Objcopy
set(CMAKE_OBJCOPY /path/to/objcopy)

# Objdump
set(CMAKE_OBJDUMP /path/to/objdump)

# Strip
set(CMAKE_STRIP /path/to/strip)

# Optional: Set compiler flags
set(CMAKE_C_FLAGS_INIT "-flag1 -flag2")
set(CMAKE_CXX_FLAGS_INIT "-flag1 -flag2")

# Optional: Set linker flags
set(CMAKE_EXE_LINKER_FLAGS_INIT "-Wl,--some-flag")
set(CMAKE_SHARED_LINKER_FLAGS_INIT "-Wl,--some-flag")
```

build.sh
Ignition: Still two cmake calls: 
`cmake --preset $PRESET` does the configure stage
`cmake --build build --preset $PRESET ...` does the build stage (but ensures same preset)

PR #2701, opened 2025-02-17
Add timeouts to SnapshotRpc

as51340:
Added timeout to SnaphotRpc message. The in-progress response message will be sent for every 1'000'000 units of work performed. Processing vertices, edges, vertices inside label index, vertices inside property index, edges inside edge type index, edges inside edge type property index is worth one unit of work. Processing vertices inside point and text indices takes a bit more time so it's worth 10 units of work. The processing of vertices inside vector indices takes longest so it's worth 1000 units of work. 

Delivered also a multi threaded circular counter using lock-free techniques. Progress observer is implemented using push observer design pattern. 

Replica should approximately send updates every 2-3s. Hence, a timeout of 60s should be more than enough.

Also delivered the bugfix for vector indices which didn't check whether vertices contain labels.

### DIFF
--- a/src/dbms/inmemory/replication_handlers.cpp
+++ b/src/dbms/inmemory/replication_handlers.cpp
@@ -317,7 +317,7 @@ void InMemoryReplicationHandlers::SnapshotHandler(DbmsHandler *dbms_handler,
   }
   spdlog::info("Received snapshot saved to {}", *maybe_snapshot_path);
 
-  auto snapshot_observer = std::make_shared<SnapshotObserver>(res_builder);
+  auto snapshot_progress_observer = std::make_shared<SnapshotObserver>(res_builder);
 
   {
     auto storage_guard = std::lock_guard{storage->main_lock_};
@@ -333,7 +333,7 @@ void InMemoryReplicationHandlers::SnapshotHandler(DbmsHandler *dbms_handler,
           &storage->repl_storage_state_.history, storage->name_id_mapper_.get(), &storage->edge_count_,
           storage->config_, &storage->enum_store_,
           storage->config_.salient.items.enable_schema_info ? &storage->schema_info_.Get() : nullptr,
-          snapshot_observer);
+          snapshot_progress_observer);
       // If this step is present it should always be the first step of
       // the recovery so we use the UUID we read from snapshot
       storage->uuid().set(snapshot_info.uuid);
@@ -347,7 +347,7 @@ void InMemoryReplicationHandlers::SnapshotHandler(DbmsHandler *dbms_handler,
       storage::durability::RecoverIndicesStatsAndConstraints(
           &storage->vertices_, storage->name_id_mapper_.get(), &storage->indices_, &storage->constraints_,
           storage->config_, recovery_info, indices_constraints, storage->config_.salient.items.properties_on_edges,
-          snapshot_observer);
+          snapshot_progress_observer);
     } catch (const storage::durability::RecoveryFailure &e) {
       spdlog::error("Couldn't load the snapshot from {} because of: {}. Storage will be cleared.", *maybe_snapshot_path,
                     e.what());
--- a/src/dbms/inmemory/replication_handlers.cpp
+++ b/src/dbms/inmemory/replication_handlers.cpp
@@ -53,7 +53,7 @@ class SnapshotObserver final : public utils::Observer<void> {
 
  private:
   slk::Builder *res_builder_;
-  // Mutex is needed because multiple RPC calls could be executed in parallel
+  // Mutex is needed because RPC execution could be concurrent
   mutable std::mutex mtx_;
 };
 
--- a/src/dbms/inmemory/replication_handlers.cpp
+++ b/src/dbms/inmemory/replication_handlers.cpp
@@ -43,13 +43,18 @@ using memgraph::storage::durability::WalDeltaData;
 
 namespace memgraph::dbms {
 
-class SnapshotObserver final : utils::Observer<void> {
+class SnapshotObserver final : public utils::Observer<void> {
  public:
   explicit SnapshotObserver(slk::Builder *res_builder) : res_builder_(res_builder) {}
-  void Update() override { rpc::SendInProgressMsg(res_builder_); }
+  void Update() override {
+    auto guard = std::lock_guard{mtx_};
+    rpc::SendInProgressMsg(res_builder_);
+  }
 
  private:
   slk::Builder *res_builder_;
+  // Mutex is needed because multiple RPC calls could be executed in parallel
+  mutable std::mutex mtx_;
 };
 
 namespace {
@@ -312,7 +317,7 @@ void InMemoryReplicationHandlers::SnapshotHandler(DbmsHandler *dbms_handler,
   }
   spdlog::info("Received snapshot saved to {}", *maybe_snapshot_path);
 
-  auto snapshot_observer = std::make_unique<SnapshotObserver>(res_builder);
+  auto snapshot_observer = std::make_shared<SnapshotObserver>(res_builder);
 
   {
     auto storage_guard = std::lock_guard{storage->main_lock_};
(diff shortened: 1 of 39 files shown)


as51340:
### Tracking


### Standard development
- [ ] Update unit/E2E tests


### CI Testing Labels


### Documentation checklist
    - **[ Release note text ]**

src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -41,6 +42,21 @@ using memgraph::storage::View;
...
+  explicit SnapshotObserver(slk::Builder *res_builder) : res_builder_(res_builder) {}
+  void Update() override {
+    auto guard = std::lock_guard{mtx_};
+    rpc::SendInProgressMsg(res_builder_);
+  }
+
+ private:
+  slk::Builder *res_builder_;
+  // Mutex is needed because multiple RPC calls could be executed in parallel
+  mutable std::mutex mtx_;
```
andrejtonev, line 57: Don't really follow this.
Wouldn't you create an observer for a single RPC execcution?

andrejtonev: Or did you mean that the RPC execution can be multi-threaded?

src/storage/v2/durability/durability.cpp
```diff
@@ -159,23 +159,29 @@ std::optional<std::vector<WalDurabilityInfo>> GetWalFiles(const std::filesystem:
...
+                           snapshot_observer);
+  RecoverTypeConstraints(constraints_metadata, constraints, vertices, parallel_exec_info, snapshot_observer);
 }
 
 void RecoverIndicesAndStats(const RecoveredIndicesAndConstraints::IndicesMetadata &indices_metadata, Indices *indices,
                             utils::SkipList<Vertex> *vertices, NameIdMapper *name_id_mapper, bool properties_on_edges,
                             const std::optional<ParallelizedSchemaCreationInfo> &parallel_exec_info,
-                            const std::optional<std::filesystem::path> &storage_dir) {
+                            const std::optional<std::filesystem::path> &storage_dir,
+                            std::shared_ptr<utils::Observer<void>> const snapshot_observer) {
```
andrejtonev, line 175: Maybe rename in general this from snapshot observer to progress observer

src/storage/v2/durability/durability.cpp
```diff
@@ -159,23 +159,29 @@ std::optional<std::vector<WalDurabilityInfo>> GetWalFiles(const std::filesystem:
...
   spdlog::info("Recreating indices from metadata.");
 
   // Recover label indices.
   spdlog::info("Recreating {} label indices from metadata.", indices_metadata.label.size());
   auto *mem_label_index = static_cast<InMemoryLabelIndex *>(indices->label_index_.get());
   for (const auto &item : indices_metadata.label) {
-    if (!mem_label_index->CreateIndex(item, vertices->access(), parallel_exec_info)) {
+    if (!mem_label_index->CreateIndex(
+            item, vertices->access(), parallel_exec_info,
+            SnapshotObserverInfo{.observer = snapshot_observer, .vertices_snapshot_progress_size = 1'000'000})) {
```
andrejtonev, line 184: There is a constant for the vertices size

src/storage/v2/durability/snapshot.cpp
```diff
@@ -312,6 +320,9 @@ void LoadPartialEdges(const std::filesystem::path &path, utils::SkipList<Edge> &
               "configured without properties on edges!");
       }
     }
+    if (snapshot_observer != nullptr && batch_counter()) {
```
andrejtonev, line 323: Can't we use i for this? Why do we need the batch counter

src/storage/v2/durability/snapshot.cpp
```diff
@@ -563,6 +579,9 @@ LoadPartialConnectivityResult LoadPartialConnectivity(const std::filesystem::pat
           }
         }
         vertex.in_edges.emplace_back(get_edge_type_from_id(*edge_type), &*from_vertex, edge_ref);
+        if (snapshot_observer != nullptr && batch_counter()) {
```
andrejtonev, line 582: Isn't this called for already batched execution? Can't we just update the progress at the end?

src/storage/v2/indices/indices_utils.hpp
```diff
@@ -338,14 +342,28 @@ inline void TryInsertLabelPropertyIndex(Vertex &vertex, std::pair<LabelId, Prope
   index_accessor.insert({std::move(value), &vertex, 0});
 }
 
+struct SnapshotObserverInfo {
+  std::shared_ptr<utils::Observer<void>> observer{nullptr};
+  uint32_t vertices_snapshot_progress_size{1'000'000};
+};
```
andrejtonev, line 348: If you want to use this structure, perhaps you can add more functionality to it.
Can we just have a function that does this
```
if (maybe_batch_counter && (*maybe_batch_counter)()) {
        snapshot_info->observer->Update();
}
```

andrejtonev (commented):
Looks good. Would be better if the observer could be cleaner. Currently you have to create the observer, batch counter, check everything and update.

src/dbms/inmemory/replication_handlers.cpp
as51340: recovery can be executed in parallel. Each recovery thread can call Update in parallel. But now am I thinking that this is already protected by RPC lock

src/storage/v2/durability/snapshot.cpp
as51340: it is but the size of the batch is controller by the configuration parameter which users can change. I want to rely all the time on the same value to have predictable timings.

src/storage/v2/durability/snapshot.cpp
as51340: Size of the batch is controlled by the configuration parameter which users can change. I want to rely all the time on the same value to have predictable timings.

src/storage/v2/durability/durability.cpp
as51340: renamed to snapshot progress observer

src/storage/v2/indices/indices_utils.hpp
as51340: I will try to improve it. Not sure if it will be better if counter gets glued with snapshot observer.

src/dbms/inmemory/replication_handlers.cpp
as51340: I think we need mutex because slk::Builder isn't thread safe

src/storage/v2/indices/indices_utils.hpp
as51340: changed this part, should be better now

src/storage/v2/durability/durability.cpp
```diff
@@ -37,11 +37,20 @@
...
 #include "utils/message.hpp"
 #include "utils/timer.hpp"
+
+namespace {
+constexpr uint32_t kEdgesSnapshotProgressSize = 1'000'000;
+constexpr uint32_t kVerticesSnapshotProgressSize = 1'000'000;
+constexpr uint32_t kVerticesTextIdxSnapshotProgressSize = 100'000;
+constexpr uint32_t kVerticesPointIdxSnapshotProgressSize = 100'000;
+constexpr uint32_t kVerticesVectorIdxSnapshotProgressSize = 1000;
+
```
andrejtonev, line 51: Could we simplify this and have a single progress size?
I realize not all progress calls are happening in the same way, but this seems overly complex.
Ideally the progress size would be defined in the observer and everyone would just call a +1 and at the point we are >= size, an RPC would be sent. 
Could we simplify this and add the observer in a different way?

as51340: Single progress size is not possible since processing some types of indices takes much longer than for others. I can change so that configuration parameters for progress sizes end inside the Observer but I don't think that's better. Why do you think this is too complex?

as51340: done. Used Push Observer pattern

andrejtonev (approved):

PR #2702, opened 2025-02-17
Robust handling around memory limits

Ignition:
- NotEnoughMemoryException and AllocationException did not need allocation.
- mgp_func_result_set_error_msg now allows temporary allocation above the memory limits.
- Use query allocator for LOAD CSV. The old reason for using different allocator because of lifetime issues no longer valid. Using query allocator will allow memory to be moved rather than copied, so should be more efficient.
- patched `nlohmann/json` to handle exception during `~basic_json`

breaking: `include/mg_exceptions.hpp` had ABI break which requires C++ modules to be rebuilt.

### DIFF
--- a/src/query/plan/operator.cpp
+++ b/src/query/plan/operator.cpp
@@ -5720,13 +5720,13 @@ std::vector<Symbol> LoadCsv::ModifiedSymbols(const SymbolTable &sym_table) const
 namespace {
 // copy-pasted from interpreter.cpp
 TypedValue EvaluateOptionalExpression(Expression *expression, ExpressionEvaluator *eval) {
-  return expression ? expression->Accept(*eval) : TypedValue();
+  return expression ? expression->Accept(*eval) : TypedValue(eval->GetMemoryResource());
 }
 
 auto ToOptionalString(ExpressionEvaluator *evaluator, Expression *expression) -> std::optional<utils::pmr::string> {
-  const auto evaluated_expr = EvaluateOptionalExpression(expression, evaluator);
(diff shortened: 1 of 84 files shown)


Ignition:
### Tracking


### Standard development
- [ ] Update unit/E2E tests


### CI Testing Labels


### Documentation checklist
    - **[ Release note text ]**

andrejtonev (approved):

src/query/plan/operator.cpp
```diff
@@ -5720,13 +5720,13 @@ std::vector<Symbol> LoadCsv::ModifiedSymbols(const SymbolTable &sym_table) const
 namespace {
 // copy-pasted from interpreter.cpp
 TypedValue EvaluateOptionalExpression(Expression *expression, ExpressionEvaluator *eval) {
-  return expression ? expression->Accept(*eval) : TypedValue();
+  return expression ? expression->Accept(*eval) : TypedValue(eval->GetMemoryResource());
 }
```
colinbarry, line 5724: Good catch re: the default initialized `TypedValue`. It's beyond the scope of this PR, but are there any places where we use `TypedValue`s with the default `utils::NewDeleteResource()`? I guess what I'm asking is: would it be worth removing the default constructor, and also the default `utils::MemoryResource` arguments from the other constructors? If we require the `MemoryResource` to be explicitly stated, we're less likely to slip-up here in the future. (I'm sure I've seen a few other PRs recently that fixed some incorrect `TypedValue` memory resources. Can't find them now, so I could be mistaken!)

colinbarry (approved):
All good.

PR #2735, opened 2025-02-24
Bugfix: Unregister replication instance could deadlock

as51340:
Added timeouts for lock acquiring in success and fail callbacks. Unregistration is also forbidden in the case when there is no current main in the cluster to prevent drifting between in-memory and raft state on coordinators. 

Instance checker will now be destroyed before RPC client to prevent possible UB.

### DIFF
--- a/src/coordination/coordinator_instance.cpp
+++ b/src/coordination/coordinator_instance.cpp
@@ -677,56 +677,89 @@ auto CoordinatorInstance::RegisterReplicationInstance(DataInstanceConfig const &
 
 auto CoordinatorInstance::UnregisterReplicationInstance(std::string_view instance_name)
     -> UnregisterInstanceCoordinatorStatus {
-  spdlog::trace("Acquiring lock to unregister instance thread {}", std::this_thread::get_id());
-  auto lock = std::lock_guard{coord_instance_lock_};
-  spdlog::trace("Acquired lock to demote instance to replica");
+  spdlog::trace("Acquiring lock to unregister instance in thread {} 1st time", std::this_thread::get_id());
 
-  if (status.load(std::memory_order_acquire) != CoordinatorStatus::LEADER_READY) {
-    return UnregisterInstanceCoordinatorStatus::NOT_LEADER;
-  }
+  std::optional<std::reference_wrapper<ReplicationInstanceConnector>> maybe_instance;
+  {
+    // 1st time we acquire the lock in order to find replication instance which the user wants to unregister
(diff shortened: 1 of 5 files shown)


src/coordination/coordinator_instance.cpp
```diff
@@ -863,69 +879,72 @@ void CoordinatorInstance::InstanceSuccessCallback(std::string_view instance_name
     return;
   }
 
-  auto lock = std::unique_lock{coord_instance_lock_};
+  auto lock = std::unique_lock{coord_instance_lock_, std::defer_lock};
+  if (!lock.try_lock_for(500ms)) {
+    spdlog::trace("Failed to acquire lock in InstanceSuccessCallback in 500ms");
+  } else {
```
andrejtonev, line 885: Do we need the if else or can we just return if the lock fails?

src/coordination/coordinator_instance.cpp
```diff
@@ -677,56 +678,71 @@ auto CoordinatorInstance::RegisterReplicationInstance(DataInstanceConfig const &
...
   auto const curr_main_uuid = raft_state_->GetCurrentMainUUID();
 
   auto coordinator_instances = raft_state_->GetCoordinatorInstancesContext();
-  if (!raft_state_->AppendClusterUpdate(std::move(cluster_state), std::move(coordinator_instances), curr_main_uuid)) {
+  // Append new cluster state. We may need to restore old state if something goes wrong.
+  if (!raft_state_->AppendClusterUpdate(std::move(new_data_instances), coordinator_instances, curr_main_uuid)) {
     return UnregisterInstanceCoordinatorStatus::RAFT_LOG_ERROR;
   }
 
-  inst_to_remove->StopStateCheck();
```
andrejtonev, line 716: Why has this been removed?

as51340: Since there are now timeouts on lock, checks can get stopped once the instance is removed from repl_instances_ and that is in line 733.

src/coordination/coordinator_instance.cpp
as51340: I thought we need else, but I think you are right, the lock should be held still no matter the else branch

src/coordination/coordinator_instance.cpp
andrejtonev: In that case update the `ReplicationInstanceClient` since currently the check will be the last to be destroyed.

as51340: makes sense

andrejtonev (approved):

PR #2754, opened 2025-02-27
Update vector index if vertex has the respective label

Josipmrden:
The check was missing when adding vector index after import.


Josipmrden:
Closing because fix is present in https://github.com/memgraph/memgraph/pull/2701

PR #2761, opened 2025-02-28
Add multiple vector indices per label property

Josipmrden:
Added support for multiple vector indices per label property for different metrics.

### DIFF
--- a/src/storage/v2/indices/vector_index.cpp
+++ b/src/storage/v2/indices/vector_index.cpp
@@ -65,9 +65,12 @@ static const std::unordered_map<unum::usearch::metric_kind_t, std::function<doub
 /// The `Impl` structure follows the PIMPL (Pointer to Implementation) idiom to separate
 /// the interface of `VectorIndex` from its implementation
 struct VectorIndex::Impl {
+  /// Maps the index name to the associated IndexItem
+  std::map<std::string, IndexItem> index_;
+
   /// The `index_` member is a map that associates a `LabelPropKey` (a combination of label and property)
-  /// with the pair of a IndexItem.
-  std::map<LabelPropKey, IndexItem> index_;
+  /// with the indexes associated with it
+  std::map<LabelPropKey, std::vector<std::string>> label_prop_to_index_names_;
 
   /// The `index_name_to_label_prop_` is a map that maps an index name (as a string) to the corresponding
   /// `LabelPropKey`. This allows the system to quickly resolve an index name to the spec
@@ -98,6 +101,8 @@ const char *VectorIndex::NameFromMetric(unum::usearch::metric_kind_t metric) {
       return "tanimoto";
     case unum::usearch::metric_kind_t::sorensen_k:
       return "sorensen";
+    case unum::usearch::metric_kind_t::jaccard_k:
+      return "jaccard";
     default:
       throw query::VectorSearchException(
           "Unsupported metric kind. Supported metrics are l2sq, ip, cos, haversine, divergence, pearson, hamming, "
@@ -133,6 +138,9 @@ unum::usearch::metric_kind_t VectorIndex::MetricFromName(std::string_view name) 
   if (name == "sorensen") {
     return unum::usearch::metric_kind_t::sorensen_k;
   }
+  if (name == "jaccard") {
+    return unum::usearch::metric_kind_t::jaccard_k;
+  }
   throw query::VectorSearchException(
       fmt::format("Unsupported metric name: {}. Supported metrics are l2sq, ip, cos, haversine, divergence, pearson, "
                   "hamming, tanimoto, and sorensen.",
@@ -141,6 +149,7 @@ unum::usearch::metric_kind_t VectorIndex::MetricFromName(std::string_view name) 
 
 bool VectorIndex::CreateIndex(const VectorIndexSpec &spec, utils::SkipList<Vertex>::Accessor &vertices,
                               std::optional<SnapshotObserverInfo> const &snapshot_info) {
+  const auto label_prop = LabelPropKey{spec.label, spec.property};
   try {
     // Create the index
     const unum::usearch::metric_punned_t metric(spec.dimension, spec.metric_kind, unum::usearch::scalar_kind_t::f32_k);
@@ -148,11 +157,10 @@ bool VectorIndex::CreateIndex(const VectorIndexSpec &spec, utils::SkipList<Verte
     // use the number of workers as the number of possible concurrent index operations
     const unum::usearch::index_limits_t limits(spec.capacity, FLAGS_bolt_num_workers);
 
-    const auto label_prop = LabelPropKey{spec.label, spec.property};
-    if (pimpl->index_.contains(label_prop) || pimpl->index_name_to_label_prop_.contains(spec.index_name)) {
+    if (pimpl->index_.contains(spec.index_name)) {
       throw query::VectorSearchException("Given vector index already exists.");
     }
-    pimpl->index_name_to_label_prop_.try_emplace(spec.index_name, label_prop);
+
     auto mg_vector_index = mg_vector_index_t::make(metric);
     if (!mg_vector_index) {
       throw query::VectorSearchException(fmt::format("Failed to create vector index {}, error message: {}",
@@ -164,7 +172,13 @@ bool VectorIndex::CreateIndex(const VectorIndexSpec &spec, utils::SkipList<Verte
       throw query::VectorSearchException(
           fmt::format("Failed to create vector index {}", spec.index_name, ". Failed to reserve memory for the index"));
     }
-    pimpl->index_.try_emplace(label_prop,
+
+    pimpl->index_name_to_label_prop_.try_emplace(spec.index_name, label_prop);
+    if (!pimpl->label_prop_to_index_names_.contains(label_prop)) {
+      pimpl->label_prop_to_index_names_.try_emplace(label_prop, std::vector<std::string>{});
+    }
+    pimpl->label_prop_to_index_names_[label_prop].push_back(spec.index_name);
+    pimpl->index_.try_emplace(spec.index_name,
                               IndexItem{std::make_shared<utils::Synchronized<mg_vector_index_t, std::shared_mutex>>(
                                             std::move(mg_vector_index.index)),
                                         spec});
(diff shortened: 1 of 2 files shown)


Josipmrden:
### Tracking


### Standard development
- [ ] Update unit/E2E tests


### CI Testing Labels


### Documentation checklist
    - **[ Release note text ]**

src/storage/v2/disk/storage.cpp
```diff
@@ -2076,14 +2076,14 @@ utils::BasicResult<storage::StorageIndexDefinitionError, void> DiskStorage::Disk
                                  kErrorMessage);
 }
 
-utils::BasicResult<storage::StorageIndexDefinitionError, void> DiskStorage::DiskAccessor::CreateVectorIndex(
-    VectorIndexSpec /*spec*/) {
+utils::BasicResult<storage::VectorIndexStorageError, VectorIndex::CreationStatus>
+DiskStorage::DiskAccessor::CreateVectorIndex(VectorIndexSpec /*spec*/) {
```
as51340, line 2080: same

src/storage/v2/indices/vector_index.hpp
```diff
@@ -175,17 +191,15 @@ class VectorIndex {
...
-  void TryInsertVertex(Vertex *vertex);
-
  private:
   /// @brief Adds a vertex to an existing index.
   /// @param vertex The vertex to be added.
   /// @param label_prop The label and property key for the index.
   /// @param value The value of the property.
   /// @throw query::VectorSearchException
-  bool UpdateVectorIndex(Vertex *vertex, const LabelPropKey &label_prop, const PropertyValue *value = nullptr);
+  utils::BasicResult<VectorIndexStorageError, void> UpdateVectorIndex(Vertex *vertex, const LabelPropKey &label_prop,
```
as51340, line 200: std::optional<error>, BasicResult is variant, you don't need it here

src/storage/v2/indices/vector_index.cpp
```diff
@@ -65,9 +65,12 @@ static const std::unordered_map<unum::usearch::metric_kind_t, std::function<doub
...
 /// the interface of `VectorIndex` from its implementation
 struct VectorIndex::Impl {
+  /// Maps the index name to the associated IndexItem
+  std::map<std::string, IndexItem> index_;
+
   /// The `index_` member is a map that associates a `LabelPropKey` (a combination of label and property)
-  /// with the pair of a IndexItem.
-  std::map<LabelPropKey, IndexItem> index_;
+  /// with the indexes associated with it
+  std::map<LabelPropKey, std::vector<std::string>> label_prop_to_index_names_;
```
as51340, line 73: I think this is not modeled in the best way. You already have index_name stored under VectorIndexSpec.

Just use
std::map<LabelPropKey, std::vector<IndexItem>>

src/storage/v2/indices/vector_index.cpp
```diff
@@ -133,142 +138,191 @@ unum::usearch::metric_kind_t VectorIndex::MetricFromName(std::string_view name)
...
-    if (!mg_vector_index) {
-      throw query::VectorSearchException(fmt::format("Failed to create vector index {}, error message: {}",
-                                                     spec.index_name, mg_vector_index.error.what()));
+  if (pimpl->index_.contains(spec.index_name)) {
+    return VectorIndex::CreationStatus::ALREADY_EXISTS;
+  }
+
+  auto mg_vector_index = mg_vector_index_t::make(metric);
+  if (!mg_vector_index) {
+    spdlog::error(fmt::format("Failed to create vector index {}, error message: {}", spec.index_name,
```
as51340, line 166: no need to use fmt::format inside spdlog::error

src/storage/v2/indices/vector_index.cpp
```diff
@@ -133,142 +138,191 @@ unum::usearch::metric_kind_t VectorIndex::MetricFromName(std::string_view name)
...
+  }
+
+  if (mg_vector_index.index.try_reserve(limits)) {
+    spdlog::info("Created vector index {}", spec.index_name);
+  } else {
+    spdlog::error(fmt::format("Unable to reserve memory for vector index {}", spec.index_name));
+    return VectorIndexStorageError::UnableToReserveMemory;
+  }
+
+  pimpl->index_name_to_label_prop_.try_emplace(spec.index_name, label_prop);
```
as51340, line 178: Check if emplaced successfully

src/storage/v2/indices/vector_index.cpp
```diff
@@ -133,142 +138,191 @@ unum::usearch::metric_kind_t VectorIndex::MetricFromName(std::string_view name)
...
+  if (!pimpl->label_prop_to_index_names_.contains(label_prop)) {
+    pimpl->label_prop_to_index_names_.try_emplace(label_prop, std::vector<std::string>{});
```
as51340, line 180: check error

src/storage/v2/indices/vector_index.cpp
```diff
@@ -133,142 +138,191 @@ unum::usearch::metric_kind_t VectorIndex::MetricFromName(std::string_view name)
...
+  }
+  pimpl->label_prop_to_index_names_[label_prop].push_back(spec.index_name);
+  pimpl->index_.try_emplace(spec.index_name,
```
as51340, line 183: check error

src/storage/v2/indices/vector_index.cpp
```diff
@@ -133,142 +138,191 @@ unum::usearch::metric_kind_t VectorIndex::MetricFromName(std::string_view name)
...
-  // try to remove entry (if it exists) and then add a new one + check if index is full
-  {
-    auto locked_index = mg_index->MutableSharedLock();
-    locked_index->remove(vertex);
-    is_index_full = locked_index->size() == locked_index->capacity();
-  }
+utils::BasicResult<VectorIndexStorageError, void> VectorIndex::UpdateVectorIndex(Vertex *vertex,
+                                                                                 const LabelPropKey &label_prop,
+                                                                                 const PropertyValue *value,
+                                                                                 const std::string *specific_index) {
```
as51340, line 238: don't pass string as pointer, use ref or std::string_view

as51340 (changes requested):

as51340:
@Josipmrden it seems that changing metrics is extremely simple in usearch and is basically free. This would mean we don't need to create multiple entries, we can just change the metric we are using when querying. @DavIvek can you check? https://github.com/unum-cloud/usearch/blob/main/include/usearch/index_dense.hpp

src/storage/v2/indices/vector_index.cpp
Josipmrden: emplace gives false if the key entry already exists. but we do check for that before. So I wouldn't say we need to check it?

src/storage/v2/indices/vector_index.cpp
```diff
@@ -133,142 +138,190 @@ unum::usearch::metric_kind_t VectorIndex::MetricFromName(std::string_view name)
...
 }
 
-bool VectorIndex::DropIndex(std::string_view index_name) {
+utils::BasicResult<VectorIndexStorageError, VectorIndex::DeletionStatus> VectorIndex::DropIndex(
+    std::string_view index_name) {
+  if (!pimpl->index_.contains(index_name.data())) {
+    return VectorIndex::DeletionStatus::NOT_FOUND;
+  }
+
   auto it = pimpl->index_name_to_label_prop_.find(index_name.data());
```
DavIvek, line 211: Can this be checked in one lookup?

src/storage/v2/indices/vector_index.cpp
```diff
@@ -133,142 +138,190 @@ unum::usearch::metric_kind_t VectorIndex::MetricFromName(std::string_view name)
...
-    if (!mg_vector_index) {
-      throw query::VectorSearchException(fmt::format("Failed to create vector index {}, error message: {}",
-                                                     spec.index_name, mg_vector_index.error.what()));
+  if (pimpl->index_.contains(spec.index_name)) {
+    return VectorIndexStorageError::VectorIndexAlreadyExists;
+  }
+
+  auto mg_vector_index = mg_vector_index_t::make(metric);
+  if (!mg_vector_index) {
+    spdlog::error("Failed to create vector index {}, error message: {}", spec.index_name, mg_vector_index.error.what());
```
DavIvek, line 166: Do we need all this logging? With these changes, the error code is returned in a clear way, and then exceptions with essentially the same messages are thrown in the interpreter.

src/storage/v2/indices/vector_index.cpp
```diff
@@ -305,23 +363,24 @@ std::vector<VectorIndexSpec> VectorIndex::ListIndices() const {
...
   auto locked_index = mg_index->ReadLock();
   return locked_index->size();
 }
 
 std::vector<std::tuple<Vertex *, double, double>> VectorIndex::Search(std::string_view index_name,
                                                                       uint64_t result_set_size,
                                                                       const std::vector<float> &query_vector) const {
-  const auto label_prop = pimpl->index_name_to_label_prop_.find(index_name);
-  if (label_prop == pimpl->index_name_to_label_prop_.end()) {
+  if (pimpl->index_.find(index_name.data()) == pimpl->index_.end()) {
```
DavIvek, line 380: Performing two lookups again. It’s not a big issue since we don’t expect these maps to grow large, but it can still be improved.
Additionally, `index_` can use `std::less<>` for `_Compare`, which enables heterogeneous lookup, eliminating the need to call .data() every time.

src/storage/v2/indices/vector_index.hpp
```diff
@@ -18,12 +18,24 @@
 #include "storage/v2/id_types.hpp"
 #include "storage/v2/property_value.hpp"
 #include "storage/v2/snapshot_observer_info.hpp"
+#include "storage/v2/storage_error.hpp"
```
DavIvek, line 21: Do we need this include?

DavIvek:
> @Josipmrden it seems that changing metrics is extremely simple in usearch and is basically free. This would mean we don't need to create multiple entries, we can just change the metric we are using when querying. @DavIvek can you check? https://github.com/unum-cloud/usearch/blob/main/include/usearch/index_dense.hpp

It's true that we can dynamically change the metric, but that could worsen the performance of the index.

src/storage/v2/indices/vector_index.cpp
Josipmrden: this is more for letting us know what happened if someone sends us logs. It will not be logged that much since errors rarely happen

src/storage/v2/indices/vector_index.hpp
Josipmrden: will remove

src/storage/v2/indices/vector_index.cpp
Josipmrden: did a new implementation with iterators

src/storage/v2/indices/vector_index.cpp
```diff
@@ -133,142 +136,203 @@ unum::usearch::metric_kind_t VectorIndex::MetricFromName(std::string_view name)
...
+  const auto label_prop = LabelPropKey{spec.label, spec.property};
+  // Create the index
+  const unum::usearch::metric_punned_t metric(spec.dimension, spec.metric_kind, unum::usearch::scalar_kind_t::f32_k);
+
+  // use the number of workers as the number of possible concurrent index operations
+  const unum::usearch::index_limits_t limits(spec.capacity, FLAGS_bolt_num_workers);
+
+  if (pimpl->index_name_to_label_prop_.find(spec.index_name) != pimpl->index_name_to_label_prop_.end()) {
+    return VectorIndexStorageError::VectorIndexAlreadyExists;
+  }
```
as51340, line 160: would be better if you would use std::set instead of vector and then rely on try_emplace whether it was inserted. In this way you have to add enum InconsistentState which hides what is actually going on

src/storage/v2/indices/vector_index.cpp
```diff
@@ -133,142 +136,203 @@ unum::usearch::metric_kind_t VectorIndex::MetricFromName(std::string_view name)
...
+      return VectorIndexStorageError::InconsistentState;
     }
-    pimpl->index_name_to_label_prop_.try_emplace(spec.index_name, label_prop);
-    auto mg_vector_index = mg_vector_index_t::make(metric);
-    if (!mg_vector_index) {
-      throw query::VectorSearchException(fmt::format("Failed to create vector index {}, error message: {}",
-                                                     spec.index_name, mg_vector_index.error.what()));
+  }
+
+  pimpl->index_[label_prop].push_back(IndexItem{
```
as51340, line 188: reuse the iterator from line 181 instead of searching again into the index

as51340 (changes requested):

Josipmrden:
image
functionality doesn't exist for v3.5. Moving milestone to v3.6

Josipmrden:
leaving for @DavIvek  as a reference

PR #2770, opened 2025-03-06
Bugfix: Avoid throwing in slk when a buffer isn't read fully

as51340:
slk::Reader would throw before if data from the buffer wouldn't be completely ready. This caused network connections between main and replica to be constantly dying when unnecessary. Now when replica doesn't read all data from the buffer, the exception thrown in the slk part of the code is handled properly, avoiding the connection drop. Additionally, the PR tries to reduce the massive amount of logs produced in HA code when trace log level is used. The PR also improves jepsen's run script so that it doesn't fail when jepsen.log cannot be read. With this fix, we should be able to see at least part of the logs which can also be helpful for debugging. Jepsen tests are also checking whether `There is still leftover data in the SLK stream` exception will be thrown during the execution.

### DIFF
--- a/src/dbms/inmemory/replication_handlers.cpp
+++ b/src/dbms/inmemory/replication_handlers.cpp
@@ -789,7 +789,7 @@ std::pair<uint64_t, uint32_t> InMemoryReplicationHandlers::ReadAndApplyDeltasSin
   auto current_durable_commit_timestamp = max_delta_timestamp;
   spdlog::trace("Current durable commit timestamp: {}", current_durable_commit_timestamp);
 
-  auto prev_printed_timestamp = 0;
+  uint64_t prev_printed_timestamp = 0;
 
   for (bool transaction_complete = false; !transaction_complete; ++current_delta_idx, ++current_batch_counter) {
     if (current_batch_counter == kDeltasBatchProgressSize) {
--- a/src/dbms/inmemory/replication_handlers.cpp
+++ b/src/dbms/inmemory/replication_handlers.cpp
@@ -30,7 +30,6 @@
 #include <range/v3/view/filter.hpp>
 #include <range/v3/view/transform.hpp>
 
-using memgraph::replication_coordination_glue::ReplicationRole;
 using memgraph::storage::Delta;
 using memgraph::storage::EdgeAccessor;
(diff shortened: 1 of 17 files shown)


as51340:
### Tracking


### Standard development
- [ ] Update unit/E2E tests


### CI Testing Labels


### Documentation checklist
    - **[ Release note text ]**

src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -282,6 +281,10 @@ void InMemoryReplicationHandlers::AppendDeltasHandler(dbms::DbmsHandler *dbms_ha
   storage::replication::AppendDeltasReq req;
   slk::Load(&req, req_reader);
 
+  // Read at the beginning so that SLK stream gets cleared even when the request is invalid
+  storage::replication::Decoder decoder(req_reader);
+  auto maybe_epoch_id = decoder.ReadString();
+
```
andrejtonev, line 287: I guess we need to read the whole message before throwing in every handler, right?

as51340: yes

andrejtonev: Is this still needed if the try catch has been added?

src/rpc/protocol.cpp
```diff
@@ -85,8 +85,11 @@ void Session::Execute() {
     throw rpc::SlkRpcFailedException();
   }
 
-  // Finalize the SLK streams.
-  req_reader.Finalize();
+  // Finalize the SLK stream. It may fail because not all data has been read, that's fine.
+  try {
+    req_reader.Finalize();
+  } catch (const slk::SlkReaderLeftoverDataException &) {
+  }
```
andrejtonev, line 92: The Finalize can be put inside the previous try block.

src/dbms/inmemory/replication_handlers.cpp
as51340: no, will revert

src/rpc/protocol.cpp
as51340: yep

andrejtonev (approved):

PR #2812, opened 2025-03-20
Improve the failover process in HA when using multitenancy

as51340:
The failover in the MT code now has a different logic. We 1st try to find the instance which has most newest databases. If there are multiple instances that are newest on the same number of databases, the sum of timestamps of all databases decides. The instance which has larger sum it is assumed to be newer.


src/coordination/coordinator_instance.cpp
```diff
@@ -1032,104 +1031,126 @@ void CoordinatorInstance::InstanceFailCallback(std::string_view instance_name,
...
 
-auto CoordinatorInstance::ChooseMostUpToDateInstance(std::span<InstanceNameDbHistories> instance_database_histories)
-    -> std::optional<NewMainRes> {
+auto CoordinatorInstance::ChooseMostUpToDateInstance(
+    std::map<std::string, replication_coordination_glue::InstanceInfo> const &instances_info)
+    -> std::optional<std::string> {
   utils::MetricsTimer const timer{metrics::ChooseMostUpToDateInstance_us};
 
-  std::optional<NewMainRes> new_main_res;
+  // Find the instance with the largest system committed timestamp
```
andrejtonev, line 1039: This is too hard to follow.
Can we just make it simpler even if less optimal?
More logging will also be useful

src/replication_coordination_glue/common.hpp
```diff
@@ -11,21 +11,18 @@
...
-  std::vector<std::pair<std::string, uint64_t>> history;
-  std::string name;  // db name
+struct InstanceDBInfo {
+  std::string db_uuid;
+  uint64_t latest_durable_timestamp;
 };
 
-using DatabaseHistories = std::vector<DatabaseHistory>;
+struct InstanceInfo {
+  std::vector<InstanceDBInfo> dbs_info;
```
andrejtonev, line 24: Making this a map between db and ts might make some processes easier down the line.

src/coordination/coordinator_instance.cpp
andrejtonev: There should always be very few instance and perhaps many tenants
I my mind we can break it like so:
1. find newest system ts and log
2. filter instances based on the ts and log
2.a. if one instance left, we are done
3. use the latest system instance to loop through dbs and their ts
3.a. loop through instances, find the db and ++ for the max one
3.b. since looping through everything, we can also do the ts sum here as well
4. check which instance has the most up to date dbs and log
4.a. if one instance, we are done
5. check which instance has the most sum ts and log
5.a. if one instance, we are done
6. pick one

.github/workflows/stress_jepsen.yaml
```diff
@@ -7,7 +7,7 @@ on:
 
 env:
   ARCH: 'amd'
-  BUILD_TYPE: 'RelWithDebInfo'
```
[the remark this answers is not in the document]
as51340: No, I don't know what was I trying

src/coordination/coordinator_instance.cpp
as51340: Partially agree. I added some lambdas to make it simpler.
1. :white_check_mark: 
2. and 3. It is much easier to preprocess and to have outer loop going over DBs for me. In this way for each DB I have a set of steps:
i.) find instances with max ts
ii) update_counters for max_instances


For steps 4 and 5 this is the code that does exactly this:
```
std::optional<std::pair<std::string, uint64_t>> newest_instance;
  for (auto const &[instance_name, cnt_newest_dbs] : total_instances_counter) {
    // If better on more DBs, update currently the best
    if (!newest_instance.has_value() || newest_instance->second < cnt_newest_dbs) {
      newest_instance.emplace(instance_name, cnt_newest_dbs);
    } else if (newest_instance->second == cnt_newest_dbs) {
      // Instances are the best over the same number of instances, let the sum of timestamps decide
      spdlog::info("Instances {} and {} are most up to date on the same number of instances.", instance_name,
                   newest_instance->first);
      if (total_instances_sum[instance_name] > total_instances_sum[newest_instance->first]) {
        spdlog::info("Instance {} has the total sum of timestamps larger than {}. It will be considered newer.",
                     instance_name, newest_instance->first);
        newest_instance.emplace(instance_name, cnt_newest_dbs);
      }
    }
  }
```

Compares counters 1st and if counters are the same, uses sum to decide. I think in your case you would have the same amount of code or more.

src/replication_coordination_glue/common.hpp
as51340: I am not searching anywhere for db's timestamp because I am looping over instances

andrejtonev (approved):

PR #2817, opened 2025-03-21
Bugfix: Handle properly network resolution failures

as51340:
Wherever we logged resolved database's endpoint, we now log non-resolved entry. This is better for users because IP could change while hostname won't. Logging non-resolved endpoint also relieves us of wrapping each log message with try-catch block.

### DIFF
--- a/src/io/network/addrinfo.cpp
+++ b/src/io/network/addrinfo.cpp
@@ -20,16 +20,16 @@ namespace memgraph::io::network {
 
 static_assert(std::forward_iterator<AddrInfo::Iterator> && std::equality_comparable<AddrInfo::Iterator>);
 
-AddrInfo::AddrInfo(const Endpoint &endpoint) : AddrInfo(endpoint.GetResolvedIPAddress(), endpoint.GetPort()) {}
+AddrInfo::AddrInfo(const Endpoint &endpoint) : AddrInfo(endpoint.GetAddress(), endpoint.GetPort()) {}
 
-AddrInfo::AddrInfo(const std::string &addr, uint16_t port) : info_{nullptr, nullptr} {
+AddrInfo::AddrInfo(const std::string &non_resolved_addr, uint16_t port) : info_{nullptr, nullptr} {
   addrinfo hints{
       .ai_flags = AI_PASSIVE,
       .ai_family = AF_UNSPEC,     // IPv4 and IPv6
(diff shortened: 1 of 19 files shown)


as51340:
### Tracking


### Standard development
- [ ] Update unit/E2E tests


### CI Testing Labels


### Documentation checklist
    - **[ Release note text ]**

src/io/network/addrinfo.cpp
```diff
@@ -20,17 +20,19 @@ namespace memgraph::io::network {
...
       .ai_flags = AI_PASSIVE,
       .ai_family = AF_UNSPEC,     // IPv4 and IPv6
       .ai_socktype = SOCK_STREAM  // TCP socket
   };
   addrinfo *info = nullptr;
-  auto status = getaddrinfo(addr.c_str(), std::to_string(port).c_str(), &hints, &info);
-  if (status != 0) throw NetworkError(gai_strerror(status));
+  auto status = getaddrinfo(non_resolved_addr.c_str(), std::to_string(port).c_str(), &hints, &info);
+  if (status != 0) {
+    throw NetworkError(gai_strerror(status));
```
andrejtonev, line 34: addrinfo can leak if this throws. 
any reason we don't just use the info_?

src/io/network/socket.cpp
```diff
@@ -148,8 +153,7 @@ bool Socket::Bind(const Endpoint &endpoint) {
     return false;
   }
 
-  // TODO: (andi) This runs again resolving, shouldn't be needed.
-  endpoint_ = Endpoint(endpoint.GetResolvedIPAddress(), ntohs(portdata.sin6_port));
+  endpoint_ = Endpoint(endpoint.GetAddress(), ntohs(portdata.sin6_port));
```
andrejtonev, line 156: Seems like `endpoint_` is mostly used for logging. Is that the case? Is there a situation where we would need to have it be resolved?

as51340: only when connecting, in socket.cpp

src/io/network/addrinfo.cpp
andrejtonev: Ignore; status should protect in this case.

andrejtonev (approved):

src/io/network/socket.cpp
andrejtonev: which is handled by AddrInfo :+1:

PR #2818, opened 2025-03-21
Snapshot creation parallelization

andrejtonev:
TODO: flag to enable/disable

### DIFF
--- a/src/storage/v2/durability/serialization.hpp
+++ b/src/storage/v2/durability/serialization.hpp
@@ -1,4 +1,4 @@
-// Copyright 2024 Memgraph Ltd.
+// Copyright 2025 Memgraph Ltd.
 //
 // Use of this software is governed by the Business Source License
 // included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
@@ -44,6 +44,7 @@ class BaseEncoder {
 /// Encoder that is used to generate a snapshot/WAL.
 class Encoder final : public BaseEncoder {
  public:
+  void Initialize(const std::filesystem::path &path);
   void Initialize(const std::filesystem::path &path, std::string_view magic, uint64_t version);
 
   void OpenExisting(const std::filesystem::path &path);
@@ -81,6 +82,8 @@ class Encoder final : public BaseEncoder {
 
   // Get the total size of the current file.
   size_t GetSize();
+
+  auto native_handle() const { return file_.fd(); }
 
  private:
   utils::OutputFile file_;
--- a/src/storage/v2/durability/snapshot.cpp
+++ b/src/storage/v2/durability/snapshot.cpp
@@ -11,6 +11,15 @@
 
 #include "storage/v2/durability/snapshot.hpp"
 
+#include <openssl/x509v3.h>
+#include <sys/sendfile.h>
+#include <sys/stat.h>
+#include <atomic>
+#include <filesystem>
+#include <future>
+#include <limits>
+#include <mutex>
+#include <string>
 #include <thread>
 #include <usearch/index_plugins.hpp>
 
@@ -42,6 +51,7 @@
 #include "utils/file_locker.hpp"
 #include "utils/logging.hpp"
 #include "utils/message.hpp"
+#include "utils/on_scope_exit.hpp"
 #include "utils/spin_lock.hpp"
 #include "utils/synchronized.hpp"
 
@@ -693,7 +703,7 @@ LoadPartialConnectivityResult LoadPartialConnectivity(
           // edges in the in/out edges list of vertices, therefore the edges has to be created here.
           if (snapshot_has_edges) {
             auto edge = edge_acc.find(Gid::FromUint(*edge_gid));
-            if (edge == edge_acc.end()) throw RecoveryFailure(":ouldn't find edge in the loaded edges!");
+            if (edge == edge_acc.end()) throw RecoveryFailure("Couldn't find edge in the loaded edges!");
             edge_ref = EdgeRef(&*edge);
           } else {
             auto [edge, inserted] = edge_acc.insert(Edge{Gid::FromUint(*edge_gid), nullptr});
@@ -3891,6 +3901,8 @@ void CreateSnapshot(Storage *storage, Transaction *transaction, const std::files
   uint64_t offset_edge_batches = 0;
   uint64_t offset_vertex_batches = 0;
 
+  const auto header_size = 0;
+
   auto write_offsets = [&] {
     snapshot.WriteUint(offset_edges);
     snapshot.WriteUint(offset_vertices);
@@ -3913,8 +3925,8 @@ void CreateSnapshot(Storage *storage, Transaction *transaction, const std::files
   }
 
   // Object counters.
-  uint64_t edges_count = 0;
-  uint64_t vertices_count = 0;
+  std::atomic<uint64_t> edges_count = 0;
+  std::atomic<uint64_t> vertices_count = 0;
 
   // Mapper data.
   std::unordered_set<uint64_t> used_ids;
@@ -3923,15 +3935,62 @@ void CreateSnapshot(Storage *storage, Transaction *transaction, const std::files
     snapshot.WriteUint(mapping.AsUint());
   };
 
+  auto write_mapping2 = [](auto &snapshot, auto &used_ids, auto mapping) {
+    used_ids.insert(mapping.AsUint());
+    snapshot.WriteUint(mapping.AsUint());
+  };
+
+  const auto kN =
+      (vertices->access().size() + edges->access().size()) / storage->config_.durability.items_per_batch + 1;
+
+  struct BatchRes {
+    std::vector<BatchInfo> batch_info{};
+    std::unordered_set<uint64_t> used_ids{};
+    uint64_t count{0};
+    std::filesystem::path snapshot_path{};
+    size_t snapshot_size{0};
+  };
+
+  auto batch = [kN, storage](auto &&acc) {
+    std::vector<int64_t> batches(kN + 1, std::numeric_limits<int64_t>::max());  // start and end gids
+    const uint64_t step = std::max(std::max(acc.size() / kN, 1UL), storage->config_.durability.items_per_batch);
+    batches[0] = 0;  // Always start from the lowest possible
+    int i = 0;
+    int t = 1;
+    if (acc.size() <= t * kN) return batches;
+    for (const auto &e : acc) {
+      if (step == i++) {
+        batches[t++] = e.gid.AsInt();
+        i = 0;
+        if (acc.size() < t * kN) break;
+        if (t == kN) break;
+      }
+    }
+    return batches;
+  };
+
   std::vector<BatchInfo> edge_batch_infos;
-  auto items_in_current_batch{0UL};
-  auto batch_start_offset{0UL};
+  auto start_egid = batch(edges->access());
+
   // Store all edges.
-  if (storage->config_.salient.items.properties_on_edges) {
-    offset_edges = snapshot.GetPosition();
-    batch_start_offset = offset_edges;
+  auto store_edges = [&](int id) -> BatchRes {
+    if (!storage->config_.salient.items.properties_on_edges) return {};
+    const auto start_gid = start_egid[id];
+    const auto end_gid = start_egid[id + 1];
+    if (start_gid >= end_gid) return {};
+
+    auto items_in_current_batch{0UL};
+    auto batch_start_offset{0UL};
+    BatchRes res;
+    res.snapshot_path = path.string() + "_edge_part_" + std::to_string(id);
+
+    Encoder edges_snapshot;
+    edges_snapshot.Initialize(res.snapshot_path);
+    batch_start_offset = edges_snapshot.GetPosition();
     auto acc = edges->access();
-    for (auto &edge : acc) {
+    auto it = acc.find_equal_or_greater(Gid::FromInt(start_gid));
+    for (; it != acc.end() && it->gid.AsInt() < end_gid; ++it) {
+      auto &edge = *it;
       // The edge visibility check must be done here manually because we don't
       // allow direct access to the edges through the public API.
       bool is_visible = true;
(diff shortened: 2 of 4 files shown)


andrejtonev:

---
___


### Tracking


### Standard development
- [ ] Update unit/E2E tests


### CI Testing Labels


### Documentation checklist
    - **[ Release note text ]**

src/storage/v2/durability/snapshot.cpp
```diff
@@ -127,6 +139,230 @@ struct BatchInfo {
...
+    std::lock_guard<std::mutex> lock(mtx_);
+    return tasks_.size();
+  }
+
+ private:
+  mutable std::mutex mtx_;
+  std::queue<task_t> tasks_;
+};
+
+void WaitAndCombine(std::vector<std::pair<SnapshotBatchRes, std::promise<bool>>> &edge_res,
```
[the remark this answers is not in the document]
as51340: does it make sense to pass type as task_results_t since you are using it in MultiThreadedWorkflow

src/storage/v2/durability/snapshot.cpp
```diff
@@ -127,6 +139,230 @@ struct BatchInfo {
...
+      if (bytes_sent == -1 || bytes_sent != size) {
+        throw RecoveryFailure("Couldn't append vertex part to snapshot!");
+      }
+      snapshot.SetPosition(current_offset + size);
+    }
+  }
+}
+
+// Return at least one batch (at least 2 elements: start and end gid)
+auto Batch(auto &&acc, const uint64_t step) {
```
as51340, line 267: maybe rename step to items_per_batch, makes it more clear to me what's the purpose of it

src/storage/v2/durability/snapshot.cpp
```diff
@@ -127,6 +139,230 @@ struct BatchInfo {
...
+  task_results_t edge_res{};
+  if (edges != nullptr) {  // No edges skiplist <=> no properties on edges
+    start_egid = Batch(edges->access(), items_per_batch);
+    edge_res = task_results_t{start_egid.size() - 1};  // last element is an end marker
+    for (int i = 0; i < edge_res.size(); ++i) {
+      tasks.AddTask([&, id = i] {
+        const auto start_gid = start_egid[id];
+        const auto end_gid = start_egid[id + 1];
+        // Create workers temporary file
+        const auto snapshot_path = snapshot.GetPath().string() + "_edge_part_" + std::to_string(id);
```
as51340, line 312: use fmt format to avoid temporary allocations.

src/storage/v2/durability/snapshot.cpp
```diff
@@ -127,6 +139,230 @@ struct BatchInfo {
...
+      i = 1;                                   // 1 on purpose, as the first element is already in the batch
+      // Check if we have enough batches
+      if (acc.size() < batch_id * n_batches) break;
+      if (batch_id == n_batches) break;
+    }
+  }
+  return batches;
+}
+
+void MultiThreadedWorkflow(utils::SkipList<Edge> *edges, utils::SkipList<Vertex> *vertices, auto &&store_edges,
```
as51340, line 292: I think you can again split this method in half for edges and vertices. Otherwise a duplication, harder to follow and If you change something in the future you will have to change it at two places

src/storage/v2/durability/snapshot.cpp
```diff
@@ -127,6 +139,230 @@ struct BatchInfo {
...
+                           uint64_t thread_count) {
+  SafeTaskQueue tasks;
+  using task_results_t = std::vector<std::pair<SnapshotBatchRes, std::promise<bool>>>;
+
+  // Generate edge tasks
+  std::vector<int64_t> start_egid{};
+  task_results_t edge_res{};
+  if (edges != nullptr) {  // No edges skiplist <=> no properties on edges
+    start_egid = Batch(edges->access(), items_per_batch);
+    edge_res = task_results_t{start_egid.size() - 1};  // last element is an end marker
```
as51340, line 306: use () brackets for this constructor instead of initializer list

src/storage/v2/durability/snapshot.cpp
```diff
@@ -127,6 +139,230 @@ struct BatchInfo {
...
+      vertex_res[id].second.set_value(true);
+    });
+  }
+
+  const auto n_workers = std::min(thread_count, tasks.size());
+  std::vector<std::jthread> workers;
+  workers.reserve(n_workers);
+  for (int i = 0; i < n_workers; ++i) {
+    workers.emplace_back([&, i] {
+      utils::ThreadSetName("snapshot" + std::to_string(i));
```
as51340, line 352: I think some thread is already called snapshot, maybe rename it to SnapshotCreation

src/storage/v2/durability/snapshot.cpp
```diff
@@ -3923,15 +4159,25 @@ void CreateSnapshot(Storage *storage, Transaction *transaction, const std::files
...
+  auto write_mapping_to = [](auto &snapshot, auto &used_ids, auto mapping) {
+    used_ids.insert(mapping.AsUint());
+    snapshot.WriteUint(mapping.AsUint());
+  };
+
   // Store all edges.
-  if (storage->config_.salient.items.properties_on_edges) {
-    offset_edges = snapshot.GetPosition();
-    batch_start_offset = offset_edges;
+  auto store_edges = [&](int64_t start_gid, int64_t end_gid, auto &edges_snapshot) -> SnapshotBatchRes {
```
as51340, line 4169: rename edges_snapshot somehow so that it signals the type of auto

src/storage/v2/durability/snapshot.cpp
```diff
@@ -127,6 +139,230 @@ struct BatchInfo {
...
+                           auto &&store_vertices, const uint64_t items_per_batch, uint64_t &offset_edges,
+                           uint64_t &offset_vertices, Encoder &snapshot, uint64_t &edges_count,
+                           uint64_t &vertices_count, std::vector<BatchInfo> &edge_batch_infos,
+                           std::vector<BatchInfo> &vertex_batch_infos, std::unordered_set<uint64_t> &used_ids,
+                           uint64_t thread_count) {
+  SafeTaskQueue tasks;
+  using task_results_t = std::vector<std::pair<SnapshotBatchRes, std::promise<bool>>>;
+
+  // Generate edge tasks
+  std::vector<int64_t> start_egid{};
```
as51340, line 302: can you rename somehow so it signals that we are working with batches

src/storage/v2/durability/snapshot.cpp
```diff
@@ -127,6 +139,230 @@ struct BatchInfo {
...
+  task_results_t edge_res{};
+  if (edges != nullptr) {  // No edges skiplist <=> no properties on edges
+    start_egid = Batch(edges->access(), items_per_batch);
+    edge_res = task_results_t{start_egid.size() - 1};  // last element is an end marker
+    for (int i = 0; i < edge_res.size(); ++i) {
+      tasks.AddTask([&, id = i] {
```
as51340, line 308: instead of having both id and id, for loop can use local iterator variable id

src/storage/v2/durability/snapshot.cpp
```diff
@@ -127,6 +139,230 @@ struct BatchInfo {
...
+    promise.get_future().wait();  // Wait for incoming result
+    spdlog::trace("Handling snapshot edge part {}, size {}, count {}...", i++, res.snapshot_size, res.count);
+    if (res.snapshot_size > 0) {
+      edges_count += res.count;
+      used_ids.merge(std::move(res.used_ids));
+      const auto current_offset = snapshot.GetPosition();
+      // Update batch positions
+      for (auto &[offset, _] : res.batch_info) {
+        offset += current_offset;
+      }
```
as51340, line 203: use std::accumulate

src/storage/v2/durability/snapshot.cpp
```diff
@@ -127,6 +139,230 @@ struct BatchInfo {
...
+    if (res.snapshot_size > 0) {
+      vertices_count += res.count;
+      used_ids.merge(std::move(res.used_ids));
+      const auto current_offset = snapshot.GetPosition();
+      // Update batch positions
+      for (auto &[offset, _] : res.batch_info) {
+        offset += current_offset;
+      }
+      // TODO Regenerate batches (currently could be less then optimal)
+      vertex_batch_infos.insert(vertex_batch_infos.end(), res.batch_info.begin(), res.batch_info.end());
```
as51340, line 242: I think you can move into vertex_batch_infos

src/storage/v2/durability/snapshot.cpp
```diff
@@ -127,6 +139,230 @@ struct BatchInfo {
...
+        close(edges_fd);
+        std::error_code ec;
+        std::filesystem::remove(res.snapshot_path, ec);
+        if (ec) spdlog::warn("Couldn't remove temporary edge snapshot part {}: {}", res.snapshot_path, ec.message());
+      }};
+      // Use sendfile for efficient copying (zero-copy)
+      off_t offset = 0;
+      const auto size = res.snapshot_size - offset;
+      ssize_t bytes_sent = sendfile(snapshot.native_handle(), edges_fd, &offset, size);
+      if (bytes_sent == -1 || bytes_sent != size) {
```
as51340, line 221: Maybe makes sense to abstract this part into some utils method or something similar?

as51340 (changes requested):

Ignition (dismissed):

src/storage/v2/durability/snapshot.cpp
Ignition: @as51340 I can give you good reasons to prefer {} over ()
- avoid accidental implicit conversions

src/storage/v2/durability/snapshot.cpp
```diff
@@ -127,6 +139,230 @@ struct BatchInfo {
   uint64_t count;
 };
 
+constexpr auto kEnd = std::numeric_limits<int64_t>::max();
+
+struct SnapshotBatchRes {
+  std::vector<BatchInfo> batch_info{};
```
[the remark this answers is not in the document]
andrejtonev: I think this is fine even in shared access.
1. Batching over GIDs
2. Batches don't need to be explicitly the same size. This is just to help split up the data.
3. Since we are using snapshot isolation, we shouldn't insert any new elements anyway
4. In case GC unlinks/removes elements, we are scanning from first equal or greater then the start GID and stopping one the read GID is larger or equal to then the end GID. Again, some threads might be utilized less then ideally, but should all be correct still. 
5. Worst case, some thread scan for longer then ideal. Last thread is scanning till the end (GID = max)

as51340:
also, I read that std::hardware_concurrency can return 0 on some machines and that it makes sense to have "back-up default". This is example from the c++ concurrency in action:
```
auto max_threads = user flag thread count
auto hardware_threads = std::thread::hardware_concurrency();
auto num_threads = std::min(hardware_threads != 0 ? hardware_threads : 2, max_threads);
```

src/flags/general.cpp
```diff
@@ -112,6 +112,16 @@ DEFINE_VALIDATED_bool(
 DEFINE_bool(storage_parallel_schema_recovery, false,
             "Controls whether the indices and constraints creation can be done in a multithreaded fashion.");
 
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
+DEFINE_bool(storage_parallel_snapshot_creation, false,
+            "If true, snapshots will be created using --storage-snapshot-thread-count number of treads.");
+
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
+DEFINE_uint64(storage_snapshot_thread_count,
```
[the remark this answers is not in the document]
andrejtonev: That is just the default value. I copied from the other flag. 
As most people just leave the default, I would rather use the number of threads as the default.

src/storage/v2/config.hpp
```diff
@@ -124,9 +124,11 @@ struct Config {
     bool restore_replication_state_on_startup{false};  // PER INSTANCE
 
     uint64_t items_per_batch{1'000'000};  // PER DATABASE
+    uint64_t snapshot_thread_count{8};    // PER INSTANCE SYSTEM FLAG
     uint64_t recovery_thread_count{8};    // PER INSTANCE SYSTEM FLAG
 
-    bool allow_parallel_schema_creation{false};  // PER DATABASE
+    bool allow_parallel_snapshot_creation{false};  // PER DATABASE
```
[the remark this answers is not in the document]
andrejtonev: Again copied as it was.
I would leave it as is, comes back to user-friendliness. Most of the time, users will want turn something on/off, then fine-tune if needed.

src/storage/v2/durability/serialization.hpp
```diff
@@ -44,6 +44,7 @@ class BaseEncoder {
 /// Encoder that is used to generate a snapshot/WAL.
 class Encoder final : public BaseEncoder {
  public:
+  void Initialize(const std::filesystem::path &path);
```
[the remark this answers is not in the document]
andrejtonev: No, this is a version of the encoder without the magic and version header. Using it for the temp files.

src/storage/v2/durability/snapshot.cpp
```diff
@@ -127,6 +139,230 @@ struct BatchInfo {
...
+  std::unordered_set<uint64_t> used_ids{};
```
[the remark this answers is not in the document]
andrejtonev: This is a shared structure. Added comments to clear it up

src/storage/v2/durability/snapshot.cpp
```diff
@@ -127,6 +139,230 @@ struct BatchInfo {
...
+  uint64_t count{0};
+  std::filesystem::path snapshot_path{};
+  size_t snapshot_size{0};
```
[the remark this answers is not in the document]
andrejtonev: Added comments

src/storage/v2/durability/snapshot.cpp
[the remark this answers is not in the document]
andrejtonev: Generalized function and calling it twice. Once for edges, once for vertices.

src/storage/v2/durability/snapshot.cpp
andrejtonev: it's not accumulating, it is applying an offset. gonna leave it as is.

src/storage/v2/durability/snapshot.cpp
```diff
@@ -127,6 +139,230 @@ struct BatchInfo {
...
+    // Single batch, nothing to do
+    return std::vector<int64_t>{0, kEnd};
+  }
+
+  std::vector<int64_t> batches(n_batches + 1, kEnd);  // start and end gids
+  batches[0] = 0;                                     // Always start from the lowest possible
+  int i = 0;
+  int batch_id = 1;
+  for (const auto &elem : acc) {
+    if (step == i++) {
```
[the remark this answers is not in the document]
andrejtonev: Gonna leave it as is

src/storage/v2/durability/snapshot.cpp
```diff
@@ -127,6 +139,230 @@ struct BatchInfo {
...
+      batches[batch_id++] = elem.gid.AsInt();  // This batch's start ID and previous batch's end ID
+      i = 1;                                   // 1 on purpose, as the first element is already in the batch
+      // Check if we have enough batches
+      if (acc.size() < batch_id * n_batches) break;
```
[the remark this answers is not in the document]
andrejtonev: Leftover from previous version. Removed it.

src/storage/v2/durability/snapshot.cpp
```diff
@@ -127,6 +139,230 @@ struct BatchInfo {
...
+      if (batch_id == n_batches) break;
```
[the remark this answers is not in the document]
andrejtonev: This just finishes earlier. No need to scan the last batch

src/storage/v2/durability/snapshot.cpp
andrejtonev: Will leave it under one function. This is the whole work that needs to be done. Simplifies the lifetimes and makes sure everything has been processed before we return.

src/storage/v2/durability/snapshot.cpp
andrejtonev: Don't see any thread being called snapshot.
Also our implementation requires a short name.

src/storage/v2/durability/snapshot.cpp
andrejtonev: Just renamed i to id. Need to copy into the lambda, so just removed the renaming

src/storage/v2/durability/snapshot.cpp
```diff
@@ -11,6 +11,17 @@
 
 #include "storage/v2/durability/snapshot.hpp"
 
+#include <openssl/x509v3.h>
+#include <sys/sendfile.h>
+#include <sys/stat.h>
```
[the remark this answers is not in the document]
andrejtonev: I'll see if I can remove some, but most of them yes

PR #2846, opened 2025-03-31
Bugfix: Concurrent snapshot creation and demotion cause replica to crash on restart

as51340:
When main becomes replica, snapshot scheduler is paused and all snapshot locks are taken. When replica gets promoted, abort is set to false and scheduler resumed. On demotion, if there is a snapshot currently getting created, it will get aborted.

### DIFF
--- a/src/replication_handler/replication_handler.cpp
+++ b/src/replication_handler/replication_handler.cpp
@@ -15,12 +15,15 @@
 #include "replication/replication_client.hpp"
 #include "replication_handler/system_replication.hpp"
 #include "replication_query_handler.hpp"
+#include "storage/v2/inmemory/storage.hpp"
 #include "utils/functional.hpp"
 #include "utils/synchronized.hpp"
 
 #include <spdlog/spdlog.h>
 
 namespace memgraph::replication {
+
+using namespace std::chrono_literals;
 
 namespace {
 #ifdef MG_ENTERPRISE
@@ -196,6 +199,21 @@ bool ReplicationHandler::SetReplicationRoleReplica(const ReplicationServerConfig
                                                    const std::optional<utils::UUID> &main_uuid) {
   try {
     auto locked_repl_state = repl_state_.TryLock();
+
+    bool snapshot_locks_taken{true};
+    dbms_handler_.ForEach([&snapshot_locks_taken](dbms::DatabaseAccess db_acc) {
+      auto *storage = static_cast<storage::InMemoryStorage *>(db_acc->storage());
+      storage->abort_snapshot_.store(true, std::memory_order_release);
+      if (!storage->snapshot_lock_.try_lock_for(10s)) {
+        snapshot_locks_taken = false;
+        return;
+      }
+    });
+
+    if (!snapshot_locks_taken) {
+      return false;
+    }
+
     return SetReplicationRoleReplica_<true>(locked_repl_state, config, main_uuid);
   } catch (const utils::TryLockException & /* unused */) {
     return false;
--- a/src/replication_handler/replication_handler.cpp
+++ b/src/replication_handler/replication_handler.cpp
@@ -206,35 +206,30 @@ bool ReplicationHandler::SetReplicationRoleReplica(const ReplicationServerConfig
       storage->snapshot_runner_.Pause();
     });
 
-    bool snapshot_locks_taken{true};
-    dbms_handler_.ForEach([&snapshot_locks_taken, timer = utils::Timer()](dbms::DatabaseAccess db_acc) {
-      if (!snapshot_locks_taken) {
-        return;
-      }
-      auto *storage = static_cast<storage::InMemoryStorage *>(db_acc->storage());
-      while (true) {
-        if (timer.Elapsed() > 10s) {
-          spdlog::error("Failed to take snapshot locks while demoting to replica");
-          snapshot_locks_taken = false;
-          // Exit from handling db
-          return;
-        }
-        if (storage->snapshot_lock_.try_lock()) {
-          // Exit from handling db
-          return;
-        }
-      }
-    });
+    std::vector<std::unique_lock<std::mutex>> snapshot_locks;
 
-    utils::OnScopeExit const unlock_snapshots{[this]() {
-      dbms_handler_.ForEach([](dbms::DatabaseAccess db_acc) {
+    std::unordered_map<std::string, bool> locks_metadata;
+    auto const timer = utils::Timer();
+    while (true) {
+      if (timer.Elapsed() > 10s) {
+        spdlog::error("Failed to take snapshot locks while demoting to replica");
+        return false;
+      }
+      dbms_handler_.ForEach([&locks_metadata, &snapshot_locks](dbms::DatabaseAccess db_acc) {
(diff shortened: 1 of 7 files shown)


as51340:
### Tracking


### Standard development
- [ ] Update unit/E2E tests


### CI Testing Labels


### Documentation checklist
    - **[ Release note text ]**

src/replication_handler/replication_handler.cpp
```diff
@@ -196,6 +199,21 @@ bool ReplicationHandler::SetReplicationRoleReplica(const ReplicationServerConfig
                                                    const std::optional<utils::UUID> &main_uuid) {
   try {
     auto locked_repl_state = repl_state_.TryLock();
+
+    bool snapshot_locks_taken{true};
+    dbms_handler_.ForEach([&snapshot_locks_taken](dbms::DatabaseAccess db_acc) {
+      auto *storage = static_cast<storage::InMemoryStorage *>(db_acc->storage());
+      storage->abort_snapshot_.store(true, std::memory_order_release);
+      if (!storage->snapshot_lock_.try_lock_for(10s)) {
```
andrejtonev, line 207: Maybe split this in two. First set the abort for everyone and then try lock

andrejtonev: Also don't wait 10s per lock, but for the whole process. In case there are many tenants, this could block for a really long time.
You could just loop and try lock or try lock for a short period and then fail after 10s.

src/storage/v2/durability/snapshot.cpp
```diff
@@ -3926,12 +3941,18 @@ void CreateSnapshot(Storage *storage, Transaction *transaction, const std::files
...
   // Store all edges.
   if (storage->config_.salient.items.properties_on_edges) {
     offset_edges = snapshot.GetPosition();
     batch_start_offset = offset_edges;
     auto acc = edges->access();
+
     for (auto &edge : acc) {
+      if (snapshot_aborted()) {
+        return;
+      }
```
andrejtonev, line 3954: Need to measure the impact on performance

as51340: I added timer specifically so that atomic isn't check all the time but probably some degradation will occur. You have an idea how to check it? Do we benchmark it?

andrejtonev: Lets just benchmark it with a decently sized dataset.

as51340: Measured it for 5mil nodes and 10mil edges. Basicaly the same. Recovering from snapshot before and after take ~19s.

src/replication_handler/replication_handler.cpp
```diff
@@ -196,6 +199,21 @@ bool ReplicationHandler::SetReplicationRoleReplica(const ReplicationServerConfig
                                                    const std::optional<utils::UUID> &main_uuid) {
   try {
     auto locked_repl_state = repl_state_.TryLock();
+
+    bool snapshot_locks_taken{true};
+    dbms_handler_.ForEach([&snapshot_locks_taken](dbms::DatabaseAccess db_acc) {
+      auto *storage = static_cast<storage::InMemoryStorage *>(db_acc->storage());
+      storage->abort_snapshot_.store(true, std::memory_order_release);
```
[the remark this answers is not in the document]
andrejtonev: Pause and then abort. That way there shouldn't be a possibility to abort and right a way start the next one.

src/replication_handler/replication_handler.cpp
```diff
@@ -196,6 +199,44 @@ bool ReplicationHandler::SetReplicationRoleReplica(const ReplicationServerConfig
...
     auto locked_repl_state = repl_state_.TryLock();
+
+    dbms_handler_.ForEach([](dbms::DatabaseAccess db_acc) {
+      auto *storage = static_cast<storage::InMemoryStorage *>(db_acc->storage());
+      storage->abort_snapshot_.store(true, std::memory_order_release);
+      storage->snapshot_runner_.Pause();
+    });
+
+    bool snapshot_locks_taken{true};
+    dbms_handler_.ForEach([&snapshot_locks_taken, timer = utils::Timer()](dbms::DatabaseAccess db_acc) {
```
andrejtonev, line 210: The ForEach is not done in parallel. Put the while loop on the outside then foreach try lock until you have them all or timeout.

andrejtonev: Is it possible to add/drop a db in the middle of this or is writing forbidden?

src/replication_handler/replication_handler.cpp
```diff
@@ -196,6 +199,44 @@ bool ReplicationHandler::SetReplicationRoleReplica(const ReplicationServerConfig
...
+          // Exit from handling db
+          return;
+        }
+      }
+    });
+
+    utils::OnScopeExit const unlock_snapshots{[this]() {
+      dbms_handler_.ForEach([](dbms::DatabaseAccess db_acc) {
+        auto *storage = static_cast<storage::InMemoryStorage *>(db_acc->storage());
+        storage->snapshot_lock_.unlock();
```
andrejtonev, line 232: You don't know here who has the lock. The snapshot_lock should have raii guard.

src/replication_handler/replication_handler.cpp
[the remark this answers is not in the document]
as51340: but if pause the scheduler thread, the snapshot that is being currently created cannot be stopped? I think the correct thing is to check whether abort_snapshot is true when starting snapshot creation and abort it if needed

src/replication_handler/replication_handler.cpp
as51340: we take unique lock when creating new DB and shared lock when doing ForEach so should be thread safe

as51340: yes at the moment. Opened the issue, we need to tackle it at some point

src/replication_handler/replication_handler.cpp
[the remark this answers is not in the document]
as51340: true. I messed up in my head, changed the order now

src/storage/v2/inmemory/storage.cpp
```diff
@@ -2553,17 +2557,17 @@ utils::BasicResult<InMemoryStorage::CreateSnapshotError> InMemoryStorage::Create
...
 
-  // TODO: Think about role check here
-  // What will happen if MAIN is being demoted at this point?
-
   utils::Timer timer;
   Transaction *transaction = accessor->GetTransaction();
   auto const &epoch = repl_storage_state_.epoch_;
   durability::CreateSnapshot(this, transaction, recovery_.snapshot_directory_, recovery_.wal_directory_, &vertices_,
-                             &edges_, uuid(), epoch, repl_storage_state_.history, &file_retainer_);
+                             &edges_, uuid(), epoch, repl_storage_state_.history, &file_retainer_, &abort_snapshot_);
```
andrejtonev, line 2564: Add a return value or use the abort_snapshot to return the correct status, which is `CreateSnapshotError::AbortSnapshot;`

andrejtonev (approved):

PR #2894, opened 2025-04-14
Add metrics for stress tests

Josipmrden:


tests/stress/metrics_metrics-monitoring.json
mattkjames7, line 1: Does this need deleting?

.github/workflows/daily_stress.yaml
```diff
@@ -0,0 +1,104 @@
+name: Daily Stress Tests
+
+on:
+  workflow_dispatch:
+  push:
```
mattkjames7, line 7: please remove this `push:` before merging or we'll end up with a stress test for every single push :joy:

.github/workflows/daily_stress.yaml
```diff
@@ -0,0 +1,104 @@
+name: Daily Stress Tests
+
```
mattkjames7, line 5: would it be worth adding some concurrency limits on this workflow, given that we are allowing a manual workflow_dispatch?

tests/stress/README.md
```diff
@@ -78,18 +78,28 @@ customWorkloads:
...
           # (Optional) Step number for phased executions. If nothing is specified, worker will have step of value (1), which means
           # it will be executed first. For phased execution, you can specify different non-negative integer numbers. Each step
           # will execute the set of workers that apply to the step one after other.
+          metrics: <list> # Metrics that will be displayed
+          # Values for metrics:
+          # "duration" -> will display how long the worker took to finish
+          # "throughput" -> will display the throughput in amount of queries per second after the execution is done
       timeout_min: <int>
       # Maximum execution time for the workload in minutes. Failing to execute the workload in this amount of minutes
       # will result in a failure of the stress test
```
mattkjames7, line 104: Not sure if this is too far out of the scope of this PR, but the example for running the stress test (below) uses a script which I don't think exists...

tests/stress/metrics_metrics-monitoring.json
Josipmrden: yes, good catch

.github/workflows/daily_stress.yaml
Josipmrden: yes definitely, will remember

Josipmrden:
@mattkjames7 will actually update the metrics worker to push only metrics requested by user, so that we don't overflow the DB with unnecessary things which we don't track

Josipmrden:
Closed for being stale

PR #2900, opened 2025-04-15
Fix release build tests

mattkjames7:
Make daily builds run the large stress test and release Jepsen test.
- Jepsen was being skipped because the input OS was being set to `ubuntu-24.04` - this input is now ignored
- Large stress test was being skipped because `run_stress_large` was not set in the `TestBuild` job, this is now fixed
- removed `cron` job for stress test large, as this will take it's place

The downside of this is that daily builds will now take 8-9 hours before triggering MAGE builds; the upside is that the previously skipped tests will be included in the aggregated test result and therefore count towards the overall result displayed on the daily build page.

### DIFF
--- a/.github/workflows/reusable_release_tests.yaml
+++ b/.github/workflows/reusable_release_tests.yaml
@@ -764,7 +764,7 @@ jobs:
 
   release_jepsen_test:
     continue-on-error: true
-    if: ${{ inputs.run_release_tests == 'true' && inputs.os == 'debian-12' }}
+    if: ${{ inputs.run_release_tests == 'true' }}
     name: "Release Jepsen Test"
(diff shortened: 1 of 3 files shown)


.github/workflows/reusable_release_tests.yaml
```diff
@@ -785,15 +785,15 @@ jobs:
         run: |
           ./release/package/mgbuild.sh \
           --toolchain $TOOLCHAIN \
-          --os $OS \
+          --os debian-12 \
```
gitbuda, line 788: is it possible to have a job-specific variable? 🤔

gitbuda (approved):
One small comment, but ready to be merged 😄

PR #2910, opened 2025-04-17
Improve perf around snapshot

Ignition:
Main work was to optimise the snapshot processing

Results:
10 million nodes + 10 million edges + 1 label property index
running `CREATE SNAPSHOT` in release mode
master:  avg 5.50 seconds
this PR:  avg 3.57 seconds

- Feature: Already running 
  - Don't block is a snapshot is already in progress 
- Feature: Snapshot unnecessary
  - If digest is identical to previous snapshot, avoid work  
- Optimise Snapshot
  - Early exit for vertices+edges processing, if object is new after the start of creating the snapshot
  - Hot code, sample snapshot_aborted less often, avoid cost of checking the current time
  - NonConcurrentOutputFile: No need for concurrent access to internal buffer, hence cheaper version of OutputFile without need for locks

Extra optimisations:
- Optimise AllVerticesIterable::Iterator
  - Common case: everything is visible
  - Avoid unnecessary writes to memory  
- Optimise ScopedProfile
  - Common case: not profiling, use optional to make it cheaper to ctr/dtr
- Optimise MemoryTracker
  - Inline the ++/-- of counter 
- Optimise OutputFile (impact WALs)
  - Fewer memory writes (use local variables) 
  - Better atomic load/store memory orders
  - Fewer lock+unlock
  - Cheaper lock, RWSpinLock
  - Better memory alignment, avoid false sharing
- Optimise In/OutEdges
  - Early exit if no edges
  - [[unlikely]] branch annotations

### DIFF
--- a/src/storage/v2/durability/snapshot.cpp
+++ b/src/storage/v2/durability/snapshot.cpp
@@ -4682,7 +4682,6 @@ bool CreateSnapshot(Storage *storage, Transaction *transaction, const std::files
   spdlog::info("Starting snapshot creation to {}", path);
   SnapshotEncoder snapshot;
   snapshot.Initialize(path, kSnapshotMagic, kVersion);
-  // TODO: snapshot is single threaded write...yet internally has a flush_lock_ protecting its buffer
 
   // Write placeholder offsets.
   uint64_t offset_offsets = 0;
--- a/src/storage/v2/durability/snapshot.cpp
+++ b/src/storage/v2/durability/snapshot.cpp
@@ -4678,6 +4678,7 @@ bool CreateSnapshot(Storage *storage, Transaction *transaction, const std::files
   spdlog::info("Starting snapshot creation to {}", path);
   Encoder snapshot;
   snapshot.Initialize(path, kSnapshotMagic, kVersion);
+  // TODO: snapshot is single threaded write...yet internally has a flush_lock_ protecting its buffer
 
   // Write placeholder offsets.
   uint64_t offset_offsets = 0;
--- a/src/storage/v2/durability/snapshot.cpp
+++ b/src/storage/v2/durability/snapshot.cpp
@@ -147,6 +147,7 @@ struct BatchInfo {
 
 using SnapshotEncoder = Encoder<utils::NonConcurrentOutputFile>;
 
+// Used for an upper limit, ie. something beyond the last item we should be processing
 constexpr auto kEnd = std::numeric_limits<int64_t>::max();
 
 // Result of a partial snapshot creation
@@ -258,7 +259,7 @@ auto Batch(auto &&acc, const uint64_t items_per_batch) {
       batches[batch_id++] = elem.gid.AsInt();  // This batch's start ID and previous batch's end ID
       i = 1;                                   // 1 on purpose, as the first element is already in the batch
       // Check if we have enough batches
-      if (batch_id == n_batches) break;
+      if (batch_id == n_batches) break;  // THIS IS IMPORTANT, we do not want to remove the kEnd as our upper limit
     }
   }
   return batches;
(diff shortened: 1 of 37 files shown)


Ignition:
### Tracking


### Standard development
- [ ] Update unit/E2E tests


### CI Testing Labels


### Documentation checklist
    - **[ Release note text ]**

src/storage/v2/inmemory/replication/recovery.cpp
```diff
@@ -201,7 +202,7 @@ std::optional<std::vector<RecoveryStep>> GetRecoverySteps(uint64_t replica_commi
   }
 
   // If we have a current wal file we need to use it
-  if (current_wal_seq_num) {
+  if (last_durable_timestamp < current_wal_timestamp && current_wal_seq_num) {
```
Ignition, line 205: @andrejtonev + @as51340 can you double check this.
Its to make sure the RecoverySteps does not include current WAL if the latest snapshot is already covers what is inside that WAL.

as51340: will think about it. Just run jepsen 2-3 times if you change something like this so we are more certain it doesn't break anything.

colinbarry (approved):
I've reviewed your changes and don't have any significant issues to raise.

src/storage/v2/durability/snapshot.cpp
```diff
@@ -256,15 +259,15 @@ auto Batch(auto &&acc, const uint64_t items_per_batch) {
       batches[batch_id++] = elem.gid.AsInt();  // This batch's start ID and previous batch's end ID
       i = 1;                                   // 1 on purpose, as the first element is already in the batch
       // Check if we have enough batches
-      if (batch_id == n_batches) break;
+      if (batch_id == n_batches) break;  // THIS IS IMPORTANT, we do not want to remove the kEnd as our upper limit
```
andrejtonev, line 262: Writing here just because it's changed line.
Have you tried using the sampling iterator instead of the forward one when scanning though the skiplist?

src/storage/v2/inmemory/storage.cpp
```diff
@@ -2593,7 +2600,10 @@ utils::BasicResult<InMemoryStorage::CreateSnapshotError> InMemoryStorage::Create
     return CreateSnapshotError::AbortSnapshot;
   }
 
-  std::lock_guard snapshot_guard(snapshot_lock_);
+  auto snapshot_guard = std::unique_lock(snapshot_lock_, std::try_to_lock);
+  if (!snapshot_guard.owns_lock()) {
```
andrejtonev, line 2604: Other places can lock this. Not sure this is the correct check

src/storage/v2/inmemory/replication/recovery.cpp
andrejtonev: The WAL is skipped, so this should be fine.
```
    // If WAL file doesn't contain any changes that need to be applied, ignore it
    if (wal_info.to_timestamp <= storage->repl_storage_state_.last_durable_timestamp_.load(std::memory_order_acquire)) {
      spdlog::trace("WAL file won't be applied since all changes already exist.");
      return {true, 0};
    }
```

andrejtonev: @as51340 
How do we update the epoch with HA?
In the current wal handler we update the epoch only if the wal's ts is higher then the local ldt.
Will we miss some role changes?

src/utils/file.hpp
```diff
@@ -256,22 +256,117 @@ class OutputFile {
...
-  void FlushBuffer(bool force_flush);
+  void FlushBuffer();
   void FlushBufferInternal();
+  void FlushBufferInternal(size_t to_write);
 
   size_t SeekFile(Position position, ssize_t offset);
 
-  int fd_{-1};
+  // put flush lock on its own cacheline
+  alignas(64) utils::RWSpinLock flush_lock_{};
```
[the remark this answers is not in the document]
Ignition: It made an observable difference

src/storage/v2/durability/snapshot.cpp
Ignition: I didn't do that yet, will get that done in an upcoming PR

src/storage/v2/inmemory/storage.cpp
Ignition: You are right, show snapshots + replication... @imilinovic we will need to review this before merge.

src/storage/v2/inmemory/replication/recovery.cpp
as51340: let's talk tomorrow about this

src/storage/v2/inmemory/storage.cpp
Ignition: Made a fix...we made it an orthogonal concern, using a separate atomic state rather than use this lock.

as51340 (changes requested):
Run jepsen

Ignition:
Merge if https://github.com/memgraph/memgraph/actions/runs/14666297744 is ok

PR #2918, opened 2025-04-22
Remove cost of vector in label+property index for non-composite indices [prototype]

colinbarry:
This is a prototype for optimising in-memory label property indices. Since replacing the single `PropertyId` for a `vector<PropertyId>`, every index - even on a single ("non-composite") index - pays the cost of the vector. This vector adds the burden of an additional allocation and deallocation.

This is the first stage in replacing the index skiplist with a variant, which will (eventually) have alternatives for various fixed-sized arrays, as well as a catch-all `vector` implementations for any indexes which have
more than n properties (for some value of n).

### DIFF
--- a/src/storage/v2/indices/label_property_index.cpp
+++ b/src/storage/v2/indices/label_property_index.cpp
@@ -64,7 +64,7 @@ auto PropertiesPermutationHelper::Extract(PropertyStore const &properties) const
 }
 
 auto PropertiesPermutationHelper::ApplyPermutation(std::vector<PropertyValue> values) const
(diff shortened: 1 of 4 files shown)


colinbarry:
This has diverged very drastically from master since we've added non-blocking index creation. 

I'm unsure how relevant this PR is in the scope of the improved index creation, and I suspect this PR can be closed. In the interim, I will leave this PR open until I get a chance to investigate, but have removed the 3.5 milestone.

colinbarry:
Closing: fixed by https://github.com/memgraph/memgraph/pull/4308

PR #2920, opened 2025-04-23
Improve ResourceLock

imilinovic:
- Refactored ResourceLock to maintain a common pattern
- Make ResourceLock prioritise READ_ONLY over WRITE
- Allow downgrading the lock from WRITE or READ_ONLY, to READ

### DIFF
--- a/src/utils/resource_lock.hpp
+++ b/src/utils/resource_lock.hpp
@@ -149,7 +149,7 @@ struct ResourceLock {
   }
 
   template <LockReq Req>
-  bool dowgrade_to_read() {
+  bool downgrade_to_read() {
     auto lock = std::unique_lock{mtx};
     if (state != SHARED) return false;
     unlock_state_updater<Req>();
@@ -294,14 +294,14 @@ struct SharedResourceLockGuard {
     if (ptr_ && locked_) {
       switch (type_) {
         case WRITE: {
-          auto res = ptr_->dowgrade_to_read<ResourceLock::LockReq::WRITE>();
+          auto res = ptr_->downgrade_to_read<ResourceLock::LockReq::WRITE>();
           if (res) type_ = READ;
           return res;
         }
         case READ:
           return true;  // can't downgrade from read to read
         case READ_ONLY: {
-          auto res = ptr_->dowgrade_to_read<ResourceLock::LockReq::READ_ONLY>();
+          auto res = ptr_->downgrade_to_read<ResourceLock::LockReq::READ_ONLY>();
           if (res) type_ = READ;
           return res;
         }
--- a/src/utils/resource_lock.hpp
+++ b/src/utils/resource_lock.hpp
@@ -120,6 +120,15 @@ struct ResourceLock {
     return false;
   }
 
+  template <LockReq Req>
+  bool dowgrade_to_read() {
+    auto lock = std::unique_lock{mtx};
+    if (state == UNIQUE) return false;
+    unlock_state_updater<Req>();
+    lock_pre_state_change<LockReq::READ>();
+    return true;
+  }
+
   template <typename Rep, typename Period, LockReq Req = LockReq::WRITE>
   bool try_lock_shared_for(std::chrono::duration<Rep, Period> const &time) {
     auto lock = std::unique_lock{mtx};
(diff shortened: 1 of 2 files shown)


src/utils/resource_lock.hpp
```diff
@@ -60,123 +64,103 @@ struct ResourceLock {
...
 
-  template <bool Read = false>
+  template <LockReq Req = LockReq::WRITE>
   bool try_lock_shared() {
     auto lock = std::unique_lock{mtx};
-    if (state != UNIQUE && ro_count == 0) {
+    if (state != UNIQUE && lock_guard_condition<Req>()) {
       state = SHARED;
-      ++w_count;
+      lock_state_updater<Req>();
```
andrejtonev, line 117: Are `lock_pre_state_change` and `lock_post_state_change` missing on purpose?

src/utils/resource_lock.hpp
```diff
@@ -60,123 +64,103 @@ struct ResourceLock {
...
-  }
-
-  template <typename Rep, typename Period, bool Read = false>
-  requires(!Read) bool try_lock_shared_for(std::chrono::duration<Rep, Period> const &time) {
-    auto lock = std::unique_lock{mtx};
-    // block until available
-    if (!cv.wait_for(lock, time, [this] { return state != UNIQUE && ro_count == 0; })) return false;
-    state = SHARED;
-    ++w_count;
+    if (state == UNIQUE) return false;
```
andrejtonev, line 126: Would prefer at least a `DMG_ASSERT` to make sure nobody is trying to downgrade an unlocked state or from READ

src/utils/resource_lock.hpp
```diff
@@ -60,123 +64,103 @@ struct ResourceLock {
...
+    unlock_state_updater<Req>();
+    lock_pre_state_change<LockReq::READ>();
```
andrejtonev, line 128: Not sure why we call `lock_pre_state_change` and not post

src/utils/resource_lock.hpp
```diff
@@ -60,123 +64,103 @@ struct ResourceLock {
...
+  template <> void lock_pre_state_change<LockReq::READ_ONLY>(){ ++ro_pending_count; }
+
+  template <LockReq Req> void lock_post_state_change(){}
+  template <> void lock_post_state_change<LockReq::READ_ONLY>(){ --ro_pending_count; }
+
+  enum class NotifyKind : uint8_t { None, One, All };
+  template <LockReq Req> NotifyKind unlock_should_notify() const;
+  template <> NotifyKind unlock_should_notify<LockReq::WRITE>() const { return w_count == 0 ? NotifyKind::All : NotifyKind::None; }
+  template <> NotifyKind unlock_should_notify<LockReq::READ>() const { return (r_count == 0 && w_count == 0 && ro_count == 0) ? NotifyKind::One : NotifyKind::None; }
+  template <> NotifyKind unlock_should_notify<LockReq::READ_ONLY>() const { return ro_count == 0 ? NotifyKind::All : NotifyKind::None; }
```
andrejtonev, line 98: Please add comments explaining why All, One or None

src/utils/resource_lock.hpp
```diff
@@ -60,123 +64,103 @@ struct ResourceLock {
...
-  void lock_shared() {
-    auto lock = std::unique_lock{mtx};
-    // block until available
-    cv.wait(lock, [this] { return state != UNIQUE && ro_count == 0; });
-    state = SHARED;
-    ++w_count;
-  }
+  // clang-format off
+  template <LockReq Req> bool lock_guard_condition() const;
+  template <> bool lock_guard_condition<LockReq::WRITE>() const     { return ro_count == 0 && ro_pending_count == 0; }
```
andrejtonev, line 69: Since WRITE is also blocked by ro_pending_count, I think you are missing a notification in case a READ_ONLY fails to lock.
So try_ro_for(/*some long time*/) does a ++ on ro_pending_count.
In the meantime, try_w_for() comes and blocks because ro_pending != 0.
First lock timesout, but the second one never receives a notification even though it could now lock because ro_pending is 0.

andrejtonev: Didn't test if that is the case, but seems like it to me. Please verify.

andrejtonev (commented):
Nice work, but a couple of details to go through

src/utils/resource_lock.hpp
Ignition: Changed to `state != SHARED`

src/utils/resource_lock.hpp
Ignition: Good catch...it should have been `lock_state_updater<LockReq::READ>();`

it was auto complete filling in the wrong function name.

src/utils/resource_lock.hpp
Ignition: Yes, we don't wait on any condition variable/give up the lock, hence no need to communicate the `ro_pending_count` change.

src/utils/resource_lock.hpp
Ignition: Fixed in new commit + test

colinbarry (approved):
All seems correct to me.

PR #2937, opened 2025-05-02
Bugfix: Improve the usage of RMW pattern in HA coordinators

as51340:
Removes unused atomic variables and fixes some of the concurrency bugs noticed in HA stack. The operation`compare_exchange_strong` in coordinator_instance.cpp now uses acquire-release memory order semantics on success and failure operations. The variable `start_idx` now uses also acquire-release semantics instead of sequentially consistent operations. The lock `last_snapshot_lock` wasn't used and atomic variable `next_idx` from coordinator_log_store.

### DIFF
--- a/src/coordination/coordinator_log_store.cpp
+++ b/src/coordination/coordinator_log_store.cpp
@@ -64,15 +64,16 @@ bool CoordinatorLogStore::HandleVersionMigration(LogStoreVersion const stored_ve
       }
 
       uint64_t const last_log_entry = std::stoull(maybe_last_log_entry.value());
-      start_idx_.store(std::stoull(maybe_start_idx.value(), std::memory_order_release);
+      auto const durable_start_idx_value = std::stoull(maybe_start_idx.value();
+      start_idx_.store(durable_start_idx_value, std::memory_order_release);
 
       // Compaction might have happened so we might be missing some logs.
-      for (auto const id : std::ranges::iota_view{start_idx_.load(std::memory_order_acquire), last_log_entry + 1}) {
+      for (auto const id : std::ranges::iota_view{durable_start_idx_value, last_log_entry + 1}) {
         auto const entry = durability_->Get(fmt::format("{}{}", kLogEntryPrefix, id));
 
         if (!entry.has_value()) {
           logger_.Log(nuraft_log_level::TRACE, fmt::format("Missing entry with id {} in range [{}:{}]", id,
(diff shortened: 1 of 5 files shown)


src/coordination/coordinator_log_store.cpp
```diff
@@ -280,19 +283,31 @@ void CoordinatorLogStore::apply_pack(uint64_t index, buffer &pack) {
...
+  auto old_start_idx = start_idx_.load(std::memory_order_acquire);
+  auto new_idx = last_log_index + 1;
+  while (
+      old_start_idx <= last_log_index &&
+      !start_idx_.compare_exchange_weak(old_start_idx, new_idx, std::memory_order_acq_rel, std::memory_order_acquire)) {
+  }
+
+  if (old_start_idx > last_log_index) {
+    return true;
+  }
```
andrejtonev, line 307: Not following this logic.

andrejtonev: If the start_idx_ is greater than the new_idx, we return. 
Does this mean another thread already did the work?

andrejtonev: Would also prefer this format
```
if (old_start_idx > last_log_index) {
  return true;
}
while (!start_idx_.compare_exchange_weak(old_start_idx, new_idx, std::memory_order_acq_rel, std::memory_order_acquire)) {
    if (old_start_idx > last_log_index) {
      return true;
    }
  }
```
Having 2 conditionals in the same line is confusing and I don't think you actually need one of the checks

as51340: yes, another thread can modify value. Will rewrite

src/coordination/coordinator_log_store.cpp
```diff
@@ -301,10 +318,8 @@ bool CoordinatorLogStore::compact(uint64_t last_log_index) {
     durability_->Delete(fmt::format("{}{}", kLogEntryPrefix, ii));
   }
 
-  if (start_idx_ <= last_log_index) {
-    start_idx_ = last_log_index + 1;
-    durability_->Put(kStartIdx, std::to_string(start_idx_.load()));
-  }
+  durability_->Put(kStartIdx, std::to_string(new_idx));
```
andrejtonev, line 321: Not sure this is correct. We are here, so the start_idx was smaller than the log_index, but nothing is preventing another thread from coming here at the same time and overwriting this value.
So a thread with a lower start idx could come along and overwrite a higher value.

PR #2941, opened 2025-05-05
Improve read-write benchmark test

DavIvek:
Added checks for the correct `vertex_count` and `edge_count` in the storage info response before proceeding with the test, ensuring that data is correctly replicated. This allows the test to fail early if the expected replication data is missing.

Additionally, the endpoints flag was previously hardcoded in `common.hpp`, but the number of database instances being tested has changed. Instead of relying on the default value, the correct value is now explicitly passed.


DavIvek:
Test [failed again](https://github.com/memgraph/memgraph/actions/runs/14880112133/job/41786036179?pr=2941) but this time the failure is due to unsuccessful replication to the async replica. Now the test reveals that we have an issue with async replication being very slow, instead of previously failing on uninformative assertions.

imilinovic (approved):

PR #2949, opened 2025-05-07
Add timeout on socket connect

as51340:
The PR adds the functionality of socket connect with timeout. The only portable way of doing this is creating a socket in a non-blocking mode, timeout check using `poll` and switching it back to a blocking mode since we all our sockets are used in blocking mode. UNIX doesn't have an API which would allow that from the start. The primary motivation for such a change is the fact that socket connect timeout can be quite high (~130s) and that it depends on the platform and on the number of SYN packets sent during connect. Two properties need to be satisfied for a system to be considered synchronous:
1. There is a known upper bound on processing delays. This includes the time needed to receive a message, perform a local computation and replying
2. There is a known upper bound on message transmission delays.

We already had timeouts on read and write message, the only thing left was a timeout on connect.

The timeout for StateCheckRpc and FrequentHeartbeatRpc is reduced to 5s.

### DIFF
--- a/src/io/network/socket.cpp
+++ b/src/io/network/socket.cpp
@@ -25,7 +25,7 @@
 #include "utils/metrics_timer.hpp"
 
 namespace {
-constexpr int timeout_ms = 5000;
+constexpr uint32_t timeout_ms = 5000;
 }  // namespace
 
 namespace memgraph::metrics {
--- a/src/io/network/socket.cpp
+++ b/src/io/network/socket.cpp
@@ -48,11 +48,7 @@ Socket::~Socket() noexcept { Close(); }
 
 void Socket::Close() noexcept {
   if (socket_ == -1) return;
-  if (close(socket_) != 0) {
-    int err_sc = errno;
-    spdlog::error("Failed to close fd for {}. Closing because 'socket_' is trying to get closed. Errno: {}",
-                  endpoint_.SocketAddress(), err_sc);
-  }
+  Close(socket_, endpoint_.SocketAddress());
   socket_ = -1;
 }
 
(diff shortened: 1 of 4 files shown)


as51340:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist

as51340:
TODO: (andi) add image from prometheus verifying that the connect is bounded

as51340:
Proof that it works 
Screenshot 2025-05-08 at 12 51 55

andrejtonev (commented):
General comment
Does it have to be so c-ish?
Why set errno and not throw? Why not use std::clock::now()? And similar comments

src/io/network/socket.cpp
```diff
@@ -76,32 +83,104 @@ bool Socket::Connect(const Endpoint &endpoint) {
...
+        }
+
+        // Failure which we don't handle
+        if (errno != EINPROGRESS && errno != EWOULDBLOCK) {
+          spdlog::error("Failed to connect to socket with err code {}", errno);
+          return false;
+        }
+
+        auto const start_time = std::chrono::steady_clock::now();
+        auto const deadline = start_time + std::chrono::milliseconds(timeout_ms);
```
andrejtonev, line 119: have the timeout_ms constant be of type milliseconds

src/io/network/socket.cpp
```diff
@@ -76,32 +83,104 @@ bool Socket::Connect(const Endpoint &endpoint) {
...
+          // Poll status of 0 indicated timeout
+          if (poll_status == 0) {
+            errno = ETIMEDOUT;
+            return false;
+          }
+
+          // A signal occurred before any requested event
+          if (errno != EINTR) {
+            return false;
+          }
```
andrejtonev, line 158: a bit confusing.
can you put if errno is EINTR continue
or write a comment

andrejtonev (approved):
just a few nitpicking comments

PR #2976, opened 2025-05-15
Bugfix: Unreleased deltas when using HA

as51340:
When an instance becomes MAIN, `RECOVER SNAPSHOT` query is used or `SnapshotRpc` is received, IDs in range [old_storage_timestamp, ldt] are marked as finished in the commit log.

### DIFF
--- a/src/replication_handler/replication_handler.cpp
+++ b/src/replication_handler/replication_handler.cpp
@@ -294,6 +294,9 @@ bool ReplicationHandler::DoToMainPromotion(const utils::UUID &main_uuid, bool fo
       storage->timestamp_ =
           std::max(storage->timestamp_, storage->repl_storage_state_.last_durable_timestamp_.load() + 1);
       spdlog::trace("New timestamp on the MAIN is {} for the database {}.", storage->timestamp_, db_acc->name());
+
+      // Mark all up to timestamp + timestamp
+      static_cast<storage::InMemoryStorage *>(storage)->commit_log_->MarkFinishedUpToId(storage->timestamp_);
     });
 
     // STEP 4) Resume TTL
--- a/src/replication_handler/replication_handler.cpp
+++ b/src/replication_handler/replication_handler.cpp
@@ -118,7 +118,6 @@ inline std::optional<query::RegisterReplicaError> HandleRegisterReplicaStatus(
     switch (instance_client.GetError()) {
       case RegisterReplicaStatus::NOT_MAIN:
         MG_ASSERT(false, "Only main instance can register a replica!");
-        return {};
       case RegisterReplicaStatus::NAME_EXISTS:
         return query::RegisterReplicaError::NAME_EXISTS;
       case RegisterReplicaStatus::ENDPOINT_EXISTS:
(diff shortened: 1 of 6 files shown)


as51340:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist

Ignition (changes requested):
Add a unit test to `tests/unit/commit_log_v2.cpp`

src/replication_handler/replication_handler.cpp
```diff
@@ -294,6 +294,10 @@ bool ReplicationHandler::DoToMainPromotion(const utils::UUID &main_uuid, bool fo
       storage->timestamp_ =
           std::max(storage->timestamp_, storage->repl_storage_state_.last_durable_timestamp_.load() + 1);
       spdlog::trace("New timestamp on the MAIN is {} for the database {}.", storage->timestamp_, db_acc->name());
+
+      // Mark all up to timestamp + timestamp
+      static_cast<storage::InMemoryStorage *>(storage)->commit_log_->MarkFinishedUpToId(storage->timestamp_);
```
andrejtonev, line 299: A couple of notes here:
1. timestamp_ is not being protected by engine_lock_. This can lead to multiple tx having the same timestamp. If a tx is created while updating timestamp_, the timestamp_ could be updated to a smaller value  and then re-read. TODO Make sure there isn't some locking mechanism in play other than repl_state and figure out lock order if we need to lock the engine_lock
2. We do not want to mark all timestamps up to timestamp_, but just the range between timestamp_ and last_durable_timestamp_. This is because all ts smaller than  timestamp_ are active tx. We want to mark only the ones we are skipping over. Marking should be [timestamp, last_durable_timestamp_]; if ldt >= timestamp_
3. This range marking also needs to happen for RECOVER SNAPSHOT both via RPC and user query

src/replication_handler/replication_handler.cpp
```diff
@@ -118,7 +118,6 @@ inline std::optional<query::RegisterReplicaError> HandleRegisterReplicaStatus(
     switch (instance_client.GetError()) {
       case RegisterReplicaStatus::NOT_MAIN:
         MG_ASSERT(false, "Only main instance can register a replica!");
-        return {};
```
andrejtonev, line 121: this shouldn't fall through

src/replication_handler/replication_handler.cpp
```diff
@@ -289,11 +288,19 @@ bool ReplicationHandler::DoToMainPromotion(const utils::UUID &main_uuid, bool fo
       auto *storage = db_acc->storage();
       storage->repl_storage_state_.epoch_ = epoch;
 
+      // Modifying storage->timestamp_ needs to be done under the lock
+      auto lock = std::lock_guard{storage->engine_lock_};
```
andrejtonev, line 292: leave a comments that engine lock needs to be acquired after the repl state

src/storage/v2/commit_log.cpp
```diff
@@ -57,11 +57,100 @@ void CommitLog::MarkFinished(uint64_t id) {
...
+
+    // Mark all fields < end_field_idx as finished
+    for (size_t i = 0; i < end_field_idx; ++i) {
+      end_block->field[i] = std::numeric_limits<uint64_t>::max();
+    }
+  }
+
+  // Apply the mask on the end field idx
+  end_block->field[end_field_idx] |= end_mask;
+}
```
andrejtonev, line 125: needs to update the oldest active in case there are no active transactions

src/storage/v2/commit_log.cpp
```diff
@@ -17,7 +17,7 @@
 namespace memgraph::storage {
 CommitLog::CommitLog() : allocator_(utils::NewDeleteResource()) {}
 
-CommitLog::CommitLog(uint64_t oldest_active) : allocator_(utils::NewDeleteResource()) {
+CommitLog::CommitLog(uint64_t const oldest_active) : allocator_(utils::NewDeleteResource()) {
```
andrejtonev, line 20: lets reuse the range marking here as well

tests/unit/commit_log_v2.cpp
```diff
@@ -79,3 +76,76 @@ TEST(CommitLog, TrackAfterInitialId) {
     check_marking_ids(&log, i);
   }
 }
+
+TEST(CommitLog, MarkAsFinishedTxn) {
+  CommitLog log;
+  log.MarkFinished(1);
```
andrejtonev, line 82: add tests for
- oldest active
- ranges that span multiple blocks

src/replication_handler/replication_handler.cpp
andrejtonev: okay, it's not a conditional assert...
maybe you'll get some compilers complaining, but if not, it's fine

tests/unit/commit_log_v2.cpp
as51340: Added for oldest active.
Test for ranges that span multiple blocks:
- TrackAfterInitialId
- MarkWholeBlockAsFinished
- MarkRangeAcrossBlocks

andrejtonev (approved):
I'll apply this fix to snapshot in another PR

PR #2984, opened 2025-05-20
Bugfix: Fix deadlock when using ASYNC replicas by skipping recovery tasks if lock cannot be obtained

as51340:
When using ASYNC replicas, MAIN could deadlock because in its thread pool there would be a commit task holding RPC lock. Commit task would wait for other tasks to finish but these tasks wouldn't not be able to get executed because RPC lock couldn't be acquired. The current fix is to try lock on RPC lock and if this is not possible, skip the task. This shouldn't be a problem because replica will stay in RECOVERY state,

### DIFF
--- a/src/rpc/client.hpp
+++ b/src/rpc/client.hpp
@@ -318,17 +318,46 @@ class Client {
           TRequestResponse::Response::Load(&response, reader);
           return response;
         },
-        std::forward<Args>(args)...);
+        /*try_lock*/ false, std::forward<Args>(args)...);
+  }
+
+  /**
+   * Tries to obtain RPC stream by try locking RPC lock, otherwise returns std::nullopt
+   * @tparam TRequestResponse RPC type
+   * @tparam Args Type of arguments to propagate to StreamWithLoad
+   * @param args  Arguments to propagate to StreamWithLoad
+   * @return nullopt if couldn't try_lock, StreamHandler otherwise
+   */
+  template <class TRequestResponse, class... Args>
+  std::optional<StreamHandler<TRequestResponse>> TryStream(Args &&...args) {
+    try {
+      return StreamWithLoad<TRequestResponse>(
+          [](auto *reader) {
+            typename TRequestResponse::Response response;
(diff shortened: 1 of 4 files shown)


src/storage/v2/replication/replication_client.cpp
```diff
@@ -71,15 +71,28 @@ void ReplicationStorageClient::UpdateReplicaState(Storage *main_storage, Databas
...
+        std::optional<replication::HeartbeatRes> res;
+        if (hb_stream.has_value()) {
+          res.emplace(hb_stream->SendAndWait());
+        }
+        return res;
+      });
+
+  if (!maybe_heartbeat_res.has_value()) {
+    spdlog::trace("Couldn't get RPC lock while trying to UpdateReplicaState");
+    return;
```
andrejtonev, line 92: Do we enqueue another UpdateReplicaState or will the frequent check do that?

src/rpc/client.hpp
```diff
@@ -318,17 +318,46 @@ class Client {
...
   StreamHandler<TRequestResponse> StreamWithLoad(
-      std::function<typename TRequestResponse::Response(slk::Reader *)> res_load, Args &&...args) {
+      std::function<typename TRequestResponse::Response(slk::Reader *)> res_load, bool const try_lock, Args &&...args) {
     typename TRequestResponse::Request request(std::forward<Args>(args)...);
     auto req_type = TRequestResponse::Request::kType;
 
-    auto guard = std::unique_lock{mutex_};
+    auto guard = std::unique_lock{mutex_, std::defer_lock};
+    if (try_lock) {
+      if (!guard.try_lock()) {
```
andrejtonev, line 355: Perhaps a try_lock_for would be better. This way this will never succeed

src/storage/v2/replication/replication_client.cpp
as51340: frequent check should do that

as51340:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist

PR #3016, opened 2025-06-02
Replace `PropertyValue` map internals

colinbarry:
Replace `PropertyValue` `map_t` with `std::map` rather than `boost::container::flat_map`. 

boost's `flat_map` does not correctly propogate allocators in nested containers. This means that nested `PropertyValue`s created for `Delta`s are created using the incorrect allocator, and leak memory which is never reclaimed.

This temporary fix replaces the internal `map_t` with `std::map`, which *does* propogate allocators. Once we have established if a more recent `boost` version fixes the issue, we can revert this fix.


colinbarry:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
  - Fixed memory leak that occurred when queries updated deeply nested properties. Users can now safely create queries that update nested map properties without risk of eventual out-of-memory errors. [#3016](https://github.com/memgraph/memgraph/pull/3016)

PR #3025, opened 2025-06-03
Support regex queries with label+property indices

colinbarry:
Queries using a label+property index now correctly return matching results when comparing properties against regular expressions. 

The problem was caused by index code rejecting ranges across different `PropertyType`s as being invalid. This is true in most cases, (e.g. `WHERE node.a > 'hello' AND node.a < 42`) but not in the case where the upper-bound is an exclusive (`<`) bound on the next (from an ordering standpoint) `PropertyType`. As the exclusive upper bound exclusive sentinel is not compared against, it means that the range is still across a single `PropertyType`.

Regexes uses a range that is inclusively lower bounded on the empty string, and exclusively upper bounded on the minimum value of whatever `PropertyType` happens to follow string.

(This bug was a regression introduced by composite indices.)

#Fixes 3010

### DIFF
--- a/src/storage/v2/inmemory/label_property_index.cpp
+++ b/src/storage/v2/inmemory/label_property_index.cpp
@@ -700,9 +700,20 @@ InMemoryLabelPropertyIndex::Iterable::Iterable(utils::SkipList<Entry>::Accessor 
         upper_bound = std::nullopt;
       }
 
+      auto const are_comparable_ranges = [](auto const &lower_bound, auto const &upper_bound) {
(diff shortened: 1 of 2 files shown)


colinbarry:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
  - Queries using regex clauses on label+property indices no longer return an empty results set. Users can now use label+property indices to improve performance of regex queries (although note that, unlike exact text match queries, regex queries require an additional filtering step.) [#3025](https://github.com/memgraph/memgraph/pull/3025)

DavIvek (approved):

PR #3026, opened 2025-06-04
Add STRICT_SYNC replication mode based on two-phase commit protocol

as51340:
Jepsen HA test can now be started in sync, sync_mixed (sync+async replicas), strict_sync and strict_sync_mixed (strict_sync + async replicas) mode. The developer can also specify time-limit upfront.

The new type of replica is added, called `STRICT_SYNC`. The main motivation of adding a new replica type is to support data-loss free automatic failover. The current solution is designed to cover a variety of failure modes and scenarios. 

The cluster can be combined of either `STRICT_SYNC` and `ASYNC` or `SYNC` and `ASYNC` replicas. The combination of `STRICT_SYNC` and `SYNC` replicas is forbidden since such combination semantically wouldn't make sense (`STRICT_SYNC` is a variation of `SYNC`). 
The semantic the of flag `--sync-failover-only` now is better described with: If this flag is true, you can failover to any replica except the `ASYNC` one. 

The commit method is now split into two phases called: `PrepareForCommitPhase` and `FinalizeCommitPhase`. In the 1st phase, we still validate existence and unique constraints same as check for auto-indices. All deltas are made durable on disk in the 1st phase so that instances could survive failure between two phases. A new delta is added `WAL_TRANSACTION_START` which allows aborting durable data, something we didn't support before. If the instance currently behaves as `MAIN`, in the 1st phase, it will collect votes from all replicas to determine whether the instance should be committed or not. If the decision is yes, the value of the delta `WAL_TRANSACTION_START` will be switched to true. The value of the delta `WAL_TRANSACTION_START` is used in two situations:
1. When replica receives WAL file, the WAL file contains deltas both from transactions which should be aborted and committed
2. When instance restarts, again some deltas shouldn't be applied because they come from aborted transactions.

If all replicas voted `YES` in the 1st phase, in the 2nd phase `ldt` (our consensus timestamp) is bumped, same as transaction's commit timestamp. All side-effects are applied also. It is important to note that all data is first made durable on main than on replicas, otherwise if main fails after sending changed to replica, the replica will contain data which main doesn't have. 

The replication's state machine stayed exactly the same as it is, so all the machinery for handling when is replica up to date or behind works as before.

From the user's perspective, if he/she receives a message:
```
At least one STRICT_SYNC replica has not confirmed committing last transaction.
```
It means that the transaction will be aborted. This is different from our `SYNC` mode in which we allow further committing on MAIN, although `SYNC` replica is down. 

Our RPC stack is extended with a functionality of upgrading the RPC stream. The functionality basically allows you take only once RPC lock and send two messages without releasing and re-taking the lock. 

There are several subtleties that got changed in regards of our durability stack that are important for future devs working on Memgraph.

1. Memgraph uses a concept of epochs to distinguish between phases in which the different instance was MAIN. From this PR on we only push an epoch to a history if in that epoch something was committed on the MAIN.
2. The requirement which says that there should always be a WAL file which contains at least one delta before snapshot is taken is not true anymore.
3. WAL file now supports interleaving of committed and aborted transactions

Durability version is bumped to 28 but snapshots should be compatible with older ones.

The abort of the unique constraints is moved to the common `Abort` function.

There is only a single scenario in which users can experience a data loss of a max one transaction which is extremely unlikely. It can happen if the all replicas and main voted YES and after voting YES, both replicas went down. Main commits, replies to the client and then goes down. If replicas get back alive before main, the coordinator has no mechanism to detect that the last txn isn't actually committed on replicas. To support this scenario we would have to glue together coordinators with data instances. We believe this scenario is so unlikely that it is not worth solving at the moment.


src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -273,38 +279,48 @@ void InMemoryReplicationHandlers::HeartbeatHandler(dbms::DbmsHandler *dbms_handl
...
-  const storage::replication::HeartbeatRes res{
-      true, storage->repl_storage_state_.last_durable_timestamp_.load(std::memory_order_acquire),
-      std::string{storage->repl_storage_state_.epoch_.id()}};
+  auto const ldt = storage->repl_storage_state_.last_durable_timestamp_.load(std::memory_order_acquire);
+
+  auto const last_epoch_with_commit = std::invoke([storage, ldt]() -> std::string {
+    if (auto &history = storage->repl_storage_state_.history; !history.empty()) {
+      auto [history_epoch, history_ldt] = history.back();
+      return history_ldt != ldt ? std::string{storage->repl_storage_state_.epoch_.id()} : history_epoch;
+    }
```
andrejtonev, line 287: Is this so we don't return epochs that don't make any changes?
Isn't it easier to just not append the epoch during PrepareForNewEpoch

src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -330,35 +346,94 @@ void InMemoryReplicationHandlers::AppendDeltasHandler(dbms::DbmsHandler *dbms_ha
...
+    rpc::SendFinalResponse(res, res_builder);
+    return;
+  }
+
+  // In this handler, we can either commit or abort. If cached accessor is nullptr, it is impossible we should commit
+  // because replying to prepare happens after assignment to the accessor
+  // If cached accessor is nullptr, and we should abort (e.g. exception was thrown while processing deltas), we can
+  // safely return here OK because it means that the abort already happened while destructing accessor during
+  // ReadAndApplyDeltasSingleTxn
+  if (!cached_commit_accessor_) {
```
andrejtonev: To be safe, there should be a way to figure out if the phase2 is part of the same phase1.
This won't be enough if the seen messages are: phase1 (tx1) and then phase2 (tx2).

andrejtonev: I suppose phase2 (tx2) can only happen if phase1 (tx2) returned ok. But this is relying on the correct logic on MAIN. There should be another simpler logic to make sure it's correct

src/storage/v2/durability/durability.cpp
```diff
@@ -596,19 +596,12 @@ std::optional<RecoveryInfo> Recovery::RecoverData(
...
           LOG_FATAL(
               "There are missing prefix WAL files and data can't be "
               "recovered without them!");
-        } else if (first_wal.from_timestamp > *snapshot_timestamp) {
-          // We recovered from a snapshot and we must have at least one WAL file
-          // that has at least one delta that was created before the snapshot in order to
-          // verify that nothing is missing from the beginning of the WAL chain.
-          LOG_FATAL(
-              "You must have at least one WAL file that contains at least one "
-              "delta that was created before the snapshot file!");
```
andrejtonev: What changed to make this irrelevant?

src/storage/v2/durability/wal.cpp
```diff
@@ -808,6 +823,9 @@ std::optional<RecoveryInfo> LoadWal(
   auto vertex_acc = vertices->access();
   spdlog::info("WAL file contains {} deltas.", info.num_deltas);
 
+  // In 2PC, we can have deltas stored on disk which shouldn't be applied when recovering
+  bool should_commit{true};
```
andrejtonev, line 827: Is this default true all that is needed for back-compatibility?

as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - **A new STRICT_SYNC replication mode is added. Its benefits are mostly observable in high-availability deployments where users can from now on, when registering replicas as strictly sync, run a high-availability cluster without the fear of experiencing a data loss during a failover**. [#3026](https://github.com/memgraph/memgraph/pull/3026)
- [x] **[ Documentation PR link ](https://github.com/memgraph/documentation/pull/1351)**

src/storage/v2/durability/wal.cpp
```diff
@@ -1194,22 +1213,33 @@ std::optional<RecoveryInfo> LoadWal(
...
+        (!last_applied_delta_timestamp || timestamp > *last_applied_delta_timestamp)) {
       // This delta should be loaded.
       auto delta = ReadWalDeltaData(&wal, *version);
-      std::visit(delta_apply, delta.data_);
-      ++deltas_applied;
-      ret.next_timestamp = std::max(ret.next_timestamp, timestamp + 1);
+
+      // We should always check if the delta is WalTransactionStart to update should_commit
+      if (auto *txn_start = std::get_if<WalTransactionStart>(&delta.data_)) {
+        should_commit = txn_start->commit;
```
andrejtonev, line 1253: won't this be handled by the delta?

src/storage/v2/inmemory/storage.cpp
```diff
@@ -71,7 +70,7 @@ extern const Event PeakMemoryRes;
 
 namespace memgraph::storage {
 namespace {
-constexpr auto ActionToStorageOperation(MetadataDelta::Action action) -> durability::StorageMetadataOperation {
+constexpr auto ActionToStorageOperation(MetadataDelta::Action const action) -> durability::StorageMetadataOperation {
```
andrejtonev, line 73: Got to here

src/dbms/inmemory/replication_handlers.cpp
as51340: Before, this would return the newest epoch with the latest ldt. PrepareForNewEpoch saves into history the last epoch not the new one.

src/storage/v2/durability/wal.cpp
as51340: yes, should be

src/storage/v2/durability/wal.cpp
as51340: yes but I don't want to call std::apply for each delta. I need to call it just if should_commit is true, therefore I need to separate the logic when the delta is waltxnstart and when is something else

src/storage/v2/durability/durability.cpp
as51340: Imagine MAIN sending WAL file with all aborts to replica. Replica will accept this WAL file but won't create its own WAL file with these timestamps because it will figure out that txns are aborted.

src/dbms/inmemory/replication_handlers.cpp
as51340: It relies on the main logic but I can make it safer by passing durability_commit_timestamp in both phases and then verifying all is good

src/storage/v2/inmemory/storage.cpp
```diff
@@ -671,213 +746,219 @@ utils::BasicResult<StorageManipulationError, void> InMemoryStorage::InMemoryAcce
...
+    if (repl_args.is_main || repl_args.desired_commit_timestamp.has_value()) {
+      auto const durability_commit_timestamp =
+          repl_args.desired_commit_timestamp.has_value() ? *repl_args.desired_commit_timestamp : *commit_timestamp_;
+
+      // Don't write things to WAL and replicate only if needed
+      if (!mem_storage->InitializeWalFile(mem_storage->repl_storage_state_.epoch_)) {
+        // No need to finalize WAL
+        FinalizeCommitPhase(durability_commit_timestamp);
+        return {};
+      }
```
andrejtonev, line 832: not really following this

src/storage/v2/inmemory/storage.cpp
```diff
@@ -671,213 +746,219 @@ utils::BasicResult<StorageManipulationError, void> InMemoryStorage::InMemoryAcce
...
-#endif
-          mem_storage->repl_storage_state_.last_durable_timestamp_.store(durability_commit_timestamp);
+      // If we are here, it means we are the main executing the commit and there are some STRICT_SYNC replicas in the
+      // cluster.
+      if (repl_prepare_phase_status) {
+        FinalizeCommitPhase(durability_commit_timestamp);
+        // Important that WAL file gets finalize after the finalize commit phase finished because we are still modifying
+        // WAL file
+        if (mem_storage->wal_file_) {
+          mem_storage->FinalizeWalFile();
```
andrejtonev, line 876: This can reset the wal_file_. Not sure that is okay in this situation. The wal file needs to be updated later on, right?

src/dbms/inmemory/replication_handlers.cpp
andrejtonev: will be handled in https://github.com/memgraph/memgraph/pull/3110

src/storage/v2/durability/durability.cpp
andrejtonev: I don't think this logic has anything to we replication. 
The idea is that the only way to know that a wal chain contains all information is to have at least one transaction in the wal chain that happened before the snapshot. Then the chain can be reconstructed and the continuity verified, even if the snapshot deleted parts of the chain

src/storage/v2/durability/wal.cpp
andrejtonev: I think that we can skip the type check by looking at the should_commit only. 
If it is true, apply. Any aborts will update the should commit automatically.
If false, check the type then and update should_commit manually.

src/storage/v2/durability/durability.cpp
as51340: I mean, I definitely managed to crash this during one of the tests with completely valid scenario. It doesn't have directly connection with replication but it happened on replica because of the above mentioned thing.

src/storage/v2/durability/wal.cpp
as51340: Yes, that is the same logic.

Maybe what you wrote is more efficient because checking bool is more efficient than variant but apart from that should be the same. You want me to change this because of perf?

src/storage/v2/inmemory/storage.cpp
as51340:
```
bool InMemoryStorage::InitializeWalFile(memgraph::replication::ReplicationEpoch &epoch) {
  if (config_.durability.snapshot_wal_mode != Config::Durability::SnapshotWalMode::PERIODIC_SNAPSHOT_WITH_WAL) {
    return false;
  }
```

This handles specific scenario

src/storage/v2/inmemory/storage.cpp
as51340: Wal file is updated only if `repl_prepare_phase_status` is true. By default the flag in wal file says: Abort this txn.

That's why it is ok to reset wal_file_ in this situation

src/storage/v2/durability/durability.cpp
andrejtonev: There needs to be a way to tell that a wal chain continues from a snapshot. Otherwise we could be missing chunks of data. 
Until now, that was done by having a wal file that starts before the snapshot.
Not sure how this assert was triggered in your case. If the wal file is all aborts and was missing, how did the timestamp check fail?

as51340: Returned the check. Issue was in comparing commit_timestamps on WAL with start timestamps of the snapshot

PR #3084, opened 2025-07-04
Rename `date.py` so it can co-exist with MAGE `date.py`

colinbarry:
Rename Memgraph's `date.py` query module to `mg_date.py` to avoid a name collision with the MAGE module with the same name. (Once we have a monorepo, these two files can be combined into a single `date.py`.)

### DIFF
--- a/config/mappings.json
+++ b/config/mappings.json
@@ -24,5 +24,5 @@
     "apoc.refactor.cloneSubgraph": "refactor.clone_subgraph",
     "apoc.refactor.cloneSubgraphFromPath": "refactor.clone_subgraph_from_path",
     "apoc.label.exists": "label.exists",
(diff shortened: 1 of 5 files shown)


colinbarry:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist

colinbarry:
Abandoning this approach, as it would mean we would forever have split date functions into `date.` and `chrono.` namespaces, with no logical (from the user's perspective) reason for the division. A better approach is simply to ensure that both Memgraph and MAGE's `date.py` implementations are the same. Yes, this means a small amount of duplication; but when we have a monorepo, this will no longer be an issue.

PR #3086, opened 2025-07-04
MT RBAC 2

andrejtonev:


andrejtonev:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Added support for multiple roles per user, enabling database-specific permissions and enhanced SSO integration. Users can now have different roles for different databases, and SSO providers can assign multiple roles that will all be recognized by the system. Existing deployments will automatically migrate. New multi-role features are available immediately, including SHOW CURRENT ROLE command and enhanced SET ROLE/CLEAR ROLE commands.
    Breaking change: System queries (auth, replication, multi-tenent) are not targeting the default database "memgraph". Users need to have both the correct privileges and access to "memgraph" in order to execute them. To verify user permissions on "memgraph", run `SHOW PRIVILEGES FOR user ON DATABASE memgraph` [3086](https://github.com/memgraph/memgraph/pull/3086)
- [x] https://github.com/memgraph/documentation/pull/1337
- [x] @gitbuda @matea16

tests/unit/auth.cpp
```diff
@@ -44,6 +46,21 @@ class AuthWithStorage : public ::testing::Test {
...
+  void SetUp() override {
+    memgraph::utils::EnsureDir(test_folder_);
+    memgraph::license::global_license_checker.EnableTesting();
+    auth.emplace(test_folder_, auth_config);
+  }
+
+  void TearDown() override {}
+
+  fs::path test_folder_{fs::path{boost::dll::program_location().parent_path().string()} / "auth_kvstore/v1"};
+  Auth::Config auth_config{};
```
colinbarry, line 60: Inconsistent naming for members here: `test_folder_` has trailing prefix, where the others don't.

src/auth/models.hpp
```diff
@@ -519,34 +736,29 @@ class User final {
...
-  }
+  const std::string &GetMain() const { return database_access_.GetMain(); }
+
+  bool DeniesDB(std::string_view db_name) const { return database_access_.Denies(db_name) || roles_.DeniesDB(db_name); }
+  bool GrantsDB(std::string_view db_name) const { return database_access_.Grants(db_name) || roles_.GrantsDB(db_name); }
+
   bool HasAccess(std::string_view db_name) const { return !DeniesDB(db_name) && GrantsDB(db_name); }
+  bool has_access(std::string_view db_name) const {
+    return !database_access_.Denies(db_name) && database_access_.Grants(db_name);
+  }
```
colinbarry, line 721: Not clear to me why we have two very similarly named functions with what appears to be slightly different semantics. Seems the `has_access` impl doesn't care about `roles_`? If that's the intent, maybe clarify so in the function name.

tests/unit/auth.cpp
```diff
@@ -243,6 +260,362 @@ TEST_F(AuthWithStorage, UserRoleFineGrainedAccessHandler) {
...
+    user->permissions().Grant(Permission::STATS);
+    user->permissions().Deny(Permission::MATCH);
+    user->permissions().Grant(Permission::CREATE);
+
+    auto db1_permissions = user->GetPermissions("db1");
+    EXPECT_FALSE(user->has_access("db1"));
+    EXPECT_TRUE(user->HasAccess("db1"));
+    EXPECT_EQ(db1_permissions.Has(Permission::INDEX), PermissionLevel::NEUTRAL);
+    EXPECT_EQ(db1_permissions.Has(Permission::STATS), PermissionLevel::NEUTRAL);
+    EXPECT_EQ(db1_permissions.Has(Permission::MATCH), PermissionLevel::GRANT);
```
colinbarry, line 542: Expect this to be `DENY`, as the user has explicit `user->permissions().Deny(Permission::MATCH)` on line 534

tests/unit/auth.cpp
```diff
@@ -243,6 +260,362 @@ TEST_F(AuthWithStorage, UserRoleFineGrainedAccessHandler) {
...
+    EXPECT_EQ(db4_permissions.Has(Permission::CREATE), PermissionLevel::GRANT);
+    ASSERT_EQ(db4_permissions.Has(Permission::SET), PermissionLevel::GRANT);
+    ASSERT_EQ(db4_permissions.Has(Permission::REMOVE), PermissionLevel::GRANT);
+
+    auto none_permissions = user->GetPermissions("nonexistent");
+    EXPECT_FALSE(user->has_access("nonexistent"));
+    EXPECT_FALSE(user->HasAccess("nonexistent"));
+    EXPECT_EQ(none_permissions.Has(Permission::INDEX), PermissionLevel::NEUTRAL);
+    EXPECT_EQ(none_permissions.Has(Permission::STATS), PermissionLevel::NEUTRAL);
+    EXPECT_EQ(none_permissions.Has(Permission::MATCH), PermissionLevel::NEUTRAL);
```
colinbarry, line 564: Also expect this to be `DENY` for same reason as above.

tests/unit/auth.cpp
```diff
@@ -243,6 +260,362 @@ TEST_F(AuthWithStorage, UserRoleFineGrainedAccessHandler) {
...
+    ASSERT_TRUE(user->CanImpersonate(*target_user));
+
+    // Test CanImpersonate with db1 filter - should only include role1
+    ASSERT_TRUE(user->CanImpersonate(*target_user, "db1"));
+
+    // Test CanImpersonate with db2 filter - should only include role2
+    ASSERT_TRUE(user->CanImpersonate(*target_user, "db2"));
+
+    // Test CanImpersonate with db3 filter - should only include role3
+    ASSERT_TRUE(user->CanImpersonate(*target_user, "db3"));
```
colinbarry, line 601: I'd suggest rather than testing all three roles as `Grant`, update the test to check
- a role with neutral impersonate permissions
- a role with deny permissions

tests/unit/auth.cpp
```diff
@@ -243,6 +260,362 @@ TEST_F(AuthWithStorage, UserRoleFineGrainedAccessHandler) {
...
+    user = auth->GetUser("user");
+
+    // Test CanImpersonate without database filter - should include all roles
+    ASSERT_TRUE(user->CanImpersonate(*target_user));
+
+    // Test CanImpersonate with db1 filter - should only include role1
+    ASSERT_TRUE(user->CanImpersonate(*target_user, "db1"));
+
+    // Test CanImpersonate with db2 filter - should only include role2
+    ASSERT_TRUE(user->CanImpersonate(*target_user, "db2"));
```
colinbarry, line 598: Be good to see an `ASSERT_FALSE` test for `CanImpersonate` with an additional user to who permissions haven't been granted.

tests/unit/auth.cpp
```diff
@@ -243,6 +260,362 @@ TEST_F(AuthWithStorage, UserRoleFineGrainedAccessHandler) {
...
+    ASSERT_TRUE(user->CanImpersonate(*target_user, "db3"));
+
+    // But should fail for db1 due to deny
+    ASSERT_FALSE(user->CanImpersonate(*target_user, "db1"));
+  }
+}
 #endif
 
 TEST_F(AuthWithStorage, RoleManipulations) {
```
colinbarry, line 689: I'd like to see a test here with multiple roles for a user.

src/auth/models.cpp
```diff
@@ -709,63 +733,163 @@ void User::UpdatePassword(const std::optional<std::string> &password,
...
 
-void User::SetRole(const Role &role) { role_.emplace(role); }
+void User::SetRole(const Role &role) {
+  // Clear all roles and add the new one
+  ClearRole();
+  roles_.AddRole(role);
+}
 
-void User::ClearRole() { role_ = std::nullopt; }
+void User::ClearRole() {
```
colinbarry: Maybe this would be better named as `ClearRoles()` or `ClearAllRoles()`

tests/unit/auth.cpp
colinbarry: OK, you have already written such tests below in a separate `TEST` called `MultipleRoles`. ✅

tests/unit/auth.cpp
```diff
@@ -764,6 +1137,104 @@ TEST_F(AuthWithStorage, UserRoleUniqueName) {
...
+  user->AddRole(*role3);
+  auth->SaveUser(*user);
+
+  // Verify user has all roles
+  auto retrieved_user = auth->GetUser("user");
+  ASSERT_TRUE(retrieved_user);
+  ASSERT_EQ(retrieved_user->roles().size(), 3);
+
+  // Check that role() returns the first role for backward compatibility
+  const auto &roles = retrieved_user->roles();
```
colinbarry, line 1230: I don't think the comment and code agree on what is being checked here. The `role()` function no longer exists (but I think the test is valid, just the comment is out dated.)

tests/unit/auth.cpp
```diff
@@ -764,6 +1137,104 @@ TEST_F(AuthWithStorage, UserRoleUniqueName) {
...
+  retrieved_user->RemoveRole("role2");
+  auth->SaveUser(*retrieved_user);
+
+  auto updated_user = auth->GetUser("user");
+  ASSERT_TRUE(updated_user);
+  ASSERT_EQ(updated_user->roles().size(), 2);
+
+  // Verify role2 was removed
+  bool has_role2 = false;
+  for (const auto &role : updated_user->roles()) {
```
colinbarry: `find_if` or `ranges::contains` with a projection would be more idiomatic for this check.

tests/unit/auth.cpp
```diff
@@ -764,6 +1137,104 @@ TEST_F(AuthWithStorage, UserRoleUniqueName) {
...
+  auto cleared_user = auth->GetUser("user");
+  ASSERT_TRUE(cleared_user);
+  ASSERT_EQ(cleared_user->roles().size(), 0);
+}
+
+TEST_F(AuthWithStorage, StorageFormatBackwardCompatibility) {
+  // Test that the new storage format maintains backward compatibility
+
+  // Create a user with a single role (old format)
+  auto user = auth->AddUser("user");
```
colinbarry: Is this actually testing what the comment says it is testing? It is one role, but isn't it being saved in the new format, as an array with one element?

colinbarry: (But it does seems that old format support is being tested correctly as part of `V1Auth.MigrationTest`.)

tests/unit/auth_checker.cpp
```diff
@@ -310,16 +310,17 @@ TEST(AuthChecker, Generate) {
   new_role.db_access().Grant("non_default");
   auth->SaveUser(new_user2);
   auth->SaveRole(new_role);
+  user2 = auth_checker.GenQueryUser("new_user2", {});
   // Session policy test
   // Session long policy
   EXPECT_FALSE(user2->IsAuthorized({AUTH}, "non_default", &memgraph::query::session_long_policy));
-  EXPECT_FALSE(user2->IsAuthorized({AUTH}, "another", &memgraph::query::session_long_policy));
+  EXPECT_TRUE(user2->IsAuthorized({AUTH}, "another", &memgraph::query::session_long_policy));
```
colinbarry: I'm probably missing something, but could you clarify why this test is now true, where previously is was false?

tests/unit/auth_checker.cpp
```diff
@@ -310,16 +310,17 @@ TEST(AuthChecker, Generate) {
   new_role.db_access().Grant("non_default");
   auth->SaveUser(new_user2);
   auth->SaveRole(new_role);
+  user2 = auth_checker.GenQueryUser("new_user2", {});
```
colinbarry: Won't `user2` be exactly the same as it was when created on line 297? Do we need to create another?

tests/unit/auth_models.cpp
```diff
@@ -1092,3 +1092,217 @@ TEST(AuthModule, RoleSerialization) {
   json[kUserImp][kUserImpDenied] = nlohmann::json::array({{{kUserImpName, "user2"}, {kUserImpId, user2.uuid()}}});
   ASSERT_EQ(json, role.Serialize());
 }
+
+#ifdef MG_ENTERPRISE
+TEST(AuthModule, UserMultiTenantRoleManagement) {
```
colinbarry: Feel we have some test duplication here: this seems to be testing the same code as is tested in the `auth.cpp` tests.

src/auth/rpc.cpp
```diff
@@ -46,24 +46,49 @@ inline void Load<auth::Role>(std::optional<auth::Role> *obj, Reader *reader) {
 // Serialize code for auth::User
 void Save(const auth::User &self, memgraph::slk::Builder *builder) {
   memgraph::slk::Save(self.Serialize().dump(), builder);
-  std::optional<auth::Role> role{};
-  if (const auto *role_ptr = self.role(); role_ptr) {
-    role.emplace(*role_ptr);
+  std::vector<auth::Role> roles;
+  for (const auto &role : self.roles()) {
```
colinbarry: Minor opportunity for an easy optimization here: either `reserve` based on the `roles()` size, or just construct the `vector` directly from `roles().cbegin()` and `roles.cend()`.

src/glue/auth_handler.hpp
```diff
@@ -63,15 +65,20 @@ class AuthQueryHandler final : public memgraph::query::AuthQueryHandler {
 
   std::vector<memgraph::query::TypedValue> GetRolenames() override;
 
-  std::optional<std::string> GetRolenameForUser(const std::string &username) override;
+  std::vector<std::string> GetRolenameForUser(const std::string &username, std::optional<std::string> db_name) override;
```
colinbarry: Should be plural: `GetRolenamesForUser`?

src/glue/auth_handler.hpp
```diff
@@ -63,15 +65,20 @@ class AuthQueryHandler final : public memgraph::query::AuthQueryHandler {
...
 
   std::vector<memgraph::query::TypedValue> GetUsernamesForRole(const std::string &rolename) override;
 
-  void SetRole(const std::string &username, const std::string &rolename, system::Transaction *system_tx) override;
+  void SetRole(const std::string &username, const std::vector<std::string> &roles,
```
colinbarry: Should be `SetRoles`

src/glue/auth_handler.hpp
```diff
@@ -63,15 +65,20 @@ class AuthQueryHandler final : public memgraph::query::AuthQueryHandler {
...
+               const std::unordered_set<std::string> &role_databases, system::Transaction *system_tx) override;
+
+  void RemoveRole(const std::string &username, const std::string &rolename, system::Transaction *system_tx) override;
 
-  void ClearRole(const std::string &username, system::Transaction *system_tx) override;
+  void ClearRole(const std::string &username, const std::unordered_set<std::string> &role_databases,
```
colinbarry: Should be `ClearRoles`

src/auth/auth.cpp
```diff
@@ -221,6 +221,51 @@ void MigrateVersions(kvstore::KVStore &store) {
...
+
+    auto puts = std::map<std::string, std::string>{{kVersion, kVersionV2}};
+    auto deletes = std::vector<std::string>{};
+
+    // Migrate all link entries from single role format to JSON array format
+    for (auto it = store.begin(kLinkPrefix); it != store.end(kLinkPrefix); ++it) {
+      auto const &[key, value] = *it;
+
+      // Check if this is already in the new format (JSON array)
+      try {
```
colinbarry: Is this something than can ever happen? If v1 was the single role, then is there any circumstance where it already could be in an array format?

src/auth/models.cpp
```diff
@@ -494,11 +494,35 @@ Permissions &Role::permissions() { return permissions_; }
 const FineGrainedAccessHandler &Role::fine_grained_access_handler() const { return fine_grained_access_handler_; }
 FineGrainedAccessHandler &Role::fine_grained_access_handler() { return fine_grained_access_handler_; }
 
-const FineGrainedAccessPermissions &Role::GetFineGrainedAccessLabelPermissions() const {
+const FineGrainedAccessPermissions &Role::GetFineGrainedAccessLabelPermissions(
+    std::optional<std::string_view> db_name) const {
+  if (!memgraph::license::global_license_checker.IsEnterpriseValidFast()) {
+    static FineGrainedAccessPermissions empty_permissions;
```
colinbarry: Lots of individual `static` `empty_permissions`. Would it be better just to use a single file/namespace scope instance?

src/query/auth_query_handler.hpp
```diff
@@ -90,18 +97,25 @@ class AuthQueryHandler {
   virtual std::vector<memgraph::query::TypedValue> GetRolenames() = 0;
 
   /// @throw QueryRuntimeException if an error ocurred.
-  virtual std::optional<std::string> GetRolenameForUser(const std::string &username) = 0;
+  virtual std::vector<std::string> GetRolenameForUser(const std::string &username,
```
colinbarry: Should be `GetRolenamesForUser`

src/query/auth_query_handler.hpp
```diff
@@ -90,18 +97,25 @@ class AuthQueryHandler {
...
+                                                      std::optional<std::string> db_name) = 0;
 
   /// @throw QueryRuntimeException if an error ocurred.
   virtual std::vector<memgraph::query::TypedValue> GetUsernamesForRole(const std::string &rolename) = 0;
 
   /// @throw QueryRuntimeException if an error ocurred.
-  virtual void SetRole(const std::string &username, const std::string &rolename, system::Transaction *system_tx) = 0;
+  virtual void SetRole(const std::string &username, const std::vector<std::string> &roles,
```
colinbarry: Should be `SetRoles`, `RemoveRoles`, `ClearRoles`/`ClearAllRoles`, etc...

src/query/query_user.hpp
```diff
@@ -34,31 +34,34 @@ extern struct UpToDatePolicy : UserPolicy {
...
-  virtual bool CanImpersonate(const std::string &target, UserPolicy *policy) const = 0;
+  virtual bool CanImpersonate(const std::string &target, UserPolicy *policy,
+                              std::optional<std::string_view> db_name = std::nullopt) const = 0;
   virtual std::string GetDefaultDB() const = 0;
 #endif
 
   std::string key() const {
     // NOTE: Each role has an associated username, that's why we check it with higher priority
-    return rolename_ ? *rolename_ : (username_ ? *username_ : "");
+    return !rolenames_.empty() ? rolenames_[0] : (username_ ? *username_ : "");
```
colinbarry: What are the implications of the `key()` being based solely on the 1st item in the `rolenames_` array? I notice `GetDatabasePrivileges` is called with the user key in `interpreter.cpp`

src/auth/models.hpp
```diff
@@ -460,6 +487,181 @@ class Role {
...
+  auto end() const { return roles_.end(); }
+  auto cbegin() const { return roles_.cbegin(); }
+  auto cend() const { return roles_.cend(); }
+
+  // Size and empty checks
+  bool empty() const { return roles_.empty(); }
+  size_t size() const { return roles_.size(); }
+
+  // Get pointer to first role (for backward compatibility)
+  const Role *first_role() const {
```
colinbarry: Is this actually used? I believe not, in which case, it can be deleted.

src/auth/models.hpp
```diff
@@ -498,19 +700,34 @@ class User final {
 
   void UpdateHash(HashedPassword hashed_password);
 
-  void SetRole(const Role &role);
+  void SetRole(const Role &role);  // Deprecated
```
colinbarry: Probably better to use the attribute `[[deprecated]]` in that case.

src/glue/SessionHL.cpp
```diff
@@ -539,38 +530,77 @@ bolt_map_t SessionHL::DecodeSummary(const std::map<std::string, memgraph::query:
...
+    if (!info.IsString()) {
+      throw memgraph::communication::bolt::ClientError("Malformed config input.");
     }
-    explicit_ = true;
-    // NOTE: Once in a transaction, the drivers stop explicitly sending the config and count on using it until commit
-  } else if (explicit_ && !in_explicit_tx) {  // Just on a switch
-    if (implicit_config_) {
-      defined_config = *implicit_config_;
+    defined_db = info.ValueString();
+    // TODO ?
```
colinbarry: Just bringing a slightly cryptic TODO to your attention. I'd either elaborate on the comment and leave it, or just delete both the TODO and commented out code.

src/glue/auth_handler.cpp
```diff
@@ -475,6 +481,34 @@ void AuthQueryHandler::DeleteDatabase(std::string_view db_name, system::Transact
     throw memgraph::query::QueryRuntimeException(e.what());
   }
 }
+
+std::optional<std::string> AuthQueryHandler::GetMainDatabase(const std::string &user_or_role) {
+  try {
+    auto locked_auth = auth_->ReadLock();
+    auto user = locked_auth->GetUser(user_or_role);
+    auto role = locked_auth->GetRole(user_or_role);
```
colinbarry: Seeing as the `role` isn't used if the `user` is not null, could we refactor to only acquire the `role` if needed, rather than computing both upfront?

src/query/frontend/opencypher/grammar/MemgraphCypher.g4
```diff
@@ -492,9 +495,9 @@ listOfColonSymbolicNames : colonSymbolicName ( ',' colonSymbolicName )* ;
 
 colonSymbolicName : COLON symbolicName ;
 
-showPrivileges : SHOW PRIVILEGES FOR userOrRole=userOrRoleName ;
+showPrivileges : SHOW PRIVILEGES FOR userOrRole=userOrRoleName ( ON ( MAIN | CURRENT | DATABASE db=symbolicName ) )? ;
```
colinbarry, line 499: There is an issue related to `SHOW PRIVILEGES`, but I haven't had time to narrow it down yet. From an empty db, run the following setup:

```cypher
CREATE USER alice;
CREATE USER bob;
CREATE USER charlie;
CREATE ROLE user;
CREATE ROLE architect;
CREATE ROLE moderator;
CREATE ROLE admin;
CREATE ROLE support;

CREATE DATABASE db1;
CREATE DATABASE db2;
CREATE DATABASE db3;

GRANT DATABASE db1 TO architect;
GRANT DATABASE db1 TO moderator;
GRANT DATABASE db2 TO support;
GRANT DATABASE db2 TO admin;
SET MAIN DATABASE db1 FOR architect;
SET MAIN DATABASE db2 FOR support;

SET ROLE FOR charlie TO user;
SET ROLE FOR charlie TO architect, moderator ON db1;
SET ROLE FOR charlie TO admin, support ON db2;

GRANT match TO architect;
DENY match TO moderator;
```

And then try:
```cypher
SHOW PRIVILEGES FOR charlie ON DATABASE db1
```

fails with "Role 'architect' is already specified as a multi-tenant role"

src/auth/models.hpp
andrejtonev: This is following the old format, where the HasAccess function generates a combined result, while the has_access gives the user only result. There are many such functions in the auth library

tests/unit/auth.cpp
andrejtonev: As discussed offline, here the user doesn't have access to db1, so its permissions are ignored.
Tested with has_access

tests/unit/auth.cpp
andrejtonev: same explanation :D

src/glue/SessionHL.cpp
andrejtonev: Was thinking I should add a check if the db even exists, but that happens later on anyway

src/query/frontend/opencypher/grammar/MemgraphCypher.g4
andrejtonev: Nice catch, the role was not being properly checked for. Fixed it!

tests/unit/auth.cpp
andrejtonev: this is just some ai nonsense

tests/unit/auth_checker.cpp
andrejtonev: It's a confusing test...
I added the line before that updates the user. 
For reasons....we have a session long permissions and up to date permissions. This is testing that. 
By updating the user I forced the permissions to update, but that is not what the test was doing before. The test was checking for session long permissions. I reverted this anyway

tests/unit/auth_checker.cpp
andrejtonev: comment above

src/auth/auth.cpp
andrejtonev: yup

src/query/query_user.hpp
andrejtonev: Removoed this key function. It doesn't make any sense now.

PR #3089, opened 2025-07-07
Added implementation for query module funtion to convert data to tree.

tonijurjevic96:
Added functionality to take a path in graph and represent it as a tree.

### DIFF
--- a/query_modules/convert.cpp
+++ b/query_modules/convert.cpp
@@ -9,13 +9,239 @@
 // by the Apache License, Version 2.0, included in the file
 // licenses/APL.txt.
 
+#include <algorithm>
 #include <iostream>
 #include <mgp.hpp>
 #include <optional>
+#include <string>
+#include <unordered_set>
+#include <vector>
 
 #include <nlohmann/json.hpp>
 
+// Forward declarations
+mgp::Value ConvertToTreeImpl(const mgp::Value &value, const mgp::Map &config, mgp_memory *memory);
 std::optional<mgp::Value> ParseJsonToMgpValue(const nlohmann::json &json_obj, mgp_memory *memory);
+
+// Helper function to check if a property should be included
+bool ShouldIncludeProperty(const std::string_view &prop_name, const mgp::Map &config, const mgp::Labels &labels) {
+  if (!config.KeyExists("nodes")) {
+    return true;
+  }
+
+  auto nodes = config.At("nodes").ValueMap();
+  bool any_exclude_mode = false;
+  std::vector<std::string_view> excluded_props;
+
+  for (size_t l = 0; l < labels.Size(); ++l) {
+    std::string_view label = labels[l];
+    if (!nodes.KeyExists(label)) continue;
+    auto props = nodes.At(label).ValueList();
+    if (props.Size() == 0) continue;
+    auto first_prop = props[0].ValueString();
+    if (first_prop.size() > 0 && first_prop[0] == '-') {
+      any_exclude_mode = true;
+      if (first_prop.size() > 1) {
+        excluded_props.push_back(first_prop.substr(1));
+      }
+      for (size_t i = 1; i < props.Size(); ++i) {
+        auto excluded = props[i].ValueString();
+        if (excluded.size() > 0 && excluded[0] == '-') {
+          excluded = excluded.substr(1);
+        }
+        excluded_props.push_back(excluded);
+      }
+    }
+  }
+
+  if (!any_exclude_mode) {
+    for (size_t l = 0; l < labels.Size(); ++l) {
+      std::string_view label = labels[l];
+      if (!nodes.KeyExists(label)) continue;
+      auto props = nodes.At(label).ValueList();
+      if (props.Size() == 0) continue;
+      if (props[0].ValueString() == "*") return true;
+      for (size_t i = 0; i < props.Size(); ++i) {
+        if (props[i].ValueString() == prop_name) return true;
+      }
+    }
+    return false;
+  }
+
+  for (const auto &excluded : excluded_props) {
+    if (prop_name == excluded) {
+      return false;
+    }
+  }
+  return true;
+}
+
+// Helper function to check if a relationship property should be included
+bool ShouldIncludeRelProperty(const std::string_view &prop_name, const mgp::Map &config,
+                              const std::string_view &rel_type) {
+  if (!config.KeyExists("rels")) {
+    return true;
+  }
+
+  auto rels = config.At("rels").ValueMap();
+  if (!rels.KeyExists(rel_type)) {
+    return true;
+  }
+
+  auto props = rels.At(rel_type).ValueList();
+  if (props.Size() == 0) {
+    return true;
+  }
+
+  auto first_prop = props[0].ValueString();
+  if (first_prop.size() > 0 && first_prop[0] == '-') {
+    std::vector<std::string_view> excluded_props;
+    if (first_prop.size() > 1) {
+      excluded_props.push_back(first_prop.substr(1));
+    }
+    for (size_t i = 1; i < props.Size(); ++i) {
+      auto excluded = props[i].ValueString();
+      if (excluded.size() > 0 && excluded[0] == '-') {
+        excluded = excluded.substr(1);
+      }
+      excluded_props.push_back(excluded);
+    }
+    for (const auto &excluded : excluded_props) {
+      if (prop_name == excluded) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  if (first_prop == "*") return true;
+  for (size_t i = 0; i < props.Size(); ++i) {
+    if (props[i].ValueString() == prop_name) return true;
+  }
+  return false;
+}
+
+// Main implementation function
+mgp::Value ConvertToTreeImpl(const mgp::Value &value, const mgp::Map &config, mgp_memory *memory) {
+  if (value.IsNull()) {
+    return value;
+  }
+
+  // If the value is a List of Paths, convert each path and merge into a tree
+  if (value.IsList()) {
+    auto paths = value.ValueList();
+    if (paths.Size() == 0) {
+      return value;
+    }
+
+    // Get the first path to extract the root node
+    auto first_path = paths[0].ValuePath();
+    auto root_node = first_path.GetNodeAt(0);
+
+    // Create root node map with properties
+    mgp::Map root_map;
+    auto root_props = root_node.Properties();
+    auto root_labels = root_node.Labels();
+
+    for (const auto &[key, value] : root_props) {
+      if (ShouldIncludeProperty(key, config, root_labels)) {
+        root_map.Insert(key, value);
+      }
+    }
+
+    // Add type information
+    if (root_labels.Size() == 1) {
+      root_map.Insert("_type", mgp::Value(std::string(root_labels[0])));
+    } else if (root_labels.Size() > 1) {
+      mgp::List label_list;
+      for (size_t i = 0; i < root_labels.Size(); ++i) {
+        label_list.Append(mgp::Value(std::string(root_labels[i])));
+      }
+      root_map.Insert("_type", mgp::Value(label_list));
+    }
+    root_map.Insert("_id", mgp::Value(root_node.Id().AsInt()));
+
+    // Process all paths to collect relationships
+    for (size_t path_idx = 0; path_idx < paths.Size(); path_idx++) {
+      auto path = paths[path_idx].ValuePath();
+
+      // Process each relationship in the path
+      for (size_t i = 0; i < path.Length(); i++) {
+        auto rel = path.GetRelationshipAt(i);
+        auto target_node = path.GetNodeAt(i + 1);
+        std::string rel_type{rel.Type()};
+
+        // Create or get relationship array
+        mgp::List rel_list;
+        if (!root_map.KeyExists(rel_type)) {
+          rel_list = mgp::List();
+        } else {
+          rel_list = root_map.At(rel_type).ValueList();
+        }
+
+        // Create target node map
+        mgp::Map target_map;
+
+        // Add target node properties
+        auto target_props = target_node.Properties();
+        auto target_labels = target_node.Labels();
+
+        // Add all properties that should be included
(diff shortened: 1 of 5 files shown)


tonijurjevic96:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Converts the provided value into a tree structure based on the specified configuration. This procedure is useful for transforming hierarchical data, nested objects, or graph structures into a standardized tree format.  [#{{PR_number}}]({{link to the PR}})
- [x] **https://github.com/memgraph/documentation/pull/1374**

Josipmrden:
@tonijurjevic96 also re-assign another developer for review. One junior + one senior from Core is required

Josipmrden (changes requested):
Added comments

query_modules/convert.cpp
```diff
@@ -9,14 +9,240 @@
...
+#include <vector>
 
 #include <nlohmann/json.hpp>
 
+// Forward declarations
+mgp::Value ConvertToTreeImpl(const mgp::Value &value, const mgp::Map &config, mgp_memory *memory);
 std::optional<mgp::Value> ParseJsonToMgpValue(const nlohmann::json &json_obj, mgp_memory *memory);
 
+// Helper function to check if a property should be included
+bool ShouldIncludeProperty(const std::string_view &prop_name, const mgp::Map &config, const mgp::Labels &labels) {
```
DavIvek: use `std::string_view` by value instead of `const &`

query_modules/convert.cpp
```diff
@@ -9,14 +9,240 @@
...
+  auto nodes = config.At("nodes").ValueMap();
+  bool any_exclude_mode = false;
+  std::vector<std::string_view> excluded_props;
+
+  for (size_t l = 0; l < labels.Size(); ++l) {
+    std::string_view label = labels[l];
+    if (!nodes.KeyExists(label)) continue;
+    auto props = nodes.At(label).ValueList();
+    if (props.Size() == 0) continue;
+    auto first_prop = props[0].ValueString();
```
DavIvek: What if first property isn't a string? Does it have to be a string?

query_modules/convert.cpp
```diff
@@ -9,14 +9,240 @@
...
+      if (props.Size() == 0) continue;
+      if (props[0].ValueString() == "*") return true;
+      for (size_t i = 0; i < props.Size(); ++i) {
+        if (props[i].ValueString() == prop_name) return true;
+      }
+    }
+    return false;
+  }
+
+  for (const auto &excluded : excluded_props) {
```
DavIvek: no need for `const &` since `std::string_view` is stored in vector

Josipmrden:
Ok I haven't seen that we have copied query module tests from MAGE also for Memgraph. Please enable CI so we can see if the tests passed for the PR because at this point all are skipped.

Josipmrden:
Test failing

query_modules/convert.cpp
```diff
@@ -9,14 +9,240 @@
...
+  bool any_exclude_mode = false;
+  std::vector<std::string_view> excluded_props;
+
+  for (size_t l = 0; l < labels.Size(); ++l) {
+    std::string_view label = labels[l];
+    if (!nodes.KeyExists(label)) continue;
+    auto props = nodes.At(label).ValueList();
+    if (props.Size() == 0) continue;
+    auto first_prop = props[0].ValueString();
+    if (first_prop.size() > 0 && first_prop[0] == '-') {
```
DavIvek: Are properties that start with "-" excluded? Is that an arbitrary rule, or does it follow something, for example, a corresponding apoc procedure?

query_modules/convert.cpp
```diff
@@ -9,14 +9,240 @@
...
+    }
+  }
+
+  if (!any_exclude_mode) {
+    for (size_t l = 0; l < labels.Size(); ++l) {
+      std::string_view label = labels[l];
+      if (!nodes.KeyExists(label)) continue;
+      auto props = nodes.At(label).ValueList();
+      if (props.Size() == 0) continue;
+      if (props[0].ValueString() == "*") return true;
```
DavIvek: Again, is "*" following a specific rule? Maybe it would make sense to define these symbols as constants.

query_modules/convert.cpp
```diff
@@ -9,14 +9,240 @@
...
+  for (const auto &excluded : excluded_props) {
+    if (prop_name == excluded) {
+      return false;
+    }
+  }
+  return true;
+}
+
+// Helper function to check if a relationship property should be included
+bool ShouldIncludeRelProperty(const std::string_view &prop_name, const mgp::Map &config,
```
DavIvek: use `std::string_view` by value.

query_modules/convert.cpp
```diff
@@ -9,14 +9,240 @@
...
+  std::vector<std::string_view> excluded_props;
+
+  for (size_t l = 0; l < labels.Size(); ++l) {
+    std::string_view label = labels[l];
+    if (!nodes.KeyExists(label)) continue;
+    auto props = nodes.At(label).ValueList();
+    if (props.Size() == 0) continue;
+    auto first_prop = props[0].ValueString();
+    if (first_prop.size() > 0 && first_prop[0] == '-') {
+      any_exclude_mode = true;
```
DavIvek: Can you change this flag with `excluded_props.empty(`) check?

query_modules/convert.cpp
```diff
@@ -9,14 +9,240 @@
...
+      }
+    }
+  }
+
+  if (!any_exclude_mode) {
+    for (size_t l = 0; l < labels.Size(); ++l) {
+      std::string_view label = labels[l];
+      if (!nodes.KeyExists(label)) continue;
+      auto props = nodes.At(label).ValueList();
+      if (props.Size() == 0) continue;
```
DavIvek: You already checked these things in the loop above, could you try to unify them or at least extract this into a lambda to make it cleaner?

query_modules/convert.cpp
```diff
@@ -9,14 +9,240 @@
...
+      excluded_props.push_back(first_prop.substr(1));
+    }
+    for (size_t i = 1; i < props.Size(); ++i) {
+      auto excluded = props[i].ValueString();
+      if (excluded.size() > 0 && excluded[0] == '-') {
+        excluded = excluded.substr(1);
+      }
+      excluded_props.push_back(excluded);
+    }
+    for (const auto &excluded : excluded_props) {
```
DavIvek: no need for `const &`

query_modules/convert.cpp
```diff
@@ -9,14 +9,240 @@
 // by the Apache License, Version 2.0, included in the file
 // licenses/APL.txt.
 
+#include <algorithm>
 #include <iostream>
 #include <mgp.hpp>
 #include <optional>
+#include <string>
+#include <unordered_set>
```
DavIvek: Looks to me that you don't need `<unordered_set>` and `<algorithm>` headers.

query_modules/convert.cpp
```diff
@@ -9,14 +9,240 @@
...
+        if (target_labels.Size() == 1) {
+          target_map.Insert("_type", mgp::Value(target_labels[0]));
+        } else if (target_labels.Size() > 1) {
+          mgp::List label_list;
+          for (size_t l = 0; l < target_labels.Size(); ++l) {
+            label_list.Append(mgp::Value(target_labels[l]));
+          }
+          target_map.Insert("_type", mgp::Value(label_list));
+        }
+        target_map.Insert("_id", mgp::Value(target_node.Id().AsInt()));
```
DavIvek: Can you wrap this into separate function or lambda function since you are using it on multiple places?

DavIvek (changes requested):
I left a few comments. Could you also add more comments to the algorithm part of the code? It’s time-consuming to understand what’s happening in each part, which makes the code harder to maintain later.

query_modules/convert.cpp
tonijurjevic96: Added comment to code to explain it

query_modules/convert.cpp
```diff
@@ -12,11 +12,252 @@
...
+// Determines if a node property should be included in the output based on config filtering rules.
+bool ShouldIncludeProperty(std::string_view prop_name, const mgp::Map &config, const mgp::Labels &labels) {
+  if (!config.KeyExists("nodes")) {
+    // No filtering config: include all properties
+    return true;
+  }
+
+  auto nodes = config.At("nodes").ValueMap();
+  // Lambda to get a valid property list for a label, or nullopt if not valid
+  auto get_valid_props = [&](const mgp::Map &nodes, const mgp::Labels &labels, size_t l) -> std::optional<mgp::List> {
```
DavIvek: `l` isn't a very expressive variable name.

query_modules/convert.cpp
```diff
@@ -12,11 +12,252 @@
...
+  };
+
+  std::vector<std::string_view> excluded_props;
+
+  // Collect excluded properties if exclusion mode is detected
+  for (size_t l = 0; l < labels.Size(); ++l) {
+    auto props_opt = get_valid_props(nodes, labels, l);
+    if (!props_opt) continue;
+    auto props = *props_opt;
+    auto first_prop = props[0].ValueString();
```
DavIvek: What if first property isn't a string?

query_modules/convert.cpp
andrejtonev: Since the result of `get_valid_props` is further manipulated (getting the first prop and the rest, reading the strings, etc)
I suggest you return the manipulated result directly from here.

query_modules/convert.cpp
```diff
@@ -12,11 +12,252 @@
...
 
+// Helper function to check if a property should be included
+// Determines if a node property should be included in the output based on config filtering rules.
+bool ShouldIncludeProperty(std::string_view prop_name, const mgp::Map &config, const mgp::Labels &labels) {
+  if (!config.KeyExists("nodes")) {
+    // No filtering config: include all properties
+    return true;
+  }
+
+  auto nodes = config.At("nodes").ValueMap();
```
andrejtonev, line 164: Either check everywhere that the types are correct or encapsulate in a try/catch

query_modules/convert.cpp
```diff
@@ -12,11 +12,252 @@
...
+    if (!props_opt) continue;
+    auto props = *props_opt;
+    auto first_prop = props[0].ValueString();
+    if (first_prop.size() > 0 && first_prop[0] == '-') {
+      if (first_prop.size() > 1) {
+        excluded_props.push_back(first_prop.substr(1));
+      }
+      for (size_t i = 1; i < props.Size(); ++i) {
+        auto excluded = props[i].ValueString();
+        if (excluded.size() > 0 && excluded[0] == '-') {
```
andrejtonev: Create a function that returns this result.
There are other places that this pops up. Or even better create a struct that encapsulates this functionality

query_modules/convert.cpp
```diff
@@ -12,11 +12,252 @@
...
+          excluded = excluded.substr(1);
+        }
+        excluded_props.push_back(excluded);
+      }
+    }
+  }
+
+  // Exclusion mode: skip if property is in excluded_props
+  if (!excluded_props.empty()) {
```
andrejtonev: This can be an unordered_set and then just do `contains(prop)`

query_modules/convert.cpp
```diff
@@ -12,11 +12,252 @@
...
+    }
+    return true;
+  }
+
+  // Inclusion mode: check if property is explicitly included
+  for (size_t l = 0; l < labels.Size(); ++l) {
+    auto props_opt = get_valid_props(nodes, labels, l);
+    if (!props_opt) continue;
+    auto props = *props_opt;
+    if (props[0].ValueString() == "*") return true;
```
andrejtonev: Create a function or a struct that encapsulates this logic. 
Or at least create a constant that names "*" to something meaningful

query_modules/convert.cpp
```diff
@@ -100,11 +341,44 @@ void str2object(mgp_list *args, mgp_func_context *ctx, mgp_func_result *res, mgp
...
+  try {
+    mgp::MemoryDispatcherGuard guard(memory);
+    const auto arguments = mgp::List(args);
+    const auto record = arguments[0];
+
+    // Create a default empty config if none provided
+    mgp::Map config;
+    if (arguments.Size() > 1) {
+      config = arguments[1].ValueMap();
+    }
```
andrejtonev, line 424: Don't see a test that is testing this scenario

query_modules/convert.cpp
```diff
@@ -12,11 +12,252 @@
...
+      if (props[i].ValueString() == prop_name) return true;
+    }
+  }
+  return false;
+}
+
+// Helper function to check if a relationship property should be included
+// Determines if a relationship property should be included in the output based on config filtering rules.
+bool ShouldIncludeRelProperty(std::string_view prop_name, const mgp::Map &config, const std::string_view &rel_type) {
+  if (!config.KeyExists("rels")) {
```
andrejtonev, line 198: I understand that the query input is a Map, but it would be better if there was a class that would just convert that into something simpler that would then be used here.

query_modules/convert.cpp
```diff
@@ -12,11 +12,252 @@
...
+        excluded = excluded.substr(1);
+      }
+      excluded_props.push_back(excluded);
+    }
+    for (auto excluded : excluded_props) {
+      if (prop_name == excluded) {
+        return false;
+      }
+    }
+    return true;
```
andrejtonev: Same comments as for `ShouldIncludeProperty`

query_modules/convert.cpp
```diff
@@ -12,11 +12,252 @@
...
+  // If the value is a List of Paths, convert each path and merge into a tree
+  if (value.IsList()) {
+    auto paths = value.ValueList();
+    if (paths.Size() == 0) {
+      // Empty list: return as is
+      return value;
+    }
+
+    // Get the first path to extract the root node
+    auto first_path = paths[0].ValuePath();
```
andrejtonev, line 225: Again, or check that all values are the expected type, or encapsulate in a try/catch block

andrejtonev (approved):

Josipmrden (approved):

PR #3093, opened 2025-07-08
Compare timestamps when updating replica state and cluster is in STRICT_SYNC mode

as51340:


as51340:
Aborted this approach, epochs will be used instead

PR #3094, opened 2025-07-08
Rapidcheck Conan

as51340:
The problem is that rapidcheck_gtest tries to include target gtest::gtest which is not available. Rather gtest target is available.


as51340:
aborting this, there were a lot of changes in our build system in between

PR #3095, opened 2025-07-09
Text search improvements: durability, replication & various fixes

DavIvek:
Upcoming improvements:
- Durability fixes and improvements
- Replication support  
- Bug fix related to concurrent updates  
- Correct handling of boolean logic and uppercase letters in text search queries

Note for reviewer: I moved e2e text index tests into gql behave tests.


DavIvek:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Several durability-related issues have been fixed, and replication support has been added. Text search queries now correctly handle boolean logic and uppercase letters. Users can now safely recover data from databases without the experimental flag or with it enabled, into a database where both the experimental flag and text search are enabled. Replication can be used if the text search flag is enabled on both the main and replica. [#3095](https://github.com/memgraph/memgraph/pull/3095)
- [x] **[ Documentation PR link memgraph/documentation/pull/1368 ]**

src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -1290,11 +1290,21 @@ std::optional<storage::SingleTxnDeltasProcessingResult> InMemoryReplicationHandl
...
-        [&](WalTextIndexDrop const &) {
-          /* NOTE: Text search doesn’t have replication in scope yet (Phases 1 and 2)*/
+        [&](WalTextIndexDrop const &data) {
+          spdlog::trace("   Delta {}. Drop text search index {}.", current_delta_idx, data.index_name);
+          auto *transaction = get_replication_accessor(delta_timestamp, kUniqueAccess);
+          if (transaction->DropTextIndex(data.index_name).HasError()) {
+            throw utils::BasicException("Failed to drop text search index {}.", data.index_name);
+          }
         },
         [&](WalExistenceConstraintCreate const &data) {
```
as51340, line 1314: Can you remove these 2 comments from the file:
```
          // NOTE: Text search doesn’t have replication in scope yet (Phases 1 and 2)

```

src/query/interpreter.cpp
```diff
@@ -3949,49 +3949,41 @@ PreparedQuery PrepareTextIndexQuery(ParsedQuery parsed_query, bool in_explicit_t
...
   MG_ASSERT(current_db.db_acc_, "Text index query expects a current DB");
   auto &db_acc = *current_db.db_acc_;
 
   MG_ASSERT(current_db.db_transactional_accessor_, "Text index query expects a current DB transaction");
   auto *dba = &*current_db.execution_db_accessor_;
 
-  // Creating an index influences computed plan costs.
-  auto invalidate_plan_cache = [plan_cache = db_acc->plan_cache()] {
-    plan_cache->WithLock([&](auto &cache) { cache.reset(); });
-  };
```
as51340, line 3962: Why text index doesn't change plan costs?

src/query/interpreter.cpp
```diff
@@ -3949,49 +3949,41 @@ PreparedQuery PrepareTextIndexQuery(ParsedQuery parsed_query, bool in_explicit_t
...
-
   auto *storage = db_acc->storage();
-  auto label = storage->NameToLabel(text_index_query->label_.name);
+  auto label_name = text_index_query->label_.name;
+  auto label = storage->NameToLabel(label_name);
```
as51340: rename label to label_id to be more clear

src/query/db_accessor.hpp
```diff
@@ -513,12 +513,6 @@ class DbAccessor final {
 
   bool TextIndexExists(const std::string &index_name) const { return accessor_->TextIndexExists(index_name); }
 
-  void TextIndexAddVertex(const VertexAccessor &vertex) { accessor_->TextIndexAddVertex(vertex.impl_); }
-
-  void TextIndexUpdateVertex(const VertexAccessor &vertex, const std::vector<storage::LabelId> &removed_labels = {}) {
-    accessor_->TextIndexUpdateVertex(vertex.impl_, removed_labels);
-  }
-
```
as51340, line 521: With what methods did you replace these 2 functionalities?

src/storage/v2/durability/durability.cpp
```diff
@@ -307,8 +307,7 @@ void RecoverIndicesAndStats(const RecoveredIndicesAndConstraints::IndicesMetadat
   spdlog::info("Global edge property indices are recreated.");
 
   // Text idx
-  if (flags::AreExperimentsEnabled(flags::Experiments::TEXT_SEARCH)) {
-    // Recover text indices.
+  {
```
as51340: Extract this:
```
  if (!storage_dir.has_value()) {


          throw RecoveryFailure("There must exist a storage directory in order to recover text indices!");
```
outside the loop

src/storage/v2/durability/snapshot.cpp
```diff
@@ -2369,11 +2369,13 @@ RecoveredSnapshot LoadSnapshotVersion17(Decoder &snapshot, const std::filesystem
     }
 
     // Recover text indices.
-    if (flags::AreExperimentsEnabled(flags::Experiments::TEXT_SEARCH)) {
-      auto size = snapshot.ReadUint();
-      if (!size) throw RecoveryFailure("Couldn't recover the number of text indices!");
-      spdlog::info("Recovering metadata of {} text indices.", *size);
-      for (uint64_t i = 0; i < *size; ++i) {
+    // NOTE: while this was experimental and hence optional
```
as51340: since this was experimental?

src/storage/v2/indices/text_index.cpp
```diff
@@ -150,6 +174,7 @@ void TextIndex::LoadNodeToTextIndices(std::int64_t gid, const nlohmann::json &pr
 
   for (auto *index_context : applicable_text_indices) {
     try {
+      std::shared_lock lock(index_writer_muex_);
```
as51340: rename to index_writer_mutex_

src/storage/v2/indices/text_index.cpp
```diff
@@ -116,12 +112,40 @@ std::string TextIndex::StringifyProperties(const std::map<PropertyId, PropertyVa
   return utils::Join(indexable_properties_as_string, " ");
 }
 
-std::vector<mgcxx::text_search::Context *> TextIndex::GetApplicableTextIndices(
-    std::span<storage::LabelId const> labels) {
-  if (!flags::AreExperimentsEnabled(flags::Experiments::TEXT_SEARCH)) {
-    throw query::TextSearchDisabledException();
+std::string TextIndex::ToLowerCasePreservingBooleanOperators(std::string_view input) {
```
as51340, line 118: SerializeProperties method doesn't need to be templated, always receives NameIdMapper

src/storage/v2/indices/text_index.cpp
```diff
@@ -175,51 +201,56 @@ void TextIndex::CommitLoadedNodes(mgcxx::text_search::Context &index_context) {
...
+void TextIndex::UpdateNode(Vertex *vertex_after_update, NameIdMapper *name_id_mapper) {
   auto applicable_text_indices = GetApplicableTextIndices(vertex_after_update->labels);
   if (applicable_text_indices.empty()) return;
-  RemoveNode(vertex_after_update, applicable_text_indices);
+  RemoveNode(vertex_after_update, applicable_text_indices);  // In order to update the node, we first remove it.
   AddNode(vertex_after_update, name_id_mapper, applicable_text_indices);
 }
 
+void TextIndex::UpdateOnAddLabel(LabelId label, Vertex *vertex, NameIdMapper *name_id_mapper) {
+  if (!label_to_index_.contains(label)) return;
```
as51340: No need for this check, you are already doing this inside GetApplicableTextIndices

src/storage/v2/indices/text_index.cpp
```diff
@@ -241,7 +268,7 @@ void TextIndex::CreateIndex(std::string const &index_name, LabelId label, storag
     }
 
     auto vertex_properties = v.Properties(View::NEW).GetValue();
-    LoadNodeToTextIndices(v.Gid().AsInt(), SerializeProperties(vertex_properties, nameIdMapper),
+    LoadNodeToTextIndices(v.Gid().AsInt(), SerializeProperties(vertex_properties, name_id_mapper),
```
as51340: Is it maybe more logical to rename LoadNodeToTextIndices -> AddNodeToTextIndices

src/storage/v2/indices/text_index.cpp
```diff
@@ -175,51 +201,56 @@ void TextIndex::CommitLoadedNodes(mgcxx::text_search::Context &index_context) {
...
+  if (!label_to_index_.contains(label)) return;
+
+  auto applicable_text_indices = GetApplicableTextIndices(std::array{label});
+  if (applicable_text_indices.empty()) return;
+
+  AddNode(vertex, name_id_mapper, applicable_text_indices);
+}
+
+void TextIndex::UpdateOnRemoveLabel(LabelId label, Vertex *vertex) {
+  if (!label_to_index_.contains(label)) return;
```
as51340: same, I think you don't need that

src/storage/v2/indices/text_index.cpp
```diff
@@ -175,51 +201,56 @@ void TextIndex::CommitLoadedNodes(mgcxx::text_search::Context &index_context) {
...
+  RemoveNode(vertex, applicable_text_indices);
+}
+
+void TextIndex::UpdateOnSetProperty(PropertyId property, const PropertyValue &value, Vertex *vertex,
+                                    NameIdMapper *name_id_mapper) {
+  // TODO(@DavIvek): This will get extended to handle specific properties in the future.
+  UpdateNode(vertex, name_id_mapper);
+}
+
 void TextIndex::RemoveNode(
```
as51340: I think the interface of this method can be simplified. Instead of using std::optional's nullopt to signal there is no value, you can use empty std::vector to signal the same thing

src/storage/v2/indices/text_index.cpp
```diff
@@ -175,51 +201,56 @@ void TextIndex::CommitLoadedNodes(mgcxx::text_search::Context &index_context) {
 void TextIndex::AddNode(
     Vertex *vertex_after_update, NameIdMapper *name_id_mapper,
     const std::optional<std::vector<mgcxx::text_search::Context *>> &maybe_applicable_text_indices) {
```
as51340: I think the interface of this method can be simplified. Instead of using std::optional's nullopt to signal there is no value, you can use empty std::vector to signal the same thing

src/storage/v2/storage.cpp
```diff
@@ -681,19 +677,30 @@ void Storage::Accessor::MarkEdgeAsDeleted(Edge *edge) {
...
+    const std::string &index_name, LabelId label) {
   MG_ASSERT(type() == UNIQUE, "Creating a text index requires unique access to storage!");
-  auto *mapper = storage_->name_id_mapper_.get();
-  storage_->indices_.text_index_.CreateIndex(index_name, label, Vertices(View::NEW), mapper);
-  transaction_.md_deltas.emplace_back(MetadataDelta::text_index_create, index_name, label);
-  memgraph::metrics::IncrementCounter(memgraph::metrics::ActiveTextIndices);
+  try {
+    storage_->indices_.text_index_.CreateIndex(index_name, label, Vertices(View::NEW), storage_->name_id_mapper_.get());
+    transaction_.md_deltas.emplace_back(MetadataDelta::text_index_create, index_name, label);
+    memgraph::metrics::IncrementCounter(memgraph::metrics::ActiveTextIndices);
```
as51340: this 2 should logically be outside the try-catch loop

src/storage/v2/storage.cpp
```diff
@@ -681,19 +677,30 @@ void Storage::Accessor::MarkEdgeAsDeleted(Edge *edge) {
...
+utils::BasicResult<storage::StorageIndexDefinitionError, void> Storage::Accessor::DropTextIndex(
+    const std::string &index_name) {
   MG_ASSERT(type() == UNIQUE, "Dropping a text index requires unique access to storage!");
-  auto deleted_index_label = storage_->indices_.text_index_.DropIndex(index_name);
-  transaction_.md_deltas.emplace_back(MetadataDelta::text_index_drop, index_name, deleted_index_label);
-  memgraph::metrics::DecrementCounter(memgraph::metrics::ActiveTextIndices);
+  try {
+    auto deleted_index_label = storage_->indices_.text_index_.DropIndex(index_name);
+    transaction_.md_deltas.emplace_back(MetadataDelta::text_index_drop, index_name, deleted_index_label);
+    memgraph::metrics::DecrementCounter(memgraph::metrics::ActiveTextIndices);
```
as51340: this 2 should logically be outside the try-catch loop

tests/e2e/replication/replicate_text_index.py
```diff
@@ -0,0 +1,158 @@
...
+    yield
+    # Stop + delete directories after running the test
+    interactive_mg_runner.kill_all(keep_directories=False)
+
+
+def show_replicas_func(cursor):
+    def func():
+        return execute_and_fetch_all(cursor, "SHOW REPLICAS;")
+
+    return func
```
as51340: Use python's partial function application

tests/e2e/replication/replicate_text_index.py
```diff
@@ -0,0 +1,158 @@
...
+
+    # 2/
+    def get_show_index_info(cursor):
+        return execute_and_fetch_all(cursor, f"SHOW INDEX INFO;")
+
+    def get_replica_cursor(name):
+        return connection(BOLT_PORTS[name], "replica").cursor()
+
+    expected_result = [("text (name: test_index)", "Node", None, None)]
+    replica_1_enums = get_show_index_info(get_replica_cursor("replica_1"))
```
as51340: rename enums

tests/e2e/replication/replicate_text_index.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+def show_replicas_func(cursor):
+    def func():
+        return execute_and_fetch_all(cursor, "SHOW REPLICAS;")
+
+    return func
+
+
+def test_text_index_replication(connection, test_name):
```
as51340, line 53: create few nodes in the test, I don't see it here?

as51340 (changes requested):
looks good

src/storage/v2/indices/text_index.cpp
```diff
@@ -21,21 +20,18 @@
...
 
+namespace r = ranges;
 namespace memgraph::storage {
 
 std::string GetPropertyName(PropertyId prop_id, NameIdMapper *name_id_mapper) {
   return name_id_mapper->IdToName(prop_id.AsUint());
 }
 
-inline std::string TextIndex::MakeIndexPath(const std::string &index_name) {
+inline std::string TextIndex::MakeIndexPath(std::string_view index_name) {
```
as51340, line 30: I think ideally:
- these should be constants in the unnamed namespace:
```
static constexpr std::string_view kTextIndicesDirectory = "text_indices";

  // Boolean operators that should be preserved in uppercase for Tantivy
  static constexpr std::string_view kBooleanAnd = "AND";
  static constexpr std::string_view kBooleanOr = "OR";
  static constexpr std::string_view kBooleanNot = "NOT";
```
- `text_index_storage_dir_` can also be constant outside the class.
- this then allows you that MakeIndexPath is function in unnamed namespace in cpp file, to avoid polluting TextIndex class

as51340: make it const

src/storage/v2/indices/text_index.hpp
```diff
@@ -34,9 +35,20 @@ class TextIndex {
  private:
   static constexpr bool kDoSkipCommit = true;
   static constexpr std::string_view kTextIndicesDirectory = "text_indices";
+
```
as51340, line 39: I think you don't need struct TextIndexData.

map<string, Context> should be enough + the other map label->index_name

src/storage/v2/indices/text_index.cpp
```diff
@@ -21,21 +20,18 @@
 #include <span>
 #include <vector>
 
+namespace r = ranges;
```
as51340, line 24: Writing here random comment:
Do we ever check:
```
  document["metadata"]["deleted"] = false;
  document["metadata"]["is_node"] = true;
```

src/storage/v2/indices/text_index.cpp
```diff
@@ -21,21 +20,18 @@
...
 std::string GetPropertyName(PropertyId prop_id, NameIdMapper *name_id_mapper) {
   return name_id_mapper->IdToName(prop_id.AsUint());
 }
 
-inline std::string TextIndex::MakeIndexPath(const std::string &index_name) {
+inline std::string TextIndex::MakeIndexPath(std::string_view index_name) {
   return (text_index_storage_dir_ / index_name).string();
 }
 
 void TextIndex::CreateEmptyIndex(const std::string &index_name, LabelId label) {
```
as51340: Check why do we need metadata, all and data

src/storage/v2/indices/text_index.cpp
as51340: Also, just add comment which types of properties are indexable

src/storage/v2/indices/text_index.cpp
as51340: If mcxx depends on it, leave it. If we can live without it remove

as51340: check also if we need data and all

src/storage/v2/indices/text_index.cpp
as51340: Would it be better to take just once lock before the for loop. I am struggling to find counterexample why this would be wrong but the gut feeling is that this is not the best, would be better to atomically process all indices for one node.

Same for RemoveNode.

DavIvek: As discussed, I think shared lock is enough in this scenario, exclusive lock is needed only in commit and rollback: 
https://github.com/quickwit-oss/tantivy/blob/main/examples/index_from_multiple_threads.rs

src/storage/v2/indices/text_index.cpp
```diff
@@ -251,10 +278,6 @@ void TextIndex::CreateIndex(std::string const &index_name, LabelId label, storag
 void TextIndex::RecoverIndex(const std::string &index_name, LabelId label,
```
as51340, line 261: check with product:
- Why do we delete all durability files to recover from text index? Should be possible to load tantivy data.

src/storage/v2/indices/text_index.cpp
```diff
@@ -278,10 +301,6 @@ void TextIndex::RecoverIndex(const std::string &index_name, LabelId label,
 }
 
 LabelId TextIndex::DropIndex(const std::string &index_name) {
```
as51340, line 275: Probably you need to take unique lock for dropping index on tangy

src/storage/v2/indices/text_index.cpp
```diff
@@ -409,21 +420,15 @@ std::string TextIndex::Aggregate(const std::string &index_name, const std::strin
 }
 
 void TextIndex::Commit() {
-  if (!flags::AreExperimentsEnabled(flags::Experiments::TEXT_SEARCH)) {
-    throw query::TextSearchDisabledException();
-  }
-
   for (auto &[_, index_data] : index_) {
+    std::lock_guard lock(index_writer_muex_);
```
as51340: Check whether we need to take lock outside or inside the loop. Same for rollback

as51340: We probably don't need Commit and CommitLoadedNodes, check if do twice the work

as51340 (changes requested):

src/query/interpreter.cpp
DavIvek: It doesn't utilize the planner, so as I see it, it doesn't need to reset the plan cache or recalculate existing plans.

src/query/db_accessor.hpp
DavIvek: They are no longer used. They were previously needed because the text index was updated directly in the operators, but I’ve now changed that to use `UpdateOnAddLabel`, `UpdateOnSetProperty`, and `UpdateOnRemoveLabel` to be consistent with other indices and to keep everything in one place for easier maintenance.

src/storage/v2/indices/text_index.hpp
DavIvek: At the moment, it's used to retrieve the label by name in the `DropIndex` method. However, the label isn't actually needed, so this retrieval is currently unnecessary. The label is only fetched to be written to the WAL file, but the DROP INDEX operation shouldn't include the label in its WAL format, as the index is fully defined by its name.

I propose we leave it as-is for now and address it in an upcoming PR, where we'll be expanding text index support to arbitrary sets of properties. That change will require updating the durability files anyway, so it's better not to introduce a new durability version now, since it will soon be overridden.

src/storage/v2/indices/text_index.cpp
DavIvek: At the moment, it's implemented as follows:

- In the metadata field, we store values we want to retrieve from search results. Currently, the only value we care about is the `gid`, but this could change in the future if we decide to include transactional metadata.

- The data field stores JSON-formatted properties. These are used when querying specific properties, for example, if we want to search based on the title property of a vertex. This allows us to extract only the properties we care about from the JSON and run the query against them.

- The all field is used for full-text search. Here, all properties are stored as a single string, and queries are executed over the entire text content.

As I understand it, the JSON format is typically used for [schema-less data](https://github.com/quickwit-oss/tantivy/blob/main/doc/src/json.md), while the text option is generally more preferred for full-text search. In our case, we're using the JSON format to efficiently ignore properties that aren't relevant to the query.

src/storage/v2/indices/text_index.cpp
DavIvek: I think you're right that we can safely recover the text index using Tantivy's storage files, since we use `Index::open_or_create` in `mgcxx` (see: [link](https://github.com/memgraph/mgcxx/blob/main/text_search/src/lib.rs#L293)).

src/storage/v2/indices/text_index.cpp
as51340: Ok. I was wondering if we could remove serialising full string or JSON but seems we need both. All good

src/storage/v2/indices/text_index.cpp
as51340: yes, do we do that now or future?

src/dbms/dbms_handler.cpp
```diff
@@ -198,6 +197,8 @@ DbmsHandler::DbmsHandler(storage::Config config,
       locked_auth->DeleteDatabase(name);
       durability_->Delete(key);
     }
+    auto text_indices_dir = root / storage::kTextIndicesDirectory;
+    std::filesystem::remove_all(text_indices_dir);
```
as51340: Write a comment that we do a specific "recovery" of text indices and that this will be handled in the future.
Also, prefer the version of `remove_all` which takes error_code as argument. If some error occurred during removal, you want to handle it. Handling could mean different things: just logging, crashing... Not sure what's the best, what do you think?

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -1981,10 +1981,6 @@ mgp_error mgp_vertex_set_property(struct mgp_vertex *v, const char *property_nam
           return impl.SetProperty(prop_key, ToPropertyValue(*property_value, name_id_mapper));
         },
         v->impl);
-    if (memgraph::flags::AreExperimentsEnabled(memgraph::flags::Experiments::TEXT_SEARCH) && !result.HasError()) {
-      auto v_impl = v->getImpl();
-      v->graph->getImpl()->TextIndexUpdateVertex(v_impl);
-    }
```
as51340, line 1987: As I understand, we will preserve experimental flag. Do we need to return this then? Same for the whole file, there are multiple such deletions

src/storage/v2/durability/durability.cpp
```diff
@@ -449,13 +444,8 @@ void RecoverIndicesStatsAndConstraints(utils::SkipList<Vertex> *vertices, NameId
                                        RecoveredIndicesAndConstraints const &indices_constraints,
                                        bool properties_on_edges,
                                        std::optional<SnapshotObserverInfo> const &snapshot_info) {
-  auto storage_dir = std::optional<std::filesystem::path>{};
-  if (flags::AreExperimentsEnabled(flags::Experiments::TEXT_SEARCH)) {
-    storage_dir = config.durability.storage_directory;
-  }
-
```
as51340, line 456: Same, should be used again right?

src/storage/v2/durability/snapshot.cpp
```diff
@@ -2369,11 +2369,13 @@ RecoveredSnapshot LoadSnapshotVersion17(Decoder &snapshot, const std::filesystem
...
+    // NOTE: while this is experimental and hence optional
+    //       it must be last in the SECTION_INDICES
     if (flags::AreExperimentsEnabled(flags::Experiments::TEXT_SEARCH)) {
-      auto size = snapshot.ReadUint();
-      if (!size) throw RecoveryFailure("Couldn't recover the number of text indices!");
-      spdlog::info("Recovering metadata of {} text indices.", *size);
-      for (uint64_t i = 0; i < *size; ++i) {
+      auto size_opt = snapshot.ReadUint();
+      const auto size = size_opt.value_or(0);
+      spdlog::info("Recovering metadata of {} text indices.", size);
```
as51340, line 2377: This is different behavior now, we don't throw anymore. What's the reason for such change?

src/storage/v2/indices/text_index.cpp
```diff
@@ -21,21 +21,18 @@
...
 
-inline std::string TextIndex::MakeIndexPath(const std::string &index_name) {
+inline std::string TextIndex::MakeIndexPath(std::string_view index_name) const {
   return (text_index_storage_dir_ / index_name).string();
 }
 
-void TextIndex::CreateEmptyIndex(const std::string &index_name, LabelId label) {
-  if (!flags::AreExperimentsEnabled(flags::Experiments::TEXT_SEARCH)) {
-    throw query::TextSearchDisabledException();
-  }
```
as51340, line 37: where do you check this now? Interpreter.cpp right?

as51340 (changes requested):
Few comments, but otherwise looks good. Test the index recovery by restarting the instance.

src/dbms/dbms_handler.cpp
DavIvek: I think `LOG_FATAL` might be the best option here since it will call `std::terminate`. If we continue using the text index with `data_recovery_on_startup` set to `false` without first deleting the durability data, it could lead to inconsistent data in both the index and the database.

src/query/procedure/mg_procedure_impl.cpp
DavIvek: This function calls `VertexAccessor::SetProperty`, which in turn calls `storage_->indices_.UpdateOnSetProperty`. That method handles the text index along with other indices in the database, and this part is guarded by a check for the experimental flag.

src/storage/v2/durability/durability.cpp
DavIvek: I'm not entirely sure why `storage_dir` was used previously. As I understand it, it was passed into the `RecoverIndicesAndStats` function only for the text index. Later, the text index was checked for a value, and if it didn't have one, an exception was thrown. However, the value was only set if the experimental flag was enabled and that flag is already checked during the text index recovery stage within the same function.
I removed this part because it seems redundant and unused.

src/storage/v2/durability/snapshot.cpp
DavIvek: This was a bug that could manifest in the following scenario:
You have a database running with the experimental flag disabled, and then you create a snapshot. That snapshot won't include any information about the text index. (We could modify that behavior to write the size as 0, but this wouldn't help with snapshots created before the change.)
Later, if you try to recover the database with the experimental flag enabled, it will attempt to read the text index size. Since it wasn’t written in the snapshot, this will throw an error, and the database won’t be able to recover.

src/storage/v2/indices/text_index.cpp
DavIvek: Yes

as51340 (approved):
💪

PR #3099, opened 2025-07-09
Support librdkafka

as51340:
Kafka cannot be built with SASL, implicit-function-declaration


as51340:
Closing this, there were a lot of changes in between in our build system

PR #3104, opened 2025-07-10
mgconsole as tool requirement

as51340:
We don't need mgconsole as build req. Automatically we also don't need mgclient.


as51340:
Closing this, there were a lot of changes in between in our build system

PR #3110, opened 2025-07-14
Check durability_commit_timestamp and remove seq_num

as51340:
Remove seq_num from PrepareCommitHandler since it was unused and add durability_commit_timestamp cache.


andrejtonev:
The timestamp part seems fine. 
Why are we not sending the sequence number anymore? How can we guarantee that the chain is correct?

as51340:
> The timestamp part seems fine. Why are we not sending the sequence number anymore? How can we guarantee that the chain is correct?

I removed `seq_num` because we were never using it in PrepareCommitHandle (ex AppendDeltas). It was unused arg.

PR #3120, opened 2025-07-17
Add fix for used symbols in existential subqueries

Josipmrden:
Used symbols in existential subqueries were wrongly implemented.

PR #3129, opened 2025-07-23
Fix bug in MemoryDispatcher when nesting query modules

### LINKED ISSUE #1947: This query using experimental text search results in docker container stopping
Memgraph version: docker image memgraph/memgraph-mage:1.16-memgraph-2.16
Environment: Controlled by docker-compose.yml

### The bug
I'm attempting to use the experimental procedure text_search.search_all() in a rather complex query. When I run the query in Memgraph Lab, I get a message saying "Query failed: Connection was closed by server", and the docker container stops. The docker container does not log any errors through docker, nor are any errors written to the memgraph log file.

#### Setup:
Start up a new docker container (and new volumes) with experimental text search enabled. Connect to it with Memgraph Lab and run the following statements one by one in order:
```cypher
CREATE TEXT INDEX searchA ON :A
CREATE TEXT INDEX searchB ON :B
CREATE TEXT INDEX searchC ON :C
CREATE TEXT INDEX searchD ON :D
CREATE TEXT INDEX searchY ON :Y
CREATE TEXT INDEX searchZ ON :Z
CREATE (:A {searchMe: 'alphabet soup', isSecret: false})-[:EXAMPLE_RELATION]->(:Z {searchMe: "Banana Bread"})
CREATE (:B {searchMe: 'Banana Bread', isSecret: false})-[:EXAMPLE_RELATION]->(:Y {searchMe: "alphabet soup"})
```
#### Full query
This is similar to the query that I want to run, and it crashes the docker container every time.
parameters:
```
{
  "searchText": "soup"
}
```
query:
```cypher
CALL { CALL do.when($searchText = '',
  "MATCH (node)
    WHERE any(label IN labels(node) WHERE label IN ['A', 'B', 'C', 'D'])
  RETURN node, null AS y, null AS z",
  "CALL text_search.search_all('searchA', $searchText) YIELD node RETURN node, null AS y, null AS z
    UNION
  CALL text_search.search_all('searchB', $searchText) YIELD node RETURN node, null AS y, null AS z
    UNION
  CALL text_search.search_all('searchC', $searchText) YIELD node RETURN node, null AS y, null AS z
    UNION
  CALL text_search.search_all('searchD', $searchText) YIELD node RETURN node, null AS y, null AS z",
  {searchText: $searchText}
) YIELD value RETURN value
  UNION
CALL do.when($searchText = '',
  "RETURN null AS node, null AS y, null AS z",
  "CALL text_search.search_all('searchY', $searchText) YIELD node
  MATCH (node)--(mainNode)
  RETURN node AS y, mainNode AS node, null AS z
    UNION
  CALL text_search.search_all('searchZ', $searchText) YIELD node
  MATCH (node)--(mainNode)
  RETURN node AS z, null AS y, mainNode AS node",
  {searchText: $searchText}
) YIELD value RETURN value}
WITH value.node AS entry, head(labels(value.node)) AS label, value.y AS y, value.z AS z
  WHERE value.node IS NOT NULL
  AND NOT value.node.isSecret
WITH entry, label, 
  CASE WHEN z IS NULL THEN null ELSE {prop: z.foo, attr: z.bar} END AS fooBarZ,
  CASE WHEN y IS NULL THEN null ELSE {prop: y.foo, attr: y.bar} END AS fooBarY
WITH
  entry {.*, fooBarX, fooBarY, fooBarX: null} AS value,
  label AS type
WITH collect({value: value, type: type}) AS results,
    count(*) AS totalResults
UNWIND (CASE WHEN totalResults > 0 THEN results ELSE [null] END) AS result
WITH totalResults, result ORDER BY result.sortable ASC SKIP 0 LIMIT 50
RETURN collect(result) AS results, totalResults
```
At this point, after attempting the above query, my docker container is no longer running. I can start the same container back up and run simple queries fine, but the above query will always make it crash again.

#### Simplified query
I tried simplifying it down to figure out what part of the query was causing the problem, but the simpler I went the more sporadic the problem became. Here's an example of a query that (with the database set up the same way with those 4 nodes and 6 indexes) _can_ crash the docker container, but only does so _rarely_ (maybe two percent of the time).
parameters (same as before):
```
{
  "searchText": "soup"
}
```
query:
```cypher
CALL do.when($searchText = '',
"MATCH (node: A)
RETURN node",
"CALL text_search.search_all('searchA', $searchText) YIELD node RETURN node
  UNION
CALL text_search.search_all('searchB', $searchText) YIELD node RETURN node
  ",
{searchText: $searchText}
) YIELD value
WITH value.node AS entry
OPTIONAL MATCH (entry)<-[:CREATED]-(person:P)<-[:OTHER_INFO]-(otherInfo:Other)
WITH entry, CASE WHEN person IS NULL THEN null ELSE {firstName: person.firstName, lastName: person.lastName, email: person.email, info: otherInfo.id} END AS creator
WITH entry {.*, creator} AS value
WITH collect({value: value}) AS results,
     count(*) AS totalResults
UNWIND (CASE WHEN totalResults > 0 THEN results ELSE [null] END) AS result
WITH totalResults, result SKIP 0 LIMIT 50
RETURN collect(result) AS results, totalResults
```
If you're having trouble getting that query to produce the bug, then try running it a lot in quick succession right after the docker container starts up.

### Expected behavior
I expect the query to either work or give a meaningful error, and I expect the docker container to stay running regardless of whether the query worked or not.

### Side note
I know that text search is very experimental right now, but I was so excited to use it that I wanted to try it out in a realistic environment. It's understandable that it has some issues, and I just hope that my bug report can help you to fix the bugs and to know that you have at least one user that really wants this feature. Thank you for all your work.

DavIvek:
This PR aims to fix bug when query modules is called inside other query module.
Example:
```
CALL do.when(
  $searchText = '',
  "MATCH (node: A) RETURN node",
  "MATCH (node) RETURN label.exists(node, 'A') AS exists;",
  {searchText: $searchText}
) YIELD value
RETURN value
```

closes #1947

### DIFF
--- a/.github/workflows/diff_jepsen.yaml
+++ b/.github/workflows/diff_jepsen.yaml
@@ -66,15 +66,7 @@ jobs:
           --toolchain $TOOLCHAIN \
           --os $OS \
           --arch $ARCH \
-          copy --binary --dest-dir build
-
-      - name: Copy libmemgraph_module_support.so
-        run: |
-          ./release/package/mgbuild.sh \
-          --toolchain $TOOLCHAIN \
-          --os $OS \
-          --arch $ARCH \
-          copy --libs --dest-dir build/src/query
+          copy --binary --use-make-install --dest-dir build
 
       - name: Run unit tests for Jepsen code
         run: |
--- a/.github/workflows/reusable_release_tests.yaml
+++ b/.github/workflows/reusable_release_tests.yaml
@@ -816,15 +816,7 @@ jobs:
           --toolchain $TOOLCHAIN \
           --os $OS \
           --arch $ARCH \
-          copy --binary --dest-dir build
-
-      - name: Copy libmemgraph_module_support.so
-        run: |
-          ./release/package/mgbuild.sh \
-          --toolchain $TOOLCHAIN \
-          --os $OS \
-          --arch $ARCH \
-          copy --libs --dest-dir build/src/query
+          copy --binary --use-make-install --dest-dir build
 
       - name: Run unit tests for Jepsen code
         run: |
--- a/.github/workflows/reusable_stress_jepsen.yaml
+++ b/.github/workflows/reusable_stress_jepsen.yaml
@@ -84,11 +84,7 @@ jobs:
 
       - name: Copy memgraph binary
         run: |
-          ./release/package/mgbuild.sh --toolchain $TOOLCHAIN --os $OS --arch $ARCH copy --binary --dest-dir build
-
-      - name: Copy libmemgraph_module_support.so
-        run: |
-          ./release/package/mgbuild.sh --toolchain $TOOLCHAIN --os $OS --arch $ARCH copy --libs --dest-dir build/src/query
+          ./release/package/mgbuild.sh --toolchain $TOOLCHAIN --os $OS --arch $ARCH copy --binary --use-make-install --dest-dir build
 
       - name: Run unit tests for Jepsen code
         run: |
--- a/.github/workflows/stress_jepsen_ha.yaml
+++ b/.github/workflows/stress_jepsen_ha.yaml
@@ -14,7 +14,7 @@ on:
           - sync
           - strict_sync
           - strict_sync_mixed
-        default: sync_mixed
+        default: strict_sync
       run-mt:
         description: "Run MT tests"
         type: boolean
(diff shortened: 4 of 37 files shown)


DavIvek:
### Tracking
- [x] **https://github.com/memgraph/memgraph/issues/1947**

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Fix bug when query module is called inside another query module. Query modules can now be safely nested without causing execution errors. Users can now structure complex logic using nested query modules like `CALL do.when(...)` without workarounds. [#3120](https://github.com/memgraph/memgraph/pull/3129)

imilinovic (approved):

PR #3153, opened 2025-07-29
Fix flaky text index unit tests cleanup

DavIvek:
This PR removes flakiness from the text index unit test cleanup by not failing the test if the cleanup step doesn't succeed.
A more correct approach would be to improve the handling of this type of error in [mgcxx](https://github.com/memgraph/mgcxx), which will be added to the TODO list for the next mgcxx release.


tests/unit/text_index.cpp
```diff
@@ -133,21 +141,17 @@ TEST_F(TextIndexTest, ConcurrencyTest) {
...
+    std::vector<std::jthread> threads;
+    threads.reserve(index_size);
+    for (int i = 0; i < index_size; i++) {
+      threads.emplace_back([this, i](std::stop_token) {
+        auto acc = this->storage->Access();
+        [[maybe_unused]] const auto vertex = TextIndexTest_ConcurrencyTest_Test::CreateVertex(
+            acc.get(), "Title" + std::to_string(i), "Content for document " + std::to_string(i));
+        ASSERT_NO_ERROR(acc->PrepareForCommitPhase());
+      });
+    }
```
imilinovic, line 154: Not sure how much this tests concurrency since the threads start immediately after creating -> creating threads is slow + CreatingVertices is maybe too fast but I guess if it passed previous reviews its fine

DavIvek: Before proper locking was introduced on the text index side, this would fail almost every time, so I guess it serves a purpose.

PR #3155, opened 2025-07-30
Text index on subset of properties

DavIvek:
`CREATE TEXT INDEX index_name ON :Label(prop1, prop2, prop3);`


DavIvek:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Now it's possible to create text index on subset of properties using next syntax:
     `CREATE TEXT INDEX index_name ON :Label(prop1, prop2, prop3);` [#3155](https://github.com/memgraph/memgraph/pull/3155)
- [x] **[ Documentation PR link memgraph/documentation/pull/1368 ]**

src/query/db_accessor.hpp
```diff
@@ -696,9 +697,9 @@ class DbAccessor final {
     return accessor_->DropPointIndex(label, property);
   }
 
-  utils::BasicResult<storage::StorageIndexDefinitionError, void> CreateTextIndex(const std::string &index_name,
-                                                                                 storage::LabelId label) {
-    return accessor_->CreateTextIndex(index_name, label);
+  utils::BasicResult<storage::StorageIndexDefinitionError, void> CreateTextIndex(
+      storage::TextIndexSpec text_index_info) {
```
as51340: pass by const ref

src/storage/v2/disk/durable_metadata.cpp
```diff
@@ -145,23 +149,30 @@ bool DurableMetadata::PersistLabelPropertyIndexAndExistenceConstraintDeletion(La
...
 }
 
-bool DurableMetadata::PersistTextIndexCreation(const std::string &index_name, LabelId label) {
-  const std::string index_name_label_pair = index_name + "," + label.ToString();
+bool DurableMetadata::PersistTextIndexCreation(const storage::TextIndexSpec &text_index) {
+  std::string properties_str;
+  if (text_index.properties_ && !text_index.properties_->empty()) {
+    properties_str = utils::Join(
+        *text_index.properties_ | rv::transform([](const auto &property_id) { return property_id.ToString(); }), ",");
+  }
```
as51340: Use IIFE

src/storage/v2/disk/durable_metadata.cpp
```diff
@@ -145,23 +149,30 @@ bool DurableMetadata::PersistLabelPropertyIndexAndExistenceConstraintDeletion(La
...
+
+  const auto index_name_label_properties =
+      text_index.index_name_ + "," + text_index.label_.ToString() + "," + properties_str;
+
```
as51340: Use fmt::format for concatenation instead of +, it's more efficient

src/storage/v2/disk/durable_metadata.cpp
```diff
@@ -145,23 +149,30 @@ bool DurableMetadata::PersistLabelPropertyIndexAndExistenceConstraintDeletion(La
...
   if (auto text_index_store = durability_kvstore_.Get(kTextIndexStr); text_index_store.has_value()) {
-    std::string &value = text_index_store.value();
+    auto &value = text_index_store.value();
     value += "|";
-    value += index_name_label_pair;
+    value += index_name_label_properties;
```
as51340: Use fmt

src/storage/v2/indices/text_index_utils.hpp
```diff
@@ -0,0 +1,55 @@
...
+inline constexpr std::string_view kTextIndicesDirectory = "text_indices";
+inline constexpr bool kDoSkipCommit = true;
+
+// Boolean operators that should be preserved in uppercase for Tantivy
+inline constexpr std::string_view kBooleanAnd = "AND";
+inline constexpr std::string_view kBooleanOr = "OR";
+inline constexpr std::string_view kBooleanNot = "NOT";
+
+struct TextIndexSpec {
+  inline bool operator==(const TextIndexSpec &other) const = default;
```
as51340: no need for inline here, it will already be inlined

src/storage/v2/indices/text_index_utils.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+  std::string index_name_;
+  LabelId label_;
+  std::optional<std::vector<PropertyId>> properties_;
```
as51340: I think std::optional<std::vector> is unnecessarily complex modelling here. You can use empty vector for the case when there are no properties being indexed (better to say index on label used). 
Hence, std::vector should be enough, it will simplify your code a bit at multiple places

src/storage/v2/durability/wal.cpp
```diff
@@ -391,6 +392,19 @@ auto Decode(utils::tag_type<std::size_t> /*unused*/, BaseDecoder *decoder, const
   }
 }
 
+template <bool is_read>
+auto Decode(utils::tag_type<std::pair<std::string, std::string>> /*unused*/, BaseDecoder *decoder,
```
as51340: What does pair of strings represent?

src/storage/v2/durability/wal.hpp
```diff
@@ -234,8 +228,28 @@ struct WalUniqueConstraintCreate : LabelUnorderedPropertiesOpInfo {};
...
 struct WalTypeConstraintDrop : TypeConstraintOpInfo {};
-struct WalTextIndexCreate : TextIndexOpInfo {};
-struct WalTextIndexDrop : TextIndexOpInfo {};
+struct WalTextIndexCreate {
+  friend bool operator==(const WalTextIndexCreate &, const WalTextIndexCreate &) = default;
+  using ctr_types =
+      std::tuple<std::string, std::string, VersionDependant<kTextIndexWithProperties, std::vector<std::string>>>;
+  std::string index_name;
+  std::string label;
+  std::optional<std::vector<std::string>> properties;  //!< Optional properties, if not set, no properties are indexed
```
as51340, line 237: same for optional as I wrote above, I think you don't need it

src/storage/v2/durability/wal.hpp
```diff
@@ -234,8 +228,28 @@ struct WalUniqueConstraintCreate : LabelUnorderedPropertiesOpInfo {};
...
+};
+
+inline auto UpgradeDropTextIndexRemoveLabel(std::pair<std::string, std::string> name_and_label) -> std::string {
+  return name_and_label.first;  // Extract only the index_name, discard the label
+};
+using UpgradableDropTextIndex =
+    VersionDependantUpgradable<kTextIndexWithProperties, std::pair<std::string, std::string>, std::string,
+                               UpgradeDropTextIndexRemoveLabel>;
```
as51340, line 245: Can you explain this part?

src/storage/v2/durability/wal.hpp
```diff
@@ -234,8 +228,28 @@ struct WalUniqueConstraintCreate : LabelUnorderedPropertiesOpInfo {};
...
+
+struct WalTextIndexDrop {
+  friend bool operator==(const WalTextIndexDrop &, const WalTextIndexDrop &) = default;
+  using ctr_types = std::tuple<UpgradableDropTextIndex>;
+  std::string index_name;
+  std::optional<std::string> label;  //!< Optional label, only used for versions < kTextIndexWithProperties
+};
```
as51340, line 252: We can only drop index based on label? We cannot drop text index with properties? Is this fine, will then all indices with the specific label get deleted

src/storage/v2/indices/text_index_utils.hpp
DavIvek: I added `std::optional` here because I thought it made it clearer that this field is optional but I agree with you that it adds unnecessary complexity and makes the code harder to maintain, so I will change it.

src/storage/v2/durability/wal.cpp
DavIvek: For each new _type_ we add to the WAL structs in `wal.cpp`, we need to define how to decode it. To decode an index name and label pair, I had to add a decode function for a pair of strings.

src/storage/v2/durability/wal.hpp
DavIvek: We need this because of the way the WAL decoder is implemented. If the version is lower than `kTextIndexWithProperties`, a text index can still be created, but its properties will not be set.
```
// Generic helper decoder, please keep after the concrete type decoders
template <bool is_read, auto MIN_VER, typename Type>
auto Decode(utils::tag_type<VersionDependant<MIN_VER, Type>> /*unused*/, BaseDecoder *decoder, const uint64_t version)
    -> std::conditional_t<is_read, std::optional<Type>, void> {
  if (MIN_VER <= version) {
    return Decode<is_read>(utils::tag_t<Type>, decoder, version);
  }
  if constexpr (is_read) {
    return std::nullopt;
  }
}
```

src/storage/v2/durability/wal.hpp
DavIvek: Before this PR, the drop text index operation was defined by the index_name and the label on which we dropped the index. However, since an index is already uniquely identified by its index_name, we no longer need the label in the WAL files. Therefore, starting from this version, the label is not required. If we are recovering from old WAL files, the label will simply be ignored.

src/storage/v2/durability/wal.hpp
DavIvek: A drop index operation is defined by the index name. We cannot have two indexes with the same name, so only the index with the given name will be dropped. The label was present before, but it was neither used nor needed.

src/storage/v2/durability/wal.cpp
as51340: ok and how do they encode properties part?

src/storage/v2/durability/wal.hpp
as51340: ok but why empty vector cannot represent the state in which no properties are indexed?

src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -1296,7 +1297,16 @@ std::optional<storage::SingleTxnDeltasProcessingResult> InMemoryReplicationHandl
           spdlog::trace("   Delta {}. Create text search index {} on {}.", current_delta_idx, data.index_name,
```
as51340: Maybe you also want to log on which properties is text index created

src/storage/v2/durability/wal.hpp
DavIvek: It can, but the decoder is currently generic, `std::nullopt` is used to represent parts of the WAL that are not used in the current version. Changing this would require implementing a specific decoder just for this object.

as51340 (approved):

src/storage/v2/durability/wal.cpp
DavIvek: It will use:
```
template <bool is_read, typename T>
auto Decode(utils::tag_type<std::vector<T>> /*unused*/, BaseDecoder *decoder, const uint64_t /*version*/)
    -> std::conditional_t<is_read, std::vector<T>, void> {
  if constexpr (is_read) {
    const auto count = decoder->ReadUint();
    if (!count) throw RecoveryFailure(kInvalidWalErrorMessage);
    std::vector<T> values;
    values.reserve(*count);
    for (uint64_t i = 0; i < *count; ++i) {
      auto value = Decode<true>(utils::tag_t<T>, decoder, 0);
      values.emplace_back(std::move(value));
    }
    return values;
  } else {
    const auto count = decoder->ReadUint();
    if (!count) throw RecoveryFailure(kInvalidWalErrorMessage);
    for (uint64_t i = 0; i < *count; ++i) {
      Decode<false>(utils::tag_t<T>, decoder, 0);
    }
  }
}
```

PR #3163, opened 2025-08-04
Speed up building by using `ccache` in CI

mattkjames7:
The CI machines will all have a local `ccache` directory `~/.cache/ccache` which will be mounted as a volume in the `mgbuild` container. When the container is started, `ccache` should be installed (using `apt` or `yum`). This should speed up building once the cache is built up for the first time on each machine. The flag `--no-ccache` can be used to build without it, as will be done in the coverage workflow.

### DIFF
--- a/release/package/mgbuild.sh
+++ b/release/package/mgbuild.sh
@@ -1134,7 +1134,7 @@ case $command in
       done
 
       # Create ccache override file if ccache is enabled (same logic as run command)
-      local compose_files=$(setup_ccache_override)
+      compose_files=$(setup_ccache_override)
 
       if [[ "$os" == "all" ]]; then
         $docker_compose_cmd $compose_files down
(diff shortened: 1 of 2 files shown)


mattkjames7:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - N/A

release/package/mgbuild.sh
```diff
@@ -262,6 +264,58 @@ version_lt() {
...
+  fi
+}
+
+setup_host_ccache_permissions() {
+  if [[ "$ccache_enabled" == "true" ]]; then
+    echo "Setting up host ccache directory permissions..."
+    mkdir -p ~/.cache/ccache
+
+    # Set open permissions on the parent .cache directory to allow other tools to create subdirectories
+    chmod 777 ~/.cache || true
```
colinbarry: Just drawing your attention to this with the question: is 777 overly permissive?

colinbarry (approved):
All looks good to me. You've probably reduced the time it takes to run tests on CI by about 50%: brilliant!

The only thing I did notice is that we are using 777 as permissions: is this access actually needed, or can we tighten it up a bit?

release/package/mgbuild.sh
mattkjames7: I do not like this very much either, I think if I change it to `-R a+rwX` - that might still work, I'll do that now and see if diff and packaging still works...

mattkjames7: It looks like it worked, let's see if it breaks things after it has merged :fearful:

PR #3172, opened 2025-08-08
feat: Add support for RPC versioning

as51340:
The goal of this PR is to support RPC versioning for the sake of ISSU. RPC versioning is handled both on global level and on message level. On the global level, we use `LoadMessageHeader` to load only appropriate fields from the stream. The first field should always be global rpc version after which follows the sequence of protocol-specific fields. The current protocol specifies that after global rpc version follows message id and message version.

Any message (Request or Response) need to specify its version. When adding a new version for a RPC request or message, the developer should not delete old structures but rather rename them e.g. `PrepareCommitReq` -> `PrepareCommitReqV1` and `PrepareCommitRes` -> `PrepareCommitResV1`. The convention is that the newest version should be called `PrepareCommitReq` and `PrepareCommitRes`. In the new version of the request, the developer should add a static method `Upgrade` with the following declaration:
```
static PrepareCommitReq(PrepareCommitReqV1 const &old_req);
```

In the new version of the response, the developer should add a member method `Downgrade` with the following declaration:
```
PrepareCommitResV1 Downgrade() const;
```

When the server receives an RPC request, it reads its version and recursively upgrades the request until we get the request with the newest version. The server should always reply with the response which is of the same version as the request received. Thus, the developer should create a response with the newest version and then recursively downgrade it until the response with the version same as the request is obtained.

### DIFF
--- a/src/rpc/utils.hpp
+++ b/src/rpc/utils.hpp
@@ -26,7 +26,7 @@ concept HasDowngrade = requires(const std::remove_cvref_t<T> &res) {
 };
 
 template <RpcMessage TResponse>
-void SaveWithUpgrade(TResponse const &res, uint64_t const response_version, slk::Builder *builder) {
+void SaveWithDowngrade(TResponse const &res, uint64_t const response_version, slk::Builder *builder) {
   if (response_version == TResponse::kVersion) {
     slk::Save(res, builder);
     return;
@@ -34,7 +34,7 @@ void SaveWithUpgrade(TResponse const &res, uint64_t const response_version, slk:
   if constexpr (HasDowngrade<TResponse>) {
     auto prev_res = res.Downgrade();
     static_assert(decltype(prev_res)::kVersion == TResponse::kVersion - 1, "Wrong response version path");
-    SaveWithUpgrade(prev_res, response_version, builder);
+    SaveWithDowngrade(prev_res, response_version, builder);
   } else {
     throw std::runtime_error("No downgrade path available for this type of response");
   }
@@ -47,7 +47,7 @@ void SendFinalResponse(TResponse const &res, uint64_t const response_version, sl
                                              .message_id = TResponse::kType.id,
                                              .message_version = response_version};
   SaveMessageHeader(message_header, builder);
-  SaveWithUpgrade(res, response_version, builder);
+  SaveWithDowngrade(res, response_version, builder);
   builder->Finalize();
   spdlog::trace("[RpcServer] sent {}, version {}. {}", TResponse::kType.name, response_version, description);
 }


as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - **The support for RPC versioning is added in the context of enabling ISSU. The user can now expect no-downtime upgrades when following a specific procedure documented TODO. Before, when exchanging messages, the server would not be able to read RPC request if not of the latest version. Additionally, support is added for updating protocol in the future releases. [#3172](https://github.com/memgraph/memgraph/pull/3172)**
- [x] **[ Documentation PR link ](https://github.com/memgraph/documentation/pull/1389)**

src/rpc/client.hpp
```diff
@@ -158,46 +159,40 @@ class Client {
...
-          slk::Load(&res_id, &res_reader);
-          slk::Load(&version, &res_reader);
-        } catch (const slk::SlkReaderException &) {
+        if (!maybe_message_header.has_value()) {
           self_->client_->ShiftData(response_data_size);
           throw SlkRpcFailedException();
+          ;
         }
 
-        if (version != rpc::current_version) {
```
andrejtonev, line 173: we still want rpc protocol version check

src/rpc/client.hpp
```diff
@@ -244,36 +240,32 @@ class Client {
...
 
-      try {
-        slk::Load(&res_id, &res_reader);
-        slk::Load(&version, &res_reader);
-      } catch (const slk::SlkReaderException &) {
+      if (!maybe_message_header.has_value()) {
         throw SlkRpcFailedException();
-      }
-
-      if (version != rpc::current_version) {
```
andrejtonev, line 258: version check

src/rpc/utils.hpp
```diff
@@ -13,28 +13,73 @@
...
-  slk::Save(rpc::current_version, builder);
-  slk::Save(res, builder);
+template <typename T>
+concept HasDowngrade = requires(const std::remove_cvref_t<T> &res) {
+  {res.Downgrade()};
+  requires RpcMessage<T> && RpcMessage<std::remove_cvref_t<decltype(res.Downgrade())>>;
+};
+
+template <RpcMessage TResponse>
+void SaveWithUpgrade(TResponse const &res, uint64_t const response_version, slk::Builder *builder) {
```
andrejtonev: SaveWithDowngrade I guess

andrejtonev (commented):
Look good. A couple of small comments

andrejtonev (approved):

src/rpc/client.hpp
as51340: Discussed on the call, this is handled with LoadMessageHeader

src/rpc/client.hpp
as51340: Discussed on the call, this is handled with LoadMessageHeader

PR #3174, opened 2025-08-09
Improve handling of epochs during data recovery

as51340:
When recovering from snapshot, there is no need to modify epoch history with the epoch from snapshot since this will modify history's state compared to the moment when snapshot is being taken.
There was a bug in one of the previous PR's in which epochs and ldt from wal would update the state even if there was no change made in that WAL (all txns were aborted).


as51340:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - **Bugfix: When recovering from snapshot, we would add into epoch history the epoch from the time when the snapshot was created although the history shouldn't contain it. This could've caused unnecessary force resets of instances. [#3174](https://github.com/memgraph/memgraph/pull/3174)** 

as51340:
> Please also check if the history is forced when the snapshot it created

Checked on the call, looks good

PR #3171, opened 2025-08-08
Start using the latest mgcxx version

DavIvek:
Move mgcxx out of toolchain and use the latest version.

PR #3190, opened 2025-08-18
[Temporary] Experiment to fix clang-tidy mgcxx include path on CI

colinbarry:
This is a temporary PR so that David and I can both work on trying to fix path issues on CI (#3171) in parallel.

Once we've solved the issue, the fix will be cherry-picked into 3171 and this PR deleted.


colinbarry:
Closing PR and branch, as this was an experiment. David solved it in the main PR.

PR #3196, opened 2025-08-20
Use Python `uv` to install tests dependencies

mattkjames7:
Python `uv` is considerably faster at managing test dependencies than `pip` alone is. This PR aims to speed up CI a little by:
- using `uv` to manage test environments
- refactor `tests/setup.sh` because it was a bit odd
- separate test prep from `init` script so that the build happens prior to setting up tests for faster feedback on build issues
- create shared cache for `pip ` on CI machines

Tested locally (before and after switching to `uv`) the `init-test` script takes:
|         |   `pip`  |   `uv`   |
|:--------|:--------:|:--------:|
|cached   |  2m 30s  |   4.4s   |
|uncached |  3m 37s  |    15s   |


mattkjames7:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - N/A

release/package/mgbuild.sh
```diff
@@ -938,8 +987,12 @@ while [[ $# -gt 0 ]]; do
         shift 2
     ;;
     --no-ccache)
-        ccache_enabled="false"
-        shift 1
+      ccache_enabled="false"
```
antejavor, line 990: `ccache_enabled` should this be `cache_enabled`?

init-test
```diff
@@ -0,0 +1,79 @@
...
+          # unknown option
+          echo "Invalid argument provided: $1"
+          print_help
+          exit 1
+      ;;
+    esac
+done
+
+DISTRO=$(operating_system)
+ARCHITECTURE=$(architecture)
```
antejavor: Looks like `ARHITECTURE` is not used here?

init-test
```diff
@@ -0,0 +1,79 @@
...
+    pushd $1 > /dev/null
+    echo "Setting up virtualenv for: $1"
+
+    # remove old virtualenv
+    if [ -d ve3 ]; then
+        rm -rf ve3
+    fi
+
+    # create new virtualenv using uv
+    uv venv ve3 || exit 1
```
antejavor, line 22: Is it better to let it fail without `exit 1`, then you get more comprehensive log output, since this fails faily default, the same goes for the ` uv pip install -r requirements.txt || exit 1`. 

If you get the same output, then my comment is pointless and I am hallucinating :D

antejavor (approved):
A few minor comments.

release/package/mgbuild.sh
mattkjames7: Not in this case, I'm not sure what, if anything, has changed on this line - that's from enabling the shared `ccache`.

init-test
mattkjames7: This appears to be from the old `init` script - I guess to ensure that it exits straight away after a fault

init-test
mattkjames7: Well spotted, I have removed it!

PR #3121, opened 2025-07-18
Support zoned datetimes in query modules

colinbarry:
Add support for `ZonedDateTime` types in query modules.

- `ZonedDateTime` with a timezone (either numeric offset or named time identifier) can be passed to query modules.
- Any `datetime`s returned by a Python query module will be converted into a `ZonedDateTime` if they have a valid `timezone` `tzinfo`, otherwise, they will be treated as `LocalDateTime`s.

PR #3200, opened 2025-08-20
fix: Use shim for PyDateTime_DATE_GET_TZINFO to support Python 3.9

colinbarry:
`PyDateTime_DATE_GET_TZINFO` is only supported in Python 3.10 and above. As we currently still support Python 3.9, this shim function does exactly the same thing, without needing the macro.

Delivered as part of [#3121](https://github.com/memgraph/memgraph/pull/3121) - see that PR for changelog, docs, etc.

### DIFF
--- a/src/query/procedure/py_module.cpp
+++ b/src/query/procedure/py_module.cpp
@@ -55,7 +55,8 @@ PyObject *INTERNAL_PyDateTime_DATE_GET_TZINFO(PyObject *obj) {
 
     if (PyDateTime_Check(obj)) {
       return (reinterpret_cast<PyDateTime_DateTime *>(obj))->tzinfo;
-    } else if (PyTime_Check(obj)) {
+    }
+    if (PyTime_Check(obj)) {
       return (reinterpret_cast<PyDateTime_Time *>(obj))->tzinfo;
     }
   }


colinbarry:
Delivered as part of [#3121](https://github.com/memgraph/memgraph/pull/3121)

src/query/procedure/py_module.cpp
```diff
@@ -3025,7 +3042,7 @@ mgp_value *PyObjectToMgpValue(PyObject *o, mgp_memory *memory) {
             PyDateTime_DATE_GET_MICROSECOND(o) %  // NOLINT(cppcoreguidelines-pro-type-cstyle-cast,hicpp-signed-bitwise)
             1000};
 
-    if (PyObject *tzinfo = PyDateTime_DATE_GET_TZINFO(o); !Py_IsNone(tzinfo)) {
+    if (PyObject *tzinfo = INTERNAL_PyDateTime_DATE_GET_TZINFO(o); !Py_IsNone(tzinfo)) {
```
mattkjames7: `error: use of undeclared identifier 'Py_IsNone'`
:joy:

colinbarry: Would you believe, `Py_IsNone` is Python 3.10 only as well. I've replaced with a raw compare.

mattkjames7 (approved):

PR #3209, opened 2025-08-26
feat: Use num committed txns for the failover

as51340:
Since we now have the information about the number of committed transactions, we want to use it for the failover decision since it provides us with the more precise information about the state in which each replica is. 
`GetDatabaseHistoriesRpc` is updated with the new version but the upgrade path is not "by the book" since we cannot construct ldt by knowing num_committed_txns.

The coordinator setting 'max_replica_lag' is by default set to std::2**32 - 1 and can be updated by using `SET COORDINATOR SETTING` query. This setting forbids failovering to the instance whose any database is behind the current main for more than N (setting's value) transactions.

The leader coordinator caches in the memory last number of committed transactions for each database of the current main. The cache is updated with `StateCheckRpc` so the information the current leader has is correct up to the last M seconds where M is equal to `--instance-health-check-frequence-sec`.

In the scenario in which we perform failover on coordinators' side (hence elect the new cluster leader) which is unable to contact main instance, it is possible that the new leader will try to perform the data failover with its cache being empty because the leader never received the response to `StateCheckRpc`. In this unlikely scenario, we will allow failovering to any replica.

TODO: (andi) Do we want to handle this?


as51340:
Work continuing under https://github.com/memgraph/memgraph/pull/3219

PR #3219, opened 2025-08-27
feat: Use num committed txns for failover and use replication lag as input for the failover decision

as51340:
Since we now have the information about the number of committed transactions, we want to use it for the failover decision since it provides us with the more precise information about the state in which each replica is. 
`GetDatabaseHistoriesRpc` is updated with the new version but the upgrade path is not "by the book" since we cannot construct ldt by knowing num_committed_txns.

The coordinator setting 'max_replica_lag' is by default set to 2**64 - 1 and can be updated by using `SET COORDINATOR SETTING` query. This setting forbids failovering to the instance whose any database is behind the current main for more than N (setting's value) transactions.

The leader coordinator caches in the memory last number of committed transactions for each database of the current main. The cache is updated with `StateCheckRpc` so the information the current leader has is correct up to the last M seconds where M is equal to `--instance-health-check-frequence-sec`.

In the scenario in which we perform failover on coordinators' side (hence elect the new cluster leader) which is unable to contact main instance, it is possible that the new leader will try to perform the data failover with its cache being empty because the leader never received the response to `StateCheckRpc`. In this unlikely scenario, we will allow failovering to any replica.


as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - **A new coordinator setting `max_failover_replica_lag` has been added which allows users to control how many transactions are they willing to risk to lose upon a data failover. Basically, if the replica is behind main for more than `max_failover_replica_lag` transactions, the coordinator will forbid failover to such an instance. [#3219](https://github.com/memgraph/memgraph/pull/3219)**   
- [x] **[Documentation PR link](https://github.com/memgraph/documentation/pull/1403)**

src/coordination/coordinator_instance.cpp
```diff
@@ -1246,9 +1278,41 @@ auto CoordinatorInstance::GetInstanceForFailover() const -> std::optional<std::s
...
+
+        if (main_db_info->second > replica_db_info.num_committed_txns &&
+            main_db_info->second - replica_db_info.num_committed_txns > max_allowed_lag) {
+          spdlog::info(
+              "Instance {} won't be used in a failover because it's much behind the current main. Main has committed "
+              "{} txns, while instance {} has committed {} txns for the database {}",
+              instance_name, main_db_info->second, instance_name, replica_db_info.num_committed_txns,
+              main_db_info->first);
+          replica_behind = true;
+          break;
```
andrejtonev, line 1285: Why break here?

src/coordination/coordinator_instance.cpp
```diff
@@ -1246,9 +1278,41 @@ auto CoordinatorInstance::GetInstanceForFailover() const -> std::optional<std::s
...
+        auto const main_db_info = main_num_txns_cache_.find(replica_db_info.db_uuid);
+        // If database got deleted on main but that change still isn't replicated, we cannot conclude that replica is
+        // too far behind using such a condition
+        if (main_db_info == main_num_txns_cache_.end()) {
+          spdlog::trace("No entry for db: {}", replica_db_info.db_uuid);
+          continue;
+        }
+
+        if (main_db_info->second > replica_db_info.num_committed_txns &&
+            main_db_info->second - replica_db_info.num_committed_txns > max_allowed_lag) {
```
andrejtonev: unsigned might underflow and trigger this.
maybe a failover happened, main is actually further behind than replica

src/coordination/include/coordination/coordinator_slk.hpp
```diff
@@ -47,6 +47,17 @@ inline void Load(ReplicationClientInfo *obj, Reader *reader) {
   Load(&obj->replication_server, reader);
 }
 
+// InstanceInfo
+inline void Save(const replication_coordination_glue::InstanceInfoV1 &obj, Builder *builder) {
```
andrejtonev, line 51: What's the difference between v1 and v2?

src/coordination/include/coordination/coordinator_rpc.hpp
```diff
@@ -178,29 +177,62 @@ struct EnableWritingOnMainRes {
...
   static void Load(GetDatabaseHistoriesRes *self, memgraph::slk::Reader *reader);
   static void Save(const GetDatabaseHistoriesRes &self, memgraph::slk::Builder *builder);
 
   explicit GetDatabaseHistoriesRes(replication_coordination_glue::InstanceInfo instance_info)
       : instance_info(std::move(instance_info)) {}
   GetDatabaseHistoriesRes() = default;
 
+  // We cannot downgrade from GetDatabaseHistoriesRes, the caller should provide function for creating both responses
+  // independently
+  GetDatabaseHistoriesResV1 Downgrade() = delete;
```
andrejtonev, line 238: Can find where the logic is different.
I assume you would need to have a duplicate logic for failover

src/coordination/coordinator_instance.cpp
as51340: it means I found at least one db for which replica is too far behind main so I don't need to check other databases, I know that this instance cannot become new main.

src/coordination/coordinator_instance.cpp
as51340: yes but that's why I am checking if main's number of committed txns is larger than replica's. If it is, reduction cannot underflow

src/coordination/include/coordination/coordinator_slk.hpp
as51340: v1 has std::vector of InstanceDBInfoV1  while v2 has std::vector of InstanceDBInfo

src/coordination/include/coordination/coordinator_rpc.hpp
as51340: yes, there is no logic for downgrade, but there is a switch depending on what you need. It is in `GetDatabaseHistories` 

```

std::variant<coordination::GetDatabaseHistoriesResV1, coordination::GetDatabaseHistoriesRes>
ReplicationHandler::GetDatabasesHistories(uint64_t const request_version) const {
  if (request_version == coordination::GetDatabaseHistoriesReqV1::kVersion) {
    replication_coordination_glue::InstanceInfoV1 results;
    results.last_committed_system_timestamp = system_.LastCommittedSystemTimestamp();
    dbms_handler_.ForEach([&results](dbms::DatabaseAccess db_acc) {
      auto const &repl_storage_state = db_acc->storage()->repl_storage_state_;
      results.dbs_info.emplace_back(std::string{db_acc->storage()->uuid()},
                                    repl_storage_state.commit_ts_info_.load(std::memory_order_acquire).ldt_);
    });
    return coordination::GetDatabaseHistoriesResV1{results};
  }

  // In the newest version we return num_committed_txns_ instead
  replication_coordination_glue::InstanceInfo results;
  results.last_committed_system_timestamp = system_.LastCommittedSystemTimestamp();
  dbms_handler_.ForEach([&results](dbms::DatabaseAccess db_acc) {
    auto const &repl_storage_state = db_acc->storage()->repl_storage_state_;
    results.dbs_info.emplace_back(
        std::string{db_acc->storage()->uuid()},
        repl_storage_state.commit_ts_info_.load(std::memory_order_acquire).num_committed_txns_);
  });

  return coordination::GetDatabaseHistoriesRes{results};
}
```

src/coordination/coordinator_instance.cpp
Ignition: C++26 std::sub_sat

src/coordination/coordinator_instance.cpp
```diff
@@ -948,30 +948,50 @@ auto CoordinatorInstance::AddCoordinatorInstance(CoordinatorInstanceConfig const
...
     return SetCoordinatorSettingStatus::UNKNOWN_SETTING;
   }
 
-  bool const value = utils::ToLowerCase(setting_value) == "true"sv;
-  CoordinatorClusterStateDelta delta_state;
+  auto const maybe_val_with_type =
+      std::invoke([setting_name, setting_value]() -> std::optional<std::variant<bool, uint64_t>> {
+        if (setting_name == kMaxFailoverLagOnReplica) {
+          try {
+            return std::stoul(setting_value.data());
```
Ignition: Prefer from_chars

src/coordination/coordinator_instance.cpp
```diff
@@ -948,30 +948,50 @@ auto CoordinatorInstance::AddCoordinatorInstance(CoordinatorInstanceConfig const
...
+    CoordinatorClusterStateDelta ret_delta_state;
+    if (setting_name == kEnabledReadsOnMain) {
+      ret_delta_state.enabled_reads_on_main_ = std::get<bool>(val_with_type);
+    } else if (setting_name == kSyncFailoverOnly) {
+      ret_delta_state.sync_failover_only_ = std::get<bool>(val_with_type);
+    } else {
+      ret_delta_state.max_failover_replica_lag_ = std::get<uint64_t>(val_with_type);
+    }
+    return ret_delta_state;
+  });
```
Ignition: maybe_val_with_type and this could be combined to make this code easier to follow/extend.

PR #3221, opened 2025-08-27
Build improvements 

Ignition:


Ignition:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist

CMakeLists.txt
```diff
@@ -15,18 +15,11 @@ endif()
...
-  set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ccache)
-  set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK ccache)
-  message(STATUS "CCache: Used")
-else ()
-  message(STATUS "CCache: Not used")
-endif(CCACHE_FOUND AND USE_CCACHE)
+# ccache setup with optimized configuration
+option(USE_CCACHE "Use ccache for faster rebuilds" ON)
+if(USE_CCACHE AND NOT TEST_COVERAGE)
+  include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/ccache.cmake)
```
mattkjames7, line 21: Not an issue, more out of curiosity: what is the reason behind not using ccache on the coverage tests?

libs/setup.sh
```diff
@@ -49,39 +49,43 @@ clone () {
...
+        # errors.
+        git cherry-pick -n "$cherry_pick_id" || exit 1
+      done
+
+      # Reapply any local changes - only pop if we actually stashed successfully.
+      if [[ $local_changes == true && $needs_pop == true ]]; then
+        git stash pop
+      fi
+    fi
+  )
```
mattkjames7, line 133: Do we want the failed exit status to propagate out of the subshell?
If so, I think we need to add ` || return 1`  or ` || exit 1`

CMakeLists.txt
Ignition: Its because the coverage build also output gcno files that get cached. And there is a concern about absolute/relative directories encoded in them.

I think its fine, but just in case did the `NOT TEST_COVERAGE`. But also we could use `gcno_cwd` and `-fcoverage-prefix-map=${PROJECT_SOURCE_DIR}=.`

libs/setup.sh
Ignition: good point, I've added more robustness to clone

CMakeLists.txt
Ignition: I added in the `gcno_cwd` and `-fcoverage-prefix-map=${PROJECT_SOURCE_DIR}=.` approach

Ignition: Actually I was unable to make the ccache changes work with coverage. I'll not deliver ccache changes yet in the PR. Will come back to ccache improvement.

src/utils/rocksdb_serialization.hpp
```diff
@@ -13,7 +13,7 @@
 
 #include "storage/v2/edge.hpp"
 #include "storage/v2/property_store.hpp"
-#include "storage/v2/vertex.hpp"
+//#include "storage/v2/vertex.hpp"
```
colinbarry: Assuming we don't need this header, remove the include completely rather than comment out.

colinbarry (approved):
All good: approved.

PR #3227, opened 2025-09-01
Debug HA memory

as51340:


as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - **Durability files that are transferred between instances in HA environment won't require anymore a big memory allocation on the receiver's end. Before, whole file had to be read into memory in order to be used, now we use a buffer of 64KiB to write bytes received on the network to the file on disk. Users with a big scale shouldn't see any memory spikes on replicas. This should help with Memgraph's stability when deployed in environments with a tight memory control, e.g K8s [#3227](https://github.com/memgraph/memgraph/pull/3227)**

src/rpc/protocol.cpp
```diff
@@ -26,32 +27,119 @@
...
+  spdlog::warn("Data starts at address: {}", static_cast<const void *>(data));
+  slk::Reader req_reader(data, size);
+  spdlog::warn("Reader pos before creating decoder: {}", req_reader.GetPos());
+  storage::replication::Decoder decoder(&req_reader);
+  spdlog::warn("Reader pos after creating decoder: {}", req_reader.GetPos());
+
+  // We should be able to always read filename and filesize since file data starts with the new segment
+  auto const maybe_filename = decoder.ReadString();
+  MG_ASSERT(maybe_filename, "Filename missing for the received file over the RPC");
+  auto const path = kTempDirectory / *maybe_filename;
```
Ignition: Are we guaranteed to be single threaded here? Is it possible for two RPC with same `kTempDirectory / *maybe_filename`?

src/coordination/coordinator_instance_management_server_handlers.cpp
```diff
@@ -16,12 +16,17 @@
...
+class FileReplicationHandler;
+}  // namespace memgraph::rpc
+
 namespace memgraph::coordination {
 
 void CoordinatorInstanceManagementServerHandlers::Register(CoordinatorInstanceManagementServer &server,
                                                            CoordinatorInstance const &coordinator_instance) {
   server.Register<coordination::ShowInstancesRpc>(
-      [&](uint64_t const request_version, slk::Reader *req_reader, slk::Builder *res_builder) -> void {
+      [&](std::optional<rpc::FileReplicationHandler> const & /*file_replication_handler*/,
```
Ignition, line 28: is there a better way to pass around this `FileReplicationHandler`? I've see that it is optional and unused in multiple callbacks so far.

src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -508,17 +521,23 @@ void InMemoryReplicationHandlers::SnapshotHandler(DbmsHandler *dbms_handler,
   }
   auto const &curr_wal_files = *maybe_curr_wal_files;
 
-  storage::replication::Decoder decoder(req_reader);
-  const auto maybe_recovery_snapshot_path = decoder.ReadFile(current_snapshot_dir);
-  if (!maybe_recovery_snapshot_path.has_value()) {
-    spdlog::error("Failed to load snapshot from {}", current_snapshot_dir);
+  auto const &active_files = file_replication_handler.GetActiveFileNames();
+
+  MG_ASSERT(active_files.size() == 1, "Received {} snapshot files but expecting only one!", active_files.size());
```
Ignition, line 525: We should not use MG_ASSERT, we need to gracefully fail the RPC without killing the REPLICA.

src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -508,17 +521,23 @@ void InMemoryReplicationHandlers::SnapshotHandler(DbmsHandler *dbms_handler,
...
+  auto const snapshot_file_name = active_files[0];
+  auto const src_snapshot_file = std::filesystem::temp_directory_path() / "memgraph" /
```
Ignition: Will this have a potential concurrent resource issue. Two RPC working on same filename.

src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -508,17 +521,23 @@ void InMemoryReplicationHandlers::SnapshotHandler(DbmsHandler *dbms_handler,
...
+                                 storage::durability::kReplicaDurabilityDirectory / snapshot_file_name;
+
+  auto const dst_snapshot_file = current_snapshot_dir / snapshot_file_name;
+
+  if (!utils::CopyFile(src_snapshot_file, dst_snapshot_file)) {
```
Ignition: Copy not move?

src/dbms/inmemory/replication_handlers.cpp
Ignition: Can't `active_files` give you path information, rather than building a path from a filename.

src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -765,11 +787,9 @@ void InMemoryReplicationHandlers::CurrentWalHandler(dbms::DbmsHandler *dbms_hand
     }
   }
 
-  storage::replication::Decoder decoder(req_reader);
-
   // Even if loading wal file failed, we return last_durable_timestamp to the main because it is not a fatal error
   // When loading a single WAL file, we don't care about saving number of deltas
-  auto const load_wal_res = LoadWal(storage, &decoder, res_builder);
+  auto const load_wal_res = LoadWal(file_replication_handler.GetActiveFileNames()[0], storage, res_builder);
```
Ignition: is `[0]` safe?

src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -797,30 +817,22 @@ void InMemoryReplicationHandlers::CurrentWalHandler(dbms::DbmsHandler *dbms_hand
...
 // If WAL file doesn't contain any new changes, we ignore it and consider WAL file as successfully applied.
-InMemoryReplicationHandlers::LoadWalStatus InMemoryReplicationHandlers::LoadWal(storage::InMemoryStorage *storage,
-                                                                                storage::replication::Decoder *decoder,
+InMemoryReplicationHandlers::LoadWalStatus InMemoryReplicationHandlers::LoadWal(std::string const &wal_file_name,
+                                                                                storage::InMemoryStorage *storage,
                                                                                 slk::Builder *res_builder,
                                                                                 uint32_t start_batch_counter) {
-  const auto temp_wal_directory =
-      std::filesystem::temp_directory_path() / "memgraph" / storage::durability::kWalDirectory;
+  auto const rpc_dir =
```
Ignition: again building path, can't we be given std::filesystem::path?

src/rpc/client.hpp
```diff
@@ -408,8 +408,11 @@ class Client {
     ProtocolMessageHeader const message_header{.protocol_version = current_protocol_version,
                                                .message_id = req_type.id,
                                                .message_version = TRequestResponse::Request::kVersion};
+    // spdlog::warn("Position before saving header: {}", handler.GetBuilder()->GetPos());
```
Ignition: delete commented code

src/storage/v2/replication/serialization.cpp
```diff
@@ -79,10 +82,15 @@ bool Encoder::WriteFile(const std::filesystem::path &path) {
     return false;
   }
   const auto &filename = path.filename().generic_string();
+  // spdlog::trace("Position before saving filename: {}", builder_->GetPos());
```
Ignition: delete commented code

src/coordination/coordinator_instance_management_server_handlers.cpp
as51340: It's used for SnapshotHandler, WalFilesHandler and CurrentWalHandler. We don't have any other mechanism to pass objects from the protocol handler to request handler except by directly sending it as part of the function

src/dbms/inmemory/replication_handlers.cpp
as51340: Agree but not in this case. If this is broken, then whole cluster is severely broken. This should never-ever happen, it would be logically very wrong. 
It's not like e.g try to create a directory on disk and if it fails, fail the RPC and try again. If this happens, I want user to see that something is broken

src/dbms/inmemory/replication_handlers.cpp
as51340: this was working exactly the same previously. There could be problems in the future if we add multiple threads on the RPC server. Currently, processing is single-threaded.

src/rpc/protocol.cpp
as51340: we are guaranteed but in order to be safer, I now added random path generation.

src/dbms/inmemory/replication_handlers.cpp
as51340: it is + now I added mg_assert. It would be a big logical mistake if there is not exactly 1 file

src/rpc/file_replication_handler.cpp
```diff
@@ -0,0 +1,96 @@
...
+  if (!file_.IsOpen()) {
+    return 0;
+  }
+
+  size_t processed_bytes{0};
+
+  auto to_write = std::min(size, file_size_ - written_);
+  while (to_write > 0) {
+    const auto chunk_size = std::min(to_write, utils::kFileBufferSize);
+    file_.Write(data, chunk_size);
```
Ignition: Should that be `data + processed_bytes`?

If so should we have a test that would have covered that?

src/rpc/file_replication_handler.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+  if (!utils::EnsureDir(kTempDirectory)) {
+    spdlog::error("Failed to create temporary directory {}", kTempDirectory);
+    return std::nullopt;
+  }
+
+  slk::Reader req_reader(data, size, size);
+  storage::replication::Decoder decoder(&req_reader);
+
+  auto const maybe_filename = decoder.ReadString();
```
Ignition, line 42: Can we do some basic validation.
- non empty
- std::filesystem::path does not throw an exception
- pure filename no directories (use `has_parent_path()`)
- maybe other edge cases

src/rpc/file_replication_handler.cpp
Ignition: Maybe we should have a fuzz test generate 0 to n files and change the protocol can send and validate what is received.

as51340: yeah you are right, it is a bug but one that would probably never be seen. We are writing 256KiB to file, however we read from socket max 64KiB. So while loop actually isn't necessary. This would manifest only if we at some point change max buffer size

as51340: actually, we resize input stream so I may have underestimated severance of this bug...

src/rpc/file_replication_handler.cpp
```diff
@@ -61,8 +61,11 @@ std::optional<size_t> FileReplicationHandler::OpenFile(const uint8_t *data, size
   MG_ASSERT(maybe_file_size, "File size missing");
   file_size_ = *maybe_file_size;
 
+  spdlog::info("Replica will be using file {} with size {}", path, file_size_);
+
   // First N bytes are file_name and file_size, therefore we don't read full size
   size_t const processed_bytes = req_reader.GetPos();
+  spdlog::info("Processed {} bytes when opening file", processed_bytes);
```
as51340: I will remove this log once all tests are done

src/rpc/protocol.cpp
```diff
@@ -82,6 +82,7 @@ void RpcMessageDeliverer::Execute() {
     }
 
     if (ret.status == slk::StreamStatus::NEW_FILE) {
+      spdlog::info("Stream size when new file is received: {}", input_stream_->size());
```
as51340: I will remove this log once all tests are done

src/storage/v2/durability/snapshot.cpp
```diff
@@ -8618,9 +8618,12 @@ auto EnsureRetentionCountSnapshotsExist(const std::filesystem::path &snapshot_di
       if (info.uuid != uuid) continue;
       old_snapshot_files.emplace_back(info.durable_timestamp, item.path());
     } catch (const RecoveryFailure &e) {
-      spdlog::warn("Found a corrupt snapshot file {} becuase of: {}. Corrupted snapshot file will be deleted.",
-                   item.path(), e.what());
+      spdlog::warn("Found a corrupt snapshot file {} becuase of: {}.", item.path(), e.what());
+      /*
+       *spdlog::warn("Found a corrupt snapshot file {} becuase of: {}. Corrupted snapshot file will be deleted.",
+       *item.path(), e.what());
```
as51340: I will remove this log once all tests are done

src/rpc/file_replication_handler.cpp
as51340: about fuzzer: we actually have Jepsen test which in this case acts like a fuzzer since files of random size and random number of files is used

src/rpc/file_replication_handler.cpp
```diff
@@ -0,0 +1,109 @@
...
+  if (!utils::EnsureDir(tmp_rnd_dir)) {
+    spdlog::error("Failed to create temporary directory {}", tmp_rnd_dir);
+    return std::nullopt;
+  }
+
+  slk::Reader req_reader(data, size, size);
+  storage::replication::Decoder decoder(&req_reader);
+
+  auto const maybe_filename = decoder.ReadString();
+  MG_ASSERT(maybe_filename.has_value(), "Filename missing for the received file over the RPC");
```
Ignition: You need to consider malicious payloads. 
Make it not abort.
```
if (!maybe_filename.has_value()) {
    spdlog::info("Filename missing for the received file over the RPC");
    return std::nullopt;
}
```

src/rpc/file_replication_handler.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+  if (maybe_filename->empty()) {
+    spdlog::error("Filename is empty");
+    return std::nullopt;
+  }
```
Ignition: Should also check for 
- trailing slashes that would indicate a directory (not the intended filename)
- any reserved names that we would reasonably see in that directory?
- "." or ".."

src/rpc/file_replication_handler.cpp
```diff
@@ -0,0 +1,109 @@
...
+  auto const file_path = std::filesystem::path(*maybe_filename);
+  if (file_path.has_parent_path()) {
+    spdlog::error("File cannot have a parent path{}", file_path.string());
+    return std::nullopt;
+  }
+
+  auto const path = tmp_rnd_dir / file_path;
+  paths_.emplace_back(path);
+
+  file_.Open(path, utils::OutputFile::Mode::OVERWRITE_EXISTING);
```
Ignition, line 52: Can we check if it is actually open, maybe some random invalid unicode is not valid on the filesystem.

src/rpc/file_replication_handler.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+  const auto maybe_file_size = decoder.ReadUint();
+  MG_ASSERT(maybe_file_size, "File size missing");
```
Ignition: Again handle malicious packets. Don't allow DoS.
```
if (!maybe_file_size.has_value()) {
    spdlog::info("File size missing in RPC message");
    ResetCurrentFile();
    return std::nullopt;
}
```

Ignition (approved):

src/rpc/file_replication_handler.cpp
as51340: We have MG_ASSERT when opening file which will do the check:
```

void OutputFile::Open(const std::filesystem::path &path, Mode mode) {
  MG_ASSERT(!IsOpen(),
            "While trying to open {} for writing the database"
            " used a handle that already has {} opened in it!",
            path, path_);
  path_ = path;
  written_since_last_sync_ = 0;

  int flags = O_WRONLY | O_CLOEXEC | O_CREAT;
  if (mode == Mode::APPEND_TO_EXISTING) flags |= O_APPEND;

  while (true) {
    // The permissions are set to ((rw-r-----) & ~umask)
    fd_ = open(path_.c_str(), flags, 0640);
    if (fd_ == -1 && errno == EINTR) {
      // The call was interrupted, try again...
      continue;
    } else {
      // All other possible errors are fatal errors and are handled in the
      // MG_ASSERT below.
      break;
    }
  }

  MG_ASSERT(fd_ != -1, "While trying to open {} for writing an error occurred: {} ({})", path_, strerror(errno), errno);
}
```

PR #3232, opened 2025-09-02
Add union find data structure

imilinovic:
Add the union data structure - one of the prerequisites for Egraphs.


imilinovic:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist

src/planner/src/core/union_find.cpp
```diff
@@ -0,0 +1,150 @@
...
+auto UnionFind::Find(id_t id) -> id_t {
+  // TODO: C++26 contract
+  assert(id < parent_.size());
+
+  // Single-pass path halving algorithm with branch prediction hints
+  // Path halving is faster than full compression for find-heavy workloads
+  auto current = id;
+  while (parent_[current] != current) [[likely]] {
+      auto const next = parent_[current];
+      parent_[current] = parent_[next];
```
andrejtonev: since you are squashing the path here, should the rank_ be updated?

src/planner/src/core/union_find.cpp
```diff
@@ -0,0 +1,150 @@
...
+  num_of_sets_--;
+
+  auto rank_comparison = rank_[root_a] <=> rank_[root_b];
+  bool a_is_parent = rank_comparison == std::strong_ordering::greater;
+  auto parent_root = a_is_parent ? root_a : root_b;
+  auto child_root = a_is_parent ? root_b : root_a;
+
+  parent_[child_root] = parent_root;
+  if (rank_comparison == std::strong_ordering::equivalent) {
+    ++rank_[parent_root];
```
andrejtonev, line 81: would prefer an overflow check since rank_ is just 8bit

src/planner/src/core/union_find.cpp
```diff
@@ -0,0 +1,150 @@
...
+  auto id = static_cast<id_t>(parent_.size());
+  parent_.push_back(id);
+  rank_.push_back(0);
+  num_of_sets_++;
+  return id;
+}
+
+auto UnionFind::Find(id_t id) -> id_t {
+  // TODO: C++26 contract
+  assert(id < parent_.size());
```
andrejtonev, line 35: do we really want to assert?

src/planner/src/core/union_find.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+#include <climits>
+#include <limits>
+
+namespace memgraph::planner::core {
+
+// Static assertion to ensure UnionFind::rank_t rank storage is sufficient
+// Maximum rank in a union-find is O(log n)
+static_assert(sizeof(UnionFind::id_t) * CHAR_BIT <= std::numeric_limits<UnionFind::rank_t>::max(),
+              "8-bit rank storage insufficient for expected maximum tree depth");
```
andrejtonev: This holds only if the trees are balanced. Which I am not sure is always the case

src/planner/src/core/union_find.cpp
```diff
@@ -0,0 +1,150 @@
...
+    unique_best = is_higher_rank || (unique_best && !(is_equal_rank && is_different_root));
+  }
+
+  // Pass 3: Union all to be the best_root
+  size_t merges_performed = 0;
+  for (auto root : roots) {
+    parent_[root] = best_root;
+    merges_performed += (root != best_root);
+  }
+  num_of_sets_ -= merges_performed;
```
andrejtonev, line 127: this is hard to follow for something so simple

src/planner/include/planner/core/union_find.hpp
```diff
@@ -0,0 +1,185 @@
...
+   **/
+  std::deque<id_t> parent_;
+
+  /**
+   * Uses std::deque with 8-bit rank storage instead of the typical
+   * size_t or int. This provides significant memory savings (8 bytes → 1 byte
+   * per element) while supporting trees up to depth 255, which is more than
+   * sufficient for practical use cases.
+   */
+  std::deque<rank_t> rank_;
```
andrejtonev, line 181: would it make sense to combine parent and rank under one structure?

src/planner/src/core/union_find.cpp
andrejtonev: also use `std::numeric_limits<T>::digits` instead of `size*8`

andrejtonev (commented):
Left a few small comments. My main concern is that this completely ignores the overall structure of the codebase (where the benchmarks, tests, src/include are generally located).

src/planner/src/core/union_find.cpp
Ignition: No need `rank_` is a pseudo proxy for height, it works even if we do not update it for a correct height, hence no need to spend the time to do a rank_ update.

imilinovic: https://en.wikipedia.org/wiki/Disjoint-set_data_structure#Union_by_rank

src/planner/src/core/union_find.cpp
Ignition: Rank will at max be the height of a fully populated balanced binary tree.
Since `log_2(2^32-1) < 2^8-1`, we will never overflow.

src/planner/src/core/union_find.cpp
Ignition: No we want C++ contracts. The assert makes sure we do not make mistakes during development. We do not want to use MG_ASSERT, because we do not want that to impact performance. We could use DMG_ASSERT, but I was keeping dependencies to a minimum (for now).

src/planner/src/core/union_find.cpp
Ignition: Its a branch-less version, its been profiled to be better. This complexity is unavoidable.

src/planner/include/planner/core/union_find.hpp
Ignition: Memory layout cache friendliness. The find will be called more than union, hence we want to be able to pull parent ids into CPU cache without wasting any of the cacheline on rank.

src/planner/src/core/union_find.cpp
Ignition: The union ensure in the worst case actual height is no higher than a balanced binary tree. 

From Wikipedia:
> It can be shown that every node has rank log n or less. Consequently each rank can be stored in O(log log n) bits.

Ignition: This made us realised the static_assert was wrong, changed to:
```
static_assert(
   std::bit_width(std::numeric_limits<UnionFind::id_t>::max()) <
   std::numeric_limits<UnionFind::rank_t>::max(),
   "8-bit rank storage insufficient for expected maximum tree depth"
);
```

PR #3264, opened 2025-09-11
Add relevance score and limit to text search. Add aggregation queries to text edge index

DavIvek:
This PR adds a relevance score to the output of text search queries, introduces a limit option as an input parameter, and adds support for aggregation queries on the text edge index.

### DIFF
--- a/tests/unit/cpp_api.cpp
+++ b/tests/unit/cpp_api.cpp
@@ -1353,7 +1353,7 @@ TYPED_TEST(CppApiTestFixture, TestTextIndexOnNodes) {
   constexpr auto index_name = "node_text_index";
   constexpr auto label_name = "Document";
   constexpr auto property_name = "content";
-
+  constexpr auto text_search_limit = 10;
   // Create text index
   {
     auto storage_acc = this->storage->UniqueAccess();
@@ -1388,8 +1388,8 @@ TYPED_TEST(CppApiTestFixture, TestTextIndexOnNodes) {
     mgp_graph raw_graph = this->CreateGraph(db_acc.get());
 
     // Test search functionality
-    auto search_results =
-        mgp::SearchTextIndex(&raw_graph, index_name, "data.content:document", text_search_mode::SPECIFIED_PROPERTIES);
+    auto search_results = mgp::SearchTextIndex(&raw_graph, index_name, "data.content:document",
+                                               text_search_mode::SPECIFIED_PROPERTIES, text_search_limit);
     ASSERT_GE(search_results.Size(), 1);
   }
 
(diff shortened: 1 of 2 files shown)


DavIvek:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
      Text search now includes a relevance score in results, a limit parameter to control result count, and support for aggregation queries on text edge indexes. These enhancements let users rank, filter, and analyze search results more effectively. [#3264](https://github.com/memgraph/memgraph/pull/3264)

- [x] **[ Documentation PR link memgraph/documentation/pull/1422]**

init
```diff
@@ -53,8 +53,12 @@ else
...
 fi
 echo "ALL BUILD PACKAGES: $($OS_SCRIPT list MEMGRAPH_BUILD_DEPS)"
-$OS_SCRIPT check MEMGRAPH_BUILD_DEPS
-echo "All packages are in-place..."
+if [[ "$check_system_packages" == "true" ]]; then
+  $OS_SCRIPT check MEMGRAPH_BUILD_DEPS
+  echo "All packages are in-place..."
+else
+  echo "Skipping package check as requested..."
+fi
```
imilinovic, line 61: is this false by default? it should probably always be checked anyway

query_modules/text_search_module.cpp
```diff
@@ -208,32 +257,40 @@ extern "C" int mgp_init_module(struct mgp_module *query_module, struct mgp_memor
                  {
                      mgp::Parameter(TextSearch::kParameterIndexName, mgp::Type::String),
                      mgp::Parameter(TextSearch::kParameterSearchQuery, mgp::Type::String),
+                     mgp::Parameter(TextSearch::kParameterLimit, mgp::Type::Int, TextSearch::kDefaultLimit),
                  },
                  {mgp::Return(TextSearch::kReturnEdge, mgp::Type::Relationship),
-                  mgp::Return(TextSearch::kReturnFromVertex, mgp::Type::Node),
-                  mgp::Return(TextSearch::kReturnToVertex, mgp::Type::Node)},
+                  mgp::Return(TextSearch::kReturnScore, mgp::Type::Double)},
```
imilinovic, line 263: return signature changing, I think this is breaking.

DavIvek: - For the score:
  This is not breaking. The return parameter can be yielded, but it doesn’t have to be.  
  Since most tests remain unchanged and still pass, this isn’t a breaking change.  

- For the out and in vertex:
  This was a bug, the procedure wasn’t setting the out and in vertex into the record, so the procedure signature was incorrect.  
  However, since this feature hasn’t been released yet, we don’t need to consider it breaking.

init
DavIvek: The `check_system_packages` flag is `true` by default. It existed before but was never actually checked.  
I ran into some local issues with `npm`, which caused this check to fail, so adding the flag helped.  
This is probably a very rare situation, but the fix was useful for me.  
I’d keep it in case anyone else needs it, since it defaults to `true` anyway.

PR #3272, opened 2025-09-15
feat: Support routing in HA+MT environment

as51340:
The field 'database' wasn't handled properly when bolt+routing request was received. This PR enables the usage of routing driver when HA+MT is used.


as51340:
Cannot deliver this independently from https://github.com/memgraph/memgraph/pull/3278. All work will be done there, closing this

PR #3275, opened 2025-09-17
Toolchain v7 remove mg dump

mattkjames7:
Remove `mg_dump` from build, as it appears to only be used in the `integration/durability` test.
Instead, use `pymgclient` to run `DUMP DATABASE`.


imilinovic (approved):
Looks good, tests should fail if something is bad... Is this only used for durability integration tests?

PR #3282, opened 2025-09-18
Add new list PropertyValue types

DavIvek:
Idea of this PR is to use more precise and efficient internal list representations instead of storing all elements as generic `PropertyValue` lists.

### DIFF
--- a/src/storage/v2/property_store.cpp
+++ b/src/storage/v2/property_store.cpp
@@ -965,7 +965,7 @@ bool CompareLists(Reader *reader, ListType list_type, uint32_t size, const Prope
         return list_val[idx];
       }
       default:
-        return std::nullopt;
+        throw PropertyValueException("Invalid list type");
     }
   };
 
@@ -1010,10 +1010,10 @@ bool CompareLists(Reader *reader, ListType list_type, uint32_t size, const Prope
           if (!double_v) return std::nullopt;
           return *double_v;
         }
-        return std::nullopt;
+        throw PropertyValueException("Expected INT or DOUBLE while decoding numeric list");
       }
       default:
-        return std::nullopt;
+        throw PropertyValueException("Invalid list type");
     }
   };
 


DavIvek:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - This change introduces more precise and efficient internal list representations instead of storing all elements as generic PropertyValue lists. As a result, operations that create deltas for numeric lists will consume less memory, improving overall efficiency. In some cases, such as lists of small integers or short non-numeric lists, memory usage might slightly increase due to additional metadata or the fixed 4-byte integer representation. [#3282](https://github.com/memgraph/memgraph/pull/3282)

src/query/typed_value.cpp
```diff
@@ -98,6 +98,37 @@ TypedValue::TypedValue(const storage::PropertyValue &value, storage::NameIdMappe
...
+    case storage::PropertyValue::Type::NumericList: {
+      type_ = Type::List;
+      const auto &vec = value.ValueNumericList();
+      alloc_trait::construct(alloc_, &list_v);
+      for (const auto &v : vec) {
+        if (std::holds_alternative<int>(v)) {
+          list_v.emplace_back(std::get<int>(v));
+        } else {
+          list_v.emplace_back(std::get<double>(v));
+        }
```
Ignition, line 111: I don't think this is needed, variant has a copy constructor. `list_v.emplace_back(v)` should be fine.

src/storage/v2/property_store.cpp
```diff
@@ -588,6 +587,51 @@ std::optional<std::pair<Type, Size>> EncodePropertyValue(Writer *writer, const P
...
+    case PropertyValue::Type::IntList: {
+      const auto &list = value.ValueIntList();
+      auto size = writer->WriteUint(list.size());
+      if (!size) return std::nullopt;
+      for (const auto &item : list) {
+        auto metadata = writer->WriteMetadata();
+        if (!metadata) return std::nullopt;
+        auto ret = writer->WriteInt(static_cast<int64_t>(item));
+        if (!ret) return std::nullopt;
+        metadata->Set({Type::INT, Size::INT8, *ret});
```
Ignition: We shouldn't need to say int per element

src/storage/v2/property_store.cpp
```diff
@@ -588,6 +587,51 @@ std::optional<std::pair<Type, Size>> EncodePropertyValue(Writer *writer, const P
...
+        }
+      }
+      return {{Type::LIST, *size}};
+    }
+    case PropertyValue::Type::IntList: {
+      const auto &list = value.ValueIntList();
+      auto size = writer->WriteUint(list.size());
+      if (!size) return std::nullopt;
+      for (const auto &item : list) {
+        auto metadata = writer->WriteMetadata();
```
Ignition: we shouldn't need metadata per element here

src/storage/v2/property_store.cpp
```diff
@@ -588,6 +587,51 @@ std::optional<std::pair<Type, Size>> EncodePropertyValue(Writer *writer, const P
...
+        if (!metadata) return std::nullopt;
+        auto ret = writer->WriteInt(static_cast<int64_t>(item));
```
Ignition: Hmm, maybe we should store 4B regardless. don't use the small int representations

src/storage/v2/property_store.cpp
```diff
@@ -588,6 +587,51 @@ std::optional<std::pair<Type, Size>> EncodePropertyValue(Writer *writer, const P
...
+        metadata->Set({Type::INT, Size::INT8, *ret});
+      }
+      return {{Type::LIST, *size}};
+    }
+    case PropertyValue::Type::DoubleList: {
+      const auto &list = value.ValueDoubleList();
+      auto size = writer->WriteUint(list.size());
+      if (!size) return std::nullopt;
+      for (const auto &item : list) {
+        auto metadata = writer->WriteMetadata();
```
Ignition: double should be stored 8B per element

src/storage/v2/property_store.cpp
```diff
@@ -1211,15 +1286,51 @@ std::optional<uint64_t> DecodeZonedTemporalDataSize(Reader &reader) {
...
-      for (uint32_t i = 0; i < *size; ++i) {
-        auto metadata = reader->ReadMetadata();
-        if (!metadata) return false;
-        if (!ComparePropertyValue(reader, metadata->type, metadata->payload_size, list[i])) return false;
+      const auto list_size = GetListSize(value);
+      if (*size != list_size) return false;
+      if (value.IsIntList()) {
+        const auto &int_list = value.ValueIntList();
+        for (uint32_t i = 0; i < *size; ++i) {
+          auto metadata = reader->ReadMetadata();
```
Ignition: if we change encode then we don't need marker here either

src/query/typed_value.cpp
DavIvek: This calls the `TypedValue` constructor, but the `TypedValue` types haven’t changed. I didn’t add new list types to `TypedValue` to avoid complicating the code further, since that hasn’t been a memory pain point. Do you think we should also introduce new types to `TypedValue`?

src/storage/v2/property_store.cpp
DavIvek: Are we ok with small int lists growing in size then?
For example: https://github.com/memgraph/memgraph/pull/3282/files#diff-585209a695df614e4525f1821567d82e23b3a203bda7b333e66ec4ba85a53aa6R152

src/query/typed_value.cpp
Ignition: Leave TypedValue as is, query memory will be as large as its ever been, but query memory is temporary. It more important to keep deltas and property store small as they are longer lived memory. Also I don't expect arbitrary queries to be looking and accessing these lists, most query usages should be using the vector index, hence never materialising a TypedValue. Only an issue if someone does something silly like `RETURN node` that happens to read the entire property store to send over bolt to the client.

src/storage/v2/property_store.cpp
```diff
@@ -782,6 +833,181 @@ std::optional<uint64_t> DecodeZonedTemporalDataSize(Reader &reader) {
...
+      return value_type == PropertyValueType::NumericList;
+    default:
+      throw PropertyValueException("Invalid list type");
+  }
+}
+
+bool CompareLists(Reader *reader, ListType list_type, uint32_t size, const PropertyValue &value) {
+  switch (list_type) {
+    case ListType::PROPERTY_VALUE: {
+      const auto &list = value.ValueList();
```
Ignition: Bad assumption, the property store maybe storing a normal list, does not mean the compare value is also a normal list. Same bad assumption for the other cases.

src/storage/v2/property_value_utils.cpp
```diff
@@ -58,6 +61,9 @@ auto LowerBoundForType(PropertyValueType type) -> std::optional<utils::Bound<Pro
     case PropertyValue::Type::String:
       return utils::MakeBoundInclusive(kSmallestString);
     case PropertyValue::Type::List:
+    case PropertyValue::Type::NumericList:
+    case PropertyValue::Type::IntList:
+    case PropertyValue::Type::DoubleList:
       return utils::MakeBoundInclusive(kSmallestList);
```
Ignition, line 67: Need to double check ordering and equivalence. 
Is [1, 2, 3, 4] == [1.0, 2.0, 3.0, 4.0] == [1, 2, 3, 4.0]

How are lists ordered?
If my index held these would they be ordered correctly?
[1,1] [1.0, 2] [2.0, 1.0] [2, 2] [3, 1.0]

src/storage/v2/property_store.cpp
DavIvek: I think this assumption is fine because we first check whether the list types are the same. If the rest of the code is correct, then whenever two lists are equal, their types must also match. Am I missing something?
```
case Type::LIST: {
      if (!value.IsAnyList()) return false;
      const auto list_type = reader->ReadUint(Size::INT8);
      if (!list_type) return false;
      const auto size = reader->ReadUint(payload_size);
      if (!size || *size != value.ListSize()) return false;
      if (!CompareListTypes(static_cast<ListType>(*list_type), value.type()) ||
          !CompareLists(reader, static_cast<ListType>(*list_type), *size, value))
        return false;
      return true;
    }
```

Btw, I think it makes sense to move `CompareListTypes` inside `CompareLists`, so that the entire responsibility for comparing lists is handled within a single function, avoiding any potential misconceptions, so I'll do that also...

src/storage/v2/property_value_utils.cpp
DavIvek: Added more tests to verify this.

tests/gql_behave/tests/memgraph_V1/features/list_operations.feature
```diff
@@ -563,3 +563,12 @@ Feature: List operators
         Then the result should be:
             | n                                                                       |
             | (:Node {prop: ['string', [1, 2], [1, 2.5], [1.3, 1.4], ['string', 1]]}) |
+
+    Scenario: Numeric list equality between int and double lists
+        When executing query:
+            """
+            RETURN [1, 2, 3, 4] = [1.0, 2.0, 3.0, 4.0] AND [1.0, 2.0, 3.0, 4.0] = [1, 2, 3, 4.0] AND [1, 2, 3, 4] = [1, 2, 3, 4.0] AS x
+            """
```
Ignition, line 571: You need a test to compare when the list is compared with the property store value. e.g.
```
CREATE INDEX ON :L(lst);
CREATE (:L {lst=[1,2,3]});
MATCH (n:L) WHERE n.lst = [1.0, 2.0, 3.0] RETURN count(*) = 1 AS result;
```

PR #3291, opened 2025-09-22
Store vectors only under vector index

DavIvek:
Idea of this PR is to move vectors out of property store and store them only in vector index.

### DIFF
--- a/src/storage/v2/indices/vector_index.cpp
+++ b/src/storage/v2/indices/vector_index.cpp
@@ -36,7 +36,10 @@ namespace memgraph::storage {
 
 // unum::usearch::index_dense_gt is the index type used for vector indices. It is thread-safe and supports concurrent
 // operations.
-using mg_vector_index_t = unum::usearch::index_dense_gt<Vertex *, unum::usearch::uint40_t>;
+using mg_vector_index_t =
+    unum::usearch::index_dense_gt<Vertex *,
+                                  unum::usearch::uint40_t>;  // TODO(DavIvek): consider making this const Vertex*
+using synchronized_mg_vector_index_t = utils::Synchronized<mg_vector_index_t, std::shared_mutex>;
 
 // NOLINTNEXTLINE(bugprone-exception-escape)
 struct IndexItem {
@@ -44,7 +47,7 @@ struct IndexItem {
   // locking because resizing the index requires exclusive access. For all other operations, we can use shared lock even
   // though we are modifying index. In the case of removing or adding elements to the index we will use
   // MutableSharedLock to acquire an shared lock.
-  std::shared_ptr<utils::Synchronized<mg_vector_index_t, std::shared_mutex>> mg_index;
+  std::shared_ptr<synchronized_mg_vector_index_t> mg_index;
   VectorIndexSpec spec;
 };
 
@@ -55,7 +58,7 @@ struct IndexItem {
 struct VectorIndex::Impl {
   /// The `index_` member is a map that associates a `LabelPropKey` (a combination of label and property)
   /// with the pair of a IndexItem.
-  std::map<LabelPropKey, IndexItem> index_;
+  absl::flat_hash_map<LabelPropKey, IndexItem> index_;
 
   /// The `index_name_to_label_prop_` is a map that maps an index name (as a string) to the corresponding
   /// `LabelPropKey`. This allows the system to quickly resolve an index name to the spec
@@ -214,6 +217,19 @@ void VectorIndex::UpdateOnSetProperty(PropertyId property, const PropertyValue &
   }
 }
 
+PropertyValue VectorIndex::GetProperty(Vertex *vertex, PropertyId property) const {
+  // first check if property_id + any of labels is in the index
+  auto has_property = [&](const auto &label_prop) { return label_prop.property() == property; };
+  auto has_label = [&](const auto &label_prop) { return utils::Contains(vertex->labels, label_prop.label()); };
+  auto view = pimpl->index_ | rv::keys | rv::filter(has_property) | rv::filter(has_label);
+  if (view.empty()) {
+    return {};
+  }
+  const auto &label_prop = *view.begin();
+  auto &[mg_index, _] = pimpl->index_.at(label_prop);
+  return GetVectorAsPropertyValue(mg_index, vertex);
+}
+
 std::vector<VectorIndexInfo> VectorIndex::ListVectorIndicesInfo() const {
   std::vector<VectorIndexInfo> result;
   result.reserve(pimpl->index_.size());
@@ -324,5 +340,7 @@ VectorIndex::IndexStats VectorIndex::Analysis() const {
 bool VectorIndex::IndexExists(std::string_view index_name) const {
   return pimpl->index_name_to_label_prop_.contains(index_name);
 }
+
+bool VectorIndex::Empty() const { return pimpl->index_.empty(); }
 
 }  // namespace memgraph::storage
--- a/src/storage/v2/indices/vector_index_utils.hpp
+++ b/src/storage/v2/indices/vector_index_utils.hpp
@@ -13,9 +13,11 @@
 
 #include <cstddef>
 #include <cstdint>
+#include <shared_mutex>
 
-#include "query/exceptions.hpp"
+#include "storage/v2/property_value.hpp"
 #include "usearch/index_plugins.hpp"
+#include "utils/synchronized.hpp"
 
 namespace memgraph::storage {
 
@@ -58,201 +60,43 @@ struct VectorIndexConfigMap {
 /// @param metric The metric kind to convert.
 /// @return A string representation of the metric kind.
 /// @throws query::VectorSearchException if the metric kind is unsupported.
-inline const char *NameFromMetric(unum::usearch::metric_kind_t metric) {
-  switch (metric) {
-    case unum::usearch::metric_kind_t::l2sq_k:
-      return "l2sq";
-    case unum::usearch::metric_kind_t::ip_k:
-      return "ip";
-    case unum::usearch::metric_kind_t::cos_k:
-      return "cos";
-    case unum::usearch::metric_kind_t::haversine_k:
-      return "haversine";
-    case unum::usearch::metric_kind_t::divergence_k:
-      return "divergence";
-    case unum::usearch::metric_kind_t::pearson_k:
-      return "pearson";
-    case unum::usearch::metric_kind_t::hamming_k:
-      return "hamming";
-    case unum::usearch::metric_kind_t::tanimoto_k:
-      return "tanimoto";
-    case unum::usearch::metric_kind_t::sorensen_k:
-      return "sorensen";
-    default:
-      throw query::VectorSearchException(
-          "Unsupported metric kind. Supported metrics are l2sq, ip, cos, haversine, divergence, pearson, hamming, "
-          "tanimoto, and sorensen.");
-  }
-}
+const char *NameFromMetric(unum::usearch::metric_kind_t metric);
 
 /// @brief Converts a metric name to its corresponding metric kind.
 /// @param name The name of the metric.
 /// @return The corresponding metric kind.
 /// @throws query::VectorSearchException if the metric name is unsupported.
-inline unum::usearch::metric_kind_t MetricFromName(std::string_view name) {
-  if (name == "l2sq" || name == "euclidean_sq") {
-    return unum::usearch::metric_kind_t::l2sq_k;
-  }
-  if (name == "ip" || name == "inner" || name == "dot") {
-    return unum::usearch::metric_kind_t::ip_k;
-  }
-  if (name == "cos" || name == "angular") {
-    return unum::usearch::metric_kind_t::cos_k;
-  }
-  if (name == "haversine") {
-    return unum::usearch::metric_kind_t::haversine_k;
-  }
-  if (name == "divergence") {
-    return unum::usearch::metric_kind_t::divergence_k;
-  }
-  if (name == "pearson") {
-    return unum::usearch::metric_kind_t::pearson_k;
-  }
-  if (name == "hamming") {
-    return unum::usearch::metric_kind_t::hamming_k;
-  }
-  if (name == "tanimoto") {
-    return unum::usearch::metric_kind_t::tanimoto_k;
-  }
-  if (name == "sorensen") {
-    return unum::usearch::metric_kind_t::sorensen_k;
-  }
-  throw query::VectorSearchException(
-      fmt::format("Unsupported metric name: {}. Supported metrics are l2sq, ip, cos, haversine, divergence, pearson, "
-                  "hamming, tanimoto, and sorensen.",
-                  name));
-}
+unum::usearch::metric_kind_t MetricFromName(std::string_view name);
 
 /// @brief Converts a scalar kind to its string representation.
 /// @param scalar The scalar kind to convert.
 /// @return A string representation of the scalar kind.
 /// @throws query::VectorSearchException if the scalar kind is unsupported.
-inline const char *NameFromScalar(unum::usearch::scalar_kind_t scalar) {
-  switch (scalar) {
-    case unum::usearch::scalar_kind_t::b1x8_k:
-      return "b1x8";
-    case unum::usearch::scalar_kind_t::u40_k:
-      return "u40";
-    case unum::usearch::scalar_kind_t::uuid_k:
-      return "uuid";
-    case unum::usearch::scalar_kind_t::bf16_k:
-      return "bf16";
-    case unum::usearch::scalar_kind_t::f64_k:
-      return "f64";
-    case unum::usearch::scalar_kind_t::f32_k:
-      return "f32";
-    case unum::usearch::scalar_kind_t::f16_k:
-      return "f16";
-    case unum::usearch::scalar_kind_t::f8_k:
-      return "f8";
-    case unum::usearch::scalar_kind_t::u64_k:
-      return "u64";
-    case unum::usearch::scalar_kind_t::u32_k:
-      return "u32";
-    case unum::usearch::scalar_kind_t::u16_k:
-      return "u16";
-    case unum::usearch::scalar_kind_t::u8_k:
-      return "u8";
-    case unum::usearch::scalar_kind_t::i64_k:
-      return "i64";
-    case unum::usearch::scalar_kind_t::i32_k:
-      return "i32";
-    case unum::usearch::scalar_kind_t::i16_k:
-      return "i16";
-    case unum::usearch::scalar_kind_t::i8_k:
-      return "i8";
-    default:
-      throw query::VectorSearchException(
-          "Unsupported scalar kind. Supported scalars are b1x8, u40, uuid, bf16, f64, f32, f16, f8, "
-          "u64, u32, u16, u8, i64, i32, i16, and i8.");
-  }
-}
+const char *NameFromScalar(unum::usearch::scalar_kind_t scalar);
 
 /// @brief Converts a scalar name to its corresponding scalar kind.
 /// @param name The name of the scalar.
 /// @return The corresponding scalar kind.
 /// @throws query::VectorSearchException if the scalar name is unsupported.
-inline unum::usearch::scalar_kind_t ScalarFromName(std::string_view name) {
-  if (name == "b1x8" || name == "binary") {
-    return unum::usearch::scalar_kind_t::b1x8_k;
-  }
-  if (name == "u40") {
-    return unum::usearch::scalar_kind_t::u40_k;
-  }
-  if (name == "uuid") {
-    return unum::usearch::scalar_kind_t::uuid_k;
-  }
-  if (name == "bf16" || name == "bfloat16") {
-    return unum::usearch::scalar_kind_t::bf16_k;
-  }
-  if (name == "f64" || name == "float64" || name == "double") {
-    return unum::usearch::scalar_kind_t::f64_k;
-  }
-  if (name == "f32" || name == "float32" || name == "float") {
-    return unum::usearch::scalar_kind_t::f32_k;
-  }
-  if (name == "f16" || name == "float16") {
-    return unum::usearch::scalar_kind_t::f16_k;
-  }
-  if (name == "f8" || name == "float8") {
-    return unum::usearch::scalar_kind_t::f8_k;
-  }
-  if (name == "u64" || name == "uint64") {
-    return unum::usearch::scalar_kind_t::u64_k;
-  }
-  if (name == "u32" || name == "uint32") {
-    return unum::usearch::scalar_kind_t::u32_k;
-  }
-  if (name == "u16" || name == "uint16") {
-    return unum::usearch::scalar_kind_t::u16_k;
-  }
-  if (name == "u8" || name == "uint8") {
-    return unum::usearch::scalar_kind_t::u8_k;
-  }
-  if (name == "i64" || name == "int64") {
-    return unum::usearch::scalar_kind_t::i64_k;
-  }
-  if (name == "i32" || name == "int32") {
-    return unum::usearch::scalar_kind_t::i32_k;
-  }
-  if (name == "i16" || name == "int16") {
-    return unum::usearch::scalar_kind_t::i16_k;
-  }
-  if (name == "i8" || name == "int8") {
-    return unum::usearch::scalar_kind_t::i8_k;
-  }
-
-  throw query::VectorSearchException(
-      fmt::format("Unsupported scalar name: {}. Supported scalars are b1x8, u40, uuid, bf16, f64, f32, f16, f8, "
-                  "u64, u32, u16, u8, i64, i32, i16, and i8.",
-                  name));
-}
+unum::usearch::scalar_kind_t ScalarFromName(std::string_view name);
 
 /// @brief Converts a distance to a similarity score based on the metric kind.
 /// @param metric The metric kind used for the distance.
 /// @param distance The distance value to convert.
 /// @return The similarity score corresponding to the distance.
 /// @throws query::VectorSearchException if the metric kind is unsupported.
-inline double SimilarityFromDistance(unum::usearch::metric_kind_t metric, double distance) {
-  switch (metric) {
-    case unum::usearch::metric_kind_t::ip_k:
-    case unum::usearch::metric_kind_t::cos_k:
-    case unum::usearch::metric_kind_t::pearson_k:
-    case unum::usearch::metric_kind_t::hamming_k:
-    case unum::usearch::metric_kind_t::tanimoto_k:
-    case unum::usearch::metric_kind_t::sorensen_k:
-    case unum::usearch::metric_kind_t::jaccard_k:
-      return 1.0 - distance;
+double SimilarityFromDistance(unum::usearch::metric_kind_t metric, double distance);
 
-    case unum::usearch::metric_kind_t::l2sq_k:
-    case unum::usearch::metric_kind_t::haversine_k:
-    case unum::usearch::metric_kind_t::divergence_k:
-      return 1.0 / (1.0 + distance);
-
-    default:
-      throw query::VectorSearchException(
-          fmt::format("Unsupported metric kind for similarity calculation: {}", NameFromMetric(metric)));
-  }
-}
+/// @brief Retrieves a vector from a USearch index using the get method and returns it as a PropertyValue of type
+/// double.
+/// @tparam IndexType The type of the USearch index (e.g., mg_vector_index_t or mg_vector_edge_index_t).
+/// @tparam KeyType The type of the key used in the index (e.g., Vertex* or EdgeIndexEntry).
+/// @param index The USearch index to retrieve the vector from.
+/// @param key The key to look up in the index.
+/// @return A PropertyValue containing the vector as a list of double values.
+/// @throws query::VectorSearchException if the key is not found in the index or if retrieval fails.
+template <typename IndexType, typename KeyType>
+PropertyValue GetVectorAsPropertyValue(const std::shared_ptr<utils::Synchronized<IndexType, std::shared_mutex>> &index,
+                                       KeyType key);
 
 }  // namespace memgraph::storage
--- a/src/storage/v2/vertex_accessor.cpp
+++ b/src/storage/v2/vertex_accessor.cpp
@@ -570,18 +570,32 @@ Result<std::map<PropertyId, PropertyValue>> VertexAccessor::ClearProperties() {
 Result<PropertyValue> VertexAccessor::GetProperty(PropertyId property, View view) const {
   bool exists = true;
   bool deleted = false;
+  bool is_in_vector_index = false;
   PropertyValue value;
   Delta *delta = nullptr;
+
   {
     auto guard = std::shared_lock{vertex_->lock};
     deleted = vertex_->deleted;
-    value = vertex_->properties.GetProperty(property);
     delta = vertex_->delta;
+
+    // Try to get property from vector index first
+    if (!storage_->indices_.vector_index_.Empty()) {
+      value = storage_->indices_.vector_index_.GetProperty(vertex_, property);
+      if (!value.IsNull()) {
+        is_in_vector_index = true;
+      }
+    }
+    // Fall back to regular vertex properties if not found in vector index
+    if (!is_in_vector_index) {
+      value = vertex_->properties.GetProperty(property);
+    }
   }
 
   // Checking cache has a cost, only do it if we have any deltas
   // if we have no deltas then what we already have from the vertex is correct.
-  if (delta && transaction_->isolation_level != IsolationLevel::READ_UNCOMMITTED) {
+  // Vector index works in read uncommitted mode so we don't need to check for it
+  if (delta && transaction_->isolation_level != IsolationLevel::READ_UNCOMMITTED && !is_in_vector_index) {
     // IsolationLevel::READ_COMMITTED would be tricky to propagate invalidation to
     // so for now only cache for IsolationLevel::SNAPSHOT_ISOLATION
     auto const useCache = transaction_->isolation_level == IsolationLevel::SNAPSHOT_ISOLATION;
--- a/src/storage/v2/CMakeLists.txt
+++ b/src/storage/v2/CMakeLists.txt
@@ -45,6 +45,7 @@ target_sources(mg-storage-v2
         indices/text_index_utils.cpp
         indices/text_index.cpp
         indices/text_edge_index.cpp
+        indices/vector_index_utils.cpp
         indices/vector_edge_index.cpp
         indices/vector_index.cpp
         inmemory/edge_property_index.cpp
--- a/src/storage/v2/indices/text_index.cpp
+++ b/src/storage/v2/indices/text_index.cpp
@@ -13,6 +13,7 @@
 
 #include "flags/experimental.hpp"
 #include "mgcxx_text_search.hpp"
+#include "query/exceptions.hpp"
 #include "storage/v2/id_types.hpp"
 #include "storage/v2/indices/text_index_utils.hpp"
 #include "storage/v2/transaction.hpp"
--- a/src/storage/v2/indices/vector_index.hpp
+++ b/src/storage/v2/indices/vector_index.hpp
@@ -111,6 +111,12 @@ class VectorIndex {
   /// @param vertex The vertex on which the property was modified.
   void UpdateOnSetProperty(PropertyId property, const PropertyValue &value, Vertex *vertex);
 
+  /// @brief Retrieves the vector of a vertex as a PropertyValue.
+  /// @param vertex The vertex to retrieve the vector from.
+  /// @param property The property of the vertex.
+  /// @return The vector of the vertex as a PropertyValue.
+  PropertyValue GetProperty(Vertex *vertex, PropertyId property) const;
+
   /// @brief Lists the info of all existing indexes.
   /// @return A vector of VectorIndexInfo objects representing the indexes.
   std::vector<VectorIndexInfo> ListVectorIndicesInfo() const;
@@ -156,6 +162,10 @@ class VectorIndex {
   /// @param index_name The name of the index to check.
   /// @return true if the index exists, false otherwise.
   bool IndexExists(std::string_view index_name) const;
+
+  /// @brief Checks if the index is empty.
+  /// @return true if the index is empty, false otherwise.
+  bool Empty() const;
 
  private:
   /// @brief Adds a vertex to an existing index.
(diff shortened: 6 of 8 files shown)


DavIvek:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Vector properties now use a new storage architecture that separates metadata from the actual vector data. Instead of duplicating vectors in the property store, Memgraph stores a lightweight reference and keeps the vector only once inside the vector index, fetching it on demand when accessed. [#3291](https://github.com/memgraph/memgraph/pull/3291)
- [x] **https://github.com/memgraph/documentation/pull/1526**

src/glue/communication.cpp
```diff
@@ -460,6 +460,15 @@ Value ToBoltValue(const storage::PropertyValue &value, const storage::Storage &s
...
       return {value.ValuePoint3d()};
     }
+    case storage::PropertyValue::Type::VectorIndexId: {
+      const auto &vector = value.ValueVectorIndexList();
+      std::vector<Value> vec;
+      vec.reserve(vector.size());
+      for (auto v : vector) {
+        vec.emplace_back(v);
+      }
+      return vec;
```
andrejtonev: use transform and ranges

src/query/frontend/ast/pretty_print.cpp
```diff
@@ -309,6 +316,9 @@ void PrintObject(std::ostream *out, const DbAccessor *dba, const storage::Extern
       break;
     case storage::ExternalPropertyValue::Type::Point3d:
       PrintObject(out, value.ValuePoint3d());
+      break;
+    case storage::ExternalPropertyValue::Type::VectorIndexId:
+      PrintObject(out, dba, value.ValueVectorIndexList());
```
andrejtonev, line 322: break

src/query/frontend/ast/pretty_print.cpp
```diff
@@ -249,6 +253,9 @@ void PrintObject(std::ostream *out, const DbAccessor *dba, const storage::Proper
       break;
     case storage::PropertyValue::Type::Point3d:
       PrintObject(out, value.ValuePoint3d());
+      break;
+    case storage::PropertyValue::Type::VectorIndexId:
+      PrintObject(out, dba, value.ValueVectorIndexList());
```
andrejtonev, line 258: break

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -832,6 +831,13 @@ mgp_value::mgp_value(const memgraph::storage::PropertyValue &pv, memgraph::stora
           "open a new issue / comment under existing one under github.com/memgraph/memgraph."};
       break;
     }
+    case memgraph::storage::PropertyValue::Type::VectorIndexId: {
+      throw std::logic_error{
+          "mgp_value for PropertyValue::Type::VectorIndexId doesn't exist. Contact Memgraph team under "
+          "team@memgraph.com or "
+          "open a new issue / comment under existing one under github.com/memgraph/memgraph."};
+      break;
```
andrejtonev: doesn't need a break

src/query/typed_value.cpp
```diff
@@ -197,6 +197,19 @@ TypedValue::TypedValue(const storage::PropertyValue &value, storage::NameIdMappe
...
+      const auto &vec = value.ValueVectorIndexList();
+      if (vec.empty()) {
+        type_ = Type::Null;
+        return;
+      }
+      type_ = Type::List;
+      alloc_trait::construct(alloc_, &list_v);
+      for (const auto &v : vec) {
+        list_v.emplace_back(static_cast<double>(v));
+      }
```
andrejtonev: transform

src/query/typed_value.cpp
```diff
@@ -336,6 +349,19 @@ TypedValue::TypedValue(storage::PropertyValue &&other, storage::NameIdMapper *na
       alloc_trait::construct(alloc_, &point_3d_v, other.ValuePoint3d());
       break;
     }
+    case storage::PropertyValue::Type::VectorIndexId: {
+      const auto &vec = other.ValueVectorIndexList();
```
andrejtonev, line 346: In theory this should be a move and not a copy

src/storage/v2/disk/storage.cpp
```diff
@@ -250,6 +250,12 @@ DiskStorage::DiskStorage(Config config, PlanInvalidatorPtr invalidator,
     column_families.emplace_back(kOutEdgesHandle, kvstore_->options_);
     column_families.emplace_back(kInEdgesHandle, kvstore_->options_);
 
+    logging::AssertRocksDBStatus(rocksdb::TransactionDB::Open(kvstore_->options_,
```
andrejtonev: what's this?

src/storage/v2/durability/serialization.cpp
```diff
@@ -237,6 +237,20 @@ void Encoder<FileType>::WriteExternalPropertyValue(const ExternalPropertyValue &
       WritePoint3d(value.ValuePoint3d());
       break;
     }
+    case ExternalPropertyValue::Type::VectorIndexId: {
+      const auto &vector_index_ids = value.ValueVectorIndexIds();
+      WriteMarker(Marker::TYPE_VECTOR_INDEX_ID);
+      WriteSize(this, vector_index_ids.size());
+      for (const auto &id : vector_index_ids) {
+        WriteString(id);
+      }
```
andrejtonev, line 246: there should be a wrtitelist or similar

src/storage/v2/durability/wal.cpp
```diff
@@ -801,40 +802,44 @@ bool SkipWalDeltaData(BaseDecoder *decoder, const uint64_t version) {
...
       // (with the `GetProperty` call). It is the only memory allocation in the
       // entire WAL file writing logic.
-      encoder->WriteExternalPropertyValue(
-          ToExternalPropertyValue(vertex.properties.GetProperty(delta.property.key), name_id_mapper));
+      auto property_value = vertex->properties.GetProperty(delta.property.key);
+      if (property_value.IsVectorIndexId()) {
+        property_value.ValueVectorIndexList() = storage->indices_.vector_index_.GetVectorPropertyFromIndex(
+            vertex, storage->name_id_mapper_->IdToName(property_value.ValueVectorIndexIds()[0]));
+      }
+      encoder->WriteExternalPropertyValue(ToExternalPropertyValue(property_value, storage->name_id_mapper_.get()));
```
andrejtonev: don't like this

src/storage/v2/indices/vector_edge_index.cpp
```diff
@@ -28,13 +28,16 @@ namespace memgraph::storage {
...
+using SyncVectorEdgeIndex = utils::Synchronized<mg_vector_edge_index_t, std::shared_mutex>;
+
 struct EdgeTypeIndexItem {
   // unum::usearch::index_dense_gt is thread-safe and supports concurrent operations. However, we still need to use
-  // locking because resizing the index requires exclusive access. For all other operations, we can use shared lock even
-  // though we are modifying index. In the case of removing or adding elements to the index we will use
-  // MutableSharedLock to acquire an shared lock.
-  std::shared_ptr<utils::Synchronized<mg_vector_edge_index_t, std::shared_mutex>> mg_index;
+  // locking because resizing the index requires exclusive access.
+  SyncVectorEdgeIndex mg_index;
```
andrejtonev, line 36: we removed the shared_ptr, is that correct?

src/storage/v2/indices/vector_edge_index.cpp
```diff
@@ -158,11 +169,7 @@ void VectorEdgeIndex::SetupIndex(const VectorEdgeIndexSpec &spec) {
   }
 
   pimpl->index_name_to_edge_type_prop_.emplace(spec.index_name, edge_type_prop);
-  pimpl->edge_index_.emplace(
-      edge_type_prop,
-      EdgeTypeIndexItem{.mg_index = std::make_shared<utils::Synchronized<mg_vector_edge_index_t, std::shared_mutex>>(
-                            std::move(mg_edge_index)),
-                        .spec = spec});
+  pimpl->edge_index_.try_emplace(edge_type_prop, std::move(mg_edge_index), spec);
```
andrejtonev: what happens if we fail to emplace?

src/storage/v2/indices/vector_index.cpp
```diff
@@ -33,16 +27,16 @@ namespace memgraph::storage {
...
 
 // NOLINTNEXTLINE(bugprone-exception-escape)
 struct IndexItem {
   // unum::usearch::index_dense_gt is thread-safe and supports concurrent operations. However, we still need to use
-  // locking because resizing the index requires exclusive access. For all other operations, we can use shared lock even
-  // though we are modifying index. In the case of removing or adding elements to the index we will use
-  // MutableSharedLock to acquire an shared lock.
-  std::shared_ptr<utils::Synchronized<mg_vector_index_t, std::shared_mutex>> mg_index;
+  // locking because resizing the index requires exclusive access.
+  synchronized_mg_vector_index_t mg_index;
```
andrejtonev, line 36: lost the shared_ptr

src/storage/v2/indices/vector_index.cpp
```diff
@@ -52,56 +46,29 @@ struct IndexItem {
...
 
-bool VectorIndex::CreateIndex(const VectorIndexSpec &spec, utils::SkipList<Vertex>::Accessor &vertices,
-                              std::optional<SnapshotObserverInfo> const &snapshot_info) {
+bool VectorIndex::CreateIndex(VectorIndexSpec &spec, utils::SkipList<Vertex>::Accessor &vertices,
+                              NameIdMapper *name_id_mapper, std::optional<SnapshotObserverInfo> const &snapshot_info) {
   const utils::MemoryTracker::OutOfMemoryExceptionEnabler oom_exception;
   try {
     SetupIndex(spec);
-    PopulateIndexOnSingleThread(vertices, spec, snapshot_info);
+    PopulateVectorIndexSingleThreaded(vertices, [&](Vertex &vertex) {
```
andrejtonev, line 65: need to pair review this file

src/storage/v2/indices/vector_index_utils.hpp
```diff
@@ -131,12 +143,6 @@ inline unum::usearch::metric_kind_t MetricFromName(std::string_view name) {
 /// @throws query::VectorSearchException if the scalar kind is unsupported.
 inline const char *NameFromScalar(unum::usearch::scalar_kind_t scalar) {
   switch (scalar) {
-    case unum::usearch::scalar_kind_t::b1x8_k:
-      return "b1x8";
-    case unum::usearch::scalar_kind_t::u40_k:
-      return "u40";
-    case unum::usearch::scalar_kind_t::uuid_k:
-      return "uuid";
```
andrejtonev, line 139: Note in the PR description and release note that we are losing support for these scalars

src/storage/v2/indices/vector_index_utils.hpp
```diff
@@ -257,30 +254,80 @@ inline double SimilarityFromDistance(unum::usearch::metric_kind_t metric, double
...
 
-  const auto vector_size = GetListSize(property);
-  if (expected_dimension != vector_size) {
-    throw query::VectorSearchException("Vector index property must have the same number of dimensions as the index.");
+  const auto list_size = value.ListSize();
+  utils::small_vector<float> vector;
+  vector.reserve(list_size);
+  for (std::size_t i = 0; i < list_size; i++) {
+    auto numeric_value = GetNumericValueAt(value, i);
+    auto float_value = std::visit([](auto val) { return static_cast<float>(val); }, *numeric_value);
```
andrejtonev, line 290: not sure what numeric_value can be, maybe it makes sense to have a auto &val?

src/storage/v2/indices/vector_index_utils.hpp
```diff
@@ -257,30 +254,80 @@ inline double SimilarityFromDistance(unum::usearch::metric_kind_t metric, double
...
+/// @param vector The vector of float values to restore.
+inline void RestoreVectorOnVertex(Vertex *vertex, PropertyId property_id, const utils::small_vector<float> &vector) {
+  vertex->properties.SetProperty(property_id, PropertyValue(std::vector<double>(vector.begin(), vector.end())));
+}
+
+/// @brief Removes an index ID from a property's vector index ID list.
+/// @param property_value The property value to modify (must be a VectorIndexId).
+/// @param index_id The index ID to remove.
+/// @return true if the property should be restored (no more index IDs), false otherwise.
+inline bool UnregisterFromIndex(PropertyValue &property_value, uint64_t index_id) {
```
andrejtonev: Is this used in the PropertyStore?

src/storage/v2/durability/snapshot.cpp
```diff
@@ -9505,73 +9491,84 @@ RecoveredSnapshot LoadCurrentVersionSnapshot(Decoder &snapshot, std::filesystem:
...
         indices_constraints.indices.vector_indices.emplace_back(
-            std::move(index_name.value()),
-            get_label_from_id(*label),
-            get_property_from_id(*property),
-            metric_kind,
-            static_cast<uint16_t>(*dimension),
-            static_cast<uint16_t>(*resize_coefficient),
-            *capacity,
-            static_cast<unum::usearch::scalar_kind_t>(*scalar_kind));
+            VectorIndexRecoveryInfo{.spec = std::move(spec), .index_entries = std::move(index_entries)});
```
andrejtonev, line 9524: How does this work in the big picture?
Do we first create the indices, then load the data?

src/storage/v2/inmemory/storage.cpp
```diff
@@ -1263,21 +1249,7 @@ void InMemoryStorage::InMemoryAccessor::Abort() {
...
-                const auto &vector_properties = index_abort_processor.vector_.l2p.find(current->label.value);
-                if (vector_properties != index_abort_processor.vector_.l2p.end()) {
-                  // label is in the vector index
-                  for (const auto &property : vector_properties->second) {
-                    if (vertex->properties.HasProperty(property)) {
-                      // it has to be removed from the index
-                      vector_label_property_cleanup[LabelPropKey{current->label.value, property}].emplace_back(vertex);
-                    }
-                  }
-                }
```
andrejtonev, line 1280: I'm assuming this is handled by the abort_processor

andrejtonev: which is defined in the vector index cpp file which I skipped for now

src/storage/v2/replication/slk.cpp
```diff
@@ -232,6 +233,15 @@ void Save(const storage::ExternalPropertyValue &value, slk::Builder *builder) {
       slk::Save(value.ValuePoint3d(), builder);
       return;
     }
+    case storage::ExternalPropertyValue::Type::VectorIndexId: {
+      slk::Save(storage::ExternalPropertyValue::Type::VectorIndexId, builder);
+      slk::Save(value.ValueVectorIndexIds().size(), builder);
+      for (const auto &id : value.ValueVectorIndexIds()) {
+        slk::Save(id, builder);
+      }
```
andrejtonev: there should be a built in serializer for vectors

src/storage/v2/replication/slk.cpp
```diff
@@ -385,6 +395,18 @@ void Load(storage::ExternalPropertyValue *value, slk::Reader *reader) {
       *value = storage::ExternalPropertyValue(v);
       return;
     }
+    case storage::ExternalPropertyValue::Type::VectorIndexId: {
+      uint32_t size{0};
+      slk::Load(&size, reader);
+      utils::small_vector<std::string> ids(size);
+      for (size_t i = 0; i < size; ++i) {
+        slk::Load(&ids[i], reader);
+      }
```
andrejtonev: should be able to automatically load like the small vector below

src/storage/v2/property_store.cpp
```diff
@@ -1144,6 +1157,19 @@ bool CompareLists(Reader *reader, ListType list_type, uint32_t size, const Prope
...
+      auto size = reader->ReadUint(payload_size);
+      if (!size) return false;
+      utils::small_vector<uint64_t> vector_index_ids;
+      vector_index_ids.reserve(*size);
+      for (uint64_t i = 0; i < *size; ++i) {
+        auto id = reader->ReadUint(Size::INT64);
+        if (!id) return false;
+        vector_index_ids.push_back(*id);
+      }
+      value = PropertyValue(std::move(vector_index_ids), utils::small_vector<float>{});
```
andrejtonev: would be better if the vector index vector + cache was encapsulated in a struct
that way you don't need to pass in the other vector and you can explicitly define the type

src/storage/v2/property_store.cpp
```diff
@@ -1558,6 +1610,17 @@ bool CompareLists(Reader *reader, ListType list_type, uint32_t size, const Prope
...
     }
+    case Type::VECTOR: {
+      if (!value.IsVectorIndexId()) return false;
+      const auto &vector_index_ids = value.ValueVectorIndexIds();
+      auto count = reader->ReadUint(payload_size);
+      if (!count || *count != vector_index_ids.size()) return false;
+      for (auto id : vector_index_ids) {
+        auto read_id = reader->ReadUint(Size::INT64);
+        if (!read_id || *read_id != id) return false;
+      }
```
andrejtonev: this forces the ordering to always be the same. is the order guaranteed somewhere?

src/storage/v2/property_store.cpp
```diff
@@ -1745,6 +1808,9 @@ enum class ExpectedPropertyStatus {
         return ExpectedPropertyStatus::EQUAL;
       }
     } break;
+    case VECTOR:
+      type = ExtendedPropertyType{PropertyValue::Type::List};
+      break;
```
andrejtonev, line 1825: you're handling this as a list of ids, right?
I don't think that's right. ExtendedPropertyType should encapsulate the PropertyType and extend the types with its subtypes. It's used for schema.

src/storage/v2/property_value.hpp
```diff
@@ -152,6 +156,39 @@ class PropertyValueImpl {
...
+                             allocator_type const &alloc = allocator_type{})
+      : alloc_{alloc}, vector_index_id_v{.vector_index_ids_ = vector_index_ids, .vector_ = std::move(vector)} {}
+
+  explicit PropertyValueImpl(vector_index_id_t &&vector_index_ids, utils::small_vector<float> &&vector,
+                             allocator_type const &alloc = allocator_type{})
+      : alloc_{alloc},
+        vector_index_id_v{.vector_index_ids_ = std::move(vector_index_ids), .vector_ = std::move(vector)} {}
+
+  explicit PropertyValueImpl(const vector_index_id_t &vector_index_ids, list_t const &list,
+                             allocator_type const &alloc = allocator_type{})
```
andrejtonev: Again, I think this would be cleaner if the vector index ids and cache was encapsulated into its own struct

src/storage/v2/property_store.cpp
andrejtonev: I got it, you are saying this type can only be used to point to a vector, so it's a list anyway. Okay

src/storage/v2/property_value.hpp
```diff
@@ -1507,6 +1634,22 @@ inline PropertyValue ToPropertyValue(const ExternalPropertyValue &value, NameIdM
...
+      }
+
+      utils::small_vector<float> vector;
+      const auto &internal_vector = value.ValueVectorIndexList();
+      vector.reserve(internal_vector.size());
+      for (auto elem : internal_vector) {
+        vector.push_back(elem);
+      }
+      return PropertyValue(std::move(vector_index_ids), std::move(vector));
+    }
```
andrejtonev, line 1619: need to check while pair reviewing

src/storage/v2/vertex_accessor.cpp
```diff
@@ -53,6 +46,15 @@ void HandleTypeConstraintViolation(Storage const *storage, ConstraintViolation c
                               storage->LabelToName(violation.label),
                               storage->PropertyToName(*violation.properties.begin()));
 }
+
+std::optional<PropertyValue> TryConvertToVectorIndexProperty(Storage *storage, Vertex *vertex, PropertyId property,
+                                                             const PropertyValue &value) {
+  if (!value.IsAnyList() || value.IsVectorIndexId()) return std::nullopt;
+  auto vector_index_ids =
```
andrejtonev, line 54: need to go over this file while pair reviewing

src/storage/v2/durability/wal.cpp
andrejtonev: as discussed, move logic to a new GetProperty function in PropertyStore, with storage as one of the args
or default arg to null and throw if mismatch between type and args

src/storage/v2/disk/storage.cpp
andrejtonev: this is a copy of the line below. probably bad merge

src/query/typed_value.cpp
andrejtonev: see if this is faster or in any way different

      list_v = decltype(list_v)(std::move_iterator(vec.begin()), std::move_iterator(vec.end()));

src/query/typed_value.cpp
```diff
@@ -336,6 +349,19 @@ TypedValue::TypedValue(storage::PropertyValue &&other, storage::NameIdMapper *na
...
+      if (vec.empty()) {
+        type_ = Type::Null;
+        break;
+      }
```
andrejtonev: should this be an assert or throw instead of a quiet conversion to null?

andrejtonev: as discussed, this should NOT happen. ProprtyStore reads and write should happen under the vertex/edge lock and so should either see NULL or VectorIndexId with a list of indices

src/storage/v2/durability/serialization.cpp
andrejtonev: nope

src/storage/v2/durability/snapshot.cpp
```diff
@@ -9505,73 +9491,84 @@ RecoveredSnapshot LoadCurrentVersionSnapshot(Decoder &snapshot, std::filesystem:
...
+                             .metric_kind = MetricFromName(*metric),
+                             .dimension = static_cast<uint16_t>(*dimension),
+                             .resize_coefficient = static_cast<uint16_t>(*resize_coefficient),
+                             .capacity = *capacity,
+                             .scalar_kind = static_cast<unum::usearch::scalar_kind_t>(*scalar_kind)};
+
+        auto entry_count = snapshot.ReadUint();
+        if (!entry_count) throw RecoveryFailure("Couldn't read vector index entry count!");
+
+        std::unordered_map<Gid, utils::small_vector<float>> index_entries;
```
andrejtonev: would this be faster if this was a flatmap?

src/storage/v2/property_store.cpp
andrejtonev: either sort in property store/value or read then sort then check

src/storage/v2/property_store.cpp
andrejtonev: add a comment

src/storage/v2/property_value.hpp
```diff
@@ -1507,6 +1634,22 @@ inline PropertyValue ToPropertyValue(const ExternalPropertyValue &value, NameIdM
...
+      for (const auto &str : external_vector_index_ids) {
+        vector_index_ids.push_back(mapper->NameToId(str));
+      }
+
+      utils::small_vector<float> vector;
+      const auto &internal_vector = value.ValueVectorIndexList();
+      vector.reserve(internal_vector.size());
+      for (auto elem : internal_vector) {
+        vector.push_back(elem);
+      }
```
andrejtonev: no need to do a loop, just copy

src/storage/v2/property_value.hpp
```diff
@@ -1507,6 +1634,22 @@ inline PropertyValue ToPropertyValue(const ExternalPropertyValue &value, NameIdM
       return PropertyValue(value.ValueDoubleList());
     case PropertyValueType::NumericList:
       return PropertyValue(value.ValueNumericList());
+    case PropertyValueType::VectorIndexId: {
+      typename PropertyValue::vector_index_id_t vector_index_ids;
+      const auto &external_vector_index_ids = value.ValueVectorIndexIds();
+      vector_index_ids.reserve(external_vector_index_ids.size());
+      for (const auto &str : external_vector_index_ids) {
+        vector_index_ids.push_back(mapper->NameToId(str));
+      }
```
andrejtonev: use a transform

src/storage/v2/property_value.hpp
andrejtonev: went over

src/storage/v2/property_value.hpp
```diff
@@ -1554,12 +1697,28 @@ inline ExternalPropertyValue ToExternalPropertyValue(const PropertyValue &value,
...
+      for (const auto &id : internal_vector_index_ids) {
+        vector_index_ids.push_back(mapper->IdToName(id));
+      }
+
+      utils::small_vector<float> vector;
+      const auto &internal_vector = value.ValueVectorIndexList();
+      vector.reserve(internal_vector.size());
+      for (const auto &elem : internal_vector) {
+        vector.push_back(elem);
+      }
```
andrejtonev: same as above
1. use transform
2. just copy

src/storage/v2/property_value.hpp
```diff
@@ -1577,7 +1736,7 @@ struct ExtendedPropertyType {
   explicit ExtendedPropertyType(EnumTypeId enum_type) : type{PropertyValueType::Enum}, enum_type{enum_type} {}
 
   explicit ExtendedPropertyType(const PropertyValue &val) {
-    if (val.IsAnyList()) {
+    if (val.IsAnyList() || val.IsVectorIndexId()) {
```
andrejtonev, line 1699: Does it make sense to change IsAnyList to return true for VectorIndexId as well?

src/storage/v2/vertex_accessor.cpp
```diff
@@ -364,7 +373,7 @@ Result<PropertyValue> VertexAccessor::SetProperty(PropertyId property, const Pro
 
   if (vertex_->deleted) return std::unexpected{Error::DELETED_OBJECT};
 
-  PropertyValue old_value;
+  auto old_value = vertex_->properties.GetProperty(property);
```
andrejtonev: leave inside lambda because it will be executed under the AtomicMemBlock

src/storage/v2/vertex_accessor.cpp
```diff
@@ -374,17 +383,20 @@ Result<PropertyValue> VertexAccessor::SetProperty(PropertyId property, const Pro
...
-    // current code always follows the logical pattern of "create a delta" and
-    // "modify in-place". Additionally, the created delta will make other
-    // transactions get a SERIALIZATION_ERROR.
-    if (skip_duplicate_write && old_value == new_value) {
-      return true;
+    if (old_value.IsVectorIndexId()) {
+      // vector is stored in the index
+      auto old_value_vector = storage_->indices_.vector_index_.GetVectorPropertyFromIndex(
+          vertex, storage_->name_id_mapper_->IdToName(old_value.ValueVectorIndexIds()[0]));
+      old_value.ValueVectorIndexList() = std::move(old_value_vector);
```
andrejtonev: This should be part of the GetProperty from PropertyStore with the storage overload

src/storage/v2/vertex_accessor.cpp
```diff
@@ -374,17 +383,20 @@ Result<PropertyValue> VertexAccessor::SetProperty(PropertyId property, const Pro
...
+    } else {
```
andrejtonev: this else is not correct. we still want to check if the old_value is equal to the new_value

src/storage/v2/vertex_accessor.cpp
```diff
@@ -421,18 +433,32 @@ Result<PropertyValue> VertexAccessor::SetProperty(PropertyId property, const Pro
...
+  if (!storage_->indices_.vector_index_.Empty()) {
+    auto transform = [&](const auto &pair) -> std::pair<storage::PropertyId, storage::PropertyValue> {
+      const auto &[property_id, property_value] = pair;
+      if (auto converted = TryConvertToVectorIndexProperty(storage_, vertex_, property_id, property_value)) {
+        return {property_id, std::move(*converted)};
+      }
+      return {property_id, property_value};
+    };
+    converted_properties = props | rv::transform(transform) | r::to<std::map>();
+  }
```
andrejtonev, line 455: This will force a copy for all properties if even one vector index exists.
Create a logic that does the conversion/copy only for those properties that require it

src/storage/v2/vertex_accessor.cpp
```diff
@@ -492,6 +519,15 @@ Result<std::vector<std::tuple<PropertyId, PropertyValue, PropertyValue>>> Vertex
   }
 
   utils::MemoryTracker::OutOfMemoryExceptionEnabler oom_exception;
+
+  if (!storage_->indices_.vector_index_.Empty()) {
```
andrejtonev, line 519: add a comment

src/storage/v2/vertex_accessor.cpp
```diff
@@ -514,11 +550,19 @@ Result<std::vector<std::tuple<PropertyId, PropertyValue, PropertyValue>>> Vertex
...
+      if (new_value.IsVectorIndexId()) {
+        auto old_value_vector = storage->indices_.vector_index_.GetVectorPropertyFromIndex(
+            vertex, storage->name_id_mapper_->IdToName(new_value.ValueVectorIndexIds()[0]));
+        if (skip_duplicate_update && old_value_vector == new_value.ValueVectorIndexList()) continue;
+        if (old_value.IsVectorIndexId()) {
+          old_value.ValueVectorIndexList() = std::move(old_value_vector);
+        }
+      } else {
+        if (skip_duplicate_update && old_value == new_value) continue;
+      }
```
andrejtonev: clean up

src/storage/v2/vertex_accessor.cpp
```diff
@@ -595,11 +640,18 @@ Result<PropertyValue> VertexAccessor::GetProperty(PropertyId property, View view
...
     deleted = vertex_->deleted;
     delta = vertex_->delta;
-    return vertex_->properties.GetProperty(property);
+
+    auto prop_value = vertex_->properties.GetProperty(property);
+    if (prop_value.IsVectorIndexId()) {
+      prop_value.ValueVectorIndexList() = storage_->indices_.vector_index_.GetVectorPropertyFromIndex(
+          vertex_, storage_->name_id_mapper_->IdToName(prop_value.ValueVectorIndexIds()[0]));
+    }
+    return prop_value;
```
andrejtonev: GetProp with storage

src/storage/v2/vertex_accessor.cpp
```diff
@@ -667,6 +719,12 @@ Result<std::map<PropertyId, PropertyValue>> VertexAccessor::Properties(View view
     auto guard = std::shared_lock{vertex_->lock};
     deleted = vertex_->deleted;
     properties = vertex_->properties.Properties();
+    r::for_each(properties, [&](auto &property_value_pair) {
+      if (property_value_pair.second.IsVectorIndexId()) {
+        property_value_pair.second.ValueVectorIndexList() = storage_->indices_.vector_index_.GetVectorPropertyFromIndex(
+            vertex_, storage_->name_id_mapper_->IdToName(property_value_pair.second.ValueVectorIndexIds()[0]));
+      }
```
andrejtonev: handle with GetProp with storage overload

src/storage/v2/indices/vector_index.cpp
```diff
@@ -110,77 +77,118 @@ bool VectorIndex::CreateIndex(const VectorIndexSpec &spec, utils::SkipList<Verte
...
+void VectorIndex::TryAddVertexToIndex(VectorIndexSpec &spec, Vertex &vertex, NameIdMapper *name_id_mapper,
+                                      std::optional<std::size_t> thread_id) {
+  if (!std::ranges::contains(vertex.labels, spec.label_id)) {
+    return;
+  }
+  auto property = vertex.properties.GetProperty(spec.property);
+  if (property.IsNull()) {
+    return;
+  }
+  const auto vector = [&]() {
```
andrejtonev: use std::invoke()

src/storage/v2/indices/vector_index.cpp
```diff
@@ -110,77 +77,118 @@ bool VectorIndex::CreateIndex(const VectorIndexSpec &spec, utils::SkipList<Verte
...
+    if (property.IsVectorIndexId()) {
+      auto &ids = property.ValueVectorIndexIds();
+      ids.push_back(name_id_mapper->NameToId(spec.index_name));
+      return GetVectorPropertyFromIndex(&vertex, name_id_mapper->IdToName(ids[0]));
+    }
+    return ListToVector(property);
```
andrejtonev: comments

src/storage/v2/indices/vector_index.cpp
```diff
@@ -110,77 +77,118 @@ bool VectorIndex::CreateIndex(const VectorIndexSpec &spec, utils::SkipList<Verte
...
+  }();
+  auto &[mg_index, _] = pimpl->index_.at({spec.label_id, spec.property});
+  UpdateVectorIndex(mg_index, spec, &vertex, vector, thread_id);
```
andrejtonev: move vector (looks like a copy at the moment)

src/storage/v2/indices/vector_index.cpp
```diff
@@ -110,77 +77,118 @@ bool VectorIndex::CreateIndex(const VectorIndexSpec &spec, utils::SkipList<Verte
...
-                                                 std::optional<SnapshotObserverInfo> const &snapshot_info) {
-  auto &[mg_index, mutable_spec] = pimpl->index_.at({spec.label_id, spec.property});
-  PopulateVectorIndexMultiThreaded(*mg_index, mutable_spec, vertices, snapshot_info, TryAddVertexToIndex);
+void VectorIndex::TryAddVertexToIndex(VectorIndexSpec &spec, Vertex &vertex, NameIdMapper *name_id_mapper,
+                                      std::optional<std::size_t> thread_id) {
+  if (!std::ranges::contains(vertex.labels, spec.label_id)) {
+    return;
+  }
+  auto property = vertex.properties.GetProperty(spec.property);
+  if (property.IsNull()) {
```
andrejtonev: what about a prop that is not an array?

src/storage/v2/indices/vector_index.cpp
andrejtonev: Will the following be simpler?
Use GetProp with storage (automatically looks up from index) 
if not already vecids, create a new prop
then pass in the vector where needed 
at the end just move the new prop

src/storage/v2/indices/vector_index.cpp
```diff
@@ -110,77 +77,118 @@ bool VectorIndex::CreateIndex(const VectorIndexSpec &spec, utils::SkipList<Verte
...
-  const auto &label_prop = it->second;
+  const auto label_prop = it->second;
+  auto &[mg_index, spec] = pimpl->index_.at(label_prop);
+  auto locked_index = mg_index.MutableSharedLock();
+
+  std::vector<double> vector(locked_index->dimensions());
+  auto restore_vector_from_index = [&](auto *vertex) {
+    locked_index->get(vertex, vector.data());
+    vertex->properties.SetProperty(label_prop.property(), PropertyValue(vector));
+  };
```
andrejtonev: no need for this to be a lambda, it's called once

src/storage/v2/indices/vector_index.cpp
```diff
@@ -192,68 +200,117 @@ void VectorIndex::Clear() {
...
+      if (old_property_value.IsVectorIndexId()) {
+        return GetVectorPropertyFromIndex(vertex,
+                                          name_id_mapper->IdToName(old_property_value.ValueVectorIndexIds()[0]));
       }
-    }
+      if (old_property_value.IsAnyList()) {
+        return ListToVector(old_property_value);
+      }
+      throw query::VectorSearchException("Vector index property must be a list of floats or integers.");
+    }();
```
andrejtonev: this looks similar to previous functions we had. think about GetProperty with storage

src/storage/v2/indices/vector_index.cpp
```diff
@@ -192,68 +200,117 @@ void VectorIndex::Clear() {
...
+      auto &index_item = pimpl->index_.at(label_prop);
+      UpdateVectorIndex(index_item.mg_index, index_item.spec, vertex, vector_property);
     }
-  });
+  } else if (value.IsNull()) {
+    auto indices = GetIndicesByProperty(property);
+    for (const auto &[_, index_name] : indices) {
+      RemoveVertexFromIndex(vertex, index_name);
+    }
+  }
```
andrejtonev, line 298: handle else
either dmg_assert or throw

src/storage/v2/indices/vector_index.cpp
```diff
@@ -314,67 +404,279 @@ VectorIndex::VectorSearchNodeResults VectorIndex::SearchNodes(std::string_view i
...
+    for (auto label : labels_to_remove) {
+      UpdateOnRemoveLabel(label, vertex, name_id_mapper);
+    }
+    for (auto label : labels_to_add) {
+      UpdateOnAddLabel(label, vertex, name_id_mapper);
+    }
+    for (const auto &[property, value] : property_to_abort) {
+      if (value.IsVectorIndexId()) {
+        UpdateOnSetProperty(property, value, vertex, name_id_mapper);
+      } else {
```
andrejtonev, line 495: add check that the value is NULL

src/query/typed_value.cpp
DavIvek: A move iterator won’t be faster here because the lists aren’t of the same type. Ideally, we should unify all these types (Bolt value, typed value, and property value).

src/storage/v2/indices/vector_index_utils.hpp
DavIvek: As discussed, there’s no need to capture by reference because these are primitive types.

src/storage/v2/indices/vector_index_utils.hpp
DavIvek: As discussed, used in add and remove labels methods.

src/storage/v2/indices/vector_index_utils.hpp
DavIvek: I need to double check why I removed this, maybe we don't have to remove it.

src/storage/v2/property_value.hpp
DavIvek: We use `IsAnyList` in other places as well, such as in the property store when decoding and encoding actual list types.

src/storage/v2/indices/vector_index.cpp
DavIvek: `ListToVector` will throw

PR #3294, opened 2025-09-24
feat: Add HA telemetry collectors

as51340:
The PR adds a support for exposing replication and coordination data to the telemetry server:
- the number of strict_sync, sync and asynchronous replicas
- coordinator's configuration: cluster size and its runtime configuration options

Problems with detecting Docker env:
- https://github.com/moby/moby/issues/47420

### DIFF
--- a/src/telemetry/telemetry.hpp
+++ b/src/telemetry/telemetry.hpp
@@ -22,7 +22,7 @@
 
 namespace memgraph::telemetry {
 
-using FuncSig = const std::function<std::optional<nlohmann::json>()>;
+using FuncSig = std::function<std::optional<nlohmann::json>()>;
 
 /**
  * This class implements the telemetry collector service. It periodically scrapes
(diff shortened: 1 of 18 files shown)


as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - **Running Memgraph instances now also expose replication and coordination data to the telemetry servers. Data which is collected is the number of strict sync, sync and asynchronous replicas from the current main while coordinators expose cluster configuration: cluster size and runtime coordinator settings. Based on the existence of environment variables KUBERNETES_SERVICE_HOST and KUBERNETES_PORT, we are trying to detect whether the user is running Memgraph in K8s or not. [#3294](https://github.com/memgraph/memgraph/pull/3294)**
- [x] **[Documentation PR link](https://github.com/memgraph/documentation/pull/1414)**

src/telemetry/telemetry.hpp
```diff
@@ -22,15 +22,17 @@
 
 namespace memgraph::telemetry {
 
+using FuncSig = const std::function<std::optional<nlohmann::json>()>;
+
```
colinbarry, line 26: Should `FuncSig` be a `const` type? Is there any reason to prefer this to:

```cpp
using FuncSig = const std::function<std::optional<nlohmann::json>()>;
void AddCollector(std::string name, FuncSig func);
```

and then

```cpp
 void AddCollector(std::string name, FuncSig func) {
    collectors_.emplace_back(std::move(name), std::move(func));
  }
```

colinbarry (approved):
Approved! Only thing I noted is the slightly non-idiomatic `const` type (but I don't know the context well enough to say if there is a reason for that.)

src/telemetry/telemetry.hpp
as51340: No good reason, I think it was const before so I kept it but you're right

PR #3309, opened 2025-09-30
Fix performance issues with text index

DavIvek:
The existing logic in the text index was affecting overall system performance, even when a text index wasn’t created. This PR fixes the issue.

### DIFF
--- a/src/storage/v2/indices/text_index.cpp
+++ b/src/storage/v2/indices/text_index.cpp
@@ -42,24 +42,34 @@ void TextIndex::CreateTantivyIndex(const std::string &index_path, const TextInde
   }
 }
 
-std::vector<TextIndexData *> TextIndex::GetApplicableTextIndices(std::span<storage::LabelId const> labels,
-                                                                 std::span<PropertyId const> properties) {
+std::vector<TextIndexData *> TextIndex::LabelApplicableTextIndices(std::span<storage::LabelId const> labels) {
   std::vector<TextIndexData *> applicable_text_indices;
-  auto matches_label = [&](const auto &text_index_data) {
-    return r::any_of(labels, [&](auto labelid) { return labelid == text_index_data.scope; });
-  };
+  for (auto &[_, index_data] : index_) {
+    if (r::any_of(labels, [&](auto label) { return label == index_data.scope; })) {
+      applicable_text_indices.push_back(&index_data);
+    }
+  }
+  return applicable_text_indices;
+}
 
-  auto matches_property = [&](const auto &text_index_data) {
+std::vector<TextIndexData *> TextIndex::GetApplicableTextIndices(const Vertex *vertex,
+                                                                 std::span<storage::LabelId const> labels,
+                                                                 std::span<PropertyId const> properties) {
+  auto matches_property = [&](const auto &text_index_data, const auto &properties_to_check) {
     if (text_index_data.properties.empty()) {  // If no properties are specified, all properties match
       return true;
     }
-    return r::any_of(properties,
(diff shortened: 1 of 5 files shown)


DavIvek:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist

src/storage/v2/indices/text_index.hpp
```diff
@@ -71,7 +73,7 @@ class TextIndex {
 
   void UpdateOnRemoveLabel(LabelId label, const Vertex *vertex, Transaction &tx);
 
-  void UpdateOnSetProperty(const Vertex *vertex, Transaction &tx);
+  void UpdateOnSetProperty(const Vertex *vertex, Transaction &tx, PropertyId property);
```
Josipmrden, line 76: const PropertyId property

src/storage/v2/indices/text_index.cpp
```diff
@@ -86,26 +82,40 @@ void TextIndex::AddNodeToTextIndex(std::int64_t gid, const nlohmann::json &prope
...
+  if (label_applicable_text_indices.empty()) return;
+  const auto vertex_properties = vertex->properties.ExtractPropertyIds();
+  auto applicable_text_indices = GetIndicesMatchingProperties(label_applicable_text_indices, vertex_properties);
   TrackTextIndexChange(tx.text_index_change_collector_, applicable_text_indices, vertex, TextIndexOp::REMOVE);
 }
 
-void TextIndex::UpdateOnSetProperty(const Vertex *vertex, Transaction &tx) {
-  auto applicable_text_indices = GetApplicableTextIndices(vertex->labels, vertex->properties.ExtractPropertyIds());
-  if (applicable_text_indices.empty()) return;
+void TextIndex::UpdateOnSetProperty(const Vertex *vertex, Transaction &tx, PropertyId property) {
```
Josipmrden, line 102: also const PropertyId

Josipmrden: check first if text index is empty, I think most of the workloads will not have an index so that is the fastest escape from this method?

src/storage/v2/indices/text_index.cpp
```diff
@@ -86,26 +82,40 @@ void TextIndex::AddNodeToTextIndex(std::int64_t gid, const nlohmann::json &prope
 }
 
 void TextIndex::UpdateOnAddLabel(LabelId label, const Vertex *vertex, Transaction &tx) {
-  auto applicable_text_indices = GetApplicableTextIndices(std::array{label}, vertex->properties.ExtractPropertyIds());
-  if (applicable_text_indices.empty()) return;
+  auto label_applicable_text_indices = LabelApplicableTextIndices(std::array{label});
```
Josipmrden, line 86: i would even here return if there is no index, but do as you wish :pray:

src/storage/v2/indices/text_index.hpp
DavIvek: I would say there is no need since we are passing by value.

src/storage/v2/indices/text_index.cpp
DavIvek: `LabelApplicableTextIndices` goes through the index, which means that if there is no index in the system, it won’t do much work. But yes, some redundant work can still be avoided with an early return, so I’ll add that 👍

src/storage/v2/indices/text_index.hpp
Josipmrden: makes sense, you're right

PR #3310, opened 2025-10-01
feat: Use arrow library in Memgraph for loading parquet files

as51340:


as51340:
Work continued under https://github.com/memgraph/memgraph/pull/3316

PR #3339, opened 2025-10-13
Return current clock time when calling `datetime` with just timezone

colinbarry:
When calling `datetime()` with only a timezone field in the given map:

```cypher
RETURN datetime({timezone:"Europe/Brussels"});
```

this returned a `datetime` with the minimum value for all numeric fields, such as year, month, day, hour, etc:

```
0000-01-01T00:00:00+00[Europe/Brussels]
```

The expected behaviour is that when *only* a timezone is given, `datetime` should compute the current clock time in the given timezone. This PR implements the correct (i.e., same as Neo4j) behaviour, but also breaks the previous `datetime()` result if anyone was relying on that specific behaviour.

The same query will now return something like:

```
2025-10-13 11:20:28.174118000[Europe/Brussels]
```

### DIFF
--- a/src/query/interpret/awesome_memgraph_functions.cpp
+++ b/src/query/interpret/awesome_memgraph_functions.cpp
@@ -1351,11 +1351,11 @@ template <typename T>
 concept IsNumberOrInteger = utils::SameAsAnyOf<T, Number, Integer>;
 
 template <IsNumberOrInteger ArgType>
-int MapNumericParameters(auto &parameter_mappings, const auto &input_parameters) {
-  int num_mapped = 0;
+bool MapNumericParameters(auto &parameter_mappings, const auto &input_parameters) {
+  bool has_mapped_any_field{false};
   for (const auto &[key, value] : input_parameters) {
     if (auto it = parameter_mappings.find(key); it != parameter_mappings.end()) {
-      ++num_mapped;
+      has_mapped_any_field = true;
       if (value.IsInt()) {
         *it->second = value.ValueInt();
       } else if (std::is_same_v<ArgType, Number> && value.IsDouble()) {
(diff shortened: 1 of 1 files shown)


colinbarry:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist

Fixed `datetime()` calls with only a `timezone` parameter (e.g., `datetime({timezone:"Europe/Brussels"})`) to return the current clock time, instead of a time with all numeric fields set to their minimum value. Breaking: this changes the behaviour of calls like `datetime({timezone:"Europe/Brussels"})`: if you need the previous behaviour, then explicitly state one or more numeric fields as `0`, e.g, `datetime({year: 0, timezone:"Europe/Brussels"}`. [#3339](https://github.com/memgraph/memgraph/pull/3339)

    - [x] https://github.com/memgraph/documentation/pull/1441

src/query/interpret/awesome_memgraph_functions.cpp
```diff
@@ -1351,9 +1351,11 @@ template <typename T>
 concept IsNumberOrInteger = utils::SameAsAnyOf<T, Number, Integer>;
 
 template <IsNumberOrInteger ArgType>
-void MapNumericParameters(auto &parameter_mappings, const auto &input_parameters) {
+int MapNumericParameters(auto &parameter_mappings, const auto &input_parameters) {
+  int num_mapped = 0;
```
DavIvek: Is there a reason this is an int? It looks like it could just be a boolean flag.

colinbarry: Yeah, you're correct. It isn't useful to know how many fields were mapped, just whether none or any were. I'll change it.

PR #3342, opened 2025-10-13
Fix in list cache invalidation + perf improvements

imilinovic:
- Fix in list cache invalidation

Perf improvements:
- ExpressionEvaluator no longer, for Identifier evaluation, does a lookup with symbol table to get a symbol to then lookup against the frame. We directly just retrieve from the frame without that lookup.
- List + Map literals now move into TypedValue rather than copy
- GetProperty does direct RVO construction rather than move assignment
- Statically optimise a branch to assume there are no deltas.

### DIFF
--- a/src/query/frontend/semantic/symbol_table.hpp
+++ b/src/query/frontend/semantic/symbol_table.hpp
@@ -68,7 +68,7 @@ class SymbolTable final {
   const auto &table() const { return table_; }
 
   int32_t position_{0};
-  absl::flat_hash_map<int32_t, Symbol> table_;
+  std::map<int32_t, Symbol> table_;
 };
 
 }  // namespace memgraph::query


imilinovic:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Fixed a cache invalidation bug which would cause the IN LIST operator to produce wrong results. [#3342](https://github.com/memgraph/memgraph/pull/3342)
    - General performance improvements around the symbol table, expression evaluations and TypedValues. Users can expect noticeable improvement in some queries. [#3342](https://github.com/memgraph/memgraph/pull/3342)

src/query/dependant_symbol_visitor.hpp
```diff
@@ -0,0 +1,228 @@
...
+    MG_ASSERT(identifier.symbol_pos_ != -1);
+    dependencies_.insert(identifier.symbol_pos_);
+  }
+
+  void Visit(ListLiteral &list_literal) override {
+    std::ranges::for_each(list_literal.elements_, [this](auto *arg) { arg->Accept(*this); });
+  }
+
+  void Visit(MapLiteral &map_literal) override {
+    std::ranges::for_each(map_literal.elements_, [this](const auto &pair) { pair.second->Accept(*this); });
```
andrejtonev, line 51: Do we only need to accept on the value?

src/query/dependant_symbol_visitor.hpp
```diff
@@ -0,0 +1,228 @@
...
+    }
+    if (slice.upper_bound_) {
+      slice.upper_bound_->Accept(*this);
+    }
+  }
+
+  void Visit(Function &func) override {
+    if (!IsFunctionPure(func.function_name_)) {
+      is_cacheable_ = false;
+      return;
```
andrejtonev, line 108: Are we not interested in the symbols of a non pure function?

src/query/dependant_symbol_visitor.hpp
```diff
@@ -0,0 +1,228 @@
...
+    // Remove temporary symbol that is only valid within the list comprehension
+    dependencies_.erase(list_comp.identifier_->symbol_pos_);
+  }
+
+  void Visit(PatternComprehension & /*pattern_comp*/) override {
+    is_cacheable_ = false;  // Not cacheable due to pattern matching
+  }
+
+  void Visit(Exists & /*exists*/) override {
+    is_cacheable_ = false;  // Not cacheable due to pattern/subquery
```
andrejtonev, line 204: Is this always not cacheable?

src/query/dependant_symbol_visitor.hpp
```diff
@@ -0,0 +1,228 @@
...
+  }
+
+  // Named expressions and other types
+  void Visit(NamedExpression &named_expr) override { named_expr.expression_->Accept(*this); }
+  void Visit(MapProjectionLiteral & /*map_proj*/) override {
+    // Safe to cache - no action needed
```
andrejtonev: Is it impossible to have symbols here?

src/query/dependant_symbol_visitor.hpp
imilinovic: Yes, we are just doing an AST walk to find all identifiers used in the IN LIST command.

src/query/dependant_symbol_visitor.hpp
imilinovic: As soon as we find a non pure function the whole in list expression in not cacheable anymore... example if you have rand() you can never cache the list.

src/query/dependant_symbol_visitor.hpp
imilinovic: not always, but this was the simplest solution

src/query/dependant_symbol_visitor.hpp
imilinovic: Good catch

PR #3343, opened 2025-10-14
Multi-label based access control

colinbarry:
**Breaking**: This PR fundamentally changes the grammar and semantics of finely-grained label authorisation. The old grammar and semantics are no longer supported.

- Rules for defining label permissions are now a list of labels and (optionally) a `MATCHING ANY` or `MATCHING EXACTLY` clause. If the clause is omitted, `MATCHING ANY` is assumed.
```cypher
  GRANT <permissions> ON NODES CONTAINING LABELS :Label1, :Label2 [MATCHING ANY | EXACTLY] TO <role>
  GRANT <permissions> ON NODES CONTAINING LABELS * TO <role>
```

where `<permissions>` can be:  `NOTHING`, or a comma-separated list containing one or more of  `READ` | `UPDATE` | `CREATE` | `DELETE`, or `*` for all permissions.

- Rules with `MATCHING ANY` will grant permissions for any nodes with any of the given labels, *regardless of the existence of other labels on the nodes*. For example:
```cypher
  GRANT READ ON NODES CONTAINING LABELS :L1, :L2 MATCHING ANY TO alice
```
will allow alice to read nodes `(:L1), (:L2), `(:L1:L2)` and even `(:L1:L3)`. Despite her having no explicit permission on `L3`, her `READ` permission on `:L1` will suffice to read `(:L1:L3)`.
 
- Rules with `MATCHING EXACTLY` will grant for permissions for nodes with *exactly* the given labels: no more, and no less. For example:

```cypher
  GRANT READ ON NODES CONTAINING LABELS :L1, :L2 MATCHING EXACTLY TO alice
```

will allow alice to read node `(:L1:L2)`, but not `(:L1)` (because it is missing labels), nor `(:L1:L2:L3)` (because of the additional `:L3` label.

- For defining permissions on relationships, use the following grammar. Note that we can't specify a `MATCHING` clause, because edges always have exactly one edge type.

```cypher
GRANT <permissions> ON EDGES CONTAINING TYPES :FOLLOWS TO <role>
GRANT <permissions> ON EDGES CONTAINING TYPES * TO <role>
```

- Permissions can also revoke using a `REVOKE` statement:
```cypher
REVOKE CREATE, DELETE ON NODES CONTAINING LABELS :L1 FROM alice
REVOKE UPDATE ON EDGES CONTAINING TYPES :BOUGHT FROM alice
```

- There is no longer a heirarchy of permissions. `CREATE`, `READ`, `UPDATE`, and `DELETE` are discrete permissions and can be granted and revoked independently.

### DIFF
--- a/src/auth/auth.cpp
+++ b/src/auth/auth.cpp
@@ -363,17 +363,18 @@ void MigrateVersions(kvstore::KVStore &store) {
         try {
           auto data = nlohmann::json::parse(value);
 
-          if (data.contains("fine_grained_access_handler") && data["fine_grained_access_handler"].is_object()) {
-            auto fg_data = data["fine_grained_access_handler"];
-
+          auto const fg_it = data.find("fine_grained_access_handler");
+          if (fg_it != data.end() && fg_it->is_object()) {
             for (auto const &perm_type : {"label_permissions", "edge_type_permissions"}) {
-              if (fg_data.contains(perm_type) && fg_data[perm_type].is_object()) {
-                auto &perm_data = fg_data[perm_type];
+              auto const perm_it = fg_it->find(perm_type);
+              if (perm_it != fg_it->end() && perm_it->is_object()) {
+                auto &perm_data = *perm_it;
 
-                if (perm_data.contains("global_permission") && perm_data["global_permission"].is_number_integer()) {
-                  auto const v2_perm = perm_data["global_permission"].template get<int64_t>();
+                auto const global_perm_it = perm_data.find("global_permission");
+                if (global_perm_it != perm_data.end() && global_perm_it->is_number_integer()) {
+                  auto const v2_perm = global_perm_it->template get<int64_t>();
                   if (v2_perm >= 0) {
-                    perm_data["global_permission"] = convert_v2_to_v3_permissions(static_cast<uint64_t>(v2_perm));
+                    *global_perm_it = convert_v2_to_v3_permissions(static_cast<uint64_t>(v2_perm));
                   }
                 }
 
@@ -381,7 +382,7 @@ void MigrateVersions(kvstore::KVStore &store) {
               }
             }
 
-            data["fine_grained_permissions"] = fg_data;
+            data["fine_grained_permissions"] = std::move(*fg_it);
             data.erase("fine_grained_access_handler");
             puts.emplace(key, data.dump());
           }


colinbarry:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Changed fine-grained access configuration syntax and semantics to be more flexible. Permissions can be set on groups of labels matching any or exactly what is set in the `GRANT` statement. `CREATE`, `READ`, `UPDATE`, and `DELETE` permissions are no longer organised in a hierarchy and can be granted or revoked independently. This is a breaking change, and the fine-grained access syntax and semantics of Memgraph 3.6 and earlier are no longer supported. Migrated databases must be manually configured to grant permissions based on the new rules.[#3343](https://github.com/memgraph/memgraph/pull/3343)
- [x] **[[ Documentation PR link memgraph/documentation#1425 ](https://github.com/memgraph/documentation/pull/1452)]**

src/auth/models.cpp
```diff
@@ -16,6 +16,9 @@
 #include <gflags/gflags.h>
 #include <nlohmann/json.hpp>
 
+#include <range/v3/algorithm/all_of.hpp>
+#include <range/v3/algorithm/any_of.hpp>
+#include <range/v3/algorithm/find_if.hpp>
```
andrejtonev: could we use the std version instead of range v3?

src/auth/models.cpp
```diff
@@ -183,51 +192,63 @@ std::string PermissionLevelToString(PermissionLevel level) {
...
+
+  if ((first_global.has_value() && first_global.value() == 0) ||
+      (second_global.has_value() && second_global.value() == 0)) {
+    global_permission = 0;
+  } else {
+    uint64_t combined_global = first_global.value_or(0) | second_global.value_or(0);
+    if (combined_global != 0) {
+      global_permission = combined_global;
+    }
   }
```
andrejtonev, line 235: This is confusing. I assume std::nullopt does not influence, while NOTHING trumps all other permissions and denies everything. Everything else is then combined with an or.
Is that right?

src/auth/models.cpp
```diff
@@ -183,51 +192,63 @@ std::string PermissionLevelToString(PermissionLevel level) {
...
 
-  for (const auto &[label_name, permission] : second.GetPermissions()) {
-    permissions[label_name] = permission;
+  auto merged_rules = first.GetPermissions();
+  for (auto &&rule2 : second.GetPermissions()) {
+    auto it = r::find_if(merged_rules, [&](auto const &rule1) {
+      return rule1.symbols == rule2.symbols && rule1.matching_mode == rule2.matching_mode;
+    });
```
andrejtonev, line 241: What complexity is this?

src/auth/models.cpp
```diff
@@ -183,51 +192,63 @@ std::string PermissionLevelToString(PermissionLevel level) {
...
+
+    if (it != merged_rules.end()) {
+      if (it->permissions == FineGrainedPermission::NOTHING || rule2.permissions == FineGrainedPermission::NOTHING) {
+        it->permissions = FineGrainedPermission::NOTHING;
+      } else {
+        it->permissions = it->permissions | rule2.permissions;
+      }
```
andrejtonev, line 248: this is the same logic as for the global one. any way to extract the logic under a single lambda?

src/auth/models.cpp
```diff
@@ -183,51 +192,63 @@ std::string PermissionLevelToString(PermissionLevel level) {
 }
 
 #ifdef MG_ENTERPRISE
-FineGrainedPermission PermissionToFineGrainedPermission(const uint64_t permission) {
-  if (permission & FineGrainedPermission::CREATE_DELETE) {
-    return FineGrainedPermission::CREATE_DELETE;
+std::string FineGrainedPermissionToString(uint64_t const permission) {
+  if (permission == 0) {
```
andrejtonev: == NOTHING

src/auth/models.cpp
```diff
@@ -323,47 +344,150 @@ bool operator==(const Permissions &first, const Permissions &second) {
...
+  for (auto &&rule : rules_) {
+    if (rule.permissions == FineGrainedPermission::NOTHING) {
+      continue;
+    }
 
-  const auto temp_permission = concrete_permission & fine_grained_permission;
+    if (rule_matches(rule)) {
+      any_rule_matched = true;
+      if ((rule.permissions & fine_grained_permission) != FineGrainedPermission::NOTHING) {
+        return PermissionLevel::GRANT;
```
andrejtonev, line 395: Since we return on any match, that means there is no way to define access to only a subset of labels/properties.
Ex: node has properties p1 and p2. I want a user to only see p1. With this API, I can let them see both p1 and p2, or not see the node with p1 and p2. 
I think we were discussing having some kind of filtering, where the user could see the properties they have access to, even if there are other properties. But these other properties would be hidden from them.

src/auth/models.cpp
```diff
@@ -323,47 +344,150 @@ bool operator==(const Permissions &first, const Permissions &second) {
...
+  }
+
+  return PermissionLevel::DENY;
+}
+
+void FineGrainedAccessPermissions::Grant(std::unordered_set<std::string> const &symbols,
+                                         FineGrainedPermission const fine_grained_permission,
+                                         MatchingMode const matching_mode) {
+  auto it = r::find_if(
+      rules_, [&](auto const &rule) { return rule.symbols == symbols && rule.matching_mode == matching_mode; });
```
andrejtonev, line 425: Could the symbols + match mode be made into a map entry and avoid the linear scans?

src/auth/models.cpp
```diff
@@ -323,47 +344,150 @@ bool operator==(const Permissions &first, const Permissions &second) {
...
+
+  if (it != rules_.end()) {
+    it->permissions = static_cast<FineGrainedPermission>(
+        static_cast<std::underlying_type_t<FineGrainedPermission>>(it->permissions) &
+        ~static_cast<std::underlying_type_t<FineGrainedPermission>>(fine_grained_permission));
+
+    // If no permissions remain, remove the rule entirely.
+    if (it->permissions == FineGrainedPermission::NOTHING) {
+      rules_.erase(it);
+    }
```
andrejtonev: But, till now, the NOTHING looked like a deliberate deny and would trump anything else. 
I get that this is 0 and that is why you are removing it. Use 0 in that case.
Also you are not checking if this is NOTHING already, so I could Revoke anything off of NOTHING and remove the rule completely.

src/auth/models.cpp
```diff
@@ -372,8 +496,22 @@ nlohmann::json FineGrainedAccessPermissions::Serialize() const {
...
+
+  nlohmann::json rules_json = nlohmann::json::array();
+  for (const auto &rule : rules_) {
+    nlohmann::json rule_json;
+    rule_json[kSymbols] = std::vector<std::string>(rule.symbols.begin(), rule.symbols.end());
+    rule_json[kGranted] = static_cast<uint64_t>(rule.permissions);
+    rule_json[kMatching] = (rule.matching_mode == MatchingMode::EXACTLY) ? "EXACTLY" : "ANY";
+    rules_json.push_back(rule_json);
+  }
+  data[kPermissions] = rules_json;
```
andrejtonev, line 531: auth.cpp has migration logic for the durable auth data. 
This should bump the version and implement the migration logic. 
We have to support automatic migration.

src/auth/models.hpp
```diff
@@ -228,34 +237,55 @@ class UserImpersonation {
...
-  void Grant(const std::string &permission, FineGrainedPermission fine_grained_permission);
+  PermissionLevel Has(std::span<const std::string> symbols, FineGrainedPermission fine_grained_permission) const;
+
+  PermissionLevel HasGlobal(FineGrainedPermission fine_grained_permission) const;
 
-  void Revoke(const std::string &permission);
+  void Grant(std::unordered_set<std::string> const &symbols, FineGrainedPermission fine_grained_permission,
+             MatchingMode matching_mode = MatchingMode::ANY);
+
+  void GrantGlobal(FineGrainedPermission fine_grained_permission);
```
andrejtonev, line 267: Could this be simplified and the global permissions treated as any other rule in rules_?

src/query/frontend/opencypher/grammar/MemgraphCypher.g4
```diff
@@ -441,11 +447,11 @@ setRole : SET ( ROLE | ROLES ) FOR user=userOrRoleName TO roles=listOfSymbolicNa
 
 clearRole : CLEAR ( ROLE | ROLES ) FOR user=userOrRoleName ( ON db=listOfSymbolicNames )? ;
 
-grantPrivilege : GRANT ( ALL PRIVILEGES | privileges=grantPrivilegesList ) TO userOrRole=userOrRoleName ;
+grantPrivilege : GRANT ( ALL PRIVILEGES | systemPrivileges=privilegesList | entityPrivileges=entityPrivilegeList ) TO userOrRole=userOrRoleName ;
 
-denyPrivilege : DENY ( ALL PRIVILEGES | privileges=privilegesList ) TO userOrRole=userOrRoleName ;
+denyPrivilege : DENY ( ALL PRIVILEGES | systemPrivileges=privilegesList ) TO userOrRole=userOrRoleName ;
```
andrejtonev, line 452: Why can't I DENY entityPrivileges?

src/auth/models.cpp
colinbarry: The global and non-global cases are similar, but different enough that forcing them into a single lambda would make the code more complicated, not less. The issue is that for global permissions, they may be missing, i.e., `nullopt`, as well as `NOTHING`, and these mean two different things. I will leave as is.

src/auth/models.cpp
colinbarry: Yes... but it wouldn't help with the "hot path" of `Has()` matching `ANY` of the symbols. If someone has a `GRANT READ ON NODES CONTAINING LABELS :L1, :L2, :L3 MATCHING ANY`, for a node with `(:L1)` we need to lookup rules that have :L1, regardless of the other symbols.

I suppose we could create a parallel index for ANY rules that are keyed on individual labels. But that would add complexity and not necessarily a speed improvment over the vector. 

Not sure if it is worth it for this PR: we could make the change later if this proves slow.

src/auth/models.hpp
colinbarry: That's how it was originally, but it made other stuff a little more complicated, because we'd have to construct an unordered sets with a single wildcard `*` as a global sentinel. I think separating the functions like this makes for a cleaner API.

src/auth/models.cpp
colinbarry: You are correct in the logic, and also correct that it isn't easy to parse at all! I'll rewrite with explicit steps that make it clear how global permissions are merged.

src/query/frontend/opencypher/grammar/MemgraphCypher.g4
colinbarry: History, I guess? It's always been this way with fine grained access: you `GRANT`, `REVOKE`, and use `GRANT NOTHING` to effectively `DENY`.  Wasn't the scope of this PR to bring that into line, so I didn't.

src/auth/models.cpp
colinbarry: Added migration to V3, and test for user and role access similar to the existing migration test. Because migration requires a manual security audit, there is no need for automated migration.

src/auth/models.cpp
colinbarry: You are correct: this is bad, from a performance stand point. In my defence, it already was bad, as this Merge is actually a "hot path" when checking label/edge access.

src/auth/models.cpp
colinbarry: No talk of property filtering in the spec for this. I'm sure it's something we'll do when we fix-up auth in the future, but it's beyond the scope of this PR.

src/auth/models.cpp
colinbarry: Good catch. Fixed, and added unit tests to capture this behaviour.

src/auth/models.cpp
colinbarry: Still `O(N*M)` on a `Merge`, but I now lazily evaluate these merged permissions and cache them in `FineGrainedAuthChecker`, so we only pay the cost once, rather than per label/edge access. Assuming that users/roles _probably_ won't have a massive list of rules, I think this is fine (and if not, we can optimise later.)

src/auth/auth.cpp
```diff
@@ -316,6 +317,15 @@ void MigrateVersions(kvstore::KVStore &store) {
 
     spdlog::info("Auth storage migration to V2 completed successfully");
   }
+
+  if (version_str == kVersionV2) {
+    spdlog::info("Migrating auth storage from V2 to V3");
+    spdlog::warn(
+        "IMPORTANT: Review your security policy and explicitly configure finely grained access rules where needed.");
```
andrejtonev, line 324: I'm still not sure why we can't migrate the previous version to the new one?
What old information can't be converted to the new api?

colinbarry: Global rules can be (and now are.) The issue is that any historic rules using specific combinations of labels or edge types don't map easily to the new rule semantics. For example:

```cypher
GRANT CREATE_DELETE TO LABELS :L1, :L2, :L3 FOR alice
```

would correspond to matching any vertex with any combination of labels `:L1`, `:L2`, `:L3`, but *no* additional labels. This is not easily expressed without creating a huge amount of rules.

We decided overall that a better approach is, for the few users who use LBA, to audit and update to the new system.

src/auth/auth.cpp
```diff
@@ -323,8 +323,83 @@ void MigrateVersions(kvstore::KVStore &store) {
...
+      for (auto it = store.begin(prefix); it != store.end(prefix); ++it) {
+        auto const &[key, value] = *it;
+        try {
+          auto data = nlohmann::json::parse(value);
+
+          if (data.contains("fine_grained_access_handler") && data["fine_grained_access_handler"].is_object()) {
+            auto fg_data = data["fine_grained_access_handler"];
+
+            for (auto const &perm_type : {"label_permissions", "edge_type_permissions"}) {
+              if (fg_data.contains(perm_type) && fg_data[perm_type].is_object()) {
```
andrejtonev: I prefer .find() since if the key exists you don't need to do another lookup

src/auth/auth.cpp
```diff
@@ -323,8 +323,83 @@ void MigrateVersions(kvstore::KVStore &store) {
...
+                  if (v2_perm >= 0) {
+                    perm_data["global_permission"] = convert_v2_to_v3_permissions(static_cast<uint64_t>(v2_perm));
+                  }
+                }
+
+                perm_data["permissions"] = nlohmann::json::array();
+              }
+            }
+
+            data["fine_grained_permissions"] = fg_data;
```
andrejtonev: can we move the data around instead of copying it twice?

PR #3356, opened 2025-10-20
feat: Support loading parquet files from local disk

as51340:
A new query is added `LOAD PARQUET FROM...`. New library `arrow` is added which is built through `libs` folder functionality. Loading parquet files uses a prefetcher thread which operates in column batches and from such column-wise oriented data, assembles rows which the main execution thread at the end just pulls. That's how we hide the cost of assembling rows from column-oriented data. To avoid copying allocated data from the background thread to the main thread, we make use of thread-safe memory resource. A new data queue is added which is used as a shared resource between the prefetcher and the main thread. 

The class `utils::Date` supports instantiation now using the number of days and the number of microseconds passed since UNIX epoch.

### DIFF
--- a/src/query/plan/operator.cpp
+++ b/src/query/plan/operator.cpp
@@ -7737,10 +7737,6 @@ class LoadParquetCursor : public Cursor {
 
     auto *mem = context.evaluation_context.memory;
     if (UNLIKELY(!reader_.has_value())) {
-      Frame local_frame(0);
-      SymbolTable const symbol_table;
-      DbAccessor *dba = nullptr;
-
       auto evaluator = PrimitiveLiteralExpressionEvaluator{context.evaluation_context};
       auto maybe_file = self_->file_->Accept(evaluator).ValueString();
 


src/query/arrow_parquet/reader.cpp
```diff
@@ -0,0 +1,526 @@
...
+
+using memgraph::query::TypedValue;
+using memgraph::utils::Date;
+using memgraph::utils::Duration;
+using memgraph::utils::LocalDateTime;
+using memgraph::utils::LocalTime;
+using memgraph::utils::MemoryResource;
+
+namespace {
+auto ToHexString(const uint8_t *data, size_t const size) -> std::string {
```
andrejtonev: this seems like it should already exist in the std, boost or our utils.
if it really is a new feature, put it under utils

src/query/arrow_parquet/reader.cpp
```diff
@@ -0,0 +1,526 @@
...
+      auto int_array = std::static_pointer_cast<arrow::Int64Array>(column);
+      return [int_array, resource](int64_t const i) -> TypedValue {
+        return int_array->IsNull(i) ? TypedValue(resource) : TypedValue(int_array->Value(i), resource);
+      };
+    }
+    case arrow::Type::UINT8: {
+      auto int_array = std::static_pointer_cast<arrow::UInt8Array>(column);
+      return [int_array, resource](int64_t const i) -> TypedValue {
+        return int_array->IsNull(i) ? TypedValue(resource)
+                                    : TypedValue(static_cast<int64_t>(int_array->Value(i)), resource);
```
andrejtonev: do we have to cast to signed int?

src/query/arrow_parquet/reader.cpp
```diff
@@ -0,0 +1,526 @@
...
+        arrow::Decimal256 const value(bytes);
+        return TypedValue(value.ToDouble(scale), resource);
+      };
+    }
+    case arrow::Type::BINARY: {
+      auto binary_array = std::static_pointer_cast<arrow::BinaryArray>(column);
+      return [binary_array, resource](int64_t const i) -> TypedValue {
+        if (binary_array->IsNull(i)) return TypedValue(resource);
+        auto const view = binary_array->GetView(i);
+        return TypedValue(ToHexString(reinterpret_cast<const uint8_t *>(view.data()), view.size()), resource);
```
andrejtonev: why are we stringifying this instead of just consuming it as a binary array?

src/query/arrow_parquet/reader.cpp
```diff
@@ -0,0 +1,526 @@
...
+        auto const view = large_binary_array->GetView(i);
+        return TypedValue(ToHexString(reinterpret_cast<const uint8_t *>(view.data()), view.size()), resource);
+      };
+    }
+    case arrow::Type::FIXED_SIZE_BINARY: {
+      auto fixed_binary = std::static_pointer_cast<arrow::FixedSizeBinaryArray>(column);
+      int32_t const width = fixed_binary->byte_width();
+      return [fixed_binary, width, resource](int64_t const i) -> TypedValue {
+        if (fixed_binary->IsNull(i)) return TypedValue(resource);
+        const uint8_t *data = fixed_binary->GetValue(i);
```
andrejtonev, line 263: Do GetValue() copy or reference the value? 
Is there a GetView()?

src/query/arrow_parquet/reader.cpp
```diff
@@ -0,0 +1,526 @@
...
+        if (list_array->IsNull(i)) return TypedValue(resource);
+
+        auto const slice = list_array->value_slice(i);
+        int64_t const list_length = slice->length();
+
+        TypedValue::TVector list_values(resource);
+        list_values.reserve(list_length);
+
+        for (int64_t k = 0; k < list_length; k++) {
+          auto const elem_converter = CreateColumnConverter(slice, resource);
```
andrejtonev: this should be out side the loop

src/query/arrow_parquet/reader.cpp
```diff
@@ -0,0 +1,526 @@
...
+        if (map_array->IsNull(i)) return TypedValue(resource);
+
+        int64_t const offset_start = map_array->value_offset(i);
+        int64_t const offset_end = map_array->value_offset(i + 1);
+
+        TypedValue::TMap map_values(resource);
+
+        for (int64_t k = offset_start; k < offset_end; k++) {
+          auto key = key_strings->GetString(k);
+          auto const val_converter = CreateColumnConverter(values_array, resource);
```
andrejtonev: should be outside the loop

src/query/arrow_parquet/reader.cpp
```diff
@@ -0,0 +1,526 @@
...
+          map_values.emplace(TypedValue::TString{key, resource}, val_converter(k));
+        }
+
+        return {std::move(map_values), resource};
+      };
+    }
+    default: {
+      // Fallback to string conversion
```
andrejtonev, line 311: Why? Why not throw or return null?

src/query/arrow_parquet/reader.cpp
```diff
@@ -0,0 +1,526 @@
...
+
+  // The user knows when to request the next batch
+  std::vector<std::shared_ptr<arrow::Array>> Next() const {
+    auto const res = rbr_->Next();
+    if (!res.ok() || !(*res)) {
+      return {};
+    }
+    auto const &batch = *res;
+
+    std::vector<std::shared_ptr<arrow::Array>> result;
```
andrejtonev, line 347: can we use unique here?

src/query/arrow_parquet/reader.cpp
```diff
@@ -0,0 +1,526 @@
...
+  // Needs to stay alive because of batch reader
+  std::unique_ptr<parquet::arrow::FileReader> file_reader_;
+  std::shared_ptr<arrow::Schema> schema_;
+  int num_columns_;
+  BatchIterator row_it_;
+  utils::MemoryResource *resource_;
+  utils::pmr::vector<Row> rows_;
+  utils::DataQueue<utils::pmr::vector<Row>> work_queue_;
+  Header header_;
+  std::jthread prefetcher_thread_;  // should get destroyed before all other variables that it uses as a reference
```
andrejtonev, line 384: don't see any way to signal stop to the background thread

src/query/arrow_parquet/reader.cpp
```diff
@@ -0,0 +1,526 @@
...
+          }
+
+          auto const num_rows = batch_ref[0]->length();
+          utils::pmr::vector<Row> queued_batch(resource_);
+          queued_batch.reserve(num_rows);
+          for (auto i = 0; i < num_rows; ++i) {
+            // temporary needs to be created
+            // NOLINTNEXTLINE
+            queued_batch.emplace_back(Row{resource_});
+          }
```
andrejtonev, line 415: couldn't this be done via a vector constructor?

src/query/frontend/semantic/required_privileges.cpp
```diff
@@ -247,6 +247,11 @@ class PrivilegeExtractor : public QueryVisitor<void>, public HierarchicalTreeVis
     return false;
   }
 
+  bool PreVisit(LoadParquet & /*unused*/) override {
+    AddPrivilege(AuthQuery::Privilege::READ_FILE);
```
andrejtonev, line 251: is it just read file? it will also write to the db, right?

src/query/interpreter.cpp
```diff
@@ -7306,11 +7306,26 @@ Interpreter::PrepareResult Interpreter::Prepare(ParseRes parse_res, UserParamete
         fmt::format("Query [{}] associated with transaction [{}]", parsed_query.query_string, *current_transaction_));
   }
 
+  auto *const cypher_query = utils::Downcast<CypherQuery>(parsed_query.query);
+  bool has_load_parquet{false};
```
andrejtonev, line 7312: comment to explain what's going on and why it is needed

src/query/plan/operator.cpp
```diff
@@ -7700,6 +7700,101 @@ std::unique_ptr<LogicalOperator> LoadCsv::Clone(AstStorage *storage) const {
...
+
+    auto frame_writer = frame.GetFrameWriter(context.frame_change_collector, context.evaluation_context.memory);
+
+    auto *mem = context.evaluation_context.memory;
+    if (UNLIKELY(!reader_.has_value())) {
+      Frame local_frame(0);
+      SymbolTable const symbol_table;
+      DbAccessor *dba = nullptr;
+      auto evaluator =
+          ExpressionEvaluator(&local_frame, symbol_table, context.evaluation_context, dba, storage::View::OLD);
```
andrejtonev: Couldn't this be PrimitiveLiteralExpressionEvaluator?

src/utils/data_queue.hpp
```diff
@@ -0,0 +1,140 @@
...
+  std::condition_variable writerCv_;
+  std::condition_variable finishCv_;
+
+  std::queue<T> queue_;
+  bool done_{false};
+  std::size_t maxSize_;
+
+  // Must have lock to call this function
+  bool full() const {
+    if (maxSize_ == 0) {
```
andrejtonev, line 36: add a dmg_assert to verify all calls have lock

src/utils/data_queue.hpp
```diff
@@ -0,0 +1,140 @@
...
+// Same as rocksdb/util/work_queue.h but moves in the pop()
+
+namespace memgraph::utils {
+
+template <typename T>
+class DataQueue {
+  std::mutex mutex_;
+  std::condition_variable readerCv_;
+  std::condition_variable writerCv_;
+  std::condition_variable finishCv_;
```
andrejtonev: do we really need 3 conditional variables?

src/utils/data_queue.hpp
```diff
@@ -0,0 +1,140 @@
...
+   * @param item  Item to push onto the queue.
+   * @returns     True upon success, false if `finish()` has been called.  An
+   *               item was pushed iff `push()` returns true.
+   */
+  template <typename U>
+  bool push(U &&item) {
+    {
+      std::unique_lock<std::mutex> lock(mutex_);
+      while (full() && !done_) {
+        writerCv_.wait(lock);
```
andrejtonev: use wait() with lambda check instead of the while + wait

src/utils/data_queue.hpp
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  /**
+   * Promise that `push()` won't be called again, so once the queue is empty
+   * there will never any more work.
+   */
+  void finish() {
+    {
+      std::lock_guard<std::mutex> lock(mutex_);
+      assert(!done_);
```
andrejtonev: use mg_assert

src/utils/data_queue.hpp
```diff
@@ -0,0 +1,140 @@
...
+   * @returns          True upon success.  False if the queue is empty and
+   *                    `finish()` has been called.
+   */
+  bool pop(T &item) {
+    {
+      std::unique_lock<std::mutex> lock(mutex_);
+      while (queue_.empty() && !done_) {
+        readerCv_.wait(lock);
+      }
+      if (queue_.empty()) {
```
andrejtonev, line 91: why shouldn't this just check done_ and return false?
the way it is now looks like we have to exhaust the queue before we can shutdown

src/query/arrow_parquet/reader.cpp
as51340: for uint8_t, uint16_t it's not, for uint32_t and uint64_t it is.

src/query/arrow_parquet/reader.cpp
as51340: Yeah I guess I wanted to make it readable but it's more efficient to use it as const char * so I guess you're right

src/query/arrow_parquet/reader.cpp
as51340: In this case it return const uint8_t *data so here it definitely doesn't copy. I think it depends on the type

src/query/arrow_parquet/reader.cpp
as51340: This is for all types that are supported in parquet but for which there is no direct support in TypedValue.

src/query/arrow_parquet/reader.cpp
as51340: Arrow's api uses return shared_ptr for batch->column(c) so I think this wouldn't be safe

src/query/arrow_parquet/reader.cpp
as51340: yep my bad

src/query/frontend/semantic/required_privileges.cpp
as51340: yes, but load csv has the same privilege that's why I used read_file

src/query/plan/operator.cpp
as51340: TBH I don't know, I think all code in operator.cpp uses ExpressionEvaluator

src/utils/data_queue.hpp
as51340: yeah I think finishCv_ is not necessary

src/utils/data_queue.hpp
as51340: because the prefetched thread is the one which determines when the queue is finished. Prefetcher thread may set the queue as finished and in this way effectively disable popping while the main thread still didn't process all batches

andrejtonev: Okay. Put some comments to make it clear

src/query/plan/operator.cpp
andrejtonev: If this is just a const string, I think you can use the PrimitiveLiterarExpressionEvaluator in the prepare phase.
Take a look in interpreter.cpp and if it works use it, if not, all good.

as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Memgraph now supports loading Parquet files directly from the local filesystem. Users can now use `LOAD PARQUET FROM /path/to/file.parquet` in their Cypher queries for efficient bulk data import. This eliminates the need for manual CSV conversion and provides significantly faster import performance for large datasets. [#3356](https://github.com/memgraph/memgraph/pull/3356)
- [x] **[Documentation PR link](https://github.com/memgraph/documentation/pull/1448)**

PR #3359, opened 2025-10-20
feat: Use data directory for storing received durability files on replica

as51340:
It's impossible to move the file from /tmp to /var/lib/memgraph when MG is deployed on K8s because tmp and /var/lib/memgraph are on different partitions. Data directory (/var/lib/memgraph) is used through PVC while tmp is locally attached storage.

### DIFF
--- a/src/memgraph.cpp
+++ b/src/memgraph.cpp
@@ -361,6 +361,7 @@ int main(int argc, char **argv) {
              .interval = std::chrono::seconds(FLAGS_storage_gc_cycle_sec)},
 
       .durability = {.storage_directory = FLAGS_data_directory,
+                     .root_data_directory = FLAGS_data_directory,
                      .recover_on_startup = FLAGS_data_recovery_on_startup,
                      .snapshot_retention_count = FLAGS_storage_snapshot_retention_count,
                      .wal_file_size_kibibytes = FLAGS_storage_wal_file_size_kib,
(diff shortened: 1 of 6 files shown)


src/rpc/file_replication_handler.cpp
```diff
@@ -46,11 +41,11 @@ std::optional<size_t> FileReplicationHandler::OpenFile(const uint8_t *data, size
   if (!ValidateFileSize(maybe_file_size)) return std::nullopt;
 
   file_size_ = *maybe_file_size;
-  auto const path = tmp_rnd_dir / *maybe_filename;
```
andrejtonev, line 49: rename from filepath to relative path or similar

as51340:
> Lets keep the general logic as is was and just use data_dir/tmp instead of the /tmp Should we clean tmp on startup?

delete on startup or at the beginning of the RPC handler for that specific db

andrejtonev (approved):

as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Replica instance will save incoming WAL files and snapshots into the tmp folder which is located in the data directory. Previously, it was using /tmp but then there was the issue in moving file from /tmp to data_dir because directories were located on different partitions. [#3359](https://github.com/memgraph/memgraph/pull/3359)

PR #3371, opened 2025-10-23
SkipList chunking iterator

andrejtonev:
Divide the SkipList into N (approximately) equal chunks.
Each chunk can be independently scanned over. 
Scanning over each chunk will be equivalent to scanning over the whole SkipList with the regular iterator.

### DIFF
--- a/src/storage/v2/edges_chunked_iterable.cpp
+++ b/src/storage/v2/edges_chunked_iterable.cpp
@@ -177,15 +177,15 @@ EdgesChunkedIterable::Iterator::Iterator(EdgesChunkedIterable::Iterator &&other)
   switch (other.type_) {
     case Type::BY_EDGE_TYPE_IN_MEMORY_CHUNKED:
       new (&in_memory_chunked_by_type_it_)
-          InMemoryEdgeTypeIndex::ChunkedIterable::Iterator(std::move(other.in_memory_chunked_by_type_it_));
+          InMemoryEdgeTypeIndex::ChunkedIterable::Iterator(other.in_memory_chunked_by_type_it_);
       break;
     case Type::BY_EDGE_TYPE_PROPERTY_IN_MEMORY_CHUNKED:
-      new (&in_memory_chunked_by_type_property_it_) InMemoryEdgeTypePropertyIndex::ChunkedIterable::Iterator(
-          std::move(other.in_memory_chunked_by_type_property_it_));
+      new (&in_memory_chunked_by_type_property_it_)
+          InMemoryEdgeTypePropertyIndex::ChunkedIterable::Iterator(other.in_memory_chunked_by_type_property_it_);
       break;
     case Type::BY_EDGE_PROPERTY_IN_MEMORY_CHUNKED:
       new (&in_memory_chunked_by_property_it_)
-          InMemoryEdgePropertyIndex::ChunkedIterable::Iterator(std::move(other.in_memory_chunked_by_property_it_));
+          InMemoryEdgePropertyIndex::ChunkedIterable::Iterator(other.in_memory_chunked_by_property_it_);
       break;
   }
 }
@@ -199,15 +199,15 @@ EdgesChunkedIterable::Iterator &EdgesChunkedIterable::Iterator::operator=(
     switch (other.type_) {
       case Type::BY_EDGE_TYPE_IN_MEMORY_CHUNKED:
         new (&in_memory_chunked_by_type_it_)
-            InMemoryEdgeTypeIndex::ChunkedIterable::Iterator(std::move(other.in_memory_chunked_by_type_it_));
+            InMemoryEdgeTypeIndex::ChunkedIterable::Iterator(other.in_memory_chunked_by_type_it_);
         break;
       case Type::BY_EDGE_TYPE_PROPERTY_IN_MEMORY_CHUNKED:
-        new (&in_memory_chunked_by_type_property_it_) InMemoryEdgeTypePropertyIndex::ChunkedIterable::Iterator(
-            std::move(other.in_memory_chunked_by_type_property_it_));
+        new (&in_memory_chunked_by_type_property_it_)
+            InMemoryEdgeTypePropertyIndex::ChunkedIterable::Iterator(other.in_memory_chunked_by_type_property_it_);
         break;
       case Type::BY_EDGE_PROPERTY_IN_MEMORY_CHUNKED:
         new (&in_memory_chunked_by_property_it_)
-            InMemoryEdgePropertyIndex::ChunkedIterable::Iterator(std::move(other.in_memory_chunked_by_property_it_));
+            InMemoryEdgePropertyIndex::ChunkedIterable::Iterator(other.in_memory_chunked_by_property_it_);
         break;
     }
   }
--- a/src/storage/v2/inmemory/label_property_index.cpp
+++ b/src/storage/v2/inmemory/label_property_index.cpp
@@ -210,7 +210,7 @@ void AdvanceUntilValid_(auto &index_iterator, const auto &end, auto *&current_ve
       if (all_null) continue;
     }
 
-    enum class InBoundResult { UNDER, IN_BOUNDS, IN_BOUNDS_AT_UB, OVER };
+    enum class InBoundResult : uint8_t { UNDER, IN_BOUNDS, IN_BOUNDS_AT_UB, OVER };
 
     auto const value_within_upper_bounds = [](std::optional<utils::Bound<PropertyValue>> const &ub,
                                               PropertyValue const &cmp_value) -> InBoundResult {
@@ -238,7 +238,7 @@ void AdvanceUntilValid_(auto &index_iterator, const auto &end, auto *&current_ve
       return value_within_upper_bounds(ub, cmp_value);
     };
 
-    enum class Result { Skip, NoMoreValidEntries, WithAllBounds };
+    enum class Result : uint8_t { Skip, NoMoreValidEntries, WithAllBounds };
 
     auto bounds_checker = [&]() {
       auto at_boundary_counter = 0;
@@ -303,7 +303,8 @@ void AdvanceUntilValid_(auto &index_iterator, const auto &end, auto *&current_ve
     skip_lower_bound_check = false;
     if (res == Result::Skip) {
       continue;
-    } else if (res == Result::NoMoreValidEntries) {
+    }
+    if (res == Result::NoMoreValidEntries) {
       index_iterator = end;
       break;
     }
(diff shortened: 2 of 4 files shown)


andrejtonev:

---
___

### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist

src/utils/skip_list.hpp
```diff
@@ -177,6 +179,26 @@ constexpr uint8_t SkipListLayerForAverageEqualsEstimation(const uint64_t N) {
...
+  if (k <= 1) return N;  // Bottom layer contains all elements
+  if (N == 0) return 0;  // Empty skip list
+
+  // Calculate (1/2)^(k-1) using bit shifting for efficiency
+  // (1/2)^(k-1) = 1 / (2^(k-1))
+  const uint8_t power = k - 1;
+  if (power >= 64) return 0;  // Result would be too small to represent
+
+  const uint64_t denominator = 1ULL << power;
+  return N / denominator;
```
as51340: return N >> shift

src/utils/skip_list.hpp
```diff
@@ -1495,6 +1642,146 @@ class SkipList final : detail::SkipListNode_base {
...
+    if (head_ == nullptr) {
+      // Return one empty chunk for empty list
+      return {{Chunk{nullptr, nullptr}}};
+    }
+
+    std::array<TNode *, kSkipListMaxHeight> start{};
+    std::array<TNode *, kSkipListMaxHeight> end{};
+    for (int layer = 0; layer < kSkipListMaxHeight; ++layer) {
+      start[layer] = head_->nexts[layer].load(std::memory_order_acquire);
+      end[layer] = nullptr;
```
as51340: I think you don't need this, end should already be initialised to nullptrs

src/utils/skip_list.hpp
```diff
@@ -1495,6 +1642,146 @@ class SkipList final : detail::SkipListNode_base {
...
+              ;
+          }
+          current = current->nexts[layer].load(std::memory_order_acquire);
+        }
+        end[layer] = current;
+      }
+    } else {
+      for (int layer = 0; layer < kSkipListMaxHeight; ++layer) {
+        end[layer] = nullptr;
+      }
```
as51340: end should already be nullptr

src/utils/skip_list.hpp
```diff
@@ -1495,6 +1642,146 @@ class SkipList final : detail::SkipListNode_base {
...
+    }
+
+    if (upper_bound) {
+      find_node<GCPolicy::DoNotRun>(upper_bound.value(), end, succs);
+      for (int layer = 0; layer < kSkipListMaxHeight; ++layer) {
+        // head_ is not a valid node (use next)
+        auto *current = end[layer] == head_ ? head_->nexts[layer].load(std::memory_order_acquire) : end[layer];
+        // Find the first element over the bound
+        while (current != nullptr && (current->obj < upper_bound.value() || current->obj == upper_bound.value())) {
```
as51340: current-> obj <= upper_bound.value(), can you combine checks?

src/utils/skip_list.hpp
```diff
@@ -1495,6 +1642,146 @@ class SkipList final : detail::SkipListNode_base {
...
+      for (int layer = 0; layer < kSkipListMaxHeight; ++layer) {
+        start[layer] = head_->nexts[layer].load(std::memory_order_acquire);
+      }
+    }
+
+    if (upper_bound) {
+      find_node<GCPolicy::DoNotRun>(upper_bound.value(), end, succs);
+      for (int layer = 0; layer < kSkipListMaxHeight; ++layer) {
+        // head_ is not a valid node (use next)
+        auto *current = end[layer] == head_ ? head_->nexts[layer].load(std::memory_order_acquire) : end[layer];
```
as51340, line 1699: end[layer] should at this point be nullptr? In that use nullptr, I think it's easier to understand in that way.
Also, could it be that head_->nexts[layer] is also a nullptr?

src/utils/skip_list.hpp
```diff
@@ -1495,6 +1642,146 @@ class SkipList final : detail::SkipListNode_base {
...
+        // Find the first element over the bound
+        while (current != nullptr && (current->obj < upper_bound.value() || current->obj == upper_bound.value())) {
+          if (!current->marked.load(std::memory_order_acquire)) {
+            while (!current->fully_linked.load(std::memory_order_acquire))
```
as51340, line 1703: wouldn't it be better to use wait(false) on fully_linked and add notify in the code instead of busy loops

src/utils/skip_list.hpp
```diff
@@ -1495,6 +1642,146 @@ class SkipList final : detail::SkipListNode_base {
...
+      cached_layer_nodes.clear();
+      cached_layer_nodes.reserve(ExpectedSizeAtLayer(size(), layer + 1));
+      // Count max-height elements (elements that appear at this layer)
+      TNode *current = start[layer];
+      TNode *layer_end = end[layer];
+      while (current != layer_end) {
+        // Only count nodes that are both fully_linked and not marked
+        if (!current->marked.load(std::memory_order_acquire)) {
+          cached_layer_nodes.push_back(current);
+          while (!current->fully_linked.load(std::memory_order_acquire))
```
as51340, line 1735: use wait

as51340: Maybe it would be more efficient that you try to count all nodes between start and end that aren't marked. If there are more than num_chunks nodes than wait on all of them until they are linked. In the current way you could be waiting for fully_linked while still not having enough nodes because they are marked

src/utils/skip_list.hpp
```diff
@@ -1495,6 +1642,146 @@ class SkipList final : detail::SkipListNode_base {
...
+    if (cached_layer_nodes.empty()) {
+      // If there are no max-height elements, return the complete list as one chunk
+      return {{Chunk{start[0], end[0]}}};
+    }
+
+    // If we have fewer max-height elements than chunks, adjust
+    if (cached_layer_nodes.size() < num_chunks) {
+      layer = 0;  // We got to the bottom layer
+      num_chunks = cached_layer_nodes.size();
+    }
```
as51340, line 1755: Does it make sense then that the loop doesn't check the layer 0

src/storage/v2/vertex_accessor.cpp
```diff
@@ -71,6 +71,7 @@ std::pair<bool, bool> IsVisible(Vertex const *vertex, Transaction const *transac
   if (delta && transaction->isolation_level != IsolationLevel::READ_UNCOMMITTED) {
     // IsolationLevel::READ_COMMITTED would be tricky to propagate invalidation to
     // so for now only cache for IsolationLevel::SNAPSHOT_ISOLATION
+    // TODO Disable for parallel execution v1
```
as51340, line 74: just a reminder for TODO

src/storage/v2/property_constants.hpp
```diff
@@ -33,6 +33,9 @@ static const auto kSmallestEnum = PropertyValue(Enum{EnumTypeId{0}, EnumValueId{
 static const auto kSmallestPoint2d = PropertyValue(Point2d{CoordinateReferenceSystem::WGS84_2d, -180, -90});
 static const auto kSmallestPoint3d =
     PropertyValue(Point3d{CoordinateReferenceSystem::WGS84_3d, -180, -90, -std::numeric_limits<double>::infinity()});
+static const auto kLargestPropertyValue =
+    PropertyValue(Point3d{CoordinateReferenceSystem::Cartesian_3d, std::numeric_limits<double>::infinity(),
+                          std::numeric_limits<double>::infinity(), std::numeric_limits<double>::infinity()});
```
as51340: what's this?

src/storage/v2/inmemory/label_index.cpp
```diff
@@ -412,4 +430,52 @@ void InMemoryLabelIndex::CleanupAllIndices() {
...
+    while (begin != end) {
+      if (prev_v != begin->vertex) break;
+      prev_v = begin->vertex;
+      ++begin;
+    }
+    // Update
+    auto &prev = chunks_[i - 1];
+    prev = utils::SkipList<Entry>::Chunk{prev.begin(), begin};
+    chunk = utils::SkipList<Entry>::Chunk{begin, end};
+  }
```
as51340: I see this logic repeated multiple times, is there a way to templatize this?

andrejtonev: cleaned up the duplicate code

src/storage/v2/property_constants.hpp
andrejtonev: That is the largest possible property value. I need it for the property index range. 
The basic implementation only has the smallest value and then iterates until the value is larger than the upper boundary.
I need to explicitly define the upper boundary and then use it when chunking to find the last valid node.
This is needed because of the new nested property indices.
Look at kSmallestProperty for the basic version

src/utils/skip_list.hpp
andrejtonev: the loop checks the layer 0 as well. layer 0 also doesn't have the correct number of elements, so the loop does --, layer goes to -1 and we exist the loop.
I'm just setting it to the correct value. We are failing back to the bottom layer. No real need for the later=0, but just to make everything consistent.

src/utils/skip_list.hpp
andrejtonev: this is how it is implemented in every other place in the SkipList. 
For the wait to work, you would need to notify on the other end. Gut feeling is that it would slow down insertions, and be useful in a very limited number of cases.

src/utils/skip_list.hpp
andrejtonev: see comment below

src/utils/skip_list.hpp
andrejtonev: a lot of the elements used in the skiplist implement only operator < and operator ==
option would be to update them to use <=> or imeplement <=
any suggestions? not sure it makes any real difference in the end though

src/utils/skip_list.hpp
andrejtonev: yeah, you are right, the `end{};` should be 0 initialized. I'm just paranoid, since without the braces it would be un-initialized.

src/storage/v2/vertex_accessor.cpp
andrejtonev: that's not relevant for this PR

src/utils/skip_list.hpp
andrejtonev: it doesn't really matter if it is a nullptr
the check here is if end (which is preds in the find node above) is defined as head_. in that case, we must use the next node, since head_ is not a valid node, but a list of first nodes

andrejtonev: added a comment to clarify

src/utils/skip_list.hpp
as51340: I think <=> is preferred nowadays but as you wish

PR #3378, opened 2025-10-24
feat: Support loading Parquet files from the s3 compatible storage directly

as51340:
Adds AWS S3 integration for loading Parquet files from S3. AWS SDK C++ library is brought transitively through arrow library. For the authentication, we use region, access_key and secret_key. Users can also use aws_endpoint_url for pointing to some other S3 compatible storage. Settings can be set through env variables, query settings and runtime settings.

Tested using `localstack` which mocks S3 locally.

### DIFF
--- a/tests/e2e/configuration/default_config.py
+++ b/tests/e2e/configuration/default_config.py
@@ -35,6 +35,7 @@ startup_config_dict = {
     ),
     "allow_load_csv": ("true", "true", "Controls whether LOAD CSV clause is allowed in queries."),
     "aws_access_key": ("", "", "Define AWS access key for the AWS integration."),
+    "aws_endpoint_url": ("", "", "Define AWS endpoint url for the AWS integration."),
     "aws_region": ("", "", "Define AWS region which is used for the AWS integration."),
     "aws_secret_key": ("", "", "Define AWS secret key for the AWS integration."),
     "audit_buffer_flush_interval_ms": (


as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Release note is under https://github.com/memgraph/memgraph/pull/3356
- [x] **[Documentation PR link](https://github.com/memgraph/documentation/pull/1448)**

src/flags/general.hpp
```diff
@@ -127,3 +127,10 @@ auto ParseQueryModulesDirectory() -> std::vector<std::filesystem::path>;
 DECLARE_string(license_key);
 // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
 DECLARE_string(organization_name);
+
+// AWS settings
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
+DECLARE_string(aws_region);
+DECLARE_string(aws_access_key);
+DECLARE_string(aws_secret_key);
+DECLARE_string(aws_endpoint_url);
```
andrejtonev, line 136: do we want this to be part of the binary command line arguments?
seems very specific. could we use a config file instead?

src/flags/run_time_configurable.cpp
```diff
@@ -93,6 +93,18 @@ DEFINE_VALIDATED_uint64(storage_snapshot_interval_sec, 300,
...
+DEFINE_string(aws_region, "", "Define AWS region which is used for the AWS integration.");
+
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables, misc-unused-parameters)
+DEFINE_string(aws_access_key, "", "Define AWS access key for the AWS integration.");
+
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables, misc-unused-parameters)
+DEFINE_string(aws_secret_key, "", "Define AWS secret key for the AWS integration.");
+
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables, misc-unused-parameters)
+DEFINE_string(aws_endpoint_url, "", "Define AWS endpoint url for the AWS integration.");
```
andrejtonev, line 106: are there some value checks we could do to return better error messages?

src/flags/run_time_configurable.cpp
```diff
@@ -414,6 +438,11 @@ void Initialize() {
         }
         return {};
       });
+
+  register_flag(kAwsRegionGFlagsKey, kAwsRegionSettingKey, kRestore);
+  register_flag(kAwsAccessGFlagsKey, kAwsAccessSettingKey, kRestore);
+  register_flag(kAwsSecretGFlagsKey, kAwsSecretSettingKey, kRestore);
+  register_flag(kAwsEndpointUrlGFlagsKey, kAwsEndpointUrlSettingKey, kRestore);
```
andrejtonev, line 445: Why are these set to be run-time flags? Don't see any callbacks that react to changes

src/memgraph.cpp
```diff
@@ -242,6 +243,11 @@ int main(int argc, char **argv) {
   python_gc_scheduler.SetInterval(std::chrono::seconds(FLAGS_storage_python_gc_cycle_sec));
   python_gc_scheduler.Run("Python GC", [] { memgraph::query::procedure::PyCollectGarbage(); });
 
+  if (auto const status = arrow::fs::EnsureS3Initialized(); !status.ok()) {
+    spdlog::error(status.message());
+    exit(1);
+  }
```
andrejtonev: Do we always want to run arrow?

src/query/plan/operator.cpp
```diff
@@ -7740,9 +7768,21 @@ class LoadParquetCursor : public Cursor {
...
       auto maybe_file = self_->file_->Accept(evaluator).ValueString();
 
+      auto maybe_config_map = ParseConfigMap(self_->config_map_, evaluator);
+
+      if (!maybe_config_map) {
+        throw QueryRuntimeException("Failed to parse config map for LOAD PARQUET clause!");
+      }
+
+      if (maybe_config_map->size() > 4) {
+        throw QueryRuntimeException("Config map cannot contain > 4 entries: {}, {}, {} and {} can be provided",
```
andrejtonev: `Config map cannot contain > 4 entries. Only {}, {}, {} and {} can be provided`

src/flags/run_time_configurable.cpp
as51340: I will check if Arrow has something for validation

src/flags/run_time_configurable.cpp
as51340: Here we don't need callbacks

src/memgraph.cpp
as51340: I will try to move this into reader.cpp so it gets initialized/finalized only if used

src/flags/run_time_configurable.cpp
as51340: Nothing as I see it. Arrow brings in AWS SDK lib so I think it's not smart for us to validate something if AWS's SDK doesn't validate. New AWS regions AFAIK can get added so I don't want to risk not support something https://docs.aws.amazon.com/global-infrastructure/latest/regions/aws-regions.html. IMO AWS SDK should care about it not us

src/memgraph.cpp
as51340: used singleton

PR #3391, opened 2025-10-30
feat: Use shared zlib, libpulsar.a and Conan's Snappy

as51340:
Use shared zlib. Use libpulsar.a instead of libpulsarwithdeps and bring Snappy from Conan

### DIFF
--- a/libs/CMakeLists.txt
+++ b/libs/CMakeLists.txt
@@ -289,7 +289,8 @@ macro(build_protobuf)
       BUILD_IN_SOURCE 1
       CONFIGURE_COMMAND true)
 endmacro()
-mg_build_if_needed(build_protobuf "protobuf")
+#mg_build_if_needed(build_protobuf "protobuf")
+mg_create_linkable_if_no_cmake(protobuf STATIC "${MG_TOOLCHAIN_ROOT}/lib/libprotobuf.a" "${MG_TOOLCHAIN_ROOT}/include")
 
 macro(build_pulsar)
     import_external_library(pulsar STATIC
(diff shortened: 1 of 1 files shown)


libs/CMakeLists.txt
```diff
@@ -311,11 +311,16 @@ macro(build_pulsar)
       -DBUILD_PYTHON_WRAPPER=OFF
       -DBUILD_PERF_TOOLS=OFF
       -DUSE_LOG4CXX=OFF
-      BUILD_COMMAND $(MAKE) pulsarStaticWithDeps)
+      BUILD_COMMAND $(MAKE))
     add_dependencies(pulsar-proj protobuf)
 endmacro()
-mg_build_if_needed(build_pulsar "pulsar")
-mg_create_linkable_if_no_cmake(pulsar STATIC "${MG_TOOLCHAIN_ROOT}/lib/libpulsarwithdeps.a" "${MG_TOOLCHAIN_ROOT}/include")
+find_library(protobuf
```
mattkjames7: We didn't need this before (I'm guessing the withdeps version of pulsar included it), but if we're going to have to link with it now, might it be worth using the one provided by [conan](https://conan.io/center/recipes/protobuf)?

mattkjames7: Potential issue with that is that the toolchain uses version `v3.12.4` which is unmaintained, the earliest version that conan has is, which is maintained, is `v3.20.3`

mattkjames7: Resolve, conan version has incompatible ABI

as51340: yep, 3.12.4 version is unmaintained on Conan and I am getting undefined symbols. Once we upgrade the Pulsar, we will be able to use Conan's protobuf

PR #3392, opened 2025-10-30
Edge property index preserve edgetype filter

Ignition:
When using global edge property indexes `CREATE GLOBAL EDGE INDEX ON :(p)`,
the query planner would incorrectly discard edge type filters from queries,
causing incorrect results.

For example, this query would return edges of all types with p=1, not just
type `:A`:
```
  CREATE GLOBAL EDGE INDEX ON :(p);
  CREATE ()-[:E{p:1}]->();
  CREATE ()-[:A{p:1}]->();
  MATCH ()-[e:A{p:1}]->() RETURN e;
```

The plan would be:
  * ScanAllByEdgePropertyValue (anon1)-[e {p}]->(anon2)

But should be:
  * Filter [e :A]
  * ScanAllByEdgePropertyValue (anon1)-[e {p}]->(anon2)

This fix:
- Introduces EdgeTypesTest AST expression (parallel to LabelsTest)
- Modifies edge_index_lookup rewriter to preserve edge type constraints
- Wraps index scans with Filter operator when edge types are specified
- Ensures correct query results when combining global edge indexes with
  edge type filters

Fixes edge property index scans to respect edge type constraints in:
- ScanAllByEdgeProperty
- ScanAllByEdgePropertyValue
- ScanAllByEdgePropertyRange

### DIFF
--- a/src/query/plan/vertex_count_cache.hpp
+++ b/src/query/plan/vertex_count_cache.hpp
@@ -33,9 +33,9 @@ class VertexCountCache {
   auto NameToLabel(const std::string &name) { return db_->NameToLabel(name); }
   auto NameToProperty(const std::string &name) { return db_->NameToProperty(name); }
   auto NameToEdgeType(const std::string &name) { return db_->NameToEdgeType(name); }
-  auto LabelToName(storage::LabelId name) { return db_->LabelToName(name); }
-  auto PropertyToName(storage::PropertyId name) { return db_->PropertyToName(name); }
-  auto EdgeTypeToName(storage::EdgeTypeId name) { return db_->EdgeTypeToName(name); }
+  auto LabelToName(storage::LabelId id) { return db_->LabelToName(id); }
+  auto PropertyToName(storage::PropertyId id) { return db_->PropertyToName(id); }
+  auto EdgeTypeToName(storage::EdgeTypeId id) { return db_->EdgeTypeToName(id); }
 
   auto GetEnumValue(std::string_view name, std::string_view value)
       -> utils::BasicResult<storage::EnumStorageError, storage::Enum> {
--- a/tests/manual/interactive_planning.cpp
+++ b/tests/manual/interactive_planning.cpp
@@ -145,6 +145,9 @@ class InteractiveDbAccessor {
   auto NameToLabel(const std::string &name) { return dba_->NameToLabel(name); }
   auto NameToProperty(const std::string &name) { return dba_->NameToProperty(name); }
   auto NameToEdgeType(const std::string &name) { return dba_->NameToEdgeType(name); }
+  auto LabelToName(memgraph::storage::LabelId id) { return dba_->LabelToName(id); }
+  auto PropertyToName(memgraph::storage::PropertyId id) { return dba_->PropertyToName(id); }
+  auto EdgeTypeToName(memgraph::storage::EdgeTypeId id) { return dba_->EdgeTypeToName(id); }
 
   auto GetStorageAccessor() { return dba_->GetStorageAccessor(); }
 


Ignition:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Bugfix, queries that use a global edge index no longer ignore any edge type filters [#3392](https://github.com/memgraph/memgraph/pull/3392)

src/query/frontend/ast/ast.hpp
```diff
@@ -1151,7 +1150,43 @@ class LabelsTest : public memgraph::query::Expression {
...
+    }
+    return visitor.PostVisit(*this);
+  }
+
+  Expression *expression_{nullptr};  // expression used to get the Edge
+  std::vector<EdgeTypeIx> valid_edgetypes_;
+
+  EdgeTypesTest *Clone(AstStorage *storage) const override {
+    EdgeTypesTest *object = storage->Create<EdgeTypesTest>();
+    object->expression_ = expression_->Clone(storage);
```
andrejtonev: expression_ ? expression_->Clone(storage) : nullptr;

src/query/interpret/eval.hpp
```diff
@@ -582,6 +581,22 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
     }
   }
 
+  TypedValue Visit(EdgeTypesTest &edgetype_test) override {
+    switch (auto expression_result = edgetype_test.expression_->Accept(*this); expression_result.type()) {
```
andrejtonev, line 589: Add a DMG_ASSERT that expression_ is valid

src/query/frontend/ast/ast.hpp
```diff
@@ -1151,7 +1150,43 @@ class LabelsTest : public memgraph::query::Expression {
...
+  DEFVISITABLE(ExpressionVisitor<void>);
+  bool Accept(HierarchicalTreeVisitor &visitor) override {
+    if (visitor.PreVisit(*this)) {
+      expression_->Accept(visitor);
+    }
+    return visitor.PostVisit(*this);
+  }
+
+  Expression *expression_{nullptr};  // expression used to get the Edge
+  std::vector<EdgeTypeIx> valid_edgetypes_;
```
andrejtonev, line 1172: could this be an unordered_set? that way we ensure no duplicates and don't have to do a linear search

src/query/interpret/eval.hpp
```diff
@@ -582,6 +581,22 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
...
+      case TypedValue::Type::Null:
+        return TypedValue(ctx_->memory);
+      case TypedValue::Type::Edge: {
+        const auto &edge_type = expression_result.ValueEdge().EdgeType();
+        auto const ix_to_storage = [&](EdgeTypeIx const &et) { return GetEdgeType(et); };
+        auto valid_edgetypes = edgetype_test.valid_edgetypes_ | std::ranges::views::transform(ix_to_storage);
+        const auto is_valid = std::ranges::contains(valid_edgetypes, edge_type);
```
andrejtonev: could be combined into a single `std::any_of`

src/query/plan/preprocess.cpp
```diff
@@ -811,7 +811,22 @@ void Filters::AnalyzeAndStoreFilter(Expression *expr, const SymbolTable &symbol_
...
+  if (auto *edgetype_test = utils::Downcast<EdgeTypesTest>(expr)) {
+    if (auto *identifier = utils::Downcast<Identifier>(edgetype_test->expression_)) {
+      if (auto it = std::ranges::find_if(all_filters_, MatchesIdentifier(identifier)); it == all_filters_.end()) {
+        // No existing EdgeTypesTest for this identifier
+        auto filter = make_filter(FilterInfo::Type::EdgeType);
+        filter.edgetypes = edgetype_test->valid_edgetypes_;
+        all_filters_.emplace_back(filter);
+      } else {
+        // We could intersect the two OR lists, but for now we don't do that processing and keep it as  separate filter
+        all_filters_.emplace_back(make_filter(FilterInfo::Type::Generic));
```
andrejtonev, line 824: Not sure how the preprocessor works, but shouldn't we modify/swap the existing filter with the generic one that has both type checks?

tests/unit/query_common.hpp
```diff
@@ -210,13 +210,24 @@ auto GetAllPropertiesLookup(AstStorage &storage, Expression *expr) { return stor
...
   }
-  return storage.Create<EdgeAtom>(storage.Create<Identifier>(name, user_declared), EdgeAtom::Type::SINGLE, dir, types);
+  auto *edge =
+      storage.Create<EdgeAtom>(storage.Create<Identifier>(name, user_declared), EdgeAtom::Type::SINGLE, dir, types);
+  if (properties) {
+    if (auto *map_literal = dynamic_cast<MapLiteral *>(properties)) {
+      edge->properties_ = map_literal->elements_;
+    } else {
+      // Assume it's a ParameterLookup
+      edge->properties_ = dynamic_cast<ParameterLookup *>(properties);
```
andrejtonev, line 228: Add a DMG_ASSERT

src/query/frontend/ast/ast.hpp
Ignition: Expected number of edge types is 0 (hence no filter) or 1. Sure you can specify more, but for now a vector is appropriate, until any typical workload would justify using a set.

src/query/plan/preprocess.cpp
Ignition: It is fine. I can't think of a situation where the same symbol has two edge-type filters. There is only one place in the grammar you can express the edge-type restriction. The Generic is there as a just in case, I would rather it was captured just in case I was wrong. If it was possible to have two or more edge type filers on the same symbol then I think a set union is how they would be merged. For now, this is fine.

PR #3400, opened 2025-11-05
fix: Take into account replica's time needed to make changes durable

as51340:
Replica's time to make deltas durable wasn't taken into account into the RPC timeout

### DIFF
--- a/src/storage/v2/inmemory/storage.cpp
+++ b/src/storage/v2/inmemory/storage.cpp
@@ -2508,7 +2508,7 @@ bool InMemoryStorage::InMemoryAccessor::HandleDurabilityAndReplicate(uint64_t du
   needs_wal_update_ = two_phase_commit;
 
   auto timer = utils::Timer{};
-  constexpr auto delta_cb_timeout = std::chrono::seconds{5};
+  constexpr auto delta_cb_timeout = std::chrono::seconds{10};
 
   // IMPORTANT: In most transactions there can only be one, either data or metadata deltas.
   //            But since we introduced auto index creation, a data transaction can also introduce a metadata delta.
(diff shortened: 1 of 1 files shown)


as51340:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Replica will now send InProgressRpc message when it takes more time to make deltas durable. Previously, it could happen that deltas were processed fast but because of writing to disk, the main thought that replica was down. This caused transactions to be aborted in STRICT SYNC mode while in SYNC mode the txn would be committed but with the exception being thrown confusing the user what actually happened in the system. [#3400](https://github.com/memgraph/memgraph/pull/3400)

src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -1189,10 +1188,21 @@ std::optional<storage::SingleTxnDeltasProcessingResult> InMemoryReplicationHandl
...
-          if (!commit_accessor || commit_timestamp != delta_timestamp)
+          if (!commit_accessor || commit_timestamp != delta_timestamp) {
             throw utils::BasicException("Invalid commit data!");
-          auto const ret = commit_accessor->PrepareForCommitPhase(
-              storage::CommitArgs::make_replica_write(commit_timestamp, two_phase_commit));
+          }
+
+          // We need to check whether the timeout passed as well when replica is making deltas durable
+          auto delta_cnt_checker = [&]() {
+            if (++current_batch_counter == kDeltasBatchProgressSize) {
```
andrejtonev: this works if each delta takes more or less the same time to process. Is that actually the case?
since there are disk operations, I would assume that we can never guarantee that
so what happens when one delta take too long?
I would rather have a watch dog then this count logic.
fail, if the number of deltas processed between two timed checks didn't change

PR #3424, opened 2025-11-12
Extend telemetry

colinbarry:
Extend telemetry to include:
- number of roles alongside number of users
- bucketed log_10 histogram of nodes per label. Labels used for proper semantic modelling should give a flat distribution, whilst those used to simulate multi-tenancy will have a narrow spike in one bucket. For in-memory, this caches the label distribution as changes are mode. For on-disk, it relies on a brute-force scan.

### DIFF
--- a/src/storage/v2/disk/storage.cpp
+++ b/src/storage/v2/disk/storage.cpp
@@ -940,6 +940,7 @@ std::unordered_map<LabelId, uint64_t> DiskStorage::GetLabelCounts() const {
   if (!config_.track_label_counts) {
     return label_counts;
   }
+  // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
   auto storage_acc = const_cast<DiskStorage *>(this)->ReadOnlyAccess();
   for (auto &&vertex : storage_acc->Vertices(View::OLD)) {
     auto const labels_result = vertex.Labels(View::OLD);


colinbarry:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
   Update telemetry to include anonymous data for role counts and label/node distributions. [#3424](https://github.com/memgraph/memgraph/pull/3424)
- [x] Update documentation: https://github.com/memgraph/documentation/pull/1492

src/dbms/dbms_handler.hpp
```diff
@@ -353,6 +363,14 @@ class DbmsHandler {
         using underlying_type = std::underlying_type_t<utils::CompressionLevel>;
         ++stats.property_store_compression_level[static_cast<underlying_type>(
             storage_info.property_store_compression_level)];
+
+        auto const label_counts = db_acc->storage()->GetLabelCounts();
```
andrejtonev, line 374: I think schema info should be able to give this information
Did you add a separate logic on purpose?

gitbuda:
@colinbarry added the docs required label because the info there has to be changed (equivalent to e.g. https://github.com/memgraph/documentation/pull/1414)

src/dbms/dbms_handler.hpp
colinbarry: Yes. Schema info is optional (`--schema-info-enabled`). As we can't rely on that being set when `--telemetry-enabled=true`, we need a separate mechanism that is always available.

andrejtonev: If the penalty for this added logic is low, then I'm fine with it.
I'll leave it to your discretion

PR #3437, opened 2025-11-14
Conan Arrow

mattkjames7:
Use Conan package for Arrow
- Removes `ExternalProject_Add` build for arrow, replacing with static lib cached by `conan`
- Updated default `conan` profile in CI to force `gnu17` C standard so that `m4` can be successfully compiled
- Added `conan.lock` to fix the packages  to be built in order to avoid build failures due to package updates from conancenter: #3450 

----------------------
## Notes on the `m4` build failure for Fedora 42 on `aarch64`:

TLDR: Fix by adding

```
[conf]
m4/*:tools.build:cflags=["-std=gnu17"]
```
to the default profile in `~/.conan2/profile/default` so that the compiler doesn't try to use a newer standard which `m4` cannot yet be built with.

`aws-sdk-cpp` has a transitive dependency which requires `m4` to build it; `m4` is pulled in by `conan`. On `x86_64` it is downloaded as a binary, but on `armv8` it has to be built from source. This means that the problem does not affect the `x86_64`  build of Fedora 42, just the `armv8` one.

`m4` seems to be built using `gcc`. It's not obvious to me whether this is using the system `gcc` or the toolchain one. If it's using the system one, then this could explain why it will not build on Fedora 42, as that is the only one with gcc-15, which defaults to `gnu23`:

```bash
gcc -dM -E -x c /dev/null | grep __STDC_VERSION__
g++ -dM -E -x c++ /dev/null | grep __cplusplus
```

```cpp
#define __STDC_VERSION__ 202311L
#define __cplusplus 201703L
```

In other distros:
```cpp
#define __STDC_VERSION__ 201710L
#define __cplusplus 201703L
```

Reproduce in `mgbuild` container on `arm`:

```bash
docker run --rm -d --name mgbuild memgraph/mgbuild:v7_fedora-42-arm
docker exec -it -u mg mgbuild bash
cd /home/mg
git clone https://github.com/memgraph/memgraph
cd memgraph
python3 -m venv env
source env/bin/activate
pip install conan
conan profile detect
export BUILD_TYPE=Release

# with the toolchain 
MG_TOOLCHAIN_ROOT=/opt/toolchain-v7 conan install --requires m4/1.4.19 --build=* -pr ./memgraph_template_profile

# without the toolchain
conan install --requires m4/1.4.19 --build=*
```

Or on `x86_64`

```bash
docker run --rm -d --name mgbuild memgraph/mgbuild:v7_fedora-42
docker exec -it -u mg mgbuild bash
cd /home/mg
git clone https://github.com/memgraph/memgraph
cd memgraph
python3 -m venv env
source env/bin/activate
pip install conan
conan profile detect
export BUILD_TYPE=Release

# with the toolchain - this builds successfully on x86_84!
MG_TOOLCHAIN_ROOT=/opt/toolchain-v7 conan install --requires m4/1.4.19 --build=* -pr ./memgraph_template_profile

# without the toolchain
conan install --requires m4/1.4.19 --build=*
```

The fix is:

```bash
conan install --requires m4/1.4.19 --build=* -c tools.build:compiler_executables='{"cxx":"g++ -std=c++23","c":"gcc -std=gnu17"}'
```

or with the toolchain:

```bash
MG_TOOLCHAIN_ROOT=/opt/toolchain-v7  conan install --requires m4/1.4.19 --build=*  -pr ./memgraph_template_profile -c tools.build:compiler_executables='{"cxx":"g++ -std=c++23","c":"gcc -std=gnu17"}'
```

Adding this to the template should probably work, but doesn't :joy: - unless, we build `m4` into the cache on it's own beforehand! (see the change to `mgbuild.sh`)
```
m4/*:tools.build:compiler_executables={"c":"gcc -std=gnu17","cxx":"g++ -std=c++23"}
```


mattkjames7:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - N/A

conanfile.py
```diff
@@ -72,7 +72,8 @@ def requirements(self):
         self.requires("range-v3/0.12.0")
         self.requires("asio/1.34.2")
         self.requires("mgclient/1.4.3", options={"with_cpp": True})
-        self.requires("snappy/1.2.1")
+        self.requires("snappy/1.1.9")
```
Ignition: https://docs.conan.io/2/reference/conanfile/methods/requirements.html#override

memgraph_template_profile
```diff
@@ -36,3 +37,19 @@ boost/*:extra_b2_flags=cxxflags=--target=aarch64-unknown-linux-gnu cflags=--targ
...
+aws-sdk-cpp/*:s3=True
+aws-sdk-cpp/*:monitoring=False
+aws-sdk-cpp/*:text-to-speech=False
+aws-sdk-cpp/*:queues=False
+aws-sdk-cpp/*:sqs=False
+aws-sdk-cpp/*:access-management=False
+aws-sdk-cpp/*:cognito-identity=True
+aws-sdk-cpp/*:iam=False
+aws-sdk-cpp/*:identity-management=True
+aws-sdk-cpp/*:transfer=False
```
Ignition: These should be added into conanfile.py default_options

https://docs.conan.io/2/reference/conanfile/attributes.html#default-options

release/package/mgbuild.sh
```diff
@@ -521,6 +521,11 @@ build_memgraph () {
   # Check if a conan profile exists and create one if needed
   docker exec -u mg "$build_container" bash -c "$CMD_START && if [ ! -f \"\$HOME/.conan2/profiles/default\" ]; then conan profile detect; fi"
 
+  # hack for fedora on arm
+  if [[ "$os" == "fedora-42-arm" ]]; then
+    docker exec -u mg "$build_container" bash -c "$CMD_START && export BUILD_TYPE=$build_type && MG_TOOLCHAIN_ROOT=/opt/toolchain-v7 conan install --requires m4/1.4.19 --build=* -pr ./memgraph_template_profile"
```
Ignition: Why not unconditionally just have `self.tool_requires("m4/1.4.19")`?

Ignition (approved):

PR #3421, opened 2025-11-12
Improve skip list GC performance by optimizing the deleted objects stack

DavIvek:
This PR improves performance of skip list GC by optimizing the deleted objects stack.

PR #3441, opened 2025-11-14
Avoid expensive GC during edge accessor destruction while vertex locks are held

DavIvek:
This PR addresses a performance bottleneck caused by skip list garbage collection (GC) occurring during edge accessor destruction while vertex locks are still held.
The skip list GC is not only expensive but also blocks other threads. This becomes critical under high concurrency, especially when queries create edges involving supernodes.

Here’s the scenario:
- Multiple threads call CreateEdge concurrently.
- One thread acquires unique locks on the vertices involved in the edge.
- Inside the locked section, an edge accessor is created.
- When the accessor goes out of scope, it may trigger skip list GC.
- If GC runs while vertex locks are still held, all threads attempting related operations are blocked.

This leads to significant performance degradation when operating on supernodes.

### DIFF
--- a/src/storage/v2/inmemory/storage.cpp
+++ b/src/storage/v2/inmemory/storage.cpp
@@ -40,7 +40,6 @@
 #include "storage/v2/replication/replication_transaction.hpp"
 #include "storage/v2/schema_info_glue.hpp"
 #include "utils/async_timer.hpp"
-#include "utils/skip_list.hpp"
 #include "utils/timer.hpp"
 
 /// REPLICATION ///


DavIvek:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Edge creation is refactored to ensure skip list GC doesn’t run while vertex locks are held. This results in better throughput and fewer stalls under heavy concurrency. No user action is required to benefit from the improved performance. [#3421](https://github.com/memgraph/memgraph/pull/3421)

PR #3448, opened 2025-11-19
Force shared lock for GC in analytical mode and add runtime flag for GC aggressiveness

DavIvek:
The `FREE MEMORY` query was forcing a unique lock on `main_lock_` in analytical mode, which would block other transactions. This was frustrating and counterintuitive for users expecting high performance with `IN_MEMORY_ANALYTICAL`.
Also, a runtime flag `storage-gc-aggressive` has been added to check whether the GC should be aggressive.

### DIFF
--- a/src/storage/v2/inmemory/storage.cpp
+++ b/src/storage/v2/inmemory/storage.cpp
@@ -421,7 +421,7 @@ InMemoryStorage::InMemoryAccessor::DetachDelete(std::vector<VertexAccessor *> no
 
   auto &[deleted_vertices, deleted_edges] = *value;
 
-  // Need to inform the next CollectGarbageCollectGarbage call that there are some
+  // Need to inform the next CollectGarbage call that there are some
   // non-transactional deletions that need to be collected
 
   auto const inform_gc_vertex_deletion = utils::OnScopeExit{[this, &deleted_vertices = deleted_vertices]() {


DavIvek:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Forced the GC to acquire a shared lock in analytical mode, preventing unnecessary blocking of other transactions and improving responsiveness. A new runtime flag, `storage-gc-aggressive`, allows users to opt into aggressive GC, which performs full cleanup of unused objects but requires taking a unique lock that will temporarily block the system. [#3448](https://github.com/memgraph/memgraph/pull/3448)

src/storage/v2/storage.hpp
```diff
@@ -700,13 +700,7 @@ class Storage {
...
   virtual void FreeMemory(std::unique_lock<utils::ResourceLock> main_guard, bool periodic) = 0;
 
-  void FreeMemory() {
-    if (storage_mode_ == StorageMode::IN_MEMORY_ANALYTICAL) {
-      FreeMemory(std::unique_lock{main_lock_}, false);
-    } else {
-      FreeMemory({}, false);
-    }
-  }
+  void FreeMemory() { FreeMemory({}, false); }
```
andrejtonev: try_lock will throw an exception
`If there is no associated mutex, [std::system_error](https://en.cppreference.com/w/cpp/error/system_error.html) with an error code of [std::errc::operation_not_permitted](https://en.cppreference.com/w/cpp/error/errc.html).`

andrejtonev: Maybe it makes sense to support aggressive GC for the user query as well?

DavIvek: I was thinking about that as well. Back in version 2.14, the GC would first try to take main_lock_ even for user queries, but that behavior was removed because it could actually hurt performance if the GC ran for a long time. However, now that we can control this with a runtime flag, we should probably bring that behavior back when the flag is enabled.

andrejtonev (approved):

PR #3459, opened 2025-11-21
feat: Modularize utils/counter

as51340:


as51340:
Modules, it turns out, slows down our codebase

PR #3502, opened 2025-12-02
Dynamically link to OpenSSL

mattkjames7:
Force memgraph and its dependencies to dynamically link to OpenSSL (currently, some parts are statically linking). The version set in `conanfile.py` should be considered to be a minimum version that the `memgraph` binary can link to (any newer minor version of OpenSSL 3.x should be ABI/API compatible with 3.0: [see here](https://openssl-library.org/policies/releasestrat/index.html)). This means that any recent-ish OpenSSL library packaged by Linux distros should work, but also allowing newer versions to be used if needed.

When packaging for Docker, Ubuntu provides an older version (3.0.14) of OpenSSL than what we want to package (3.5.4). To address this, `mgbuild.sh` now has the `build-ssl` argument which uses `conan` to build the binaries for the version that we want to package. We then package them binaries as `.deb` packages, in the same format as those provided by Ubuntu -> these packages are then installed in the Docker image build by `apt`.

The reason for building and installing using `apt` is simply because `apt` will not overwrite a newer version of OpenSSL with an older one found in Ubuntu's repo, and anything else within the container that depends on it should use our custom version. A couple of alternatives to this would be to either statically link to OpenSSL, or to place the custom libs directly within `/usr/lib/memgraph` (within `$RUNPATH`) - the downside of this would be that the Ubuntu-provided package will still be needed for other `apt` and Python packages that depend on OpenSSL -> so we would end up with two versions of OpenSSL and a (slightly) larger container image.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -81,13 +81,16 @@ class Memgraph(ConanFile):
         self.requires("fmt/11.2.0")
         self.requires("libcurl/8.17.0", override=True)
         self.requires("mgclient/1.4.3", options={"with_cpp": True})
-        self.requires("openssl/3.0.18", override=True, options={"shared": True})
         self.requires("range-v3/0.12.0")
         self.requires("simdjson/4.2.2")
         self.requires("snappy/1.2.1", override=True)
         self.requires("spdlog/1.15.3")
         self.requires("strong_type/v15")
         self.requires("zlib/1.3.1")
+        # We force Memgraph and all its dependencies to dynamically link to OpenSSL. We use version
+        # 3.0 here so that the binary can dynamically link to any version of OpenSSL >=3 and < 4,
+        # therefore allowing it to work with any package provided by our supported Linux distributions.
+        self.requires("openssl/3.0.18", override=True, options={"shared": True})
 
     def build_requirements(self):
         self.tool_requires("cmake/4.1.2")
--- a/tools/openssl/build-libssl3-deb.sh
+++ b/tools/openssl/build-libssl3-deb.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 set -euo pipefail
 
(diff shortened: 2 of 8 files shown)


mattkjames7:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - N/A

tools/openssl/build-libssl3-deb.sh
```diff
@@ -0,0 +1,42 @@
...
+Package: libssl3t64
+Version: 3.5.4-0ubuntu0custom1
+Section: libs
+Priority: required
+Architecture: $ARCH
+Maintainer: Matt James <matthew.james@memgraph.io>
+Conflicts: libssl3t64
+Replaces: libssl3t64
+Provides: libssl3
+Description: Custom libssl/libcrypto from OpenSSL 3.5.4 (Conan build)
```
gitbuda: centralize the OpenSSL version somewhere

conanfile.py
```diff
@@ -87,6 +87,10 @@ def requirements(self):
         self.requires("spdlog/1.15.3")
         self.requires("strong_type/v15")
         self.requires("zlib/1.3.1")
+        # We force Memgraph and all its dependencies to dynamically link to OpenSSL. We use version
+        # 3.0 here so that the binary can dynamically link to any version of OpenSSL >=3 and < 4,
+        # therefore allowing it to work with any package provided by our supported Linux distributions.
+        self.requires("openssl/3.0.18", override=True, options={"shared": True})
```
andrejtonev, line 93: why 3.0.18 specifically?

mattkjames7: This is the only version of 3.0 that they currently have on conancenter.

PR #3503, opened 2025-12-02
feat: allow triggers to use permissions of trigger definer or trigger invoker

imilinovic:
Allow triggers to use permissions of security definer or security invoker.

Previously triggers would use the permissions of the security definer (user who created the trigger). Now by default it will be the security definer (user who wrote the trigger).


imilinovic:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Triggers now support a SECURITY clause to control which user's permissions are used during execution. [#3503](https://github.com/memgraph/memgraph/pull/3503)
- [x] **[ Documentation PR link memgraph/documentation#1502 ]**

tests/e2e/triggers/privilege_check.cpp
```diff
@@ -0,0 +1,265 @@
+// Copyright 2025 Memgraph Ltd.
+//
```
imilinovic, line 2: I've switched this to GTest in later PRs due to easier readability from both logs & code

src/query/interpreter.cpp
```diff
@@ -4518,39 +4554,46 @@ Callback CreateTrigger(TriggerQuery *trigger_query, const storage::ExternalPrope
...
+      .fn = [trigger_name = std::move(trigger_query->trigger_name_),
+             trigger_statement = std::move(trigger_query->statement_), event_type = trigger_query->event_type_,
+             before_commit = trigger_query->before_commit_, trigger_store, interpreter_context, dba, user_parameters,
+             owner = std::move(owner), db_name,
+             privilege_context = trigger_query->privilege_context_]() mutable -> std::vector<std::vector<TypedValue>> {
+        trigger_store->AddTrigger(std::move(trigger_name), trigger_statement, user_parameters,
+                                  ToTriggerEventType(event_type),
+                                  before_commit ? TriggerPhase::BEFORE_COMMIT : TriggerPhase::AFTER_COMMIT,
+                                  &interpreter_context->ast_cache, dba, interpreter_context->config.query,
+                                  std::move(owner), db_name, static_cast<TriggerPrivilegeContext>(privilege_context));
```
andrejtonev: can we remove the static cast?

src/query/interpreter.cpp
```diff
@@ -8082,9 +8164,9 @@ void Interpreter::Commit() {
   // ordered execution of after commit triggers are not guaranteed.
   if (trigger_context && db->trigger_store()->AfterCommitTriggers().size() > 0) {
     db->AddTask([db_acc = *current_db_.db_acc_, interpreter_context = interpreter_context_,
-                 trigger_context = std::move(*trigger_context),
-                 user_transaction = std::move(current_db_.db_transactional_accessor_)]() {
-      RunTriggersAfterCommit(db_acc, interpreter_context, std::move(trigger_context));
+                 trigger_context = std::move(*trigger_context), triggering_user = user_or_role_,
+                 user_transaction = std::shared_ptr(std::move(current_db_.db_transactional_accessor_))]() {
```
andrejtonev: why the std::shared_ptr?

src/query/trigger.cpp
```diff
@@ -177,30 +184,42 @@ std::shared_ptr<Trigger::TriggerPlan> Trigger::GetPlan(DbAccessor *db_accessor,
...
 
     auto logical_plan = MakeLogicalPlan(std::move(ast_storage), utils::Downcast<CypherQuery>(parsed_statements_.query),
                                         parsed_statements_.parameters, db_accessor, predefined_identifiers);
 
     trigger_plan_ = std::make_shared<TriggerPlan>(std::move(logical_plan), std::move(identifiers));
   }
-  if (owner_ && !owner_->IsAuthorized(parsed_statements_.required_privileges, db_name, &up_to_date_policy)) {
-    throw utils::BasicException("The owner of trigger '{}' is not authorized to execute the query!", name_);
+
+  // GenQueryUser always returns a non-null shared_ptr and
```
andrejtonev, line 196: lets add a DMG_ASSERT to verify the creator and user are valid

src/query/trigger.cpp
```diff
@@ -221,6 +240,7 @@ void Trigger::Execute(DbAccessor *dba, dbms::DatabaseAccess db_acc, utils::Memor
   ctx.evaluation_context.memory = execution_memory;
   ctx.protector = dbms::DatabaseProtector{db_acc}.clone();
   ctx.is_main = is_main;
+  ctx.user_or_role = triggering_user;
```
andrejtonev: 1. Should we use creator or user based on the security requirements?
2. Should we reuse the whole execution context in case of before commit triggers?

src/query/trigger.cpp
```diff
@@ -264,8 +286,10 @@ void TriggerStore::RestoreTrigger(utils::SkipList<QueryCacheEntry> *query_cache,
     spdlog::warn(invalid_state_message);
     return;
   }
-  if (json_trigger_data["version"] != kVersion) {
-    spdlog::warn(get_failed_message("Invalid version of the trigger data."));
+  const auto version = json_trigger_data["version"].get<uint64_t>();
```
andrejtonev, line 323: Create a migration function that is called at the start and always updates the durable data to follow the latest format. That way you don't need to care about the version later on in the code, you just support the latest one.

src/query/trigger.cpp
```diff
@@ -242,7 +262,9 @@ void Trigger::Execute(DbAccessor *dba, dbms::DatabaseAccess db_acc, utils::Memor
 
 namespace {
 // When the format of the persisted trigger is changed, increase this version
-inline constexpr uint64_t kVersion{2};
+// also update tests/integration/triggers
+inline constexpr uint64_t kDefinerVersion{2};
```
andrejtonev: kDefinerOnlyVersion
kPreSecurityVersion
or something else that better explains what's going on

src/query/trigger.cpp
```diff
@@ -322,10 +346,27 @@ void TriggerStore::RestoreTrigger(utils::SkipList<QueryCacheEntry> *query_cache,
...
+  auto const privilege_context = std::invoke([&]() -> std::optional<TriggerPrivilegeContext> {
+    if (version == kDefinerVersion) {
+      return TriggerPrivilegeContext::DEFINER;
+    }
+
+    const auto privilege_context_json = json_trigger_data["privilege_context"];
+    if (!privilege_context_json.is_number_unsigned()) {
+      return std::nullopt;
+    }
+    return static_cast<TriggerPrivilegeContext>(privilege_context_json.get<uint8_t>());
```
andrejtonev: Unsafe. Check if within range or create a serializer for the enum and directly use it

andrejtonev: Also think about storing strings instead of ints, that would be independent of the enum values and any future changes.

src/query/trigger.hpp
```diff
@@ -33,15 +33,23 @@ namespace memgraph::query {
 struct QueryCacheEntry;
 
 enum class TransactionStatus;
+
 struct Trigger {
+  enum class PrivilegeContext : uint8_t {
+    INVOKER = 0,  // trigger is executed with the permissions of the user who invoked the trigger
+    DEFINER = 1   // trigger is executed with the permissions of the user who defined the trigger
+  };
```
andrejtonev: This same thing is defined in 3 places. Can we converge on a single specification?

src/query/interpreter.cpp
imilinovic: because i pressed tab and didn't look :(
reverted

src/query/trigger.cpp
imilinovic: 1. Currently doesn't matter because its only used for triggers in case of INVOKER but YES! Fixed
2. Not sure, currently I think no because it has stuff like trigger_context_collector which you don't want to happen during triggers (recursive triggering would happen I think).

PR #3509, opened 2025-12-04
Fix various concurrency issues

colinbarry:
Fixes various TSAN-identified concurrency issues:

- Added double-checked locking to `DiskNameIdMapper`'s `NameToId` and `IdToName`. Prevents races where concurrent threads could generate duplicate IDs.
- Ensures GC `commit_log_` always exists before the GC thread starts.
- Ensure all our `Synchronized` `Lock` methods return decayed values, not references. This fixes several issues where we were taking copies out synchronized objects without holding the lock.
- Suppress `WalDeathResilience` tests when running with TSAN, as TSAN does not support `fork()`ing with active threads.

### DIFF
--- a/src/storage/v2/disk/name_id_mapper.hpp
+++ b/src/storage/v2/disk/name_id_mapper.hpp
@@ -72,6 +72,12 @@ class DiskNameIdMapper final : public NameIdMapper {
       return maybe_name.value();
     }
 
+    std::lock_guard<std::mutex> const lock{name_id_maps_mutex_};
+    maybe_name = NameIdMapper::MaybeIdToName(id);
+    if (maybe_name) {
+      return maybe_name.value();
+    }
+
     auto maybe_name_from_disk = id_to_name_storage_->Get(std::to_string(id));
     MG_ASSERT(maybe_name_from_disk.has_value(), "Trying to get a name from disk for an invalid ID!");
 


colinbarry:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Fix various unlikely crashes related to concurrency and thread-safety. [#3509](https://github.com/memgraph/memgraph/pull/3509)

src/storage/v2/disk/name_id_mapper.hpp
```diff
@@ -42,6 +43,12 @@ class DiskNameIdMapper final : public NameIdMapper {
     if (auto maybe_id = MaybeNameToId(name); maybe_id.has_value()) {
       return maybe_id.value();
     }
+
+    std::lock_guard<std::mutex> const lock{name_id_maps_mutex_};
+    if (auto maybe_id = MaybeNameToId(name); maybe_id.has_value()) {
+      return maybe_id.value();
+    }
+
```
DavIvek, line 51: Which part here wasn’t safe? It looks to me like `MaybeNameToId` is using a skip list, which is thread-safe, so I assume only the update part wasn’t safe.

DavIvek (approved):
Left one comment, but looks good!

src/storage/v2/disk/name_id_mapper.hpp
colinbarry: `MaybeNameToId` is safe, but we need to ensure that the whole sequence of: "check cache, check disk, generate id" is atomic. Otherwise, concurrent threads could generate different IDs for the same name.

PR #3530, opened 2025-12-10
feat: Update HA instance config

as51340:
You can now update bolt server of all coordinators and data instances after starting the whole cluster deployment.

### DIFF
--- a/src/query/interpreter.cpp
+++ b/src/query/interpreter.cpp
@@ -1,4 +1,4 @@
-// Copyright 2025 Memgraph Ltd.
+// Copyright 2026 Memgraph Ltd.
 //
 // Use of this software is governed by the Business Source License
 // included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
@@ -1778,7 +1778,7 @@ Callback HandleCoordinatorQuery(CoordinatorQuery *coordinator_query, const Param
 
       // TODO: MemoryResource for EvaluationContext, it should probably be passed as
       // the argument to Callback.
-      EvaluationContext evaluation_context{.timestamp = QueryTimestamp(), .parameters = parameters};
+      EvaluationContext const evaluation_context{.timestamp = QueryTimestamp(), .parameters = parameters};
       auto evaluator = PrimitiveLiteralExpressionEvaluator{evaluation_context};
       auto config_map = ParseConfigMap(coordinator_query->configs_, evaluator);
 


as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - A new query `UPDATE CONFIG FOR` is added to high availability stack that can be executed on the leader coordinator. The query allows modification of the bolt server of each instance. The main motivation for this feature is to support one-step installation of the HA helm chart. [#3530](https://github.com/memgraph/memgraph/pull/3530)
- [x] **[ Documentation PR link ](https://github.com/memgraph/documentation/pull/1506)**

src/coordination/coordinator_instance.cpp
```diff
@@ -260,26 +263,31 @@ auto CoordinatorInstance::ShowInstancesStatusAsFollower() const -> std::vector<I
...
   };
 
-  auto const stringify_repl_health = [](auto &&instance) -> std::string { return instance.IsAlive() ? "up" : "down"; };
+  auto const stringify_repl_health = [](auto const &connector) -> std::string {
+    return connector.IsAlive() ? "up" : "down";
+  };
+
+  auto get_instance_status = [&](auto const &instance) -> InstanceStatus {
+    auto maybe_connector = FindReplicationInstance(instance.config.instance_name, repl_instances_);
+    DMG_ASSERT(maybe_connector.has_value(), "Couldn't find ReplicationInstanceConnector for instance {}",
```
andrejtonev, line 278: throw would be better here since this path could be triggered by a user query.

src/coordination/coordinator_instance.cpp
```diff
@@ -630,24 +644,23 @@ auto CoordinatorInstance::DemoteInstanceToReplica(std::string_view instance_name
...
-
-  auto data_instance = std::ranges::find_if(cluster_state, [instance_name](auto &&data_instance) {
+  auto data_instances_cache = raft_state_->GetDataInstancesContext();
+  auto data_instance = std::ranges::find_if(data_instances_cache, [instance_name](auto const &data_instance) {
     return data_instance.config.instance_name == instance_name;
   });
   data_instance->status = ReplicationRole::REPLICA;
 
+  if (!repl_instance->SendRpc<DemoteMainToReplicaRpc>(data_instance->config.replication_client_info)) {
+    return DemoteInstanceCoordinatorStatus::RPC_FAILED;
```
andrejtonev, line 659: Looks like we are using the find if to find the first main. Is that right?
In any case, return bool here, not the status. 
If I'm following correctly `return DemoteInstanceCoordinatorStatus::RPC_FAILED;` is returning to the find_if and should return false

src/coordination/coordinator_instance.cpp
```diff
@@ -1392,5 +1429,84 @@ auto CoordinatorInstance::GetTelemetryJson() const -> nlohmann::json {
...
+
+    // If we managed to add it to the NuRaft configuration but not to our app logs.
+    if (!raft_state_->AppendClusterUpdate(delta_state)) {
+      spdlog::error("Couldn't append application log when updating the config for the coordinator {}.", coord_id);
+      return UpdateConfigStatus::RAFT_FAILURE;
+    }
+
+  } else {
+    // Updating a config for the replication instance
+    auto const instance_name = std::get<std::string>(instance);
```
andrejtonev, line 1473: check that it is a string

src/coordination/coordinator_instance.cpp
```diff
@@ -1392,5 +1429,84 @@ auto CoordinatorInstance::GetTelemetryJson() const -> nlohmann::json {
...
+    }
+
+    auto repl_instance_it = std::ranges::find_if(data_instances_context, [instance_name](auto const &repl_instance) {
+      return repl_instance.config.instance_name == instance_name;
+    });
+    if (repl_instance_it == std::ranges::end(data_instances_context)) {
+      return UpdateConfigStatus::NO_SUCH_REPL_INSTANCE;
+    }
+
+    repl_instance_it->config.bolt_server = bolt_endpoint;
```
andrejtonev, line 1496: is this thread safe?

src/coordination/coordinator_instance.cpp
as51340: I disagree because if maybe_connector doesn't exist my app logic is completely wrong. This is not something that should be handled or user should know about it, this is just me trying to be defensive.

src/coordination/coordinator_instance.cpp
as51340: maybe I am misunderstanding but I am using find_if to find the instance with the name instance_name. Then I try to send rpc and if rpc fails, I return from the function enum.

src/coordination/coordinator_instance.cpp
as51340: instance is a variant of string and int32_t. In line 1434 I checked if the variant contains int32_t. This code is in else branch so should be fine

src/coordination/coordinator_instance.cpp
as51340: yes, this is just local object on stack. From that object I am creating delta object which gets converted to raft log and then raft log commits in the bg thread.

PR #3562, opened 2025-12-15
Add parallel recovery for vector index

DavIvek:
This PR adds a parallel index recovery option for vector indices, along with several additional code improvements, including better handling of index resizing and entry removal.

From now on, an entry is removed only if it is found in the index. Although `remove()` already checks whether the entry is present, it acquires a unique lock, whereas the `contains()` method does not. Since most vector index workloads do not update indices frequently, it should be beneficial to check `contains()` first before performing removal.

### DIFF
--- a/src/storage/v2/indices/vector_edge_index.cpp
+++ b/src/storage/v2/indices/vector_edge_index.cpp
@@ -117,7 +117,7 @@ bool VectorEdgeIndex::CreateIndex(const VectorEdgeIndexSpec &spec, utils::SkipLi
   return true;
 }
 
-bool VectorEdgeIndex::RecoverIndex(const VectorEdgeIndexSpec &spec, utils::SkipList<Vertex>::Accessor &vertices,
+void VectorEdgeIndex::RecoverIndex(const VectorEdgeIndexSpec &spec, utils::SkipList<Vertex>::Accessor &vertices,
                                    std::optional<SnapshotObserverInfo> const &snapshot_info) {
   const utils::MemoryTracker::OutOfMemoryExceptionEnabler oom_exception;
   try {
@@ -132,7 +132,6 @@ bool VectorEdgeIndex::RecoverIndex(const VectorEdgeIndexSpec &spec, utils::SkipL
     CleanupFailedIndex(spec);
     throw;
   }
-  return true;
 }
 
 void VectorEdgeIndex::SetupIndex(const VectorEdgeIndexSpec &spec) {
--- a/src/storage/v2/indices/vector_edge_index.hpp
+++ b/src/storage/v2/indices/vector_edge_index.hpp
@@ -98,8 +98,8 @@ class VectorEdgeIndex {
   /// @param spec The specification for the index to be recovered.
   /// @param vertices vertices from which to recover the index.
   /// @param snapshot_info Optional snapshot observer for progress tracking.
-  /// @return true if the index was recovered successfully, false otherwise.
-  bool RecoverIndex(const VectorEdgeIndexSpec &spec, utils::SkipList<Vertex>::Accessor &vertices,
+  /// @param thread_id Optional thread ID hint for usearch's internal optimizations.
+  void RecoverIndex(const VectorEdgeIndexSpec &spec, utils::SkipList<Vertex>::Accessor &vertices,
                     std::optional<SnapshotObserverInfo> const &snapshot_info = std::nullopt);
 
   /// @brief Drops an existing index.
--- a/src/storage/v2/indices/vector_index.cpp
+++ b/src/storage/v2/indices/vector_index.cpp
@@ -109,7 +109,7 @@ bool VectorIndex::CreateIndex(const VectorIndexSpec &spec, utils::SkipList<Verte
   return true;
 }
 
-bool VectorIndex::RecoverIndex(const VectorIndexSpec &spec, utils::SkipList<Vertex>::Accessor &vertices,
+void VectorIndex::RecoverIndex(const VectorIndexSpec &spec, utils::SkipList<Vertex>::Accessor &vertices,
                                std::optional<SnapshotObserverInfo> const &snapshot_info) {
   const utils::MemoryTracker::OutOfMemoryExceptionEnabler oom_exception;
   try {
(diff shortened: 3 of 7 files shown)


DavIvek:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Vector index recovery can now run in parallel, with additional improvements to index resizing and safer entry removal logic. This results in faster startup and schema recovery for databases with large vector indexes. Users can enable this behavior by starting Memgraph with --storage-parallel-schema-recovery=true. [#3562](https://github.com/memgraph/memgraph/pull/3562).

src/storage/v2/indices/vector_index.cpp
```diff
@@ -61,59 +61,124 @@ struct VectorIndex::Impl {
...
+  auto property = vertex.properties.GetProperty(spec.property);
+  if (property.IsNull()) {
+    return;
+  }
+  auto vector = PropertyToFloatVector(property, spec.dimension);
+  if (vector_index.size() >= vector_index.capacity()) {
+    throw query::VectorSearchException(
+        "Vector index is full. Try increasing the capacity.");  // TODO(@DavIvek): Should we do this for user?
+  }
+  vector_index.add(&vertex, vector.data());
```
colinbarry: I think we need to check the result of this `add` and handle errors. Because it can be called concurrently via`PopulateIndexOnMultipleThreads`, we could have two threads concurrently check for capacity, both try to add an item, and for the second one to silently fail.

src/storage/v2/indices/vector_edge_index.cpp
```diff
@@ -52,72 +54,141 @@ struct VectorEdgeIndex::Impl {
...
+    }
+    auto property = edge->properties.GetProperty(spec.property);
+    if (property.IsNull()) {
+      continue;
+    }
+    auto vector = PropertyToFloatVector(property, spec.dimension);
+    if (vector_index.size() >= vector_index.capacity()) {
+      throw query::VectorSearchException("Vector index is full. Try increasing the capacity.");
+    }
+    vector_index.add({&from_vertex, to_vertex, edge}, vector.data());
```
colinbarry: Think we need to check this as well. See comment on `TryAddVertexToIndex` below.

src/storage/v2/indices/vector_edge_index.cpp
```diff
@@ -196,10 +269,22 @@ void VectorEdgeIndex::UpdateOnSetProperty(Vertex *from_vertex, Vertex *to_vertex
...
   if (std::ranges::any_of(pimpl->edge_index_ | rv::keys | rv::filter(has_property),
                           [&](const auto &edge_type_prop) { return edge_type_prop.edge_type() == edge_type; })) {
-    UpdateVectorIndex({from_vertex, to_vertex, edge}, EdgeTypePropKey{edge_type, property}, &value);
+    UpdateVectorIndex({.from_vertex = from_vertex, .to_vertex = to_vertex, .edge = edge},
+                      EdgeTypePropKey{edge_type, property}, &value);
   }
 }
 
+void VectorEdgeIndex::RemoveEdge(Edge *edge, EdgeTypeId edge_type, Vertex *from_vertex, Vertex *to_vertex) {
+  r::for_each(pimpl->edge_index_ | rv::keys, [&](const auto &edge_type_prop) {
```
colinbarry: `RemoveEdge` doesn't seem to be called from anywhere.

src/storage/v2/indices/vector_index.cpp
```diff
@@ -61,59 +61,124 @@ struct VectorIndex::Impl {
...
+    return;
+  }
+  auto property = vertex.properties.GetProperty(spec.property);
+  if (property.IsNull()) {
+    return;
+  }
+  auto vector = PropertyToFloatVector(property, spec.dimension);
+  if (vector_index.size() >= vector_index.capacity()) {
+    throw query::VectorSearchException(
+        "Vector index is full. Try increasing the capacity.");  // TODO(@DavIvek): Should we do this for user?
```
colinbarry: I think the TODO is correct: we should automatically do it. Otherwise, isn't the following scenario possible:
- snapshot taken with 100 vertices in index, and capacity of 100
- we add 50 more vertices, and capacity is automatically increased
- crash!
- we recover the snapshot, with a capacity of 100, but when we replay the WAL, the 50 additional vertices exceed the capacity, and so recovery will fail with "Vector index is full. Try increasing the capacity."

colinbarry (changes requested):
Good stuff! Just a few observations about:
- what happens if we exceed capacity during recovery
- what happens if two threads concurrently exceed capacity

src/storage/v2/indices/vector_edge_index.hpp
```diff
@@ -88,12 +88,20 @@ class VectorEdgeIndex {
...
   /// @return true if the index was created successfully, false otherwise.
   bool CreateIndex(const VectorEdgeIndexSpec &spec, utils::SkipList<Vertex>::Accessor &vertices,
                    std::optional<SnapshotObserverInfo> const &snapshot_info = std::nullopt);
 
+  /// @brief Recovers a vector edge index based on the provided specification.
+  /// @param spec The specification for the index to be recovered.
+  /// @param vertices vertices from which to recover the index.
+  /// @param snapshot_info Optional snapshot observer for progress tracking.
+  /// @return true if the index was recovered successfully, false otherwise.
+  bool RecoverIndex(const VectorEdgeIndexSpec &spec, utils::SkipList<Vertex>::Accessor &vertices,
```
colinbarry: Not sure we need the `bool` return type here, as is appears that `RecoverIndex` either succeeds, or throws an exception. So (currently), it will never return `false`. But... this is consistent with all the other `CreateIndex` functions, so there is argument for leaving as is. Your choice!

colinbarry (approved):
All good. Only observation is that the `RecoverIndex` func has an unused `bool` success return value that isn't actually used (but I understand why it's there for consistency with other index code.)

src/storage/v2/indices/vector_edge_index.hpp
DavIvek: We don't need a bool, you are correct.

PR #1981, opened 2024-04-25
Expose coordinator's internal functionality to manage cluster

antoniofilipovic:
### Description

This diff adds the option for users to run actions like DEMOTE INSTANCE and FORCE RESET CLUSTER STATE, which are run internally but now open for users to reset cluster state, and choose their instance to be MAIN.

TODO:
- [ ] learn how to trim commits in merge to master, ping @as51340 
- [x] refactor force reset cluster 

[master < Task] PR
- [x] Provide the full content or a guide for the final git message
    - Expose some internal functionality to manage cluster

### CI Testing Labels
Please select the appropriate CI test labels _(CI -build=**build-name** -test=**test-suite**)_

### Documentation checklist
- [x] Add the documentation label tag
- [x] Add the bug / feature label tag
- [x] Add the milestone for which this feature is intended
    - If not known, set for a later milestone
- [x] Write a release note, including added/changed clauses
    - You can now run demote instance to replica and force reset cluster
- [ ] Link the documentation PR here
    - **[Documentation PR link]**
- [x] Tag someone from docs team in the comments

PR #3567, opened 2025-12-16
Robust shutdown process

andrejtonev:
Moved Settings to be held under main (from a static global object)
Shutting down the whole CoordinatorState instead of just setting an atomic flag
Removed logs from ~KVStore

### DIFF
--- a/src/query/interpreter.cpp
+++ b/src/query/interpreter.cpp
@@ -5140,6 +5140,7 @@ PreparedQuery PrepareShowNextSnapshotQuery(ParsedQuery parsed_query, bool in_exp
                        RWType::NONE};
 }
 
+namespace {
 PreparedQuery PrepareSettingQuery(ParsedQuery parsed_query, const bool in_explicit_transaction,
                                   InterpreterContext *interpreter_context) {
   if (in_explicit_transaction) {
@@ -5167,6 +5168,7 @@ PreparedQuery PrepareSettingQuery(ParsedQuery parsed_query, const bool in_explic
   // False positive report for the std::make_shared above
   // NOLINTNEXTLINE(clang-analyzer-cplusplus.NewDeleteLeaks)
 }
+}  // namespace
 
 template <typename Func>
 auto ShowTransactions(const std::unordered_set<Interpreter *> &interpreters, QueryUserOrRole *user_or_role,


andrejtonev:

---
___

### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Complete shutdown of coordinators and Settings held locally under main(). Avoids segfaults while destroying static objects in an uncontrolled order. Avoids unexpected errors from nuraft while shutting down or exiting. [#3567](https://github.com/memgraph/memgraph/pull/3567)

src/kvstore/kvstore.cpp
```diff
@@ -39,11 +39,8 @@ KVStore::KVStore(std::filesystem::path storage) : pimpl_(std::make_unique<impl>(
 
 KVStore::~KVStore() {
   if (pimpl_ == nullptr) return;
-  spdlog::debug("Destroying KVStore at {}", pimpl_->storage.string());
-  const auto sync = pimpl_->db->SyncWAL();
-  if (!sync.ok()) spdlog::error("KVStore sync failed!");
-  const auto close = pimpl_->db->Close();
-  if (!close.ok()) spdlog::error("KVStore close failed!");
+  (void)pimpl_->db->SyncWAL();
+  (void)pimpl_->db->Close();
```
Ignition, line 43: Is this still needed?

src/memgraph.cpp
```diff
@@ -624,7 +621,7 @@ int main(int argc, char **argv) {
   auto db_acc = dbms_handler.Get();
 
   memgraph::query::InterpreterContextLifetimeControl interpreter_context_lifetime_control(
-      interp_config, &dbms_handler, repl_state, system,
+      interp_config, settings.get(), &dbms_handler, repl_state, system,
```
Ignition, line 625: Does raw ptr vs weak ptr matter here. I've not thought hard enough to figure out if it matters or not.

Ignition (approved):

src/memgraph.cpp
andrejtonev: weak pointer is only for the license checker.  License checker is another static object and it would be very hard to make it not be static.
Using weak pointer so the license checker's background thread can run on even if the settings gets destroyed. That way I don't need to explicitly sync anything on shutdown or exit.

andrejtonev: Everything else is owned by the main and is created after the settings, so will be destroyed before settings.

src/kvstore/kvstore.cpp
andrejtonev: ~~yes, that's the correct way to close rocksdb. the issue was the logging, not the closing itself~~
you meant the change. no, not really.

src/memgraph.cpp
```diff
@@ -772,7 +769,8 @@ int main(int argc, char **argv) {
 #ifdef MG_ENTERPRISE
     metrics_server.Shutdown();
     if (coordinator_state.has_value() && coordinator_state->IsCoordinator()) {
-      coordinator_state->ShutDownCoordinator();
+      // Coordinator instance destruction will handle the complete shutdown
```
as51340, line 773: Not sure if we can remove this ShutDownCoordinator(). We introduced it here https://github.com/memgraph/memgraph/pull/1981, I think the reason was to stop NuRaft from executing a leader callback during destruction

as51340:
if all tests pass successfully, delete is_shutting_down from coordinator_instance.hpp

src/kvstore/kvstore.cpp
andrejtonev: decided to remove them in the end

andrejtonev:
@as51340 I'll leave the atomic because of the shutdown process of the CoordinatorInsatnce
```
CoordinatorInstance::~CoordinatorInstance() {
  // Coordinator instance can be stuck in infinite loop of doing reconciliation of cluster state. We need to notify
  // thread that we are shutting down, and we need to stop state checks and thread pool before anything is destructed
  // otherwise thread pool can get task become leader, when raft_state_ ptr is already destructed
  // Order is important:
  // 1. Thread pool might be running with reconcile cluster state, or become follower, so we should shut that down
  // 2. Await shutdown of coordinator thread pool so that coordinator can't become follower or do reconcile cluster
  // state
  // 3. State checks running, we need to stop them before raft_state_ is destroyed
  ShuttingDown();
  thread_pool_.ShutDown();
  // We don't need to take lock as reconcile cluster state can't be running, coordinator can't become follower,
  // user queries can't be running as memgraph awaits server shutdown
  std::ranges::for_each(repl_instances_, [](auto &repl_instance) { repl_instance.StopStateCheck(); });
}
```

PR #3570, opened 2025-12-17
Fix text index recovery from snapshot

DavIvek:
Currently, text index recovery works only when the text index directories exist; otherwise, the index is not recovered. This PR fixes this issue by checking whether the index folder exists and, if it does not, recovering it from the snapshot.
Additionally, it improves the safety of the Clear() function (which is called during DROP GRAPH). Previously, this function was not safe because it could fail during execution and leave the system in an intermediate state if a failure occurred while clearing.

### DIFF
--- a/src/storage/v2/indices/text_edge_index.cpp
+++ b/src/storage/v2/indices/text_edge_index.cpp
@@ -261,19 +261,25 @@ std::vector<TextEdgeIndexSpec> TextEdgeIndex::ListIndices() const {
 }
 
 void TextEdgeIndex::Clear() {
-  std::vector<TextEdgeIndexSpec> index_specs_to_recover;
-  index_specs_to_recover.reserve(index_.size());
+  // Collect all index specs before modifying the map -> we will need to recover them if we fail to drop any of them
+  std::vector<TextEdgeIndexSpec> all_index_specs;
+  all_index_specs.reserve(index_.size());
   for (const auto &[index_name, index_data] : index_) {
+    all_index_specs.emplace_back(index_name, index_data.scope, index_data.properties);
+  }
+
+  std::vector<TextEdgeIndexSpec> successfully_dropped;
+  successfully_dropped.reserve(all_index_specs.size());
+  for (const auto &index_spec : all_index_specs) {
     try {
-      // We have to keep this order of operations because DropIndex removes the index from the map
-      index_specs_to_recover.emplace_back(index_name, index_data.scope, index_data.properties);
-      DropIndex(index_name);
+      DropIndex(index_spec.index_name);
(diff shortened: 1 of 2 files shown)


DavIvek:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Text index recovery now works correctly from snapshots even when the index directories are missing, allowing users to restore full-text indexes using only a snapshot, which was not possible before. Additionally, the Clear() function used during DROP GRAPH has been made safer to prevent intermediate system states in case of failures. [#3570](https://github.com/memgraph/memgraph/pull/3570)

src/storage/v2/indices/text_edge_index.cpp
```diff
@@ -222,15 +261,21 @@ std::vector<TextEdgeIndexSpec> TextEdgeIndex::ListIndices() const {
...
-    if (ec) {
-      spdlog::error("Error removing text edge index directory '{}': {}", text_index_storage_dir_, ec.message());
-      return;
+  std::vector<TextEdgeIndexSpec> index_specs_to_recover;
+  index_specs_to_recover.reserve(index_.size());
+  for (const auto &[index_name, index_data] : index_) {
+    try {
+      // We have to keep this order of operations because DropIndex removes the index from the map
+      index_specs_to_recover.emplace_back(index_name, index_data.scope, index_data.properties);
+      DropIndex(index_name);
```
colinbarry: Using `index_.extract` (inside `DropIndex`) whilst iterating a range-based for loop is undefined behaviour. 

See https://en.cppreference.com/w/cpp/container/map/extract.html - "Extracting a node invalidates only the iterators to the extracted element." 

Not sure the best way to refactor these; maybe a two-pass collect and then drop?

src/storage/v2/indices/text_index.cpp
```diff
@@ -236,14 +271,20 @@ std::optional<uint64_t> TextIndex::ApproximateVerticesTextCount(std::string_view
...
-    std::filesystem::remove_all(text_index_storage_dir_, ec);
-    if (ec) {
-      spdlog::error("Error removing text index directory '{}': {}", text_index_storage_dir_, ec.message());
-      return;
+  std::vector<TextIndexSpec> index_specs_to_recover;
+  index_specs_to_recover.reserve(index_.size());
+  for (const auto &[index_name, index_data] : index_) {
+    try {
+      // We have to keep this order of operations because DropIndex removes the index from the map
+      index_specs_to_recover.emplace_back(index_name, index_data.scope, index_data.properties);
```
colinbarry: Same issue here with `DropIndex` invalidating the current iterator. See the notes I made in `TextEdgeIndex::Clear()`

PR #3575, opened 2025-12-19
testing: Run Jepsen test for 6h

as51340:
Since we use now larger concurrency for Jepsen test, 1 in 4 times approx. tests go OOM. Therefore, trying it to run it for a shorter time.


as51340:
Fixed by reducing the number of concurrent clients that runs on top of Memgraph

Issue #1909, opened 2024-04-05
Implement Point in Time Recovery


gitbuda:

antejavor:
+1 in community

PR #3596, opened 2025-12-30
feat: Recover snapshot from remote server

as51340:
Users can now download a file from the remote S3, http, https and ftp server. If S3 is used, authentication parameters can be provided using run-time database settings, env variables or through the query using WITH CONFIG query construct. If downloading the file fails, the file won't visible after the recover snapshot operation finishes. An optimisation is added to the recover snapshot which if the file is being loaded from the local directory, it is copied to the .old directory but was never deleted from the old directory if all went well. Now we delete it.

Curl code now uses FAILONERROR property to handle invalid curl return codes. Aws module doesn't throw exceptions but rather now returns bools status. The caller can throw exceptions as it wishes based on bool status codes.


as51340:
### Tracking
- [x] **[Link to Epic/Issue](https://github.com/memgraph/memgraph/issues/1909)**

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Users can load a snapshot directly from any s3 storage and from any http/https/ftp URL using RECOVER SNAPSHOT query. AWS auth parameters can be provided through the query config map, env variables or run-time settings. [#3596](https://github.com/memgraph/memgraph/pull/3596) 
- [x] **[ Documentation PR link ](https://github.com/memgraph/documentation/pull/1507)**

src/query/interpreter.cpp
```diff
@@ -5082,27 +5101,60 @@ PreparedQuery PrepareRecoverSnapshotQuery(ParsedQuery parsed_query, bool in_expl
...
+  if (maybe_config_map->size() > 4) {
+    throw QueryRuntimeException("Config map cannot contain > 4 entries. Only {}, {}, {} and {} can be provided",
+                                utils::kAwsRegionQuerySetting, utils::kAwsAccessKeyQuerySetting,
+                                utils::kAwsSecretKeyQuerySetting, utils::kAwsEndpointUrlQuerySetting);
+  }
+
+  constexpr auto s3_matcher = ctre::starts_with<"s3://">;
+
+  std::optional<utils::S3Config> s3_config;
+  if (s3_matcher(path_value.ValueString())) {
```
andrejtonev, line 5107: we could build the config only for the s3 case

src/storage/v2/inmemory/storage.cpp
```diff
@@ -3074,32 +3079,61 @@ std::expected<std::filesystem::path, InMemoryStorage::CreateSnapshotError> InMem
...
+      return std::unexpected{InMemoryStorage::RecoverSnapshotError::S3GetFailure};
+    }
+
+  } else {  // local filesystem path
+    if (!std::filesystem::exists(uri) || std::filesystem::is_directory(uri)) {
+      return std::unexpected{InMemoryStorage::RecoverSnapshotError::MissingFile};
+    }
+
+    // Copy to local snapshot dir
+    std::error_code ec{};
```
andrejtonev, line 3127: we were talking about trying to load and then copying on success. is that possible? does it actually make sense?

src/storage/v2/inmemory/storage.cpp
```diff
@@ -3074,32 +3079,61 @@ std::expected<std::filesystem::path, InMemoryStorage::CreateSnapshotError> InMem
...
+    if (!file_in_local_dir) {
+      std::filesystem::copy_file(uri, local_path, std::filesystem::copy_options::overwrite_existing, ec);
+      if (ec) {
+        spdlog::warn("Failed to copy snapshot into local snapshots directory. Err: {}", ec.message());
+        return std::unexpected{InMemoryStorage::RecoverSnapshotError::CopyFailure};
+      }
+    }
+  }
+
```
andrejtonev, line 3136: putting the comment here because I can't add it to a stable line
below there is some logic that copies a local file to the old before we touch it (change the name and uuid)
perhaps that is too aggressive. 
renaming will not change the file
perhaps uuid could break the file, but even in that case, we should copy it only if the uuid is different

src/query/interpreter.cpp
as51340: it is currently built only for the s3 scenario. Can you explain?

src/storage/v2/inmemory/storage.cpp
as51340: it's possible but first this is not a proper solution because you could still end up with invalid disk state. Second, I can do that before doing the manifest thing but I will definitely then separate it from this pr

src/storage/v2/inmemory/storage.cpp
as51340: I checked that part, it makes sense to me as it is now. We are copying to .old only if you're doing recover snapshot from the data directory folder. In this situation, it makes sense to me to somehow preserver the old snapshot file so I would leave it as it is.

src/query/interpreter.cpp
andrejtonev: we build the runtime config regardless if we are actually fetching from s3
we could built it only if the s3 fetch has been requested or throw if config was specified for other types or if not specified in case of s3

PR #3602, opened 2026-01-02
fix: Delete partially written aborted snapshot file

as51340:
Snapshots which are getting aborted, should get deleted immediately. Otherwise, we lose processing time on reading this snapshot and figuring out that it wasn't fully written. This PR currently solves this.

There is also a question how to delete snapshots which aren't fully written because an instance went down during snapshot creation. One way to do that would be to conclude that the snapshot is partially written if we cannot read section metadata, since that section is the last one in a snapshot file.

Another thing which is I think going on at the moment is that there could be snapshot getting created and in parallel GetSnapshotFiles which will read all snapshot files. This won't cause bugs but we again lose time on reading snapshot + in logs we are seeing all the time 'couldn't read marker' error messages => Bad UX.

### DIFF
--- a/src/storage/v2/durability/durability.cpp
+++ b/src/storage/v2/durability/durability.cpp
@@ -1,4 +1,4 @@
-// Copyright 2025 Memgraph Ltd.
+// Copyright 2026 Memgraph Ltd.
 //
 // Use of this software is governed by the Business Source License
 // included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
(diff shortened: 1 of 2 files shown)


as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Aborted snapshots will now be deleted immediately after abortion. They will not pollute anymore disk space hence the user should observer more efficient disk management. [#3602](https://github.com/memgraph/memgraph/pull/3602)

andrejtonev (approved):

PR #3608, opened 2026-01-05
Cache GetExecutablePath() result

Ignition:
Cache the result of reading /proc/self/exe in a static local variable, eliminating a readlink syscall on every call after the first.

Impact: Reduces syscall overhead during startup and runtime. No user-visible behavior change.

### DIFF
--- a/src/utils/file.cpp
+++ b/src/utils/file.cpp
@@ -1,4 +1,4 @@
-// Copyright 2025 Memgraph Ltd.
+// Copyright 2026 Memgraph Ltd.
 //
 // Use of this software is governed by the Business Source License
 // included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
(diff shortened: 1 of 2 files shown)


Ignition:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Cache GetExecutablePath(), minor performance improvement [#3608](https://github.com/memgraph/memgraph/pull/3608)

colinbarry (commented):
Code looks good to me. My question is: what performance benefit do we get from this? As far as I can tell `GetExecutablePath` is only ever called once per Memgraph instance - during setup in `main` - so doesn't benefit from memoization in the current code base, as is? Maybe I'm just missing the bigger picture and you forsee us using calling this function in other places in the future?

Ignition:
> As far as I can tell `GetExecutablePath` is only ever called once per Memgraph instance

Your right, it came up as an AI assisted suggestion. Locally it looked like a reasonable suggestion, I didn't check to see that it was only called once and hence doesn't matter.

PR #3634, opened 2026-01-16
fix: Thread-safe usage of spdlog

as51340:
Before, the swapping of the logger wasn't thread safe. This cannot be mitigated by just adding new sink because sinks are not protected.

The fix is to use dist_sink_mt which overrides base_sink. Sink dist_sink_mt takes lock when logging and when adding a new sink. Additionally, the code should now be more performant because for logging, we are now taking only a single lock compared to previously 3 (stderr, file and web socket locks). 

inline stderr sink is used because sinks vector is still not protected from reallocations so one thread could cause a vector reallocation while the other thread is relying on vector's pointer stability.

### DIFF
--- a/src/flags/log_level.cpp
+++ b/src/flags/log_level.cpp
@@ -79,10 +79,10 @@ std::optional<spdlog::level::level_enum> memgraph::flags::LogLevelToEnum(std::st
 // This allows us MT safe
 void memgraph::flags::InitializeLogger() {
   // stderr subsink
-  stderr_sink->set_level(spdlog::level::off);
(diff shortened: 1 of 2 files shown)


as51340:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - We now use `dist_sink_mt` which should fix the race condition between the main thread and any other thread when adding a sink. Logging should also be more efficient because we now use a single mutex compared to previously 3. [#3634](https://github.com/memgraph/memgraph/pull/3634)

Ignition (approved):

PR #3638, opened 2026-01-19
fix: Restart repl server on UUID change

as51340:


as51340:
Not an issue at the moment, aborting work on this currently

PR #3644, opened 2026-01-20
fix: type constraint validation and MVCC cleanup

### LINKED ISSUE #2935: Crash (SIGSEGV/Exit Code 139) when setting NULL with TYPED STRING constraint
**Memgraph version**
I'm using `latest` docker images from official compose template:
- memgraph/memgraph-mage `3.2`
- memgraph/lab `3.2.0`

**Environment**
- Architecture: x86_64
- Deployment: Docker (using memgraph/memgraph-mage:latest)
- Connection Method: tried on both Memgraph Lab, neo4rs client

**Describe the bug**
Memgraph crashes with a Segmentation Fault (SIGSEGV), indicated by exit code 139 in Docker, when executing a query that set a property to `NULL`. This crash only occurs if that property has an `IS TYPED` constraint. 
The crash does not happen if the constraint is removed.

**To Reproduce**
1. Run the following constraints first
```cypher
CREATE INDEX ON :Node(id);
CREATE CONSTRAINT ON (n:Node) ASSERT EXISTS (n.id);
CREATE CONSTRAINT ON (n:Node) ASSERT n.id IS UNIQUE;
CREATE INDEX ON :Node(id);
CREATE CONSTRAINT ON (n:Node) ASSERT n.id IS TYPED STRING;
CREATE CONSTRAINT ON (n:Node) ASSERT n.name IS TYPED STRING;
```
2. Execute now this write operation
```cypher
MERGE (n:Node {id: "42"})
SET n.name = "test"
RETURN n;

// This below will cause the crash
MATCH (n:Node)
WHERE contains(n.id, "42")
set n.name = NULL
RETURN n;
```
3. Now the operation in hanging till crash happens
4. If we run again without `CREATE CONSTRAINT ON (n:Node) ASSERT n.name IS TYPED STRING;` everything work as expected.

**Expected behavior**
This bug seems preventing the reliable use of `IS TYPED` constraints on properties that might need to be conditionally set to `NULL` during data ingestion or updates (e.g., handling optional fields). 

Am I missing something indeed?

**Logs**
From Memgraph Lab we are able to see this after a while:

> Query failed
> Connection was closed by server

Ignition:
Bug Fix:
Setting a property to NULL on a vertex with a type constraint would crash with an assertion failure: "Unexpected conversion from PropertyValueType::Null to TypeConstraint::Type". Type constraints only enforce the type when the property has a value, not its existence - NULL (property removal) is now allowed.

MVCC Refactoring:
This commit completes the MVCC constraints cleanup following the race condition fix in commit 9a8f48c82. The refactoring consolidates constraint interfaces and moves validation to use transaction-scoped active constraints.

File Rename:
- schema_status.hpp → population_status.hpp (struct SchemaStatus → PopulationStatus)

Interface Consolidation:
- Moved constraint ActiveConstraints interfaces from active_constraints.hpp to their respective headers (existence_constraints.hpp, type_constraints.hpp, unique_constraints.hpp)
- active_constraints.hpp now only contains the aggregating struct
- Added UniqueConstraints::ActiveConstraints virtual base class with InMemory/Disk implementations

Type Constraints:
- Added PropertyValueMatchesTypeConstraint helper to type_constraints_kind.hpp
- Added Validate methods to TypeConstraints::ActiveConstraints
- Moved internal validation to file-local ValidateVertex function
- Removed unused GetSnapshot() and Validate(vertex) methods
- vertex_accessor.cpp now uses transaction->active_constraints_.type_->Validate()

Performance Optimization:
- UpdateProperties now validates each property with extracted values instead of re-reading from the property store via Validate(*vertex)

Transaction Changes:
- Removed has_constraints parameter from Transaction constructor
- Replaced with active_constraints.empty() check

Cleanup:
- Removed Constraints::AbortPopulating() - the only production failure path is validation failure, which self-cleans via explicit DropConstraint call before returning the error. Timeout and cancellation during constraint creation are not currently supported.
- Removed Constraints::HasTypeConstraints() - replaced with empty() check
- Removed 3 tests for old AbortPopulating mechanism, updated test comments

OOM Exception Handling:
- Added try-catch for OutOfMemoryException in constraint creation methods
- CreateExistenceConstraint, CreateUniqueConstraint, and CreateTypeConstraint now clean up registered constraints on OOM before re-throwing
- This prevents resource leaks where constraints remain in POPULATING state if OOM is thrown during validation

Test Changes:
- Added TypeConstraintsSetPropertyToNull test
- Added TypeConstraintsUpdatePropertiesToNull test
- Updated MVCC test comments to reference new status-based cleanup

resolves #2935

### DIFF
--- a/src/storage/v2/constraints/type_constraints_kind.hpp
+++ b/src/storage/v2/constraints/type_constraints_kind.hpp
@@ -69,6 +69,7 @@ inline std::string_view TypeConstraintKindToString(TypeConstraintKind type) {
     case TypeConstraintKind::POINT:
       return "POINT"sv;
   }
+  __builtin_unreachable();
 }
 
 inline PropertyStoreType TypeConstraintsKindToPropertyStoreType(TypeConstraintKind type) {
@@ -97,6 +98,7 @@ inline PropertyStoreType TypeConstraintsKindToPropertyStoreType(TypeConstraintKi
     case TypeConstraintKind::POINT:
       return PropertyStoreType::POINT;
   }
+  __builtin_unreachable();
 }
 
 inline bool TemporalMatch(TemporalType type, TypeConstraintKind expected_type) {
@@ -110,6 +112,7 @@ inline bool TemporalMatch(TemporalType type, TypeConstraintKind expected_type) {
     case TemporalType::Duration:
       return expected_type == TypeConstraintKind::DURATION;
   }
+  __builtin_unreachable();
 }
 
 /// Convert a PropertyValue to its corresponding TypeConstraintKind.
@@ -154,6 +157,7 @@ inline TypeConstraintKind PropertyValueToTypeConstraintKind(const PropertyValue 
     case PropertyValueType::Null:
       MG_ASSERT(false, "Unexpected conversion from PropertyValueType::Null to TypeConstraint::Type");
   }
+  __builtin_unreachable();
 }
 
 /// Check if a PropertyValue matches a TypeConstraintKind.
--- a/src/storage/v2/inmemory/unique_constraints.hpp
+++ b/src/storage/v2/inmemory/unique_constraints.hpp
@@ -92,11 +92,13 @@ class InMemoryUniqueConstraints : public UniqueConstraints {
     void AbortEntries(AbortableInfo &&info, uint64_t exact_start_timestamp) override;
     bool empty() const override;
 
-    void UpdateOnRemoveLabel(LabelId removed_label, const Vertex &vertex_before_update,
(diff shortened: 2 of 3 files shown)


Ignition:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Fixed crash when removing properties on type-constrained vertices. Setting a property to NULL (i.e., removing it) on a vertex with a type constraint no longer causes an assertion failure. Type constraints enforce the property's type only when a value is present, they don't require the property to exist. [#3644](https://github.com/memgraph/memgraph/pull/3644)

PR #3713, opened 2026-01-23
fix: Destroy dbms handler after global repl state

as51340:
Dbms handler doesn't have a reference to the repl state anymore. TTL is set-up so that writes are disallowed and the user_check_ function is set-up in memgraph.cpp after the initialisation and in the interpreter after the database creation. Shutdown protocol is modified a bit in a sense that we first shutdown repl state, then all background jobs (streams, triggers, TTL) and then all storage repl clients. Repl state gets destroyed before dbms handler. 

SwapMainUUIDRpc when received from the main instance will now lock a global repl state and then change the uuid.  All other inmemory/replication_handlers.cpp work with cached RoleReplicaData because we have a guarantee that the reference won't change for as long as the instance is replica.

### DIFF
--- a/src/memgraph.cpp
+++ b/src/memgraph.cpp
@@ -834,6 +834,7 @@ int main(int argc, char **argv) {
                       &server,
                       &interpreter_context_,
                       &dbms_handler,
+                      &repl_state,
                       &worker_pool_] {
     // Server needs to be shutdown first and then the database. This prevents
     // a race condition when a transaction is accepted during server shutdown.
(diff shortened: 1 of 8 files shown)


as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - In situations where users had a lot of databases (500+) the segfault on the shutdown could've happened because databases were destroyed while global replication clients were still alive. This is now fixed and global replication clients will be shutdown before databases. [#3713](https://github.com/memgraph/memgraph/pull/3713)

andrejtonev (approved):

PR #3717, opened 2026-01-26
Add global server side parameters store

DavIvek:
This PR adds global store for server side parameters.

### DIFF
--- a/src/replication_handler/system_rpc.cpp
+++ b/src/replication_handler/system_rpc.cpp
@@ -19,23 +19,19 @@
 
 namespace memgraph::slk {
 
-namespace {
-
-// ParameterInfo SLK (for SystemRecoveryReq::parameters); defined here so replication_handler stays independent of
-// mg-parameters.
+// NOLINTNEXTLINE(misc-use-internal-linkage)
 void Save(const memgraph::parameters::ParameterInfo &self, memgraph::slk::Builder *builder) {
   memgraph::slk::Save(self.name, builder);
   memgraph::slk::Save(self.value, builder);
   memgraph::slk::Save(self.scope, builder);
 }
 
+// NOLINTNEXTLINE(misc-use-internal-linkage)
 void Load(memgraph::parameters::ParameterInfo *self, memgraph::slk::Reader *reader) {
   memgraph::slk::Load(&self->name, reader);
   memgraph::slk::Load(&self->value, reader);
   memgraph::slk::Load(&self->scope, reader);
 }
-
-}  // namespace
 
 // Serialize code for SystemRecoveryReqV1
 void Save(const memgraph::replication::SystemRecoveryReqV1 &self, memgraph::slk::Builder *builder) {


DavIvek:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Added support for server-side global parameters, enabling users to define named values that persist across sessions and are automatically resolved in queries. Users can manage them with `SET GLOBAL PARAMETER`, `UNSET GLOBAL PARAMETER`, and `SHOW PARAMETERS`, and reference them in any Cypher query using the `$param` syntax, client-provided parameters take precedence if both exist. Global parameters are replicated to replicas and recovered on restart, so users can rely on them as a durable, cluster-wide configuration mechanism for frequently used query values. [#3717](https://github.com/memgraph/memgraph/pull/3717)

src/query/interpreter_context.hpp
```diff
@@ -1,4 +1,5 @@
 // Copyright 2026 Memgraph Ltd.
+// Copyright 2026 Memgraph Ltd.
```
DavIvek: delete this

src/query/serialization/typed_value.hpp
```diff
@@ -0,0 +1,50 @@
...
+ * @return Vector of TypedValues
+ */
+std::vector<TypedValue> DeserializeTypedValueList(const nlohmann::json::array_t &data);
+
+/**
+ * @brief Deserialize a map of TypedValues from JSON object.
+ * @param data The JSON object to deserialize
+ * @return Map of string to TypedValue
+ */
+std::map<std::string, TypedValue> DeserializeTypedValueMap(const nlohmann::json::object_t &data);
```
DavIvek: probably don't need this two

src/query/interpreter_context.hpp
```diff
@@ -121,8 +129,8 @@ class InterpreterContextHolder {
   }
 
  private:
-  static void Initialize(InterpreterConfig interpreter_config, utils::Settings *settings,
-                         dbms::DbmsHandler *dbms_handler,
+  static void Initialize(InterpreterConfig interpreter_config, memgraph::utils::Settings *settings,
```
DavIvek: no need for introducing memgraph namespace

src/memgraph.cpp
```diff
@@ -88,7 +89,7 @@ constexpr const char *kMgExperimentalEnabled = "MEMGRAPH_EXPERIMENTAL_ENABLED";
 constexpr const char *kMgBoltPort = "MEMGRAPH_BOLT_PORT";
 constexpr const char *kMgHaClusterInitQueries = "MEMGRAPH_HA_CLUSTER_INIT_QUERIES";
 
-constexpr uint64_t kMgVmMaxMapCount = 262'144;
+constexpr uint64_t kMgVmMaxMapCount = 262144;
```
as51340: let's leave it as it is, someone else will then have problems with formatting

src/memgraph.cpp
```diff
@@ -765,16 +766,11 @@ int main(int argc, char **argv) {
   auto server_endpoint = memgraph::communication::v2::ServerEndpoint{boost::asio::ip::make_address(FLAGS_bolt_address),
                                                                      static_cast<uint16_t>(extracted_bolt_port)};
 #ifdef MG_ENTERPRISE
-  memgraph::glue::Context session_context{.endpoint = server_endpoint,
-                                          .ic = &interpreter_context_,
-                                          .auth = auth_.get(),
-                                          .audit_log = &audit_log,
-                                          .worker_pool_ = worker_pool_ ? &*worker_pool_ : nullptr};
+  memgraph::glue::Context session_context{
```
as51340: I think clang-tidy will fail here because it wants member initialisation in format .ic = ...

src/dbms/dbms_handler.cpp
```diff
@@ -412,6 +416,8 @@ struct CreateDatabase : memgraph::system::ISystemAction {
     // Done during dbms execution
   }
 
+  bool IsEnterpriseOnly() const override { return true; }
```
as51340: this looks fishy

src/auth/auth.cpp
```diff
@@ -163,6 +165,8 @@ struct DropAuthData : memgraph::system::ISystemAction {
 
   void DoDurability() override { /* Done during Auth execution */ }
 
+  bool IsEnterpriseOnly() const override { return true; }
```
as51340: this looks fishy

src/query/frontend/semantic/required_privileges.cpp
```diff
@@ -165,6 +165,8 @@ class PrivilegeExtractor : public QueryVisitor<void>, public HierarchicalTreeVis
 
   void Visit(ShowDatabaseQuery & /*unused*/) override { /* no privilege needed to show current database */ }
 
+  void Visit(ParameterQuery & /*parameter_query*/) override {}  // TODO(@DavIvek): Add new privelege type here
```
as51340: TODO: reminder not sure if you need this for v1

tests/gql_behave/tests/memgraph_V1/features/server_side_parameters.feature
```diff
@@ -0,0 +1,90 @@
...
+            SET GLOBAL PARAMETER x="value"
+            """
+        When executing query:
+            """
+            SHOW PARAMETERS
+            """
+        Then the result should be (ignoring element order for lists):
+            | name | value        | scope   |
+            | 'x'  | '"value"'    | 'global' |
+            | 'y'  | '"my_config"'| 'global' |
```
as51340: I think you have string within string, is this expected?

src/query/frontend/ast/ast.hpp
```diff
@@ -3778,6 +3778,36 @@ class SettingQuery : public memgraph::query::Query {
   friend class AstStorage;
 };
 
+class ParameterQuery : public memgraph::query::Query {
+ public:
+  static const utils::TypeInfo kType;
+
+  const utils::TypeInfo &GetTypeInfo() const override { return kType; }
+
+  enum class Action { SET_PARAMETER, UNSET_PARAMETER, SHOW_PARAMETERS, DELETE_ALL_PARAMETERS };
```
as51340:
```suggestion
  enum class Action : uint8_t { SET_PARAMETER, UNSET_PARAMETER, SHOW_PARAMETERS, DELETE_ALL_PARAMETERS };
```

src/query/frontend/ast/ast.hpp
```diff
@@ -3778,6 +3778,36 @@ class SettingQuery : public memgraph::query::Query {
...
+
+  ParameterQuery() = default;
+
+  DEFVISITABLE(QueryVisitor<void>);
+
+  memgraph::query::ParameterQuery::Action action_;
+  std::string parameter_name_;
+  memgraph::query::Expression *parameter_value_{nullptr};
+
+  // TODO(@DavIvek): Add scope information (GLOBAL, DATABASE, SESSION) when implementing scopes
```
as51340: You need scopes now?

src/query/serialization/typed_value.hpp
```diff
@@ -0,0 +1,33 @@
...
+#include "query/typed_value.hpp"
+
+namespace memgraph::query::serialization {
+
+/**
+ * @brief Serialize a TypedValue to JSON.
+ * @param value The TypedValue to serialize
+ * @return JSON representation of the TypedValue
+ */
+nlohmann::json SerializeTypedValue(const TypedValue &value);
```
as51340: What we usually use it to_json and from_json constructs to serialise and deserialise classes/structs. For example, take a look at coordinator_communication_config.hpp

as51340: Not sure if you already have it but if you don't make sure to add unit tests specifically just for (de)serialisation.

as51340: you can declare to_json and from_json in typed_value files, no need to add another library for this

src/query/serialization/typed_value.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+  if (data.is_object()) {
+    TypedValue::TMap typed_map;
+    for (const auto &item : data.items()) {
+      const auto &key = item.key();
+      const auto &value = item.value();
+      typed_map.emplace(key, DeserializeTypedValue(value));
+    }
+    return TypedValue(std::move(typed_map));
+  }
```
as51340: if you use to_json and from_json, you should be able to use something like:
```
else if (j.is_array()) {
      value = TypedValue{j.get<TypedValue::TVector>()};  // Recursively uses from_json
    } else if (j.is_object()) {
      value = TypedValue{j.get<TypedValue::TMap>()};     // Recursively uses from_json
    } 
```

src/query/serialization/typed_value.cpp
```diff
@@ -0,0 +1,90 @@
...
+        arr.push_back(SerializeTypedValue(item));
+      }
+      return arr;
+    }
+    case TypedValue::Type::Map: {
+      nlohmann::json obj = nlohmann::json::object();
+      for (const auto &[key, val] : value.ValueMap()) {
+        obj.emplace(key, SerializeTypedValue(val));
+      }
+      return obj;
```
as51340: I think if you use to_json, you will be able to use this simpler function:

```
case TypedValue::Type::List:
        j = value.ValueList();  // Recursively uses to_json
        return;
      case TypedValue::Type::Map:
        j = value.ValueMap();   // Recursively uses to_json
```

src/system/include/system/transaction.hpp
```diff
@@ -109,11 +107,14 @@ struct Transaction {
   std::list<std::unique_ptr<ISystemAction>> actions_;
 };
 
-#ifdef MG_ENTERPRISE
 struct DoReplication {
   explicit DoReplication(replication::RoleMainData &main_data) : main_data_{main_data} {}
 
   auto ApplyAction(ISystemAction const &action, Transaction const &system_tx) -> AllSyncReplicaStatus {
+    if (action.IsEnterpriseOnly() && !license::global_license_checker.IsEnterpriseValidFast()) {
```
as51340: you don't need to check whether license is valid here because ApplyAction is called from system/transaction.hpp which is called from interpreter and in the interpreter we already check if we are enterprise and if the license is valid. That means you can also remove the header license/license.hpp

as51340: I think what you want is to modify interpreter.cpp, this part:
```
 auto const main_commit = [&](replication::RoleMainData &mainData) {
    // Only enterprise can do system replication
#ifdef MG_ENTERPRISE
      if (license::global_license_checker.IsEnterpriseValidFast()) {
        return system_transaction_->Commit(memgraph::system::DoReplication{mainData});
      }
#endif
      return system_transaction_->Commit(memgraph::system::DoNothing{});
    };
```

before if it was community and system txn it was DoNothing (I think the correct thing to do was crash the db because it should've never happened) and you want to replicate even if in community because parameters are community feature

src/system/include/system/action.hpp
```diff
@@ -26,13 +26,15 @@ struct ISystemAction {
   /// Durability step which is defered until commit time
   virtual void DoDurability() = 0;
 
-#ifdef MG_ENTERPRISE
+  /// If true, replication is only performed when the enterprise license is valid.
+  /// If false (community feature), replication is always attempted.
+  virtual bool IsEnterpriseOnly() const = 0;
```
as51340: Look at transaction.hpp, I think you don't need this, looks fishy to me

src/replication_handler/include/replication_handler/system_rpc.hpp
```diff
@@ -48,11 +85,12 @@ struct SystemRecoveryReq {
   std::vector<auth::User> users;
   std::vector<auth::Role> roles;
   std::vector<auth::UserProfiles::Profile> profiles;
+  std::vector<utils::ParameterInfo> parameters;
 };
 
 struct SystemRecoveryRes {
   static constexpr utils::TypeInfo kType{.id = utils::TypeId::REP_SYSTEM_RECOVERY_RES, .name = "SystemRecoveryRes"};
-  static constexpr uint64_t kVersion{1};
+  static constexpr uint64_t kVersion{2};
```
as51340, line 114: This is ok but I think you also need to add SystemRecoveryResV1. We don't have a way of checking whether struct changed between versions v1 and v2. You can add a unit test to tests/unit/rpc_versioning to check what will happen if there is a request with 2 versions but a single-versioned response

src/utils/replication_handlers.cpp
```diff
@@ -0,0 +1,161 @@
...
+  if (parameters.DeleteAllParameters()) {
+    res = DeleteAllParametersRes(true);
+  }
+
+  rpc::SendFinalResponse(res, request_version, res_builder);
+}
+
+}  // namespace
+
+bool ApplyParametersRecovery(Parameters &parameters, const std::vector<ParameterInfo> &params) {
```
as51340: What's the purpose of having here a bool if you always return true

src/utils/parameters.hpp
```diff
@@ -0,0 +1,88 @@
...
+#include "kvstore/kvstore.hpp"
+#include "utils/rw_lock.hpp"
+
+namespace memgraph::system {
+struct Transaction;
+}  // namespace memgraph::system
+
+namespace memgraph::utils {
+
+enum class ParameterScope { GLOBAL, DATABASE, SESSION };
```
as51340: uint8_t

src/utils/replication_handlers.cpp
as51340: also, this method is logically part of Parameters class if you ask me, I don't see why this would be a free method

as51340: also now I see that SetParameter returns bool, what do you want to do if Setting the parameter failed (not all passed successfully)

src/replication_handler/system_replication.cpp
```diff
@@ -57,13 +57,49 @@ void SystemRecoveryHandler(memgraph::system::ReplicaHandlerAccessToState &system
...
+  if (current_main_uuid != req.main_uuid) [[unlikely]] {
+    LogWrongMain(current_main_uuid, req.main_uuid, SystemRecoveryReq::kType.name);
+    return;
+  }
+
+  /*
+   * PARAMETERS
+   */
+  if (!utils::ApplyParametersRecovery(parameters, req.parameters)) return;
+
```
as51340: same as above

src/utils/replication_handlers.cpp
as51340: you probably want some kind of RocksDB transaction so it either all passed or nothing, you don't want partial states

src/replication_handler/include/replication_handler/replication_handler.hpp
```diff
@@ -75,28 +81,30 @@ void SystemRestore(ReplicationClient &client, system::System &system, dbms::Dbms
       return std::nullopt;
     });
 
+    bool const is_enterprise = license::global_license_checker.IsEnterpriseValidFast();
```
as51340: why this got moved

src/utils/replication_handlers.cpp
```diff
@@ -0,0 +1,161 @@
...
+  for (const auto &p : params) {
+    parameters.SetParameter(p.name, p.value, p.scope);
+  }
+  return true;
+}
+
+std::vector<ParameterInfo> GetParametersSnapshotForRecovery(Parameters &parameters) {
```
as51340: Don't see why is this a free method and not a member method

src/replication_handler/include/replication_handler/replication_handler.hpp
```diff
@@ -75,28 +81,30 @@ void SystemRestore(ReplicationClient &client, system::System &system, dbms::Dbms
...
     return DbInfo{{dbms_handler.Get()->config().salient}, system.LastCommittedSystemTimestamp()};
   });
   try {
     utils::MetricsTimer const timer{metrics::SystemRecoveryRpc_us};
+    auto const params_snapshot = utils::GetParametersSnapshotForRecovery(parameters);
     auto stream = std::invoke([&]() {
-      // Handle only default database is no license
-      if (!is_enterprise) {
+#ifdef MG_ENTERPRISE
+      if (!license::global_license_checker.IsEnterpriseValidFast()) {
```
as51340: not sure if you need to check license validity here, maybe it's already checked at the caller's side

src/replication_handler/system_rpc.cpp
```diff
@@ -15,10 +15,47 @@
...
+  memgraph::slk::Save(static_cast<uint8_t>(self.scope), builder);
+}
+
+void Load(memgraph::utils::ParameterInfo *self, memgraph::slk::Reader *reader) {
+  memgraph::slk::Load(&self->name, reader);
+  memgraph::slk::Load(&self->value, reader);
+  uint8_t scope = 0;
+  memgraph::slk::Load(&scope, reader);
+  self->scope = static_cast<utils::ParameterScope>(scope);
+}
```
as51340: There is already a specialization on how to (de)serialise enums, check src/slk/serialization.hpp

src/utils/parameters.hpp
```diff
@@ -0,0 +1,88 @@
...
+ * Parameters are dynamic key-value pairs that can be set, retrieved, and deleted.
+ * Unlike Settings, parameters don't need to be pre-registered and can be created on-the-fly.
+ */
+struct Parameters {
+  /**
+   * @brief Construct Parameters with storage path and recovery flag.
+   * @param storage_path Path to the storage directory for parameters
+   * @param recovery_on_startup If false, clears all existing parameters on startup
+   */
+  explicit Parameters(std::filesystem::path storage_path);
```
as51340: accept by const ref since you don't move in cpp file

src/utils/parameters.cpp
```diff
@@ -0,0 +1,176 @@
...
+  bool IsEnterpriseOnly() const override { return false; }
+
+  void PostReplication(replication::RoleMainData & /*main_data*/) const override {}
+};
+
+void AddDeleteAllParametersAction(system::Transaction &txn) { txn.AddAction<DeleteAllParameters>(); }
+
+Parameters::Parameters(std::filesystem::path storage_path) {
+  std::lock_guard parameters_guard{parameters_lock_};
+  if (!FLAGS_data_recovery_on_startup) {
```
as51340: use && instead of nested if

src/utils/parameters.cpp
```diff
@@ -0,0 +1,176 @@
...
+      throw utils::BasicException("Invalid parameter scope");
+  }
+}
+
+bool Parameters::SetParameter(std::string_view name, std::string_view value, ParameterScope scope) {
+  std::lock_guard parameters_guard{parameters_lock_};
+  if (!storage_) return false;
+
+  if (!storage_->Put(name, value)) {
+    SPDLOG_ERROR("Failed to set parameter '{}' with scope '{}'", name, static_cast<int>(scope));
```
as51340: use spdlog::error, you want to log not just in debug mode

src/utils/parameters.cpp
```diff
@@ -0,0 +1,176 @@
...
+    case ParameterScope::SESSION:
+      return "session";
+    default:
+      throw utils::BasicException("Invalid parameter scope");
+  }
+}
+
+bool Parameters::SetParameter(std::string_view name, std::string_view value, ParameterScope scope) {
+  std::lock_guard parameters_guard{parameters_lock_};
+  if (!storage_) return false;
```
as51340: I claim this should be DMG_ASSERT or MG_ASSERT. If you don't have KV store initialised and invoke some operation on it, I would like to see this as an app error not just normal process continuation

as51340: same for other methods

src/utils/parameters.hpp
```diff
@@ -0,0 +1,88 @@
...
+   */
+  std::vector<ParameterInfo> GetAllParameters(ParameterScope scope) const;
+
+  /**
+   * @brief Delete all parameters.
+   */
+  bool DeleteAllParameters();
+
+ private:
+  mutable utils::RWLock parameters_lock_{RWLock::Priority::WRITE};
```
as51340: You don't need lock, rocksdb is atomic. Make concurrency test.

What do you want to achieve with lock, maybe I am misunderstanding?

src/utils/parameters.cpp
```diff
@@ -0,0 +1,176 @@
...
+bool Parameters::DeleteAllParameters() {
+  std::lock_guard parameters_guard{parameters_lock_};
+  if (!storage_) return false;
+
+  for (const auto &[key, _] : *storage_) {
+    if (!storage_->Delete(key)) {
+      return false;
+    }
+  }
+  return true;
```
as51340: Use DeleteMultiple

src/utils/parameters_rpc.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+namespace memgraph::slk {
+
+void Save(const memgraph::storage::replication::SetParameterReq &self, memgraph::slk::Builder *builder) {
+  memgraph::slk::Save(self.main_uuid, builder);
+  memgraph::slk::Save(self.expected_group_timestamp, builder);
+  memgraph::slk::Save(self.new_group_timestamp, builder);
+  memgraph::slk::Save(self.parameter.name, builder);
+  memgraph::slk::Save(self.parameter.value, builder);
+  memgraph::slk::Save(static_cast<uint8_t>(self.parameter.scope), builder);
```
as51340: enum should already be handled

as51340:
Also, I would add more labels for community testing to make sure it's ok before attempting merge

src/query/frontend/opencypher/grammar/MemgraphCypher.g4
```diff
@@ -698,6 +708,18 @@ showSetting : SHOW DATABASE SETTING settingName ;
 
 showSettings : SHOW DATABASE SETTINGS ;
 
+parameterName : symbolicName ;
+
+parameterValue : literal | parameter ;
+
+setParameter : SET GLOBAL PARAMETER parameterName '=' parameterValue ;
+
+unsetParameter : UNSET PARAMETER parameterName ;
```
gitbuda: add the optional [GLOBAL]

src/memgraph.cpp
DavIvek: Weird, it wasn’t building for me because of this.

src/query/frontend/ast/ast.hpp
DavIvek: No, that will be added in second iteration.

src/query/frontend/semantic/required_privileges.cpp
DavIvek: I discussed this with @gitbuda, and since it’s required for the first version, I’ll add it.

src/replication_handler/include/replication_handler/replication_handler.hpp
DavIvek: It’s used on line 85, so there’s no need to evaluate it earlier. I think the code is easier to understand this way.
I’ll inline this call into the `if`, that makes more sense, I assume. Feel free to correct me if I’m misunderstanding something.

DavIvek: Aha no, I see now, I shouldn't move it... It was reused on line 92 but I accidentally deleted that...

src/replication_handler/include/replication_handler/replication_handler.hpp
DavIvek: I didn't add this, just moved it. But looks to me it's needed.

src/utils/parameters.hpp
DavIvek: No need for locking, you’re right. Additionally, all parameter writes will be in unique transactions, so there’s no need for protection around this.

tests/gql_behave/tests/memgraph_V1/features/server_side_parameters.feature
DavIvek: That happens because we convert a typed value to JSON, then convert the JSON to a string and store it in RocksDB. It returns "value". When we reference it later (for example using $x), it won’t be quoted twice.

src/auth/auth.cpp
```diff
@@ -128,6 +128,8 @@ struct UpdateAuthData : memgraph::system::ISystemAction {
 
   void DoDurability() override { /* Done during Auth execution */ }
 
+  bool ShouldReplicateInCommunity() const override { return false; }
```
as51340, line 131: No need for this, enterprise actions will be added only if in enterprise version of the code

src/dbms/dbms_handler.cpp
```diff
@@ -235,6 +235,8 @@ struct DropDatabase : memgraph::system::ISystemAction {
 
   void DoDurability() override { /* Done during DBMS execution */ }
 
+  bool ShouldReplicateInCommunity() const override { return false; }
```
as51340, line 238: No need for this, enterprise actions will be added only if in enterprise version of the code

src/dbms/dbms_handler.cpp
```diff
@@ -257,6 +259,8 @@ struct RenameDatabase : memgraph::system::ISystemAction {
...
```
as51340, line 262: No need for this, enterprise actions will be added only if in enterprise version of the code

src/dbms/dbms_handler.cpp
```diff
@@ -412,6 +416,8 @@ struct CreateDatabase : memgraph::system::ISystemAction {
     // Done during dbms execution
   }
 
+  bool ShouldReplicateInCommunity() const override { return false; }
```
as51340, line 419: No need for this, enterprise actions will be added only if in enterprise version of the code

src/dbms/dbms_handler.hpp
```diff
@@ -42,6 +42,7 @@
 #include "utils/uuid.hpp"
 
 namespace memgraph::dbms {
+// TODO: add to stats
```
as51340: handle todo

src/dbms/dbms_handler.hpp
```diff
@@ -65,6 +66,7 @@ struct Statistics {
   uint64_t property_store_compression_enabled;  //!< Number of databases with property store compression enabled
   std::array<uint64_t, 3>
       property_store_compression_level{};  //!< Number of databases with each compression level [LOW, MID, HIGH]
+  uint64_t parameters;                     //!< Number of server-side parameters
```
as51340: rename to num_parameters

src/dbms/dbms_handler.hpp
as51340: or if handled delete the comment

src/dbms/dbms_handler.hpp
```diff
@@ -98,6 +100,7 @@ static inline nlohmann::json ToJson(const Statistics &stats) {
                                        {"10K-99.9K", stats.label_node_count_histogram[4]},
                                        {"100K-999K", stats.label_node_count_histogram[5]},
                                        {"1M+", stats.label_node_count_histogram[6]}};
+  res["parameters"] = stats.parameters;
```
as51340: num_parameters

src/utils/typeinfo.hpp
```diff
@@ -130,6 +130,12 @@ enum class TypeId : uint64_t {
   REP_DROP_DATABASE_RES,
   REP_RENAME_DATABASE_REQ,
   REP_RENAME_DATABASE_RES,
+  REP_SET_PARAMETER_REQ,
+  REP_SET_PARAMETER_RES,
+  REP_UNSET_PARAMETER_REQ,
+  REP_UNSET_PARAMETER_RES,
+  REP_DELETE_ALL_PARAMETERS_REQ,
+  REP_DELETE_ALL_PARAMETERS_RES,
```
as51340: Add this at the end of the section not in the middle. Could be a problem during no downtime upgrade because previous version would think this is some other rpc message

src/utils/typeinfo.hpp
```diff
@@ -282,6 +288,7 @@ enum class TypeId : uint64_t {
   AST_CREATE_SNAPSHOT_QUERY,
   AST_STREAM_QUERY,
   AST_SETTING_QUERY,
+  AST_PARAMETER_QUERY,
```
as51340: I would also move this to the end

src/system/include/system/transaction.hpp
```diff
@@ -91,6 +88,13 @@ struct Transaction {
     actions_.clear();
   }
 
+  /// True if every action in this transaction is replicable (e.g. only parameter actions). Auth and dbms actions are
+  /// not.
+  [[nodiscard]] bool CanReplicateInCommunity() const {
+    return !actions_.empty() &&
+           std::ranges::all_of(actions_, [](auto const &action) { return action->ShouldReplicateInCommunity(); });
+  }
+
```
as51340: no need for this

src/system/include/system/action.hpp
```diff
@@ -26,13 +26,15 @@ struct ISystemAction {
   /// Durability step which is defered until commit time
   virtual void DoDurability() = 0;
 
-#ifdef MG_ENTERPRISE
+  /// If true, this action may be replicated (e.g. parameter changes). If false, replication is skipped (e.g. auth,
+  /// dbms).
+  virtual bool ShouldReplicateInCommunity() const = 0;
```
as51340, line 31: no need for this

src/parameters/parameters.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+  void PostReplication(replication::RoleMainData &) const override {}
+};
+
+std::string_view ParameterScopeToString(ParameterScope scope) {
+  switch (scope) {
+    case ParameterScope::GLOBAL:
+      return "global";
+  }
+  throw utils::BasicException("Invalid parameter scope");
```
as51340: use default: std::unreachable instead https://en.cppreference.com/w/cpp/utility/unreachable.html

src/parameters/parameters.cpp
```diff
@@ -0,0 +1,186 @@
...
+namespace memgraph::parameters {
+
+namespace {
+
+std::string_view ScopePrefix(ParameterScope scope) {
+  switch (scope) {
+    case ParameterScope::GLOBAL:
+      return "global/";
+  }
+  throw utils::BasicException("Invalid parameter scope");
```
as51340: use default: std::unreachable instead https://en.cppreference.com/w/cpp/utility/unreachable.html

src/parameters/parameters.cpp
```diff
@@ -0,0 +1,186 @@
...
+std::string MakeKey(ParameterScope scope, std::string_view name) {
+  return fmt::format("{}{}", ScopePrefix(scope), name);
+}
+
+std::filesystem::path PrepareStoragePath(const std::filesystem::path &storage_path) {
+  if (!FLAGS_data_recovery_on_startup && utils::DirExists(storage_path)) {
+    utils::DeleteDir(storage_path);
+  }
+  return storage_path;
+}
```
as51340: I don't think this is the best way to do that. Would be better if you can somehow ignore it instead of deleting it.

src/parameters/parameters.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+// --- System actions for replication (defined before Parameters methods that use them) ---
+
+struct SetParameterAction : memgraph::system::ISystemAction {
+  explicit SetParameterAction(std::string_view name, std::string_view value, ParameterScope scope)
+      : name_{name}, value_{value}, scope_{scope} {}
+
+  void DoDurability() override { /* Done during Parameters execution */ }
+
+  bool ShouldReplicateInCommunity() const override { return true; }
```
as51340, line 48: delete

src/parameters/parameters.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+  bool DoReplication(replication::ReplicationClient &client, const utils::UUID &main_uuid,
+                     memgraph::system::Transaction const &txn) const override {
+    auto check = [](const storage::replication::SetParameterRes &res) { return res.success; };
+    return client.StreamAndFinalizeDelta<storage::replication::SetParameterRpc>(
+        check,
+        main_uuid,
+        txn.last_committed_system_timestamp(),
+        txn.timestamp(),
+        ParameterInfo{.name = std::string(name_), .value = std::string(value_), .scope = scope_});
```
as51340: no need to use std::string, it will be copy assigned any way

src/parameters/parameters.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+  void DoDurability() override { /* Done during Parameters execution */ }
+
+  bool ShouldReplicateInCommunity() const override { return true; }
+
+  bool DoReplication(replication::ReplicationClient &client, const utils::UUID &main_uuid,
+                     memgraph::system::Transaction const &txn) const override {
+    auto check = [](const storage::replication::UnsetParameterRes &res) { return res.success; };
+    return client.StreamAndFinalizeDelta<storage::replication::UnsetParameterRpc>(
+        check, main_uuid, txn.last_committed_system_timestamp(), txn.timestamp(), std::string(name_), scope_);
```
as51340: no need to use std::string, it will be copy assigned any way

src/parameters/parameters.cpp
```diff
@@ -0,0 +1,186 @@
...
+  for (auto it = storage_.begin(prefix); it != storage_.end(prefix); ++it) {
+    std::string name = it->first.substr(prefix.size());
+    parameters.emplace_back(ParameterInfo{.name = std::move(name), .value = it->second, .scope = scope});
+  }
+  return parameters;
+}
+
+size_t Parameters::CountParameters(ParameterScope scope) const {
+  return storage_.Size(std::string(ScopePrefix(scope)));
+}
```
as51340: unused delete

src/parameters/parameters.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+std::vector<ParameterInfo> Parameters::GetSnapshotForRecovery() const {
+  std::vector<ParameterInfo> out;
+  for (auto scope : {ParameterScope::GLOBAL}) {
+    for (const auto &p : GetAllParameters(scope)) {
+      out.push_back(p);
+    }
+  }
+  return out;
```
as51340, line 173: Better version  std::vector<ParameterInfo> Parameters::GetSnapshotForRecovery() const {
    std::vector<ParameterInfo> out;
    for (auto scope : {ParameterScope::GLOBAL}) {
      auto params = GetAllParameters(scope);
      out.insert(out.end(), std::make_move_iterator(params.begin()), std::make_move_iterator(params.end()));
    }
    return out;
  }

src/parameters/replication_handlers.cpp
```diff
@@ -0,0 +1,153 @@
...
+                                                auto *res_builder) mutable {
+        DeleteAllParametersHandler(
+            system_state_access, data.uuid_, parameters, request_version, req_reader, res_builder);
+      });
+}
+
+void Parameters::RegisterReplicationHandlers(replication::RoleReplicaData const &data,
+                                             system::ReplicaHandlerAccessToState &system_state_access) {
+  Register(data, system_state_access, *this);
+}
```
as51340: Unused I think

src/parameters/parameters.hpp
```diff
@@ -0,0 +1,102 @@
...
+   * @brief Return snapshot of all parameters for SystemRecoveryReq (main side).
+   */
+  std::vector<ParameterInfo> GetSnapshotForRecovery() const;
+
+  /**
+   * @brief Register replication RPC handlers for parameters on the replica.
+   */
+  void RegisterReplicationHandlers(replication::RoleReplicaData const &data,
+                                   system::ReplicaHandlerAccessToState &system_state_access);
+
```
as51340: Unused I think

src/query/frontend/opencypher/grammar/MemgraphCypher.g4
```diff
@@ -139,6 +139,7 @@ memgraphCypherKeyword : cypherKeyword
                       | ON_DISK_TRANSACTIONAL
                       | PARALLEL
                       | PARALLEL_EXECUTION
+                      | SERVER_SIDE_PARAMETERS
```
as51340: use alphabetical order

src/query/frontend/opencypher/grammar/MemgraphCypher.g4
```diff
@@ -172,6 +173,8 @@ memgraphCypherKeyword : cypherKeyword
                       | SESSION
                       | SETTING
                       | SETTINGS
+                      | PARAMETER
+                      | PARAMETERS
```
as51340: use alphabetical order

src/query/interpret/eval.hpp
```diff
@@ -127,6 +127,23 @@ class PrimitiveLiteralExpressionEvaluator : public ExpressionVisitor<TypedValue>
...
+    return {std::move(result), ctx_->memory};
+  }
+
+  TypedValue Visit(MapLiteral &map_literal) override {
+    TypedValue::TMap result(ctx_->memory);
+    for (const auto &pair : map_literal.elements_) {
+      result.emplace(TypedValue::TString(pair.first.name, ctx_->memory), pair.second->Accept(*this));
+    }
+    return {std::move(result), ctx_->memory};
+  }
```
as51340, line 146: why do you need this?

src/query/interpreter.cpp
```diff
@@ -2585,6 +2600,99 @@ Callback HandleSettingQuery(SettingQuery *setting_query, const Parameters &param
...
+
+      callback.fn = [parameter_name = parameter_query->parameter_name_,
+                     value_str,
+                     parameters = interpreter_context->parameters,
+                     interpreter]() {
+        if (!parameters) {
+          throw QueryRuntimeException("Parameters are not available");
+        }
+        MG_ASSERT(interpreter->system_transaction_, "System transaction is not available");
+        if (!parameters->SetParameter(parameter_name, value_str, kParamScope, &*interpreter->system_transaction_)) {
```
as51340, line 2788: Why assert if system txn can be nullptr (non-replication scenario)

src/query/interpreter.cpp
```diff
@@ -2585,6 +2600,99 @@ Callback HandleSettingQuery(SettingQuery *setting_query, const Parameters &param
...
+    case ParameterQuery::Action::UNSET_PARAMETER: {
+      callback.fn = [parameter_name = parameter_query->parameter_name_,
+                     parameters = interpreter_context->parameters,
+                     interpreter]() {
+        if (!parameters) {
+          throw QueryRuntimeException("Parameters are not available");
+        }
+        MG_ASSERT(interpreter->system_transaction_, "System transaction is not available");
+        if (!parameters->UnsetParameter(parameter_name, kParamScope, &*interpreter->system_transaction_)) {
+          throw utils::BasicException("Parameter '{}' does not exist", parameter_name);
```
as51340: Why assert if system txn can be nullptr (non-replication scenario)

src/query/interpreter.cpp
```diff
@@ -2585,6 +2600,99 @@ Callback HandleSettingQuery(SettingQuery *setting_query, const Parameters &param
...
+      callback.fn = [parameter_name = parameter_query->parameter_name_,
+                     value_str,
+                     parameters = interpreter_context->parameters,
+                     interpreter]() {
+        if (!parameters) {
+          throw QueryRuntimeException("Parameters are not available");
+        }
+        MG_ASSERT(interpreter->system_transaction_, "System transaction is not available");
+        if (!parameters->SetParameter(parameter_name, value_str, kParamScope, &*interpreter->system_transaction_)) {
+          throw utils::BasicException("Failed to set parameter '{}'", parameter_name);
```
as51340: throw queryruntimaexception

src/query/interpreter.cpp
as51340: throw queryruntimaexception

src/query/interpreter.cpp
```diff
@@ -2585,6 +2600,99 @@ Callback HandleSettingQuery(SettingQuery *setting_query, const Parameters &param
...
+        return results;
+      };
+      return callback;
+    }
+    case ParameterQuery::Action::DELETE_ALL_PARAMETERS: {
+      callback.fn = [parameters = interpreter_context->parameters, interpreter]() {
+        if (!parameters) {
+          throw QueryRuntimeException("Parameters are not available");
+        }
+        MG_ASSERT(interpreter->system_transaction_, "System transaction is not available");
```
as51340, line 2835: throw queryruntimaexception and why assert?

src/query/interpreter.cpp
```diff
@@ -8692,12 +8843,10 @@ void Interpreter::Commit() {
     });
 
     auto const main_commit = [&](replication::RoleMainData &mainData) {
-    // Only enterprise can do system replication
-#ifdef MG_ENTERPRISE
-      if (license::global_license_checker.IsEnterpriseValidFast()) {
+      // With valid enterprise license: replicate all. Without license: only parameter-only transactions replicate.
+      if (license::global_license_checker.IsEnterpriseValidFast() || system_transaction_->CanReplicateInCommunity()) {
```
as51340, line 9025: I don't think you need canreplicateincommunity already handled with ifdef MG_ENTERPRISE

src/replication_handler/include/replication_handler/system_rpc.hpp
```diff
@@ -62,16 +121,25 @@ struct SystemRecoveryRes {
...
 
+  SystemRecoveryResV1 Downgrade() const {
+    return SystemRecoveryResV1{static_cast<SystemRecoveryResV1::Result>(result)};
+  }
+
   Result result;
 };
 
 using SystemRecoveryRpc = rpc::RequestResponse<SystemRecoveryReq, SystemRecoveryRes>;
+using SystemRecoveryRpcV1 = rpc::RequestResponse<SystemRecoveryReqV1, SystemRecoveryResV1>;
```
as51340: not needed

src/replication_handler/system_rpc.cpp
```diff
@@ -37,6 +74,20 @@ void Load(memgraph::replication::SystemRecoveryReq *self, memgraph::slk::Reader
   memgraph::slk::Load(&self->users, reader);
   memgraph::slk::Load(&self->roles, reader);
   memgraph::slk::Load(&self->profiles, reader);
+  memgraph::slk::Load(&self->parameters, reader);
+}
+
+// Serialize code for SystemRecoveryResV1 (same layout as Res)
+void Save(const memgraph::replication::SystemRecoveryResV1 &self, memgraph::slk::Builder *builder) {
+  memgraph::slk::Save(std::to_underlying(self.result), builder);
```
as51340: do you need to_underlying, we already have impl for enum serialisation?

tests/e2e/system_replication/replicate_parameters.py
```diff
@@ -0,0 +1,164 @@
...
+    mg_sleep_and_assert_collection(_expected_replicas_ts(3), show_replicas_func(cursor))
+
+    for port in (BOLT_PORTS["replica_1"], BOLT_PORTS["replica_2"]):
+        repl_cursor = connection(port, "replica").cursor()
+        rows = _show_parameters(repl_cursor)
+        assert len(rows) == 1
+        assert rows[0][0] == "x"
+
+
+def test_unset_parameter_replication(connection, test_name, clean_dirs):
```
as51340, line 199: add instance restarts in replication test

tests/unit/rpc_messages.hpp
```diff
@@ -116,3 +116,19 @@ void Load(EchoMessage *echo, Reader *reader);
 }  // namespace memgraph::slk
 
 using Echo = memgraph::rpc::RequestResponse<EchoMessage, EchoMessage>;
+
+// Used by rpc_versioning test: response type with single version (no Downgrade).
+namespace rpc_versioning_test {
```
as51340: No need for new namespace

as51340 (changes requested):
Much better 💪

src/parameters/parameters.cpp
DavIvek: Why would that be better, then user will have stale rocksdb files?

src/query/interpret/eval.hpp
DavIvek: We want to be able to save maps and lists as parameters. Currently, we can’t use `ExpressionEvaluator` because at the moment a parameter is created, we don’t have a `Frame`, `SymbolTable`, or `View`, and we don’t actually need them. I believe a better approach is to extend `PrimitiveExpressionEvaluator` with support for evaluating maps and lists, since it already has everything needed to handle them.

src/parameters/parameters.cpp
DavIvek: Should be used in stats, will fix

src/query/interpreter.cpp
DavIvek: I thought that we always create system transaction.

DavIvek: I think it is always created (non-replication also):
```
// System queries require strict ordering; since there is no MVCC-like thing, we allow single queries
    bool system_queries =
        utils::Downcast<AuthQuery>(parsed_query.query) || utils::Downcast<MultiDatabaseQuery>(parsed_query.query) ||
        utils::Downcast<ReplicationQuery>(parsed_query.query) ||
        utils::Downcast<UserProfileQuery>(parsed_query.query) || utils::Downcast<ParameterQuery>(parsed_query.query);
```

src/parameters/parameters.cpp
DavIvek: As discussed on the call, we should always restore.

as51340 (approved):

PR #3718, opened 2026-01-27
feat: Improve snapshots' and WALs' retention policy

as51340:
GetSnapshotFiles and GetWalFiles will now crash the database only on the failure when the instance restarts. WALs with wrong storage UUID will get deleted. Snapshots with wrong storage UUID will now be treated same as any other corrupted snapshot, only last N --storage-snapshot-retention-count will get preserved.

### DIFF
--- a/src/storage/v2/durability/snapshot.cpp
+++ b/src/storage/v2/durability/snapshot.cpp
@@ -10956,7 +10956,7 @@ std::optional<std::filesystem::path> CreateSnapshot(Storage *storage, Transactio
   spdlog::info("Snapshot creation successful!");
 
   auto const old_snapshot_files =
-      EnsureRetentionCountSnapshotsExist(snapshot_directory, path, uuid_str, file_retainer, storage);
+      EnsureRetentionCountSnapshotsExist(snapshot_directory, uuid_str, path, file_retainer, storage);
 
   // -1 needed because we skip the current snapshot
   if (old_snapshot_files.size() == storage->config_.durability.snapshot_retention_count - 1 &&


src/storage/v2/durability/snapshot.cpp
```diff
@@ -10086,7 +10083,6 @@ void EnsureNecessaryWalFilesExist(const std::filesystem::path &wal_directory, co
     if (!item.is_regular_file()) continue;
     try {
       auto info = ReadWalInfo(item.path());
-      if (info.uuid != uuid) continue;
```
andrejtonev, line 10089: I think we can just delete these.

src/storage/v2/durability/snapshot.cpp
```diff
@@ -10127,22 +10123,20 @@ void EnsureNecessaryWalFilesExist(const std::filesystem::path &wal_directory, co
...
+                                        const std::string &current_snapshot_path, utils::FileRetainer *file_retainer,
+                                        Storage *storage) -> OldSnapshotFiles {
   OldSnapshotFiles old_snapshot_files;
   std::error_code error_code;
   for (const auto &item : std::filesystem::directory_iterator(snapshot_directory, error_code)) {
     if (!item.is_regular_file()) continue;
     if (item.path() == current_snapshot_path) continue;
     try {
       auto info = ReadSnapshotInfo(item.path());
-      if (info.uuid != uuid) continue;
```
andrejtonev, line 10139: we can just put its timestamp to 0.
threat it as if it was corrupted

src/storage/v2/durability/snapshot.cpp
andrejtonev: If we want to be safe, we could move them to .old at instance startup.

as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - WALs with a wrong UUID will be deleted. Corrupted/incomplete snapshots won't be preserved anymore but will rather be treated as any other snapshot. The policy is to keep the latest --storage-snapshot-retention-count snapshots in the system independently from the fact whether such snapshots are valid or invalid. [#3718](https://github.com/memgraph/memgraph/pull/3718)

PR #3723, opened 2026-01-27
Add cuGraph builld to CI

mattkjames7:
- Added the cuGraph MAGE image build to CI - can be built manually, part of daily build and RC build.
- Added a new mgbuild image to build MAGE modules with `memgraph/mgbuild:v7_ubuntu-24.04-cugraph`.
- Updated *all* MAGE images to use `torch==2.9.0`.
- The script for installing Python dependencies for MAGE will now pull `torch_*`  packages from S3 because official binaries that work with our particular chosen versions of `torch` and CUDA do not exist yet.
- Updated expected outcomes of some cuGraph e2e tests - the behaviour seems to have changed a little again, since the change in the build (this also happened [here](https://github.com/memgraph/mage/pull/710)). Perhaps we need to devise a more reliable way of testing the cuGraph module.
- Updated script for building custom MAGE images to be able to build the cuGraph image.

NOTE: the cuGraph MAGE image cannot currently be tested in CI, as it requires the existence of an NVIDIA GPU - so testing was done locally.

Requirements for running this new image:
- NVIDIA GPU _after but not including Pascal_ (e.g. RTX)
- Drivers >= 590
- CUDA toolkit >= 13.1

Related Infra PR here: https://github.com/memgraph/infra/pull/443

### DIFF
--- a/.github/workflows/daily_build.yml
+++ b/.github/workflows/daily_build.yml
@@ -94,7 +94,7 @@ jobs:
     secrets: inherit
 
   DailyPackageArtifact:
-    if: ${{ (github.event_name == 'schedule' && (success() || failure())) || (github.event_name == 'workflow_dispatch' && !inputs.skip_package)  }}
+    if: ${{ (github.event_name == 'schedule' && always()) || (github.event_name == 'workflow_dispatch' && !inputs.skip_package)  }}
     needs: [CurrentDateJob]
     strategy:
       fail-fast: false
@@ -154,7 +154,7 @@ jobs:
     secrets: inherit
 
   DailyDockerArtifact:
-    if: ${{ (github.event_name == 'schedule' && (success() || failure())) || (github.event_name == 'workflow_dispatch' && !inputs.skip_docker_artifact)  }}
+    if: ${{ (github.event_name == 'schedule' && always()) || (github.event_name == 'workflow_dispatch' && !inputs.skip_docker_artifact)  }}
     needs: [CurrentDateJob]
     strategy:
       fail-fast: false
@@ -177,7 +177,7 @@ jobs:
     secrets: inherit
 
   AggregateBuildTests:
-    if: ${{ (github.event_name == 'schedule' || (github.event_name == 'workflow_dispatch' && !inputs.skip_tests && !inputs.mock)) && (success() || failure()) }}
+    if: ${{ (github.event_name == 'schedule' || (github.event_name == 'workflow_dispatch' && !inputs.skip_tests && !inputs.mock)) && always() }}
     needs: [CurrentDateJob,TestBuild,DailyPackageArtifact,DailyDockerArtifact]
     runs-on: [self-hosted]
     steps:
@@ -216,7 +216,7 @@ jobs:
             -d "$payload"
 
   BuildAndTestMAGE:
-    if: ${{ (github.event_name == 'schedule' && (success() || failure())) || (github.event_name == 'workflow_dispatch' && !inputs.skip_mage) }}
+    if: ${{ (github.event_name == 'schedule' && always()) || (github.event_name == 'workflow_dispatch' && !inputs.skip_mage) }}
     name: Build and Test MAGE
     needs: [CurrentDateJob]
     uses: ./.github/workflows/reusable_package_mage.yaml
@@ -269,7 +269,7 @@ jobs:
     secrets: inherit
 
   AggregateMAGEBuildTests:
-    if: ${{ success() || failure() }}
+    if: ${{ always() }}
     needs: [CurrentDateJob,BuildAndTestMAGE]
     runs-on: [self-hosted]
     steps:
@@ -311,7 +311,7 @@ jobs:
   SetupDailyBuildImageOutputs:
     name: Setup Daily Build Image Outputs
     needs: [CurrentDateJob, DailyDockerArtifact]
-    if: ${{ (github.event_name == 'schedule' && (success() || failure())) || (github.event_name == 'workflow_dispatch' && !inputs.skip_issu_tests || !inputs.skip_stress_tests) }}
+    if: ${{ (github.event_name == 'schedule' && always()) || (github.event_name == 'workflow_dispatch' && !inputs.skip_issu_tests || !inputs.skip_stress_tests) }}
     runs-on: ubuntu-24.04
     outputs:
       amd_image_url: ${{ steps.set_image_urls.outputs.amd_image_url }}
@@ -353,7 +353,7 @@ jobs:
   RunISSUTest:
     name: Run ISSU Tests
     needs: [SetupDailyBuildImageOutputs]
-    if: ${{ (github.event_name == 'schedule' && (success() || failure()) || (github.event_name == 'workflow_dispatch' && !inputs.skip_issu_tests)) && needs.SetupDailyBuildImageOutputs.result == 'success' }}
+    if: ${{ (github.event_name == 'schedule' && always() || (github.event_name == 'workflow_dispatch' && !inputs.skip_issu_tests)) && needs.SetupDailyBuildImageOutputs.result == 'success' }}
     uses: ./.github/workflows/reusable_issu_test.yaml
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


mattkjames7:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - New cuGraph images for MAGE to be built regularly. Users may try the new image in order to use the `cugraph` module by pulling `memgraph/memgraph-mage:<version>-relwithdebinfo-cugraph` and then running with access to their hosts GPU, e.g. `docker run --gpus all -p 7687:7687 --name mage-cugraph memgraph/memgraph-mage:<version>-relwithdebinfo-cugraph`. Requires an NVIDIA GPU supported by driver version >=590 and CUDA toolkit >= 13.1 [#3723](https://github.com/memgraph/memgraph/pull/3723).
- [x] https://github.com/memgraph/documentation/pull/1522

.github/workflows/daily_build.yml
```diff
@@ -94,7 +94,7 @@ jobs:
     secrets: inherit
 
   DailyPackageArtifact:
-    if: ${{ (github.event_name == 'schedule' && always()) || (github.event_name == 'workflow_dispatch' && !inputs.skip_package)  }}
+    if: ${{ (github.event_name == 'schedule' && always() && !cancelled()) || (github.event_name == 'workflow_dispatch' && !inputs.skip_package)  }}
```
imilinovic: why use !cancelled() ?
From what I see its main use is to do cleanup jobs if cancelled()
otherwise actions should always behave as if !cancelled() ?

.github/workflows/daily_build.yml
```diff
@@ -227,38 +227,49 @@ jobs:
...
     with:
       arch: ${{ matrix.build_arch }}
       build_type: ${{ matrix.build_type }}
       s3_dest_dir: "daily-build/mage${{ inputs.mock && '_mock' || '' }}/${{ needs.CurrentDateJob.outputs.current_date }}"
       cuda: ${{ matrix.cuda }}
+      cugraph: ${{ matrix.cugraph }}
       push_to_s3: true
       malloc: ${{ matrix.malloc }}
-      run_smoke_tests: ${{ !inputs.skip_smoke_tests }}
+      run_smoke_tests: ${{ (matrix.cugraph == false && true || false) && !inputs.skip_smoke_tests }}
```
imilinovic: `run_smoke_tests: ${{ !matrix.cugraph && !inputs.skip_smoke_tests }}`
this should be same?

.github/workflows/daily_build.yml
```diff
@@ -300,7 +311,7 @@ jobs:
   SetupDailyBuildImageOutputs:
     name: Setup Daily Build Image Outputs
     needs: [CurrentDateJob, DailyDockerArtifact]
-    if: ${{ (github.event_name == 'schedule' && always()) || (github.event_name == 'workflow_dispatch' && !inputs.skip_issu_tests || !inputs.skip_stress_tests) }}
+    if: ${{ (github.event_name == 'schedule' && always() && !cancelled()) || (github.event_name == 'workflow_dispatch' && !inputs.skip_issu_tests || !inputs.skip_stress_tests) }}
```
imilinovic: always() should always evaluate to true, also same comment as above for !cancelled()

so I think you can remove both?

.github/workflows/reusable_package_mage.yaml
```diff
@@ -115,7 +120,7 @@ jobs:
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
       - name: Pull e2e Docker images in background
-        if: ${{ inputs.run_tests == true }}
+        if: ${{ inputs.run_tests == true && inputs.cugraph == false }}
```
imilinovic, line 123: `inputs.cugraph == false` not needed because you embedded that into run_tests,

imilinovic: `   run_tests: ${{ matrix.run_tests == 'true' && matrix.cugraph == 'false' }}`

.github/workflows/reusable_package_mage.yaml
```diff
@@ -143,10 +148,18 @@ jobs:
 
       - name: Check mgdeps-cache availability
         run: |
-          if [ "$ARCH" == "amd" ]; then
-            echo "mgdeps-cache is temporary disabled for amd64"
-            echo "CACHE_PRESENT=false" >> $GITHUB_ENV
-          elif curl --silent --fail "http://mgdeps-cache:8000/wheels/" -o /dev/null; then
+          if [[ "$CUGRAPH" == true  || "$CUDA" == true ]]; then
+            echo "Using wheels build for CUDA"
+            WHEEL_PATH="cuda-13.0"
```
imilinovic: does this work for both arm and amd?

.github/workflows/reusable_package_mage.yaml
```diff
@@ -431,7 +461,7 @@ jobs:
           path: "sbom/"
 
       - name: Start container for e2e tests
-        if: ${{ inputs.run_tests == true }}
+        if: ${{ inputs.run_tests == true && inputs.cugraph == false }}
```
imilinovic, line 460: same as before, holds for below as well

tools/ci/package_mage_setup.py
```diff
@@ -3,45 +3,58 @@
...
+        "cugraph": False,
         "malloc": True,
     },
+    {
+        "arch": "amd",
+        "build_type": "Release",
+        "cuda": False,
+        "cugraph": True,
+        "malloc": False,
+    },
```
imilinovic, line 57: aha cugraph is amd only, thats answers question from above, I suppose its hard to support on arm?

.github/workflows/daily_build.yml
mattkjames7: When it's just `always()` it will keep going, even if I attempt to cancel the worflow run - I knew this could be an issue, and it was a pain last night while I was testing the daily build :facepalm:. With `always() && !cancelled()` it means that the workflow will always run, _unless_ it's cancelled - or at least, thats what the docs led me to believe...

.github/workflows/daily_build.yml
mattkjames7: I must have fell asleep here, you're correct!

.github/workflows/daily_build.yml
mattkjames7: I think that GitHub reevaluates the `if` when the job gets cancelled, so without the `&& !cancelled()` this job still runs, if even if manually cancelled.

.github/workflows/reusable_package_mage.yaml
mattkjames7: I made a mistake in the call to this workflow from `package_mage.yml` - I actually want `run_tests: true` in most cases, even with the cuGraph build, because we are still able to run unit tests inside the build container, just not e2e tests. I'll push up a fix for this.

.github/workflows/reusable_package_mage.yaml
mattkjames7: I think that installing CUDA/cuGraph on ARM is totally possible - we have no way to test it though :joy:, so I only build wheels for AMD.
As with the existing CUDA image for MAGE - we only currently build for AMD. If the demand is there though, I'm sure we could change it to support CUDA.
It might be worth me changing the path in the cache to support this eventuality - I will push a change for this so it uses `cuda-13.0/amd64`.

.github/workflows/reusable_package_mage.yaml
mattkjames7: We can't currently run e2e in CI for this image, but can run unit tests.

tools/ci/package_mage_setup.py
mattkjames7: As mentioned above - we have no way of testing this yet, and the existing CUDA image is also AMD-only for now.

.github/workflows/daily_build.yml
mattkjames7: So it turns out that this also does not work. It's not a big problem as far as I am concerned, I have a script which I can use to force-cancel if needed. I'll revert to the original behaviour for now.

PR #3727, opened 2026-01-30
fix: Add self coord instance shouldn't call std::terminate

as51340:
Adding a coordinator instance could fail. When adding a coordinator instance on which the query is actually run (self), there is no need to call std::terminate in the case of the failure.


as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
   - When adding itself as a coordinator, there is no need to call std::terminate on the failure, throwing an exception is enough because the rest of the coordinator state is valid. [#3727](https://github.com/memgraph/memgraph/pull/3727)

PR #3574, opened 2025-12-18
Memgraph + MAGE Monorepo

mattkjames7:
Merge [MAGE](https://github.com/memgraph/mage) into this repo.

Changes:
- [x] Move `cpp/`, `python/` and `rust/` from MAGE into `mage/` subdirectory
- [x] Remove `cpp/memgraph` submodule
- [x] Deduplicate modules which exist in both repos -> https://github.com/memgraph/memgraph/pull/3623
- [x] Add MAGE build to diff workflow (AMD and ARM and CUDA tests added to diff)
- [x] Add git labels for testing MAGE on push to pull request
- [x] Add git labels for packaging MAGE on push to pull request
- [x] Remove repository dispatch triggers from RC, Daily build and PR build workflows that would have triggered MAGE build; replace with calls to new workflows
- [x] Add MAGE to daily build
- [x] Add MAGE to RC build + add more inputs for testing purposes
- [x] Add MAGE to promote RC workflow + more inputs for testing
- [x] Update PR build workflow
- [x] Move smoke tests
- [x] Move CVE scans
- [x] Add MAGE to flakiness workflow
- [x] Add MAGE modules to clang-tidy check
- [x] Cleanup unused workflows (`promote_daily_rc.yml` was never needed; `release_build_test.yaml` has been replaced with `build_rc.yml`)
- [x] Add external package mage workflow
- [x] Add setup script to package MAGE workflow
- [x] Move `e2e/`, `e2e_correctness/`, `e2e_migration/`, `smoke-release-testing/` test directories and related scripts under `mage/tests` subdirectory
- [x] Add MAGE tests to mgbuild.sh
  - [x] Unit tests
  - [x] E2E
  - [x] Correctness
  - [x] Migration
  - [x] Smoke
- [x] Move `Dockerfile.release`, `setup` script and `make-dev-container` script into `mage/` directory (we will build container images from here)
- [x] Build modules in `mgbuild.sh`
- [x] Package docker container in `mgbuild.sh`
- [x] Package DEB in `mgbuild.sh`
- [x] Build `pymgclient` for arm64 in `mgbuild.sh`
- [x] Handle duplicate code for building `heaptrack`
- [x] Deduplicate code for SBOM generation
- [x] Add custom OpenSSL build to MAGE containers
- [x] Copy README from MAGE repo under `mage/` subdirectory
  - [x] Check all references to items int he repo
  - [x] Check main README for references to MAGE
- [x] Copy third-party license from mage to the `licenses` directory
[...]

PR #3742, opened 2026-02-02
Packaging fixes and improvements

mattkjames7:
- Fix `[2026-02-10 11:02:43.406] [memgraph_log] [error] Unable to load module /usr/lib/memgraph/query_modules/libmgclient.so; /usr/lib/memgraph/query_modules/libmgclient.so: undefined symbol: mgp_init_module. For more details, visit https://memgr.ph/modules.`
- Fix incorrect query module count for the Memgraph Docker image: some files were being copied from MAGE when they shouldn't have been. NOTE: the [monorepo PR](https://github.com/memgraph/memgraph/pull/3574) added three new query modules when merging/deduplicating code: 
    - `date.format`
    - `date.parse`
    - `nxalg.weakly_connected_components_subgraph`
- Added smoke test to check for `[error]` within the logs to stop the previous two issues slipping into our builds unnoticed.
- Fix broken query module naming: importing `weakly_connected_components_subgraph` from `nxalg.py` within `wcc.py` caused all functions from `nxalg` to be renamed, e.g. `nxalg.pagerank` -> `wcc.pagerank`.
- Updated some Python packages to address some CVEs found in older versions.
For Memgraph (x86_64):
```
|      This PR        |    |        Daily        |
| Severity   | Counts |    | Severity   | Counts |
|------------|--------|    |------------|--------|
| UNKNOWN    |      0 |    | UNKNOWN    |      0 |
| NEGLIGIBLE |      0 |    | NEGLIGIBLE |      0 |
| LOW        |     22 |    | LOW        |     24 |
| MEDIUM     |     29 |    | MEDIUM     |     46 |
| HIGH       |      0 |    | HIGH       |      4 |
| CRITICAL   |      0 |    | CRITICAL   |      0 |
```
and MAGE (x86_64):
```
|      This PR        |    |        Daily        |
| Severity   | Counts |    | Severity   | Counts |
|------------|--------|    |------------|--------|
| UNKNOWN    |      0 |    | UNKNOWN    |      0 |
| NEGLIGIBLE |      0 |    | NEGLIGIBLE |      0 |
| LOW        |     26 |    | LOW        |     28 |
| MEDIUM     |     28 |    | MEDIUM     |     41 |
| HIGH       |      0 |    | HIGH       |      4 |
| CRITICAL   |      0 |    | CRITICAL   |      0 |
```
- Moved Python package installation into `python-base` stage of Dockerfiles so that `wheel` and `setuptools` provided by Ubuntu/APT don't end up in the final image.
- Moved `setuptools` out of auth module requirements - it appears that this isn't needed here, but it is needed by MAGE because `torchmetrics` depends on the deprecated `pkg_resources` module that it provides.
- Improved Dockerfiles by replacing `COPY` with bind mounts to reduce final image size. 
```
|      Image        | Daily Build |  This PR   |
|-------------------|-------------|------------|
| Memgraph (x86_64) |      262 MB |     208 MB |
| MAGE (x86_64)     |      651 MB |     697 MB |
```


mattkjames7:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - N/A

mage/python/requirements.txt
```diff
@@ -3,12 +3,13 @@ boto3==1.36.23
...
 duckdb==1.2.1
 elasticsearch==8.17.0
-filelock==3.19.1
+filelock==3.20.3
 gekko==1.2.1
 gensim==4.3.3
 gqlalchemy==1.8.0
 huggingface-hub==0.35.1
 igraph==0.11.8
+jaraco-context==6.1.0
```
imilinovic, line 12: why do we need this? don't see it used anywhere and it wasn't in mage before

mage/tests/smoke-release-testing/mgconsole/query_modules.bash
```diff
@@ -7,7 +7,7 @@ test_query_modules() {
   if [ "$IMAGE_TYPE" == "mage" ]; then
     expected_module_count=322
   elif [ "$IMAGE_TYPE" == "memgraph" ]; then
-    expected_module_count=142
+    expected_module_count=138
```
imilinovic: not sure what this is counting, in the description it says 3 were added, but here its -4
date.format
date.parse
nxalg.weakly_connected_components_subgraph

also just to be precise date is a query module, date.format is a function/procedure
so date.format and date.parse would fall under same query module

query_modules/wcc.py
```diff
@@ -48,4 +45,6 @@ def get_components(
     procedure name for better compatibility with NetworkX algorithms.
     """
     # Call the implementation from nxalg
+    from nxalg import weakly_connected_components_subgraph
+
```
imilinovic, line 49: do you have an explanation why this caused the problem? I think the procedure names are generated in format `filename.procedure_name` so in this case should be `wcc.get_components`, moving the import shouldn't impact this?

mage/python/requirements.txt
mattkjames7: It's from a dependency (I don't remember which) - but an older version gets installed by default  and has a high severity CVE. At some point we should probably have a lockfile or something instead of relying on pip alone.

query_modules/wcc.py
colinbarry: By moving it to function scope, we import nxalg at call time, not at module load time. I assume this ensures that the module's symbols are registered under the correct module name/scope?

mage/tests/smoke-release-testing/mgconsole/query_modules.bash
mattkjames7: So `3.7.2` had 135, this test was only added a few days ago, after the monorepo merge - the 142 included some functions/procedures that were mistakenly copied from mage; without them it should be 138.

.github/workflows/package_memgraph.yaml
```diff
@@ -48,7 +48,7 @@ on:
       s3_dest_dir:
         type: string
         description: "Target dir path in chosen S3 bucket, don't end the path with '/'. For bucket 'deps.memgraph.io' use format 'memgraph/vX.Y.Z'."
-        default: ''
+        default: 'memgraph-unofficial'
```
colinbarry, line 51: Just to check this is expected behaviour: I notice we have different s3 default dirs for `workflow_call` and `workflow_dispatch`.

query_modules/wcc.py
mattkjames7: I'm not sure exactly, but I noticed that wcc.py was loaded before nxalg.py in the logs - which meant that the functions in nxalg.py would have been interpreted at the point where wcc.py imported them.

.github/workflows/package_memgraph.yaml
mattkjames7: Yep - I only wanted to change this for manual runs of the workflow directly

query_modules/wcc.py
mattkjames7: Ahh your comment wasn't there when I wrote my response 😂 - yes Colin is correct, and the module's symbols have the correct names again.

mage/tests/smoke-release-testing/mgconsole/query_modules.bash
mattkjames7: I have also renamed this...

imilinovic (approved):

PR #3746, opened 2026-02-03
fix: Perf regression on all-shortest paths

as51340:
throw+catch is too expensive, use direct null typed value instead.

### DIFF
--- a/src/query/interpret/eval.hpp
+++ b/src/query/interpret/eval.hpp
@@ -232,17 +232,6 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
     }                                                                                                          \
   }
 
-#define BINARY_COMPARISON_OPERATOR_VISITOR(OP_NODE, CPP_OP, CYPHER_OP)                                         \
-  TypedValue Visit(OP_NODE &op) override {                                                                     \
-    auto val1 = op.expression1_->Accept(*this);                                                                \
-    auto val2 = op.expression2_->Accept(*this);                                                                \
-    try {                                                                                                      \
-      return val1 CPP_OP val2;                                                                                 \
-    } catch (const TypedValueException &) {                                                                    \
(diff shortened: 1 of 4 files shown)


src/query/interpret/eval.hpp
```diff
@@ -262,12 +260,12 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
...
-  BINARY_COMPARISON_OPERATOR_VISITOR(LessOperator, <, <);
-  BINARY_COMPARISON_OPERATOR_VISITOR(GreaterOperator, >, >);
-  BINARY_COMPARISON_OPERATOR_VISITOR(LessEqualOperator, <=, <=);
-  BINARY_COMPARISON_OPERATOR_VISITOR(GreaterEqualOperator, >=, >=);
+  BINARY_OPERATOR_VISITOR(NotEqualOperator, !=, <>);
+  BINARY_OPERATOR_VISITOR(EqualOperator, ==, =);
+  BINARY_OPERATOR_VISITOR(LessOperator, <, <);
+  BINARY_OPERATOR_VISITOR(GreaterOperator, >, >);
+  BINARY_OPERATOR_VISITOR(LessEqualOperator, <=, <=);
+  BINARY_OPERATOR_VISITOR(GreaterEqualOperator, >=, >=);
```
imilinovic, line 257: I think BINARY_COMPARISON_OPERATOR_VISITOR is now unused and can be deleted

imilinovic (approved):

src/query/typed_value.hpp
```diff
@@ -797,4 +797,22 @@ class IncompatibleTypesComparisonException : public TypedValueException {
   SPECIALIZE_GET_EXCEPTION_NAME(IncompatibleTypesComparisonException)
 };
```
imilinovic: also you can now delete IncompatibleTypesComparisonException as well

PR #3750, opened 2026-02-04
feat: Coordinator request forwarding

as51340:
All cluster management queries can now be executed on any coordinator. Queries will be automatically forwarded to the current leader. Especially useful in deployments where there is a single load balancer on top of all coordinators.


as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Raft specifies that all queries should be executed on the current leader. This has some consequences on the UX side where users would have to first check which instance is the current leader and only after they find out, execute query on top of it. With this PR, all cluster management queries run on followers will be automatically routed to the current leader. This is especially useful in K8s deployments where there is only a single load balancer on top of all coordinators. [#3750](https://github.com/memgraph/memgraph/pull/3750)
- [x] **[Documentation PR link](https://github.com/memgraph/documentation/pull/1535)**

src/coordination/coordinator_instance.cpp
```diff
@@ -905,9 +969,16 @@ auto CoordinatorInstance::RemoveCoordinatorInstance(int coordinator_id) const ->
...
+  auto const leader_id = raft_state_->GetLeaderId();
+
+  if (ShouldForward(leader_id)) {
+    if (auto *leader = FindClientConnector(leader_id); leader != nullptr) {
+      return leader->SendBoolRpc<AddCoordinatorRpc>(config) ? AddCoordinatorInstanceStatus::SUCCESS
+                                                            : AddCoordinatorInstanceStatus::LEADER_NOT_FOUND;
+    }
+
+    return AddCoordinatorInstanceStatus::LEADER_FAILED;
+  }
```
andrejtonev: can we combine this logic into some function?
looks like we repeat it often

src/coordination/coordinator_instance.cpp
```diff
@@ -905,9 +969,16 @@ auto CoordinatorInstance::RemoveCoordinatorInstance(int coordinator_id) const ->
 
 auto CoordinatorInstance::AddCoordinatorInstance(CoordinatorInstanceConfig const &config) const
```
andrejtonev, line 949: I assume that now we forward the command to the leader, the leader executes it and then update raft, which in turn updates the state on all coordinators. Right?

src/coordination/coordinator_instance.cpp
```diff
@@ -534,9 +538,17 @@ auto CoordinatorInstance::ReconcileClusterState_() -> ReconcileClusterStateStatu
...
   if (!status.compare_exchange_strong(
           expected, CoordinatorStatus::LEADER_NOT_READY, std::memory_order_acq_rel, std::memory_order_acquire)) {
+    if (auto *leader = FindClientConnector(leader_id); leader != nullptr) {
+      return leader->SendBoolRpc<ForceResetRpc>() ? ReconcileClusterStateStatus::SUCCESS
+                                                  : ReconcileClusterStateStatus::LEADER_NOT_FOUND;
+    }
+
+    return ReconcileClusterStateStatus::LEADER_FAILED;
+
     return ReconcileClusterStateStatus::NOT_LEADER_ANYMORE;
```
andrejtonev: 2 returns. which one is correct?
Should there be an if need to forward?

src/coordination/coordinator_instance_management_server_handlers.cpp
```diff
@@ -42,6 +104,81 @@ void CoordinatorInstanceManagementServerHandlers::Register(CoordinatorInstanceMa
...
 }
 
+void CoordinatorInstanceManagementServerHandlers::AddCoordinatorHandler(CoordinatorInstance const &coordinator_instance,
+                                                                        uint64_t request_version,
+                                                                        slk::Reader *req_reader,
+                                                                        slk::Builder *res_builder) {
+  AddCoordinatorReq req;
+  rpc::LoadWithUpgrade(req, request_version, req_reader);
+  auto const res = coordinator_instance.AddCoordinatorInstance(req.arg_);
+  AddCoordinatorRes const rpc_res{res == AddCoordinatorInstanceStatus::SUCCESS};
```
andrejtonev: should we support all status types or an optional error message we can print on the callers side?
this goes for all of these messages

andrejtonev (commented):
nice 💪

src/coordination/coordinator_instance.cpp
as51340: exactly

src/coordination/coordinator_instance.cpp
as51340: none, correct thing is to return leader_not_found

src/coordination/coordinator_instance_management_server_handlers.cpp
as51340: I chose the simple approach of just returning bool and logging errors on the callee (leader side) to avoid sending strings around the network just for logging purposes. At the end it's pretty much the same whether the user sends us a single log file or files from all coordinators

src/coordination/raft_state.cpp
```diff
@@ -34,6 +34,147 @@
...
+    case RESULT_NOT_EXIST_YET: {
+      spdlog::error("Request failed because the result doesn't exist yet");
+      break;
+    }
+    case FAILED: {
+      spdlog::error("Generic request failure");
+      break;
+    }
+    default: {
+      std::unreachable();
```
andrejtonev, line 98: do we need this?
usually we just force all the cases to be handled

as51340: This is just a good practice nothing more https://en.cppreference.com/w/c/program/unreachable.html

PR #3751, opened 2026-02-04
fix: Cost model for BFS algorithm + index lookup

imilinovic:
1. Only use STShortestPathCursor if both destination and source index are present.
2. If 1. is true use a cost model to decide between STShortestPath and SingleSourceBFS

Better way to resolve this would be unification of BFS operator and deciding which BFS to use at runtime. Due to deciding at plan time we can't cache the query. Unification wasn't done as part of this PR due to the complexity it would add - explain query on BFS operator would potentially produce wrong results due the runtime plan being different.

### DIFF
--- a/src/query/plan/rewrite/index_lookup.hpp
+++ b/src/query/plan/rewrite/index_lookup.hpp
@@ -379,7 +379,7 @@ class IndexLookupRewriter final : public HierarchicalLogicalOperatorVisitor {
       auto destination_result = FindBestScanByIndex(dst_scan);
       // If it has an index check that its not a point index
       bool destination_has_index =
-          destination_result.has_value() ? HasIndexedSource(destination_result->operator_->input()) : false;
+          destination_result.has_value() ? HasIndexedSource(destination_result->operator_) : false;
       bool source_has_index = HasIndexedSource(expand.input());
 
       if (destination_has_index && source_has_index) {


imilinovic:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Produce better plans for some queries which use BFS and index lookup. [#3751](https://github.com/memgraph/memgraph/pull/3751)

src/query/plan/operator.cpp
```diff
@@ -2795,7 +2798,10 @@ class SingleSourceShortestPathCursor : public query::plan::Cursor {
         to_visit_next_.clear();
         processed_.clear();
 
-        const auto &vertex_value = frame[self_.input_symbol_];
+        // When existing_node is true, we expand from destination (node_symbol) instead of source (input_symbol_)
+        // This happens when only destination index exists (single-source BFS from destination)
+        const auto &vertex_value =
+            self_.common_.existing_node ? frame[self_.common_.node_symbol] : frame[self_.input_symbol_];
```
Ignition: We should normalise code so that we are guaranteed to know which is src vs dst. Everything else should change around that.

imilinovic:
Longer explanation

1. BFS Cost Model and Bidirectional BFS
The most significant change is in src/query/plan/rewrite/index_lookup.hpp, where a heuristic-based cost model is introduced to decide whether to use bidirectional BFS (STShortestPath) or single-source BFS.
Heuristic Logic: The rewriter now checks if both the source and destination of a BFS have indices. If they do, it uses ShouldUseSTShortestPath to evaluate if a bidirectional search is beneficial. This check includes:
Pair Explosion Guard: Ensures the product of source and destination cardinalities is within limits (≤ 1024).
Balance Heuristic: Prefers bidirectional BFS when the source and destination sets are somewhat balanced (ratio ≤ 4.0).
Graph Statistics: Uses the average branching factor of the graph to estimate if a meeting in the middle is likely to be faster.
Benefit: This allows Memgraph to use the more efficient bidirectional BFS algorithm when it's predicted to outperform a single-source search.
2. BFS Operator Flexibility (Flipping)
The branch enables the planner to "flip" BFS expansions, which was previously forbidden.
Flipping Safety: In src/query/plan/variable_start_planner.cpp, flipping is now allowed for BFS/DFS as long as:
The expansion doesn't use accumulated_path (which is order-dependent).
The expansion's filter doesn't reference the inner_node (the intermediate node during expansion). This is a crucial safety check because DFS/BFS only filter the target node of each step; flipping would change which node is filtered.
Implementation: src/query/plan/operator.cpp was updated to remove the assertion preventing reversed BFS. SingleSourceShortestPathCursor now correctly handles the is_reverse_ flag and the is_flipped state from the planner.
3. Bug Fixes and Improvements
ExpandVariable::Clone: Fixed a bug where the limit_ member was not being cloned.

src/query/plan/operator.cpp
```diff
@@ -2687,13 +2695,7 @@ class SingleSourceShortestPathCursor : public query::plan::Cursor {
         input_cursor_(self_.input()->MakeCursor(mem)),
         processed_(mem),
         to_visit_next_(mem),
-        to_visit_current_(mem) {
-    MG_ASSERT(!self_.common_.existing_node,
-              "Single source shortest path algorithm "
-              "should not be used when `existing_node` "
-              "flag is set, s-t shortest path algorithm "
-              "should be used instead!");
```
Ignition, line 2695: If you put this back use DMG_ASSERT

PR #3754, opened 2026-02-05
Fix performance regression for vector index creation/import

DavIvek:
I believe I’ve identified the cause of the ~30% regression: it comes from using SimSIMD for the vector index metric. Possible explanation is that in that path, the distance kernel is invoked through a function pointer and can’t be inlined, so every comparison pays additional call overhead. When SimSIMD is disabled, the builtin path is inlined by the compiler and autovectorized, so we still benefit from SIMD but without the extra call cost (benchmarked on Apple M4).

Additionally, simplified the vector index implementation and improved performance by switching storage to a single map keyed by index ID instead of maintaining separate maps (label+property and name→key), which reduces lookups and avoids extra NameIdMapper usage.

### DIFF
--- a/src/storage/v2/indices/vector_index.cpp
+++ b/src/storage/v2/indices/vector_index.cpp
@@ -79,10 +79,10 @@ std::optional<uint64_t> VectorIndex::SetupIndex(const VectorIndexSpec &spec, Nam
   if (pimpl->index_by_id_.contains(index_id)) {
     return std::nullopt;
   }
-  if (r::find_if(pimpl->index_by_id_, [&](const auto &id_index_item) {
-        const auto &spec = id_index_item.second.spec;
-        return spec.label_id == spec.label_id && spec.property == spec.property;
-      }) != pimpl->index_by_id_.end()) {
+  if (r::any_of(pimpl->index_by_id_, [&](const auto &id_index_item) {
+        auto &index_spec = id_index_item.second.spec;
+        return spec.label_id == index_spec.label_id && spec.property == index_spec.property;
+      })) {
     return std::nullopt;
   }
 
--- a/src/storage/v2/indices/vector_index_utils.hpp
+++ b/src/storage/v2/indices/vector_index_utils.hpp
@@ -11,6 +11,7 @@
 
 #pragma once
 
+#include <concepts>
 #include <optional>
 #include "flags/bolt.hpp"
 #include "flags/general.hpp"
@@ -368,6 +369,7 @@ void UpdateVectorIndex(utils::Synchronized<Index, std::shared_mutex> &mg_index, 
 /// @param vertices The vertices accessor to iterate over.
 /// @param process The function to call for each vertex (thread_id is std::nullopt).
 template <typename ProcessFunc>
+  requires std::invocable<ProcessFunc, Vertex &, std::optional<std::size_t>>
 void PopulateVectorIndexSingleThreaded(utils::SkipList<Vertex>::Accessor &vertices, ProcessFunc &&process) {
   const utils::MemoryTracker::OutOfMemoryExceptionEnabler oom_exception;
   for (auto &vertex : vertices) {
@@ -380,6 +382,7 @@ void PopulateVectorIndexSingleThreaded(utils::SkipList<Vertex>::Accessor &vertic
 /// @param vertices The vertices accessor to iterate over.
 /// @param process The function to call for each vertex (thread_id is the chunk index).
 template <typename ProcessFunc>
+  requires std::invocable<ProcessFunc, Vertex &, std::optional<std::size_t>>
 void PopulateVectorIndexMultiThreaded(utils::SkipList<Vertex>::Accessor &vertices, ProcessFunc &&process) {
   const utils::MemoryTracker::OutOfMemoryExceptionEnabler oom_exception;
   auto vertices_chunks = vertices.create_chunks(FLAGS_storage_recovery_thread_count);


DavIvek:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Disabled SimSIMD for the vector index metric to eliminate extra call overhead and simplified index storage by using a single map keyed by index ID, improving performance and reducing the ~30% regression while maintaining SIMD benefits. [#3754](https://github.com/memgraph/memgraph/pull/3754)

libs/CMakeLists.txt
```diff
@@ -354,10 +354,18 @@ add_library(tantivy_text_search ALIAS mgcxx_text_search)
...
+# SimSIMD variant (USEARCH_USE_SIMSIMD=1) for usearch_index_add_simsimd benchmark; uncomment with benchmark block.
+#[[
+add_library(usearch_simsimd INTERFACE IMPORTED GLOBAL)
+set_property(TARGET usearch_simsimd PROPERTY INTERFACE_INCLUDE_DIRECTORIES
   ${CMAKE_CURRENT_SOURCE_DIR}/usearch/include
   ${CMAKE_CURRENT_SOURCE_DIR}/usearch/fp16/include
   ${CMAKE_CURRENT_SOURCE_DIR}/usearch/simsimd/include
   ${CMAKE_CURRENT_SOURCE_DIR}/usearch/stringzilla/include)
-target_compile_definitions(usearch INTERFACE
-  USEARCH_USE_FP16LIB=1
```
andrejtonev, line 362: could we still use `USEARCH_USE_FP16LIB` in the regular build

src/storage/v2/indices/indices.cpp
```diff
@@ -77,11 +77,10 @@ void Indices::UpdateOnRemoveLabel(LabelId label, Vertex *vertex, Transaction &tx
       IndexedPropertyDecoder<Vertex>{.indices = this, .name_id_mapper = name_id_mapper, .entity = vertex});
 }
 
-void Indices::UpdateOnSetProperty(PropertyId property, const PropertyValue &value, Vertex *vertex, Transaction &tx,
-                                  NameIdMapper *name_id_mapper) {
+void Indices::UpdateOnSetProperty(PropertyId property, const PropertyValue &value, Vertex *vertex, Transaction &tx) {
   tx.active_indices_.label_properties_->UpdateOnSetProperty(property, value, vertex, tx);
   text_index_.UpdateOnSetProperty(vertex, tx, property);
-  vector_index_.UpdateOnSetProperty(property, value, vertex, name_id_mapper);
```
andrejtonev, line 84: where did name_id_mapper go?

src/storage/v2/indices/vector_index_utils.hpp
```diff
@@ -334,15 +319,16 @@ inline std::size_t GetVectorIndexThreadCount() {
...
 /// @param key The key to add/update in the index.
-/// @param vector The float vector data (taken by value to enable move semantics). If empty, entry is only removed.
+/// @param vector The vector to insert into the index.
 /// @param thread_id Optional thread ID hint for usearch's internal thread-local optimizations.
 /// @throws query::VectorSearchException if dimension mismatch or add fails for reasons other than capacity.
-template <typename Index, typename Key, typename Spec>
+template <typename Index, typename Key, typename Spec, typename Vector>
 void UpdateVectorIndex(utils::Synchronized<Index, std::shared_mutex> &mg_index, Spec &spec, const Key &key,
-                       utils::small_vector<float> vector, std::optional<std::size_t> thread_id = std::nullopt) {
+                       Vector &&vector, std::optional<std::size_t> thread_id = std::nullopt) {
```
andrejtonev: are you forcing takeover of the ownership or just want to move?
if it's just about the move, leave if without && and allow the caller to move

src/storage/v2/indices/vector_index_utils.hpp
```diff
@@ -379,29 +365,23 @@ void UpdateVectorIndex(utils::Synchronized<Index, std::shared_mutex> &mg_index,
...
-/// @tparam Args Additional argument types to forward to the process function.
+/// @tparam ProcessFunc Callable with signature void(Vertex&, std::optional<std::size_t> thread_id).
 /// @param vertices The vertices accessor to iterate over.
-/// @param process The function to call for each vertex.
-/// @param args Arguments to forward to the process function.
-template <typename ProcessFunc, typename... Args>
-void PopulateVectorIndexSingleThreaded(utils::SkipList<Vertex>::Accessor &vertices, const ProcessFunc &process,
-                                       Args &&...args) {
+/// @param process The function to call for each vertex (thread_id is std::nullopt).
+template <typename ProcessFunc>
```
andrejtonev, line 371: add constraint that the function is callable with vector and optional

andrejtonev: same for the other places where we removed `args...`

src/storage/v2/indices/vector_index.hpp
```diff
@@ -17,7 +17,19 @@
 #include "storage/v2/property_value.hpp"
 #include "storage/v2/snapshot_observer_info.hpp"
 #include "storage/v2/vertex.hpp"
+
+// Suppress usearch library warnings
+#if defined(__clang__)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-W#warnings"
```
andrejtonev: can we target the specific warning?

src/storage/v2/indices/vector_index.cpp
```diff
@@ -42,16 +42,10 @@ struct IndexItem {
...
-  /// The `index_` member is a map that associates a `LabelPropKey` (a combination of label and property)
-  /// with the pair of a IndexItem.
-  std::unordered_map<LabelPropKey, IndexItem> index_;
-
-  /// The `index_name_to_label_prop_` is a map that maps an index name (as a string) to the corresponding
-  /// `LabelPropKey`. This allows the system to quickly resolve an index name to the spec
-  /// associated with that index.
-  std::map<std::string, LabelPropKey, std::less<>> index_name_to_label_prop_;
+  /// Map from index ID to IndexItem.
+  std::unordered_map<uint64_t, IndexItem> index_by_id_;
```
andrejtonev, line 48: This was being done via name id mapper, right?
Is this thread-safe now?
Is it correctly initialized?

src/storage/v2/indices/vector_index.cpp
```diff
@@ -79,8 +74,16 @@ bool VectorIndex::CreateIndex(VectorIndexSpec &spec, utils::SkipList<Vertex>::Ac
...
 }
 
-bool VectorIndex::SetupIndex(VectorIndexSpec &spec) {
-  const auto label_prop = LabelPropKey{spec.label_id, spec.property};
+std::optional<uint64_t> VectorIndex::SetupIndex(const VectorIndexSpec &spec, NameIdMapper *name_id_mapper) {
+  const auto index_id = name_id_mapper->NameToId(spec.index_name);
+  if (pimpl->index_by_id_.contains(index_id)) {
+    return std::nullopt;
+  }
+  for (const auto &[_, item] : pimpl->index_by_id_) {
```
andrejtonev: use find_if

src/storage/v2/indices/vector_index.cpp
```diff
@@ -96,17 +99,8 @@ bool VectorIndex::SetupIndex(VectorIndexSpec &spec) {
...
-    return false;
-  }
-
-  auto [index_it, index_inserted] = pimpl->index_.try_emplace(label_prop, std::move(mg_vector_index.index), spec);
-  if (!index_inserted) {
-    pimpl->index_name_to_label_prop_.erase(name_it);
-    return false;
-  }
-  return true;
+  const auto [_, inserted] = pimpl->index_by_id_.try_emplace(index_id, std::move(mg_vector_index.index), spec);
```
andrejtonev, line 103: You are protecting the name via the id?
So you avoid going to name id mapper after the initial setup?

andrejtonev: Sorry for the comments, just thinking out loud.
But do answer them :D

src/storage/v2/indices/vector_index.cpp
```diff
@@ -115,43 +109,48 @@ void VectorIndex::RecoverIndex(VectorIndexRecoveryInfo &recovery_info, utils::Sk
...
+            thread_id);
+      }
+      if (snapshot_info) {
+        snapshot_info->Update(UpdateType::VECTOR_IDX);
+      }
+    };
 
     if (FLAGS_storage_parallel_schema_recovery && FLAGS_storage_recovery_thread_count > 1) {
-      PopulateVectorIndexMultiThreaded(vertices, process_vertex_for_recovery, spec);
+      PopulateVectorIndexMultiThreaded(vertices, process_vertex_for_recovery);
```
andrejtonev, line 141: Why did we remove spce?
I though we introduced it to make recovery simpler/faster?

src/storage/v2/indices/vector_index.cpp
```diff
@@ -274,54 +272,56 @@ void VectorIndex::UpdateOnRemoveLabel(LabelId label, Vertex *vertex, const Index
...
+void VectorIndex::UpdateOnSetProperty(PropertyId property, const PropertyValue &value, Vertex *vertex) {
   // Property should already be updated to the vector index id if it has vector index defined on it.
   if (value.IsVectorIndexId()) {
     const auto &vector_property = value.ValueVectorIndexList();
     const auto &index_ids = value.ValueVectorIndexIds();
     for (auto index_id : index_ids) {
-      const auto &idx_name = name_id_mapper->IdToName(index_id);
-      const auto &label_prop = pimpl->index_name_to_label_prop_.at(idx_name);
-      auto &index_item = pimpl->index_.at(label_prop);
+      auto &index_item = pimpl->index_by_id_.at(index_id);
```
andrejtonev, line 286: I'm not really following.
We used to have to find the label and then find the correct index.
Now we just have these index_id that does everything for us?

src/storage/v2/indices/vector_index.cpp
```diff
@@ -338,69 +338,82 @@ std::vector<VectorIndexInfo> VectorIndex::ListVectorIndicesInfo() const {
...
+      std::ranges::for_each(buffer, [&](auto value) { encoder->WriteDouble(value); });
+    }
   }
 }
 
 std::optional<uint64_t> VectorIndex::ApproximateNodesVectorCount(LabelId label, PropertyId property) const {
-  auto it = pimpl->index_.find(LabelPropKey{label, property});
-  if (it == pimpl->index_.end()) {
-    return std::nullopt;
+  for (const auto &[_, index_item] : pimpl->index_by_id_) {
```
andrejtonev: find_if

src/storage/v2/indices/vector_index.cpp
```diff
@@ -436,43 +449,41 @@ void VectorIndex::RemoveObsoleteEntries(std::stop_token token) const {
...
-           const auto &[_, spec] = pimpl->index_.at(label_prop);
-           return name_id_mapper->NameToId(spec.index_name);
-         }) |
-         r::to<utils::small_vector<uint64_t>>();
+bool VectorIndex::Empty() const { return pimpl->index_by_id_.empty(); }
+
+utils::small_vector<uint64_t> VectorIndex::GetVectorIndexIdsForVertex(Vertex *vertex, PropertyId property) const {
+  utils::small_vector<uint64_t> result;
+  result.reserve(static_cast<uint32_t>(pimpl->index_by_id_.size()));
+  for (const auto &[index_id, index_item] : pimpl->index_by_id_) {
```
andrejtonev, line 469: You are building the prop+label to index map on the fly?

src/storage/v2/indices/indices.cpp
DavIvek: Not needed anymore in this function since we already resolved property values into vector index ids.

src/storage/v2/indices/vector_index.cpp
DavIvek: Yes, but I removed it because I believe there’s no need to have two maps, and it’s better to avoid hashing strings. Additionally, we don’t expect users to have that many indexes, so there’s no real benefit to those maps. It’s thread-safe because the map is modified only when creating or dropping an index, which must happen within a single transaction.

src/storage/v2/indices/vector_index.cpp
DavIvek: Yes, the idea is to access the nameId mapper as little as possible. It might even be better if the vector index had its own nameId mapper, but let’s keep that idea for future improvements for now.

src/storage/v2/indices/vector_index.cpp
DavIvek: We had mapping id -> name -> label_prop -> index
Now it's simplified and we can get the index just by using id.

src/storage/v2/indices/vector_index.cpp
DavIvek: Since we don’t expect to have many indexes, this shouldn’t be expensive. Inside IndexItem we store the spec, and based on that spec we can filter by label and property.

src/storage/v2/indices/vector_index.cpp
DavIvek: We take spec on line 109:
`auto &spec = recovery_info.spec;`
And then in lambda we capture by ref:
`auto process_vertex_for_recovery = [&](Vertex &vertex, std::optional<std::size_t> thread_id)`

libs/CMakeLists.txt
DavIvek: It looks to me like usearch selects that based on the architecture (from index_plugins.hpp):
```
#if !defined(USEARCH_USE_FP16LIB)
#if defined(__AVX512F__)
#define USEARCH_USE_FP16LIB 0
#elif defined(USEARCH_DEFINED_ARM)
#include <arm_fp16.h> // `__fp16`
#define USEARCH_USE_FP16LIB 0
#else
#define USEARCH_USE_FP16LIB 1
#endif
#endif
```
Forcing it on my machine actually made it slower.

src/storage/v2/indices/vector_index.hpp
DavIvek: Not that I know, but I can move it to CMake.

src/storage/v2/indices/vector_index_utils.hpp
DavIvek: I wanted to move it so that during the recovery process we could move vectors and avoid duplicated memory, but that was actually wrong, good catch. Does it make sense to add overloads for a const & and for a &&?

DavIvek: But actually, I’m thinking it might be better to have a single function and, in this specific recovery case, just call `clear` and `shrink_to_fit` on the vector.

src/storage/v2/indices/vector_index.cpp
andrejtonev: But if I have a label prop,  I need to do a linear search, right?
And we are expecting low number of vector indices, so that is fine.

DavIvek: yes 👍

DavIvek:
I ran the same benchmark on AMD EPYC “Rome” (Zen 2, 16-core, AVX2) CPU and got the same results.

PR #3802, opened 2026-02-17
feat: Avoid exceptions in replication and handle specifically timeout replication error

as51340:
## What

Replaces exception-based error handling across the entire RPC/SLK communication stack with `std::expected`-based error propagation. Introduces a new `utils::RpcError` enum
(`UNSUPPORTED_RPC_VERSION_ERROR`, `GENERIC_RPC_ERROR`, `FAILED_TO_GET_RPC_STREAM`, `TIMEOUT_ERROR`) and threads it through socket I/O, SLK serialization, RPC client, replication,
 and coordinator layers. Adds a new `TimeoutReplicationError` that surfaces to users when an RPC timeout occurs during replication commit, advising them to adjust
`deltas_batch_progress_size`.

## Why

The existing exception-based RPC error handling made it impossible to distinguish timeout errors from generic connection failures. This distinction is critical: a timeout during
replication needs a different user-facing message (suggesting a configuration change) than a hard connection failure (which triggers recovery). Switching to `std::expected` also
eliminates the overhead and control-flow complexity of exceptions in a hot path (every RPC call), and makes error provenance explicit at each layer boundary.

## How

The refactoring follows a bottom-up approach through the communication stack:

- **Socket/Communication** (`io/network/socket.{hpp,cpp}`, `communication/client.{hpp,cpp}`): `Socket::Write()` and `Client::Read()`/`Client::Write()` now return
`std::expected<void, ClientCommunicationError>` instead of `bool`. A new `ClientCommunicationError` enum distinguishes `TIMEOUT_ERROR` from `GENERIC_ERROR`.

- **SLK streams** (`slk/streams.{hpp,cpp}`): `Builder` tracks errors internally via an `std::optional<RpcError> error_` member; writes become no-ops once an error is latched. The
 write callback signature changes from `void(...)` to `std::expected<void, RpcError>(...)`. `Reader` similarly enters an error state on malformed data instead of throwing,
surfaced via `GetError()`. `SlkBuilderException`, `SlkReaderLeftoverDataException` are removed.

- **RPC client** (`rpc/client.hpp`): `Stream()`, `TryStream()`, `UpgradeStream()`, `SendAndWait()`, `SendAndWaitProgress()` all return `std::expected` instead of throwing.
Template parameter renamed from `TRequestResponse` to `RPC` constrained by `IsRpc` concept. Exception classes `GenericRpcFailedException`, `UnsupportedRpcVersionException`,
`FailedToGetRpcStreamException` are deleted.

- **Replication** (`storage/v2/replication/`): `ReplicaStream::Finalize()`, `ReplicationStorageClient::FinalizePrepareCommitPhase()`, `FinalizeTransactionReplication()`,
`ShipDeltas()`, `UpdateReplicaState()` return `std::expected<void, ReplicationError>` where `ReplicationError` is a new enum with `TIMEOUT_ERROR` and `GENERIC_ERROR`.
`HandleDurabilityAndReplicate()` returns `std::expected<void, ReplicationError>`, enabling the commit path in `InMemoryStorage::Commit()` to distinguish timeout from generic
replication failure.

- **Query layer** (`query/interpreter.cpp`, `query/plan/operator.cpp`): New `TimeoutReplicationError` variant in `StorageManipulationError` is handled in commit,
trigger-after-commit, and periodic-commit paths. A timeout throws `ReplicationException` with actionable guidance.

- **Coordination** (`coordination/`): All `SendRpc` patterns updated from try/catch to `std::expected` checking. `InProgressRes` moved from `storage::replication` to `rpc`
namespace.

- **SLK bug fix** (`slk/streams.cpp`): Fixed `CheckStreamStatus()` to handle TCP segmentation when transitioning between files — requests additional data when the file-segment
mask arrives but file metadata hasn't yet.

## Testing

- Extensive updates to existing unit tests (`slk_streams.cpp`, `rpc.cpp`, `rpc_in_progress.cpp`, `rpc_timeouts.cpp`, `rpc_versioning.cpp`, `license.cpp`,
`file_replication_handler.cpp`, `replication_rpc_progress.cpp`, `snapshot_rpc_progress.cpp`) to use `std::expected`-based APIs.
- New SLK unit tests added in `slk_streams.cpp` covering builder/reader error propagation scenarios.
- Jepsen tests updated (`bank/test.clj`, `create/test.clj`, `mtenancy/test.clj`) to handle the new `replica-timeout?` error classification.
- E2E `durability.py` test adjusted for snapshot count assertion relaxation.
- Integration test files updated with missing `#include "communication/init.hpp"`.

## Notes for reviewers

- **No ISSU/durability version bump needed**: The on-wire SLK protocol format is unchanged; only the C++ error-handling mechanism changed.
- The `Builder` error-latching pattern (write becomes no-op after first error) is intentional to avoid cascading failures — callers check `Finalize()` or `GetError()` at the
boundary.
- `Reader::GetSegment()` no longer throws on leftover data — this is documented as benign (e.g., rejected `PrepareCommitReq` leaves unread deltas in the stream).
- The "last error wins" approach in `TransactionReplication::ShipDeltas()` mirrors the previous `success &= finalized` semantics but preserves the error type.
- `slk::Builder` moved-from state now returns `GENERIC_RPC_ERROR` from its write function instead of silently succeeding.

### DIFF
--- a/src/rpc/version.hpp
+++ b/src/rpc/version.hpp
@@ -61,6 +61,9 @@ struct ProtocolMessageHeader {
 inline auto LoadMessageHeader(slk::Reader *reader) -> std::expected<ProtocolMessageHeader, utils::RpcError> {
   ProtocolMessageHeader header;
   slk::Load(&header.protocol_version, reader);
+  if (auto err = reader->GetError(); err.has_value()) {
+    return std::unexpected{*err};
+  }
   switch (header.protocol_version) {
     case V1:
     case V2:
--- a/src/slk/streams.cpp
+++ b/src/slk/streams.cpp
@@ -26,7 +26,8 @@ bool Builder::IsEmpty() const { return pos_ == 0; }
 void Builder::Save(const uint8_t *data, uint64_t size) {
   size_t offset = 0;
   while (size > 0 && !error_) {
-    FlushSegment(false);
+    // NOLINTNEXTLINE(bugprone-unused-return-value)
+    FlushSegment(false);  // we check error_ that's why it is ok to use ignore [[nodiscard]]
     if (error_) return;
 
     size_t const to_write = std::min(size, kSegmentMaxDataSize - pos_);
@@ -94,7 +95,8 @@ void Builder::SaveFooter(uint64_t const total_size) {
   memcpy(segment_.data() + total_size, &kFooter, sizeof(SegmentSize));
 }
 
-auto Builder::FlushSegment(bool const final_segment, bool const force_flush) -> BuilderWriteFunction::result_type {
+[[nodiscard]] auto Builder::FlushSegment(bool const final_segment, bool const force_flush)
+    -> BuilderWriteFunction::result_type {
   if (error_) return std::unexpected(*error_);
   if (!force_flush && !final_segment && pos_ < kSegmentMaxDataSize) return {};
   MG_ASSERT(pos_ > 0, "Trying to flush out a segment that has no data in it!");


src/coordination/include/coordination/coordinator_instance_connector.hpp
```diff
@@ -24,13 +24,19 @@ class CoordinatorInstanceConnector {
   template <rpc::IsRpc Rpc, typename... Args>
   auto SendRpc(Args &&...args) {
     using ReturnType = decltype(std::declval<typename Rpc::Response>().arg_);
-    try {
```
andrejtonev, line 27: I would still leave the try catch block in order to limit any stray exceptions

src/storage/v2/inmemory/replication/recovery.hpp
```diff
@@ -70,39 +70,56 @@ bool WriteFiles(const T &paths, std::filesystem::path const &root_data_dir, repl
...
+  // if ASYNC mode, we shouldn't block on transferring durability files because there could be a commit task which
+  // holds rpc stream and which needs to be executed
+  auto maybe_stream_result = (mode == replication_coordination_glue::ReplicationMode::ASYNC)
+                                 ? client.TryStream<Rpc>(kRecoveryRpcTimeout, std::forward<Args>(args)...)
+                                 : client.Stream<Rpc>(std::forward<Args>(args)...);
 
   // If dealing with ASYNC replica and couldn't obtain the lock
   if (!maybe_stream_result) {
-    return std::nullopt;
+    return std::unexpected{utils::RpcError::FAILED_TO_GET_RPC_STREAM};
```
andrejtonev, line 107: Do TryStream and Stream fail in the same way?
Will we miss some ASYNC errors this way?

src/io/network/socket.cpp
```diff
@@ -360,24 +361,27 @@ bool Socket::Write(const uint8_t *data, size_t len, bool have_more, std::optiona
...
         // Terminal error, return failure.
-        return false;
+        return std::unexpected{ClientCommunicationError::GENERIC_ERROR};
       }
       // Non-fatal error, retry after the socket is ready. This is here to
       // implement a non-busy wait. If we just continue with the loop we have a
       // busy wait.
-      if (!WaitForReadyWrite(timeout_ms)) return false;
+      if (!WaitForReadyWrite(timeout_ms)) {
+        return std::unexpected{ClientCommunicationError::TIMEOUT_ERROR};
```
andrejtonev: poll error would now cause a TIMEOUT_ERROR. we should distinguish timeout errors (ret == 0) and connection errors (ret == -1)

src/storage/v2/replication/replication_transaction.cpp
```diff
@@ -46,22 +48,25 @@ auto TransactionReplication::ShipDeltas(uint64_t durability_commit_timestamp, Co
...
       // SYNC replica cannot be part of 2PC
-      return true;
+      return {};
     });
 
-    success &= finalized;
+    // Last error wins approach
+    if (!finalized.has_value()) {
+      status.swap(finalized);
+    }
```
andrejtonev, line 67: should we prioritize timeout?

src/slk/streams.cpp
```diff
@@ -43,45 +45,58 @@ void Builder::Save(const uint8_t *data, uint64_t size) {
...
-void Builder::FlushInternal(size_t const size, bool const has_more) {
-  write_func_(segment_.data(), size, has_more);
-  pos_ = 0;
+auto Builder::FlushInternal(size_t const size, bool const has_more) -> BuilderWriteFunction::result_type {
+  // Setting position to 0 is fine here because for the success, it means builder is fone
+  // If error, then we rely on short-circuit if (error_) return ...
+  utils::OnScopeExit const on_exit{[this]() { pos_ = 0; }};
+  if (error_) return std::unexpected(*error_);
+  if (auto const res = write_func_(segment_.data(), size, has_more); !res.has_value()) {
+    error_ = res.error();
```
andrejtonev, line 79: Not sure this will work.
Before we would throw and that way stop any work being done higher up. Now we store the error and ignore any data, but the higher up logic does not check for this and so would work on garbage data.

andrejtonev: For example ReadDelta

src/coordination/include/coordination/coordinator_instance_connector.hpp
as51340: Not sure what we would get with this. We would see it then only in the logs and we wouldn't know that there is a bug in the code because of which this exception is thrown.

src/storage/v2/inmemory/replication/recovery.hpp
as51340: failed to get rpc stream can only happen for async replicas. For sync replicas we block until we can obtain the lock.

src/io/network/socket.cpp
as51340: agree

src/storage/v2/replication/replication_transaction.cpp
as51340: makes sense

src/slk/streams.cpp
as51340: this is builder code, ReadDelta should should use reader

andrejtonev: both reader and builder have the same issue. they were throwing, now they are internally erroring

src/rpc/version.hpp
```diff
@@ -53,8 +58,7 @@ struct ProtocolMessageHeader {
       0};  // request/response version. 0 is safe default because 1 is the first version for all messages
 };
 
-// @throws UnsupportedRpcVersionException
-inline auto LoadMessageHeader(slk::Reader *reader) -> ProtocolMessageHeader {
+inline auto LoadMessageHeader(slk::Reader *reader) -> std::expected<ProtocolMessageHeader, utils::RpcError> {
   ProtocolMessageHeader header;
   slk::Load(&header.protocol_version, reader);
```
andrejtonev, line 63: should we be checking for reader error?

src/rpc/utils.hpp
```diff
@@ -49,20 +50,29 @@ void SendFinalResponse(TResponse const &res, uint64_t const response_version, sl
                                              .message_version = response_version};
   SaveMessageHeader(message_header, builder);
   SaveWithDowngrade(res, response_version, builder);
-  builder->Finalize();
-  spdlog::trace("[RpcServer] sent {}, version {}. {}", TResponse::kType.name, response_version, description);
+  if (auto const res = builder->Finalize(); !res.has_value()) {
```
andrejtonev, line 53: do we check for this error at a higher level, or we just log and silently fail?

src/storage/v2/replication/replication_client.cpp
```diff
@@ -75,47 +75,51 @@ ReplicationStorageClient::ReplicationStorageClient(::memgraph::replication::Repl
...
+    if (!hb_stream.has_value()) {
+      return std::unexpected{hb_stream.error()};
+    }
 
-  if (!maybe_heartbeat_res) {
+    return hb_stream->SendAndWait();
+  });
+
+  if (!maybe_heartbeat_res.has_value()) {
     spdlog::trace("Couldn't get RPC lock while trying to UpdateReplicaState");
```
andrejtonev, line 121: lets add the error message into the log

src/slk/streams.cpp
```diff
@@ -43,45 +45,58 @@ void Builder::Save(const uint8_t *data, uint64_t size) {
...
 }
 
 void Builder::SaveFooter(uint64_t const total_size) {
+  if (error_) return;
   memcpy(segment_.data() + total_size, &kFooter, sizeof(SegmentSize));
 }
 
-void Builder::FlushSegment(bool const final_segment, bool const force_flush) {
-  if (!force_flush && !final_segment && pos_ < kSegmentMaxDataSize) return;
+auto Builder::FlushSegment(bool const final_segment, bool const force_flush) -> BuilderWriteFunction::result_type {
```
andrejtonev: lets add a [[nodiscard]] so we have to check this result. that way we know there isn't a lingering error that is ignored.

src/rpc/utils.hpp
as51340: this is log + silent failure. That's fine here, because it happens at the end of the replica handler

src/storage/v2/replication/replication_client.cpp
as51340: the only thing that can happen here is that the lock couldn't have been obtained. So this log message should be fine

src/rpc/version.hpp
as51340: yes

as51340:
Done under https://github.com/memgraph/memgraph/pull/3942

PR #3807, opened 2026-02-18
Fix startup recovery to restore databases as system metadata

DavIvek:
Currently, when `--data-recovery-on-startup` is enabled, we recover users but not databases. Since both are system metadata rather than user data, we should recover both. However, actual user data should not be recovered.


DavIvek:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - When `--data-recovery-on-startup` is enabled, the system now recovers both users and databases, whereas previously only users were recovered. This ensures that all system metadata is restored consistently on startup, while actual user data is still not recovered. [#3807](https://github.com/memgraph/memgraph/pull/3807)

tests/e2e/durability/multitenancy.py
```diff
@@ -89,5 +89,66 @@ def test_mt_with_hidden_files(test_name):
     execute_and_fetch_all(cursor, "RETURN 0")
 
 
+def test_recovery_on_startup_false_recovers_tenants_but_not_data(test_name):
```
as51340, line 92: Also check that .backup directories exist

PR #3813, opened 2026-02-20
Add server side parameters per database

DavIvek:
This PR extends the server-side parameter store to be defined per database.


src/parameters/parameters.cpp
```diff
@@ -22,26 +22,15 @@ namespace memgraph::parameters {
...
-      return "global/";
-    default:
-      std::unreachable();
-  }
-}
-
-std::string MakeKey(ParameterScope scope, std::string_view name) {
-  return fmt::format("{}{}", ScopePrefix(scope), name);
-}
+std::string MakeKey(std::string_view name, std::string_view scope) { return fmt::format("{}/{}", scope, name); }
```
as51340, line 25: return type std::string_view to avoid allocations

imilinovic:
bugbot run

src/parameters/parameters.cpp
```diff
@@ -102,73 +92,71 @@ struct DeleteAllParametersAction : memgraph::system::ISystemAction {
...
-}
-
 Parameters::Parameters(const std::filesystem::path &storage_path) : storage_(storage_path) {}
 
-bool Parameters::SetParameter(std::string_view name, std::string_view value, ParameterScope scope,
-                              system::Transaction *txn) {
-  if (!storage_.Put(MakeKey(scope, name), value)) return false;
+SetParameterResult Parameters::SetParameter(std::string_view name, std::string_view value, std::string_view scope,
+                                            system::Transaction *txn) {
+  if (scope != kGlobalScope && GetParameter(name, kGlobalScope).has_value())
```
as51340: Can you explain this? if scope is not global and GetParameter from global?

src/query/cypher_query_interpreter.cpp
```diff
@@ -41,44 +41,45 @@ namespace memgraph::query {
...
-      auto opt = global_parameters->GetParameter(param_key, parameters::ParameterScope::GLOBAL);
-      if (opt) {
-        TypedValue value;
-        query::from_json(nlohmann::json::parse(*opt), value);
-        parameters.Add(param_index, static_cast<storage::ExternalPropertyValue>(value));
-        continue;
-      }
-    }
+    if (try_server_param(param_index, param_key, database_uuid)) continue;
+    if (try_server_param(param_index, param_key, parameters::kGlobalScope)) continue;
```
as51340, line 65: what if user didn't specify GLOBAL and try_server_param for database_uuid failed? You will still try to set parameter on the global scope

src/parameters/parameters.cpp
DavIvek: If I'm correct, we can't return `std::string_view` here. `fmt::format` returns a `std::string` and a `std::string_view` would point to that temporary.

src/parameters/parameters.cpp
DavIvek: The idea was to prevent the user from setting a database-level parameter if a global parameter with the same name already exists. However, I’ll remove this restriction. If we introduce this kind of logic, we should also support the opposite case, which would significantly complicate the implementation and could potentially introduce performance degradation. So let’s not do that for now.

src/query/cypher_query_interpreter.cpp
DavIvek: This code is only for evaluating parameters, not for setting them. First, we check whether a database-level parameter exists. If it doesn’t, we then check whether a global parameter exists.

DavIvek:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Added support for database-scoped server-side parameters (`SET PARAMETER` / `UNSET PARAMETER`) in addition to the existing global parameters (`SET GLOBAL PARAMETER`). Database parameters are tied to the current database and take precedence over global parameters of the same name during query execution, allowing users to customize parameter values per database while maintaining shared defaults globally. Users should use `SET PARAMETER` for database-specific values and `SET GLOBAL PARAMETER` for values shared across all databases; `SHOW PARAMETERS` will display both scopes for the active database. [#3813](https://github.com/memgraph/memgraph/pull/3813)

PR #3817, opened 2026-02-20
Memory optimization: Allow user to specify float property precision

andrejtonev:
New flag `--storage-floating-point-resolution-bits` allows the user to define double(64), float(32) or half(16) precision for property store values.
This is system wide and not replicated/made durable. This only affects PropertyStore which is an ephemeral structure, so restarts with changed resolution is possible.

### DIFF
--- a/src/storage/v2/property_store.cpp
+++ b/src/storage/v2/property_store.cpp
@@ -36,6 +36,8 @@
 #include "utils/logging.hpp"
 #include "utils/temporal.hpp"
 
+#include <fp16.h>  // Taken via usearch (seems like _Float16 is broken on some platforms)
+
 // NOLINTNEXTLINE (cppcoreguidelines-avoid-non-const-global-variables)
 DEFINE_bool(storage_property_store_compression_enabled, false,
             "Controls whether the properties should be compressed in the storage.");
@@ -140,9 +142,9 @@ constexpr uint32_t SizeToByteSize(Size size) {
   }
 }
 
-inline double HalfToDouble(uint16_t bits) { return static_cast<double>(std::bit_cast<_Float16>(bits)); }
+inline double HalfToDouble(uint16_t bits) { return static_cast<double>(fp16_ieee_to_fp32_value(bits)); }
 
-inline uint16_t DoubleToHalf(double value) { return std::bit_cast<uint16_t>(static_cast<_Float16>(value)); }
+inline uint16_t DoubleToHalf(double value) { return fp16_ieee_from_fp32_value(static_cast<float>(value)); }
 
 inline uint64_t FloatingPointResolution() { return FLAGS_storage_floating_point_resolution_bits; }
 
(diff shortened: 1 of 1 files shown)


andrejtonev:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Introduced `--storage-floating-point-resolution-bits` a system wide PropertyStore precision definition. This lowers the memory used by floating point numbers. Allows the user to select between double, float and half precisions. [#3817](https://github.com/memgraph/memgraph/pull/3817)
- [x] https://github.com/memgraph/documentation/pull/1541

src/storage/v2/property_store.cpp
```diff
@@ -884,8 +961,16 @@ bool DecodeList(Reader *reader, ListType list_type, uint32_t size, PropertyValue
...
       for (uint32_t i = 0; i < size; ++i) {
-        auto double_v = reader->ReadDouble(Size::INT64);
+        std::optional<double> double_v;
+        if (res <= 16) {
+          double_v = reader->ReadHalf(Size::INT16);
+        } else if (res <= 32) {
+          double_v = reader->ReadFloat(Size::INT32);
+        } else {
+          double_v = reader->ReadDouble(Size::INT64);
+        }
```
DavIvek: This logic is same/similar on multiple places, maybe it makes sense to extract it to a function.

src/storage/v2/property_store.cpp
```diff
@@ -903,7 +988,15 @@ bool DecodeList(Reader *reader, ListType list_type, uint32_t size, PropertyValue
           if (!int_v) return false;
           list.emplace_back(static_cast<int>(*int_v));
         } else if (metadata->type == Type::DOUBLE) {
-          auto double_v = reader->ReadDouble(metadata->payload_size);
+          auto const res = FloatingPointResolution();
+          std::optional<double> double_v;
+          if (res <= 16) {
```
DavIvek: `<= 16` and `<= 32` can be a bit misleading? Why not using `==`?

src/storage/v2/property_store.cpp
```diff
@@ -32,13 +32,24 @@
...
 DEFINE_bool(storage_property_store_compression_enabled, false,
             "Controls whether the properties should be compressed in the storage.");
 
+// NOLINTNEXTLINE (cppcoreguidelines-avoid-non-const-global-variables)
+DEFINE_VALIDATED_uint64(storage_floating_point_resolution_bits, 64,
+                        "Max bits for floating-point property storage (16, 32, or 64). "
+                        "Smaller values save space but reduce precision (32=float, 16=half).",
+                        {
+                          if (value == 16 || value == 32 || value == 64) return true;
+                          std::cout << "Expected --" << flagname << " to be one of 16, 32, 64\n";
```
DavIvek, line 51: Can we use `spdlog` instead of a `std::cout`?

andrejtonev:
for the std::float16_t
yes and no :D 
the support is very limited. The C version has a better support for now. 
Even the C version isn't that well supported. Using a library for f16
I'll leave a comment to revisit

src/storage/v2/property_store.cpp
andrejtonev: using == and addeding a DMG ASSERT

src/storage/v2/property_store.cpp
andrejtonev: we are using std::cout in other places when printing an error for the flag. 
spdlog needs to be enabled at the right level and also be pointer to a file or stderr. but this is not granted, so std::cout seems like a safe place to print the error message associated with the flag.
perhaps this isn't true when running in some environments, but will leave as is, since this is the way we do it in other places. we can think about changing it in a dedicated pr.

andrejtonev:
bugbot run

PR #3818, opened 2026-02-20
feat: Add e-graph rewrite system for planner v2

### LINKED ISSUE #3886: Add rewriter subsystem

imilinovic:
Adds an e-graph-based rewrite system to plan_v2 for equality-saturation-driven optimisation. `Pattern<Symbol>` is expressed through a Builder DSL; an `EMatcher` uses a symbol index for O(1) candidate lookup; slot-based `PatternMatch` bindings live in an arena-backed `MatchArena`. `RewriteRule<Symbol, Analysis>` composes multiple patterns via a hash-join multi-pattern matcher and hands rules a safe `RuleContext`. First production rule: `InlineRule`, which merges Identifier e-classes with their bound expressions (copy propagation).

The e-graph core gains duplicate e-node removal during congruence closure, `num_live_nodes()`/`num_dead_nodes()` tracking, and a `MergeResult` struct that surfaces what actually merged. The planner internals are reorganised into focused modules (`core/` → `egraph/`, `pattern/`, `rewrite/`, `extract/`), the cost estimator's magic numbers are pulled into `cost_constants.hpp` (`CostParam`, `CardParam`, `MiscParam`), and a small `symbol_make_traits` layer cleans up the plan_v2 `Make<S>()` facade.

Includes a unit-test suite and benchmarks for the new infrastructure, plus two rounds of post-review fixes (parts 1 and 2) covering deferred duplicate-removal in `repair_hashcons_enode`, the `PatternCompiler`/`CompiledPattern` → `PatternsCompiler`/`CompiledMatcher` rename, a correctness fix returning a compact `SymbolTable` from `ConvertToLogicalOperator`, and the traits-driven collapse of the symbol-emplace dispatch.


imilinovic:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist

imilinovic:
bugbot run

src/planner/egraph.cppm
```diff
@@ -0,0 +1,748 @@
...
+template <typename Symbol, typename Analysis>
+auto EGraph<Symbol, Analysis>::repair_hashcons_eclass(EClass<Analysis> const &eclass, EClassId eclass_id,
+                                                      BaseProcessingContext &ctx) -> EClassId {
+  // Collect canonical IDs to merge. We must iterate fully before merging to
+  // avoid iterator invalidation - merge() can modify/move the eclass.
+  auto &canonical_ids = ctx.canonical_eclass_ids;
+  canonical_ids.clear();
+
+  for (auto const &enode_id : eclass.nodes()) {
+    if (auto merge_target = repair_hashcons_enode(enode_id, eclass_id)) {
```
colinbarry, line 573: Whilst iterating the `span` on `eclass.nodes()`, `repair_hashcons_enode` can call `remove_duplicate_enode()`. This could invalidate the underlying iterators.

src/planner/egraph.cppm
```diff
@@ -0,0 +1,748 @@
...
+      merge_all(canonical_eclass_ids, ctx.union_find_context);
+      // hashcons update deferred to next rebuild iteration - duplicates will be
+      // detected and removed then via repair_hashcons_enode
+    } else if (canonical_eclass_ids.size() == 1) {
+      // All parent e-nodes in this group belong to the same e-class after canonicalization.
+      // We still need to check for congruence with e-nodes in OTHER e-classes via hashcons lookup.
+      // repair_hashcons_enode returns a merge target if it finds a duplicate in a different e-class.
+      auto current_eclass = EClassId{canonical_eclass_ids[0]};
+      for (auto enode_id : enode_ids) {
+        auto merge_target = repair_hashcons_enode(enode_id, current_eclass);
```
colinbarry, line 660: If we have two or more e-nodes in the same `canonical_to_parents` e-class, `repair_hashcons_enode` will canonicalize the first and remove the `hashcons` entry. The means the second e-node could hit the assert (in `repair_hashcons_enode`):

```
auto old_it = hashcons_.find(ENodeRef{enode});
assert(old_it != hashcons_.end() && "canonical e-node must exist in hashcons");
```

src/planner/include/planner/pattern/vm/executor.hpp
```diff
@@ -0,0 +1,601 @@
...
+      &&op_Halt,
+  };
+
+  // Cache current instruction - fetched once per dispatch, used by all handlers
+  Instruction instr;
+
+  // clang-format off
+#define DISPATCH()                                                      \
+  do {                                                                  \
+    if (state_.pc >= code_.size()) return; /*DMG_ASSERT*/               \
```
colinbarry, line 301: I think the `DMG_ASSERT` is the better choice here. In theory, should never hit, but as is could mask bytecode generation bugs that would be easier to debug if we can catch them here.

src/planner/include/planner/pattern/vm/executor.hpp
```diff
@@ -0,0 +1,601 @@
...
+  return true;
+}
+
+template <typename Symbol, typename Analysis, bool DevMode>
+auto VMExecutor<Symbol, Analysis, DevMode>::exec_iter_symbol_eclasses(Instruction instr) -> bool {
+  assert(matcher_index_ && "MatcherIndex must be set for IterSymbolEClasses");
+  auto const *set = matcher_index_->eclasses_set_for_symbol(symbols_[instr.arg]);
+
+  if constexpr (DevMode) {
+    collector_.on_iter_all_eclasses_start(state_.pc, set ? set->size() : 0);
```
colinbarry, line 481: Misleading trace callback name (albeit, one just used for `DevMode`.) Maybe add a dedicated `on_iter_symbol_eclasses_start`?

src/planner/include/planner/pattern/vm/tracer.hpp
```diff
@@ -0,0 +1,486 @@
...
+#include "planner/pattern/vm/instruction.hpp"
+
+namespace memgraph::planner::core::pattern::vm {
+
+/// Statistics collected during VM execution (for benchmarking/profiling)
+struct VMStats {
+  std::size_t instructions_executed{0};
+  std::size_t iter_enode_calls{0};
+  std::size_t iter_parent_calls{0};
+  std::size_t parent_symbol_hits{0};    // Parents that matched symbol filter
```
colinbarry, line 33: `parent_symbol_hits` actually track any `CheckSymbol`, not just parents. Same for misses.

src/planner/include/planner/pattern/vm/tracer.hpp
```diff
@@ -0,0 +1,486 @@
...
+    if (tracer) tracer->on_iter_start(pc, count);
+  }
+
+  void on_iter_parent_start(std::size_t pc, std::size_t count) {
+    ++stats.iter_parent_calls;
+    if (tracer) tracer->on_iter_start(pc, count);
+  }
+
+  void on_iter_all_eclasses_start(std::size_t pc, std::size_t count) {
+    if (tracer) tracer->on_iter_start(pc, count);
```
colinbarry, line 323: Unlike the other `iter_` hooks, there is no `VMstats` update here.

src/planner/include/planner/pattern/vm/instruction.hpp
```diff
@@ -0,0 +1,355 @@
...
+enum class VMOp : uint8_t {
+  // ===== Navigation =====
+  /// Load child e-class from current e-node: dst = enode[src].children[arg]
+  LoadChild,
+  /// Get e-class containing e-node in src register: dst = eclass_of(enode[src])
+  GetENodeEClass,
+
+  // ===== E-Node Iteration =====
+  /// Begin iterating e-nodes in e-class[src], store first in dst
+  /// Pushes iteration state. If empty, jumps to target.
```
colinbarry, line 129: "If empty, jumps to target" is not true: An e-class must always have at least one  e-node, so the code (correctly) doesn't have to handle that case.

src/planner/include/planner/pattern/vm/state.hpp
```diff
@@ -0,0 +1,255 @@
...
+
+  // Program counter
+  std::size_t pc{0};
+
+  // Iteration state: separate homogeneous vectors indexed by register
+  // Each type uses exhausted state as inactive indicator (no explicit deactivation needed)
+  // The compiler statically determines which type each register uses
+  // When an iteration exhausts, its state is naturally inert; fresh iterations overwrite
+  std::vector<ENodesIter> enodes_iters;                   // E-node iterations (span-based)
+  std::vector<ParentsIter> parents_iters;                 // Parent iterations (index-based)
```
colinbarry, line 121: `parents_iters` is actually set-based, not index-based

src/planner/include/planner/pattern/vm/executor.hpp
```diff
@@ -0,0 +1,601 @@
...
+  if constexpr (DevMode) {
+    collector_.on_iter_advance(state_.pc, iter.remaining());
+  }
+  out_reg = iter.current();
+  return true;
+}
+
+template <typename Symbol, typename Analysis, bool DevMode>
+auto VMExecutor<Symbol, Analysis, DevMode>::exec_iter_all_eclasses(Instruction instr) -> bool {
+  // E-graph maintains canonical e-class list - set in execute_impl()
```
colinbarry, line 426: Comment drift: set in `execute()`

src/planner/include/planner/pattern/vm/compiler.hpp
```diff
@@ -0,0 +1,886 @@
...
+
+template <typename Symbol>
+auto PatternCompiler<Symbol>::get_symbol_index(Symbol const &sym) -> uint8_t {
+  // TODO: maybe a strong type (so we know uint8_t is in relation to symbols_)
+  //       better name than `get_symbol_index`
+  //       Not a linear search
+  for (std::size_t i = 0; i < symbols_.size(); ++i) {
+    if (symbols_[i] == sym) return static_cast<uint8_t>(i);
+  }
+  symbols_.push_back(sym);
```
colinbarry, line 882: Might be worth explicitly checking for overflow here and raising an exception, like you do with `alloc_eclass_reg`, rather than silently returning the incorrect index.

src/planner/include/planner/pattern/vm/compiled_pattern.hpp
```diff
@@ -0,0 +1,119 @@
...
+namespace memgraph::planner::core::pattern::vm {
+
+/// Non-templated base class for CompiledPattern.
+///
+/// Contains all Symbol-independent state and accessors. This reduces template
+/// instantiation overhead and allows code that only needs bytecode/register
+/// metadata to work with a non-templated interface.
+///
+/// @see CompiledPattern for the full templated interface
+class CompiledPatternBase {
```
colinbarry, line 33: Wondering about the name here, as it is not a compiled pattern, but a complete compiled rule (or compiled patterns?)

colinbarry (commented):
Batching some change requests. More to come in the next batch!

src/query/plan_v2/egraph_converter.cpp
```diff
@@ -110,10 +110,13 @@ struct Builder {
     return std::make_shared<Produce>(input, std::vector{named_expression});
   }
 
-  auto Build(utils::tag_value<symbol::Symbol> /*tag*/, enode_ref /*node*/, children_ref /*children*/) -> BuildResult {
+  auto Build(utils::tag_value<symbol::Symbol> /*tag*/, enode_ref node, children_ref /*children*/) -> BuildResult {
     auto const frame_position = next_frame_position_++;
+    auto const sym_pos = static_cast<int32_t>(node.disambiguator());
+    auto const it = reverse_symbol_name_store_.find(sym_pos);
+    std::string name = (it != reverse_symbol_name_store_.end()) ? it->second : "unknown";
```
colinbarry, line 117: I don't believe in current code that this lookup can ever fail. But if it ever does, this would silently create a dummy signal, whereas below (line 126), we instead `throw QueryException{"Planner error, literal not found in store"};` when the lookup fails. I think the exception (or assert, seeing as if it did that would be an invariant violation) would be preferable.

src/planner/eclass.cppm
colinbarry, line 94: Maybe worth a `static_assert` if ever used with some non-trivial Analysis type to catch this at compile-time?

src/planner/enode.cppm
colinbarry, line 161: Code is correct,  but having the `rhs` and `lhs` swapped is surprising.

src/query/plan_v2/egraph_internal.hpp
colinbarry, line 61: `symbol_make_traits<S>` already specialises per symbol and knows exactly what arguments `make()` takes, so it implicitly encodes the same information as this. I wonder if it be better if this table be removed and the overload selection done directly through the traits, rather than splitting the logical between two places?

src/planner/include/planner/pattern/vm/compiler.hpp
```diff
@@ -0,0 +1,886 @@
...
+}
+
+// ============================================================================
+// Emission: top-level
+// ============================================================================
+
+template <typename Symbol>
+auto PatternCompiler<Symbol>::emit_patterns(std::span<Pattern<Symbol> const> patterns, EmitPlan const &plan)
+    -> InstrAddr {
+  InstrAddr innermost = kHaltPlaceholder;
```
colinbarry, line 598: Not sure `innermost` is a precise enough name, as this is actually the address of the target when iteration is exhausted or fails. `backtrack_addr` maybe?

src/planner/include/planner/pattern/vm/compiler.hpp
```diff
@@ -0,0 +1,886 @@
...
+  struct Candidate {
+    std::size_t depth;
+    PatternNodeId node;
+    std::vector<PathStep> ancestors;
+  };
+
+  std::optional<Candidate> symbol;
+  std::optional<std::pair<Candidate, PatternVar>> shared;
+
+  auto search = [&](auto &self, PatternNodeId node_id, std::vector<PathStep> &steps) -> void {
```
colinbarry, line 486: Think this is a candidate for `deducing this`. If this first argument was `this auto& self`, we wouldn't need to pass `self` at all the call sites.

src/planner/include/planner/extract/extractor.hpp
```diff
@@ -180,11 +150,24 @@ auto TopologicalSort(EGraph<Symbol, Analysis> const &egraph,
...
+/**
+ * @brief Extracts the lowest-cost expression tree from an e-graph
+ *
+ * The extraction result contains ENodeIds that are valid only as long as no
+ * rebuild() is performed on the e-graph. Use get_enode() to access e-node
+ * data immediately after extraction.
+ */
 template <typename Symbol, typename Analysis, typename CostModel>
 struct Extractor {
   Extractor(EGraph<Symbol, Analysis> const &egraph, CostModel &&cost_model)
```
colinbarry, line 162: Is taking `CostModel` by rvalue ref right here? It's enforcing the parameter being a sink parameter, but I think taking by value and allowing the caller to `std::move` or copy would be more idiomatic. (Doesn't make any difference in current code as this is only ever called with a `CostModel` temporary, but worth considering?)

src/planner/egraph.cppm
Ignition: good catch

src/query/plan_v2/egraph_converter.cpp
```diff
@@ -110,10 +110,13 @@ struct Builder {
     return std::make_shared<Produce>(input, std::vector{named_expression});
   }
 
-  auto Build(utils::tag_value<symbol::Symbol> /*tag*/, enode_ref /*node*/, children_ref /*children*/) -> BuildResult {
+  auto Build(utils::tag_value<symbol::Symbol> /*tag*/, enode_ref node, children_ref /*children*/) -> BuildResult {
     auto const frame_position = next_frame_position_++;
```
colinbarry, line 114: Rebuild assigns `Symbol.position_` by discovery order, rather than preserving the original position, so rebuilt AST `symbol_pos_` can diverge from the stored `SymbolTable` indices it is later resolved against. During execution, `symbol_table.at(...)` could retrieve the wrong symbol.

Probably worth adding a unit test for this situation to ensure symbols retain their identity after extraction.

colinbarry (commented):
Another pass done. Only critical issue, I think, is the symbol position potentially changing after rebuild. Everything else is suggestions, etc.

src/planner/include/planner/pattern/vm/compiler.hpp
```diff
@@ -0,0 +1,889 @@
...
+
+template <typename Symbol>
+auto PatternsCompiler<Symbol>::get_symbol_index(Symbol const &sym) -> uint8_t {
+  // TODO: maybe a strong type (so we know uint8_t is in relation to symbols_)
+  //       better name than `get_symbol_index`
+  //       Not a linear search
+  for (std::size_t i = 0; i < symbols_.size(); ++i) {
+    if (symbols_[i] == sym) return static_cast<uint8_t>(i);
+  }
+  if (symbols_.size() == std::numeric_limits<uint8_t>::max()) [[unlikely]] {
```
colinbarry: Off by one. Should be `> std::numeric_limits<uint8_t>::max()` (i.e, `> 255` rather than `== 255`).

colinbarry (approved):
Great stuff guys! 💪  Approved, with one note about a benign off-by-one error when it comes to the maximum allowed symbol count.

PR #3819, opened 2026-02-20
Add Properties.to_dict() and mgp.to_dict() for JSON serialization

### LINKED ISSUE #313: Implement mgp object serialization
It would be useful if all objects from [`mgp`](https://github.com/memgraph/memgraph/blob/master/include/mgp.py) are `JSON` serializable, that is, that there is some kind of method that returns Python `Dict` of `mgp` objects. For example, when I'm writing a [query module for database export](https://github.com/memgraph/mage/pull/100), I'm actually creating a `JSON` file with a list of nodes and relationships. Therefore, I have to reach every label, property, etc. of each `Vertex` and `Edge`. I would like to be able to just add properties as a list to my properties key, but then `mgp.Vertex.properties` is not `JSON` serializable. I solved this by going through all properties and creating new dictionary, but I think that the right place for this code would be inside mgp and not query module. Since we're going to create new utility modules similar to the export one, this feature may come in handy.

C-you-know:
Fixes #313 

This PR adds Properties.to_dict() and a module-level mgp.to_dict() helper to make vertex and edge properties JSON serializable without manual boilerplate.

Right now every query module that does any kind of export has to iterate through properties and handle temporal type conversion on its own. This puts that logic in mgp so it only needs to be written once.

Changes:
  - Properties.to_dict() on both mgp and mgp_mock
  - Module-level to_dict() helper for converting standalone values
  - Simplified json_util.py to use the new methods
  - 27 tests covering primitives, temporal types, nested structures,
    edge properties, empty dicts, and round-trip JSON serialization
    
Happy to iterate and improve based on any feedback or guidance from the maintainers. :)


C-you-know:
Gentle ping @gitbuda 🙂

PR #3820, opened 2026-02-23
fix: Repl state deadlock on replica

as51340:
## What

Fix a deadlock on replica data instances by acquiring the `repl_state` read lock at the very start of replication RPC handlers (`PrepareCommitHandler`, `WalFilesHandler`,
`CurrentWalHandler`) using a non-blocking `TryReadLock`. If the lock cannot be acquired (e.g., a main promotion holds the write lock), the handler returns a failure response
immediately instead of blocking. Also filters out `.old` directories from `GetFilesFromDir()` to prevent stale directory entries from interfering during file moves.

## Why

A deadlock could occur when a main promotion (which takes a write lock on `repl_state`) ran concurrently with replication handlers that acquired the `repl_state` read lock
*after* already holding other resources. By the time the handler reached the `ReadLock()` call on `repl_state` deep inside `LoadWal`, the lock ordering was violated, causing the
promotion and the handler to wait on each other indefinitely.

## How

- **`repl_state` lock hoisted to handler entry point**: Each of the three replication handlers (`PrepareCommitHandler`, `WalFilesHandler`, `CurrentWalHandler`) now calls
`repl_state.TryReadLock()` as its first action. The lock is held for the entire handler duration via `maybe_locked_repl_state`. `deltas_batch_progress_size` is read from the
locked state immediately, eliminating the mid-handler `ReadLock()` calls that existed before inside `LoadWal`.
- **`LoadWal` signature changed**: `LoadWal` no longer takes a `repl_state` reference. Instead it receives the pre-fetched `deltas_batch_progress_size` value, removing its need
to lock `repl_state` independently.
- **`RWSpinLock::try_lock_shared()` added** (`rw_spin_lock.hpp`): A non-blocking shared lock acquisition that atomically increments the reader count and backs off if a unique
lock is held. This enables the `TryReadLock` path on `Synchronized<..., RWSpinLock>`.
- **`TrySharedLockable` concept added** (`synchronized.hpp`): `TryReadLock` now requires `TrySharedLockable` instead of just `SharedMutex`, making the constraint precise.
- **`.old` directory filtering** (`file.cpp`): `GetFilesFromDir` now filters out entries named `.old` to avoid picking up transient directories created during file moves.

## Testing

No new tests were added. The deadlock is a race condition in the replication protocol that is difficult to reproduce deterministically in unit tests. The fix is a straightforward
 lock-ordering correction validated by code inspection. Existing replication jepsen, e2e and HA tests cover the affected handlers.

## Notes for reviewers

- The `TryReadLock` failure path sends a failure response (`PrepareCommitRes{false}`, `WalFilesRes{nullopt, 0}`, `CurrentWalRes{nullopt, 0}`)
- The `try_lock_shared()` implementation on `RWSpinLock` does a single atomic attempt with no spinning. If the unique lock bit is set, it immediately decrements the reader count
and returns `false`. This is intentional — spinning would reintroduce the deadlock window.
- The `.old` filtering in `GetFilesFromDir` is a global change affecting all callers

### DIFF
--- a/src/utils/file.cpp
+++ b/src/utils/file.cpp
@@ -110,7 +110,9 @@ auto GetFilesFromDir(std::filesystem::path const &dir) -> std::vector<std::files
   }
   std::error_code error_code;
   return std::filesystem::directory_iterator(dir, error_code) |
-         std::views::transform([](auto const &dir_entry) { return dir_entry.path(); }) | std::ranges::to<std::vector>();
+         std::views::transform([](auto const &dir_entry) { return dir_entry.path(); }) |
+         std::views::filter([](std::filesystem::path const &path) { return path.filename() != ".old"; }) |
+         std::ranges::to<std::vector>();
 }
 
 bool DeleteFile(const std::filesystem::path &file) noexcept {


as51340:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Fixed a deadlock on replica instances that could occur when a main promotion ran concurrently with replication data handlers (commit, WAL apply). The replication state lock is
  now acquired non-blockingly at the start of each handler — if the lock is unavailable (e.g., because a role transition is in progress), the handler returns a transient failure. This improves stability of high-availability clusters during failover, particularly under write load. No user action is
  required; the fix is transparent to existing replication configurations. [#3820](https://github.com/memgraph/memgraph/pull/3820)

PR #3829, opened 2026-02-24
testing: Fix Jepsen error msgs

as51340:
Fixes unhandled Jepsen error messages.

### DIFF
--- a/tests/jepsen/src/memgraph/high_availability/bank/test.clj
+++ b/tests/jepsen/src/memgraph/high_availability/bank/test.clj
@@ -209,7 +209,7 @@
                   (assoc op :type :info :value "SYNC replica is down")
 
                   (utils/cannot-get-shared-access? e)
-                  (assoc op :type :ok :value {:str "Cannot get shared access to the storage."})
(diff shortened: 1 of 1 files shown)


tests/jepsen/src/memgraph/high_availability/bank/test.clj
```diff
@@ -199,20 +199,20 @@
...
 
-                    (utils/sync-replica-down? e)
-                    (assoc op :type :info :value "SYNC replica is down")
+                  (utils/sync-replica-down? e)
+                  (assoc op :type :info :value "SYNC replica is down")
 
-                    (utils/main-unwriteable? e)
-                    (assoc op :type :info :value {:str "Cannot commit because main is currently non-writeable."})
+                  (utils/cannot-get-shared-access? e)
+                  (assoc op :type :ok :value {:str "Cannot get shared access to the storage."})
```
colinbarry: Should this be `:info` rather than `:ok`?

as51340: Being consistent is better, in terms of functionality it is the same at the end I just check there were no :fail replies

PR #3837, opened 2026-02-25
fix: Ensure GIL is taken before destroying Object

as51340:
## What

All special member functions and `FromBorrow` in `py::Object` (`src/py/py.hpp`) now acquire the Python GIL via `EnsureGIL` before calling any `Py_INCREF`/`Py_XDECREF`.
Previously, only callers were responsible for holding the GIL, which broke when `py::Object` was destroyed on a C++ worker thread that had no awareness of Python. A regression
test and a GIL-requirement comment on `ProcessFileDependencies` are also included.

## Why

When a Cypher query uses a Python user-defined function, `UserFunction()` deep-copies the `mgp_func` (including the lambda capturing a `py::Object`) into the `Function` AST node.
 After the query completes, `Interpreter::Pull` resets the `QueryExecution` on a `PriorityThreadPool` worker thread, destroying the AST and ultimately calling
`py::Object::~Object()` → `Py_XDECREF` without the GIL, causing a segfault inside `libpython`.

## How

- Every `py::Object` method that manipulates Python refcounts (`FromBorrow`, destructor, copy constructor, copy assignment, move assignment) now guards with `if (ptr_ &&
!_Py_IsFinalizing()) { EnsureGIL gil; ... }` before calling `Py_INCREF`/`Py_XDECREF`.
- The `_Py_IsFinalizing()` check prevents `PyGILState_Ensure` from terminating the thread during Python shutdown, as documented in the `EnsureGIL` class comment.
- `PyGILState_Ensure` is safe to call when the GIL is already held (it bumps an internal counter), so all existing call sites are unaffected.
- The move constructor is left unprotected — it only transfers the pointer without touching Python refcounts.
- Copy assignment acquires a single `EnsureGIL` covering both `Py_XDECREF` (old) and `Py_XINCREF` (new) to avoid acquiring the GIL twice.

## Testing

Added `TEST(PyObjectLifecycle, DestructorAcquiresGIL)` in `tests/unit/query_procedure_py_module.cpp`. It creates a `py::Object` wrapping a `PyCapsule` while holding the GIL, then
 destroys it on a `std::jthread` that never held the GIL. Without the fix, this segfaults in `Py_DECREF` during capsule deallocation; with the fix, the destructor acquires the
GIL internally and the test passes.

## Notes for reviewers

- The deep copy of `mgp_func` into the AST (via `UserFunction` in `awesome_memgraph_functions.cpp:2038`) is intentional — it makes the query independent of module reloads. The
fix belongs in `py::Object`, not in the copy path.
- `PythonModule::Close()` (`module.cpp:1025-1033`) already acquired the GIL before clearing its maps. The bug only manifested because the deep copy escapes `PythonModule` into
the AST, where destruction happens on an unrelated worker thread.
- All other direct `Py_DECREF`/`Py_INCREF` calls in the codebase (in `memgraph.cpp`, `module.cpp`, `py_module.cpp`) were audited and are already within GIL-held contexts (Python
`tp_dealloc`, module init, or explicit `EnsureGIL`).

### DIFF
--- a/tests/unit/query_procedure_py_module.cpp
+++ b/tests/unit/query_procedure_py_module.cpp
@@ -340,6 +340,93 @@ TEST(PyObjectLifecycle, DestructorAcquiresGIL) {
   }
 }
 
+TEST(PyObjectLifecycle, FromBorrowGIL) {
+  std::optional<memgraph::py::Object> py_obj;
+  {
+    auto gil = memgraph::py::EnsureGIL();
+    static int dummy;
+    py_obj.emplace(PyCapsule_New(&dummy, "gil_test", GILCheckCapsuleDestructor));
+    ASSERT_TRUE(*py_obj);
+  }
+  // GIL is released. Destroy on a thread that never held the GIL.
+  {
+    std::jthread t([&py_obj]() { auto b = memgraph::py::Object::FromBorrow(py_obj.value().Ptr()); });
+  }
+}
+
+TEST(PyObjectLifecycle, CopyAssignGIL) {
+  std::optional<memgraph::py::Object> py_obj;
+  {
+    auto gil = memgraph::py::EnsureGIL();
+    static int dummy;
+    py_obj.emplace(PyCapsule_New(&dummy, "gil_test", GILCheckCapsuleDestructor));
+    ASSERT_TRUE(*py_obj);
+  }
+  {
+    std::jthread t([&py_obj]() {
+      memgraph::py::Object b;
+      {
+        auto gil = memgraph::py::EnsureGIL();
+        static int dummy2;
+        b = memgraph::py::Object(PyCapsule_New(&dummy2, "old_capsule", GILCheckCapsuleDestructor));
+      }
+      // b holds a PyCapsule with refcount 1, GIL released
+      b = py_obj.value();  // copy assignment: DECREF on old capsule → refcount 0 → _Py_Dealloc without GIL → crash
+    });
+  }
+}
+
+TEST(PyObjectLifecycle, CopyCtrGIL) {
+  std::optional<memgraph::py::Object> py_obj;
+  {
+    auto gil = memgraph::py::EnsureGIL();
+    static int dummy;
+    py_obj.emplace(PyCapsule_New(&dummy, "gil_test", GILCheckCapsuleDestructor));
+    ASSERT_TRUE(*py_obj);
+  }
+  // GIL is released. Destroy on a thread that never held the GIL.
+  {
+    std::jthread t([&py_obj]() { auto b = py_obj.value(); });
+  }
+}
+
+TEST(PyObjectLifecycle, MoveCtrGIL) {
+  std::optional<memgraph::py::Object> py_obj;
+  {
+    auto gil = memgraph::py::EnsureGIL();
+    static int dummy;
+    py_obj.emplace(PyCapsule_New(&dummy, "gil_test", GILCheckCapsuleDestructor));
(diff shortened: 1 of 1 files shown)


as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Fixed a crash that could occur when queries using Python user-defined functions were cleaned up on internal worker threads. The crash happened because Python's Global Interpreter
   Lock (GIL) was not held when releasing references to Python callback objects during query teardown. Memgraph now correctly acquires the GIL before manipulating Python reference
  counts in all code paths, preventing the segfault. No user action is required; upgrading to this version resolves the issue for anyone experiencing intermittent crashes when
  using Python query modules that register custom functions. [#3837](https://github.com/memgraph/memgraph/pull/3837)

tests/unit/query_procedure_py_module.cpp
```diff
@@ -316,6 +318,28 @@ TYPED_TEST(PyModule, PyObjectToMgpValue) {
...
 
+namespace {
+void GILCheckCapsuleDestructor(PyObject * /*capsule*/) {}
+}  // namespace
+
+// Regression test: py::Object must acquire the GIL before calling Py_XDECREF,
+// even when destroyed on a thread that doesn't hold the GIL.
+// Uses a PyCapsule with a C-level destructor that checks PyGILState_Check()
+// at the exact moment of deallocation.
+TEST(PyObjectLifecycle, DestructorAcquiresGIL) {
```
andrejtonev, line 329: Add tests for constructor, copy, move, etc

src/py/py.hpp
```diff
@@ -69,27 +69,48 @@ class [[nodiscard]] Object final {
...
   Object(Object &&other) noexcept : ptr_(other.ptr_) { other.ptr_ = nullptr; }
 
   Object &operator=(const Object &other) noexcept {
     if (this == &other) return *this;
-    Py_XDECREF(ptr_);
-    ptr_ = other.ptr_;
-    Py_XINCREF(ptr_);
+    if (!_Py_IsFinalizing()) {
+      EnsureGIL gil;
+      Py_XDECREF(ptr_);
```
andrejtonev, line 99: if ptr_

src/py/py.hpp
```diff
@@ -69,27 +69,48 @@ class [[nodiscard]] Object final {
...
+      ptr_ = other.ptr_;
+      Py_XINCREF(ptr_);
```
andrejtonev, line 101: if ptr_

src/py/py.hpp
```diff
@@ -69,27 +69,48 @@ class [[nodiscard]] Object final {
...
+    if (ptr_ && !_Py_IsFinalizing()) {
+      EnsureGIL gil;
+      Py_INCREF(ptr_);
+    }
+  }
 
   Object(Object &&other) noexcept : ptr_(other.ptr_) { other.ptr_ = nullptr; }
 
   Object &operator=(const Object &other) noexcept {
     if (this == &other) return *this;
```
andrejtonev, line 96: or ptr_ == other.ptr_?

src/py/py.hpp
as51340: no need, xdecref won't fail even ptr_ is null

src/py/py.hpp
as51340: no need, xincref won't fail even ptr_ is null

src/py/py.hpp
as51340: I think usually this is checked?

PR #3842, opened 2026-02-26
Fix: Python module close can segfault on shutdown

andrejtonev:


andrejtonev:
BugBot run

colinbarry:
BugBot run

andrejtonev:
@cursor push a8672e17f6

src/py/py.hpp
```diff
@@ -32,14 +32,19 @@
 #define Py_Is(x, y) ((x) == (y))
 #endif
 
+// Python 3.13 compatibility macros
+#if PY_VERSION_HEX < 0x030d0000 && !defined(Py_IsFinalizing)
+#define Py_IsFinalizing() _Py_IsFinalizing()
+#endif
+
```
as51340, line 39: Not sure if this is needed. Don't we force users to always use some specific Python version? Or for modules we always use same Python version?

andrejtonev: I had 3.13 locally, so just future proofed

src/query/procedure/module.cpp
```diff
@@ -1025,16 +1025,33 @@ bool PythonModule::Load(const std::filesystem::path &file_path) {
...
+  if (Py_IsFinalizing()) {
+    procedures_.clear();
+    transformations_.clear();
+    functions_.clear();
+    py_module_ = py::Object(nullptr);
+    return true;
+  }
+
   auto gil = py::EnsureGIL();
+
```
colinbarry, line 1042: The race window is narrowed, but there is still a window between us checking `Py_IsFinalizing()` and calling `PyGILState_Ensure` via `py::EnsureGIL`. 

I think the totally robust solution is our own managed synchronisation, so that we can atomically:
- Set some flag before we call `Py_Finalize` under a write on a shared mutex
- When we call `py::EnsureGIL`, we need to hold the mutex as read whilst we hold the GIL. Mutex would need to be held until we release the GIL.

Or we could do this with an `atomic<int>` counter if you wanted something lighter than a mutex?

colinbarry: Forget that: I think we're safe because by the time we `UnloadAllModules` all worker threads have joined. The fix is correct. Ignore my above comment.

andrejtonev:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Fix: Python module shutdown segfault. Added a missing python api state check that prevents a segfault while shutting down. [#3842](https://github.com/memgraph/memgraph/pull/3842)

PR #3844, opened 2026-02-27
feat: Add support for asynchronous logging 

as51340:
## What

Add support for asynchronous logging via a new `--logger-type` flag. When set to `async`, Memgraph uses spdlog's `async_logger` backed by a dedicated thread pool instead of the
default synchronous logger. The flag files are renamed from `log_level.{cpp,hpp}` to `logging.{cpp,hpp}` to better reflect their broadened scope.

## Why

Synchronous logging on hot paths can become a bottleneck under high throughput because every log call blocks until the message is written to the sink. An asynchronous logger
decouples the caller from I/O, reducing latency for write-heavy workloads.

## How

- **New flag `--logger-type`** (`sync` | `async`, default `sync`): defined with `DEFINE_VALIDATED_string` in `src/flags/logging.cpp`. Validation rejects any value other than the
two allowed strings (case-insensitive).
- **Async logger creation**: when `--logger-type=async`, `InitializeLogger()` initialises an spdlog thread pool (8192-slot ring buffer, 1 background thread) and creates an
`spdlog::async_logger` with `overrun_oldest` overflow policy. The `dist_sink_mt` distribution sink is shared identically between both paths.
- **File rename**: `log_level.{cpp,hpp}` → `logging.{cpp,hpp}`; all includes and CMake source lists updated accordingly.
- **Namespace cleanup**: replaced explicit `memgraph::flags::` qualifiers with a wrapping `namespace memgraph::flags { … }` block.

## Testing

No new tests were added. The change is behind a new flag that defaults to the existing synchronous behaviour, so all current tests exercise the unmodified code path. Manual
testing of `--logger-type=async` is recommended. A follow-up could add a unit test that verifies the default logger type after `InitializeLogger()` with each flag value.

## Notes for reviewers

- The `overrun_oldest` overflow policy silently drops the oldest queued message when the 8192-slot ring buffer is full. Under extreme log pressure this means log lines can be
lost. Consider whether `block` would be a safer default, or whether the buffer size should be configurable.
- `flush_on(spdlog::level::trace)` is set on the async logger as well, which triggers a flush after every message — this partially negates the async benefit. Reviewers should
decide whether a less aggressive flush policy (e.g. `flush_on(spdlog::level::warn)` or periodic flush) is more appropriate.
- No durability version implications; the flag only affects runtime logger initialisation.

### DIFF
--- a/tests/e2e/configuration/default_config.py
+++ b/tests/e2e/configuration/default_config.py
@@ -92,6 +92,11 @@ startup_config_dict = {
         "",
         "List of default Kafka brokers as a comma separated list of broker host or host:port.",
     ),
+    "logger_type": (
+        "sync",
+        "sync",
+        "Controls whether synchronous or asynchronous logger will be used. Options: sync, async",
+    ),
     "log_file": ("", "", "Path to where the log should be stored."),
     "nuraft_log_file": ("", "", "Path to the file where NuRaft logs are saved."),
     "log_level": (


as51340:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
  - Memgraph now supports asynchronous logging via a new `--logger-type` flag that accepts `sync` (default) or `async`. When set to `async`, log messages are handed off to a
  background thread instead of being written inline, reducing latency on hot paths under high throughput. No action is required — the default behaviour is unchanged; to opt in,
  start Memgraph with `--logger-type=async`. [#3844](https://github.com/memgraph/memgraph/pull/3844
- [x] **[Documentation PR link memgraph/documentation](https://github.com/memgraph/documentation/pull/1549)**

PR #3850, opened 2026-03-02
fix: Data race in coordinator state manager

as51340:
## What

Fix a data race in `CoordinatorStateManager` on `cluster_config_` by adding a `utils::RWSpinLock` (`config_mutex_`) to protect concurrent access from `load_config()`,
`save_config()`, and `GetCoordinatorInstancesAux()`. The previously unprotected `cluster_config_` shared pointer could be read and written simultaneously from NuRaft internal
threads and user-facing query threads.

## Why

TSAN-instrumented HA e2e tests revealed a data race on `CoordinatorStateManager::cluster_config_`. NuRaft calls `save_config()` from its commit/election thread while
`load_config()` and `GetCoordinatorInstancesAux()` can be invoked concurrently from other threads, resulting in unsynchronized reads and writes to the same `shared_ptr`.

## How

- **`config_mutex_` (`utils::RWSpinLock`)** added as a `mutable` member on `CoordinatorStateManager`. `save_config()` and `load_config()` acquire it exclusively
(`std::lock_guard`); `GetCoordinatorInstancesAux()` acquires it in shared mode when reading.

- **`GetCoordinatorInstancesAux` templatized** with a `bool LockNeeded` parameter (default `true`). The `<false>` variant is called from `save_config()` where the exclusive lock
is already held, avoiding double-locking. The method is moved from the `.cpp` to the header to support the template.

- **`TryUpdateClusterConfigFromDisk()`** no longer acquires its own lock — the caller is responsible. Both call sites (`load_config` and the constructor) already hold the
appropriate lock or are in single-threaded context.

## Testing

No new automated tests were added. The race was identified and verified fixed using TSAN-instrumented HA e2e test runs. Existing unit tests in
`tests/unit/coordinator_state_manager.cpp` cover single-threaded `load_config`/`save_config` round-trips. A concurrent stress test (multiple threads calling `load_config`,
`save_config`, and `GetCoordinatorInstancesAux` simultaneously under TSAN) would strengthen coverage.

## Notes for reviewers

- The `.gitignore` change (`CLAUDE.md`) is unrelated tooling noise — consider dropping it from the final PR.
- No durability version bump needed; no new persistent state is introduced.
- Verify that no other callers of `GetCoordinatorInstancesAux` exist outside the lock scope — currently the only `<false>` call site is inside `save_config` under the exclusive
lock.

### DIFF
--- a/src/coordination/coordinator_state_manager.cpp
+++ b/src/coordination/coordinator_state_manager.cpp
@@ -116,6 +116,7 @@ CoordinatorStateManager::CoordinatorStateManager(CoordinatorStateManagerConfig c
   cluster_config_->get_servers().push_back(my_srv_config_);
 
   HandleVersionMigration();
+  auto lock = std::lock_guard{config_mutex_};
   TryUpdateClusterConfigFromDisk();
 }
 


src/coordination/coordinator_state_manager.cpp
```diff
@@ -143,32 +143,14 @@ void CoordinatorStateManager::TryUpdateClusterConfigFromDisk() {
 
 // Called when application is starting up
 auto CoordinatorStateManager::load_config() -> std::shared_ptr<cluster_config> {
-  spdlog::trace("Got request to update config from disk");
+  auto lock = std::lock_guard{config_mutex_};
```
andrejtonev, line 147: CoordinatorStateManager constructors call TryUpdateClusterConfigFromDisk without lock

src/coordination/include/coordination/coordinator_state_manager.hpp
```diff
@@ -64,6 +96,7 @@ class CoordinatorStateManager final : public state_mgr {
 
   void TryUpdateClusterConfigFromDisk();
 
+  mutable utils::RWSpinLock config_mutex_;
```
andrejtonev, line 99: Could we start annotating these locks with clang thread-safety?
Too many locks to look out for

as51340: I think we should yes

as51340:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - A data race in the high-availability coordinator's cluster configuration handling has been fixed by adding proper synchronization around concurrent reads and writes of the Raft
  cluster config; previously, simultaneous config loads and saves could produce inconsistent state, potentially causing coordinator instability during cluster membership changes.
  No user action is required — the fix is internal and automatic upon upgrade. [#3850](https://github.com/memgraph/memgraph/pull/3850)

PR #3856, opened 2026-03-03
fix: Concurrent workloads on vector indices

as51340:
## What

Replace the coarse `std::shared_mutex`-based locking in both vertex and edge vector indices with a 3-tier `utils::ResourceLock` (READ_ONLY / WRITE / UNIQUE). This enables concurrent writes via the
 WRITE tier while reserving UNIQUE for remove/resize operations. The GC cleanup strategy is changed from a full index scan (`RemoveObsoleteEntries`) to targeted removal (`RemoveVertices` /
`RemoveEdges`) called before skip-list entries are freed, and the serialization path now extracts data under a read lock and writes outside the lock.

## Why

Can't rely on deleted() solely to determine whether the key should be deleted and usearch doesn't allow for concurrent inserts and searches.

## How

- **`vector_index_utils.hpp`**: Moved `mg_vector_index_t` and `synchronized_mg_vector_index_t` (with `ResourceLock`) into this shared header. The `UpdateVectorIndex` template is generalized over
`SyncIndex` so both vertex and edge sync wrappers dispatch through the same happy-path (WRITE lock for adds) / hard-path (UNIQUE lock for remove+resize) logic. 

- **`vector_index.cpp`**: All `ReadLock()`/`MutableSharedLock()`/`Lock()` calls replaced with `SharedResourceLockGuard(READ_ONLY)` / `SharedResourceLockGuard(WRITE)` / `std::lock_guard` (UNIQUE).
`RemoveObsoleteEntries` replaced with `RemoveVertices(vector<Vertex*>)` using a two-phase approach: READ_ONLY to find which vertices exist in each index, then UNIQUE only on indices that need
removal. Serialization extracts keys+vectors under READ_ONLY into a local buffer, then writes to the encoder outside the lock.

- **`vector_edge_index.cpp`**: Same `ResourceLock` treatment. Defines a local `synchronized_mg_vector_edge_index_t` mirroring the vertex struct. `RemoveObsoleteEntries` replaced with
`RemoveEdges(list<Gid>)` — a two-phase approach that exports keys under READ_ONLY, filters by GID, then removes under UNIQUE. Fixed a bug where `GetVectorFromEdge` was taking an exclusive lock
instead of READ_ONLY.

- **`indices.cpp` / `indices.hpp`**: Removed `RemoveObsoleteEntries` calls from the periodic GC path. Added `RemoveVerticesFromVectorIndices` and `RemoveEdgesFromVectorEdgeIndices` as targeted GC
entry points.

- **`storage.cpp` (GC flow)**: Targeted removal is called in four places — both the transactional path (`current_deleted_vertices` / `current_deleted_edges` before skip-list removal) and the
`IN_MEMORY_ANALYTICAL` full-scan path (`need_full_scan_vertices` / `need_full_scan_edges` before skip-list removal). All guarded by `Empty()` checks to avoid unnecessary work when no vector
indices exist.

## Testing

**Unit tests** (`tests/unit/vector_edge_index.cpp`):
- Added `VectorEdgeIndexGCTest` fixture with periodic GC (100ms interval).
- `AnalyticalModeDeleteEdgeGCCleansVectorIndex`: Creates edge in transactional mode, switches to analytical, deletes, switches back (triggering GC), verifies index is cleaned.
- `AnalyticalModeDeleteMultipleEdgesGCCleansVectorIndex`: Deletes 2 of 3 edges in analytical mode, verifies only the surviving edge remains after GC.
- `TransactionalModeDeleteEdgeGCCleansVectorIndex`: Deletes edge in transactional mode, waits for periodic GC, verifies cleanup.
- Updated existing `RemoveEntriesTest` to use the new `RemoveEdges` API.

**Stress tests** (new workloads):
- `vector_index_concurrent_ops`: HA stress test with concurrent create/delete operations on vertex vector indices across multiple tenants, followed by restart and replica consistency checks.
- `vector_edge_index_recovery`: Same pattern for edge vector indices.

### DIFF
--- a/src/storage/v2/indices/vector_index_utils.hpp
+++ b/src/storage/v2/indices/vector_index_utils.hpp
@@ -375,7 +375,10 @@ void UpdateVectorIndex(SyncIndex &mg_index, Spec &spec, const Key &key, const ut
   auto guard = std::lock_guard{mg_index.mutex};
   if (mg_index.index.contains(key)) {  // check again freshly if index contains key
     mg_index.index.remove(key);
+  }
 
+  // Try to add without resizing (another thread may have already resized)
+  {
     auto result = mg_index.index.add(key, vector.data(), thread_id_for_adding);
     if (!result.error) {
       return;


as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
   - Concurrent operations on vector indexes (both node and edge) are now significantly less contention-prone, as the internal locking has been reworked to allow multiple writers in parallel while
  reserving exclusive access only for resize and removal operations. Garbage collection of deleted entries in vector indexes is now targeted rather than scanning the entire index, which improves GC
  performance and fixes a potential crash when edges or vertices were deleted in analytical storage mode and then garbage-collected before the vector index was cleaned up. No user action is
  required; existing vector indexes benefit automatically after upgrading. [#3856](https://github.com/memgraph/memgraph/pull/3856

as51340:
Check UpdateOnRemoveLabel if you need locked_index->get

as51340:
Check also if you need at other places pinned accessor

as51340:
Check search nodes

imilinovic:
bugbot run

src/storage/v2/indices/vector_index.cpp
```diff
@@ -260,13 +258,13 @@ void VectorIndex::UpdateOnRemoveLabel(LabelId label, Vertex *vertex, const Index
       ids.erase(ranges::remove(ids, index_id), ids.end());
       auto &index_item = pimpl->index_by_id_.at(index_id);
 
-      auto locked_index = index_item.mg_index.MutableSharedLock();
+      auto locked_index = index_item.mg_index.Lock();
```
as51340: is this now unique lock on utils::Synchronised or usearch's unique lock

DavIvek:
still crashes when testing with https://github.com/memgraph/memgraph/pull/3857

src/storage/v2/inmemory/storage.cpp
```diff
@@ -2935,6 +2932,20 @@ void InMemoryStorage::CollectGarbage(std::unique_lock<utils::ResourceLock> main_
...
+      // Remove from vector indices BEFORE skip list removal while Vertex* is still valid.
+      auto const vertices_to_remove = current_deleted_vertices |
+                                      std::ranges::views::transform([&vertex_acc](auto const gid) {
+                                        auto it = vertex_acc.find(gid);
+                                        DMG_ASSERT(it != vertex_acc.end(), "Invalid database state!");
+                                        return &*it;
+                                      }) |
+                                      std::ranges::to<std::vector>();
+
+      indices_.RemoveVerticesFromVectorIndices(vertices_to_remove);
```
DavIvek, line 2949: No need to do this if there is no vector index in the system

src/storage/v2/indices/vector_edge_index.cpp
```diff
@@ -245,26 +251,26 @@ std::optional<uint64_t> VectorEdgeIndex::ApproximateEdgesVectorCount(EdgeTypeId
     return std::nullopt;
   }
   auto &[mg_index, _] = it->second;
-  auto locked_index = mg_index.ReadLock();
-  return locked_index->size();
+  auto guard = utils::SharedResourceLockGuard(mg_index.mutex, utils::SharedResourceLockGuard::READ_ONLY);
```
andrejtonev, line 254: isn't this a unique access operation?

src/storage/v2/indices/vector_edge_index.cpp
```diff
@@ -290,27 +296,39 @@ void VectorEdgeIndex::RestoreEntries(
...
-      locked_index->remove(entry);
+
+    // Phase 1: READ_ONLY — export keys and find entries matching deleted GIDs
+    std::vector<EdgeIndexEntry> entries_to_remove;
+    {
+      auto guard = utils::SharedResourceLockGuard(mg_index.mutex, utils::SharedResourceLockGuard::READ_ONLY);
+      auto const index_size = mg_index.index.size();
+      if (index_size == 0) continue;
+      std::vector<EdgeIndexEntry> all_entries(index_size);
+      mg_index.index.export_keys(all_entries.data(), 0, index_size);
```
andrejtonev, line 315: did we conclude that the export keys is broken?

src/storage/v2/indices/vector_edge_index.cpp
andrejtonev: exclusive against add/delete, so read only is fine

src/storage/v2/indices/vector_edge_index.cpp
```diff
@@ -290,27 +296,39 @@ void VectorEdgeIndex::RestoreEntries(
...
 }
 
-void VectorEdgeIndex::RemoveObsoleteEntries(std::stop_token token) const {
-  auto maybe_stop = utils::ResettableCounter(2048);
+bool VectorEdgeIndex::Empty() const { return pimpl->edge_index_.empty(); }
+
+void VectorEdgeIndex::RemoveEdges(std::list<Gid> const &deleted_edge_gids) const {
+  std::unordered_set<uint64_t> gids_to_remove;
+  gids_to_remove.reserve(deleted_edge_gids.size());
+  for (auto const &gid : deleted_edge_gids) {
```
andrejtonev: check if there is an insert (begin, end) from a list

src/storage/v2/indices/vector_edge_index.cpp
```diff
@@ -290,27 +296,39 @@ void VectorEdgeIndex::RestoreEntries(
...
+        }
+      }
+    }
+
+    if (entries_to_remove.empty()) continue;
+
+    // Phase 2: UNIQUE — remove matched entries
+    auto guard = std::lock_guard{mg_index.mutex};
+    for (auto const &entry : entries_to_remove) {
+      mg_index.index.remove(entry);
```
andrejtonev: use remove_all

src/storage/v2/indices/vector_index.cpp
```diff
@@ -262,18 +258,20 @@ void VectorIndex::UpdateOnRemoveLabel(LabelId label, Vertex *vertex, const Index
...
-          locked_index->get(vertex, vector.data());
+          std::vector<double> vector(index_item.mg_index.index.dimensions());
+          if (!index_item.mg_index.index.get(vertex, vector.data())) return PropertyValue();
           return PropertyValue(std::move(vector));
         }
         return old_vertex_property_value;
       });
-      locked_index->remove(vertex);
+      auto remove_result = index_item.mg_index.index.remove(vertex);
+      MG_ASSERT(remove_result.completed, "Vertex not found in vector index during label removal");
```
andrejtonev: move this do DMG_ASSERT or a throw or just ignore failure

src/storage/v2/indices/vector_index.cpp
```diff
@@ -370,26 +370,30 @@ void VectorIndex::SerializeAllVectorIndices(durability::BaseEncoder *encoder,
...
+      auto const size = mg_index.index.size();
+      if (size == 0) return {};
+
+      std::vector<Vertex *> keys(size);
+      mg_index.index.export_keys(keys.data(), 0, size);
+
+      std::vector<Entry> result;
+      std::vector<float> buffer(mg_index.index.dimensions());
+      for (auto *vertex : keys) {
+        if (vertex->deleted()) continue;
```
andrejtonev: need to check in vertex == nullptr

src/storage/v2/indices/vector_index.cpp
```diff
@@ -370,26 +370,30 @@ void VectorIndex::SerializeAllVectorIndices(durability::BaseEncoder *encoder,
...
 
-    std::vector<float> buffer(dimension);
-    for (auto *vertex : vertices) {
-      encoder->WriteUint(vertex->gid.AsUint());
-      locked_index->get(vertex, buffer.data());
-      std::ranges::for_each(buffer, [&](auto value) { encoder->WriteDouble(value); });
+    encoder->WriteUint(entries.size());
+    for (const auto &[gid, vector] : entries) {
+      encoder->WriteUint(gid);
+      for (auto value : vector) encoder->WriteDouble(value);
```
andrejtonev, line 395: don't we have a WriteList

src/storage/v2/indices/vector_index.cpp
```diff
@@ -433,32 +437,37 @@ VectorIndex::VectorSearchNodeResults VectorIndex::SearchNodes(std::string_view i
...
-    // size() and export_keys() are not atomic — a concurrent add/remove can cause
-    // size() > slot_lookup_.size(), leaving trailing nullptr entries in the buffer.
-    auto deleted =
-        vertices_to_remove | rv::filter([](const Vertex *vertex) { return vertex != nullptr && vertex->deleted(); });
-    for (const auto &vertex : deleted) {
-      locked_index->remove(vertex);
+    // read only to check which vertices should be removed in that index
+    auto guard = utils::SharedResourceLockGuard(mg_index.mutex, utils::SharedResourceLockGuard::READ_ONLY);
+
+    auto index_vertices =
```
andrejtonev: lets have both the edge and vertex variants use the same logic

src/storage/v2/indices/vector_index_utils.hpp
```diff
@@ -326,42 +339,67 @@ inline std::size_t GetVectorIndexThreadCount() {
...
-    // In order to resize the index, we need to acquire an exclusive lock.
-    auto exclusively_locked_index = mg_index.Lock();
-    if (exclusively_locked_index->size() >= exclusively_locked_index->capacity()) {
-      const auto new_size = static_cast<std::size_t>(spec.resize_coefficient * exclusively_locked_index->capacity());
-      const unum::usearch::index_limits_t new_limits(new_size, GetVectorIndexThreadCount());
-      if (!exclusively_locked_index->try_reserve(new_limits)) {
-        throw query::VectorSearchException("Failed to resize vector index.");
-      }
-      spec.capacity = exclusively_locked_index->capacity();
+    auto result = mg_index.index.add(key, vector.data(), thread_id_for_adding);
```
andrejtonev, line 382: this can only succeed without resizing if the key already existed, so put it under the if {}

src/storage/v2/inmemory/storage.cpp
```diff
@@ -2966,6 +2996,20 @@ void InMemoryStorage::CollectGarbage(std::unique_lock<utils::ResourceLock> main_
   // EXPENSIVE full scan, is only run if an IN_MEMORY_ANALYTICAL transaction involved any deletions
   if (need_full_scan_edges) {
     auto edge_acc = edges_.access();
+
+    if (!indices_.vector_edge_index_.Empty()) {
+      // Collect deleted edge GIDs BEFORE skip-list removal while Edge* is still valid.
+      std::list<Gid> analytical_deleted_edge_gids;
```
andrejtonev, line 3002: use unorederd_set

src/storage/v2/indices/vector_index.cpp
```diff
@@ -433,32 +437,37 @@ VectorIndex::VectorSearchNodeResults VectorIndex::SearchNodes(std::string_view i
...
   }
+
+  for (auto &[index_id, loc_vertices_to_remove] : index_to_vertices) {
+    auto index_it = pimpl->index_by_id_.find(index_id);
+    DMG_ASSERT(index_it != pimpl->index_by_id_.end());
+    // Take unique lock for removing
+    auto guard = std::lock_guard{index_it->second.mg_index.mutex};
+    auto &index = index_it->second.mg_index.index;
+    index.remove(loc_vertices_to_remove.begin(), loc_vertices_to_remove.end());
+  }
```
andrejtonev, line 469: if you already have a list of vertex* (which are keys in the index), can't you just lock and call remove_all?
make sure remove_all supports keys that don't exit

src/storage/v2/indices/vector_edge_index.cpp
as51340: no, bug on our side

src/storage/v2/indices/vector_index.cpp
as51340: Nope AFAICT

src/storage/v2/inmemory/storage.cpp
as51340: my bad on the meet, has to be list because current_deleted_edges is list<gid>

src/storage/v2/indices/vector_index.cpp
as51340: yep, will be the same logic now

PR #3858, opened 2026-03-03
Add env var to stress tests

Josipmrden:
Add env vars to stress tests.

### DIFF
--- a/.github/workflows/reusable_stress_tests.yaml
+++ b/.github/workflows/reusable_stress_tests.yaml
@@ -22,6 +22,11 @@ on:
         required: false
         type: string
         default: ''
+      publications_dataset_path:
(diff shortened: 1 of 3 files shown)


.github/workflows/reusable_stress_tests.yaml
mattkjames7, line 381: We should probably add this as an option to `mgbuild.sh` e.g.
```
./release/package/mgbuild.sh \
          --enterprise-license $MEMGRAPH_ENTERPRISE_LICENSE \
          --organization-name $MEMGRAPH_ORGANIZATION_NAME \
          test-memgraph stress-eks-ha --publications-dataset-path "$PUBLICATIONS_DATASET_PATH"
```
Then add to the `print_help()` function so that it is documented, rather than relying on an exported environmental variable. We can export it before running the `continuous_integration` script within `mgbuild.sh`

Josipmrden:
Will re-do later

PR #3862, opened 2026-03-04
feat: Add configuration flag --log-retention-days

as51340:
## What

Introduces a new `--log-retention-days` configuration flag that controls how many days daily log files are preserved. The previously hard-coded retention count of 35 is now a
user-configurable `gflags` parameter with a validated range of 1–max(uint16_t)

## Why

Log retention was hard-coded to 35 days with no way to adjust it at runtime or deployment time. Users running Memgraph in environments with stricter storage constraints or
compliance requirements had no control over how long log files were kept.

## How

- Replaced the `inline constexpr auto log_retention_count = 35` constant in `src/flags/logging.cpp` with a `DEFINE_VALIDATED_uint64(log_retention_days, ...)` flag
- The `daily_file_sink_st` constructor call now uses `FLAGS_log_retention_days` instead of the old constant.
- Minor reformatting of the `DEFINE_VALIDATED_string` for `logger_type` (namespace qualification of `utils::ToLowerCase` → `memgraph::utils::ToLowerCase` since the flag
definition sits outside `namespace memgraph::flags`).

## Testing

- Updated `tests/e2e/configuration/default_config.py` to include `log_retention_days` with its expected default value (`"35"`), ensuring the default-config e2e test validates the
 new flag.
- No new unit tests added — the flag uses `DEFINE_VALIDATED_uint64` with `FLAG_IN_RANGE`, which is a well-tested gflags validation macro, and the downstream behaviour (spdlog
daily sink retention) is third-party code.

### DIFF
--- a/src/flags/logging.cpp
+++ b/src/flags/logging.cpp
@@ -60,7 +60,7 @@ DEFINE_VALIDATED_string(logger_type, "sync",
 // default set to 35 days
 // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
 DEFINE_VALIDATED_uint64(log_retention_days, 35, "Controls for how many days will daily log files be preserved.",
-                        FLAG_IN_RANGE(1, std::numeric_limits<uint64_t>::max()));
+                        FLAG_IN_RANGE(1, std::numeric_limits<uint16_t>::max()));
 
 inline constexpr std::array log_level_mappings{std::pair{"TRACE"sv, spdlog::level::trace},
                                                std::pair{"DEBUG"sv, spdlog::level::debug},


as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Memgraph now exposes a --log-retention-days configuration flag that controls how many days daily log files are kept before being automatically removed; the default remains 35
  days (5 weeks), matching the previous hard-coded behavior, so no action is needed unless you want to customize retention by setting the flag to a value larger than one.
  [#3862](https://github.com/memgraph/memgraph/pull/3862)
- [x] **[Documentation PR link memgraph/documentation#XXXX](https://github.com/memgraph/documentation/pull/1550)**

src/flags/logging.cpp
```diff
@@ -49,29 +46,30 @@ spdlog::level::level_enum ParseLogLevel() {
...
                           }
                           return true;
-                        }
+                        });
 
-);
+// default set to 35 days
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
+DEFINE_VALIDATED_uint64(log_retention_days, 35, "Controls for how many days will daily log files be preserved.",
+                        FLAG_IN_RANGE(1, std::numeric_limits<uint64_t>::max()));
```
andrejtonev: Is there an OFF value? -1, 0, something
And do we want to allow the user to persist till the end of the universe? :D

src/flags/logging.cpp
```diff
@@ -49,29 +46,30 @@ spdlog::level::level_enum ParseLogLevel() {
...
 
+inline constexpr std::array log_level_mappings{std::pair{"TRACE"sv, spdlog::level::trace},
+                                               std::pair{"DEBUG"sv, spdlog::level::debug},
+                                               std::pair{"INFO"sv, spdlog::level::info},
+                                               std::pair{"WARNING"sv, spdlog::level::warn},
+                                               std::pair{"ERROR"sv, spdlog::level::err},
+                                               std::pair{"CRITICAL"sv, spdlog::level::critical}};
```
andrejtonev, line 70: Not related to this, but would be nice to change the log-level check to be case insensitive. 
Currently you have to write --log-level INFO

as51340: agree will write this on todo list

src/flags/logging.cpp
as51340: I don't think there is off value. Off value would mean use some other type for file logger. We don't have support for any other type of logger currently so off doesn't count.

Till the end of the universe no 😄 You have some better max boundary

andrejtonev: `// create daily file sink which rotates on given time
    daily_file_sink(filename_t base_filename,
                    int rotation_hour,
                    int rotation_minute,
                    bool truncate = false,
                    uint16_t max_files = 0,
                    const file_event_handlers &event_handlers = {})`

not 100% sure I am looking at the exact constructor, but looks like 0 means keep files forever. 
Plus looks like it's a unit16_t so if nothing else use uint16_t max

PR #3868, opened 2026-03-04
Fix: Periodic commit safe abort

### LINKED ISSUE #3616: Memgraph crashes when using unique constraints and periodic commit
Query:

```
USING PERIODIC COMMIT 10000
CALL migrate.neo4j
(
  'MATCH p=(m:M)-[r:O]->(d:D) WHERE d.d>=20251231 AND m.pi is not null RETURN m.pi as pi,properties(m) AS mp,d.d as d',
  { 
    host: "",
    port: 7687,
    username: "",
    password: ""
  }
)
YIELD row
MERGE (m:M {pi: row.pi})
SET m += row.mp
MERGE (d:D {d: row.d})
MERGE (m)-[:O]->(d);
```

Logs:
```
[2026-01-08 06:36:54.536] [memgraph_log] [critical]
Assertion failed in file /home/mg/memgraph/src/storage/v2/inmemory/storage.cpp at line 1068.
        Expression: 'is_transaction_active_'
        Message: 'The transaction is already terminated!'
[2026-01-08 06:36:54.538] [memgraph_log] [trace] Storage GC on 'memgraph' started [periodic]
[2026-01-08 06:36:54.538] [memgraph_log] [trace] Storage GC on 'memgraph' finished [periodic]. Duration: 0.000s
PeriodicCommit failed: Unable to commit due to constraint violation.
```

Version: 3.7.2 MAGE

Need to use unique constraint, and somehow execute periodic commit writing queries.

Reproduction steps:
1. Run Memgraph MAGE (don't need to run Neo4j as the migrate.neo4j is compatible).
2. Queries:
```
CREATE CONSTRAINT ON (n:Node) ASSERT n.id IS UNIQUE;

UNWIND range(1, 10000) as x
CREATE (:Node {id: x});

USING PERIODIC COMMIT 100
CALL migrate.neo4j("MATCH (n)RETURN n.id as nid, properties(n) as props_n", {host: "", password: ""})
YIELD row
CREATE (n:Node {id: row.nid})
SET n += row.props_n;
```

andrejtonev:
PeriodicCommit handles commit internally inside the Pull. 
This means that any failure to commit aborts right there, in the Pull.
However, we are still a query, so the interpreter will also try to abort after the failure and this will cause a double abort and an assert. 
closes #3616

### DIFF
--- a/tests/e2e/periodic_commit/periodic_commit.py
+++ b/tests/e2e/periodic_commit/periodic_commit.py
@@ -107,7 +107,7 @@ def test_periodic_commit_unique_constraint_violation_does_not_crash(memgraph):
 
     # The session must still be usable after the error (no crash, no broken connection).
     result = execute_and_fetch_all(cursor, "MATCH (n:Node) RETURN count(n) AS cnt")
-    assert result[0][0] >= 100
+    assert result[0][0] == 100
 
 
 if __name__ == "__main__":


tests/e2e/periodic_commit/periodic_commit.py
```diff
@@ -87,5 +87,28 @@ def test_periodic_commit_acid_guarantee_in_batches(memgraph):
...
+    # Before the fix, this caused a double-Abort and crashed with an assertion failure.
+    with pytest.raises(DatabaseError):
+        execute_and_fetch_all(
+            cursor,
+            "USING PERIODIC COMMIT 10 UNWIND range(1, 100) AS i CREATE (:Node {id: i})",
+        )
+
+    # The session must still be usable after the error (no crash, no broken connection).
+    result = execute_and_fetch_all(cursor, "MATCH (n:Node) RETURN count(n) AS cnt")
+    assert result[0][0] >= 100
```
Josipmrden: we have a constraint, shouldn't this be exactly 100?

src/query/interpreter.hpp
```diff
@@ -656,10 +656,14 @@ std::map<std::string, TypedValue> Interpreter::Pull(TStream *result_stream, std:
...
     LogQueryMessage(e.what());
-    // Trigger first failed query
     metrics::FirstFailedQuery();
     memgraph::metrics::IncrementCounter(memgraph::metrics::FailedQuery);
     memgraph::metrics::IncrementCounter(memgraph::metrics::FailedPull);
+    // PeriodicCommitException means the storage layer already aborted the transaction internally.
+    // Null the accessor first so AbortCommand does not call Abort() a second time.
+    if (dynamic_cast<const PeriodicCommitException *>(&e)) {
+      current_db_.CleanupDBTransaction(false);
+    }
```
imilinovic, line 666: What if instead of this in Abort() we change
```
MG_ASSERT(is_transaction_active_, "The transaction is already
  terminated!");```
to
```
if (!is_transaction_active_) return;  // Already aborted
```

this would be cleaner and prevent this kind of stuff in the future if something else needs it?

imilinovic (approved):
One suggestion, otherwise LGTM

Josipmrden (approved):

src/query/interpreter.hpp
andrejtonev: Lets leave the general logic as is and implement just the special case for PeriodicCommit.
We can come back to this and perhaps redesign the flow in the future.

andrejtonev:

---
___

### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Fix: Periodic commit failure cases double abort (which in turn cases an assert). Periodic commit's failure is a special case and handled correctly by the interpreter. Users can run periodic commit without worrying if the query will always succeed. [#3868](https://github.com/memgraph/memgraph/pull/3868)

PR #3879, opened 2026-03-09
Add concurrent index creation + responsive reads stress test

Josipmrden:
Concurrent index creation stress test
Responsive reads stress test

### DIFF
--- a/tests/stress/ha/workloads/publications/workload.py
+++ b/tests/stress/ha/workloads/publications/workload.py
@@ -23,7 +23,7 @@ S3_REGION = "eu-west-1"
 
 # Concurrent workload settings
 NUM_WORKERS = 10
-ITERATIONS = 50_000  # total iterations across all workers
+ITERATIONS = 10_000  # total iterations across all workers
 
 LABELS = ["Concept", "Journal", "Publication", "Date", "Section", "Author"]
 LABEL_UNIQUE_PROP = {


tests/stress/ha/workloads/concurrent_edge_creation_on_supernode/workload.py
```diff
@@ -126,21 +128,48 @@ def verify_results() -> None:
...
+        print("=" * 60)
+        print(f"Completed in {total_elapsed:.1f}s ({total_elapsed / 60:.1f} min)")
         print("Workload completed successfully!")
-    finally:
-        cleanup()
+    except SystemExit:
+        raise
+    # finally:
+    # TODO: Inspect why this is failing to replicate
+    #     cleanup()
```
mattkjames7, line 172: Can we add the cleanup back in?

tests/stress/standalone/native/deployment/deployment.sh
```diff
@@ -115,15 +115,7 @@ stop_memgraph() {
...
-    echo "Waiting ${force_kill_settle_sec} seconds before final stop check..."
-    sleep "$force_kill_settle_sec"
-
-    if kill -0 "$MG_PID" 2>/dev/null; then
-        echo "ERROR: Failed to stop Memgraph process $MG_PID"
-        return 1
-    fi
-
-    echo "Memgraph has stopped after force kill."
+    echo "Warning: Memgraph process $MG_PID may still be in D-state (kernel I/O), continuing anyway."
```
mattkjames7, line 117: What happens if this is the case? Does this still block Memgraph from binding to 7687  if relaunched while in this state?

Josipmrden: no it seems not to

tests/stress/ha/workloads/concurrent_edge_creation_on_supernode/workload.py
Josipmrden: no, that's another issue

PR #3889, opened 2026-03-10
feat: Add built-in default roles

colinbarry:
Extends the auth system with several related improvements:                                                                                                                     

- Adds three built-in roles: `admin`, `readonly`, and `readwrite`. The first enterprise user is automatically assigned the `admin` role. Built-in roles are created atomically and only once.                                       
- Allows users and roles to share the same name (as `admin` is both a common username and rolename). Disambiguates auth commands (`SHOW PRIVILEGES`, `GET MAIN DATABASE`, etc.) by accepting an optional `USER` or `ROLE` keyword.           
- Stack roles in pieces with `GRANT ROLE[S] <roles> TO <user>` and `REVOKE ROLE[S] <roles> FROM <user>` commands. `SET ROLE FOR` is kept for backwards compatibility.
- Roles can only be deleted when no users are assigned to them.

### DIFF
--- a/src/auth/auth.cpp
+++ b/src/auth/auth.cpp
@@ -954,6 +954,7 @@ void Auth::InitialiseFirstUser(User &user, system::Transaction *system_tx) {
       "{} is the first created user. Granting all privileges. The official advice and intention is to use this "
       "first user as the superuser with full privileges and capabilities on the Memgraph database.",
       user.username());
+#ifdef MG_ENTERPRISE
   if (license::global_license_checker.IsEnterpriseValidFast()) {
     auto admin = GetRole("admin");
     if (admin && admin->IsBuiltIn()) {
@@ -961,14 +962,13 @@ void Auth::InitialiseFirstUser(User &user, system::Transaction *system_tx) {
       SaveUser(user, system_tx);
       return;
     }
-  }
-  for (auto const permission : kPermissionsAll) {
-    user.permissions().Grant(permission);
-  }
-  if (license::global_license_checker.IsEnterpriseValidFast()) {
     user.fine_grained_access_handler().label_permissions().GrantGlobal(kAllLabelPermissions);
     user.fine_grained_access_handler().edge_type_permissions().GrantGlobal(kAllEdgeTypePermissions);
     user.db_access().GrantAll();
+  }
+#endif
+  for (auto const permission : kPermissionsAll) {
+    user.permissions().Grant(permission);
   }
   SaveUser(user, system_tx);
 }
@@ -1240,6 +1240,7 @@ std::optional<Role> Auth::AddRole(const std::string &rolename, system::Transacti
   return new_role;
 }
 
+#ifdef MG_ENTERPRISE
 bool Auth::CreateBuiltinRoles(system::Transaction *system_tx) {
   if (!license::global_license_checker.IsEnterpriseValidFast()) return false;
   if (!AllRolenames().empty()) {
@@ -1289,6 +1290,7 @@ bool Auth::CreateBuiltinRoles(system::Transaction *system_tx) {
 
   return true;
 }
+#endif
 
 bool Auth::RemoveRole(const std::string &rolename_orig, bool force, system::Transaction *system_tx) {
   auto rolename = utils::ToLowerCase(rolename_orig);
--- a/src/auth/auth.cpp
+++ b/src/auth/auth.cpp
@@ -966,10 +966,8 @@ void Auth::InitialiseFirstUser(User &user, system::Transaction *system_tx) {
     user.permissions().Grant(permission);
   }
   if (license::global_license_checker.IsEnterpriseValidFast()) {
-    constexpr auto kFullFineGrained = FineGrainedPermission::READ | FineGrainedPermission::UPDATE |
-                                      FineGrainedPermission::CREATE | FineGrainedPermission::DELETE;
-    user.fine_grained_access_handler().label_permissions().GrantGlobal(kFullFineGrained);
-    user.fine_grained_access_handler().edge_type_permissions().GrantGlobal(kFullFineGrained);
+    user.fine_grained_access_handler().label_permissions().GrantGlobal(kAllLabelPermissions);
+    user.fine_grained_access_handler().edge_type_permissions().GrantGlobal(kAllEdgeTypePermissions);
     user.db_access().GrantAll();
   }
   SaveUser(user, system_tx);
@@ -1253,10 +1251,6 @@ bool Auth::CreateBuiltinRoles(system::Transaction *system_tx) {
     SaveRole(role, system_tx);
   };
 
-  constexpr auto kFullFineGrained = static_cast<FineGrainedPermission>(
-      std::to_underlying(FineGrainedPermission::READ) | std::to_underlying(FineGrainedPermission::UPDATE) |
-      std::to_underlying(FineGrainedPermission::CREATE) | std::to_underlying(FineGrainedPermission::DELETE));
-
   auto const grant_privileges =
       [](Role &role, FineGrainedPermission labelPermissions, FineGrainedPermission edgePermissions) {
         role.fine_grained_access_handler().label_permissions().GrantGlobal(labelPermissions);
@@ -1267,7 +1261,7 @@ bool Auth::CreateBuiltinRoles(system::Transaction *system_tx) {
     for (auto permission : kPermissionsAll) {
       role.permissions().Grant(permission);
     }
-    grant_privileges(role, kFullFineGrained, kFullFineGrained);
+    grant_privileges(role, kAllLabelPermissions, kAllEdgeTypePermissions);
     role.db_access().GrantAll();
   });
 
@@ -1281,7 +1275,7 @@ bool Auth::CreateBuiltinRoles(system::Transaction *system_tx) {
                             Permission::MATCH}) {
       role.permissions().Grant(permission);
     }
-    grant_privileges(role, kFullFineGrained, kFullFineGrained);
+    grant_privileges(role, kAllLabelPermissions, kAllEdgeTypePermissions);
   });
 
   make_role("readonly", [&](Role &role) {
--- a/src/auth/auth.cpp
+++ b/src/auth/auth.cpp
@@ -942,10 +942,7 @@ std::optional<User> Auth::AddUser(const std::string &username, const std::option
   if (!NameRegexMatch(username)) {
     throw AuthException("Invalid user name.");
   }
-  auto existing_user = GetUser(username);
-  if (existing_user) return std::nullopt;
-  auto existing_role = GetRole(username);
-  if (existing_role) return std::nullopt;
+  if (GetUser(username)) return std::nullopt;
   auto new_user = User(username);
   UpdatePassword(new_user, password);
   SaveUser(new_user, system_tx);
@@ -957,15 +954,17 @@ void Auth::InitialiseFirstUser(User &user, system::Transaction *system_tx) {
       "{} is the first created user. Granting all privileges. The official advice and intention is to use this "
       "first user as the superuser with full privileges and capabilities on the Memgraph database.",
       user.username());
-#ifdef MG_ENTERPRISE
-  auto admin = GetRole("admin");
-  MG_ASSERT(admin, "admin role must exist when initialising first user");
-  user.AddRole(*admin);
-#else
+  if (license::global_license_checker.IsEnterpriseValidFast()) {
+    auto admin = GetRole("admin");
+    if (admin && admin->IsBuiltIn()) {
+      user.AddRole(*admin);
+      SaveUser(user, system_tx);
+      return;
+    }
+  }
   for (auto const permission : kPermissionsAll) {
     user.permissions().Grant(permission);
   }
-#endif
   SaveUser(user, system_tx);
 }
 
@@ -1230,14 +1229,16 @@ std::optional<Role> Auth::AddRole(const std::string &rolename, system::Transacti
   if (!NameRegexMatch(rolename)) {
     throw AuthException("Invalid role name.");
   }
-  if (auto existing_role = GetRole(rolename)) return std::nullopt;
-  if (auto existing_user = GetUser(rolename)) return std::nullopt;
+  if (GetRole(rolename)) return std::nullopt;
   auto new_role = Role(rolename);
   SaveRole(new_role, system_tx);
   return new_role;
 }
 
 void Auth::CreateBuiltinRoles(system::Transaction *system_tx) {
+  if (!license::global_license_checker.IsEnterpriseValidFast()) return;
+  if (!AllRolenames().empty()) return;
+
   auto const make_role = [&](const std::string &name, auto grant_permissions) {
     Role role{name};
     grant_permissions(role);
--- a/src/auth/auth.cpp
+++ b/src/auth/auth.cpp
@@ -1293,38 +1293,40 @@ bool Auth::CreateBuiltinRoles(system::Transaction *system_tx) {
   return true;
 }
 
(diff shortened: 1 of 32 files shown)


colinbarry:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Three built-in roles (admin, readonly, readwrite) are now created automatically on first enterprise user creation, with the first user assigned admin. Roles can now be granted and revoked incrementally using `GRANT ROLE[S] <roles> TO <user>` and `REVOKE ROLE[S] <roles> FROM <user>`. [#3889](https://github.com/memgraph/memgraph/pull/3889)
    - [x] https://github.com/memgraph/documentation/pull/1601

src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -2141,7 +2152,32 @@ antlrcpp::Any CypherMainVisitor::visitClearRole(MemgraphCypher::ClearRoleContext
   auto *auth = storage_->Create<AuthQuery>();
   auth->action_ = AuthQuery::Action::CLEAR_ROLE;
   auth->user_ = std::any_cast<std::string>(ctx->user->accept(this));
-  // Optionally limit the role to specific databases
+  if (ctx->db) {
+    auto db_names = std::any_cast<std::vector<std::string>>(ctx->db->accept(this));
+    auth->role_databases_ = std::unordered_set<std::string>(std::make_move_iterator(db_names.begin()),
+                                                            std::make_move_iterator(db_names.end()));
```
andrejtonev, line 2158: Looks like we do this in a couple of places. Can we unify?

src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -2141,7 +2152,32 @@ antlrcpp::Any CypherMainVisitor::visitClearRole(MemgraphCypher::ClearRoleContext
...
+  }
+  return auth;
+}
+
+antlrcpp::Any CypherMainVisitor::visitGrantRole(MemgraphCypher::GrantRoleContext *ctx) {
```
andrejtonev, line 2163: the grant and revoke code looks very similar, unify?

src/query/frontend/ast/query/auth_query.hpp
```diff
@@ -11,6 +11,7 @@
 
 #pragma once
 
+#include "auth/models.hpp"
```
andrejtonev, line 14: Don't think we should be mixing these. 
Usually we duplicate any enums that might be needed in both libs. Example FineGrainedPrivilege

src/query/auth_query_handler.hpp
```diff
@@ -24,6 +24,12 @@
 
 namespace memgraph::query {
 
+struct CreateUserResult {
+  bool created;
```
andrejtonev, line 28: missing default values

src/query/auth_query_handler.hpp
```diff
@@ -104,11 +127,11 @@ class AuthQueryHandler {
   virtual std::vector<memgraph::query::TypedValue> GetUsernames() = 0;
 
   /// @throw QueryRuntimeException if an error ocurred.
-  virtual std::vector<memgraph::query::TypedValue> GetRolenames() = 0;
+  virtual std::vector<std::pair<std::string, bool>> GetRolenames() = 0;
```
andrejtonev, line 130: any reason we use a pair instead of a separate struct?

src/replication_handler/auth_replication_handlers.cpp
```diff
@@ -125,7 +125,7 @@ void DropAuthDataHandler(memgraph::system::ReplicaHandlerAccessToState &system_s
         auth->RemoveUser(req.name);
       } break;
       case ROLE: {
-        auth->RemoveRole(req.name);
+        auth->RemoveRole(req.name, true);
```
andrejtonev, line 128: add inline comment `/*force = */ true`
or create a constexpr

src/auth/models.hpp
```diff
@@ -571,6 +618,7 @@ class Role {
  private:
   std::string rolename_;
   Permissions permissions_;
+  bool is_builtin_{false};
```
andrejtonev, line 621: How are we protecting against users creating roles with the same names as the built in ones?

src/auth/models.cpp
```diff
@@ -798,9 +771,14 @@ Role Role::Deserialize(const nlohmann::json &data) {
...
-  return {*role_name_it, permissions, std::move(fine_grained_access_handler), std::move(db_access), std::move(usr_imp)};
+  auto role = Role{
+      *role_name_it, permissions, std::move(fine_grained_access_handler), std::move(db_access), std::move(usr_imp)};
+  role.SetBuiltIn(is_builtin);
+  return role;
 #else
-  return {*role_name_it, permissions};
+  auto role = Role{*role_name_it, permissions};
+  role.SetBuiltIn(is_builtin);
+  return role;
```
andrejtonev, line 781: could be moved outside the #ifdef

src/auth/auth.cpp
```diff
@@ -942,16 +942,37 @@ std::optional<User> Auth::AddUser(const std::string &username, const std::option
   if (!NameRegexMatch(username)) {
     throw AuthException("Invalid user name.");
   }
-  auto existing_user = GetUser(username);
-  if (existing_user) return std::nullopt;
-  auto existing_role = GetRole(username);
-  if (existing_role) return std::nullopt;
+  if (GetUser(username)) return std::nullopt;
```
andrejtonev, line 945: The GetUser and GetRole were there to check if the username is already taken.
How are we handling the roles now?

src/auth/auth.cpp
```diff
@@ -1213,81 +1234,98 @@ std::optional<Role> Auth::AddRole(const std::string &rolename, system::Transacti
   if (!NameRegexMatch(rolename)) {
     throw AuthException("Invalid role name.");
   }
-  if (auto existing_role = GetRole(rolename)) return std::nullopt;
-  if (auto existing_user = GetUser(rolename)) return std::nullopt;
+  if (GetRole(rolename)) return std::nullopt;
```
andrejtonev, line 1237: Same comment as for the User

src/auth/models.cpp
```diff
@@ -773,6 +740,12 @@ Role Role::Deserialize(const nlohmann::json &data) {
     throw AuthException("Couldn't load role data!");
   }
   auto permissions = Permissions::Deserialize(*permissions_it);
+
+  bool is_builtin = false;
```
andrejtonev, line 744: Trying to get the full picture.
We introduced the built in roles. But we are also marking each role with is_builtin. Why?
Is it for back-compatibility? But I saw that we create the builtin roles only is there aren't any roles.
There is also no migration logic. This all gets weird because we give the admin role to the first user only if it is the built in one.
I think we need to simplify. What are we tying to accomplish?

src/glue/auth_handler.cpp
```diff
@@ -551,23 +528,46 @@ void AuthQueryHandler::RevokeDatabase(const std::string &db_name, const std::str
 }
 
 std::vector<std::vector<memgraph::query::TypedValue>> AuthQueryHandler::GetDatabasePrivileges(
-    const std::string &user, const std::vector<std::string> &roles) {
+    const std::string &user, const std::vector<std::string> &roles, auth::UserOrRoleType type) {
```
andrejtonev, line 531: Here (I think) the user is the user and roles are the roles, we can't be UNSPECIFIED.

src/glue/auth_handler.cpp
```diff
@@ -551,23 +528,46 @@ void AuthQueryHandler::RevokeDatabase(const std::string &db_name, const std::str
...
   try {
     auto locked_auth = auth_->ReadLock();
-    if (auto local_user = locked_auth->GetUser(user)) {
+    auto local_user = (type != auth::UserOrRoleType::ROLE) ? locked_auth->GetUser(user) : std::nullopt;
```
andrejtonev, line 534: Are we allowing for username and rolenames to overlap?

src/query/frontend/ast/cypher_main_visitor.cpp
colinbarry: Agreed. Moved to a static namespace-scoped helper function in `cypher_main_visitor.cpp`

src/query/frontend/ast/cypher_main_visitor.cpp
colinbarry: I've moved the meaty part of the code into the `GetRoleDatabases` helper, leaving these visits as simple as possible. Seems to be an acceptable amount of similarity now.

src/query/auth_query_handler.hpp
colinbarry: Changed to `RolenameResult`

src/replication_handler/auth_replication_handlers.cpp
colinbarry: Done: `/*force = */ true`

src/auth/models.hpp
colinbarry: `AddRole` doesn't allow duplicate roles (but as we said, roles and users are now in two different namespaces, so an `admin` role and and `admin` user won't collide.)

src/auth/auth.cpp
colinbarry: Intentional change. Usernames and rolenames are in two different namespace and won't collide with each other.

src/auth/models.cpp
colinbarry: `is_builtin_` exists for one specific purpose: in `InitialiseFirstUser`, we only auto-assign admin to the first user if it is the built-in role, not if someone happened to create a plain role named `admin` beforehand. Without the flag we can't distinguish the two. (Discussed and resolved in sync.)

src/glue/auth_handler.cpp
colinbarry: Two namespaces, so user `sales` and role `sales` are unique entities, and can be targetted in truly ambiguous situations with explicit `ROLE sales` or `USER sales`.

src/glue/auth_handler.cpp
colinbarry: `UNSPECIFIED` here means the user has not specified either `USER sales` or `ROLE sales`. `UNSPECIFIED` is fine provided there is no ambiguity. I've updated the comments for the `UNSPECIFIED` symbol so it is clear what this means.

src/auth/models.cpp
colinbarry: Done

src/query/auth_query_handler.hpp
colinbarry: Done.

src/query/frontend/ast/query/auth_query.hpp
colinbarry: Done. Duplicated those symbols and added some mappings between the scoped matching symbols.

src/auth/auth.cpp
colinbarry: Resolved

src/auth/auth.cpp
```diff
@@ -1213,81 +1234,98 @@ std::optional<Role> Auth::AddRole(const std::string &rolename, system::Transacti
...
+  make_role("readonly", [&](Role &role) {
+    role.permissions().Grant(Permission::MATCH);
+    role.permissions().Grant(Permission::STATS);
+    grant_privileges(role, FineGrainedPermission::READ, FineGrainedPermission::READ);
+  });
+
+  return true;
+}
+
+bool Auth::RemoveRole(const std::string &rolename_orig, bool force, system::Transaction *system_tx) {
```
andrejtonev, line 1295: I assume we allow built in roles to be deleted, since we can modify them anyway

src/auth/auth.cpp
```diff
@@ -1213,81 +1234,98 @@ std::optional<Role> Auth::AddRole(const std::string &rolename, system::Transacti
...
+        if (!json_data.is_array()) continue;
+        for (auto const &role_name : json_data) {
+          if (role_name.is_string() && utils::ToLowerCase(role_name.get<std::string>()) == rolename) {
+            throw AuthException(
+                "Cannot delete role '{}' as it is assigned to one or more users. Drop or reassign the users first.",
+                rolename);
           }
         }
+      } catch (AuthException const &) {
+        throw;
```
andrejtonev: not need

src/auth/auth.cpp
```diff
@@ -1213,81 +1234,98 @@ std::optional<Role> Auth::AddRole(const std::string &rolename, system::Transacti
...
+  if (GetRole(rolename)) return std::nullopt;
   auto new_role = Role(rolename);
   SaveRole(new_role, system_tx);
   return new_role;
 }
 
-bool Auth::RemoveRole(const std::string &rolename_orig, system::Transaction *system_tx) {
-  auto rolename = utils::ToLowerCase(rolename_orig);
-  if (!storage_.Get(kRolePrefix + rolename)) return false;
+bool Auth::CreateBuiltinRoles(system::Transaction *system_tx) {
```
andrejtonev, line 1244: could you add some debug/trace logs that specify how the function failed?

src/auth/auth.cpp
```diff
@@ -1213,81 +1234,98 @@ std::optional<Role> Auth::AddRole(const std::string &rolename, system::Transacti
...
       }
     }
-    if (update) {
-      if (json_data.empty()) {
-        storage_.Delete(it->first);
-      } else {
-        storage_.Put(it->first, json_data.dump());
+
+    for (auto it = storage_.begin(kMtLinkPrefix); it != storage_.end(kMtLinkPrefix); ++it) {
+      auto username = it->first.substr(kMtLinkPrefix.size());
```
andrejtonev, line 1319: RolePrefix loop had a try catch block. Do we need it here as well?

src/auth/auth.cpp
```diff
@@ -1408,13 +1446,17 @@ std::vector<std::string> Auth::AllUsernamesForRole(const std::string &rolename_o
...
 
 #ifdef MG_ENTERPRISE
-Auth::Result Auth::GrantDatabase(const std::string &db, const std::string &name, system::Transaction *system_tx) {
+Auth::Result Auth::GrantDatabase(const std::string &db, const std::string &name, UserOrRoleType type,
+                                 system::Transaction *system_tx) {
   using enum Auth::Result;
-  if (auto user = GetUser(name)) {
+  auto user = (type != UserOrRoleType::ROLE) ? GetUser(name) : std::nullopt;
+  auto role = (type != UserOrRoleType::USER) ? GetRole(name) : std::nullopt;
+  if (user && role) throw AuthException("Ambiguous: '{}' is both a user and a role. Specify USER or ROLE.", name);
```
andrejtonev: this is used a lot, can we implement a function that handles the logic instead of copy/pasting it everywhere?
maybe passing in the user and/or role callbacks
there was also a type called something like UserOrRole, not sure if that helps here

andrejtonev (approved):
just a few minor comments

src/auth/auth.cpp
colinbarry: Correct: built-in roles can be deleted.

src/auth/auth.cpp
colinbarry: Refactored to remove this.

src/auth/auth.cpp
colinbarry: Done.

src/auth/auth.cpp
colinbarry: We do. Added the catch.

src/auth/auth.cpp
colinbarry: Added a namespace-scoped helper to dispatch to the relevant function depending on whether it the entity is a user or role.

PR #3895, opened 2026-03-11
feat: make comparisons and order by respect OpenCypher 9 specifications

imilinovic:
## Summary

Brings TypedValue equality, comparability, orderability, and equivalence into line with the openCypher 9 specification. Previously, `=`, `<`, `<=`, `>`, `>=` and ORDER BY had ad-hoc null/NaN handling that produced incorrect or non-spec results for cross-type comparisons, nested nulls, durations, and NaN. This PR replaces that with the four-relation model from the openCypher CIP "Define comparability and equality as well as orderability and equivalence":

- **Equality** (`=`, `<>`) — propagates null per three-valued logic.
- **Comparability** (`<`, `<=`, `>`, `>=`) — null for any null operand, NaN, incomparable types, or types that are not comparable in the spec (Duration).
- **Orderability** (ORDER BY) — total order across all types using a fixed type rank, nulls last in ASC.
- **Equivalence** (DISTINCT, GROUP BY, hash-set membership) — equality except null is equivalent to null and NaN is equivalent to NaN, recursively into containers.

## What changed

`src/query/typed_value.{cpp,hpp}` and `src/query/common.hpp`:

- `operator<` rewritten to the spec contract — null/incomparable types yield null instead of throwing or producing implementation-defined results. Lists, maps, and paths get dedicated three-valued comparison helpers.
- `operator==` for List and Map uses three-valued conjunction: a definite false element dominates over null (`[1, null] = [2, 0]` is false), and null is returned only when no element compared false.
- `operator<=`, `>`, `>=` are now one-line forwarders derived from a single `<` call (`!(b<a)`, `b<a`, `!(a<b)`). Avoids the redundant element-walk that the previous `(a<b) || (a==b)` form did for containers, and lets null propagate through `!` per three-valued logic.
- `TypedValueCompare` (`common.hpp`) — total order over every TypedValue type for ORDER BY, using the spec's type-rank hierarchy (`MAP < NODE < REL < LIST < PATH < POINT < temporal types < STRING < BOOL < NUMBER < NULL`). `std::unreachable()` after the rank switch so a future enum entry fails-fast instead of returning garbage.
- `BoolEqual` (DISTINCT/GROUP BY) reimplemented as recursive equivalence: any two nulls collapse, any two NaNs collapse, container elements compared element-wise. Switch lists every TypedValue type explicitly so a new enum entry fails to compile rather than silently routing through the old `default: → operator==` path.
- `TypedValue::Hash` maps every NaN double to a fixed sentinel, since `std::hash<double>` is unspecified for NaN bit patterns and a divergence between `Hash` and `BoolEqual` would silently break DISTINCT/GROUP BY on NaN keys.

`tests/gql_behave/tests/openCypher_M09/features/`:

- `ComparabilityEquivalence.feature` (770 lines, 75 scenarios) — equality, comparability, orderability, DISTINCT, GROUP BY, aggregation, node/rel/path equality, temporal types, points, mixed types.
- `WorkingWithNull.feature` (182 lines, 36 scenarios) — null propagation through arithmetic, logical operators, comparisons, IN, list/map ops, CASE, COALESCE.

## Performance

Net improvement vs `master`, no regressions on hot paths:

| Operator | master | this PR |
|---|---|---|
| `<=` | `(a<b) \|\| (a==b)` — two element walks (TypedValue `\|\|` does **not** short-circuit, it's a free function) | `!(b<a)` — one walk |
| `>` | `!(a<=b)` — two walks | `b<a` — one walk |
| `>=` | `b<=a` → two walks | `!(a<b)` — one walk |
| `==` for List/Map | per-element TypedValue construction | same allocation count, more spec-correct |
| `BoolEqual` (DISTINCT/GROUP BY) | calls public `==` per element, allocating a TypedValue per nesting level | recursive bool-returning helper, only allocates at scalar leaves |

For ORDER BY of list/map-typed expressions and DISTINCT/GROUP BY on container keys, this halves element-walk cost on the comparison path and removes per-element TypedValue allocations.


imilinovic:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist

imilinovic:
fixed in other PRs

PR #3900, opened 2026-03-12
fix: Coordinators config logs marshalling

as51340:
## What

Fix `CoordinatorLogStore` to correctly persist NuRaft `conf` (cluster config) log entries. Previously, `StoreEntryToDisk` discarded the buffer content for non-`app_log` entries, storing an empty
string. On restart, the reloaded 4-byte buffer causes `cluster_config::deserialize()` to throw `std::overflow_error("not enough space")`, crash-looping the coordinator.

Config entries are now serialized as JSON via the existing `to_json(cluster_config)` on write, and reconstructed via `from_json` + `cluster_config::serialize()` on reload.

## Why

When a new coordinator joins an HA cluster, NuRaft appends `conf` entries for the membership change. If the coordinator restarts before those entries are compacted by a snapshot, the corrupted
entries are loaded from disk. NuRaft's ASIO worker thread dies trying to deserialize them, the Raft server fails to initialize within the 120-second timeout, and `memgraph.cpp` returns
`EXIT_FAILURE` — repeating on every restart.

This affects **all** coordinators in the cluster, but long-running ones typically survive because snapshots compact away the conf entries before a restart occurs.

## How

- **`StoreEntryToDisk`**: For `log_val_type::conf` entries, deserialize the NuRaft binary buffer via `cluster_config::deserialize()`, convert to JSON with `to_json()`, and store the JSON string in
 the existing `"data"` field — instead of storing an empty string.
- **`HandleVersionMigration`** (reload path): For `conf` entries with non-empty `"data"`, parse the JSON with `from_json()` to reconstruct a `cluster_config`, then call `serialize()` to produce
the exact NuRaft binary buffer. Old entries with empty data (written by the buggy code) fall through to the existing `put_str` path — they remain broken but cannot be recovered regardless.
- No `LogStoreVersion` bump needed. The change is backward-compatible: new code reads both old (empty) and new (JSON) formats. Old empty conf entries are already irrecoverably corrupted.

## Testing

- **`TestConf`**: Updated to use a real `cluster_config` buffer instead of an arbitrary string that could never exist in a real NuRaft conf entry.
- **`TestConfEntryBufferPreservedAfterRestart`** (new): Creates a 3-server `cluster_config`, stores it as a `conf` entry, destroys and recreates the log store (simulating restart), then verifies:
(1) buffer size matches the original, (2) `cluster_config::deserialize()` succeeds without throwing, (3) deserialized config fields match the original. This test fails without the fix (4-byte
buffer vs 115-byte original) and passes with it.

## Notes for reviewers

- **Affected users**: All coordinators in a cluster store corrupted conf entries. A coordinator that restarts before snapshot compaction will crash-loop. Recovery requires wiping the raft data
directory and rejoining.  Recommended procedure: rolling wipe — followers first, leader last.
- The fix relies on the existing `to_json`/`from_json` for `cluster_config` in `coordinator_state_manager.hpp`. These are already used for snapshot persistence, so the serialization format is
proven.
- NuRaft does **not** call `state_machine::commit()` for `conf` entries — only for `app_log` entries — so the state machine commit path is unaffected.

### DIFF
--- a/src/coordination/coordinator_log_store.cpp
+++ b/src/coordination/coordinator_log_store.cpp
@@ -100,10 +100,12 @@ bool CoordinatorLogStore::HandleVersionMigration(LogStoreVersion const stored_ve
             std::shared_ptr<nuraft::cluster_config> config;
             from_json(config_json, config);
             log_term_buffer = config->serialize();
-          } else {
+          } else if (val_type_enum == nuraft::log_val_type::app_log) {
             log_term_buffer = buffer::alloc(sizeof(uint32_t) + data.size());
             buffer_serializer bs{log_term_buffer};
             bs.put_str(data);
+          } else {
+            LOG_FATAL("Invalid type of logs found in log store durability");
           }
 
           logs_[id] = std::make_shared<log_entry>(term, log_term_buffer, val_type_enum);


as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Fixed a bug where coordinator Raft cluster configuration log entries were not persisted correctly, causing coordinators to enter a crash-loop on restart if uncompacted configuration entries
  existed on disk — most commonly triggered when a newly added coordinator restarts before a Raft snapshot compacts those entries. Users running HA clusters who experience a coordinator crash-loop
  after adding a new coordinator should upgrade to this version; all coordinators should have its raft data directory wiped before restarting and setup queries should be run again on the new coordinator leader. [#3900](https://github.com/memgraph/memgraph/pull/3900)

src/coordination/coordinator_log_store.cpp
```diff
@@ -336,17 +347,17 @@ bool CoordinatorLogStore::flush() {
 bool CoordinatorLogStore::StoreEntryToDisk(const std::shared_ptr<log_entry> &clone, uint64_t key_id,
                                            bool is_newest_entry) {
   auto const data_string = [&clone, logger = &logger_]() -> std::string {
-    if (clone->get_val_type() != nuraft::log_val_type::app_log) {
-      // this is only our log, others nuraft creates and we
-      // don't have actions for them
-      logger->Log(nuraft_log_level::TRACE, "Received non-application log, data will be empty string.");
-      // behavior.
-      return {};
+    if (clone->get_val_type() == nuraft::log_val_type::conf) {
```
andrejtonev: we want from checking != app_log to explicitly checking if it is conf. what about all the other types?

src/coordination/coordinator_log_store.cpp
```diff
@@ -85,10 +83,23 @@ bool CoordinatorLogStore::HandleVersionMigration(LogStoreVersion const stored_ve
...
           auto const data = j.at(kLogEntryDataKey).get<std::string>();
           auto const value_type = j.at("val_type").get<int>();
-          auto log_term_buffer = buffer::alloc(sizeof(uint32_t) + data.size());
-          buffer_serializer bs{log_term_buffer};
-          bs.put_str(data);
-          logs_[id] = std::make_shared<log_entry>(term, log_term_buffer, static_cast<nuraft::log_val_type>(value_type));
+          auto const val_type_enum = static_cast<nuraft::log_val_type>(value_type);
+
+          nuraft::ptr<buffer> log_term_buffer;
+          if (val_type_enum == nuraft::log_val_type::conf && !data.empty()) {
```
andrejtonev: If I understood correctly, we have a conf entry in the log, but it has not been committed (since we store the conf in commit via save_config). Then when we restart after the leg entry, but before the commit, we read the log entry (which was badly serialized) and crash.
If we already store it on commit and log entries that have not been committed are not valid, why go through the trouble of making sure the log store is correct, instead of just ignoring it and fixing the deserialization instead?

src/coordination/coordinator_log_store.cpp
as51340: only these 2 types of logs are actually committed and persisted. Only ones we should care about

src/coordination/coordinator_log_store.cpp
as51340: the problem was that conf logs that are committed were invalid. Snapshot saved us usually but if there were logs which were not yet put under the snapshot, we would see a failure

src/coordination/coordinator_log_store.cpp
andrejtonev: I'd rather have a switch and handle all other cases with mg_assert or throw

src/coordination/coordinator_log_store.cpp
```diff
@@ -85,10 +83,23 @@ bool CoordinatorLogStore::HandleVersionMigration(LogStoreVersion const stored_ve
...
+            // Config entries store a JSON-serialized cluster_config.
+            // Reconstruct the NuRaft binary buffer from it.
+            auto const config_json = nlohmann::json::parse(data);
+            std::shared_ptr<nuraft::cluster_config> config;
```
andrejtonev, line 100: why is this a shared_ptr?

andrejtonev: or try to overload the from/to to support std::shared_ptr<T>

src/coordination/coordinator_log_store.cpp
andrejtonev: wrong check, this way empty conf will execute the old logic and crash

tests/unit/coordinator_log_store.cpp
```diff
@@ -356,14 +356,78 @@ TEST_F(CoordinatorLogStoreTests, TestConf) {
...
-  auto log_term_buffer = buffer::alloc(sizeof(uint32_t) + data.size());
-  buffer_serializer bs{log_term_buffer};
-  bs.put_str(data);
-  auto log_entry_obj = cs_new<log_entry>(1, log_term_buffer, nuraft::log_val_type::conf);
+  // Create a real cluster_config — conf entries always carry serialized cluster configs
+  auto config = nuraft::cs_new<nuraft::cluster_config>(10, 9);
+  config->get_servers().push_back(nuraft::cs_new<nuraft::srv_config>(1, 0, "127.0.0.1:12000", "", false, 1));
+  auto conf_buf = config->serialize();
+
+  auto log_entry_obj = cs_new<log_entry>(1, conf_buf, nuraft::log_val_type::conf);
```
andrejtonev, line 366: add tests for version migration and also for empty conf log

src/coordination/coordinator_log_store.cpp
as51340: actually it needs to be shared_ptr because cluster_config Is non-movable and non-copyable and there is no method to set_log_prev_idx. I will have to keep it this way. I knew there had to be a reason why we did like this

src/coordination/coordinator_log_store.cpp
as51340: yep, I think this is correct

src/coordination/coordinator_log_store.cpp
```diff
@@ -85,10 +83,30 @@ bool CoordinatorLogStore::HandleVersionMigration(LogStoreVersion const stored_ve
...
+                  "Please stop all coordinators, remove their data directories, and re-establish the HA cluster.",
+                  id);
+            }
+            // Config entries store a JSON-serialized cluster_config.
+            // Reconstruct the NuRaft binary buffer from it.
+            auto const config_json = nlohmann::json::parse(data);
+            std::shared_ptr<nuraft::cluster_config> config;
+            from_json(config_json, config);
+            log_term_buffer = config->serialize();
+          } else {
```
andrejtonev: lets else if == app_info and then else dmg_assert

PR #3903, opened 2026-03-13
fix: Destroy DataInstanceManagementServer before repl state

as51340:
## What

Explicitly shut down the `DataInstanceManagementServer` during SIGTERM handling before tearing down replication state, and harden the `Shutdown()` contract across `communication::Server`,
`rpc::Server`, `ReplicationServer`, `DataInstanceManagementServer`, and `CoordinatorInstanceManagementServer` so that it returns `bool` (true = this caller performed the shutdown) and is safe to
call multiple times via atomic CAS on the `alive_` flag.

## Why

On SIGTERM, the `DataInstanceManagementServer` (coordinator→data-instance RPC server) was never stopped for data instances. It kept accepting RPCs (e.g. `StateCheckReq`) while `repl_state` was
being destroyed, causing a use-after-free that manifested as `std::get: variant is valueless` when the handler read the destroyed `ReplicationState` variant.

## How

- **`communication::Server::Shutdown()`** — replaced the plain `alive_.store(false)` with `compare_exchange_strong`, making the method return `bool` and ensuring only one thread performs the
socket/listener shutdown. This eliminates the TOCTOU between `IsRunning()` and `Shutdown()` that all callers previously relied on.
- **`rpc::Server::Shutdown()`** — now returns `bool`, forwarding from `communication::Server`.
- **`DataInstanceManagementServer`** — extracted a public `Shutdown()` method (called from the shutdown lambda and from the destructor). Uses the new `bool` return from `rpc_server_.Shutdown()` to
 guard `AwaitShutdown()` and tracing.
- **`CoordinatorInstanceManagementServer`** — same pattern applied to the destructor.
- **`ReplicationServer`** — same pattern: `Shutdown()` returns `bool`, uses CAS result to guard await and tracing.
- **`memgraph.cpp` shutdown lambda** — added an explicit `DataInstanceManagementServer::Shutdown()` call after `server.Shutdown()` (Bolt) but before `repl_state->Shutdown()`, with a comment
explaining the lifetime dependency between `coordinator_state` and `repl_state`.

## Testing

All unit tests calling `Shutdown()` on `rpc::Server` or `communication::Server` were updated to wrap the call in `ASSERT_TRUE(...)`, validating that exactly one shutdown occurs. Affected test
files:

- `rpc.cpp` (8 call sites)
- `network_timeouts.cpp` (2)
- `rpc_in_progress.cpp` (4)
- `rpc_timeouts.cpp` (5)
- `rpc_versioning.cpp` (4)
- `replication_rpc_progress.cpp` (8)
- `snapshot_rpc_progress.cpp` (3)

No new test was added for the SIGTERM race itself; deterministic reproduction requires a coordinator sending RPCs during shutdown. The Jepsen suite on `testing/jepsen-sigterm` covers this scenario
 end-to-end.

## Notes for reviewers

- The `.gitignore` change adding `GEMINI.md` is unrelated housekeeping.
- `CoordinatorInstanceManagementServer` does not yet expose a public `Shutdown()` — its destructor delegates directly. If coordinator-side handlers ever develop similar lifetime issues, a public
method would be needed.
- No durability version bump or ISSU implications; the change is purely runtime shutdown ordering.

### DIFF
--- a/src/communication/server.hpp
+++ b/src/communication/server.hpp
@@ -125,7 +125,7 @@ class Server final {
   bool Shutdown() {
     // This should be as simple as possible, so that it can be called inside a
     // signal handler.
-    auto expected{true};
+    bool expected{true};
     // If I am the thread which earned the right to do the shutdown
     auto const res =
         alive_.compare_exchange_strong(expected, false, std::memory_order_acq_rel, std::memory_order_acquire);
--- a/src/replication/replication_server.cpp
+++ b/src/replication/replication_server.cpp
@@ -56,7 +56,7 @@ bool ReplicationServer::Shutdown() {
   // if I am the thread which did the shutdown
   if (rpc_server_.Shutdown()) {
     try {
-      spdlog::trace("Closing replication server on");
+      spdlog::trace("Closing replication server");
       // NOLINTNEXTLINE(bugprone-empty-catch)
     } catch (std::exception const &) {
     }
--- a/src/rpc/server.cpp
+++ b/src/rpc/server.cpp
@@ -18,7 +18,7 @@ Server::Server(io::network::Endpoint endpoint, communication::ServerContext *con
 
 bool Server::Start() { return server_.Start(); }
 
-bool Server::Shutdown() { return server_.Shutdown(); }
+[[nodiscard]] bool Server::Shutdown() { return server_.Shutdown(); }
 
 void Server::AwaitShutdown() { server_.AwaitShutdown(); }
 


as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Fixed a race condition during graceful shutdown (SIGTERM) of high-availability data instances where the internal management RPC server could process incoming coordinator health-check requests
  after the replication state had already been destroyed, resulting in a crash with the error "std::get: variant is valueless"; the shutdown sequence now correctly stops the management server before
   tearing down replication state, so users running HA clusters should no longer observe unexpected crashes or error logs when stopping data instances with SIGTERM. No configuration changes or user
  action is required. [#3903](https://github.com/memgraph/memgraph/pull/3903)

src/replication/replication_server.cpp
```diff
@@ -52,18 +52,18 @@ ReplicationServer::~ReplicationServer() { Shutdown(); }
...
-  if (rpc_server_.IsRunning()) {
+bool ReplicationServer::Shutdown() {
+  // if I am the thread which did the shutdown
+  if (rpc_server_.Shutdown()) {
     try {
-      // trace can throw
-      auto const &endpoint = rpc_server_.endpoint();
-      spdlog::trace("Closing replication server on {}", endpoint.SocketAddress());
-      rpc_server_.Shutdown();
+      spdlog::trace("Closing replication server on");
```
andrejtonev: update log to something better, since the endpoint is no longer read

src/rpc/server.cpp
```diff
@@ -18,7 +18,7 @@ Server::Server(io::network::Endpoint endpoint, communication::ServerContext *con
 
 bool Server::Start() { return server_.Start(); }
 
-void Server::Shutdown() { server_.Shutdown(); }
+bool Server::Shutdown() { return server_.Shutdown(); }
```
andrejtonev: add [[nodiscard]]

src/communication/server.hpp
```diff
@@ -122,14 +122,20 @@ class Server final {
...
-  void Shutdown() {
+  bool Shutdown() {
     // This should be as simple as possible, so that it can be called inside a
     // signal handler.
-    alive_.store(false);
-    // Shutdown the socket to return from any waiting `Accept` calls.
-    socket_.Shutdown();
-    // Shutdown the listener.
-    listener_.Shutdown();
+    auto expected{true};
```
andrejtonev: use bool

src/memgraph.cpp
andrejtonev, line 900: Is this fine to do while the server is still running and potentially bolt workers are executing queries?

src/memgraph.cpp
```diff
@@ -863,6 +867,15 @@ int main(int argc, char **argv) {
     if (worker_pool_) worker_pool_->ShutDown();  // Workers can enqueue io tasks, so they need to be stopped first
     // Shutdown communication server
     server.Shutdown();
```
andrejtonev, line 869: Issue #3: Full call chain
The shutdown side (main thread)
[src/memgraph.cpp:869](vscode-webview://0p08eprq938dgi4aelr8aqplfauvj9415dksbs5lea80jf0a72st/src/memgraph.cpp#L869) — Bolt server signaled to stop, but NOT awaited:


server.Shutdown();
[src/memgraph.cpp:888](vscode-webview://0p08eprq938dgi4aelr8aqplfauvj9415dksbs5lea80jf0a72st/src/memgraph.cpp#L888) — write-locks the client list, destroys all ReplicationStorageClient objects (including their rpc::Client):


acc->storage()->repl_storage_state_.replication_storage_clients_.WithLock([](auto &clients) { clients.clear(); });
[src/memgraph.cpp:941](vscode-webview://0p08eprq938dgi4aelr8aqplfauvj9415dksbs5lea80jf0a72st/src/memgraph.cpp#L941) — Bolt threads finally joined, far too late:


server.AwaitShutdown();
The commit side (Bolt thread, concurrently)
[src/storage/v2/inmemory/storage.cpp:1033-1034](vscode-webview://0p08eprq938dgi4aelr8aqplfauvj9415dksbs5lea80jf0a72st/src/storage/v2/inmemory/storage.cpp#L1033) — creates a TransactionReplication, which immediately acquires a read lock on replication_storage_clients_:


auto replicating_txn =
    mem_storage->repl_storage_state_.StartPrepareCommitPhase(durability_commit_timestamp, mem_storage, commit_args);
[src/storage/v2/replication/replication_storage_state.cpp:19-21](vscode-webview://0p08eprq938dgi4aelr8aqplfauvj9415dksbs5lea80jf0a72st/src/storage/v2/replication/replication_storage_state.cpp#L19) — passes the list by reference; TransactionReplication constructor will call ReadLock() on it:


return {durability_commit_timestamp, storage, commit_args, replication_storage_clients_};
[src/storage/v2/replication/replication_transaction.hpp:79-82](vscode-webview://0p08eprq938dgi4aelr8aqplfauvj9415dksbs5lea80jf0a72st/src/storage/v2/replication/replication_transaction.hpp#L79) — the critical member ordering:


private:
  std::vector<std::optional<ReplicaStream>> streams;  // declared 1st → destroyed LAST
  ReadLockedPtr locked_clients;                        // declared 2nd → destroyed 2nd-to-last
  bool run_two_phase_commit = false;
C++ destroys members in reverse declaration order. So:

run_two_phase_commit (trivial)
locked_clients destructs → read lock released → main thread's WithLock() can now acquire the write lock → clients.clear() runs → ReplicationStorageClient and its rpc::Client are destroyed
streams destructs (still has live ReplicaStream objects if commit threw or returned early)
[src/storage/v2/replication/replication_client.hpp:73-75](vscode-webview://0p08eprq938dgi4aelr8aqplfauvj9415dksbs5lea80jf0a72st/src/storage/v2/replication/replication_client.hpp#L73) — ReplicaStream holds:


private:
  Storage *storage_;
  rpc::Client::StreamHandler<replication::PrepareCommitRpc> stream_;  // holds Client* and a lock on it
[src/rpc/client.hpp:303-308](vscode-webview://0p08eprq938dgi4aelr8aqplfauvj9415dksbs5lea80jf0a72st/src/rpc/client.hpp#L303) — StreamHandler holds a raw Client* and a unique_lock that will unlock() the client's ResourceLock in its destructor:


Client *self_;                                    // raw pointer — dangling after clients.clear()
std::unique_lock<utils::ResourceLock> guard_;     // destructor calls self_->lock_.unlock() → UB
slk::Builder req_builder_;                        // callback captures Client* — also dangling
The race

Bolt thread                            Main thread (shutdown lambda)
──────────────────────────────────────────────────────────────────
replicating_txn out of scope:
  ~run_two_phase_commit (trivial)
  ~locked_clients                  →  [read lock released]
                                       WithLock() acquires write lock
                                       clients.clear() runs
                                       ~ReplicationStorageClient()
                                       ~rpc::Client()               ← rpc::Client destroyed
  ~streams:
    ~ReplicaStream()
      ~StreamHandler()
        ~guard_: calls unlock()
          on destroyed ResourceLock  ← USE AFTER FREE
The root cause is the declaration order in [replication_transaction.hpp:79-82](vscode-webview://0p08eprq938dgi4aelr8aqplfauvj9415dksbs5lea80jf0a72st/src/storage/v2/replication/replication_transaction.hpp#L79): streams must be destroyed before locked_clients releases the read lock — but as declared, it's destroyed after. Reversing the two declarations (put locked_clients first, streams last) would fix it: streams would then destruct while the read lock is still held, keeping the clients alive.

andrejtonev: this maybe correct, but I think the solution is to shutdown and awaitshiutdown in the same place. since now we don't use sighandlers, but await the singals in main

src/memgraph.cpp
andrejtonev: same as the other comment

andrejtonev (approved):

src/memgraph.cpp
as51340: will check in other PR

src/memgraph.cpp
as51340: also checking in other PR

src/memgraph.cpp
as51340: Logically, I think this should be fine because coordinator state should be independent from the bolt server and executing such queries

PR #3908, opened 2026-03-16
fix: Close Python module

as51340:
## What

Fixes a crash during Memgraph shutdown caused by the `PythonModule` destructor throwing exceptions. The destructor now catches exceptions gracefully, and all filesystem operations in
`PythonModule::Close()` (specifically `remove_all` and `recursive_directory_iterator` increment) use non-throwing error-code overloads instead of their throwing counterparts.

## Why

During shutdown, `ModuleRegistry::UnloadAllModules()` clears the module map, which destroys each `PythonModule` via its `shared_ptr`. The `PythonModule::Close()` method performs filesystem cleanup
 (`__pycache__` removal, directory iteration) that can throw — e.g. when cached files have already been deleted. Since destructors are implicitly `noexcept` in C++, any escaping exception triggers
 `std::terminate` → `abort`, producing a core dump on every shutdown when Python modules are loaded.

## How

Two layers of defense in `src/query/procedure/module.cpp`:

- **Destructor safety net**: `~PythonModule()` now wraps the `Close()` call in a `try/catch(std::exception&)` block, logging the error instead of terminating.
- **Non-throwing filesystem ops in `Close()`**: `std::filesystem::remove_all` calls for `__pycache__` directories now use the `std::error_code` overload. The `recursive_directory_iterator` loop
was rewritten from a range-for (which uses throwing `operator++`) to a manual loop with `increment(ec)`, so deleted entries during iteration don't cause exceptions. All error codes are checked and
 logged as warnings.

## Testing

No new tests added. The crash only manifests during process shutdown when Python query modules with submodule directories are loaded and their `__pycache__` has already been cleaned up or is
otherwise inaccessible — this is a teardown-order issue that is difficult to reproduce in unit tests. The fix is purely defensive (catch + error-code overloads) with no behavioral change to the
happy path.

## Notes for reviewers

- The outer `try/catch` in the destructor is intentionally kept as a safety net even though the inner filesystem operations are now non-throwing — `Close()` also calls `py::EnsureGIL()`,
`ProcessFileDependencies()`, and Python C API functions that could still throw in edge cases.
- The `recursive_directory_iterator` invalidation was caused by `remove_all` on `__pycache__` dirs modifying the tree mid-iteration. The manual `increment(ec)` loop handles this gracefully by
breaking on error

### DIFF
--- a/src/query/procedure/module.cpp
+++ b/src/query/procedure/module.cpp
@@ -1094,7 +1094,7 @@ bool PythonModule::Close() {
         break;
       }
       auto const &rec_dir_entry = *rec_it;
-      std::string rec_dir_entry_stem = rec_dir_entry.path().stem().string();
+      std::string const rec_dir_entry_stem = rec_dir_entry.path().stem().string();
       if (rec_dir_entry.is_directory() && rec_dir_entry_stem != "__pycache__") {
         ec.clear();
         std::filesystem::remove_all(rec_dir_entry.path() / "__pycache__", ec);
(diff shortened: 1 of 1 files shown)


as51340:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Python module cleanup during shutdown no longer crashes the process; previously, filesystem operations that failed while removing __pycache__ directories would throw exceptions inside a
  destructor, triggering std::terminate and an abrupt abort — now those operations use non-throwing error-code overloads and log warnings instead, so shutdown completes gracefully even when cached
  files have already been removed. [#3908](https://github.com/memgraph/memgraph/pull/3908)

PR #3911, opened 2026-03-16
feat: Add per-tenant metrics to OpenMetrics endpoint

### LINKED ISSUE #3087: Implement per database memory usage control
- [ ] Track memory usage per each database
- [ ] Expose memory usage under monitoring
- [ ] Limit memory usage per database

## Related

* https://github.com/memgraph/memgraph/issues/1255

colinbarry:
**Changes:**
- Adds a metrics endpoint with native OpenMetrics/Prometheus export format via prometheus-cpp. This is served at both `/` and `/metrics`.
- `--metrics-format` flag can be set to `JSON` (default) or `OpenMetrics.` `JSON` remains the default for backwards compatibility, but is now deprecated.
- Per-database metric handles covering query counters, latency histograms, storage gauges (vertex/edge counts, disk/memory usage), and operator counters (`ScanAll` variants, `CreateNode`, `CreateExpand`, etc.). All labelled by database name and UUID.
- Global aggregates preserved for backwards compatibility on the JSON endpoint.
- RAII metric wrappers (`ScopedHistogramTimer`, `ScopedGauge`, `GaugeHandle`, `HistogramHandle`, `CounterHandle`) for safe metric lifetime management.
- `DatabaseMetricsRegistration` RAII guard; automatically deregisters per-database metrics on database drop.

**Breaking**:
- Histogram buckets for JSON endpoint have changed since 3.10, so may report slightly different values for p50, p90, and p99 times.
- `SHOW METRICS INFO` now shows consistent per-database metrics for the currently used database, rather than a mix of current database (for vertex and edge counts) and aggregate metrics (for eveything else).


colinbarry:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Added per-tenant metrics through native OpenMetrics endpoint. **Breaking**: `SHOW METRICS INFO` now displays metrics for the current database, rather than in aggregate. JSON metrics histograms may report slightly different value due to recalibrated buckets. [#3911](https://github.com/memgraph/memgraph/pull/3911)
    - [x] https://github.com/memgraph/documentation/pull/1635

gitbuda:
How much does the Memgraph binary size change with this (if I'm not wrong, we should have that info somewhere under the benchmarks 👀)?

src/flags/general.cpp
```diff
@@ -45,6 +46,26 @@ DEFINE_VALIDATED_int32(monitoring_port, 7444,
 // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
 DEFINE_VALIDATED_int32(metrics_port, 9091, "Port on which the Memgraph server for exposing metrics should listen.",
                        FLAG_IN_RANGE(0, std::numeric_limits<uint16_t>::max()));
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
+DEFINE_VALIDATED_string(metrics_format, "JSON",
+                        "Format for the metrics endpoint. Supported values: OpenMetrics, JSON. JSON is deprecated.", {
+                          if (value == "OpenMetrics") return true;
+                          if (value == "JSON") {
```
as51340, line 53: Do these 2 warnings happen at the startup or 1 at startup and second when the flag is parsed? If

src/metrics/prometheus_metrics.hpp
```diff
@@ -0,0 +1,507 @@
...
+  std::vector<MetricInfo> GetGlobalMetricsInfo() const;
+  std::vector<MetricInfo> GetGlobalMetricsInfoForLegacyJson();
+
+  prometheus::Registry &registry() { return registry_; }
+
+  GlobalMetricHandles global;
+
+ private:
+  struct DatabaseEntry {
+    std::string db_name;
```
as51340, line 278: shouldn't you use db uuid instead of db name +ref count?

src/metrics/prometheus_metrics.cpp
```diff
@@ -0,0 +1,1922 @@
...
+      "SwapMainUUIDRpcSuccess",
+      "GetDatabaseHistoriesRpcFail",
+      "GetDatabaseHistoriesRpcSuccess",
+      "UpdateDataInstanceConfigRpcFail",
+      "UpdateDataInstanceConfigRpcSuccess",
+  };
+  return std::ranges::find(kLegacyCoordinatorDeltaMetrics, name) != kLegacyCoordinatorDeltaMetrics.end();
+}
+
+// 16 buckets covering 10µs to 120s
```
as51340, line 70: either wrong comment or code? up to 60s or 120s

src/metrics/prometheus_metrics.cpp
```diff
@@ -0,0 +1,1922 @@
...
+    r::transform(databases_, std::back_inserter(names), &DatabaseEntry::db_name);
+  }
+
+  auto const snaps = names | rv::transform([&](auto const &name) { return ResolveStorageSnapshot(name); }) |
+                     r::to<std::vector<StorageSnapshot>>();
+
+  {
+    std::shared_lock const lock{databases_mutex_};
+    for (size_t i = 0; i < names.size(); ++i) {
+      auto it = std::ranges::find_if(databases_, [&](auto const &e) { return e.db_name == names[i]; });
```
as51340, line 1077: yeah I think you should use db uuid

src/metrics/prometheus_metrics.cpp
```diff
@@ -0,0 +1,1922 @@
...
+      it->handles.edge_count->Set(static_cast<double>(snapshot.edge_count));
+      it->handles.disk_usage_bytes->Set(static_cast<double>(snapshot.disk_usage));
+      it->handles.memory_res_bytes->Set(static_cast<double>(snapshot.memory_res));
+    }
+  }
+
+#ifdef MG_ENTERPRISE
+  std::vector<coordination::InstanceStatus> instances;
+  {
+    std::lock_guard const lock{instance_resolver_mutex_};
```
as51340, line 1090: why is this under a mutex?

as51340: because prometheus thread reading while we are updating?

src/metrics/prometheus_metrics.cpp
```diff
@@ -0,0 +1,1922 @@
...
+        ->Set(static_cast<double>(inst.last_succ_resp_ms) / 1000.0);
+  }
+#endif
+}
+
+namespace {
+
+// Compute percentile from a prometheus histogram's cumulative bucket data.
+// Returns 0 if there are no observations.
+double HistogramPercentile(prometheus::Histogram const &h, double quantile) {
```
as51340, line 1135: Doesn't prometheus support an API in which we just "push" observation instead of us calculating percentiles?

src/query/plan/operator.cpp
```diff
@@ -549,8 +493,8 @@ bool Once::OnceCursor::Pull(Frame &, ExecutionContext &context) {
   return false;
 }
 
-UniqueCursorPtr Once::MakeCursor(utils::MemoryResource *mem) const {
-  memgraph::metrics::IncrementCounter(memgraph::metrics::OnceOperator);
+UniqueCursorPtr Once::MakeCursor(utils::MemoryResource *mem, metrics::DatabaseMetricHandles *metric_handles) const {
+  if (metric_handles) metric_handles->once_operator->Increment();
```
as51340, line 497: shouldn't query events go into global metrics instead of per-db?

src/query/plan/operator.hpp
```diff
@@ -612,7 +616,7 @@ class ScanAllByEdge : public memgraph::query::plan::ScanAll {
                 Symbol node2_symbol, EdgeAtom::Direction direction, const std::vector<storage::EdgeTypeId> &edge_types,
                 storage::View view = storage::View::OLD);
   bool Accept(HierarchicalLogicalOperatorVisitor &visitor) override;
-  UniqueCursorPtr MakeCursor(utils::MemoryResource *) const override;
+  UniqueCursorPtr MakeCursor(utils::MemoryResource *, metrics::DatabaseMetricHandles * = nullptr) const override;
```
as51340, line 619: same for streams, shouldn't they be part of global metrics?

src/storage/v2/durability/durability.cpp
```diff
@@ -776,8 +769,6 @@ std::optional<RecoveryInfo> Recovery::RecoverData(
                                     indices_constraints,
                                     config.salient.items.properties_on_edges);
 
-  memgraph::metrics::Measure(memgraph::metrics::SnapshotRecoveryLatency_us,
-                             std::chrono::duration_cast<std::chrono::microseconds>(timer.Elapsed()).count());
```
as51340, line 780: Where did this go?

src/storage/v2/inmemory/edge_property_index.cpp
```diff
@@ -253,27 +254,23 @@ auto InMemoryEdgePropertyIndex::PopulateIndex(PropertyId property, utils::SkipLi
 bool InMemoryEdgePropertyIndex::PublishIndex(PropertyId property, uint64_t commit_timestamp) {
   auto index = GetIndividualIndex(property);
   if (!index) return false;
-  index->Publish(commit_timestamp);
+  auto *gauge = metric_handles_ ? metric_handles_->active_edge_property_indices : nullptr;
```
as51340, line 257: why would metric_handles_ ever be a nullptr?

src/storage/v2/inmemory/edge_type_index.cpp
```diff
@@ -193,27 +194,23 @@ bool InMemoryEdgeTypeIndex::RegisterIndex(EdgeTypeId edge_type, ActiveIndicesUpd
...
-  memgraph::metrics::IncrementCounter(memgraph::metrics::ActiveEdgeTypeIndices);
+  gauge_ = metrics::ScopedGauge{gauge};
 }
 
-InMemoryEdgeTypeIndex::IndividualIndex::~IndividualIndex() {
-  if (status_.IsReady()) {
-    memgraph::metrics::DecrementCounter(memgraph::metrics::ActiveEdgeTypeIndices);
-  }
-}
+InMemoryEdgeTypeIndex::IndividualIndex::~IndividualIndex() = default;
```
as51340, line 207: default it in the declaration

src/storage/v2/inmemory/edge_type_index.cpp
```diff
@@ -327,6 +324,22 @@ void InMemoryEdgeTypeIndex::DropGraphClearIndices() {
   });
 }
 
+void InMemoryEdgeTypeIndex::SetMetricHandles(metrics::DatabaseMetricHandles *metric_handles) {
```
as51340, line 327: can we somehow avoid this function and give to the storage in the constructor DatabaseMetricHandle? Then in the construction we can pass this handle to constraints, indices etc. 
Also, why are we reading now count in idx?

src/storage/v2/inmemory/storage.cpp
```diff
@@ -352,6 +345,7 @@ InMemoryStorage::InMemoryStorage(Config config, std::optional<free_mem_fn> free_
     // Disable ttl until after recovery and role switch / write enabled
     ttl_.SetUserCheck([]() -> bool { return false; });
     // Recover data
+    utils::Timer recovery_timer;
```
as51340, line 348: found where it went 😄

src/storage/v2/storage.hpp
```diff
@@ -1128,6 +1107,9 @@ class Storage {
   IsolationLevel isolation_level_;
   StorageMode storage_mode_;
 
+  metrics::DatabaseMetricHandles *metric_handles_{nullptr};
+  std::optional<double> snapshot_recovery_latency_s_;
```
as51340, line 1111: why do we need to cache it?

src/utils/metrics_timer.hpp
```diff
@@ -11,26 +11,26 @@
...
-#include "utils/event_histogram.hpp"
-
 #include <chrono>
 
+#include <prometheus/histogram.h>
+
 namespace memgraph::utils {
 
-template <typename TDuration = std::chrono::microseconds>
 class MetricsTimer {
```
as51340, line 20: rename so it's clear it is connected to Prometheus

conan.lock
```diff
@@ -28,6 +29,7 @@
         "ctre/3.10.0#3f970a74e5cc3ba1a8ccb8e7f513062f%1755247138.192",
         "croncpp/2023.03.30#7ca459a1acfa234e759182a8bafa7ddf%1722442009.811",
         "cppitertools/2.2#387562fa5016e7692d673955f66d2e7a%1741517127.057",
+        "civetweb/1.16#f4a42c8aa0bc0fe4bbe2a2270af0cea2%1750148475.207",
```
as51340, line 32: what's this lib for?

src/metrics/prometheus_metrics.hpp
```diff
@@ -0,0 +1,507 @@
...
+ public:
+  PrometheusMetrics();
+
+  PrometheusMetrics(PrometheusMetrics const &) = delete;
+  PrometheusMetrics(PrometheusMetrics &&) = delete;
+  PrometheusMetrics &operator=(PrometheusMetrics const &) = delete;
+  PrometheusMetrics &operator=(PrometheusMetrics &&) = delete;
+  ~PrometheusMetrics() = default;
+
+  DatabaseMetricHandles *AddDatabase(std::string_view db_name);
```
as51340, line 259: db UUID

src/metrics/prometheus_metrics.hpp
```diff
@@ -0,0 +1,507 @@
...
+#include <unordered_map>
+#include <variant>
+#include <vector>
+
+#include <prometheus/counter.h>
+#include <prometheus/gauge.h>
+#include <prometheus/histogram.h>
+#include <prometheus/registry.h>
+
+#include "coordination/include/coordination/instance_status.hpp"
```
as51340, line 32: metrics now rely on coordination lib, this will influence build time right? Can you fwd declare it

src/http_handlers/metrics.hpp
as51340, line 121: remove return, it is void function

src/metrics/prometheus_metrics.cpp
as51340: ok I figured out this is for legacy JSON. In that case can we keep using old histogram that was in event_histogram? That one was pretty good if I remember correctly

src/metrics/prometheus_metrics.cpp
```diff
@@ -0,0 +1,1922 @@
...
+                                prometheus::Histogram const &h) {
+  // Histograms are stored in seconds; multiply by 1e6 to emit microseconds in SHOW METRICS INFO output.
+  for (auto const [quantile, label] :
+       {std::pair{0.50, "_us_50p"}, std::pair{0.90, "_us_90p"}, std::pair{0.99, "_us_99p"}}) {
+    out.push_back({name + label, type, "Histogram", HistogramPercentile(h, quantile) * 1e6});
+  }
+}
+
+// Compute percentile from merged cumulative bucket counts (same boundaries across all histograms).
+double MergedHistogramPercentile(std::vector<prometheus::ClientMetric::Histogram> const &hdatas, double quantile) {
```
as51340, line 1166: if I understand correctly this is used to calculate global histogram, kinda like a combination from each histogram on its own. Why would that be useful? For example, I don't see why it is relevant to see combined histogram for snapshot creation for all databases

src/metrics/prometheus_metrics.cpp
```diff
@@ -0,0 +1,1922 @@
...
+#include "utils/logging.hpp"
+
+namespace r = std::ranges;
+namespace rv = r::views;
+
+namespace memgraph::metrics {
+
+namespace {
+
+using prometheus::ClientMetric;
```
as51340, line 34: unused

src/metrics/prometheus_metrics.hpp
```diff
@@ -0,0 +1,507 @@
...
+    // sharing the same prometheus objects. This occurs only in unit tests that
+    // simulate multi-node setups (e.g. main + replicas) in-process.
+    uint32_t ref_count{1};
+  };
+
+  StorageSnapshot ResolveStorageSnapshot(std::string_view db_name) const;
+
+  prometheus::Registry registry_;
+  mutable std::shared_mutex databases_mutex_;
+  mutable std::mutex legacy_json_delta_mutex_;
```
as51340, line 290: what does this mutex protect?

src/metrics/prometheus_metrics.hpp
```diff
@@ -0,0 +1,507 @@
...
+  std::unordered_map<std::string, int64_t> legacy_json_prev_ha_counter_values_;
+  mutable std::mutex snapshot_resolver_mutex_;
```
as51340, line 292: if we only set lambda at the startup while metrics server isn't yet started, we can remove this mutex and the mutex for instance_status_resolver_. I don't see how these lambdas would change throughout the code

src/dbms/database.cpp
```diff
@@ -70,6 +71,28 @@ Database::Database(storage::Config config, std::function<storage::DatabaseProtec
   } else {
     storage_ = dbms::CreateInMemoryStorage(std::move(config), std::move(invalidator), database_protector_factory);
   }
+
+  auto const should_register_database_metrics =
```
as51340, line 75: for JSON output don't we also now send per-db info?

as51340: I think this makes sense but because of this I have a feeling storage has at a bunch of places nullptr check. Proper thing would be that we don't even create dbms handler and storage for coordinators.

I actually think the tradeoff of seeing 0 for coordinator's default DB is ok compared to bunch of nullptr checks (if this is the reason of them and not me hallucinating)

src/dbms/database.cpp
```diff
@@ -70,6 +71,28 @@ Database::Database(storage::Config config, std::function<storage::DatabaseProtec
...
+      !(FLAGS_metrics_format == "OpenMetrics" && flags::CoordinationSetupInstance().IsCoordinator());
+  if (should_register_database_metrics) {
+    metrics_.reset(metrics::Metrics().AddDatabase(storage_->name()));
+  }
+  storage_->SetMetricHandles(metrics_.get());
```
as51340, line 80: Also one design point which I think could simplify the code. Instead of having SetMetricHandles method at a lot of places, can you all the work you're currently doing manually at the storage constructor? There you could glue it with constraints, indices...

as51340 (changes requested):
Good work! I will do another round next week after you reply/fix

src/dbms/database.cpp
as51340: does legacy JSON now do aggregation for all data across all DBs? Did somebody ask that or ? I don't see this idea being very fluent, why combine everything together 🤔

as51340:
Observation: I started HA cluster with the following configuration:
```yaml
instance_1:
    args:
        - "--bolt-port=7687"
        - "--management-port=13011"
        - "--storage-wal-enabled=true"
        - "--storage-snapshot-interval-sec=3600"
        - "--storage-snapshot-on-exit=false"
        - "--telemetry-enabled=false"
        - "--monitoring-port=7444"
        - "--metrics-port=9091"
        - "--metrics-format=JSON"
        - "--log-level=TRACE"
        - "--schema-info-enabled=true"
        - "--storage-gc-cycle-sec=3"
          #        - "--query-modules-directory=/home/andi.linux/Memgraph/code/memgraph/mage/python"
    log_file: "/tmp/ha_demo/instance_1.log"
    data_directory: "/tmp/ha_demo/instance_1"
    setup_queries: []
instance_2:
    args:
        - "--bolt-port=7688"
        - "--management-port=13012"
        - "--storage-wal-enabled=true"
        - "--storage-snapshot-interval-sec=3600"
        - "--storage-snapshot-on-exit=false"
        - "--telemetry-enabled=false"
        - "--monitoring-port=7445"
        - "--metrics-port=9092"
        - "--metrics-format=JSON"
        - "--log-level=TRACE"
        - "--schema-info-enabled=true"
        - "--storage-gc-cycle-sec=3"
          #- "--query-modules-directory=/home/andi.linux/Memgraph/code/memgraph/mage/python"
    log_file: "/tmp/ha_demo/instance_2.log"
    data_directory: "/tmp/ha_demo/instance_2"
    setup_queries: []
instance_3:
    args:
        - "--bolt-port=7689"
        - "--management-port=13013"
        - "--storage-wal-enabled=true"
        - "--storage-snapshot-interval-sec=3600"
        - "--storage-snapshot-on-exit=false"
        - "--telemetry-enabled=false"
        - "--monitoring-port=7446"
        - "--metrics-port=9093"
        - "--metrics-format=JSON"
        - "--log-level=TRACE"
        - "--schema-info-enabled=true"
        - "--storage-gc-cycle-sec=3"
          #- "--query-modules-directory=/home/andi.linux/Memgraph/code/memgraph/mage/python"
    log_file: "/tmp/ha_demo/instance_3.log"
    data_directory: "/tmp/ha_demo/instance_3"
    setup_queries: []
coordinator_2:
    args:
        - "--bolt-port=7692"
        - "--coordinator-id=2"
        - "--coordinator-port=10112"
        - "--management-port=12122"
        - "--coordinator-hostname=localhost"
        - "--telemetry-enabled=false"
        - "--log-level=TRACE"
        - "--monitoring-port=7448"
        - "--metrics-port=9095"
        - "--metrics-format=JSON"
        - "--instance-health-check-frequency-sec=5"
        - "--instance-down-timeout-sec=30"
          #- "--query-modules-directory=/home/andi.linux/Memgraph/code/memgraph/mage/python"
    log_file: "/tmp/ha_demo/coordinator_2.log"
    data_directory: "/tmp/ha_demo/coordinator_2"
    setup_queries: []
coordinator_3:
    args:
        - "--bolt-port=7693"
        - "--coordinator-id=3"
        - "--coordinator-port=10113"
        - "--management-port=12123"
        - "--coordinator-hostname=localhost"
        - "--telemetry-enabled=false"
        - "--log-level=TRACE"
        - "--monitoring-port=7449"
        - "--metrics-port=9096"
        - "--metrics-format=JSON"
        - "--instance-health-check-frequency-sec=5"
        - "--instance-down-timeout-sec=30"
          #- "--query-modules-directory=/home/andi.linux/Memgraph/code/memgraph/mage/python"
    log_file: "/tmp/ha_demo/coordinator_3.log"
    data_directory: "/tmp/ha_demo/coordinator_3"
    setup_queries: []
coordinator_1:
    args:
        - "--bolt-port=7691"
        - "--coordinator-id=1"
        - "--coordinator-port=10111"
        - "--management-port=12121"
        - "--coordinator-hostname=localhost"
        - "--telemetry-enabled=false"
        - "--log-level=TRACE"
        - "--monitoring-port=7447"
        - "--metrics-port=9094"
        - "--metrics-format=JSON"
        - "--instance-health-check-frequency-sec=5"
        - "--instance-down-timeout-sec=30"
          #- "--query-modules-directory=/home/andi.linux/Memgraph/code/memgraph/mage/python"
    log_file: "/tmp/ha_demo/coordinator_1.log"
    data_directory: "/tmp/ha_demo/coordinator_1"
    setup_queries: [
      "ADD COORDINATOR 1 WITH CONFIG {'bolt_server': 'localhost:7691', 'coordinator_server': 'localhost:10111', 'management_server': 'localhost:12121'}",
      "ADD COORDINATOR 2 WITH CONFIG {'bolt_server': 'localhost:7692', 'coordinator_server': 'localhost:10112', 'management_server': 'localhost:12122'}",
      "ADD COORDINATOR 3 WITH CONFIG {'bolt_server': 'localhost:7693', 'coordinator_server': 'localhost:10113', 'management_server': 'localhost:12123'}",
      "REGISTER INSTANCE instance_1 WITH CONFIG {'bolt_server': 'localhost:7687', 'management_server': 'localhost:13011', 'replication_server': 'localhost:10001'}",
      "REGISTER INSTANCE instance_2 WITH CONFIG {'bolt_server': 'localhost:7688', 'management_server': 'localhost:13012', 'replication_server': 'localhost:10002'}",
      "REGISTER INSTANCE instance_3 WITH CONFIG {'bolt_server': 'localhost:7689', 'management_server': 'localhost:13013', 'replication_server': 'localhost:10003'}",
      "SET INSTANCE instance_1 TO MAIN"
    ]
```

and prometheus-exporter with:
```bash
python3 mg_exporter.py --type=HA --config-file=ha_config.yaml
```

File `ha_config.yaml` being:
```yaml
# Template for the configuration
exporter:
  port: 9115
  pull_frequency_seconds: 5
instances:
  - name: coord1
    url: http://127.0.0.1
    port: 9094
    type: coordinator
  - name: coord2
    url: http://127.0.0.1
    port: 9095
    type: coordinator
  - name: coord3
    url: http://127.0.0.1
    port: 9096
    type: coordinator
  - name: data1
    url: http://127.0.0.1
    port: 9091
    type: data_instance
  - name: data2
    url: http://127.0.0.1
    port: 9092
    type: data_instance
  - name: data3
    url: http://127.0.0.1
    port: 9093
    type: data_instance
```

These are log errors I am seeing:

```txt
2026-04-17 12:21:03,804 [ERROR]: Error occurred while updating metrics: No label names were set when constructing counter:ReplicaRecoveryFail
2026-04-17 12:21:03,804 [INFO]: Send update to Prometheus for instance data1
2026-04-17 12:21:08,816 [ERROR]: Error occurred while updating metrics: No label names were set when constructing counter:ReplicaRecoveryFail
2026-04-17 12:21:08,816 [INFO]: Send update to Prometheus for instance data2
2026-04-17 12:21:13,831 [ERROR]: Error occurred while updating metrics: No label names were set when constructing counter:ReplicaRecoveryFail
2026-04-17 12:21:13,831 [INFO]: Send update to Prometheus for instance data3
2026-04-17 12:21:18,842 [INFO]: Send update to Prometheus for instance coord1
```

as51340:
Managed to connect full stack: Memgraph HA + Prometheus + Grafana 💪 
Screenshot 2026-04-17 at 12 44 43

src/metrics/prometheus_metrics.cpp
colinbarry: Wrong comment. Fixed.

src/http_handlers/metrics.hpp
colinbarry: Fixed! (In my defence, this weird `return void` was already in origin/master!)

src/metrics/prometheus_metrics.cpp
colinbarry: Confirmed. `ClientMetric` is now fully qualified, so I'll remove the alias.

conan.lock
colinbarry: Transitive dependency of prometheus-cpp. Can see this at https://conan.io/center/recipes/prometheus-cpp under dependencies.

src/utils/metrics_timer.hpp
colinbarry: Agree. In fact, I think this isn't really generic enough for utils - it is tightly coupled to prometheus histograms, so I will move it into the metrics folder.

src/query/plan/operator.cpp
colinbarry: I used the approach of any metric in the context of a database be a per-database metric. I think it makes more sense this way, as if customers want the aggregate they can easily just use the PromQL sum in Prometheus/Victoria Metrics to compute it. Can discuss if you have other ideas?

src/query/plan/operator.hpp
colinbarry: I used the approach of any metric in the context of a database be a per-database metric. I think it makes more sense this way, as if customers want the aggregate they can easily just use the PromQL sum in Prometheus/Victoria Metrics to compute it. Can discuss if you have other ideas?

src/metrics/prometheus_metrics.hpp
colinbarry: Protects the metrics' list of database. We need to iterate the list when the endpoint is scraped, but this can occur concurrently with an `AddDatabase` or `RemoveDatabase`. Will make this clearer by putting the mutex and its protected state into a struct.

src/storage/v2/inmemory/edge_type_index.cpp
colinbarry: This pattern, with the destructor for index structs defaulted in the cpp, already existed in the code base. But you're right, there is no good reason, and so I've changed this (and the others).

src/metrics/prometheus_metrics.hpp
colinbarry: You're right - I've remove the pointless synchronisation.

src/metrics/prometheus_metrics.hpp
colinbarry: Switched to a forward declaration.

src/flags/general.cpp
colinbarry: Once at startup, once again when the flag is parsed. Which means, if they specify `--open-metrics=JSON`, which is (currently) the default, they get duplicate warnings.

src/storage/v2/storage.hpp
colinbarry: The caching is necessary because recovery happens before the metric handle exists. Not obvious though, so have added a comment.

src/metrics/prometheus_metrics.cpp
colinbarry: I'm not sure of the merit of supporting the old histogram code as well as the Prometheus based one. In theory, we're computing the same calculation here as the old code did, albeit with slightly different bucketing. Thoughts?

src/metrics/prometheus_metrics.cpp
colinbarry: This preserves the pre-multi-tenancy behaviour for the deprecated/backwards compatible JSON endpoint. The merged result is equivalent to the old behaviour, where there was a single global histogram for the whole process.

src/metrics/prometheus_metrics.cpp
colinbarry: Done - using UUID internally to reference databases uniquely.

src/storage/v2/inmemory/storage.cpp
colinbarry: ✅

src/storage/v2/durability/durability.cpp
colinbarry: Resolving, as you found it. Replaced these with an RAII wrapper to for timings.

src/metrics/prometheus_metrics.hpp
colinbarry: Done - using UUID internally to reference databases uniquely.

src/metrics/prometheus_metrics.hpp
colinbarry: Done - using UUID internally to reference databases uniquely.

src/storage/v2/inmemory/edge_property_index.cpp
colinbarry: Unit tests. The alternative is to instantiate a Prometheus metrics object in all unit tests that use storage in any capacity. Which isn't a big deal - it's a lightweight object, we don't need to spin up the endpoint - but adds more noise to (currently) 60+ difference unit test files. I'm on the fence about this, so open to discussion -  essentially, we decide what is worse: having to create a metrics objects in every storage facing unit test, which introduces a fair amount of test boilerplate; or having nullptr checks.

src/dbms/database.cpp
colinbarry: This PR keeps the behaviour of the old JSON endpoint intact, so no multi-database support. OpenMetrics for multi-tenant metrics thus becomes a forcing function.

My bad: I assumed the JSON endpoint aggregated these metrics, I didn't realise it only observed the "memgraph" db. Restored the old behaviour in this case.

src/metrics/prometheus_metrics.cpp
colinbarry: Refactored, mutex no longer needed.

src/storage/v2/inmemory/edge_type_index.cpp
colinbarry: Good idea: refactored.

src/dbms/database.cpp
colinbarry: Done

src/metrics/prometheus_metrics.cpp
as51340: I mean yes but why would we change the old behavior? My intuition: leave old JSOn as it was.. I hope you understand what I mean, this PR adds per-db tracking with open metrics format. Why would it change histogram for old code?

src/query/plan/operator.cpp
as51340: Currently I think both is fine. But if our queries start becoming smarter and we add cross-database queries then this doesn't hold anymore. In my view of databases, query >> storage. Ask product for opinion maybe

src/query/plan/operator.hpp
as51340: same

src/storage/v2/inmemory/edge_property_index.cpp
as51340: I am leaning towards ugly unit tests compared to null check in src. Maybe you can have some small wrapper which by default does nothing (imitates nullptr) and in the src does what it needs to do.

src/metrics/prometheus_metrics.cpp
as51340: I think there was some comment above: in the old code, there was no aggregation

src/metrics/prometheus_metrics.hpp
as51340: I think usage around db_name can still be better. In general I don't like we are tracking when the last database with the name goes out of scope. I will write here few things that are all related to the same thing:
- StorageSnapshotResolved doesn't need db_name as argument, it can leave with UUID
- UpdateGauges and GetDbMetricsInfo hence don't need db_name
- For RemoveDatabase instead of tracking when the last db goes out of scope, you can instead add UUID as Prometheus's internal label. Such labels start with double underscore https://training.promlabs.com/training/relabeling/introduction-to-relabeling/hidden-labels-and-metadata/

src/metrics/prometheus_metrics.hpp
```diff
@@ -0,0 +1,527 @@
...
+  StorageSnapshot ResolveStorageSnapshot(std::string_view db_name) const;
+
+  prometheus::Registry registry_;
+
+  struct {
+    mutable std::shared_mutex mutex;
+    std::list<DatabaseEntry> entries;
+  } databases_;
+
+  mutable std::mutex json_ha_metrics_delta_mutex_;
```
as51340: if only one thread at the time can invoke GetGlobalMetricsForJson, is this mutex really needed?

src/metrics/prometheus_metrics.cpp
```diff
@@ -0,0 +1,1983 @@
...
+              .gc_latency_seconds = &gc_latency_family_.Add(labels, kLatencyBuckets),
+              .gc_skiplist_cleanup_latency_seconds = &gc_skiplist_cleanup_latency_family_.Add(labels, kLatencyBuckets),
+          },
+  });
+  return &databases_.entries.back().handles;
+}
+
+void PrometheusMetrics::RemoveDatabase(DatabaseMetricHandles const *handles) {
+  std::lock_guard const lock{databases_.mutex};
+  auto it = r::find_if(databases_.entries, [handles](auto const &e) { return &e.handles == handles; });
```
as51340: A bit unintuitive, I would rather remove based on UUID than based on pointer. list is stable but UUID is more intuitive

as51340: you will probably need friend class for DatabseMetricsRegistration

src/storage/v2/indices/edge_property_index.hpp
```diff
@@ -12,6 +12,11 @@
 #pragma once
 
 #include <map>
+
+namespace memgraph::metrics {
+struct DatabaseMetricHandles;
+}  // namespace memgraph::metrics
+
```
as51340: not needed anymore?

src/storage/v2/indices/edge_type_index.hpp
```diff
@@ -20,6 +20,10 @@
 #include <variant>
 #include <vector>
 
+namespace memgraph::metrics {
+struct DatabaseMetricHandles;
+}  // namespace memgraph::metrics
+
```
as51340: not needed anymore?

src/storage/v2/indices/edge_type_property_index.hpp
```diff
@@ -11,6 +11,10 @@
 
 #pragma once
 
+namespace memgraph::metrics {
+struct DatabaseMetricHandles;
+}  // namespace memgraph::metrics
+
```
as51340: not needed anymore?

src/storage/v2/indices/indices.cpp
```diff
@@ -111,15 +111,15 @@ void Indices::UpdateOnEdgeCreation(Vertex *from, Vertex *to, EdgeRef edge_ref, E
...
-      label_index_ = std::make_unique<InMemoryLabelIndex>();
-      label_property_index_ = std::make_unique<InMemoryLabelPropertyIndex>();
-      edge_type_index_ = std::make_unique<InMemoryEdgeTypeIndex>();
-      edge_type_property_index_ = std::make_unique<InMemoryEdgeTypePropertyIndex>();
-      edge_property_index_ = std::make_unique<InMemoryEdgePropertyIndex>();
+      label_index_ = std::make_unique<InMemoryLabelIndex>(metric_handles);
+      label_property_index_ = std::make_unique<InMemoryLabelPropertyIndex>(metric_handles);
+      edge_type_index_ = std::make_unique<InMemoryEdgeTypeIndex>(metric_handles);
+      edge_type_property_index_ = std::make_unique<InMemoryEdgeTypePropertyIndex>(metric_handles);
+      edge_property_index_ = std::make_unique<InMemoryEdgePropertyIndex>(metric_handles);
```
as51340: small improvement but you can pass only the needed field from metric_handles, e.g metric_handles->active_edge_type_indices. Why would some index have access to whole metric handle... I hope you understand what I mean

src/storage/v2/indices/label_index.hpp
```diff
@@ -14,6 +14,10 @@
 #include "storage/v2/vertex.hpp"
 #include "storage/v2/vertex_accessor.hpp"
 
+namespace memgraph::metrics {
+struct DatabaseMetricHandles;
+}  // namespace memgraph::metrics
+
```
as51340: aha, I guess this is fwd declared here so it can be used in in memory and disk versions?

src/storage/v2/inmemory/edge_type_index.cpp
```diff
@@ -193,27 +194,21 @@ bool InMemoryEdgeTypeIndex::RegisterIndex(EdgeTypeId edge_type, ActiveIndicesUpd
...
 
 bool InMemoryEdgeTypeIndex::DropIndex(EdgeTypeId edge_type, ActiveIndicesUpdater const &updater) {
   auto const result = index_.WithLock([&](std::shared_ptr<IndicesContainer const> &indices_container) {
-    {
-      auto const it = indices_container->indices_.find(edge_type);
-      if (it == indices_container->indices_.cend()) return false;
-    }
+    auto const it = indices_container->indices_.find(edge_type);
+    if (it == indices_container->indices_.cend()) return false;
+    it->second->gauge_ = {};
```
as51340: why is this reset needed?

as51340 (changes requested):

src/metrics/prometheus_metrics.hpp
colinbarry: Done - we now use UUID as the source of truth for uniqely identifying dbs. This has the advantage that if someone DROPs and CREATEs a database with the same name, the second db will be treated as a different time series than the first. 👍

src/metrics/prometheus_metrics.hpp
colinbarry: Agreed: Removed in subsequent refactor.

src/storage/v2/indices/edge_property_index.hpp
colinbarry: Removed

src/storage/v2/indices/edge_type_index.hpp
colinbarry: Removed

src/storage/v2/indices/edge_type_property_index.hpp
colinbarry: Removed

src/storage/v2/indices/indices.cpp
colinbarry: Done - only passing the actual gauge through to indices, etc.

src/storage/v2/inmemory/edge_type_index.cpp
colinbarry: This is intentional: resetting the `ScopedGauge` decrements the active index-count gauge when an index is dropped. Without it, the gauge would remain stale until the evicted entry is fully destroyed (which may be deferred to some arbitrary point in the future). I think as the metric name is "active", this reflects what the metric covers: once the index is dropped, it is no longer active, even if some clean up is deferred.

src/storage/v2/indices/label_index.hpp
colinbarry: Removed

src/dbms/inmemory/storage_helper.hpp
```diff
@@ -20,15 +21,14 @@ namespace memgraph::dbms {
...
   // Use default safe factory from Storage constructor for basic usage
-  auto storage = std::make_unique<storage::InMemoryStorage>(std::move(config),
-                                                            std::nullopt,
-                                                            std::move(invalidator),
-                                                            std::move(database_protector_factory),
-                                                            db_arena,
-                                                            db_embedding_memory_tracker);
+  auto storage = std::make_unique<storage::InMemoryStorage>(
+      std::move(config), std::nullopt, std::move(invalidator), std::move(metric_handles),
+      std::move(database_protector_factory), db_arena, db_embedding_memory_tracker);
```
as51340, line 30: I guess you could also create this object in the InMemoryStorage, without polluting this dbms helper

src/dbms/database.cpp
```diff
@@ -94,6 +97,17 @@ memory::ArenaPool &Database::Arena() const noexcept { return *db_arena_; }
 
 Database::Database(storage::Config config, std::function<storage::DatabaseProtectorPtr()> database_protector_factory)
     : db_arena_(std::make_unique<memory::ArenaPool>(&db_memory_tracker_)),
+      metrics_(config.salient.uuid, ([&config]() -> metrics::DatabaseMetricHandles {
+                 if (!config.register_metrics) {
+                   return {};
+                 }
+                 auto const should_register =
+                     !(FLAGS_metrics_format == "OpenMetrics" && flags::CoordinationSetupInstance().IsCoordinator());
```
as51340, line 105: Why it's ok for registration to be old JSON format and coordinator?

src/storage/v2/constraints/type_constraints.cpp
```diff
@@ -226,6 +223,8 @@ TypeConstraints::IndividualConstraintPtr TypeConstraints::DropConstraint(LabelId
     auto it = container->constraints_.find({label, property});
     if (it == container->constraints_.end()) return nullptr;
     if (it->second->type != type) return nullptr;
+    // Clear ScopedGauge, which decrements the gauge backing this metric.
+    it->second->gauge_ = {};
```
as51340: can this be done somehow simpler, in some destructor maybe? Looks like error-prone

as51340 (changes requested):

src/dbms/inmemory/storage_helper.hpp
colinbarry: I assume you mean the `metric_handles`? I don't think that's possible - (per-tenant) metrics are a natural part of the `Database`, and creating them in the storage level would invert the natural relationship, and also introduce dependencies between the prometheus lib and storage.

src/dbms/database.cpp
colinbarry: Backwards compatibilty with the legacy JSON endpoint.

With the new OpenMetrics endpoint, we skip per-db registration for coordinators because they don't have a database and so don't report any per-tenant metrics.

But the JSON endpoint, even on coordinators, delivered all metrics - even things like operator counts - despite having no database. These metric values would be 0

src/storage/v2/constraints/type_constraints.cpp
colinbarry: Yeah, on reflection, you're right. The previous code decremented these counters when the shared_ptr object destructs, rather than immedaitely on `DROP`, so let's match that behaviour.

colinbarry: Done. Removed these `= {}`, and instead rely on natural destruction to decrement the object via RAII.

as51340 (approved):

PR #3914, opened 2026-03-16
Fix embeddings crash on parallel requests 

antejavor:
### What?

The embeddings.py query module crashes when multiple Bolt sessions call `embeddings.text()` or `embeddings.node_sentence()` concurrently. This PR fixes the crash and adds e2e parallel tests.

### Why?

Each request created a new `SentenceTransformer` instance from scratch. Under concurrent load this caused:
- **Race conditions** — multiple threads importing and instantiating `SentenceTransformer` simultaneously led to crashes.
- **GPU memory waste** — every request loaded a fresh model copy, then attempted fragile manual cleanup (`model.to("cpu")`, `del model`, `gc.collect()`, `torch.cuda.empty_cache()`, `torch.cuda.synchronize()`).
- **Unnecessary latency** — model loading takes several seconds per request, even when the same model is reused.

### How?

Replaced per-request model instantiation with a **thread-safe global model cache** using double-checked locking:

- **Added `_get_or_load_model()`** — caches models by `(model_name, device)` key, protected by `threading.Lock` with double-check pattern to ensure a single load under contention.
- **Replaced all `SentenceTransformer()` calls** in `cpu_compute()` and `single_gpu_compute()` with `_get_or_load_model()`.
- **Removed manual GPU cleanup** — the `finally` block that did `model.to("cpu")` / `del model` / `gc.collect()` / `torch.cuda.empty_cache()` / `torch.cuda.synchronize()` is no longer needed since models stay cached.
- **Removed dead `try/finally: pass`** and flattened control flow in `single_gpu_compute()`.
- **Fixed typo** in comment: `shoiuld` → `should`.

### Tests?

Added test_parallel_embeddings.py — runs against a live Memgraph instance with 4 concurrent Bolt connections synchronized via `threading.Barrier`:

| Test | Description |
|---|---|
| `test_concurrent_text_embed` | 4 threads call `embeddings.text()` with different sentences in parallel, assert all return `success=True` with dimension 384 |
| `test_concurrent_node_sentence` | 4 threads call `embeddings.node_sentence()` on shared graph nodes in parallel, assert correctness and clean up |

Auto-skips if Memgraph isn't running or the embeddings module isn't loaded. The base embedding test will fail in this case. 

### Notes to reviewer

- The model cache is intentionally **never evicted** — models stay loaded for the lifetime of the Memgraph process. This matches the expected usage pattern (same model across requests). If memory pressure becomes a concern, an LRU eviction policy could be added later.
- `multi_gpu_compute()` is unaffected — it already spawns separate worker processes via `ProcessPoolExecutor` with `embed_worker`, so each worker loads its own model.
- The removed GPU cleanup code was unreliable (referenced `freed_memory` before assignment on error paths) and is no longer needed with caching.

### DIFF
--- a/mage/tests/e2e/test_module.py
+++ b/mage/tests/e2e/test_module.py
@@ -273,7 +273,8 @@ def _test_online(test_dir: Path, db: Memgraph):
 
 def _test_standalone(test_file: Path):
     """
-    Run a standalone Python test script via pytest subprocess.
+    Run a self-contained Python test script via pytest subprocess.
+    These scripts manage their own connections and cleanup.
     """
     result = subprocess.run(
         ["python3", "-m", "pytest", str(test_file), "-v"],


antejavor:
The embeddings.py query module crashes when multiple Bolt sessions call embeddings.text() or embeddings.node_sentence() concurrently. This PR fixes the crash and adds e2e parallel tests.


---

### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    -  Fixed crash in the `embeddings` query module when multiple Bolt sessions called `embeddings.text()` or `embeddings.node_sentence()` in parallel, which fails on the model init. Model init is now thread-safe with a global cache using double-checked locking, preventing race conditions and eliminating redundant model instantiation per request.  https://github.com/memgraph/memgraph/pull/3914

mage/python/embeddings.py
```diff
@@ -16,6 +17,26 @@
...
 
+_model_cache = {}
+_model_lock = threading.Lock()
+
+
+def _get_or_load_model(model_name: str, device: str = "cpu"):
+    """Thread-safe model loading with caching."""
+    key = (model_name, device)
+    if key in _model_cache:
+        return _model_cache[key]
```
imilinovic, line 28: You are accessing the cache without locking it

imilinovic: From what I see this is fine because GIL makes dict lookup atomic, but still prefer locking to avoid stale data

mage/tests/e2e/embeddings_test/test_parallel_embeddings.py
```diff
@@ -0,0 +1,158 @@
...
+            try:
+                barrier.wait(timeout=10)
+                results[idx] = run_node_sentence()
+            except Exception as exc:
+                errors[idx] = exc
+
+        with ThreadPoolExecutor(max_workers=NUM_WORKERS) as pool:
+            futs = [pool.submit(worker, i) for i in range(NUM_WORKERS)]
+            for f in as_completed(futs):
+                f.result()
```
imilinovic: i don't think this is truly parallel just concurrent due to GIL, is this an issue?

mage/tests/e2e/embeddings_test/test_parallel_embeddings.py
```diff
@@ -0,0 +1,158 @@
...
+    conn.close()
+    return {
+        "success": row[0],
+        "num": row[1],
+        "dim": row[2],
+        "dimension": row[3],
+    }
+
+
+@pytest.fixture(scope="module")
```
mattkjames7, line 50: I don't see this test running in the logs in CI (at least, it's not jumping out at me). I see the others, e.g.:

```
2026-03-16T13:47:12.9565888Z test_module.py::test_end2end[util_module_test-util_func_test_int] PASSED [ 87%]
2026-03-16T13:47:17.0599567Z test_module.py::test_end2end[embeddings_test-test_embed] PASSED          [ 87%]
2026-03-16T13:47:17.0968518Z test_module.py::test_end2end[embeddings_test-test_cpu_compute] PASSED    [ 88%]
2026-03-16T13:47:17.1479942Z test_module.py::test_end2end[embeddings_test-test_verify_embeddings] PASSED [ 88%]
2026-03-16T13:47:17.1586028Z test_module.py::test_end2end[embeddings_test-test_info] PASSED           [ 88%]
2026-03-16T13:47:17.1695668Z test_module.py::test_end2end[temporal_test-test_date] PASSED             [ 88%]
```
 ~~Does the script `mage/tests/test_e2e` need updating in order to pick this test up?~~
Sorry wrong file, this one probably needs editing to pick it up: `mage/tests/e2e/test_module.py`

mage/python/embeddings.py
antejavor: Opus actually suggested this because it is a problem with the initial construction of the model, avoiding overhead on each subsequent lock if you just want to get the model from the cache. Initially, I had the same thought process as you did here. 

So if there is no local model, we are locking since just the construction is the problem -> https://en.wikipedia.org/wiki/Double-checked_locking

mage/tests/e2e/embeddings_test/test_parallel_embeddings.py
antejavor: This is a good catch, I think it is concurrent sufficiently (not parallel) to trigger the problem since fetching the model takes a lot of time, but this is a good catch.  I have used the actual process pool before to trigger the error, will make this use processes so it is truly parallel and we are certain.

mage/tests/e2e/embeddings_test/test_parallel_embeddings.py
antejavor: This is correct. 

I have run this locally.  This is where the test is coming from. I was thinking, should this be a part of CI or manual, and after that, how to even incorporate this into a mage testing infra. I had the same problem in another PR.  

IMO, the whole structure of e2e testing is a bit rigid, since it expects input -> output; you cannot build parallelism inside. 

I am all in for opening a specific issue, where inside the e2e tests we would have a custom Python test (we should probably think about this a bit) 

```(env) antejavor@lima-memgraph:~/repos/memgraph/mage/tests/e2e/embeddings_test$  cd /home/antejavor.linux/repos/memgraph/mage/tests/e2e/embeddings_test && source /home/antejavor.linux/repos/memgraph/env/bin/activate && python -m pytest test_parallel_embeddings.py -v --no-header 2>&1
================================================================================================================================================================================================== test session starts ===================================================================================================================================================================================================
collected 2 items                                                                                                                                                                                                                                                                                                                                                                                                        

test_parallel_embeddings.py::TestParallelEmbeddings::test_concurrent_text_embed PASSED                                                                                                                                                                                                                                                                                                                             [ 50%]
test_parallel_embeddings.py::TestParallelEmbeddings::test_concurrent_text_embed_multiple_strings PASSED                                                                                                                                                                                                                                                                                                            [100%]

=================================================================================================================================================================================================== 2 passed in 0.29s ====================================================================================================================================================================================================```

Comment (mattkjames7) [reply mage/tests/e2e/embeddings_test/test_parallel_embeddings.py:50]: I agree that it is very rigid, and that custom Python scripts like this are probably the way to handle it.

So, if I remember correctly, the current test logic looks for subdirectory names ending in `_test` and then the input/output files nested within those. Could we just add an extra bit of logic to also scan for self-contained Python test scripts within the `e2e` directory? Would that work for the other PR you said you had this problem with?

Comment (antejavor) [reply mage/tests/e2e/embeddings_test/test_parallel_embeddings.py:50]: Yes, added a change here.

File: mage/tests/e2e/test_module.py [hunk 4]
```diff
@@ -254,8 +271,26 @@ def _test_online(test_dir: Path, db: Memgraph):
...
+
+
 @pytest.mark.parametrize("test_dir", tests)
 def test_end2end(test_dir: Path, db: Memgraph):
+    # Standalone Python test scripts run in a subprocess
+    if test_dir.is_file() and test_dir.suffix == ".py":
+        _test_standalone(test_dir)
+        return
+
     db.drop_database()
```
mattkjames7, line 295: should this appear before the addition above?

antejavor: I have added the small doc string for contex, so the full Python script should be fully contained.  Since you don't want to depend on the cypher harness that depends on the gqldriver session outside of test, that is why the drop is happening after.

Speaking more broadly, I am more than happy to restructure the test infra here since I will need it, but I think that is for the next PR.  It is confusing because there are now 4 different versions of the e2e test. I don't want this PR to become Mage testing infra change. 

Wdyt?  

My next step would be to open an issue for this, and I can work on a change, and you can review it?

mattkjames7: I think that is a good idea - let's keep as is for now and do the test restructuring elsewhere :joy:

antejavor: Here is the issue -> https://github.com/memgraph/memgraph/issues/3918

PR #3916, opened 2026-03-16
Add Kerberos SSO authentication module

### LINKED ISSUE #3637: Implement Kerberos Auth

DavIvek:
Add Kerberos SSO authentication module                                                                                                                                                                              
                                                                                                                                                                                                                    
  Adds a Kerberos reference auth module that validates client-supplied                                                                                                                                                
  service tickets via GSSAPI and maps the authenticated principal to
  Memgraph roles. Registers "kerberos" as an SSO scheme and plugs into                                                                                                                                                
  the existing auth-module dispatch (same mechanism as saml.py and                                                                                                                                                    
  oidc.py).                                                                                                                                                                                                           
                                                                                                                                                                                                                      
  Role mapping supports two modes via MEMGRAPH_SSO_KERBEROS_ROLE_MAPPING_MODE:                                                                                                                                        
  - ldap (default): queries AD/LDAP for the user's group membership and
    maps groups to roles; optional nested-group resolution via AD's                                                                                                                                                   
    LDAP_MATCHING_RULE_IN_CHAIN                                                                                                                                                                                       
  - principal: static mapping from principal name to roles                                                                                                                                                            
                                                                                                                                                                                                                      
  Tested end-to-end against a real Samba AD DC (KDC + LDAP); GSSAPI and                                                                                                                                               
  ldap3 are mocked in CI via tests/e2e/sso/test_kerberos.py.

Closes #3637


DavIvek:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Added a Kerberos SSO authentication module that validates service tickets via GSSAPI and maps authenticated principals to Memgraph roles (via LDAP group lookup or static principal mapping), enabling seamless enterprise authentication; users can enable the "kerberos" SSO scheme and configure role mapping through `MEMGRAPH_SSO_KERBEROS_ROLE_MAPPING_MODE` based on their environment. [#3916](https://github.com/memgraph/memgraph/pull/3916)
- [x] **https://github.com/memgraph/documentation/pull/1607**

tests/e2e/sso/test_kerberos.py
```diff
@@ -0,0 +1,162 @@
...
+        defaults.update(overrides)
+    for key, value in defaults.items():
+        monkeypatch.setenv(key, value)
+
+
+def _setup_mock_ctx(principal="david@EXAMPLE.COM", complete=True):
+    mock_ctx = MagicMock()
+    mock_ctx.complete = complete
+    mock_ctx.initiator_name = MagicMock()
+    mock_ctx.initiator_name.__str__ = lambda self: principal
```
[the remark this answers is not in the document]
DavIvek: The tests pass — MagicMock specifically supports overriding __str__ via instance assignment (it's one of the magic-method exceptions documented in the unittest.mock docs). Verified that the                  principal/realm/role-mapping assertions get the correct string.

tests/e2e/sso/test_kerberos.py
```diff
@@ -0,0 +1,162 @@
...
+
+mock_gssapi = ModuleType("gssapi")
+mock_gssapi.Name = MagicMock()
+mock_gssapi.NameType = MagicMock()
+mock_gssapi.Credentials = MagicMock()
+mock_gssapi.SecurityContext = MagicMock()
+mock_gssapi.exceptions = ModuleType("gssapi.exceptions")
+mock_gssapi.exceptions.GSSError = type("GSSError", (Exception,), {})
+sys.modules["gssapi"] = mock_gssapi
+sys.modules["gssapi.exceptions"] = mock_gssapi.exceptions
```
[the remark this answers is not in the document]
DavIvek: Same as before, these tests run in their own pytest process via pytest_runner.sh, not in a shared session with other tests.

src/auth/reference_modules/kerberos.py
```diff
@@ -0,0 +1,252 @@
...
+                "{user_dn}", ldap3.utils.conv.escape_filter_chars(user_dn)
+            )
+            conn.search(base_dn, nested_filter, attributes=["cn"])
+            user_groups = {entry.cn.value for entry in conn.entries if entry.cn}
+            member_of = []
+        else:
+            group_entry = conn.entries[0][group_attr] if group_attr in conn.entries[0] else None
+            member_of = group_entry.values if group_entry else []
+            user_groups = None
+        conn.unbind()
```
imilinovic: should do this in finally block of try, except, final,
want to unbind even with exception.

actually cleanest would be to do 

`
try:
       with ldap3.Connection(server, ...) as conn:
               do stuff
       # it should unbind automatically out of scope
`

src/auth/reference_modules/kerberos.py
```diff
@@ -0,0 +1,252 @@
...
+    try:
+        import gssapi
+    except ImportError:
+        return {"authenticated": False, "errors": "gssapi package not installed. Run: pip install gssapi"}
+
+    try:
+        server_name = gssapi.Name(service_principal, gssapi.NameType.kerberos_principal)
+        server_creds = gssapi.Credentials(name=server_name, usage="accept")
+        ctx = gssapi.SecurityContext(creds=server_creds, usage="accept")
+        ctx.step(token_bytes)
```
imilinovic: maybe do something like

`output_token = ctx.step(token_bytes)
  if not ctx.complete:
      return {"authenticated": False, "errors": "GSSAPI
  negotiation incomplete (multi-leg not supported)"}
  if output_token:
      return {"authenticated": False, "errors": "Mutual
  authentication not supported"}`

to reject Kerberos with mutual auth since its not supported atm

src/auth/reference_modules/kerberos.py
```diff
@@ -0,0 +1,252 @@
...
+import re
+
+
+def _load_role_mappings(raw_role_mappings: str) -> dict:
+    if raw_role_mappings:
+        role_mapping = {}
+        for mapping in raw_role_mappings.strip().split(";"):
+            if not mapping.strip():
+                continue
+            mapping_list = mapping.split(":")
```
imilinovic: splitting by `:` is bad in case of bare kerberos (not using LDAP), we can have something like
`MSSQLSvc/db.example.com:1433@CORP.COM : admin` which will get wrongly split/fail... I don't think you'd ever want to use non LDAP kerberos though so might be worth just noting in docs

src/auth/reference_modules/kerberos.py
```diff
@@ -0,0 +1,252 @@
...
+                "authenticated": False,
+                "errors": f"Client realm {client_realm} does not match expected realm {realm}",
+            }
+
+    role_mapping = config["role_mapping"]
+    role_mapping_mode = config["role_mapping_mode"]
+
+    if role_mapping_mode == "ldap":
+        resolved = _resolve_roles_ldap(client_principal, role_mapping, config)
+    else:
```
imilinovic:
```
else if "principal"

else 
     return error
```

src/auth/reference_modules/kerberos.py
```diff
@@ -0,0 +1,252 @@
...
+def _load_config_from_env():
+    config = {}
+    config["keytab"] = os.environ.get("MEMGRAPH_SSO_KERBEROS_KEYTAB", "")
+    config["service_principal"] = os.environ.get("MEMGRAPH_SSO_KERBEROS_SERVICE_PRINCIPAL", "")
+    config["realm"] = os.environ.get("MEMGRAPH_SSO_KERBEROS_REALM", "")
+    config["username_field"] = os.environ.get("MEMGRAPH_SSO_KERBEROS_USERNAME_FIELD", "name")
+    config["role_mapping"] = _load_role_mappings(os.environ.get("MEMGRAPH_SSO_KERBEROS_ROLE_MAPPING", ""))
+
+    # LDAP group-based role mapping (optional)
+    config["role_mapping_mode"] = os.environ.get("MEMGRAPH_SSO_KERBEROS_ROLE_MAPPING_MODE", "principal")
```
imilinovic: maybe it would be better to default to nothing so the user doesn't get weird errors in cause of forgotting to set this

imilinovic: or default to LDAP -> using Kerberos without LDAP (or some other auth protocol) doesn't generally make sense from security standpoint

DavIvek: I thought about keeping it at principal because minimal-config path works out of the box. Defaulting to ldap means a bare `--auth-module-mappings=kerberos` + keytab + role mapping fails until the user also sets LDAP_URI, LDAP_BASE_DN, bind credentials, etc., but I agree overall that putting principal as default doesn't make sense in enterprise context so I'll change that part 👍

imilinovic (approved):

PR #3919, opened 2026-03-17
fix: Unregistering instance moved to the reconciliation loop

as51340:


as51340:
Continuing work here https://github.com/memgraph/memgraph/pull/3922

PR #3922, opened 2026-03-18
fix: Extend and enhance reconciliation loop logic inside HA coordinators

as51340:
## What

Rearchitects coordinator cluster operations to follow a **Raft-first** pattern: commit the desired state to the Raft log and block until majority acknowledgement before returning success. RPCs to data instances (promote, demote, register/unregister replica) are sent
optimistically after commit; failures are retried automatically via the reconciliation loop in `InstanceSuccessCallback`. Removes `EnableWritingOnMainRpc` from the promotion path, fixes a dangling-pointer bug in `ReplicationInstanceConnector`, and simplifies the
`ReconcileClusterState_` early-return guard.

## Why

The previous RPC-first approach required Raft rollbacks on RPC failure, which could themselves fail — leading to `LOG_FATAL` in `UnregisterReplicationInstance` and unrecoverable states. Users also received errors for transient RPC failures that the coordinator could
self-heal. By making the Raft log the single source of truth and treating RPCs as best-effort with automatic retry, the system becomes more resilient and the error handling is dramatically simpler.

## How

- **`RaftState::AppendLogAndWaitForCommit`** (renamed from `AppendClusterUpdate`): blocks via `res->get()` until NuRaft delivers a definitive commit result. If commit fails, the log was not applied — the user can safely retry.

- **`SetReplicationInstanceToMain`**: looks up the instance in Raft state (not in-memory connectors), commits the role change and new UUID to Raft first, then sends `SwapAndUpdateUUID` and `PromoteToMainRpc` best-effort. Removes `EnableWritingOnMainRpc`. If the
connector doesn't exist yet (startup race), returns a soft error instead of asserting.

- **`DemoteInstanceToReplica`**: commits role change to Raft first, then sends `DemoteMainToReplicaRpc` best-effort. Adds `ALREADY_REPLICA` early return.

- **`RegisterReplicationInstance`**: commits to Raft first, then creates the connector and sends `DemoteMainToReplicaRpc` + `RegisterReplicaOnMainRpc` best-effort. Removes the `pop_back` + `RPC_FAILED` rollback path.

- **`UnregisterReplicationInstance`**: commits to Raft first, then sends `UnregisterReplicaRpc` best-effort. Removes the dangerous Raft state revert that could `LOG_FATAL`.

- **`InstanceSuccessCallback`** (reconciliation loop): two new paths reconcile Raft vs. main's actual state:
  1. Replicas in Raft (with `REPLICA` role) but missing from main's `replicas_num_txns` → sends `RegisterReplicaOnMainRpc`.
  2. Replicas reported by main but absent from Raft → sends `UnregisterReplicaRpc`.

- **`ReconcileClusterState_`**: the `kDisconnectedCluster` early-return guard now only exits when there are also no data instances, fixing a bug where a single coordinator with registered data instances would skip reconciliation entirely (no connectors created, no
health checks).

- **`ReplicationInstanceConnector`**: `repl_client_info_` changed from `const ReplicationClientInfo*` to `ReplicationClientInfo` (stored by value), fixing a dangling-pointer bug where the pointer referenced a destroyed caller-local object.

- **`coordinator_ops_status.hpp`**: removed `RPC_FAILED`, `COULD_NOT_PROMOTE_TO_MAIN`, `ENABLE_WRITING_FAILED`. Added `ALREADY_REPLICA`. Added trailing commas to all enums.

- **`interpreter.cpp`**: unified all `RAFT_LOG_ERROR` messages to "Writing to Raft log failed. Please retry the operation." — since the blocking `get()` call guarantees a definitive result, there is no ambiguity about whether the operation took effect.

## Testing

- Updated `tests/e2e/high_availability/distributed_coords.py` (`test_even_number_coords`) to match the new error message.
- Updated `tests/e2e/high_availability/coord_cluster_registration.py` (`test_register_one_coord_with_env_vars_no_instances_alive_on_start`) — the expected cluster state now includes data instances because `ReconcileClusterState_` no longer skips them on a single
coordinator.
- Removed two e2e tests (`test_force_reset_works_after_failed_registration`, `test_force_reset_works_after_failed_registration_and_replica_down`) that tested the old RPC-failure-triggers-force-reset path which no longer applies.

### DIFF
--- a/src/coordination/include/coordination/replication_instance_connector.hpp
+++ b/src/coordination/include/coordination/replication_instance_connector.hpp
@@ -87,7 +87,8 @@ class ReplicationInstanceConnector {
  private:
   ReplicationInstanceClient client_;
   mutable TimedFailureDetector timed_failure_detector_;
+  ReplicationClientInfo repl_client_info_;
 };
 
 }  // namespace memgraph::coordination
-#endif `
+#endif


as51340:
### Tracking

### Standard development
- [] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
   - Coordinator cluster operations (registering, unregistering, promoting, and demoting instances) now wait for the Raft log to be committed by a majority of coordinators before returning success, and treat follow-up RPCs to data instances as best-effort with automatic
  retry via the reconciliation loop. This means users will no longer see transient RPC failures surfaced as errors for operations that the cluster can self-heal; if a Raft commit fails, the error message now clearly states it is safe to retry. No user action is required
  — the change is fully backward compatible and the cluster will automatically recover from transient network issues during instance registration, unregistration, promotion, and demotion. [#3922](https://github.com/memgraph/memgraph/pull/3922)    

src/coordination/include/coordination/coordinator_ops_status.hpp
```diff
@@ -118,12 +115,12 @@ enum class UpdateConfigStatus : uint8_t {
 enum class DemoteInstanceCoordinatorStatus : uint8_t {
   NO_INSTANCE_WITH_NAME = 0,
   NOT_LEADER,
-  RPC_FAILED,
   RAFT_LOG_ERROR,
   SUCCESS,
   NOT_COORDINATOR,
   LEADER_NOT_FOUND,
   LEADER_FAILED,
+  ALREADY_REPLICA
```
andrejtonev: missing comma at the end

src/coordination/coordinator_instance.cpp
```diff
@@ -873,30 +868,19 @@ auto CoordinatorInstance::UnregisterReplicationInstance(std::string_view instanc
...
     return UnregisterInstanceCoordinatorStatus::RAFT_LOG_ERROR;
   }
 
   auto const name_matches = [instance_name](auto const &instance) { return instance.InstanceName() == instance_name; };
 
-  // The network could be down or the request could fail because of some strange reason. In that case, we try to bring
-  // back old raft state
+  // Even if RPC fails, this is still a success from user's perspective because we will retry this operation in the
+  // reconciliation loop
   if (curr_main->SendRpc<UnregisterReplicaRpc>(instance_name)) {
```
andrejtonev: Bug: dangling connector when unregister RPC fails after Raft commit

When SendRpc<UnregisterReplicaRpc> fails, repl_instances_ still contains the connector for instance_name. The comment says "we will retry in the reconciliation loop", but the retry in InstanceSuccessCallback only fires when main still reports the replica in replicas_num_txns. If the RPC failure was a timeout and main actually processed it (sub-case A: main unregistered the replica), main will stop reporting it — so the retry path in InstanceSuccessCallback is never entered and the connector lingers indefinitely.

The dangling connector keeps firing InstanceSuccessCallback. In that callback, raft_state_->IsCurrentMain(instance_name) returns false (the instance is no longer in Raft), so we enter the else branch. If the instance then reports is_replica = false (e.g., it was demoted elsewhere and hasn't fully transitioned), the code reaches:


// line ~1185
DMG_ASSERT(instance_raft_cache != std::ranges::end(data_instances),
           "Data instance {} not found in Raft cache", instance_name);
This fires because the instance is not in Raft → coordinator crashes.

Even without the crash path: if is_replica = true but UUID is stale, the coordinator will keep calling SendSwapAndUpdateUUID and potentially RegisterReplicaRpc on the unregistered instance — re-registering it with main against the user's intent.

Fix: Always remove from repl_instances_ here regardless of RPC result (the connector should go away once Raft has committed the removal). The reconciliation loop should be responsible for re-registering instances that appear in Raft but are missing a connector — not for cleaning up connectors whose Raft entry has already been deleted.


curr_main->SendRpc<UnregisterReplicaRpc>(instance_name);  // best-effort
std::erase_if(repl_instances_, name_matches);              // always remove
return UnregisterInstanceCoordinatorStatus::SUCCESS;

src/coordination/coordinator_instance.cpp
```diff
@@ -1136,8 +1120,24 @@ void CoordinatorInstance::InstanceSuccessCallback(std::string_view instance_name
       main_num_txns_cache_ = *instance_state.inner_state.main_num_txns;
     }
 
+    // Use this information for checking the number of committed txns and also to figure out if some replica got deleted
     if (instance_state.inner_state.replicas_num_txns.has_value()) {
```
andrejtonev, line 1172: replicas_num_txns_cache_ accumulates stale entries for removed replicas

The old code replaced the entire cache on each tick:


replicas_num_txns_cache_ = *instance_state.inner_state.replicas_num_txns;
The new code only insert_or_assigns entries that are both in Raft and reported by main. An entry is erased only in the else branch — meaning only when the replica is NOT in Raft and main still reports it.

There's a third case that's never cleaned up: a replica removed from Raft that main has already unregistered (sub-case A from UnregisterReplicationInstance). Main won't report it in replicas_num_txns, so neither the insert_or_assign nor the erase branch fires. The entry persists in replicas_num_txns_cache_ until ReconcileClusterState_ or coordinator restart.

Routing impact is nil — CreateRoutingTable iterates raft_log_data_instances (Raft state) for instance membership; the cache is only used as a lag filter, and extra entries in it are harmless.

But it's a behavioral regression from the old code (which kept the cache perfectly in sync) and will grow unboundedly if instances are repeatedly registered/unregistered over the coordinator's lifetime.

Fix: before the loop, prune entries from replicas_num_txns_cache_ that are no longer in data_instances:


std::erase_if(replicas_num_txns_cache_, [&](auto const &kv) {
    return std::ranges::find(data_instances, kv.first,
        [](auto const &c) -> std::string const & { return c.config.instance_name; })
        == std::ranges::end(data_instances);
});

src/coordination/coordinator_instance.cpp
```diff
@@ -809,7 +807,7 @@ auto CoordinatorInstance::RegisterReplicationInstance(DataInstanceConfig const &
   // NOLINTNEXTLINE
   CoordinatorClusterStateDelta const delta_state{.data_instances_ = std::move(data_instances)};
 
-  if (!raft_state_->AppendClusterUpdate(delta_state)) {
+  if (!raft_state_->AppendLogAndWaitForCommit(delta_state)) {
```
andrejtonev: Timeout here leaves the user stuck: "safe to retry" but retry returns NAME_EXISTS

On a 10s timeout, AppendLogAndWaitForCommit returns false and we pop the connector. The interpreter surfaces this as "Writing to Raft log failed — it is safe to retry". But before this call, two RPCs already succeeded:

DemoteMainToReplicaRpc to instance_x — it is now acting as replica
RegisterReplicaOnMainRpc to main — main is now replicating to instance_x
If the Raft entry commits after the timeout (which is the indeterminate case the comment describes), Raft now has instance_x. When the user retries REGISTER INSTANCE, the duplicate check at the top of this function reads from Raft and returns NAME_EXISTS — directly contradicting the "safe to retry" message.

Additionally: the coordinator has no connector for instance_x (popped back, StartStateCheck never called). The cluster is in an inconsistent state — Raft and main both know about the replica, but the coordinator has no health monitoring for it. This self-heals only when ReconcileClusterState_ fires on the next leadership change; there is no automatic recovery path.

Two things to fix:

The "safe to retry" wording is only accurate for operations that are purely idempotent from the user's perspective (like SET INSTANCE TO MAIN which returns MAIN_ALREADY_EXISTS on retry as an informational status). For REGISTER INSTANCE, retry is not safe — consider a more specific message like "Check SHOW INSTANCES to determine if the registration succeeded before retrying".

After a timeout, consider calling TryVerifyOrCorrectClusterState() asynchronously (or exposing it as a hint to the user) so the connector inconsistency is healed without requiring a leadership change.

src/coordination/coordinator_instance.cpp
```diff
@@ -1136,8 +1120,24 @@ void CoordinatorInstance::InstanceSuccessCallback(std::string_view instance_name
...
     }
 
+    // Use this information for checking the number of committed txns and also to figure out if some replica got deleted
     if (instance_state.inner_state.replicas_num_txns.has_value()) {
-      replicas_num_txns_cache_ = *instance_state.inner_state.replicas_num_txns;
+      auto const data_instances = raft_state_->GetDataInstancesContext();
+
+      // Handle the case when data instance was unregistered
+      for (auto const &[replica_name, replica_data] : *instance_state.inner_state.replicas_num_txns) {
+        if (std::ranges::find(data_instances, replica_name, [](auto const &context) {
```
andrejtonev: Projection returns std::string by value — unnecessary heap allocation per element per tick

context.config.instance_name is std::string. Returning it by value from the projection causes std::ranges::find to heap-allocate a copy for every element it examines before finding a match (or reaching end). This runs on every InstanceSuccessCallback tick — once per second by default — for each replica reported by main.

Fix: return const& from the projection:


[](auto const &context) -> std::string const & {
    return context.config.instance_name;
}
std::string const& compares equally against replica_name (std::string const&) without any allocation.

src/coordination/coordinator_instance.cpp
```diff
@@ -1131,13 +1115,50 @@ void CoordinatorInstance::InstanceSuccessCallback(std::string_view instance_name
...
+      auto const data_instances = raft_state_->GetDataInstancesContext();
+
+      // Handle the case when an instance was registered with a log successfully committed to Raft logs but RPC message
+      // for registering replica on main failed
+      for (auto const &raft_replica_instance : data_instances) {
+        if (instance_name != raft_replica_instance.config.instance_name &&
+            !instance_state.inner_state.replicas_num_txns->contains(raft_replica_instance.config.instance_name)) {
+          auto replica_inmem_instance =
+              FindReplicationInstance(raft_replica_instance.config.instance_name, repl_instances_);
+          MG_ASSERT(replica_inmem_instance.has_value(), "Inconsistency found between in-memory state and Raft logs");
```
andrejtonev, line 1183: Do we have to assert here? Can't we prioritize the raft log over the local durability?

andrejtonev: Or just continue?

src/coordination/coordinator_instance.cpp
```diff
@@ -1131,13 +1115,50 @@ void CoordinatorInstance::InstanceSuccessCallback(std::string_view instance_name
...
+    // Use this information for checking the number of committed txns and also to figure out if some replica got
+    // unregistered
     if (instance_state.inner_state.replicas_num_txns.has_value()) {
-      replicas_num_txns_cache_ = *instance_state.inner_state.replicas_num_txns;
+      auto const data_instances = raft_state_->GetDataInstancesContext();
+
+      // Handle the case when an instance was registered with a log successfully committed to Raft logs but RPC message
+      // for registering replica on main failed
+      for (auto const &raft_replica_instance : data_instances) {
+        if (instance_name != raft_replica_instance.config.instance_name &&
```
andrejtonev, line 1178: Missing status filter: this loop should only consider REPLICA-status
instances from Raft.

Right now it iterates ALL data_instances and filters only by name
(instance_name != ...). If Raft ever has an instance with status=MAIN
that is not the current main by UUID — possible if status and
current_main_uuid_ diverge during a transition or future code change —
this loop will call RegisterReplicaOnMainRpc for that MAIN instance,
telling the actual main to register another MAIN as its replica.

The invariant "status=MAIN iff current_main_uuid matches" holds today
but is not enforced here, making this a latent trap for anyone who
touches the delta-building logic later.

One-line fix:
  if (raft_replica_instance.status != ReplicationRole::REPLICA) continue;

Add it as the first statement inside the loop body, before the
replicas_num_txns contains-check.

src/coordination/coordinator_instance.cpp
```diff
@@ -1131,13 +1115,50 @@ void CoordinatorInstance::InstanceSuccessCallback(std::string_view instance_name
...
+                          instance_name);
+          } else {
+            spdlog::error("Failed to register replica {} on the current main {}",
+                          replica_inmem_instance->get().InstanceName(),
+                          instance_name);
+          }
+        }
+      }
+
+      // Handle the case when data instance was unregistered (Raft log committed) but RPC UnregisterReplicaRpc failed
```
andrejtonev, line 1200: Semantics change from old code: the old one-liner
  replicas_num_txns_cache_ = *instance_state.inner_state.replicas_num_txns;
was a full replacement — any replica absent from the main's report had its
cache entry cleared, which is the safe/correct behaviour.

The new incremental update only removes entries for instances absent from
Raft. Registered replicas that are temporarily absent from replicas_num_txns
(down, mid-reconnect, not yet re-registered on main) keep their stale cache
entry indefinitely.

This directly affects routing: CreateRoutingTable's lag_filter treats a
missing cache entry as "allow routing" (see utils.hpp:69, 75), but with a
stale entry it compares the old lag value against max_replica_read_lag. If
the replica was lagging when it went down (stale_lag > max_replica_read_lag),
it is permanently excluded from the READ routing table until the main
reports it again — even after it comes back fully synced.

The simplest fix: at the start of the second loop, build the new cache from
scratch (only entries present in both Raft and replicas_num_txns), then
handle the unregister-retry for Raft-absent entries. This restores the
full-replacement semantics while keeping the new unregister-retry logic.

Alternatively: at the top of the `if (instance_state.inner_state.replicas_num_txns.has_value())` block, clear replicas_num_txns_cache_ entries for
any replica in Raft that is absent from replicas_num_txns (i.e., the
complement of the second loop's condition).

src/coordination/coordinator_instance.cpp
```diff
@@ -623,66 +619,62 @@ auto CoordinatorInstance::SetReplicationInstanceToMain(std::string_view new_main
...
-  auto const not_main_raft = [new_main_name](auto &&instance) {
-    return instance.config.instance_name != new_main_name;
+  auto const is_new_main_connector = [new_main_name](auto const &instance) {
+    return instance.InstanceName() == new_main_name;
   };
-  // replicas already have status replica
-  for (auto &data_instance : data_instances | ranges::views::filter(not_main_raft)) {
+  auto new_main_connector = std::ranges::find_if(repl_instances_, is_new_main_connector);
+  // we already checked if instance exists in Raft log
+  DMG_ASSERT(new_main_connector != std::ranges::end(repl_instances_),
```
andrejtonev: DMG_ASSERT here is reachable in practice and will crash the coordinator.

The guard at L614 only blocks FOLLOWER status; LEADER_NOT_READY passes
through. Two scenarios reach this assert with repl_instances_ not
matching Raft:

1. Race at leadership start: coordinator just won election, status =
   LEADER_NOT_READY, ReconcileClusterState_() is scheduled but hasn't
   taken coord_instance_lock_ yet. User query arrives, acquires the
   lock first (reconcile hasn't taken it), repl_instances_ is still
   empty (fresh leader), but Raft already has the instance → assert fires.

2. Failed reconcile: ReconcileClusterState_() cleared repl_instances_
   (L467) but an exception during rebuild left it partially populated.
   Status remains LEADER_NOT_READY. Next user query hits the assert.

The comment "we already checked if instance exists in Raft log" is the
wrong justification — the Raft check (L623) and the connector check (L638)
are independent invariants. Raft having an entry does not imply a connector
exists in repl_instances_ during LEADER_NOT_READY.

Since the assert fires before any Raft commit, the cluster state is
unchanged — a crash here is pure lost availability with no data risk.
Replace with a graceful return:

  if (new_main_connector == std::ranges::end(repl_instances_)) {
    spdlog::error("Instance {} found in Raft but no connector exists "
                  "(coordinator not ready yet?). Retry the operation.",
                  new_main_name);
    return SetInstanceToMainCoordinatorStatus::NO_INSTANCE_WITH_NAME;
  }

src/coordination/raft_state.cpp
```diff
@@ -506,20 +506,40 @@ auto RaftState::YieldLeadership() const -> void { raft_server_->yield_leadership
...
   spdlog::trace("Request for updating cluster state accepted.");
 
+  // Block until the majority appends log.
+  // Coordinator could get blocked if the log successfully replicated but never committed and we don't want to block
+  // the user query
+  // The problem here is we don't know whether the majority was reached or not so we cannot say to users whether they
+  // should repeat the operation or not
+  constexpr auto kCommitTimeout = std::chrono::seconds{10};
+  auto const start = std::chrono::steady_clock::now();
+  while (!res->has_result()) {
```
andrejtonev: The timeout here is not a clean "operation did not happen" signal — it is
a "we stopped watching" signal. NuRaft continues replicating after we
return false, and if it reaches quorum at t=10s+ε the log commits and the
state machine applies the delta. The caller gets false and skips creating
the connector / updating repl_instances_, but Raft now has the change.

This creates a persistent split that does NOT self-heal while the
coordinator remains leader:
- ReconcileClusterState_ only runs on leadership change.
- InstanceSuccessCallback reconciliation requires a connector to exist
  in repl_instances_ (to get ReplicationClientInfo). Without a connector
  the retry loop either hits the MG_ASSERT (see L1135) or skips the
  instance silently.

Practical user impact on RegisterReplicationInstance timeout:
  User sees RAFT_LOG_ERROR → checks SHOW INSTANCES → instance IS there →
  retries REGISTER INSTANCE → gets INSTANCE_ALREADY_EXISTS → stuck until
  leadership re-election.

The NuRaft API has get_result_with_timeout() which blocks in NuRaft's
internals and returns a definitive result code (RESULT_NOT_EXIST_YET on
timeout) without abandoning the result object mid-flight. Using that
avoids the "still writing into the result after we returned" problem.

If the polling loop is kept, on timeout the coordinator should attempt a
read of the current Raft state to determine whether the log was actually
committed (GetDataInstancesContext + compare) and proceed as success if
the delta is already reflected, rather than unconditionally returning false.

src/coordination/coordinator_instance.cpp
```diff
@@ -780,14 +777,25 @@ auto CoordinatorInstance::RegisterReplicationInstance(DataInstanceConfig const &
...
+
+  // NOLINTNEXTLINE
+  CoordinatorClusterStateDelta const delta_state{.data_instances_ = std::move(data_instances)};
+
+  if (!raft_state_->AppendLogAndWaitForCommit(delta_state)) {
+    spdlog::error("Aborting instance registration. Writing to Raft failed.");
+    return RegisterInstanceCoordinatorStatus::RAFT_LOG_ERROR;
+  }
+
   auto *new_instance =
```
andrejtonev: Minor: emplace_back at L792 is now a "point of no return". The old code
used pop_back to roll back the connector if anything after it failed.
That was correct when connectors came first; now that Raft is first, a
throw from emplace_back (OOM) leaves Raft with the instance and
repl_instances_ without a connector — the same gap as a 10s timeout
(Issue in AppendLogAndWaitForCommit). The coordinator stays in this
split state until a leadership re-election triggers ReconcileClusterState_.

Probability is very low (std::list::emplace_back OOM), but worth a
comment explaining that connector creation failing here has no cleanup
path and will be resolved by the next ReconcileClusterState_ run.

src/coordination/replication_instance_connector.cpp
```diff
@@ -50,7 +50,9 @@ ReplicationInstanceConnector::ReplicationInstanceConnector(
     const std::chrono::seconds instance_health_check_frequency_sec)
     : client_(ReplicationInstanceClient(config.instance_name, config.mgt_server, coord_instance,
                                         instance_health_check_frequency_sec)),
-      timed_failure_detector_(instance_down_timeout_sec) {}
+      timed_failure_detector_(instance_down_timeout_sec),
+      // takes a copy because config is a temporary, stack-allocated object
```
andrejtonev: Comment is misleading on two counts:
  1. config is DataInstanceConfig const& — a reference, not a temporary or
     stack object. It may outlive this constructor or not, depending on
     the caller.
  2. The reason to copy is not about config's lifetime at all. The old code
     stored const ReplicationClientInfo* pointing into a DataInstanceContext
     inside a temporary std::vector returned by GetDataInstancesContext().
     That vector (and the pointer into it) was destroyed at the end of the
     calling scope, leaving a dangling pointer. Storing by value eliminates
     the dangling pointer regardless of what config's lifetime is.

Suggested replacement:
  // Stored by value to avoid a dangling pointer: callers build
  // DataInstanceContext vectors as temporaries, and a raw pointer into
  // those would dangle after the vector is destroyed.

src/coordination/coordinator_instance.cpp
as51340: should be assert because something is very wrong if this happens

src/coordination/coordinator_instance.cpp
as51340: I think this should never happen because there should never be more than one main in Raft logs. I can still add this check as a defensive measure though.

src/coordination/coordinator_instance.cpp
as51340: this cannot happen because StateCheckHandler always returns all replicas, there is no reason to temporarily not send data about replica.

src/coordination/coordinator_instance.cpp
as51340: Not sure if this can actually occur but defensive measure won't hurt

src/coordination/raft_state.cpp
as51340: I will do a blocking call now

src/coordination/coordinator_instance.cpp
andrejtonev: As long as the comment `Minor: emplace_back at L792 is now a "point of no return".` is not fixed, the state could be wrong. 
Raft commits the register-replica log, then emplace_back creates the in-memory connector. If emplace_back throws (OOM) after the Raft commit, the replica is permanently in Raft but never in repl_instances_. The assert at L1145 would then fire every reconciliation cycle for that replica — the coordinator crash-loops with no way to recover short of manual intervention.

If the team's position is "OOM on coordinator = game over anyway", then the assert is acceptable. But that position should be documented explicitly and Issue 7 acknowledged. Right now, the assert fires before reconciliation can fix the inconsistency.

src/coordination/coordinator_instance.cpp
andrejtonev: these checks seem redundant. 
what are we exactly trying to do?

as51340: no need to do reconciliation if there is nothing to reconcile

PR #3923, opened 2026-03-18
feat: Virtual graphs via derive() aggregate

DavIvek:
## What this PR adds

A new `derive(path, options)` aggregate function that builds an in-query
virtual graph as an overlay over the real graph. Procedures called in
that scope iterate the virtual nodes/edges, enabling graph-algorithm
modules to operate on derived relationships without materialising real
edges.

```cypher
MATCH p=(a:Issue)-[*]-(b:Issue)
WHERE id(a) < id(b)
WITH derive(p, {
  virtualEdgeType: 'RELATES_TO',
  relationshipProperties: {weight: 1}
}) AS subgraph
CALL community_detection.get(subgraph) YIELD node, community_id
RETURN node.name, community_id
```

`derive()` is an aggregation, so multi-row MATCH accumulates into one
VirtualGraph. Supports per-row dynamic edge type via `type(r)`,
`undirectedEdgeTypes` (with `'*'` wildcard), and source/target
node label and property overlays.

## Architecture

- `src/query/virtual_{node,edge,graph}.{hpp,cpp}` — new in-memory
  derived-graph structures. Synthetic gids; per-endpoint `shared_ptr`
  on edges so edges that escape their source graph keep endpoints alive.
- `src/query/db_accessor.hpp` — `VirtualGraphDbAccessor` joins
  `DbAccessor` and `SubgraphDbAccessor` as the third arm of the accessor
  family. Wraps the outer `DbAccessor` by value plus a non-owning
  `VirtualGraph *`.
- `src/query/procedure/mg_procedure_impl.{hpp,cpp}` — variant arms:
  - `mgp_graph::impl` is a 3-arm `std::variant`.
  - `mgp_vertex::impl` is 3-arm (`VertexAccessor` / `SubgraphVertexAccessor` / `VirtualNode`).
  - `mgp_edge::impl` is 2-arm (`EdgeAccessor` / `VirtualEdge`).
  - Every `std::visit` enumerates all arms; virtual arms throw
    `ImmutableObjectException` for mutation paths and `std::logic_error`
    for invariant-violation paths.
- Iterators (`mgp_edges_iterator`, `mgp_vertices_iterator`) hold state
  as `std::variant<monostate, RealCursor, VirtualCursor>` — one cursor
  kind, chosen at construction.
- `src/query/plan/operator.cpp` — new `Aggregation::Op::DERIVE` op
  invoking `ProjectPathWithOptions` per row, with per-aggregation
  `real_gid → canonical_synth_gid` dedup map.

## Tests

- `tests/unit/query_virtual_graph.cpp` (new, 8 tests) — data-structure
  layer: storage, dedup, alias-merge, properties.
- `tests/unit/query_procedures_mgp_graph.cpp` (+3 tests) — C-API:
  scope visibility, `IsVirtual()` on real entities, mutation rejection
  across vertex / graph / 6 schema ops.
- `tests/e2e/write_procedures/virtual_graph.py` (new) — full
  match→derive→call pipeline.
- `tests/gql_behave/.../aggregations.feature` — Cypher semantics:
  property/label overlays, `undirectedEdgeTypes` (incl. `'*'` wildcard),
  per-row `type(r)` accumulation, error paths.

## Compat

No public-API breaks. `include/mg_procedure.h` (the user-visible C API)
is unchanged in shape. All variant-related changes are behind
forward-declared types that user-written query modules can only access
through opaque pointers.


DavIvek:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Added a new derive(path, options) aggregate function that constructs an in-query virtual graph overlay, allowing graph algorithms to operate on derived relationships without creating real edges; users can now run procedures on dynamically built subgraphs (with support for custom edge types, undirected patterns, and property overlays) directly within a query, simplifying advanced graph analysis workflows. [#3923](https://github.com/memgraph/memgraph/pull/3923)
- [x] **https://github.com/memgraph/documentation/pull/1664**

src/query/virtual_graph.cpp
```diff
@@ -0,0 +1,115 @@
...
+    in_index_ = std::move(other.in_index_);
+  } else {
+    nodes_ = node_map(other.nodes_, alloc);
+    edges_ = edge_set(other.edges_, alloc);
+    RebuildEdgeIndexes();
+  }
+}
+
+void VirtualGraph::IndexEdge(const VirtualEdge *edge) {
+  out_index_[edge->FromGid()].push_back(edge);
```
colinbarry, line 58: Because `VirtualEdge`s are stored in an `unordered_set`, their addresses are not stable, and so we shouldn't store by pointer. The danger is that an insert on `edges_` could rehash the set, moving the existing edges about in memory, and leaving the raw pointers in `out_index_` and `in_index_` referencing now invalid memory.

src/query/plan/operator.cpp
```diff
@@ -6604,6 +6636,123 @@ class AggregateCursor : public Cursor {
...
+
+    const auto &path_vertices = path_value.ValuePath().vertices();
+    if (path_vertices.empty()) return;
+
+    const auto alloc = projected_graph.get_allocator();
+    auto canonical = [&](const VertexAccessor &real_vertex,
+                         std::string_view labels_key,
+                         std::string_view props_key) -> std::shared_ptr<const VirtualNode> {
+      const auto real_gid = real_vertex.Gid();
+      if (const auto it = dedup.find(real_gid); it != dedup.end()) {
```
colinbarry, line 6748: Not sure if this is an issue, but we deduplicate here by `Gid` only. A vertex could be the first vertex on one path and the last on another. The projected node only keeps either the source overlays or the target overlays, whichever we hit first. 

I guess the alternative is that a vertex which is both source and target would have *both* sets of labels/properties? That would be an easy fix - dedup by both `Gid` and source/target type. Probably a product decision, unless this edge case was specced?

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -3220,7 +3199,22 @@ mgp_error mgp_edge_set_properties(struct mgp_edge *e, struct mgp_map *properties
         },
         properties->items);
 
-    const auto result = e->impl.UpdateProperties(props);
+    auto &ea = std::get<memgraph::query::EdgeAccessor>(e->impl);
+    auto *ctx = e->from.graph->ctx;
+
+#ifdef MG_ENTERPRISE
```
colinbarry, line 3195: Consider moving the auth check upwards, as we currently do the work and then fail, when we could just as easily fail before doing the `std::visit` etc.

src/query/db_accessor.hpp
```diff
@@ -1155,4 +1156,36 @@ class SubgraphDbAccessor final {
...
+
+  const std::string &PropertyToName(storage::PropertyId prop) const;
+
+  const std::string &LabelToName(storage::LabelId label) const;
+
+  const std::string &EdgeTypeToName(storage::EdgeTypeId type) const;
+
+  [[nodiscard]] std::shared_ptr<const VirtualNode> FindNode(storage::Gid synthetic_gid) const;
+
+  [[nodiscard]] VirtualGraph *getGraph() const;
```
colinbarry, line 1184: camelCase naming rather than PascalCase for `getGraph()`

src/query/virtual_node.hpp
```diff
@@ -0,0 +1,113 @@
...
+#include "utils/pmr/string.hpp"
+#include "utils/pmr/unordered_map.hpp"
+#include "utils/pmr/vector.hpp"
+
+namespace memgraph::query {
+
+// Synthetic Gids for VirtualNode and VirtualEdge share a single counter counting down from UINT64_MAX,
+// so node and edge Gids are drawn from the same space and can never collide with each other
+// (nor with real Gids, which count up from 0).
+inline storage::Gid NextSyntheticGid() {
```
colinbarry: Not a big deal, but I wonder if this file (in `virtual_node`) is the right place for this helper, seeing as it is used by both `VirtualNode`s and `VirtualEdge`s.

src/query/interpret/awesome_memgraph_functions.cpp
```diff
@@ -468,8 +469,22 @@ TypedValue Properties(const TypedValue *args, int64_t nargs, const FunctionConte
...
+  } else if (value.IsVirtualNode()) {
+    TypedValue::TMap properties(ctx.memory);
+    for (const auto &[prop_id, prop_value] : value.ValueVirtualNode().Properties()) {
+      properties.emplace(TypedValue::TString(dba->PropertyToName(prop_id), ctx.memory),
+                         TypedValue(prop_value, dba->GetStorageAccessor()->GetNameIdMapper(), ctx.memory));
+    }
+    return TypedValue(std::move(properties));
   } else if (value.IsVertex()) {
     return get_properties(value.ValueVertex());
+  } else if (value.IsVirtualEdge()) {
```
colinbarry: The virtual node block (472) and virtual edge block look identical, same map build from `.Properties()`. Might be worth a small shared helper/lambda (and `Keys`/`Values` has a similar duplicated pattern).

src/glue/communication.cpp
```diff
@@ -115,6 +115,43 @@ storage::Result<communication::bolt::Edge> ToBoltEdge(const query::EdgeAccessor
...
+                                   .element_id = std::move(element_id),
+                                   .from_element_id = std::move(from_element_id),
+                                   .to_element_id = std::move(to_element_id)};
+}
+
+communication::bolt::Vertex ToBoltVertex(const query::VirtualNode &node, const storage::Storage &db) {
+  auto id = communication::bolt::Id::FromUint(node.Gid().AsUint());
+  std::vector<std::string> labels;
+  labels.reserve(node.Labels().size());
+  for (const auto &label : node.Labels()) labels.emplace_back(label);
```
colinbarry, line 145: Consider using `std::ranges::copy` rather than a raw `for` loop

src/query/typed_value.hpp
```diff
@@ -21,6 +21,9 @@
 #include <vector>
 
 #include "query/path.hpp"
+#include "query/virtual_edge.hpp"
```
colinbarry: Seems like `VirtualEdge`, `VirtualNode`, and `VirtualGraph` are only used as `unique_ptr`, so we can get away for forward declaring them rather than introducing these "query/virtual_*.hpp" headers into this header, and then moving the includes to the .cpp.

src/query/procedure/mg_procedure_impl.hpp
```diff
@@ -715,15 +754,37 @@ struct mgp_edge {
            const memgraph::query::SubgraphVertexAccessor &to_v, mgp_graph *graph, allocator_type alloc)
       : alloc(alloc), impl(impl), from(from_v, graph, alloc), to(to_v, graph, alloc) {}
 
+  mgp_edge(const memgraph::query::VirtualEdge &ve, mgp_graph *graph, allocator_type alloc)
+      : alloc(alloc), impl(ve), from(ve.From(), graph, alloc), to(ve.To(), graph, alloc) {}
+
+  mgp_edge(const memgraph::query::VirtualEdge &ve, const memgraph::query::SubgraphVertexAccessor &from_v,
```
colinbarry: Is this correct/useful? It seems this constructs an edge wrapper for a virtual edge with actual vertex endpoints.

src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -3918,6 +3918,9 @@ antlrcpp::Any CypherMainVisitor::visitFunctionInvocation(MemgraphCypher::Functio
       return static_cast<Expression *>(
           storage_->Create<Aggregation>(expressions[0], nullptr, Aggregation::Op::PROJECT_PATH, is_distinct));
     }
+    if (upper_function_name == Aggregation::kDerive) {
+      throw SemanticException("derive() requires exactly 2 arguments: a path and an options map.");
```
colinbarry: `derive()` called with 3 or more arguments isn't caught by this exception, so we get less specific details if we accidentally call it with extra arguments.

src/query/interpret/awesome_memgraph_functions.cpp
```diff
@@ -794,6 +817,20 @@ TypedValue Keys(const TypedValue *args, int64_t nargs, const FunctionContext &ct
   if (value.IsEdge()) {
     return get_keys(value.ValueEdge());
   }
+  if (value.IsVirtualEdge()) {
+    TypedValue::TVector keys(ctx.memory);
+    for (const auto &[prop_id, prop_value] : value.ValueVirtualEdge().Properties()) {
```
colinbarry: Consider using `std::range::transform` ?

src/query/interpret/awesome_memgraph_functions.cpp
```diff
@@ -852,6 +905,13 @@ TypedValue Labels(const TypedValue *args, int64_t nargs, const FunctionContext &
   FType<Or<Null, Vertex>>("labels", args, nargs);
   auto *dba = ctx.db_accessor;
   if (args[0].IsNull()) return TypedValue(ctx.memory);
+  if (args[0].IsVirtualNode()) {
+    TypedValue::TVector labels(ctx.memory);
```
colinbarry, line 897: Consider using `std::ranges::copy`

src/query/interpret/eval.cpp
```diff
@@ -13,10 +13,21 @@
...
+#include <ranges>
 #include <regex>
 
 namespace memgraph::query {
 
+namespace {
+template <std::ranges::input_range R>
+TypedValue RangeToTypedValueList(R &&range, utils::MemoryResource *memory) {
+  utils::pmr::vector<TypedValue> out(memory);
+  if constexpr (std::ranges::sized_range<R>) out.reserve(std::ranges::size(range));
```
colinbarry, line 26: Nice use of `if constexpr` 👍

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -2822,47 +2855,25 @@ mgp_error mgp_vertex_get_property(mgp_vertex *v, const char *name, mgp_memory *m
...
+        auto maybe_prop = v->VisitReal([v, key](const auto &impl) { return impl.GetProperty(v->graph->view, key); });
+        if (!maybe_prop) ThrowOnReadError(maybe_prop.error(), "get a property of", ObjectKind::Vertex);
         return NewRawMgpObject<mgp_value>(memory, std::move(*maybe_prop), GetNameIdMapper(v->graph));
       },
       result);
 }
 
 mgp_error mgp_vertex_iter_properties(mgp_vertex *v, mgp_memory *memory, mgp_properties_iterator **result) {
-  // NOTE: This copies the whole properties into the iterator.
-  // TODO: Think of a good way to avoid the copy which doesn't just rely on some
```
colinbarry, line 2846: Think the TODO should remain as we still have the same behaviour.

colinbarry (commented):
Looks good so far - the only major issue is storing addresses of items in an `unordered_set`, which are not stable. Everything else is style/suggestions. But overall, I approve of this solution. Good job 🚀

src/query/virtual_graph.cpp
DavIvek: Hm, but `edges_` set is `std::unordered_set` which is is a node-based container. As I'm aware of it is stable on inserts, only removals make pointers to elements not valid anymore. (https://en.cppreference.com/cpp/container/unordered_set).

src/query/plan/operator.cpp
DavIvek: This needs documenting, but currently it's following [Neo4j's cypher projection spec](https://neo4j.com/docs/graph-data-science/current/management-ops/graph-creation/graph-project-cypher-projection/#_selection_of_node_properties_and_labels).

src/glue/communication.cpp
DavIvek: I think we can't use `std::ranges::copy` on this code as it currently is because node.Labels() are `utils::pmr`. We could use transform but that doesn't make code any cleaner imo.

src/query/virtual_graph.cpp
colinbarry: Yeah, you are absolutely right, my bad. Resolving this and closing the issue.

colinbarry (approved):
Awesome stuff - this is a fantastic feature, really well implemented. 💪

PR #3925, opened 2026-03-18
fix: Handle TCP segmentation in SLK file transitions

as51340:
## What

Fix a bug in SLK stream processing where TCP segmentation during file transitions caused `CheckStreamStatus` to incorrectly return `NEW_FILE` instead of `PARTIAL`. The fix adds metadata completeness validation
in `CheckStreamStatus()` (`src/slk/streams.cpp`) when a `kFileSegmentMask` is encountered during a file transition (i.e., `remaining_file_size` is set), ensuring the full file metadata (filename string + file
size uint) is present before signaling the transition.

## Why

When the SLK protocol transitions between files in a multi-file transfer, the sender emits `kFileSegmentMask` followed by the new file's metadata (filename + size). TCP can split the segment boundary such that
the mask arrives in one packet but the metadata arrives later. Without this fix, the `OpenFile` path in the Reader would attempt to parse incomplete metadata and fail with "Size data missing in SLK stream!".

## How

In `CheckStreamStatus`, when `kFileSegmentMask` is encountered and `remaining_file_size` is set (indicating a file-to-file transition, not the first file), the code now performs a two-stage check before
returning `NEW_FILE`:

1. **String prefix check** — verifies at least 9 bytes (1 marker + 8 length) are available after the mask.
2. **Full metadata check** — reads the string length, then verifies the entire metadata block (string prefix + string data + uint marker + uint value) fits in the remaining buffer.

If either check fails, `PARTIAL` is returned with the appropriate `stream_size` hint so the caller knows how much more data to wait for. The check is gated on `remaining_file_size` being set, so first-file
detection is unaffected.

## Testing

Added 5 unit tests in `tests/unit/slk_streams.cpp`:

- **`FileTransitionMaskOnly`** — mask present but zero metadata bytes; expects `PARTIAL`
- **`FileTransitionPartialStringPrefix`** — fewer than 9 bytes after mask; expects `PARTIAL`
- **`FileTransitionPartialStringData`** — string prefix present but string body truncated; expects `PARTIAL`
- **`FileTransitionSufficientMetadata`** — complete metadata present; expects `NEW_FILE` (positive case)
- **`FirstFileMaskNotAffectedByCheck`** — no `remaining_file_size`; verifies the new guard doesn't regress first-file detection

## Notes for reviewers

- The `stream_size` values returned in the `PARTIAL` branches are best-effort hints (`pos + kSegmentMaxTotalSize` when the string length is unknown, `pos + needed` when it is). Verify these are sufficient for
the caller's read-more logic.
- The guard `str_len <= kSegmentMaxDataSize` protects against corrupted length fields — if the length is impossibly large, the code falls through to the existing `NEW_FILE` return, which will let the Reader
surface a more specific error during actual parsing

### DIFF
--- a/tests/unit/slk_streams.cpp
+++ b/tests/unit/slk_streams.cpp
@@ -646,18 +646,16 @@ TEST(CheckStreamStatus, FirstFileMaskNotAffectedByCheck) {
   // The metadata check should NOT apply when remaining_file_size is not set (first file case).
   // This verifies backward compatibility with WholeFileInSegment-style buffers.
   std::vector<uint8_t> buffer;
-  memgraph::slk::Builder builder(
-      [&buffer](const uint8_t *data, size_t size, bool have_more) -> memgraph::slk::BuilderWriteFunction::result_type {
-        for (size_t i = 0; i < size; ++i) {
-          buffer.push_back(data[i]);
-        }
-        return {};
-      });
+  memgraph::slk::Builder builder([&buffer](const uint8_t *data, size_t size, bool have_more) {
+    for (size_t i = 0; i < size; ++i) {
+      buffer.push_back(data[i]);
+    }
+  });
 
   auto const input = GetRandomData(5);
   builder.PrepareForFileSending();
-  ASSERT_TRUE(builder.SaveFileBuffer(input.data(), input.size()).has_value());
-  ASSERT_TRUE(builder.Finalize().has_value());
+  builder.SaveFileBuffer(input.data(), input.size());
+  builder.Finalize();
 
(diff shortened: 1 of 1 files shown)


as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Fixed a bug where large file transfers between Memgraph instances (such as snapshots during replication) could fail with a "Size data missing in SLK stream!" error when TCP split the data at an unfortunate
  boundary between consecutive files. This was most likely to occur under high network load or with large snapshot files. No action is required — upgrading to this version resolves the issue automatically. [#3925](https://github.com/memgraph/memgraph/pull/3925)

PR #3929, opened 2026-03-18
Apply single-store pattern to VectorEdgeIndex

DavIvek:
Refactor VectorEdgeIndex to store VectorIndexId in edge properties instead of raw vector lists, matching the vertex VectorIndex pattern from PR #3291. The actual vector data lives only in the uSearch index, avoiding data duplication.

Key changes:
- Restructure Impl to use index_by_id_ map keyed by NameIdMapper ID
- Convert edge properties to VectorIndexId on SetProperty and CreateIndex
- Wire IndexedPropertyDecoder<Edge> into EdgeAccessor for transparent decoding
- Add AbortProcessor with removed_vectors_cache for correct tx abort
- Serialize edge index entries in snapshots (with backward-compat reading)
- Add VectorEdgeIndexRecovery for WAL replay with property tracking
- Update durability (snapshot/WAL/recovery) for new format


DavIvek:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Refactors the vector edge index to align with the node vector index storage model, eliminating duplication by storing vector data only once in the index and keeping lightweight references on edges. This improves memory efficiency and consistency, with no required changes for users. [#3929](https://github.com/memgraph/memgraph/pull/3929)
- [x] **https://github.com/memgraph/documentation/pull/1608**

src/storage/v2/indices/vector_edge_index.cpp
```diff
@@ -238,131 +313,288 @@ std::vector<VectorEdgeIndexInfo> VectorEdgeIndex::ListVectorIndicesInfo() const
...
+    if (edge->deleted()) return false;
+    auto ep_it = pimpl->edge_endpoints_.find(edge);
+    if (ep_it == pimpl->edge_endpoints_.end()) return false;
+    return !ep_it->second.first->deleted() && !ep_it->second.second->deleted();
+  });
   for (std::size_t i = 0; i < result_keys.size(); ++i) {
-    const auto &entry = static_cast<EdgeIndexEntry>(result_keys[i].member.key);
+    auto *edge = static_cast<Edge *>(result_keys[i].member.key);
+    auto [from_vertex, to_vertex] = pimpl->edge_endpoints_.at(edge);
     result.emplace_back(
```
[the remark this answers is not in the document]
DavIvek: Not a concern for now; we don’t expect to have many indices, so it shouldn’t become a bottleneck.

src/storage/v2/edge_accessor.cpp
```diff
@@ -35,6 +37,19 @@ namespace rv = r::views;
...
+std::optional<PropertyValue> TryConvertToVectorEdgeIndexProperty(Storage *storage, Edge * /*edge*/,
+                                                                 EdgeTypeId edge_type, PropertyId property,
+                                                                 const PropertyValue &value) {
+  if (!value.IsAnyList() || value.IsVectorIndexId()) return std::nullopt;
+  if (storage->indices_.vector_edge_index_.Empty()) return std::nullopt;
+  auto indices_by_et = storage->indices_.vector_edge_index_.GetIndicesByEdgeType(edge_type);
+  auto it = indices_by_et.find(property);
+  if (it == indices_by_et.end()) return std::nullopt;
+  return PropertyValue(PropertyValue::VectorIndexIdData{.ids = utils::small_vector<uint64_t>{it->second},
+                                                        .vector = ListToVector(value)});
```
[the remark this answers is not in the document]
DavIvek: Not a concern for now; we don’t expect to have many indices, so it shouldn’t become a bottleneck.

src/storage/v2/durability/snapshot.cpp
andrejtonev, line 10988: we should rebase to the latest master and then handle the new logic in v34 and keep the separate v33 version

src/storage/v2/indices/vector_edge_index.cpp
```diff
@@ -39,189 +43,275 @@ struct synchronized_mg_vector_edge_index_t {
...
+        } else {
+          edge->properties.SetProperty(spec.property, vector_property);
+        }
+      }
+    }
+  }
+  // Collect edges from the dropped index before erasing
+  std::vector<Edge *> dropped_edges;
+  {
+    auto guard = utils::SharedResourceLockGuard(mg_index.mutex, utils::SharedResourceLockGuard::READ_ONLY);
```
andrejtonev: why unlock and lock again, instead of doing all under the one lock?

src/storage/v2/edge_accessor.cpp
```diff
@@ -35,6 +37,19 @@ namespace rv = r::views;
 namespace memgraph::storage {
 
 namespace {
+
+std::optional<PropertyValue> TryConvertToVectorEdgeIndexProperty(Storage *storage, Edge * /*edge*/,
+                                                                 EdgeTypeId edge_type, PropertyId property,
+                                                                 const PropertyValue &value) {
+  if (!value.IsAnyList() || value.IsVectorIndexId()) return std::nullopt;
+  if (storage->indices_.vector_edge_index_.Empty()) return std::nullopt;
+  auto indices_by_et = storage->indices_.vector_edge_index_.GetIndicesByEdgeType(edge_type);
```
andrejtonev: TryConvertToVectorEdgeIndexProperty / GetIndicesByEdgeType allocates and returns a std::unordered_map by value on every SetProperty, InitProperties, and UpdateProperties call. Only a single lookup (find(property)) is performed on the result and the map is immediately discarded.

The vertex analog avoids this by calling GetVectorIndexIdsForVertex which takes both the label and property as filters and returns a small_vector — no intermediate map. The edge version should do the same: a direct lookup by (edge_type, property) pair through index_by_id_ would avoid the heap allocation entirely.

src/storage/v2/durability/snapshot.cpp
```diff
@@ -9856,14 +9949,38 @@ RecoveredSnapshot LoadCurrentVersionSnapshot(Decoder &snapshot, std::filesystem:
   return {.snapshot_info = info, .recovery_info = recovery_info, .indices_constraints = std::move(indices_constraints)};
 }
 
+RecoveredSnapshot LoadSnapshotVersion33(Decoder &snapshot, std::filesystem::path const &path,
```
andrejtonev: I get that current load snapshot has some logic to make it compatible with v33 and v34, but we have always just copy-pasted the old implementation and that way are sure the code is as it was before.

andrejtonev: For example, you read entry_count in the vector edge index unconditionally. I don't think that works if the snapshot is v33

tests/unit/vector_edge_index.cpp
```diff
@@ -163,8 +163,7 @@ TEST_F(VectorEdgeIndexTest, UpdatePropertyValueTest) {
     ASSERT_NO_ERROR(acc->PrepareForCommitPhase(memgraph::tests::MakeMainCommitArgs()));
     const auto search_result = acc->VectorIndexSearchOnEdges(test_index.data(), 1, std::vector<float>{2.0, 2.0});
     EXPECT_EQ(search_result.size(), 1);
-    EXPECT_EQ(std::get<0>(search_result[0]).GetProperty(acc->NameToProperty(test_property), View::OLD).value(),
-              updated_value);
+    EXPECT_EQ(std::get<0>(search_result[0]).Gid(), edge_gid);
```
andrejtonev, line 166: Verify IndexedPropertyDecoder<Edge> correctly decodes VectorIndexId -> float list

src/storage/v2/indices/vector_edge_index.cpp
```diff
@@ -239,131 +324,298 @@ std::vector<VectorEdgeIndexInfo> VectorEdgeIndex::ListVectorIndicesInfo() const
...
   VectorSearchEdgeResults result;
   result.reserve(result_set_size);
 
   auto guard = utils::SharedResourceLockGuard(mg_index.mutex, utils::SharedResourceLockGuard::READ_ONLY);
-  const auto result_keys =
-      mg_index.index.filtered_search(query_vector.data(), result_set_size, [](const EdgeIndexEntry &entry) {
-        auto guard = std::shared_lock{entry.edge->lock};
-        return !entry.from_vertex->deleted() && !entry.to_vertex->deleted() && !entry.edge->deleted();
-      });
+  auto ep_lock = std::shared_lock{pimpl->edge_endpoints_mutex_};
```
andrejtonev, line 365: Lock ordering deadlock: SearchEdges acquires mg_index.mutex (read) at line 364, then edge->lock (shared) inside the filtered_search callback. SetProperty acquires edge->lock (unique) first, then mg_index.mutex (write) via UpdateVectorIndex. Inverted order → deadlock.

src/storage/v2/indices/vector_edge_index.cpp
```diff
@@ -239,131 +324,298 @@ std::vector<VectorEdgeIndexInfo> VectorEdgeIndex::ListVectorIndicesInfo() const
...
+  }
+}
+
+// VectorEdgeIndexRecovery implementation
+void VectorEdgeIndexRecovery::UpdateOnIndexDrop(std::string_view index_name, NameIdMapper *name_id_mapper,
+                                                std::vector<VectorEdgeIndexRecoveryInfo> &recovery_info_vec,
+                                                utils::SkipList<Vertex>::Accessor &vertices) {
+  for (auto &recovery_info : recovery_info_vec) {
+    if (recovery_info.spec.index_name == index_name) {
+      auto index_id = name_id_mapper->NameToId(index_name);
```
andrejtonev: use NameToIdIfExists as to not spam the id map

src/storage/v2/indexed_property_decoder.hpp
```diff
@@ -26,8 +28,13 @@ struct IndexedPropertyDecoder {
   void DecodeProperty(PropertyValue &value) const {
     switch (value.type()) {
       case PropertyValueType::VectorIndexId: {
-        value.ValueVectorIndexList() = indices->vector_index_.GetVectorPropertyFromIndex(
-            entity, name_id_mapper->IdToName(value.ValueVectorIndexIds()[0]), name_id_mapper);
+        if constexpr (std::is_same_v<T, Vertex>) {
+          value.ValueVectorIndexList() = indices->vector_index_.GetVectorPropertyFromIndex(
+              entity, name_id_mapper->IdToName(value.ValueVectorIndexIds()[0]), name_id_mapper);
```
andrejtonev, line 34: are we sure there is at least 1 element?

DavIvek: I'll add dmg_assert since if this happens we have corrupted property store which

src/storage/v2/indices/vector_index_utils.hpp
andrejtonev: The function both mutates the property value AND returns a decision. The name ShouldUnregisterFromIndex suggests a pure predicate ("should we?") but it actually performs the unregistration itself.

Caller pattern (both sites are identical):

auto vector_property = edge->properties.GetProperty(spec.property);
if (ShouldUnregisterFromIndex(vector_property, index_id)) {
    // ids now empty → restore raw vector
    edge->properties.SetProperty(spec.property, PropertyValue(raw_vector));
} else {
    // ids still remain → write back mutated VectorIndexId (index_id removed)
    edge->properties.SetProperty(spec.property, vector_property);
}
The else branch is the subtle part: it unconditionally calls SetProperty with the already-mutated vector_property. This is correct because ShouldUnregisterFromIndex already erased index_id from the list in step 2. But if a caller ever skips the else (assuming it's a pure check), they'd silently leave the stale index_id in the property.

The real concern:
The !IsVectorIndexId() → return true early-exit path is odd. If property_value is not a VectorIndexId (e.g. it's already a raw list, or null), the function returns true (restore) without touching it. At both call sites this triggers SetProperty with the result of index.get(edge, ...) — which reads from the uSearch index to reconstruct the raw float vector. That path is correct if the uSearch index still contains the edge. In DropIndex, mg_index.index.contains(edge) is checked before ShouldUnregisterFromIndex (line 226), so it's safe. In UpdateOnIndexDrop (recovery), same guard at line 575 (recovery_info.index_entries.find). So the early-exit doesn't cause a bug today, but it's fragile logic.

Summary of the issue:
The name is misleading — it should be UnregisterFromIndex (imperative, signals mutation). The return true early-exit for non-VectorIndexId properties also hides an implicit assumption that the caller has already validated the edge is in the uSearch index before calling. A comment or rename would make this contract explicit.

andrejtonev (approved):

PR #3949, opened 2026-03-23
feat: Move timeout and health check frequency flags to run-time coordinator settings

as51340:
## What

Moves `instance_down_timeout_sec` and `instance_health_check_frequency_sec` from per-instance startup flags on `CoordinatorInstance` to Raft-replicated coordinator settings in `CoordinatorClusterState`. Both settings are now queryable via `SHOW COORDINATOR SETTINGS`
and modifiable at run time via `SET COORDINATOR SETTING`. The old command-line flags are preserved as hidden/deprecated — they emit a warning on use and their values are ignored.

## Why

These two settings were the only coordinator-level knobs that still required a restart to change. Making them run-time settings brings them in line with the other coordinator settings (`enabled_reads_on_main`, `sync_failover_only`, `max_failover_replica_lag`,
`max_replica_read_lag`, `deltas_batch_progress_size`) and allows operators to tune health-check behaviour on a live cluster without downtime.

## How

The implementation follows the exact same pattern used for existing Raft-replicated coordinator settings:

- **`CoordinatorClusterState`**: Added `instance_down_timeout_sec_` (`uint32_t`, default 5) and `instance_health_check_frequency_sec_` (`uint32_t`, default 1) as member fields. Updated all special member functions (copy/move constructors, copy/move assignment,
`operator==`), `DoAction`, `to_json`/`from_json` (with `j.value()` for backward-compatible deserialization), and getters/setters.
- **`CoordinatorClusterStateDelta`**: Added corresponding `std::optional` fields so Raft log entries can carry updates to these settings.
- **`CoordinatorStateMachine`**: Added `SerializeUpdateClusterState`/`DecodeLog` entries and getter pass-throughs to `CoordinatorClusterState`.
- **`RaftState`**: Added getter pass-throughs to `CoordinatorStateMachine`.
- **`CoordinatorInstance`**: Removed the two member fields. `SetCoordinatorSetting` and `ShowCoordinatorSettings` now include the new settings. All call sites that previously read from instance fields (`ReconcileClusterState_`, `RegisterReplicationInstance`,
`GetCoordinatorsInstanceStatus`, `GetTelemetryJson`) now read from `raft_state_->Get...()`.
- **`CoordinatorInstanceInitConfig`**: Removed the two `std::chrono::seconds` fields; updated construction sites in `memgraph.cpp` and `tests/integration/telemetry/client.cpp`.
- **`src/flags/coordination.cpp`**: Replaced `DEFINE_uint32` with `DEFINE_VALIDATED_HIDDEN_uint32` (new macro in `flag_validation.hpp`) that logs a deprecation warning pointing users to the `SET COORDINATOR SETTING` query.

## Testing

- **Unit — `coordinator_cluster_state`**: `Marshalling` test extended to assert default values for the two new settings, include them in the delta, and verify serialization round-trip. `RoutingPoliciesSwitch` test extended to assert correct defaults when deserializing
legacy JSON without the new fields.
- **Unit — `coordinator_raft_log_serialization`**: Two new tests added — `SerializeUpdateClusterStateWithInstanceSettings` (delta with only the two new fields) and `SerializeUpdateClusterStateAllSettings` (delta with all fields populated) — verifying
`SerializeUpdateClusterState`/`DecodeLog` round-trip.
- **E2E — `distributed_coords.py`**: `test_coord_settings` extended to check default values, SET new values, and verify distribution across all three coordinators for both settings.

## Notes for reviewers

- **Upgrade migration**: Values previously set via `--instance_down_timeout_sec` / `--instance_health_check_frequency_sec` flags are **not** automatically migrated into Raft state. After upgrading, the settings will have their defaults (5s / 1s). Users who customized
these flags must run `SET COORDINATOR SETTING` queries manually post-upgrade.
- **No durability version bump**: The new fields are added to `CoordinatorClusterState` JSON with `j.value(key, default)` in `from_json`, so older snapshots/logs deserialize safely with defaults.
- The `DEFINE_VALIDATED_HIDDEN_uint32` macro is new in `flag_validation.hpp`, mirroring the existing `DEFINE_VALIDATED_HIDDEN_int32` pattern.

### DIFF
--- a/tests/unit/coordinator_cluster_state.cpp
+++ b/tests/unit/coordinator_cluster_state.cpp
@@ -199,7 +199,7 @@ TEST_F(CoordinatorClusterStateTest, Marshalling) {
   // test default settings
   ASSERT_TRUE(cluster_state.GetSyncFailoverOnly());
   ASSERT_EQ(cluster_state.GetDeltasBatchProgressSize(), 100'000);
-  ASSERT_EQ(cluster_state.GetInstanceDownTimeoutSec(), std::chrono::seconds{5});
+  ASSERT_EQ(cluster_state.GetInstanceDownTimeoutSec(), 5);
   ASSERT_EQ(cluster_state.GetInstanceHealthCheckFrequencySec(), std::chrono::seconds{1});
 
   // NOLINTNEXTLINE
@@ -225,7 +225,7 @@ TEST_F(CoordinatorClusterStateTest, Marshalling) {
   ASSERT_EQ(cluster_state.GetMaxFailoverReplicaLag(), 25);
   ASSERT_EQ(cluster_state.GetMaxReplicaReadLag(), 10);
   ASSERT_EQ(cluster_state.GetDeltasBatchProgressSize(), 50'000);
-  ASSERT_EQ(cluster_state.GetInstanceDownTimeoutSec(), std::chrono::seconds{10});
+  ASSERT_EQ(cluster_state.GetInstanceDownTimeoutSec(), 10);
   ASSERT_EQ(cluster_state.GetInstanceHealthCheckFrequencySec(), std::chrono::seconds{3});
 }
 
@@ -269,6 +269,6 @@ TEST_F(CoordinatorClusterStateTest, RoutingPoliciesSwitch) {
   ASSERT_EQ(deserialized_cluster_state.GetMaxFailoverReplicaLag(), std::numeric_limits<uint64_t>::max());
   ASSERT_EQ(deserialized_cluster_state.GetMaxReplicaReadLag(), std::numeric_limits<uint64_t>::max());
   // by default read 5 and 1
-  ASSERT_EQ(deserialized_cluster_state.GetInstanceDownTimeoutSec(), std::chrono::seconds{5});
+  ASSERT_EQ(deserialized_cluster_state.GetInstanceDownTimeoutSec(), 5);
   ASSERT_EQ(deserialized_cluster_state.GetInstanceHealthCheckFrequencySec(), std::chrono::seconds{1});
 }


as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - The instance_down_timeout_sec and instance_health_check_frequency_sec settings are now managed as run-time coordinator settings replicated through Raft, rather than startup flags. Users can view and change these values on a live cluster using SHOW COORDINATOR SETTINGS
  and SET COORDINATOR SETTING 'instance_down_timeout_sec' TO '<value>' (and likewise for instance_health_check_frequency_sec), with changes automatically distributed to all coordinators. The old command-line flags still exist but are hidden and ignored with a deprecation
   warning guiding users to the new queries. When upgrading from a previous version, the values previously set via flags will not be automatically migrated into the Raft state — the new settings will start with their defaults (5 seconds for instance_down_timeout_sec and
  1 second for instance_health_check_frequency_sec), so users who had customized these flags must manually run the corresponding SET COORDINATOR SETTING queries after the upgrade to restore their desired values. [#3949](https://github.com/memgraph/memgraph/pull/3949)

src/coordination/coordinator_instance.cpp
andrejtonev, line 1132: The Raft write succeeds and GetInstanceDownTimeoutSec() / GetInstanceHealthCheckFrequencySec() will immediately return the new values, but the existing ReplicationInstanceConnector objects in repl_instances_ are unaffected. Their timeouts were captured at emplace_back time — TimedFailureDetector::instance_down_timeout_sec_ and ReplicationInstanceClient::instance_health_check_frequency_sec_ are set once at construction and never updated. The scheduler interval (instance_checker_.SetInterval(...)) was called once in StartStateCheck() and does not re-read the field dynamically.

Practical effect: after SET COORDINATOR SETTING 'instance_down_timeout_sec' TO X, currently-running health checkers still use the old timeout until the next ReconcileClusterState_() call (i.e., the next leader election or restart). The claim that this allows "tuning on a live cluster without downtime" is misleading for already-registered instances.

Possible fixes:

After writing to Raft, iterate repl_instances_ on the leader and call StopStateCheck() / restart with new values.
Have TimedFailureDetector and ReplicationInstanceClient hold a const RaftState& and read values dynamically.
At minimum, document this in the PR notes and in the SET COORDINATOR SETTING query docs.

src/coordination/coordinator_state_machine.cpp
```diff
@@ -228,6 +230,16 @@ auto CoordinatorStateMachine::DecodeLog(buffer &data) -> CoordinatorClusterState
...
 
+    if (json.contains(kInstanceDownTimeoutSec.data())) {
+      auto const instance_down_timeout_sec = json.value(kInstanceDownTimeoutSec.data(), uint32_t{5});
+      delta_state.instance_down_timeout_sec_ = instance_down_timeout_sec;
+    }
+
+    if (json.contains(kInstanceHealthCheckFreqSec.data())) {
+      auto const instance_health_check_frequency_sec = json.value(kInstanceHealthCheckFreqSec.data(), uint32_t{1});
+      delta_state.instance_health_check_frequency_sec_ = instance_health_check_frequency_sec;
+    }
```
andrejtonev, line 233: The fallback values in json.value(key, uint32_t{5}) and json.value(key, uint32_t{1}) are unreachable — we already checked json.contains(key) on the line above. Replace with json[kInstanceDownTimeoutSec.data()].get<uint32_t>() to express that the key is guaranteed present at this point.
Or rewrite to avoid the double lookup

src/coordination/coordinator_state_machine.cpp
```diff
@@ -228,6 +230,16 @@ auto CoordinatorStateMachine::DecodeLog(buffer &data) -> CoordinatorClusterState
       delta_state.deltas_batch_progress_size_ = deltas_batch_progress_size;
     }
 
+    if (json.contains(kInstanceDownTimeoutSec.data())) {
+      auto const instance_down_timeout_sec = json.value(kInstanceDownTimeoutSec.data(), uint32_t{5});
+      delta_state.instance_down_timeout_sec_ = instance_down_timeout_sec;
+    }
+
+    if (json.contains(kInstanceHealthCheckFreqSec.data())) {
+      auto const instance_health_check_frequency_sec = json.value(kInstanceHealthCheckFreqSec.data(), uint32_t{1});
```
andrejtonev: 5 and 1 are magic numbers, store as named constants

PR #3950, opened 2026-03-23
feat: Eliminate ORDER BY when index scan provides ascending order

imilinovic:
## Summary

Eliminate redundant `OrderBy` operators when an index scan already provides the required ascending order, avoiding an unnecessary O(n log n) sort. This is especially impactful with `LIMIT` — instead of sorting all rows and taking the first N, the index provides rows in order and the query stops after N rows.

### What's supported
- **Simple cases**: `ORDER BY n.prop` with index on `(prop)` and a range/equality filter
- **Composite indexes**: `ORDER BY n.a, n.b` with index `(a, b)`, including equality-pinned column skipping (`WHERE a = 5 ORDER BY b`)
- **Variable renames**: `WITH n AS m ORDER BY m.prop` — tracked through Produce operators
- **Alias resolution**: `WITH n.prop AS a ORDER BY a` and `RETURN n.prop AS a ORDER BY a` — aliases resolved through Produce chains
- **Mixed expressions**: `ORDER BY a, n.b` where some columns are aliases and some are direct PropertyLookups
- **Alias rename chains**: `WITH n.prop AS a WITH a AS b ORDER BY b` — multi-level alias resolution
- **Dual-scope**: `RETURN n AS m ORDER BY n.prop` and `ORDER BY m.prop` both work
- **DISTINCT**: `RETURN DISTINCT n.prop AS a ORDER BY a` — Distinct preserved, OrderBy eliminated
- **WITH ORDER BY RETURN**: `WITH n ORDER BY n.prop RETURN n.prop AS p` — parent Produce from outer clause included in alias resolution
- **Multi-scan sequential plans**: `MATCH (b:B), (a:A) ORDER BY b.id, a.id` — multiple nested scans matched in outermost-first order
- **Edge property index scans**: `ORDER BY e.prop` with edge type+property or global edge property indexes — works for both range scans (`ScanAllByEdgeTypePropertyRange`, `ScanAllByEdgePropertyRange`) and exact-value scans (`ScanAllByEdgeTypePropertyValue`, `ScanAllByEdgePropertyValue`)
- **Mutation safety**: `SET n.x = 1 ORDER BY n.y` — SetProperty/RemoveProperty preserve order when the mutated property doesn't overlap ORDER BY columns; SetLabels/RemoveLabels always preserve order

### What correctly bails out
- DESC ordering (index is ASC only)
- IN-list filters (Unwind rewrite breaks global order)
- Non-order-preserving operators between scan and OrderBy (Aggregate, CallProcedure, etc.)
- Partial index coverage (ORDER BY has more columns than the index covers)
- Aliases resolving to different symbols or computed expressions
- Wrong nesting order in multi-scan plans (inner before outer)
- Edge property ORDER BY on a different property than the index provides
- `USING PARALLEL EXECUTION` — parallel scans split work across threads, so index order is not preserved

### Implementation
- Shared `OrderByEliminator<TDbAccessor>` helper (`order_by_elimination.hpp`) used by both `IndexLookupRewriter` (vertex scans) and `EdgeIndexRewriter` (edge scans)
- Stack-based approach with `PreVisit`/`PostVisit(OrderBy)` hooks
- Walks `prev_ops_` (visitor path) from scan upward to OrderBy — avoids relying on `input()` which is only valid for `HasSingleInput()` operators
- Two-phase alias resolution in `ResolveDownward`: step 1 tracks scan symbol renames through Produce (e.g. `n AS m`), step 2 resolves ORDER BY aliases to property paths (e.g. `n.prop AS a`)
- Edge scans recorded via `TryRecordEdgeScan` which looks through one level of Filter wrapper (for global edge property indexes with edge-type filtering)
- Secondary branches (Cartesian, RollUpApply, Optional) are isolated via `RewriteBranch`
- `OrderBy` cost estimation uses `IncrementCost` for consistency with other operators

### Testing
- 61 unit tests (`tests/unit/query_plan_orderby_index.cpp`)
- 35 e2e tests (`tests/e2e/query_planning/query_planning_orderby_index.py`) — plan verification and correctness checks with real data

### DIFF
--- a/src/query/plan/planner.hpp
+++ b/src/query/plan/planner.hpp
@@ -71,12 +71,15 @@ class PostProcessor final {
     auto &symbol_table = context->symbol_table;
     auto &db = context->db;
 
+    const bool parallel_exec = context->query->pre_query_directives_.parallel_execution_;
+
     return std::move(plan) | [&](auto p) { return RewriteEnumAccess(std::move(p), symbol_table, ast, db); } |
            [&](auto p) {
-             return RewriteWithIndexLookup(std::move(p), symbol_table, ast, db, index_hints_, parameters_);
+             return RewriteWithIndexLookup(
+                 std::move(p), symbol_table, ast, db, index_hints_, parameters_, parallel_exec);
            } |
            [&](auto p) { return RewriteWithJoinRewriter(std::move(p), symbol_table, ast, db); } |
-           [&](auto p) { return RewriteWithEdgeIndexRewriter(std::move(p), symbol_table, ast, db); } |
+           [&](auto p) { return RewriteWithEdgeIndexRewriter(std::move(p), symbol_table, ast, db, parallel_exec); } |
            [&](auto p) { return RewritePeriodicDelete(std::move(p), symbol_table, ast, db); }
 #ifdef MG_ENTERPRISE
            |
--- a/src/query/plan/rewrite/edge_index_lookup.hpp
+++ b/src/query/plan/rewrite/edge_index_lookup.hpp
@@ -38,8 +38,12 @@ namespace impl {
 template <class TDbAccessor>
 class EdgeIndexRewriter final : public HierarchicalLogicalOperatorVisitor {
  public:
-  EdgeIndexRewriter(SymbolTable *symbol_table, AstStorage *ast_storage, TDbAccessor *db)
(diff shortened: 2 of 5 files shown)


imilinovic:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - ORDER BY on indexed properties is now automatically optimized away to skip sorting when the index already provides the right order, improving query performance. [#3950](https://github.com/memgraph/memgraph/pull/3950)

PR #3960, opened 2026-03-25
Fix INT64_MIN literal parsing (issue #471)

### LINKED ISSUE #471: [BUG] Min int value throws error
**Memgraph version**
2.3.1

**Environment**
Windows, Docker, Bolt, PHP Bolt library

**Describe the bug**
When I run query with min int value I get error:
Integer literal exceeds 64 bits. (Memgraph.ClientError.MemgraphError.MemgraphError)

Min int value on x64 is -9223372036854775808

**To Reproduce**
1. Run query `RETURN $i` over bolt protocol with min int value as parameter.

**Expected behavior**
I expect to get min int value instead of error.

queelius:
## Summary

- Fixes the bug where Memgraph throws "Integer literal exceeds 64 bits" for the valid value `-9223372036854775808` (INT64_MIN)
- The root cause: the Cypher parser tokenizes `-9223372036854775808` as unary minus + the literal `9223372036854775808`, but `9223372036854775808` overflows `int64_t` before negation can be applied
- Fix detects this edge case at three levels: `ParseIntegerLiteral` (accepts a `negated` flag), query stripping (merges the preceding `-` token), and AST construction (produces `PrimitiveLiteral(INT64_MIN)` directly in `visitExpression4`)

## Changes

- `src/query/frontend/parsing.hpp` / `parsing.cpp`: Add `bool negated` parameter to `ParseIntegerLiteral`; when set and parsing overflows, retry with prepended `-`
- `src/query/frontend/stripped.cpp`: When an INT token overflows and the prior token is `-`, merge them and store INT64_MIN
- `src/query/frontend/ast/cypher_main_visitor.cpp`: In `visitExpression4` (unary +/- handler), detect the INT64_MIN pattern and produce the literal directly without going through `UnaryMinusOperator`

## Test plan

- [ ] New unit tests in `tests/unit/stripped.cpp`: `NegativeInteger`, `Int64Min`, `Int64MinInExpression`, `Int64Max`, `IntegerOverflow`
- [ ] New unit tests in `tests/unit/cypher_main_visitor.cpp`: `Int64MinLiteral`, `Int64MinInComparison`
- [ ] Existing tests should continue to pass (no behavioral change for values within int64 range)
- [ ] Manual test: `RETURN $i = -9223372036854775808` should no longer throw

Fixes #471

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/query/frontend/ast/cypher_main_visitor.cpp
+++ b/src/query/frontend/ast/cypher_main_visitor.cpp
@@ -11,8 +11,10 @@
 
 #include <algorithm>
 #include <any>
+#include <cstdint>
 #include <cstring>
 #include <iterator>
+#include <limits>
 #include <string>
 #include <tuple>
 #include <type_traits>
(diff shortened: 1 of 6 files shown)


queelius:
Closing this. I missed the CLA requirement before submitting. The fix approach is sound if someone with a signed CLA wants to pick it up (the branch and tests are all here).

PR #3981, opened 2026-03-27
feat: Vector index on combination of properties and labels

### LINKED ISSUE #3042: Add the ability to define vector index on any combination of labels and props (0 or more)
This was already mentioned, but adding it on GH as well, so we have it noted as community member requested it.

DavIvek:
Closes #3042

## What

Adds three new shapes to vector index DDL on top of the existing `:Label(prop)` form, for both vertex (`CREATE VECTOR INDEX`) and edge (`CREATE VECTOR EDGE INDEX`):

| Syntax | Mode | Semantics |
|---|---|---|
| `:L(p)` | `SINGLE` (existing) | exactly one label / edge type |
| `:L1\|L2(p)` | `ANY_OF` | matches if entity has any of the listed ids |
| `:L1&L2(p)` | `ALL_OF` | vertex has every listed label (rejected for edges — an edge has exactly one type) |
| `(p)` | `WILDCARD` | every entity that has the property, no label/edge-type filter |

`SHOW INDEX INFO` displays the filter as `*` for wildcard, `:L`, `:L1\|L2`, `:L1&L2` for the others.

## Where it lives

- **Grammar** — new `vectorIndexLabels` rule in `MemgraphCypher.g4` with three alternatives (wildcard `(p)`, OR-list, AND-list).
- **AST / visitor** — `VectorIndexQuery` and `CreateVectorEdgeIndexQuery` carry `(mode, vector<LabelIx/EdgeTypeIx>)`. The visitor derives mode from grammar shape (presence of `&` ⇒ AND, multiple labels ⇒ OR, none ⇒ wildcard, single ⇒ SINGLE) and **rejects** AND on edges with a `SemanticException`.
- **Storage** — `VectorMembershipFilter<IdT>` template (`storage::VectorMatchMode`, `vector<IdT>`) shared between vertex (`VectorLabelFilter`) and edge (`VectorEdgeTypeFilter`) sides. `Matches`, `IsAffectedBy`, `Format` all live there.
- **Edge index** — `EdgeIndexEntry` carries `EdgeTypeId` and the per-edge map `edge_endpoints_` widened to `EdgeEndpoints{from, to, edge_type}` so `VectorIndexSearchOnEdges` no longer scans `from_vertex->out_edges` to recover the edge type per result.
- **Hot paths** — `UpdateOnAddLabel` / `UpdateOnRemoveLabel` / `UpdateOnSetProperty` etc. call `filter.Matches(vertex->labels)` instead of `==`. Multi-index overlap on a property is supported: `GetIndicesByLabel` returns every match and the iteration covers all of them. `UpdateOnRemoveLabel` keeps a vertex in an ANY_OF index when removing one of multiple matching labels still leaves it as a member.
- **Auth** — `SHOW SCHEMA INFO` requires `HasUnrestrictedAccessToVertices()` / `Edges()` to expose a wildcard index, and READ on **every** listed label/edge type for non-wildcard indexes (so a `:A&B(emb)` row doesn't leak to a user who has READ only on `:A`).
- **AbortProcessor** — both vertex and edge sides now have a `wildcard_properties` set so abort correctly undoes property changes for wildcard indices.

## Durability

- Bumped WAL/snapshot version to **`v35` (`kVectorIndexMultiLabel`)**. `v34` (the released `kDescriptionAndDescIndexSupport`) keeps the legacy single-label byte layout; only `v35+` carries the new packed `(mode + count + ids…)` layout.
- WAL: `WalVectorIndexCreate` / `WalVectorEdgeIndexCreate` use `VersionDependantUpgradable<kVectorIndexMultiLabel, std::string, VectorFilterInfo, UpgradeForVectorMultiLabel>`. v < 35 reads a single string and the upgrader promotes it to `{SINGLE, {name}}`; v ≥ 35 reads the packed struct directly.
- Snapshot: `LoadCurrentVersionSnapshot` takes `version` and gates the new layout on `version >= kVectorIndexMultiLabel`. Encoder writes the packed shape; v34 reader still gets a single id.

## How indexes are tracked on entities

Vector index ids are stored alongside the property as `VectorIndexIdData{ids[], vector}`. With overlapping indexes (e.g. wildcard plus specific) **all** matching index ids are stored — `GetIndicesByLabel` and `GetIndexIdsForEdgeTypeProperty` return every match, so updates and deletes touch every overlapping index.

## What changed structurally 

- Renamed `VectorLabelMode` → `VectorMatchMode` and dropped the per-side `VectorEdgeTypeMode` alias; query layer keeps its own AST-level `query::VectorLabelMode` linked to storage by `static_assert`.
- `ApproximateVerticesVectorCount` / `ApproximateEdgesVectorCount` now take an `index_name` (works for every mode — old `(label, prop)` key only fit `SINGLE`).


DavIvek:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Extended vector index to support multi-label and wildcard filters (`ANY_OF`, `ALL_OF`, and `(property)` forms) for both vertex and edge indexes, enabling more flexible indexing across multiple labels or all entities with a given property; users can now define broader or more precise index scopes without duplicating indexes, with automatic compatibility for existing indexes and no manual migration required. [#3981](https://github.com/memgraph/memgraph/pull/3981)
- [x] **https://github.com/memgraph/documentation/pull/1665**

src/storage/v2/durability/snapshot.cpp
DavIvek, line 1: TODO: update once we merge https://github.com/memgraph/memgraph/pull/4106

src/storage/v2/durability/version.hpp
DavIvek, line 1: TODO: Update once we merge https://github.com/memgraph/memgraph/pull/4106

src/query/frontend/ast/cypher_main_visitor.hpp
```diff
@@ -41,6 +41,12 @@ struct ParsingContext {
   bool is_query_cached = false;
 };
 
+struct VectorIndexLabelsInfo {
+  VectorLabelMode mode;
+  std::vector<std::string> names;
```
imilinovic: I'd maybe template this so you can use EdgeTypeIx/LabelIx instead of pure string

src/query/frontend/ast/ast.hpp
```diff
@@ -2386,6 +2387,20 @@ class CreateTextEdgeIndexQuery : public memgraph::query::Query {
 
 using ConfigMap = std::unordered_map<Expression *, Expression *>;
 
+enum class VectorLabelMode : uint8_t {
+  SINGLE = 0,
+  WILDCARD = 1,
+  ANY_OF = 2,
+  ALL_OF = 3,
+};
```
imilinovic: just use storage version, then you dont need casting inbetween later, same convention was used already

src/query/frontend/ast/ast.hpp
```diff
@@ -36,6 +36,7 @@
 #include "storage/v2/constraints/type_constraints.hpp"
 #include "storage/v2/description_store.hpp"
 #include "storage/v2/indices/index_order.hpp"
+#include "storage/v2/indices/vector_index.hpp"
```
imilinovic: can we avoid this? I think this is pretty heavy include... I think if you move VectorLabelMode to a singular .hpp this can be avoided (comment below mentions this)

src/storage/v2/indices/vector_edge_index.cpp
```diff
@@ -297,17 +300,15 @@ void VectorEdgeIndex::UpdateOnSetProperty(Vertex *from_vertex, Vertex *to_vertex
...
-    // If value is null, we have to remove the edge from all indices that contain it (by edge type).
-    auto indices = GetIndicesByProperty(property);
-    for (const auto &[et, idx_id] : indices) {
-      if (et != edge_type) continue;
+  } else {
+    const auto indices = GetIndicesByProperty(property);
+    for (const auto &[idx_id, filter] : indices) {
+      if (!filter->Matches(edge_type)) continue;
       RemoveEdgeFromIndex(edge, idx_id);
     }
```
imilinovic, line 310: if edge is removed you should also remove the entry from edge_endpoints_

DavIvek: nice catch

imilinovic (approved):

PR #3207, opened 2025-08-25
Add more accurate way to get number of logical CPU cores

antaljanosbenjamin:
Add a function to detect number of logical CPU cores reading from `/proc/cpuinfo` when `std::thread::hardware_concurrency` returns 0.

PR #3982, opened 2026-03-27
Fix: std::hardware_concurrency can return 0 on some machines 

### LINKED ISSUE #2891: std::hardware_concurrency can return 0 on some machines
I read that std::hardware_concurrency can return 0 on some machines and that it makes sense to have "back-up default". This is example from the c++ concurrency in action:
```
auto max_threads = user flag thread count
auto hardware_threads = std::thread::hardware_concurrency();
auto num_threads = std::min(hardware_threads != 0 ? hardware_threads : 2, max_threads);
```

Maybe would be good that we change this at all places in the code and provide a utility function for obtaining the number of threads since std::thread::hardware_concurrency is only a hint https://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency

_Originally posted by @as51340 in https://github.com/memgraph/memgraph/issues/2818#issuecomment-2796408673_

DavIvek:
Fixes #2891

**Root cause:**
`rpc/server.hpp` and `communication/server.hpp` used `std::thread::hardware_concurrency()` as a default parameter with no fallback. On machines where this returns 0, servers could be created with 0 workers.

**Fix:**
Added `GetSafeHardwareConcurrency(unsigned fallback = 2)` utility in `utils/system_info.hpp` that wraps `hardware_concurrency()` with a guaranteed non-zero fallback. Replaced all 5 raw callsites:
- `src/rpc/server.hpp` — unprotected default parameter
- `src/communication/server.hpp` — unprotected default parameter
- `src/flags/bolt.cpp` — had `std::max(..., 1U)`, now uses utility
- `src/flags/general.cpp` — snapshot + recovery thread counts
- `src/utils/priority_thread_pool.cpp` — wakeup thread calculation

**Test:**
Added unit test `tests/unit/safe_hardware_concurrency.cpp` — verifies `GetSafeHardwareConcurrency()` always returns non-zero with both default and custom fallbacks.

### DIFF
--- a/src/utils/system_info.cpp
+++ b/src/utils/system_info.cpp
@@ -221,19 +221,11 @@ RuntimeEnv DetectRuntimeEnv() {
   return RuntimeEnv::NO_KUBERNETES;
 }
 
-namespace {
-unsigned LogicalCPUCoresFromProcCpuinfo() {
-  auto lines = utils::ReadLines("/proc/cpuinfo");
-  return static_cast<unsigned>(
-      std::ranges::count_if(lines, [](const auto &line) { return line.starts_with("processor"); }));
-}
-}  // namespace
-
 unsigned GetSafeHardwareConcurrency(unsigned fallback) {
   auto hw = std::thread::hardware_concurrency();
   if (hw != 0) return hw;
 
-  hw = LogicalCPUCoresFromProcCpuinfo();
+  hw = static_cast<unsigned>(GetCPUInfo("").cpu_count);
   return hw != 0 ? hw : std::max(fallback, 1U);
 }
 


DavIvek:
### Tracking
- [x] **Fixes #2891**

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels
- [x] Select the appropriate CI test labels

### Documentation checklist
    - Fixes an issue where certain servers could start with zero worker threads on systems reporting no hardware concurrency, improving overall reliability. [#3982](https://github.com/memgraph/memgraph/pull/3982)

gitbuda:
I think this https://github.com/memgraph/memgraph/pull/3207 PR tried to do the same? The approach here is simpler? Is the complexity of the closed PR worth it? CC @mattkjames7 because you were commenting on the closed PR 😃

DavIvek:
> I think this #3207 PR tried to do the same? The approach here is simpler? Is the complexity of the closed PR worth it? CC @mattkjames7 because you were commenting on the closed PR 😃

That approach was more correct, let's go with that (I can't reopen a PR)...

src/utils/system_info.cpp
```diff
@@ -219,4 +221,20 @@ RuntimeEnv DetectRuntimeEnv() {
   return RuntimeEnv::NO_KUBERNETES;
 }
 
+namespace {
+unsigned LogicalCPUCoresFromProcCpuinfo() {
+  auto lines = utils::ReadLines("/proc/cpuinfo");
```
mattkjames7: This logic might be duplicated to some degree here https://github.com/memgraph/memgraph/pull/3982/changes#diff-56e32e9f5c8d4386ec1d39d6de32d7c577e1798de89cdb7207e563754c59d891L147
 in `CPUInfo GetCPUInfo(const std::string &machine)`

PR #3985, opened 2026-03-30
feat: Instance-specific replication exception handling

as51340:
## What

Replaces the three separate replication error types (`SyncReplicationError`, `StrictSyncReplicationError`, `TimeoutReplicationError`) with a single unified `ReplicationError` struct that carries per-replica failure details (replica name, replication mode, failure
reason). Error messages now identify exactly which replicas failed and why, instead of generic "at least one SYNC/STRICT_SYNC replica has not confirmed" messages. Also adds `RpcFailedToConnectException` to distinguish connection failures from other RPC errors, and
changes `StartTransactionReplication` to return `std::expected` instead of `std::optional` to preserve error context.

## Why

The previous error types were opaque — they indicated that *some* replica failed but not *which* one or *why*. This made debugging replication issues in production clusters difficult, especially when multiple replicas with different failure modes were involved. The new
 structured errors give operators actionable information (e.g., "Failed to replicate to SYNC replica 'instance_1': replica is not reachable or not in sync with the main").

## How

- **`storage_error.hpp/cpp`**: Defines `ReplicaFailure` (name + mode + reason enum), `ReplicationError` (vector of failures + whether main committed), and `FormatReplicationError()` which groups failures by mode/reason for human-readable messages. The three old error
structs are removed from `StorageManipulationError`.
- **`replication_client.cpp`**: `StartTransactionReplication` returns `std::expected<ReplicaStream, StartTxnReplicationError>` instead of `std::optional<ReplicaStream>`. Each failure path (RECOVERY, REPLICATING, MAYBE_BEHIND, DIVERGED, connection failure, RPC failure)
returns a typed error variant.
- **`replication_transaction.cpp/hpp`**: `TransactionReplication` stores start-txn errors in a parallel `errors_` vector. `CollectStartTxnErrors()` filters out ASYNC replica errors (fire-and-forget semantics) and maps internal error types to `ReplicaFailure`.
`ShipDeltas` collects per-replica failures instead of returning a single `ClientCommunicationError`.
- **`inmemory/storage.cpp`**: `PrepareForCommitPhase` uses the new error types. For the 2PC path, start-txn failures are checked after `FinalizeTransaction()` and properly call `AbortAndResetCommitTs()` to avoid GC timestamp leaks. The `PeriodicCommit` fatal-error
check uses `transaction_committed` to decide whether the error is fatal.
- **`rpc/exceptions.hpp`**: Adds `RpcFailedToConnectException` (subclass of `RpcFailedException`). Removes `kRpcTimeoutMsg` constant — the timeout guidance is now part of `FormatReplicationError`.
- **`interpreter.cpp`**: Commit path simplified — replication errors are thrown immediately via `FormatReplicationError` instead of accumulating boolean flags and throwing after commit.
- **`operator.cpp`**: `HandlePeriodicCommitError` uses `transaction_committed` to decide whether to throw or warn.

## Testing

- **New**: `tests/unit/storage_v2_error.cpp` — 14 unit tests covering `FormatReplicationError`: single/multiple replicas, all failure reasons, grouping by mode/reason, plural forms, timeout suffix, committed vs aborted messages.
- **New**: `tests/e2e/high_availability/distributed_coords.py::test_main_cannot_connect` — verifies error messages when one or two SYNC replicas are killed.
- **Modified**: All E2E (HA single/distributed coordinator, strict_sync), Jepsen (bank, create, mtenancy, replication), and stress test assertions updated to match new error message format (`"Failed to replicate to SYNC replica"` instead of `"At least one SYNC replica
has not confirmed"`).
- **Modified**: Jepsen `utils.clj` — `sync-replica-down?` and `strict-sync-replica-down?` merged into `is-replica-down?` matching all three replication modes.

## Notes for reviewers

- The 2PC path in `PrepareForCommitPhase` checks `CollectStartTxnErrors()` *after* `FinalizeTransaction()`. This is safe because if any replica failed to start the transaction, `repl_prepare_phase_status` cannot have a value (the replica couldn't have voted), so
`FinalizeCommitPhase` is never called on that path — only `FinalizeTransaction(false, ...)` (abort) runs before the check.
- `FormatReplicationError` uses string-prefix matching (`starts_with("replica is")` → `"replicas are"`) for pluralization — somewhat fragile if reason strings are refactored.
- No durability version bump needed — these changes are in-memory error handling only, no persistent state changes.

### DIFF
--- a/tests/e2e/high_availability/distributed_coords.py
+++ b/tests/e2e/high_availability/distributed_coords.py
@@ -304,7 +304,10 @@ def test_main_cannot_connect(test_name):
     try:
         execute_and_fetch_all(main_cursor, "create ()")
     except Exception as e:
-        assert "Failed to replicate to SYNC replicas 'instance_1' and 'instance_2'" in str(e)
+        # Depending on exact replication timing, there could be a different reason for a failure for instance_1 and instance_2, that's why we check for a failure in this way
+        assert "Failed to replicate to SYNC replica" in str(e)
+        assert "instance_1" in str(e)
+        assert "instance_2" in str(e)
 
 
 def test_global_edge_index_drop_replication(test_name):


as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Replication error messages now identify exactly which replicas failed and why, replacing the previous generic "at least one SYNC/STRICT_SYNC replica has not confirmed" messages with detailed per-replica information such as the replica name, its replication mode, and
  the specific failure reason (e.g., not reachable, RPC timeout, diverged). Users debugging replication issues in clusters will see actionable error messages that pinpoint the problematic replicas, making it easier to diagnose and resolve failures without needing to
  inspect logs on each instance individually. No configuration changes or user action is required — the improved messages appear automatically in query error responses and logs. [#3985](https://github.com/memgraph/memgraph/pull/3985)
- [x] **[ Documentation PR link ](https://github.com/memgraph/documentation/pull/1581)**

src/storage/v2/inmemory/storage.cpp
```diff
@@ -1058,51 +1058,64 @@ std::expected<void, StorageManipulationError> InMemoryStorage::InMemoryAccessor:
...
+  auto res = commit_args.apply_if_main([&](DatabaseProtector const &protector)
+                                           -> std::expected<void, StorageManipulationError> {
+    // From this point on, only main executes this
+    // If there are no STRICT_SYNC replicas for the current txn
+    if (!replicating_txn.ShouldRunTwoPC()) {
+      // WAL file is already finalized
+      FinalizeCommitPhase(durability_commit_timestamp);
+
+      auto start_failures = replicating_txn.CollectStartTxnErrors();
+      if (!start_failures.empty()) {
```
andrejtonev: Bug: missed replica failures when start-phase and ship-phase failures coexist

In PrepareForCommitPhase (non-2PC path ~line 1069 and 2PC abort path ~line 1100) a replica failure can be silently dropped from the reported error.

Scenario: Two SYNC replicas A and B. A fails at StartTransactionReplication (ReplicaNotInSyncErr → errors_[A] set, stream is nullopt). B starts successfully but fails mid-stream in FinalizeTransactionReplication (e.g. network drop → TIMEOUT_ERROR).

This produces:

CollectStartTxnErrors() → {A(NOT_IN_SYNC)}
repl_prepare_phase_status.error().failures → {A(RPC_ERROR), B(TIMEOUT)} — A appears here too but with a degraded reason (null stream falls into the GENERIC_ERROR branch), B appears with its real reason
Current code (non-2PC path):


auto start_failures = replicating_txn.CollectStartTxnErrors();
if (!start_failures.empty()) {
    // Returns {A(NOT_IN_SYNC)} — B is never reported
    return std::unexpected{ReplicationError{.failures = std::move(start_failures), .transaction_committed = true}};
}
if (!repl_prepare_phase_status.has_value()) {
    // Would return {A, B} but never reached when start_failures non-empty
    ...
}
The early return preserves the precise reason for A (good), but drops B entirely (bad).

Impact: observability only. B is marked MAYBE_BEHIND and recovered by heartbeat regardless. transaction_committed is still correct. The operator sees an incomplete error — potentially missing B's TIMEOUT advisory with the deltas_batch_progress_size hint entirely.

Suggested fix — merge both sets, letting start_failures override the reason for replicas that appear in both:


auto start_failures = replicating_txn.CollectStartTxnErrors();
if (!repl_prepare_phase_status.has_value()) {
    auto merged = repl_prepare_phase_status.error().failures;
    // Prefer the precise start-error reason for overlapping replicas
    for (auto const &sf : start_failures) {
        if (auto it = std::ranges::find(merged, sf.name, &ReplicaFailure::name); it != merged.end())
            it->reason = sf.reason;
    }
    return std::unexpected{ReplicationError{.failures = std::move(merged), .transaction_committed = true}};
}
if (!start_failures.empty()) {
    return std::unexpected{ReplicationError{.failures = std::move(start_failures), .transaction_committed = true}};
}
return {};
Apply the same merge in the 2PC abort path (~line 1100). In 2PC the scenario is less likely (a STRICT_SYNC start failure will typically cause FinalizePrepareCommitPhase to fail for the same replica, making the sets overlap), but the logic is identical.

src/storage/v2/replication/replication_transaction.cpp
andrejtonev, line 65: If `StartTransactionReplication()` already failed for this replica, `streams[i]` is `nullopt` and `errors_[i]` already carries the start-txn failure. We still call `FinalizePrepareCommitPhase()` / `FinalizeTransactionReplication()` here, which turns that same replica into an additional generic ship-delta failure. Later `PrepareForCommitPhase()` concatenates `CollectStartTxnErrors()` with `repl_prepare_phase_status.error().failures`, so the same replica can be reported twice with conflicting reasons. Should we skip `ShipDeltas()` entirely for replicas whose start phase already failed?

src/query/interpreter.cpp
```diff
@@ -9542,17 +9528,10 @@ void Interpreter::Commit() {
...
+        [&execution_db_accessor = current_db_.execution_db_accessor_]<typename T>(const T &arg) {
           using ErrorType = std::remove_cvref_t<T>;
-          if constexpr (std::is_same_v<ErrorType, storage::SyncReplicationError>) {
-            commit_confirmed_by_all_sync_replicas = false;
-          } else if constexpr (std::is_same_v<ErrorType, storage::StrictSyncReplicationError>) {
-            commit_confirmed_by_all_strict_sync_replicas = false;
-          } else if constexpr (std::is_same_v<ErrorType, storage::TimeoutReplicationError>) {
-            rpc_timeout = true;
+          if constexpr (std::is_same_v<ErrorType, storage::ReplicationError>) {
+            throw ReplicationException(storage::FormatReplicationError(arg));
```
andrejtonev: This changes behavior for committed-on-main replication failures. `PrepareForCommitPhase()` can now return `ReplicationError{transaction_committed=true}` after `FinalizeCommitPhase()` has already committed locally (see the non-2PC path in storage), but we throw here before reaching the after-commit trigger scheduling block below. On the old path, SYNC/timeout replication errors were deferred until after triggers were enqueued. Is it intentional that a locally committed transaction now skips after-commit triggers whenever replication returns a non-fatal error?

PR #3952, opened 2026-03-23
Database specific memory tracking

andrejtonev:
Adds per-database memory tracking and a **Tenant Profiles** feature that lets operators cap memory usage per database. Every allocation inside a database — storage, vector indices, query execution — is attributed to a per-DB tracker hierarchy. That hierarchy feeds into both the global domain trackers (license enforcement) and a per-DB cap tracker (tenant enforcement).

---

### Tracker Hierarchy

Each `Database` owns four `MemoryTracker` instances. Declaration order in `database.hpp` is destruction-order-sensitive — C++ destructs members in reverse declaration order, so `db_arena_` must be declared last. Its destructor purges the arena and fires extent hooks that dereference tracker pointers; those trackers must still be alive at that point.

**Global license enforcement** (`total_memory_tracker` — no parent)
```
total_memory_tracker
├── graph_memory_tracker              (storage + query domain; AI Platform license limit)
│   ├── db_memory_tracker_            (skip lists, delta slabs, indices)  ─┐
│   │                                                                       ├─ also child of db_total_memory_tracker_
│   └── [query bytes via arena hooks, not a tracker child — see below]     │
│                                                                           │
└── vector_index_memory_tracker       (vector index domain; AI Platform license limit)
    └── db_embedding_memory_tracker_  (usearch mmap allocations)          ─┘
```

**Per-DB tenant cap** (`db_total_memory_tracker_` — no parent)
```
db_total_memory_tracker_
├── db_memory_tracker_            (also child of graph_memory_tracker above)
├── db_embedding_memory_tracker_  (also child of vector_index_memory_tracker above)
└── db_query_memory_tracker_      (PMR allocations inside query execution)
      ↳ only one parent (db_total_memory_tracker_); no graph_memory_tracker parent.
        Query bytes reach graph_memory_tracker via the default arena's extent hooks.
        A second tracker parent would double-count them.
```

Storage and embedding allocations are counted in exactly three trackers — per-DB domain tracker → global domain tracker → per-DB cap. Query allocations are counted in two — `db_query_memory_tracker_` → `db_total_memory_tracker_`, with domain aggregation happening via arena extent hooks rather than a tracker parent. Each tracker represents a distinct logical scope; no byte is counted twice. No lock overhead beyond the tracker atomics.

---

#### `DbArena` — jemalloc extent hooks

[...]

PR #3999, opened 2026-04-02
Add AI Platform license type

DavIvek:
Added new `AI_PLATFORM` license type that enforces memory limits only on graph memory, allowing vector index memory to grow unconstrained up to the system `--memory-limit`.


DavIvek:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Added a new `AI_PLATFORM` license type that enforces memory limits only on graph memory while allowing vector index memory to grow up to the system `--memory-limit`, enabling more flexible scaling for vector workloads; users can switch to this license type if they need to prioritize vector index capacity without being constrained by graph memory limits. [#3999](https://github.com/memgraph/memgraph/pull/3999)
- [x] **https://github.com/memgraph/documentation/pull/1605**

src/license/license.cpp
```diff
@@ -83,10 +83,13 @@ LicenseCheckResult IsValidLicenseInternal(const License &license, std::string_vi
 std::string LicenseTypeToString(const LicenseType license_type) {
   switch (license_type) {
     case LicenseType::ENTERPRISE: {
-      return "enterprise";
+      return "Enterprise";
```
gitbuda: no no, this has to stay as-is because then the previous license keys will stop working...

so the exact license types should be (to be backward compatible, without a need to reissue the license keys):
* `enterprise`
* `oem`
* `ai-platform`

gitbuda (changes requested):

src/license/license.cpp
DavIvek: The license type is stored as a uint8_t in the SLK serialized key, not as a string. This is used only for logging.

DavIvek: But I’ll revert this due to telemetry backward compatibility.

src/license/license.cpp
```diff
@@ -125,18 +128,34 @@ LicenseChecker::~LicenseChecker() { Finalize(); }
...
+      utils::total_memory_tracker.SetHardLimit(0);
+      utils::graph_memory_tracker.SetHardLimit(memory_limit);
+    } else {
+      // ENTERPRISE / no license: limit total memory, clear graph limit
+      utils::graph_memory_tracker.ResetLimit();
       utils::total_memory_tracker.SetHardLimit(memory_limit);
-      *locked = memory_limit;
     }
+
+    *locked = current;
```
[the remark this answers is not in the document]
DavIvek: This is the same pattern that existed before this changes. Only tests are reseting limits when doing a cleanup.

src/license/license.cpp
```diff
@@ -125,18 +128,34 @@ LicenseChecker::~LicenseChecker() { Finalize(); }
...
+  const auto set_memory_limit = [](int64_t memory_limit, std::optional<LicenseType> license_type = std::nullopt) {
     auto locked = previous_memory_limit.Lock();
-    if (!*locked || **locked != memory_limit) {
+    PreviousMemoryState current{.limit = memory_limit, .type = license_type};
+    if (*locked && **locked == current) return;
+
+    if (license_type == LicenseType::AI_PLATFORM && memory_limit > 0) {
+      // AI_PLATFORM: limit graph memory only; total falls back to --memory-limit
+      utils::total_memory_tracker.SetHardLimit(0);
+      utils::graph_memory_tracker.SetHardLimit(memory_limit);
```
andrejtonev: make it clear that set 0 is actually unlimitted. you can also create a named const so its clear that way

src/license/license.cpp
```diff
@@ -88,6 +88,9 @@ std::string LicenseTypeToString(const LicenseType license_type) {
     case LicenseType::OEM: {
       return "oem";
     }
+    case LicenseType::AI_PLATFORM: {
```
andrejtonev, line 97: just putting this here, but should be on src/query/interpreter.cpp (line 6772):

For `AI_PLATFORM`, this field becomes misleading.

`RevalidateLicense()` moves the license cap onto `graph_memory_tracker` and intentionally leaves `total_memory_tracker` at the startup `--memory-limit` (`SetHardLimit(0)` falls back to `maximum_hard_limit_`). As a result, `SHOW STORAGE INFO` reports `allocation_limit` as the system cap even though graph allocations are enforced by the lower license cap.

That means an operator can see `allocation_limit=10GiB` here while graph writes start failing at `2GiB`. We should either expose the effective graph cap separately for `AI_PLATFORM` or make this field reflect the active enforcement model.

DavIvek: Nice catch. I’d rather adjust allocation_limit, since with the new changes for the storage info query (https://github.com/memgraph/memgraph/pull/3952) we’re adding a lot of new fields to the output. I’d prefer to limit that a bit, but I’m not completely sure.

src/query/interpreter.cpp
```diff
@@ -6766,11 +6766,14 @@ PreparedQuery PrepareSystemInfoQuery(ParsedQuery parsed_query, bool in_explicit_
...
-            {TypedValue("disk_usage"), TypedValue(utils::GetReadableSize(static_cast<double>(info.disk_usage)))},
-            {TypedValue("memory_tracked"),
+            {TypedValue("global_disk_usage"), TypedValue(utils::GetReadableSize(static_cast<double>(info.disk_usage)))},
+            {TypedValue("global_memory_tracked"),
              TypedValue(utils::GetReadableSize(static_cast<double>(utils::total_memory_tracker.Amount())))},
-            {TypedValue("allocation_limit"),
-             TypedValue(utils::GetReadableSize(static_cast<double>(utils::total_memory_tracker.HardLimit())))},
+            {TypedValue("global_runtime_allocation_limit"),
+             TypedValue(utils::GetReadableSize(static_cast<double>(utils::total_memory_tracker.MaximumHardLimit())))},
+            {TypedValue("global_license_allocation_limit"),
```
andrejtonev: memory_limit == 0 means "unlimited" for ENTERPRISE, but GetReadableSize(0.0) returns "0 B". A user querying SHOW STORAGE INFO on an ENTERPRISE instance will see global_license_allocation_limit: "0 B" which looks like a bug or zero-byte limit rather than unconstrained.

Suggest: mirror the pattern already used a few lines down in SHOW LICENSE INFO (lines 6811–6812):


const auto lic_limit = license::global_license_checker.GetDetailedLicenseInfo().memory_limit;
TypedValue(lic_limit != 0 ? utils::GetReadableSize(static_cast<double>(lic_limit)) : "unlimited")

src/license/license.cpp
```diff
@@ -125,18 +128,37 @@ LicenseChecker::~LicenseChecker() { Finalize(); }
...
+
+    *locked = current;
   };
 
   if (enterprise_enabled_) [[unlikely]] {
     is_valid_.store(true, std::memory_order_release);
-    set_memory_limit(0);
+    set_memory_limit(0, license_type_);
     return;
   }
```
andrejtonev, line 170: GetDetailedLicenseInfo() reads previous_license_info_ (line ~357), but the enterprise_enabled_ fast-path never populates it. In tests using EnableTesting(LicenseType::AI_PLATFORM), SHOW STORAGE INFO will return a stale or zero-initialized global_license_allocation_limit — not the value routed by set_memory_limit.

Fix: populate previous_license_info_ in the enterprise_enabled_ branch, or in EnableTesting directly:


void LicenseChecker::EnableTesting(const LicenseType license_type) {
    enterprise_enabled_ = true;
    license_type_ = license_type;
    {
        auto locked = previous_license_info_.Lock();
        locked->emplace("test-key", "test-org");
        (*locked)->is_valid = true;
        (*locked)->license.type = license_type;
        // memory_limit left 0 — matches what set_memory_limit(0, license_type) will do
    }
    is_valid_.store(true, std::memory_order_release);
}
Without this, AiPlatform_TestingFlag and similar unit tests exercise memory routing but leave GetDetailedLicenseInfo() broken for the same path.

src/license/license.cpp
```diff
@@ -125,18 +128,37 @@ LicenseChecker::~LicenseChecker() { Finalize(); }
 void LicenseChecker::RevalidateLicense(utils::Settings &settings) {
   spdlog::trace("License revalidation started");
 
-  static utils::Synchronized<std::optional<int64_t>, utils::SpinLock> previous_memory_limit;
-  const auto set_memory_limit = [](const int64_t memory_limit) {
+  // Passing 0 to SetHardLimit restores the limit to maximum_hard_limit_ (the --memory-limit flag value).
```
andrejtonev: The comment is only true when --memory-limit was configured (i.e. maximum_hard_limit_ > 0). Tracing SetHardLimit(0) through memory_tracker.cpp:


maximum_hard_limit_ == 0  →  next_limit = 0
next_limit <= 0           →  spdlog::warn("Invalid memory limit.") + return (no-op)
So when --memory-limit is not set, passing kUseDefaultMemoryLimit silently warns and does nothing — it does not restore anything.

Suggested fix:


// Passing 0 to SetHardLimit restores the limit to maximum_hard_limit_ (the --memory-limit flag value)
// if --memory-limit was configured. If it was not, this is a no-op (logs a warning).
constexpr int64_t kUseDefaultMemoryLimit = 0;

andrejtonev (approved):
looks good, a few comments

DavIvek:
`breaking` because we changed names of `disk_usage` into `global_disk_usage`, `memory_tracked` into `global_memory_tracked`, `allocation_limit` into `global_runtime_allocation_limit` in `SHOW STORAGE INFO` output, and added a new field `global_license_allocation_limit`. 
Same changes coming in https://github.com/memgraph/memgraph/pull/3952.

gitbuda (approved):

PR #4017, opened 2026-04-09
Feature: Void procedures

Josipmrden:
Implementation of void query procedures for better Cypher compatibility.

Example:
```
WITH true AS predicate
CALL mgps.validate(predicate,'message %d',[42])
RETURN count(*);
```

Behaviour of void procedures is that they execute and return nothing.


Josipmrden:
---
___

### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist

src/query/frontend/ast/cypher_main_visitor.cpp
```diff
@@ -1887,28 +1887,20 @@ antlrcpp::Any CypherMainVisitor::visitCallProcedure(MemgraphCypher::CallProcedur
...
+    call_proc->void_procedure_ = true;
   }
 
   auto *yield_ctx = ctx->yieldProcedureResults();
+  if (yield_ctx && call_proc->void_procedure_) {
+    throw SemanticException("YIELD may not be used on void procedures which do not return any result fields.");
+  }
   if (!yield_ctx) {
-    if ((maybe_found && !maybe_found->second->results.empty()) && !call_proc->void_procedure_) {
+    if (!maybe_found->second->results.empty() && !call_proc->void_procedure_) {
```
imilinovic: void_procedure_ = results.empty() so you can just if check void, dont need both

src/query/plan/operator.cpp
```diff
@@ -7981,6 +7981,14 @@ class CallProcedureCursor : public Cursor {
         memgraph::utils::MemoryTracker::OutOfMemoryExceptionBlocker blocker;
         throw QueryRuntimeException("{}: {}", self_->procedure_name_, *result_.error_msg);
       }
+
+      if (self_->void_procedure_) {
+        // we do not throw if a void procedure returns something, so we clear the iterators here
+        result_.rows.clear();
+        result_row_it_ = result_.rows.end();
+        return true;
+      }
```
imilinovic, line 7937: if we have a batched procedure there would be an infinite loop since we never set stream_exhausted = true, disallow batched void procedures in module registration (or cypher_main_visitor if not possible there, but it should be)

query_modules/mgps_cpp.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+}  // namespace Mgps
+
+extern "C" int mgp_init_module(struct mgp_module *module, struct mgp_memory *memory) {
+  try {
+    mgp::MemoryDispatcherGuard guard{memory};
+
+    AddProcedure(Mgps::Validate,
+                 Mgps::kProcedureValidate,
+                 mgp::ProcedureType::Write,
```
imilinovic: i think its a read procedure not write

query_modules/mgps.py
```diff
@@ -19,3 +19,9 @@ def validate_predicate(predicate: bool, message: str, params: list):
         raise Exception(message % tuple(params))
     else:
         return True
+
+
+@mgp.write_proc
```
imilinovic: think its a read procedure not write

query_modules/mgps_cpp.cpp
```diff
@@ -0,0 +1,91 @@
...
+  const auto params = arguments[2].ValueList();
+
+  // Replace format specifiers: %d (int), %f (double), %s (string)
+  std::size_t pos = 0;
+  for (const auto &param : params) {
+    pos = message.find('%', pos);
+    if (pos == std::string::npos || pos + 1 >= message.size()) break;
+
+    const char spec = message[pos + 1];
+    if (!std::isalpha(spec)) {
```
imilinovic: undefined behavior if spec isn't in range of unsigned char, cast to unsigned char

src/query/procedure/py_module.cpp
```diff
@@ -1193,6 +1193,10 @@ void CallPythonProcedure(const py::Object &py_cb, mgp_list *args, mgp_graph *gra
     if (!py_args) return py::FetchError();
     auto py_res = py_cb.Call(py_graph, py_args);
     if (!py_res) return py::FetchError();
+    if (py_res.Ptr() == Py_None) {
+      // Void procedure - no records to process.
+      return std::nullopt;
+    }
```
imilinovic, line 1205: its possible that somebody wrote a signature but forgot to return a value in the python module, should throw an error in that case, something like below, add a test for it
` 
 if (py_res.Ptr() == Py_None) {
    if (!result->signature.empty()) {
      result->error_msg.emplace(
          "Procedure implementation returned None, but its signature declares "
          "result fields. Did you forget a 'return mgp.Record(...)' statement?");
      return std::nullopt;
    }
    // Void procedure - no records to process.
    return std::nullopt;
  }
`

imilinovic (changes requested):

src/query/procedure/py_module.cpp
Josipmrden: what about mgp.nullable returns?

imilinovic (approved):

PR #4019, opened 2026-04-10
fix: Log rotation when instance restarts

as51340:
Spdlog caches next rotation time in-memory which means that if the instance restarts, it gets a new rotation time. it is easy to imagine then that if the instance often restarts, log files will never be rotated. Fixed by manually running rotation of log files at the Memgraph startup.

### DIFF
--- a/src/flags/logging.cpp
+++ b/src/flags/logging.cpp
@@ -173,7 +173,7 @@ void TurnOnStdErr() {
 // Deletes old log files that should've been rotated by spdlog but aren't. Spdlog saves next rotation time in the memory
 // so if the instance restarts before the scheduled rotations the rotation event will never trigger. Therefore, we are
 // trying to fix this behavior by running manually rotation on the Memgraph startup
-// Assumes there are only log files in the drectory.
+// Assumes there are only log files in the directory.
 // Deletes all files whose last_write_time is older than --log-retention-days
 void CleanLogsDir() {
   if (FLAGS_log_file.empty()) return;
@@ -182,6 +182,7 @@ void CleanLogsDir() {
   auto const log_directory = log_path.parent_path();
   auto const cutoff = std::filesystem::file_time_type::clock::now() - std::chrono::days(FLAGS_log_retention_days);
 
+  // Logs error only at the end, doesn't log for each file
   std::error_code ec;
   for (auto const &entry : std::filesystem::directory_iterator(log_directory, ec)) {
     if (!entry.is_regular_file(ec)) continue;
--- a/src/flags/logging.cpp
+++ b/src/flags/logging.cpp
@@ -14,7 +14,6 @@
 #include <spdlog/logger.h>
 #include <spdlog/sinks/sink.h>
 #include <spdlog/spdlog.h>
-#include <algorithm>
 #include <array>
 #include <cstdint>
 #include <ctime>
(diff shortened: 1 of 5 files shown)


as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - The --log-retention-days flag was not being honored when Memgraph instances restarted frequently (e.g. daily in Kubernetes), because spdlog's internal log rotation only triggers after 24 hours of continuous uptime — each restart reset that timer, so old log files
  accumulated indefinitely. Memgraph now explicitly cleans up log files older than the configured retention period at startup, ensuring the flag works correctly regardless of restart frequency. No action is needed; the fix applies automatically on upgrade.[#4019](https://github.com/memgraph/memgraph/pull/4019)

src/flags/logging.cpp
```diff
@@ -170,4 +170,28 @@ void TurnOnStdErr() {
...
+void CleanLogsDir() {
+  if (FLAGS_log_file.empty()) return;
+
+  auto const log_path = std::filesystem::path{FLAGS_log_file};
+  auto const log_directory = log_path.parent_path();
+  auto const cutoff = std::filesystem::file_time_type::clock::now() - std::chrono::days(FLAGS_log_retention_days);
+
+  std::error_code ec;
+  for (auto const &entry : std::filesystem::directory_iterator(log_directory, ec)) {
+    if (!entry.is_regular_file(ec)) continue;
```
colinbarry, line 188: Because the `ec` is being reused, any failures will be overwritten by a success that follows. As `ec` is only used for logging, not a problem from a correctness point of view, but it does mean that warnings could be swallowed.

(I see why you've done it this way, because you don't want to flood the log on every clean-up failure, and I agree that one warning per `CleanLogsDir()` call is fine.)

src/flags/logging.cpp
```diff
@@ -170,4 +170,28 @@ void TurnOnStdErr() {
...
+    if (entry.last_write_time(ec) < cutoff) {
+      memgraph::utils::DeleteFile(entry.path());
```
colinbarry, line 190: Seeing as we use `std::filesystem` here, may be worth using `std::filesystem::remove(entry.path(), ec)` to remove the file (and catch any errors).

src/flags/logging.cpp
```diff
@@ -170,4 +170,28 @@ void TurnOnStdErr() {
   stderr_sink()->set_level(spdlog::level::trace);
 }
 
+// Deletes old log files that should've been rotated by spdlog but aren't. Spdlog saves next rotation time in the memory
+// so if the instance restarts before the scheduled rotations the rotation event will never trigger. Therefore, we are
+// trying to fix this behavior by running manually rotation on the Memgraph startup
+// Assumes there are only log files in the drectory.
```
colinbarry: Tiny typo in comment: `drectory` -> `directory`

colinbarry (approved):
Approved, with a few minor suggestions. 🔥

src/flags/logging.cpp
as51340: this is what DeleteFiles does actually, it uses filesystem::remove and logs if there is an error

src/flags/logging.cpp
as51340: I documented the behaviour

PR #4020, opened 2026-04-10
fix: Fix use-after-free due with non-sequential deltas in stress tests

colinbarry:
Fixes an issue discovered during testing of #4016. Delta chain traversals stop at a committed sequential delta, but skip past a committed non-sequential one via `next`. The old "fast-break" left `P->next` dangling to a delta about to be freed.  Restricting the fast-break to sequential predecessors and routing non-sequential ones through the `prev_delta->next.store(nullptr)` (under vertex lock) path correctly terminates such traversals at delta `P`, rather than reading into freed memory.

An example scenario in which this problem occurs:
- We have two vertices, `v1` and `v2`
- `TxA` creates an edge on `v1`, and prepends an edge delta to `v1`
- `TxB` creates an edge on `v2`, and prepends an edge delta to `v2`
- `TxC` creates edges on both `v1` and `v2`, prepending `NON_SEQ` edge deltas to both vertices. (`TxA's` and `TxB's` existing deltas are upgraded to `NON_SEQ`, but that is irrelevant for this issue.)
- `TxA` commits at `ts=40`, and its `GCDeltas` are moved to `committed_transactions_` with an unlinkable timestamp of `40`
- `TxC` commits at `ts=50`, and its `GCDeltas` are moved to `waiting_gc_deltas_` as it has the `has_non_sequential_deltas` flag set.
- `TxD` starts at `start_timestamp=55`
- `TxB` commits at `ts=60`, and its `GCDeltas` are moved to `committed_transactions_` with unlinkable timestamp of `60`
- The only in-flight transaction is now `TxD`, which establishes `oldest_active_timestamp=55`
- Later, periodic `CollectGarbage` runs:
  - `waiting_gc_deltas_` are processed. `TxC`s contributors are committed, so its `GCDelta`s are moved to `committed_transactions_` with an unlinkable timestamp of `60` (from the highest timestamp of its downstream contributors, i.e., `TxB`s delta)
  - `committed_transactions_` are swapped into the local `linked_undo_buffers`, which now contains all three `GCDelta` containers from `TxA`, `TxB`, and `TxC`
  - The unlinker loop processes `TxA`s `GCDelta`s, which have an unlinkable timestamp of 40. `40 < 55`, so we know we can unlink it.
  - This delta's `prev` is `TxC`'s delta with a commit timestamp of `50`, which hits the break at:
```cpp
if (prev.delta->commit_info->timestamp.load() < oldest_active_start_timestamp)
```
  - `TxA`s delta is moved to `unlinked_undo_buffers_`
  - But, `TxC`s `GCDeltas`, which have an unlinkable timestamp of 60 are *not* unlinked, and instead are moved back to `committed_transactions_` for later processing in another GC cycle. This leaves `TxC`s delta `.next` on `v1` pointing to a delta that is now supposedly unlinked and will be destructed in a later GC cycle.
- Later, a reader walks `v1`s chain, follows the `v1.delta.next` and arrives at the destructed delta. Boom.

### DIFF
--- a/src/storage/v2/inmemory/storage.cpp
+++ b/src/storage/v2/inmemory/storage.cpp
@@ -2916,27 +2916,6 @@ void InMemoryStorage::CollectGarbage(std::unique_lock<utils::ResourceLock> main_
     }
   });
 
-  {
-    auto guard = std::unique_lock{engine_lock_};
-    uint64_t mark_timestamp = timestamp_;  // a timestamp no active transaction can currently have
-
-    // Deltas from previous GC runs or from aborts can be cleaned up here
-    garbage_undo_buffers_.WithLock([&](auto &garbage_undo_buffers) {
-      guard.unlock();
-      if (main_guard.owns_lock() || mark_timestamp == oldest_active_start_timestamp) {
-        // We know no transaction is active, it is safe to simply delete all the garbage undos
-        // Nothing can be reading them
-        garbage_undo_buffers.clear();
-      } else {
-        // garbage_undo_buffers is ordered, pop until we can't
-        while (!garbage_undo_buffers.empty() &&
-               garbage_undo_buffers.front().mark_timestamp_ <= oldest_active_start_timestamp) {
-          garbage_undo_buffers.pop_front();
-        }
-      }
-    });
-  }
-
   // We don't move undo buffers of unlinked transactions to garbage_undo_buffers
   // list immediately, because we would have to repeatedly take
   // garbage_undo_buffers lock.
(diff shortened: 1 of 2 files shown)


colinbarry:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Fix rare crashes that could occur during shutdown or under heavy load after concurrent edge imports. [#4020](https://github.com/memgraph/memgraph/pull/4020)

colinbarry:
For posterity, PR used to have following additional changes, since reverted:

Fixes various issues discovered during testing of #4016.

- Moves `garbage_undo_buffers_` reclamation to after the unlinking loop. Previously, it could free an upstream delta that the unlinking routine later defererences via `prev.delta->commit_info->timestamp`. Doing the `pop` afterwards ensures every `prev` remains while the unlinking loop may still read it.

- Stop gc alongside other stopped background tasks. Before this, dropping a database left `gc_runner_` running against a partially torn-down `InMemoryStorage`.

Ignition (approved):

PR #4021, opened 2026-04-10
feat: Retry locking data dir to allow distributed storage system to propagate file lock release

as51340:
When users force kill a pod it can happen that the new pod is started while the old one didn't finish shutdown and didn't yet release file lock on data directory. Also, when using distributed storage systems like CephFs, there can be a small delay before ceph server accepts the request from the old pod's client for releasing a file lock. Fix contains of retrying in a loop acquiring the lock for the user-defined timeout.

### DIFF
--- a/src/memgraph.cpp
+++ b/src/memgraph.cpp
@@ -390,23 +390,13 @@ int main(int argc, char **argv) {
   // holding the file opened after timeout occurs
   memgraph::utils::OutputFile lock_file_handle;
   lock_file_handle.Open(data_directory / ".lock", memgraph::utils::OutputFile::Mode::OVERWRITE_EXISTING);
-  auto const start_time = std::chrono::steady_clock::now();
-  auto const lock_file_timeout = std::chrono::seconds{FLAGS_data_dir_lock_acquisition_timeout_sec};
-  constexpr uint16_t sleep_time_ms = 250;
-  while (true) {
-    if (lock_file_handle.AcquireLock()) break;
-    if (std::chrono::steady_clock::now() - start_time > lock_file_timeout) {
-      LOG_FATAL(
-          "Couldn't acquire lock on the storage directory {} within {}s"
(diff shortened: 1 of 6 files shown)


src/memgraph.cpp
```diff
@@ -387,15 +387,28 @@ int main(int argc, char **argv) {
...
-            "Couldn't acquire lock on the storage directory {}"
-            "!\nAnother Memgraph process is currently running with the same "
-            "storage directory, please stop it first before starting this "
-            "process!",
-            data_directory);
+  auto const start_time = std::chrono::steady_clock::now();
+  auto const lock_file_timeout = std::chrono::seconds{FLAGS_data_dir_lock_acquisition_timeout_sec};
+  constexpr uint16_t sleep_time_ms = 250;
+  while (true) {
+    if (lock_file_handle.AcquireLock()) break;
```
Ignition: Push this capability down into `OutputFile` or lower

Ignition (approved):

as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - This PR fixes at least 2 situations in which Memgraph couldn't be started because the exclusive file lock wasn't released fast enough from the previous Memgraph process. Such situations were happening often in K8s when for example users force killed a pod without waiting for a shutdown. Another situation which can trigger such a bug could've happened with distributed storage systems like CephFs because there is a time window dependent on the network round trip before storage client successfully updates the file release on the storage server.  Fix contains of waiting in a loop for the predefined time before actually crashing the process. The timeout is defined with a new flag `--data-dir-lock-acquistion-timeout-sec` set to 30s by default. [#4021](https://github.com/memgraph/memgraph/pull/4021)
- [x] **[ Documentation PR link ](https://github.com/memgraph/documentation/pull/1590)**

PR #4025, opened 2026-04-13
feat: Handle transient disk failures when opening OutputFile

as51340:
Failures to open output files and non-concurrent output files are not fatal anymore in all situations. If unable to open WAL file, this is considered a fatal failure. If unable to open snapshot file for writing whether in a single-threaded env or multi-thread writing env, this is not considered a fatal failure since the next snapshot will be created on the next scheduled interval.

### DIFF
--- a/src/memgraph.cpp
+++ b/src/memgraph.cpp
@@ -394,7 +394,9 @@ int main(int argc, char **argv) {
   // database if it can't open the file for writing or if any other process is
   // holding the file opened after timeout occurs
   memgraph::utils::OutputFile lock_file_handle;
-  lock_file_handle.Open(data_directory / ".lock", memgraph::utils::OutputFile::Mode::OVERWRITE_EXISTING);
+  MG_ASSERT(lock_file_handle.Open(data_directory / ".lock", memgraph::utils::OutputFile::Mode::OVERWRITE_EXISTING),
+            "Failed to open {}/.lock file",
+            data_directory);
   MG_ASSERT(lock_file_handle.AcquireLockWithTimeout(FLAGS_data_dir_lock_acquisition_timeout_sec),
             "Couldn't acquire lock on the storage directory {} within {}s!"
             "Another Memgraph process is currently running with the same "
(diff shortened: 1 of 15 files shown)


as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Memgraph no longer crashes when a file required for snapshot creation cannot be opened due to a transient storage failure (for example, a brief NFS hiccup or a full disk). Instead, the error is logged and snapshot creation is automatically retried at the next scheduled
   interval, keeping the database available throughout. Similarly, if a replica fails to open a file during replication, the error is logged and the main instance will retry sending the file, rather than crashing the replica. WAL and audit log file failures remain fatal,
   as these are critical for data integrity and must succeed. No configuration changes or user action is required. [#4025](https://github.com/memgraph/memgraph/pull/4025)

src/storage/v2/durability/snapshot.cpp
```diff
@@ -213,7 +215,7 @@ void WaitAndCombine(task_results_t &partial_results, SnapshotEncoder &snapshot_e
       if (ec) spdlog::warn("Couldn't remove temporary snapshot part {}: {}", path, ec.message());
```
colinbarry, line 215: Would this warning also be logged if the snapshot file also couldn't be created in the first place? I assume `remove` will try to remove a non-existent file and assign `ec` to some error code.

src/memgraph.cpp
```diff
@@ -444,12 +446,18 @@ int main(int argc, char **argv) {
...
-    audit_log.Start();
+    MG_ASSERT(audit_log.Start(), "Failed to open audit file {}", data_directory / "audit");
   }
   // Setup SIGUSR2 to be used for reopening audit log files, when e.g. logrotate
   // rotates our audit logs.
   MG_ASSERT(memgraph::utils::SignalHandler::RegisterHandler(memgraph::utils::Signal::User2,
-                                                            [&audit_log]() { audit_log.ReopenLog(); }),
+                                                            [&audit_log]() {
+                                                              if (audit_log.ReopenLog()) {
+                                                                spdlog::info("Succesfully reopened audit log");
```
colinbarry: `ReopenLog` will return `false` if the log was never opened due to auditing being disabled. The code will warn "Failed to reopen audit log",  which is technically true, but perhaps not obvious why. Maybe it would be better to log something else in that case, like "Failed to reopen audit log because auditing is not enabled"?

colinbarry: Typo on "Succesfully" -> "Successfully"

colinbarry (approved):
Approved. Only suggestions were for some typos and for you to double check when some things are logged.

src/storage/v2/durability/snapshot.cpp
as51340: no, std::filesystem::remove doesn't set error code if the file doesn't exist, it just returns false

PR #4032, opened 2026-04-14
fix: Network leak and epoll exception on construction

as51340:
Fixed leak in listener epoll. Avoid the crash if epoll_create1 fails. We now throw an exception instead of crashing the database.

### DIFF
--- a/src/io/network/epoll.hpp
+++ b/src/io/network/epoll.hpp
@@ -40,7 +40,7 @@ class Epoll {
     }
   }
 
-  // This is important to be delete because otherwise we could get into a double-close bug for the same file descriptor
+  // "Copy/move deleted: epoll_fd_ is const, and ownership of the FD is non-transferable."
   Epoll(const Epoll &other) = delete;
   Epoll &operator=(const Epoll &other) = delete;
   Epoll(Epoll &&other) = delete;
--- a/src/io/network/epoll.hpp
+++ b/src/io/network/epoll.hpp
@@ -34,19 +34,21 @@ class Epoll {
 
   // Prevent FD from leaking to child processes by using EPOLL_CLOEXEC
   Epoll() : epoll_fd_(epoll_create1(EPOLL_CLOEXEC)) {
-    // epoll_create1 returns an error if there is a logical error in our code
-    // (for example invalid flags) or if there is irrecoverable error. In both
-    // cases it is best to terminate.
     if (epoll_fd_ == -1) {
-      // The exception should be caught wherever the listener is being created
+      // Exception is currently propagated: Epoll -> Listener -> communication::Server -> RpcServer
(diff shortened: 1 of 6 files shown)


as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Fixed a file descriptor leak in the internal epoll mechanism that could cause the system to gradually exhaust available file descriptors over time. Additionally, if epoll initialization fails when creating a replication server, Memgraph now handles the error gracefully and logs a warning instead of crashing. No user action is required. [#4032](https://github.com/memgraph/memgraph/pull/4032)

src/io/network/epoll.hpp
```diff
@@ -32,11 +32,23 @@ class Epoll {
...
+  Epoll() : epoll_fd_(epoll_create1(EPOLL_CLOEXEC)) {
+    if (epoll_fd_ == -1) {
+      // Exception is currently propagated: Epoll -> Listener -> communication::Server -> RpcServer
+      throw utils::BasicException("Error on epoll create: ({}) {}", errno, strerror(errno));
+    }
+  }
+
+  // This is important to be delete because otherwise we could get into a double-close bug for the same file descriptor
+  Epoll(const Epoll &other) = delete;
+  Epoll &operator=(const Epoll &other) = delete;
```
andrejtonev, line 45: && other

src/replication/state.cpp
```diff
@@ -57,6 +57,10 @@ ReplicationState::ReplicationState(std::optional<std::filesystem::path> durabili
         LOG_FATAL("Cannot parse previously saved configuration of replication role.");
         return;
       }
+      case REPL_SERVER_FAILURE: {
+        LOG_FATAL("Couldn't initialize replication server");
```
andrejtonev: FetchReplicationData returns REPL_SERVER_FAILURE when epoll_create1 fails — a resource exhaustion condition. Mapping that to LOG_FATAL (process death) on startup restore defeats the stated goal of "avoid crash." If the intent was graceful degradation, this should fall back to RoleMainData{} (like NOTHING_FETCHED) with a warning, not terminate.

The PR description says "throw instead of crashing" but on the durability restore path, it still crashes via LOG_FATAL.

andrejtonev (approved):
looks good, a few comments

src/replication/state.cpp
as51340: yes but I think that makes sense here. It follows the logic of durability files (snapshots and WALs). If you cannot restore replication on startup, don't even allow Memgraph to be started. What do you think?

as51340: actually, now I am thinking that in HA env I can do graceful degradation to RoleMainData{} because coordinators will demote automatically. In plain replication, I will leave LOG_FATAL since these roles aren't automatically managed

PR #4034, opened 2026-04-14
fix: Remove spdlog::shutdown() before calling std::abort

as51340:
Calling spdlog::shutdown() is not necessary for synchronous logger but it is the only way to flush messages when using async logger. The reason why we don't use spdlog::shudown is because there is then a time window between the invocation of spdlog::shutdown and std::abort which means that the program could segfault at any logging place in the codebase. In the core dump, it would therefore be hard to see the proper reason of the core dump

It is possible if using asynchronous logger that the log line of failure won't be seen. We consider this a reasonable tradeoff.

Changes std::cout to std:cerr for diagnostic debugging because when stdout is redirected to a file or pipe (common in production/containers), it becomes fully buffered, and std::flush won't help if std::abort kills the process before the kernel writes the buffer to disk.

Context https://github.com/gabime/spdlog/wiki/Flush-policy

### DIFF
--- a/src/utils/logging.cpp
+++ b/src/utils/logging.cpp
@@ -12,7 +12,6 @@
 #include "utils/logging.hpp"
 
 #include <fmt/format.h>
-#include <spdlog/async_logger.h>
 #include <spdlog/sinks/stdout_color_sinks.h>
 
 #include <iostream>
@@ -32,9 +31,9 @@ std::string memgraph::logging::MaskSensitiveInformation(std::string_view const i
 }
 
 // It is possible if using asynchronous logger that this log line won't be seen because there is no way force flush
-// messages when using asynchronous queue except calling spdlog::shutdown. That's why we also log err msg on std::cout.
+// messages when using asynchronous queue except calling spdlog::shutdown. That's why we also log err msg on std::cerr.
 // Calling spdlog::shutdown() is not necessary for synchronous logger but it is the only way to flush messages when
-// using async logger. The reason why we don't use spdlog::shudown is because there is then a time window between the
+// using async logger. The reason why we don't use spdlog::shutdown is because there is then a time window between the
 // invocation of spdlog::shutdown and std::abort which means that the program could segfault at any logging place in the
 // codebase. In the core dump, it would therefore be hard to see the proper reason of the core dump
 void memgraph::logging::AssertFailed(std::source_location const loc, char const *expr, std::string const &message) {
@@ -46,9 +45,9 @@ void memgraph::logging::AssertFailed(std::source_location const loc, char const 
       loc.line(),
       expr,
       !message.empty() ? fmt::format("\n\tMessage: '{}'", message) : "");
-  spdlog::critical(msg);
+  spdlog::critical("{}", msg);
   if (std::dynamic_pointer_cast<spdlog::async_logger>(spdlog::default_logger())) {
-    std::cout << msg << '\n';
+    std::cerr << msg << '\n';
   }
   std::terminate();
 }
--- a/src/utils/logging.cpp
+++ b/src/utils/logging.cpp
@@ -28,6 +28,12 @@ std::string memgraph::logging::MaskSensitiveInformation(std::string_view const i
   return std::regex_replace(str, re_all, std::string{kRegexFmt});
 }
 
+// It is possible if using asynchronous logger that this log line won't be seen because there is no way force flush
+// messages when using asynchronous queue except calling spdlog::shutdown. Calling spdlog::shutdown() is not necessary
+// for synchronous logger but it is the only way to flush messages when using async logger. The reason why we don't use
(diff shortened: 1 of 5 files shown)


as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - The PR removes spdlog::shutdown at the aborted application exited. The reason for this is because the program could segfault in the time window between calling spdlog::shutdown and std::abort since the default logger would become nullptr. On fatal failures and when using async logger, we will additionally log the error to std:cerr to make sure the log is seen because spdlog doesn't guarantee all the logs will be flushed on shutdown without using spdlog::shutdown. [#4034](https://github.com/memgraph/memgraph/pull/4034)

src/utils/logging.cpp
```diff
@@ -28,15 +31,25 @@ std::string memgraph::logging::MaskSensitiveInformation(std::string_view const i
...
-  spdlog::critical(
+  auto const msg = fmt::format(
       "\nAssertion failed in file {} at line {}."
       "\n\tExpression: '{}'"
       "{}",
       loc.file_name(),
       loc.line(),
       expr,
       !message.empty() ? fmt::format("\n\tMessage: '{}'", message) : "");
+  spdlog::critical(msg);
```
andrejtonev: msg is a pre-formatted string that may contain { or } from the assertion expression or user message. spdlog::critical forwards to fmt::format(msg, args...), so any stray braces cause fmt::format_error or UB — ironically crashing the crash handler before the actual message is written.

Fix: spdlog::critical("{}", msg);

src/utils/terminate_handler.hpp
```diff
@@ -12,15 +12,17 @@
 #pragma once
 
 #include <execinfo.h>
+#include <spdlog/async_logger.h>
```
andrejtonev: do we need this include?

src/utils/logging.cpp
```diff
@@ -28,15 +31,25 @@ std::string memgraph::logging::MaskSensitiveInformation(std::string_view const i
   return std::regex_replace(str, re_all, std::string{kRegexFmt});
 }
 
+// It is possible if using asynchronous logger that this log line won't be seen because there is no way force flush
+// messages when using asynchronous queue except calling spdlog::shutdown. That's why we also log err msg on std::cout.
+// Calling spdlog::shutdown() is not necessary for synchronous logger but it is the only way to flush messages when
+// using async logger. The reason why we don't use spdlog::shudown is because there is then a time window between the
```
andrejtonev: typo spdlog::shudown

src/utils/logging.cpp
```diff
@@ -28,15 +31,25 @@ std::string memgraph::logging::MaskSensitiveInformation(std::string_view const i
...
       "\nAssertion failed in file {} at line {}."
       "\n\tExpression: '{}'"
       "{}",
       loc.file_name(),
       loc.line(),
       expr,
       !message.empty() ? fmt::format("\n\tMessage: '{}'", message) : "");
+  spdlog::critical(msg);
+  if (std::dynamic_pointer_cast<spdlog::async_logger>(spdlog::default_logger())) {
+    std::cout << msg << '\n';
```
andrejtonev: std::cout is fully buffered when stdout is not a TTY (systemd service, pipe to journald). std::abort() does not flush stdio buffers. The fallback message will be silently lost in every production deployment — exactly the scenario it exists to handle.

Fix:
std::cout << msg << '\n' << std::flush;

src/utils/terminate_handler.hpp
```diff
@@ -29,14 +31,11 @@ inline void TerminateHandler(std::ostream &stream) noexcept {
     try {
       std::rethrow_exception(exc);
     } catch (std::exception &ex) {
-      stream << ex.what() << std::endl << std::endl;
+      stream << ex.what() << "\n\n";
```
andrejtonev, line 33: The previous std::endl flushed ex.what() to the stream before std::abort(). Replacing it with "\n\n" removes that guarantee. The no-arg overload passes std::cout (line 42), which is fully buffered when stdout is not a TTY — i.e. in every production deployment running under systemd. std::abort() does not flush stdio buffers, so the exception message and stacktrace are lost.

Fix: restore the flush explicitly on the last write before abort:


stream << ex.what() << "\n\n";
utils::Stacktrace stacktrace;
stacktrace.dump(stream);
stream << std::flush;
(Or change the no-arg overload to pass std::cerr which is unbuffered by default.)

src/utils/terminate_handler.hpp
as51340: no

src/utils/logging.cpp
as51340: changes to cerr which is unbuffered

src/utils/terminate_handler.hpp
as51340: changed to std::cerr

andrejtonev (approved):

PR #4035, opened 2026-04-14
feat: Disable snapshots, --init-file, --init-data-file. Enable system info and config info queries on coords

as51340:
Snapshots should not be run on coordinators because there is no data stored on coordinators. Using `--init-file` and `--init-data-file` is explicitly prohibited on coordinators and data instances managed by coordinators because we cannot guarantee safe execution in all situations. System info and config queries (`SHOW CONFIG`, `SHOW LICENSE INFO`, `SHOW BUILD INFO`...) are now enabled on coordinators.

### DIFF
--- a/src/dbms/inmemory/storage_helper.hpp
+++ b/src/dbms/inmemory/storage_helper.hpp
@@ -27,17 +27,15 @@ inline std::unique_ptr<storage::Storage> CreateInMemoryStorage(
   auto storage = std::make_unique<storage::InMemoryStorage>(
       std::move(config), std::nullopt, std::move(invalidator), std::move(database_protector_factory));
 
-  if (storage->config_.durability.snapshot_wal_mode != storage::Config::Durability::SnapshotWalMode::DISABLED) {
-    // Connect replication state and storage
-    storage->CreateSnapshotHandler(
-        [storage = storage.get()]() -> std::expected<void, storage::InMemoryStorage::CreateSnapshotError> {
-          auto result = storage->CreateSnapshot();
-          if (!result) {
-            return std::unexpected(result.error());
-          }
-          return {};
-        });
-  }
+  // Connect replication state and storage
+  storage->CreateSnapshotHandler(
+      [storage = storage.get()]() -> std::expected<void, storage::InMemoryStorage::CreateSnapshotError> {
+        auto result = storage->CreateSnapshot();
+        if (!result) {
+          return std::unexpected(result.error());
+        }
+        return {};
+      });
 
   return std::move(storage);
 }
--- a/src/dbms/inmemory/storage_helper.hpp
+++ b/src/dbms/inmemory/storage_helper.hpp
@@ -21,23 +21,24 @@ inline std::unique_ptr<storage::Storage> CreateInMemoryStorage(
     storage::Config config,
     storage::PlanInvalidatorPtr invalidator = std::make_unique<storage::PlanInvalidatorDefault>(),
     std::function<storage::DatabaseProtectorPtr()> database_protector_factory = nullptr) {
-  const auto name = config.salient.name;
+  const bool is_coordinator = config.durability.is_coordinator;
 
   // Use default safe factory from Storage constructor for basic usage
   auto storage = std::make_unique<storage::InMemoryStorage>(
       std::move(config), std::nullopt, std::move(invalidator), std::move(database_protector_factory));
 
-  // Connect replication state and storage
-  storage->CreateSnapshotHandler(
-      [storage = storage.get()]() -> std::expected<void, storage::InMemoryStorage::CreateSnapshotError> {
-        auto result = storage->CreateSnapshot();
-        if (!result) {
-          return std::unexpected(result.error());
-        }
-        return {};
-      });
+  if (!is_coordinator) {
+    storage->CreateSnapshotHandler(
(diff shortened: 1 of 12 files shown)


as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
   - Snapshots will not be run on coordinators anymore because there is no data stored on coordinators. Using `--init-file` and `--init-data-file` is now explicitly prohibited in HA mode because we cannot guarantee safe execution in all situations neither on coordinators or data instances. System info and config queries (`SHOW CONFIG`, `SHOW LICENSE INFO`, `SHOW BUILD INFO`...) are now enabled on coordinators. [#4035](https://github.com/memgraph/memgraph/pull/4035)

src/memgraph.cpp
```diff
@@ -562,6 +555,33 @@ int main(int argc, char **argv) {
...
+                                             !coordination_setup.coordinator_hostname.empty();
+
+  if (is_valid_coordinator_instance) {
+    if (!FLAGS_init_file.empty()) {
+      LOG_FATAL(
+          "Coordinator instances don't support --init-file flag. Please restart the instance by removing this flag.");
+    }
+    if (!FLAGS_init_data_file.empty()) {
+      LOG_FATAL(
+          "Coordinator instances don't support --init-data_file flag. Please restart the instance by removing this "
```
andrejtonev: type --init-data_file

tests/e2e/high_availability/durability.py
```diff
@@ -222,6 +228,20 @@ def cleanup_after_test():
...
+        test_name=test_name, snapshot_interval_sec="1"
+    )
+    interactive_mg_runner.start_all(instances_description, keep_directories=False)
+
+    build_dir = os.path.join(interactive_mg_runner.PROJECT_DIR, "build", "e2e", "data")
+    data_dir_coord_1 = f"{build_dir}/{get_data_path(file, test_name)}/coordinator_1"
+    snapshot_dir_coord_1 = f"{data_dir_coord_1}/snapshots"
+
+    assert count_files(snapshot_dir_coord_1) == 0
+
```
andrejtonev, line 246: The test checks count_files == 0 immediately after start_all, before any snapshot interval
could elapse. This passes trivially regardless of whether snapshot disabling actually works.
Should sleep for a few seconds (interval is 1s, so ~5s) before asserting, similar to how
test_snapshots_on_replica uses mg_sleep_and_assert_eval_function.

Also the comment on line 232 says "every 3s" but snapshot_interval_sec="1".

src/replication_handler/replication_handler.cpp
```diff
@@ -64,7 +64,8 @@ void RecoverReplication(utils::Synchronized<ReplicationState, utils::RWSpinLock>
           "The instance has the MAIN replication role, but durability logs and snapshots are disabled. Please "
           "consider "
           "enabling durability by using --storage-snapshot-interval-sec and --storage-wal-enabled flags because "
-          "without write-ahead logs this instance is not replicating any data.");
+          "without write-ahead logs this instance is not replicating any data. Ignore this warning if this is "
```
andrejtonev: "Ignore this warning if this is coordinator instance." leaks coordinator awareness into
replication_handler. Better to suppress the warning at the call site (main.cpp) when
is_valid_coordinator_instance is true, rather than patching the message here.

andrejtonev: The warning fires inside RecoverReplication, which is called from ReplicationHandler's constructor, which is called in main() at line 739 — after is_valid_coordinator_instance is already known.

The cleanest suppression point is ReplicationHandler's constructor — pass is_coordinator as a bool parameter and thread it into RecoverReplication. That way:

replication_handler.cpp has no coordinator logic leaking in — the bool is just a generic "suppress durability warning"
main.cpp is the only place that knows about coordinator topology and passes the flag
The warning stays in the right layer (replication handler), just conditionally
Alternatively, name the parameter suppress_durability_warning to keep it fully generic — RecoverReplication doesn't need to know why it's suppressed.

Concretely:


// replication_handler.hpp
ReplicationHandler(..., bool suppress_durability_warning = false);

// replication_handler.cpp - RecoverReplication signature gains the flag
// the spdlog::warn becomes:
if (!suppress_durability_warning && snapshot_wal_mode == DISABLED) {
    spdlog::warn(...);
}

// memgraph.cpp line 739
auto replication_handler = memgraph::replication::ReplicationHandler{
    repl_state, dbms_handler, system,
#ifdef MG_ENTERPRISE
    *auth_,
#endif
    *parameters,
    is_valid_coordinator_instance  // suppress durability warning on coordinators
};
This is the minimal, clean fix — zero coordinator awareness inside replication_handler.cpp.

andrejtonev (approved):

src/dbms/inmemory/storage_helper.hpp
```diff
@@ -21,23 +21,24 @@ inline std::unique_ptr<storage::Storage> CreateInMemoryStorage(
     storage::Config config,
     storage::PlanInvalidatorPtr invalidator = std::make_unique<storage::PlanInvalidatorDefault>(),
     std::function<storage::DatabaseProtectorPtr()> database_protector_factory = nullptr) {
-  const auto name = config.salient.name;
+  const bool is_coordinator = config.durability.is_coordinator;
```
andrejtonev: I'd just move it to the root config (instead of the durability config)

andrejtonev (approved):

PR #4038, opened 2026-04-15
fix: Harden Python code

as51340:
## What

Fix a cluster of Python-embedding safety issues in Memgraph's procedure layer, including the `gilstate_tss_set` fatal error at shutdown, refcount corruption around `PyTuple_SetItem`, missing bounds checks on list access, corruption of `__main__` by the dependency
scanner. Also caches the `mgp` module import on hot conversion paths and releases the GIL during file I/O inside `ProcessFileDependencies`.

## Why

Third-party Python extensions (DGL, PyTorch, NumPy) spawn background threads and open-MP worker pools that race with CPython's TSS teardown during `Py_Finalize`, producing `gilstate_tss_set: failed to set current tstate` fatal aborts on exit (bpo-42969). Separately, several long-standing Python C API misuses (stolen refs double-decremented, borrowed refs treated as owned, out-of-bounds `PyList_GET_ITEM`) risked refcount corruption
and crashes, and the dependency scanner was mutating `__main__`'s global dict.

## How

Changes are grouped by subsystem:

- **Process entry point (`src/memgraph.cpp`)**
  - Skip `Py_Finalize()` at shutdown. The OS reclaims resources and this avoids racing third-party threads through CPython's TSS teardown (bpo-42969).
  - Fix `PyTuple_SetItem` refcount misuse in the `dlopenflags` setup: the tuple steals `flag`, and `setdl` is borrowed from `PySys_GetObject` — neither should be `Py_DECREF`'d.

- **Python helper (`src/py/py.hpp`)**
  - `Object::FromBorrow` re-checks `Py_IsFinalizing()` around the `Py_INCREF` to avoid incrementing refs during teardown.
  - Replace unchecked `PyList_GET_ITEM` in `FormatException` with `PyList_GetItem` and skip/clear on nullptr.
  - Document on `SetAttr` overloads that callers must `FetchError()` or `PyErr_Clear()` on failure.

- **Module loading (`src/query/procedure/module.cpp`)**
  - Release the GIL (`Py_BEGIN_ALLOW_THREADS`) around the `ReadFile` call in `ProcessFileDependencies` and assert the GIL is held on entry.
  - Run the dependency scanner in a `PyDict_Copy` sandbox instead of `__main__`'s globals, so the scanner script cannot corrupt the main namespace.
  - Wrap `PyDict_GetItemString`'s borrowed `modules` result in `Object::FromBorrow`.
  - Log scan failures via `spdlog::warn` with `FormatException` output, and log stale `sys.modules` cleanup failures.

- **Value conversion (`src/query/procedure/py_module.cpp`)**
  - Introduce `MgpValueToPyObjectImpl` / `MgpListToPyTupleImpl` that accept a pre-imported `mgp` module, so recursive conversion no longer re-imports `mgp` for every vertex/edge/path/point/enum element. Public entry points import once and thread the reference through.
  - Cache the `mgp` import across the up-to-6 `is_mgp_instance` calls per value in `PyObjectToMgpValue`.
  - Fix `AddRecordFromPython`: wrap the borrowed `PyTuple_GetItem` result in `FromBorrow` so the cached value survives later dict/tuple mutations.
  - Replace the `std::bad_alloc` thrown on `PyUnicode_AsUTF8` failure with `std::invalid_argument` (and add the same check on the string-value path) — the actual failure mode is UTF-8 decoding, not allocation.
  - Add `MG_ASSERT(gMgpUnknownError != nullptr)` in `RaiseExceptionFromErrorCode` and document the one-time-write threading model for the `gMgp*` globals.
  - Replace the stale "not thread safe" note in `WithMgpModule` with an explicit contract that `ModuleRegistry::LoadModuleLibrary` serialises loading.


docs/python_bindings_audit.md
```diff
@@ -0,0 +1,486 @@
+# Python Bindings Audit — Memgraph `mgp` / Call Procedures
```
andrejtonev: remove this file

src/py/py.hpp
```diff
@@ -76,7 +76,7 @@ class [[nodiscard]] Object final {
   static Object FromBorrow(PyObject *ptr) noexcept {
     if (ptr && !Py_IsFinalizing()) {
       EnsureGIL gil;
-      Py_INCREF(ptr);
+      if (!Py_IsFinalizing()) Py_INCREF(ptr);
```
as51340, line 79: Don't we already check for finalising?

src/memgraph.cpp
```diff
@@ -296,6 +296,11 @@ int main(int argc, char **argv) {
   // Set program name, so Python can find its way to runtime libraries relative
   // to executable.
   Py_SetProgramName(program_name);
+  // Prevent NumPy/BLAS/MKL/OpenMP from spawning O(100) threads that contend
+  // with Memgraph workers. The 0 overwrite flag lets users override via env.
+  setenv("OMP_NUM_THREADS", "1", 0);
+  setenv("MKL_NUM_THREADS", "1", 0);
+  setenv("OPENBLAS_NUM_THREADS", "1", 0);
```
as51340, line 303: Isn't this kinda dangerous? I mean why setting OMP NUM THREADS to 0? Maybe set to hardware concurrency

andrejtonev (approved):

as51340:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Memgraph's embedded Python runtime has been hardened against crashes and resource contention that could occur when running Python query procedures, especially those using libraries like NumPy, PyTorch, or DGL. Users should see more reliable shutdowns (no more sporadic
  fatal errors on exit) and better isolation when loading procedure modules, since the dependency scanner no longer pollutes the interpreter's main namespace. No user action is required; existing Python procedures
  continue to work unchanged. [#4038](https://github.com/memgraph/memgraph/pull/4038)

src/memgraph.cpp
andrejtonev: the value is 1
the 0 is a behavioral flag so that we don't overwrite if already defined
not completely wrong, but will remove for now.

PR #4039, opened 2026-04-15
testing: Jepsen test  num_txns_behind_main should be non-negative

as51340:
TODO: Needs fixing because the test it passed even when I've set it to pos value (for testing purposes)


as51340:
Replication lag can be negative when one of replicas becomes main without previously committing all txns. Especially possible with SYNC replication mode. Hence, closing this PR

PR #4040, opened 2026-04-16
fix: Routing table with the negative lag and improve ReplicationException

as51340:
Replication exception will now take into account also failures when starting txns in STRICT SYNC mode. Main's number of committed transactions is now updated before replicas', to prevent negative lag situations. Negative lag can occur however if an instance without having all txns committed becomes main. Such a situation is possible in SYNC replication mode. Previously, if a situation with the negative lag occurred, that number would overflow. We now explicitly disable routing to replicas which are "in front" of main. This in front should last for a short time but it is still possible.

### DIFF
--- a/src/coordination/include/coordination/utils.hpp
+++ b/src/coordination/include/coordination/utils.hpp
@@ -79,8 +79,9 @@ auto CreateRoutingTable(std::vector<DataInstanceContext> const &raft_log_data_in
           return true;
         }
         // return true if cached lag is smaller than max_allowed_replica_read_lag. This also allows routing to replicas
-        // with negative lag. negative lag can occur when an instance becomes main without comitting all txns
-        return db_it->second < max_replica_read_lag;
+        // with negative lag. negative lag can occur when an instance becomes main without comitting all txns. Check for
+        // overflow also
+        return db_it->second < 0 || db_it->second < max_replica_read_lag;
       };
 
   auto readers = raft_log_data_instances | ranges::views::filter(std::not_fn(is_instance_main_func)) |
--- a/src/coordination/include/coordination/utils.hpp
+++ b/src/coordination/include/coordination/utils.hpp
@@ -78,9 +78,6 @@ auto CreateRoutingTable(std::vector<DataInstanceContext> const &raft_log_data_in
         if (db_it == replica_it->second.end()) {
           return true;
         }
-
-        spdlog::trace("Comparing: {} < {}", db_it->second, max_replica_read_lag);
-
         // return true if cached lag is smaller than max_allowed_replica_read_lag. This also allows routing to replicas
         // with negative lag. negative lag can occur when an instance becomes main without comitting all txns
         return db_it->second < max_replica_read_lag;
--- a/src/coordination/include/coordination/utils.hpp
+++ b/src/coordination/include/coordination/utils.hpp
@@ -79,6 +79,8 @@ auto CreateRoutingTable(std::vector<DataInstanceContext> const &raft_log_data_in
           return true;
         }
 
+        spdlog::trace("Comparing: {} < {}", db_it->second, max_replica_read_lag);
+
         // return true if cached lag is smaller than max_allowed_replica_read_lag
         return db_it->second < max_replica_read_lag;
       };
--- a/src/coordination/include/coordination/utils.hpp
+++ b/src/coordination/include/coordination/utils.hpp
@@ -81,7 +81,8 @@ auto CreateRoutingTable(std::vector<DataInstanceContext> const &raft_log_data_in
 
         spdlog::trace("Comparing: {} < {}", db_it->second, max_replica_read_lag);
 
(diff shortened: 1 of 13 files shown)


src/coordination/include/coordination/utils.hpp
```diff
@@ -79,6 +79,8 @@ auto CreateRoutingTable(std::vector<DataInstanceContext> const &raft_log_data_in
           return true;
         }
 
+        spdlog::trace("Comparing: {} < {}", db_it->second, max_replica_read_lag);
+
         // return true if cached lag is smaller than max_allowed_replica_read_lag
         return db_it->second < max_replica_read_lag;
```
as51340, line 85: Add check for negative regardless of the fix

src/coordination/include/coordination/utils.hpp
```diff
@@ -78,9 +78,10 @@ auto CreateRoutingTable(std::vector<DataInstanceContext> const &raft_log_data_in
         if (db_it == replica_it->second.end()) {
           return true;
         }
-
-        // return true if cached lag is smaller than max_allowed_replica_read_lag
-        return db_it->second < max_replica_read_lag;
+        // return true if cached lag is smaller than max_allowed_replica_read_lag. This also allows routing to replicas
+        // with negative lag. negative lag can occur when an instance becomes main without comitting all txns. Check for
+        // overflow also
+        return db_it->second < 0 || db_it->second < max_replica_read_lag;
```
andrejtonev: 1. can max_replica_read_lag be 0?
2. should this be < or <=?

src/coordination/include/coordination/utils.hpp
```diff
@@ -78,9 +78,10 @@ auto CreateRoutingTable(std::vector<DataInstanceContext> const &raft_log_data_in
         if (db_it == replica_it->second.end()) {
           return true;
         }
-
-        // return true if cached lag is smaller than max_allowed_replica_read_lag
-        return db_it->second < max_replica_read_lag;
+        // return true if cached lag is smaller than max_allowed_replica_read_lag. This also allows routing to replicas
+        // with negative lag. negative lag can occur when an instance becomes main without comitting all txns. Check for
```
andrejtonev: "negative lag can occur when an instance becomes main without comitting all txns."
What do you mean by this?
Coordinators chose a main that is behind?

src/replication_handler/replication_handler.cpp
```diff
@@ -439,19 +439,29 @@ std::pair<ReplicationHandler::MainResT, ReplicationHandler::ReplicasResT> Replic
...
+              repl_storage_state.commit_ts_info_.load(std::memory_order_acquire).num_committed_txns_;
+          main.emplace(std::string{db_acc->storage()->uuid()}, num_main_committed_txns);
+
+          for (auto &repl_storage_client : storage_clients) {
+            auto const replica_name = repl_storage_client->Name();
+            auto replica_it = replicas.find(replica_name);
+            DMG_ASSERT(replica_it != replicas.end(), "No info for replica {}", replica_name);
+            auto old_value_it = replica_it->second.find(db_name);
+            DMG_ASSERT(old_value_it != replica_it->second.end(), "No info for db {}", db_name);
+            old_value_it->second = num_main_committed_txns - old_value_it->second;
```
andrejtonev, line 464: We should check for underflow here

src/coordination/include/coordination/utils.hpp
andrejtonev: If this is a split brain, we shouldn't return the instances that are in the future

src/replication_handler/replication_handler.cpp
andrejtonev: another possible underflow

src/coordination/include/coordination/utils.hpp
as51340: <= max_replica_read_lag

src/coordination/include/coordination/utils.hpp
as51340: situation: main committed but replicas didn't. main goes down, one of replicas becomes main. Old main becomes replica.


Split brain is something else. I think you are right though about allowing reading txns that will get removed.

src/replication_handler/replication_handler.cpp
as51340: no need because old_value_it->second is int64_t not unsigned hence it will at the end be correct value

andrejtonev (approved):

as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Main's number of committed transactions is now incremented always before replicas' on commit in order to avoid negative lag situation occurring. A negative lag can however occur for a short period of time if an instance that didn't commit all transactions becomes main. In such situations, we forbid routing read queries to replicas which are in front of main. [#4040](https://github.com/memgraph/memgraph/pull/4040)

PR #4042, opened 2026-04-16
fix: Stop assigned user roles leaking across databases

### LINKED ISSUE #3835: User with 2 roles gets assigned wrong privileges on a certain database
Report shows that user with role A, and role B (with SSO) 
can create all the nodes on database a (connected with role A), although he doesn't have all the privileges with role A. 

However, the role B (for only the database b) -> he has there all the privileges and that might be the issue where the role privileges are getting mixed up.

colinbarry:
Fixes fine-grained access control role leakage across databases for users with multiple roles, ensuring permission checks are evaluated only against roles that apply to the currently active database. 

This ensures that we no longer combine roles in either of the following situations:
- Per-database roles for databases other than the current.
- Global roles which do not have access to the current database.

#fixes https://github.com/memgraph/memgraph/issues/3835


colinbarry:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Fixes fine-grained access control role leakage across databases for users with multiple roles, ensuring permission checks are evaluated only against roles that apply to the currently active database. [#4042](https://github.com/memgraph/memgraph/pull/4042)

src/glue/auth_checker.cpp
```diff
@@ -206,13 +208,14 @@ bool AuthChecker::CanImpersonate(const memgraph::auth::Roles &roles, const memgr
 #endif
 
 #ifdef MG_ENTERPRISE
-FineGrainedAuthChecker::FineGrainedAuthChecker(auth::UserOrRole user_or_role, const memgraph::query::DbAccessor *dba)
-    : user_or_role_{std::move(user_or_role)}, dba_(dba) {};
+FineGrainedAuthChecker::FineGrainedAuthChecker(auth::UserOrRole user_or_role, const memgraph::query::DbAccessor *dba,
+                                               std::string db_name)
+    : user_or_role_{std::move(user_or_role)}, dba_(dba), db_name_{std::move(db_name)} {};
```
DavIvek: Why are we explicitly passing `db_name` and fetching it from `DbAccessor` before calling the constructor, if it’s already part of `DbAccessor` and we can access it when needed?

src/glue/auth_checker.cpp
```diff
@@ -206,13 +208,14 @@ bool AuthChecker::CanImpersonate(const memgraph::auth::Roles &roles, const memgr
...
 
 auth::FineGrainedAccessPermissions const &FineGrainedAuthChecker::GetCachedLabelPermissions() const {
   if (!cached_label_permissions_) {
-    cached_label_permissions_ = std::visit(memgraph::utils::Overloaded{[](auto const &user_or_role) {
-                                             return user_or_role.GetFineGrainedAccessLabelPermissions();
+    cached_label_permissions_ = std::visit(memgraph::utils::Overloaded{[this](auto const &user_or_role) {
+                                             return user_or_role.GetFineGrainedAccessLabelPermissions(db_name_);
```
DavIvek, line 216: `models.cpp`:
```
const FineGrainedAccessPermissions &Roles::GetFineGrainedAccessLabelPermissions(
    std::optional<std::string_view> db_name) const {
  if (roles_.empty()) return empty_permissions;

  FineGrainedAccessPermissions combined_permissions;
  for (const auto &role : roles_) {
    if (!db_name || role.HasAccess(*db_name)) {
      combined_permissions = Merge(combined_permissions, role.fine_grained_access_handler().label_permissions());
    }
  }
  static FineGrainedAccessPermissions result;
  result = combined_permissions;
  return result;
}
```

Different threads pass different db_names, so the computed values genuinely differ. A race now means thread A on db1 can silently end up with thread B's db2 permissions cached

DavIvek (changes requested):

src/glue/auth_checker.cpp
colinbarry: Yeah, good catch. Fixed.

src/glue/auth_checker.cpp
colinbarry: Fixed: removed that awful, pointless`static` cache!

DavIvek (approved):
🚀

PR #4051, opened 2026-04-20
Label property index not cleaned in analytical mode

Josipmrden:
Test added in this PR shows the behaviour, leading to memory not being cleaned inside the index.


Josipmrden:
@Ignition  fixes with:
https://github.com/memgraph/memgraph/pull/4071

PR #4052, opened 2026-04-21
refactor: Remove DBMS handler, repl state and replication handler from coordinators

as51340:
Code components dbms handler, replication state and replication handler are now nullopt for coordinator instances. This should make coordinators more stable because of independence with the disk for example, easier to run and manage. 
Fixes the bug in which storage was always required for List TypedValue.

### DIFF
--- a/src/glue/SessionHL.cpp
+++ b/src/glue/SessionHL.cpp
@@ -548,6 +548,8 @@ void SessionHL::BeginTransaction(const bolt_map_t &extra) {
 
 void SessionHL::Configure(const bolt_map_t &run_time_info) {
 #ifdef MG_ENTERPRISE
+  // Coordinators have no dbms_handler and no databases to switch to, so there is nothing to configure.
+  if (flags::CoordinationSetupInstance().IsCoordinator()) return;
   runtime_config_.Configure(run_time_info, interpreter_.in_explicit_transaction_);
 #else
   (void)run_time_info;
@@ -615,9 +617,6 @@ bolt_map_t SessionHL::DecodeSummary(const std::map<std::string, memgraph::query:
 
 #ifdef MG_ENTERPRISE
 void RuntimeConfig::Configure(const bolt_map_t &run_time_info, bool in_explicit_tx) {
-  // Coordinators have no dbms_handler and no databases to switch to, so there is nothing to configure.
-  if (flags::CoordinationSetupInstance().IsCoordinator()) return;
-
   // NOTE: Once in a transaction, the drivers stop explicitly sending the config and count on using it until commit
   // Runtime config is sent at the beginning of the transaction, but is missing during the transaction
   if (in_explicit_tx || (previous_run_time_info_ && run_time_info == *previous_run_time_info_)) return;
--- a/src/glue/SessionHL.cpp
+++ b/src/glue/SessionHL.cpp
@@ -19,6 +19,7 @@
 #include "auth/auth.hpp"
 #include "auth/exceptions.hpp"
 #include "dbms/constants.hpp"
+#include "flags/coord_flag_env_handler.hpp"
 #include "flags/run_time_configurable.hpp"
 #include "frontend/ast/ast.hpp"
 #include "glue/SessionHL.hpp"
@@ -614,6 +615,9 @@ bolt_map_t SessionHL::DecodeSummary(const std::map<std::string, memgraph::query:
 
 #ifdef MG_ENTERPRISE
 void RuntimeConfig::Configure(const bolt_map_t &run_time_info, bool in_explicit_tx) {
+  // Coordinators have no dbms_handler and no databases to switch to, so there is nothing to configure.
+  if (flags::CoordinationSetupInstance().IsCoordinator()) return;
+
   // NOTE: Once in a transaction, the drivers stop explicitly sending the config and count on using it until commit
   // Runtime config is sent at the beginning of the transaction, but is missing during the transaction
   if (in_explicit_tx || (previous_run_time_info_ && run_time_info == *previous_run_time_info_)) return;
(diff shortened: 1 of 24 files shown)


as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist

src/glue/communication.cpp
```diff
@@ -157,7 +157,6 @@ storage::Result<Value> ToBoltValue(const query::TypedValue &value, const storage
 
     // Database is required
     case query::TypedValue::Type::List: {
-      check_db();
```
andrejtonev, line 160: don't think we should remove this

src/glue/SessionHL.cpp
```diff
@@ -614,6 +615,9 @@ bolt_map_t SessionHL::DecodeSummary(const std::map<std::string, memgraph::query:
 
 #ifdef MG_ENTERPRISE
 void RuntimeConfig::Configure(const bolt_map_t &run_time_info, bool in_explicit_tx) {
+  // Coordinators have no dbms_handler and no databases to switch to, so there is nothing to configure.
+  if (flags::CoordinationSetupInstance().IsCoordinator()) return;
```
andrejtonev: I'd rather put this in the SessionHL logic that in the RuntimeConfig

src/query/interpreter.cpp
```diff
@@ -6824,6 +6824,11 @@ PreparedQuery PrepareSystemInfoQuery(ParsedQuery parsed_query, bool in_explicit_
 
   switch (info_query->info_type_) {
     case SystemInfoQuery::InfoType::STORAGE: {
+#ifdef MG_ENTERPRISE
```
andrejtonev, line 6827: I feel like there will many places we need to do this check. 
How are we limiting the queries we can execute in general?

src/query/interpreter.cpp
```diff
@@ -8833,6 +8838,19 @@ Interpreter::PrepareResult Interpreter::Prepare(ParseRes parse_res, UserParamete
       return system_txn;
     });
 
+#ifdef MG_ENTERPRISE
+    // TODO(antoniofilipovic) extend to cover Lab queries
+    // Must run before SetupDatabaseTransaction below: coordinators have no db_acc, so any query that
+    // requests a storage accessor (e.g. CypherQuery) would otherwise throw a generic "Database required"
+    // error instead of this clearer coordinator-specific message.
+    if (interpreter_context_->coordinator_state_ && interpreter_context_->coordinator_state_->IsCoordinator() &&
```
andrejtonev, line 8846: I guess the one before is a special case

src/memgraph.cpp
```diff
@@ -144,9 +144,16 @@ void CheckSuspiciousPositionalArgs(int argc, char **argv) {
...
-                         std::string cypherl_file_path, memgraph::audit::Log *audit_log = nullptr) {
-  memgraph::query::Interpreter interpreter(&ctx, db_acc);
+void InitFromCypherlFile(memgraph::query::InterpreterContext &ctx,
+                         std::optional<memgraph::dbms::DatabaseAccess> &db_acc, std::string cypherl_file_path,
+                         memgraph::audit::Log *audit_log = nullptr) {
+  auto interpreter = std::invoke([&ctx, &db_acc]() {
+    if (db_acc.has_value()) {
+      return memgraph::query::Interpreter{&ctx, *db_acc};
+    }
+    return memgraph::query::Interpreter{&ctx};
```
andrejtonev, line 154: Is this so we still allow cypher init files but limit the scope to database-less queries only?

src/memgraph.cpp
```diff
@@ -715,38 +721,58 @@ int main(int argc, char **argv) {
 
 #endif
 
-  memgraph::dbms::DbmsHandler dbms_handler(db_config);
+  std::optional<memgraph::dbms::DbmsHandler> dbms_handler;
+#ifdef MG_ENTERPRISE
+  if (!is_coordinator_instance) {
```
andrejtonev, line 727: you could just have the bool default to false and not have the #if #else here

src/glue/communication.cpp
as51340: it's fine, we need storage only for storage types not for generic ones like null, bool, int, double... We don't check_db for map neither, check lines above

src/glue/SessionHL.cpp
as51340: makes sense

src/query/interpreter.cpp
as51340: We forbid everything except:
```
if (interpreter_context_->coordinator_state_ && interpreter_context_->coordinator_state_->IsCoordinator() &&
        !utils::Downcast<CoordinatorQuery>(parsed_query.query) && !utils::Downcast<SettingQuery>(parsed_query.query) &&
        !utils::Downcast<ReloadSSLQuery>(parsed_query.query) && !utils::Downcast<ShowConfigQuery>(parsed_query.query) &&
        !utils::Downcast<SystemInfoQuery>(parsed_query.query)) {
      throw QueryRuntimeException("Coordinator can run only coordinator queries!");
    }
```

This change is specific because the query is allowed on coords and data instances but part of the query show storage info, is not allowed on coords while for example show build info is

src/query/interpreter.cpp
as51340: exactly 😄

src/memgraph.cpp
as51340: not sure exactly what you mean but we can create interpreter both with db_acc and without. We have a flag --ha-init-file which needs interpreter but doesn't need db access. I think nobody uses it though so probably good candidate for deprecation in the next few releases.

src/memgraph.cpp
as51340: yes

andrejtonev (approved):

PR #4065, opened 2026-04-23
fix: Support LIMIT on both sides of a UNION

### LINKED ISSUE #3909: Query hangs when both UNION branches contain LIMIT
**Memgraph version**
Memgraph version: `3.8.1`

**Environment**
- Docker image used locally: `memgraph/memgraph:latest`
- Current explicit Docker tag for reproducible setup: `memgraph/memgraph:3.8.1-relwithdebinfo`
- Deployment: Docker
- Host OS: Windows
- Client used: `mgconsole`

**Describe the bug**
A query using `UNION` where both branches contain their own `LIMIT` does not return on Memgraph 3.8.1.
The same query returns immediately on Neo4j.

**To Reproduce**
```cypher
RETURN 1 AS x LIMIT 1
UNION
RETURN 2 AS x LIMIT 1;
```

I also reproduced the same behavior with a graph-backed variant:

```cypher
MATCH (n) DETACH DELETE n;
CREATE (:T {v: 1}), (:T {v: 2});
MATCH (n:T) RETURN n LIMIT 1
UNION
MATCH (n:T) RETURN n LIMIT 1;
```

**Expected behavior**
The query should return quickly.

**Actual behavior**
the query does not return within the timeout window and appears to hang indefinitely.

**What I checked**
I minimized the original failing query and verified the following:

- `UNION` without per-branch `LIMIT` does not hang.
- A single `LIMIT` on only one branch does not hang.
- A single branch with `LIMIT` does not hang.
- The hang appears only when both `UNION` branches contain their own `LIMIT`.

Examples that do **not** hang:

```cypher
RETURN 1 AS x UNION RETURN 2 AS x;
```

```cypher
RETURN 1 AS x LIMIT 1 UNION RETURN 2 AS x;
```

```cypher
RETURN 1 AS x UNION RETURN 2 AS x LIMIT 1;
```

**Verification Environment**
 Plain memgraph Docker image

colinbarry:
Fixes `UNION` so that it no longer tries `Pull`ing from an already exhausted left-hand branch.

#fix [3909](https://github.com/memgraph/memgraph/issues/3909)

### DIFF
--- a/src/query/plan/operator.cpp
+++ b/src/query/plan/operator.cpp
@@ -11129,6 +11129,7 @@ bool Limit::LimitCursor::Pull(Frame &frame, ExecutionContext &context) {
   }
 
   // check we have not exceeded the limit before pulling
+  if (!shared_quota_) return false;
   if (shared_quota_->Decrement() == 0) {
     shared_quota_.reset();  // Important to release any remaining resource for other threads
     return false;
--- a/src/query/plan/operator.cpp
+++ b/src/query/plan/operator.cpp
@@ -7398,18 +7398,18 @@ bool Union::UnionCursor::Pull(Frame &frame, ExecutionContext &context) {
   AbortCheck(context);
 
   utils::pmr::unordered_map<std::string, TypedValue> results(context.evaluation_context.memory);
-  if (left_cursor_->Pull(frame, context)) {
+  if (!is_left_exhausted_ && left_cursor_->Pull(frame, context)) {
     // collect values from the left child
     for (const auto &output_symbol : self_.left_symbols_) {
       results[output_symbol.name()] = frame[output_symbol];
     }
-  } else if (right_cursor_->Pull(frame, context)) {
+  } else {
+    is_left_exhausted_ = true;
+    if (!right_cursor_->Pull(frame, context)) return false;
     // collect values from the right child
     for (const auto &output_symbol : self_.right_symbols_) {
       results[output_symbol.name()] = frame[output_symbol];
     }
-  } else {
-    return false;
   }
 
   // put collected values on frame under union symbols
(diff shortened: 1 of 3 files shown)


colinbarry:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Fixes hang or crash when using a `LIMIT` on both sides of a `UNION` or `UNION ALL` query. [#4065](https://github.com/memgraph/memgraph/pull/4065)

src/query/plan/operator.cpp
```diff
@@ -11181,6 +11181,7 @@ bool Limit::LimitCursor::Pull(Frame &frame, ExecutionContext &context) {
   }
 
   // check we have not exceeded the limit before pulling
+  if (!shared_quota_) return false;
```
andrejtonev: The first pull initializes the shared quota.
Subsequent pulls reset the quota, but only when returning false, so no subsequent pulls should happen. What is happening exactly that this fixes?

This is different from Skip where Skip continues pulling after the quota has been exhausted.

tests/gql_behave/tests/memgraph_V1/features/union.feature
```diff
@@ -134,3 +134,34 @@ Feature: Union
...
+
+    Scenario: Union all with limit on each branch
+        Given an empty graph
+        When executing query:
+            """
+            RETURN 1 AS x LIMIT 1 UNION ALL RETURN 2 AS x LIMIT 1
+            """
+        Then the result should be:
+            | x |
+            | 1 |
```
Josipmrden, line 166: maybe add one with limit, other without limit, and vice versa?

src/query/plan/operator.cpp
colinbarry: Fixed in `UnionCursor` instead

tests/gql_behave/tests/memgraph_V1/features/union.feature
colinbarry: Done

andrejtonev (approved):

PR #4066, opened 2026-04-23
feat: Ignore --query-modules-directory on coordinators

as51340:
The flag --query-modules-directory is ignored on coordinators. It is not completely removed because our packaging flags set it by default so every user would need to explicitly disable it when starting coordinators.

### DIFF
--- a/src/flags/run_time_configurable.cpp
+++ b/src/flags/run_time_configurable.cpp
@@ -461,47 +461,46 @@ void Initialize(utils::Settings &settings) {
    * Register periodic snapshot setting. In the case both flags are defined, --storage-snapshot-interval flag will be
    * used. Ideally, we rely on just a single flag but --storage-snapshot-interval-sec is for community,
    * --storage-snapshot-interval for enterprise.
+   *
+   * Coordinators don't support snapshots, so the setting is not registered on them. This also avoids validating
+   * a non-empty default value that the coordinator-side validator would otherwise reject.
    */
-  if (FLAGS_storage_snapshot_interval_sec != 0) {
-    if (FLAGS_storage_snapshot_interval.empty()) {
-      FLAGS_storage_snapshot_interval = std::to_string(FLAGS_storage_snapshot_interval_sec);
-    } else {
-      spdlog::warn(
-          "Periodic snapshot schedule defined via both --storage-snapshot-interval-sec and "
-          "--storage-snapshot-interval. Memgraph will use the configuration flag from --storage-snapshot-interval!");
+  if (!memgraph::flags::CoordinationSetupInstance().IsCoordinator()) {
+    if (FLAGS_storage_snapshot_interval_sec != 0) {
+      if (FLAGS_storage_snapshot_interval.empty()) {
+        FLAGS_storage_snapshot_interval = std::to_string(FLAGS_storage_snapshot_interval_sec);
+      } else {
+        spdlog::warn(
+            "Periodic snapshot schedule defined via both --storage-snapshot-interval-sec and "
+            "--storage-snapshot-interval. Memgraph will use the configuration flag from --storage-snapshot-interval!");
+      }
     }
+
+    // FATAL validation at startup; can't be part of the flag defintion, since we need to check for license
+    ValidPeriodicSnapshot<true>(FLAGS_storage_snapshot_interval);
+    register_flag(
+        kSnapshotPeriodicGFlagsKey,
+        kSnapshotPeriodicSettingKey,
+        !kRestore,
+        [](std::string_view val) {
+          try {
+            const auto period = ValidPeriod(val);
+            snapshot_periodic_.Modify(std::chrono::seconds{period});
+          } catch (const std::invalid_argument & /* unused */) {
+            // String is not a period; pass in as a cron expression
+            // Expression is guaranteed to be valid
+            snapshot_periodic_.Modify(std::string{val});
+          }
+        },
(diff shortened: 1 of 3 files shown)


src/memgraph.cpp
```diff
@@ -818,12 +828,12 @@ int main(int argc, char **argv) {
   auto &interpreter_context_ = memgraph::query::InterpreterContextHolder::GetInstance();
   if (!is_coordinator_instance) {
```
andrejtonev, line 828: do we want to disable the flag instead of just disabling the load?

andrejtonev: maybe at this point it makes sense to split the main() into two implementations: data instance and coordinator

src/flags/run_time_configurable.cpp
```diff
@@ -461,47 +461,46 @@ void Initialize(utils::Settings &settings) {
...
+   * a non-empty default value that the coordinator-side validator would otherwise reject.
    */
-  if (FLAGS_storage_snapshot_interval_sec != 0) {
-    if (FLAGS_storage_snapshot_interval.empty()) {
-      FLAGS_storage_snapshot_interval = std::to_string(FLAGS_storage_snapshot_interval_sec);
-    } else {
-      spdlog::warn(
-          "Periodic snapshot schedule defined via both --storage-snapshot-interval-sec and "
-          "--storage-snapshot-interval. Memgraph will use the configuration flag from --storage-snapshot-interval!");
+  if (!memgraph::flags::CoordinationSetupInstance().IsCoordinator()) {
```
andrejtonev, line 468: what about expanding the register flag to take in an enable bool that can be is_coord
or add a list of coord flags and then pass in the search results

andrejtonev: this also changes the logic from "flag not implemented on coord" to "no flag"
is that what we want?

src/memgraph.cpp
as51340: the problem with disabling it is that config/flags.yaml which is our packaging file has a default set to /usr/lib/memgraph/query_modules so all users would have to explicitly disable this flag. I find it more intuitive for users to just not do anything.

as51340: For splitting the main() into 2 implementations: maybe but not now

src/flags/run_time_configurable.cpp
as51340: I find it better to not call at all register_flag instead of calling it with boolean value.

And yes, changing flag not implemented on coord to no flag is what I wanted to achieve.

src/memgraph.cpp
```diff
@@ -297,57 +297,73 @@ int main(int argc, char **argv) {
...
-      if (maybe_exc) {
-        spdlog::error(memgraph::utils::MessageWithLink(
-            "Unable to load support for embedded Python: {}.", *maybe_exc, "https://memgr.ph/python"));
-      } else {
-        // Change how we load dynamic libraries on Python by using RTLD_NOW flag.
-        // This solves an issue with using the wrong version of libstdc++.
+#ifdef MG_ENTERPRISE
+  memgraph::flags::SetFinalCoordinationSetup();
+  auto const &coordination_setup = memgraph::flags::CoordinationSetupInstance();
+  bool const is_coordinator_instance = coordination_setup.management_port && coordination_setup.coordinator_port &&
```
andrejtonev, line 303: can we define a helper function that encapsulates this logic and is used everywhere?

src/memgraph.cpp
```diff
@@ -1067,16 +1076,19 @@ int main(int argc, char **argv) {
...
+  if (!is_coordinator_instance) {
+    try {
+      memgraph::query::procedure::gModuleRegistry.UnloadAllModules();
+    } catch (memgraph::query::QueryException &) {
+      spdlog::warn("Failed to unload query modules while shutting down.");
+    }
+    python_gc_scheduler->Stop();
+    PyEval_RestoreThread(python_thread_state);
+    // Shutdown Python
+    Py_Finalize();
```
andrejtonev: don't we now skip py finalize?

src/memgraph.cpp
```diff
@@ -1067,16 +1076,19 @@ int main(int argc, char **argv) {
...
-    spdlog::warn("Failed to unload query modules while shutting down.");
+
+  if (!is_coordinator_instance) {
+    try {
+      memgraph::query::procedure::gModuleRegistry.UnloadAllModules();
+    } catch (memgraph::query::QueryException &) {
+      spdlog::warn("Failed to unload query modules while shutting down.");
+    }
+    python_gc_scheduler->Stop();
+    PyEval_RestoreThread(python_thread_state);
```
andrejtonev, line 1092: add MG_ASSERT(python_thread_state); before so we know the logic is sound

andrejtonev (approved):

src/memgraph.cpp
as51340: will be, not merged yet

src/memgraph.cpp
as51340: where everywhere? I think it should be defined here and reused everywhere else

as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - The flag --query-modules-directory cannot be set on coordinators anymore because query modules are supported only on data instances. [#4066](https://github.com/memgraph/memgraph/pull/4066)
- [x] **[ Documentation PR link ](https://github.com/memgraph/documentation/pull/1636)**

PR #4071, opened 2026-04-23
fix: FREE MEMORY now reclaims skiplist nodes and stale index entries

Ignition:
SkipListGc had an off-by-one: nodes were tagged with the next-to-allocate accessor id but erased only when tag < largest-released-id, needing two extra accessor cycles past the releasing one. FREE MEMORY left hundreds of MiB per 1M churn pinned. Fixed by stamping with counter - 1 and tracking live_horizon (exclusive upper bound of the dead prefix).

IN_MEMORY_ANALYTICAL's label+property index accumulated stale entries on every SET because CollectGarbage's cleanup sweep is gated on vertex deletions. Thread old_value through UpdateOnSetProperty and, in analytical mode, remove the stale entry for this vertex before inserting the new one.


Ignition:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - FREE MEMORY now frees significantly more memory after heavy update/delete workloads. Also in IN_MEMORY_ANALYTICAL mode where repeated SETs on indexed properties previously grew memory mode was changed to IN_MEMORY_TRANSACTIONAL, these are now handled while still in IN_MEMORY_ANALYTICAL. [#4071](https://github.com/memgraph/memgraph/pull/4071)

src/utils/skip_list.hpp
```diff
@@ -357,7 +349,9 @@ class SkipListGc final {
     auto guard = std::unique_lock{lock_, std::defer_lock};
     if (!guard.try_lock()) return;
     Block *tail = tail_.load(std::memory_order_acquire);
-    uint64_t last_dead = 0;
+    // Exclusive upper bound of the dead prefix: nodes with tag < live_horizon
+    // are safe to free (their owning accessor has been released).
```
imilinovic: maybe inclusive upper bound on alive ids is better since name is already live_horizon

src/storage/v2/inmemory/label_property_index.cpp
```diff
@@ -604,8 +611,28 @@ void InMemoryLabelPropertyIndex::ActiveIndices::UpdateOnSetProperty(PropertyId p
     for (auto &lookup : it->second | rv::filter(relevant_index)) {
       auto &[property_ids, index] = lookup.second;
       auto values = index->permutations_helper.Extract(vertex->properties);
-      if (r::any_of(values, [](auto &&value) { return !value.IsNull(); })) {
-        auto acc = index->skiplist.access();
+
+      auto acc = index->skiplist.access();
+
+      if (analytical) [[unlikely]] {
+        auto old_values = values;
```
imilinovic: since this is hot path maybe create accessor inside ifs (conditionally) instead of always

src/storage/v2/inmemory/label_property_index.cpp
```diff
@@ -604,8 +611,28 @@ void InMemoryLabelPropertyIndex::ActiveIndices::UpdateOnSetProperty(PropertyId p
     for (auto &lookup : it->second | rv::filter(relevant_index)) {
       auto &[property_ids, index] = lookup.second;
       auto values = index->permutations_helper.Extract(vertex->properties);
-      if (r::any_of(values, [](auto &&value) { return !value.IsNull(); })) {
-        auto acc = index->skiplist.access();
+
+      auto acc = index->skiplist.access();
+
+      if (analytical) [[unlikely]] {
```
imilinovic, line 673: I think this fix is needed for `UpdateOnRemoveLabel` as well

imilinovic (changes requested):

src/storage/v2/inmemory/label_property_index.cpp
imilinovic: since then you'd also need to remove it from skiplist

src/utils/skip_list.hpp
Ignition: I updated the comment

src/storage/v2/inmemory/label_property_index.cpp
Ignition: Added code for UpdateOnRemoveLabel

imilinovic (approved):

PR #4077, opened 2026-04-24
Fix: INT64_MIN integer literal fails to parse

### LINKED ISSUE #2089: Off-By-One in MIN_INT
When running the following query:
```cypher
RETURN -9223372036854775808
```

I encountered the following error:
```
Neo4jError: Memgraph.ClientError.MemgraphError.MemgraphError (Integer literal exceeds 64 bits.)
```

Since `MIN_INT` is `-9223372036854775808`, I believe the query mentioned above is semantically and syntactically correct and thus no error should be thrown here.

This seems like an off-by-one error, as `RETURN -9223372036854775807` runs successfully.

I encountered this issue when testing queries against a locally compiled instance of the memgraph master branch in a Docker container running **debian bullseye**.

### Steps to reproduce
Run the following query and observe it throws an error:
```cypher
RETURN -9223372036854775808
```

### Expected behavior
The query should run successfully

### Actual behavior
The query fails with the error message `Neo4jError: Memgraph.ClientError.MemgraphError.MemgraphError (Integer literal exceeds 64 bits.)`

DavIvek:
Closes #2089

## Root cause
The Cypher grammar treats a leading `-` as a unary operator separate from the integer literal, so `RETURN -9223372036854775808;` is evaluated as `unary_minus(9223372036854775808)`. `ParseIntegerLiteral` then tries to fit the magnitude `9223372036854775808` (== 2^63) into `int64_t` via `std::stoll`, which overflows and raises `Integer literal exceeds 64 bits`, even though the intended value `INT64_MIN` is representable.

## Fix
`visitExpression4` now detects a net-negative unary-minus chain applied to a bare integer literal whose magnitude equals 2^63 (in decimal, octal, or hex form) and emits `PrimitiveLiteral(INT64_MIN)` directly, bypassing the normal unary-minus wrapping and the unsigned-range check.

Cases that still error intentionally:
- `- -9223372036854775808` (even number of minuses → positive 2^63, overflow)
- `-(9223372036854775808)` (parenthesized expression is not a bare integer literal)
- `-9223372036854775808 + 0` etc. — the literal is evaluated inside its own `expression4`, so the special case still fires and the outer operator sees `INT64_MIN`.

## Test
Added behave scenarios covering the fix:
- `RETURN -9223372036854775808` returns `INT64_MIN`
- `RETURN - -9223372036854775808` still errors
- `RETURN 9223372036854775807` (INT64_MAX) still works

See `tests/gql_behave/tests/memgraph_V1/features/min_int_literal.feature`.

## Verification notes
The local build tree in this environment was a stale hard-link of a prior build (built against an older commit, missing `usearch_DIR` / `nuraft_DIR` resolution and the new `memgraph.storage.property_value` C++20 module), so `cmake --build` could not be re-run without a full re-provisioning. The behave test was exercised against the pre-fix binary and reproduced the reported failure (`Integer literal exceeds 64 bits.`). Full CI should validate the fix on a clean build.

### DIFF
--- a/src/query/frontend/ast/cypher_main_visitor.cpp
+++ b/src/query/frontend/ast/cypher_main_visitor.cpp
@@ -11,12 +11,16 @@
 
 #include <algorithm>
 #include <any>
+#include <charconv>
 #include <cstring>
 #include <functional>
 #include <iterator>
+#include <limits>
+#include <optional>
 #include <range/v3/all.hpp>
 #include <ranges>
 #include <string>
+#include <system_error>
 #include <tuple>
 #include <type_traits>
 #include <unordered_map>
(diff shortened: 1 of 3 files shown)


DavIvek:
## Verification update — fix is incomplete

After getting a working build of this branch, the new behave test in this PR still fails on master+this fix:
```
RETURN -9223372036854775808 AS x;
-- Error: Integer literal exceeds 64 bits.
```

**Root cause of the gap:** `ParseIntegerLiteral` is called from two places, and this PR only handles one:
- `cypher_main_visitor.cpp:3771` (`visitIntegerLiteral`) — the bypass in `visitExpression4` works here ✓
- `src/query/frontend/stripped.cpp:147` — the query stripper runs **before** the visitor and calls `ParseIntegerLiteral` directly. It throws "Integer literal exceeds 64 bits" before `visitExpression4` ever runs.

The fix needs to also handle stripping. Options:
1. Make the stripper look at the preceding token; if it's `-`, allow the magnitude `9223372036854775808` and combine
2. Have `ParseIntegerLiteral` return a sentinel (e.g. INT64_MIN) for the magnitude `9223372036854775808` instead of throwing, and let callers either accept or reject

The test added here is a good regression test — it's failing as expected. The visitor-side fix is also correct as far as it goes; it just needs a sibling change to the stripper.

Marking as draft — needs follow-up work before merge.

DavIvek:
## Update — verified end-to-end

Pushed a follow-up commit (`58f3390ca`) that closes the gap I noted earlier. `ParseIntegerLiteral` now returns `INT64_MIN` for the magnitude `9223372036854775808` rather than throwing, so the query stripper accepts it. The visitor's `visitExpression4` special case still binds the literal to the leading minus as `PrimitiveLiteral(INT64_MIN)`, so the runtime never computes `-INT64_MIN` (which is UB in C++).

**Verified:**
```
RETURN -9223372036854775808;        -- INT64_MIN ✓
RETURN -9223372036854775807;        -- INT64_MIN+1 ✓
RETURN  9223372036854775807;        -- INT64_MAX ✓
RETURN -9223372036854775809;        -- error: "Integer literal exceeds 64 bits." ✓
```

**Behave tests (locally on a fresh build of this branch):**
- `min_int_literal.feature` — 3/3 scenarios pass
- `expressions.feature` — 13/13 scenarios pass (no regressions)
- `functions.feature` — 104/104 scenarios pass (no regressions)

**Behavior note:** the bare degenerate query `RETURN 9223372036854775808` now returns `INT64_MIN` instead of erroring. Previously this case errored; the new behavior is at worst harmless and consistent with how `-INT64_MIN` itself wraps. If we want to keep that case as an error for backwards compatibility, we'd need to add a sign-aware check in the stripper — happy to do that if reviewers prefer.

DavIvek:
I don't really like this approach, will reopen if needed.

PR #4095, opened 2026-04-30
fix: Ensure jemalloc TLS state is set before enforcing allocation limits

andrejtonev:
jemalloc sets up its TLS environment lazily (on first allocation for the thread)
This setup can fail if allocation limits are enforced, leading to a broken state that jemalloc does not handle.
We need to ensure the TLS setup by forcing an allocation on each thread. This is handled inside the malloc interface via EnsureJemallocThreadStateInitialized().

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,9 @@ CLAUDE.md
 .claude/
 CLAUDE.md
 
+# OpenCode
+.opencode/
+
 # Conan files
 conanbuild.sh
 conanbuildenv*
--- a/src/memgraph.cpp
+++ b/src/memgraph.cpp
@@ -94,7 +94,7 @@ constexpr const char *kMgExperimentalEnabled = "MEMGRAPH_EXPERIMENTAL_ENABLED";
 constexpr const char *kMgBoltPort = "MEMGRAPH_BOLT_PORT";
 constexpr const char *kMgHaClusterInitQueries = "MEMGRAPH_HA_CLUSTER_INIT_QUERIES";
 
-constexpr uint64_t kMgVmMaxMapCount = 262'144;
+constexpr uint64_t kMgVmMaxMapCount = 524'288;
 
 void WarnDeprecatedFlags() {
   auto warn_if_set = [](std::string_view name, std::string_view message) {
--- a/src/memory/malloc_free.cpp
+++ b/src/memory/malloc_free.cpp
@@ -22,7 +22,7 @@
 
 namespace {
 inline auto alloc_tracking(size_t size, int flags = 0) -> bool {
-  if (memgraph::memory::IsQueryTracked()) [[unlikely]] {
+  if (size > 0 && memgraph::memory::IsQueryTracked()) [[unlikely]] {
     auto actual_size = je_nallocx(size, flags);
     return memgraph::memory::TrackAllocOnCurrentThread(actual_size);
   }
@@ -30,7 +30,7 @@ inline auto alloc_tracking(size_t size, int flags = 0) -> bool {
 }
 
 inline void failed_alloc_tracking(size_t size, int flags = 0) {
-  if (memgraph::memory::IsQueryTracked()) [[unlikely]] {
+  if (size > 0 && memgraph::memory::IsQueryTracked()) [[unlikely]] {
     const auto actual_size = je_nallocx(size, flags);
     memgraph::memory::TrackFreeOnCurrentThread(actual_size);
   }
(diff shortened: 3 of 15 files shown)


andrejtonev:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - fix: jemalloc TLS state is lazily setup on first allocation. This can silently fail in case we are at the memory limit and cause UB or segafult. First allocations now trigger a dummy allocation that sets up the TLS state outside our memory limiting logic, ensuring the state is correctly setup. [#4095](https://github.com/memgraph/memgraph/pull/4095)

.gitignore
```diff
@@ -6,6 +6,9 @@ CLAUDE.md
 .claude/
 CLAUDE.md
 
+# OpenCode
```
colinbarry, line 9: Unrelated .gitignore change?

src/memory/malloc_free.cpp
```diff
@@ -103,6 +110,8 @@ extern "C" void *calloc(size_t count, size_t size) {
 }
 
 extern "C" void *realloc(void *ptr, size_t size) {
+  // Warm up jemalloc thread state before any tracking helper can trigger OOM handling.
```
colinbarry, line 113: LTO will likely inline these `EnsureJemallocThreadStateInitialized` calls, but maybe if the cross-translation unit function calls on the hot path are noticeable in performance monitoring, we could move the check locally:

```cpp
// malloc_free.cpp
 namespace {                                 
  constinit thread_local bool jemalloc_tls_ready [[gnu::tls_model("initial-exec")]]{false;}
                                            
  inline void EnsureJemallocReady() {
    if (!jemalloc_tls_ready) [[unlikely]] {
      memgraph::memory::EnsureJemallocThreadStateInitialized();
      jemalloc_tls_ready = true;
    }
  }
  }  // namespace 
```

colinbarry (approved):
Approved. I noted an accidental .gitignore change, and added a comment about performance of cross-TU function calls on a hot-path - deal with it later if it is an issue - but good to merge as is. 💪

src/memory/malloc_free.cpp
andrejtonev: noted, but will leave as is currently.

.gitignore
andrejtonev: just gonna sneak that in 🫢

release/package/mgbuild.sh
```diff
@@ -393,6 +393,23 @@ setup_host_cache_permissions() {
   fi
 }
```
mattkjames7, line 395: This won't work in CI because `sudo` cannot be used by the actions runner user. If this increase is needed then we will need to run an ansible playbook over all of the CI machines to update it. We should remove this function and open an issue in `memgraph/infra`.

src/memgraph.cpp
```diff
@@ -94,7 +94,7 @@ constexpr const char *kMgExperimentalEnabled = "MEMGRAPH_EXPERIMENTAL_ENABLED";
 constexpr const char *kMgBoltPort = "MEMGRAPH_BOLT_PORT";
 constexpr const char *kMgHaClusterInitQueries = "MEMGRAPH_HA_CLUSTER_INIT_QUERIES";
 
-constexpr uint64_t kMgVmMaxMapCount = 262'144;
```
mattkjames7, line 97: I'm not sure what this is used for - is this the new minimum that should be used? If so, we will need to update this page: https://memgraph.com/docs/database-management/system-configuration

andrejtonev: that specific var is just used for some warning as the lowest sensible count.
@gitbuda how did we generate the table matt is talking about? I think this will need to be a matrix with ram size and number of tenants.

release/package/mgbuild.sh
andrejtonev: yeah, I though so :D 
will do

src/memory/malloc_free.cpp
```diff
@@ -17,26 +17,27 @@
...
-  if (memgraph::memory::IsQueryTracked()) [[unlikely]] {
+  if (size > 0 && memgraph::memory::IsQueryTracked()) [[unlikely]] {
     const auto actual_size = je_nallocx(size, flags);
     memgraph::memory::TrackFreeOnCurrentThread(actual_size);
   }
 }
 
 inline void free_tracking(void *ptr, int flags = 0) {
-  if (memgraph::memory::IsQueryTracked()) [[unlikely]] {
+  if (ptr != nullptr && memgraph::memory::IsQueryTracked()) [[unlikely]] {
```
colinbarry: We already check for `ptr != nullptr`at every call site of `free_tracking`, so this is dead code.

colinbarry (approved):
Approved (again).

PR #4126, opened 2026-05-07
perf: Fix slow delta chain walk in garbage collection

colinbarry:
Optimises the case of severing the `next` link when unlinking a non-sequential delta in `CollectGarbage`. Consider the case of `TxA`'s deltas being unlinked, and one of `TxA`'s delta to be downstream from an inactive non-sequential delta from `TxB`, i.e.:

```
(vertex) ->...->[TxB's delta]->[TxA's delta]->...
```

#4020 fixed a correctness issue by ensuring that TxB's `next` was correctly set to `nullptr`, but it did so in a way that required both an `O(n)` walk up the chain to locate the vertex, and for the vertex's lock to be acquired. Correct, but regressive from a performance standpoint.

This approach instead just sets `TxB`'s delta's `.next` to `nullptr` atomically, bypassing the `O(n)` walk and the lock. Doing so is safe because:
- There can be no concurrent modification of `TxB's` `delta->next`. Prepends only happen at the chain head, garbage collection is serialized by `gc_lock_`, and the abort path only operates on active deltas. We know both the deltas in this operation are inactive because they cannot escape `waiting_gc_deltas_` unless all downstream contributors are committed, so we can be the only thing modifying `next`.
- No concurrent reader needs to read anything beyond `TxB`'s delta, as any deltas that would be needed are protected by the `oldest_active_start_timestamp` horizon.

Regession discovered by stress test #4115.

### DIFF
--- a/src/storage/v2/inmemory/storage.cpp
+++ b/src/storage/v2/inmemory/storage.cpp
@@ -3073,17 +3073,24 @@ void InMemoryStorage::CollectGarbage(std::unique_lock<utils::ResourceLock> main_
             }
 
             if (prev.delta->commit_info->timestamp.load() < oldest_active_start_timestamp) {
-              // For a committed non-sequential predecessor, readers skip delta
-              // via next, so we must clear delta->next before freeing
-              // downstream delta to stop traversal into freed deltas.
-              if (!IsDeltaNonSequential(*prev.delta)) {
-                break;
+              if (IsDeltaNonSequential(*prev.delta)) {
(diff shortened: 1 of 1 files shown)


colinbarry:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Improve performance of garbage collector after concurrent import of datasets with significant number of edges on one or more supernodes. [#4126](https://github.com/memgraph/memgraph/pull/4126)

Ignition (approved):

PR #4140, opened 2026-05-13
feat: Intra-cluster TLS communication

as51340:
Adds intra-cluster TLS in HA deployment. New flags are added --cluster-cert-file, --cluster-key-file and --cluster-ca-file. The communication is now secure on all management server, replication and coordinator server. Bolt server independent from this, was delivered previously. SSL_connect works with a non-blocking connect so the timeout could be used.

Reloading certificates functionality will be added later.


as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - By adding flags --cluster-key-file, --cluster-cert-file and --cluster-ca-file, users can enable secure TLS communication between instances in the HA cluster. All 3 flags needs to be configured in order to consider mTLS well configured. [#4140](https://github.com/memgraph/memgraph/pull/4140)
- [x] **[ Documentation PR link ](https://github.com/memgraph/documentation/pull/1647)**

src/communication/context.cpp
```diff
@@ -49,6 +49,21 @@ ClientContext::ClientContext(const std::string &key_file, const std::string &cer
   }
 }
 
+ClientContext::ClientContext(const std::string &key_file, const std::string &cert_file, const std::string &ca_file)
+    : ClientContext(key_file, cert_file) {
+  if (!ca_file.empty()) {
+    MG_ASSERT(SSL_CTX_use_certificate_file(ctx_, cert_file.c_str(), SSL_FILETYPE_PEM) == 1,
```
andrejtonev: the key and cert files should have been handled by the : ClientContext(key_file, cert_file)

src/communication/context.cpp
```diff
@@ -49,6 +49,21 @@ ClientContext::ClientContext(const std::string &key_file, const std::string &cer
...
+              "Couldn't load client certificate from file: {}",
+              cert_file);
+    MG_ASSERT(SSL_CTX_use_PrivateKey_file(ctx_, key_file.c_str(), SSL_FILETYPE_PEM) == 1,
+              "Couldn't load client private key from file: ",
```
andrejtonev: missing {} for the filename

src/communication/context.cpp
andrejtonev: Missing {} for the filename

src/communication/client.cpp
```diff
@@ -74,19 +66,78 @@ bool Client::Connect(const io::network::Endpoint &endpoint) {
...
+    }
+
+    if (!handshake_ok) {
+      socket_.Close();
+      return false;
+    }
+
+    // This is safe to do because we only ever modify blocking and non-blocking mode. Be careful if you mess with
+    // O_APPEND, O_ASYNC, O_DIRECT or O_NOATIME
+    if (fcntl(fd, F_SETFL, 0) < 0) {
```
andrejtonev: Lets be explicit and safe
```suggestion
int flags = fcntl(fd, F_GETFL, 0);
if (flags == -1) { /* handle error */ }
flags &= ~O_NONBLOCK;  // Clear only O_NONBLOCK
if (fcntl(fd, F_SETFL, flags) < 0) { /* handle error */ }
```

src/flags/general.cpp
```diff
@@ -225,3 +228,34 @@ DEFINE_string(query_callable_mappings_path, "",
...
+
+auto memgraph::flags::IsIntraClusterTLSEnabled() -> bool {
+  return !FLAGS_cluster_cert_file.empty() && !FLAGS_cluster_key_file.empty() && !FLAGS_cluster_ca_file.empty();
+}
+
+void memgraph::flags::ValidateIntraClusterTLSFlags() {
+  const int set_count = static_cast<int>(!FLAGS_cluster_cert_file.empty()) +
+                        static_cast<int>(!FLAGS_cluster_key_file.empty()) +
+                        static_cast<int>(!FLAGS_cluster_ca_file.empty());
+  if (set_count != 0 && set_count != 3) {
```
andrejtonev, line 270: Lets expand the validation logic:
1. do files exist
2. are they the correct format

src/coordination/data_instance_management_server_handlers.cpp
```diff
@@ -191,9 +193,15 @@ auto DataInstanceManagementServerHandlers::DoRegisterReplica(replication::Replic
                                                              coordination::ReplicationClientInfo const &config)
     -> bool {
   auto const converter = [&config](const auto &repl_info_config) {
+    std::optional<utils::TlsConfig> maybe_ssl;
+    if (flags::IsIntraClusterTLSEnabled()) {
+      maybe_ssl.emplace(FLAGS_cluster_key_file, FLAGS_cluster_cert_file, FLAGS_cluster_ca_file);
```
andrejtonev: The pattern maybe_ssl.emplace(FLAGS_cluster_key_file, FLAGS_cluster_cert_file, FLAGS_cluster_ca_file) is repeated at: 

1 | data_instance_management_server_handlers.cpp | 198 | DoRegisterReplica lambda
2 | data_instance_management_server_handlers.cpp | 287 | DemoteMainToReplicaHandler
3 | query/interpreter.cpp | 451 | ReplQueryHandler::RegisterReplica
4 | query/interpreter.cpp | 491 | ReplQueryHandler::SetReplicationRole
5 | memgraph.cpp | 702 | main() HA setup
6 | replication/state.cpp | 202 | FetchReplicationData

Refactoring opportunity: Create utils::TlsConfig FromClusterFlags() helper in flags/general.hpp.

src/replication/status.cpp
```diff
@@ -107,25 +107,16 @@ void to_json(nlohmann::json &j, const ReplicationReplicaEntry &p) {
                                {kCheckFrequency, p.config.replica_check_frequency.count()}};
 
   common[kReplicaServer] = p.config.repl_server_endpoint;  // non-resolved
+                                                           // Serialize always to nullptr for backward compatiblity
```
andrejtonev: compatibility

src/coordination/coordinator_instance.cpp
andrejtonev: empty scope

src/utils/tls.hpp
```diff
@@ -0,0 +1,26 @@
...
+#include <string>
+
+namespace memgraph::utils {
+
+struct TlsConfig {
+  std::string key_file;
+  std::string cert_file;
+  std::string ca_file;
+
+  friend bool operator==(TlsConfig const &, TlsConfig const &) = default;
```
andrejtonev, line 23: these are paths, right? can we use paths instead?

src/flags/general.cpp
as51340: I think this is not needed because in the initialisation we call ClientContext's constructor and do MG_ASSERT on SSL_CTX_use_certificate_file and SSL_CTX_use_PrivateKey_file. So if something is wrong, this won't pass and we get same behavior as if I do the explicit check in memgraph.cpp. It would just be a duplicated logic.

src/utils/tls.hpp
as51340: we are passing all around strings in SSL code so I think it doesn't make sense to change to paths

src/communication/client.cpp
```diff
@@ -74,19 +66,84 @@ bool Client::Connect(const io::network::Endpoint &endpoint) {
...
+      socket_.Close();
+      return false;
+    }
+    flags &= ~O_NONBLOCK;
+    if (fcntl(fd, F_SETFL, flags) < 0) {
+      spdlog::error("Failed to restore socket to blocking mode after SSL handshake");
       socket_.Close();
       return false;
     }
+    return true;
```
andrejtonev: this "return true" doesn't seem to be correct. we will miss the setup at the bottom

src/communication/client.cpp
```diff
@@ -74,19 +66,84 @@ bool Client::Connect(const io::network::Endpoint &endpoint) {
...
+      return false;
+    }
 
-    // Perform the TLS handshake.
-    auto ret = SSL_connect(ssl_);
-    if (ret != 1) {
-      SPDLOG_WARN("Couldn't connect to SSL server: {}", SslGetLastError());
+    int flags = fcntl(fd, F_GETFL, 0);
+    if (flags < 0) {
+      spdlog::error("Failed to read flags when restoring socket to non-blocking mode");
```
andrejtonev: I assume the comment is wrong. we are trying to restore blocking mode.

src/communication/client.cpp
```diff
@@ -74,19 +66,84 @@ bool Client::Connect(const io::network::Endpoint &endpoint) {
...
+      if (poll_status == 0) {
+        spdlog::warn("SSL handshake timed out after {}ms", connect_timeout_ms_.count());
+        break;
+      }
+      if (poll_status < 0) {
+        if (errno == EINTR) continue;
+        spdlog::warn("poll() during SSL handshake failed: {}", std::strerror(errno));
+        break;
+      }
+    }
```
andrejtonev: we are handling just <= 0. If >0
events can return POLLERR/POLLHUP and we should handle those

src/communication/client.cpp
```diff
@@ -74,19 +66,84 @@ bool Client::Connect(const io::network::Endpoint &endpoint) {
...
-    // Clear all leftover errors.
     ERR_clear_error();
+    spdlog::trace("Trying to do SSL_connect");
+    // Drive the TLS handshake — socket is already non-blocking because we
+    // asked Socket::Connect to leave it that way.
+    const int fd = socket_.fd();
+    auto const deadline = std::chrono::steady_clock::now() + connect_timeout_ms_;
+    bool handshake_ok = false;
+
+    while (true) {
```
andrejtonev: a bit hard to follow, can we put logics into lambdas/functions or somehow reorder?

src/communication/client.cpp
```diff
@@ -74,19 +66,84 @@ bool Client::Connect(const io::network::Endpoint &endpoint) {
...
+        break;
+      }
+      auto const ms_remaining = std::chrono::duration_cast<std::chrono::milliseconds>(deadline - now).count();
+
+      std::array<pollfd, 1> pfds{{{.fd = fd, .events = events, .revents = 0}}};
+      const int poll_status = poll(pfds.data(), pfds.size(), static_cast<int>(ms_remaining));
+
+      if (poll_status == 0) {
+        spdlog::warn("SSL handshake timed out after {}ms", connect_timeout_ms_.count());
+        break;
```
andrejtonev: do we want to break here or return false?

src/communication/client.cpp
andrejtonev, line 188: are these missing the optional timeouts on purpose?

src/replication/state.cpp
andrejtonev, line 187: are we ignoring the tls config for clients on purpose?

src/communication/client.cpp
```diff
@@ -24,37 +24,29 @@
...
   if (context_->use_ssl()) {
     // Release leftover SSL objects.
     ReleaseSslObjects();
 
     // Create a new SSL object that will be used for SSL communication.
     ssl_ = SSL_new(context_->context());
     if (ssl_ == nullptr) {
-      SPDLOG_ERROR("Couldn't create client SSL object!");
+      spdlog::error("Couldn't create client SSL object!");
       socket_.Close();
```
andrejtonev, line 59: proactively call ReleaseSslObjects on failures

src/io/network/socket.cpp
```diff
@@ -78,7 +74,8 @@ void Socket::Close(int const sfd, std::string_view socket_addr) {
   }
 }
 
-bool Socket::Connect(const Endpoint &endpoint) {
+bool Socket::Connect(const Endpoint &endpoint, std::chrono::milliseconds const connect_timeout_ms,
+                     bool const keep_non_blocking) {
```
andrejtonev, line 77: Not a huge fan of the keep_non_blocking

src/communication/client.cpp
as51340: true, my bad happened after refactoring

src/communication/client.cpp
as51340: yes

src/communication/client.cpp
as51340: it's ok to break here because handshake_ok will be false and then we handle this at line 117 with socket_.Close and return false;

src/communication/client.cpp
as51340: no, I have to handle this and I plan to do in the next PR

src/communication/client.cpp
as51340: if ssl is nullptr then there is nothing to release because ReleaseSslObjects checks if (ssl). I can add this though further down this release of ssl objects

src/io/network/socket.cpp
as51340: suggestion for some other naming? 

The reason why I did it like this is to avoid setting to non blocking, back to blocking, non-blocking again and then blocking again because F_GETFL and F_SETFL are sys calls which could be expensive.

src/replication/state.cpp
as51340: I think you caught good bug

src/communication/client.cpp
as51340: I will refactor it

src/replication/status.cpp
```diff
@@ -107,25 +107,16 @@ void to_json(nlohmann::json &j, const ReplicationReplicaEntry &p) {
                                {kCheckFrequency, p.config.replica_check_frequency.count()}};
 
   common[kReplicaServer] = p.config.repl_server_endpoint;  // non-resolved
+                                                           // Serialize always to nullptr for backward compatibility
+  common[kSSLKeyFile] = nullptr;
+  common[kSSLCertFile] = nullptr;
```
andrejtonev, line 112: why do we need to set to nullptr for backwards compatibility? if we have an older json, we just ignore it, no need for the newer  version to set the value. or do we need to have the older memgraph version handle the newer json?
and do we need to bump the version and add migration logic?

src/io/network/socket.cpp
```diff
@@ -277,6 +274,22 @@ void Socket::SetNonBlocking() {
...
+  if (flags < 0) {
+    spdlog::error("Failed to read flags when setting socket to blocking mode");
+    return false;
+  }
+  flags &= ~O_NONBLOCK;
+  if (fcntl(socket_, F_SETFL, flags) < 0) {
+    spdlog::error("Failed to restore socket to blocking mode after SSL handshake");
+    return false;
+  }
+  return true;
```
andrejtonev, line 290: looks like the other Set functions just MG_ASSERT. Should we handle these errors or just assert?

andrejtonev (approved):
looks good, just some smaller comments

src/io/network/socket.cpp
as51340: SetNonBlocking is uses in session constructor, I think we should be able to handle these errors but I will handle this in the next PR. IMO failure to set blocking or non-blocking aren't critical failures, at least not on the network level.

src/replication/status.cpp
as51340: I want to support this: or do we need to have the older memgraph version handle the newer json?

However, I realised that none of the replication code is actually backwards compatible if I remove some field.
Migration logic is not needed, we use migration logic only for forward compatibility

PR #4149, opened 2026-05-15
feat: unify logs into one stream and add per session log settings

imilinovic:
## Summary

Replaces per-session query-log files with a single main log stream tagged per-session. `SET SESSION TRACE ON|OFF` toggles structured emission for the current session; tagged events emit into the main log at **INFO** with `[session=<uuid>] [user=<u>] [tx=<id>]`, respecting `--log-level`.

Generic `spdlog::*` call sites are untouched — no wrapper, no per-call TLS read on the hot path. The session machinery is one header in `mg-utils`; the logger lifecycle and flags stay in `mg-flags`.

## What changed

**New** `src/utils/session_context.hpp`:
- `SessionLogContext` — per-session tag fields (`session_uuid`, `user`, `tx_id`) and `trace_enabled` toggle.
- `ScopedSessionLog` — thread-local RAII guard installed in `Session::Execute_` per Bolt message.
- `EmitSessionTraceEvent` — gated structured emit at INFO; args evaluate only when the gate passes.
- `IsSessionTraceEnabled` — call-site predicate for gating expensive arg construction (JSON dumps, plan prints) before `EmitSessionTraceEvent` even sees them.

**Removed** `QueryLogger`, `--query-log-directory`, per-session `{uuid}.log` files, `tests/e2e/query_log/`.

Worker threads spawned by a handler (parallel index creation, replication, GC, NuRaft) do NOT inherit the guard — trace coverage is bounded to the bolt dispatcher thread.

## Breaking changes

1. **`--query-log-directory` removed.** Startup rejects it; drop from configs.
2. **No per-session log files.** Tooling parsing `{uuid}.log` must switch to `grep '\[session=<uuid>\]' memgraph.log`.
3. **`SET SESSION TRACE` destination + level changed.** Same syntax — events now interleave into the main log at INFO with session tags, instead of a per-session file at trace level. Visible when `--log-level` is `info` or lower.


imilinovic:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Session trace now emits tagged `[session=<uuid>]` events into the main Memgraph log at INFO (when `--log-level` is `info` or lower) instead of a per-session `{uuid}.log` file, and the `--query-log-directory` flag is removed; drop it from configs and filter the main log by session UUID. [#4149](https://github.com/memgraph/memgraph/pull/4149)
- [x] **[ Documentation PR link memgraph/documentation#1661 ]**

src/query/interpreter.cpp
```diff
@@ -3492,20 +3484,20 @@ PreparedQuery PrepareCypherQuery(ParsedQuery parsed_query, std::map<std::string,
   auto hints = plan::ProvidePlanHints(&plan->plan(), plan->symbol_table());
   for (const auto &hint : hints) {
     notifications->emplace_back(SeverityLevel::INFO, NotificationCode::PLAN_HINTING, hint);
-    interpreter.LogQueryMessage(hint);
+    memgraph::logging::EmitSessionTraceEvent("{}", hint);
   }
 
-  if (interpreter.IsQueryLoggingActive()) {
+  if (memgraph::logging::IsSessionTraceEnabled()) {
```
as51340, line 3514: you don't need this if check because EmitSessionTraceEvent will just skip this logging if session logging is not enabled

src/query/interpreter.cpp
```diff
@@ -3610,12 +3601,15 @@ PreparedQuery PrepareExplainQuery(ParsedQuery parsed_query, std::vector<Notifica
...
-    interpreter.LogQueryMessage(hint);
+    memgraph::logging::EmitSessionTraceEvent("{}", hint);
   }
 
   std::stringstream printed_plan;
   plan::PrettyPrint(*dba, &cypher_query_plan->plan(), &printed_plan);
-  interpreter.LogQueryMessage(fmt::format("Explain plan:\n{}", printed_plan.str()));
+  // PrettyPrint is needed for the EXPLAIN result rows below; only the trace emit is gated.
+  if (memgraph::logging::IsSessionTraceEnabled()) {
+    memgraph::logging::EmitSessionTraceEvent("Explain plan:\n{}", printed_plan.str());
```
as51340, line 3635: same as above

src/utils/session_context.hpp
```diff
@@ -0,0 +1,108 @@
...
+template <typename... Args>
+void EmitSessionTraceEvent(fmt::format_string<Args...> fmt_str, Args &&...args) {
+  auto *ctx = detail::ActiveTraceContext();
+  if (ctx == nullptr) return;
+  auto &buf = detail::log_assembly_buf;
+  buf.clear();
+  ctx->AppendTraceTags(buf);
+  buf.push_back(' ');
+  fmt::format_to(std::back_inserter(buf), fmt_str, std::forward<Args>(args)...);
+  spdlog::default_logger_raw()->log(spdlog::level::info, std::string_view{buf.data(), buf.size()});
```
as51340: this can be simplified to spdlog::info(...)

src/query/interpreter.cpp
```diff
@@ -8122,22 +8114,18 @@ PreparedQuery PrepareSessionTraceQuery(ParsedQuery parsed_query, CurrentDB &curr
...
-      interpreter->LogQueryMessage("Session initialized!");
-    } else {
-      interpreter->query_logger_.reset();
+    interpreter->GetLogContext()->SetTrace(session_trace_enabled);
+    if (session_trace_enabled) {
+      // Trace events emit at INFO — warn if --log-level filters them out.
+      if (!spdlog::should_log(spdlog::level::info)) {
+        spdlog::warn(
+            "SET SESSION TRACE ON: trace events emit at INFO, but --log-level is filtering INFO out. "
+            "Lower --log-level to info (or below) to see them.");
```
as51340: this is good warning but point them to use run-time database setting instead of --log-level

src/utils/session_context.hpp
```diff
@@ -0,0 +1,108 @@
...
+
+namespace memgraph::logging {
+
+// Per-session log state. Mutated and read only on the bolt thread; relaxed atomics suffice.
+class SessionLogContext {
+ public:
+  // SET SESSION TRACE ON|OFF toggle. Events emit at INFO.
+  void SetTrace(bool on) noexcept { trace_enabled_.store(on, std::memory_order_relaxed); }
+
+  bool trace_enabled() const noexcept { return trace_enabled_.load(std::memory_order_relaxed); }
```
as51340: If using relaxed is intentional please add a comment why it's enough to use relaxed instead of standard acquire-release semantic

src/query/interpreter.hpp
```diff
@@ -672,11 +671,11 @@ std::map<std::string, TypedValue> Interpreter::Pull(TStream *result_stream, std:
       }
     }
   } catch (const ExplicitTransactionUsageException &e) {
-    LogQueryMessage(e.what());
+    memgraph::logging::EmitSessionTraceEvent("{}", e.what());
```
as51340: I think you can just memgraph::logging::EmitSessionTraceEvent(e.what()); or if not make an overload which will support this

src/query/interpreter.hpp
```diff
@@ -672,11 +671,11 @@ std::map<std::string, TypedValue> Interpreter::Pull(TStream *result_stream, std:
...
     query_execution.reset(nullptr);
     throw;
   } catch (const utils::BasicException &e) {
-    LogQueryMessage(e.what());
+    memgraph::logging::EmitSessionTraceEvent("{}", e.what());
```
as51340: same

src/utils/session_context.hpp
as51340: also, why trace_enabled_ needs to be atomic and user_, tx_id_ not?

src/utils/session_context.hpp
```diff
@@ -0,0 +1,108 @@
...
+ private:
+  std::string session_uuid_;
+  std::string user_;
+  uint64_t tx_id_ = 0;
+  std::atomic<bool> trace_enabled_{false};
+};
+
+namespace detail {
+inline thread_local SessionLogContext *current_session_log = nullptr;
+inline thread_local fmt::memory_buffer log_assembly_buf;  // reused across emits
```
as51340: I think this is fine but maybe unnecessarily complex since fmt::memory_buffer has 500B of space allocated

as51340 (changes requested):

src/query/interpreter.cpp
imilinovic: if we dont have a check then printed_plan will get converted to string even if we dont print, possibly expensive

src/query/interpreter.cpp
imilinovic: if we dont have a check then we allocate the big string even if we dont have session trace on, so t his check was put at places where the allocation is potentially expensive.

src/utils/session_context.hpp
imilinovic: you are right, atomic not needed since this is per session

as51340 (approved):

PR #4152, opened 2026-05-18
Shutdown forces telemetry/license updates to abort

andrejtonev:
Telemetry data requests are aborted, while shutdown request has a lower 10 second timeout.
License requests are simply aborted during shutdown.

### DIFF
--- a/src/requests/requests.cpp
+++ b/src/requests/requests.cpp
@@ -87,11 +87,22 @@ auto DownloadProgressCb(void *clientp, curl_off_t dltotal, curl_off_t dlnow, cur
 
 size_t CurlWriteCallback(char * /*ptr*/, size_t /*size*/, size_t nmemb, void * /*userdata*/) { return nmemb; }
 
+// Progress callback for POST requests - checks per-request abort flag
+auto PostProgressCallback(void *clientp, curl_off_t /*dltotal*/, curl_off_t /*dlnow*/, curl_off_t /*ultotal*/,
+                          curl_off_t /*ulnow*/) -> int {
+  auto const *abort_flag = static_cast<std::atomic<bool> const *>(clientp);
+  if (abort_flag && abort_flag->load(std::memory_order_relaxed)) {
+    return 1;  // Return non-zero to abort transfer
+  }
+  return 0;  // Continue transfer
+}
+
 }  // namespace
 
 void Init() { curl_global_init(CURL_GLOBAL_ALL); }
 
-bool RequestPostJson(const std::string &url, const nlohmann::json &data, int timeout_in_seconds) {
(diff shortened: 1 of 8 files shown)


andrejtonev:

---
___

### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Any Telemetry/License in-progress communication is abort during shutdown. This gives allows for faster shutdowns. [#4152](https://github.com/memgraph/memgraph/pull/4152)

src/telemetry/telemetry.cpp
```diff
@@ -156,7 +166,9 @@ void Telemetry::CollectData(const std::string &event) {
     StoreData(event, data);
   }
   if (num_ % send_every_n_ == 0 || event == "shutdown") {
-    SendData();
+    // Use shorter timeout for shutdown event
+    int timeout = (event == "shutdown") ? 10 : (2 * 60);
```
as51340: extract into common global variable. Hard to know here from where the "shutdown" comes from

src/requests/requests.cpp
```diff
@@ -87,11 +87,22 @@ auto DownloadProgressCb(void *clientp, curl_off_t dltotal, curl_off_t dlnow, cur
...
 size_t CurlWriteCallback(char * /*ptr*/, size_t /*size*/, size_t nmemb, void * /*userdata*/) { return nmemb; }
 
+// Progress callback for POST requests - checks per-request abort flag
+auto PostProgressCallback(void *clientp, curl_off_t /*dltotal*/, curl_off_t /*dlnow*/, curl_off_t /*ultotal*/,
+                          curl_off_t /*ulnow*/) -> int {
+  auto const *abort_flag = static_cast<std::atomic<bool> const *>(clientp);
+  if (abort_flag && abort_flag->load(std::memory_order_relaxed)) {
+    return 1;  // Return non-zero to abort transfer
+  }
+  return 0;  // Continue transfer
```
as51340, line 97: maybe add that this is not checked on every iteration of the loop but rather every 100/1000 times.

as51340 (changes requested):

src/requests/requests.cpp
andrejtonev: Can't do just every 100th time because if there is a problem with the communication it can be slow and this will be triggered every second.
I'll add a timed cool off period. 👍

andrejtonev: just ran some benchmarks and I don't think it makes a difference. the callback is called every couple of ms, so a relaxed atomic should be fine.

andrejtonev: Is any performance critical code going to hit this? RPC?

as51340: I think RPC doesn't hit this, only telemetry

as51340 (approved):

PR #4153, opened 2026-05-18
feat: richer output for schema node/rel type-properties procedures

imilinovic:
## Summary

Enriches the output of `schema.node_type_properties()` and `schema.rel_type_properties()` so consumers get a more complete picture of the graph's shape.

### Output changes

**`rel_type_properties()`** now emits one row per `(relType, sourceNodeLabels, targetNodeLabels, propertyName)` instead of one row per `(relType, propertyName)`. The same edge type connecting different endpoint label sets produces separate rows. New columns:

- `sourceNodeLabels: LIST<STRING>`, `targetNodeLabels: LIST<STRING>` — labels on the start / end node of edges in the partition.
- `propertyObservations: INTEGER` — edges in the partition that carried this property.
- `totalObservations: INTEGER` — edges in the partition that were examined.

**`node_type_properties()`** keeps its `(nodeType, propertyName)` row granularity and gains `propertyObservations` / `totalObservations` with the analogous meaning for nodes.

`propertyTypes` is now declared `LIST<STRING>` on both procedures (was `Any`).

### Implementation notes

- New `RelKey { rel_type, source_labels, target_labels }` keys the rel-properties scan, hashed via `boost::hash_combine` over the sorted label sets.
- Filter lookups now use a transparent `StringSet`, so `includeRels`/`excludeRels`/`includeLabels`/`excludeLabels` take `std::string_view` without per-call allocations; `rel.Type()` is materialised into a `std::string` only after the filter accepts it.
- `RelKey`, `RelKeyHash`, and `LabelsHash` moved into the file's anonymous namespace; dead `LabelsComparator` removed.

## BREAKING

The output schemas of `schema.node_type_properties()` and `schema.rel_type_properties()` are not backwards-compatible:

1. **New columns** — `CALL ... YIELD *` returns extra fields. Code that binds positionally or asserts column counts will break; use explicit `YIELD <names>`.
2. **Row granularity for `rel_type_properties` changed** — a relationship type connecting multiple endpoint-label combinations now produces one row per `(sourceLabels, targetLabels, property)` rather than one per `(relType, property)`. Row counts and per-row `mandatory` / count values can differ.
3. **`propertyTypes` declared type narrowed** from `Any` to `LIST<STRING>` — clients binding by declared type will see a schema change.

### DIFF
--- a/query_modules/schema.cpp
+++ b/query_modules/schema.cpp
@@ -237,13 +237,41 @@ struct RelKey {
   bool operator==(const RelKey &) const = default;
 };
 
+struct RelKeyView {
+  std::string_view rel_type;
+  const std::set<std::string> &source_labels;
+  const std::set<std::string> &target_labels;
+};
+
 struct RelKeyHash {
-  std::size_t operator()(const RelKey &k) const noexcept {
-    std::size_t seed = std::hash<std::string>{}(k.rel_type);
-    boost::hash_combine(seed, boost::hash_range(k.source_labels.begin(), k.source_labels.end()));
-    boost::hash_combine(seed, boost::hash_range(k.target_labels.begin(), k.target_labels.end()));
+  using is_transparent = void;
+
+  std::size_t operator()(const RelKey &k) const noexcept { return Hash(k.rel_type, k.source_labels, k.target_labels); }
+
+  std::size_t operator()(const RelKeyView &k) const noexcept {
+    return Hash(k.rel_type, k.source_labels, k.target_labels);
+  }
+
+ private:
+  static std::size_t Hash(std::string_view rt, const std::set<std::string> &s,
+                          const std::set<std::string> &t) noexcept {
+    std::size_t seed = std::hash<std::string_view>{}(rt);
+    boost::hash_combine(seed, boost::hash_range(s.begin(), s.end()));
+    boost::hash_combine(seed, boost::hash_range(t.begin(), t.end()));
     return seed;
   }
+};
+
+struct RelKeyEqual {
+  using is_transparent = void;
+
+  bool operator()(const RelKey &a, const RelKey &b) const noexcept { return a == b; }
+
+  bool operator()(const RelKey &a, const RelKeyView &b) const noexcept {
+    return a.rel_type == b.rel_type && a.source_labels == b.source_labels && a.target_labels == b.target_labels;
+  }
+
+  bool operator()(const RelKeyView &a, const RelKey &b) const noexcept { return (*this)(b, a); }
 };
 
 mgp::List LabelsToList(const std::set<std::string> &labels) {
@@ -323,10 +351,6 @@ void Schema::NodeTypeProperties(mgp_list *args, mgp_graph *memgraph_graph, mgp_r
         continue;
       }
 
-      if (node.Properties().empty()) {
-        continue;
-      }
-
       for (const auto &[key, prop] : node.Properties()) {
         auto prop_type = TypeOf(prop.Type());
         if (current_labels_info.properties.find(key) == current_labels_info.properties.end()) {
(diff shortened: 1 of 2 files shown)


imilinovic:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - `schema.node_type_properties()` and `schema.rel_type_properties()` now return additional columns (propertyObservations, totalObservations, plus sourceNodeLabels / targetNodeLabels for rels, which are also now split per endpoint-label combination). Clients binding results positionally or assuming one row per (relType, propertyName) should switch to explicit YIELD <names> and account for the finer-grained rows. [#4153](https://github.com/memgraph/memgraph/pull/4153)
- [x] **[ Documentation PR link memgraph/documentation#1638 ]**

query_modules/schema.cpp
DavIvek: I know you didn't add this but we don't need this check anymore.

query_modules/schema.cpp
```diff
@@ -349,47 +414,63 @@ void Schema::RelTypeProperties(mgp_list *args, mgp_graph *memgraph_graph, mgp_re
...
+        // source_labels is reused across this node's edges, so it must be copied (not moved).
+        RelKey key{.rel_type = std::string(rel_type_view),
+                   .source_labels = source_labels,
+                   .target_labels = std::move(target_labels)};
+        auto &rel_info = rel_types_properties[std::move(key)];
         rel_info.number_of_occurrences++;
 
         if (rel.Properties().empty()) {
           continue;
         }
```
DavIvek: Same here, we don't need this check

query_modules/schema.cpp
```diff
@@ -349,47 +414,63 @@ void Schema::RelTypeProperties(mgp_list *args, mgp_graph *memgraph_graph, mgp_re
...
+        std::set<std::string> target_labels;
+        for (const auto label : rel.To().Labels()) {
+          target_labels.emplace(label);
+        }
+
+        // source_labels is reused across this node's edges, so it must be copied (not moved).
+        RelKey key{.rel_type = std::string(rel_type_view),
+                   .source_labels = source_labels,
+                   .target_labels = std::move(target_labels)};
+        auto &rel_info = rel_types_properties[std::move(key)];
```
DavIvek: We currently have to create a key for each entry before checking if it already exists. If this becomes a bottleneck in the future we could do something like:
```
struct RelKeyView {
    std::string_view rel_type;
    const std::set<std::string> &source_labels;
    const std::set<std::string> &target_labels;
  };

  struct RelKeyHash {
    using is_transparent = void;

    std::size_t operator()(const RelKey &k)     const noexcept { return Hash(k.rel_type, k.source_labels, k.target_labels); }
    std::size_t operator()(const RelKeyView &k) const noexcept { return Hash(k.rel_type, k.source_labels, k.target_labels); }

   private:
    static std::size_t Hash(std::string_view rt,
                            const std::set<std::string> &s,
                            const std::set<std::string> &t) noexcept {
      std::size_t seed = std::hash<std::string_view>{}(rt);
      boost::hash_combine(seed, boost::hash_range(s.begin(), s.end()));
      boost::hash_combine(seed, boost::hash_range(t.begin(), t.end()));
      return seed;
    }
  };
  
  struct RelKeyEqual {
    using is_transparent = void;
    bool operator()(const RelKey &a, const RelKey &b)        const noexcept { return a == b; }
    bool operator()(const RelKey &a, const RelKeyView &b)    const noexcept {
      return a.rel_type == b.rel_type && a.source_labels == b.source_labels && a.target_labels == b.target_labels;
    }
    bool operator()(const RelKeyView &a, const RelKey &b)    const noexcept { return (*this)(b, a); }
  };

  std::unordered_map<RelKey, LabelOrRelTypeInfo, RelKeyHash, RelKeyEqual> rel_types_properties;
```

and then:
```
const RelKeyView probe{rel_type_view, source_labels, target_labels};
  auto it = rel_types_properties.find(probe);
  if (it == rel_types_properties.end()) {
    RelKey key{.rel_type      = std::string(rel_type_view),
               .source_labels = source_labels,
               .target_labels = std::move(target_labels)};
    it = rel_types_properties.emplace(std::move(key), LabelOrRelTypeInfo{}).first;
  }
  auto &rel_info = it->second;
```

DavIvek (approved):
A few suggestions but looks good 🚀

PR #4154, opened 2026-05-19
feat: Add query RELOAD INTRA_CLUSTER TLS

as51340:
SSL code uses non-blocking sockets + poll in order to support timeouts on read and write, no-ssl uses blocking sockets as before. 

Breaking change because `RELOAD BOLT_SERVER TLS` now requires RELOAD_TLS privilege.


as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
   - Users can now reload their certificates for intra-cluster secured communication without the need for restarting processes.  Small breaking change since now both `RELOAD BOLT_SERVER TLS` and `RELOAD INTRA_CLUSTER TLS` require `RELOAD_TLS` privilege. [#4154]
- [x] **[ Documentation PR link ](https://github.com/memgraph/documentation/pull/1647)**

src/communication/context.cpp
```diff
@@ -102,27 +122,49 @@ ServerContext::ServerContext(std::string key_file, std::string cert_file, std::s
...
   auto ptr = ctx_.load(std::memory_order_acquire);
   MG_ASSERT(ptr, "Trying to use uninitialized SSL context");
   return ptr->native_handle();
 }
 
 boost::asio::ssl::context &ServerContext::context_clone() {
+  if (mode_ == Mode::ClusterView) {
+    auto ptr = ClusterServerSsl::Instance().CurrentContext();
+    MG_ASSERT(ptr, "ClusterView ServerContext used before ClusterServerSsl::Init");
+    return *ptr;
```
andrejtonev: If I am following correctly, we are using an atomic to store the shared pointer and then return a reference of the object. 
Why not pass back the shared pointer instead of the reference? 
Since we can change the ctx_ we might dereference a deleted object. Even if the current flow was correct and didn't uaf, it's just waiting to break.

src/communication/context.cpp
```diff
@@ -58,42 +59,61 @@ ClientContext::ClientContext(const std::string &key_file, const std::string &cer
...
     SSL_CTX_free(ctx_);
   }
 }
 
-SSL_CTX *ClientContext::context() { return ctx_; }
+SSL_CTX *ClientContext::context() {
+  if (mode_ == Mode::ClusterView) {
+    auto ptr = ClusterClientSsl::Instance().CurrentContext();
+    MG_ASSERT(ptr, "ClusterView ClientContext used before ClusterClientSsl::Init");
+    return ptr->native_handle();
```
andrejtonev: Similar comment to one before. Can we make is safer?

src/communication/client.cpp
```diff
@@ -176,31 +175,37 @@ auto Client::Read(size_t len, bool exactly_len, const std::optional<int> timeout
...
+          // POLLIN was set but no app data was decryptable (e.g., a TLS 1.3 NewSessionTicket
+          // consumed internally). Loop; the poll at the top of the next iteration enforces the timeout.
           continue;
         } else if (err == SSL_ERROR_WANT_WRITE) {
-          // The OpenSSL library probably wants to perform some kind of
-          // handshake so we wait for the socket to become ready for a write
-          // and call the read again.
-          socket_.WaitForReadyWrite();
+          // OpenSSL needs to write (renegotiation). Wait for write readiness with the same timeout.
+          if (!socket_.WaitForReadyWrite(timeout_ms)) {
```
andrejtonev: From the docs looks like we should call SSL_read again when the socket is ready for write. The SSL_read should do an internal write. But now we wait for read before calling SSL_read. This will then block until the timeout and fail.

src/query/interpreter.cpp
```diff
@@ -5036,14 +5038,68 @@ PreparedQuery PrepareReloadSSLQuery(ParsedQuery parsed_query, bool in_explicit_t
...
+              // the previous good state. After NuRaft commits successfully, the remaining two
+              // commits are pure atomic stores and genuinely cannot fail.
+              if (auto const res = coordination::CoordinatorCertReloader::Instance().Commit(std::move(*raft_prep));
+                  !res.has_value()) {
+                throw QueryRuntimeException("NuRaft TLS commit: " + res.error().msg);
+              }
+              communication::ClusterServerSsl::Instance().Commit(std::move(*srv_prep));
+              communication::ClusterClientSsl::Instance().Commit(std::move(*cli_prep));
             }
+#endif
```
andrejtonev, line 5102: should we do #else and throw if we happen to pass the wrong type in community builds?

src/coordination/coordinator_cert_reloader.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+  return data;
+}
+
+auto CoordinatorCertReloader::RefreshIfStale() -> std::expected<void, utils::SSL_CTX_Error> {
+  auto snapshot = current_.load(std::memory_order_acquire);
+
+  auto const cert_mtime_now = SafeMtime(cfg_.cert_file);
+  auto const key_mtime_now = SafeMtime(cfg_.key_file);
+  auto const ca_mtime_now = SafeMtime(cfg_.ca_file);
```
andrejtonev, line 182: any way we cache this or create a cooldown timer?
Not sure how often this is called in practice, but looks like a lot of disk calls. May become a bottleneck

src/coordination/coordinator_cert_reloader.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+  current_.store(*loaded, std::memory_order_release);
+  spdlog::info("Coordinator TLS reloaded via lazy mtime check (cert={}, key={}, ca={}).",
+               cfg_.cert_file,
+               cfg_.key_file,
+               cfg_.ca_file);
+  return {};
+}
+
+int CoordinatorCertReloader::CertCb(SSL *ssl, void *arg) {
```
andrejtonev, line 209: AI reviewed datarace:
[2026-05-27--certcb-thread-safety-verification.html](https://github.com/user-attachments/files/28301223/2026-05-27--certcb-thread-safety-verification.html)

src/coordination/coordinator_cert_reloader.cpp
```diff
@@ -0,0 +1,252 @@
...
+  // pure atomic store and cannot fail.
+  current_.store(std::move(prepared.next), std::memory_order_release);
+  spdlog::info("Coordinator TLS committed (cert={}, key={}, ca={}).", cfg_.cert_file, cfg_.key_file, cfg_.ca_file);
+  return {};
+}
+
+auto CoordinatorCertReloader::RefreshCa() -> std::expected<void, utils::SSL_CTX_Error> {
+  auto refresh = [&](SSL_CTX *ctx) -> std::expected<void, utils::SSL_CTX_Error> {
+    if (ctx == nullptr) return {};
+    if (SSL_CTX_load_verify_locations(ctx, cfg_.ca_file.c_str(), nullptr) != 1) {
```
andrejtonev: Looking around a bit at best practices looks like people usually create a new ctx instead of reloading the old ctx (in part because the ctx is not thread safe, in part in order to not break in flight connections). Example: https://yuneta.io/guide-ytls/#hot-reloading-certificates
I don't have the full picture in my head, so maybe this is irrelevant for our usecase.

src/coordination/coordinator_cert_reloader.cpp
```diff
@@ -0,0 +1,252 @@
...
+      return std::unexpected{utils::SSL_CTX_Error{
+          .err_type = utils::SSL_CTX_ERR_TYPE::FAIL_LOAD_CA,
+          .msg = fmt::format("SSL_CTX_load_verify_locations('{}') failed: {}", cfg_.ca_file, OpenSslErrorString())}};
+    }
+    return {};
+  };
+  if (auto r = refresh(server_ctx_); !r.has_value()) return r;
+  if (auto r = refresh(client_ctx_); !r.has_value()) return r;
```
andrejtonev: Can we succeed on server and fail on client? What will this mean for the connections?

src/io/network/socket.cpp
```diff
@@ -267,27 +267,31 @@ bool Socket::Bind(const Endpoint &endpoint) {
 
 // Not const because of C-API
 // NOLINTNEXTLINE
-void Socket::SetNonBlocking() {
-  const unsigned flags = fcntl(socket_, F_GETFL);
-  constexpr unsigned o_nonblock = O_NONBLOCK;
-  MG_ASSERT(flags != -1, "Can't get socket mode");
-  MG_ASSERT(fcntl(socket_, F_SETFL, flags | o_nonblock) != -1, "Can't set socket nonblocking");
+auto Socket::SetNonBlocking() -> std::expected<void, std::string> {
```
andrejtonev: Please audit all the usages of SetNonBlocking() and check if they handle the error correctly (we didn't need to think about it before due to the assert). 
Looks like at least the tests/unit/scoket.cpp is not checking the return value. 
Good practice to add [[nodiscard]]

src/io/network/socket.cpp
```diff
@@ -267,27 +267,31 @@ bool Socket::Bind(const Endpoint &endpoint) {
...
+    spdlog::error("Failed to restore socket to the non-blocking mode");
+    return std::unexpected{"Failed to restore socket to the non-blocking mode"};
+  }
+  return {};
 }
 
 // Not const because of C-API
 // NOLINTNEXTLINE
-auto Socket::SetBlocking() -> bool {
+auto Socket::SetBlocking() -> std::expected<void, std::string> {
```
andrejtonev: Remove, since looks like we are not using this anymore

src/io/network/socket.hpp
```diff
@@ -108,14 +108,14 @@ class Socket {
...
   /**
-   * Sets the socket to non-blocking.
+   * Sets the socket to non-blocking. Returns false on failure.
    */
-  void SetNonBlocking();
+  auto SetNonBlocking() -> std::expected<void, std::string>;
 
   /**
-  Sets the socket to the blocking mode.
+  Sets the socket to the blocking mode. Returns false on failure.
```
andrejtonev: returns expected, not bool

src/communication/cluster_tls.cpp
```diff
@@ -0,0 +1,210 @@
...
+  // NOLINTNEXTLINE(bugprone-unused-return-value)
+  new_ctx->use_private_key_file(cfg_.key_file, ssl::context::pem, ec);
+  if (ec) {
+    auto msg = fmt::format("Cluster server TLS: couldn't load private key from {}: {}", cfg_.key_file, ec.message());
+    spdlog::error(msg);
+    return std::unexpected{
+        utils::SSL_CTX_Error{.err_type = utils::SSL_CTX_ERR_TYPE::FAIL_KEY_FILE, .msg = std::move(msg)}};
+  }
+  // NOLINTNEXTLINE(bugprone-unused-return-value)
+  new_ctx->set_options(SSL_OP_NO_SSLv3, ec);
```
andrejtonev: didn't we already disable it on line 63?

src/communication/cluster_tls.cpp
```diff
@@ -0,0 +1,210 @@
...
+  if (cfg_.key_file.empty() || cfg_.cert_file.empty()) {
+    return std::unexpected{utils::SSL_CTX_Error{.err_type = utils::SSL_CTX_ERR_TYPE::FLAGS_NOT_CONFIGURED,
+                                                .msg = "Cluster client TLS: key or certificate file is empty."}};
+  }
+
+  namespace ssl = boost::asio::ssl;
+  auto new_ctx = std::make_shared<boost::asio::ssl::context>(ssl::context::tls_client);
+
+  // NOLINTNEXTLINE(bugprone-unused-return-value)
+  new_ctx->set_options(SSL_OP_NO_SSLv3);
```
andrejtonev: the server ctx had default_workarounds, no_sslv2, no_sslv3, single_dh_use. 
the client just has no_sslv3. is that correct? shouldn't they have the same setup?

src/communication/client.cpp
andrejtonev, line 171: we keep passing the timeout_ms to each poll, meaning the read could take up to N*timeout_ms. Should we switch to a deadline approach?

src/communication/context.cpp
as51340: you are right, the point of such a design is for a session to store shared ptr and to rely on refcount

src/communication/context.cpp
as51340: fixed

src/coordination/coordinator_cert_reloader.cpp
as51340: this is invoked only on socket connect for NuRaft. First, this should be rarely because connections are established and then hold for a long time. Second, this is still cheap operation (order of microseconds) compared to TLS handshake which is in order of milliseconds.

src/io/network/socket.cpp
as51340: I checked, the only usage in code is in src/communication/session.hpp which I changed. Will change unit tests though

src/coordination/coordinator_cert_reloader.cpp
as51340: solved by adding X509 store to the snapshot

src/coordination/coordinator_cert_reloader.cpp
as51340: we do this for all connections except NuRaft's since NuRaft only exposes pointer to the context, that's why we do it in this way,

src/coordination/coordinator_cert_reloader.cpp
as51340: this is now removed

src/communication/cluster_tls.cpp
as51340: yes, you are right

src/communication/cluster_tls.cpp
as51340: I will explicitly disable sslv2. single_dh_use is only server config

src/communication/client.cpp
as51340: solved

src/communication/client.cpp
```diff
@@ -168,47 +167,76 @@ auto Client::Read(size_t len, bool exactly_len, const std::optional<int> timeout
...
         if (err == SSL_ERROR_WANT_READ) {
-          // OpenSSL want's to read more data from the socket. We wait for
-          // more data to be ready and retry the call.
-          socket_.WaitForReadyRead();
+          // POLLIN was set but no app data was decryptable (e.g., a TLS 1.3 NewSessionTicket
+          // consumed internally).
+          // Wait for the underlying socket to have data before SSL_read. With a non-blocking socket,
+          // polling unconditionally (timeout_ms == nullopt -> poll(-1)) avoids a busy loop when
+          // SSL_read returns WANT_READ on the retry path below.
+          if (!socket_.WaitForReadyRead(remaining_timeout_ms)) {
```
andrejtonev, line 215: Issue not caused by this PR but should fix:
WaitForReadyRead/Write are not handling EINTR correctly, they just return failure instead or retrying.

src/communication/context.cpp
andrejtonev: this still seems to give back a raw pointer. am I missing something?

src/query/interpreter.cpp
andrejtonev: this is unconditional so will always throw. you need the else

src/query/interpreter.cpp
```diff
@@ -5036,19 +5038,71 @@ PreparedQuery PrepareReloadSSLQuery(ParsedQuery parsed_query, bool in_explicit_t
...
+                               communication::ClusterServerSsl::Instance().Commit(std::move(*srv_prep));
+                               communication::ClusterClientSsl::Instance().Commit(std::move(*cli_prep));
+                               coordination::CoordinatorCertReloader::Instance().Commit(std::move(*raft_prep));
+                             }
+#else
+                             throw QueryRuntimeException(
+                                 "RELOAD INTRA_CLUSTER TLS cannot be invoked in the community build");
+#endif
+                             notifications->emplace_back(
+                                 SeverityLevel::INFO, NotificationCode::RELOAD_SSL, "Reloading SSL for Bolt server");
```
andrejtonev: different text based on the exact target?

src/communication/client.cpp
andrejtonev: this goes for write as well

src/auth/models.hpp
```diff
@@ -81,6 +81,7 @@ enum class Permission : uint64_t {
   PARALLEL_EXECUTION     = 1U << 28U,
   SERVER_SIDE_PARAMETERS = 1U << 29U,
   SERVER_SIDE_DESCRIPTIONS = 1U << 30U,
+  RELOAD_TLS             = 1U << 31U
```
andrejtonev: use ULL so the next person uses ULL automatically (force uint64t)

src/query/frontend/semantic/required_privileges.cpp
```diff
@@ -189,7 +189,7 @@ class PrivilegeExtractor : public QueryVisitor<void>, public HierarchicalTreeVis
 
   void Visit(AlterEnumUpdateValueQuery & /*enum_query*/) override { AddPrivilege(AuthQuery::Privilege::CREATE); }
 
-  void Visit(ReloadSSLQuery & /*reload_ssl_query*/) override { /* no privilege needed to reload SSL */ }
+  void Visit(ReloadSSLQuery & /*reload_ssl_query*/) override { AddPrivilege(AuthQuery::Privilege::RELOAD_TLS); }
```
andrejtonev, line 192: should we mark this as breaking? or there wasn't any reload in 3.10?

as51340: There was for bolt server, I will add a breaking change

src/query/interpreter.cpp
as51340: But I added else, not sure what do you mean?

src/communication/client.cpp
as51340: yes, good resource here https://www.hardening-consulting.com/en/posts/20210322-select-poll-and-eintr.html, fixed it now

src/query/interpreter.cpp
as51340: I added else not sure what do you mean

andrejtonev: you have
```
if (action_type == ReloadSSLQuery::Type::BOLT_SERVER) {
...
}
#ifdef MG_ENTERPRISE
else {
...
}
#else
throw ...
#endif
```
the if action type does not return and the else branch is completely inside the enterprise so in community we go to next line and throw unconditionally.

src/communication/client.cpp
```diff
@@ -168,47 +173,76 @@ auto Client::Read(size_t len, bool exactly_len, const std::optional<int> timeout
...
 
       // Handle errors that might have occurred.
       if (got < 0) {
         auto err = SSL_get_error(ssl_, got);
+        auto const remaining_timeout_ms = std::invoke([&deadline]() -> std::optional<int> {
+          if (!deadline) return std::nullopt;
+          return std::chrono::duration_cast<std::chrono::milliseconds>(*deadline - std::chrono::steady_clock::now())
+              .count();
+        });
+        if (remaining_timeout_ms && *remaining_timeout_ms < 0) {
```
andrejtonev: lets use <= 0, no need to pass if timeout is 0

src/coordination/coordinator_cert_reloader.cpp
```diff
@@ -0,0 +1,254 @@
...
+  static CoordinatorCertReloader instance;
+  return instance;
+}
+
+auto CoordinatorCertReloader::Register(SSL_CTX *server_ctx, SSL_CTX *client_ctx, utils::TlsConfig cfg)
+    -> std::expected<void, utils::SSL_CTX_Error> {
+  std::lock_guard const guard{reload_mu_};
+  server_ctx_ = server_ctx;
+  client_ctx_ = client_ctx;
+  cfg_ = std::move(cfg);
```
andrejtonev, line 61: we read the cfg_ without the lock. is that safe because we only register once at the start?
can we add an assert here to verify the cfg_ is not being changed?

src/communication/cluster_tls.cpp
```diff
@@ -0,0 +1,203 @@
...
+void ClusterClientSsl::Commit(ClusterClientReloadPrepared prepared) noexcept {
+  ctx_.store(std::move(prepared.next_ctx), std::memory_order_release);
+  spdlog::info("Cluster client TLS reloaded (cert={}, key={}, ca={}).", cfg_.cert_file, cfg_.key_file, cfg_.ca_file);
+}
+
+std::shared_ptr<boost::asio::ssl::context> ClusterClientSsl::CurrentContext() const {
+  return ctx_.load(std::memory_order_acquire);
+}
+
+auto ClusterClientSsl::Build() -> std::expected<std::shared_ptr<boost::asio::ssl::context>, utils::SSL_CTX_Error> {
```
andrejtonev, line 170: can the server and client build be combined into a single free function?

andrejtonev (approved):
a few smaller issues to fix and then good to go

src/communication/cluster_tls.cpp
as51340: not into a single function because they are differences but I will deduplicate as much as possible

src/coordination/coordinator_cert_reloader.cpp
as51340: Added a check that the Register is called at most once.

PR #4156, opened 2026-05-19
feat: Implement Property-based access control

colinbarry:
Implement property-based access control on nodes and relationships.

`GRANT READ {}` is used to grant read access to nodes or relationships for the given user or role. Properties are either specified as a `{ }` delimited list of properties, or `{*}` as a wildcard for all properties.

```cypher
GRANT READ {prop, ...} ON NODES CONTAINING LABELS :Label TO user_or_role;
GRANT READ {*} ON NODES CONTAINING LABELS :Label MATCHING ANY TO user_or_role;
GRANT READ, SET PROPERTY {prop} ON NODES CONTAINING LABELS :Label MATCHING EXACTLY TO user_or_role;
```

Similarly, we can `DENY` and `REVOKE` permissions using similar syntax:                                                  
```cypher
DENY READ {prop} ON NODES CONTAINING LABELS :Label TO user_or_role;
REVOKE READ {prop} ON NODES CONTAINING LABELS :Label FROM user_or_role;                                                                                                                                   
```
     
`GRANT`, `DENY`, and `REVOKE` of properties also applies to edges:
```cypher
GRANT READ {prop, ...} ON EDGES OF TYPE :Type TO user_or_role;
DENY SET PROPERTY {prop} ON EDGES OF TYPE :Type TO user_or_role;
REVOKE READ {prop} ON EDGES OF TYPE * FROM user_or_role;
```


colinbarry:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Added: Property-Based Access Control (PBAC). Administrators can now control which properties a user or role can read or write, scoped per node label or relationship type. Denied properties can be redacted on read and rejected on write.  [#4156](https://github.com/memgraph/memgraph/pull/4156})

src/glue/auth_checker.cpp
```diff
@@ -311,13 +308,102 @@ bool FineGrainedAuthChecker::HasAllGlobalPrivilegesOnEdges() const {
...
+    return permissions.HasGlobal(prop_name, perm_type) == auth::PermissionLevel::GRANT;
+  }
+  std::vector<std::string> label_names;
+  label_names.reserve(labels.size());
+  for (auto label : labels) {
+    label_names.push_back(dba_->LabelToName(label));
+  }
+  auto level = permissions.Has(label_names, prop_name, perm_type);
+  return level == auth::PermissionLevel::GRANT;
+}
```
andrejtonev, line 392: User with `GRANT READ ON NODES CONTAINING LABELS :A TO u` (label-FGA only, no property rule), running `MATCH (n:A) RETURN n.name`:
1. Per-label LBAC ⇒ `HasUnrestrictedAccessToVertices()` = false ⇒ `NeedsFineGrainedAuthChecker()` (340–349) returns **true** (the `is_unconfigured` escape only guards the *property* dimension; the LBAC-vertices term defeats it) ⇒ checker attached.
2. `HasPropertyPermission` has **no** `is_unconfigured` escape hatch — only the `HasUnrestrictedAccessToVertexProperties()` early-out (line 374), which is **false** for an empty `global_`. License valid ⇒ no early-out ⇒ `permissions.Has([:A], "name", READ)` over empty `rules_`+`global_` ⇒ **NEUTRAL**.
3. `NEUTRAL != GRANT` ⇒ **false** ⇒ `n.name` nulled. Every property of every labeled entity is nulled.

src/auth/models.cpp
```diff
@@ -660,6 +663,297 @@ bool operator==(const FineGrainedAccessHandler &first, const FineGrainedAccessHa
...
+void PropertyAccessPermissions::Deny(std::unordered_set<std::string> const &entities, std::string const &property,
+                                     PropertyPermissionType type, MatchingMode matching_mode) {
+  auto const bit = type;
+  auto &rule = FindOrCreateRule(entities, matching_mode);
+  auto &perm = rule.properties[property];
+  perm.denies |= bit;
+  perm.grants &= ~bit;
+  if (property == "*") {
+    std::erase_if(rule.properties,
+                  [](auto const &kv) { return kv.first != "*" && kv.second.grants == PropertyPermissionType::NONE; });
```
andrejtonev: `erase_if` predicate checks only **one** bit, while `Revoke` (723–725) correctly checks both:
- **Deny side** (predicate `kv.second.grants == NONE`): `Deny("ssn", READ)` ⇒ `ssn:{grants=0, denies=READ}`; then `Deny("*", WRITE)` ⇒ predicate `ssn.grants==NONE` true ⇒ **ssn entry erased** ⇒ READ deny lost ⇒ `Has("ssn", READ)` falls through to wildcard `{denies=WRITE}` ⇒ **NEUTRAL** ⇒ ssn becomes readable. **Security-weakening.**
- **Grant side** (mirror, predicate `denies == NONE`): a prior named grant on the other bit is dropped ⇒ **wrongful denial**.

src/auth/models.cpp
```diff
@@ -660,6 +663,297 @@ bool operator==(const FineGrainedAccessHandler &first, const FineGrainedAccessHa
...
+void PropertyAccessPermissions::Grant(std::unordered_set<std::string> const &entities, std::string const &property,
+                                      PropertyPermissionType type, MatchingMode matching_mode) {
+  auto const bit = type;
+  auto &rule = FindOrCreateRule(entities, matching_mode);
+  auto &perm = rule.properties[property];
+  perm.grants |= bit;
+  perm.denies &= ~bit;
+  if (property == "*") {
+    std::erase_if(rule.properties,
+                  [](auto const &kv) { return kv.first != "*" && kv.second.denies == PropertyPermissionType::NONE; });
```
andrejtonev: same issue as deny

src/query/plan/operator.cpp
```diff
@@ -5373,14 +5342,41 @@ void SetPropertiesOnRecord(TRecordAccessor *record, const TypedValue &rhs, SetPr
     }
   };
 
+  auto const mask_denied_properties [[maybe_unused]] = [&](PropertiesMap &props, auto const &check_read_fn) {
+    for (auto &[prop_id, value] : props) {
+      if (!check_read_fn(prop_id)) {
+        value = storage::PropertyValue{};
+      }
+    }
+  };
```
andrejtonev, line 5352: `SET v += n` silently deletes a target property that is READ-denied on the source

**Trace** (`SET v += n`, target `v.secret` exists, source `n.secret` READ-denied to caller):
1. `mask_denied_properties` sets the denied source entry to a **NULL `PropertyValue{}`** but leaves the key in the map (`operator.cpp:5348`).
2. `UpdateProperties` clears the store (`:3004`), then `properties.emplace(secret, old_value)` is a **no-op** because `secret` is already present as the NULL sentinel (`:3015`).
3. `DoInitProperties` skips `IsNull()` entries (`:2976`) ⇒ `secret` never re-encoded.
4. Net: `v.secret` is **silently wiped** — data loss, not redaction.

src/glue/auth_handler.cpp
```diff
@@ -342,6 +342,87 @@ std::vector<std::vector<memgraph::query::TypedValue>> ShowFineGrainedRolePrivile
...
+                             std::vector<std::vector<memgraph::query::TypedValue>> &result) {
+  if (!perms.GetGlobalRules().empty()) {
+    EmitPropertyMap(perms.GetGlobalRules(), global_entity_kind, true, user_or_role_str, result);
+  }
+  for (auto const &rule : perms.GetRules()) {
+    std::vector<std::string> sorted_entities(rule.entities.begin(), rule.entities.end());
+    std::ranges::sort(sorted_entities);
+    auto formatted = sorted_entities | std::views::transform([](auto const &e) { return fmt::format(":{}", e); });
+    auto entity_str = memgraph::utils::Join(formatted, ", ");
+    if (rule.entities.size() > 1) {
```
andrejtonev: `SHOW PRIVILEGES` hides `MATCHING EXACTLY`/`ANY` on single-label rules
**`src/glue/auth_handler.cpp:410`.** The `(EXACTLY)/(ANY)` suffix is only appended when `rule.entities.size() > 1`, so a single-label `MATCHING EXACTLY` rule renders identically to `MATCHING ANY` — an admin auditing least-privilege can't distinguish them (the modes differ semantically for a node carrying extra labels). **Verified by:** review-A, Skeptic-B, guided-D. **Fix:** drop the `>1` guard (or always render the mode when `EXACTLY`).

src/glue/auth_checker.cpp
colinbarry: By design. Property-based access control follows the same model as Label-based access control: if you don't GRANT anything, you don't access anything.

For new users, you need to add a `GRANT READ {*} ON NODES CONTAINING LABELS * TO alice` or whatever. Generally, the easiest thing to do is just assign a new user either a `readwrite` or `readonly` role.

Note that users/roles created with any Memgraph version without PBAC are automagically updated to have the global READ/SET PROPERTY grants. So queries will work the same as before.

src/auth/models.cpp
colinbarry: Added test to confirm this issue, and fixed the code.

src/auth/models.cpp
colinbarry: Added test to confirm this issue, and fixed the code.

src/query/plan/operator.cpp
colinbarry: By design (admittedly, not obvious.) Essentially, `SET v += n` , assuming `n` has `n.prop1 and n.prop2` is logically equivalent to:

```
SET v.prop1 = n.prop1, v.prop2 = n.prop2
```

If the current user does not have `READ` permission on `n.prop2`, the read will return `null`, and hence null/remove `v.prop2`, same as if we did a direct `SET v.prop2 = null`

Think this is fine and the correct behaviour. I did add a unit test to capture this behaviour.

src/glue/auth_handler.cpp
colinbarry: Yeah, you're right. Fixed.

andrejtonev (approved):

PR #4182, opened 2026-05-26
feat(plan_v2): Improve planner extraction

### LINKED ISSUE #4190: Improve cost estimation and extraction
Our egraph plan_v2 for CYPHER execution has semantics around storing symbols in a frame. To support dead store elimination and other global level optimisations, we need Pareto frontier approach in the cost model and extraction.

Given an inline rewrite rule we can turn
```cypher
WITH 42 AS x RETURN x AS result;
```

into effectively:
```cypher
WITH 42 AS x RETURN 42 AS result;
```

the Pareto frontier allows us to do dead store elimination, giving us effectively:
```cypher
RETURN 42 AS result;
```

Ignition:
Bring planner v2 extraction to the point where it carries full expression trees (arithmetic, comparison, boolean, function, unwind, subquery) and selects plans on cost. The extractor now searches the equivalent plans held in the e-graph and picks the cheapest, scoring candidates with a cost model and a pluggable cardinality estimator. Candidates can be better on different axes - estimated cost, output cardinality, and demand (whether later operators actually use a value) - so it keeps a Pareto frontier of non-dominated candidates instead of committing early. Propagating demand lets it drop work nothing reads: unreferenced alternatives are pruned, and an operator like Bind takes its cheaper variant that skips evaluating the bound value wherever nothing downstream consumes it.

The per-symbol behaviour is organised into trait families spanning the make, cost, resolve, and build axes, with a single dispatcher across them. Adding a new operator is driven from one X-macro symbol list that flows into the enum, predicates, and round-trip; a missing trait specialisation is a compile error rather than a silent gap. plan_v2 is reorganised along these axes into egraph, cost, frontend, and resolve layers.

Extraction state is threaded through the interpreter via a planner context, replacing the previous ad-hoc plumbing. Test and benchmark support is consolidated under shared test_support libraries, with new algebra, cardinality, and extractor suites covering the Pareto cases and full operator round-trips, plus new extraction and match benchmarks.


src/planner/bench/bench_common.hpp
```diff
@@ -40,40 +32,6 @@ using core::ProcessingContext;
...
-
-// Variables for eclass-level hoisting benchmarks
-constexpr PatternVar kHoistR{30};  // ?r = F(?x) root binding
-constexpr PatternVar kHoistY{31};  // Mul(?r, ?y) sibling binding
-
-// ============================================================================
-// Benchmark Size Constants
-// ============================================================================
-
 namespace sizes {
```
Ignition: why have its own namespace if we have a using statement straight away?

src/planner/include/planner/extract/extract.hpp
```diff
@@ -0,0 +1,165 @@
...
+
+/// Non-template half of the resolver's output bundle: the visit-ordered entries
+/// plus the CSR child-index buffer.  Downstream consumers (e.g. a builder that
+/// walks `build_order` children-before-parents) only need this view of the
+/// resolver's result and don't depend on the cost-model or resolver-key types.
+///
+/// `append_children` and `children_of` are the only intended way to write into
+/// or read from `child_indices`; they own the half-open-window invariant.
+struct ResolvedBuildContext {
+  std::vector<ResolvedEntry> build_order;
```
Ignition, line 67: Is there useful methods we can add for interacting with `build_order` rather than manipulating it directly?

src/planner/include/planner/extract/extractor.hpp
```diff
@@ -11,174 +11,286 @@
...
-      best_node = Selection<CostResult>{enode_id, current_cost};
-    }
-  }
+    auto enode_frontier = cost_model(enode, enode_id, children_frontiers);
 
-  // Update with the actual computed cost
-  if (best_node) {
-    it2->second = *best_node;
-    return best_node->cost_result;
+    if (!merged_frontier) {
```
Ignition: this should be the eclass_frontier

src/planner/include/planner/extract/extractor.hpp
```diff
@@ -11,174 +11,286 @@
...
+    if (!inserted) {
+      assert(it->second != kCycleSentinel && "cycle in DfsPostOrder");
+      return it->second;
     }
-  }
-  return result;
+    auto entry = make_entry(key, self);
+    auto const idx = static_cast<std::uint32_t>(out.size());
+    out.push_back(std::move(entry));
+    auto const write_it = seen.find(key);
```
Ignition: `write_it` should be same as `it`, we inserted it, no need for 2nd find.
The assert by design is impossible.

src/planner/include/planner/extract/extractor.hpp
```diff
@@ -11,174 +11,286 @@
...
+  // re-lookup the slot for the final write after `make_entry` returns.
+  auto recurse = [&](this auto const &self, Key key) -> std::uint32_t {
+    auto [it, inserted] = seen.try_emplace(key, kCycleSentinel);
+    if (!inserted) {
+      assert(it->second != kCycleSentinel && "cycle in DfsPostOrder");
+      return it->second;
     }
-  }
-  return result;
+    auto entry = make_entry(key, self);
```
Ignition, line 285: better name? this is where recursion happens, what contractually makes this DFS?

src/planner/include/planner/extract/pareto_frontier.hpp
```diff
@@ -0,0 +1,506 @@
...
+// some `<=>` ordering type.  Totality is detected at compile time from that
+// return type: `strong_ordering` / `weak_ordering` ⇒ totally ordered.
+// Comparators that need set-based reasoning return `partial_ordering`.
+
+/// "Lower is better" for any three-way-comparable type.  Returns whatever the
+/// underlying `<=>` yields (typically `strong_ordering`) so the totally-ordered
+/// trait is detectable.  Swapping operands inverts the ordering so a < b ⇒
+/// "a dominates b" ⇒ `greater`.
+struct LowerIsBetter {
+  static constexpr auto compare(auto const &a, auto const &b) { return b <=> a; }
```
Ignition, line 40: `b <=> a` how does that correlate with LowerIsBetter?

src/query/plan_v2/cost/builtin_estimator.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+}  // namespace
+
+BuiltinEstimator::BuiltinEstimator(egraph const &f) : CardinalityEstimator(impl_of(f).graph.core()), facade(f) {}
+
+auto BuiltinEstimator::Estimate(planner::core::ENode<symbol> const &enode,
+                                std::span<planner::core::EClassId const> arg_eclasses) const -> double {
+  switch (enode.symbol()) {
+    case symbol::Function:
+      return EstimateFunction(eg_, facade, enode, arg_eclasses);
```
Ignition: makes sense for function that these are `arg_eclasses` but in general, that won't hold. maybe just call them children for now.

src/query/plan_v2/cost/cost_model.hpp
```diff
@@ -0,0 +1,49 @@
...
+namespace memgraph::query::plan::v2 {
+
+/// CostFrontier: ParetoFrontier with resolve/min_cost for the extraction contract.
+struct CostFrontier
+    : planner::core::extract::CostResultBase<Alternative, AlternativeDim_Cost, AlternativeDim_Cardinality,
+                                             AlternativeDim_Required, AlternativeDim_Introduces> {
+  using CostResultBase::CostResultBase;
+};
+
+/// Per-extraction cost-model state passed to every `symbol_cost_traits<S>::cost`.
```
Ignition: Hidden detail we should cleanup comments

src/query/plan_v2/cost/cost_model.hpp
```diff
@@ -0,0 +1,49 @@
...
+struct CostCtx {
+  using CostResult = CostFrontier;
+
+  // NOLINTBEGIN(cppcoreguidelines-avoid-const-or-ref-data-members)
+  CardinalityEstimator const &estimator;
+  ExtractionEnv const &env;
```
Ignition: Why reference ExtractionEnv? we are in cost's code.

src/query/plan_v2/cost/cost_model.cpp
```diff
@@ -0,0 +1,478 @@
...
+        DMG_ASSERT(a.required.empty(), "operator Alt must have empty required");
+        return b.required.is_subset_of(a.introduces);
+      },
+      [enode_id](Alternative const &a, Alternative const &b) {
+        return Alternative{.cost = a.cost + (a.cardinality * b.cost),
+                           .cardinality = a.cardinality,
+                           .required = {},
+                           .introduces = a.introduces,
+                           .enode_id = enode_id};
+      });
```
Ignition, line 138: maintenance, `a` and `b` should be named to help keep track if its an Alt of `row_pipe` or of `named_out`

src/query/plan_v2/cost/cost_model.cpp
```diff
@@ -0,0 +1,478 @@
...
+      named_out,
+      [](Alternative const &a, Alternative const &b) {
+        DMG_ASSERT(a.required.empty(), "operator Alt must have empty required");
+        return b.required.is_subset_of(a.introduces);
+      },
+      [enode_id](Alternative const &a, Alternative const &b) {
+        return Alternative{.cost = a.cost + (a.cardinality * b.cost),
+                           .cardinality = a.cardinality,
+                           .required = {},
+                           .introduces = a.introduces,
```
Ignition: Do we need a comment to explain these semantics, does the parent Output node clean up the .introduces + .required?

src/query/plan_v2/cost/cost_model.cpp
```diff
@@ -0,0 +1,478 @@
...
+    -> CostFrontier {
+  return CostFrontier::cartesian_product_if(
+      row_pipe,
+      named_out,
+      [](Alternative const &a, Alternative const &b) {
+        DMG_ASSERT(a.required.empty(), "operator Alt must have empty required");
+        return b.required.is_subset_of(a.introduces);
+      },
+      [enode_id](Alternative const &a, Alternative const &b) {
+        return Alternative{.cost = a.cost + (a.cardinality * b.cost),
```
Ignition: `UNWIND range(0,100) as x RETURN x*2 as result;`
For `x*2 as result`
- `x*2` is b's expression, it has a cost
- stream of `x` is what `a` pulls, it has a cost and cardinality

Hence `a.cost + (a.cardinality * b.cost)`

src/query/plan_v2/cost/cost_model.cpp
```diff
@@ -0,0 +1,478 @@
...
+/// so by the time a Bind is constructed, every relevant in-scope symbol is in
+/// `input_alt.introduces`.
+///
+/// (input, expr) combinations where `expr.required ⊄ input.introduces` are
+/// skipped: they cannot form a satisfiable plan, but other (input, expr) pairs
+/// in the cartesian cross do (e.g. the alive-Bind variant of the row pipe with
+/// the same expr), so the frontier still emits the valid combinations.  This
+/// also subsumes the post-inline-rewrite self-referential case where
+/// `Identifier(sym)` is unioned into expr's e-class.
+auto BindFlatMap(CostFrontier const &input, CostFrontier const &expr, planner::core::EClassId sym_eclass,
```
Ignition, line 166: BindFlatMap, OutputCombine, UnwindFlatMap, SubqueryFlatMap, FunctionCombine. Do they deserve there own named functions? Is there a common generalisation that they can use, or do we inline into their symbol_cost_traits?

src/query/plan_v2/cost/cost_model.cpp
```diff
@@ -0,0 +1,478 @@
...
+                 double sym_cost, VariableSet const &referenced_syms, VariableIndex const &idx,
+                 planner::core::ENodeId enode_id) -> CostFrontier {
+  auto const sym_bit = idx.bit_of(sym_eclass);
+  return CostFrontier::flat_map(input, [&, enode_id, sym_bit](Alternative const &input_alt, auto emit) {
+    DMG_ASSERT(input_alt.required.empty(), "operator Alt must have empty required");
+    // Alive: sym has at least one Identifier reference somewhere in the e-graph.
+    if (referenced_syms.test(sym_bit)) {
+      for (Alternative const &expr_alt : expr.alts()) {
```
Ignition, line 174: is this not a combine_if or cartesian_if pattern?

src/query/plan_v2/cost/cost_model.cpp
```diff
@@ -0,0 +1,478 @@
...
+/// applies: emitted `required = ∅`.  (input, list)
+/// combinations whose list-expression demand isn't satisfied by `input.introduces`
+/// are skipped (see BindFlatMap header for rationale).
+auto UnwindFlatMap(CostFrontier const &input, CostFrontier const &list, planner::core::EClassId sym_eclass,
+                   double sym_cost, VariableIndex const &idx, planner::core::ENodeId enode_id) -> CostFrontier {
+  auto const sym_bit = idx.bit_of(sym_eclass);
+  return CostFrontier::flat_map(input, [&, enode_id, sym_bit](Alternative const &input_alt, auto emit) {
+    DMG_ASSERT(input_alt.required.empty(), "operator Alt must have empty required");
+    for (Alternative const &list_alt : list.alts()) {
+      if (!list_alt.required.is_subset_of(input_alt.introduces)) continue;  // demand unsatisfiable by this input
```
Ignition: This pattern of checking that operators alt satisfies the expression alt is common. Bind, Unwind, etc. And will be more common as we add in other operators that use expressions (eg. Filter). Can this be generalised into something that can be reused?

src/query/plan_v2/cost/cost_model.cpp
```diff
@@ -0,0 +1,478 @@
...
+/// introductions are STRIPPED at the boundary; only `exposed_syms`
+/// (children[2..]) cross into the outer scope, unioned with outer's own
+/// introductions.
+auto SubqueryFlatMap(CostFrontier const &outer, CostFrontier const &inner, VariableSet exposed_syms,
+                     planner::core::ENodeId enode_id) -> CostFrontier {
+  return CostFrontier::flat_map(
+      outer, [&inner, exposed_syms = std::move(exposed_syms), enode_id](Alternative const &outer_alt, auto emit) {
+        for (Alternative const &inner_alt : inner.alts()) {
+          // Non-importing subquery: inner must be self-contained.
+          if (!inner_alt.required.empty()) continue;
```
Ignition: This will not always hold true, `inner_alt` might require something.

```
EXPLAIN WITH "test" AS x CALL { WITH x RETURN x AS y } RETURN y;
+--------------------+
| QUERY PLAN         |
+--------------------+
| " * Produce {y}"   |
| " * Apply"         |
| " |\\ "            |
| " | * Produce {y}" |
| " | * Produce {x}" |
| " | * Once"        |
| " * Produce {x}"   |
| " * Once"          |
+--------------------+
```

src/query/plan_v2/cost/cost_model.cpp
```diff
@@ -0,0 +1,478 @@
...
+
+          // BARRIER: outer.introduces ∪ exposed_syms; inner.introduces is dropped.
+          emit({.cost = outer_alt.cost + (outer_alt.cardinality * inner_alt.cost),
+                .cardinality = outer_alt.cardinality * inner_alt.cardinality,
+                .required = outer_alt.required,
```
Ignition: Need to consider `inner_alt.required`

src/query/plan_v2/cost/cost_model.cpp
```diff
@@ -0,0 +1,478 @@
...
+    DMG_ASSERT(input_alt.required.empty(), "operator Alt must have empty required");
+    // Alive: sym has at least one Identifier reference somewhere in the e-graph.
+    if (referenced_syms.test(sym_bit)) {
+      for (Alternative const &expr_alt : expr.alts()) {
+        if (!expr_alt.required.is_subset_of(input_alt.introduces)) continue;  // demand unsatisfiable by this input
+        // Bind is one-shot, not a row-pipe: passes input's cardinality
+        // through unchanged.  expr is evaluated once at bind-time.
+        emit({.cost = BindAliveCost(input_alt.cost, sym_cost, expr_alt.cost, input_alt.cardinality),
+              .cardinality = input_alt.cardinality,
+              .required = {},
```
Ignition, line 180: Does this need reconsidering? will operators on inner branches need to communicate their requirements?

src/query/plan_v2/cost/cost_model.cpp
```diff
@@ -0,0 +1,478 @@
...
+  auto result = args.empty() ? LeafFrontier(0.0, enode_id) : Restamp(*args[0], 0.0, enode_id);
+  for (auto const *arg : args.empty() ? args : args.subspan(1)) {
+    result = BinaryCombine(result, *arg, 0.0, enode_id);
+  }
+  result.mutate_pruning_invariant_preserving([&](Alternative &alt) {
+    alt.cost += 1.0;
+    alt.cardinality = cardinality;
+    alt.enode_id = enode_id;
+  });
+  return result;
```
Ignition, line 262: This is not general, it special cases to handle the arity of the functions we have been testing (ie. `range(_,_)`). This requires code to be more general.

src/query/plan_v2/cost/cost_model.cpp
```diff
@@ -0,0 +1,478 @@
...
+  // Scope-barrier row-pipe.  Children layout is [outer, inner, exposed_sym...].
+  // Inner's introduces are STRIPPED at the barrier; only exposed_syms cross
+  // into the outer scope.  Importing-CALL is unsupported - guard up front so
+  // a failing query surfaces the real cause instead of a downstream
+  // "no self-contained alternative".
+  static auto cost(ENodeT const &n, ENodeId id, CostChildren children, CostCtx const &ctx) -> CostFrontier {
+    bool const has_self_contained_inner =
+        std::ranges::any_of(children[1]->alts(), [](Alternative const &a) { return a.required.empty(); });
+    if (!has_self_contained_inner) {
+      throw NotYetImplemented{"importing CALL subqueries"};
```
Ignition, line 371: What will it take to make CALL subquery complete? What is fundamental that need fixing first?

src/query/plan_v2/cost/cost_model.cpp
```diff
@@ -0,0 +1,478 @@
...
+// NOLINTEND(cppcoreguidelines-macro-usage,bugprone-macro-parentheses)
+
+}  // namespace
+
+auto ExtractOutputOwnSyms(planner::core::ENode<symbol> const &output_enode, EGraph const &egraph,
+                          VariableIndex const &idx) -> VariableSet {
+  VariableSet out;
+  for (auto const no_eclass : output_enode.children().subspan(1)) {
+    auto const &cls = egraph.eclass(no_eclass);
+    DMG_ASSERT(!cls.nodes().empty(), "NamedOutput e-class must have at least one enode");
```
Ignition: Can we assert there should only be 1. This way we can fix if this assumption breaks in the future.

src/query/interpreter.hpp
```diff
@@ -20,6 +20,7 @@
 #include "memory/db_arena_fwd.hpp"
 #include "query/context.hpp"
 #include "query/db_accessor.hpp"
+#include "query/plan_v2/frontend/egraph_converter.hpp"
```
Ignition: TODO: QueryPlannerContext coming from "frontend/egraph_converter.hpp" looks strange, change its header

tests/unit/CMakeLists.txt
```diff
@@ -845,11 +867,16 @@ if (MG_ENTERPRISE)
         LINK_TARGETS mg-auth mg-license Boost::filesystem
     )
 
-    add_custom_target(memgraph__unit__auth_kvstore_dir ALL
+    file(GLOB _auth_kvstore_files "${CMAKE_CURRENT_SOURCE_DIR}/auth_kvstore/*")
```
Ignition, line 883: Why not use the cmake function to do this stamp?

tests/unit/query_plan_v2_rewrites.cpp
```diff
@@ -255,5 +202,185 @@ TEST_F(InlineRewriteTest, FixedPoint) {
...
+// → Identifier should be merged with Literal(1)
+TEST_F(InlineThroughOperatorTest, InlineThroughAdd) {
+  auto once = eg_.MakeOnce();
+  auto sym = eg_.MakeSymbol(0, "a");
+  auto lit1 = eg_.MakeLiteral(storage::ExternalPropertyValue{1});
+  [[maybe_unused]] auto bind = eg_.MakeBind(once, sym, lit1);
+
+  auto ident = eg_.MakeIdentifier(sym);
+  auto lit2 = eg_.MakeLiteral(storage::ExternalPropertyValue{2});
+  [[maybe_unused]] auto add = eg_.MakeAdd(ident, lit2);
```
Ignition, line 267: These are not valuable tests, the InlineRewrite we know happens regardless of the expression that holds the Identifier expression. We can cull most of these tests.

tests/unit/query_plan_v2_algebra.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+TEST(BindAlgebra_IsAlive, SymNotInRequiredYieldsDead) {
+  auto required = MakeSet({1, 2, 3});
+  EXPECT_FALSE(required.test(4));
+}
+
+TEST(BindAlgebra_IsAlive, EmptyRequiredIsAlwaysDead) {
+  auto required = MakeSet({});
+  EXPECT_FALSE(required.test(1));
+}
```
Ignition: This is just testing around the VariableSet, no need to refer to IsAlive. Maybe its own test file.

tests/unit/query_plan_v2_algebra.cpp
```diff
@@ -0,0 +1,157 @@
...
+                    AliveRequiredCase{"ExprDemandOverlapsInputProducesUnion", {1, 2, 3}, 2, {3, 4}, {1, 3, 4}},
+                    AliveRequiredCase{"ExprDemandReintroducesSym", {1, 2}, 2, {2}, {1, 2}},
+                    AliveRequiredCase{"EmptyInputAndExprYieldsEmpty", {}, 2, {}, {}},
+                    AliveRequiredCase{"SymIsMinElementOfInput_NoDemands", {1, 2, 3}, 1, {}, {2, 3}},
+                    AliveRequiredCase{"SymIsMinElementOfInput_WithDemands", {1, 2, 3}, 1, {4}, {2, 3, 4}},
+                    AliveRequiredCase{"SymIsMaxElementOfInput_NoDemands", {1, 2, 3}, 3, {}, {1, 2}},
+                    AliveRequiredCase{"SymIsMaxElementOfInput_WithDemands", {1, 2, 3}, 3, {4}, {1, 2, 4}},
+                    AliveRequiredCase{"SymIsSoleElementOfInput_NoDemands", {2}, 2, {}, {}},
+                    AliveRequiredCase{"SymIsSoleElementOfInput_WithDemands", {2}, 2, {1, 3}, {1, 3}}),
+    AliveRequiredCaseName);
```
Ignition, line 111: We should have unit test that check difference_bit and set_union independently

tests/unit/query_plan_v2_algebra.cpp
```diff
@@ -0,0 +1,157 @@
...
+                                                          std::partial_ordering::greater},
+                                         AltDominanceCase{"LowerCardinalityDominates_Forward",
+                                                          MakeAlt(5.0, 6.0, {1}, 0),
+                                                          MakeAlt(5.0, 1000.0, {1}, 1),
+                                                          std::partial_ordering::greater},
+                                         AltDominanceCase{"LowerCardinalityDominates_Reverse",
+                                                          MakeAlt(5.0, 1000.0, {1}, 0),
+                                                          MakeAlt(5.0, 6.0, {1}, 1),
+                                                          std::partial_ordering::less}),
+                         AltDominanceCaseName);
```
Ignition, line 138: Can we test this properly with only 2 fields?

tests/unit/query_plan_v2_cardinality.cpp
```diff
@@ -0,0 +1,278 @@
...
+  auto b = IntLiteral(eg, 0);
+  EXPECT_DOUBLE_EQ(EstimateBuiltin(eg, eg.MakeFunction("range", {a, b}), {AsCoreId(a), AsCoreId(b)}), 0.0);
+}
+
+TEST(BuiltinEstimator, RangeWithParameterFallsBackToDefault) {
+  egraph eg;
+  auto a = IntLiteral(eg, 0);
+  auto b = eg.MakeParameterLookup(0);
+  EXPECT_DOUBLE_EQ(EstimateBuiltin(eg, eg.MakeFunction("range", {a, b}), {AsCoreId(a), AsCoreId(b)}),
+                   kDefaultRowEstimate);
```
Ignition: We are Estimating the list size not a cardinality of number of rows.

tests/unit/query_plan_v2_cardinality.cpp
```diff
@@ -0,0 +1,278 @@
...
+  egraph eg;
+  BuiltinEstimator estimator{eg};
+  auto const &core_eg = CoreOf(eg);
+  auto fn = eg.MakeFunction("dummy", {});
+  auto const fn_eclass = core_eg.find(EClassId{fn.value_of()});
+  auto const fn_enode_id = core_eg.eclass(fn_eclass).nodes()[0];
+  auto fn_enode = core_eg.get_enode(fn_enode_id);
+  fn_enode = planner::core::ENode{fn_enode.symbol(), fn_enode.children(), 0};
+  EXPECT_DOUBLE_EQ(estimator.Estimate(fn_enode, {}), kDefaultRowEstimate);
+}
```
Ignition: not needed

tests/unit/query_plan_v2_cardinality.cpp
```diff
@@ -0,0 +1,278 @@
...
+  EXPECT_GT(result.cost, 0.0);
+  EXPECT_LT(result.cost, 1e9);
+  EXPECT_DOUBLE_EQ(result.cardinality, 6.0);
+}
+
+// ============================================================================
+// Subquery scope barrier
+// ============================================================================
+
+TEST(SubqueryBarrier, InnerBindingsStripped) {
```
Ignition: Should this kind of test be classified under another unit test file?

tests/unit/query_plan_v2_pipeline.cpp
```diff
@@ -281,28 +414,441 @@ INSTANTIATE_TEST_SUITE_P(
...
+            .min_rewrites = 1,
+            .should_saturate = true,
+        }
+    ),
+    TestCaseName
+);
+
+// --- Expression operator pipeline round-trip tests ---
+
+INSTANTIATE_TEST_SUITE_P(
```
Ignition, line 453: We need to define what is an interesting threshold for tests. 
It MUST either:
- Capture a working optimisation
- Document a missing optimisation opportunity

src/planner/test/unittest__extractor.cpp
```diff
@@ -378,268 +384,1013 @@ TEST(Extract_Cost, FullyCyclicEClassInfiniteCost) {
...
+      {.cost = 3.0, .required = {}, .enode_id = ENodeId{2}},      // expensive, no demands
+  }};
+  ASSERT_EQ(frontier.alts().size(), 3) << "All three Pareto-incomparable alts should survive";
+}
 
-  ASSERT_EQ(result.size(), 3);
-  // Root should come first
-  ASSERT_EQ(result[0].first, root_class);
-  // Left and right can come in any order, but both should be after root
+TEST(ParetoFrontier_Prune, AllDominatedByOne) {
```
Ignition: TODO: review this whole file, there is too much, do all theses test hold important value?

imilinovic (approved):

PR #4211, opened 2026-05-29
feat: add slow and failed-query logger + add grammar to set session wide settings

imilinovic:
## What

Two new operational log streams, sharing the unified main log sink:

- `[slow-query]` at **WARN** — one line per successful query whose end-to-end
  user-visible time meets a configurable threshold, optionally followed by an
  indented EXPLAIN plan.
- `[failed-query]` at **ERROR** — one line per user-initiated query that throws.

Both gates are runtime-configurable globally and overridable per-session.

## Configuration

| gflag                   | setting key           | type    | default | persistent | session-overridable |
| ----------------------- | --------------------- | ------- | ------- | ---------- | ------------------- |
| `--log_min_duration_ms` | `log.min_duration_ms` | `int64` | `-1`    | yes        | yes                 |
| `--log_failed_queries`  | `log.failed_queries`  | `bool`  | `false` | yes        | yes                 |
| `--log_query_plan`      | `log.query_plan`      | `bool`  | `true`  | yes        | yes                 |

`log.min_duration_ms` semantics: `-1` = off, `0` = log every successful query,
`>0` = threshold in ms.

### Per-session API

```cypher
SET SESSION SETTING "log.min_duration_ms" TO "500"
RESET SESSION SETTING "log.min_duration_ms"
```

The overlay is a hard allow-list of the three keys above; anything else raises
`Setting "<key>" cannot be set per-session`. `log.level` was deliberately
rejected when the unified-trace stream landed and remains rejected here.

## Line format

```
[slow-query] duration_ms=<N> user=<u> db=<d> query="<q>"
PLAN:
  <indented plan tree, when log.query_plan is true>
[failed-query] user=<u> db=<d|<none>> error="<msg>" query="<q>"
```

Duration is the sum of `parsing_time + planning_time + plan_execution_time`
from the existing summary; commit and AFTER_COMMIT triggers are excluded. The
`query=` and `error=` fields are quoted and self-delimiting: `"` and `\` are
escaped, and `\n`/`\r`/`\t` become escape sequences so a field never spans log
lines. Only the slow-query `PLAN:` block is intentionally multi-line (indented
continuation). `user=` is the effective auth user (matching the trace stream's
`[user=]` tag), empty for unauthenticated sessions; `db=<none>` when no
database is selected.

## Known limitation: DDL timing

DDL queries (`CREATE INDEX`, `DROP INDEX`, constraints, etc.) do their work in
a `query_handler` lambda rather than a `PullPlan`, so they never populate
`plan_execution_time`. A long index build therefore reports only
parse + plan time and is effectively never picked up as "slow".

This is intentional and acceptable: the slow-query logger targets user
query latency (the read/write Cypher workload), not schema-maintenance
operations whose cost is already understood and tracked elsewhere. Regular
Cypher queries populate all three timing fields and are timed correctly.

## Performance

The common case (logging disabled) is gated on cheap atomic reads before any
logging string work. Plan rendering is lazy — armed at prepare time but only
executed when a query actually crosses the threshold, after which the line is
emitted post-commit (so a query that fails to commit is logged as failed, not
slow). The per-session overlay uses transparent hashing to avoid per-query
allocation. The one unconditional cost is a small per-query copy of the query
text, kept so it survives until the deferred emit (see the `query_string` TODO
in `Pull`); the disabled path does no further string work.


imilinovic:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Memgraph can now log slow successful queries (`[slow-query]` at WARNING, gated by `--log-min-duration-ms`, with an optional EXPLAIN plan) and failed queries (`[failed-query]` at ERROR, gated by `--log-failed-queries`); both gates are runtime-configurable globally and overridable per session via `SET SESSION SETTING`, letting operators surface latency outliers and errors without enabling full session trace. [#4211](https://github.com/memgraph/memgraph/pull/4211)
- [x] **[ Documentation PR link memgraph/documentation#1654 ]**

src/flags/run_time_configurable.cpp
```diff
@@ -183,6 +200,18 @@ std::atomic<const std::chrono::time_zone *> timezone_{nullptr};
...
+std::atomic<bool> log_query_plan_{true};
+
+memgraph::utils::Settings::ValidatorResult ValidInt64Str(std::string_view in) {
+  int64_t value{};
+  auto [ptr, ec] = std::from_chars(in.data(), in.data() + in.size(), value);
+  if (ec != std::errc{} || ptr != in.data() + in.size()) {
+    return std::unexpected{"Value must be an integer."};
+  }
+  return {};
+}
```
as51340, line 213: there is already a similar function ParseInt which essentially does the same thing

src/flags/run_time_configurable.cpp
```diff
@@ -610,4 +662,70 @@ void SnapshotPeriodicDetach(std::shared_ptr<utils::Observer<utils::SchedulerInte
...
+int64_t GetLogMinDurationMs() { return log_min_duration_ms_.load(std::memory_order_acquire); }
+
+bool GetLogFailedQueries() { return log_failed_queries_.load(std::memory_order_acquire); }
+
+bool GetLogQueryPlan() { return log_query_plan_.load(std::memory_order_acquire); }
+
+namespace {
+// Per-session-overridable settings; IsSessionSettable and ValidateSessionSettingValue
+// derive from it. GetEffective keeps its own per-key fallback (see below).
+enum class SessionSettingType { kInt64, kBool };
```
as51340: : uint8_t

src/flags/run_time_configurable.cpp
```diff
@@ -610,4 +662,70 @@ void SnapshotPeriodicDetach(std::shared_ptr<utils::Observer<utils::SchedulerInte
...
+
+struct SessionSettableSetting {
+  std::string_view key;
+  SessionSettingType type;
```
as51340: I think this is overly complex, just query by setting name, depending on the setting name, use that function in ValidateSessionSettingValue

src/query/frontend/ast/ast.hpp
```diff
@@ -4405,6 +4405,34 @@ class SessionTraceQuery : public memgraph::query::Query {
   friend class AstStorage;
 };
 
+class SessionSettingQuery : public memgraph::query::Query {
+ public:
+  static const utils::TypeInfo kType;
+
+  const utils::TypeInfo &GetTypeInfo() const override { return kType; }
+
+  enum class Action { SET_SETTING, RESET_SETTING };
```
as51340: uint8_t

src/flags/run_time_configurable.cpp
```diff
@@ -610,4 +662,70 @@ void SnapshotPeriodicDetach(std::shared_ptr<utils::Observer<utils::SchedulerInte
...
+template <>
+bool GetEffective<bool>(std::string_view key, const logging::SessionLogContext &ctx) {
+  // ValidateSessionSettingValue guarantees the stored value is canonical lowercase.
+  if (auto overlay = ctx.GetSetting(key); overlay.has_value()) {
+    return *overlay == "true";
+  }
+  if (key == kLogFailedQueriesKey) return GetLogFailedQueries();
+  if (key == kLogQueryPlanKey) return GetLogQueryPlan();
+  LOG_FATAL("GetEffective<bool> called for unregistered key");
+}
```
as51340: I would rather do get function for each setting than group them by type -> if someone adds a similar config, this needs to be changed -> easy to forget

as51340 (approved):

PR #4212, opened 2026-05-29
Feature (Light Edges / 4) - feat(utils): add standalone EpochTracker for epoch-based reclamation

andrejtonev:
Add src/utils/epoch_tracker.hpp, an epoch-based safe-drain reclamation
primitive (Acquire/Release readers; IsSafeToFree gated on a guard-epoch
snapshot). It is a header-only primitive with no callers yet — the
LightEdgeIterableGuard/EdgePin consumers and the storage members that use
it land in later PRs of the light-edge stack — so it is dead, unreachable
code regardless of the storage_light_edge flag.

Fixes applied on top of the original implementation:
  - Use __builtin_ffsll instead of __builtin_ffsl in ComputeLastDead.
    field is uint64_t; __builtin_ffsl takes long (32-bit on LLP64) and
    would truncate the upper 32 bits. No-op on LP64/ARM64; portability
    hardening only.
  - epoch_ uses acq_rel (Acquire fetch_add) / acquire (CurrentEpoch load),
    matching the SkipListGc reference (skip_list.hpp AllocateId). The id is a
    reclamation guard: a producer snapshotting CurrentEpoch() after a reader's
    Acquire() must observe a guard_epoch greater than that reader's id, or
    IsSafeToFree could free a still-live reader's object. relaxed left that
    producer<->reader ordering unenforced on weak-memory targets (ARM/POWER) — a
    latent UAF invisible under x86-TSO.
  - Isolate head_ on its own cache line (alignas(64)). head_ is RMW'd under
    lock_ on every block allocation; sharing epoch_'s line would let block-
    allocating writers ping-pong the line that the lock-free Acquire() hot path
    hammers. tail_/last_id_ ride along on head_'s line (lock_-only, no contention
    with the lock-free epoch_ path).
  - Document the teardown precondition: ~EpochTracker()/Clear() free all
    blocks without waiting for outstanding readers, so every Acquire()d id
    must be Release()d before destruction (otherwise UB).

Dead-prefix watermark cache: ComputeLastDead previously rescanned the whole
block list from tail_ on every IsSafeToFree call. The dead prefix is monotonic
(bits only ever set, never cleared until Clear()), so the scan resumes from the
block where the prefix last stalled. The cache fields (cached_last_dead_,
scan_resume_) are non-atomic and the scan is a multi-step read-modify, so
ComputeLastDead serialises its whole body under a dedicated spin lock
(cache_lock_) — IsSafeToFree/ComputeLastDead are therefore safe under concurrent
callers. This is required because the real consumer (the graveyard drain) is NOT
a single scanner: a second collector can refill the graveyard between one drain's
swap and completion, so two drains can call ComputeLastDead concurrently. Clear()
resets the cache (scan_resume_ would otherwise dangle across the block frees).

Batch drain support: add a DeadPrefixWatermark() snapshot accessor and a
two-arg IsSafeToFree(guard_epoch, watermark) overload so a caller freeing many
guard_epochs in one pass (the graveyard drain) snapshots the watermark once and
turns an O(entries * blocks) sweep into O(entries) + a single O(blocks) scan. The
dead prefix is monotonic, so the snapshot is a safe lower bound: an entry not yet
freeable under it is simply retried next drain, never freed early. The one-arg
IsSafeToFree now delegates to the overload.

Memory-bound documentation: correct the per-block figure to sizeof(Block) == 576
bytes (24 bytes of links/first_id padded to a 64-byte line + a 512-byte field[]
bitmap) and the worst-case retention to ~140 MB at 1e9 IDs. Blocks are never
reclaimed by design — safe block reclamation reintroduces exactly the removal
races this layout avoids — so the figure is guarded by a
static_assert(sizeof(Block) == 576) rather than chasing reclamation.

The author's TODO about unifying this with the skiplist GC allocator is
kept verbatim as a deferred follow-up; src/utils/skip_list.hpp is left
byte-for-byte unchanged.

Add a self-proving unit test (tests/unit/utils_epoch_tracker.cpp) covering
Acquire/Release id monotonicity, contiguous-dead-prefix reclamation, the 64-id
field boundary, multi-block allocation across the 4096-id block boundary, the
safe teardown precondition, the watermark cache (WatermarkCacheAdvancesAndStalls,
ClearResetsWatermarkCache), and concurrent stress tests
(SafetyContractUnderConcurrency, ConcurrentReleaseRacesBlockAllocation,
AcquireReleaseChurn, and ConcurrentScannersShareWatermarkCache — multiple
IsSafeToFree scanners against live churn, a TSan regression for the cache_lock_
serialisation).

### DIFF
--- a/src/utils/epoch_tracker.hpp
+++ b/src/utils/epoch_tracker.hpp
@@ -1,0 +1,353 @@
+// Copyright 2026 Memgraph Ltd.
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
+// License, and you may not use this file except in compliance with the Business Source License.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#pragma once
+
+#include <array>
+#include <atomic>
+#include <bit>
+#include <cstdint>
+#include <limits>
+#include <new>
+
+#include "utils/logging.hpp"
+#include "utils/rw_spin_lock.hpp"
+#include "utils/spin_lock.hpp"
+
+namespace memgraph::utils {
+
+/// Epoch-based safe-drain tracker for deferred reclamation.
+///
+/// Readers call Acquire() on entry and Release(id) on exit.  A producer
+/// records `guard_epoch = CurrentEpoch()` when an object becomes eligible for
+/// reclamation.  The object may be safely freed once IsSafeToFree(guard_epoch)
+/// returns true — meaning every reader that was alive at the time of the
+/// guard_epoch snapshot has since called Release().
+///
+/// The implementation is a simplified variant of the SkipListGc bitmap
+/// approach: accessor IDs are packed into 64-bit fields inside fixed-size
+/// Blocks linked in a doubly-linked list.  Unlike SkipListGc, fully-dead
+/// blocks are never reclaimed, so the implementation is free of the tricky
+/// races that accompany block removal.
+///
+/// Memory bound: each block holds 4096 IDs and costs sizeof(Block) == 576 bytes
+/// (24 bytes of links + first_id, padded out to a 64-byte cache line, plus a
+/// 512-byte field[] bitmap).  Blocks are never reclaimed (see above), so a server
+/// processing 1 billion unique edge-index scans will retain ~140 MB of tracker
+/// memory for the lifetime of the storage instance.  If this becomes a concern
+/// in practice, safe block reclamation would have to be added — but that
+/// reintroduces exactly the block-removal races this design deliberately avoids,
+/// so it is intentionally left out (TODO only if the bound is ever hit).
+///
+/// Teardown precondition: ~EpochTracker() / Clear() free all Blocks WITHOUT
+/// waiting for outstanding readers. The caller MUST guarantee that every id
+/// handed out by Acquire() has been Release()d (equivalently: no reader holds a
+/// live id) before the tracker is destroyed or Clear()ed. Destroying with live
+/// readers is undefined behaviour (a subsequent Release() would touch freed
+/// memory).
+///
+/// Dead-prefix watermark cache (amortized rescan):
+///   The dead prefix is monotonic — bits in field[] are only ever set (never
+///   cleared) until Clear().  ComputeLastDead() caches (cached_last_dead_,
+///   scan_resume_) so subsequent calls resume from the stall point instead of
+///   re-walking the entire block list from tail_.  These cache fields are
+///   non-atomic and the scan is a multi-step read-modify, so ComputeLastDead()
+///   serialises its whole body under cache_lock_ — IsSafeToFree/ComputeLastDead
+///   are therefore safe to call from multiple threads.  This matters because the
+///   only caller, the light-edge graveyard drain (DrainLightEdgeGraveyard), is
+///   NOT a single dedicated thread: FreeMemory reaches it from the periodic GC
+///   runner, FREE MEMORY queries, and SetStorageMode, and those drains are not
+///   mutually exclusive (a second collector can refill the graveyard between one
+///   drain's swap and completion), so two drains can scan concurrently.
+
+// TODO This is just the skiplist allocator's epoch tracker. Can we move the skiplist's out and use this there as well?
+// Audit to make sure the logic is EXACTLY the same.
+class EpochTracker {
+ private:
+  static constexpr uint64_t kIdsInField = sizeof(uint64_t) * 8;        // 64 bits per field
+  static constexpr uint64_t kBlockFields = 64;                         // 64 fields per block
+  static constexpr uint64_t kIdsInBlock = kBlockFields * kIdsInField;  // 4096 IDs per block
+
+  struct Block {
+    std::atomic<Block *> prev{nullptr};
+    std::atomic<Block *> succ{nullptr};
+    uint64_t const first_id;  // immutable after construction; safe to read without lock
+    // Placed on its own cache line to prevent false sharing with prev/succ/first_id.
+    alignas(64) std::array<std::atomic<uint64_t>, kBlockFields> field;
+
+    explicit Block(uint64_t id) : first_id{id}, field{} {}
+  };
+
+  // Guards the memory-bound figure documented above (576 bytes/block, ~140 MB at
+  // 1e9 IDs). If the layout changes, update the class doc accordingly.
+  static_assert(sizeof(Block) == 576, "EpochTracker::Block layout changed; revisit the memory-bound doc");
+
+  Block *AllocateBlock(Block *expected_head) noexcept {
+    auto guard = std::lock_guard{lock_};
+    // relaxed: the lock provides the necessary acquire/release ordering here.
+    Block *curr_head = head_.load(std::memory_order_relaxed);
+    if (curr_head != expected_head) return curr_head;  // someone else allocated while we waited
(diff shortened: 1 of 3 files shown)


andrejtonev:

---
___

### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist

as51340 (changes requested):
Great stuff!

1. I would like to see concurrent test with multiple threads doing something on top of EpochTracker
2. Use std::numeric_limits<uint64_t> instead of -1
3. I think you could cache last_dead_ so you don't need to start every time from the tracker's tail to speed up ComputeLastDead

src/utils/epoch_tracker.hpp
```diff
@@ -0,0 +1,339 @@
...
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#pragma once
+
+#include <atomic>
+#include <bit>
+#include <cstdint>
+#include <cstring>
```
as51340: Shouldn't be needed

src/utils/epoch_tracker.hpp
```diff
@@ -0,0 +1,339 @@
...
+  static constexpr uint64_t kIdsInField = sizeof(uint64_t) * 8;        // 64 bits per field
+  static constexpr uint64_t kBlockFields = 64;                         // 64 fields per block
+  static constexpr uint64_t kIdsInBlock = kBlockFields * kIdsInField;  // 4096 IDs per block
+
+  struct Block {
+    std::atomic<Block *> prev{nullptr};
+    std::atomic<Block *> succ{nullptr};
+    uint64_t const first_id;  // immutable after construction; safe to read without lock
+    // Placed on its own cache line to prevent false sharing with prev/succ/first_id.
+    alignas(64) std::atomic<uint64_t> field[kBlockFields];
```
as51340: Use std::array instead

src/utils/epoch_tracker.hpp
```diff
@@ -0,0 +1,339 @@
...
+  /// A caller freeing many guard_epochs in one pass (e.g. the graveyard drain)
+  /// should snapshot once and pass it to the two-arg overload below, turning an
+  /// O(entries * blocks) sweep into O(entries) + a single O(blocks) scan. The
+  /// dead prefix is monotonic, so the snapshot is a safe lower bound: an entry
+  /// not yet freeable under it is simply retried on the next drain, never freed
+  /// early.
+  uint64_t DeadPrefixWatermark() const noexcept { return ComputeLastDead(); }
+
+  /// O(1) IsSafeToFree against a watermark obtained from DeadPrefixWatermark().
+  bool IsSafeToFree(uint64_t guard_epoch, uint64_t dead_prefix_watermark) const noexcept {
```
as51340: should be static method

src/utils/epoch_tracker.hpp
```diff
@@ -0,0 +1,339 @@
...
+
+    explicit Block(uint64_t id) : first_id{id}, field{} {}
+  };
+
+  // Guards the memory-bound figure documented above (576 bytes/block, ~140 MB at
+  // 1e9 IDs). If the layout changes, update the class doc accordingly.
+  static_assert(sizeof(Block) == 576, "EpochTracker::Block layout changed; revisit the memory-bound doc");
+
+  Block *AllocateBlock(Block *expected_head) {
+    auto guard = std::lock_guard{lock_};
```
as51340, line 94: Why do you need lock here?

src/utils/epoch_tracker.hpp
```diff
@@ -0,0 +1,339 @@
...
+  // both fields, so that whole method is serialised under cache_lock_ (atomics
+  // alone would not make the RMW atomic). The earlier assumption that the
+  // graveyard-swap made the drain a single scanner is FALSE — a second collector
+  // can refill the graveyard between one drain's swap and completion, so two
+  // drains can call ComputeLastDead concurrently; cache_lock_ closes that race.
+  //
+  // Lifecycle: initialized to "no cache" state; updated by ComputeLastDead();
+  // MUST be reset to "no cache" state in Clear() because scan_resume_ points
+  // into a Block that Clear() deallocates (dangling pointer otherwise).
+  mutable uint64_t cached_last_dead_{kNoDeadPrefix};
```
as51340, line 349: I don't see where you read from cached_last_dead_?

as51340: same for scan_resume_

src/utils/epoch_tracker.hpp
```diff
@@ -0,0 +1,339 @@
...
+  /// Record this as the guard_epoch when an object becomes reclaimable.
+  uint64_t CurrentEpoch() const noexcept {
+    // acquire: pairs with Acquire()'s acq_rel fetch_add so a guard_epoch recorded
+    // here observes every reader id allocated before this snapshot (see Acquire).
+    return epoch_.load(std::memory_order_acquire);
+  }
+
+  /// Returns true if all readers with id < guard_epoch have called Release().
+  /// Thread-safe: ComputeLastDead() serialises the cache RMW under cache_lock_.
+  bool IsSafeToFree(uint64_t guard_epoch) const noexcept { return IsSafeToFree(guard_epoch, ComputeLastDead()); }
```
as51340: Where is this used?

src/utils/epoch_tracker.hpp
```diff
@@ -0,0 +1,339 @@
...
+
+  /// Snapshot of the dead-prefix watermark, for batch IsSafeToFree() checks.
+  /// A caller freeing many guard_epochs in one pass (e.g. the graveyard drain)
+  /// should snapshot once and pass it to the two-arg overload below, turning an
+  /// O(entries * blocks) sweep into O(entries) + a single O(blocks) scan. The
+  /// dead prefix is monotonic, so the snapshot is a safe lower bound: an entry
+  /// not yet freeable under it is simply retried on the next drain, never freed
+  /// early.
+  uint64_t DeadPrefixWatermark() const noexcept { return ComputeLastDead(); }
```
as51340: I don't see where this is used?

src/utils/epoch_tracker.hpp
```diff
@@ -0,0 +1,339 @@
...
+        auto prev_val = head->field[field_idx].fetch_or(mask, std::memory_order_release);
+        MG_ASSERT(!(prev_val & mask), "EpochTracker: id {} released twice", id);
+        return;
+      }
+    }
+  }
+
+  /// Returns the epoch value that will be returned by the next Acquire() call.
+  /// Record this as the guard_epoch when an object becomes reclaimable.
+  uint64_t CurrentEpoch() const noexcept {
```
as51340, line 172: Where is this used?

as51340 (changes requested):

src/utils/epoch_tracker.hpp
andrejtonev: ComputeLastDead reads them at the start to update the local last_dead

src/utils/epoch_tracker.hpp
andrejtonev: currently in tests only. this is setting up the epoch tracker that will be used in future prs.

src/utils/epoch_tracker.hpp
andrejtonev: same, future prs will use it.

src/utils/epoch_tracker.hpp
andrejtonev: later pr

src/utils/epoch_tracker.hpp
andrejtonev: this is the same logic as the skiplist uses for its reclamation. I'm just extracting, so I can use the same thing for light edges.

as51340 (approved):

Issue #4216, opened 2026-06-01
Add CRC32-per-transaction checksums to WAL files (verified on recovery and replica replay)


as51340:
## Problem Statement

As a Memgraph operator, I have no way to detect whether the data inside a WAL transaction has been silently corrupted (bit rot on disk, a bad block, a faulty transfer) before that transaction is applied. Today a corrupted-but-structurally-parseable transaction will be applied as-is — during local crash recovery and when a replica replays WAL files received via `CurrentWalRpc`/`WalFilesRpc` — which can silently propagate wrong data into the in-memory graph and onward to replicas.

## Solution

Add a CRC32 checksum (computed with zlib) per transaction in WAL files. Each committed transaction in a WAL file carries a checksum covering its full byte span. On read, both local crash recovery and replicas verify the checksum per transaction and refuse to apply a transaction whose bytes do not match. This turns silent corruption into a loud, detectable failure: recovery refuses to start on a corrupt WAL, and a replica rejects a corrupt transferred WAL non-fatally so the main retries.

Snapshots are out of scope. No feature flag is added for now — the expected overhead is small and will be measured with benchmarks later.

## User Stories

1. As a Memgraph operator, I want each WAL transaction to carry a CRC32 checksum, so that silent on-disk corruption can be detected rather than applied blindly.
2. As a Memgraph operator, I want local crash recovery to verify the checksum of every transaction it reads, so that a corrupted WAL does not produce a silently-wrong in-memory graph.
3. As a Memgraph operator, I want recovery to refuse to start (fatal) when it detects a checksum mismatch, so that a node never serves data it cannot vouch for.
4. As a Memgraph operator running replication, I want a replica receiving WAL files via `WalFilesRpc` to verify each transaction before committing it, so that corruption introduced in transit or on the replica disk is caught.
5. As a Memgraph operator running replication, I want a replica receiving the current WAL via `CurrentWalRpc` to verify each transaction before committing it, so that catch-up recovery cannot apply corrupt data.
6. As a Memgraph operator, I want a replica that detects a checksum mismatch to fail non-fatally and let the main retry, so that a transient transfer corruption self-heals without crashing the replica.
7. As a Memgraph operator, I want transactions that were already committed on the replica before a mismatch to remain committed, so that a single bad transaction does not roll back valid progress.
[...]

PR #4225, opened 2026-06-02
feat: WAL CRC32 per transaction

### LINKED ISSUE #4218: WAL CRC32-per-txn format plumbing (write + skip-on-read, no verification)
## Parent

memgraph/memgraph#4216

## What to build

Add the CRC32-per-transaction field to the WAL file format end-to-end, with no rejection behavior yet — this slice establishes the format plumbing so later slices can verify against it. After this slice, every WAL transaction written at v36 carries a checksum, all readers consume it without desyncing, and pre-v36 WALs still load.

Scope:
- A small, stable `CrcAccumulator` deep module wrapping zlib's incremental `crc32(crc, buf, len)` API: `reset()`, `update(bytes, len)`, `value() -> uint32`. Embedded by the WAL encoder and decoder.
- Running checksum accumulated in the low-level WAL `Encoder::Write()`. Reset at `AppendTransactionStart` (before the `SECTION_DELTA` of `TRANSACTION_START` is written) and the finalized value written as a raw 4-byte little-endian `uint32` immediately after the `DELTA_TRANSACTION_END` delta. Coverage span is `[SECTION_DELTA of TRANSACTION_START, end of TRANSACTION_END)`, inclusive of the commit-flag byte; the checksum field itself is excluded. Reset-at-start naturally excludes file headers/metadata sections.
- Every reader that walks the delta stream — local recovery (`durability::LoadWal`), the WAL info skip-scan (`ReadWalInfo`), and the replica apply loop (`ReadAndApplyDeltasSingleTxn`) — consumes the 4-byte field, gated on `version >= 36`, without counting it as a delta (so delta counts and section offsets stay aligned). No verification/rejection in this slice.
- The live `PrepareCommitRpc` network delta stream is unchanged (it carries no checksum field).
- Regenerate the v36 WAL integration fixtures so they include the checksum.

## Acceptance criteria

- [ ] `CrcAccumulator` exists with `reset`/`update`/`value` and matches zlib `crc32` output on known vectors; incremental updates equal a one-shot computation.
- [ ] WAL transactions written at v36 carry a 4-byte LE `uint32` checksum immediately after `TRANSACTION_END`, covering the full transaction span.
- [ ] A clean multi-transaction WAL (data-only, DDL/metadata, and an empty `START`/`END` transaction) round-trips and all checksum fields are present and consumed.
- [ ] Backward-compat: a pre-v36 WAL (no checksum field) loads correctly with the field read skipped via the version gate.
- [ ] `ReadWalInfo` reports correct `num_deltas` and section offsets for v36 WALs (the checksum field is not counted as a delta).
- [ ] The `PrepareCommitRpc` path is unaffected.
- [ ] v36 WAL integration fixtures are regenerated to include the checksum.

## Blocked by

- memgraph/memgraph#4217

### LINKED ISSUE #4219: Verify WAL transaction CRC32 on local crash recovery (fatal on mismatch)
## Parent

memgraph/memgraph#4216

## What to build

Make local crash recovery verify each transaction's CRC32 and refuse to start on a mismatch, turning silent on-disk WAL corruption into a loud, fatal failure.

Scope:
- Running checksum accumulated in the low-level WAL `Decoder::Read()`, reset per transaction.
- In `durability::LoadWal`, at `TRANSACTION_END`: snapshot the running checksum, read the trailing 4-byte stored checksum (read after the snapshot so it is excluded from accumulation), and compare. On mismatch, throw `RecoveryFailure`, which the recovery caller already turns into `LOG_FATAL`.
- Apply-as-you-go is acceptable here: the fatal crash discards the corrupt in-memory state, so it is never served. No tail-truncation special case.

## Acceptance criteria

- [ ] A clean v36 WAL recovers successfully with all transaction checksums validated.
- [ ] Flipping a single byte inside a transaction payload causes recovery to throw `RecoveryFailure` (fatal), rather than applying the corrupt transaction.
- [ ] A byte-flip fuzz/property test over every byte position of a transaction asserts detection at each position.
- [ ] Pre-v36 WALs continue to recover with verification skipped (no regression from #4218).

## Blocked by

- memgraph/memgraph#4218

### LINKED ISSUE #4220: Verify WAL transaction CRC32 on replica replay (WalFilesRpc/CurrentWalRpc), non-fatal retry
## Parent

memgraph/memgraph#4216

## What to build

Make a replica verify each transaction's CRC32 before committing it when replaying WAL files received via `WalFilesRpc` or `CurrentWalRpc`, and reject a corrupt transaction non-fatally so the main retries.

Scope:
- In `ReadAndApplyDeltasSingleTxn` (gated on `loading_wal && version >= 36`): after reading the `TRANSACTION_END` delta but before the transaction-end handler runs `PrepareForCommitPhase`, snapshot the decoder's running checksum, read the trailing 4-byte stored checksum, and compare. On mismatch, return `std::nullopt` so the transaction is never committed.
- The mismatch surfaces as `LoadWal` `success=false` — a non-fatal condition that returns the un-advanced `last_durable_timestamp` to the main, which retries. Transactions committed earlier in the same file remain committed.
- The live `PrepareCommitRpc` network path (no checksum field) is left unchanged.

## Acceptance criteria

- [ ] A replica replaying a clean transferred WAL commits all transactions as today (existing replication suites pass unchanged).
- [ ] A transaction with a corrupted payload in a transferred WAL is not committed; the replica reports failure non-fatally and the main retries.
- [ ] Transactions committed before the corrupt one in the same WAL file stay committed.
- [ ] An end-to-end replication test corrupts a transferred WAL file and asserts non-fatal rejection + main retry.
- [ ] The `PrepareCommitRpc` live path is unaffected.

## Blocked by

- memgraph/memgraph#4218

### LINKED ISSUE #4221: Recompute WAL transaction CRC32 on STRICT_SYNC (2PC) commit-flag flip
## Parent

memgraph/memgraph#4216

## What to build

Keep the per-transaction CRC32 correct for STRICT_SYNC (2PC) transactions, whose commit-flag byte is flipped after the transaction (including `TRANSACTION_END` and its checksum) has already been written.

Scope:
- Track each transaction's `txn_start_pos` and `crc_field_pos` alongside the existing commit-flag position.
- In `UpdateCommitStatus` (called from `FinalizeCommitPhase` while `engine_lock_` is held, so the transaction is still the stable tail of the WAL): after flipping and flushing the commit flag, recompute the checksum by `pread`-ing the transaction's byte span `[txn_start_pos, crc_field_pos)` and rewrite the 4-byte checksum field at `crc_field_pos`.
- A 2PC abort leaves the commit flag false and requires no recompute — the originally written checksum (computed over the `commit=false` bytes) remains valid.

HITL — concurrency review required: confirm the flag-flip-then-checksum-rewrite is not observable mid-update by a concurrent `CurrentWalRpc` stream read (which reads the current file buffer). `engine_lock_` is held during the rewrite; verify and document whether the stream read path is serialized against it, and address if not.

## Acceptance criteria

- [ ] For a 2PC commit (flag flipped false->true), the rewritten checksum validates against the final on-disk bytes (including the flipped flag).
- [ ] For a 2PC abort (flag stays false), the original checksum validates with no recompute.
- [ ] Unit tests cover both the 2PC commit-recompute and abort paths.
- [ ] The concurrency interaction between `UpdateCommitStatus`'s rewrite and a concurrent `CurrentWalRpc` stream read is reviewed and documented (and fixed if a mid-update read is possible).

## Blocked by

- memgraph/memgraph#4218

as51340:
Protects WAL file header with a checksum and each transaction with its own 4 bytes checksum.
For strict sync replication where commit decision is overriden, we use crc32_combine with no performance penalties to overwrite the previous commit vote.
Deltas streamed for PrepareCommitRpc (a.k.a commit) are not protected with a checksum, we are using TCP network protocol which should give us enough guarantees to consider deltas correct.
WAL files sent to replica are checksummed and later on replica check because we are buffering them on disk.
To be as efficient as possible, we are relying on InputFile's buffer and calculate a checksum once we fill the buffer or when the need for writing the transaction comes.

Snapshots are not checksummed yet, the change will come later.


as51340:
### Tracking
- [x] **[Link to Epic/Issue](https://github.com/memgraph/memgraph/issues/4216)**

### Standard development
- [x] Update unit/E2E tests
- [] Compare the [benchmarking results](https://bench-graph.memgraph.com/) between the master branch this branch

### CI Testing Labels

### Documentation checklist
    - WAL files are now protected with a checksum and won't be loaded in the case of a mismatch. The 4 bytes checksum is added for WAL file init section and for each transaction. The user shouldn't see any differences in the performance or behaviour. [#4225](https://github.com/memgraph/memgraph/pull/4225)
- [x] **[ Documentation PR link ](https://github.com/memgraph/documentation/pull/1669)**

src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -1974,6 +1973,15 @@ std::optional<storage::SingleTxnDeltasProcessingResult> InMemoryReplicationHandl
         },
     };
 
+    if (auto const *txn_end = std::get_if<WalTransactionEnd>(&delta.data_)) {
+      // We don't do CRC verification on PrepareCommitRpc because we are already using TCP sockets
+      if (loading_wal && txn_end->txn_crc.has_value() && !utils::CrcAccumulator::Verify(decoder->CrcAccValue())) {
+        throw utils::BasicException(
```
andrejtonev: this code is outside the apply delta, so is not covered by the same try/catch and will not return nullopt on error, but will throw and be caught at a higher level.
1. we should return error like with other deltas
2. why have this outside the delta_apply visitor?

src/storage/v2/durability/wal.cpp
```diff
@@ -1205,30 +1238,29 @@ auto convert_to_transaction_access_type(StorageAccessType access_type) -> Transa
...
-  encoder->WriteMarker(Marker::DELTA_TRANSACTION_START);
-  encoder->WriteBool(commit);
-  encoder->WriteUint(static_cast<uint8_t>(convert_to_transaction_access_type(access_type)));
-}
-
-void EncodeTransactionEnd(BaseEncoder *encoder, uint64_t timestamp) {
+WalTxnEndPos EncodeTransactionEnd(BaseEncoder *encoder, uint64_t timestamp) {
+  // Capture the CRC of everything accumulated since the matching EncodeTransactionStart (txn start + deltas) *before*
+  // writing the transaction-end frame, so the stored value matches what the reader can recompute. The reader snapshots
+  // its accumulator before consuming the transaction-end frame, hence the frame markers must be excluded here.
```
andrejtonev: Is this comment true? Don't see where we skip the marker.

src/utils/file.cpp
```diff
@@ -267,6 +289,10 @@ size_t InputFile::GetPosition() {
 }
 
 std::optional<size_t> InputFile::SetPosition(Position position, ssize_t offset) {
+  // Settle the CRC over the bytes consumed so far while they are still available in the buffer; bytes skipped over by
+  // the seek itself are never folded.
+  FoldPendingCrc();
```
andrejtonev: what is this trying to solve? 
If we are moving around the file, we need some (probably complex) way of keeping track of what has been crced and where we are and what should be added to the crc in the current section. I think snapshot jumps around a lot, so will be important later on.

src/storage/v2/durability/serialization.cpp
```diff
@@ -343,6 +362,10 @@ std::optional<uint64_t> ReadSize(Decoder *decoder) {
   uint64_t size;
   if (!decoder->Read(reinterpret_cast<uint8_t *>(&size), sizeof(size))) return std::nullopt;
   size = utils::LittleEndianToHost(size);
+  // Bound check
+  if (auto const file_size = decoder->GetSize(); !file_size || size >= *file_size) {
```
andrejtonev: not sure why decoder returns optional, since the file returns size_t.

src/storage/v2/durability/wal.hpp
```diff
@@ -490,17 +491,27 @@ void EncodeDelta(BaseEncoder *encoder, Storage *storage, SalientConfig::Items it
 void EncodeDelta(BaseEncoder *encoder, Storage *storage, const Delta &delta, Edge *edge, uint64_t timestamp,
                  Gid in_vertex_gid, EdgeTypeId edge_type_id);
 
-/// Function used to encode the transaction start
-/// Returns the position in the WAL where the flag 'commit' is about to be written
-uint64_t EncodeTransactionStart(Encoder<utils::OutputFile> *encoder, uint64_t timestamp, bool commit,
-                                StorageAccessType access_type);
+struct WalTxnDataPos {
+  uint64_t commit_flag_wal_position_{0};  // position of the commit flag inside the transaction-start frame
+  uint64_t txn_end_wal_pos_{0};           // one-past-last byte of the CRC-protected region (start of the txn-end frame)
```
andrejtonev: where is txn_end_wal_pos_ used?

src/storage/v2/durability/wal.cpp
```diff
@@ -953,19 +965,40 @@ WalInfo ReadWalInfo(const std::filesystem::path &path) {
...
+          spdlog::error("Durability CRC mismatch");
+          return std::nullopt;
+        }
+        wal.ResetCrcAcc();
+      }
+
       return {{timestamp, is_transaction_end}};
     } catch (const RecoveryFailure &e) {
-      spdlog::error("Error occurred while reading WAL info: {}", e.what());
+      spdlog::info("Error occurred while reading WAL info: {}", e.what());
```
andrejtonev: why lower it to info?

src/storage/v2/durability/wal.cpp
```diff
@@ -2050,21 +2098,44 @@ void WalFile::AppendDelta(const Delta &delta, Edge *edge, uint64_t timestamp, St
...
+  // The TYPE_BOOL marker precedes the actual value byte, which is the only byte that changes.
+  auto const changed_byte_pos = wal_positions.commit_flag_wal_position_ + sizeof(Marker);
+  // The stored CRC covers everything up to and including the TYPE_INT marker that introduces the CRC trailer;
+  // crc_wal_pos_ points at that marker. bytes_after = bytes following the flipped value byte within that region.
+  auto const bytes_after = wal_positions.crc_wal_pos_ - changed_byte_pos;
+  auto const new_crc = utils::CrcAccumulator::PatchByte(wal_positions.stored_crc_, t_delta, bytes_after);
+
+  // Overwrite the stored CRC value in place.
+  wal_.SetPosition(wal_positions.crc_wal_pos_);
+  wal_.WriteUint(new_crc);
```
andrejtonev: could be WriteCrcAt()

src/storage/v2/durability/wal.cpp
```diff
@@ -1205,30 +1238,29 @@ auto convert_to_transaction_access_type(StorageAccessType access_type) -> Transa
...
+  auto const txn_end_wal_pos = encoder->GetPosition();  // end of the CRC-protected region
   encoder->WriteMarker(Marker::SECTION_DELTA);
   encoder->WriteUint(timestamp);
   encoder->WriteMarker(Marker::DELTA_TRANSACTION_END);
+  auto const crc_wal_pos = encoder->GetPosition();  // position where the CRC value is stored
+  auto const txn_crc = encoder->WriteCrc();
+  return {.txn_end_wal_pos_ = txn_end_wal_pos, .crc_wal_pos_ = crc_wal_pos, .stored_crc_ = txn_crc};
 }
 
 std::optional<RecoveryInfo> LoadWal(
```
andrejtonev, line 1261: Document that the crc verification is done in read wal info and is not needed later on. 
If the verification is not expensive, we could verify again in the apply loop as well.

src/utils/crc_accumulator.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+class CrcAccumulator {
+ public:
+  CrcAccumulator();
+
+  ~CrcAccumulator() = default;
+  CrcAccumulator(CrcAccumulator const &) = default;
+  CrcAccumulator &operator=(CrcAccumulator const &) = default;
+  CrcAccumulator(CrcAccumulator &&) = default;
+  CrcAccumulator &operator=(CrcAccumulator &&) = default;
```
andrejtonev: since all are default, you can just remove all of them

src/utils/crc_accumulator.cpp
```diff
@@ -0,0 +1,48 @@
...
+  return static_cast<uint32_t>(crc32_combine(crc_a, crc_b, static_cast<z_off_t>(len_b)));
+}
+
+auto CrcAccumulator::Verify(uint32_t const crc) -> bool {
+  // `crc` must be the running CRC after every byte UP TO AND INCLUDING the stored CRC trailer has been folded in.
+  // CRC-32 is self-checking: feeding a block followed by its own CRC reduces to a fixed residue. Because the trailer is
+  // stored as a uint64 (TYPE_INT marker + 8 little-endian bytes, with the 32-bit CRC in the low 4 and zero padding in
+  // the high 4 -- matching how WriteUint/ReadUint move a uint), the residue is the 8-byte-append value below, not the
+  // textbook 4-byte 0x2144DF1C. zlib's CRC-32 carries init/xorout = 0xFFFFFFFF, which is what shifts it off zero.
+  constexpr uint32_t kCrc32Residue = 0x6522DF69U;
```
andrejtonev, line 44: Ideally we have some compile time verification of this number. not sure how.

src/storage/v2/durability/wal.cpp
as51340: was true before, will remove it now

src/storage/v2/durability/wal.cpp
as51340: I will document it

src/storage/v2/durability/wal.cpp
as51340: debugging purposes, will revert

src/storage/v2/durability/wal.hpp
as51340: left from all experiments, forgot to remove it.

src/storage/v2/durability/serialization.cpp
as51340: fixed, this was from before not sure why

src/dbms/inmemory/replication_handlers.cpp
as51340: agree, this was again because of some old experiments and I forgot to change it

src/utils/file.cpp
as51340: It was trying to solve SetPosition(offset_metadata) in ReadWalInfo.
Went looking and realised we don't even need SetPosition(offset_metadata) because we are decoding in order. Added TODO for implementing CRC for SetPosition since it will be used in snapshots

src/utils/crc_accumulator.cpp
as51340: I agree but the only idea I have is to manually compute this in compile-time and I don't think this is good. As long as zlib's version doesn't change this should be fine. When it changes, we will see tests failing

andrejtonev: is there a base number zlib uses that we can static assert is the same as when we calculated this?
same goes for the size of the added fields

as51340: no

andrejtonev (approved):

PR #4229, opened 2026-06-03
feat: show garbage collector in SHOW TRANSACTIONS

imilinovic:
## Overview

`SHOW TRANSACTIONS` now reports an in-progress storage **garbage collection** as a synthetic transaction, the same way a running periodic snapshot already appears. This gives operators visibility into GC activity that was previously invisible.

## Motivation

Garbage collection runs continuously in the background. When it is healthy it completes in well under a millisecond and nobody needs to think about it. But when GC becomes a problem — a large backlog, very long version chains, or aggressive mode where it holds the storage lock and blocks all transactions — there was no way to observe it from a query. Operators had to infer a misbehaving GC indirectly.

This feature makes a slow, stuck, or blocking GC **directly observable**: it shows up in `SHOW TRANSACTIONS` exactly when it is running long enough to matter.

## Behavior

While a GC is actively running, `SHOW TRANSACTIONS` includes an extra row alongside the real client transactions:

| Column | Value |
|---|---|
| `username` | *(empty)* |
| `transaction_id` | `gc` |
| `query` | `["GARBAGE COLLECTION"]` |
| `status` | `running` |
| `start_time` | when this GC run started |
| `elapsed_ms` | how long it has been running — the key signal for a stuck GC |
| `metadata` | see below |

The `metadata` map carries the diagnostic detail:

- **`db_name`** — the database the GC is running on.
- **`phase`** — which stage of the GC cycle is currently running (see below). Helps answer "stuck where?".
- **`trigger`** — `periodic` (the background cycle) or `forced` (a `FREE MEMORY` query or a storage-mode switch).
- **`exclusive_lock`** — `true` when GC is holding the storage lock exclusively and therefore blocking all transactions. The clearest "GC is freezing the database" signal.

### GC phases

A GC run moves through these stages in order; the `phase` value tells you which one it is in right now:

- **`unlink`** — GC walks the obsolete record versions left behind by committed/aborted transactions and detaches the ones no active transaction can still see. This is also where the initial bookkeeping for the run happens.
- **`index_cleanup`** — stale entries for deleted objects are removed from indexes and constraints. This is typically the **most expensive** stage, since it scans index structures; a GC that appears stuck is most often sitting here.
- **`delete`** — the final reclamation: the unlinked versions are freed and deleted vertices/edges are removed from storage.

### Notes

- One `gc` row per database that currently has a GC running; multiple databases can each show their own.
- The row appears only while GC is actively working. A healthy fast GC will rarely be caught — that is expected; the value is in catching a pathological one.
- No change to the result schema: the existing 7 columns are reused, exactly like the snapshot row. Existing clients and drivers are unaffected.
- During a `FREE MEMORY` query, both the user's own transaction and the synthetic `gc` row may be visible at once, distinguished by `trigger`.

## Example

```
SHOW TRANSACTIONS;
```

```
username | transaction_id | query                  | status  | metadata                                                                                 | start_time           | elapsed_ms
---------+----------------+------------------------+---------+------------------------------------------------------------------------------------------+----------------------+-----------
         | gc             | ["GARBAGE COLLECTION"] | running | {db_name: "memgraph", phase: "index_cleanup", trigger: "periodic", exclusive_lock: true} | 2026-06-03T09:48:... | 395
```

### DIFF
--- a/src/storage/v2/gc_status.cpp
+++ b/src/storage/v2/gc_status.cpp
@@ -13,19 +13,20 @@
 
 namespace memgraph::storage {
 
-const char *GcProgress::PhaseToString(GcPhase phase) {
+std::string_view GcProgress::PhaseToString(GcPhase phase) {
+  using namespace std::string_view_literals;
   switch (phase) {
     using enum GcPhase;
     case IDLE:
-      return "idle";
+      return "idle"sv;
     case UNLINK:
-      return "unlink";
+      return "unlink"sv;
     case INDEX_CLEANUP:
-      return "index_cleanup";
+      return "index_cleanup"sv;
     case DELETE:
-      return "delete";
+      return "delete"sv;
   }
-  return "unknown";
+  return "unknown"sv;
 }
 
 }  // namespace memgraph::storage
--- a/src/storage/v2/gc_status.cpp
+++ b/src/storage/v2/gc_status.cpp
@@ -1,0 +1,31 @@
+// Copyright 2026 Memgraph Ltd.
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
+// License, and you may not use this file except in compliance with the Business Source License.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+#include "storage/v2/gc_status.hpp"
+
+namespace memgraph::storage {
+
+const char *GcPhaseToString(GcPhase phase) {
+  switch (phase) {
+    using enum GcPhase;
+    case IDLE:
+      return "idle";
+    case UNLINK:
+      return "unlink";
+    case INDEX_CLEANUP:
+      return "index_cleanup";
+    case DELETE:
+      return "delete";
+  }
+  return "unknown";
+}
+
+}  // namespace memgraph::storage
--- a/src/storage/v2/gc_status.cpp
+++ b/src/storage/v2/gc_status.cpp
@@ -15,19 +15,17 @@
 
 namespace memgraph::storage {
 
-bool GcProgress::IsRunning() const { return running.load(std::memory_order_acquire); }
-
 void GcProgress::Start(bool is_periodic, bool is_exclusive) {
   start_steady_ms.store(
       std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now().time_since_epoch())
           .count(),
-      std::memory_order_relaxed);
+      std::memory_order_release);
   start_time_us.store(
       std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::system_clock::now().time_since_epoch())
           .count(),
-      std::memory_order_relaxed);
-  periodic.store(is_periodic, std::memory_order_relaxed);
-  exclusive_lock.store(is_exclusive, std::memory_order_relaxed);
+      std::memory_order_release);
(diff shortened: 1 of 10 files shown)


imilinovic:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - `SHOW TRANSACTIONS` now shows a running storage garbage collection as a synthetic `gc` row (`trigger`, `phase`, `exclusive_lock`, `db_name` metadata), so operators can directly observe a slow or blocking GC instead of inferring it. [#4229](https://github.com/memgraph/memgraph/pull/4229)
- [x] **[ Documentation PR link memgraph/documentation#1652 ]**

src/storage/v2/inmemory/storage.hpp
```diff
@@ -754,6 +755,19 @@ class InMemoryStorage final : public Storage {
             .start_steady_ms = snapshot_progress_.start_steady_ms.load(std::memory_order_acquire)};
   }
 
+  // Coherent read: nullopt unless a snapshot is running. Re-checks the flag after
+  // the fields so a snapshot ending mid-read reads as not-running, never a torn row.
+  std::optional<SnapshotProgressView> TryGetSnapshotProgress() const {
+    if (!snapshot_running_.load(std::memory_order_acquire)) return std::nullopt;
+    SnapshotProgressView progress = GetSnapshotProgress();
+    if (!snapshot_running_.load(std::memory_order_acquire)) return std::nullopt;
```
DavIvek: Do you need this double check? Maybe only the second one is enough?

src/storage/v2/gc_status.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+void GcProgress::SetPhase(GcPhase p) { phase.store(p, std::memory_order_release); }
+
+void GcProgress::Reset() {
+  running.store(false, std::memory_order_release);
+  phase.store(GcPhase::IDLE, std::memory_order_relaxed);
+  exclusive_lock.store(false, std::memory_order_relaxed);
+  periodic.store(false, std::memory_order_relaxed);
+  start_time_us.store(0, std::memory_order_relaxed);
+  start_steady_ms.store(0, std::memory_order_relaxed);
```
DavIvek: Here using `std::memory_order_relaxed` and in `src/storage/v2/snapshot_progress.hpp` `std::memory_order_release`

tests/unit/storage_v2_gc.cpp
DavIvek, line 1: It looks to me like nothing is testing concurrency at the moment, even though that’s probably the riskiest part.

src/storage/v2/inmemory/storage.hpp
```diff
@@ -754,6 +755,19 @@ class InMemoryStorage final : public Storage {
...
+    return progress;
+  }
+
+  bool IsGcRunning() const { return gc_progress_.IsRunning(); }
```
DavIvek: Looks to me this is not used? Same for snapshot

DavIvek (changes requested):
A few comments, but nice work 🚀

src/storage/v2/gc_status.cpp
imilinovic: using release everywhere now

src/storage/v2/inmemory/storage.hpp
imilinovic: it was just some early out, but not performance critical and maybe not even good, removed

tests/unit/storage_v2_gc.cpp
imilinovic: hard to test without being flaky due to the periodic nature of gc (and it being very short)

DavIvek (approved):

PR #4230, opened 2026-06-03
feat: add toBooleanOrNull, toFloatOrNull, toIntegerOrNull, toStringList

imilinovic:
Add the missing OrNull scalar conversion functions and the toStringList list converter. The OrNull variants return null for unsupported argument types instead of throwing, delegating to the existing strict conversion functions for the supported types. toStringList converts each element via toStringOrNull so non-stringifiable elements become null rather than aborting.

### Behavior changes

- **BREAKING: `toBooleanList` / `toFloatList` / `toIntegerList` no longer throw on a non-convertible element** — the offending element becomes `null` in the returned list, matching the documented casting semantics ("if any values are not convertible they will be `null` in the returned list"). Previously these aborted the whole query. Each now routes its per-element conversion through the corresponding `*OrNull` variant.
- **`toString` / `toStringOrNull` now stringify points** (`POINT({ x:.., y:.., srid:.. })`) via the shared canonical point representation used by Bolt output, dump, and audit.
- **`toStringOrNull` returns null (not throw)** when an enum value cannot be converted to a string, matching the OrNull contract.

### Notes

- The strict scalar conversions (`toBoolean` / `toFloat` / `toInteger`) keep throwing on a rejected argument type; the `*OrNull` variants return null. A parse failure on an accepted type (e.g. `toInteger("abc")`) returns null in both forms.
- The three scalar `*OrNull` wrappers now share a `ConvertOrNull` helper. `toString` and `toStringOrNull` share a `ToStringTypes` alias, and `toStringOrNull` delegates to `toString` (with the unresolvable-enum case still returning null) rather than duplicating its switch.
- Extend the CompareList unit-test helper to handle string elements.

### DIFF
--- a/src/query/interpret/awesome_memgraph_functions.cpp
+++ b/src/query/interpret/awesome_memgraph_functions.cpp
@@ -605,9 +605,8 @@ TypedValue OutDegree(const TypedValue *args, int64_t nargs, const FunctionContex
 
 // Type-set shared by each strict to* and its *OrNull variant: strict throws on a rejected type, *OrNull
 // returns null; a parse failure on an accepted type returns null in both.
-using ToBooleanTypes = Or<Null, Bool, Integer, String>;
-using ToFloatTypes = Or<Null, Bool, Number, String>;
-using ToIntegerTypes = Or<Null, Bool, Number, String>;
+using ToBooleanTypes = Or<Null, Bool, Integer, String>;  // Integer, not Number: toBoolean rejects floats.
+using ToNumericTypes = Or<Null, Bool, Number, String>;   // shared by toFloat and toInteger.
 using ToStringTypes =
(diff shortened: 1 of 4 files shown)


imilinovic:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Adds the `toBooleanOrNull`, `toFloatOrNull`, `toIntegerOrNull`, and `toStringList` conversion functions. Lets `toString`/`toStringOrNull` convert spatial points. Use the `*OrNull` scalars and the list converters when input may contain values that cannot be converted. BREAKING: `toBooleanList`/`toFloatList`/`toIntegerList` return `null` for non-convertible elements instead of failing the whole query[#4230](https://github.com/memgraph/memgraph/pull/4230)
- [x] **[ Documentation PR link memgraph/documentation#1651 ]**

DavIvek (approved):

PR #4246, opened 2026-06-05
fix(plan): prevent server crash on label disjunctions over many indexed labels

### LINKED ISSUE #4238: Segfault when a label OR over many indexed labels compiles to a deeply nested Union inside a write query
## Environment

- Memgraph 3.9.0 and 3.10.1 (latest at the time of writing). Reproduced on a fresh, empty 3.10.1 container.
- Image: `memgraph/memgraph-mage:latest`, single node, Docker.
- OS: Linux x86_64.
- Client: Bolt. Also reproducible from `mgconsole`.
- Default configuration. Not OOM killed (`OOMKilled=false`, no memory limit).
- Raising the OS thread stack ulimit does not help. I tested `docker run --ulimit stack=536870912:536870912` (512 MB) and the engine still segfaults, so the executor worker thread is using a fixed internal stack size that the OS limit does not reach.

## Summary

When a node is matched by an indexed property and constrained to one of many labels with an `OR`, for example `OPTIONAL MATCH (g {id: $x}) WHERE g:L0 OR g:L1 OR ... OR g:L35`, and every label in the disjunction has a label-property index on that property, the planner compiles the match into a left-deep tree of N-1 nested `Union` operators, one `ScanAllByLabelProperties` per label, each wrapped in `Union` and `Distinct`.

When that plan is part of a write query (it feeds `FOREACH ( ... | MERGE ... )` and is preceded by an `OPTIONAL MATCH ... DELETE` cleanup), executing it segfaults the engine. A worker thread overflows its stack. Memgraph takes SIGSEGV and the container exits. No error is returned to the client and nothing is written to Memgraph's log. The client only sees the connection drop.

The crash is governed by operator tree depth (the nested `Union`s plus the surrounding write operators), not by data size. It reproduces on an empty database.

It does not crash when any one of these holds:

- the `OR` has few labels (roughly 8 or fewer nested `Union`s),
- at least one label in the `OR` has no property index, in which case the planner falls back to a single flat `ScanAll`,
- the disjunction is replaced by a single concrete label (one index seek, no `Union`),
- the query is read only, or has fewer surrounding operators.

## Reproducer

A one-shot shell script that starts a container and triggers the crash is at the bottom of this report. The manual steps are below.

### 1. Setup

Create an `id` index for each of 36 labels `L0` through `L35`, two helper labels, and one owner node:

```cypher
CREATE INDEX ON :L0(id);
// L1 through L35 ...
CREATE INDEX ON :L35(id);
CREATE INDEX ON :N(id);
CREATE INDEX ON :M(id);
CREATE (:L0 {id:'owner1'});
CREATE (:M  {id:'m1'});
```

### 2. Crashing query

Run this twice. The first run creates `n` and the `:IN` relationship. The second run exercises the `OPTIONAL MATCH ... DELETE` cleanup branch and segfaults the server:

```cypher
MERGE (n:N {id:'n1'})
WITH n
OPTIONAL MATCH (n)<-[r1:IN]-(:M)
OPTIONAL MATCH (n)-[r2:OUT]->(:M)
DELETE r1, r2
WITH n
OPTIONAL MATCH (g {id:'owner1'})
WHERE (g:L0 OR g:L1 OR g:L2 OR g:L3 OR g:L4 OR g:L5 OR g:L6 OR g:L7 OR g:L8
    OR g:L9 OR g:L10 OR g:L11 OR g:L12 OR g:L13 OR g:L14 OR g:L15 OR g:L16
    OR g:L17 OR g:L18 OR g:L19 OR g:L20 OR g:L21 OR g:L22 OR g:L23 OR g:L24
    OR g:L25 OR g:L26 OR g:L27 OR g:L28 OR g:L29 OR g:L30 OR g:L31 OR g:L32
    OR g:L33 OR g:L34 OR g:L35)
FOREACH (_ IN CASE WHEN g IS NOT NULL THEN [1] ELSE [] END |
  MERGE (n)-[:OWNS]->(g))
WITH n
MATCH (a:M {id:'m1'})
MERGE (n)<-[:IN]-(a);
```

## EXPLAIN

The owner match becomes 35 nested `Union` operators, one per extra label, each over a `ScanAllByLabelProperties` and a `Distinct`:

```
 * Foreach
 |\
 | * Merge ( (n)-[:OWNS]->(g) )
 * Optional
   * Distinct
   * Union { g : g }
     |\
     | * ScanAllByLabelProperties (g :L0 {id})
     * Distinct
     * Union { g : g }
       |\
       | * ScanAllByLabelProperties (g :L1 {id})
       // 33 more nested Union levels, one per remaining label
```

The owner match alone is about 70 operators deep (35 `Union` plus 35 `Distinct`), nested inside the write pipeline (`Accumulate`, `Delete`, `Foreach`, `Merge`, `Unwind`, `Merge`).

## Crash signature (dmesg)

```
low prior.[...]: segfault at 7fb32c633ff8  ip ...  sp 00007fb32c634000  error 6 in memgraph[...]
low prior.[...]: segfault at 0             ip ...                       error 4 in memgraph[...]
```

The fault address (`...633ff8`) is 8 bytes below the stack pointer (`...634000`), which is a stack overflow in the executor (the `low prior.` worker thread). Other restarts fault at `0` or small offsets, which is the null or wild pointer dereference that follows the overflow. `OOMKilled` is false, nothing is logged, and the process takes SIGSEGV.

## Expected behavior

A label `OR` over indexed properties should compile to a flat or N-ary union, or otherwise be executed without O(N) stack depth, so that a larger label set cannot overflow the thread stack. At a minimum Memgraph should return a query error rather than segfaulting the server.

## Why it matters

This is a plain, idempotent upsert that re-links a node to its owner, where the owner can be one of many labels. That is a common shape when a reference points at a polymorphic type. With indexes on all candidate labels, which is the recommended configuration, an ordinary write reliably takes the database down in the middle of a transaction.

## Observations that localize it to plan depth

- 2 to 8 labels in the `OR` run fine (shallow nesting).
- 10 or more labels, all indexed, inside a write pipeline segfault.
- The same `OR` with one label unindexed produces a single flat `ScanAll` and does not crash.
- The same query, read only or with fewer surrounding operators, does not crash.
- A single concrete label (`(g:L0 {id})`) is one `ScanAllByLabelProperties`, no `Union`, and does not crash.
- 3.10.1 still crashes. Raising the stack ulimit to 512 MB does not change it.

## Workarounds for anyone hitting this

- Match the target by a single concrete label instead of a label `OR`.
- Keep the label disjunction small (8 or fewer).
- Leave one candidate label unindexed to force a flat `ScanAll`. Not desirable at scale.

## One-shot reproduction script

Save as `repro.sh`, then run `bash repro.sh`. It needs Docker. It starts a throwaway Memgraph container on port 7699, runs the query twice, prints the container state (which is `exited` after the crash), and removes the container.

```bash
#!/usr/bin/env bash
# Reproduce a Memgraph segfault: a label OR over many indexed labels compiles to
# a deeply nested Union, and running it inside a write query overflows the
# executor thread stack and crashes the server.
set -uo pipefail

IMAGE="memgraph/memgraph-mage:latest"
NAME="mg-union-segfault-repro"
PORT="7699"
LABELS=36

docker rm -f "$NAME" >/dev/null 2>&1 || true
echo "starting $IMAGE as $NAME on port $PORT"
docker run -d --name "$NAME" -p "${PORT}:7687" "$IMAGE" --telemetry-enabled=false >/dev/null

mg() { docker exec -i "$NAME" mgconsole 2>/dev/null; }

echo "waiting for bolt"
for _ in $(seq 1 60); do
  if echo "RETURN 1;" | mg >/dev/null 2>&1; then break; fi
  sleep 2
done

echo "version:"
echo "SHOW VERSION;" | mg

echo "creating 36 label indexes, two helper indexes, and seed nodes"
{
  for i in $(seq 0 $((LABELS-1))); do echo "CREATE INDEX ON :L${i}(id);"; done
  echo "CREATE INDEX ON :N(id);"
  echo "CREATE INDEX ON :M(id);"
  echo "CREATE (:L0 {id:'owner1'});"
  echo "CREATE (:M  {id:'m1'});"
} | mg >/dev/null

# build the OR over all 36 labels
OR=""
for i in $(seq 0 $((LABELS-1))); do OR="${OR}g:L${i} OR "; done
OR="${OR%% OR }"

QUERY="MERGE (n:N {id:'n1'})
WITH n
OPTIONAL MATCH (n)<-[r1:IN]-(:M)
OPTIONAL MATCH (n)-[r2:OUT]->(:M)
DELETE r1, r2
WITH n
OPTIONAL MATCH (g {id:'owner1'}) WHERE (${OR})
FOREACH (_ IN CASE WHEN g IS NOT NULL THEN [1] ELSE [] END | MERGE (n)-[:OWNS]->(g))
WITH n
MATCH (a:M {id:'m1'})
MERGE (n)<-[:IN]-(a);"

echo "run 1 (create)"
echo "$QUERY" | timeout 20 docker exec -i "$NAME" mgconsole >/dev/null 2>&1 || true
sleep 1
echo "run 2 (re-upsert, this segfaults the server)"
echo "$QUERY" | timeout 20 docker exec -i "$NAME" mgconsole 2>&1 | tail -2 || true

sleep 2
echo "container state (expect status=exited after the crash):"
docker inspect -f 'status={{.State.Status}} exitcode={{.State.ExitCode}} oomkilled={{.State.OOMKilled}}' "$NAME"

docker rm -f "$NAME" >/dev/null 2>&1 || true
```

Ignition:
A query like `MATCH (n) WHERE n:L0 OR ... OR n:Ln RETURN n` (all labels indexed) could crash the whole server (OOM-killed or stack-overflow segfault), scaling with the number of labels. 

Three fixes:

1. **Exponential index validation (OOM).** `UsedIndexChecker`'s composite-operator `PreVisit`s were traversing children manually *and* returned `true`, hence `Accept` re-traversed every branch, causing 2^N over the deep Union chain. I've dropped the manual traversal so each branch is visited only once.

2. **Linear tree depth (stack overflow).** The disjunction compiled to a left-deep `Union`+`Distinct` chain, depth linear in N. Now folds into a balanced `Union` tree (depth `ceil(log2 N)`) with a single top `Distinct`.

3. **Memory limit bypassed while planning.** `MemoryTrackerCanThrow()` was false during preparation, so the runaway allocation skipped `--memory-limit`. Added an `OutOfMemoryExceptionEnabler` around `CypherQueryToPlan`; a runaway now aborts the query instead of killing the server.

No behaviour change for valid queries: the plan is set-equivalent, and the union-fold is still only used when every disjoined label is indexed.

### DIFF
--- a/src/query/cypher_query_interpreter.cpp
+++ b/src/query/cypher_query_interpreter.cpp
@@ -27,6 +27,7 @@
 #include "query/plan/used_index_checker.hpp"
 #include "query/plan/vertex_count_cache.hpp"
 #include "utils/flag_validation.hpp"
+#include "utils/memory_tracker.hpp"
 #include "utils/string.hpp"
 
 #include "parameters/parameters.hpp"
@@ -224,6 +225,13 @@ std::shared_ptr<PlanWrapper> CypherQueryToPlan(frontend::StrippedQuery const &st
                                                PlanCacheLRU *plan_cache, DbAccessor *db_accessor,
                                                plan::v2::QueryPlannerContext &planner_context,
                                                const std::vector<Identifier *> &predefined_identifiers) {
+  // Enforce the global memory limit during query preparation. Without this,
+  // MemoryTrackerCanThrow() is false here (the per-cursor
+  // OutOfMemoryExceptionEnablers only cover execution), so a runaway allocation
+  // while planning would bypass --memory-limit and let the kernel OOM-kill the
+  // whole server instead of aborting just the query.
+  utils::MemoryTracker::OutOfMemoryExceptionEnabler const oom_exception_enabler;
+
   // Skip plan cache when using experimental v2 planner - plans may change as v2 evolves
   const bool use_plan_cache = plan_cache && !flags::AreExperimentsEnabled(flags::Experiments::PLANNER_V2);
   if (use_plan_cache) {
--- a/src/query/plan/rewrite/balanced_union.hpp
+++ b/src/query/plan/rewrite/balanced_union.hpp
@@ -1,0 +1,88 @@
+// Copyright 2026 Memgraph Ltd.
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
+// License, and you may not use this file except in compliance with the Business Source License.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0, included in the file
+// licenses/APL.txt.
+
+/// @file
+/// Lowering for an indexed disjunction over a single node: fold N index scans
+/// into a balanced `Union` tree topped by a single `Distinct`.
+
+#pragma once
+
+#include <memory>
+#include <utility>
+#include <vector>
+
+#include "query/frontend/semantic/symbol.hpp"
+#include "query/plan/operator.hpp"
+#include "utils/logging.hpp"
+
+namespace memgraph::query::plan {
+
+/// Fold per-symbol index `scans` into a balanced binary `Union` tree topped by a
+/// single deduplicating `Distinct`, and return the root.
+///
+/// Preconditions (all hold for an indexed label or edge-type disjunction over
+/// one node, and are exactly what make the balancing + single `Distinct` valid):
+///   - every scan produces the same single output symbol `node_symbol`, so the
+///     symbol mapping is identical at every level of the tree;
+///   - the branches combine by pure set-union (`Union` concatenates), so one
+///     global `Distinct` is equivalent to deduplicating at every level;
+///   - the combine operator (`Union`) is order-independent in cost --
+///     `Union(A, B)` costs the same as `Union(B, A)`, and the total cardinality
+///     is the sum of the branches regardless of how they nest -- so the pairing
+///     and nesting are free to optimize for depth. Note this does NOT require
+///     the branches to be equal-cost: they routinely are not (one label may hit
+///     a handful of vertices and another millions), and they are paired by
+///     index order, not by cost.
+///
+/// Do NOT reuse this for query-level `UNION` (per-combinator symbols, `UNION
+/// ALL`) or `Cartesian` (left/right order is the cost decision): those violate
+/// the preconditions, and balancing would change semantics or cost.
+///
+/// Balancing keeps the operator-tree depth at O(log N) in the number of scans
+/// instead of the O(N) of a left-deep chain, which otherwise overflows the
+/// executor thread stack at large N. Balancing trades nothing in logical
+/// (cardinality) cost; what it bounds is the executor stack depth. The single
+/// top `Distinct` removes the duplicates a vertex matching several labels would
+/// otherwise produce.
+///
+/// `scans` must be non-empty. With a single scan the result is that scan
+/// unchanged (no `Union`, no `Distinct`).
+inline std::unique_ptr<LogicalOperator> BalancedDisjunctionUnion(std::vector<std::unique_ptr<LogicalOperator>> scans,
+                                                                 const Symbol &node_symbol) {
+  MG_ASSERT(!scans.empty(), "BalancedDisjunctionUnion requires at least one scan");
+  // More than one branch means a vertex can match several of them, so the
+  // combined output needs deduplicating exactly once at the top.
+  const bool needs_distinct = scans.size() > 1;
+  // Combine adjacent pairs each round; an odd one out is carried over unchanged.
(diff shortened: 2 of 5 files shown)


Ignition:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Large OR-ed indexed-label queries (MATCH (n:A|B|...)) no longer crash the server. [#4246](https://github.com/memgraph/memgraph/pull/4246)

Kevin-Mc-Callister:
Hello! This was such a quick turn-around for a fix.

Just as a side note, I built this locally and it appears to have fixed the issue #4238

DavIvek (approved):
Nice work! 🚀

PR #4269, opened 2026-06-15
fix: prevent infinite loop/OOM in replace() with empty search string

### LINKED ISSUE #4266: Crash by infinite loop/OOM in replace() function when search string is empty
**Memgraph version**
v3.10.1 Latest

**Environment**
Ubuntu 22.04 and mgconsole 1.5.2

**Describe the bug**
Denial of Service via infinite loop/OOM in replace() function when search string is empty.

**To Reproduce**
```
# Executing this query in mgconsole:
RETURN replace("A", "", "B");
# mg_raw_transport_recv: connection closed by server

# Also reproducible with RETURN replace("", "", "a");
```

**Root Cause Analysis**
The crash is triggered explicitly when the search_string argument is an empty string "", regardless of the original string. This strongly indicates an infinite loop and subsequent out-of-memory (OOM) memory exhaustion in the underlying C++ string replacement implementation. The standard std::string::find("", pos) will return a match at every position. If the internal loop fails to correctly advance the search cursor by at least 1 character when the search length is 0, it will continuously append the replacement string, infinitely allocating heap memory until the OS abruptly kills the Memgraph daemon.

imilinovic:
### Description

Fixes #4266.

`replace(original, search, replacement)` crashed the server (infinite loop → OOM) whenever `search` was an empty string, e.g. `RETURN replace("A", "", "B")`. In `utils::Replace`, `find("")` matches at every position and the loop never advances, so it appends the replacement indefinitely until the process exhausts memory.

### Fix

Special-case an empty `match`: insert the replacement at every byte boundary in a single pass with an exact `reserve()`, matching `std::regex_replace`. So `replace("A", "", "B")` → `"BAB"` and `replace("abc", "", "-")` → `"-a-b-c-"`. This converts an unbounded loop into a bounded, single allocation.

### DIFF
--- a/src/utils/string.hpp
+++ b/src/utils/string.hpp
@@ -186,6 +186,16 @@ template <class TAllocator>
 std::basic_string<char, std::char_traits<char>, TAllocator> *Replace(
     std::basic_string<char, std::char_traits<char>, TAllocator> *out, const std::string_view src,
     const std::string_view match, const std::string_view replacement) {
+  if (match.empty()) {  // empty match inserts replacement at every byte boundary, matching std::regex_replace
+    out->clear();
+    out->reserve(src.size() + (src.size() + 1) * replacement.size());
+    out->append(replacement);
+    for (const char c : src) {
+      out->push_back(c);
+      out->append(replacement);
+    }
+    return out;
+  }
   // TODO: This could be implemented much more efficiently.
   *out = src;
   for (size_t pos = out->find(match); pos != std::string::npos; pos = out->find(match, pos + replacement.size())) {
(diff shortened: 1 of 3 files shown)


imilinovic:
### Tracking
- [x] **[Link to Epic/Issue]** #4266

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - `replace()` no longer crashes the server when the search string is empty; an empty search string now inserts the replacement around every character (e.g. `replace("abc", "", "-")` returns `"-a-b-c-"`), so such queries return a result instead of taking the instance down. [#4269](https://github.com/memgraph/memgraph/pull/4269)
- [x] **[ Documentation PR link memgraph/documentation#1658 ]**

as51340 (approved):

PR #4273, opened 2026-06-15
fix: terminate label-property index scan at value bound before visibility checks (#4089)

### LINKED ISSUE #4089: Bad performance on concurrent MERGE queries
https://github.com/memgraph/best-practices/pull/60

`uv run iceberg_to_memgraph.py --source pyiceberg --batch-size 20000 --workers 16 --user-write merge`

Really bad performance with some of the merge queries (146 seconds for MERGE of 10k users was the worst).
Indices were already created beforehand.

--user-write create 
works perfectly normal

imilinovic:
## Description

Concurrent `MERGE` ingestion was pathologically slow — O(N²) per batch — while the equivalent `CREATE` was fast (#4089). The cause is the order of per-entry filters in the in-memory property-index scan (`AdvanceUntilValid_`).

A `MERGE` lookup for `prop = x` seeks to the first index entry with value `>= x` (the skip list is value-sorted) and iterates forward, applying filters in this order:

1. `gid >= max_gid` — visibility
2. `CanSeeEntityWithTimestamp` — visibility
3. value-bounds check — if past the upper bound, **terminate the scan** (`break`)

The terminator ran **last**. So for `MERGE {prop: 42}` on a miss, the scan walks forward into `43, 44, …`; under concurrency those higher entries were just created by other in-flight transactions and are invisible, so each one hits the `gid` filter and `continue`s **before reaching the bounds terminator** — the scan never stops at `42` and walks every invisible entry to the end of the index. One lookup becomes O(N), and N concurrent lookups make a batch O(N²). (Serial loads are fine: the neighbouring entries are already committed and visible, so the first one reaches the bounds check and stops.)

### Fix

Move the value-bounds check **ahead of** the visibility filters so the scan terminates at the bound first:

- The bounds check reads only the entry's value, which is immutable and available regardless of visibility.
- Early termination depends on value ordering, which is orthogonal to MVCC visibility, so stopping at the bound is always correct.
- Invisible entries are still never returned — the visibility filters still run before any match is emitted.

Applied to all three scans sharing this pattern: `InMemoryLabelPropertyIndex`, `InMemoryEdgePropertyIndex`, `InMemoryEdgeTypePropertyIndex`.

## Root cause analysis

- Only **concurrent** `MERGE` was slow → cost was in the index match-phase scan under concurrency.
- Query plan confirmed correct (`ScanAllByLabelProperties`).
- Off-CPU profiling: workers not blocked on a lock (~2.5/12 cores busy).
- On-CPU profiling: **99.4%** of samples in `AdvanceUntilValid()` — the scan loop itself (iterator advance + `gid` compare), not the expensive delta-chain check.
- Reproduces in both `IN_MEMORY_ANALYTICAL` and `IN_MEMORY_TRANSACTIONAL`.

## Performance

| Scenario | Before | After |
|---|---|---|
| 16-worker MERGE, 200k nodes (analytical) | 9.9–18 s | **0.56 s** |
| 16-worker MERGE, 200k nodes (transactional) | 16.1 s | **0.59 s** |
| 16-worker MERGE, 600k nodes (analytical) | 46.9 s | **1.36 s** (~35×) |
| CREATE / serial paths | — | unchanged |

### Verification with the issue's workload

Reproduced with the exact command from #4089 (`iceberg_to_memgraph.py` from memgraph/best-practices, 16 worker processes, `--batch-size 20000`), 200k users, `IN_MEMORY_ANALYTICAL`, Release builds:

```
python3 iceberg_to_memgraph.py --source pyiceberg --batch-size 20000 --workers 16 --user-write {merge|create}
```

| 200k users, 16 workers, batch=20k | master | this PR |
|---|---|---|
| `--user-write merge`  | 22.62 s | **0.93 s** (~24×) |
| `--user-write create` | 0.74 s  | 0.58 s (unchanged) |

On master, `MERGE` is ~30× slower than the equivalent `CREATE`; the fix closes the gap (`MERGE` ≈ `CREATE`).

Fixes #4089


imilinovic:
### Tracking
- [ ] **[Link to Epic/Issue]** Fixes #4089

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Concurrent `MERGE` ingestion into indexed labels and edge types is now dramatically faster. [#4273](https://github.com/memgraph/memgraph/pull/4273)
- [x] **[ Documentation PR link N/A ]** — performance fix only; no user-facing syntax, flags, or behaviour to document.

andrejtonev (approved):
💪

PR #4275, opened 2026-06-16
fix: Update `UUID` label on default DB gauges when data instance join cluster

colinbarry:
Ensure OpenMetrics gauges use the correct shared default database UUID when a data instance joins a cluster. 

Previously, the default database gauges would use the UUID of their local (discarded) default database:
```
memgraph_accumulate_operator_total{database="memgraph", instance="localhost:9091", instance_name="instance_1", job="memgraph", uuid="00039767-6ba8-4c12-af7d-1dfebd3f0d9e"}    0           
memgraph_accumulate_operator_total{database="memgraph", instance="localhost:9092", instance_name="instance_2", job="memgraph", uuid="ac1ff97f-c893-4709-9bd3-8a7cee375e9e"}    0           
memgraph_accumulate_operator_total{database="memgraph", instance="localhost:9093", instance_name="instance_3", job="memgraph", uuid="10958d52-a951-43c4-9609-387e8fdcf57f"}    0
```                                                                                                                                                   

Now HA instances update the label to the UUID of the shared default database:

```
memgraph_accumulate_operator_total{database="memgraph", instance="localhost:9091", instance_name="instance_1", job="memgraph", uuid="0311f880-1363-448c-8489-5a0618aba107"}    0         
memgraph_accumulate_operator_total{database="memgraph", instance="localhost:9092", instance_name="instance_2", job="memgraph", uuid="0311f880-1363-448c-8489-5a0618aba107"}    0           
memgraph_accumulate_operator_total{database="memgraph", instance="localhost:9093", instance_name="instance_3", job="memgraph", uuid="0311f880-1363-448c-8489-5a0618aba107"}    0 
```

Also see:
https://github.com/memgraph/memgraph/pull/4774
https://github.com/memgraph/memgraph/pull/4805

### DIFF
--- a/src/dbms/database.hpp
+++ b/src/dbms/database.hpp
@@ -236,6 +236,10 @@ class Database {
 
   metrics::DatabaseMetricHandles *metric_handles() { return &metrics_.handles(); }
 
+  void RebindMetrics(utils::UUID const &uuid, metrics::DatabaseMetricHandles const &handles) {
+    metrics_.Rebind(uuid, handles);
+  }
+
  private:
   // Declared first so it is released last: the storage's garbage collector writes to these metrics
   // until it is joined, which happens while the storage is being destroyed.
--- a/src/dbms/database.hpp
+++ b/src/dbms/database.hpp
@@ -236,9 +236,7 @@ class Database {
 
   metrics::DatabaseMetricHandles *metric_handles() { return &metrics_.handles(); }
 
-  void RebindMetrics(utils::UUID const &uuid, metrics::DatabaseMetricHandles const &handles) {
-    metrics_.Rebind(uuid, handles);
-  }
+  void RebindMetrics(metrics::DatabaseMetricHandles const &handles) { metrics_.Rebind(handles); }
 
  private:
   // Declared first so it is released last: the storage's garbage collector writes to these metrics
--- a/src/dbms/database.hpp
+++ b/src/dbms/database.hpp
@@ -236,8 +236,6 @@ class Database {
 
   metrics::DatabaseMetricHandles *metric_handles() { return &metrics_.handles(); }
 
-  void RebindMetrics(metrics::DatabaseMetricHandles const &handles) { metrics_.Rebind(handles); }
-
  private:
   // Declared first so it is released last: the storage's garbage collector writes to these metrics
   // until it is joined, which happens while the storage is being destroyed.
--- a/src/dbms/database.hpp
+++ b/src/dbms/database.hpp
@@ -232,7 +232,9 @@ class Database {
     ~GatekeeperGuard() noexcept { memory::tls_db_arena_state = prev_; }
   };
 
-  std::shared_ptr<metrics::DatabaseMetricHandles> metric_handles() const { return metrics_.handles(); }
+  metrics::DatabaseMetricHandles const *metric_handles() const { return &metrics_.handles(); }
+
+  metrics::DatabaseMetricHandles *metric_handles() { return &metrics_.handles(); }
 
   void RebindMetrics(metrics::DatabaseMetricHandles const &handles) { metrics_.Rebind(handles); }
 
--- a/src/dbms/database.hpp
+++ b/src/dbms/database.hpp
@@ -232,9 +232,7 @@ class Database {
     ~GatekeeperGuard() noexcept { memory::tls_db_arena_state = prev_; }
   };
 
-  metrics::DatabaseMetricHandles const *metric_handles() const { return &metrics_.handles(); }
-
-  metrics::DatabaseMetricHandles *metric_handles() { return &metrics_.handles(); }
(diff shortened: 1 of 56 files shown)


colinbarry:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Ensure the correct UUID of the default database is shown in the OpenMetrics scrapes on replica data instances. [#4275](https://github.com/memgraph/memgraph/pull/4275)

src/dbms/database.hpp
```diff
@@ -236,6 +236,10 @@ class Database {
 
   metrics::DatabaseMetricHandles *metric_handles() { return &metrics_.handles(); }
 
+  void RebindMetrics(utils::UUID uuid, metrics::DatabaseMetricHandles handles) {
+    metrics_.Rebind(uuid, std::move(handles));
```
as51340: I think you can remove move, handles are trivially copyable

src/dbms/database.hpp
```diff
@@ -270,6 +274,11 @@ class Database {
 
     metrics::DatabaseMetricHandles &handles() { return handles_; }
 
+    void Rebind(utils::UUID uuid, metrics::DatabaseMetricHandles handles) {
+      uuid_ = uuid;
+      handles_ = std::move(handles);
```
as51340: trivially copyable

as51340: also in the constructor

src/dbms/dbms_handler.hpp
```diff
@@ -208,7 +208,8 @@ class DbmsHandler {
       spdlog::debug("Updated default db's UUID");
       // Default db cannot be deleted and remade, have to just update the UUID
       storage->config_.salient.uuid = config.uuid;
-      metrics::Metrics().RebindDefaultDatabaseUUID(config.uuid);
+      auto new_handles = metrics::Metrics().RebindDefaultDatabaseUUID(config.uuid);
+      db->RebindMetrics(config.uuid, std::move(new_handles));
```
as51340: no need for move

src/metrics/prometheus_metrics.cpp
```diff
@@ -1148,14 +1148,10 @@ void PrometheusMetrics::RemoveDatabase(utils::UUID const &uuid) {
...
-  if (!default_db_uuid_) return;  // metrics not registered for the default DB
-  auto const &old_uuid = *default_db_uuid_;
-  auto it = r::find_if(databases_.entries, [&old_uuid](auto const &e) { return e.uuid == old_uuid; });
-  MG_ASSERT(it != databases_.entries.end(), "RebindDefaultDatabaseUUID: default UUID not found in metrics registry");
-  it->uuid = new_uuid;
-  default_db_uuid_ = new_uuid;
+DatabaseMetricHandles PrometheusMetrics::RebindDefaultDatabaseUUID(utils::UUID const &new_uuid) {
+  auto const old_uuid = *default_db_uuid_;
+  RemoveDatabase(old_uuid);
+  return AddDatabase(new_uuid, dbms::kDefaultDB);
```
as51340: 1. Logically I think this should be all under the lock
2. I think these change would UAF. For example storage uses GaugeHandle* which would just get destroyed with RemoveDatabase. You probably need to use shared_ptr.

as51340: maybe even atomic<shared_ptr> if multiple threads are accessing the same pointer?

as51340 (changes requested):

src/dbms/database.hpp
colinbarry: Fixed

src/dbms/database.hpp
colinbarry: Fixed

src/dbms/dbms_handler.hpp
colinbarry: Fixed

src/metrics/prometheus_metrics.cpp
colinbarry: You are right - this is UAF, and a bigger refactor than I initially thought.

colinbarry: Fixed: I decided against using `shared_ptr` indirection as I was concerned about the performance cost on relatively hot code.

Instead, I'm relying on the fact that rebinding only occurs on an clean database. We know that there are no indices, constraints, or TTL, so the scope of the stale underlying metrics pointers is quite small. The only concurrent users would potentially be GC or snapshot creator, and both of these are suspended during the switch.

as51340 (changes requested):
This is still not safe for indices,constraint, ttl (maybe something else). They are still able to UAF.

Additionally, I am a bit concerned about SystemRecoveryRpc and if something bad can happen if the request takes a long time to be received on replica (network issue). Therefore, please start Jepsen a few times to see if everything is fine

PR #4280, opened 2026-06-18
feat: Hot/Cold Tenants - suspend/resume a tenant to a durable COLD shell

andrejtonev:
Adds an admin-driven ability to **`SUSPEND DATABASE <name>`** (tear a tenant's in-memory storage down to a durable, metadata-only COLD shell, freeing its RAM) and **`RESUME DATABASE <name>`** (rebuild it HOT from disk with all data intact). Suspend/resume is **system-replicated like `CREATE`/`DROP DATABASE`**, survives restarts, and is an **enterprise-only multi-database feature** (privilege-checked; rejected outside enterprise edition).

## Why

Idle/cold tenants in a multi-tenant instance hold full in-memory storage. This lets an operator reclaim that RAM on demand without dropping the tenant or its data, and bring it back when needed — replicated consistently across the cluster.

## Design

- **`utils::Gatekeeper<Database>` 4-state FSM** (HOT → SUSPENDING → COLD → RESUMING). `access()` is HOT-only; suspend reaches a sole-accessor state before teardown; resume is single-flight (concurrent resumers rebuild once).
- **SUSPEND** is fail-fast: it does not kill in-flight queries; it bounded-waits for sole-access and otherwise returns a retriable error.
- **RESUME** is synchronous in the bolt worker (MVP threading model), with an OOM-safe rollback path: a failed rebuild leaves the tenant COLD and retriable. Resuming an already-HOT database is idempotent success and returns a distinct `"…is already resumed (HOT)."` status (not an error).
- **Managed features** are stopped on suspend and restored on resume: streams (stop+restore), TTL (user-check re-armed); the periodic-snapshot scheduler and GC are torn down with the storage on suspend and recreated on resume.
- **Snapshots**: SUSPEND does **not** take a proactive snapshot. Durability at suspend is whatever the periodic snapshot + WAL already provide, plus — if `--storage-snapshot-on-exit` is enabled — the snapshot the storage writes during its normal teardown. RESUME recovers via `recover_on_startup` (latest snapshot, if any, + WAL replay), so data is intact regardless of the flag.
- **Durability**: a `cold:true` marker + per-tenant metadata (uuid, rel_dir, as-of-suspend stats) persist the COLD state in the V2 durability schema (purely additive over V1); a restart restores COLD tenants as shells and HOT tenants normally.
- **Replication**: dedicated `Suspend`/`ResumeDatabaseRpc` apply handlers, plus a **V3 `SystemRecovery` wire** carrying the HOT/COLD set (each COLD tenant as its salient config + as-of-suspend stats) so a reconnecting/lagging/restarted replica converges to MAIN's authoritative state. A resumed cold tenant trusts its own on-disk WAL/snapshot epoch (rebuilt by `BuildDetached`) — no epoch is carried on the wire. Backward-compatible with pre-V3 peers via the SystemRecovery `V1→V2→V3` version chain.
- COLD tenants stay first-class for **`SHOW DATABASES`** (tagged COLD) and **`SHOW STORAGE INFO ON`** (serves the full as-of-suspend `StorageInfo` snapshot).


specs/hot-cold-databases.md
```diff
@@ -0,0 +1,335 @@
...
+    └──────────────────────────────┘
+```
+
+- A database can be suspended only when **nothing is actively using it**. If a client
+  session is connected to it, or a query/transaction is in flight, `SUSPEND` **fails
+  fast** with a clear "database is in use" error rather than waiting or forcibly killing
+  the in-flight work (see §6, decision **D6**). The operator retries once the database is
+  idle.
+- `SUSPEND` requires the database to have durable storage (periodic snapshots + WAL) and
+  to be an in-memory transactional database. Analytical/on-disk databases cannot be
```
as51340, line 127: What's the challenge of suspending analytical database?

specs/hot-cold-databases.md
```diff
@@ -0,0 +1,335 @@
...
+  suspended (suspending them could lose data — that would violate the core principle).
+- States are surfaced to operators via the `status` column of `SHOW DATABASES` (present
+  only when the feature is enabled).
+
+---
+
+## 6. Product decisions and rationale
+
+This feature went through two iterations. The first ("v1") was broad: it included
```
as51340: I would remove this first iteration, I think nobody will care about that in the future that we tried to do that.

specs/hot-cold-databases.md
```diff
@@ -0,0 +1,335 @@
...
+*Rationale.* Automatic eviction was the largest and riskiest part of v1: a watermark
+scheduler picking "the coldest idle tenant" introduced thrash risk, hard-to-reason-about
+timing, and the single most complex concurrency surface in the system (the idle-session
+reaper's claim protocol). It also made behavior non-deterministic from the operator's
+point of view — a database could vanish from memory without anyone asking. The product
+decision was that **the operator knows their workload better than a watermark heuristic
+does.** Memgraph provides the mechanism (suspend/resume) and the visibility (`SHOW
+DATABASES`, stats, metrics); the policy belongs to the operator or an external control
+plane. Removing the automatic machinery also removed an entire class of races and made
+the feature small enough to reason about and ship with confidence.
```
as51340: similar, I would just keep everything that is related to this current feature

specs/hot-cold-databases.md
```diff
@@ -0,0 +1,335 @@
...
+
+*Rationale.* In a replicated cluster, "which databases exist and in what state" is
+cluster-wide truth, exactly like the set of databases itself. If hot/cold were a
+node-local decision, a replica could disagree with its MAIN about whether a database is
+queryable, which is confusing and operationally fragile. Treating it as replicated DDL
+makes the behavior predictable: the cluster has one hot/cold map, and it is the MAIN's.
+
+A consequence, accepted deliberately: suspending a database on the MAIN also tears it
+down on replicas, evicting any reader connected to a replica's copy. That is the price of
+a single coherent cluster-wide state, and it is consistent with how other DDL behaves.
```
as51340: What happens if nobody is using database on main but the database is used on replica?
Should we maybe do this through consensus?

specs/hot-cold-databases.md
```diff
@@ -0,0 +1,335 @@
...
+predict that up front, `RESUME` attempts the rebuild and, if it runs out of memory (or
+recovery otherwise fails), **rolls the database back to cold** and returns a retriable
+error. The instance does not crash and no data is lost; the operator can free memory and
+retry.
+
+*Rationale.* A footprint pre-check would be both unreliable (memory is shared and moving)
+and redundant with the rollback path that has to exist anyway for genuine recovery
+failures. Attempt-and-roll-back is simpler, and it degrades gracefully: a failed resume
+is a clean, retriable error, never a half-built database or a downed instance. The same
+philosophy extends to startup: if recovering a hot database at boot would exhaust memory,
```
as51340: I am not sure if this is expected. Maybe that should be under the future flag --storage-allow-recovery-failure which I will add for tolerating tenant failures?

specs/hot-cold-databases.md
```diff
@@ -0,0 +1,335 @@
...
+| Metric | Type | Meaning |
+|---|---|---|
+| `memgraph_database_suspends_total` | counter | Successful suspends (operator + replica-apply). |
+| `memgraph_database_resumes_total` | counter | Successful resumes. |
+| `memgraph_cold_databases` | gauge | Currently-cold database count. |
+| `memgraph_database_boot_recovery_failures_total` | counter | Databases left cold at boot due to a recovery failure. |
+| `memgraph_database_boot_recovery_oom_failures_total` | counter | Subset of the above where the cause was out-of-memory. |
+
+Metrics are global rather than per-database because a cold database has no live storage
+to attach per-database metrics to. Per-database labelling is a possible future addition.
```
as51340: Not sure how useful is to have these as global metrics. I see a value in operator being able to see hey this DB is often cold, maybe I should think about migrating it somewhere else.

specs/hot-cold-databases.md
```diff
@@ -0,0 +1,335 @@
...
+- No data is lost by suspending and resuming (§2).
+- The default `memgraph` database can never be suspended (§4).
+- A failed resume rolls back to cold and is retriable; it never crashes the instance or
+  leaves a half-built database (D8).
+- Hot/cold state is consistent cluster-wide (D4) and durable across restart (D5).
+
+**Non-goals (explicitly out of scope for this release)**
+
+- **Automatic eviction / idle reaping** — removed by D1. Policy lives with the operator.
+- **Transparent reheat on access** — removed; cold access is an error by D1/§4.
```
as51340: remove old refs

src/utils/skip_list.hpp
```diff
@@ -273,6 +273,12 @@ class SkipListGc final {
   // Allocate a new head block if the caller's snapshot of head_ still matches.
   // Otherwise another thread already extended the chain; return that new head.
   Block *AllocateBlock(Block *expected_head) {
```
as51340: Should the method be noexcept then if we are trying to block OOM exceptions?

src/utils/skip_list.hpp
```diff
@@ -273,6 +273,12 @@ class SkipListGc final {
...
+    // Block allocation is internal GC bookkeeping called exclusively from
+    // ReleaseId, which is invoked from noexcept Accessor/ConstAccessor
+    // destructors and move-assignment operators. A tracked allocation that
+    // throws OutOfMemoryException out of a destructor causes std::terminate.
+    // Suppress OOM exceptions here for the same reason Run() does at line ~374.
```
as51340: Remove line ref

src/utils/skip_list.hpp
as51340: Also ReleaseId should then be noexcept method?

src/dbms/database.hpp
```diff
@@ -236,7 +237,15 @@ class Database {
 
   metrics::DatabaseMetricHandles *metric_handles() { return &metrics_.handles(); }
 
+  /// Used by hot/cold idle detection and SHOW DATABASES. Relaxed atomics: a coarse stat, not a fence.
+  void MarkUsed() noexcept {
+    last_used_ns_.store(std::chrono::steady_clock::now().time_since_epoch().count(), std::memory_order_relaxed);
+  }
```
as51340: Looks like unused code

src/query/frontend/ast/ast.hpp
```diff
@@ -4069,7 +4069,7 @@ class MultiDatabaseQuery : public memgraph::query::Query {
 
   DEFVISITABLE(QueryVisitor<void>);
 
-  enum class Action { CREATE, DROP, RENAME };
+  enum class Action { CREATE, DROP, RENAME, SUSPEND, RESUME };
```
as51340: Action : uint8_t

src/storage/v2/replication/replication_storage_state.cpp
```diff
@@ -45,11 +45,19 @@ void ReplicationStorageState::SaveLatestHistory() {
...
 }
 
+std::string ReplicationStorageState::LastEpochWithCommit(uint64_t const ldt) const {
+  if (!history.empty()) {
+    auto const &[history_epoch, history_ldt] = history.back();
+    return history_ldt != ldt ? std::string{epoch_.id()} : history_epoch;
+  }
+  return std::string{epoch_.id()};
+}
+
```
as51340: Shouldn't this do a search over deque? I mean what if ldt corresponds to some older epoch

src/dbms/rpc.cpp
```diff
@@ -146,6 +178,66 @@ void Load(memgraph::storage::replication::DropDatabaseRes *self, memgraph::slk::
...
+void Save(const memgraph::storage::replication::ResumeDatabaseRes &self, memgraph::slk::Builder *builder) {
+  memgraph::slk::Save(self.result, builder);
+}
+
+void Load(memgraph::storage::replication::ResumeDatabaseRes *self, memgraph::slk::Reader *reader) {
+  uint8_t res = 0;
+  memgraph::slk::Load(&res, reader);
+  if (!utils::NumToEnum(res, self->result)) {
+    throw SlkReaderException("Unexpected result line:{}!", __LINE__);
+  }
```
as51340: I think we have already functions for decoding and encoding enum 
```
inline void Load(utils::TypeId *obj, Reader *reader) {
  using enum_type = std::underlying_type_t<utils::TypeId>;
  enum_type obj_encoded;
  slk::Load(&obj_encoded, reader);
  *obj = utils::TypeId(std::bit_cast<enum_type>(obj_encoded));
}

template <utils::Enum T>
void Save(const T &enum_value, slk::Builder *builder) {
  slk::Save(std::to_underlying(enum_value), builder);
}
```

src/dbms/rpc.cpp
```diff
@@ -146,6 +178,66 @@ void Load(memgraph::storage::replication::DropDatabaseRes *self, memgraph::slk::
...
+  memgraph::slk::Save(self.result, builder);
+}
+
+void Load(memgraph::storage::replication::SuspendDatabaseRes *self, memgraph::slk::Reader *reader) {
+  uint8_t res = 0;
+  memgraph::slk::Load(&res, reader);
+  if (!utils::NumToEnum(res, self->result)) {
+    throw SlkReaderException("Unexpected result line:{}!", __LINE__);
+  }
+}
```
as51340, line 197: Same as below

as51340 (changes requested):
First batch

src/dbms/dbms_handler.hpp
```diff
@@ -306,6 +328,294 @@ class DbmsHandler {
...
+    // marker so the instance never silently drops a tenant.
+    for (const auto &[name, entry] : suspended_) {
+      std::string_view st = "COLD";
+      switch (entry.cold_reason) {
+        case SuspendedEntry::ColdReason::kRecoveryFailed:
+          st = "COLD (recovery failed)";
+          break;
+        case SuspendedEntry::ColdReason::kRecoveryFailedOom:
+          st = "COLD (recovery failed: out of memory)";
+          break;
```
as51340: Is that really important to handle it specifically?

src/dbms/dbms_handler.hpp
```diff
@@ -306,6 +328,294 @@ class DbmsHandler {
...
+        case SuspendedEntry::ColdReason::kSuspended:
+          st = "COLD";
```
as51340: not needed

src/dbms/dbms_handler.hpp
```diff
@@ -306,6 +328,294 @@ class DbmsHandler {
...
+      }
+      out.emplace_back(name, std::string{st});
+    }
+    for (auto &name : db_handler_.All()) {  // HOT names only (Handler::All() skips no-value shells)
+      if (!suspended_.contains(name)) out.emplace_back(std::move(name), "HOT");
+    }
+    return out;
+  }
+
+  /// C11/C12: minimal projection of a COLD tenant for SHOW STORAGE INFO ON <cold> (reverses HC-5).
```
as51340: can you remove these cryptic codes 😂 C11/C12, HC-5 I don't understand why we need them

src/dbms/dbms_handler.hpp
```diff
@@ -306,6 +328,294 @@ class DbmsHandler {
...
+      switch (it->second.cold_reason) {
+        case SuspendedEntry::ColdReason::kRecoveryFailed:
+          st = "COLD (recovery failed); stats unavailable";
+          break;
+        case SuspendedEntry::ColdReason::kRecoveryFailedOom:
+          st = "COLD (recovery failed: out of memory); stats unavailable";
+          break;
+        case SuspendedEntry::ColdReason::kSuspended:
+          st = "COLD (as-of-suspend snapshot)";
+          break;
```
as51340: IIFE could be useful here (same as above somewhere where I left the comment)

src/dbms/dbms_handler.hpp
```diff
@@ -306,6 +328,294 @@ class DbmsHandler {
...
+      return ColdShowInfo{.uuid = it->second.salient.uuid, .stats = it->second.cold_stats, .status = std::string{st}};
+    }
+    return std::nullopt;
+  }
+
+  /**
+   * @brief Resume a COLD tenant during replica SystemRecovery (rewire_replication=false).
+   *
+   * SR-1′(1): when MAIN's incoming HOT config names a tenant this replica holds COLD, Update() would
+   * throw UnknownDatabaseException on the COLD shell. Resume it first. rewire=false because recovery
```
as51340: Do we really throw UnknownDatabaseException? If yes, can we throw something more intuitive to the users

src/dbms/dbms_handler.hpp
```diff
@@ -306,6 +328,294 @@ class DbmsHandler {
...
+    for (const auto &[name, entry] : suspended_) {
+      out.push_back(storage::ColdTenantRecovery{.salient = entry.salient,
+                                                .stats = entry.cold_stats,
+                                                .has_epoch_meta = entry.has_epoch_meta,
+                                                .last_durable_timestamp = entry.last_durable_timestamp,
+                                                .current_epoch = entry.current_epoch,
+                                                .epoch_history = entry.epoch_history});
+    }
+    return out;
+  }
```
as51340, line 642: looks like a good candidate for ranges::transform

src/dbms/dbms_handler.hpp
```diff
@@ -526,6 +850,48 @@ class DbmsHandler {
...
+    int64_t last_used_ns;            //!< last-used stamp captured at suspend time
+    // Heartbeat metadata captured at suspend time (P4 — for C7 replication).
+    uint64_t last_durable_timestamp{0};  //!< repl_storage_state_.commit_ts_info_.ldt_ at suspend
+    uint64_t num_committed_txns{0};      //!< repl_storage_state_.commit_ts_info_.num_committed_txns_
+    // C10 (RE-9′): the FULL replication epoch state captured at suspend, so a promotion can rewrite it
+    // and a resume can restore it verbatim (a single epoch string is insufficient — the replica
+    // continuous-history check needs the whole deque; see PromoteColdTenants + Resume_).
+    std::string current_epoch;            //!< repl_storage_state_.epoch_.id() at suspend (post-promotion: new epoch)
+    storage::EpochHistory epoch_history;  //!< copy of repl_storage_state_.history at suspend
+    bool has_epoch_meta{false};           //!< false for pre-C10 V2 cold entries (no epoch fields) — skip override
```
as51340: Can you explain why we need to cache this information? Isn't this saved in replication storage state?

src/dbms/dbms_handler.hpp
```diff
@@ -683,6 +1061,15 @@ class DbmsHandler {
...
       return *db;
     }
+    // Cold-access query seam: a suspended (COLD) tenant still has an in-map gatekeeper, but
+    // access() refuses (it is not HOT), so db_handler_.Get returned nullopt. Distinguish "exists
+    // but suspended" from "truly unknown" with a clear, actionable message. We reuse
+    // UnknownDatabaseException so existing fallback catch sites (SetupDefault_, RestoreTenantProfiles_)
+    // keep treating a COLD tenant as not-currently-usable; the message is what the user sees.
+    if (suspended_.contains(name)) {
+      throw UnknownDatabaseException(
+          "Database \"{}\" is suspended (cold); run RESUME DATABASE {} before using it.", name, name);
```
as51340, line 1100: Yeah, maybe UnknownDatabaseException is not the best type for users

src/dbms/dbms_handler.cpp
```diff
@@ -103,6 +116,73 @@ struct Durability {
...
+  // F3: serialize via the single StorageInfoForEachField field list (storage.hpp) so adding a field
+  // updates this path AND the V3 SLK wire at once. Enums are stored as their underlying integer.
+  static nlohmann::json StatsToJson(const storage::StorageInfo &s) {
+    nlohmann::json j;
+    storage::StorageInfoForEachField(s, [&](const char *key, const auto &v) {
+      using T = std::remove_cvref_t<decltype(v)>;
+      if constexpr (std::is_enum_v<T>) {
+        j[key] = std::to_underlying(v);
+      } else {
+        j[key] = v;
```
as51340: I guess nlohmann has some smarter way of serialising enums

src/dbms/dbms_handler.cpp
```diff
@@ -103,6 +116,73 @@ struct Durability {
...
+        const auto raw = j.value(key, std::underlying_type_t<T>{0});
+        // storage_mode has an Enum::N sentinel (NumToEnum-validatable); isolation/compression do not, so
+        // direct-cast their underlying integer (mirrors the StorageInfo SLK Load in system_rpc.cpp).
+        if constexpr (std::is_same_v<T, storage::StorageMode>) {
+          if (!utils::NumToEnum(raw, v)) v = storage::StorageMode::IN_MEMORY_TRANSACTIONAL;
+        } else {
+          v = static_cast<T>(raw);
+        }
+      } else {
+        v = j.value(key, T{});
```
as51340: Same as above.
Also, I think it would be better to implement from_json and to_json functions instead of these kinda custom functions

src/dbms/dbms_handler.cpp
```diff
@@ -166,35 +250,169 @@ DbmsHandler::DbmsHandler(storage::Config config) : default_config_{std::move(con
...
   auto directories = std::set{std::string{kDefaultDB}};
 
-  // Restore databases
+  // Reconstruct a COLD tenant's rebuild metadata from a durable entry. The salient is rebuilt from
+  // the instance defaults (per-tenant salient persistence is out of scope, D10) with the durable name
+  // + uuid overlaid — identical to how New_(name, uuid, ...) builds it for a HOT tenant.
+  auto make_cold_entry =
+      [&](std::string_view nm, utils::UUID id, std::filesystem::path rel_dir, const nlohmann::json &json) {
+        SuspendedEntry entry;
+        entry.salient = default_config_.salient;
```
as51340, line 234: This smells again to me that it can be simplified using to_json and from_json

src/dbms/dbms_handler.cpp
```diff
@@ -166,35 +250,169 @@ DbmsHandler::DbmsHandler(storage::Config config) : default_config_{std::move(con
...
+        entry.last_durable_timestamp = json.value("last_durable_timestamp", uint64_t{0});
+        entry.num_committed_txns = json.value("num_committed_txns", uint64_t{0});
+        // C10 (RE-9′): restore the full epoch state. has_epoch_meta is false for a pre-C10 V2 entry
+        // (no current_epoch key) -> resume leaves the disk-recovered epoch/history intact (R15
+        // tolerant read), and PromoteColdTenants skips it so it is never corrupted with a "" boundary.
+        if (json.contains("current_epoch")) {
+          entry.has_epoch_meta = true;
+          entry.current_epoch = json.value("current_epoch", std::string{});
+          if (json.contains("epoch_history")) {
+            for (const auto &e : json.at("epoch_history")) {
```
as51340: Why we have to check if current_epoch exists and if epoch_history exists, don't we serialise them always?

src/dbms/dbms_handler.cpp
```diff
@@ -166,35 +250,169 @@ DbmsHandler::DbmsHandler(storage::Config config) : default_config_{std::move(con
...
-    const auto &name = entry.path().filename().string();
-    if (entry.is_directory() && !name.empty() && name.front() != '.') {
-      auto itr = directories.find(name);
-      if (itr == directories.end()) {
-        std::error_code dummy;
-        std::filesystem::remove_all(entry, dummy);
-      } else {
-        directories.erase(itr);
+  // Clean the unused directories. C14 (F1): if any durable entry failed to parse, the keep-set is
+  // incomplete — skip cleanup entirely so we never delete the data dir of a tenant we could not read.
```
as51340, line 337: What do we get by preserving such files? I guess we won't be ever able to read them?

as51340 (changes requested):
Second batch of comments

src/replication_handler/system_rpc.cpp
```diff
@@ -33,6 +33,80 @@ void Load(memgraph::parameters::ParameterInfo *self, memgraph::slk::Reader *read
...
+// F3 wire-contract tripwire: the StorageInfo SLK Save/Load below write/read each enum as
+// std::underlying_type_t<T>, and a peer must agree on that width. The encoding has always assumed a
+// 1-byte base; widening any of these enums would silently change the wire width and desync a same-version
+// peer. Make that a COMPILE error instead of a runtime corruption.
+static_assert(sizeof(std::underlying_type_t<memgraph::storage::StorageMode>) == 1);
+static_assert(sizeof(std::underlying_type_t<memgraph::storage::IsolationLevel>) == 1);
+static_assert(sizeof(std::underlying_type_t<memgraph::utils::CompressionLevel>) == 1);
+
+void Save(const memgraph::storage::StorageInfo &self, memgraph::slk::Builder *builder) {
+  memgraph::storage::StorageInfoForEachField(self, [&](const char * /*key*/, const auto &v) {
```
as51340: I see this function being repeated, can we deduplicate it?

src/replication_handler/system_rpc.cpp
```diff
@@ -33,6 +33,80 @@ void Load(memgraph::parameters::ParameterInfo *self, memgraph::slk::Reader *read
...
+    if constexpr (std::is_enum_v<T>) {
+      memgraph::slk::Save(std::to_underlying(v), builder);
+    } else {
+      memgraph::slk::Save(v, builder);
+    }
+  });
+}
+
+void Load(memgraph::storage::StorageInfo *self, memgraph::slk::Reader *reader) {
+  memgraph::storage::StorageInfoForEachField(*self, [&](const char * /*key*/, auto &v) {
```
as51340: Same, deduplicate the function

src/replication_handler/system_rpc.cpp
```diff
@@ -33,6 +33,80 @@ void Load(memgraph::parameters::ParameterInfo *self, memgraph::slk::Reader *read
...
+void Save(const memgraph::storage::ColdTenantRecovery &self, memgraph::slk::Builder *builder) {
+  memgraph::slk::Save(self.salient, builder);
+  memgraph::slk::Save(self.stats, builder);
+  memgraph::slk::Save(self.current_epoch, builder);
+  // EpochHistory is a std::deque (no direct slk support). Serialize element-by-element in the SAME wire
+  // layout slk uses for a std::vector (uint64 size + each (epoch, ldt) pair) so the format is unchanged,
+  // without materializing an intermediate vector copy.
+  const uint64_t history_size = self.epoch_history.size();
+  memgraph::slk::Save(history_size, builder);
+  for (const auto &entry : self.epoch_history) memgraph::slk::Save(entry, builder);
```
as51340: Add a function in the slk/serializaiton.hpp for deque serialisation

src/replication_handler/system_rpc.cpp
```diff
@@ -33,6 +33,80 @@ void Load(memgraph::parameters::ParameterInfo *self, memgraph::slk::Reader *read
...
+  memgraph::slk::Load(&self->salient, reader);
+  memgraph::slk::Load(&self->stats, reader);
+  memgraph::slk::Load(&self->current_epoch, reader);
+  uint64_t history_size = 0;
+  memgraph::slk::Load(&history_size, reader);
+  self->epoch_history.clear();
+  for (uint64_t i = 0; i < history_size; ++i) {
+    std::pair<std::string, uint64_t> entry;
+    memgraph::slk::Load(&entry, reader);
+    self->epoch_history.push_back(std::move(entry));
```
as51340: Add a function in the slk/serialization

src/replication_handler/system_rpc.cpp
```diff
@@ -90,6 +189,19 @@ void Load(memgraph::replication::SystemRecoveryResV1 *self, memgraph::slk::Reade
...
+void Save(const memgraph::replication::SystemRecoveryResV2 &self, memgraph::slk::Builder *builder) {
+  memgraph::slk::Save(self.result, builder);
+}
+
+void Load(memgraph::replication::SystemRecoveryResV2 *self, memgraph::slk::Reader *reader) {
+  uint8_t res = 0;
+  memgraph::slk::Load(&res, reader);
+  if (!utils::NumToEnum(res, self->result)) {
+    throw SlkReaderException("Unexpected result line:{}!", __LINE__);
+  }
```
as51340: this looks repeated from somewhere in the dbms, try to deduplicate it

src/replication_handler/replication_handler.cpp
```diff
@@ -306,6 +306,14 @@ bool ReplicationHandler::DoToMainPromotion(const utils::UUID &main_uuid, bool co
       spdlog::trace("New timestamp is {} for the database {}.", storage->timestamp_, db_acc->name());
     });
 
+#ifdef MG_ENTERPRISE
+    // STEP 4b) C10 (PR-1): COLD (suspended) tenants are invisible to the ForEach loops above (no-value
+    // gatekeeper), so rewrite their durable epoch metadata to the same new epoch directly. Metadata-only
+    // (no force-resume); takes only DbmsHandler::lock_ + KVStore, never repl_state, so it is safe to call
+    // while holding the repl_state write lock here. See DbmsHandler::PromoteColdTenants.
+    dbms_handler_.PromoteColdTenants(new_epoch.id());
+#endif
```
as51340: Would it be simpler maybe to rely on RPC reconciling the tenant state than overwriting durability?

specs/hot-cold-databases.md
```diff
@@ -0,0 +1,362 @@
...
+
+Everything else in this spec follows from that sentence. A suspended (cold) database
+holds exactly the same data as it did when hot — it simply isn't materialized in memory.
+Resuming reconstructs the identical graph from durable storage. If any decision in this
+document seems surprising, check it against this principle: it is almost always the
+reason.
+
+The feature is part of enterprise multi-tenancy and is on by default in enterprise
+builds — there is no opt-in flag. (It originally shipped behind an
+`--experimental-enabled=hot-cold-databases` flag; once it stabilized the flag was
```
as51340: Not relevant anymore that there was experimental flag

specs/hot-cold-databases.md
```diff
@@ -0,0 +1,362 @@
...
+queryable, which is confusing and operationally fragile. Treating it as replicated DDL
+makes the behavior predictable: the cluster has one hot/cold map, and it is the MAIN's.
+
+A consequence, accepted deliberately: suspending a database on the MAIN also tears it
+down on replicas, evicting any reader connected to a replica's copy. That is the price of
+a single coherent cluster-wide state, and it is consistent with how other DDL behaves.
+
+This is also why hot/cold is the MAIN's decision rather than a per-node or consensus one.
+A database that is idle on the MAIN but actively read on a replica still follows the MAIN
+when the MAIN suspends it — the replica's readers are evicted. A node-local or quorum
```
as51340: What does evicted mean? Sessions dropped or?

src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -207,6 +207,25 @@ std::optional<DatabaseAccess> GetDatabaseAccessor(dbms::DbmsHandler *dbms_handle
...
     return std::optional{std::move(acc)};
   } catch (const dbms::UnknownDatabaseException &) {
+#ifdef MG_ENTERPRISE
+    // A data or recovery RPC referenced a tenant that is not HOT on this replica. If it
+    // is a known COLD (suspended) tenant, reheat it from disk INLINE on this (single-threaded) RPC
+    // thread — never by waiting for another RPC, which would self-deadlock the one server thread.
+    // ResumeByUUID runs synchronous local recovery (snapshot + WAL), so this is a bounded,
+    // self-contained reheat. This barrier lets a data delta that arrives for a still-COLD tenant
+    // (reconnect/lag, before the RESUME system delta is applied) be applied rather than dropped.
+    if (auto resumed = dbms_handler->ResumeByUUID(uuid); resumed.has_value()) {
```
as51340: Isn't this wrong in a sense that only the explicit request for resuming the database will resume it? Otherwise, this drifts relatively simply from main

specs/hot-cold-databases.md
andrejtonev: analytical database has no wals, so we can't guarantee the durability is correct. we only support snapshot + wal enabled.
ignoring on disk in general.

src/dbms/database.hpp
```diff
@@ -11,6 +11,7 @@
 
 #pragma once
 
+#include <atomic>
```
as51340: not needed

specs/hot-cold-databases.md
andrejtonev: we are replicating the HOT/COLD status, so the whole cluster should see the same hot databases. replicas should not be able to read from cold. 
but we should have a follow up for HA, where this is a consensus and split on replication groups.

specs/hot-cold-databases.md
andrejtonev: makes sense. but we also need to do this at runtime, so we can't abort if a db cannot recover at runtime.

specs/hot-cold-databases.md
andrejtonev: I'll ping product on this.

src/storage/v2/replication/replication_storage_state.cpp
andrejtonev: dead code and removed

src/dbms/dbms_handler.hpp
```diff
@@ -306,6 +329,321 @@ class DbmsHandler {
...
+   * @brief Is @p name suspended here AND under this exact @p uuid? Distinguishes a tenant that is
+   * still the same COLD tenant (refresh its metadata) from one MAIN drop+recreated under the same
+   * name while this replica kept the OLD one COLD (a stale shell that must be dropped and rebuilt —
+   * otherwise per-uuid Suspend/Resume RPC for the new uuid misses forever and the replica stays
+   * permanently BEHIND).
+   */
+  bool IsSuspendedWithUuid(std::string_view name, const utils::UUID &uuid) const {
+    auto rd = std::shared_lock{lock_};
+    auto it = suspended_.find(name);
+    return it != suspended_.end() && it->second.salient.uuid == uuid;
```
as51340: Isn't a solution to these kinda duplicated functions that suspended_ saves databases' UUID instead of names

src/dbms/dbms_handler.hpp
andrejtonev: trying to make it as obvious as possible. this will be used by agents, so more informative errors are expected.

src/dbms/dbms_handler.hpp
andrejtonev: distinguishing now

src/dbms/dbms_handler.hpp
```diff
@@ -306,6 +329,321 @@ class DbmsHandler {
...
+
+  /**
+   * @brief Is @p name suspended here AND under this exact @p uuid? Distinguishes a tenant that is
+   * still the same COLD tenant (refresh its metadata) from one MAIN drop+recreated under the same
+   * name while this replica kept the OLD one COLD (a stale shell that must be dropped and rebuilt —
+   * otherwise per-uuid Suspend/Resume RPC for the new uuid misses forever and the replica stays
+   * permanently BEHIND).
+   */
+  bool IsSuspendedWithUuid(std::string_view name, const utils::UUID &uuid) const {
+    auto rd = std::shared_lock{lock_};
```
as51340: could there be some concurrency problem for reusing lock_ for suspended_ and db_handler_?

src/dbms/dbms_handler.hpp
```diff
@@ -306,6 +329,321 @@ class DbmsHandler {
...
+      const std::string_view st = [&]() -> std::string_view {
+        switch (entry.cold_reason) {
+          case SuspendedEntry::ColdReason::kRecoveryFailed:
+            return "COLD (recovery failed)";
+          case SuspendedEntry::ColdReason::kRecoveryFailedOom:
+            return "COLD (recovery failed: out of memory)";
+          case SuspendedEntry::ColdReason::kSuspended:
+            return "COLD";
+        }
+        return "COLD";  // unreachable; silences -Wreturn-type
```
as51340: if unreachable, use std::unreachable

src/dbms/dbms_handler.hpp
andrejtonev: since the cold tenant has no in memory representation, we need to cache this so we have the last know info before suspending it. that way we can give the user useful info without recovering the whole tenant.

src/dbms/dbms_handler.hpp
```diff
@@ -306,6 +329,321 @@ class DbmsHandler {
...
+      const std::string_view st = [&]() -> std::string_view {
+        switch (it->second.cold_reason) {
+          case SuspendedEntry::ColdReason::kRecoveryFailed:
+            return "COLD (recovery failed); stats unavailable";
+          case SuspendedEntry::ColdReason::kRecoveryFailedOom:
+            return "COLD (recovery failed: out of memory); stats unavailable";
+          case SuspendedEntry::ColdReason::kSuspended:
+            return "COLD (as-of-suspend snapshot)";
+        }
+        return "COLD (as-of-suspend snapshot)";  // unreachable
```
as51340: This looks like it could be reused in AllWithHotColdStatus function

src/dbms/dbms_handler.hpp
andrejtonev: new exception and is non retrievable client error.

src/dbms/dbms_handler.cpp
andrejtonev: backward compatibly

src/dbms/dbms_handler.cpp
andrejtonev: trying to not loose random data 🤷

src/replication_handler/replication_handler.cpp
andrejtonev: COLD tenant has no live storage and is invisible to the replication `ForEach` loops, so there is no RPC channel to carry the new epoch to it. So the epoch must be written somewhere the cold tenant can read on resume

src/dbms/replication_handlers.cpp
```diff
@@ -196,7 +204,168 @@ void RenameDatabaseHandler(memgraph::system::ReplicaHandlerAccessToState &system
...
+    // replica does NOT silently diverge — the system timestamp still advances via FinalizeSystemTxRpc
+    // and the failed apply flags this replica BEHIND, so it reconciles on the next ordered re-sync /
+    // SystemRecovery.
+  } catch (const std::exception &e) {
+    // FAILURE: log so a diverged replica has a diagnostic anchor.
+    spdlog::error(
+        "SuspendDatabaseHandler: failed to apply suspend for database {}: {}", std::string{req.uuid}, e.what());
+  } catch (...) {
+    spdlog::error("SuspendDatabaseHandler: failed to apply suspend for database {}", std::string{req.uuid});
+  }
```
as51340, line 261: do we need both std::exception and ...?

src/dbms/replication_handlers.cpp
```diff
@@ -196,7 +204,168 @@ void RenameDatabaseHandler(memgraph::system::ReplicaHandlerAccessToState &system
...
+  ResumeDatabaseRes res(ResumeDatabaseRes::Result::FAILURE);
+
+  // Ignore if no license
+  if (!license::global_license_checker.IsEnterpriseValidFast()) {
+    spdlog::error(
+        "Handling ResumeDatabase, an enterprise RPC message, without license. Check your license status by running "
+        "SHOW LICENSE INFO.");
+    rpc::SendFinalResponse(res, request_version, res_builder);
+    return;
+  }
```
as51340: can you move this maybe to the place where dispatching of handlers is being done?

src/dbms/replication_handlers.cpp
```diff
@@ -196,7 +204,168 @@ void RenameDatabaseHandler(memgraph::system::ReplicaHandlerAccessToState &system
...
+      // latches BEHIND and SystemRecovery supplies it.
+      res = ResumeDatabaseRes(ResumeDatabaseRes::Result::NO_NEED);
+    }
+    // RECOVERY_FAILED leaves FAILURE -> diverge-then-reconcile via SystemRecovery.
+  } catch (const std::exception &e) {
+    // FAILURE: log so a diverged replica has a diagnostic anchor.
+    spdlog::error("ResumeDatabaseHandler: failed to apply resume for database {}: {}", std::string{req.uuid}, e.what());
+  } catch (...) {
+    spdlog::error("ResumeDatabaseHandler: failed to apply resume for database {}", std::string{req.uuid});
+  }
```
as51340, line 312: Do we need both ... and std::exception

src/dbms/replication_handlers.cpp
```diff
@@ -221,24 +390,105 @@ bool SystemRecoveryHandler(DbmsHandler &dbms_handler, const std::vector<storage:
...
+  // tenant MAIN has dropped while this replica holds it COLD would otherwise be in neither the
+  // incoming sets nor `old` and would never be reconciled away (permanent divergence). The DROP of
+  // a COLD tenant is handled cold-aware by Delete() (see DeleteCold_). AllWithHotColdStatus() is
+  // de-duplicated (a SUSPENDING-transient tenant is listed once, as COLD).
+  std::vector<std::string> old;
+  {
+    auto hot_cold = dbms_handler.AllWithHotColdStatus();
+    old.reserve(hot_cold.size());
+    for (auto &[name, _status] : hot_cold) old.emplace_back(std::move(name));
+  }
```
as51340: std::ranges::keys + std::ranges::transform

src/replication_handler/include/replication_handler/replication_handler.hpp
```diff
@@ -67,6 +67,10 @@ void SystemRestore(ReplicationClient &client, system::System &system, dbms::Dbms
   struct DbInfo {
     std::vector<storage::SalientConfig> configs;
     uint64_t last_committed_timestamp;
+    // Hot/cold: COLD set carried so a reconnecting/lagging replica converges to {HOT ∪ COLD}.
+    // One ColdTenantRecovery per suspended tenant (salient + stats + epoch); empty on non-enterprise /
+    // no-license.
+    std::vector<storage::ColdTenantRecovery> cold_databases;
```
as51340, line 77: If this is a structure being serialized and sent over the network then we need to version it, e.g DbInfoV1, DbInfoV2

src/storage/v2/storage.hpp
```diff
@@ -249,6 +302,19 @@ class Storage {
 
   StorageMode GetStorageMode() const noexcept;
 
+  // True iff this storage's durability mode keeps BOTH periodic snapshots AND a WAL
+  // chain — the precondition for hot/cold suspend (suspend tears down RAM and relies
+  // on {snapshot + WAL} on disk to recover). PERIODIC_SNAPSHOT-only or DISABLED is
+  // NOT suspendable.
+  [[nodiscard]] bool IsDurabilityCompleteForSuspend() const {
+    return config_.durability.snapshot_wal_mode == Config::Durability::SnapshotWalMode::PERIODIC_SNAPSHOT_WITH_WAL;
+  }
```
as51340, line 306: Why it's not ok to have only periodic snapshots?

src/storage/v2/inmemory/storage.hpp
```diff
@@ -912,6 +917,12 @@ class InMemoryStorage final : public Storage {
   std::filesystem::path lock_file_path_;
   std::unique_ptr<utils::OutputFile> lock_file_handle_ = std::make_unique<utils::OutputFile>();
 
+  // Hot/cold suspend: when Suspend_ takes a consolidating snapshot before teardown, it calls
+  // DisableExitSnapshot() so the destructor (triggered by finish_suspend()) does NOT write a
+  // second snapshot-on-exit that would push the WAL position past the consolidated snapshot's
+  // durable timestamp and break fast resume (WAL replay must start from after the snapshot).
+  bool exit_snapshot_enabled_{true};
```
as51340: Does this need to be atomic?

src/utils/gatekeeper.hpp
```diff
@@ -278,13 +325,127 @@ struct Gatekeeper {
...
+
+  // COLD -> RESUMING (single-flight).
+  // Only the first caller wins the COLD->RESUMING transition; concurrent
+  // callers that already see RESUMING get false.  This acts as a single-flight
+  // token: the winner is responsible for loading the value and then
+  // move-assigning a freshly-built HOT Gatekeeper over this shell to complete
+  // the transition.  Call abort_resume() if loading fails.
+  bool begin_resume() {
+    auto guard = std::unique_lock{pimpl_->mutex_};
+    if (pimpl_->state_ != GatekeeperState::COLD) return false;
```
as51340, line 420: I guess we could change this if condition to MG_ASSERT since there is unique lock taken on mutex_ that is protecting us?

src/utils/gatekeeper.hpp
```diff
@@ -278,13 +325,127 @@ struct Gatekeeper {
...
-      pimpl_->cv_.wait(lock, [this] { return pimpl_->count_ == 0; });
+      auto const terminal_and_drained = [this] {
+        return (pimpl_->state_ == GatekeeperState::HOT || pimpl_->state_ == GatekeeperState::COLD) &&
+               pimpl_->count_ == 0;
+      };
+      // Behaviour-preserving diagnosability: keep waiting unboundedly for the real condition (a slow
+      // accessor drain on shutdown is legitimate and must not be cut short), but if it has not been met
+      // after a generous interval, log WHY. Destroying a gatekeeper stuck in SUSPENDING/RESUMING is a
+      // caller ordering error that would otherwise hang here silently forever with no diagnostic.
+      while (!pimpl_->cv_.wait_for(lock, std::chrono::seconds{30}, terminal_and_drained)) {
```
as51340: let's decrease this timeout a bit, we have termination grace period seconds parameter on helm charts set to 30s.

as51340 (changes requested):

specs/hot-cold-databases.md
as51340: what's the issue in guaranteeing correct durability with snapshots only?

specs/hot-cold-databases.md
as51340: let's sync about this tomorrow

src/dbms/replication_handlers.cpp
```diff
@@ -221,24 +390,105 @@ bool SystemRecoveryHandler(DbmsHandler &dbms_handler, const std::vector<storage:
...
+    const auto name = std::string{*config.name.str_view()};
+    // Exempt COLD names from the leftover-delete loop below — a replica-HOT tenant that MAIN
+    // now lists COLD must be reconciled to COLD, not dropped.
+    std::erase(old, name);
+
+    if (dbms_handler.IsSuspended(name)) {
+      // Same tenant still COLD here: refresh MAIN's as-of-suspend stats snapshot AND epoch metadata
+      // (a converged-then-promoted replica needs MAIN's epoch to record the right boundary).
+      if (dbms_handler.IsSuspendedWithUuid(name, config.uuid)) {
+        dbms_handler.ApplyColdRecoveryMeta(name, cold);
```
as51340, line 451: Shouldn't this update the state on disk, not only in-memory state?

src/dbms/dbms_handler.cpp
```diff
@@ -672,4 +988,623 @@ void DbmsHandler::RestoreTriggers(query::InterpreterContext *ic) {
...
+      if (!durability_->Put(Durability::GenKey(name),
+                            Durability::GenColdVal(entry.salient.uuid,
+                                                   entry.rel_dir,
+                                                   entry.last_durable_timestamp,
+                                                   entry.num_committed_txns,
+                                                   entry.current_epoch,
+                                                   entry.epoch_history,
+                                                   entry.cold_stats))) {
+        spdlog::warn("hot/cold promotion: failed to persist new epoch for cold database '{}'.", name);
+      }
```
as51340: What's the worst thing that can happen if this writing to the disk fails?

specs/hot-cold-databases.md
andrejtonev: problem: someone is using a db on replica, but main has suspended the db.
solutions (possible): 
1. do as drop database force with deferred delete
2. fail on replica and require a system recovery
3. 2 phase commit with cluster concensus

+ needs stress tests and jepsen maybe

specs/hot-cold-databases.md
andrejtonev: agreed, on recovery hot database failure is terminal. note for future dev to enable recovery failure on this path.

specs/hot-cold-databases.md
andrejtonev: need product feedback

specs/hot-cold-databases.md
andrejtonev: need product feedback

specs/hot-cold-databases.md
andrejtonev: Need to follow up on this. depends on what we decide to do when a replica needs to suspend. do we force disconnect users or defer the suspension or fail.

src/storage/v2/storage.hpp
andrejtonev: need product feedback

src/dbms/inmemory/replication_handlers.cpp
andrejtonev: this is left from an automatic version. we fail here and wait for system recovery.

src/dbms/replication_handlers.cpp
```diff
@@ -221,24 +390,105 @@ bool SystemRecoveryHandler(DbmsHandler &dbms_handler, const std::vector<storage:
...
+        continue;
+      }
+      // Same name, DIFFERENT uuid: MAIN drop+recreated this tenant while this replica kept the OLD
+      // one COLD. The local shell points at the old tenant's data — drop it so the rebuild below
+      // recreates MAIN's new tenant. Leaving it would make every per-uuid Suspend/Resume RPC for the
+      // new uuid return NON_EXISTENT, latching this replica BEHIND permanently with no self-heal.
+      if (const auto del = dbms_handler.Delete(name); !del && del.error() != DeleteError::NON_EXISTENT) {
```
andrejtonev, line 469: shouldn't we update config instead of delete and leaving it without a db?

src/dbms/dbms_handler.cpp
andrejtonev: I'll check if all of these are already persisted in snapshots and wals, so are irrelevant for recovery.

andrejtonev: most of these should not be duplicated, since already present in other durable data. save only runtime info that we present to the user as a best guess.

specs/hot-cold-databases.md
andrejtonev: product is fine with transactional with wal only

specs/hot-cold-databases.md
andrejtonev: replica needs to follow main, any deviation causes a failure on the replica and a system/db recovery from main

specs/hot-cold-databases.md
andrejtonev: product defined wanted metrics

specs/hot-cold-databases.md
andrejtonev: we fail and let the main recover replica

src/dbms/dbms_handler.hpp
andrejtonev: Re-keying would turn the single uuid scan into O(log n) but turn all the name lookups into linear scans, and it would break the intentional noexcept find(string_view) window (the std::less<> transparent comparator, with a comment explicitly warning not to change it). It also wouldn't actually collapse the two functions — they answer different questions (existence vs. existence-and-same-uuid).

src/dbms/replication_handlers.cpp
andrejtonev: just a split so we cover all exceptions and write out the exception when present.

src/storage/v2/storage.hpp
andrejtonev: product confirmed transactional with wal for now.

src/dbms/dbms_handler.cpp
andrejtonev: cleaned the structs

src/dbms/dbms_handler.hpp
andrejtonev: No — it's deliberate. One lock makes the HOT↔COLD handoff atomic to readers; separate locks would introduce lost/double-visibility windows. Lock order is consistently lock_→gatekeeper, never inverted, so no deadlock.

src/dbms/replication_handlers.cpp
andrejtonev: made a helper function

src/replication_handler/include/replication_handler/replication_handler.hpp
andrejtonev: this is local representation only

src/storage/v2/inmemory/storage.hpp
andrejtonev: no need, the sequence is in order on a single thread

src/utils/gatekeeper.hpp
andrejtonev: the asserts are only for failure rollback paths. the higher level functions just fail

src/dbms/replication_handlers.cpp
andrejtonev: fixed

src/dbms/replication_handlers.cpp
andrejtonev: have to drop and rebuild

src/dbms/dbms_handler.hpp
```diff
@@ -306,6 +338,269 @@ class DbmsHandler {
...
+   * consumers run on their own threads); a future arm must too.
+   */
+  void SetOnResume(std::function<void(DatabaseAccess)> cb) { on_resume_ = std::move(cb); }
+
+  /**
+   * @brief Test-only override of Resume_'s retry/timeout knobs (see ResumeRetryPolicy). Defaults to
+   *        the production values; a test can shrink them to exercise the bounded single-flight
+   *        loser-retry path deterministically in milliseconds instead of minutes.
+   */
+  void SetResumeRetryPolicyForTest(ResumeRetryPolicy policy) { resume_retry_policy_ = policy; }
```
as51340: I think this is a bad practice to add methods just for tests. You want to use dependency injection

src/dbms/dbms_handler.cpp
```diff
@@ -134,8 +185,12 @@ struct Durability {
...
     }
 
-    // Set version
-    durability->Put("version", "V1");
+    // V1 -> V2 is purely additive: V1 entries are all HOT ({uuid, rel_dir}) and remain valid as-is;
+    // V2 only ADDS optional COLD entries (written by SUSPEND). Reads default `cold` to false and
+    // `cold_stats` to zeros, so an unmigrated V1 entry is read correctly with no data movement.
+
+    // Set version to the current schema (V2).
+    durability->Put("version", "V2");
```
as51340: Make sure to run ISSU test just in case

src/dbms/dbms_handler.cpp
```diff
@@ -166,35 +221,132 @@ DbmsHandler::DbmsHandler(storage::Config config) : default_config_{std::move(con
...
+      // and silently skip a valid tenant — mirrors the OOM-aware classification of the HOT path below.
+      durability_view_incomplete = true;
+      spdlog::error(
+          "Corrupt durable database entry '{}' ({}); skipping it. Its on-disk data is left untouched and "
+          "directory cleanup is SKIPPED this boot so no data is deleted while the durable view is incomplete.",
+          key,
+          e.what());
+      continue;
+    }
+    // The on-disk directory must survive the cleanup pass below for BOTH hot and cold tenants — a
```
as51340, line 276: What if you're recovering v1 durability version?

src/dbms/dbms_handler.cpp
```diff
@@ -166,35 +221,132 @@ DbmsHandler::DbmsHandler(storage::Config config) : default_config_{std::move(con
...
-    MG_ASSERT(new_db.has_value(), "Failed while creating database {}.", name);
+    // Parse defensively. A corrupt/truncated entry must leave the instance starting DEGRADED
+    // (skip the entry, keep the data on disk) rather than throwing out of the ctor and bricking boot.
+    nlohmann::json json;
+    utils::UUID uuid;
+    std::filesystem::path rel_dir;
+    try {
+      json = nlohmann::json::parse(config_json);
+      uuid = json.at("uuid").get<utils::UUID>();
+      rel_dir = json.at("rel_dir").get<std::filesystem::path>();
```
as51340: extract "uuid", "rel_dir" into common constants

src/dbms/dbms_handler.cpp
```diff
@@ -315,6 +478,16 @@ DbmsHandler::DeleteResult DbmsHandler::TryDelete(std::string_view db_name, syste
...
+  // Cold-tenant fast path.
+  if (suspended_.contains(db_name)) {
+    auto res = DeleteCold_(db_name);
+    if (!res) return std::unexpected{res.error()};
+    if (transaction) {
+      transaction->AddAction<DropDatabase>(*res);
+    }
+    return {};
+  }
+
```
as51340, line 495: Extract into a common function since it is the same as lines 423-432

src/dbms/dbms_handler.cpp
```diff
@@ -334,11 +507,25 @@ DbmsHandler::DeleteResult DbmsHandler::Delete(std::string_view db_name, system::
...
   auto wr = std::lock_guard(lock_);
+  // Cold-tenant fast path: scan suspended_ by uuid (replica apply path — no transaction arg).
+  {
+    auto it = FindSuspendedByUuid_(uuid);
+    if (it != suspended_.end()) {
+      auto res = DeleteCold_(it->first);
+      if (!res) return std::unexpected{res.error()};
+      return {};
+    }
+  }
```
as51340: Maybe first check if in-memory and only if not go to suspended? I saw that order somewhere before in the code, seems more logical to me

src/dbms/dbms_handler.cpp
```diff
@@ -672,4 +919,649 @@ void DbmsHandler::RestoreTriggers(query::InterpreterContext *ic) {
...
+    std::optional<DatabaseAccess> sacc;
+    {
+      auto rd = std::shared_lock{lock_};
+      auto a = db_handler_.Get(name);
+      if (!a) return std::unexpected{SuspendError::NON_EXISTENT};
+      sacc = std::move(*a);
+    }  // release lock_ BEFORE joining consumer threads
+    if (on_suspend_) {
+      on_suspend_(*sacc);
+      streams_stopped = true;
```
as51340: I would rename this flag to on_suspend_run since in the future the callback might do more than just streams

src/dbms/dbms_handler.cpp
```diff
@@ -672,4 +919,649 @@ void DbmsHandler::RestoreTriggers(query::InterpreterContext *ic) {
...
+    constexpr auto kSnapPollStep = std::chrono::milliseconds(10);
+    constexpr auto kSnapWaitTimeout = std::chrono::seconds(10);
+    auto snap = inmem->CreateSnapshot(/*force=*/false, "suspend");
+    if (!snap.has_value() && snap.error() == CreateSnapshotError::AlreadyRunning) {
+      const auto snap_deadline = std::chrono::steady_clock::now() + kSnapWaitTimeout;
+      while (!snap.has_value() && snap.error() == CreateSnapshotError::AlreadyRunning &&
+             std::chrono::steady_clock::now() < snap_deadline) {
+        std::this_thread::sleep_for(kSnapPollStep);
+        snap = inmem->CreateSnapshot(/*force=*/false, "suspend");
+      }
```
as51340: 1. If we do take snapshot on exit, what's the problem with supporting analytical mode.
2. kSnapWaitTimeout set to 10s doesn't seem correct.
3. Why even force snapshot creation at suspend, I would just suspend it with the durability state that exists at that moment

src/dbms/dbms_handler.cpp
```diff
@@ -672,4 +919,649 @@ void DbmsHandler::RestoreTriggers(query::InterpreterContext *ic) {
...
+    // — whether it returns false OR throws (KVStore::Put is not noexcept; bad_alloc under memory pressure)
+    // — degrades to a warning and never rolls back. Doing it BEFORE the suspended_ insert means a throwing
+    // Put unwinds through the rollback guard with suspended_ still clean, so the tenant cleanly stays HOT
+    // instead of becoming HOT-but-listed-COLD (HC-2). A missing marker only makes the tenant recover HOT
+    // on restart; MAIN's data stays durable on disk regardless.
+    if (durability_) {
+      bool marker_persisted = false;
+      try {
+        marker_persisted = durability_->Put(cold_key, cold_val);
+      } catch (...) {  // NOLINT(bugprone-empty-catch) — best-effort marker; fall through to the warning
```
as51340: I would catch the exception and log it

src/dbms/dbms_handler.cpp
```diff
@@ -672,4 +919,649 @@ void DbmsHandler::RestoreTriggers(query::InterpreterContext *ic) {
...
+  // SUSPENDING -> COLD; if the guard fired afterwards it would abort_suspend() on a non-SUSPENDING
+  // state and trip the debug assert.
+  rollback.Disable();
+
+  // Disable exit snapshot NOW — past the abort point-of-no-return (rollback is disabled, so the
+  // tenant is committed to teardown). The storage object lives in the gatekeeper until
+  // finish_suspend() destroys it, so inmem_for_disable is still valid here. Doing this after
+  // rollback.Disable() ensures an aborted suspend never leaves the tenant re-HOT with exit-snapshot
+  // permanently disabled (there is no EnableExitSnapshot() API to undo it).
+  if (inmem_for_disable) inmem_for_disable->DisableExitSnapshot();
```
as51340: + to add what I wrote above: I think it is enough to rely on users setting --snapshot-on-exit=true if they want snapshot before suspend

as51340: or a separate flag because snapshot creation could take literally hours

src/dbms/dbms_handler.cpp
```diff
@@ -672,4 +919,649 @@ void DbmsHandler::RestoreTriggers(query::InterpreterContext *ic) {
...
+        // COLD-fallback path restarts Phase A and counts it, so a persistent on_resume_ failure (e.g.
+        // corrupt trigger metadata) is surfaced as RECOVERY_FAILED instead of looping forever.
+      }
+      // A live (RESUMING) winner: keep waiting — do NOT advance the abort bound (this is the HC-1 fix:
+      // the absolute_ceiling can fire purely because the winner stayed RESUMING, and that must not count).
+      if (winner_still_resuming) continue;
+      // Bound the retry — a single transient abort is retried; a persistent on_resume_ failure
+      // (e.g. corrupt trigger metadata) is surfaced as RECOVERY_FAILED instead of looping forever.
+      if (++cold_fallback_restarts > resume_retry_policy_.max_cold_fallback_restarts)
+        return std::unexpected{ResumeError::RECOVERY_FAILED};
```
as51340: relying on counters is problematic, I would rather rely on specific time passed = timeout.

src/dbms/dbms_handler.cpp
```diff
@@ -672,4 +919,649 @@ void DbmsHandler::RestoreTriggers(query::InterpreterContext *ic) {
...
+          // the load-bearing note on Gatekeeper::operator=(Gatekeeper&&) in utils/gatekeeper.hpp.
+          *gk = std::move(fresh);
+          if (it != suspended_.end()) suspended_.erase(it);
+          // The tenant is now HOT. Counter increment + gauge set are atomic/non-throwing, so they
+          // do not break the publish block's no-throw guarantee; done under lock_ so the gauge reads a
+          // consistent suspended_ size.
+          metrics::Metrics().global.database_resumes->Increment();
+          UpdateColdGauge_();
+        }
+      } catch (...) {
```
as51340: log the exception

as51340: or add both catching std::exception (first) and then ... if not a standard exception was thrown

as51340: there are multiple places for this, just change everywhere, don't want to write the same comment N times

src/dbms/dbms_handler.cpp
```diff
@@ -672,4 +919,649 @@ void DbmsHandler::RestoreTriggers(query::InterpreterContext *ic) {
...
+      // the gatekeeper is already HOT here, so any throw escaping to the outer catch(...) would call
+      // abort_resume() on a HOT gatekeeper and trip its DMG_ASSERT(state == RESUMING) -> terminate.
+      // GenKey/GenVal allocate (fmt/nlohmann) and can throw bad_alloc. A failure (bool or throw)
+      // degrades to a warning: the tenant is live HOT regardless; on the next restart a stale cold
+      // marker only makes it recover COLD, and it is resumable again.
+      try {
+        if (durability_) {
+          auto wr = std::lock_guard{lock_};
+          if (!suspended_.contains(name)) {
+            if (!durability_->Put(Durability::GenKey(name), Durability::GenVal(salient.uuid, rel_dir))) {
```
as51340, line 1470: is this fatal? What if failing to disk fails and then the instance restarts? The durability would say it's still cold

src/dbms/handler.hpp
```diff
@@ -139,7 +203,14 @@ class Handler {
...
       auto task = [gk = std::move(itr->second), post_delete_func = std::forward<Func>(post_delete_func)]() mutable {
-        gk.~Gatekeeper<T>();
+        // Destroy the gatekeeper exactly once, via natural scope — NOT an explicit gk.~Gatekeeper<T>()
+        // followed by the captured gk being destructed again when this lambda is destroyed (that is a
+        // double-destruction: [basic.life] UB, reading a destroyed object's pimpl_). Moving into a
+        // block-scoped local runs the blocking ~Gatekeeper once here; the moved-from capture then
+        // destroys cleanly (null pimpl_) with the lambda.
+        {
+          auto dying = std::move(gk);
+        }
```
as51340, line 213: Was that a bug before?

tests/stress/standalone/native/workloads/hot_cold/workload.py
```diff
@@ -0,0 +1,475 @@
...
+Hammers the hot/cold 4-state gatekeeper (HOT -> SUSPENDING -> COLD ->
+RESUMING -> HOT) under two-way concurrent contention:
+
+  1. WRITER threads   — open a per-tenant connection, USE DATABASE tenant_k,
+                        BEGIN / CREATE M nodes / COMMIT; track committed count.
+  2. SUSPENDER thread — issue SUSPEND DATABASE tenant_k on random tenants via
+                        the default-db control connection.
+  3. RESUMER thread   — issue RESUME DATABASE tenant_k on random tenants via
+                        the default-db control connection.
+  4. READER thread    — USE DATABASE tenant_k / MATCH (n) RETURN count(n);
```
as51340, line 27: I would extend this test with dbms operations: create tenant, drop tenant, rename ... with force no force whatever

src/dbms/dbms_handler.cpp
```diff
@@ -672,4 +919,649 @@ void DbmsHandler::RestoreTriggers(query::InterpreterContext *ic) {
...
+            "hot/cold suspend: failed to persist the cold durability marker for '{}'; the database is "
+            "suspended in memory but will recover HOT on restart.",
+            name);
+      }
+    }
+    // In-memory commit — the tenant is now COLD. Counts every successful Suspend_ (user SUSPEND, replica
+    // RPC apply, and recovery force-suspend all funnel through here); the gauge tracks the live cold-set
+    // size. insert_or_assign / Increment / UpdateColdGauge_ are all noexcept, so once we reach here the
+    // suspend cannot throw-and-roll-back with a stale suspended_ entry.
+    suspended_.insert_or_assign(std::string{name}, std::move(entry));
```
as51340, line 1197: I think you want to put this before durability_->Put because it can throw and you get into an inconsistent state between in-memory and disk state

as51340 (changes requested):

as51340:
Screenshot 2026-07-02 at 07 39 44

I think there is a problem with storage mode when running `show storage info on database db1` where db1 is cold

as51340:
Screenshot 2026-07-02 at 07 43 01
Multiple resumes seem to be passing normally, maybe not the best UX?

as51340:
Also, should we skip cold databases when deciding to which instance to failover on? We need num committed txns which I think is not saved as part of the cold metadata

as51340:
> Also, should we skip cold databases when deciding to which instance to failover on? We need num committed txns which I think is not saved as part of the cold metadata

I guess this is fine because of dbms_handler.ForEach?

src/dbms/dbms_handler.hpp
andrejtonev: removed

src/dbms/dbms_handler.cpp
andrejtonev: added

src/dbms/dbms_handler.cpp
andrejtonev: those keys are v1

src/dbms/dbms_handler.cpp
andrejtonev: update as defined by product. just transactional with wal and have the snapshot on exit. planned improvements as users start using.

src/dbms/dbms_handler.cpp
andrejtonev: using snapshot on exit. async hot/cold planned for v2

src/dbms/handler.hpp
andrejtonev: we were calling destructor directly, this is cleaner

src/dbms/dbms_handler.cpp
andrejtonev: data is correct, we just might restart in cold

src/dbms/handler.hpp
as51340: the comment says it was UB before? // Destroy the gatekeeper exactly once, via natural scope — NOT an explicit gk.~Gatekeeper<T>()
        // followed by the captured gk being destructed again when this lambda is destroyed (that is a
        // double-destruction: [basic.life] UB, reading a destroyed object's pimpl_). Moving into a
        // block-scoped local runs the blocking ~Gatekeeper once here; the moved-from capture then
        // destroys cleanly (null pimpl_) with the lambda.

as51340 (approved):

andrejtonev:

---
___

### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Added SUSPEND DATABASE and RESUME DATABASE for multi-tenant Enterprise deployments. Suspend tears down an idle tenant's in-memory storage to a durable COLD shell to free RAM while keeping its data on disk, and resume rebuilds it HOT with all data intact; both replicate across the cluster like CREATE/DROP DATABASE and survive restarts. SHOW DATABASES now reports each database's HOT/COLD state, and SHOW STORAGE INFO ON a suspended database returns its as-of-suspend statistics. [#4280](https://github.com/memgraph/memgraph/pull/4280)
- [x] https://github.com/memgraph/documentation/pull/1679

PR #4283, opened 2026-06-19
fix: Allow currency symbols (Sc) as non-leading characters in identifiers

colinbarry:
OpenCypher grammar states that currency character such as `$` can be used as an identifier character providing it is not used as the leading character.

If the query stripper encountered such a currency character mid-identifier, it would treat it as a parameter marker and fail to parse. This would conflict with the original query, which did parse successfully, and a `LOG_FATAL` would occur.

We now allow all currency codepoints in `kUnescapeNameAllowedParts`, so the tokenzier agrees with the ANTLR grammar.


colinbarry:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Allow currency characters, such as `$`, to be used in identifiers provided they are not the leading character. This brings Memgraph's identifier support in line with OpenCypher. [#4283](https://github.com/memgraph/memgraph/pull/4283)

DavIvek (approved):

PR #4284, opened 2026-06-19
feat: Tolerate tenant recovery failure

as51340:
With the new flag --storage-allow-recovery-failure, we allow an instance to get up even in the case of tenant failure.


src/utils/typeinfo.hpp
```diff
@@ -128,6 +128,8 @@ enum class TypeId : uint64_t {
   REP_CREATE_DATABASE_RES,
   REP_DROP_DATABASE_REQ,
   REP_DROP_DATABASE_RES,
+  REP_REPAIR_DATABASE_REQ,
```
andrejtonev: I think we only append, otherwise the enum values change and we stop being backwards compatible.

src/query/interpreter.cpp
```diff
@@ -9749,6 +9819,57 @@ Interpreter::PrepareResult Interpreter::Prepare(ParseRes parse_res, UserParamete
       auto transaction_requirements = QueryTransactionRequirements{
           parse_info.parsed_query.using_schema_assert, parsed_query.is_cypher_read, storage_mode};
       parsed_query.query->Accept(transaction_requirements);
+
+      // Fail-closed gate for broken databases (those that failed durability recovery and
+      // came up empty under --storage-allow-recovery-failure). Any query that operates on
+      // the current database's data is rejected until it is recovered via RECOVER SNAPSHOT
+      // or REPAIR DATABASE. Meta queries (USE/SHOW DATABASES, SHOW STORAGE INFO, auth,
+      // replication, ...) do not touch the tenant graph and are allowed through.
+      if (current_db_.db_acc_ && (*current_db_.db_acc_)->storage()->IsBroken()) {
```
andrejtonev, line 9943: We are in implicit transaction here, what about explicit transactions? looks like there are no guard against:
BEGIN; CREATE(); COMMIT on a broken db.

src/query/interpreter.cpp
```diff
@@ -9528,6 +9593,10 @@ struct QueryTransactionRequirements : QueryVisitor<void> {
 
   void Visit(RecoverSnapshotQuery & /*unused*/) override { accessor_type_ = storage::StorageAccessType::UNIQUE; }
 
+  // REPAIR DATABASE runs as a system transaction (it replicates the tenant reset to the replicas); it must
+  // not open a data accessor on the current DB, otherwise the system-transaction commit path is skipped.
+  void Visit(RepairDatabaseQuery & /*unused*/) override {}
```
andrejtonev: I don't like this, since other things in the background could be handling storage in parallel. Shouldn't this be UNIQUE regardless if system or not?

src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -439,6 +439,13 @@ void InMemoryReplicationHandlers::PrepareCommitHandler(
   const memory::DbArenaScope db_arena_scope{db_acc->get()};
   auto *storage = static_cast<storage::InMemoryStorage *>(db_acc->get()->storage());
 
+  // No broken check is needed here. A broken replica tenant reports commit-ts 0 with a fresh epoch (a consequence of
```
andrejtonev: I feel like there is an edge case here where the tenant is empty, but replica recovers a broken tenant for some reason, so ldt is 0 but broken.

src/storage/v2/inmemory/storage.cpp
```diff
@@ -4511,9 +4531,83 @@ std::expected<void, InMemoryStorage::RecoverSnapshotError> InMemoryStorage::Reco
...
+    std::filesystem::remove(dir / old_dir, ec);  // remove dir if empty
+  }
+
+  // Reset the tenant to an empty working state (mirrors the broken scrub in the constructor).
+  ResetTenant();
+
+  return {};
+}
+
+void InMemoryStorage::ResetTenant() {
```
andrejtonev: ResetTenant()/Clear() on the replica reset path clears memory+epoch but leaves the tenant's old WAL on disk (RepairBroken purges it, this path
  doesn't). Reproduced: replica restart after an RPC reset either SIGABRTs (GID collision across epochs) or silently recovers the stale pre-repair  epoch.
  Purge the on-disk WAL/snapshot in the replica reset path too.

src/storage/v2/inmemory/storage.cpp
```diff
@@ -475,6 +475,18 @@ InMemoryStorage::InMemoryStorage(Config config, std::optional<free_mem_fn> free_
...
             &ttl_,
             &description_store_);
+      } catch (const durability::RecoveryFailure &e) {
+        // --storage-allow-recovery-failure: instead of crashing the process, bring this
+        // database up empty and broken. RecoverData only reads durability files, so the
+        // on-disk snapshot/WAL are left untouched for the operator to RECOVER SNAPSHOT,
+        // REPAIR DATABASE, or restore the whole data directory from a backup.
+        if (!config_.durability.allow_recovery_failure) throw;
+        spdlog::warn("Database '{}' failed to recover ({}); bringing it up in the broken state.", name(), e.what());
+        Clear();
```
andrejtonev, line 501: A broken tenant that had TTL configured keeps a live storage-ttl thread — verified it created a label+property index on the 0-vertex broken tenant
   via Access(WRITE). Clear() doesn't disarm ttl_ and Access() has no IsBroken() gate. Call ttl_.Shutdown() in the broken-scrub path (ctor catch +
  ResetTenant).

src/storage/v2/inmemory/storage.cpp
```diff
@@ -4511,9 +4531,83 @@ std::expected<void, InMemoryStorage::RecoverSnapshotError> InMemoryStorage::Reco
...
+
   return {};
 }
 
+std::expected<void, InMemoryStorage::RepairError> InMemoryStorage::RepairBroken() {
+  if (!IsBroken()) {
+    return std::unexpected{InMemoryStorage::RepairError::NotBroken};
+  }
+
+  auto const use_old_dir = FLAGS_storage_backup_dir_enabled;
```
andrejtonev: Do we need to lock snapshot and gc? I think clear has a comment explaining the lock ordering that you need for full control.

src/storage/v2/storage.hpp
```diff
@@ -216,6 +220,44 @@ class Storage {
...
+  // learns it must reset that tenant. The flag is in-memory and lives only for the current run.
+  bool WasRepaired() const noexcept { return was_repaired_.load(std::memory_order_acquire); }
+
+  void SetRepaired(bool value) noexcept { was_repaired_.store(value, std::memory_order_release); }
+
+  // Records that replica `replica_name` confirmed (applied) the most recent repair of this tenant, so the
+  // main stops advertising the repair to it (see WasRepaired()). Prevents a replica that already reset and
+  // re-synced from being needlessly reset again on a later reconnect.
+  void MarkRepairConfirmedBy(std::string_view replica_name) {
+    repaired_confirmed_replicas_.WithLock([&](auto &names) { names.emplace(std::string{replica_name}); });
```
andrejtonev: do replicas have ids? is a key via name safe?

src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -716,6 +723,11 @@ void InMemoryReplicationHandlers::SnapshotHandler(rpc::FileReplicationHandler co
   }
   spdlog::debug("Snapshot from {} loaded successfully.", dst_snapshot_file);
 
+  // A successful snapshot load is the moment a broken replica tenant becomes healthy: the main has just full-synced
+  // it. Clearing the broken flag re-enables background durability and lets queries touch the tenant again (slice 1
+  // guards writes behind IsBroken(), so clearing the flag is sufficient to resume normal operation).
+  storage->SetBroken(false);
```
andrejtonev: this is set outside the main lock. is that correct?

andrejtonev (commented):
first pass

src/storage/v2/inmemory/storage.cpp
as51340: I take snapshot lock and main lock in ResetTenant before calling Clear. Main lock is needed for Clear() and I take snapshot lock to block periodic snapshot runner while resetting tenant. We take snapshot lock and then main lock in the same order in SnapshotHandler, WalFilesHandler and CurrentWalHandler

src/storage/v2/storage.hpp
as51340: We don't have ids, name is the best what we can do currently

src/dbms/inmemory/replication_handlers.cpp
as51340: broken_ is atomic so should be ok. But I will move it under main lock for consistency with ResetTenant

src/storage/v2/inmemory/storage.cpp
as51340: good catch

src/storage/v2/inmemory/storage.cpp
as51340: agree, fixed

src/dbms/inmemory/replication_handlers.cpp
as51340: I think that's not possible but it's not a problem to add a check here

src/query/interpreter.cpp
as51340: 1. I think we enforce in the code that there is either a system transaction taking place or data txn
2. I am not sure what else could be run in parallel since the db is in broken state

as51340: conclusion: this query should do drop force tenant + create tenant

as51340: waiting for product to confirm

src/utils/typeinfo.hpp
```diff
@@ -349,6 +352,7 @@ enum class TypeId : uint64_t {
   AST_TENANT_PROFILE_QUERY,
   AST_SHOW_MEMORY_INFO_QUERY,
   AST_SESSION_SETTING_QUERY,
+  AST_RESET_DATABASE_QUERY,
```
andrejtonev: Wire-compat break. AST_RESET_DATABASE_QUERY sits in the unanchored run between AST_LABELIX = 3000 and SYMBOL = 4000, and REP_TENANT_PROFILE_REQ/RES are embedded after it — so this shifts their ordinals from 3149/3150 (master) to 3150/3151. Those are live TenantProfileRpc wire ids dispatched by message_id equality, so a mixed-version main↔replica pair mis-dispatches tenant-profile replication. The REP_ pair was correctly appended before COORD_FAILOVER_REQ = 2500; please do the same for the AST entry — move it after REP_TENANT_PROFILE_RES (end of the block), or numerically pin REP_TENANT_PROFILE_REQ/RES.

src/storage/v2/inmemory/storage.cpp
```diff
@@ -4511,9 +4534,86 @@ std::expected<void, InMemoryStorage::RecoverSnapshotError> InMemoryStorage::Reco
...
+}
+
+std::expected<void, InMemoryStorage::ResetError> InMemoryStorage::ResetBroken() {
+  if (!IsBroken()) {
+    return std::unexpected{InMemoryStorage::ResetError::NotBroken};
+  }
+  return ClearDurabilityAndReset();
+}
+
+std::expected<void, InMemoryStorage::ResetError> InMemoryStorage::ClearDurabilityAndReset() {
```
andrejtonev: Data-loss race. The snapshots/+wal/ move/delete loop runs holding no lock until ResetTenant() at the end. RESET DATABASE (system txn, no data accessor), RECOVER SNAPSHOT (UNIQUE main_lock_), and the periodic snapshot writer (snapshot_lock_, gated only on !IsBroken()) are mutually unserialized. On the replica side ResetDatabaseHandler/SystemRecoveryHandler call this on non-broken tenants where the writer is live → a snapshot landing after the dir scan survives and resurrects stale data on restart. Please take snapshot_lock_ across the whole file loop and abort/await any in-flight snapshot before removing files.

src/dbms/dbms_handler.cpp
```diff
@@ -235,6 +236,45 @@ struct DropDatabase : memgraph::system::ISystemAction {
...
+
+  bool ShouldReplicateInCommunity() const override { return false; }
+
+  bool DoReplication(replication::ReplicationClient &client, const utils::UUID &main_uuid,
+                     memgraph::system::Transaction const &txn) const override {
+    auto check_response = [](const storage::replication::ResetDatabaseRes &response) {
+      return response.result != storage::replication::ResetDatabaseRes::Result::FAILURE;
+    };
+
+    auto const completed = client.StreamAndFinalizeDelta<storage::replication::ResetDatabaseRpc>(
```
andrejtonev: Stale ASYNC replica. For ASYNC replicas StreamAndFinalizeDelta returns true at enqueue time (replication_client.hpp:108-110), before the RPC is sent, and we call MarkResetConfirmedBy on that. If the enqueued reset later fails the false confirmation sticks, SystemRestore never re-advertises (!ResetConfirmedBy), and since the reset main is at ts 0 / fresh epoch while the replica is ahead at ts>0, normal recovery won't wipe it → stale data served indefinitely. Mark confirmed only inside the background task after check(SendAndWait()) succeeds, like the SYNC path.

src/storage/v2/inmemory/storage.cpp
```diff
@@ -4511,9 +4534,86 @@ std::expected<void, InMemoryStorage::RecoverSnapshotError> InMemoryStorage::Reco
...
+    std::filesystem::remove(dir / old_dir, ec);  // remove dir if empty
+  }
+
+  // Reset the tenant to an empty working state (mirrors the broken scrub in the constructor).
+  ResetTenant();
+
   return {};
 }
 
+void InMemoryStorage::ResetTenant() {
```
andrejtonev: Deadlock. ResetTenant() takes main_lock_ exclusively via a no-timeout lock_guard, but unlike PrepareCommit/Snapshot/WalFiles/CurrentWal handlers, the reset paths don't call AbortPrevTxnIfNeeded first. An orphaned 2PC accessor in the process-wide two_pc_cache_ (from a main that died mid-PrepareCommit before failover) holds main_lock_ shared → the reset RPC thread hangs forever, the replica connection leaks, and all SYNC/STRICT_SYNC commits to that replica stall. Call AbortPrevTxnIfNeeded()/DestroyReplAccessor() before ClearDurabilityAndReset() in both reset handlers.

src/dbms/replication_handlers.cpp
```diff
@@ -252,6 +311,28 @@ bool SystemRecoveryHandler(DbmsHandler &dbms_handler, const std::vector<storage:
     }
   }
 
+  // Reset any tenant the main has reset but whose ResetDatabaseRpc this replica missed. A tenant
+  // already at commit ts 0 was reset (e.g. by the ResetDatabaseRpc) and is skipped to avoid a needless
+  // re-sync; one still holding stale data (ts != 0) is reset so the main can re-sync it from scratch.
+  for (const auto &uuid : reset_uuids) {
```
andrejtonev: False-confirm → stale replica. If ClearDurabilityAndReset() fails for a reset_uuids entry the loop only logs and continues, then the function returns true (SUCCESS). Main then MarkResetConfirmedBy for every advertised uuid incl. the failed one → never re-advertised → replica keeps stale files. Same outcome for a licensed main + unlicensed replica (no-license branch returns success without running the loop). Track a failure flag and return false if any reset failed, mirroring the Update/Delete-failure pattern already in this function; and don't let an unlicensed replica report success.

src/storage/v2/durability/durability.cpp
```diff
@@ -541,12 +542,20 @@ void RecoverDerivedState(utils::SkipListDb<Vertex> *vertices, [[maybe_unused]] u
...
+      }
+    } else if (edges_metadata->size() != edges->size()) {
+      // Every recovered edge must have an edge - metadata entry;
+      //  fail at recovery rather
+      //  than later in GC or via lookups.
+      if (config.durability.allow_recovery_failure) {
+        throw RecoveryFailure("Edge metadata count does not match edge count after recovery!");
+      }
+      LOG_FATAL("Edge metadata count does not match edge count after recovery!");
     }
```
andrejtonev, line 553: Default-config process crash + release regression. These edge-consistency checks throw RecoveryFailure only when --storage-allow-recovery-failure is on, else LOG_FATAL. But the helper is  shared by RECOVER SNAPSHOT (storage.cpp:4429) and the replica SnapshotHandler (inmemory/replication_handlers.cpp:710), whose catch(RecoveryFailure) is dead code when the flag is off (the default) — so a corrupt/edge-inconsistent snapshot crashes the whole process for users not using the feature. These were DMG_ASSERT (compiled out under NDEBUG) on master, so it's also a release regression. Throw RecoveryFailure unconditionally (all callers already catch it), or keep DMG_ASSERT for the flag-off path.

andrejtonev (commented):
first batch

src/storage/v2/inmemory/storage.cpp
andrejtonev, line 1280: STRICT_SYNC livelock. A broken STRICT_SYNC replica at ldt 0 while its main is also ldt 0 is classed READY (0>=0 tie; branching-point skipped for kTimestampInitialId). The first write is rejected by the new broken-guard → 2PC prepare fails → main AbortAndResetCommitTs → main ldt never advances → tie never breaks → every write to that tenant is permanently aborted, no self-heal. Reachability is narrow (crash-during-init / restored corrupt backup / bit-rot on a fresh empty tenant) but there's no recovery. Needs the fresh-broken STRICT_SYNC tie to force RECOVERY.

src/storage/v2/inmemory/storage.cpp
andrejtonev: TTL silently lost after the headline cure query. ResetTenant() clears broken_ but never calls ttl_.Enable()/Configure()/Resume(), so a tenant cured via RESET DATABASE loses TTL until an operator manually re-ENABLEs it — inconsistent with RECOVER SNAPSHOT, which restores TTL from snapshot metadata. Decide intended behavior and restore/clear TTL config explicitly.

src/query/interpreter.cpp
```diff
@@ -6305,6 +6317,58 @@ PreparedQuery PrepareRecoverSnapshotQuery(ParsedQuery parsed_query, bool in_expl
...
+            !result.has_value()) {
+          throw QueryException("{}", result.error());
+        }
+        notifications->emplace_back(
+            SeverityLevel::INFO,
+            NotificationCode::RESET_DATABASE,
+            fmt::format("Database '{}' reset.", db_name),
+            "The database has been reset to an empty state. You can now run your import queries to repopulate it.");
+        return QueryHandlerResult::COMMIT;
+      },
```
andrejtonev: Silent main/replica divergence on exception. The irreversible local reset (ResetBroken()) and AddAction run before the outer Commit(). A BasicException (e.g. OOM — including the handler's own notifications.emplace_back) routes to Interpreter::Abort(), which discards the queued ResetDatabaseAction without replicating. Local reset is done, IsBroken() is now false so manual retry is blocked, and self-heal only fires on an incidental reconnect (was_reset_/SystemRestore is event-triggered, not periodic). Consider ordering replication before the irreversible local reset, or making the local reset idempotent/retryable.

src/storage/v2/replication/replication_client.cpp
andrejtonev, line 731: Failover leaves a stale replica READY. was_reset_/reset_confirmed_replicas_ are in-memory per-run. If the main restarts or HA promotes a new main between RESET and a lagging replica's reconnect, reset_uuids is never advertised and the empty-steps do_reset gap fires → replica marked READY on stale pre-reset data. Persist the reset intent, or drive the reset off epoch identity rather than the in-memory flag. (Also trace the unconfirmed secondary risk: post-reset writes to a READY-but-stale replica may skip branching-point reconciliation on the next commit.)

src/query/interpreter.cpp
andrejtonev, line 6943: DatabaseInfoQuery (SHOW INDEX/CONSTRAINT/NODE LABELS/EDGE TYPES INFO) is allowlisted on a broken tenant but reads the empty storage with no IsBroken() check and returns a clean 0-row result — misleading, unlike SystemInfoQuery/SHOW DATABASES which surface a status. Gate these on IsBroken() or emit broken status.

src/replication_handler/include/replication_handler/replication_handler.hpp
```diff
@@ -124,10 +140,25 @@ void SystemRestore(ReplicationClient &client, system::System &system, dbms::Dbms
                                                           params_snapshot);
 #endif
```
andrejtonev, line 141: A v3 SystemRecoveryReq sent to a not-yet-upgraded v2 replica throws (requests have no downgrade) and the replica stalls BEHIND until upgraded. Inherent to the RPC scheme, but this feature targets rolling upgrades — document "upgrade replicas before the main."

src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -867,6 +890,13 @@ void InMemoryReplicationHandlers::WalFilesHandler(
   }
```
andrejtonev, line 946: A partial WAL load applying a WalTtlOperation delta configures/enables TTL even when a later file fails and SetBroken(false) is never reached → live (paused) TTL thread on a still-broken tenant. Guard TTL (re)configuration on !IsBroken().

src/storage/v2/inmemory/storage.cpp
```diff
@@ -475,6 +475,21 @@ InMemoryStorage::InMemoryStorage(Config config, std::optional<free_mem_fn> free_
...
+        spdlog::warn("Database '{}' failed to recover ({}); bringing it up in the broken state.", name(), e.what());
+        Clear();
+        name_id_mapper_->Clear();
+        description_store_.Clear();
+        // Snapshot recovery may have already started the storage-ttl background thread; stop it so a
+        // broken database doesn't keep firing TTL jobs against cleared storage.
+        ttl_.Disable();
+        SetBroken(true);
+        return std::nullopt;
       } catch (...) {
```
andrejtonev: The flag-off catch(RecoveryFailure) throw; bypasses the sibling catch(...)'s ClearLightEdges(), leaking pool-allocated edges. Mostly moot (startup abort), but call ClearLightEdges() before rethrow.

src/storage/v2/inmemory/storage.cpp
```diff
@@ -4511,9 +4534,86 @@ std::expected<void, InMemoryStorage::RecoverSnapshotError> InMemoryStorage::Reco
...
+}
+
+std::expected<void, InMemoryStorage::ResetError> InMemoryStorage::ClearDurabilityAndReset() {
+  auto const use_old_dir = FLAGS_storage_backup_dir_enabled;
+  constexpr std::string_view old_dir = ".old";
+
+  // Move the corrupt snapshots and WAL files aside (to .old) when backup directories
+  // are enabled, otherwise delete them. This leaves the durability directory clean so
+  // the tenant recovers as an empty database on the next restart rather than re-breaking.
+  if (use_old_dir) {
```
andrejtonev: The prep loop can remove_all the first dir's .old then fail on the second, returning BackupFailure ("safe to retry") after already destroying the first backup.

tests/unit/rpc_versioning.cpp
```diff
@@ -233,7 +233,8 @@ TEST(RpcVersioning, RequestTwoVersionsSingleVersionResponse_ThrowsWhenSendingV1)
   EXPECT_THROW(memgraph::rpc::SaveWithDowngrade(res, 1, &builder), std::runtime_error);
 }
 
-// SystemRecoveryRpc: with ResV1 and Downgrade, both v1 and v2 requests get correct response.
+// SystemRecoveryRpc: requests with and without the optional trailing fields (parameters, reset
+// tenant uuids) all load and get a correct response.
 TEST(RpcVersioning, SystemRecoveryRpc_V1AndV2Request_BothSucceed) {
```
andrejtonev: All sub-cases stream the v3 type, so SystemRecoveryReqV2/ResV2 and the reset_uuids Upgrade/Downgrade paths get zero cross-version decode coverage despite the test's name. Add genuine ...RpcV1/...RpcV2 aliases (mirror GetDatabaseHistoriesRpcV1).

andrejtonev (commented):
final wave

src/storage/v2/durability/durability.cpp
as51340: I think this is correct as it is currently

src/replication_handler/include/replication_handler/replication_handler.hpp
as51340: that's how ISSU works, update your AI agent

src/storage/v2/inmemory/storage.cpp
as51340: changed now, I catch ...

src/storage/v2/inmemory/storage.cpp
as51340: same logic is in RecoverSnapshot and probably at other places as well. I don't plan changing this here.

src/dbms/inmemory/replication_handlers.cpp
as51340: I think this is fine because as soon as WAL request is received, SetBroken(false) will be executed

src/storage/v2/inmemory/storage.cpp
as51340: can you explain this better, when and how can this occur?

andrejtonev (approved):

as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Recovery failures in a single database no longer have to take down the whole instance: with the new --storage-allow-recovery-failure flag (disabled by default), a database that fails to recover its durability files at startup now comes up in a broken (defunct) state rather than crashing the process, keeping every healthy database available. A broken database rejects queries until it is repaired, its health is surfaced through SHOW DATABASES and SHOW STORAGE INFO, and it can be cured in place by loading a good snapshot with RECOVER SNAPSHOT or simply removed with DROP DATABASE; in a high-availability cluster a defunct tenant on a replica additionally self-heals by re-syncing its snapshot from the main. To take advantage of this, enable --storage-allow-recovery-failure on startup; leaving it off preserves the previous fail-fast behavior, so no action is required for existing deployments. #4284 (https://github.com/memgraph/memgraph/pull/4284)

PR #4292, opened 2026-06-23
Enhanced projection

Ignition:
## Cypher over subgraphs and projections

Run declarative Cypher over a derived view of the graph instead of the whole store, and write computed values back. Two kinds of view:

- **Subgraph** with `project(path)`: a membership-filtered set of real nodes and edges, so writes persist.
- **Projection** with `derive(path, config)` or `virtualGraph(nodes, edges)`: virtual nodes and edges, where an overlay node maps back to a real origin, reads its properties lazily, and can shadow them with computed values.

A view is bound as the ambient graph for a block with `CALL { USE g ... }`; match, expand, built-in functions, and the variable-length and shortest-path family all run over it.

### New capabilities

- `CALL { USE g ... }` runs a subquery against a bound view `g`.
- `derive(path, {...})` overlays computed nodes over a matched path, with per-property control via `propertyPolicy` (`origin` / `overlay` / `hidden`), plus `sourceNodeProperties` / `targetNodeProperties` and `undirectedEdgeTypes`.
- `virtualGraph(nodes, edges, config?)` assembles a projection from lists built with `virtualNode(...)` and `virtualEdge(...)`, importing an external graph in one statement.
- Variable-length and shortest-path expansion over a view.
- `id()` and `virtual_id()` return stable, query-local ids for virtual elements; per-property fine-grained access control is enforced on overlay reads.

### Examples

Each of these does something plain Cypher cannot express on the stored graph.

**Mixed edge direction in a multi-hop traversal.** A single pattern cannot make one relationship type undirected while another stays directed. A projection can: here `CITES` stays directed but `SIMILAR_TO` counts as mutual, across a three-hop walk for recommendations. The papers carry large embedding vectors, and the projection reads a property only when it is used (here for the rows returned), never while walking:

```cypher
MATCH p = (:Paper)-[r:CITES|SIMILAR_TO]->(:Paper)
WITH derive(p, {virtualEdgeType: type(r), undirectedEdgeTypes: ['SIMILAR_TO']}) AS g
CALL { USE g
       MATCH (:Paper {id: 1})-[:CITES|SIMILAR_TO *1..3]->(rec)
       WHERE rec.id <> 1
       RETURN DISTINCT rec.id AS id, rec.embedding AS embedding }
RETURN id, embedding
```

Return the node `rec` instead and add `SET rec.recommended = true` after the block to persist the result on the real papers.

**Shortest path with per-type direction.** Shortest hop count where `FRIEND` is mutual but `MENTOR` is one-way, which no single shortest-path pattern expresses:

```cypher
MATCH p = (:Person)-[r:FRIEND|MENTOR]->(:Person)
WITH derive(p, {virtualEdgeType: type(r), undirectedEdgeTypes: ['FRIEND']}) AS g
CALL { USE g
       MATCH (:Person {name: 'B'})-[e:FRIEND|MENTOR *BFS]->(:Person {name: 'A'})
       RETURN size(e) AS hops }
RETURN hops
```

**Query a graph that is not stored.** Assemble nodes and edges on the fly (for example rows exported from another system) and run a query over them, without first creating them in the database:

```cypher
WITH virtualGraph(
       [virtualNode(1, 'Person', {name: 'CEO'}), virtualNode(2, 'Person', {name: 'VP'}), virtualNode(3, 'Person', {name: 'IC'})],
       [virtualEdge('MANAGES', 1, 2), virtualEdge('MANAGES', 2, 3)]) AS org
CALL { USE org MATCH (:Person {name: 'CEO'})-[:MANAGES*]->(report) RETURN report.name AS name }
RETURN name
```

**Remove a field from what a query can see.** Give a subquery a view of the org graph with `salary` removed, so even a generic read like `properties(n)` cannot surface it:

```cypher
MATCH p = (:Employee)-[:REPORTS_TO]->(:Employee)
WITH derive(p, {virtualEdgeType: 'REPORTS_TO', propertyPolicy: {salary: 'hidden'}}) AS org
CALL { USE org MATCH (e) RETURN properties(e) AS visible }
RETURN visible
```

A `project()` subgraph is different in kind: it behaves like the real graph restricted to a chosen slice, so most single queries over it can also be written with plain filters. Its distinct value is handing that slice to a graph algorithm, which then sees only the subgraph.

### Current limitations (version 1)

These apply inside a `CALL { USE g ... }` block and are reported as clear errors, not silent wrong results. A `project()` subgraph is made of the real graph's own nodes and edges, so it behaves like a normal graph; the restrictions are mainly about `derive()` projections, whose elements are virtual.

- **The block is read-only.** Inside `USE` you can `MATCH`, traverse, and `RETURN`, but not `CREATE` / `SET` / `REMOVE` / `DELETE`. Apply writes after the block, on what it returns (as in the recommendation example).
- **No nested `USE`.** One view is bound at a time; a `USE` block cannot contain another.
- **No index scans inside the block.** Matches scan the whole view rather than using label or property indexes, so keep the view small (a projection usually is).
- **Over a `derive()` projection: no named paths and no variable-length lambdas.** You cannot capture a named path (`MATCH path = (...)`) or use the inline filter/weight form `[:REL *1..3 (e, n | ...)]` while traversing a projection, because both need real graph elements; return endpoints, node and edge values, or a relationship-list length instead (as the shortest-path example does). Both work over a `project()` subgraph.

### DIFF
--- a/ADRs/008_cypher_execution_over_projections.md
+++ b/ADRs/008_cypher_execution_over_projections.md
@@ -1,0 +1,141 @@
+# Cypher Execution over Subgraphs and Projections ADR
+
+**Author**
+Gareth Lloyd (https://github.com/Ignition)
+
+**Status**
+ACCEPTED (target model; implementation staged, see Consequences)
+
+**Date**
+July 15, 2026
+
+**Problem**
+
+Memgraph can hand a derived view of the graph to an algorithm procedure, but
+users want more: to run declarative Cypher over a view, and to write computed
+values back onto its nodes. Two kinds of derived view exist. A **subgraph**
+(`project(path)`) is a membership-filtered set of real vertices and edges; its
+elements carry real accessors, so a write persists. A **projection**
+(`derive(path, config)`, or `virtualGraph(nodes, edges)` assembled from lists) is
+built from nodes and edges that are not real vertices.
+
+Three forces shape the design. First, the original projection copied every
+property of every node into compute, which is untenable for embedding-heavy
+nodes. Second, a projection's nodes must still map back to a real node where one
+exists, so a written value lands on the right vertex. Third, adding derived-view
+nodes as a second element kind forces every consumer (built-in functions, the
+`SET` operator, Bolt serialization, every scan and expand operator) to branch on
+kind, and each new capability multiplies the branching.
+
+**Criteria**
+
+- **Real-graph path is not regressed** (highest weight). The overwhelming
+  majority of queries touch no derived view; they must pay nothing.
+- **One code path, not per-kind forks.** Adding a derived view must not fork
+  every operator and function.
+- **No eager data copy.** A projection over real nodes must not materialise
+  properties it never reads.
+- **Declarative expressiveness.** The whole query engine, not a fixed set of
+  procedures, should run over a view.
+
+**Decision**
+
+Seven decisions, each addressing part of the problem above.
+
+## 1. One projected-node type
+
+Every node in a derived view is one in-memory type (`query::VirtualNode`) holding
+an **optional origin** real vertex and its own **overlay** property store. An
+*overlay node* has an origin: reads fall through to it lazily (an origin property
+is never copied unless read), and declared overlay keys shadow it. A *synthetic
+node* has no origin: reads and writes hit its overlay only. These are two
+configurations of one type, not two types, which collapses the per-kind branching
+to a single path. This directly serves "no eager data copy" and "one code path".
+
+## 2. Per-property binding with a static schema
+
+Each property is bound to one store: `origin` (read-through and write-through,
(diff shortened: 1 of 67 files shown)


Ignition:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist

.scratch/projections/demo/projection_demo.cypher
```diff
@@ -0,0 +1,165 @@
+// ============================================================================
```
gitbuda: This is very nice, but do we move that under the `specs`? I would keep this as examples -> it would probably be useful as the test cases 🤔

Josipmrden:
I see the negative index with virtual nodes/edges is auto-incrementing. Do we rotate (it can overflow quickly if we automate analysis) CC @Ignition , not sure how that is handled in the code

PR #4299, opened 2026-06-25
feat(plan): lower id(n) IN <list> to an indexed scan

### LINKED ISSUE #2234: Improve planner in the WHERE id(n) IN $idArray case
"We often run queries like `match (n) where id(n) in $idArray;`. Such queries currently trigger a full scan of all nodes (ScanAll) instead of taking advantage of ScanAllById. I tried using `(id(n) = 1 or id(n) = 2 or ...)`, but it does not seem to work either. Do you have a suggestion for a workaround?"

Workaround:
```
UNWIND $idArray as id
MATCH (n) WHERE ID(n) = id
RETURN n;
```

Most likely the problem is that we don't deduce that $idArray is actually array (NOTE: there is the planner rule to leverage `ScanAllById` in case of an array).

Ignition:
MATCH (n) WHERE id(n) IN <list> now lowers to a per-element ScanAllById (Unwind feeding the id scan) instead of a full ScanAll plus membership Filter. It accepts parameter lists and any list-valued expression, closing the common $ids case that previously fell through to a full scan.

The existing n.prop IN <list> label+property unwind, which only fired for literal lists, is extended the same way: a parameter or other expression now drives the indexed scan too, and a value repeated in the list emits its node once rather than duplicating it.

Both paths wrap the driving list as toSet(coalesce(value, [])) so a null list returns zero rows instead of throwing and duplicates collapse; a statically-known non-list literal and a self-referential list stay on the Filter path. The rewrites stay result-identical to the membership Filter across the full input contract, proven by an end-to-end equivalence suite on in-memory and on-disk storage.

### DIFF
--- a/src/query/plan/preprocess.cpp
+++ b/src/query/plan/preprocess.cpp
@@ -273,8 +273,9 @@ PropertyFilter::PropertyFilter(Symbol symbol, PropertyIx property, Type type)
 PropertyFilter::PropertyFilter(Symbol symbol, PropertyIxPath properties, Type type)
     : symbol_(std::move(symbol)), property_ids_(std::move(properties)), type_(type) {}
 
-IdFilter::IdFilter(const SymbolTable &symbol_table, const Symbol &symbol, Expression *value, bool expects_string_id)
-    : symbol_(symbol), value_(value), expects_string_id_(expects_string_id) {
+IdFilter::IdFilter(const SymbolTable &symbol_table, const Symbol &symbol, Expression *value, bool expects_string_id,
+                   Type type)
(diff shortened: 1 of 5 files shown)


Ignition:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - `MATCH ... WHERE id(n) IN <list>` and `n.prop IN <list>` now lower to a per-element indexed scan (id index, or label+property index) driven by a parameter or any list-valued expression, not just a literal list, so these lookups no longer fall back to a full scan. No query change is needed to benefit. [#4299](https://github.com/memgraph/memgraph/pull/4299)

DavIvek (approved):

PR #4301, opened 2026-06-25
fix(ha/auth): Migrate users during rolling upgrade between Memgraph instances of different versions

colinbarry:
User/role data sent over the wire was not being migrated.

During a rolling upgrade from `3.9` to `3.10.1`, LBAC grants are lost when a `3.9` MAIN replicates auth data to an upgraded `3.10.1` REPLICA. The old MAIN serializes user/role structs in `V3` format, but the REPLICA's `Deserialize` expects `V4`; the result is that FGA fields are silently dropped. When failover occurs and REPLICA becomes MAIN, it does so with sheared user configurations.

- Per-entity migration on deserialize: Extracted existing `V2` -> `V3` and `V3` -> `V4` migration logic into `MigrateAuthJson()`. Each user or role now carries a `"version"` field.
- Separated auth store-level from entity-level versioning: Auth store versions (`V1` -> `V2` -> `V5`) are now reserved for structural KV store changes such as key prefix migrations. Entity schema changes (such as fine-grained access) are handled per-entity on deserialize.
- ISSU test coverage: Added `verify_fga_post_upgrade.sh` to confirm FGA grants survive mid-upgrade (on REPLICA while old MAIN is still running) and post-upgrade. ISSU test need further enhancement, but that will come in successive PRs.

### DIFF
--- a/src/auth/auth.cpp
+++ b/src/auth/auth.cpp
@@ -215,6 +215,7 @@ constexpr auto kStoreV2 = "V2";  // Single role links, to JSON arrays for multi-
 constexpr auto kStoreV3 = "V3";  // V2 FGA -> V3 FGA (fine_grained_access_handler to fine_grained_permissions)
 constexpr auto kStoreV4 = "V4";  // V3 FGA -> V4 FGA (global_permission → global_grants/global_denies)
 constexpr auto kStoreV5 = "V5";  // Eagerly migrates entity JSON via MigrateAuthJson over RPC (V3/V4 -> current)
+constexpr auto kCurrentStoreVersion = kStoreV5;
 }  // namespace
 
 /**
@@ -398,7 +399,7 @@ Auth::Auth(std::string storage_directory, Config config
     MigrateVersions(storage_);
   } else {
     // Clean storage; put the version
-    storage_.Put(kStoreVersionKey, kStoreV5);
+    storage_.Put(kStoreVersionKey, kCurrentStoreVersion);
   }
 
 #ifdef MG_ENTERPRISE
--- a/src/auth/auth.cpp
+++ b/src/auth/auth.cpp
@@ -328,11 +328,15 @@ void MigrateVersions(kvstore::KVStore &store) {
     auto const migrate_entities = [&](auto const &prefix) {
       for (auto it = store.begin(prefix); it != store.end(prefix); ++it) {
         auto const &[key, value] = *it;
-        auto data = nlohmann::json::parse(value);
-        auto const original = data;
-        auth::MigrateAuthJson(data);
-        if (data != original) {
-          puts.emplace(key, data.dump());
+        try {
+          auto data = nlohmann::json::parse(value);
+          auto const original = data;
+          auth::MigrateAuthJson(data);
+          if (data != original) {
+            puts.emplace(key, data.dump());
+          }
+        } catch (nlohmann::json::exception const &e) {
+          throw auth::AuthException("Failed to migrate auth data for '{}': {}", key, e.what());
         }
       }
     };
--- a/src/auth/auth.cpp
+++ b/src/auth/auth.cpp
@@ -206,11 +206,15 @@ const std::string kMtLinkPrefix = "mtlink:";
 // Profile linking is now handled by UserProfiles class
 const std::string kVersion = "version";
 
-constexpr auto kVersionV1 = "V1";
-constexpr auto kVersionV2 = "V2";
-constexpr auto kVersionV3 = "V3";
-constexpr auto kVersionV4 = "V4";
-constexpr auto kVersionV5 = "V5";
+// Auth KV store versions. These track structural changes to the store layout
+// (key prefixes, link format, etc.). Entity-level schema changes (e.g. FGA
+// format) are handled per-entity via MigrateAuthJson and the entity "version"
+// field: do NOT bump the store version for those (anymore).
+constexpr auto kVersionV1 = "V1";  // Added password hash algorithm field
+constexpr auto kVersionV2 = "V2";  // Single role links, to JSON arrays for multi-role support
+constexpr auto kVersionV3 = "V3";  // V2 FGA -> V3 FGA (fine_grained_access_handler to fine_grained_permissions)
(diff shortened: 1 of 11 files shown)


colinbarry:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Fixed LBAC permissions being silently dropped if failover occurred during rolling migration. [#4301](https://github.com/memgraph/memgraph/pull/4301)

src/auth/models.cpp
```diff
@@ -1824,4 +1827,164 @@ PropertyAccessPermissions Roles::GetPropertyEdgeTypePermissions(std::optional<st
...
+  if (!version.has_value() || version == kCurrentAuthVersion) {
+    if (!data.contains(kVersion)) data[kVersion] = kCurrentAuthVersion;
+    return;
+  }
+
+  // V2 to V3: rename fine_grained_access_handler to fine_grained_permissions,
+  //           convert global_permission values from V2 enum to V3 bitmask,
+  //           add empty permissions array per sub-object.
+  if (version == 2) {
+    auto fg_it = data.find("fine_grained_access_handler");
```
as51340: extract fine_grained_access_handler to a constant. Move whole block if (version == 2) to a separate function, hard to read it currently

src/auth/models.hpp
```diff
@@ -1119,4 +1119,9 @@ FineGrainedAccessPermissions Merge(const FineGrainedAccessPermissions &first,
                                    const FineGrainedAccessPermissions &second);
 #endif
 
+constexpr int kCurrentAuthVersion = 4;
```
as51340: why is here kCurrrentAuthVersion set to 4 while we added kVersion =5 to auth.cpp?

as51340:
Discussed on the call, the proper fix is RPC versioning

tests/issu/test_upgrade.sh
```diff
@@ -862,6 +862,13 @@ kubectl delete pod memgraph-data-1-0
 wait_memgraph_pods_ready 90s
 echo "Upgrade of pod memgraph-data-1-0 passed successfully"
 
+echo "Waiting for old MAIN to sync auth to upgraded replica..."
+sleep 30
```
mattkjames7: Is there a more reliable way to check this, rather then using an arbitrary sleep?

colinbarry: Yes - this a quick n dirty "I just want to see a failure" hack. I'll add something more robust short term, and then at some point soon do a complete pass on enhancing the ISSU checks.

src/auth/auth.cpp
```diff
@@ -553,7 +398,7 @@ Auth::Auth(std::string storage_directory, Config config
     MigrateVersions(storage_);
   } else {
     // Clean storage; put the version
-    storage_.Put(kVersion, kVersionV4);
+    storage_.Put(kStoreVersionKey, kStoreV5);
```
as51340: add a pointer to the current version, otherwise this is to mistake. Something like
```
constexpr auto kCurrentVersion = kStoreV5;
```

src/auth/models.cpp
```diff
@@ -1824,4 +1827,192 @@ PropertyAccessPermissions Roles::GetPropertyEdgeTypePermissions(std::optional<st
...
+          new_rule["granted"] = new_granted;
+          new_rule["denied"] = 0;
+        }
+        new_permissions.push_back(std::move(new_rule));
+      }
+      perm_data["permissions"] = std::move(new_permissions);
+    }
+  }
+
+  data[kVersion] = kCurrentEntityVersion;
```
as51340, line 2008: I assume there is a difference between kCurrentEntityVersion and current version which is 5?

tests/unit/auth.cpp
```diff
@@ -3940,4 +3940,291 @@ TEST_F(AuthWithStorage, GetUsersForProfile) {
...
+constexpr uint64_t kRead = 1;
+constexpr uint64_t kSetProperty = 2;
+constexpr uint64_t kReadOrSetProperty = kRead | kSetProperty;  // 3
+constexpr uint64_t kAllLabel = 507;
+constexpr uint64_t kAllEdgeType = 27;
+constexpr uint64_t kSetLabel = 32;
+constexpr uint64_t kRemoveLabel = 64;
+constexpr uint64_t kDeleteEdge = 128;
+constexpr uint64_t kCreateEdge = 256;
+constexpr uint64_t kLabelUpdateExpanded = kRead | kSetProperty | kSetLabel | kRemoveLabel | kDeleteEdge | kCreateEdge;
```
as51340: why not just move these in models.hpp outside the enterprise block?

src/auth/models.cpp
colinbarry: Done. Added `kFineGrainedAccessHandler` constant. Extracted `MigrateV2ToV3()` and `MigrateV3ToV4()` into the anonymous namespace.

src/auth/models.hpp
colinbarry: These are two separate version tracks. `kCurrentEntityVersion` is the per-entity JSON schema version. Store version V5 tracks KV store layout changes. Added comments to both clarifying the distinction. Essentially, changing the JSON schema of users doesn't require a store bump, and changing the KVStore schema doesn't require a user bump.

src/auth/auth.cpp
colinbarry: Done

src/auth/models.cpp
colinbarry: See comments about auth kv store/vs auth JSON entity versioning.

tests/unit/auth.cpp
colinbarry: Moved the entire `FineGrainedPermission` enum, operators, and `kAllLabelPermissions`/`kAllEdgeTypePermissions` constants outside `#ifdef MG_ENTERPRISE`. Also moved the JSON key constants used by migration code outside the guard in `models.cpp`, as these are part of the serialization format, not the enterprise auth data. Tests now use the enum directly.

as51340 (approved):

PR #4304, opened 2026-06-25
feat(text_search): fuzzy phrase search 

DavIvek:
## What

Adds two text-search procedures for **fuzzy phrase** search:

- `text_search.fuzzy_phrase_search(index, 'data.<property>:<terms>', config)`
- `text_search.fuzzy_phrase_search_edges(index, 'data.<property>:<terms>', config)`

They match an **ordered, adjacent** sequence of terms on a single property where the **last term is a prefix**, tolerating up to `fuzzy_distance` typos **across the whole input** (one shared budget, not per word) — e.g. `data.title:big bad wo` matches `big bad wolf` / `big bad world`; `big bd wo` matches at `fuzzy_distance:1` (1 edit total), but two separate typos need `fuzzy_distance:2`; order/adjacency are enforced (`bad big wo` does **not** match `big bad wolf`).

## How

Implemented **natively in mgcxx (Rust)** behind a new `text_search_mode::FUZZY_PHRASE` threaded through the existing search path (`mgp.hpp` → `mg_procedure` → storage `TextIndex::Search` → mgcxx). New `fuzzy_phrase_search_gids_pinned` / `fuzzy_phrase_search_edge_gids_pinned`:

- parse `data.<property>:<terms>` and tokenize the terms with the **field's own analyzer** (`tokenizer_for_field`) — so tokenization matches the index exactly (Unicode segmentation, `RemoveLongFilter`, lowercasing);
- build the candidate query **programmatically** — a `BooleanQuery` of per-term `FuzzyTermQuery` on `Term::from_field_json_path` (leading terms whole-word fuzzy, last term a fuzzy prefix). Building it directly (rather than a `+data.<prop>:t` query string) lets **numeric/boolean tokens match as string terms** (e.g. `report 2024`) instead of failing tantivy's JSON fast-value conversion, and needs no escaping for unusual property names;
- **adaptively page** the candidate pool in score order — the first page is sized to `limit` (with a small over-fetch factor), doubling only when the post-filter is starved, capped at `FUZZY_PHRASE_MAX_CANDIDATES` — reading each hit's stored `data[property]`, tokenizing it with the same analyzer, and keeping docs where the terms appear **adjacent and in order** with the **sum** of their per-word edit distances (optimal-string-alignment; leading whole-word, last a prefix) within `fuzzy_distance` — one budget shared across the phrase, not per term. The summed budget is bounded by the per-term candidate net, so the candidate is a genuine superset (no recall loss).

## Behavior / config

- **Single property only** (`data.<property>:<terms>`); any other query form is rejected.
- `fuzzy_prefix:false` is rejected (the last term is always a prefix); `fuzzy_distance` is validated to `0–2`.

### DIFF
--- a/mgcxx/text_search/src/lib.rs
+++ b/mgcxx/text_search/src/lib.rs
@@ -480,11 +480,11 @@ mod tests {
         let _ = std::fs::remove_dir_all(&dir);
     }
 
-    // Adaptive candidate paging: 300 real matches exceed SEQUENCE_BASE_PAGE (256), so returning all of
-    // them forces a second page regardless of score order -- exercising the widening + and_offset
-    // boundary. Returning exactly the 300 (no more, no fewer) proves no boundary drop or duplicate; the
-    // interleaved decoys carry every query term in the wrong order, so they must be rejected across
-    // pages by the order/adjacency post-filter.
+    // Adaptive candidate paging: the 10 real matches sit between 150 decoys on each side, so the first
+    // page (limit 10 * SEQUENCE_OVERFETCH_FACTOR = 80 candidates) holds only decoys whichever way score
+    // ties break -- reaching the matches requires widening past the first page via and_offset. The
+    // decoys carry every query term in the wrong order, so the post-filter must reject them across
+    // pages; returning exactly the 10 proves no boundary drop or duplicate.
     #[test]
     fn search_sequence_pages_past_the_first_candidate_page() {
         let dir = std::env::temp_dir().join(format!("mgcxx_seq_paging_{}", std::process::id()));
@@ -497,14 +497,19 @@ mod tests {
             .to_string();
         let mut ctx = create_index(&path, &ffi::IndexConfig { mappings }).unwrap();
 
-        let real: Vec<u64> = (1u64..=300).collect();
-        for &gid in &real {
-            let data = format!(r#"{{"data":{{"name":"big bad wolf"}},"all":"big bad wolf","gid":{gid}}}"#);
+        let mut add_doc = |gid: u64, name: &str| {
+            let data = format!(r#"{{"data":{{"name":"{name}"}},"all":"{name}","gid":{gid}}}"#);
             add_document(&mut ctx, &ffi::DocumentInput { data }, true).unwrap();
+        };
+        for gid in 1u64..=150 {
+            add_doc(gid, "wolf bad big");
         }
-        for gid in 1000u64..=1049 {
-            let data = format!(r#"{{"data":{{"name":"wolf bad big"}},"all":"wolf bad big","gid":{gid}}}"#);
-            add_document(&mut ctx, &ffi::DocumentInput { data }, true).unwrap();
+        let real: Vec<u64> = (1000..1010).collect();
+        for &gid in &real {
+            add_doc(gid, "big bad wolf");
+        }
+        for gid in 2000u64..=2150 {
+            add_doc(gid, "wolf bad big");
         }
         commit(&mut ctx).unwrap();
 
@@ -514,7 +519,7 @@ mod tests {
             search_query: "data.name:big bad wo".to_string(),
             return_fields: vec![],
             aggregation_query: String::new(),
-            limit: 0, // unlimited -> first page is SEQUENCE_BASE_PAGE, must widen to reach all matches
+            limit: 10,
             fuzzy_distance: 0,
             fuzzy_prefix: true,
             fuzzy_transpositions: false,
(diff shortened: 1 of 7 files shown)


mgcxx/text_search/src/lib.rs
```diff
@@ -808,6 +957,131 @@ fn search_get_fields(
...
+// Reads the string value of `property` from a hit's stored `data` JSON object, if present.
+fn extract_property_string(doc: &TantivyDocument, data_field: Field, property: &str) -> Option<String> {
+    match doc.get_first(data_field)?.into() {
+        OwnedValue::Object(entries) => entries.into_iter().find_map(|(k, v)| match v {
+            OwnedValue::Str(s) if k == property => Some(s),
+            _ => None,
+        }),
+        _ => None,
+    }
+}
```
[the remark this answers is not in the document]
DavIvek: It's not our bug and not sequence-specific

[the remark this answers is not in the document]
DavIvek: Verified this is not caused by `extract_property_string`. A property name with a dot is stored flat by nlohmann (`{"a.b": ...}`, not nested), the JSON field has `expand_dots` disabled, and the candidate query uses `from_field_json_path(..., expand_dots=false)` — so the candidate query, the index, and the post-filter all agree on the literal top-level key.

Empirically: `data.a.b:...` returns empty **even via the regular `text_search.search` proc**, while the same node is found via a normal property (`data.title:...`). So the empty result occurs at the candidate/json-path stage (a pre-existing limitation shared by all text-search procedures), before the sequence post-filter runs — not in this code.

mgcxx/text_search/src/lib.rs
```diff
@@ -808,6 +957,129 @@ fn search_get_fields(
...
+            Error::new(ErrorKind::Other, format!("search_sequence expects 'data.<property>:<terms>', got: {}", query))
+        })
+}
+
+// Runs the field's own analyzer over `text`, yielding the exact tokens tantivy indexed (Unicode
+// segmentation, RemoveLongFilter, lowercasing -- no approximation).
+fn analyzer_tokenize(analyzer: &mut TextAnalyzer, text: &str) -> Vec<Vec<char>> {
+    let mut tokens = Vec::new();
+    let mut stream = analyzer.token_stream(text);
+    while stream.advance() {
```
imilinovic: Adjacency check is position-blind → silent false positives. analyzer_tokenize (lib.rs:985) discards Token::position; tantivy's RemoveLongFilter drops tokens ≥40 bytes but keeps their position slot. So "big <45-char-blob> bad wolf" matches data.title:big bad wo even though the words aren't adjacent.

mgcxx/text_search/src/lib.rs
```diff
@@ -808,6 +957,129 @@ fn search_get_fields(
...
+            if total > budget {
+                return false;
+            }
+        }
+        true
+    })
+}
+
+// Reads the string value of `property` from a hit's stored `data` JSON object, if present.
+fn extract_property_string(doc: &TantivyDocument, data_field: Field, property: &str) -> Option<String> {
```
imilinovic, line 1080: extract_property_string deep-clones the entire stored JSON doc per candidate (lib.rs:1054). Reading data.title alone materializes every indexed property (description, body, …) into owned Strings — on every one of up to 50k candidates. Fix: use tantivy's zero-copy Value/ReferenceValue API.

mgcxx/text_search/src/lib.rs
```diff
@@ -1022,6 +1294,148 @@ fn search_edge_gids_pinned(
...
+            };
+            (Occur::Must, Box::new(fuzzy) as Box<dyn Query>)
+        })
+        .collect();
+    let query = BooleanQuery::new(subqueries);
+
+    // Page through score-ranked candidates, post-filtering each page, until we've collected
+    // `result_limit` matches or the candidate stream runs out. Selective queries fill the first page;
+    // we widen only when the post-filter is starved, capped at SEQUENCE_MAX_CANDIDATES.
+    let result_limit = if input.limit == 0 { usize::MAX } else { input.limit };
```
imilinovic: - limit: 0 diverges — means "unlimited (up to 50k)" for search_sequence but "default 1000" for every other search mode.

imilinovic (approved):

DavIvek:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Adds new fuzzy phrase search procedures that enable matching ordered, adjacent terms with typo tolerance, improving search flexibility, especially for partial and slightly misspelled queries. Users can now perform more forgiving and intuitive text searches (including prefix matching on the last term). [#4304](https://github.com/memgraph/memgraph/pull/4304)
- [x] **[memgraph/documentation/pull/1677]**

PR #4308, opened 2026-06-26
feat(storage): arity-specialize label-property index entry storage

Ignition:
Single- and two-property label-property index entries store their values in a fixed std::array inlined into the skiplist node, instead of a std::vector (a header plus a separate heap allocation) per entry; higher composite arities keep the vector. This removes the per-entry heap indirection that composite index support introduced for the common fixed-arity cases and restores the single-property entry footprint.

Entry storage is parameterized on composite arity and sort direction, with the per-arity index objects behind a variant chosen once at index creation. The query entry points are type-erased: they pick the entry type from the index's arity and direction and return the wrapper the executor consumes, so callers never name a per-arity type. A single dispatch layer holds the arity-and-order-to-entry-type policy that creation, population, publication, and the query paths share, so supporting another fixed arity is one edit. A strong IndexOrderedValuesView keeps index-ordered values distinct from storage order across the match and comparison helpers, and a static_assert pins each array-backed entry to its no-indirection size.

The ingestion path builds a fixed-arity entry straight into a stack array and moves it into the node, with no transient heap vector: PropertyStore and PropertiesPermutationHelper gain into-buffer extract and in-place permutation overloads, and the permutation carry moves rather than copies each value.

VectorIndex and VectorEdgeIndex return early from UpdateOnSetProperty when no vector index is defined, removing per-write overhead on databases without one.

### DIFF
--- a/src/storage/v2/disk/edge_import_mode_cache.cpp
+++ b/src/storage/v2/disk/edge_import_mode_cache.cpp
@@ -29,19 +29,20 @@ InMemoryLabelIndex::Iterable EdgeImportModeCache::Vertices(LabelId label, View v
       ->Vertices(label, vertices_.access(), view, storage, transaction);
 }
 
-InMemoryLabelPropertyIndex::Iterable<InMemoryLabelPropertyIndex::Entry> EdgeImportModeCache::Vertices(
+InMemoryLabelPropertyIndex::Iterable<InMemoryLabelPropertyIndex::Entry<1>> EdgeImportModeCache::Vertices(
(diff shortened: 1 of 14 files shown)


Ignition:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Label-property indexes on one or two properties use less memory, values are stored inline per entry instead of in a per-entry heap allocation, with no user-facing change. [#4308](https://github.com/memgraph/memgraph/pull/4308)

colinbarry (approved):
Approved 🚀

PR #2980, opened 2025-05-19
Support defining user profiles and resource limitation

andrejtonev:

PR #4309, opened 2026-06-26
perf(auth): Remove performance regressions caused by property-based access control

colinbarry:
Fix performance regressions introduced by property-based access control:

- Cache the `FineGrainedAuthChecker` at the session level during Prepare, avoiding recreation each query. Refresh only the `DbAccessor` pointer on cache hit.                                                                                   
- Add a fast-path in `FineGrainedAuthChecker::HasPropertyRestrictions()` to skip property permission checks when no property-level restrictions exist.                                                                         
- Reduce header coupling by pimpl'ing `CachedFineGrainedAuth` behind a `unique_ptr` in `Interpreter`, removing `auth_checker.hpp` from `interpreter.hpp`.
- Add e2e test for trigger property-based access control.

### DIFF
--- a/src/communication/bolt/v1/states/handlers.hpp
+++ b/src/communication/bolt/v1/states/handlers.hpp
@@ -588,11 +588,15 @@ State HandleRoute(TSession &session, const Marker marker) {
 
 template <typename TSession>
 State HandleLogOff(TSession &session) {
-  session.LogOff();
-  if (!session.encoder_.MessageSuccess({})) {
-    spdlog::trace("Couldn't send success message!");
-    return State::Close;
+  try {
+    session.LogOff();
+    if (!session.encoder_.MessageSuccess({})) {
+      spdlog::trace("Couldn't send success message!");
+      return State::Close;
+    }
+    return State::Init;
+  } catch (const std::exception &e) {
+    return HandleFailure(session, e);
   }
-  return State::Init;
 }
 }  // namespace memgraph::communication::bolt
--- a/src/communication/bolt/v1/states/handlers.hpp
+++ b/src/communication/bolt/v1/states/handlers.hpp
@@ -588,8 +588,11 @@ State HandleRoute(TSession &session, const Marker marker) {
 
 template <typename TSession>
 State HandleLogOff(TSession &session) {
-  // No arguments and cannot fail
   session.LogOff();
+  if (!session.encoder_.MessageSuccess({})) {
+    spdlog::trace("Couldn't send success message!");
+    return State::Close;
+  }
   return State::Init;
 }
 }  // namespace memgraph::communication::bolt
--- a/src/glue/SessionHL.cpp
+++ b/src/glue/SessionHL.cpp
@@ -352,6 +352,7 @@ std::expected<void, communication::bolt::AuthFailure> SessionHL::SSOAuthenticate
 }
 
 void SessionHL::LogOff() {
+  interpreter_.cached_auth_checker_ = nullptr;
   cached_auth_checker_.reset();
 #ifdef MG_ENTERPRISE
   interpreter_.ResetDB();
@@ -361,6 +362,7 @@ void SessionHL::LogOff() {
 }
 
 void SessionHL::Abort() {
+  interpreter_.cached_auth_checker_ = nullptr;
   cached_auth_checker_.reset();
   interpreter_.Abort();
 }
@@ -404,6 +406,7 @@ bolt_map_t SessionHL::Pull(std::optional<int> n, std::optional<int> qid) {
 }
 
 void SessionHL::InterpretParse(const std::string &query, bolt_map_t params, const bolt_map_t &extra) {
+  interpreter_.cached_auth_checker_ = nullptr;
   cached_auth_checker_.reset();
 #ifdef MG_ENTERPRISE
   if (memgraph::license::global_license_checker.IsEnterpriseValidFast()) {
@@ -455,6 +458,7 @@ std::pair<std::vector<std::string>, std::optional<int>> SessionHL::InterpretPrep
 
 #ifdef MG_ENTERPRISE
     cached_auth_checker_.reset();
+    interpreter_.cached_auth_checker_ = nullptr;
     auto &db = interpreter_.current_db_.db_acc_;
     auto *storage = db ? db->get()->storage() : nullptr;
     if (storage && interpreter_context_->auth_checker && interpreter_.user_or_role_ && *interpreter_.user_or_role_ &&
@@ -466,6 +470,7 @@ std::pair<std::vector<std::string>, std::optional<int>> SessionHL::InterpretPrep
       if (!cached_auth_checker_->NeedsFineGrainedAuthChecker()) {
         cached_auth_checker_.reset();
       }
+      interpreter_.cached_auth_checker_ = cached_auth_checker_.get();
     }
 #endif
 
--- a/src/glue/SessionHL.cpp
+++ b/src/glue/SessionHL.cpp
@@ -352,7 +352,7 @@ std::expected<void, communication::bolt::AuthFailure> SessionHL::SSOAuthenticate
 }
 
 void SessionHL::LogOff() {
-  interpreter_.cached_fga_.Reset();
+  interpreter_.ResetCachedFga();
 #ifdef MG_ENTERPRISE
   interpreter_.ResetDB();
 #endif
@@ -361,7 +361,7 @@ void SessionHL::LogOff() {
 }
 
 void SessionHL::Abort() {
-  interpreter_.cached_fga_.Reset();
+  interpreter_.ResetCachedFga();
   interpreter_.Abort();
 }
 
@@ -388,7 +388,7 @@ bolt_map_t SessionHL::Pull(std::optional<int> n, std::optional<int> qid) {
         communication::bolt::Encoder<communication::bolt::ChunkedEncoderBuffer<communication::v2::OutputStream>>;
     auto &db = interpreter_.current_db_.db_acc_;
     auto *storage = db ? db->get()->storage() : nullptr;
-    TypedValueResultStream<TEncoder> stream(&encoder_, storage, interpreter_.cached_fga_.get());
+    TypedValueResultStream<TEncoder> stream(&encoder_, storage, interpreter_.GetCachedFga());
     return DecodeSummary(interpreter_.Pull(&stream, n, qid));
   } catch (const memgraph::query::QueryException &e) {
     RewrapQueryException(e);
@@ -619,7 +619,7 @@ void RuntimeConfig::Configure(const bolt_map_t &run_time_info, bool in_explicit_
   // Runtime config is sent at the beginning of the transaction, but is missing during the transaction
   if (in_explicit_tx || (previous_run_time_info_ && run_time_info == *previous_run_time_info_)) return;
 
-  session_->interpreter_.cached_fga_.Reset();
+  session_->interpreter_.ResetCachedFga();
 
   db_explicit_ = false;
   user_explicit_ = false;
--- a/src/glue/SessionHL.cpp
+++ b/src/glue/SessionHL.cpp
@@ -352,8 +352,7 @@ std::expected<void, communication::bolt::AuthFailure> SessionHL::SSOAuthenticate
 }
 
 void SessionHL::LogOff() {
-  interpreter_.cached_auth_checker_.reset();
-  interpreter_.auth_checker_db_name_.clear();
+  interpreter_.cached_fga_.Reset();
 #ifdef MG_ENTERPRISE
   interpreter_.ResetDB();
 #endif
@@ -362,8 +361,7 @@ void SessionHL::LogOff() {
 }
 
 void SessionHL::Abort() {
-  interpreter_.cached_auth_checker_.reset();
-  interpreter_.auth_checker_db_name_.clear();
+  interpreter_.cached_fga_.Reset();
   interpreter_.Abort();
 }
 
@@ -390,7 +388,7 @@ bolt_map_t SessionHL::Pull(std::optional<int> n, std::optional<int> qid) {
         communication::bolt::Encoder<communication::bolt::ChunkedEncoderBuffer<communication::v2::OutputStream>>;
     auto &db = interpreter_.current_db_.db_acc_;
     auto *storage = db ? db->get()->storage() : nullptr;
-    TypedValueResultStream<TEncoder> stream(&encoder_, storage, interpreter_.cached_auth_checker_.get());
+    TypedValueResultStream<TEncoder> stream(&encoder_, storage, interpreter_.cached_fga_.get());
     return DecodeSummary(interpreter_.Pull(&stream, n, qid));
   } catch (const memgraph::query::QueryException &e) {
     RewrapQueryException(e);
@@ -621,8 +619,7 @@ void RuntimeConfig::Configure(const bolt_map_t &run_time_info, bool in_explicit_
   // Runtime config is sent at the beginning of the transaction, but is missing during the transaction
   if (in_explicit_tx || (previous_run_time_info_ && run_time_info == *previous_run_time_info_)) return;
 
-  session_->interpreter_.cached_auth_checker_.reset();
-  session_->interpreter_.auth_checker_db_name_.clear();
+  session_->interpreter_.cached_fga_.Reset();
 
   db_explicit_ = false;
   user_explicit_ = false;
--- a/src/glue/SessionHL.cpp
+++ b/src/glue/SessionHL.cpp
@@ -354,6 +354,7 @@ std::expected<void, communication::bolt::AuthFailure> SessionHL::SSOAuthenticate
 void SessionHL::LogOff() {
   interpreter_.cached_auth_checker_ = nullptr;
   cached_auth_checker_.reset();
+  auth_checker_db_name_.clear();
 #ifdef MG_ENTERPRISE
   interpreter_.ResetDB();
 #endif
@@ -364,6 +365,7 @@ void SessionHL::LogOff() {
 void SessionHL::Abort() {
   interpreter_.cached_auth_checker_ = nullptr;
   cached_auth_checker_.reset();
+  auth_checker_db_name_.clear();
   interpreter_.Abort();
 }
 
@@ -406,8 +408,6 @@ bolt_map_t SessionHL::Pull(std::optional<int> n, std::optional<int> qid) {
 }
 
 void SessionHL::InterpretParse(const std::string &query, bolt_map_t params, const bolt_map_t &extra) {
-  interpreter_.cached_auth_checker_ = nullptr;
-  cached_auth_checker_.reset();
 #ifdef MG_ENTERPRISE
   if (memgraph::license::global_license_checker.IsEnterpriseValidFast()) {
     auto &db = interpreter_.current_db_.db_acc_;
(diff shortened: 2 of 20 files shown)


colinbarry:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Remove performance regressions added by the property-based access control, and improve overall performance of queries with label-based or property-based access controls. [#4309](https://github.com/memgraph/memgraph/pull/4309)

src/glue/SessionHL.cpp
```diff
@@ -343,14 +343,18 @@ std::expected<void, communication::bolt::AuthFailure> SessionHL::SSOAuthenticate
 }
 
 void SessionHL::LogOff() {
+  interpreter_.ResetCachedFga();
```
andrejtonev: `LogOff()` frees the cached FGA checker (`ResetCachedFga()`) and tears down the DB (`ResetDB()`) but, unlike `Abort()`, never clears `in_explicit_transaction_` or `query_executions_`. If a client is in an explicit transaction, then logs off and back on (the neo4j driver's standard `re_auth()` pipelines LOGOFF+LOGON on the same connection), the next `RUN` takes the `in_explicit_transaction_` branch at `interpreter.cpp:9662` (`AdvanceCommand()` only, no `SetupInterpreterTransaction`) — but the DB was already torn down, so `current_db.execution_db_accessor_` is null and `MG_ASSERT` at `interpreter.cpp:3586` aborts the process. `MG_ASSERT` is **not** NDEBUG-gated, so this fires in release builds — confirmed crashing on both ASan-Debug and RelWithDebInfo. This is a 100%-reproducible, remotely-triggerable DoS reachable by any authenticated client with ordinary query privileges, using a standard driver feature. Please make `LogOff()` mirror `Abort()` — `CleanRuntimeData()` over `query_executions_` and reset `in_explicit_transaction_` — or reject LOGOFF while `in_explicit_transaction_` has live executions. (This also closes the freed-`auth_checker` dangle the caching change newly introduced; the crash just happens first.)

andrejtonev: **`src/communication/bolt/v1/states/handlers.hpp:590`** (`HandleLogOff`) `HandleLogOff` returns `State::Init` without sending a SUCCESS/FAILURE for the LOGOFF message. A spec-compliant client that waits for the LOGOFF reply before sending LOGON hangs forever. The neo4j driver happens to pipeline both so it's not hit in practice, but this is a protocol-conformance gap.

src/query/auth_checker.hpp
```diff
@@ -211,4 +224,32 @@ class AllowEverythingAuthChecker final : public AuthChecker {
...
+struct CachedFineGrainedAuth {
+  std::unique_ptr<FineGrainedAuthChecker> checker;
+  std::string db_name;
+  bool checked{false};
+
+  FineGrainedAuthChecker const *get() const { return checker.get(); }
+
+  void Refresh(AuthChecker const &auth_checker, QueryUserOrRole const &user, DbAccessor const *dba,
+               std::string current_db) {
+    if (checked && db_name == current_db) {
```
andrejtonev: The reuse branch `if (checked && db_name == current_db)` at `auth_checker.hpp:236` only calls `UpdateDbAccessor` and never re-evaluates the license — `IsEnterpriseValidFast()` is checked *only* in the rebuild branch, inside `GetFineGrainedAuthChecker` (`glue/auth_checker.cpp:124`). A session that runs its first query under an invalid license caches `{checker=nullptr, checked=true}` (`checked = true` is set unconditionally at `auth_checker.hpp:244`); if the license is then activated mid-session (`SET DATABASE SETTING "enterprise.license" …`), same-DB queries keep hitting the reuse branch and FGA/PBAC is never enforced for that session until logoff/abort/DB-switch. Verified live: with the license valid (`SHOW LICENSE INFO: is_valid=True`), a query on the original connection still returned DENY-restricted `:Secret` nodes, while a fresh session correctly hid them. Master re-checked the license on every `PullPlan` construction, so this is a fail-open regression, and nothing resets the cache on a license-state change. Please re-evaluate the license in the reuse branch (rebuild when `checker==nullptr` but the license is now valid), or reset the cache on a license transition.

andrejtonev: `Refresh` invalidates the cached checker only on DB-name change (`db_name == current_db`), never on an auth-epoch bump. An already-connected session therefore keeps using the checker built at its first query and does **not** honor a later external GRANT/REVOKE/DENY. Verified live (RelWithDebInfo): after an external `DENY READ ... :Secret` plus a 2s wait (to rule out auth-propagation latency), the same session still returned the denied nodes; a fresh session enforced correctly. Because the row engine (`ctx_.auth_checker` used by ScanAll/Expand) and the Bolt output encoder share the *same* cached pointer, this is a **row-level** staleness (returns nodes/edges the current policy would deny), not just output redaction. Master rebuilt the checker per-Pull, so this is likely a regression (not confirmed against a master build). Note: the review's specific "SHOW/TERMINATE TRANSACTIONS leave the checker stale" claim did **not** reproduce — running `SHOW TRANSACTIONS` actually makes the session enforce, and those commands are rejected inside explicit transactions. Suggest keying the cache on the auth epoch so it rebuilds on GRANT/REVOKE (this also subsumes #2).

src/glue/auth_checker.cpp
andrejtonev, line 432: `MakeThreadSafe()` (`auth_checker.cpp:414`) populates the base-class memo `has_property_restrictions_` via `(void)HasPropertyRestrictions()`, but `IsThreadSafe()` → `IsCachedPermissionsPopulated()` (`:432`) only checks the 8 glue caches and cannot check that memo — it's `private` in the base (`auth_checker.hpp:122`) and unreachable from the glue subclass. So the `DMG_ASSERT(... IsThreadSafe())` guard at `operator.cpp:10450` can't actually verify what `MakeThreadSafe` populated (and `DMG_ASSERT` is compiled out under RelWithDebInfo/NDEBUG anyway). Race-free today only by convention — the memo is populated single-threaded before task submission — but a future `MakeThreadSafe`/subclass change could silently reintroduce a `mutable` data race that this guard would miss. Suggest making the field `protected` or adding a `protected bool IsPropertyRestrictionsCached()` accessor and folding it into `IsCachedPermissionsPopulated()`.

src/query/auth_checker.hpp
```diff
@@ -211,4 +224,32 @@ class AllowEverythingAuthChecker final : public AuthChecker {
...
 
+struct CachedFineGrainedAuth {
+  std::unique_ptr<FineGrainedAuthChecker> checker;
+  std::string db_name;
+  bool checked{false};
+
+  FineGrainedAuthChecker const *get() const { return checker.get(); }
+
+  void Refresh(AuthChecker const &auth_checker, QueryUserOrRole const &user, DbAccessor const *dba,
+               std::string current_db) {
```
andrejtonev, line 243: `Refresh` takes `std::string current_db` by value and the call site at `interpreter.cpp:9801` passes `dba->DatabaseName()`, which `shared_lock`s and heap-copies the name on **every** query — including the cache-hit path (`auth_checker.hpp:236`) that only compares and discards it (the arg is materialized before the hit/miss decision; it's only moved into `db_name` on the miss path at `:243`). Microbench (RelWithDebInfo): 104 ns for a 38-char non-SSO name vs 4.68 ns for `DatabaseNameView()` — ~99 ns wasted per `Prepare()` for non-SSO DB names (SSO names hide it). Change the param to `std::string_view` and pass `*dba->DatabaseNameView()`, materializing the `std::string` only in the rebuild branch. (History note: this alloc was reintroduced by `8a196a3fb`, which reverted `DatabaseNameView()` back to `DatabaseName()`; `ceb2d9147`'s literal diff didn't compile — `ConstSafeWrapper` has no `operator==`.)

tests/e2e/triggers/privilege_check.cpp
```diff
@@ -414,6 +423,51 @@ TEST_P(PrivilegeCheckTest, DefinerFineGrainedNoSet) {
...
+
+  auto pbac_client = ConnectWithUser(kPbacUser);
+
+  // Trigger copies v.secret into v.result. If PBAC is enforced, v.secret is
+  // null.
+  pbac_client->Execute(
+      fmt::format("CREATE TRIGGER PbacReadDeny SECURITY INVOKER ON CREATE "
+                  "{} COMMIT EXECUTE "
+                  "UNWIND createdVertices AS v "
+                  "SET v.result = v.secret",
```
andrejtonev: The trigger copies `v.secret` (a denied read) into `v.result`, and the test asserts `n.result IS NULL`. But `result` is also null if the trigger never wrote it. Mutation-verified: changing the trigger to `SET v.mutation_a_unrelated = v.secret` (so `.result` is never written) still passes both BeforeCommit and AfterCommit — the null means "property never existed", not "denied read yielded null". Add a positive control in the same trigger, e.g. `SET v.result = v.secret, v.result_ok = v.id` on a non-denied property, and assert `v.result_ok == kVertexId` alongside `v.result IS NULL`, so a null `result` uniquely means the read was denied. Separately (follow-up, not this PR): while mutating the DENY→GRANT axis, the AfterCommit phase read a *granted* property as null when it ran right after BeforeCommit in the same process — sequential BEFORE→AFTER trigger instances sharing a name/vertex-id appear to leak auth/plan state. It's dormant under the DENY config (correct and buggy outcomes are both null) and would only be caught by a positive (GRANT) test. Worth a separate investigation.

src/query/auth_checker.hpp
colinbarry: Re: epoch-bump. The stale auth is an effect of the `session_long_policy`, which is used in the query path. Under that policy, auth privileges and permissions are fixed for the duration of the session, and this PR reflects that. (Future PR should remove `session_long_policy` and we should use `up_to_date_policy`, but that's beyond the scope of this change.)

colinbarry: Re: Setting enterprise licence mid-session: fixed

src/glue/SessionHL.cpp
colinbarry: Fixed `LogOff()`. This was pre-existing crash where where `LogOff()` didn't clean-up the transaction. Now, `LogOff()` calls `Abort()` at the start, which safely tears down in-flight transactions before releasing the DB and user.

colinbarry: `HandleLogOff` was a pre-existing issue from #2980 . Fixed anyway to match protocol.

src/glue/auth_checker.cpp
colinbarry: Moved `has_property_restrictions_` to protected and added `IsPropertyRestrictionsCached()` accessor. `DMG_ASSERT(IsThreadSafe())` guard now covers everything that  `MakeThreadSafe` populates.

src/query/auth_checker.hpp
colinbarry: Fixed

tests/e2e/triggers/privilege_check.cpp
colinbarry: Yeah, good idea. Expanded the test to have a non-write-restricted property as a check that the trigger actually ran.

src/communication/bolt/v1/states/handlers.hpp
```diff
@@ -588,8 +588,11 @@ State HandleRoute(TSession &session, const Marker marker) {
 
 template <typename TSession>
 State HandleLogOff(TSession &session) {
```
andrejtonev, line 590: HandleLogOff calls session.LogOff() with no try/catch, unlike HandleReset (:403-412) which wraps the identical session.Abort(). This PR made LogOff() call Abort() → interpreter_.Abort() (not noexcept; can throw std::bad_alloc from storage teardown), so a throw during LOGOFF now escapes to the top-level loop and drops the connection instead of returning a Bolt FAILURE. 
Please wrap session.LogOff() + the success send in try { … } catch (const std::exception &e) { return HandleFailure(session, e); } to match HandleReset.

src/query/interpreter.cpp
andrejtonev, line 10893: CachedFineGrainedAuth::Refresh (auth_checker.hpp:237) keys the reuse guard on db_name only — never on user identity — and Interpreter::SetUser doesn't reset the cache. Safe today only because every user-transition path happens to reset first (LogOff→Abort→ResetCachedFga, Configure at :613), but that invariant is unenforced: a future SetUser caller that skips the reset would serve one user's permissions to another on the same DB. Note this is not the session_long_policy staleness — it's a user-identity issue. Cheap defense: add a user-identity fingerprint to the reuse guard, and/or call ResetCachedFga() in SetUser. (A guard test is included: tests/unit/cached_fga_user_identity.cpp.)

src/query/auth_checker.hpp
```diff
@@ -211,4 +225,34 @@ class AllowEverythingAuthChecker final : public AuthChecker {
...
+        return;
+      }
+    }
+    checker = auth_checker.GetFineGrainedAuthChecker(user, dba);
+    if (!checker || !checker->NeedsFineGrainedAuthChecker()) {
+      checker.reset();
+    }
+    db_name = current_db;
+    checked = true;
+  }
```
andrejtonev, line 268: Refresh gives no caching benefit to a licensed no-restriction user: {checker=nullptr, checked=true} never hits the if (checker) short-circuit, so GetFineGrainedAuthChecker is re-run every Prepare. Not a regression — master was worse — but commit 88aa2d2c8 "Skip auth check when no auth applied" doesn't actually skip. Either distinguish "checked, genuinely no restrictions" from "null because unlicensed" (a sentinel, or a cheap up-front IsEnterpriseValidFast() gate) so the common case caches too.

src/query/interpret/eval.cpp
```diff
@@ -568,14 +568,24 @@ TypedValue ExpressionEvaluator::Visit(PropertyLookup &property_lookup) {
 
 #ifdef MG_ENTERPRISE
 bool ExpressionEvaluator::IsPropertyAllowed(VertexAccessor const &accessor, storage::PropertyId prop) const {
-  if (!auth_checker_) return true;
+  if (!auth_checker_ || !auth_checker_->HasPropertyRestrictions()) return true;
```
andrejtonev, line 571: The property fast-path skips the accessor.Labels() fetch, so for an LBAC-only user a concurrently-deleted object now throws QueryRuntimeException where master returned a silent null. No data leaks; it's an exception-vs-null divergence on a race edge. Add a regression test or note it as intended.

src/glue/auth_checker.cpp
andrejtonev, line 397: Now reached per restricted-property-read from eval.cpp:580, this rebuilds a std::vector<std::string> label_names via dba_->LabelToName(...) on every call. Previously ~once per privilege check.
Consider a std::views::transform to string_view or caching the name resolution.

src/query/interpreter.cpp
```diff
@@ -9782,6 +9795,15 @@ Interpreter::PrepareResult Interpreter::Prepare(ParseRes parse_res, UserParamete
       };
     };
 
+#ifdef MG_ENTERPRISE
+    if (current_db_.execution_db_accessor_ && interpreter_context_->auth_checker && user_or_role_ && *user_or_role_) {
+      auto *dba = &*current_db_.execution_db_accessor_;
+      cached_fga_->Refresh(*interpreter_context_->auth_checker, *user_or_role_, dba, *dba->DatabaseNameView());
```
andrejtonev: Reentrant shared_lock on the DB-name mutex in CachedFineGrainedAuth::Refresh (miss path) src/query/interpreter.cpp:9801 · src/query/auth_checker.hpp:235-249 · src/glue/auth_checker.cpp:213 · src/utils/safe_string.hpp:68,79-88 

Refresh(..., *dba->DatabaseNameView()) acquires the DB name's shared_mutex and holds it for the entire Refresh body, and on the rebuild branch we then lock that same mutex a second time on the same thread. Both locks target the one SafeString::mutex_ behind config_.salient.name.

Lock #1 — the argument (held across the whole call). *dba->DatabaseNameView() → SafeString::str_view() (safe_string.hpp:88) returns a ConstSafeWrapper by value, and that wrapper is the lock:
 struct ConstSafeWrapper {
   ConstSafeWrapper(const SafeString &s) : safe_str_(s), lock_(s.mutex_) {}  // safe_string.hpp:79
   std::shared_lock<std::shared_mutex> lock_;                                // :85
 };
It's a temporary in the Refresh(...) full-expression, so it isn't destroyed until Refresh returns — the shared_lock is held for the whole call, not just the db_name compare.

Lock #2 — the rebuild ctor. On a cache miss Refresh calls GetFineGrainedAuthChecker(user, dba), which constructs glue::FineGrainedAuthChecker whose initializer db_name_{dba_->DatabaseName()} (auth_checker.cpp:213) calls DatabaseName() → SafeString::str() (safe_string.hpp:68), which does std::shared_lock lock(mutex_) on the same mutex, on the same thread, while Lock #1 is still held.

Why it matters. Recursive lock_shared() while the thread already owns the mutex is undefined behavior per [thread.sharedmutex.requirements]. It happens to be harmless on glibc (default pthread_rwlock is reader-preferring, so the nested reader just succeeds), which is why nothing fails today — but on a writer-preferring implementation it self-deadlocks the moment a concurrent RENAME DATABASE (SafeString::operator=, unique_lock) queues between the two shared acquisitions: the writer waits for reader #1 to drain, reader #2 waits behind the writer. Even on glibc it needlessly serializes Refresh against concurrent renames. (Hit path is unaffected in release — UpdateDbAccessor's DMG_ASSERT(dba->DatabaseName()==db_name_) also re-locks, but it's compiled out under NDEBUG; the live reentrancy is the miss-path ctor.)

src/communication/bolt/v1/states/handlers.hpp
colinbarry: Fixed: wrapped `session.LogOff()` in `try`/`catch` to match `HandleReset`.

src/query/interpreter.cpp
colinbarry: Fixed. `SetUser` now calls `ResetCachedFga()` defensively.

src/query/interpret/eval.cpp
colinbarry: As intended. `master`'s silent null was a side effect of the auth check swallowing `DELETED_OBJECT` via `Labels()`. The non-auth path has always thrown `QueryRuntimeException` on deleted objects. Our code is now consistent with the normal code path.

src/query/auth_checker.hpp
colinbarry: Fixed. Replaced the checked bool with an explicit State enum (`EMPTY`, `NO_LICENSE`, `NO_RESTRICTIONS`, `ACTIVE`). `NO_RESTRICTIONS` now caches correctly. `NO_LICENSE` re-evaluates every query so license activation mid-session is detected

src/query/interpreter.cpp
colinbarry: Fixed. Refresh now takes the `db_name` `std::string` by value; the caller passes `dba->DatabaseName()`. This eliminates the reentrant lock. One string copy per query: negligible vs query execution cost.

src/glue/auth_checker.cpp
colinbarry: Acknowledged, but unresolved. `LabelToName` returns `const std::string &` from the name-id mapper, so a `string_view` cache is feasible. However, the proper fix touches the `Has` signature in the auth model and has lots of plumbing for operators to invalidate a cache. Too begin of a change for this PR, but will implement in a future PR.

andrejtonev (approved):

PR #4312, opened 2026-06-29
fix: Negative replication lag in the routing table

as51340:
Fixes multiple issues that caused negative replication lag in the routing table and as a consequence, invalid routing table.

### DIFF
--- a/src/replication_handler/replication_handler.cpp
+++ b/src/replication_handler/replication_handler.cpp
@@ -413,29 +413,47 @@ auto ReplicationHandler::GetReplicationLag() const -> coordination::ReplicationL
     auto &repl_storage_state = db_acc->storage()->repl_storage_state_;
     auto const db_name = db_acc->name();
 
-    repl_storage_state.replication_storage_clients_.WithReadLock(
-        [&db_name, &lag_info, &repl_storage_state](auto &storage_clients) {
-          // First observe replicas' num committed txns, then main's, so a concurrent commit can't make
-          // main's value smaller than the replica's and produce a spurious negative lag.
-          std::map<std::string, uint64_t> repl_committed_txns;
-          for (auto &repl_storage_client : storage_clients) {
-            repl_committed_txns.emplace(repl_storage_client->Name(), repl_storage_client->GetNumCommittedTxns());
-          }
+    repl_storage_state.replication_storage_clients_.WithReadLock([&db_name, &lag_info, &repl_storage_state](
+                                                                     auto &storage_clients) {
+      // First observe replicas' num committed txns, then main's, so a concurrent commit can't make
+      // main's value smaller than the replica's and produce a spurious negative lag. This ordering only
+      // helps SYNC replicas, whose cache is bumped in program order after main's counter; an ASYNC
+      // replica records its progress on a worker thread with no ordering against main's bump, so track
+      // its mode to floor its lag below.
+      struct ReplicaProgress {
+        uint64_t num_committed_txns;
+        bool is_async;
+      };
+      std::map<std::string, ReplicaProgress> repl_committed_txns;
+      for (auto &repl_storage_client : storage_clients) {
+        repl_committed_txns.emplace(repl_storage_client->Name(),
+                                    ReplicaProgress{.num_committed_txns = repl_storage_client->GetNumCommittedTxns(),
+                                                    .is_async = repl_storage_client->Mode() ==
+                                                                replication_coordination_glue::ReplicationMode::ASYNC});
+      }
 
-          auto const num_main_committed_txns =
-              repl_storage_state.commit_ts_info_.load(std::memory_order_acquire).num_committed_txns_;
-          lag_info.dbs_main_committed_txns_.emplace(db_name, num_main_committed_txns);
+      auto const num_main_committed_txns =
+          repl_storage_state.commit_ts_info_.load(std::memory_order_acquire).num_committed_txns_;
+      lag_info.dbs_main_committed_txns_.emplace(db_name, num_main_committed_txns);
 
-          for (auto const &[replica_name, num_committed_txns_repl] : repl_committed_txns) {
-            auto const replica_lag = static_cast<int64_t>(num_main_committed_txns - num_committed_txns_repl);
-            // Insert or find the already inserted element
-            auto [replica_it, _] = lag_info.replicas_info_.try_emplace(
-                replica_name, std::map<std::string, coordination::ReplicaDBLagData>{});
-            replica_it->second.emplace(db_name,
-                                       coordination::ReplicaDBLagData{.num_committed_txns_ = num_committed_txns_repl,
-                                                                      .num_txns_behind_main_ = replica_lag});
-          }
-        });
+      for (auto const &[replica_name, progress] : repl_committed_txns) {
+        auto replica_lag =
+            static_cast<int64_t>(num_main_committed_txns) - static_cast<int64_t>(progress.num_committed_txns);
+        // An ASYNC replica can transiently appear one txn ahead of main (its progress is recorded on a
+        // worker thread with no happens-before against main's counter bump). A non-divergent replica can't
+        // truly lead main, so floor the artifact to 0. A SYNC replica bumps after main in program order, so
+        // a negative there would be a real invariant violation and is intentionally left visible.
+        if (progress.is_async && replica_lag < 0) {
+          replica_lag = 0;
+        }
+        // Insert or find the already inserted element
+        auto [replica_it, _] =
+            lag_info.replicas_info_.try_emplace(replica_name, std::map<std::string, coordination::ReplicaDBLagData>{});
+        replica_it->second.emplace(db_name,
+                                   coordination::ReplicaDBLagData{.num_committed_txns_ = progress.num_committed_txns,
+                                                                  .num_txns_behind_main_ = replica_lag});
+      }
+    });
   });
   return lag_info;
 }
@@ -470,7 +488,15 @@ std::pair<ReplicationHandler::MainResT, ReplicationHandler::ReplicasResT> Replic
             DMG_ASSERT(replica_it != replicas.end(), "No info for replica {}", replica_name);
             auto old_value_it = replica_it->second.find(db_name);
             DMG_ASSERT(old_value_it != replica_it->second.end(), "No info for db {}", db_name);
-            old_value_it->second = num_main_committed_txns - old_value_it->second;
+            auto lag = static_cast<int64_t>(num_main_committed_txns) - old_value_it->second;
+            // An ASYNC replica records its progress on a worker thread with no happens-before against main's
+            // counter bump, so it can transiently appear one txn ahead of main. Floor that artifact to 0; a
+            // SYNC replica bumps after main in program order, so a negative there is a real invariant violation
+            // and is left visible.
+            if (repl_storage_client->Mode() == replication_coordination_glue::ReplicationMode::ASYNC && lag < 0) {
+              lag = 0;
+            }
+            old_value_it->second = lag;
           }
         });
   });
--- a/src/replication_handler/replication_handler.cpp
+++ b/src/replication_handler/replication_handler.cpp
@@ -74,6 +74,18 @@ void RecoverReplication(utils::Synchronized<ReplicationState, utils::RWSpinLock>
   auto const result = std::visit(utils::Overloaded{replica, main}, repl_state->ReplicationData());
   MG_ASSERT(result, "Replica recovery failure!");
 }
+
+// Lag of a replica behind main, in committed transactions. An ASYNC replica records its progress on a
+// worker thread with no happens-before against main's counter bump, so it can transiently appear one txn
+// ahead of main; floor that artifact to 0. A SYNC replica bumps after main in program order, so a negative
+// there is a real invariant violation and is left visible.
+int64_t ReplicaLagBehindMain(uint64_t num_main_committed_txns, uint64_t num_replica_committed_txns, bool is_async) {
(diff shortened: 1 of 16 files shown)


as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Fixes several issues that could've caused negative replication lag and consequently invalid routing table. **[#4312]()**

src/replication_handler/replication_handler.cpp
```diff
@@ -412,15 +412,21 @@ auto ReplicationHandler::GetReplicationLag() const -> coordination::ReplicationL
...
+          std::map<std::string, uint64_t> repl_committed_txns;
           for (auto &repl_storage_client : storage_clients) {
-            auto const replica_name = repl_storage_client->Name();
-            auto const num_committed_txns_repl = repl_storage_client->GetNumCommittedTxns();
+            repl_committed_txns.emplace(repl_storage_client->Name(), repl_storage_client->GetNumCommittedTxns());
+          }
+
+          auto const num_main_committed_txns =
+              repl_storage_state.commit_ts_info_.load(std::memory_order_acquire).num_committed_txns_;
+          lag_info.dbs_main_committed_txns_.emplace(db_name, num_main_committed_txns);
```
andrejtonev: On the no-STRICT_SYNC commit path, `ShipDeltas`→`FinalizeTransactionReplication` *enqueues* the ASYNC replica's cache bump to the per-client `thread_pool_` (client.cpp:609); that bump runs later on a worker thread (client.cpp:584). Main bumps its *own* counter only afterwards, in `FinalizeCommitPhase` (storage.cpp:1208→1334), on the committing thread. There is no happens-before edge between the two, so the worker can reach N before main reaches N. This function reads all replica caches first (`repl_committed_txns`, line 422) then main last (line 426), so it can compute `replica(N) − main(N-1) = -1`.

andrejtonev: can we just clamp?

src/storage/v2/replication/replication_client.cpp
```diff
@@ -227,7 +224,15 @@ void ReplicationStorageClient::UpdateReplicaState(Storage *main_storage, Databas
     return;
   }
 
-  // No branching point
+  // No branching point. The replica's history is consistent with main's, so its reported progress can't exceed
+  // main's; record it now.
+  commit_ts_info_.store(CommitTsInfo{.ldt_ = heartbeat_res.current_commit_timestamp_,
+                                     .num_committed_txns_ = heartbeat_res.num_txns_committed_},
+                        std::memory_order_release);
```
andrejtonev: The `!heartbeat_res.success_` early-`return` is inside `#ifdef MG_ENTERPRISE` (lines 113-123), so in a community build a *failed* heartbeat falls through the empty `#endif`. A failed `HeartbeatRes{false,0,"",0}` has `current_commit_timestamp_ == 0 == kTimestampInitialId`, so the branching-point check is skipped and control now reaches the **relocated** unconditional store at line 229, overwriting the cached replica progress with `{0,0}`. On master the store lived inside `if (success_)` and never ran on a failed heartbeat. Effect: a transient heartbeat failure zeroes `.ldt_`, producing a spurious `SHOW REPLICAS ... behind` / positive lag until the next good heartbeat.

src/storage/v2/inmemory/storage.cpp
```diff
@@ -2825,26 +2825,30 @@ Transaction InMemoryStorage::CreateTransaction(IsolationLevel isolation_level, S
...
+                          storage_mode,
+                          false,
+                          *std::move(point_index_context),
+                          std::move(active_indices),
+                          std::move(active_constraints),
+                          std::move(async_index_helper),
+                          commit_ts_info.ldt_,
+                          metric_handles_.unreleased_delta_objects};
+  transaction.last_durable_num_committed_txns_ = commit_ts_info.num_committed_txns_;
+  return transaction;
```
andrejtonev: random placement

please check why we use the old_info here line 4400
```
auto const update_func =
        [new_ldt = recovered_snapshot.snapshot_info.durable_timestamp](CommitTsInfo const &old_info) -> CommitTsInfo {
      return CommitTsInfo{.ldt_ = new_ldt, .num_committed_txns_ = old_info.num_committed_txns_};
    };
```

src/storage/v2/replication/replication_client.cpp
```diff
@@ -227,7 +224,15 @@ void ReplicationStorageClient::UpdateReplicaState(Storage *main_storage, Databas
...
+  spdlog::trace("Set num committed txns to {}", heartbeat_res.num_txns_committed_);
```
andrejtonev, line 259: this will read the atomic value regardless of the log level. encapsulate with if (should_log) or create a wrapper that fires only if printing (there is already some such structure somewhere).

src/replication_handler/replication_handler.cpp
```diff
@@ -412,15 +412,21 @@ auto ReplicationHandler::GetReplicationLag() const -> coordination::ReplicationL
...
-    auto const num_main_committed_txns =
-        repl_storage_state.commit_ts_info_.load(std::memory_order_acquire).num_committed_txns_;
-    lag_info.dbs_main_committed_txns_.emplace(db_name, num_main_committed_txns);
 
     repl_storage_state.replication_storage_clients_.WithReadLock(
-        [&db_name, &lag_info, &num_main_committed_txns](auto &storage_clients) {
+        [&db_name, &lag_info, &repl_storage_state](auto &storage_clients) {
+          // First observe replicas' num committed txns, then main's, so a concurrent commit can't make
+          // main's value smaller than the replica's and produce a spurious negative lag.
+          std::map<std::string, uint64_t> repl_committed_txns;
```
andrejtonev: do we really need a map here?

src/storage/v2/inmemory/storage.cpp
```diff
@@ -2825,26 +2825,30 @@ Transaction InMemoryStorage::CreateTransaction(IsolationLevel isolation_level, S
...
-          isolation_level,
-          storage_mode,
-          false,
-          *std::move(point_index_context),
-          std::move(active_indices),
-          std::move(active_constraints),
-          std::move(async_index_helper),
-          last_durable_ts,
-          metric_handles_.unreleased_delta_objects};
+  Transaction transaction{transaction_id,
```
andrejtonev: why not just return as before?

tests/e2e/high_availability/replication_lag.py
```diff
@@ -413,5 +434,64 @@ def test_replication_lag_failover(test_name):
     mg_sleep_and_assert(leader_data, partial(show_instances, coord3_cursor))
 
 
+def test_replication_lag_not_negative_after_divergent_main_rejoins(test_name):
```
andrejtonev: should we test all replication types?

as51340: yes

src/storage/v2/inmemory/storage.cpp
as51340: same, I will revert

src/replication_handler/replication_handler.cpp
as51340: yes, this was the same before, why not?

src/storage/v2/replication/replication_client.cpp
as51340: makes sense

src/storage/v2/inmemory/storage.cpp
as51340: bug, fixed it now

src/storage/v2/replication/replication_client.cpp
as51340: fixed now

src/replication_handler/replication_handler.cpp
as51340: Agree and clamped for async replicas

src/replication_handler/replication_handler.cpp
```diff
@@ -464,7 +488,15 @@ std::pair<ReplicationHandler::MainResT, ReplicationHandler::ReplicasResT> Replic
             DMG_ASSERT(replica_it != replicas.end(), "No info for replica {}", replica_name);
             auto old_value_it = replica_it->second.find(db_name);
             DMG_ASSERT(old_value_it != replica_it->second.end(), "No info for db {}", db_name);
-            old_value_it->second = num_main_committed_txns - old_value_it->second;
+            auto lag = static_cast<int64_t>(num_main_committed_txns) - old_value_it->second;
```
andrejtonev: The dev's "SYNC bumps after main in program order" is true only for the commit-thread UpdateCommitTsInfo path — but a second, unsynchronized writer, the background heartbeat (replication_client.cpp:231, no engine_lock_), stores the replica's self-count. A plain-SYNC replica finalizes its own commit (count→N) before main bumps its counter at storage.cpp:1208, so a concurrent heartbeat makes GetReplicationLag read cache=N, main=N−1 = −1, unclamped. STRICT_SYNC (2PC) is the only safe mode — opposite of the dev's comment. Fix: clamp all non-STRICT_SYNC.

src/storage/v2/inmemory/storage.cpp
```diff
@@ -2843,7 +2845,8 @@ Transaction InMemoryStorage::CreateTransaction(IsolationLevel isolation_level, S
           std::move(active_indices),
           std::move(active_constraints),
           std::move(async_index_helper),
-          last_durable_ts,
+          commit_ts_info.ldt_,
+          commit_ts_info.num_committed_txns_,
```
andrejtonev, line 2841: SnapshotInfo::num_committed_txns (snapshot.hpp:63) has no default initializer, and the pre-v30 readers never set it, so RECOVER SNAPSHOT of a pre-upgrade snapshot seeds the counter from indeterminate stack memory — reintroducing arbitrary lag on the exact upgrade path this branch targets. Fix: num_committed_txns{0} at snapshot.hpp:63.

src/replication_handler/replication_handler.cpp
as51340: fixed by taking engine lock before updating in UpdateReplicaState

src/storage/v2/replication/replication_client.cpp
```diff
@@ -227,16 +226,32 @@ void ReplicationStorageClient::UpdateReplicaState(Storage *main_storage, Databas
...
-  // No branching point
   // Lock engine lock in order to read main_storage timestamp and synchronize with any active commits
   auto engine_lock = std::unique_lock{main_storage->engine_lock_};
+
+  // No branching point. The replica's history is consistent with main's, so its reported progress can't exceed
+  // main's; record it now.
+  // This needs to be done under the engine lock, otherwise this function could bump replica's num committed txns before
+  // commit thread on main. This would be possible because there is a time window between we release RPC lock and update
+  // main's number of committed txns in the commit thread
+  commit_ts_info_.store(CommitTsInfo{.ldt_ = heartbeat_res.current_commit_timestamp_,
```
andrejtonev: this .store() isn't monotonic. num_txns_committed_ is captured before we take engine_lock_; if a commit lands in that window the commit thread bumps this client's count via the +1 path and this store overwrites it back down. engine_lock_ stops torn writes but not stale-overwrite, so the cached count can regress → SHOW REPLICATION LAG jumps positive and we may kick off an unnecessary RecoverReplica. Suggest an advance-only atomic_struct_update instead of a blind store.

src/replication_handler/replication_handler.cpp
```diff
@@ -412,24 +412,48 @@ auto ReplicationHandler::GetReplicationLag() const -> coordination::ReplicationL
...
+      lag_info.dbs_main_committed_txns_.emplace(db_name, num_main_committed_txns);
+
+      for (auto const &[replica_name, progress] : repl_committed_txns) {
+        auto replica_lag =
+            static_cast<int64_t>(num_main_committed_txns) - static_cast<int64_t>(progress.num_committed_txns);
+        // An ASYNC replica can transiently appear one txn ahead of main (its progress is recorded on a
+        // worker thread with no happens-before against main's counter bump). A non-divergent replica can't
+        // truly lead main, so floor the artifact to 0. A SYNC replica bumps after main in program order, so
+        // a negative there would be a real invariant violation and is intentionally left visible.
+        if (progress.is_async && replica_lag < 0) {
```
andrejtonev: Clamp/policy duplicated in 3 sites (replication_handler.cpp:446/496 + utils.hpp:82) — extract a helper.

src/replication_handler/replication_handler.cpp
```diff
@@ -464,7 +488,15 @@ std::pair<ReplicationHandler::MainResT, ReplicationHandler::ReplicasResT> Replic
...
-            old_value_it->second = num_main_committed_txns - old_value_it->second;
+            auto lag = static_cast<int64_t>(num_main_committed_txns) - old_value_it->second;
+            // An ASYNC replica records its progress on a worker thread with no happens-before against main's
+            // counter bump, so it can transiently appear one txn ahead of main. Floor that artifact to 0; a
+            // SYNC replica bumps after main in program order, so a negative there is a real invariant violation
+            // and is left visible.
+            if (repl_storage_client->Mode() == replication_coordination_glue::ReplicationMode::ASYNC && lag < 0) {
+              lag = 0;
+            }
+            old_value_it->second = lag;
```
andrejtonev: GetNumCommittedTxns (replication_handler.cpp:499) overwrites the map value with lag, contradicting the field name/comment in instance_state.hpp:32 (behavior correct; naming misleading).

andrejtonev (approved):

src/storage/v2/replication/replication_client.cpp
as51340: makes sense

PR #4316, opened 2026-06-29
fix: enforce fine-grained label permissions in text and vector search

### LINKED ISSUE #4296: Text search bypasses fine-grained RBAC label permissions
Text search (`text_search.search`, `text_search.search_all`, and presumably `*_edges`/`aggregate`) bypasses fine-grained RBAC label permissions. A user with no read access to a label still gets full node contents back through the text index.

Verified on v3.10.0: a user granted only `READ ON NODES CONTAINING LABELS :Public` gets `count 0` from `MATCH (n:Document)`, but `CALL text_search.search('doc_text_index', ...)` returns the `:Document` nodes including their properties.

Root cause: `mgp_graph_search_text_index` in `src/query/procedure/mg_procedure_impl.cpp` calls `TextIndexSearch` then `WrapTextSearch` with no `auth_checker->Has(..., FineGrainedPrivilege::READ)` filtering — unlike normal vertex iteration (`NextPermitted`), which does filter. Same gap exists in vector search.

### LINKED ISSUE #4297: Vector search bypasses fine-grained RBAC label permissions
Vector search (`vector_search.search`, `vector_search.search_edges`) bypasses fine-grained RBAC label permissions. A user with no read access to a label still gets full node contents back through the vector index.

Verified on v3.10.0: a user granted only `READ ON NODES CONTAINING LABELS :Public` gets nothing from `MATCH (n:Document)`, but `CALL vector_search.search('doc_index', ...)` returns the `:Document` nodes including their properties — identical results to an admin.

Root cause: `mgp_graph_search_vector_index` in `src/query/procedure/mg_procedure_impl.cpp` calls `VectorIndexSearchOnNodes` then `WrapVectorSearchResults` with no `auth_checker->Has(..., FineGrainedPrivilege::READ)` filtering — unlike normal vertex iteration (`NextPermitted`), which does filter. Same gap exists in text search.

DavIvek:
Closes #4296.
Closes #4297.

## What

Enforces fine-grained label/property RBAC in `text_search.*` and `vector_search.*` so search results match what normal vertex/edge iteration would return, closing a bypass where a hit's membership + BM25/similarity score leaked content the caller can't read.

### Per-hit row filter
Every hit is filtered by `auth_checker` against the entity's **real** labels/type (not the index's declared label): the entity must be READ-permitted (both endpoints, for edges) and the properties that could have produced the match must be readable on those labels. This handles the cases the old code leaked — a `DENY` on a co-label of a multi-label node, and NEUTRAL/ungranted properties under deny-by-default (the common case once any fine-grained rule is attached).

### Mode-aware property gating
- `search` (specified properties): gates on the propert(ies) **named in the query** (`data.<prop>:…`, including boolean `AND`/`OR`/`NOT` expressions) — a hit surfaces when the queried field is readable, regardless of other indexed properties.
- `search_all` / `regex` (blob): gates on the index's full property set (any indexed property could have matched).
- In both cases only properties the entity **actually has** are checked — a declared/queried property the entity lacks can't have produced the match, so it is skipped rather than dropping the hit.

### Aggregates
`text_search.aggregate{,_edges}` runs inside Tantivy with no per-row hook, so it can't honor per-row perms: it returns an `error_msg` unless the caller can read every indexed property. Multi-label label-dimension over-counting is a documented, accepted limitation.

## Storage fix 
A second vector **edge** index on the same property couldn't back-fill edges already indexed by another vector-edge index — the second index silently indexed nothing (e.g. a WILDCARD edge index created over an existing graph came up empty). The property stores only the index ids, not the vector, so the raw read yielded an empty vector. Fixed by recovering the vector from the first index's uSearch during backfill, mirroring the node path. Node indexes and the on-insert path were unaffected.


src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -4083,6 +4084,166 @@ mgp_edge *GetEdgeByGid(mgp_graph *graph, memgraph::storage::Gid edge_gid, memgra
...
 
+#ifdef MG_ENTERPRISE
+namespace {
+// A search hit is visible iff the entity is label/type-readable AND every candidate property it actually
+// has is READ-permitted on its real labels. `searched` holds the candidates: the fields named in a
+// specified-property query, or the index's full property set for blob (search_all/regex) modes; empty ⇒
+// property-less index, where every string property present is a candidate. A declared property the
+// entity lacks is skipped — it could not have produced the match.
+bool VertexSearchHitReadable(const memgraph::query::VertexAccessor &v, const mgp_graph &graph,
+                             std::span<const memgraph::storage::PropertyId> searched) {
```
colinbarry, line 4113: Optimisation opportunity here: when `searched` in non-empty, instead of reading the full `Properties()` map, you could just check the given property names.

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -4083,6 +4084,166 @@ mgp_edge *GetEdgeByGid(mgp_graph *graph, memgraph::storage::Gid edge_gid, memgra
...
+  if (!searched.empty())
+    return std::ranges::all_of(searched,
+                               [&](const auto property) { return !props->contains(property) || readable(property); });
+  return std::ranges::all_of(*props, [&](const auto &kv) {
+    return kv.second.type() != memgraph::storage::PropertyValue::Type::String || readable(kv.first);
+  });
+}
+
+// the index's indexed properties — the set a hit could have matched on (empty ⇒ property-less text index)
+std::vector<memgraph::storage::PropertyId> TextIndexSearchedProperties(const mgp_graph &graph,
```
colinbarry, line 4142: `vector` return type, but we only ever return 0 or 1 `PropertyId`. Maybe an `optional` would be better, as we could avoid the `vector` allocation.

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -4083,6 +4084,166 @@ mgp_edge *GetEdgeByGid(mgp_graph *graph, memgraph::storage::Gid edge_gid, memgra
...
+                                                                       std::string_view index_name) {
+  const auto indices = graph.getImpl()->ListAllIndices();
+  const auto it = std::ranges::find_if(indices.vector_indices_spec, [&](const memgraph::storage::VectorIndexSpec &s) {
+    return s.index_name == index_name;
+  });
+  if (it == indices.vector_indices_spec.end()) return {};
+  return {it->property};
+}
+
+std::vector<memgraph::storage::PropertyId> VectorEdgeIndexSearchedProperty(const mgp_graph &graph,
```
colinbarry: Same note as above re: `optional` vs `vector` for returning 0 or 1 `PropertyId`.

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -4083,6 +4084,166 @@ mgp_edge *GetEdgeByGid(mgp_graph *graph, memgraph::storage::Gid edge_gid, memgra
...
+// specified-property query, or the index's full property set for blob (search_all/regex) modes; empty ⇒
+// property-less index, where every string property present is a candidate. A declared property the
+// entity lacks is skipped — it could not have produced the match.
+bool VertexSearchHitReadable(const memgraph::query::VertexAccessor &v, const mgp_graph &graph,
+                             std::span<const memgraph::storage::PropertyId> searched) {
+  const auto *ctx = graph.ctx;
+  if (!ctx || !ctx->auth_checker) return true;
+  const auto *auth_checker = ctx->auth_checker;
+  if (!auth_checker->Has(v, graph.view, memgraph::query::AuthQuery::FineGrainedPrivilege::READ)) return false;
+  if (!auth_checker->HasPropertyRestrictions()) return true;
```
colinbarry, line 4118: This inner logic (lines 4100-4112) is identical to the same in `EdgeSearchHitReadable:4126-4137`, other than accessor types. I would consider extracting to a reusable template, if you feel removing the 15-line duplication is worth it?

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -4083,6 +4084,166 @@ mgp_edge *GetEdgeByGid(mgp_graph *graph, memgraph::storage::Gid edge_gid, memgra
...
+  const auto props = v.Properties(graph.view);
+  if (!props) return false;
+  const auto readable = [&](const memgraph::storage::PropertyId property) {
+    return memgraph::query::PropertyReadAllowed(auth_checker, v, graph.view, property);
+  };
+  if (!searched.empty())
+    return std::ranges::all_of(searched,
+                               [&](const auto property) { return !props->contains(property) || readable(property); });
```
colinbarry: I think there are probably enough uses of `std::ranges::` to replace with a `namespace r = std::ranges` alias

colinbarry (approved):
Approved, with a few notes about some potential code hygiene changes and performance optimisations you could consider. 🚀

DavIvek:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    -  Enforces fine-grained RBAC for text_search.* and vector_search.*, ensuring search results respect label and property permissions and no longer expose data users aren’t allowed to read. Additionally, fixes an issue where creating a second vector edge index on the same property failed to backfill existing data, ensuring consistent and correct indexing behavior. [#4316](https://github.com/memgraph/memgraph/pull/4316)

PR #4317, opened 2026-06-29
fix(rpc): correct file-bearing RPCs reads

Ignition:
A file-bearing RPC is a request-args segment, one or more file segments, and a footer. RpcMessageDeliverer::Execute() re-parses the receive buffer from the start on every read and advances it only by clearing, while OpenFile() has a side effect: it registers a received file. So when a file was fully received but a following segment (such as the footer) was split across reads, the next call re-parsed and re-registered it, and a single-file transfer arrived as two. That trips the receiver's "exactly one file" check and aborts the replica during current-WAL recovery.

Drop the consumed prefix from the buffer before waiting for more data so a completed file is not re-parsed, and treat an engaged file handler as "message in progress" so a footer arriving alone is the message end, not an empty stream. Two adjacent splits are fixed too: the file-metadata look-ahead now covers the first file (not just file-to-file transitions), so a mask split from its metadata waits for more data; and the request header is captured at its fixed length (the args segment plus the file mask, which the reader accepts as the final segment) rather than from how much has arrived, which truncated it when the args were longer than a file's metadata.

Add unit tests that split a built wire at every structural boundary and assert the parse is identical however it is fragmented (file count, byte-for-byte contents, one delivery) across single and multi file, empty, sub- and multi-segment files, and args longer than the metadata, plus two messages on one connection.

### DIFF
--- a/src/rpc/protocol.cpp
+++ b/src/rpc/protocol.cpp
@@ -58,10 +58,13 @@ void RpcMessageDeliverer::Execute() {
   while (true) {
     auto const remaining_file_size = GetRemainingFileSize();
 
-    slk::StreamInfo const ret = slk::CheckStreamStatus(input_stream_->data() + consumed_bytes,
-                                                       input_stream_->size() - consumed_bytes,
-                                                       remaining_file_size,
-                                                       consumed_bytes);
+    // Non-zero marks the message in progress, so a lone footer ends it instead of
+    // parsing as an empty (INVALID) stream.
+    slk::StreamInfo const ret =
+        slk::CheckStreamStatus(input_stream_->data() + consumed_bytes,
+                               input_stream_->size() - consumed_bytes,
+                               remaining_file_size,
+                               consumed_bytes + (file_replication_handler_.has_value() ? 1U : 0U));
 
     if (ret.status == slk::StreamStatus::INVALID) {
       input_stream_->Clear();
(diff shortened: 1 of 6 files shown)


Ignition:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Fixed file-bearing replication RPCs being mis-framed across network reads, which could abort a replica during current-WAL recovery. [#4317](https://github.com/memgraph/memgraph/pull/4317)

as51340 (approved):

as51340 (changes requested):
```
1. tests/unit/slk_streams.cpp:467 — WholeFileInSegment is now ~1/16384 flaky.
The test builds [mask][5 random bytes][footer] via raw SaveFileBuffer (no real file metadata) and asserts NEW_FILE. With the if (remaining_file_size) guard removed, the first-file look-ahead now reads bytes pos+1..pos+8 (4 of the random file bytes + 4 footer zero-bytes) as a filename length. Whenever that little-endian uint32 is ≤ kSegmentMaxDataSize (262144) — probability ≈ 262145/2³² ≈ 1/16384 per run — the call returns PARTIAL, so ASSERT_EQ(res.status, NEW_FILE) fails. The PR does not update this test. (On a feat/flake_investigation branch, shipping a new ~1/16k flake is worth catching.)

2. tests/unit/slk_streams.cpp:645 — FirstFileMaskNotAffectedByCheck documents the opposite of the new contract (and is flaky for the same reason).
This test's comment states "The metadata check should NOT apply when remaining_file_size is not set (first file case)" and asserts NEW_FILE on the same raw [mask][5 random][footer] buffer. That precondition is exactly what this PR deliberately reverses — the look-ahead now does apply to the first file. The test still passes only by luck (random str_len usually exceeds kSegmentMaxDataSize) and shares the ~1/16384 flake. It should be updated to feed valid metadata after the mask (asserting NEW_FILE), or rewritten to assert the new PARTIAL-on-split behavior — otherwise it encodes a contract the code no longer honors.

```

as51340 (approved):

PR #4322, opened 2026-07-01
fix(auth): guard FGAC checker build in SessionHL::Pull on valid license

andrejtonev:
SessionHL::Pull built a fine-grained auth checker whenever an authenticated user was present, but -- unlike the three call sites in interpreter.cpp -- did not first check that the enterprise license is valid. AuthChecker::GetFineGrainedAuthChecker returns a null unique_ptr the moment the license is invalid, so on an enterprise binary running without a valid license the following `auth_checker->NeedsFineGrainedAuthChecker()` dereferenced null and crashed the server (the DMG_ASSERT guarding it is compiled out in release builds).


andrejtonev:
Delivered via a different bugger auth pr

PR #4318, opened 2026-06-30
feat: add e-class analysis and constant folding to plan_v2

imilinovic:
Adds e-class analysis to plan_v2 — per-e-class facts seeded on insert and combined on e-class union — and uses it for plan-time constant folding during equality saturation. First of a stacked series; everything here sits behind the experimental `PLANNER_V2` flag, so there is no user-visible behavior change.

**What.** Every e-class now carries `ExpressionAnalysis` facts (`known_constant_value`, `known_list_length`). A fold rewrite evaluates an operator whose operands are all known constants and merges the result back in, so constant subexpressions collapse during saturation (e.g. `1 + 1` → `2`, cascading through nesting).

**How.**
- `resolve/analysis.{hpp,cpp}`: a `variant<ExpressionAnalysis, OperatorAnalysis, SymbolAnalysis>` whose arm is the e-class kind. Facts are seeded on insert and combined on e-class union with an agree-or-bug rule — contradictory constants throw `PlannerBug`. Kind partition is enforced at compile time (`symbol.hpp`). (`known_list_length` is seeded here but not read in this PR; a later stack PR consumes it — the fold consumes `known_constant_value`.)
- `resolve/constant_identity.hpp`: a strict type-then-structure identity (`1` ≠ `1.0`, `NaN` ≡ `NaN`, deep on lists/maps) used both for analysis merges and for keying the literal interning store — so interning never coerces `Int` and `Double` together.
- `rewrite/fold.{cpp,hpp}` + `rewrite/rewrites.cpp`: one fold rule per operator, generated from the `op_ast_lists.hpp` registry so evaluation and rule generation can't drift. `FoldConstant` reuses the runtime `TypedValue` operators, so NULL three-valued logic, divide-by-zero, and type coercion match runtime by construction; non-scalar or error results decline.
- Engine prerequisites in `src/planner/`: the rewrite engine is unified on a single `RewritableGraph` concept (rules can read analysis and mint typed nodes via `ctx.Make<S>`), and the core e-graph gains an `analysis_of` accessor. Literal interning moved to a node-stable hash map so the id→value reverse index stays valid.

**Why.** Constant folding is a self-contained plan-time optimization: constant subexpressions collapse so the plan carries the value, not the computation (`RETURN 1 + 1` plans as `2`), all behind the experimental flag. As the bottom of the planner-v2 stack it also establishes the e-class analysis substrate later work reuses — but it stands alone; nothing here needs a later PR to be correct or useful.

PR #4324, opened 2026-07-02
feat: make plan_v2 equality saturation incremental

imilinovic:
Makes plan_v2's equality saturation **incremental**: each pass arms only the rules that could newly fire, instead of re-running every rule over every e-class. The two schedules are `ArmingMode::Full` (every rule every pass — the reference behaviour and differential oracle) and `ArmingMode::Incremental` (the optimisation).

**How**
- The e-graph records a per-pass *touched* e-class set — every e-class inserted or merged, including cascade merges during rebuild (`egraph.cppm`).
- A symbol→rules *arming index* selects candidate rules; variable-rooted rules stay always-armed (`arming_index.hpp`, `rule.hpp`).
- The *active set* is the touched e-classes closed under parents up to the deepest pattern binding depth — the bound that keeps incremental arming sound (`active_set.hpp`).
- For a single-pattern rule whose only symbol node is its root, the pattern VM additionally restricts the root candidates to the active set (`executor.hpp`). Rules that don't qualify — multi-pattern, or any pattern with a symbol below the root (whose VM entry is the deepest symbol, not the root) — fall back to symbol-granularity arming and match every candidate (`RewriteRule::supports_active_root_restriction`, `rewriter.hpp`). Re-saturation arms only what changed.

**Why**: a steady-state no-op pass previously re-ran every rule over the whole graph, growing worse than linearly with graph size. Incremental arming keeps long saturations sparse (added steady-state benchmark), while a differential fuzzer (driving an inserting rule against a partition oracle) plus unit tests assert that `Full` and `Incremental` saturation reach the identical fixpoint — including a deep-entry regression for the root-entry precondition above, a mid-session `set_rules` re-arm test, and an always-armed rule driven end to end.

**Equivalence note**: `Incremental` reaches the same fixpoint as `Full` *given enough iterations*. Because it derives a pass's active set from the previous pass's merges, a deep multi-rule cascade can need a few more passes; under a truncated `max_iterations` it may stop less-saturated — a valid, never incorrect, plan. The first pass is always full, so a fresh single saturation (the production path) tracks `Full` pass-for-pass until the first within-pass re-enable.

Stacked on #4318 — **base branch is `feat/plan-v2-eclass-analysis`, not `master`.** planner_v2 stays experimental and off by default, so there is no effect on the default planner.


imilinovic:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist

PR #4334, opened 2026-07-03
fix(query): bound AST cache with LRU and fold signed list literals

DavIvek:
## Why

`InterpreterContext::ast_cache` was an unbounded skip list keyed on a bare 64-bit hash. Query stripping replaces literal values with placeholders, but every *distinct stripped text* permanently retained its own `AstStorage` — a stream of structurally-varying queries (e.g. inlined embedding vectors, whose sign patterns alone produce a distinct stripped text per query) grew process memory without limit. The reported leak.

## How

- Replace the unbounded skip list with an LRU sized by `--query-ast-cache-max-size` (default 1000, 0 disables the cache). Entries are `shared_ptr<const CachedQuery>`: the lock is released before the (potentially heavy) AST clone, a concurrent eviction can't free an entry mid-clone, and shared entries are compiler-enforced immutable. A parse race on the same uncached query converges on one entry (the loser adopts the winner's).
- Key the cache on `HashedString` instead of the raw 64-bit hash: a bucket collision is resolved by comparing text, so a colliding query is a real hit or a fresh parse, never a wrong AST.
- Fold a unary sign into the numeric literal it precedes when the previous non-space token is `[`, or `,` at bracket depth > 0. In those positions a `-`/`+` cannot be a subtraction — nothing it could subtract from precedes it — so folding needs no heuristics and cannot misfire. Sign variants of a list (`[-1, 2]`, `[1, -2]`, `[1, 2]`) then strip to the same text and hit both the AST and plan caches. Binary minus is untouched (`[1-2, 3]`, `[a-1]`), as are scalars, maps, and parens (`RETURN -1`, `{k: -1}`, `abs(-1)`).
- Make `Parameters::AtTokenPosition` O(1) by keying parameters on their token position — list-literal evaluation looked up one parameter per element per row, so an N-element list literal did O(N²) work per row.
- Validate coordinator ids where they are evaluated: `EvaluateCoordinatorId` goes through `EvaluateUint` and rejects values above `INT32_MAX` instead of silently truncating at the `static_cast` call sites.

## Known trade-offs

Sign variants of scalars/maps (`RETURN -1` vs `RETURN 1`) still take separate entries — 2 entries per shape, LRU-bounded, unlike the 2^N list case this PR targets.

### DIFF
--- a/tests/unit/stripped.cpp
+++ b/tests/unit/stripped.cpp
@@ -532,6 +532,7 @@ TEST(QueryStripper, SignFoldInList) {
   {
     StrippedQuery stripped("RETURN [-1e-3]");
     EXPECT_FLOAT_EQ(SingleLiteral(stripped).ValueDouble(), -1e-3);
+    EXPECT_EQ(stripped.stripped_query().str(), "RETURN [ " + kStrippedDoubleToken + " ]");
   }
   {
     StrippedQuery stripped("RETURN [+42]");
--- a/tests/unit/stripped.cpp
+++ b/tests/unit/stripped.cpp
@@ -593,4 +593,37 @@ TEST(QueryStripper, BinaryMinusInListIsNotFolded) {
   }
 }
 
+TEST(Parameters, LookupByTokenPosition) {
+  Parameters params;
+  params.Add(3, memgraph::storage::ExternalPropertyValue(int64_t{30}));
+  params.Add(7, memgraph::storage::ExternalPropertyValue(int64_t{70}));
+  params.Add(5, memgraph::storage::ExternalPropertyValue(int64_t{50}));
+
+  EXPECT_EQ(params.size(), 3);
+  // Lookup is by token position, independent of insertion order.
+  EXPECT_EQ(params.AtTokenPosition(3).ValueInt(), 30);
+  EXPECT_EQ(params.AtTokenPosition(5).ValueInt(), 50);
+  EXPECT_EQ(params.AtTokenPosition(7).ValueInt(), 70);
+  // At() still addresses storage in insertion order.
+  EXPECT_EQ(params.At(0).first, 3);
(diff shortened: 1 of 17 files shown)


Ignition:
Related #4372

tests/unit/stripped.cpp
```diff
@@ -510,4 +518,166 @@ TEST(QueryStripper, KeywordsCanBeUsedInStrippedQueries) {
...
+    EXPECT_EQ(stripped.stripped_query().str(), "RETURN [ " + kStrippedIntToken + " ]");
+  }
+  {
+    StrippedQuery stripped("RETURN [+0.5]");
+    EXPECT_FLOAT_EQ(SingleLiteral(stripped).ValueDouble(), 0.5);
+    EXPECT_EQ(stripped.stripped_query().str(), "RETURN [ " + kStrippedDoubleToken + " ]");
+  }
+  {
+    StrippedQuery stripped("RETURN [-1e-3]");
+    EXPECT_FLOAT_EQ(SingleLiteral(stripped).ValueDouble(), -1e-3);
```
colinbarry, line 534: I notice that this test doesn't check for `kStrippedDoubleToken`, whilst the other 3 cases do.

colinbarry (approved):
Approved, with a tiny comment about an inconsistency in one unit test. 💪

DavIvek:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    -  Fixes a memory growth issue in the query AST cache by introducing a bounded LRU cache and improving cache key handling, preventing unbounded memory usage from structurally unique queries. This also improves performance through better cache hits and optimizations in parameter lookup, with no required user changes (configurable via `--query-ast-cache-max-size`). [#4334](https://github.com/memgraph/memgraph/pull/4334)

PR #4336, opened 2026-07-03
fix(plan): correctness fixes for edge IN-list, ORDER BY elimination, and nested edge property filters

Ignition:
Six silent wrong-results bugs in the planner's index-lookup and
ORDER BY-elimination rewrites, each a case the vertex path got right but the
edge or composite-index path did not. Fixed test-first (in-memory + on-disk).

- `id(e) IN [...]` on edges returned **zero rows**, now lowered to
  `Unwind -> ScanAllByEdgeId`.
  `MATCH ()-[e]->() WHERE id(e) IN [0,1,2] RETURN e`
- Edge `e.prop IN [...]` **duplicated rows** for repeated values, now deduped.
  `MATCH ()-[e:R]->() WHERE e.prop IN [1,1] RETURN e`
- `ORDER BY` **silently dropped** over an Unwind-fed value scan (edge + vertex).
  `... WHERE e.prop IN [3,1,2] RETURN e.prop ORDER BY e.prop`
- Composite-index nullable suffix **misordered NULL** (wrong rows under LIMIT).
  `MATCH (n:L) WHERE n.a = 5 RETURN n.b ORDER BY n.b LIMIT 2`
- Nested edge property `e.a.b = 5` **misread as `e.a = 5`**, wrong rows.
  `MATCH ()-[e:T]->() WHERE e.a.b = 5 RETURN e`

### DIFF
--- a/src/query/plan/rewrite/order_by_elimination.hpp
+++ b/src/query/plan/rewrite/order_by_elimination.hpp
@@ -287,6 +287,13 @@ class OrderByEliminator {
             size_t ob_ptr = 0;
             for (size_t i = 0; i < s->properties_.size() && ob_ptr < group_size; ++i) {
               if (s->properties_[i] == entries[group_start + ob_ptr].resolved) {
+                // A composite index stores a missing property as NULL and sorts
+                // NULL first, but ORDER BY places NULL last, so the index order
+                // only matches ORDER BY when the sort column cannot be NULL. A
+                // column is guaranteed non-null when it carries a filter (every
+                // filter type -- EQUAL/RANGE/REGEX/IN/IS_NOT_NULL -- excludes
+                // NULL); an unconstrained column (no range entry) may be NULL.
+                if (i >= s->expression_ranges_.size()) return false;
                 ++ob_ptr;
               } else if (i < s->expression_ranges_.size() &&
                          s->expression_ranges_[i].type_ == PropertyFilter::Type::EQUAL) {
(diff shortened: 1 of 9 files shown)


Ignition:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Fixed several planner correctness bugs where edge-index and composite-index queries (edge IN-lists, id(e) IN, nested edge properties, ORDER BY on a composite-index suffix) could silently return wrong or misordered rows; affected queries now return correct results after upgrading. [#4336](https://github.com/memgraph/memgraph/pull/4336)

src/query/plan/rewrite/order_by_elimination.hpp
```diff
@@ -287,6 +287,13 @@ class OrderByEliminator {
             size_t ob_ptr = 0;
             for (size_t i = 0; i < s->properties_.size() && ob_ptr < group_size; ++i) {
               if (s->properties_[i] == entries[group_start + ob_ptr].resolved) {
+                // A composite index stores a missing property as NULL and sorts
+                // NULL first, but ORDER BY places NULL last, so the index order
+                // only matches ORDER BY when the sort column cannot be NULL. A
+                // column is guaranteed non-null when it carries a filter (every
+                // filter type -- EQUAL/RANGE/REGEX/IN/IS_NOT_NULL -- excludes
+                // NULL); an unconstrained column (no range entry) may be NULL.
+                if (i >= s->expression_ranges_.size()) return false;
```
colinbarry, line 296: Think we could refine to:

```cpp
if (s-index_order_ == ASC && i >= s->expression_ranges_.size()) return false;
```

For DESC index and ORDER BY DESC, the ordering of `NULL` types is the same in both `PropertyValue` ordering and `ORDER BY` expectations.

colinbarry (approved):
Approved (with one minor comment about a potential further optimisation)

src/query/plan/rewrite/order_by_elimination.hpp
Ignition: I tested it and the DESC case is actually inverted the same way as ASC, so restricting to ASC would re-introduce the misordering for DESC indexes.

imilinovic (approved):

PR #4314, opened 2026-06-29
fix: Abort replication on shutdown

as51340:
Shutdown main instance even in the case of long ongoing snapshot/wal replication by aborting the replication client.
Throw SessionException if replica cannot send response to main.
If `InProgressMsg` cannot be delivered to main, we catch and process the exception. The next such message can be sent normally on the clean state.

PR #4347, opened 2026-07-06
fix: Sending final response msg failure handling

as51340:
Regardless of whether the response is successfully sent for SnapshotRpc, WalFilesRpc and CurrentWalRpc, files should be moved to .old directory. The issue exists since a long time but wasn't seen until changes in https://github.com/memgraph/memgraph/pull/4314.

### DIFF
--- a/src/dbms/inmemory/replication_handlers.cpp
+++ b/src/dbms/inmemory/replication_handlers.cpp
@@ -388,7 +388,7 @@ void InMemoryReplicationHandlers::HeartbeatHandler(dbms::DbmsHandler *dbms_handl
 
   const storage::replication::HeartbeatRes res{
       true, commit_info.ldt_, last_epoch_with_commit, commit_info.num_committed_txns_};
-  rpc::SendFinalResponse(res, request_version, res_builder, fmt::format("db: {}", storage->name()));
+  rpc::SendFinalResponse(res, request_version, res_builder, storage->name());
 }
 
 void InMemoryReplicationHandlers::PrepareCommitHandler(
@@ -475,7 +475,7 @@ void InMemoryReplicationHandlers::PrepareCommitHandler(
     }
 
     const storage::replication::PrepareCommitRes res{false};
-    rpc::SendFinalResponse(res, request_version, res_builder, fmt::format("db: {}", storage->name()));
+    rpc::SendFinalResponse(res, request_version, res_builder, storage->name());
     return;
   }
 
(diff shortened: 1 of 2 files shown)


as51340:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Replica will move its durability files to the backup directory or delete them regardless of whether response is successfully sent back to main. This occurred when SnapshotRpc, WalFilesRpc or CurrentWalRpc were received for resetting the replica. [#4347](https://github.com/memgraph/memgraph/pull/4347)

src/dbms/inmemory/replication_handlers.cpp
andrejtonev, line 890: Not introduced here (master does the same), but worth a follow-up: this early return after a reset + Clear() leaves the old WAL chain on disk — the same orphaned-chain  situation the guard was added to prevent. Declaring the cleanup guard before the load  loop would also cover the partial-load failure.

src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -726,37 +727,42 @@ void InMemoryReplicationHandlers::SnapshotHandler(rpc::FileReplicationHandler co
...
+        return snapshot_path != dst_snapshot_file;
+      };
+      auto snapshots_to_process = curr_snapshot_files | rv::filter(not_recovery_snapshot) | r::to_vector;
+
+      if (FLAGS_storage_backup_dir_enabled) {
+        auto const maybe_backup_dirs = CreateBackupDirectories(current_snapshot_dir, current_wal_directory);
+        if (!maybe_backup_dirs) {
+          spdlog::error("Couldn't create backup directories. Old durable files won't be moved to .old directory.");
+          return;
+        }
```
andrejtonev, line 746: Behavior change vs master here is a net improvement (removes the double-send), but this branch still leaves stale files while reporting success, which is the exact  orphaned-chain case the commit warns about. Not blocking (pre-existing), but consider  falling back to deletion on backup-dir failure, or at least a comment stating the  leave-behind is intentional.

src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -726,37 +727,42 @@ void InMemoryReplicationHandlers::SnapshotHandler(rpc::FileReplicationHandler co
...
-        snapshots_to_process, backup_snapshot_dir, curr_wal_files, backup_wal_dir, &(storage->file_retainer_));
-  } else {
-    DeleteFiles(snapshots_to_process, &storage->file_retainer_);
-    DeleteFiles(curr_wal_files, &storage->file_retainer_);
-  }
+  }};
 
-  spdlog::debug("Replication recovery from snapshot finished!");
+  const storage::replication::SnapshotRes res{ldt, num_committed_txns};
+  rpc::SendFinalResponse(res, request_version, res_builder, fmt::format("db: {}", storage->name()));
```
andrejtonev: Minor: this format string allocates unconditionally even though it's only used in a trace log. Passing db_name as a string_view and moving the "db: " prefix into the trace  format would avoid the per-RPC allocation. Pre-existing pattern, optional. Other places with same issue

src/dbms/inmemory/replication_handlers.cpp
as51340: There is already a log.
```
"Couldn't create backup directories. Old durable files won't be moved to .old directory."
```

andrejtonev (approved):

PR #4348, opened 2026-07-06
refactor: build and package MAGE from the unified CMake tree

mattkjames7:
Changes to the way that MAGE query modules are built and installed:
- MAGE is now built using `conan` in the same way that Memgraph is.
- MAGE can also be built directly using `build.sh` (using `--mage on` or `--mage only`).
- The old Python script `mage/setup` has been removed.
- MAGE code has relocated to `src/mage`.
- Testing code has moved from `mage/tests` to `tests/mage` and `tests/smoke`.
- Changed the packaging method of MAGE to use `CPack`, in the same way as Memgraph, removing the shell scripts under `tools/ci/mage-build/package` that were used to create DEB and RPM packages.
- Packaging code and Dockerfiles for MAGE now exist in `release/package/mage`.
- Added a script to simplify packaging DEBs and RPMs of Memgraph/MAGE locally: `package.sh`.
- We now bundle `libomp.so` with MAGE.
- Some small changes to C++ modules were required to allow MAGE to build using `clang++` from the toolchain (it used GCC before).
- An unfortunate consequence of moving many of the source files meant that they all had there formatting re-checked by the pre-commit hooks, meaning that a number of files have been reformatted.


mattkjames7:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Refactored the MAGE build and packaging logic to unify with Memgraph. Only affects those building MAGE from source. [#4348](https://github.com/memgraph/memgraph/pull/4348)
- [x] **https://github.com/memgraph/documentation/pull/1700**

imilinovic (commented):
## Review — comments only (not requesting changes, not approving)

Nice refactor — the mechanical move is clean: I verified stale-path integrity across the whole
tree, the Conan `mage_only` graph trim, the three build modes, the module-count scanner, the
`mage__` ctest prefix, and the deb/rpm packaging all hold up. Leaving these as **comments** rather
than blocking — but two of them (the first two) look like real breakages worth a look before merge.

---

### 🔴 clang-tidy CI ends up configured MAGE-only — memgraph source is no longer linted

`.github/workflows/diff_coverage.yaml:299` (and `:399` for the enterprise job)

The clang-tidy jobs now run two configure steps into the *same* `build/`:
1. `build-memgraph --cmake-only --community` → configures memgraph, generates its `compile_commands.json` + `generated_code`.
2. `build-memgraph --mage only --cmake-only` → reconfigures.

But `build_memgraph()` wipes the tree on every call — `rm -rf $MGBUILD_ROOT_DIR` (`release/package/mgbuild.sh:690`) and `rm -rf build/*` (`:739`) — and skips `generated_code` for mage-only (`:925`). So step 2 clobbers step 1 and leaves `build/` configured `MG_BUILD_MEMGRAPH=OFF` with **only `src/mage` translation units** in the compile DB. `tests/code_analysis/clang_tidy.sh:125` then lints `git diff -- src` (memgraph *and* src/mage) against that DB, so changed memgraph files aren't in it and get silently skipped — the gate goes green while linting nothing on ordinary PRs.

Suggestion: a single unified configure — `build-memgraph --mage on --cmake-only` (both `MG_BUILD_MEMGRAPH=ON` and `MG_BUILD_MAGE=ON` in one tree) — instead of two clobbering steps.

---

### 🔴 `make-dev-container.sh` clones into the wrong directory

`release/package/mage/make-dev-container.sh:61-62`

```sh
git clone https://github.com/memgraph/memgraph.git   # → /root/memgraph
cd /root/mage
```

The clone URL changed from `mage.git` to `memgraph.git` but the following `cd /root/mage` wasn't updated, so it now `cd`s into a directory that doesn't exist. There's no `set -e`, so it continues silently and every later step (`cp -r /root/mage/. /mage/`, `install_deps.sh`, `./build.sh --mage on`) runs against an empty path. This script is `COPY`'d into `Dockerfile.cugraph:123` and `Dockerfile.release:141` as the documented prod→dev-container flow.

Suggestion: `cd /root/memgraph` (and update the later `/root/mage` references), or clone into `/root/mage` explicitly.

---

### 🟡 Dead artifact-upload step

`.github/workflows/reusable_package_mage.yaml:1129`

"Push query module archive to GitHub" still uploads `path: "mage/mage.tar.gz"`, but this PR deleted the only producer (`compress-query-modules.sh`) — and the sibling comment at `:547` even says "the old mage.tar.gz hand-off is gone." `upload-artifact@v7` defaults to `if-no-files-found: warn`, so the job won't fail; it just publishes an empty/absent `<name>-query-modules` artifact (nothing downloads it today). Consider removing the step or repointing it at `output/memgraph-mage*.{deb,rpm}`.

---

### 🟡 Stale old-path references in comments

Non-executing, but they point at pre-move paths:
- `tools/ci/mage-build/offline-installer/download-debs.sh:16` — `mage/install_runtime_requirements.sh` (note its sibling `:17` *was* updated in this PR)
- `tools/ci/mage-build/offline-installer/download-wheels.sh:6,59,84`
- `tools/ci/mage-build/offline-installer/install.sh:109`
- `tools/ci/build-gssapi.sh:16`

All reference `mage/install_*_requirements.sh`, now under `release/package/mage/`.

---

### 🟡 Comment / flag drift

- `conanfile.py:27` and `CMakeLists.txt:219` say `./build.sh --mage-only`; the real flag is `--mage only` (a space-separated MODE arg).
- `release/package/mage/Dockerfile.rpm:18` references `build-rpm.sh` (deleted).
- `tools/ci/mage-build/build-docker-image.sh:6` references the removed `build-mage` subcommand.

---

### 🔵 FYI — mode-switch cache staleness (already documented)

`--mage on`/`off` never set `MG_BUILD_MEMGRAPH` (conanfile.py `:211-213` sets it only under `mage_only`; `build.sh:204-205` only adds `-DMG_BUILD_MAGE=ON`). So reusing a `build/` across `--mage only` → `--mage on`/`off` with `--keep-build` leaves a stale `MG_BUILD_MEMGRAPH=OFF` that nothing re-asserts (`option()` won't overwrite an existing cache entry), silently skipping the memgraph build. The reverse (`on`→`only`) is safe. Already called out in `build.sh:27-29` ("use a fresh build directory when switching to/from full builds") — just flagging that it's enforced only by that note. Not asking for a change.

---

### ⚪ Out of scope (pre-existing, not this PR)

- `.clang-tidy:41` — `HeaderFilterRegex` still lists a bare `mage/cpp` alternative; harmless (the `src` alternative already covers `src/mage/cpp`), and the file is untouched by this PR.
- `.github/workflows/reusable_package_mage.yaml:1139` — the "Push RPM package" `if:` is missing parens around the OS `||` group, so it binds as `(push && MAGE && fedora) || centos || rocky` and fires on any centos/rocky build. This is byte-identical to `origin/master:1090` (unchanged by this PR); worth a separate stacked fix, not this one.

imilinovic (approved):

PR #4353, opened 2026-07-07
feat: Global vertex-property indices

colinbarry:
Implement indices on global vertex properties, i.e., properties across all vertices. Enables index-backed lookups like `MATCH (n) WHERE n.uid = '123' RETURN n` when matching vertices have no labels in common.

Syntax:
```cypher
CREATE GLOBAL INDEX ON :(prop);
DROP GLOBAL INDEX ON :(prop);
```

Changes:
- Storage: `InMemoryVertexPropertyIndex` (skip-list backed), GC, abort processing, `DiskVertexPropertyIndex` stub
- Durability: snapshot/WAL encoding, recovery (v37 via #4408)
- Grammar + interpreter: `CREATE/DROP GLOBAL INDEX`
- Planner: `ScanAllByVertexProperty`[`Value`|`Range`] operators, cost estimation, label fallback (uses global index + label filter when no label+property index exists)
- Parallel runtime: chunked scan support
- `DUMP DATABASE`, C API, Prometheus metrics


colinbarry:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Add global vertex-property indices for matching vertices with only properties in common, rather than labels in common. [#4353](**{{link to the PR}}**)

specs/global-vertex-property-index.md
```diff
@@ -0,0 +1,261 @@
...
+---
+
+## 3. User-facing surface
+
+### Creating and dropping
+
+```cypher
+CREATE GLOBAL INDEX ON ':'(prop);
+DROP GLOBAL INDEX ON ':'(prop);
+```
```
imilinovic, line 52: I think this is a typo, global edge index has
`CREATE GLOBAL EDGE INDEX ON :(property_name);`

colinbarry: I agree. Fixed in the spec.

include/mgp.hpp
```diff
@@ -5405,6 +5405,22 @@ inline List ListAllLabelPropertyIndices(mgp_graph *memgraph_graph) {
   return List(label_property_indices, StealType{});
 }
 
+inline bool CreateVertexPropertyIndex(mgp_graph *memgraph_graph, const std::string_view property) {
```
imilinovic, line 5415: drop the const

include/mgp.hpp
```diff
@@ -5405,6 +5405,22 @@ inline List ListAllLabelPropertyIndices(mgp_graph *memgraph_graph) {
...
+  return create_vertex_property_index(memgraph_graph, property.data());
+}
+
+inline bool DropVertexPropertyIndex(mgp_graph *memgraph_graph, const std::string_view property) {
```
imilinovic, line 5419: drop const

src/query/interpreter.cpp
```diff
@@ -4441,6 +4441,68 @@ PreparedQuery PrepareIndexQuery(ParsedQuery parsed_query, bool in_explicit_trans
   auto *dba = &*current_db.execution_db_accessor_;
 
   auto *storage = db_acc->storage();
+
+  // Global vertex property index: no label, just a property.
+  if (index_query->label_.name.empty()) {
+    MG_ASSERT(index_query->properties_.size() == 1, "Global vertex property index requires exactly one property.");
```
imilinovic, line 4439: queries like
`CREATE INDEX ON :``;`
now crash the db, maybe add a bool to index_query whether its boolean? not a big cost.

src/query/plan/operator.hpp
```diff
@@ -133,6 +133,9 @@ class ScanAllByEdgeProperty;
 class ScanAllByEdgePropertyValue;
 class ScanAllByEdgePropertyRange;
 class ScanAllByEdgeId;
+class ScanAllByVertexProperty;
+class ScanAllByVertexPropertyValue;
+class ScanAllByVertexPropertyRange;
```
imilinovic: I'd create only one operator, its cleaner -> how LabelProperty index does it, dont think there is any benefit to this approach

src/query/plan/operator.cpp
```diff
@@ -10184,6 +10312,140 @@ std::unique_ptr<LogicalOperator> ScanParallelByEdgePropertyRange::Clone(AstStora
...
+
+UniqueCursorPtr ScanParallelByVertexPropertyRange::MakeCursor(utils::MemoryResource *mem,
+                                                              metrics::DatabaseMetricHandles &metric_handles) const {
+  metric_handles.scan_all_by_vertex_property_range_operator.Increment();
+#ifdef MG_ENTERPRISE
+  auto get_chunks = [this](Frame &frame, ExecutionContext &context) {
+    auto *db = context.db_accessor;
+    ExpressionEvaluator evaluator = ExpressionEvaluator{&frame, context, view_, nullptr, &context.number_of_hops};
+
+    auto [maybe_lower, maybe_upper] = ConvertBoundsAndCheckNull(lower_bound_, upper_bound_, evaluator);
```
imilinovic: I think if you pass [null, null] for upper bound and lower bound all nodes get returned, 

`**Failure.** Global index on `p`, parallel execution enabled (enterprise + a `PARALLEL` directive).
`MATCH (n) WHERE n.p > $x RETURN n` with `$x = null` — or any bound expression that evaluates to Null
at runtime, e.g. fed by an unmatched `OPTIONAL MATCH`:

- serial plan → **0 rows**
- parallel plan → **every vertex that has `p`**`

src/storage/v2/inmemory/vertex_property_index.cpp
```diff
@@ -0,0 +1,527 @@
...
+    PropertyId property, std::optional<utils::Bound<PropertyValue>> const &lower,
+    std::optional<utils::Bound<PropertyValue>> const &upper) const {
+  auto it = index_container_->indices_.find(property);
+  MG_ASSERT(it != index_container_->indices_.end(), "Index for vertex property {} doesn't exist", property.AsUint());
+  auto acc = it->second->skip_list_.access();
+  // NOLINTNEXTLINE(bugprone-narrowing-conversions,cppcoreguidelines-narrowing-conversions)
+  return acc.estimate_range_count(lower, upper, utils::SkipListLayerForCountEstimation(acc.size()));
+}
+
+void InMemoryVertexPropertyIndex::RemoveObsoleteEntries(uint64_t oldest_active_start_timestamp, std::stop_token token) {
```
imilinovic: https://github.com/memgraph/memgraph/pull/4475
check this, there was some gc bug, in edge index implementation, I think you have to do the same

src/query/procedure/mg_procedure_impl.cpp
```diff
@@ -3508,6 +3508,86 @@ mgp_error mgp_list_all_label_property_indices(mgp_graph *graph, mgp_memory *memo
   });
 }
 
+mgp_error mgp_create_vertex_property_index(mgp_graph *graph, const char *property, int *result) {
```
imilinovic, line 3514: this will MG_ASSERT if called with wrong accessors, not sure if this is fine behavior, all previous ones have this behavior but we should think about it

src/query/plan/pretty_print.cpp
```diff
@@ -391,6 +397,9 @@ PRE_VISIT_DBA_TS(ScanAllByEdgeProperty);
 PRE_VISIT_DBA_TS(ScanAllByEdgePropertyValue);
 PRE_VISIT_DBA_TS(ScanAllByEdgePropertyRange);
 PRE_VISIT_DBA_TS(ScanAllByEdgeId);
+PRE_VISIT_DBA_TS(ScanAllByVertexProperty);
+PRE_VISIT_DBA_TS(ScanAllByVertexPropertyValue);
+PRE_VISIT_DBA_TS(ScanAllByVertexPropertyRange);
```
imilinovic: this also might break memgraph
check https://github.com/memgraph/memgraph/pull/4489

src/query/plan/operator.hpp
```diff
@@ -848,6 +856,69 @@ class ScanAllByEdgePropertyRange : public memgraph::query::plan::ScanAllByEdge {
...
+
+  const utils::TypeInfo &GetTypeInfo() const override { return kType; }
+
+  ScanAllByVertexProperty() = default;
+  ScanAllByVertexProperty(const std::shared_ptr<LogicalOperator> &input, Symbol output_symbol,
+                          storage::PropertyId property, storage::View view = storage::View::OLD);
+  bool Accept(HierarchicalLogicalOperatorVisitor &visitor) override;
+  UniqueCursorPtr MakeCursor(utils::MemoryResource *, metrics::DatabaseMetricHandles &) const override;
+
+  std::string ToString() const override;
```
imilinovic: same comment for check https://github.com/memgraph/memgraph/pull/4489

src/query/plan/rewrite/order_by_elimination.hpp
```diff
@@ -195,7 +195,8 @@ class OrderByEliminator {
            type_info == ScanAllByEdgeTypeProperty::kType || type_info == ScanAllByEdgeTypePropertyValue::kType ||
            type_info == ScanAllByEdgeTypePropertyRange::kType || type_info == ScanAllByEdgeProperty::kType ||
            type_info == ScanAllByEdgePropertyValue::kType || type_info == ScanAllByEdgePropertyRange::kType ||
-           type_info == ScanAllByEdgeId::kType;
+           type_info == ScanAllByEdgeId::kType || type_info == ScanAllByVertexProperty::kType ||
```
imilinovic: using ProvidedScan = std::variant<const ScanAllByLabelProperties *, const ScanAllByEdgeTypePropertyRange *,
                                    const ScanAllByEdgePropertyRange *, const ScanAllByEdgeTypePropertyValue *,
                                    const ScanAllByEdgePropertyValue *>;

this doesn't work currently, you also ned to update above I think, also add a test that checks this

src/storage/v2/inmemory/vertex_property_index.cpp
```diff
@@ -0,0 +1,527 @@
...
+  }
+
+  try {
+    auto const accessor_factory = [&] { return index->skip_list_.access(); };
+    if (tx) {
+      auto const insert_function = [&](Vertex &vertex, auto &index_accessor) {
+        TryInsertVertexPropertyIndex(vertex, property, index_accessor, snapshot_info, *tx);
+      };
+      PopulateIndexDispatch(
+          vertices, accessor_factory, insert_function, std::move(cancel_check), {} /*TODO: parallel*/);
```
imilinovic: i think its worth adding parallel population for recovery here

src/storage/v2/inmemory/vertex_property_index.cpp
```diff
@@ -0,0 +1,527 @@
...
+    } else {
+      auto const insert_function = [&](Vertex &vertex, auto &index_accessor) {
+        TryInsertVertexPropertyIndex(vertex, property, index_accessor, snapshot_info);
+      };
+      PopulateIndexDispatch(
+          vertices, accessor_factory, insert_function, std::move(cancel_check), {} /*TODO: parallel*/);
```
imilinovic: same as above

src/query/plan/rewrite/index_lookup.hpp
```diff
@@ -1749,17 +1881,25 @@ class IndexLookupRewriter final : public HierarchicalLogicalOperatorVisitor {
...
+    auto vertex_prop_result = FindBestVertexPropertyScan(node_symbol, bound_symbols, input, view, metadata);
+
     if (!labels.empty()) {
       auto maybe_label = FindBestLabelIndex(labels);
       if (maybe_label) {
         const auto &label = *maybe_label;
-        if (!max_vertex_count || db_->VerticesCount(GetLabel(label)) <= *max_vertex_count) {
+        auto const label_count = db_->VerticesCount(GetLabel(label));
+        if (!max_vertex_count || label_count <= *max_vertex_count) {
+          if (vertex_prop_result && vertex_prop_result->estimated_count < label_count) {
```
imilinovic, line 1900: there is USING INDEX syntax which I think you haven't added, not sure where it is
`MATCH (n:L) USING INDEX :L WHERE n.p = 1 RETURN n`
Can you add tests so you show it doesnt override other indices / gets used when needed

src/query/plan/rewrite/index_lookup.hpp
```diff
@@ -1749,17 +1881,25 @@ class IndexLookupRewriter final : public HierarchicalLogicalOperatorVisitor {
...
+            return std::move(*vertex_prop_result);
+          }
           metadata.labels_to_erase.push_back(label);
           auto op = std::make_unique<ScanAllByLabel>(input, node_symbol, GetLabel(label), view);
           return ScanByIndexResult{std::move(op), std::move(metadata)};
         }
       }
     }
+    if (vertex_prop_result) return std::move(*vertex_prop_result);
```
imilinovic: I think now
`MATCH (n) WHERE (n:A OR n:B) AND n.p = 1 ` will always prefer property index vs two label props (or labels case), maybe worth changing

src/query/plan/rewrite/index_lookup.hpp
```diff
@@ -1654,13 +1789,10 @@ class IndexLookupRewriter final : public HierarchicalLogicalOperatorVisitor {
...
       }
     }
-    // Now try to see if we can use label+property index. If not, try to use
-    // just the label index.
     auto labels = filters_.FilteredLabels(node_symbol);
     auto or_labels = filters_.FilteredOrLabels(node_symbol);
     if (labels.empty() && or_labels.empty()) {
-      // Without labels, we cannot generate any indexed ScanAll.
-      return std::nullopt;
+      return FindBestVertexPropertyScan(node_symbol, bound_symbols, input, view, metadata);
```
imilinovic: I think you don't respect `max_vertex_count`, not sure is it respected elsewhere or why it exists

src/query/plan/rewrite/index_lookup.hpp
imilinovic: also I'm not sure do we do it here, but you pick index only on its size while if you have a predicate you can scan/estimate range (ResolveAtPlantime function)

imilinovic (changes requested):

include/mgp.hpp
colinbarry: Agree (in theory), but this is consistent with the other functions in this file, so will leave as is. Made a note to perhaps tidy in some future clean-up PR.

src/query/interpreter.cpp
colinbarry: Fixed. Added `is_global` to detect label-property vs vertex-property indices.

src/query/plan/operator.cpp
colinbarry: Fixed for ranges, and added an e2e test which matches the value code/test.

src/storage/v2/inmemory/vertex_property_index.cpp
colinbarry: Fixed, same pattern at in #4475

src/query/plan/pretty_print.cpp
colinbarry: Fixed by rebase

src/query/plan/operator.hpp
colinbarry: Fixed by rebase

include/mgp.hpp
colinbarry: Same as above - this is the pattern used elsewhere.

src/query/plan/rewrite/index_lookup.hpp
colinbarry: Consistent with how `label-property` candidate selection works. `ResolveAtPlantime` is used in the cost estimator after operator creation, not during candidate selection. Improving this would benefit all index types, but is out of scope for this PR. Added a note to address this in a future PR.

colinbarry: Implemented `USING INDEX` hints.

src/query/plan/rewrite/index_lookup.hpp
colinbarry: `max_vertex_count` respected for this now, similar to other label scans

src/query/procedure/mg_procedure_impl.cpp
colinbarry: Agree, but also beyond the scope of this feature PR as it fixes all the front-facing mg_api calls. I've noted this, and will implement in a separate PR.

src/storage/v2/inmemory/vertex_property_index.cpp
colinbarry: Done

src/query/plan/operator.hpp
colinbarry: Done. I've merged the three individual operators into a single scan operator, same as label-property indices.

src/storage/v2/inmemory/vertex_property_index.cpp
colinbarry: Parallel recovery implemented

src/query/plan/rewrite/index_lookup.hpp
colinbarry: Fixed: I now check the estimated count of a vertex-property scan against the OR-labels cardinality, and use this decide on the likely-best scan.

imilinovic (approved):

PR #4374, opened 2026-07-09
fix: show source text in plan_v2 result headers and plan columns

imilinovic:
Two fixes so plan_v2 shows an unaliased column by its **source text** instead of a stripping placeholder. planner_v2 stays experimental and off by default, so there is no effect on the production planner; these are display-name fixes, not result-value fixes.

**What / how**
- **Result header** (`symbol.hpp`, `ast_converter`, `builder`): an unaliased column's header is its source text (`RETURN 1 + 1` → `1 + 1`), recovered from the stripped query by the symbol's `token_position`. plan_v2 dropped `token_position` on reconstructed symbols, so the header fell back to the stripped placeholder (`RETURN 42` showed header `0`). `token_position` now rides on the reconstructed symbol, independent of expression folding, so a folded column keeps its source text as its name.
- **Plan display name** (`cypher_query_interpreter`, `ast_converter`): stripping replaces an unaliased column's expression with a placeholder, so `EXPLAIN` showed `Produce {0+0}` for `RETURN 1 + 2`. The stripped per-column source text is threaded into the converter and used as the display name for unaliased, unfolded columns; aliased columns keep their alias, and a raw (unstripped) parse falls back to the AST name.

**Why**: without these, plan_v2 mislabels result-column headers and plan output for unaliased expression columns.

Stacked PR — **base branch is `feat/plan-v2-cardinality-scale`, not `master`.**

### DIFF
--- a/src/query/cypher_query_interpreter.cpp
+++ b/src/query/cypher_query_interpreter.cpp
@@ -180,7 +180,8 @@ ParsedQuery ParseQuery(const std::string &raw_query_string, UserParameters const
 
 auto MakeLogicalPlan(AstStorage ast_storage, CypherQuery *query, const Parameters &parameters, DbAccessor *db_accessor,
                      const std::vector<Identifier *> &predefined_identifiers,
-                     plan::v2::QueryPlannerContext &planner_context) -> std::unique_ptr<LogicalPlan> {
+                     plan::v2::QueryPlannerContext &planner_context,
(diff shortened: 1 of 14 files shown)


imilinovic:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - [ ]

PR #4399, opened 2026-07-13
feat: SSO authentication on coordinators

as51340:
Enables SSO HA authentication on coords. SSO is enabled when `--auth-module-mappings` flag is set to true and role with write privilege exists. In that case basic auth using username and password is forbidden.

Also:
- fixes flaky tests/drivers.
- removes unused EnableWritingOnMainRpc.
- adds forwarding for SetCoordinatorSetting.
- Fixes RPC timeouts for follower->leader communication and system db rpc messages

## ⚠️ Breaking change: the `COORDINATOR` privilege is removed

This PR removes the `COORDINATOR` privilege (`Permission::COORDINATOR`, `AuthQuery::Privilege::COORDINATOR`) and drops it from the `privilege` grammar rule. It never gated any operation, but it was grantable, so removing it is user visible:

- **`GRANT` / `DENY` / `REVOKE COORDINATOR` no longer parses.** Any provisioning script, migration, or IaC step that runs e.g. `GRANT COORDINATOR TO some_role;` now fails with a syntax error instead of succeeding. Such statements must be deleted — they were always no-ops.
- **`SHOW PRIVILEGES` no longer reports `COORDINATOR`,** and it is gone from `kPermissionsAll` / `kPrivilegesAll`, so `GRANT ALL PRIVILEGES` no longer includes it. Tooling that snapshots or diffs the full privilege list sees one fewer entry.
- **Existing grants need no migration.** Bit 25 is left reserved and is deliberately not reused, so a durably stored mask with bit 25 set is never reinterpreted as some other privilege. `Permissions::GetGrants()`/`GetDenies()` iterate `kPermissionsAll`, so a stale bit is simply never reported and never checked — it is inert. It also can no longer be revoked via Cypher, which is harmless for the same reason.
- **The new `COORDINATOR_READ` / `COORDINATOR_WRITE` privileges are not a drop-in replacement.** They are coordinator-only: `EditPermissions` rejects them on a data instance, and they gate the coordinator Cypher surface for SSO sessions. A `GRANT COORDINATOR` on a data instance has no equivalent to rewrite it to, because it never granted anything in the first place.

Documented as a breaking change in the v3.13.0 release notes.


specs/coordinator-sso-auth/issues/04-role-query-forwarding.md
```diff
@@ -0,0 +1,37 @@
...
+
+## What to build
+
+Make `CREATE ROLE`, `DROP ROLE`, and `SHOW ROLES` work from any coordinator by transparently
+forwarding to the leader, instead of erroring with not-leader.
+
+End-to-end behavior:
+
+- Introduce coordinator→coordinator RPCs for the three role operations (create, drop, and a
+  read for `SHOW ROLES`).
```
Ana-Marija-design: Being nitpicky, but does Create/drop/read include the set/grant/clear/revoke..?

as51340: no because there won't be privileges on coords, at least until someone asks for it

src/coordination/coordinator_instance.cpp
```diff
@@ -1149,6 +1160,187 @@ auto CoordinatorInstance::SetCoordinatorSetting(std::string_view const setting_n
   return SetCoordinatorSettingStatus::SUCCESS;
 }
 
+auto CoordinatorInstance::CreateRole(std::string_view const role_name) const -> CreateRoleStatus {
```
andrejtonev, line 1177: This applies equally to DropRole :1190, GrantPrivilege :1245, RevokePrivilege :1275.
Lost-update: these four methods do `GetRoles()` → mutate → `AppendLogAndWaitForCommit`
(whole-vector delta) with no `coord_instance_lock_`, unlike `RegisterReplicationInstance`
(:735) which holds it across exactly this pattern. The RPC server is multi-threaded, so
two concurrent admin writes read the same pre-image and the later commit clobbers the
earlier (DoAction overwrites the whole vector, coordinator_cluster_state.cpp:172). Both
queries return SUCCESS but one change — e.g. a REVOKE — is silently lost.

src/coordination/include/coordination/coordinator_rpc.hpp
```diff
@@ -639,6 +615,52 @@ struct UpdateDataInstanceConfigRes {
 
 using UpdateDataInstanceConfigRpc = rpc::RequestResponse<UpdateDataInstanceConfigReq, UpdateDataInstanceConfigRes>;
 
+// Coordinator->leader role/privilege RPCs. Follower coordinators forward role and privilege queries to the leader so
+// the query works from any coordinator. Write responses carry a bool success flag, matching the other management RPCs;
+// a false value (the write was rejected by the leader, or the RPC itself failed on a dead or mixed-version leader) is
+// surfaced to the client as a forwarding error rather than a crash.
+using CreateRoleReq = SingleArgMsg<utils::TypeId::COORD_CREATE_ROLE_REQ, "CreateRoleReq", 1, std::string>;
+using CreateRoleRes = SingleArgMsg<utils::TypeId::COORD_CREATE_ROLE_RES, "CreateRoleRes", 1, bool>;
```
andrejtonev: Write responses are bool-only, so the leader's status collapses: `ROLE_ALREADY_EXISTS` →
false → `LEADER_FAILED` on the follower. Result: `CREATE ROLE IF NOT EXISTS x` (x exists)
is a silent no-op on the leader but throws on a follower — the `if_not_exists` arm at
interpreter.cpp:687-692 is unreachable, it hits the `LEADER_FAILED` throw at :697. Same
collapse masks NO_SUCH_ROLE / ROLE_ALREADY_EXISTS for DROP/GRANT/REVOKE on followers.

src/query/frontend/ast/query/auth_query.hpp
```diff
@@ -224,7 +231,57 @@ class AuthQuery : public memgraph::query::Query {
...
+      }
+      for (auto const privilege : query.privileges_) {
+        if (!IsCoordinatorPrivilege(privilege)) {
+          return false;
+        }
+      }
+      return true;
+    }
+    case AuthQuery::Action::SHOW_PRIVILEGES:
+      return query.entity_type_ != AuthQuery::UserOrRoleType::USER;
```
andrejtonev: This branch checks only `entity_type_`, never `database_specification_`, so
`SHOW PRIVILEGES FOR ROLE r ON MAIN|CURRENT|DATABASE` is accepted on a coordinator and
returns rows instead of the coordinator-not-allowed error. Not a bypass (data is identical),
but it violates the "ON clause rejected"

src/rpc/client.hpp
andrejtonev, line 52: The 7 new forwarded RPCs (CreateRole/DropRole/Grant/Revoke/GetRoles/GetRolePrivileges/
SetCoordinatorSetting) aren't in this map, so `timeout_ms = nullopt` → `poll(&p,1,-1)`
blocks forever against a reachable-but-stuck leader, hanging the caller's Bolt session
thread (the connector's `catch(std::exception)` can't rescue a blocked poll).

src/coordination/coordinator_instance.cpp
```diff
@@ -1149,6 +1160,187 @@ auto CoordinatorInstance::SetCoordinatorSetting(std::string_view const setting_n
...
+  if (leader_id != raft_state_->GetMyCoordinatorId() && leader_id != -1) {
+    if (auto *leader = FindClientConnector(leader_id); leader != nullptr) {
+      if (auto res = leader->SendRpc<GetRolesRpc>(); res.has_value()) {
+        return std::move(*res);
+      }
+    }
+  }
+
+  spdlog::trace("Couldn't reach leader for SHOW ROLES; serving from local replicated state.");
+  return raft_state_->GetRoles();
```
andrejtonev: SSO login (SessionHL.cpp:372) reads roles through here, and this falls back to local
replicated state when the leader is unreachable. During a leader outage a partitioned
follower can authenticate against a role that was already DROPped (or at a mask already
revoked) on the former leader — fail-open for an auth decision. The reverse (new role not
yet replicated) fails closed and is fine.

src/glue/SessionHL.cpp
```diff
@@ -351,6 +358,49 @@ std::expected<void, communication::bolt::AuthFailure> SessionHL::SSOAuthenticate
...
+  auto auth_result = authenticator.Authenticate(scheme, identity_provider_response);
+  if (!auth_result) {
+    return std::unexpected{communication::bolt::AuthFailure::kGeneric};
+  }
+
+  // The session carries the effective privilege mask; the coordinator privilege checks (query gate and ROUTE handler)
+  // consult it rather than the auth kvstore. Authentication already rejects any identity without COORDINATOR_READ or
+  // COORDINATOR_WRITE, so a surviving session can run at least read queries. The matched role names are carried too so
+  // SHOW CURRENT ROLE can report the session's roles.
+  interpreter_.SetCoordinatorPrivileges(auth_result->effective_mask);
```
andrejtonev, line 429: The effective mask is captured once here at login and never refreshed; the gate
(interpreter.cpp:10288) and ROUTE (:9842) read it directly. So `REVOKE COORDINATOR_WRITE
FROM ROLE x` doesn't downgrade an already-connected session until it reconnects — unlike
data-instance sessions, which re-check on the auth epoch. Given routing connections are
long-lived, that's real revocation latency.

src/glue/coordinator_sso_authenticator.hpp
```diff
@@ -0,0 +1,67 @@
...
+    uint64_t effective_mask;
+    std::vector<std::string> roles;
+  };
+
+  CoordinatorSSOAuthenticator(ModuleRunner module_runner, RoleMaskProvider role_mask_provider);
+
+  // Authenticates the SSO connection. On success returns the session's effective privilege mask (the union of the
+  // matched roles' masks) together with the matched role names, or nullopt on rejection. Rejection cases: the module
+  // fails / returns an invalid token, the module returns no roles, or any returned role does not exist in the committed
+  // role set. A returned role that exists but carries no grant contributes a zero mask, so a bare-role session
```
andrejtonev: Stale: this says a bare-role session authenticates with an all-denying (zero) mask, but
the impl rejects a zero-mask login (coordinator_sso_authenticator.cpp:50). Update the comment.

src/coordination/include/coordination/coordinator_ops_status.hpp
```diff
@@ -18,7 +18,55 @@
...
+  RAFT_LOG_ERROR,
+  UNKNOWN_SETTING,
+  INVALID_ARGUMENT,
+  NOT_LEADER,
+  LEADER_NOT_FOUND,
+  LEADER_FAILED
+};
+
+// Role/privilege ops forward to the leader (see CoordinatorInstance). SUCCESS/LEADER_FAILED/LEADER_NOT_FOUND make the
+// write enums satisfy the ForwardableStatus concept; the RPC round-trip preserves the leader's exact status so a
```
andrejtonev: This says the RPC round-trip "preserves the leader's exact status" — it doesn't; the write
Res is bool-only (coordinator_rpc.hpp:623), so non-SUCCESS collapses to LEADER_FAILED.
This comment is what makes the IF-NOT-EXISTS-on-follower bug easy to miss.

src/communication/bolt/v1/states/init.hpp
andrejtonev: `utils::Split(mapping, ":")[0]` has no empty-check; a trailing `;` in --auth-module-mappings
yields an empty element → out-of-bounds. Pre-existing, but this branch newly routes
coordinator connections through here, so it's newly reachable with a non-basic scheme.
Cheap empty guard warranted.

src/glue/SessionHL.cpp
```diff
@@ -351,6 +358,49 @@ std::expected<void, communication::bolt::AuthFailure> SessionHL::SSOAuthenticate
...
+  interpreter_.SetCoordinatorRoles(std::move(auth_result->roles));
+  return {};
+}
+#endif
+
 void SessionHL::LogOff() {
```
andrejtonev, line 465: LogOff resets DB/user but not `coordinator_permissions_`/`coordinator_roles_`. Verified
not exploitable today (post-LOGOFF the state machine only accepts LOGON before any query),
but a cheap defense-in-depth reset — SetCoordinatorPrivileges(0) + SetCoordinatorRoles({}) —
is worth adding beside the ResetUser().

src/coordination/coordinator_instance.cpp
as51340: Agree, fixed

src/glue/SessionHL.cpp
as51340: Let's talk about this because this potentially slows down every ROUTE request because we have to contact the current leader

as51340: actually thought twice about this: This will potentially slow down every coordinator query but those aren't important to be fast. ROUTE is a different story but routing table is refetched every 5min so not an issue.

as51340: fixed

src/communication/bolt/v1/states/init.hpp
```diff
@@ -135,14 +158,78 @@ std::optional<State> AuthenticateUser(TSession &session, Value &metadata) {
...
+    const bool sso_configured = !FLAGS_auth_module_mappings.empty();
+    if (schema == "basic" || schema == "none") {
+      if (sso_configured) {
+        // The full (non-cached) check, matching the gate the SSO path uses: the cached flag is only refreshed every
+        // few minutes, so a license that expired by date would keep basic denied while SSO already rejects every
+        // login, locking every Bolt scheme out of the coordinator until the next refresh.
+        bool deny_basic = license::global_license_checker.IsEnterpriseValid().has_value();
+        if (deny_basic) {
+          // nullopt (leader unreachable / no coordinator state) => keep basic denied (fail-closed).
+          deny_basic = session.CoordinatorHasWritableRole().value_or(true);
```
andrejtonev, line 196: This .value_or(true) makes a leaderless coordinator refuse every auth scheme. On a live 3-coordinator cluster with two of three killed, all seven probes were denied: fresh SSO, basic, credential-less none, and an already-open SSO session losing SHOW INSTANCES, FORCE RESET CLUSTER STATE, YIELD LEADERSHIP and SHOW ROLES (mask 0 via interpreter.cpp:9843). Those last three were deliberately built to work without a leader.

It also doesn't self-heal: LEADER_READY needs ReconcileClusterState_(), which fails with no MAIN and no reachable data instance — the state DEMOTE INSTANCE <main> persists — and the escape hatch SET INSTANCE … TO MAIN needs COORDINATOR_WRITE. That contradicts specs/coordinator-sso-auth.md:174-176 ("always at least one working auth path").

Worth noting the damage isn't in this line alone — it's this plus fail-closed on per-query enforcement for already-open sessions (interpreter.cpp:9843-9846), which is what gates the recovery queries. Either alone is defensible.

src/rpc/client.hpp
andrejtonev, line 52: This table went 18 → 47 entries, and only 7 of the 30 additions are the new role RPCs — the other 23 are pre-existing calls that were previously unbounded. Five of those are now bounded below their leader-side work:

ForceResetReq 10s vs ReconcileClusterState's while (IsLeader()) loop under ExponentialBackoff{1s, 60s} (effectively unbounded) · SetInstanceToMainReq 10s vs ≤33s of sequential SwapMainUUIDRpc · RegisterInstanceReq 10s vs ≤23s · Demote/UnregisterInstanceReq 10s vs ≤13s.

Since CoordinatorInstanceConnector::SendRpc (coordinator_instance_connector.hpp:33-36) catches everything and returns a default-constructed response, a timeout is indistinguishable from a genuine failure and surfaces as LEADER_FAILED — so we report failure for an operation the leader already committed (Raft itself caps at client_req_timeout_ = 3000ms).

Worth noting a missing entry is worse than a wrong one: nullopt → poll(&p, 1, -1) (src/io/network/socket.cpp:413-441), i.e. hang forever — and tests/unit/rpc_timeouts.cpp:313-329 injects its own hand-written 11-pair map and never reads default_rpc_timeouts_ms, so CI can't catch one.

src/communication/bolt/v1/states/init.hpp
```diff
@@ -135,14 +158,78 @@ std::optional<State> AuthenticateUser(TSession &session, Value &metadata) {
...
+    // every login) or when the committed role set has no COORDINATOR_WRITE role (SSO validates the module's roles
+    // against that set, and only a WRITE role can create the first administrator). In either case basic stays open as
+    // the break-glass path so an admin is never permanently locked out -- covering a fresh boot with mappings and no
+    // roles, and dropping the last writable role on a live cluster. A transient leader outage leaves the role set
+    // unknown; that case is fail-closed (basic denied), matching the SSO path, since it is temporary and SSO is
+    // unavailable then too. An SSO scheme present in --auth-module-mappings runs the coordinator SSO path
+    // (enterprise-gated); any other/unknown scheme is rejected.
+    const bool sso_configured = !FLAGS_auth_module_mappings.empty();
+    if (schema == "basic" || schema == "none") {
+      if (sso_configured) {
```
andrejtonev, line 189: Flagging this as a product question rather than a bug, because your own e2e asserts it (auth.py:1544) so I'm reading it as deliberate.

The third conjunct means the basic-auth deny is conditional on a COORDINATOR_WRITE role existing anywhere in the committed set. So REVOKE COORDINATOR_WRITE FROM <last writable role> or DROP ROLE — both of which read as tightening operations — re-open credential-less full-WRITE access on every coordinator, with only an spdlog::warn. Same for license expiry. And an operator who deliberately grants only COORDINATOR_READ roles never satisfies the condition at all, so basic stays wide open permanently.

Worth being explicit that CoordinatorPassthroughAuthenticate() (SessionHL.cpp:410-414) validates nothing — no principal, no credentials, no user records exist — so "basic auth is allowed" means "any client that can reach the port is a cluster admin". I ran both your tests and they pass, so the transition is real.

This is strictly better than master, which ignores coordinator auth entirely — the only concern is that the protection disengages silently on a routine admin action. If it's intended, suggest (a) an opt-out flag, (b) narrowing the escape to roles.empty() so it's monotonic and can never re-open once bootstrapped, and (c) surfacing "break-glass open" in SHOW INSTANCES/metrics rather than a log line.

src/coordination/coordinator_cluster_state.cpp
```diff
@@ -164,6 +168,10 @@ auto CoordinatorClusterState::DoAction(CoordinatorClusterStateDelta delta_state)
   if (delta_state.global_read_only_.has_value()) {
     global_read_only_ = *delta_state.global_read_only_;
   }
+
+  if (delta_state.roles_.has_value()) {
+    roles_ = std::move(*delta_state.roles_);
```
andrejtonev, line 189: Replacing roles_ wholesale from a whole-vector delta is safe on a single leader — coord_instance_lock_ plus nuraft's blocking return really do serialise same-leader read-modify-writes. The gap is cross-term: that lock is process-local and guarantees nothing across a leadership change.

The interleaving: L1 appends roles={A} at index N, gets a majority ack, applies it, returns SUCCESS to the client — then dies before its heartbeat carries commit_idx=N. L2 holds N in its log unapplied, wins the election, and becomes leader without appending anything. A subsequent GRANT reads a role set with no A and appends the whole vector again, so applying N then N+1 erases a role we already acknowledged. The same window serves pre-revocation privileges, which makes the "strong read from the leader" comment at SessionHL.cpp:421-422 untrue there.

src/coordination/coordinator_instance_management_server_handlers.cpp
```diff
@@ -176,6 +176,105 @@ void CoordinatorInstanceManagementServerHandlers::Register(CoordinatorInstanceMa
...
+
+  // Role/privilege reads forwarded from a follower. An unengaged optional means this coordinator couldn't serve the
+  // read as the ready leader (or reach one), which the follower surfaces as a forwarding error.
+  server.Register<GetRolesRpc>([&](std::optional<rpc::FileReplicationHandler> const & /*file_replication_handler*/,
+                                   uint64_t const request_version,
+                                   slk::Reader *req_reader,
+                                   slk::Builder *res_builder) -> void {
+    CoordinatorInstanceManagementServerHandlers::FwdRequestHandler<GetRolesRpc>(
+        [&coordinator_instance]() -> std::optional<std::vector<CoordinatorRole>> {
+          return coordinator_instance.GetRoles();
```
andrejtonev, line 258: This binds the forwarding coordinator_instance.GetRoles(), while the sibling handler 14 lines below binds the leader-only GetRolePrivilegesAsLeader. Since GetRoles() itself forwards when not LEADER_READY, a follower with a stale leader id can bounce A→B→A, each hop costing up to the 10s timeout, on a server with kCoordInstanceManagementServerThreads = 1.

The bigger issue is upstream: EffectiveCoordinatorPermissions() (interpreter.cpp:9833-9853) has no cache, so every SSO session pays a synchronous leader RPC on every query and every ROUTE — and driver routing pools re-ROUTE continuously. CoordinatorHasWritableRole puts the same RPC on the pre-auth handshake.

Note GetRolesAsLeader() existed and was removed in dbbc420ce; restoring and binding it is a one-liner for the bounce. For the per-query cost, the epoch-based revalidation pattern at glue/query_user.cpp:23-36 gives you live revocation without an RPC per query — that was suggested in the 07‑20 round too.

src/glue/SessionHL.cpp
```diff
@@ -351,10 +358,86 @@ std::expected<void, communication::bolt::AuthFailure> SessionHL::SSOAuthenticate
...
+    return std::unexpected{communication::bolt::AuthFailure::kGeneric};
+  }
+
+  // The session carries the matched role names; the coordinator privilege checks (query gate and ROUTE handler)
+  // recompute the effective mask from these roles against the leader's committed role set on every check, so a
+  // later REVOKE or DROP ROLE downgrades this session without a reconnect. The login-time mask is stored too, but it
+  // is authoritative only for role-less (basic-auth passthrough) sessions. Authentication already rejects any identity
+  // without COORDINATOR_READ or COORDINATOR_WRITE, so a surviving session starts with at least read privileges.
+  interpreter_.SetCoordinatorPrivileges(auth_result->effective_mask);
+  interpreter_.SetCoordinatorRoles(std::move(auth_result->roles));
```
andrejtonev, line 430: This sets the privilege mask and role names but never SetSessionInfo, unlike SSOAuthenticate at :355. The module's username is right there in the IdP response and Auth::SSOGetRoleNames (auth.cpp:640-668) drops it on the coordinator path.

Consequence: an audit entry is written for every query (:490-501) but with an empty username and no role, and session_user_or_role_ stays null. So after somebody runs FORCE RESET CLUSTER STATE or SET INSTANCE x TO MAIN, the logs can tell you which role did it (from the INFO line in auth.cpp) but never who. For an access-control feature on the control plane that seems worth closing.

Looks like ~3 lines: carry username through SSOGetRoleNames → AuthResult → SetSessionInfo. (Master had no principal to record, so this is a new-capability gap rather than a regression.)

src/communication/bolt/v1/states/init.hpp
andrejtonev, line 156: Because scheme_in_module_mappings excludes "basic" and the basic/none branch at :189 is tested before the SSO dispatch at :217, a config of --auth-module-mappings=basic:/path/mod.py leaves the coordinator with no usable auth path once a COORDINATOR_WRITE role exists: sso_configured is true (it's just flag-non-emptiness at :188) so basic is denied, and no SSO scheme is mapped so nothing else works.

I reproduced it — after creating one WRITE role, all five of basic-with-credentials, credential-less none, scheme="basic" through the SSO handshake, scheme="oidc", and credential-less auth after a restart are refused. The role persists via the Raft log, so the node stays unloggable across restarts.

What makes it nasty is the ordering: an operator who thinks they've configured authentication gets fail-open until they create their first admin role, and a bricked node immediately after. Suggest validating --auth-module-mappings at startup on coordinators — reject basic as a scheme, and reject empty/whitespace-only entries.

src/coordination/coordinator_instance_management_server_handlers.cpp
```diff
@@ -176,6 +176,105 @@ void CoordinatorInstanceManagementServerHandlers::Register(CoordinatorInstanceMa
             req_reader,
             res_builder);
       });
+
```
andrejtonev, line 179: These seven handlers put role and privilege administration on the coordinator management port, which has no authentication — there's no credential field in the wire header at all (rpc/version.hpp:52-58) — and TLS/peer-verification is off unless all three --cluster-* flags are set (communication/context.hpp:173-178).

To be clear about what is and isn't new: the port already accepts RegisterInstanceRpc, ForceResetRpc, SetInstanceToMainRpc etc., so full cluster control is already unauthenticated there and this branch changes nothing about the port itself. What's new is the capability class — CreateRoleRpc + GrantPrivilegeRpc mint a durable, Raft-replicated, Bolt-reachable COORDINATOR_WRITE role that survives restart, and DropRoleRpc on the last writable role flips the basic-auth break-glass open (H1), turning management-port reachability into Bolt-port access.

Not asking for a redesign here. Suggest documenting the trust boundary explicitly in the spec, and considering requiring mTLS for these specific write handlers.

src/coordination/coordinator_instance.cpp
```diff
@@ -834,6 +834,7 @@ auto CoordinatorInstance::RegisterReplicationInstance(DataInstanceConfig const &
 auto CoordinatorInstance::UnregisterReplicationInstance(std::string_view instance_name)
     -> UnregisterInstanceCoordinatorStatus {
   metrics::Metrics().global.unregister_repl_instance->Increment();
+  auto lock = std::lock_guard{coord_instance_lock_};
```
andrejtonev: This lock_guard moved from after ForwardToLeader (master has it at :848) to before it, so on a follower we now hold the coordinator's central lock across a 10s UnregisterInstanceRpc. The lock isn't needed until FindReplicationInstance.

While it's held, ReconcileClusterState_ (:452) and the nuraft become-follower callback (:176) both block on unconditional unique_lock — so a promotion to LEADER_READY, and therefore failover, can be delayed by the RPC timeout. (I initially thought this also starved the health-check callbacks; it doesn't — they return on status != LEADER_READY before reaching try_lock_for.)

The two methods added alongside it (:911, :959) correctly take the lock after forwarding, so this looks unintentional. Unless it was fixing a specific race, suggest reverting to master's ordering.

tests/e2e/high_availability/auth.py
```diff
@@ -548,5 +552,1055 @@ def test_ha_mt_auth_scenario(test_name):
...
+    def bootstrap(cursor):
+        create_role_with_privilege(cursor, "reader", grant="COORDINATOR_READ")
+        create_role_with_privilege(cursor, "writer", grant="COORDINATOR_WRITE")
+        create_role_with_privilege(cursor, "bare", grant=None)
+
+    start_sso_cluster(test_name, bootstrap)
+    leader_port = sso_wait_for_ready_leader_port()
+
+    # READ-only session: read/introspection queries succeed.
+    assert len(sso_run(leader_port, "oidc", "reader", "SHOW INSTANCES")) >= 0
```
andrejtonev: assert len(sso_run(..., "SHOW INSTANCES")) >= 0 can't fail — len() is never negative — and this is the only place a READ-only session exercises a cluster-management read.

More broadly, the coverage gaps line up with the risky parts: nothing covers the leader-outage path (which is what 8499eb4de was written to harden — see B1), nothing covers the missing-handler forwarding path, none of the six role RPCs or the role SLK round-trip has a unit test, and a READ-only session is denied exactly one of 14 mutating actions (CREATE ROLE, :1366). test_lagging_follower_catch_up (:1088-1092) forwards SHOW ROLES to the leader, so it asserts the leader's state rather than the rejoined follower's reconstruction. The new tests/unit/bolt_session.cpp:140-149 hooks are stubs (CoordinatorHasWritableRole always nullopt), so a green Bolt unit suite isn't evidence about handshake policy.

The suite itself is good — real drivers, message-checked errors, and the follower-forwarding test the last round asked for is there at :979-985. I've attached repros for the leaderless path and the snapshot wipe (auth_repro.py + unit additions); reuse or rewrite as you prefer.

src/coordination/coordinator_instance.cpp
```diff
@@ -1149,6 +1163,178 @@ auto CoordinatorInstance::SetCoordinatorSetting(std::string_view const setting_n
   return SetCoordinatorSettingStatus::SUCCESS;
 }
 
+auto CoordinatorInstance::CreateRole(std::string_view const role_name) const -> CreateRoleStatus {
```
andrejtonev, line 1177: CreateRole here accepts whatever the parser produced — no NameRegexMatch, no --auth-user-or-role-name-regex, no length cap, no role-count cap — while Auth::AddRole (auth.cpp:1154-1157) validates all of that on data instances. A backticked identifier can carry arbitrary bytes and arbitrary length, and since every role op appends the whole role vector to the Raft log, a very long name amplifies into the log and every snapshot.

Low security impact — it needs COORDINATOR_WRITE, and that actor can already FORCE RESET CLUSTER STATE. The part I'd call a real bug is the consistency gap: an operator who sets --auth-user-or-role-name-regex will expect it to apply to coordinator roles and it silently doesn't. Suggest applying the same regex plus a length and count cap.

src/communication/bolt/v1/states/init.hpp
```diff
@@ -111,18 +122,30 @@ std::optional<State> SSOAuthentication(TSession &session, memgraph::communicatio
...
+  auto credentials = ExtractSSOCredentials(data);
+  if (!credentials) {
+    return State::Close;
+  }
+  const auto &[scheme, identity_provider_response] = *credentials;
+  const auto auth_res = session.CoordinatorSSOAuthenticate(scheme, identity_provider_response);
+  if (!auth_res) {
+    // Rejected: invalid token, a role the module returned that doesn't exist on the coordinator, or a missing
+    // enterprise license. HandleAuthFailure sends the failure message and throws to close the connection.
+    HandleAuthFailure(session,
```
andrejtonev: This message covers "invalid token", "unknown role" and "missing license", but since 630b18524 it also absorbs a fourth case: the role exists and simply has no coordinator privilege. That's probably the most common real-world IdP misconfiguration — a group mapped to an ungranted role — and the operator is told their token is invalid.

coordinator_sso_authenticator.cpp:30-52 returns a bare std::nullopt for all four, so distinguishing them means a small reason enum rather than a message tweak. Worth it, I think — this is the error people will actually hit during rollout.

src/query/interpreter.hpp
```diff
@@ -311,6 +329,14 @@ class Interpreter final {
...
       user_or_role_{};  // Deep copy is not needed here, since it is only used in the current thread
 #ifdef MG_ENTERPRISE
+  // Coordinator privilege mask captured at login (auth::Permission bits). Consulted directly only for role-less
+  // (basic-auth passthrough) sessions, which carry full WRITE; sessions with coordinator roles recompute their mask
+  // per check via EffectiveCoordinatorPermissions. Zero denies everything, so an interpreter that never authenticated
+  // grants nothing: every privileged path must call SetCoordinatorPrivileges explicitly.
+  uint64_t coordinator_permissions_{0};
+  // Role names the session authenticated with on a coordinator (empty for a basic-auth passthrough session). Reported
+  // by SHOW CURRENT ROLE; coordinator roles live in Raft, not the auth kvstore, so they can't be looked up on demand.
+  std::vector<std::string> coordinator_roles_;
```
andrejtonev, line 341: coordinator_roles_ is only ever written at authentication (SessionHL.cpp:406/:413/:440), so SHOW CURRENT ROLE keeps reporting a role after it's been dropped — while every privileged query from that session is correctly denied, since the mask is re-derived from live state.

Harmless (it reveals only the caller's own claimed identity and grants nothing) but confusing in exactly the situation where someone runs it. Either cross-check against live GetRoles() when serving it, or note the login-time semantics in the docs.

src/query/frontend/opencypher/grammar/MemgraphCypher.g4
```diff
@@ -590,12 +592,13 @@ privilege : CREATE
           | STORAGE_MODE
           | MULTI_DATABASE_EDIT
           | MULTI_DATABASE_USE
-          | COORDINATOR
```
andrejtonev, line 594: Dropping COORDINATOR from the privilege rule makes GRANT COORDINATOR TO <role> a hard syntax error. That's spec-documented and intended, so just flagging the upgrade surface: any provisioning script or auth dump containing that statement now fails to parse rather than warning.

Two things soften it — .cypherl init files catch per line and warn (memgraph.cpp:181-191) rather than failing startup, and --init-file is rejected on coordinators regardless — so the realistic blast radius is interactive/driver-driven provisioning.

Also worth noting a stale bit 25 in an existing kvstore is now permanently unrevocable and invisible: REVOKE ALL PRIVILEGES expands to kPrivilegesAll and SHOW PRIVILEGES iterates the same list, and COORDINATOR is in neither. Inert, but it can only be cleared by dropping and recreating the role. A release note covers both. (Minor: the justifying comment at models.hpp:78 says the privilege "never gated any operation" — master did attach it to every CoordinatorQuery.)

src/coordination/coordinator_instance.cpp
andrejtonev: Not introduced by this PR — FindClientConnector and UpdateClientConnectors are byte-identical to master — but this branch makes it much easier to hit, so flagging it here.

The function takes the coordinator_connectors_ spinlock inside an inner block, then returns a raw pointer to a list element after the lock is released; both forwarding templates (coordinator_instance.hpp:172-174, :202-206) dereference it across a blocking SendRpc. Meanwhile erase_if at :197 can destroy that element from the nuraft config-change path. std::list gives address stability against insertion, not erasure — the precise bug is pointee destruction, which is why it reads as safe.

Master has 10 such call sites, all requiring a human to type a cluster command; this branch adds seven more on the pre-auth Bolt handshake and the per-query authorization gate, so the exposure changes character. Suggest fixing the primitive — hold a shared_ptr<CoordinatorInstanceConnector> and return a copy — rather than carrying it forward.

coords_auth.md
```diff
@@ -0,0 +1,75 @@
+One thing is how to authenticate routing connection the other, if we're logging in only to the coord.
```
andrejtonev: A few housekeeping items before merge:

coords_auth.md at the repo root is 75 lines of working notes — "TODO: check Clickhouse", "question for Ivan: is v1/states/init.hpp the only entrypoint for auth?", some Croatian fragments, and open security questions about intra-cluster TLS. Its substance belongs in specs/coordinator-sso-auth.md; the file itself shouldn't ship.

7f99ed541 ("docs: Add spec for SSO on coords") also deletes ten spec files belonging to two other features — specs/issues/01-08 (recovery-failure) and specs/read_only/issues/01-03. Looks like an accidental working-tree inclusion; worth confirming with those owners or restoring.

Bundled changes: SetCoordinatorSettingRpc forwarding, the driver-metadata flakiness fixes, and the RPC-timeout table rewrite (B4) are three unrelated behaviour changes in an SSO PR. The settings forwarding should stay — a previous review round asked for it — but the timeout table probably wants its own PR.

Metrics: removing 3 Prometheus families and 2 legacy metric names is monitoring-visible and needs a changelog entry. (The removal itself is clean — the kLegacyCoordinatorDeltaMetrics 28 → 26 count is right and the test files match exactly.)

src/communication/bolt/v1/states/init.hpp
as51340: That's fine. The comment from a week ago was that follower should contact a leader for the authentication purposes. Which means directly that this is not supported. Why you should be able to log-in to a follower which doesn't have a quorum?

src/communication/bolt/v1/states/init.hpp
as51340: intended

src/coordination/coordinator_cluster_state.cpp
as51340: NuRaft guarantees that this works as expected. You cannot become a leader if you don't have all logs from before

src/coordination/coordinator_instance_management_server_handlers.cpp
as51340: expected and intended

src/communication/bolt/v1/states/init.hpp
as51340: again totally expected. When there is no COORDINATOR_WRITE role, we accept anything.

src/coordination/coordinator_instance_management_server_handlers.cpp
as51340: Users have an option of using TLS....

src/coordination/coordinator_instance.cpp
as51340: will revert it

tests/e2e/high_availability/auth.py
as51340: I will fix the assertion. I believe that the test of the test suite is fine, there is no need to test everything on every level, we never do that

src/coordination/coordinator_instance.cpp
as51340: will add it

src/query/frontend/opencypher/grammar/MemgraphCypher.g4
as51340: that's fine, this privilege was never used, it's fine to delete it

src/communication/bolt/v1/states/init.hpp
as51340: will add it

src/query/interpreter.hpp
as51340: I will cross-check it against live GetRoles

src/coordination/coordinator_instance.cpp
as51340: true

src/rpc/client.hpp
as51340: makes sense

andrejtonev (approved):

as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Adds support for SSO authentication on coordinators using coordinators' roles. Users are not supported. In order to support HA routing connection, the same role needs to exist on both coordinators and data instances. Two privileges are added: COORDINATOR_READ and COORDINATOR_WRITE. Roles without any privilege cannot be used for authentication. This PR removes the `COORDINATOR` privilege (`Permission::COORDINATOR`, `AuthQuery::Privilege::COORDINATOR`) and drops it from the `privilege` grammar rule. It never gated any operation, but it was grantable, so removing it is user visible. [#4399](https://github.com/memgraph/memgraph/pull/4399)
- [x] **[ Documentation PR link ](https://github.com/memgraph/documentation/pull/1708)**

PR #4415, opened 2026-07-14
feat: Add collections.disjunction, collections.subtract, collections.duplicates

imilinovic:
Adds three list functions to the MAGE `collections` query module:

- **`collections.disjunction(list1, list2)`** — the symmetric difference of two lists (elements in exactly one of them), deduplicated. `disjunction([1,2,3,4,5], [3,4,5]) → [1,2]`.
- **`collections.subtract(list1, list2)`** — the first list treated as a set with every element of the second list removed. `subtract([1,2,3,4,5,6,6], [3,4,5]) → [1,2,6]`.
- **`collections.duplicates(coll)`** — the values that occur more than once, each reported a single time, in order of first repeat. `duplicates([1,3,5,7,9,9]) → [9]`.

`disjunction` and `subtract` results are set-style (unordered), consistent with existing `collections.intersection` / `collections.to_set`; `duplicates` preserves input order.

### NULL handling
The parameters are registered as nullable (via the low-level `func_add_arg` + `type_nullable`, since the high-level `mgp::AddFunction` only builds non-nullable types), so a `NULL` argument reaches the body and is handled explicitly instead of being rejected during argument validation:

- `disjunction`: `NULL` list resolves to the other list (returned unchanged, duplicates kept); both `NULL` → `NULL`.
- `subtract`: `NULL` first → `NULL`; `NULL` second → first list unchanged (duplicates kept).
- `duplicates`: `NULL` (or a list too small to hold a duplicate) → empty list.

`NULL` *elements inside* the lists are treated as ordinary values in the set path (two `NULL`s compare equal), so a `NULL` present in both lists is dropped by `disjunction`/`subtract`, and a `NULL` present in only one survives.

### Changes
- `mage/cpp/collections_module/` — constants, implementations, and registration (the three functions are registered with the shared nullable-list argument type).
- `config/mappings.json` — three new compatibility-map entries.
- `mage/tests/e2e/collections_test/` — 24 e2e cases: basic, dedup, empty second list, single-element, the symmetric-difference second loop, int/float equality, and each NULL path (including NULL elements that survive vs. are dropped).


imilinovic:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Adds three `collections` list functions — `collections.disjunction` (symmetric difference of two lists), `collections.subtract` (first list minus the elements of the second, deduplicated), and `collections.duplicates` (values occurring more than once) — callable like any other `collections` function, matching the compatibility-layer behavior. [#4415](https://github.com/memgraph/memgraph/pull/4415)
- [x] **[ Documentation PR link memgraph/documentation#1694 ]**

mage/cpp/collections_module/algorithm/collections.cpp
```diff
@@ -728,3 +728,118 @@ void Collections::FrequenciesAsMap(mgp_list *args, mgp_func_context *ctx, mgp_fu
...
+  try {
+    // Null first list -> null; null second list -> first list returned unchanged.
+    if (arguments[0].IsNull()) {
+      result.SetValue();
+      return;
+    }
+    if (arguments[1].IsNull()) {
+      result.SetValue(arguments[0].ValueList());
+      return;
+    }
```
DavIvek, line 789: `subtract(x, null)` keeps duplicates while `subtract(x, [])` dedups ([1,2,2,3] vs [1,2,3]). Since subtract treats the first list as a set, null-second should dedup too, fall through to the set path. `test_subtract_null2` needs updating to [1,2,3].

DavIvek (approved):

mage/cpp/collections_module/algorithm/collections.cpp
imilinovic: made to match apoc

PR #4416, opened 2026-07-14
fix: Align collections null-argument handling with compatibility layer

imilinovic:
Every `collections.*` function declared its list parameters as non-nullable `LIST`, so passing `NULL` was rejected during argument validation (a runtime error) before the function body ran — diverging from the behavior expected by the compatibility layer, which accepts `NULL` and returns a defined result per function.

This registers list (and the relevant scalar) parameters as nullable — via the low-level `func_add_arg` / `proc_add_arg` + `type_nullable`, since the high-level `mgp::AddFunction` only builds non-nullable types — and handles `NULL` explicitly in each body, returning the compatibility-equivalent result:

| Functions | Result on `NULL` argument |
|---|---|
| `sum`, `sum_longs`, `avg`, `min`, `max`, `to_set`, `pairs` | `NULL` |
| `union`, `union_all` | the other list; both `NULL` → `NULL` |
| `remove_all` | `NULL` first → `NULL`; `NULL` second → first list |
| `intersection`, `sort`, `flatten` | empty list |
| `contains`, `contains_all`, `contains_sorted` | `false` |
| `frequencies_as_map` | empty map |
| `split`, `partition` (procedures) | no rows |

Non-`NULL` and empty-list semantics are unchanged; this only fixes the `NULL`-argument path. Each function's `NULL` result was checked against the reference behavior of its equivalent (per `config/mappings.json`).

> Note: a few functions also diverge on the **empty-list** path (`sum([])`/`avg([])`/`min([])`/`max([])` should be `null`; `containsAll([], [])` should be `false`; `split([])` should yield no rows). Those are a separate axis from `NULL` (and `split([])` has an existing test asserting current behavior), so they are intentionally out of scope here and left for a follow-up.

### Changes
- `mage/cpp/collections_module/collections_module.cpp` — nullable parameter registration.
- `mage/cpp/collections_module/algorithm/collections.cpp` — per-function `NULL` guards.
- `mage/tests/e2e/collections_test/` — an e2e case for each `NULL` path.

### DIFF
--- a/mage/cpp/collections_module/algorithm/collections.cpp
+++ b/mage/cpp/collections_module/algorithm/collections.cpp
@@ -10,6 +10,8 @@
 // licenses/APL.txt.
 
 #include "collections.hpp"
+#include <fmt/format.h>
+#include <fmt/ostream.h>
 #include <algorithm>
 #include <list>
 #include <ranges>
@@ -33,9 +35,8 @@ void Collections::SumLongs(mgp_list *args, mgp_func_context *ctx, mgp_func_resul
 
     for (const auto list_item : list) {
       if (!list_item.IsNumeric()) {
-        std::ostringstream oss;
-        oss << list_item.Type();
-        throw mgp::ValueException("Unsupported type for this operation, received type: " + oss.str());
+        throw mgp::ValueException(
+            fmt::format("Unsupported type for this operation, received type: {}", fmt::streamed(list_item.Type())));
       }
       sum += static_cast<int64_t>(list_item.ValueNumeric());
     }
@@ -63,9 +64,8 @@ void Collections::Avg(mgp_list *args, mgp_func_context *ctx, mgp_func_result *re
 
     for (const auto list_item : list) {
       if (!list_item.IsNumeric()) {
-        std::ostringstream oss;
-        oss << list_item.Type();
-        throw mgp::ValueException("Unsupported type for this operation, received type: " + oss.str());
+        throw mgp::ValueException(
+            fmt::format("Unsupported type for this operation, received type: {}", fmt::streamed(list_item.Type())));
       }
       average += list_item.ValueNumeric();
     }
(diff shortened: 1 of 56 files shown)


imilinovic:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - The `collections.*` functions and procedures now accept `NULL` list and scalar arguments instead of raising an argument-validation error, returning a defined per-function result (e.g. `collections.sum(null)` → `null`, `collections.sort(null)` → `[]`, `collections.contains(null, x)` → `false`); no query changes are needed. [#4416](https://github.com/memgraph/memgraph/pull/4416)
- [x] **[ Documentation PR link memgraph/documentation#1694 ]**

mage/cpp/collections_module/collections_module.cpp
```diff
@@ -13,89 +13,135 @@
...
+    {
+      auto *func = mgp::module_add_function(module, Collections::kProcedureSort.data(), Collections::Sort);
+      mgp::func_add_arg(func, Collections::kArgumentSort.data(), ListOfAny());
+    }
+    {
+      // contains_sorted: keep the whole list nullable (contains_sorted(null, x) -> false) but reject a
+      // NULL element and a NULL search value at the type layer (both would NPE in the compatibility layer).
+      auto *func = mgp::module_add_function(module, Collections::kProcedureCS.data(), Collections::ContainsSorted);
+      mgp::func_add_arg(func, Collections::kArgumentInputList.data(), ListOfAny());
+      mgp::func_add_arg(func, Collections::kArgumentElement.data(), mgp::type_any());
```
DavIvek: contains_sorted(l, null) errors here (non-nullable ANY → rejected at validation), while contains(l, null) / contains_all(l, [null]) return false. Intended? If not, register NullableAny() and early-return false to match contains.

mage/cpp/collections_module/algorithm/collections.cpp
```diff
@@ -492,32 +534,46 @@ void Collections::Min(mgp_list *args, mgp_func_context *ctx, mgp_func_result *re
...
-      if (list[i].Type() != type && !(isListNumeric && list[i].IsNumeric())) {
+      if (list[i].IsNull()) {
+        continue;
+      }
+      if (!seeded) {
+        type = list[i].Type();
+        if (type == mgp::Type::Map || type == mgp::Type::Path || type == mgp::Type::List) {
+          std::ostringstream oss;
+          oss << type;
+          throw mgp::ValueException("Unsuppported type for this operation, receieved type: " + oss.str());
```
DavIvek: Maybe you can fmt instead so it's cleaner. Something like: `fmt::format("Unsupported type for this operation, received type: {}", fmt::streamed(type)));`

mage/cpp/collections_module/algorithm/collections.cpp
```diff
@@ -492,32 +534,46 @@ void Collections::Min(mgp_list *args, mgp_func_context *ctx, mgp_func_result *re
...
-    }
 
-    const bool isListNumeric = list[0].IsNumeric();
-    mgp::Value min{list[0]};
+    // NULL elements are skipped; the reference type is taken from the first non-null element and a
+    // min over only NULLs yields NULL.
+    mgp::Value min{};
+    mgp::Type type{mgp::Type::Null};
+    bool is_numeric{false};
+    bool seeded{false};
```
DavIvek: You can get rid of seeded flag I think.
`min` starts as a default-constructed `mgp::Value{}`, which is `Null`, and it's only ever assigned inside the loop from a value we've already confirmed is non-null.

So we can change it with `if (value.IsNull()) continue; `

DavIvek (approved):
Left a few comments but looks good! 🚀

mage/cpp/collections_module/collections_module.cpp
imilinovic: intended, matches apoc behavior

PR #4417, opened 2026-07-14
feat: add map.get and map.merge_list; align map functions on null/coercion behavior

imilinovic:
## Summary
Adds two new `map` module functions and brings several existing ones in line with the
expected compatibility behavior (null handling, node/relationship coercion, key
formatting). Also fixes a latent iterator-invalidation bug in the recursive key removal
path.

## New functions
- **`map.get(map, key, value = null, fail = true)`** — returns `map[key]`; falls back to a
  non-null `value`, otherwise throws when `fail` is true or returns null when it isn't.
  An existing key wins even when its stored value is null.
- **`map.merge_list(maps)`** — merges a list of maps left-to-right (last key wins); an empty
  list yields `{}`.

Both are exposed under their compatibility names via `config/mappings.json`.

## Behavior changes to existing functions
- **`map.merge`** — arguments renamed to `map1`/`map2`; null arguments treated as empty
  maps; nodes/relationships coerced to their property map.
- **`map.set_key`** — a null map is treated as empty and a null key is a no-op (previously
  threw); nodes/relationships coerced.
- **`map.remove_key` / `map.remove_keys` / `map.flatten`** — nodes/relationships coerced to
  their property map.
- **`map.from_values`** — pairs with a null key are skipped (the value is still consumed).
- **`map.from_lists`** — empty lists now yield an empty map instead of erroring; clearer
  size-mismatch message; value elements may be null.
- **Non-string keys** (e.g. from `from_values`/`from_nodes`) are stringified via the
  standard value-to-string conversion; doubles use their shortest round-tripping form
  (`2.0` → `"2"`, `0.1` → `"0.1"`, `1e-10` → `"1e-10"`).

## Bug fix
- **`RemoveRecursionSet`** (used by `map.remove_key`/`map.remove_keys`) previously erased the
  element it was iterating and then advanced an invalidated iterator (UB). Mutations are now
  collected and applied after iteration.


imilinovic:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Adds `map.get` (fetch a key's value, with an optional default and a fail flag) and `map.merge_list` (merge a list of maps into one, last key wins), and makes `set_key`, `from_values`, and `from_lists` handle null and empty inputs gracefully; call the new functions directly in Cypher. [#4417](https://github.com/memgraph/memgraph/pull/4417)
- [x] **[ Documentation PR link memgraph/documentation#1696 ]**

DavIvek (approved):

PR #4435, opened 2026-07-16
feat: add text.compare_cleaned function to text module

imilinovic:
Adds `text.compare_cleaned(text1, text2)` to the MAGE `text` module.

**What:** returns whether two strings are equal after a normalization pass. `NULL` arguments return `false`.

**How:** normalization keeps only ASCII letters and digits, lowercased, and drops everything else (accents, punctuation, whitespace, and non-ASCII characters); the two cleaned strings are then compared for equality. This ASCII-only scope keeps the module free of a Unicode-library dependency and consistent with its other string functions — accents and non-ASCII scripts are not folded to a base letter (e.g. `café` cleans to `caf`, not equal to `cafe`), which is documented at the top of `text.cpp`. Parameters are registered as nullable strings via the low-level `module_add_function` + `func_add_arg` + `type_nullable` path so a `NULL` argument reaches the body instead of being rejected during argument validation. Adds a compatibility-map entry and e2e coverage.

**Why:** fills a gap in the string-function compatibility surface — cleaned/normalized string comparison had no equivalent.

### DIFF
--- a/config/mappings.json
+++ b/config/mappings.json
@@ -102,6 +102,7 @@
   "apoc.refactor.renameType": "refactor.rename_type",
   "apoc.refactor.to": "refactor.to",
   "apoc.temporal.format": "temporal.format",
+  "apoc.text.compareCleaned": "text.compare_cleaned",
   "apoc.text.format": "text.format",
   "apoc.text.indexOf": "text.indexOf",
   "apoc.text.join": "text.join",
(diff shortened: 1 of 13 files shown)


imilinovic:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Adds `text.compare_cleaned(text1, text2)`, which returns whether two strings are equal after ASCII normalization (keeps only ASCII letters and digits, lowercased, and drops everything else) [#4435](https://github.com/memgraph/memgraph/pull/4435)
- [x] **[ Documentation PR link memgraph/documentation#1691 ]**

DavIvek (approved):

PR #4443, opened 2026-07-20
feat: add JSON conversion functions (from_json_map/list, to_map, to_json) to the convert module

imilinovic:
**⚠️ Breaking**

The APOC compatibility names `apoc.convert.fromJsonList` and `apoc.convert.toJson` now resolve to the new C++ `convert` module (`convert.from_json_list` / `convert.to_json`) instead of the previous Python `json_util` implementations (`json_util.from_json_list` / `json_util.to_json`). Output can differ — in particular, `to_json` serializes graph entities, spatial, and temporal types in the structured/ISO-8601 form described below. The `json_util.*` functions themselves remain available under their own names.

---

Adds four JSON/map conversion functions to the `convert` query module and repoints the corresponding compatibility names to them.

**What**
- `convert.from_json_map(map [, path])` — parses a JSON-object string into a Cypher map, with an optional `path` selecting a nested part of the document first.
- `convert.from_json_list(list [, path])` — parses a JSON-array string into a Cypher list, with the same optional `path` selector; the selection must resolve to an array.
- `convert.to_map(value)` — returns a map unchanged, a node/relationship as its property map, and null for null or any other value.
- `convert.to_json(value)` — serializes any value to a JSON string.

**How**
- `query_modules/convert.cpp`:
  - `from_json_map`/`from_json_list` reuse the `ParseJsonToMgp*` helpers (nlohmann_json) and share a `ResolveJsonPath` helper that applies the optional `path`. The `path` is translated into a JSON pointer and resolved against the parsed document; reference-token escaping is handled by nlohmann `json_pointer`. Both accept a nullable first argument and an optional `path` (default `""`), so they use the low-level `module_add_function`/`func_add_arg`/`func_add_opt_arg` path.
  - `to_json` serializes with nlohmann_json: scalars/lists/maps directly; nodes as `{id, type, labels, properties}`; relationships as `{id, type, label, start, end, properties}` (endpoints are full node objects); paths as a flat `[node, rel, node, ...]` array; points as `{crs, x/y/z}` (cartesian) or `{crs, longitude/latitude/height}` (geographic); and temporals via the engine's canonical string form (ISO-8601 for durations and zoned datetimes). Empty `labels`/`properties` are omitted.
- Semantics for the `from_*` functions: null input, an unresolved path, or a JSON-null selection yields null; a selection of the wrong shape (non-object for `from_json_map`, non-array for `from_json_list`) is rejected; unsupported path constructs (wildcards, filters, recursive descent) are rejected.
- `config/mappings.json`: repoints the `fromJsonList`/`fromJsonMap`/`toMap`/`toJson` compatibility names to the new C++ functions. The previous `json_util` (Python) implementations are left in place.
- `tests/e2e/query_modules/`: e2e coverage for all four functions — every `path` branch, and node/relationship/path/point/temporal serialization; the convert workload runs with properties-on-edges enabled so relationship/path serialization can be exercised.

**Path option (`from_json_map` / `from_json_list`)**

The optional second argument selects a nested part of the JSON document before conversion. It is a restricted path syntax that is translated into a JSON pointer.

Supported:

| Syntax | Meaning | Example (on `{"a": 1, "b": {"c": 2, "e": [10, 20]}}`) |
|---|---|---|
| `$` / empty / `null` | the whole document | `$` → the whole object |
| `.key` | object key step | `$.b` → `{"c": 2, "e": [10, 20]}` |
| `['key']` or `["key"]` | quoted key step (keys with dots, spaces, special chars) | `$['b']` → `{"c": 2, "e": [10, 20]}` |
| `[index]` | array element, 0-based | `$.b.e[1]` → `20` |

- Steps chain left to right: `$.b.e[0]` → `10`.
- A leading `$` is optional — `a.b` behaves the same as `$.a.b`.

Not supported (rejected with an error):

- wildcards — `$.e[*]`, `$.*`
- recursive descent — `$..x`
- filter expressions — `$.e[?(@.x > 1)]`
- array slices / unions — `$.e[0:2]`

Resolution rules: an unresolved path (`$.missing`) or a JSON-null selection yields `null`; a selection whose shape does not match the function (a non-object for `from_json_map`, a non-array for `from_json_list`) is rejected with an error.

To pull out a scalar leaf, index the returned map with Cypher rather than the `path` argument — e.g. `convert.from_json_map(candidate)['mode']` (or `.mode`), which returns `null` for a missing key.

**Why**
- Consolidates the JSON/map conversion helpers into one C++ module with consistent, structured serialization for graph entities, spatial, and temporal types, and lets users drill into nested JSON via a simple `path`.


imilinovic:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - The `convert` module gains four JSON/map conversion functions: `convert.from_json_map(map [, path])` and `convert.from_json_list(list [, path])` parse a JSON string into a map or list, with an optional path to select a nested part of the document first; `convert.to_map(value)` coerces a map, node, or relationship into a map; and `convert.to_json(value)` serializes any value — including nodes, relationships, paths, points, and temporals — into a JSON string. Call them like any other query-module function. [#4443](https://github.com/memgraph/memgraph/pull/4443)
- [x] **[ Documentation PR link memgraph/documentation#1699 ]**

tests/e2e/query_modules/convert_test.py
```diff
@@ -18,22 +19,20 @@
...
-    result = execute_and_fetch_all(cursor, f"RETURN convert.str2object('[2, 4, 8, [2]]') AS result;")[0][0]
+    result = execute_and_fetch_all(cursor, "RETURN convert.str2object('[2, 4, 8, [2]]') AS result;")[0][0]
     assert (result) == [2, 4, 8, [2]]
 
 
 def test_convert_list_wrong():
     cursor = connect().cursor()
-    result = execute_and_fetch_all(cursor, f"RETURN convert.str2object('[2, 4, 8, [2]]') AS result;")[0][0]
+    result = execute_and_fetch_all(cursor, "RETURN convert.str2object('[2, 4, 8, [2]]') AS result;")[0][0]
     assert (result) != [3, 4, 8, [2]]
```
as51340: what's the point of this test, isn't it the same as the previous one?

tests/e2e/query_modules/convert_test.py
```diff
@@ -9,6 +9,7 @@
 # by the Apache License, Version 2.0, included in the file
 # licenses/APL.txt.
 
+import json
```
as51340, line 12: Wouldn't these tests be more suitable as GQL ones?

as51340 (approved):

tests/e2e/query_modules/convert_test.py
imilinovic: preexisting thing, removed

tests/e2e/query_modules/convert_test.py
imilinovic: maybe, but I think that requires changing CI a bit to support loading query modules, and they were already located here so I'd keep it

imilinovic: edit: its possible but a bit clumsy for json, I'd keep here for now

PR #4445, opened 2026-07-20
refactor(utils): index the LRU cache by node iterators

Ignition:
LRUCache kept a std::list of entries and a std::unordered_map from key to list iterator. The map's key was a second copy of the entry key, so a large key (the AST and plan caches key on a stripped query, its full normalized text) was stored twice per entry.

Index by the list node instead. The list owns each entry (key and value, both const) and carries the LRU order; the index is a std::unordered_set of list iterators, hashed and compared by the key they point at through transparent functors. A lookup takes the bare key and locates the node in O(1) with no second key copy and no allocation. Each entry now holds one key and one index locator.

Entries are also immutable: a put for a key already present keeps the stored value and only refreshes recency, matching how both caches use it (they insert only after a lookup miss). The unit tests cover the contract, including that a key is materialized once and that a lookup constructs none.

### DIFF
--- a/src/utils/lru_cache.hpp
+++ b/src/utils/lru_cache.hpp
@@ -11,73 +11,103 @@
 
 #pragma once
 
+#include <cstddef>
 #include <list>
 #include <optional>
-#include <unordered_map>
+#include <unordered_set>
 #include <utility>
 
 namespace memgraph::utils {
 
 /// A simple LRU cache implementation.
 /// It is not thread-safe.
-
+///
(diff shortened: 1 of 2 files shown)


Ignition:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Memory reduction for LRU cache entries. [#4445](https://github.com/memgraph/memgraph/pull/4445)

imilinovic (approved):

PR #4447, opened 2026-07-20
fix: correct filterStartNode/minHops and bidirectional filter in path.subgraph_all/subgraph_nodes

imilinovic:
Fixes config-option handling in `path.subgraph_all` / `path.subgraph_nodes`.

**What**
- `filterStartNode` now defaults to `false`. When the start node is unfiltered it is treated as a plain whitelisted node instead of being returned unconditionally, so it no longer leaks into end/termination result sets while traversal through it still continues.
- `minHops` is now honored. The subgraph BFS previously enforced only the upper hop bound, so nodes closer than `minHops` were still returned. Nodes are returned when their distance falls within `[minHops, maxHops]`.
- The bidirectional (`<TYPE>`) relationship filter no longer returns duplicate nodes (and, in `subgraph_all`, duplicate relationships), and now respects the hop-range and label filters like every other traversal branch.
- The bidirectional dedup set previously keyed on a `string_view` whose backing wasn't guaranteed to outlive the traversal (a per-iteration local string, later a view into edge-type storage that still dangled for projected-graph edges); it now holds an owning key, with the per-edge direction lookup kept allocation-free.

**How**
In `path_module/algorithm/path.cpp`:
- Flip the `filterStartNode` default in the config reader, and in `PathSubgraph::TryInsertNode` gate insertion on `PathSizeOk(hop)` (full hop range) while treating an unfiltered start node as whitelisted-only. Closer-than-`minHops` nodes are still traversed, just not returned.
- In `ExpandFromRelationships`, the bidirectional-match branch now only enqueues the matched node instead of also appending it directly to the result set; the node is returned once via `TryInsertNode` on dequeue, which applies the hop-range and label filters and avoids the double insert.
- Also in `ExpandFromRelationships`, `seen` now keys on `std::string`; the transparent-hash `relationship_sets` lookup still takes a `string_view`, so the hot path allocates nothing.

E2E coverage updated/added under `tests/e2e/path_test`, including the previously-untested bidirectional path (node dedup, `minHops` window, and the `subgraph_all` node/relationship sets).

**Why**
The config options were mis-handled: `filterStartNode` had an inverted default and over-included the start node, `minHops` was silently ignored, and the bidirectional relationship filter duplicated results while bypassing the hop-range and label filters — all producing incorrect subgraph results.

**Note**
`minHops >= 2` returns nodes whose distance is within `[minHops, maxHops]` (a distance window), rather than rejecting the call.

### DIFF
--- a/src/mage/cpp/path_module/algorithm/path.cpp
+++ b/src/mage/cpp/path_module/algorithm/path.cpp
@@ -488,6 +488,11 @@ void Path::PathSubgraph::ExpandFromRelationships(const std::pair<mgp::Node, int6
 }
 
 void Path::PathSubgraph::TryInsertNode(const mgp::Node &node, int64_t hop_count, LabelBools &label_bools) {
+  // Nodes closer than minHops are still traversed but must not be returned.
+  if (!path_data_.helper_.PathSizeOk(hop_count)) {
+    return;
+  }
+
   if (path_data_.helper_.IsNotStartOrFiltersStartNode(hop_count == 0)) {
     if (path_data_.helper_.AreLabelsValid(label_bools)) {
       to_be_returned_nodes_.AppendExtend(mgp::Value(node));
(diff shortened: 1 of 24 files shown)


imilinovic:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - `path.subgraph_all` and `path.subgraph_nodes` now handle their config options correctly — `filterStartNode` defaults to `false`, `minHops` is honored, and the bidirectional relationship filter (`<TYPE>`) no longer returns duplicate nodes and relationships — so these procedures return correct subgraphs with no query changes needed. [#4447](https://github.com/memgraph/memgraph/pull/4447)

DavIvek (approved):

PR #4448, opened 2026-07-21
testing: Improve interactive_mg_runner.py script

as51340:
Improvements:
- Starting an already started instance won't fail.
- You can now start/stop multiple instances
- There is a persisted history through which you can iterate using keys up and down
- If setup queries fail, the exception is caught and a warning caught
- Instances are now started in parallel so that more than coordinator can be restarted at once

### DIFF
--- a/tests/e2e/interactive_mg_runner.py
+++ b/tests/e2e/interactive_mg_runner.py
@@ -26,6 +26,7 @@
 # approaches have to be employed.
 # NOTE: The instance description / context should be compatible with tests/e2e/runner.py
 
+import concurrent.futures
 import logging
 import os
 import secrets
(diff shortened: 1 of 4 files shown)


tests/e2e/interactive_mg_runner.py
```diff
@@ -70,7 +71,46 @@
 }
 
 
+HISTORY_MAX_SIZE = 1000
+HISTORY_FILE = os.path.expanduser("~/.interactive_mg_runner_history")
+ACTION_HISTORY = []
+history_loaded = False
+
+
+def load_history():
```
DavIvek, line 79: Consider replacing the hand-rolled reader with the stdlib `readline` module instead of building history on top of the termios byte-reader.

DavIvek (approved):
Nice work!

PR #4450, opened 2026-07-21
fix(utils): give a waiting UNIQUE priority over new shared acquirers

andrejtonev:
A `ResourceLock` acquirer waiting for UNIQUE could be starved indefinitely by a continuous stream of shared acquirers (READ, WRITE, READ_ONLY), since nothing made new shared acquisitions yield to it.

A waiting UNIQUE now registers as pending for as long as it waits, and a pending UNIQUE gates new shared acquisitions, mirroring the existing READ_ONLY-over-WRITE priority. The registration is RAII, so it survives a timeout or an exception out of the wait rather than blocking every shared acquisition for the lifetime of the lock. Both pending counters are plain integers mutated only under the mutex: a waiter reads them inside its wait predicate, so an off-mutex mutation can land between that check and the waiter's enqueue on the condition variable and lose the wake.

`UniquePendingScope` and `ReadOnlyPendingScope` give the same priority to callers that cannot block on the condition variable, such as a coroutine that must yield to a scheduler. The scope registers as pending once, then polls a non-blocking `try_acquire()` that hands ownership over on success. This is the foundation for non-blocking accessor acquisition in the stacked storage PR.

Analytical garbage collection took READ, read the storage mode under it, then acquired WRITE without releasing READ. With shared acquisition now gated on a pending UNIQUE that escalation deadlocks, so GC releases READ before requesting WRITE and re-reads the mode under the hold it will use.

Internally there is now one acquire path and one release path for all four ways to hold the lock, parameterised on the wait strategy, in place of a separate implementation per acquisition kind.

### DIFF
--- a/src/storage/v2/inmemory/storage.cpp
+++ b/src/storage/v2/inmemory/storage.cpp
@@ -2954,10 +2954,16 @@ void InMemoryStorage::CollectGarbage(std::unique_lock<utils::ResourceLock> main_
     auto read_guard = Guard{main_lock_, Guard::READ};
     if (storage_mode_ == StorageMode::IN_MEMORY_TRANSACTIONAL) return read_guard;
 
-    // Analytical: acquire WRITE before read_guard releases READ at scope exit, so the mode stays
-    // pinned across a continuous hold (one thread may hold READ+WRITE: lock_guard_condition<WRITE>
-    // checks only ro_count/ro_pending).
-    return Guard{main_lock_, Guard::WRITE};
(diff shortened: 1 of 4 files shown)


andrejtonev:

---
___

### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Under heavy concurrent load, operations that need exclusive access to the database could hang indefinitely while shared readers kept arriving. Those exclusive waiters now take priority so the hang no longer happens. [#4450](https://github.com/memgraph/memgraph/pull/4450)

Ignition (approved):

PR #4451, opened 2026-07-21
fix(storage): SetStorageMode double-unlock + non-blocking Try*Access primitives

andrejtonev:
SetStorageMode held main_lock_ in UNIQUE and passed that hold to FreeMemory by
adopting it into a second lock, so one hold had two owners and was released
twice. The second release drops the lock while another thread may already hold
it in UNIQUE. The hold is now handed off, so exactly one owner releases it.

The storage mode and isolation level were read outside any hold: by the accessor
factories, as constructor arguments evaluated before the constructor acquired
anything, and by replication. They are now atomic, and the constructor reads
them under the hold it keeps, so a transaction runs in the mode in force for the
life of that hold. Query preparation has to read the mode before taking the
access it decides, so it re-checks against the pinned copy afterwards and raises
a transient error on mismatch.

An Accessor now holds one ResourceLockGuard covering all four lock modes, and is
constructed from the hold it will keep rather than a tag. That makes the
non-blocking TryAccess, TryUniqueAccess and TryReadOnlyAccess expressible: one
probe, reporting busy instead of blocking or throwing. No caller yet; they are
for the parkable query preparation that follows.

Carried along: PrepareDeletableEdges read edge endpoints' deleted() with no
vertex lock, and that bit shares a non-atomic word with the delta pointer.
ResourceLock now asserts release balance, since the unsigned hold counts wrap on
an unmatched release and silently block a mode for the life of the process.

### DIFF
--- a/src/query/db_accessor.hpp
+++ b/src/query/db_accessor.hpp
@@ -676,7 +676,7 @@ class DbAccessor final {
 
   void Abort() { accessor_->Abort(); }
 
-  storage::StorageMode GetStorageMode() const noexcept { return accessor_->GetCreationStorageMode(); }
+  storage::StorageMode GetStorageMode() const noexcept { return accessor_->GetPinnedStorageMode(); }
 
   bool LabelIndexReady(storage::LabelId label) const { return accessor_->LabelIndexReady(label); }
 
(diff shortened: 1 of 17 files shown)


andrejtonev:

---
___

### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Fixed concurrent `SET STORAGE MODE` handling so mode switches are safe under load. [#4451](https://github.com/memgraph/memgraph/pull/4451)

Ignition (approved):

PR #4453, opened 2026-07-21
feat(query): parkable Prepare — a query blocked on a contended accessor parks its Bolt worker (flag-gated, off)

andrejtonev:
When a query needs a contended storage accessor — a UNIQUE accessor (index/constraint DDL, SetStorageMode) or a READ_ONLY accessor behind active writers — its Bolt worker blocks inside Prepare(), pinning that worker thread for the whole wait. With a fixed Bolt worker pool, a handful of such queries can pin every worker and stall otherwise-independent traffic, even though the CPU is idle.

What this does

Turns Prepare()'s accessor acquisition into a C++20 coroutine that parks when the accessor isn't immediately available and yields its worker back to the pool instead of blocking it. The worker services other queries; when the lock is released, the parked query is rescheduled and resumes exactly where it left off. Net effect: a single query waiting on a contended accessor no longer costs a whole worker.

Flag-gated OFF (--experimental-coro-prepare-accessor-yield=false by default). The flag-off path is byte-for-byte the original blocking Prepare() — this is inert until explicitly enabled.

Design

- Coroutine substrate (src/utils/): Task<T> (lazy, symmetric-transfer coroutine), WorkerResumeEvent (epoch-gated register-before-recheck wake), ParkState + single-owner ClaimPark CAS, deadline_park_registry (timeout via a monitor sweep, preserving --query-execution-timeout), and a pinned per-worker pool queue + RescheduleTaskOnWorker.
- Storage hook (src/storage/v2/): a per-Storage WorkerResumeEvent; NotifyMainLockReleased() fired on every accessor release; non-blocking TryAccess* probes (from the ResourceLock/Try*Access fix) drive the retry campaign. SupportsParkAcquire() is InMemory-only — disk/HIGH-isolation fall back to the blocking path.
- Bolt driving (src/communication/, src/glue/): a tri-state ExecuteResult (kNoMoreData / kMoreData / kNeedsCoroPrepare), HandlePrepareCoro, Session::parked_prepare_ + RunLoop / DrivePreparedRun, and a shutdown Drain so parked work can't leak past teardown.

Correctness & safety

- Result-identical to blocking Prepare: flag-on ≡ flag-off outputs across every profile tested; durable across restart/recover.
- Crash-free under a 150-client / 8-worker customer-shaped stress profile and a 40-scenario adversarial QA campaign (0 crashes).
- TSan-clean: substrate suites + full-server flag-on e2e under contention.
- Adversarially hardened: two independent review rounds fixed a shutdown-window UAF (F1), a vacuous e2e (F2, rewritten to be discriminating: flag-on 0.008 s vs flag-off 4.31 s), and a wake-on-release gap + shutdown hang (F5).


andrejtonev:
moved to #4497

PR #4454, opened 2026-07-21
fix(query): prevent LOAD PARQUET from crashing on invalid values

andrejtonev:
The parquet reader converts each Arrow record batch to Memgraph rows on a background prefetcher jthread. A converter can throw (e.g. a TIME64/TIMESTAMP/ DURATION value out of Memgraph's temporal range fails LocalTime/LocalDateTime/ Duration validation). That exception escaped the thread's top-level function, which calls std::terminate and crashes the whole server (SIGABRT) instead of returning a query error.

Capture any exception thrown by the prefetcher into an exception_ptr, finish the work queue, and rethrow it from GetNextRow on the consumer thread so it surfaces as a regular QueryRuntimeException. finish() happens-before pop() returning false, so the stored pointer is visible to the consumer without extra synchronization.

Adds an e2e regression test (time_overflow.parquet: a TIME64[us] column with a 24h value) asserting the query errors gracefully and the server stays alive.


andrejtonev:

---
___

### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Fixed a crash when `LOAD PARQUET` hit invalid values (for example a time outside Memgraph's supported range). The query now fails with an error instead of taking down the server. [#4454](https://github.com/memgraph/memgraph/pull/4454)

as51340 (approved):
Just check if need to clear prefetch_error ?

as51340:
> Just check if need to clear prefetch_error ?

all good here

PR #4460, opened 2026-07-21
feat: add search.node/search.node_all query-module procedures

imilinovic:
Adds `search.node` and `search.node_all` query-module procedures that find nodes by a `{label: property}` map, a comparison operator, and a value.

**What.** `CALL search.node(labelPropertyMap, operator, value) YIELD node` returns nodes whose property satisfies the operator. A property list per label is a disjunction (results unioned). `search.node` deduplicates by node id; `search.node_all` keeps duplicates. Operators: `=`/`exact`, `<>`, `<`, `<=`, `>`, `>=`, `starts with`, `ends with`, `contains`, `=~`. The label-property map is accepted as a map or a JSON string (parsed with nlohmann_json); `value` is a nullable string (null → no rows); operators are trimmed and case-insensitive.

**How.** Each `(label, property)` pair is turned into a Cypher query run through the interpreter (`mgp_execute_query`), so the planner selects the access path (label-property index where available, label scan otherwise) and each search runs in its own transaction. Going through the interpreter also enforces label- and property-level read permissions natively via the evaluator.

**Why.** Delegating to the interpreter is faster than a hand-rolled scan in nearly every case and inherits future planner improvements. Note: the planner does not yet use the label-property index for `STARTS WITH` (implemented in [#4464](https://github.com/memgraph/memgraph/pull/4464)) that operator currently falls back to a label scan + filter.

Kernel-free — no changes outside `mage/` and `config/mappings.json`. An earlier revision added a label-property range mgp C API; that has been split out to #4465 and is not required by these procedures.


imilinovic:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - New `search.node` and `search.node_all` procedures find nodes by a `{label: property}` map, a comparison operator, and a value; a property list per label is OR-ed, `search.node` de-duplicates by node while `search.node_all` keeps duplicates. Use them to look up nodes by property value without writing the equivalent `MATCH`/`WHERE`; an existing label-property index is used automatically. [#4460](https://github.com/memgraph/memgraph/pull/4460)
- [x] **[ Documentation PR link memgraph/documentation#1698 ]**

mage/cpp/search_module/algorithm/search.cpp
```diff
@@ -0,0 +1,191 @@
...
+  if (op == ">") return ">";
+  if (op == ">=") return ">=";
+  if (op == "starts with") return "STARTS WITH";
+  if (op == "ends with") return "ENDS WITH";
+  if (op == "contains") return "CONTAINS";
+  if (op == "=~") return "=~";
+  throw mgp::ValueException(
+      "operator `" + op +
+      "` invalid, it must be one of (case insensitive): [<=, =~, contains, <>, ends with, starts with, exact, <, =, "
+      ">, >=].");
```
as51340: Extract operators to constants

as51340: also you can extract a function from here and Run to check if the operator is string only

mage/cpp/search_module/algorithm/search.cpp
```diff
@@ -0,0 +1,191 @@
...
+    std::unordered_set<int64_t> seen;
+    // A sub-query node accessor is bound to the sub-query's transaction, freed with its result. Re-resolve
+    // each id against the caller's graph so the emitted node outlives that transaction.
+    const auto emit = [&](int64_t node_id) {
+      if (deduplicate && !seen.insert(node_id).second) return;
+      const auto id = mgp::Id::FromInt(node_id);
+      if (!graph.ContainsNode(id)) return;  // skip nodes not visible in the caller's snapshot
+      auto record = record_factory.NewRecord();
+      record.Insert(kResultNode, graph.GetNodeById(id));
+    };
```
as51340, line 167: Do we need both checks?

mage/cpp/search_module/algorithm/search.cpp
```diff
@@ -0,0 +1,191 @@
...
+        // The guard must be a CASE, not `valueType(...) = 'STRING' AND n.p <op> ...`: the planner does not
+        // preserve AND operand order, so the operator can be evaluated before the type check and still throw.
+        // A CASE evaluates only the branch it takes, so the operator is applied to string properties only.
+        std::string predicate = node_property + " " + comparison + " $value";
+        if (string_only) {
+          predicate = "CASE WHEN valueType(" + node_property + ") = 'STRING' THEN " + predicate + " ELSE false END";
+        }
+        const std::string query =
+            "MATCH (n:`" + EscapeIdentifier(label) + "`) WHERE " + predicate + " RETURN id(n) AS node_id";
+
```
as51340: Use fmt::format if possible, this creates a lot of temporary strings

as51340 (changes requested):

mage/cpp/search_module/algorithm/search.cpp
imilinovic: yes, first check is for deduplication,
second is if mgp_execute_query saw something which snapshot isolation of this transaction is preventing

imilinovic: on second thought we do two lookups here (containsnode + getnodebyid), we can reduce to one, good catch

as51340 (approved):

PR #4464, opened 2026-07-22
feat: use label-property index for STARTS WITH prefix seek

imilinovic:
**What:** the query planner now serves `WHERE x.p STARTS WITH v` with an index prefix seek instead of a full scan + `Filter` — for node label-property indexes as well as edge-type-property and global edge-property indexes.

**How:** `x.p STARTS WITH v` is detected during filter preprocessing and lowered to a new `PropertyFilter::Type::PREFIX`. At scan time it produces the half-open range `[v, PrefixSuccessor(v))` restricted to strings (`utils::PrefixSuccessor` drops trailing `0xFF` bytes and increments the last byte; the upper bound is left unset when there is no successor, so the index supplies the string-type upper bound). The range is exact for strings, so the residual `Filter` is dropped, and it is scored between an equality and a bounded range during index selection. Nodes scan via `ScanAllByLabelProperties`; edges via `ScanAllByEdgeTypePropertyRange` / `ScanAllByEdgePropertyRange`, which gain an optional `ExpressionRange`. The `startsWith`/`endsWith`/`contains` predicates now treat a null or non-string operand as a non-match (return `null`) rather than throwing, so index and no-index paths return identical results on mixed-type data; a non-string search term produces an empty range (matches nothing).

**Why:** `STARTS WITH` on an indexed property previously scanned every node/edge and filtered, which is unnecessarily slow when an index exists.

### DIFF
--- a/src/query/interpret/awesome_memgraph_functions.cpp
+++ b/src/query/interpret/awesome_memgraph_functions.cpp
@@ -1189,8 +1189,13 @@ TypedValue Rand(const TypedValue *args, int64_t nargs, const FunctionContext &ct
 
 template <class TPredicate>
 TypedValue StringMatchOperator(const TypedValue *args, int64_t nargs, const FunctionContext &ctx) {
-  FType<Or<Null, String>, Or<Null, String>>(TPredicate::name, args, nargs);
-  if (args[0].IsNull() || args[1].IsNull()) return TypedValue(ctx.memory);
+  if (nargs != 2) {
+    throw QueryRuntimeException("'{}' requires exactly 2 arguments.", TPredicate::name);
+  }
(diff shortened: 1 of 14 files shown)


imilinovic:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - `STARTS WITH` on an indexed string property now uses that index for a prefix seek instead of scanning the whole label/edge-type (label-property, edge-type-property, and global edge-property indexes), so such queries run faster with no query changes needed; `STARTS WITH`/`ENDS WITH`/`CONTAINS` also now return no match (rather than raising an error) when an operand is null or not a string. [#4464](https://github.com/memgraph/memgraph/pull/4464)

PR #4468, opened 2026-07-23
perf(storage): GC takes main_lock_ in READ instead of WRITE (transactional)

andrejtonev:
In `IN_MEMORY_TRANSACTIONAL` mode, `InMemoryStorage::CollectGarbage` now takes `main_lock_` in **READ** instead of **WRITE**. Analytical mode is unchanged (it keeps the WRITE-mode shared hold).

## Why

GC's shared hold defaulted to WRITE, which is mutually exclusive with READ_ONLY. `CREATE INDEX` / `CREATE CONSTRAINT` take READ_ONLY, so a running GC blocked that DDL, and a *pending* READ_ONLY request stalled a fresh GC acquisition (the `ro_pending_count` gate). GC only *reads* the indices and constraints (each independently synchronized and copy-on-write), so READ suffices, and READ is compatible with READ_ONLY: GC and index/constraint DDL no longer wait on each other. This addresses the GC-correlated latency stalls seen under index-heavy workloads.

## Correctness

The storage mode selects the lock mode, so GC must read it without racing a concurrent `SetStorageMode`. The mode is now read only while holding a shared lock: GC takes READ optimistically, reads `storage_mode_` under that hold, and swaps to WRITE if the mode is analytical. Because any shared hold blocks `SetStorageMode`'s UNIQUE acquisition, the mode cannot change while GC holds the lock, so the value GC reads stays valid for the whole pass.

Reading the mode *before* acquiring the lock was wrong in two ways. First, the non-atomic `storage_mode_` was read with no lock held while `SetStorageMode` could be writing it under UNIQUE: a data race (undefined behavior, and flagged by ThreadSanitizer). Second, even setting the race aside, the mode could change between the check and the acquire (a time-of-check to time-of-use gap): GC could read TRANSACTIONAL, decide on READ, and by the time it acquired the lock the store had flipped to analytical, running analytical GC under a READ hold. Reading under the hold closes both.

## Refactor

The lock-mode handling is consolidated behind a new `utils::ResourceLockGuard`: a move-only RAII guard whose runtime mode tag spans UNIQUE plus the three shared modes. It replaces the hand-rolled `OnScopeExit` release in `CollectGarbage`, so one guard always releases in whatever mode it holds. Unit tests cover its constructors, move semantics, `downgrade_to_read`, and `try_lock_for`.

### DIFF
--- a/src/storage/v2/inmemory/storage.cpp
+++ b/src/storage/v2/inmemory/storage.cpp
@@ -2929,50 +2929,36 @@ void InMemoryStorage::SetStorageMode(StorageMode new_storage_mode) {
 }
 
 void InMemoryStorage::CollectGarbage(std::unique_lock<utils::ResourceLock> main_guard, bool periodic) {
-  // NOTE: You do not need to consider cleanup of deleted object that occurred in
-  // different storage modes within the same CollectGarbage call. This is because
-  // SetStorageMode will ensure CollectGarbage is called before any new transactions
-  // with the new storage mode can start.
+  // NOTE: A single call need not handle objects deleted under a different storage mode: SetStorageMode
+  // runs GC before any transaction in the new mode can start.
 
-  // SetStorageMode will pass its unique_lock of main_lock_. We will use that lock,
-  // as reacquiring the lock would cause deadlock. Otherwise, we need to get our own
-  // lock.
-  // GC only reads the indices/constraints (each independently synchronized + COW), so in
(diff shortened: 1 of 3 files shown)


Ignition (approved):

Ignition:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Garbage collection and index/constraint creation no longer block each other in in-memory transactional mode. [#4468](https://github.com/memgraph/memgraph/pull/4468)

PR #4470, opened 2026-07-23
refactor(utils): share contiguous iterator between small_vector and static_vector

seuros:


Ignition (dismissed):

Ignition:
I've been through this on a different PR before, we want strong types to prevent bugs. We do not want `typeof(small_vector::iterator) == typeof(static_vector::iterator)`

PR #4471, opened 2026-07-23
refactor(utils): dedup /proc/meminfo parsing into utils::sysinfo

seuros:


src/utils/system_info.cpp
```diff
@@ -42,24 +43,9 @@ std::string GetMachineId() {
 }
 
 MemoryInfo GetMemoryInfo() {
-  // Parse `/proc/meminfo`.
-  uint64_t memory{0};
-  uint64_t swap{0};
-  auto mem_data = utils::ReadLines("/proc/meminfo");
```
Ignition, line 48: We want to move away from using `/proc/meminfo` + string processing. Use `<sys/sysinfo.h>` instead. For now leave this as is, as it is a single read over `/proc/meminfo`.

seuros: This PR doesn't change it the behaviour.  You have the exact code in 2 places.

This is a PR so when i touch the helper for FreeBSD, you don't have big diffs.

Ignition (changes requested):
I'd rather fix this properly in one PR than have an intermediate step which is worse execution.

ATM we have one pass over a single read of `/proc/meminfo`. This PR changes it to two passes over two reads.

seuros:
ping @Ignition @as51340 .

This is the last commit that clear the path to FreeBSD.

Ignition (approved):

PR #4475, opened 2026-07-23
fix(storage): pin object stores during index-cleanup GC to prevent edge-index use-after-free

### LINKED ISSUE #4473: Engine SIGSEGV in Storage GC — `InMemoryEdgeTypePropertyIndex::RemoveObsoleteEntries` reads a freed edge's PropertyStore under concurrent writes (v3.12.0, IN_MEMORY_TRANSACTIONAL)
Yes, the following bug report was written by Claude Code. We spent a week instrumenting and measuring to get to this point though.

## Summary

Under sustained concurrent write load, the engine intermittently takes a **SIGSEGV in its background `Storage GC` thread** and the whole process dies -- the container reports `Exited (139)` with `oom_killed=false`. It is a genuine engine crash, not a client error and not an OOM: no query returns an error, the process simply vanishes mid-run.

The crash is rare on any single instance but **reproduces reliably across a high-concurrency fleet** (our CI runs dozens of independent ephemeral Memgraph instances in parallel on shared hosts, each under a heavy short-transaction write workload). It survived our upgrade from 3.10 to 3.12.0, but the bug remains albeit with less frequency.

We finally captured a core dump of the crashing engine process and symbolized it. The faulting thread is `Storage GC`, and the stack points squarely at the **edge-type-property index garbage-collection path reading an edge's property store**.

## Backtrace (symbolized, from the core; aarch64, v3.12.0)

```
Program terminated with signal SIGSEGV, Segmentation fault.
Core was generated by `/usr/lib/memgraph/memgraph --log-level WARNING --also-log-to-stderr=true --stor'.
[Current thread is 1 (Thread 0xffff80bf4a40 (LWP 19))]   # the "Storage GC" thread

#0  memgraph::storage::(anonymous namespace)::HasExpectedProperty(Reader*, PropertyId)
#1  memgraph::storage::PropertyStore::IsPropertyEqual(PropertyId, PropertyValue const&) const
#2  memgraph::storage::AnyVersionHasProperty(Edge const&, PropertyId, PropertyValue const&, uint64_t)
#3  memgraph::storage::InMemoryEdgeTypePropertyIndex::RemoveObsoleteEntries(uint64_t, std::stop_token)
#4  memgraph::storage::Indices::RemoveObsoleteEdgeEntries(uint64_t, std::stop_token) const
#5  memgraph::storage::InMemoryStorage::CollectGarbage(std::unique_lock<memgraph::utils::ResourceLock>, bool)
#6  memgraph::storage::InMemoryStorage::InMemoryStorage(...)::$_2  (the GC callback)
#7  std::__invoke_impl<...>(...)
#8  memgraph::storage::InMemoryStorage::InMemoryStorage(...)::$_3
#9  memgraph::utils::Scheduler::ThreadRun(std::string, std::function<void()>, std::stop_token)
#10 std::thread::_State_impl<...>::_M_run()
#11 std::execute_native_thread_routine (thread.cc:104)
#12 start_thread (pthread_create.c:447)
#13 thread_start (clone.S:79)
```

The instruction pointer is at `HasExpectedProperty(...)+44`. Frame #0 faults dereferencing through the `Reader*` into the edge's `PropertyStore` buffer.

## Interpretation

The `Storage GC` scheduler thread runs `InMemoryStorage::CollectGarbage`, which calls `Indices::RemoveObsoleteEdgeEntries`   `InMemoryEdgeTypePropertyIndex::RemoveObsoleteEntries`. To decide whether an index entry is now obsolete, that routine calls `AnyVersionHasProperty(Edge const&, PropertyId, value, timestamp)`, which reads the edge's `PropertyStore` (`IsPropertyEqual`   `HasExpectedProperty`).

The segfault reading that property buffer means the `Edge` (or the property buffer it points at) has been **concurrently freed or reallocated** -- by a committing/aborting transaction mutating/deleting that edge, or by delta GC reclaiming the old version in the same cycle -- while `RemoveObsoleteEntries` still holds a raw reference to it. In other words, a **use-after-free / torn read between edge-type-property-index GC and concurrent edge mutation**: the index-GC path appears to read edge properties without the lock or the "already deleted / already reclaimed" guard that would make that read safe.

This looks analogous to the class of bug fixed for the delta path in PR #4323 ("fix(storage): register committed deltas before MarkFinished", shipped in 3.12.0) -- that change measurably reduced our crash rate but did **not** eliminate this edge-index variant, which still faults on 3.12.0. A natural question is whether the **edge** index's `RemoveObsoleteEntries` takes the same edge-lock / deleted-guard that the **node** label-property index path takes before touching `PropertyStore`   the crash is consistent with the edge path missing it.

## Environment

- **Memgraph v3.12.0**   official `memgraph/memgraph:3.12.0` Docker image (current latest release).
- Storage mode: default **IN_MEMORY_TRANSACTIONAL**.
- OS/arch: Linux, **aarch64** (ARM64).
- Launch flags: `--log-level WARNING --also-log-to-stderr=true --storage-snapshot-interval-sec=86400` (periodic snapshots effectively disabled).
- Indexes in play include **edge-type + property** indexes (the subsystem in the stack).

## How it reproduces for us

- A parallel integration-test suite. Each test process spins up its **own** short-lived Memgraph container and drives it with a high rate of small transactions that `CREATE` / `MERGE` / `DELETE` nodes **and relationships**, with node/edge property writes and edge-type+property indexes, then tears the instance down. Dozens of these run concurrently on one host.
- The crash **only** appears under this concurrency pressure -- it scales with how many engines share a host (i.e. with CPU oversubscription / scheduling jitter), which is the classic trigger for a latent GC-vs-writer race. A single instance under light load does not hit it.
- Frequency: rare per individual instance, but with a large enough concurrent fleet it recurs regularly (it is currently the dominant crash mode we see).

## What we ruled out (measured, not guessed)

- **Not OOM.** Engine RSS stays flat (tens to a few hundred MB). The host reports ~7 10 GB free at crash time, there are **no OOM/kill records in `dmesg`**, and Docker reports `oom_killed=false`. Swap is 0.
- **Not durability.** We disabled WAL, snapshots, and telemetry independently   the engine still exits 139. Only the 3.10 3.12 bump (i.e. PR #4323) measurably helped, and only partially.
- **Not memory-map exhaustion.** `vm.max_map_count` headroom is ~130x the observed map count.
- **No readable fatal-signal trace anywhere.** Despite `--also-log-to-stderr=true`, both `docker logs` (stdout/stderr) **and** the in-container `/var/log/memgraph/*.log` go completely silent ~1 minute before the process exits   the engine writes no crash handler / signal backtrace to any log channel on this crash. The **only** artifact that carried a stack was a core dump of the engine process itself (host-level and per-worker core capture were blind, because the crash is inside the container's PID/mount namespace; we had to bind-mount a core directory into the container and lift its core ulimit to catch it).

I have the **full core dump** (aarch64, v3.12.0, ~650 MB) and can share it, upload additional `gdb` output (`thread apply all bt full`, register/memory dumps around the faulting address, etc.), or provide a reduced reproduction on request. Happy to help however is most useful.
``

strichter:
The index and unique-constraint cleanup sweeps dereference the raw Vertex/Edge pointers in their entries while holding no accessor on vertices_/edges_. A concurrent accessor's run_gc can free an object mid-scan and crash the storage GC thread.

Each sweep now holds an accessor on its store for the scan: nothing present at scan start is reclaimed under it, and since an object's index entries are removed before it leaves the store, every entry the sweep sees is live. Edge sweeps pin via MakeEdgePin, which also covers light edges. The accessor is taken per-sweep at the dereference, only when there is an index or constraint to scan.

Existing skip-list GC tests cover the invariant; the index, constraint, and light-edge suites pass.

### DIFF
--- a/src/storage/v2/indices/indices.cpp
+++ b/src/storage/v2/indices/indices.cpp
@@ -27,19 +27,22 @@
 
 namespace memgraph::storage {
 
-void Indices::RemoveObsoleteVertexEntries(uint64_t oldest_active_start_timestamp, std::stop_token token) const {
-  static_cast<InMemoryLabelIndex *>(label_index_.get())->RemoveObsoleteEntries(oldest_active_start_timestamp, token);
+void Indices::RemoveObsoleteVertexEntries(Storage *storage, uint64_t oldest_active_start_timestamp,
+                                          std::stop_token token) const {
+  static_cast<InMemoryLabelIndex *>(label_index_.get())
+      ->RemoveObsoleteEntries(storage, oldest_active_start_timestamp, token);
   static_cast<InMemoryLabelPropertyIndex *>(label_property_index_.get())
-      ->RemoveObsoleteEntries(oldest_active_start_timestamp, token);
+      ->RemoveObsoleteEntries(storage, oldest_active_start_timestamp, token);
 }
 
-void Indices::RemoveObsoleteEdgeEntries(uint64_t oldest_active_start_timestamp, std::stop_token token) const {
+void Indices::RemoveObsoleteEdgeEntries(Storage *storage, uint64_t oldest_active_start_timestamp,
+                                        std::stop_token token) const {
   static_cast<InMemoryEdgeTypeIndex *>(edge_type_index_.get())
-      ->RemoveObsoleteEntries(oldest_active_start_timestamp, token);
+      ->RemoveObsoleteEntries(storage, oldest_active_start_timestamp, token);
   static_cast<InMemoryEdgeTypePropertyIndex *>(edge_type_property_index_.get())
-      ->RemoveObsoleteEntries(oldest_active_start_timestamp, token);
+      ->RemoveObsoleteEntries(storage, oldest_active_start_timestamp, token);
   static_cast<InMemoryEdgePropertyIndex *>(edge_property_index_.get())
-      ->RemoveObsoleteEntries(oldest_active_start_timestamp, token);
+      ->RemoveObsoleteEntries(storage, oldest_active_start_timestamp, token);
 }
 
 void Indices::RemoveVerticesFromVectorIndices(std::vector<Vertex *> const &vertices_to_remove) const {
--- a/src/storage/v2/indices/indices.hpp
+++ b/src/storage/v2/indices/indices.hpp
@@ -34,6 +34,8 @@ class MemoryTracker;
 
 namespace memgraph::storage {
 
+class Storage;
+
 struct Indices {
   Indices(const Config &config, StorageMode storage_mode, utils::MemoryTracker *db_embedding_memory_tracker = nullptr,
           metrics::GaugeHandle active_label_indices = {}, metrics::GaugeHandle active_label_property_indices = {},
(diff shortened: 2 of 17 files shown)


Ignition (approved):

Ignition:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Fixed a rare garbage-collector crash caused by a concurrency race under heavy concurrent writes when indexes or unique constraints are defined. [#4475](https://github.com/memgraph/memgraph/pull/4475)

PR #4480, opened 2026-07-24
fix(tests): use Abort in Rpc.Abort instead of concurrent Shutdown

seuros:
The Rpc.Abort unit test triggers heap corruption during concurrent Client::Shutdown(). 

This is latent on Linux with glibc malloc (which does not validate freed metadata), but is caught by stricter allocators.

### DIFF
--- a/tests/unit/rpc.cpp
+++ b/tests/unit/rpc.cpp
@@ -74,8 +74,8 @@ TEST(Rpc, Abort) {
 
   std::thread thread([&client]() {
     std::this_thread::sleep_for(100ms);
-    spdlog::info("Shutting down the connection!");
-    client.Shutdown();
+    spdlog::info("Aborting the connection!");
+    client.Abort();
   });
 
   memgraph::utils::Timer const timer;
@@ -83,6 +83,7 @@ TEST(Rpc, Abort) {
   EXPECT_LT(timer.Elapsed(), 200ms);
 
   thread.join();
+  client.Shutdown();
 
   ASSERT_TRUE(server.Shutdown());
   server.AwaitShutdown();


tests/unit/rpc.cpp
```diff
@@ -74,15 +74,16 @@ TEST(Rpc, Abort) {
...
+    spdlog::info("Aborting the connection!");
+    client.Abort();
   });
 
   memgraph::utils::Timer const timer;
   EXPECT_THROW(client.Call<SumV1>(10, 20), RpcFailedException);
   EXPECT_LT(timer.Elapsed(), 200ms);
 
   thread.join();
+  client.Shutdown();
```
as51340, line 86: this is not needed because client.Abort already calls Shutdown

as51340 (changes requested):

tests/unit/rpc.cpp
seuros: That incorrect. 

It calls client_->Shutdown(); (socket-level) instead Client::Shutdown() (object-level)

https://github.com/memgraph/memgraph/blob/master/src/rpc/client.cpp#L35

We could fix by editing client.cpp to add 'client_ = std::nullopt;'

as51340: I agree, that's what I meant, but for the purpose of this test, I don't see why setting client_ to nullopt would help. Anyway both variations (with or without Shutdown) should be fine so I will approve it

as51340 (approved):

tests/unit/rpc.cpp
seuros: @as51340 , i'm doing the FreeBSD compatibility, it is much stricter than current glibc malloc.

Those passes are to avoid having code that diverge between OSes or to present you a diff with thousands of LOCs.

PR #4487, opened 2026-07-27
fix(query): correlate pattern comprehensions and scope CALL subquery variables

### LINKED ISSUE #3955: CALL subquery using an outer variable without WITH returns a single null-grouped row instead of preserving outer rows
**Memgraph version**
Memgraph `3.8.1`

**Environment**
- Host OS: Windows 10
- Architecture: x86_64
- Deployment: Docker
- Connection method: Bolt
- Client used for direct reproduction: Python `neo4j` Bolt driver, `cypher-shell`, and `mgconsole`
- Differential comparison target: Neo4j Community `2026.02.2`

**Describe the bug**
A `CALL { ... }` subquery that references an outer variable `p` without explicitly importing it with `WITH p` is accepted by Memgraph, but the final result collapses the outer grouping column to `null`.

**To Reproduce**
2. Run the following query:

```cypher
MATCH (n) DETACH DELETE n;
CREATE (:Person {name: 'Alice'}), (:Person {name: 'Bob'}), (:Person {name: 'Charlie'});

MATCH (p:Person)
CALL {
  OPTIONAL MATCH (p)-[:ACTED_IN]->(m)
  RETURN m AS m
}
RETURN p.name AS person, count(m) AS c
ORDER BY person;
```
**Expected behavior**
The query should not silently collapse the outer grouping column to `null`.

Reasonable outcomes would be:
- preserve one row per outer `p`, or
- reject the query explicitly if referencing `p` inside the subquery without `WITH p` is not supported

On Neo4j, the query returns:

```text
Alice, 0
Bob, 0
Charlie, 0
```
**Actual behavior**
On Memgraph, the query succeeds but returns:

```text
person = null, c = 0
```

Observed CSV output from `mgconsole`:

```text
"person","c"
"Null","0"
```
**Logs**
Memgraph does not raise an explicit query error for the reproducer. It executes successfully but returns a single null-grouped row..

**Verification Environment**
Plain Memgraph Docker image is enough.
Preferred verification target:
- Docker image
- x86_64
- `memgraph/memgraph`
I can verify a fix using:
- plain Memgraph Docker image on Windows host
- Bolt connection using `mgconsole` and Python Bolt driver

### LINKED ISSUE #4160: Deprecated CALL write subquery without explicit import may nullify outer variables afterward
**Memgraph version**
Observed on:

- Memgraph `3.9.0`
- Memgraph `3.10.1`

Docker images used locally:
- `memgraph/memgraph:3.9.0`
- `memgraph/memgraph:3.10.1` 

**Environment**
- Host OS: Windows 10
- Architecture: x86_64
- Deployment: Docker
- Connection method: Bolt
- Client used for direct reproduction: Python `neo4j` Bolt driver

**Describe the bug**
Memgraph may nullify an outer variable after a deprecated-style `CALL { ... }` subquery if:

- the outer variable is carried into the subquery implicitly from the surrounding pipeline,
- the subquery body performs a write, and
- the subquery does **not** explicitly re-import the outer variable with `WITH`.

In the minimized repro below, Neo4j returns `Alice`. Memgraph instead returns `null`, even though the outer variable `a` was already bound before entering the subquery.

This does not look like a general problem with `CALL { ... }`, because Memgraph behaves normally if the subquery is read-only, or if the write subquery explicitly imports `a` (and `b`) with `WITH`.

**To Reproduce**
2. Run:

```cypher
MATCH (n) DETACH DELETE n;
CREATE (:Person {name:'Alice'});
CREATE (:Person {name:'Bob'});

MATCH (a:Person {name:'Alice'})
MATCH (b:Person {name:'Bob'})
WITH a, b
CALL {
  CREATE (a)-[:LIKES]->(b)
}
RETURN a.name AS name;
```

**Expected behavior**
Observed Neo4j result:

```text
name = Alice
```

Memgraph should return the same value.

**Actual behavior**
Observed Memgraph result:

```text
name = null
```

So the outer variable remains present as a row, but its value is lost after the write subquery.

**Control query 1**
If the same write is done without `CALL`, both engines behave normally:

```cypher
MATCH (n) DETACH DELETE n;
CREATE (:Person {name:'Alice'});
CREATE (:Person {name:'Bob'});

MATCH (a:Person {name:'Alice'})
MATCH (b:Person {name:'Bob'})
CREATE (a)-[:LIKES]->(b)
RETURN a.name AS name;
```

```text
name = Alice
```

So the problem is not simply “write followed by return of a bound node”.

**Control query 2**
If the subquery explicitly re-imports the outer variables, both engines also behave normally:

```cypher
MATCH (n) DETACH DELETE n;
CREATE (:Person {name:'Alice'});
CREATE (:Person {name:'Bob'});

MATCH (a:Person {name:'Alice'})
MATCH (b:Person {name:'Bob'})
WITH a, b
CALL {
  WITH a, b
  CREATE (a)-[:LIKES]->(b)
}
RETURN a.name AS name;
```

```text
name = Alice
```

This makes the boundary much sharper: the failure is tied to the deprecated implicit-import form, not to the write itself.

**Control query 3**
If the deprecated-style subquery is read-only, both engines also behave normally:

```cypher
MATCH (n) DETACH DELETE n;
CREATE (:Person {name:'Alice'});
CREATE (:Person {name:'Bob'});

MATCH (a:Person {name:'Alice'})
MATCH (b:Person {name:'Bob'})
WITH a, b
CALL {
  RETURN 1 AS x
}
RETURN a.name AS name;
```

```text
name = Alice
```

So the issue is not a blanket inability to preserve outer variables across deprecated-style `CALL`.

**Additional variant**
The same outer-variable nullification also appears with a single outer variable and a self-relationship write:

```cypher
MATCH (n) DETACH DELETE n;
CREATE (:Person {name:'Alice'});

MATCH (a:Person {name:'Alice'})
WITH a
CALL {
  CREATE (a)-[:SELF]->(a)
}
RETURN a.name AS name;
```

Observed Neo4j result:

```text
name = Alice
```

Observed Memgraph result:

```text
name = null
```

This suggests the issue is not specific to the presence of a second variable `b`.

### LINKED ISSUE #4161: Inline size(pattern-comprehension) filter after WITH may ignore correlation and admit unrelated rows
**Memgraph version**
Observed on:

- Memgraph `3.9.0`
- Memgraph `3.10.1` 

Docker images used locally:
- `memgraph/memgraph:3.9.0`
- `memgraph/memgraph:3.10.1` 

**Environment**
- Host OS: Windows 10
- Architecture: x86_64
- Deployment: Docker
- Connection method: Bolt
- Client used for direct reproduction: Python `neo4j` Bolt driver

**Describe the bug**
Memgraph may evaluate a correlated filter incorrectly when:

- an earlier `MATCH` already bound `p`
- `p` is carried through a `WITH`
- the later `WHERE` uses an inline `size([...]) > 0` check over a correlated pattern-comprehension-like list expression

In the minimized repro below, Neo4j returns only `Bob`, because only `Bob` is adjacent to a `Person` whose name starts with `A`.

Memgraph instead returns:

```text
Alice
Bob
Charlie
```

This suggests the inline filter is being evaluated as if the correlation on `p` had been lost, even though the same `size(...)` expression computes the correct numeric values on both engines when returned directly.

**To Reproduce**
2. Run:

```cypher
MATCH (n) DETACH DELETE n;
CREATE (:Person {name:'Alice'})-[:KNOWS]->(:Person {name:'Bob'})-[:KNOWS]->(:Person {name:'Charlie'});

MATCH (p:Person)-[:KNOWS]-(q:Person)
WITH p
WHERE size([(p)-[:KNOWS]-(r:Person) WHERE r.name STARTS WITH 'A' | r]) > 0
RETURN DISTINCT p.name AS p
ORDER BY p;
```

**Expected behavior**
Observed Neo4j result:

```text
Bob
```

Memgraph should return the same row.

**Actual behavior**
Observed Memgraph result:

```text
Alice
Bob
Charlie
```

So rows that should fail the correlated filter are still kept.

**Control query 1**
If the same correlated filter is used directly on `MATCH (p:Person)` without the earlier `MATCH ... WITH p` pipeline, both engines behave correctly:

```cypher
MATCH (n) DETACH DELETE n;
CREATE (:Person {name:'Alice'})-[:KNOWS]->(:Person {name:'Bob'})-[:KNOWS]->(:Person {name:'Charlie'});

MATCH (p:Person)
WHERE size([(p)-[:KNOWS]-(r:Person) WHERE r.name STARTS WITH 'A' | r]) > 0
RETURN p.name AS p
ORDER BY p;
```

```text
Bob
```

So the inline `size(...) > 0` expression is not broken in complete generality.

**Control query 2**
If the `size(...)` value is first materialized in `WITH` and only then filtered, both engines also behave correctly:

```cypher
MATCH (n) DETACH DELETE n;
CREATE (:Person {name:'Alice'})-[:KNOWS]->(:Person {name:'Bob'})-[:KNOWS]->(:Person {name:'Charlie'});

MATCH (p:Person)-[:KNOWS]-(q:Person)
WITH p, size([(p)-[:KNOWS]-(r:Person) WHERE r.name STARTS WITH 'A' | r]) AS c
WHERE c > 0
RETURN DISTINCT p.name AS p
ORDER BY p;
```

```text
Bob
```

This makes the boundary much sharper: the problem is tied to the **inline** filter form, not to the `size(...)` computation itself.

**Control query 3**
If the list itself is first materialized and then tested with `hits <> []`, both engines also line up:

```cypher
MATCH (n) DETACH DELETE n;
CREATE (:Person {name:'Alice'})-[:KNOWS]->(:Person {name:'Bob'})-[:KNOWS]->(:Person {name:'Charlie'});

MATCH (p:Person)-[:KNOWS]-(q:Person)
WITH p, [(p)-[:KNOWS]-(r:Person) WHERE r.name STARTS WITH 'A' | r] AS hits
WHERE hits <> []
RETURN DISTINCT p.name AS p
ORDER BY p;
```

```text
Bob
```

So Memgraph can handle the correlated pattern expression once the list result is stabilized before filtering.

### LINKED ISSUE #4180: OPTIONAL MATCH + collect(...) followed by ORDER BY size([pattern comprehension]) may trigger internal "expected a list, got null" behavior
**Memgraph version**
Observed on:

- Memgraph `3.10.1`

**Environment**
- Host OS: Windows
- Architecture: x86_64
- Deployment: Docker
- Connection method: Bolt
- Client used for direct reproduction: Python `neo4j` Bolt driver

**Describe the bug**
Memgraph can raise an internal-style error when a query combines:

- an `OPTIONAL MATCH`
- an aggregate `collect(...)` in `WITH`
- and an `ORDER BY size([pattern comprehension])`

even though nearby variants of the same query shape behave normally.

The following small query currently fails on Memgraph `3.10.1`:

```cypher
MATCH (n) DETACH DELETE n;
CREATE (a:A {name: 'Alice', age: 30}), (b:B {name: 'Bob', age: 25}), (c:C {name: 'Charlie', age: 35}), (d:D {name: 'Diana', age: 28}), (e:E {name: 'Eve', age: 32});
CREATE (a)-[:KNOWS]->(b), (a)-[:KNOWS]->(c), (b)-[:KNOWS]->(d), (c)-[:KNOWS]->(e), (d)-[:KNOWS]->(e);
CREATE (a)-[:FRIEND]->(c), (b)-[:FRIEND]->(e);

MATCH (n:A)
OPTIONAL MATCH (n)-[:KNOWS]->(m)
WITH n, collect(m.name) AS friends
ORDER BY size([(n)-[:FRIEND]->(f) | f.name]) DESC
RETURN n.name AS name, n.age AS age, friends
```

Observed Memgraph error:

```text
Unexpected behavior: Pattern Comprehension expected a list, got null. Please report the problem on GitHub issues
```

The error text itself already says this is unexpected behavior and asks for a bug report.

**To Reproduce**
Run:

```cypher
MATCH (n) DETACH DELETE n;
CREATE (a:A {name: 'Alice', age: 30}), (b:B {name: 'Bob', age: 25}), (c:C {name: 'Charlie', age: 35}), (d:D {name: 'Diana', age: 28}), (e:E {name: 'Eve', age: 32});
CREATE (a)-[:KNOWS]->(b), (a)-[:KNOWS]->(c), (b)-[:KNOWS]->(d), (c)-[:KNOWS]->(e), (d)-[:KNOWS]->(e);
CREATE (a)-[:FRIEND]->(c), (b)-[:FRIEND]->(e);

MATCH (n:A)
OPTIONAL MATCH (n)-[:KNOWS]->(m)
WITH n, collect(m.name) AS friends
ORDER BY size([(n)-[:FRIEND]->(f) | f.name]) DESC
RETURN n.name AS name, n.age AS age, friends
```

**Expected behavior**
The query should not fail with internal-style "unexpected behavior" messaging.

On Neo4j `2026.04.0`, the same query returns:

```text
name = Alice, age = 30, friends = []
```

If Memgraph considers some part of this shape unsupported, it should return a normal user-facing semantic/runtime error instead of an internal-style pattern-comprehension failure.

**Actual behavior**
Observed on Memgraph `3.10.1`:

```text
Unexpected behavior: Pattern Comprehension expected a list, got null. Please report the problem on GitHub issues
```

**Control query 1**
Pattern comprehension by itself behaves normally:

```cypher
MATCH (n) DETACH DELETE n;
CREATE (a:A {name: 'Alice', age: 30}), (c:C {name: 'Charlie', age: 35});
CREATE (a)-[:FRIEND]->(c);

MATCH (n:A)
RETURN size([(n)-[:FRIEND]->(f) | f.name]) AS cnt
```

Observed Memgraph result:

```text
cnt = 0
```

So this is not simply "pattern comprehension is unsupported".

**Control query 2**
If the `OPTIONAL MATCH + collect(...)` is removed, ordering by the same pattern-comprehension expression works:

```cypher
MATCH (n) DETACH DELETE n;
CREATE (a:A {name: 'Alice', age: 30}), (b:B {name: 'Bob', age: 25}), (c:C {name: 'Charlie', age: 35}), (d:D {name: 'Diana', age: 28}), (e:E {name: 'Eve', age: 32});
CREATE (a)-[:KNOWS]->(b), (a)-[:KNOWS]->(c), (b)-[:KNOWS]->(d), (c)-[:KNOWS]->(e), (d)-[:KNOWS]->(e);
CREATE (a)-[:FRIEND]->(c), (b)-[:FRIEND]->(e);

MATCH (n:A)
WITH n
ORDER BY size([(n)-[:FRIEND]->(f) | f.name]) DESC
RETURN n.name AS name, n.age AS age
```

Observed Memgraph result:

```text
name = Alice, age = 30
```

**Control query 3**
If the `ORDER BY size([pattern comprehension])` is removed, the `OPTIONAL MATCH + collect(...)` shape also behaves normally:

```cypher
MATCH (n) DETACH DELETE n;
CREATE (a:A {name: 'Alice', age: 30}), (b:B {name: 'Bob', age: 25}), (c:C {name: 'Charlie', age: 35}), (d:D {name: 'Diana', age: 28}), (e:E {name: 'Eve', age: 32});
CREATE (a)-[:KNOWS]->(b), (a)-[:KNOWS]->(c), (b)-[:KNOWS]->(d), (c)-[:KNOWS]->(e), (d)-[:KNOWS]->(e);
CREATE (a)-[:FRIEND]->(c), (b)-[:FRIEND]->(e);

MATCH (n:A)
OPTIONAL MATCH (n)-[:KNOWS]->(m)
WITH n, collect(m.name) AS friends
RETURN n.name AS name, n.age AS age, friends
```

Observed Memgraph result:

```text
name = Alice, age = 30, friends = []
```

This boundary makes the trigger much sharper:

- pattern comprehension alone: OK
- `OPTIONAL MATCH + collect(...)` alone: OK
- both combined via `WITH ... ORDER BY size([pattern comprehension])`: internal-style failure

### LINKED ISSUE #4339: Pattern comprehension `size()` in WHERE clause silently drops all rows with cartesian product
**Memgraph version**
- Memgraph 3.11.0 (Docker)

**Environment**
- Host OS: Linux
- Deployment: Docker (official image)
- Endpoint: Bolt (`bolt://127.0.0.1:17689`)
- Differential comparison targets:
  - Neo4j 2026.05.0 (correct)
  - Neo4j 5.26.28 (correct)
  - ArcadeDB 26.7.1 (correct)

**Describe the bug**
When two independent `MATCH` clauses produce a cartesian product, and `size([(a)-[:R]->(b) | 1])` is used as a WHERE filter, Memgraph silently filters out ALL rows — even rows that should pass the condition.

Per-pair diagnostics confirm that `size([... | 1])` returns correct values when queried individually for each (a,b) pair. The bug only manifests in bulk cartesian-product execution.

**To Reproduce**
```cypher
MATCH (n) DETACH DELETE n;

CREATE
  (a1:User {id: 1, active: true}),
  (a2:User {id: 2, active: true}),
  (b1:User {id: 3, active: false}),
  (b2:User {id: 4, active: false});

CREATE (a1)-[:FRIEND]->(b1);

-- The cartesian product has 4 pairs: (a1,b1), (a1,b2), (a2,b1), (a2,b2)
-- (a1,b1) has an existing FRIEND edge; the other 3 do not
-- Expected: 3 MONITOR edges created (all pairs except the one with existing FRIEND)

MATCH (a:User {active: true})
MATCH (b:User {active: false})
WITH a, b
WHERE size([(a)-[:FRIEND]->(b) | 1]) = 0
CREATE (a)-[:MONITOR]->(b);

MATCH ()-[r:MONITOR]->()
RETURN count(r) AS cnt;
```

**Expected behavior**
```cypher
MATCH (n) DETACH DELETE n;

CREATE
  (a1:User {id: 1, active: true}),
  (a2:User {id: 2, active: true}),
  (b1:User {id: 3, active: false}),
  (b2:User {id: 4, active: false});

CREATE (a1)-[:FRIEND]->(b1);

-- The cartesian product has 4 pairs: (a1,b1), (a1,b2), (a2,b1), (a2,b2)
-- (a1,b1) has an existing FRIEND edge; the other 3 do not
-- Expected: 3 MONITOR edges created (all pairs except the one with existing FRIEND)

MATCH (a:User {active: true})
MATCH (b:User {active: false})
WITH a, b
WHERE size([(a)-[:FRIEND]->(b) | 1]) = 0
CREATE (a)-[:MONITOR]->(b);

MATCH ()-[r:MONITOR]->()
RETURN count(r) AS cnt;
```
### Actual behavior

```text
cnt = 0
```

No MONITOR edges are created — the WHERE clause eliminated all 4 rows, including the 3 that should have passed `size(...) = 0`.

### Control case 1: per-pair check works correctly

```cypher
MATCH (a:User {id: 1}) MATCH (b:User {id: 4})
RETURN size([(a)-[:FRIEND]->(b) | 1]) AS cnt;
```

Returns `0` as expected (no edge from a1 to b2).

### Control case 2: OPTIONAL MATCH approach works correctly

```cypher
MATCH (n) DETACH DELETE n;

CREATE
  (a1:User {id: 1, active: true}),
  (a2:User {id: 2, active: true}),
  (b1:User {id: 3, active: false}),
  (b2:User {id: 4, active: false});
CREATE (a1)-[:FRIEND]->(b1);

MATCH (a:User {active: true})
MATCH (b:User {active: false})
OPTIONAL MATCH (a)-[r:FRIEND]->(b)
WITH a, b, r
WHERE r IS NULL
CREATE (a)-[:MONITOR]->(b);

MATCH ()-[r:MONITOR]->()
RETURN count(r) AS cnt;
```

Returns `3` (optically equivalent logic using `OPTIONAL MATCH` works correctly).

**Logs**
No server-side error logged. The query executes successfully and returns `(empty result)` / `0`.

```
memgraph> MATCH (a:User {active: true}) MATCH (b:User {active: false})
         WITH a, b WHERE size([(a)-[:FRIEND]->(b) | 1]) = 0
         CREATE (a)-[:MONITOR]->(b);
Empty set (0 rows)

memgraph> MATCH ()-[r:MONITOR]->() RETURN count(r) AS cnt;
+-------+
| cnt   |
+-------+
| 0     |
+-------+
```

**Additional context**
- Per-pair `size([... | 1])` evaluates correctly when queried individually.
- The same semantic expressed via `OPTIONAL MATCH ... WHERE r IS NULL` works correctly.
- Neo4j (both 2026.05 and 5.26) and ArcadeDB all produce the correct result of 3.
- The failure manifests only in Memgraph's cartesian-product batch execution context.

**Verification Environment**
- Plain Memgraph Docker image (x86_64)
- Same repro above: `CREATE` + `MATCH` + `WHERE size(...)` + `CREATE` → verify MONITOR count
- Also verify Neo4j 2026.05 / 5.26 returns `cnt = 3` on the same repro

### LINKED ISSUE #4350: Pattern comprehension guard with outer MATCH variables incorrectly filters all rows before CREATE
**Memgraph version**

Observed on:

```text
memgraph version 3.11.0
```

This was confirmed inside the running Docker container with:

```bash
docker exec memgraph_test sh -c "echo 'SHOW VERSION;' | mgconsole --no-history"
# -> version "3.11.0"
```

**Environment**

- Host OS: Linux 5.4.0
- Architecture: x86_64
- Deployment: Docker (`memgraph/memgraph:latest`)
- Connection method: Bolt
- Endpoint used locally: `bolt://127.0.0.1:17689`
- Driver: Neo4j Python driver
- Authentication: disabled in the local test instance
- Differential comparison targets:
  - Neo4j Community `2026.05.0`
  - Neo4j Community `5.26.28`

**Describe the bug**

A pattern comprehension used as a guard in a CREATE context returns an incorrect
filter result. `WHERE size([(a)-[:LINK]->(b) | 1]) = 0` — which should pass rows
where no `:LINK` edge exists between `a` and `b` — instead filters out ALL rows,
causing zero edges to be created when edges should have been created.

Six other semantically-equivalent guard mechanisms all produce the correct result
on the same Memgraph instance and graph. Only the pattern comprehension guard fails.

This is a silent wrong-result bug: the query succeeds, but the guard evaluates
to false for every row, causing the CREATE to be silently skipped.

**To Reproduce**

Run the following Cypher from a clean database:

```cypher
MATCH (n) DETACH DELETE n;

CREATE
  (a1:Node {id: 1, active: true}),
  (a2:Node {id: 2, active: true}),
  (a3:Node {id: 3, active: false}),
  (a3)-[:LINK]->(a1);
// Graph: 1 existing :LINK edge (a3→a1).
// Active nodes: a1 (id=1), a2 (id=2).
// Active-active pairs without edge: (1→2), (2→1) = 2 pairs.
// Expected after CREATE: 1 existing + 2 new = 3 total :LINK edges.

// --- Failing query: pattern comprehension guard ---
MATCH (a:Node {active: true}), (b:Node {active: true})
WHERE a.id <> b.id
WITH a, b
WHERE size([(a)-[:LINK]->(b) | 1]) = 0
CREATE (a)-[:LINK]->(b);

MATCH ()-[r:LINK]->() RETURN count(r) AS total;
```

Actual Memgraph result:

```text
total = 1
```

Expected (based on Neo4j 5.26.28, 2026.05.0, and the NOT EXISTS guard on Memgraph):

```text
total = 3
```

**Expected behavior**

For each active-active pair `(a, b)` where `a.id <> b.id`, the pattern comprehension
`[(a)-[:LINK]->(b) | 1]` should evaluate against the graph with `a` and `b` bound
to the current row's values. The guard `size(...) = 0` should pass rows where no
`:LINK` edge exists.

In the graph above, pairs `(a1, a2)` and `(a2, a1)` both have no existing `:LINK`
edge. The guard should pass both. 2 new edges should be created, for a total of 3.

Neo4j Community `5.26.28` creates 3 edges for the exact same query. The NOT EXISTS
guard on Memgraph itself also creates 3 edges (see below). Only the pattern
comprehension guard fails.

**Control queries (pass on Memgraph)**

All six equivalent guard mechanisms produce the correct result (3 total edges)
on the same Memgraph instance:

```cypher
// NOT EXISTS subquery — CORRECT
MATCH (a:Node {active: true}), (b:Node {active: true})
WHERE a.id <> b.id
  AND NOT EXISTS { MATCH (a)-[:LINK]->(b) }
CREATE (a)-[:LINK]->(b);
// → total = 3 ✓

// OPTIONAL MATCH + IS NULL — CORRECT
MATCH (a:Node {active: true}), (b:Node {active: true})
WHERE a.id <> b.id
OPTIONAL MATCH (a)-[r:LINK]->(b)
WITH a, b, r WHERE r IS NULL
CREATE (a)-[:LINK]->(b);
// → total = 3 ✓

// NOT pattern — CORRECT
MATCH (a:Node {active: true}), (b:Node {active: true})
WHERE a.id <> b.id AND NOT ((a)-[:LINK]->(b))
CREATE (a)-[:LINK]->(b);
// → total = 3 ✓

// FOREACH + NOT EXISTS — CORRECT
MATCH (a:Node {active: true}), (b:Node {active: true})
WHERE a.id <> b.id AND NOT EXISTS { MATCH (a)-[:LINK]->(b) }
WITH collect([a, b]) AS pairs
FOREACH (p IN pairs | CREATE (p[0])-[:LINK]->(p[1]));
// → total = 3 ✓

// CALL (a, b) subquery — CORRECT
MATCH (a:Node {active: true}), (b:Node {active: true})
WHERE a.id <> b.id AND NOT EXISTS { MATCH (a)-[:LINK]->(b) }
CALL (a, b) { CREATE (a)-[:LINK]->(b) };
// → total = 3 ✓

// CALL { WITH a, b } (deprecated) — CORRECT
MATCH (a:Node {active: true}), (b:Node {active: true})
WHERE a.id <> b.id AND NOT EXISTS { MATCH (a)-[:LINK]->(b) }
CALL { WITH a, b CREATE (a)-[:LINK]->(b) };
// → total = 3 ✓
```

**Logs**

No crash or server-side error was observed. The query executes successfully and
returns a wrong result.

Observed local output:

```text
Memgraph NOT EXISTS guard:      total = 3 ✓
Memgraph pattern comp guard:    total = 1 ✗
Neo4j-5 pattern comp guard:     total = 3 ✓
```

**Additional context**

The pattern comprehension `[(a)-[:LINK]->(b) | 1]` should be evaluated once for
each `(a, b)` row with `a` and `b` bound to the current row. The fact that `0`
edges are created (rather than `2`) suggests the inner pattern is not correctly
binding `a` and `b` from the outer scope — it may be evaluating against unbound
or incorrectly scoped variables.

This may be related to issue #4335 (nested pattern comprehension inside
list/reduce returns global counts), but the triggering mechanism differs:

| | #4335 | This issue |
|---|---|---|
| Variable source | list comprehension (`x IN xs`) | MATCH clause (`a`, `b`) |
| Context | RETURN / reduce expression | CREATE WHERE guard |
| Manifestation | wrong count returned | guard filters all rows |

If the root cause is the same, this reproducer provides a write-context variant
that may help isolate the fix. If the root cause differs, this is a standalone bug.

**Verification Environment**

A plain Memgraph Docker image (`memgraph/memgraph:latest`) is sufficient to verify
the fix. The reproducer only requires Cypher over Bolt and does not require the
full Memgraph platform.

### LINKED ISSUE #4364: Wrong empty result for WITH variable followed by WHERE size(pattern-comprehension WHERE ...) = 0
### Memgraph version

`3.11.0`

### Environment

- Host OS: Linux `5.4.0-216-generic`, x86_64
- Deployment: Docker `memgraph/memgraph:latest`
- Query interface: Bolt, Python driver `neo4j` 5.x
- Query language: Cypher
- Differential comparison target: Neo4j Community `2026.05.0`

### Description

Memgraph returns the wrong result when `size()` over a pattern comprehension with an inner `WHERE` filter is used as a `= 0` check after a bare `WITH <variable>`.

The failing shape is:

```cypher
MATCH (u:User {active:true})
WITH u
WHERE size([(u)-[:FRIEND]->(f:User) WHERE f.active IS NULL OR f.active = true | f]) = 0
WITH u
ORDER BY u.id
RETURN collect(u.id) AS ids;
```

Expected:

```text
[1, 3]
```

Actual on Memgraph `3.11.0`:

```text
[]
```

The same expression works correctly in other equivalent contexts:

| Form | Memgraph result |
|---|---|
| Direct `MATCH ... WHERE size(PC WHERE ...) = 0` without `WITH` | `[1, 3]` |
| `WITH u, size(...) AS bad_count WHERE bad_count = 0` | `[1, 3]` |
| `WITH u WHERE size(PC WHERE ...) > 0` | `[2]` |
| `RETURN size(PC WHERE ...)` diagnostic | `0, 1, 0` |

So the failure appears specific to:

```cypher
WITH u
WHERE size(pattern_comprehension_with_inner_where) = 0
```

### To reproduce

Run from a clean database.

```cypher
MATCH (n) DETACH DELETE n;

CREATE
  (u1:User {id:1, active:true}),
  (u2:User {id:2, active:true}),
  (u3:User {id:3, active:true}),
  (u4:User {id:4, active:false}),
  (u5:User {id:5, active:false}),
  (u6:User {id:6, active:false}),
  (u1)-[:FRIEND]->(u4),
  (u1)-[:FRIEND]->(u5),
  (u2)-[:FRIEND]->(u4),
  (u2)-[:FRIEND]->(u1),
  (u3)-[:FRIEND]->(u5),
  (u3)-[:FRIEND]->(u6);
```

Graph meaning:

- `u1`: friends are `u4(false)`, `u5(false)` → should match
- `u2`: friends are `u4(false)`, `u1(true)` → should not match
- `u3`: friends are `u5(false)`, `u6(false)` → should match

The query is intended to find active users with no friend whose `active` value is `NULL` or `true`.

So the expected result is:

```text
[1, 3]
```

### Read-only diagnostic query

```cypher
MATCH (u:User {active:true})
RETURN
  u.id AS id,
  [(u)-[:FRIEND]->(f:User) | f.id] AS friends,
  [(u)-[:FRIEND]->(f:User) WHERE f.active IS NULL OR f.active = true | f.id] AS bad_friends,
  size([(u)-[:FRIEND]->(f:User) WHERE f.active IS NULL OR f.active = true | f]) AS bad_count
ORDER BY id;
```

Expected result:

```text
id | friends | bad_friends | bad_count
---+---------+-------------+----------
1  | [4, 5]  | []          | 0
2  | [4, 1]  | [1]         | 1
3  | [5, 6]  | []          | 0
```

Neo4j `2026.05.0` returns this expected result.

Memgraph `3.11.0` also computes this diagnostic correctly.

### Failing query

```cypher
MATCH (u:User {active:true})
WITH u
WHERE size([(u)-[:FRIEND]->(f:User) WHERE f.active IS NULL OR f.active = true | f]) = 0
WITH u
ORDER BY u.id
RETURN collect(u.id) AS ids;
```

### Expected behavior

```text
[1, 3]
```

Neo4j `2026.05.0` returns:

```text
[1, 3]
```

### Actual behavior

Memgraph `3.11.0` returns:

```text
[]
```

All rows are filtered out, even though the same `size(...)` expression evaluates to `0` for users `1` and `3` in the diagnostic query.

### Control 1: direct WHERE without WITH works

```cypher
MATCH (u:User {active:true})
WHERE size([(u)-[:FRIEND]->(f:User) WHERE f.active IS NULL OR f.active = true | f]) = 0
WITH u
ORDER BY u.id
RETURN collect(u.id) AS ids;
```

Memgraph result:

```text
[1, 3]
```

### Control 2: WITH alias works

```cypher
MATCH (u:User {active:true})
WITH u, size([(u)-[:FRIEND]->(f:User) WHERE f.active IS NULL OR f.active = true | f]) AS bad_count
WHERE bad_count = 0
WITH u
ORDER BY u.id
RETURN collect(u.id) AS ids;
```

Memgraph result:

```text
[1, 3]
```

### Control 3: positive check works

```cypher
MATCH (u:User {active:true})
WITH u
WHERE size([(u)-[:FRIEND]->(f:User) WHERE f.active IS NULL OR f.active = true | f]) > 0
WITH u
ORDER BY u.id
RETURN collect(u.id) AS ids;
```

Memgraph result:

```text
[2]
```

### Control 4: simple WHERE after WITH works

```cypher
MATCH (u:User {active:true})
WITH u
WHERE u.id IN [1, 3]
WITH u
ORDER BY u.id
RETURN collect(u.id) AS ids;
```

Memgraph result:

```text
[1, 3]
```

### Workaround

Compute the pattern-comprehension size as a `WITH` alias, then filter by the alias:

```cypher
MATCH (u:User {active:true})
WITH u, size([(u)-[:FRIEND]->(f:User) WHERE f.active IS NULL OR f.active = true | f]) AS bad_count
WHERE bad_count = 0
SET u.flagged = true;
```

### Cross-engine comparison

| Form | Neo4j `2026.05.0` | Memgraph `3.11.0` |
|---|---:|---:|
| Direct `WHERE size(PC WHERE ...) = 0` | `[1, 3]` | `[1, 3]` |
| `WITH u, size(...) AS bad_count WHERE bad_count = 0` | `[1, 3]` | `[1, 3]` |
| `WITH u WHERE size(PC WHERE ...) > 0` | `[2]` | `[2]` |
| `WITH u WHERE size(PC WHERE ...) = 0` | `[1, 3]` | `[]` |

### Why this looks like a bug

1. The reproducer is small: 6 nodes and 6 relationships.
2. The expression `size(PC WHERE ...)` is computed correctly in `RETURN`, direct `WHERE`, and `WITH ... AS` contexts.
3. Only the bare `WITH u` followed by `WHERE size(PC WHERE ...) = 0` path returns the wrong result.
4. A trivial alias workaround restores correct behavior.
5. Neo4j returns the expected result for all equivalent forms.
6. The failure is deterministic on Memgraph `3.11.0`.

### Suspected issue

The planner or expression evaluator appears to mishandle this specific form:

```cypher
WITH u
WHERE size([(u)-[:FRIEND]->(f:User) WHERE ... | f]) = 0
```

The same pattern-comprehension expression is evaluated correctly when materialized as an alias first:

```cypher
WITH u, size([(u)-[:FRIEND]->(f:User) WHERE ... | f]) AS bad_count
WHERE bad_count = 0
```

### LINKED ISSUE #4365: Pattern comprehension ignores bound endpoint variable and returns non-empty list for non-matching pairs
### Memgraph version

`3.11.0`

### Environment

- Host OS: Linux `5.4.0-216-generic`, x86_64
- Deployment: Docker `memgraph/memgraph:latest`
- Query interface: Bolt, Python driver `neo4j` 5.x
- Query language: Cypher
- Differential comparison target: Neo4j Community `2026.05.0`

### Description

Memgraph returns the wrong result when `size()` is used over a pattern comprehension whose pattern uses already-bound outer variables as both endpoints.

The failing expression is:

```cypher
size([(a)-[:FRIEND]->(b) | 1])
```

Here, both `a` and `b` are already bound by the outer `MATCH`:

```cypher
MATCH (a:User {active:true}), (b:User {active:false})
```

For each `(a, b)` row, the expression should return:

- `1` if that specific `a` has a `:FRIEND` edge to that specific `b`
- `0` if no such edge exists

However, on Memgraph `3.11.0`, the pattern comprehension returns `[1]` for every `(a, b)` row, even when no such relationship exists.

As a result:

```cypher
size([(a)-[:FRIEND]->(b) | 1]) = 0
```

is never true.

The equivalent direct predicates work correctly on Memgraph:

```cypher
WHERE NOT (a)-[:FRIEND]->(b)
```

and:

```cypher
WHERE NOT exists((a)-[:FRIEND]->(b))
```

This isolates the issue to the pattern-comprehension form.

### To reproduce

Run from a clean database.

```cypher
MATCH (n) DETACH DELETE n;

CREATE
  (a1:User {id:1, active:true}),
  (a2:User {id:2, active:true}),
  (b1:User {id:3, active:false}),
  (b2:User {id:4, active:false}),
  (a1)-[:FRIEND]->(b1);
```

The graph has:

- two active users: `a1`, `a2`
- two inactive users: `b1`, `b2`
- one `:FRIEND` edge: `(a1)-[:FRIEND]->(b1)`

Therefore, among the four active-to-inactive pairs:

| Pair | `:FRIEND` exists? |
|---|---|
| `(a1, b1)` | yes |
| `(a1, b2)` | no |
| `(a2, b1)` | no |
| `(a2, b2)` | no |

So only one row should produce a non-empty pattern-comprehension result.

### Read-only diagnostic query

```cypher
MATCH (a:User {active:true}), (b:User {active:false})
RETURN
  a.id AS a,
  b.id AS b,
  [(a)-[:FRIEND]->(b) | 1] AS pc,
  size([(a)-[:FRIEND]->(b) | 1]) AS sz
ORDER BY a, b;
```

### Expected behavior

```text
a | b | pc  | sz
--+---+-----+---
1 | 3 | [1] | 1
1 | 4 | []  | 0
2 | 3 | []  | 0
2 | 4 | []  | 0
```

Neo4j `2026.05.0` returns this expected result.

### Actual behavior

Memgraph `3.11.0` returns a non-empty pattern-comprehension result for every row:

```text
a | b | pc  | sz
--+---+-----+---
1 | 3 | [1] | 1
1 | 4 | [1] | 1
2 | 3 | [1] | 1
2 | 4 | [1] | 1
```

So `size([(a)-[:FRIEND]->(b) | 1]) = 0` is never true, even for pairs without a `:FRIEND` relationship.

### Failing write query

```cypher
MATCH (a:User {active:true}), (b:User {active:false})
WITH a, b
WHERE size([(a)-[:FRIEND]->(b) | 1]) = 0
CREATE (a)-[:MONITOR]->(b);

MATCH ()-[m:MONITOR]->()
RETURN count(m) AS cnt;
```

### Expected behavior

There are 3 active-to-inactive pairs without an existing `:FRIEND` edge:

- `(a1, b2)`
- `(a2, b1)`
- `(a2, b2)`

So 3 `:MONITOR` edges should be created.

```text
cnt = 3
```

Neo4j `2026.05.0` returns:

```text
cnt = 3
```

### Actual behavior

Memgraph `3.11.0` returns:

```text
cnt = 0
```

No `:MONITOR` edges are created because the `size(...) = 0` condition is never true.

### Control 1: direct `NOT` pattern works correctly

```cypher
MATCH (n) DETACH DELETE n;

CREATE
  (a1:User {id:1, active:true}),
  (a2:User {id:2, active:true}),
  (b1:User {id:3, active:false}),
  (b2:User {id:4, active:false}),
  (a1)-[:FRIEND]->(b1);

MATCH (a:User {active:true}), (b:User {active:false})
WHERE NOT (a)-[:FRIEND]->(b)
CREATE (a)-[:MONITOR2]->(b);

MATCH ()-[m:MONITOR2]->()
RETURN count(m) AS cnt;
```

Memgraph result:

```text
cnt = 3
```

This is correct.

### Control 2: `NOT exists(...)` also works correctly

```cypher
MATCH (n) DETACH DELETE n;

CREATE
  (a1:User {id:1, active:true}),
  (a2:User {id:2, active:true}),
  (b1:User {id:3, active:false}),
  (b2:User {id:4, active:false}),
  (a1)-[:FRIEND]->(b1);

MATCH (a:User {active:true}), (b:User {active:false})
WHERE NOT exists((a)-[:FRIEND]->(b))
CREATE (a)-[:MONITOR3]->(b);

MATCH ()-[m:MONITOR3]->()
RETURN count(m) AS cnt;
```

Memgraph result:

```text
cnt = 3
```

This is also correct.

#### Read-only diagnostic query

| Engine | Non-empty pattern-comprehension rows | `sz = 0` rows |
|---|---:|---:|
| Neo4j `2026.05.0` | 1 | 3 |
| Memgraph `3.11.0` | 4 | 0 |

#### Write query

| Engine | `MONITOR` count |
|---|---:|
| Neo4j `2026.05.0` | 3 |
| Memgraph `3.11.0` | 0 |

### Why this looks like a bug

1. The reproducer is minimal: 4 nodes and 1 relationship.
2. The read-only diagnostic query isolates the wrong result before any write occurs.
3. The incorrect value comes from the pattern-comprehension expression itself, not from `CREATE`.
4. Equivalent direct pattern predicates work correctly on Memgraph.
5. Neo4j returns the expected result for the same Cypher pattern comprehension.
6. The failure is deterministic on this repro.

### Suspected issue

The pattern comprehension appears to ignore one or both bound endpoint variables when evaluating:

```cypher
[(a)-[:FRIEND]->(b) | 1]
```

It behaves as if the existence of some `:FRIEND` relationship in the graph is enough to produce `[1]` for every outer `(a, b)` row, even when the specific pair does not match.

### LINKED ISSUE #4376: Pattern comprehension in `WHERE` after aggregation in `WITH` triggers internal error
### Memgraph version

Observed on the latest Docker image:

```text
docker pull memgraph/memgraph:latest
Pulled: 2026-07-10

Memgraph version: 3.11.0

Endpoint:
bolt://127.0.0.1:17689
```

### Environment

* Host OS: Linux `5.4.0-216-generic`
* Deployment: Docker `memgraph/memgraph:latest`
* Query interface: Bolt
* Query language: Cypher
* Differential comparison target:
  * Neo4j Community `2026.05.0`

### Description

Memgraph throws an internal error when:

1. an aggregation function is evaluated in a `WITH` clause; and
2. a pattern comprehension is evaluated in the following `WHERE` clause.

The failure is reproducible with multiple aggregation functions, including:

* `count()`
* `sum()`
* `collect()`

The error is:

```text
Pattern Comprehension expected a list, got null.
Please report the problem on GitHub issues
```

Each feature works correctly in isolation:

* aggregation in `WITH` works without the subsequent pattern comprehension;
* the pattern comprehension works in `WHERE` without a preceding aggregation;
* the pattern comprehension also works when materialized before filtering.

The same queries execute correctly on Neo4j Community `2026.05.0`.

### To reproduce

Run each statement separately and in order.

#### 1. Clean the database

```cypher
MATCH (n) DETACH DELETE n;
```

#### 2. Create test data

```cypher
CREATE
  (a:U {id: 1, active: true}),
  (b:U {id: 2, active: false}),
  (c:U {id: 3, active: false}),
  (a)-[:KNOWS]->(b),
  (a)-[:KNOWS]->(c);
```

Node `a` has two outgoing `KNOWS` relationships, both leading to inactive
nodes.

### Failing query 1: `count()`

```cypher
MATCH (a:U)-[:KNOWS]->(b:U)
WITH a, count(b) AS cnt
WHERE cnt > 0
  AND size([
    (a)-[:KNOWS]->(f:U)
    WHERE f.active = false
    | f
  ]) > 0
RETURN a.id AS id;
```

### Failing query 2: `sum()`

```cypher
MATCH (a:U)-[:KNOWS]->(b:U)
WITH a, sum(b.id) AS total
WHERE total > 0
  AND size([
    (a)-[:KNOWS]->(f:U)
    WHERE f.active = false
    | f
  ]) > 0
RETURN a.id AS id;
```

### Failing query 3: `collect()`

```cypher
MATCH (a:U)-[:KNOWS]->(b:U)
WITH a, collect(b.active) AS acts
WHERE size(acts) > 0
  AND size([
    (a)-[:KNOWS]->(f:U)
    WHERE f.active = false
    | f
  ]) > 0
RETURN a.id AS id;
```

### Expected behavior

All three queries should return:

```text
[{id: 1}]
```

The aggregation condition is true, and node `a` has outgoing `KNOWS`
relationships to inactive nodes.

Neo4j Community `2026.05.0` returns the expected result for all three
queries.

### Actual behavior

All three queries fail on Memgraph `3.11.0` with:

```text
Pattern Comprehension expected a list, got null.
Please report the problem on GitHub issues
```

No normal Cypher semantic or runtime error is returned.

#### Control 1: aggregation without pattern comprehension

```cypher
MATCH (a:U)-[:KNOWS]->(b:U)
WITH a, count(b) AS cnt
WHERE cnt > 0
RETURN a.id AS id;
```

Result:

```text
[{id: 1}]
```

#### Control 2: pattern comprehension in `WHERE` without prior aggregation

```cypher
MATCH (a:U {id: 1})
WHERE size([
  (a)-[:KNOWS]->(f:U)
  WHERE f.active = false
  | f
]) > 0
RETURN a.id AS id;
```

Result:

```text
[{id: 1}]
```

#### Control 3: materialize the pattern comprehension before filtering

```cypher
MATCH (a:U {id: 1})
WITH a, [
  (a)-[:KNOWS]->(f:U)
  WHERE f.active = false
  | f
] AS inactive
WHERE size(inactive) > 0
RETURN a.id AS id;
```

Result:

```text
[{id: 1}]
```

### Trigger boundary

| Query shape | Memgraph 3.11.0 |
|---|---|
| Aggregation in `WITH` only | Works |
| Pattern comprehension in `WHERE` only | Works |
| Pattern comprehension materialized in `WITH` | Works |
| Aggregation in `WITH` followed by pattern comprehension in `WHERE` | Internal error |

This indicates that the failure is caused by the interaction between the
aggregation boundary and subsequent evaluation of a pattern comprehension,
rather than either feature independently.

### Related but distinct issue

This may be related to #4180, which reports the same internal error message
with a different query shape:

```text
OPTIONAL MATCH
→ collect(...) in WITH
→ pattern comprehension in ORDER BY
```

This report is distinct because:

* it reproduces on Memgraph `3.11.0`;
* it does not require `OPTIONAL MATCH`;
* the pattern comprehension appears in `WHERE`, not `ORDER BY`;
* it affects `count()` and `sum()` as well as `collect()`.

The common failure may indicate a broader issue with evaluating pattern
comprehensions after an aggregation boundary.

### LINKED ISSUE #4378: Pattern comprehension in `WHERE` produces wrong results after another pattern comprehension in `WITH`
### Memgraph version

Observed on the latest Docker image:

```text
docker pull memgraph/memgraph:latest
Pulled: 2026-07-10

Memgraph version: 3.11.0

Endpoint:
bolt://127.0.0.1:17689
```

Also reproduced in a fresh isolated Docker container, 3 out of 3 runs.

### Environment

- Host OS: Linux `5.4.0-216-generic`
- Deployment: Docker `memgraph/memgraph:latest`
- Query interface: Bolt
- Query language: Cypher / OpenCypher
- Differential comparison target:
  - Neo4j Community `2026.05.0`

### Description

Memgraph produces incorrect results when:

1. a pattern comprehension is evaluated in a `WITH` clause; and
2. another pattern comprehension is evaluated in the following `WHERE`
   clause.

The issue reproduces with several predicate forms, including:

- `AND`;
- `CASE`;
- `NOT`.

Each pattern comprehension works correctly when evaluated independently.
The equivalent query also works when both pattern comprehensions appear in
the same `WHERE` clause.

This is a silent wrong-result bug: the queries complete successfully but
incorrectly include or exclude rows.

### To reproduce

Run each statement separately and in order.

#### 1. Clean the database

```cypher
MATCH (n) DETACH DELETE n;
```

#### 2. Create test data

```cypher
CREATE
  (a:U {id: 1, v: 10, active: true}),
  (b:U {id: 2, v: 20, active: true}),
  (c:U {id: 3, v: 30, active: false}),
  (d:U {id: 4, v: 40, active: false}),
  (a)-[:KNOWS]->(b),
  (a)-[:KNOWS]->(c),
  (b)-[:KNOWS]->(c),
  (b)-[:KNOWS]->(d);
```

The graph has the following relevant state:

```text
Node 1:
- 2 outgoing KNOWS relationships
- 1 active friend: node 2

Node 2:
- 2 outgoing KNOWS relationships
- 0 active friends: nodes 3 and 4 are inactive

Nodes 3 and 4:
- no outgoing KNOWS relationships
```

### Failing query 1: `AND` across the `WITH` / `WHERE` boundary

```cypher
MATCH (a:U)
WITH a, size([
  (a)-[:KNOWS]->(f:U)
  | f
]) AS friend_count
WHERE friend_count > 0
  AND size([
    (a)-[:KNOWS]->(f:U)
    WHERE f.active = true
    | f
  ]) > 0
RETURN a.id AS id
ORDER BY id;
```

Expected:

```text
[1]
```

Only node 1 has both outgoing friends and at least one active friend.

Actual on Memgraph:

```text
[1, 2]
```

Node 2 is incorrectly included even though it has no active friends.

Neo4j `2026.05.0` returns:

```text
[1]
```

### Failing query 2: `CASE` in `WHERE`

```cypher
MATCH (a:U)
WITH a, size([
  (a)-[:KNOWS]->(f:U)
  | f
]) AS friend_count
WHERE CASE
  WHEN friend_count > 0
  THEN size([
    (a)-[:KNOWS]->(f:U)
    WHERE f.active = true
    | f
  ]) > 0
  ELSE false
END
RETURN a.id AS id
ORDER BY id;
```

Expected:

```text
[1]
```

Actual on Memgraph:

```text
[1, 2]
```

Node 2 is again incorrectly included.

Neo4j `2026.05.0` returns:

```text
[1]
```

### Failing query 3: negated predicate across the boundary

```cypher
MATCH (a:U)
WITH a, size([
  (a)-[:KNOWS]->(f:U)
  | f
]) AS friend_count
WHERE friend_count > 0
  AND NOT (
    size([
      (a)-[:KNOWS]->(f:U)
      WHERE f.active = true
      | f
    ]) > 0
  )
RETURN a.id AS id
ORDER BY id;
```

Expected:

```text
[2]
```

Node 2 has outgoing friends but no active friends.

Actual on Memgraph:

```text
[]
```

Node 2 is incorrectly excluded.

Neo4j `2026.05.0` returns:

```text
[2]
```

### Control: both pattern comprehensions in the same `WHERE` clause

The equivalent query works when both pattern comprehensions are evaluated
in the same clause:

```cypher
MATCH (a:U)
WHERE size([
    (a)-[:KNOWS]->(f:U)
    | f
  ]) > 0
  AND size([
    (a)-[:KNOWS]->(f:U)
    WHERE f.active = true
    | f
  ]) > 0
RETURN a.id AS id
ORDER BY id;
```

Memgraph result:

```text
[1]
```

This is correct.

### Expected behavior

The second pattern comprehension should be evaluated independently against
the graph.

Moving the first pattern comprehension from `WHERE` into `WITH` should not
change the meaning of the second pattern comprehension.

Expected results:

```text
AND query:   [1]
CASE query:  [1]
NOT query:   [2]
```

### Actual behavior

Memgraph returns:

```text
AND query:   [1, 2]
CASE query:  [1, 2]
NOT query:   []
```

The incorrect behavior only appears in the tested query shape where one
pattern comprehension is evaluated in `WITH` and another is evaluated in
the following `WHERE`.

### Cross-engine comparison

| Query form | Memgraph `3.11.0` | Neo4j `2026.05.0` | Expected |
|---|---|---|---|
| `AND` across `WITH` / `WHERE` | `[1, 2]` | `[1]` | `[1]` |
| `CASE` in `WHERE` | `[1, 2]` | `[1]` | `[1]` |
| Negated predicate | `[]` | `[2]` | `[2]` |
| Both comprehensions in `WHERE` | `[1]` | `[1]` | `[1]` |

### Isolated recheck

The same behavior was reproduced in a fresh isolated Docker container,
3 out of 3 runs.

### Why this looks like a bug

The following two query structures should be semantically equivalent:

```text
pattern comprehension 1 in WHERE
+
pattern comprehension 2 in WHERE
```

and:

```text
pattern comprehension 1 evaluated in WITH
+
pattern comprehension 2 evaluated in the following WHERE
```

Memgraph returns the correct result for the first structure but incorrect
results for the second.

The direction of the error also changes with predicate structure:

- positive predicates incorrectly include node 2;
- the negated predicate incorrectly excludes node 2.

This suggests an evaluation or planning issue involving multiple pattern
comprehensions across a `WITH` clause boundary. The exact internal cause is
unknown.

### LINKED ISSUE #4388: `CALL ()` subquery-local variable overwrites an outer variable with the same name after an unmatched `OPTIONAL MATCH`
### Memgraph version

Observed on the following local Docker image:

```text
docker images memgraph/memgraph --format "{{.ID}} {{.CreatedAt}}"
(latest, pulled 2026-07-11)

docker exec mg_verify mgconsole --version
Memgraph 3.11.0

Interface: Bolt, accessed via docker exec -i mg_verify mgconsole
```

Also reproduced in a fresh isolated Docker container:

```text
container: mg_verify
image: memgraph/memgraph:latest
pulled: 2026-07-11
version: 3.11.0
reproduction rate: 3/3
```

### Environment

* Host OS: Linux x86_64
* Deployment: Docker
* Query language: Cypher
* Reference engine: Neo4j Community `2026.05.0`

### Description

A variable local to a `CALL ()` subquery can overwrite an outer variable with the same name, even though `CALL ()` imports no outer variables.

In the reproducer below, the outer query binds `a` to an `:A {id: 1}` node. The subquery uses `CALL ()`, so no outer variables are imported. Its `OPTIONAL MATCH` therefore introduces a separate subquery-local variable that is also named `a`.

When the optional pattern does not match, Memgraph returns `null` for the outer `a` after the subquery completes:

```text
aid=Null
```

Renaming the local variable to `x` preserves the outer `a`. Explicitly importing the outer `a` using `CALL (a)` also preserves it.

Neo4j `2026.05.0` preserves the outer binding in all three forms.

### To reproduce

Run each statement separately from a clean database.

#### 1. Clean the database

```cypher
MATCH (n) DETACH DELETE n;
```

#### 2. Create test data

```cypher
CREATE (:A {id: 1});
```

#### 3. Run the failing query

```cypher
MATCH (a:A {id: 1})
CALL () {
  OPTIONAL MATCH (a)-[:NO_SUCH]->(c)
  RETURN c
}
RETURN a.id AS aid, c IS NULL AS c_is_null;
```

The outer `a` is not imported into the subquery because the scope clause is empty:

```cypher
CALL ()
```

The `a` used in the inner pattern is therefore a separate variable local to the subquery.

### Expected behavior

The subquery-local `a` must not modify the outer variable with the same name.

Because the `:NO_SUCH` relationship does not exist, `c` should be `null`, while the outer `a` should remain bound to the node with `id = 1`:

```text
aid=1, c_is_null=true
```

### Actual behavior

Memgraph returns:

```text
+------+-----------+
| aid  | c_is_null |
+------+-----------+
| Null | true      |
+------+-----------+
```

The outer `a` is unexpectedly replaced with `null` after the subquery returns.

### Control 1: rename the subquery-local variable

```cypher
MATCH (a:A {id: 1})
CALL () {
  OPTIONAL MATCH (x)-[:NO_SUCH]->(c)
  RETURN c
}
RETURN a.id AS aid, c IS NULL AS c_is_null;
```

Memgraph correctly returns:

```text
+-----+-----------+
| aid | c_is_null |
+-----+-----------+
| 1   | true      |
+-----+-----------+
```

Changing only the inner variable name from `a` to `x` prevents the outer binding from being overwritten.

### Control 2: explicitly import the outer variable

```cypher
MATCH (a:A {id: 1})
CALL (a) {
  OPTIONAL MATCH (a)-[:NO_SUCH]->(c)
  RETURN c
}
RETURN a.id AS aid, c IS NULL AS c_is_null;
```

Memgraph correctly returns:

```text
+-----+-----------+
| aid | c_is_null |
+-----+-----------+
| 1   | true      |
+-----+-----------+
```

Here, the inner `a` intentionally refers to the imported outer variable, and its outer binding remains intact.

### Results summary

| Subquery form                         | Inner variable | Imported? | Memgraph `aid` | Expected |
| ------------------------------------- | -------------- | :-------: | :------------: | :------: |
| `CALL () { OPTIONAL MATCH (a)-... }`  | `a`            |     No    |    **Null**    |     1    |
| `CALL () { OPTIONAL MATCH (x)-... }`  | `x`            |     No    |        1       |     1    |
| `CALL (a) { OPTIONAL MATCH (a)-... }` | `a`            |    Yes    |        1       |     1    |

### Cross-engine comparison

For the failing query:

| Engine            |   `aid`  | `c_is_null` |
| ----------------- | :------: | :---------: |
| Neo4j `2026.05.0` |     1    |     true    |
| Memgraph `3.11.0` | **Null** |     true    |

### Isolated recheck

The behavior was reproduced in a fresh isolated Docker container:

```text
container: mg_verify
image: memgraph/memgraph:latest
pulled: 2026-07-11
version: 3.11.0

Run 1: aid=Null
Run 2: aid=Null
Run 3: aid=Null

Reproduction rate: 3/3
```

Controls in the same isolated container:

```text
CALL () with inner variable renamed to x: aid=1
CALL (a) with a explicitly imported:      aid=1
```

### Why this appears to be a bug

`CALL ()` imports no variables from the enclosing query scope. Variables introduced inside that subquery are local to the subquery and must not modify identically named variables in the outer scope.

The failing query and the two controls differ only in the subquery variable name or import declaration:

```cypher
CALL () {
  OPTIONAL MATCH (a)-[:NO_SUCH]->(c)
}
```

returns:

```text
aid=Null
```

while:

```cypher
CALL () {
  OPTIONAL MATCH (x)-[:NO_SUCH]->(c)
}
```

and:

```cypher
CALL (a) {
  OPTIONAL MATCH (a)-[:NO_SUCH]->(c)
}
```

both return:

```text
aid=1
```

This isolates the failure to a name collision between an outer variable and an unimported subquery-local variable.

The unmatched `OPTIONAL MATCH` assigns `null` to the local variable, and that local value appears to leak across the subquery boundary and overwrite the unrelated outer binding.

This is a semantic correctness issue: a local variable from a non-importing `CALL ()` subquery must not mutate or replace an outer variable solely because they share the same name.

### LINKED ISSUE #4394: CALL subquery with implicit variable import fails with "Unbound variable" when inner MATCH references same variable name
### Memgraph version

3.11.0

### Environment

Docker, x86_64, Linux, Bolt

### To reproduce

```cypher
CREATE (a:A {id:1});

-- Implicit CALL import + MATCH (a) → "Unbound variable: a"
MATCH (a:A {id:1})
CALL { MATCH (a) WHERE true RETURN a.id AS x }
RETURN a.id AS aid, x;
```

### Expected behavior

Outer variable `a` should be accessible inside the CALL subquery. Neo4j returns `[{aid: 1, x: 1}]`.

### Actual behavior

Memgraph throws: `Unbound variable: a.`

### Workaround

Use explicit CALL variable import:

```cypher
MATCH (a:A {id:1})
CALL (a) { MATCH (a) WHERE true RETURN a.id AS x }
RETURN a.id AS aid, x;
-- → [{aid: 1, x: 1}] (correct)
```

### Control

Using a different variable name inside CALL with implicit import works correctly:

```cypher
MATCH (a:A {id:1})
CALL { MATCH (b:A {id:1}) RETURN b.id AS x }
RETURN a.id AS aid, x;
-- → [{aid: 1, x: 1}] (correct)
```

The bug is specific to the combination of: (1) legacy CALL syntax without explicit import, and (2) inner query referencing the outer variable by the same name.

imilinovic:
A pattern comprehension in the `WHERE` or `ORDER BY` of a `WITH` did not correlate with the bound variable — it re-scanned it, so `size([(p)-->(m)|m])` silently meant *"every edge in the graph"*. Filters admitted rows they should have excluded, `ORDER BY` did not sort, and with an aggregation in the same `WITH` the comprehension was never planned at all (`Pattern Comprehension expected a list, got null`). Eight reported issues are this one defect in different shapes.

## How

`VisitReturnBody` mints the `WITH`'s output symbols *after* the named expressions but *before* `order_by`/`where`, so a comprehension there binds to those symbols — while `GenReturnBody` spliced every `RollUpApply` *below* the `Produce` that writes them. Unbound at planning time ⇒ uncorrelated `ScanAll`.

`ReturnBodyContext` now records each comprehension's position and splices per position. `OrderByCursor` restores only its `output_symbols_`, and `Filter(where)` sits above it, so one splice point cannot serve both:

```
Filter(where) → RollUpApply(where) → Limit → Skip → OrderBy → RollUpApply(orderby) → Distinct → Produce
```

Five chains can evaluate a comprehension. Three drain through one helper (main clause chain, FOREACH body, each MERGE branch); a WITH/RETURN body drains on demand, sharing the same view rule; a `CallProcedure` clause has no drain — a known gap. `ExpandVariable` can only read `View::OLD`, so one rule at the single planning funnel forces that for variable-length patterns, rejecting only when the pattern's root is a node a write clause of the same query part binds.

## Crash fixes

Two `MG_ASSERT`s on the planning path — unconditional in every build, so a whole-process abort dropping every session, and reachable from **`EXPLAIN` alone** without write permission. Both fired on `master`; both are now `QueryException`s, with the shapes that reached them rejected earlier by name.

- `EXPLAIN CREATE (a)-[:R]->(b)-[:R]->(c) RETURN [(a)-[*1..2]->(x) | x.name]`
- `EXPLAIN MATCH (a)-[r]->(b) WHERE EXISTS { MATCH (a)-[r]->(y) RETURN y } RETURN a`

## Fixes with no comprehension involved

An un-imported name of the enclosing query is now out of scope inside a `CALL {}` subquery, and a named path no longer escapes into an outer one. Each measured against a merge-base binary and a reference implementation:

- **Silent wrong answers for a plain `MATCH`.** `MATCH (n) CALL { MATCH (n)-[:R]->(x) RETURN count(x) AS c } RETURN n.id, c` returned `n.id` as `null` in *every* row — the subquery's scan resolved `n` to the caller's symbol and wrote through the frame slot they share. Closes #4394, #4388, #3955, #4160.
- **A named path was overwritten by an inner one of the same name**, in a subquery or in a comprehension's own path variable — the latter needs no subquery at all.
- **A `WITH` with both `ORDER BY` and `WHERE` could return wrong rows.** `OrderBy` restores only the symbols it is given and the `WHERE` sits above it, so a symbol the projection dropped was frozen at the last input row. `MATCH (p) WITH p.name AS n ORDER BY n WHERE p.age > 25 RETURN n` returned every row.

## Breaking changes

1. **Silent results change.** Inside a `CALL {}`, a *pattern* mentioning an un-imported variable of the enclosing query declares that name afresh instead of reusing the caller's, so such a query returns **different results with no error**. A bare `CALL {}` now matches its documented scope, identical to `CALL () {}`; every non-pattern reference was already rejected as unbound. Import explicitly — `CALL (n) {`, `CALL (*) {`, or a leading `WITH n`. The `exists()` *pattern* form raises instead.
2. `CREATE (q:T {c: size([(q)-->(m)|m])})` → `Entity 'q' cannot be created and referenced by a pattern comprehension in the same clause`. No placement works: the operator that binds the symbol is the one that reads the result.
3. Reusing an **already bound relationship** in a comprehension or `EXISTS {}` pattern → `Cannot use the already bound relationship 'r' in a pattern here`. `Expand` has no `existing_edge` counterpart, so there is no operator to emit. Bound *nodes* still correlate. Previously a process abort.
4. A **variable-length** comprehension expanding from a node a write clause of the same query part binds → `A variable-length pattern cannot expand from 'a', …`. Closing the write clause with a `WITH` works. Previously a process abort, or a runtime storage error.

Also: `CALL { CREATE (p) … }` over an un-imported outer `p` now succeeds (unlabelled form only); `EXPLAIN` output changes.

## Verification

Measured against a binary built at the merge base and against a reference implementation.

- `query_plan` **231/231**, `query_semantic` **238/238**, plus six neighbouring `query_*` suites.
- `gql_behave memgraph_V1` with `USING PARALLEL EXECUTION`: **1158 of 1227, zero regressions** — the residual 69 are `text_*`/`vector_*` needing experimental flags. `openCypher_M09`: 113 failures, identical on both binaries.
- Every fix's test was verified to **fail when its own hunk is reverted**.

## Known gaps, deliberately not fixed here

Each is pre-existing and behaves identically on the merge-base binary.

- A `CallProcedure` clause never drains its comprehensions — arguments as much as `YIELD … WHERE`. Also why `pending_comprehensions` cannot yet be asserted empty.
- `MakePatternComprehensionFilter` is a second planning path, so a comprehension in a `MATCH … WHERE` bypasses the variable-length rule.
- `Accumulate` does not restore a comprehension's anonymous result symbol, so a write clause followed by a non-reprojected symbol freezes it.
- The main clause chain drains before a write clause, so `MATCH (p) CREATE (p)-[:R]->(x) RETURN size([(p)-[:R]->(m)|m])` counts pre-existing edges only. A FOREACH body likewise splices at the earliest satisfying clause rather than the reading one.
- Comprehension in `SKIP`/`LIMIT`; the CREATE rejection is order-dependent; `MERGE` has no counterpart to it.
- Same scoping family, still open: `check_node_semantic` uses raw `HasSymbol`, so `CREATE (p)` shadows but `CREATE (p:L)` errors; a comprehension's *filter* still reaches an un-imported outer name; `CALL (v) { WITH … MATCH (v) … }` still resolves to the caller's symbol.
- One scenario lives in `query_semantic.cpp` rather than `subqueries.feature`: an aggregation inside a `CALL {}` trips a pre-existing parallel-executor bug, filed separately.

Closes #4394
Closes #4388
Closes #4378
Closes #4376
Closes #4365
Closes #4364
Closes #4350
Closes #4339
Closes #4180
Closes #4161
Closes #4160
Closes #3955

### DIFF
--- a/tests/unit/query_semantic.cpp
+++ b/tests/unit/query_semantic.cpp
@@ -1604,3 +1604,165 @@ TYPED_TEST(TestSymbolGenerator, ListComprehensionInWith) {
   // 0, we erase the x symbol as it is only mentioned in the list comprehension
   ASSERT_EQ(collector.symbols_.size(), 0);
 }
+
+// Shadowing of un-imported outer variables inside a `CALL {}` subquery. Such a name is out of scope in the subquery,
+// so a pattern occurrence of it declares a fresh variable rather than referencing the outer one. Referencing it would
+// resolve to the outer symbol and make the branch write through its frame slot, which the subquery shares with its
+// caller. `SUBQUERY_COMPREHENSION(name)` builds `[(<name>)-->() | 1]`.
+//
+// Note the subquery RETURNs below list two NamedExpressions rather than a bare name plus one: `RETURN("t", NEXPR(...))`
+// selects the `RETURN(expr, AS(name))` overload, which would overwrite the comprehension with an identifier.
+#define SUBQUERY_COMPREHENSION(name)                                                                                   \
+  PATTERN_COMPREHENSION(                                                                                               \
+      nullptr,                                                                                                         \
+      PATTERN(                                                                                                         \
+          NODE(name), EDGE("anon_edge", EdgeAtom::Direction::OUT, {}, false), NODE("anon_node", std::nullopt, false)), \
+      nullptr,                                                                                                         \
+      LITERAL(1))
+
+namespace {
+
+const Identifier &ComprehensionStartNode(PatternComprehension *pc) {
+  auto *node = dynamic_cast<NodeAtom *>(pc->pattern_->atoms_[0]);
+  MG_ASSERT(node, "expected the comprehension's first atom to be a node");
+  return *node->identifier_;
+}
+
+// A subquery's RETURN rejects an unaliased expression, and the NEXPR helper does not set the flag.
+NamedExpression *Aliased(NamedExpression *named_expr) {
+  named_expr->is_aliased_ = true;
+  return named_expr;
+}
+
+}  // namespace
+
+TYPED_TEST(TestSymbolGenerator, PatternComprehensionInSubqueryShadowsUnimportedOuterVariable) {
+  // MATCH (p) CALL { MATCH (t) RETURN t, [(p)-->() | 1] AS k } RETURN p, k
+  auto *outer_p = NODE("p");
+  auto *comprehension = SUBQUERY_COMPREHENSION("p");
+  auto *subquery = SINGLE_QUERY(MATCH(PATTERN(NODE("t"))),
+                                RETURN(Aliased(NEXPR("t", IDENT("t"))), Aliased(NEXPR("k", comprehension))));
+  auto *query = QUERY(SINGLE_QUERY(MATCH(PATTERN(outer_p)), CALL_SUBQUERY(subquery), RETURN("p", "k")));
+
+  auto symbol_table = MakeSymbolTable(query);
+
+  const auto outer_symbol = symbol_table.at(*outer_p->identifier_);
+  const auto inner_symbol = symbol_table.at(ComprehensionStartNode(comprehension));
+  EXPECT_NE(outer_symbol, inner_symbol) << "the un-imported name must be declared afresh, not bound to the outer `p`";
+  EXPECT_EQ(outer_symbol.name(), inner_symbol.name());
+}
+
+TYPED_TEST(TestSymbolGenerator, PatternComprehensionInSubqueryCorrelatesExplicitlyImportedVariable) {
+  // MATCH (p) CALL (p) { MATCH (t) RETURN t, [(p)-->() | 1] AS k } RETURN p, k
+  auto *outer_p = NODE("p");
+  auto *comprehension = SUBQUERY_COMPREHENSION("p");
+  auto *subquery = SINGLE_QUERY(MATCH(PATTERN(NODE("t"))),
+                                RETURN(Aliased(NEXPR("t", IDENT("t"))), Aliased(NEXPR("k", comprehension))));
+  auto *call_sub = CALL_SUBQUERY(subquery);
+  call_sub->has_variable_scope_ = true;
+  call_sub->scoped_variables_.push_back(NEXPR("p", IDENT("p")));
+  auto *query = QUERY(SINGLE_QUERY(MATCH(PATTERN(outer_p)), call_sub, RETURN("p", "k")));
+
+  auto symbol_table = MakeSymbolTable(query);
+
+  EXPECT_EQ(symbol_table.at(*outer_p->identifier_), symbol_table.at(ComprehensionStartNode(comprehension)))
+      << "an explicitly imported name is in scope, so it must still correlate";
+}
+
+TYPED_TEST(TestSymbolGenerator, PatternComprehensionInSubqueryCorrelatesWithCallStar) {
+  // MATCH (p) CALL (*) { MATCH (t) RETURN t, [(p)-->() | 1] AS k } RETURN p, k
+  auto *outer_p = NODE("p");
+  auto *comprehension = SUBQUERY_COMPREHENSION("p");
+  auto *subquery = SINGLE_QUERY(MATCH(PATTERN(NODE("t"))),
+                                RETURN(Aliased(NEXPR("t", IDENT("t"))), Aliased(NEXPR("k", comprehension))));
+  auto *call_sub = CALL_SUBQUERY(subquery);
+  call_sub->has_variable_scope_ = true;
+  call_sub->all_variables_scoped_ = true;
+  auto *query = QUERY(SINGLE_QUERY(MATCH(PATTERN(outer_p)), call_sub, RETURN("p", "k")));
+
+  auto symbol_table = MakeSymbolTable(query);
+
+  EXPECT_EQ(symbol_table.at(*outer_p->identifier_), symbol_table.at(ComprehensionStartNode(comprehension)))
+      << "CALL (*) imports every user-declared outer variable";
+}
+
+TYPED_TEST(TestSymbolGenerator, PatternComprehensionInNestedSubqueryReshadowsImportedVariable) {
+  // MATCH (p) CALL (p) { CALL { MATCH (t) RETURN t, [(p)-->() | 1] AS k } RETURN t, k } RETURN p, k
+  // The inner `CALL {}` imports nothing, so `p` is out of scope again even though the outer subquery imported it.
+  auto *outer_p = NODE("p");
+  auto *comprehension = SUBQUERY_COMPREHENSION("p");
+  auto *inner_subquery = SINGLE_QUERY(MATCH(PATTERN(NODE("t"))),
+                                      RETURN(Aliased(NEXPR("t", IDENT("t"))), Aliased(NEXPR("k", comprehension))));
+  auto *outer_subquery = SINGLE_QUERY(CALL_SUBQUERY(inner_subquery), RETURN("t", "k"));
+  auto *call_sub = CALL_SUBQUERY(outer_subquery);
+  call_sub->has_variable_scope_ = true;
+  call_sub->scoped_variables_.push_back(NEXPR("p", IDENT("p")));
+  auto *query = QUERY(SINGLE_QUERY(MATCH(PATTERN(outer_p)), call_sub, RETURN("p", "k")));
+
+  auto symbol_table = MakeSymbolTable(query);
+
+  EXPECT_NE(symbol_table.at(*outer_p->identifier_), symbol_table.at(ComprehensionStartNode(comprehension)))
+      << "the innermost CALL {} imported nothing, so `p` is out of scope there too";
+}
+
+TYPED_TEST(TestSymbolGenerator, PatternComprehensionInSubqueryCorrelatesItsOwnScopeVariable) {
+  // MATCH (p) CALL { MATCH (t) RETURN t, [(t)-->() | 1] AS k } RETURN t, k
+  // `t` is declared inside the subquery, so it is in scope there and must correlate, not be shadowed.
+  auto *inner_t = NODE("t");
+  auto *comprehension = SUBQUERY_COMPREHENSION("t");
+  auto *subquery = SINGLE_QUERY(MATCH(PATTERN(inner_t)),
+                                RETURN(Aliased(NEXPR("t", IDENT("t"))), Aliased(NEXPR("k", comprehension))));
+  auto *query = QUERY(SINGLE_QUERY(MATCH(PATTERN(NODE("p"))), CALL_SUBQUERY(subquery), RETURN("t", "k")));
+
+  auto symbol_table = MakeSymbolTable(query);
+
+  EXPECT_EQ(symbol_table.at(*inner_t->identifier_), symbol_table.at(ComprehensionStartNode(comprehension)))
+      << "a name declared inside the subquery is in scope there, so shadowing must not fire";
+}
+
+TYPED_TEST(TestSymbolGenerator, PatternComprehensionAtTopLevelDeclaresUndeclaredNameAfresh) {
+  // MATCH (t) RETURN [(zz)-->() | 1] AS k -- no CALL {} anywhere, so the rule must not fire differently here
+  auto *comprehension = SUBQUERY_COMPREHENSION("zz");
+  auto *query = QUERY(SINGLE_QUERY(MATCH(PATTERN(NODE("t"))), RETURN(NEXPR("k", comprehension))));
+
+  auto symbol_table = MakeSymbolTable(query);
+
+  EXPECT_EQ(symbol_table.at(ComprehensionStartNode(comprehension)).name(), "zz");
+}
+
+TYPED_TEST(TestSymbolGenerator, CreateInSubqueryMayRedeclareUnimportedOuterVariable) {
+  // MATCH (p) CALL { CREATE (p) RETURN p AS q } RETURN q
+  // `p` is out of scope in the subquery, so the CREATE declares a fresh node instead of raising a redeclaration
+  // error, and the outer `p` keeps its own frame slot.
+  auto *outer_p = NODE("p");
+  auto *created_p = NODE("p");
+  auto *subquery = SINGLE_QUERY(CREATE(PATTERN(created_p)), RETURN(IDENT("p"), AS("q")));
+  auto *query = QUERY(SINGLE_QUERY(MATCH(PATTERN(outer_p)), CALL_SUBQUERY(subquery), RETURN("q")));
+
+  auto symbol_table = MakeSymbolTable(query);
+
+  EXPECT_NE(symbol_table.at(*outer_p->identifier_), symbol_table.at(*created_p->identifier_))
+      << "the created node must be a fresh symbol, so the outer `p` survives the subquery";
+}
+
+TYPED_TEST(TestSymbolGenerator, SubqueryReturningShadowingNameStillCollidesWithOuterScope) {
+  // MATCH (p) CALL { CREATE (p) RETURN p } RETURN p
+  // Returning the shadowing name under its own name would collide with the caller's `p`; still rejected.
+  auto *subquery = SINGLE_QUERY(CREATE(PATTERN(NODE("p"))), RETURN("p"));
+  auto *query = QUERY(SINGLE_QUERY(MATCH(PATTERN(NODE("p"))), CALL_SUBQUERY(subquery), RETURN("p")));
+
+  EXPECT_THROW(MakeSymbolTable(query), SemanticException);
+}
+
+TYPED_TEST(TestSymbolGenerator, ExistsPatternInSubqueryRejectsUnimportedOuterVariable) {
+  // MATCH (p) CALL { MATCH (t) WHERE exists((p)-->()) RETURN t } RETURN t
+  // A pattern expression may not introduce variables, so the out-of-scope `p` is an unbound reference.
+  auto *exists = EXISTS(PATTERN(
+      NODE("p"), EDGE("anon_edge", EdgeAtom::Direction::OUT, {}, false), NODE("anon_node", std::nullopt, false)));
+  auto *subquery = SINGLE_QUERY(MATCH(PATTERN(NODE("t"))), WHERE(exists), RETURN("t"));
+  auto *query = QUERY(SINGLE_QUERY(MATCH(PATTERN(NODE("p"))), CALL_SUBQUERY(subquery), RETURN("t")));
+
+  EXPECT_THROW(MakeSymbolTable(query), SemanticException);
+}
+
+#undef SUBQUERY_COMPREHENSION
--- a/tests/unit/query_semantic.cpp
+++ b/tests/unit/query_semantic.cpp
@@ -1608,11 +1608,11 @@ TYPED_TEST(TestSymbolGenerator, ListComprehensionInWith) {
 // Shadowing of un-imported outer variables inside a `CALL {}` subquery. Such a name is out of scope in the subquery,
 // so a pattern occurrence of it declares a fresh variable rather than referencing the outer one. Referencing it would
 // resolve to the outer symbol and make the branch write through its frame slot, which the subquery shares with its
-// caller. `SUBQUERY_COMPREHENSION(name)` builds `[(<name>)-->() | 1]`.
+// caller. `COMPREHENSION_OVER(name)` builds `[(<name>)-->() | 1]`.
 //
 // Note the subquery RETURNs below list two NamedExpressions rather than a bare name plus one: `RETURN("t", NEXPR(...))`
 // selects the `RETURN(expr, AS(name))` overload, which would overwrite the comprehension with an identifier.
-#define SUBQUERY_COMPREHENSION(name)                                                                                   \
+#define COMPREHENSION_OVER(name)                                                                                       \
   PATTERN_COMPREHENSION(                                                                                               \
       nullptr,                                                                                                         \
       PATTERN(                                                                                                         \
@@ -1639,7 +1639,7 @@ NamedExpression *Aliased(NamedExpression *named_expr) {
 TYPED_TEST(TestSymbolGenerator, PatternComprehensionInSubqueryShadowsUnimportedOuterVariable) {
   // MATCH (p) CALL { MATCH (t) RETURN t, [(p)-->() | 1] AS k } RETURN p, k
   auto *outer_p = NODE("p");
-  auto *comprehension = SUBQUERY_COMPREHENSION("p");
+  auto *comprehension = COMPREHENSION_OVER("p");
   auto *subquery = SINGLE_QUERY(MATCH(PATTERN(NODE("t"))),
                                 RETURN(Aliased(NEXPR("t", IDENT("t"))), Aliased(NEXPR("k", comprehension))));
   auto *query = QUERY(SINGLE_QUERY(MATCH(PATTERN(outer_p)), CALL_SUBQUERY(subquery), RETURN("p", "k")));
@@ -1655,7 +1655,7 @@ TYPED_TEST(TestSymbolGenerator, PatternComprehensionInSubqueryShadowsUnimportedO
 TYPED_TEST(TestSymbolGenerator, PatternComprehensionInSubqueryCorrelatesExplicitlyImportedVariable) {
   // MATCH (p) CALL (p) { MATCH (t) RETURN t, [(p)-->() | 1] AS k } RETURN p, k
   auto *outer_p = NODE("p");
-  auto *comprehension = SUBQUERY_COMPREHENSION("p");
+  auto *comprehension = COMPREHENSION_OVER("p");
   auto *subquery = SINGLE_QUERY(MATCH(PATTERN(NODE("t"))),
                                 RETURN(Aliased(NEXPR("t", IDENT("t"))), Aliased(NEXPR("k", comprehension))));
   auto *call_sub = CALL_SUBQUERY(subquery);
@@ -1672,7 +1672,7 @@ TYPED_TEST(TestSymbolGenerator, PatternComprehensionInSubqueryCorrelatesExplicit
 TYPED_TEST(TestSymbolGenerator, PatternComprehensionInSubqueryCorrelatesWithCallStar) {
   // MATCH (p) CALL (*) { MATCH (t) RETURN t, [(p)-->() | 1] AS k } RETURN p, k
   auto *outer_p = NODE("p");
-  auto *comprehension = SUBQUERY_COMPREHENSION("p");
+  auto *comprehension = COMPREHENSION_OVER("p");
   auto *subquery = SINGLE_QUERY(MATCH(PATTERN(NODE("t"))),
                                 RETURN(Aliased(NEXPR("t", IDENT("t"))), Aliased(NEXPR("k", comprehension))));
   auto *call_sub = CALL_SUBQUERY(subquery);
@@ -1690,7 +1690,7 @@ TYPED_TEST(TestSymbolGenerator, PatternComprehensionInNestedSubqueryReshadowsImp
   // MATCH (p) CALL (p) { CALL { MATCH (t) RETURN t, [(p)-->() | 1] AS k } RETURN t, k } RETURN p, k
   // The inner `CALL {}` imports nothing, so `p` is out of scope again even though the outer subquery imported it.
   auto *outer_p = NODE("p");
-  auto *comprehension = SUBQUERY_COMPREHENSION("p");
+  auto *comprehension = COMPREHENSION_OVER("p");
   auto *inner_subquery = SINGLE_QUERY(MATCH(PATTERN(NODE("t"))),
                                       RETURN(Aliased(NEXPR("t", IDENT("t"))), Aliased(NEXPR("k", comprehension))));
   auto *outer_subquery = SINGLE_QUERY(CALL_SUBQUERY(inner_subquery), RETURN("t", "k"));
@@ -1709,7 +1709,7 @@ TYPED_TEST(TestSymbolGenerator, PatternComprehensionInSubqueryCorrelatesItsOwnSc
   // MATCH (p) CALL { MATCH (t) RETURN t, [(t)-->() | 1] AS k } RETURN t, k
   // `t` is declared inside the subquery, so it is in scope there and must correlate, not be shadowed.
   auto *inner_t = NODE("t");
-  auto *comprehension = SUBQUERY_COMPREHENSION("t");
+  auto *comprehension = COMPREHENSION_OVER("t");
   auto *subquery = SINGLE_QUERY(MATCH(PATTERN(inner_t)),
                                 RETURN(Aliased(NEXPR("t", IDENT("t"))), Aliased(NEXPR("k", comprehension))));
   auto *query = QUERY(SINGLE_QUERY(MATCH(PATTERN(NODE("p"))), CALL_SUBQUERY(subquery), RETURN("t", "k")));
@@ -1720,37 +1720,24 @@ TYPED_TEST(TestSymbolGenerator, PatternComprehensionInSubqueryCorrelatesItsOwnSc
       << "a name declared inside the subquery is in scope there, so shadowing must not fire";
 }
 
-TYPED_TEST(TestSymbolGenerator, PatternComprehensionAtTopLevelDeclaresUndeclaredNameAfresh) {
-  // MATCH (t) RETURN [(zz)-->() | 1] AS k -- no CALL {} anywhere, so the rule must not fire differently here
-  auto *comprehension = SUBQUERY_COMPREHENSION("zz");
-  auto *query = QUERY(SINGLE_QUERY(MATCH(PATTERN(NODE("t"))), RETURN(NEXPR("k", comprehension))));
+TYPED_TEST(TestSymbolGenerator, PatternComprehensionAtTopLevelCorrelatesToABoundName) {
+  // MATCH (zz) RETURN [(zz)-->() | 1] AS k
+  // Outside a subquery there is no boundary to shadow against, so a bound name in a comprehension pattern must resolve
+  // to the same symbol. Asserting the symbol's *name* would pass either way - both branches name it "zz".
+  auto *outer = NODE("zz");
+  auto *comprehension = COMPREHENSION_OVER("zz");
+  auto *query =
+      QUERY(SINGLE_QUERY(MATCH(PATTERN(outer)), RETURN(NEXPR("k", static_cast<Expression *>(comprehension)))));
 
   auto symbol_table = MakeSymbolTable(query);
 
-  EXPECT_EQ(symbol_table.at(ComprehensionStartNode(comprehension)).name(), "zz");
+  EXPECT_EQ(symbol_table.at(ComprehensionStartNode(comprehension)), symbol_table.at(*outer->identifier_))
+      << "at top level the shadowing rule must not fire";
 }
 
 // Moved here from `subqueries.feature` because that suite also runs with `USING PARALLEL EXECUTION`, and this shape
 // (an aggregation inside a `CALL {}`) hits a pre-existing parallel-executor bug that nulls every outer variable --
 // unrelated to what the scenario is checking. Asserting the symbol identity pins the rule directly instead.
-TYPED_TEST(TestSymbolGenerator, PatternComprehensionInAggregatingSubqueryWhereShadowsUnimportedOuterVariable) {
-  // MATCH (p) CALL { MATCH (t) WITH t, count(*) AS c WHERE size([(p)-->() | 1]) > 0 RETURN t } RETURN p
-  auto *outer_p = NODE("p");
-  auto *comprehension = SUBQUERY_COMPREHENSION("p");
-  auto *subquery = SINGLE_QUERY(MATCH(PATTERN(NODE("t"))),
-                                WITH(IDENT("t"), AS("t"), COUNT(LITERAL(1), false), AS("c")),
-                                WHERE(GREATER(comprehension, LITERAL(0))),
-                                RETURN(Aliased(NEXPR("t", IDENT("t")))));
-  auto *query = QUERY(SINGLE_QUERY(MATCH(PATTERN(outer_p)), CALL_SUBQUERY(subquery), RETURN("p")));
-
-  auto symbol_table = MakeSymbolTable(query);
-
-  const auto outer_symbol = symbol_table.at(*outer_p->identifier_);
-  const auto inner_symbol = symbol_table.at(ComprehensionStartNode(comprehension));
-  EXPECT_NE(outer_symbol, inner_symbol) << "an aggregating WHERE must not correlate the un-imported outer `p` either";
-  EXPECT_EQ(outer_symbol.name(), inner_symbol.name());
-}
-
 TYPED_TEST(TestSymbolGenerator, PlainMatchInSubqueryShadowsUnimportedOuterVariable) {
   // MATCH (a) CALL { MATCH (a)-[:R]->(x) RETURN x } RETURN a, x
   // The subquery's own MATCH pattern is the same case as the comprehension: `a` is out of scope, so the pattern
@@ -1795,7 +1782,8 @@ TYPED_TEST(TestSymbolGenerator, ComprehensionInWhereRejectsAlreadyBoundRelations
                                    WHERE(EQ(pattern_comp, LIST())),
                                    RETURN("a")));
 
-  EXPECT_THROW(MakeSymbolTable(query), SemanticException);
+  EXPECT_THAT([&] { MakeSymbolTable(query); },
+              ThrowsMessage<SemanticException>(::testing::HasSubstr("already bound relationship 'r'")));
 }
 
 TYPED_TEST(TestSymbolGenerator, ComprehensionInWhereStillCorrelatesAnAlreadyBoundNode) {
@@ -1850,8 +1838,6 @@ TYPED_TEST(TestSymbolGenerator, ExistsPatternInSubqueryRejectsUnimportedOuterVar
   EXPECT_THROW(MakeSymbolTable(query), SemanticException);
 }
 
-#undef SUBQUERY_COMPREHENSION
-
 // A pattern comprehension may not reference a symbol the same CREATE clause declares. The operator that binds the
 // symbol is the one that reads the comprehension's result, so the frame slot is unwritten no matter where the
 // RollUpApply is spliced. A comprehension over a symbol bound by an *earlier* clause is fine and must keep working -
@@ -1867,18 +1853,9 @@ NodeAtom *WithProperty(AstStorage &storage, NodeAtom *node, const std::string &n
 
 }  // namespace
 
-// `[(<name>)-->() | 1]`
-#define SELF_COMPREHENSION(name)                                                                                       \
-  PATTERN_COMPREHENSION(                                                                                               \
-      nullptr,                                                                                                         \
-      PATTERN(                                                                                                         \
-          NODE(name), EDGE("anon_edge", EdgeAtom::Direction::OUT, {}, false), NODE("anon_node", std::nullopt, false)), \
-      nullptr,                                                                                                         \
-      LITERAL(1))
-
 TYPED_TEST(TestSymbolGenerator, PatternComprehensionOverNodeCreatedBySameClauseIsRejected) {
   // CREATE (q:L {c: [(q)-->() | 1]})
-  auto *created = WithProperty(this->storage, NODE("q", "L"), "c", SELF_COMPREHENSION("q"));
+  auto *created = WithProperty(this->storage, NODE("q", "L"), "c", COMPREHENSION_OVER("q"));
   auto *query = QUERY(SINGLE_QUERY(CREATE(PATTERN(created))));
 
   EXPECT_THROW(MakeSymbolTable(query), SemanticException);
@@ -1888,7 +1865,7 @@ TYPED_TEST(TestSymbolGenerator, PatternComprehensionOverNodeCreatedBySameClauseI
   // FOREACH (i IN [1] | CREATE (q:L {c: [(q)-->() | 1]}))
   // The FOREACH-wrapped form is the one that regressed: before the fix the merge-base planned it with an uncorrelated
   // scan and completed with a wrong count, and afterwards it failed with an internal error.
-  auto *created = WithProperty(this->storage, NODE("q", "L"), "c", SELF_COMPREHENSION("q"));
+  auto *created = WithProperty(this->storage, NODE("q", "L"), "c", COMPREHENSION_OVER("q"));
   auto *query = QUERY(SINGLE_QUERY(FOREACH(NEXPR("i", LIST(LITERAL(1))), {CREATE(PATTERN(created))})));
 
   EXPECT_THROW(MakeSymbolTable(query), SemanticException);
@@ -1896,7 +1873,7 @@ TYPED_TEST(TestSymbolGenerator, PatternComprehensionOverNodeCreatedBySameClauseI
 
 TYPED_TEST(TestSymbolGenerator, PatternComprehensionOverNodeCreatedByEarlierPatternInSameClauseIsRejected) {
   // CREATE (a:L), (q:L {c: [(a)-->() | 1]})
-  auto *created = WithProperty(this->storage, NODE("q", "L"), "c", SELF_COMPREHENSION("a"));
+  auto *created = WithProperty(this->storage, NODE("q", "L"), "c", COMPREHENSION_OVER("a"));
   auto *query = QUERY(SINGLE_QUERY(CREATE(PATTERN(NODE("a", "L")), PATTERN(created))));
 
   EXPECT_THROW(MakeSymbolTable(query), SemanticException);
@@ -1921,7 +1898,7 @@ TYPED_TEST(TestSymbolGenerator, PatternComprehensionReferencingCreatedNodeOnlyIn
 TYPED_TEST(TestSymbolGenerator, PatternComprehensionOverNodeCreatedByAnEarlierCreateClauseIsAccepted) {
   // CREATE (a:L) CREATE (q:L {c: [(a)-->() | 1]})
   // A separate clause binds `a`, so the comprehension is drained after it and correlates normally.
-  auto *created = WithProperty(this->storage, NODE("q", "L"), "c", SELF_COMPREHENSION("a"));
+  auto *created = WithProperty(this->storage, NODE("q", "L"), "c", COMPREHENSION_OVER("a"));
   auto *query = QUERY(SINGLE_QUERY(CREATE(PATTERN(NODE("a", "L"))), CREATE(PATTERN(created))));
 
   EXPECT_NO_THROW(MakeSymbolTable(query));
@@ -1929,7 +1906,7 @@ TYPED_TEST(TestSymbolGenerator, PatternComprehensionOverNodeCreatedByAnEarlierCr
 
 TYPED_TEST(TestSymbolGenerator, PatternComprehensionOverMatchedNodeInsideCreateIsAccepted) {
   // MATCH (p) CREATE (q:L {c: [(p)-->() | 1]})
-  auto *created = WithProperty(this->storage, NODE("q", "L"), "c", SELF_COMPREHENSION("p"));
+  auto *created = WithProperty(this->storage, NODE("q", "L"), "c", COMPREHENSION_OVER("p"));
   auto *query = QUERY(SINGLE_QUERY(MATCH(PATTERN(NODE("p"))), CREATE(PATTERN(created))));
 
   EXPECT_NO_THROW(MakeSymbolTable(query));
@@ -1938,7 +1915,7 @@ TYPED_TEST(TestSymbolGenerator, PatternComprehensionOverMatchedNodeInsideCreateI
 TYPED_TEST(TestSymbolGenerator, PatternComprehensionOverItsOwnNodesInsideCreateIsAccepted) {
   // CREATE (q:L {c: [(z)-->() | 1]})
   // Nothing the CREATE binds is referenced, so the comprehension stands alone.
-  auto *created = WithProperty(this->storage, NODE("q", "L"), "c", SELF_COMPREHENSION("z"));
+  auto *created = WithProperty(this->storage, NODE("q", "L"), "c", COMPREHENSION_OVER("z"));
   auto *query = QUERY(SINGLE_QUERY(CREATE(PATTERN(created))));
 
   EXPECT_NO_THROW(MakeSymbolTable(query));
--- a/tests/unit/query_semantic.cpp
+++ b/tests/unit/query_semantic.cpp
@@ -1921,4 +1921,4 @@ TYPED_TEST(TestSymbolGenerator, PatternComprehensionOverItsOwnNodesInsideCreateI
   EXPECT_NO_THROW(MakeSymbolTable(query));
 }
 
-#undef SELF_COMPREHENSION
+#undef COMPREHENSION_OVER
--- a/tests/unit/query_semantic.cpp
+++ b/tests/unit/query_semantic.cpp
@@ -1730,6 +1730,27 @@ TYPED_TEST(TestSymbolGenerator, PatternComprehensionAtTopLevelDeclaresUndeclared
   EXPECT_EQ(symbol_table.at(ComprehensionStartNode(comprehension)).name(), "zz");
 }
 
+// Moved here from `subqueries.feature` because that suite also runs with `USING PARALLEL EXECUTION`, and this shape
+// (an aggregation inside a `CALL {}`) hits a pre-existing parallel-executor bug that nulls every outer variable --
+// unrelated to what the scenario is checking. Asserting the symbol identity pins the rule directly instead.
+TYPED_TEST(TestSymbolGenerator, PatternComprehensionInAggregatingSubqueryWhereShadowsUnimportedOuterVariable) {
+  // MATCH (p) CALL { MATCH (t) WITH t, count(*) AS c WHERE size([(p)-->() | 1]) > 0 RETURN t } RETURN p
+  auto *outer_p = NODE("p");
+  auto *comprehension = SUBQUERY_COMPREHENSION("p");
+  auto *subquery = SINGLE_QUERY(MATCH(PATTERN(NODE("t"))),
+                                WITH(IDENT("t"), AS("t"), COUNT(LITERAL(1), false), AS("c")),
+                                WHERE(GREATER(comprehension, LITERAL(0))),
+                                RETURN(Aliased(NEXPR("t", IDENT("t")))));
+  auto *query = QUERY(SINGLE_QUERY(MATCH(PATTERN(outer_p)), CALL_SUBQUERY(subquery), RETURN("p")));
+
+  auto symbol_table = MakeSymbolTable(query);
+
+  const auto outer_symbol = symbol_table.at(*outer_p->identifier_);
+  const auto inner_symbol = symbol_table.at(ComprehensionStartNode(comprehension));
+  EXPECT_NE(outer_symbol, inner_symbol) << "an aggregating WHERE must not correlate the un-imported outer `p` either";
+  EXPECT_EQ(outer_symbol.name(), inner_symbol.name());
+}
+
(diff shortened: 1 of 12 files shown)


imilinovic:
### Tracking
- [x] **[Link to Epic/Issue]** #4161 #4364 #4376 #4180 #4378 #4339 #4350 #4365 #4394 #4388 #3955 #4160

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - **Fixed:** a pattern comprehension in the `WHERE` or `ORDER BY` of a `WITH` now correlates with the bound variable instead of re-scanning the whole graph, so `size([(n)-->(m) | ...])` filters and sorts on the right rows. With an aggregation in the same `WITH` it previously failed with an internal error. [#4487](https://github.com/memgraph/memgraph/pull/4487)
    - **Fixed:** a `CALL {}` subquery no longer overwrites a variable of the enclosing query that it did not import. A plain `MATCH`, a pattern comprehension or a write inside the subquery could leave the outer variable as `null` in every row. [#4487](https://github.com/memgraph/memgraph/pull/4487)
    - **Fixed:** a named path is no longer overwritten by a path of the same name declared inside a subquery or inside a pattern comprehension. [#4487](https://github.com/memgraph/memgraph/pull/4487)
    - **Fixed:** a `WITH` with both `ORDER BY` and `WHERE` no longer evaluates the filter against a stale row when the `WHERE` reads a variable the projection did not output. [#4487](https://github.com/memgraph/memgraph/pull/4487)
    - **Fixed:** two queries could abort the server process and drop every session, reachable through `EXPLAIN` alone and so without write permission — a variable-length pattern comprehension after a write in the same query part, and an `EXISTS {}` subquery reusing an already bound relationship. Both now raise a query error. [#4487](https://github.com/memgraph/memgraph/pull/4487)
    - **Breaking:** inside a `CALL {}` subquery, a pattern that mentions a variable of the enclosing query without importing it now declares that name afresh rather than silently reusing the caller's, so a query relying on the old behaviour returns **different results without raising an error**. This makes a bare `CALL {}` match its documented scope, identical to `CALL () {}`. Import explicitly with `CALL (n) {`, `CALL (*) {`, or a leading `WITH n`. Non-pattern references to an un-imported variable were already rejected as unbound. The `exists()` pattern form now raises a clear error instead of an internal one. [#4487](https://github.com/memgraph/memgraph/pull/4487)
    - **Breaking:** a pattern comprehension may no longer read a node the same `CREATE` clause binds — `CREATE (q:T {c: size([(q)-->(m) | m])})` now raises `Entity 'q' cannot be created and referenced by a pattern comprehension in the same clause.` No operator placement can work, since the operator that binds the symbol is the one that reads the result. Use an earlier clause: `CREATE (q:T) SET q.c = size([(q)-->(m) | m])`. [#4487](https://github.com/memgraph/memgraph/pull/4487)
    - **Breaking:** a pattern comprehension may no longer reuse an already bound relationship variable — `MATCH (a)-[r]->(b) WITH a, r WHERE size([(a)-[r]->(y) | y]) > 0` now raises `Cannot use the already bound relationship 'r' in a pattern here.` Bound *nodes* still correlate; only relationships are affected. Previously this aborted the server process. [#4487](https://github.com/memgraph/memgraph/pull/4487)
    - **Breaking:** a variable-length pattern comprehension may no longer expand from a node a write clause of the same query part binds — for example `CREATE (a)-[:R]->(b) RETURN [(a)-[*1..2]->(x) | x]`. Variable-length expansion cannot read data written by its own query part, so no valid plan exists; close the write clause with a `WITH` first. Previously this aborted the server process. [#4487](https://github.com/memgraph/memgraph/pull/4487)

tests/unit/query_semantic.cpp
```diff
@@ -1604,3 +1604,321 @@ TYPED_TEST(TestSymbolGenerator, ListComprehensionInWith) {
...
+TYPED_TEST(TestSymbolGenerator, PatternComprehensionOverItsOwnNodesInsideCreateIsAccepted) {
+  // CREATE (q:L {c: [(z)-->() | 1]})
+  // Nothing the CREATE binds is referenced, so the comprehension stands alone.
+  auto *created = WithProperty(this->storage, NODE("q", "L"), "c", COMPREHENSION_OVER("z"));
+  auto *query = QUERY(SINGLE_QUERY(CREATE(PATTERN(created))));
+
+  EXPECT_NO_THROW(MakeSymbolTable(query));
+}
+
+#undef SELF_COMPREHENSION
```
colinbarry: Typo: assume you meant `COMPREHENSION_OVER`. (Not that `undef`'ing at the end of a .cpp means much!)

colinbarry (approved):
Approved (with one typo to fix.) 🚀

Issue #4478, opened 2026-07-24
Memgraph container crashes (SIGSEGV) when executing a list comprehension with an EXISTS subquery


hkjiang26:
**Memgraph version**

Memgraph `3.12.0`

**Environment**

- Memgraph running in Docker
- Query executed through the Memgraph Web UI
- Host OS: Windows with Docker Desktop
- Architecture: x86_64
- No external driver

**Describe the bug**

Executing the following query causes the Memgraph Docker container to stop:

```cypher
RETURN [p = ()-[]->() WHERE EXISTS { RETURN 1 } | p];
```

The issue can be reproduced on an empty database without creating any data.

The same query executed in Neo4j returns `0`.

**To Reproduce**

1. Start Memgraph `3.12.0` in Docker.
2. Open the Memgraph Web UI.
3. Execute:

```cypher
RETURN [p = ()-[]->() WHERE EXISTS { RETURN 1 } | p];
```

4. The Memgraph container stops.

Container status:

```text
exited ExitCode=139 OOMKilled=false Error=
Exit code 139 indicates a segmentation fault (SIGSEGV).
```

**Expected behavior**

The query should return:

```text
[]
```

The container should remain running.

**Logs**

No additional query error was printed before the container stopped.

**Additional context**

The query was executed on an empty database. No dataset or external driver is required.

Neo4j executes the same query successfully and returns:

```text
[]
```

**Verification Environment**

Please verify the fix with:

- Memgraph Docker image
- Memgraph Web UI
- No external driver

PR #4490, opened 2026-07-28
fix: prevent SIGSEGV for RETURN-only EXISTS subqueries

### LINKED ISSUE #4485: Memgraph crashes (SIGSEGV) when executing exists {} subquery expression
**Memgraph version**

Memgraph v3.12.0

**Environment**

- OS: Windows with Docker Desktop
- Architecture: x86_64
- Deployment: Docker container
- Client: Memgraph Lab (web interface)

**Describe the bug**

Memgraph crashes when executing the following valid Cypher query:

```cypher
MATCH (n) WHERE exists { RETURN n } RETURN n
````

The same query executes successfully in Neo4j.

The crash happens regardless of whether the database contains nodes or is empty.

After the crash, the container exits with:

```
ExitCode=139
OOMKilled=false
```

The Memgraph process terminates due to a segmentation fault.

**To Reproduce**

1. Start Memgraph v3.12.0 in Docker Desktop.
2. Open Memgraph Lab.
3. Execute:

```cypher
MATCH (n) WHERE exists { RETURN n } RETURN n
```

4. Memgraph crashes and the container exits.

**Expected behavior**

Memgraph should execute the query successfully or return an error without crashing.

**Logs**

```
PS E:\workspace> docker inspect memgraph-poc --format "{{.State.Status}} ExitCode={{.State.ExitCode}} OOMKilled={{.State.OOMKilled}} Error={{.State.Error}}"

```

**Additional context**

The query uses the Cypher `exists {}` subquery expression.

Neo4j executes this query successfully, but Memgraph v3.12.0 crashes during query processing.

This issue might be related to #4478 , where a similar SIGSEGV occurs when EXISTS {} is used inside a list comprehension predicate.

Nithin078:
## Summary

Fixes a segmentation fault when planning Cypher queries that use an `EXISTS` subquery with only a `RETURN` clause (no `MATCH`), for example:

```cypher
MATCH (n) WHERE exists { RETURN n } RETURN n
```

### Root cause

In existential subqueries, `RETURN` is intentionally omitted (EXISTS only cares whether a row is produced). For **RETURN-only** subqueries that left `input_op` as `nullptr`, the planner then called `input_op->OutputSymbols(...)`, which null-dereferenced and terminated Memgraph with SIGSEGV (exit code 139).

The same path affected pattern-comprehension filters such as:

```cypher
RETURN [p = ()-[]->() WHERE EXISTS { RETURN 1 } | p];
```

### Fix

1. **GenReturn**: when inside an EXISTS subquery and there is no prior operator, create a `Once` with outer bound symbols so the subquery still has a valid row source.
2. **Plan / ExtractPatternFilters**: null-guard and defensive `Once` so a null plan cannot reach `OutputSymbols` / `Limit` / `EvaluatePatternFilter`.
3. **UsedSymbolsCollector**: also collect free symbols from `RETURN` / `WITH` / `WHERE` inside EXISTS subqueries so e.g. `EXISTS { RETURN n }` waits for outer `n` to be bound before the filter is applied.

### Tests

- Unit planner tests: `ExistsSubqueryReturnOnly`, `ExistsSubqueryReturnOnlyInPatternComprehension`
- GQL Behave scenarios in `memgraph_exists.feature` for RETURN-only EXISTS (with data, empty graph) and EXISTS inside a pattern comprehension filter

## Related issues

- Fixes https://github.com/memgraph/memgraph/issues/4485
- Related to https://github.com/memgraph/memgraph/issues/4478

## Test plan

- [ ] `memgraph__unit__query_plan` — filter `*ExistsSubqueryReturnOnly*`
- [ ] Behave: `memgraph_exists.feature` RETURN-only / pattern comprehension scenarios
- [ ] Manual: `MATCH (n) WHERE exists { RETURN n } RETURN n` does not crash (empty and non-empty DB)
- [ ] Manual: `RETURN [p = ()-[]->() WHERE EXISTS { RETURN 1 } | p]` does not crash

### DIFF
--- a/src/query/plan/preprocess.hpp
+++ b/src/query/plan/preprocess.hpp
@@ -102,9 +102,12 @@ class UsedSymbolsCollector : public HierarchicalTreeVisitor {
         atom->Accept(*this);
       }
     } else if (exists.HasSubquery()) {
-      // For subqueries, we need to collect symbols from the subquery
-      auto *single_query = exists.GetSubquery()->single_query_;
-      if (single_query) {
+      // Collect free (outer) symbols referenced by the subquery. Newly introduced MATCH
+      // variables are marked user_declared_=false by SymbolGenerator and are skipped by Visit.
+      // RETURN/WITH expressions must also be visited so that e.g. EXISTS { RETURN n } waits for
+      // outer `n` to be bound (and does not crash / mis-order the filter). See #4485.
+      auto collect_from_single_query = [this](SingleQuery *single_query) {
+        if (!single_query) return;
         for (auto *clause : single_query->clauses_) {
           if (auto *match = utils::Downcast<Match>(clause)) {
             for (auto *pattern : match->patterns_) {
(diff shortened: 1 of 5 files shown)


hkjiang26 (approved):
Thanks for the fix!

Nithin078:
No Problem @hkjiang26 , please do let me know if there is anything else you want me to do before the merge.

PR #4497, opened 2026-07-31
perf(query): park the Bolt worker on a contended accessor acquire instead of pinning it

andrejtonev:
A query that needs a contended UNIQUE/READ_ONLY storage accessor during `Prepare` parks its
Bolt worker — returning it to the `PriorityThreadPool` — instead of blocking it in
`try_lock_for` for the whole `--storage-access-timeout-sec`. One slow DDL/index/storage-mode
operation no longer costs the instance a worker per waiting query.

Measured on the discriminating shape (one long-held WRITE accessor, more READ_ONLY contenders
than there are Bolt workers):

| | probes completed | p50 |
|---|---|---|
| parking on | 373 | **0.0018s** |
| parking off | 26 | 0.9501s |

~500x, and the probe count matters as much as the latency: the same window serviced 14x more
unrelated work.

**What the feature doanges *how* a blockedquery waits, not
*whether* it is admitannot acquire withinits timeout still
throws the same errorsn't holding a workerhostage while it
waits. The win is worck admission.

## Shape

- **Substrate** — `Ta`DeadlineParkRegistry`, `ParkState` with a                                                       two-party delivery must_run_` queue forresumes.                                                               - **Storage** — virtucquire`; parked waiters are woken from                                                           `ResourceLock`'s owobserver, so noadmitting transition can                                                 forget to wake.
- **Query/Bolt** — `Interpreter::PrepareCoro` splits Prepare into threephases around a singl
  `co_await`; `Session::DrivePreparedRun`/`RunLoop` drive it on a pool worker.

A resume is **not** pPinning conflated*when* a resume may run
with *where*; only thw a `ParkState` gaterather than an affinity.
That un-pin is what stency — and thereforeits timeout
enforcement — from beacklog.

## Scope limits, all

- One park site: the e. Nothing parks during Pull/execute.
- HIGH-priority querilocking path unchanged.
- `SHOW TRANSACTIONS` cannot see a parked autocommit query — the tx-id is
assigned in Phase 3.

## Verification

- 216 unit tests / 9 ** (216 tests).
- e2e: 7 tests / 3 workloads, including a flag-off control asserting the
*opposite* outcome on
  the identical scenario, and a graceful-shutdown-with-a-park-in-flight
test against a real
  process.
- Two stress arms(`tests/stress/standalone/native/workloads/coro_prepare_park*`) with both
  gates mutation-veri fails against aflag-off build, and the
  session-leak gate fed past its expectedlifetime.

Reviewed in three adversarial passes. Defect yield declined each time — the third found one
MEDIUM in the second's own fix, plus nits — which is why review stopped there.


andrejtonev:
replaced with #4685 and #4777

Issue #4414, opened 2026-07-14
`EXISTS {}` correlated via a WHERE predicate: internal error "Expected to generate all filters!", or silently wrong results with an importing WITH


GregoireIO:
**Memgraph version**

Hit on **3.7.1**. Also reproduced on **3.11.0** (`memgraph/memgraph:latest` at time of writing), so it is present in the current release.

**Environment**

* OS: Debian GNU/Linux 12 (bookworm), kernel 6.1
* Architecture: x86_64
* Deployment: official Docker images — `memgraph/memgraph:3.7.1` and `memgraph/memgraph:latest`, started with `--telemetry-enabled=false`
* Client/driver: `mgconsole` over Bolt (also observed through the Go Bolt driver)
* Storage mode: default `IN_MEMORY_TRANSACTIONAL` (also reproduces in `IN_MEMORY_ANALYTICAL`)

**Describe the bug**

An `EXISTS { ... }` subquery that correlates with an outer variable **through a `WHERE` predicate** (rather than through a relationship pattern) misbehaves in two different ways:

1. **Without an importing `WITH`** — the query fails with what looks like an internal planner error:

   ```
   Expected to generate all filters! Please contact Memgraph support as this
   scenario should not happen and is very likely a bug in the query engine!
   ```

2. **With an importing `WITH`** — no error is raised, but the query **silently returns wrong results**. The correlated predicate behaves as though the outer variable were unbound/NULL, so no rows qualify and the subquery evaluates to false for every row.

The second case is the more dangerous of the two: it produces incorrect data with no error, no warning, and nothing in the server log.

Correlating through a **relationship pattern** (`EXISTS { MATCH (p)-[:R]->(c) }`) works correctly, and an **uncorrelated** `EXISTS { ... }` works correctly. Only correlation via a `WHERE` predicate on an otherwise-unconnected pattern is affected.

**To Reproduce**

Full self-contained reproduction:

```cypher
// --- dataset ---
CREATE (:Parent {name: 'a'});
CREATE (:Parent {name: 'b'});
CREATE (:Child  {name: 'a'});

// Only Parent 'a' has a same-named Child, so every query below should yield 1.

// --- ground truth: plain correlated join (works, returns 1) ---
MATCH (p:Parent)
MATCH (c:Child) WHERE c.name = p.name
RETURN count(DISTINCT p);
// => 1   ✅ correct

// --- BUG 1: correlated EXISTS without importing WITH -> internal error ---
[...]

PR #4499, opened 2026-08-03
fix(query): drain pattern comprehensions at the clause that reads them

### LINKED ISSUE #4134: Pattern comprehension may ignore a relationship created earlier in the same clause
**Memgraph version**
Observed on:

- Memgraph `3.9.0`

Docker image used locally:
- `memgraph/memgraph:3.9.0`

**Environment**
- Host OS: Windows 10
- Architecture: x86_64
- Deployment: Docker
- Connection method: Bolt
- Client used for direct reproduction: Python `neo4j` Bolt driver via the local differential-testing harness

**Describe the bug**
Memgraph may fail to expose a relationship to a pattern comprehension when that relationship was created earlier in the same clause.

In the minimized repro below, the query:

1. matches an existing `Person` and an existing `City`
2. creates a `WORKS_IN` relationship between them
3. immediately returns a pattern comprehension over that new relationship

Neo4j returns the expected city name.
Memgraph returns an empty list.

This does not look like a generic write-visibility problem, because a nearby control that inserts a `WITH p` pipeline boundary before the `RETURN` makes Memgraph behave correctly.

**To Reproduce**
2. Run:

```cypher
MATCH (n) DETACH DELETE n;
CREATE (:Person {name:'Charlie'}), (:City {name:'New York'});

MATCH (p:Person {name:'Charlie'}), (c:City {name:'New York'})
CREATE (p)-[:WORKS_IN]->(c)
RETURN p.name AS name,
       [ (p)-[:WORKS_IN]->(city) | city.name ] AS cities;
```

**Expected behavior**
Observed Neo4j result:

```text
Charlie, ["New York"]
```

Memgraph should return the same row.

**Actual behavior**
Observed Memgraph result:

```text
Charlie, []
```

So the relationship is created, but the immediately following pattern comprehension does not see it.

**Control query 1**
If the query returns the matched endpoint directly instead of using a pattern comprehension, both engines behave correctly:

```cypher
MATCH (n) DETACH DELETE n;
CREATE (:Person {name:'Charlie'}), (:City {name:'New York'});

MATCH (p:Person {name:'Charlie'}), (c:City {name:'New York'})
CREATE (p)-[:WORKS_IN]->(c)
RETURN p.name AS name, c.name AS city;
```

```text
Charlie, New York
```

So the failure is not simply that the `CREATE` did not happen.

**Control query 2**
If the same query inserts a `WITH p` boundary after the relationship creation, Memgraph behaves correctly:

```cypher
MATCH (n) DETACH DELETE n;
CREATE (:Person {name:'Charlie'}), (:City {name:'New York'});

MATCH (p:Person {name:'Charlie'}), (c:City {name:'New York'})
CREATE (p)-[:WORKS_IN]->(c)
WITH p
RETURN p.name AS name,
       [ (p)-[:WORKS_IN]->(city) | city.name ] AS cities;
```

```text
Charlie, ["New York"]
```

This makes the boundary much sharper:

- the relationship creation itself succeeds
- the same pattern comprehension works after a pipeline boundary
- the failure is specifically that the pattern comprehension cannot see the newly created relationship immediately in the same clause

**Control query 3**
This is not limited to `CREATE`. `MERGE` shows the same behavior:

```cypher
MATCH (n) DETACH DELETE n;
CREATE (:Person {name:'Charlie'}), (:City {name:'New York'});

MATCH (p:Person {name:'Charlie'}), (c:City {name:'New York'})
MERGE (p)-[:WORKS_IN]->(c)
RETURN p.name AS name,
       [ (p)-[:WORKS_IN]->(city) | city.name ] AS cities;
```

Observed behavior:

- Neo4j returns:
  - `Charlie, ["New York"]`
- Memgraph returns:
  - `Charlie, []`

So the problem does not look specific to `CREATE` versus `MERGE`.

**Control query 4**
This is also not a blanket failure of pattern comprehensions after writes. If the entire path is created in a single `CREATE` pattern, both engines behave correctly:

```cypher
MATCH (n) DETACH DELETE n;

CREATE (p:Person {name:'Charlie'})-[:WORKS_IN]->(:City {name:'New York'})
RETURN p.name AS name,
       [ (p)-[:WORKS_IN]->(city) | city.name ] AS cities;
```

```text
Charlie, ["New York"]
```

So the boundary seems to be more specific: the failure appears when the relationship is created between already matched endpoints and then immediately consumed by a pattern comprehension in the same clause.

imilinovic:
## The defect

A pattern comprehension is planned as a `RollUpApply` spliced onto an operator chain, and it must land on the chain that *reads* its result. `pending_comprehensions` kept no record of which clause a comprehension came from, so the drain could answer *"can this be planned yet?"* but never *"should it be, here?"* — it fired at whatever clause happened to satisfy the dependencies. Every symptom below is that one gap.

**The Accumulate freeze.** Drained at the write clause, so the `RollUpApply` lands below the `Accumulate` that clause forces. `Accumulate`'s remember set is `body.used_symbols()` and never holds a comprehension *result* symbol, so on replay every row sees the last input row's list.

```cypher
CREATE (a:Person {name:'Alice'}), (b:Person {name:'Bob'}), (c:Person {name:'Carol'});
MATCH (a:Person {name:'Alice'})
CREATE (a)-[:ACTED_IN]->(:Movie {title:'M1'}), (a)-[:ACTED_IN]->(:Movie {title:'M2'});
MATCH (b:Person {name:'Bob'}) CREATE (b)-[:ACTED_IN]->(:Movie {title:'M3'});

MATCH (p:Person) SET p.z = 1
WITH p.name AS n, [(p)-[:ACTED_IN]->(m) | m] AS lst
RETURN n, size(lst) AS movies ORDER BY n;
```

| | Alice | Bob | Carol |
|---|---|---|---|
| master | 0 | 0 | 0 |
| this PR | **2** | **1** | 0 |

The plans differ only in where the `RollUpApply` sits:

```
master                            this PR
 * Produce {n, size(lst)}          * Produce {n, size(lst)}
 * Produce {n, lst}                * Produce {n, lst}
 * Accumulate                      * RollUpApply        <-- above the write
 * SetProperty                     |\
 * RollUpApply   <-- below it      | * Produce {anon2}
 |\                                | * Expand (p)-[anon2:ACTED_IN]->(m)
 | * Produce {anon2}               | * Once
 | * Expand (p)-[anon2:...]->(m)   * Accumulate
 | * Once                          * SetProperty
 * Filter (p :Person)              * Filter (p :Person)
 * ScanAll (p)                     * ScanAll (p)
```

**Same clause, wrong side of the write (#4134).** The `CREATE` stacks above the `RollUpApply`, so the branch sees the pre-write graph.

```cypher
CREATE (:Person {name:'Charlie'}), (:City {name:'New York'});

MATCH (p:Person {name:'Charlie'}), (c:City {name:'New York'})
CREATE (p)-[:WORKS_IN]->(c)
RETURN p.name AS name, [(p)-[:WORKS_IN]->(city) | city.name] AS cities;
-- master:  Charlie, []
-- this PR: Charlie, ["New York"]
```

Closes #4134

## What this does

**Records the originating clause.** `PatternComprehensionMatching` gains `origin_clause`, stamped at the per-clause collect site where the attribution was already available and being discarded. Every main-chain drain is restricted to the comprehensions its own clause originates, so one clause can no longer take another's.

**Gives `CALL … YIELD` a drain.** It had none: `CallProcedure::Accept` descends into `where_`, so the matching was collected and then silently abandoned, and the clause is a query-part boundary so nothing downstream could take it either. It now drains twice, because arguments and `YIELD … WHERE` need opposite splice points relative to the operator. `DepsSatisfied` chooses: YIELD symbols are registered as bound-by-this-query-part but not yet bound, so a comprehension reading one is refused below the operator — where the slot is unwritten, and stale from the previous row after the first — and accepted above it, below the `Filter`, evaluated per procedure output row. Everything else stays below, evaluated once per input row.

**Makes the view the write history alone.** The early drain reached for `View::NEW` only when a now-deleted helper also held, while the on-demand path had no such conjunct, so the same comprehension got a different view depending on which site planned it. A clause sees the effects of the clauses before it, and within one command no `AdvanceCommand` folds a write into `View::OLD`. The FOREACH body now inherits the caller's write history — without it a semantically inert `WHERE` in the pattern flipped the view — and a MERGE marks itself a write before its own drain, as every other write site does. The variable-length carve-out is untouched.

**Splices MERGE branch comprehensions into their branch.** `ON CREATE`/`ON MATCH` comprehensions originate at the `MERGE`, so they are subtracted from the main chain and left to `GenMerge`, which splices each after `GenCreateForPattern` — so `ON CREATE` sees what the MERGE just created, and nothing is evaluated on rows whose branch never runs. Applies to a MERGE nested in a FOREACH body too.

**Completes `external_symbols`.** `DepsSatisfied` decides from it, but it was built from the comprehension's filter and result expression only — a correlation in a pattern property map, a variable-length bound, or a nested comprehension's filter was invisible, and the silent wrong answer above was reachable from those positions. Fixed at the set rather than by widening a predicate, which would have taught one caller four AST positions with no guarantee of a fifth and left `DepsSatisfied` wrong everywhere else.

## Behaviour changes

- A comprehension read by a later clause is evaluated at that clause. Reading `size([(n)-->(m) | ...])` after a write in the same query part previously returned `0` for every row, or no rows.
- A comprehension in `CALL … YIELD … WHERE` is evaluated instead of abandoned; it previously failed with an internal error. When it reads a yielded name it previously returned wrong results silently, order-dependently.
- An uncorrelated comprehension now sees writes made earlier in the same query part, and one in a `MERGE` pattern sees what earlier rows of that `MERGE` created.
- **A comprehension expanding from a node a preceding clause deleted now raises** `Trying to get relationships of a deleted node`, where it previously returned a count of relationships that no longer existed. It now behaves as Memgraph's own plain expansion from a deleted node already does — `MATCH (n:A) DETACH DELETE n WITH n MATCH (n)-[:R]->(m)` raises the same error on master. Measured against Neo4j 2026.02.2, which returns `0` for both that plain expansion and the comprehension while raising on property and label reads: so Memgraph's divergence here is pre-existing and wider than comprehensions, and this change removes an inconsistency inside Memgraph rather than adding one against Neo4j. Aligning with Neo4j would mean making `Expand` treat a deleted source as no rows, which is an operator change beyond this PR.

## Not fixed here

- A comprehension in a **`MATCH`/`MERGE` pattern property** with a user-declared internal atom throws *"Expected to generate all filters!"*: `Filters::CollectPatternFilters` puts the comprehension's internal pattern symbols into the property filter's `used_symbols` and nothing binds them. Pre-existing, affects `MATCH` equally, confirmed by re-running with this PR's changes stashed. Related to #4414 only by sharing that error message. The all-anonymous form works.
- `write_occurred` resets at any query-part boundary that does not advance the command (`UNWIND`, `CALL … YIELD`, `LOAD *`), so the first drain after one reads `View::OLD`. Pre-existing.
- The FOREACH main chain and body share one gating set, so an uncorrelated body comprehension is hoisted out of the loop and evaluated once. Forced by `origin_clause` granularity: `ForeachCursor::Pull` evaluates the list expression before writing the loop variable, so a list-expression comprehension *must* drain on the main chain, and the collect site records which clause but not where in it. Stamping position instead would delete the bespoke subtractions and unify every site — its own PR, since it changes body hoisting.
- A nested FOREACH body restarts `GenMerge`'s ON MATCH narrowing, so a first body clause erases the merge symbols even when an enclosing body wrote. Master's behaviour too.

### DIFF
--- a/src/query/frontend/ast/ast.hpp
+++ b/src/query/frontend/ast/ast.hpp
@@ -2566,7 +2566,7 @@ class CallProcedure : public memgraph::query::Clause {
   std::vector<memgraph::query::Identifier *> result_identifiers_;
   memgraph::query::Expression *memory_limit_{nullptr};
   size_t memory_scale_{1024U};
-  bool is_write_;
+  bool is_write_{false};
   bool void_procedure_{false};
   memgraph::query::Where *where_{nullptr};
 
--- a/src/query/frontend/semantic/symbol_generator.cpp
+++ b/src/query/frontend/semantic/symbol_generator.cpp
@@ -242,11 +242,13 @@ bool SymbolGenerator::PostVisit(CypherUnion &cypher_union) {
 
 bool SymbolGenerator::PreVisit(Create &) {
   scopes_.back().in_create = true;
+  create_clause_symbols_.clear();
   return true;
 }
 
 bool SymbolGenerator::PostVisit(Create &) {
   scopes_.back().in_create = false;
+  create_clause_symbols_.clear();
   return true;
 }
 
@@ -523,6 +525,10 @@ SymbolGenerator::ReturnType SymbolGenerator::Visit(Identifier &ident) {
     // A shadowed outer name must not resolve to the enclosing query's symbol, so declare it in this scope.
     symbol = shadows_outer_name ? CreateSymbol(ident.name_, ident.user_declared_, type)
                                 : GetOrCreateSymbol(ident.name_, ident.user_declared_, type);
+    // An atom that was not already in scope is declared here, so this CREATE clause is what binds it.
+    if (scope.in_create && !name_in_scope) {
+      create_clause_symbols_.insert(symbol);
+    }
   } else if (scope.in_pattern && !scope.in_pattern_atom_identifier && scope.in_match) {
     if (scope.in_edge_range && scope.visiting_edge && scope.visiting_edge->identifier_ &&
         scope.visiting_edge->identifier_->name_ == ident.name_) {
@@ -549,6 +555,13 @@ SymbolGenerator::ReturnType SymbolGenerator::Visit(Identifier &ident) {
       throw UnboundVariableError(ident.name_);
     }
     symbol = GetOrCreateSymbol(ident.name_, ident.user_declared_, Symbol::Type::ANY);
+  }
+
+  // The operator that binds a symbol declared by this CREATE is the same one that reads the comprehension's result, so
+  // the frame slot is still unwritten when the clause needs it. There is no placement that works - reject instead.
+  if (scope.in_pattern_comprehension && create_clause_symbols_.contains(symbol)) {
+    throw SemanticException(
+        "Entity '{}' cannot be created and referenced by a pattern comprehension in the same clause.", ident.name_);
   }
 
   ident.MapTo(symbol);
--- a/src/query/frontend/semantic/symbol_generator.cpp
+++ b/src/query/frontend/semantic/symbol_generator.cpp
@@ -208,6 +208,7 @@ bool SymbolGenerator::PreVisit(CypherUnion &) {
   // Currently only CALL and EXISTS subqueries can contain complete queries with UNION.
   next_scope.in_call_subquery = prev_scope.in_call_subquery;
   next_scope.in_exists_subquery = prev_scope.in_exists_subquery;
+  next_scope.call_subquery_base = prev_scope.call_subquery_base;
   // Carry over explicit `CALL (v1, v2) { ... }` imports so each UNION branch
   // within the subquery still sees the imported variables.
   next_scope.call_subquery_imports = prev_scope.call_subquery_imports;
@@ -271,6 +272,8 @@ bool SymbolGenerator::PostVisit(CallProcedure &call_proc) {
 
 bool SymbolGenerator::PreVisit(CallSubquery &call_sub) {
   Scope new_scope{.in_call_subquery = true};
+  // The scope about to be pushed is the subquery's outermost one; names below it need an explicit import.
+  new_scope.call_subquery_base = scopes_.size();
 
   if (call_sub.has_variable_scope_) {
     // `CALL (...) { ... }`: resolve imports against the current outer scope
@@ -444,8 +447,10 @@ bool SymbolGenerator::PostVisit(Match &) {
 
 bool SymbolGenerator::PreVisit(Foreach &for_each) {
   const auto &name = for_each.named_expression_->name_;
+  auto const call_subquery_base = scopes_.back().call_subquery_base;
   scopes_.emplace_back(Scope());
   scopes_.back().in_foreach = true;
+  scopes_.back().call_subquery_base = call_subquery_base;
   for_each.named_expression_->MapTo(
       CreateSymbol(name, true, Symbol::Type::ANY, for_each.named_expression_->token_position_));
   return true;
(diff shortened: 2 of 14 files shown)


imilinovic:
### Tracking
- [x] #4487 has merged; this PR is based on `master` and its diff is its own seven commits.

### Standard development
- [x] Update unit/E2E tests
- [ ] Compare the [benchmarking results](https://bench-graph.memgraph.com/) between the master branch and this branch

### CI Testing Labels

### Documentation checklist
    - A pattern comprehension whose value a later clause reads is now evaluated at that clause rather than at whichever earlier clause happened to satisfy its dependencies. Queries that read `size([(n)-->(m) | ...])` after a write clause in the same query part previously returned `0` for every row, or no rows at all, because the value was frozen by the accumulation step; they now return correct results. A comprehension in a `RETURN` alongside a `CREATE` in the same query part now also sees that write. [#4499](https://github.com/memgraph/memgraph/pull/4499)
    - A pattern comprehension in the `WHERE` of a `CALL ... YIELD` clause is now evaluated instead of silently abandoned; it previously failed with an internal error. When such a comprehension reads a yielded name, it previously returned wrong results with no error - an empty list on the first row and the previous row's value afterwards, so the answer depended on row order. [#4499](https://github.com/memgraph/memgraph/pull/4499)
    - A pattern comprehension that correlates to nothing now also sees writes made earlier in the same query part, and one in a `MERGE` pattern sees what earlier rows of the same `MERGE` created. [#4499](https://github.com/memgraph/memgraph/pull/4499)
    - **Changed behaviour:** a pattern comprehension that expands from a node a preceding clause deleted now raises `Trying to get relationships of a deleted node`. It previously returned a count of relationships that no longer existed, because the comprehension was evaluated before the delete. [#4499](https://github.com/memgraph/memgraph/pull/4499)

colinbarry (approved):
Approved 👍

PR #4504, opened 2026-08-03
feat(query): support EXISTS subqueries in WITH/RETURN projections, WHERE and ORDER BY

### LINKED ISSUE #3385: "Exists cannot be used within WITH" error when using subquery with WHERE pattern
**Memgraph version** 3.6.1
**Environment**
Operating System: macOS
Deployment: Docker

**Describe the bug**
When running a subquery that includes a WITH clause followed by a WHERE pattern predicate, Memgraph throws the following error:
Not yet implemented: Exists cannot be used within WITH!
This indicates that pattern predicates inside a WHERE clause after a WITH statement are currently not supported, even though they are part of standard Cypher semantics.

**To Reproduce**
Run the following query:
CALL {
  MATCH (n1)--(n2)
  WITH n1.id AS alias0
  WHERE (n2) <-[:rt1]- (n2)
  RETURN n2 AS alias1
}
RETURN alias1
Observe the error message:
Not yet implemented: Exists cannot be used within WITH!

### LINKED ISSUE #4485: Memgraph crashes (SIGSEGV) when executing exists {} subquery expression

imilinovic:
`EXISTS { … }` was usable only in a `WHERE` hanging off `MATCH`/`OPTIONAL MATCH`. Everywhere else it failed with `Not yet implemented: Exists can only be used inside the WHERE clause!` — including the shape users most often want:

```cypher
MATCH (p:Person {name: 'Regina King'})
WITH p, EXISTS { MATCH (p)-[:ACTED_IN]->(m:Movie) WHERE m.title STARTS WITH 'J' } AS hasActed
RETURN p, hasActed;
```

This PR makes `EXISTS { … }` work in a `WITH`/`RETURN` projection, in `WITH … WHERE`, in `ORDER BY`, and inside an aggregate argument. The pattern form `exists((a)-[]->(b))` gains the same positions.

## How

The blocker was never the semantic gate. `EvaluatePatternFilter` — the operator that evaluates EXISTS today — is not a pipeline operator: its single `input_` **is** the branch, and its `Pull` ignores it and returns `true` unconditionally. It functions only as a `Filter` side branch, and `Produce`/`Aggregate` have no such vector. Making it pipeline-capable is a large refactor.

`RollUpApply` is already a pipeline operator, and its fold is a four-line loop. So instead: **give `RollUpApply` a bool fold** and splice it onto the main chain like any other operator. Every predicate is pushed inside the branch, so one pull is a sound witness.

This is not a shortcut around the intended design — it *is* the design. Timing is a static, per-position planner decision: a value going into a returned column has nothing to short-circuit, so **forced is the correct timing for a projection**. The `WHERE` path keeps the deferred closure and is untouched; its `EXPLAIN` output is byte-identical.

Around that:

- The per-clause collector's `getFilterMatchings()` was being discarded; it is now kept on `SingleQueryPart` and keyed by result symbol.
- `ReturnBodyContext` gains an `Exists` visit pair mirroring the pattern-comprehension one, with the same three positional buckets — projection below the `Produce`, `ORDER BY` below the `OrderBy` that reads it in its collection sweep, `WHERE` *above* the `OrderBy` and directly below the `Filter`. With an aggregation present every branch goes before the `Aggregate`, and its result symbol joins the remember set unless the aggregate consumes it.
- `MakeExistsBranch` is factored out of `MakeExistsFilter`, so both folds build the branch identically and the subquery form keeps its recursive `Plan()` under `in_exists_subquery`. That flag drives three separate behaviours — the body's `RETURN` being dropped, the `EmptyResult` wrapper being suppressed, and `GenWith` preserving outer-scope vertex/edge symbols — and sharing the builder means the new path inherits all three rather than diverging.
- A branch spliced after a write reads `View::NEW`. Because the branch's body is planned as a fresh query part with no write history of its own, that fact is carried explicitly, and every view decision inside the branch consults it — the body's own `MATCH`, the comprehensions and nested EXISTS its `WITH`/`RETURN` plans on demand, and the filter a body `MATCH`'s `WHERE` builds. Outside an EXISTS branch it all collapses to `View::OLD`.
- The symbol collector now walks an EXISTS body in full, so `Accumulate` and `OrderBy` remember the outer names the branch correlates to — including a correlation that lives only in the body's `WHERE`.
- The blanket `!in_where` precondition becomes a **positive allow-list, default-deny**. Dropping `in_with`/`in_return` would have accomplished nothing: the checks ran in a fixed order with `!in_where` first, so those two were already dead for the projection forms. A new per-scope lambda depth refuses the forced fold inside `all`/`any`/`none`/`single`, `reduce`, `extract`, a list comprehension and an edge lambda, where the branch would run once above the `Produce` and read an unbound element variable. `in_reduce` and `num_if_operators` stay as narrowed refusals with their own messages.
- Renames to match the shared vocabulary: `PatternComprehensionView` → `SubqueryView`, `PatternComprehensionPlanner` → `SubqueryBranchPlanner`, `PatternComprehensionContext` → `SubqueryContext`.
## Deliberate scope calls

- **`UNWIND` is refused**, though it is a natural member of the allow-list. It is not a return body, so there is no splice point — allowing it would land a matching that nothing drains, leaving the frame slot unwritten. The expression then reads it and throws *"Unexpected behavior: Exists expected a function, got null. Please report the problem on GitHub issues"* — the failure mode `CALL … YIELD … WHERE EXISTS{}` exhibits today, since it is the same undrained shape. A clear refusal beats an internal error inviting a bug report, which is what default-deny is for. Pinned as a refusal; opening it belongs with the `CASE` work, and needs a drain for EXISTS — the same change that would make `external_symbols` worth computing.
- **`external_symbols` is not computed for the EXISTS matchings.** Every allowed position is planned on demand at the clause that owns it, so no EXISTS reaches a drain and the dependency check is never consulted for one — the field would be write-only.
- **`CALL … YIELD … WHERE EXISTS{…}` is unchanged**: still accepted, still never evaluated, because that clause's `Filter` gets no pattern-filter vector and it is a query-part boundary. Closing it needs a dependency check for EXISTS, which is what would justify computing `external_symbols`.

## Not covered

The body-`RETURN` discard and the `CASE`/`SET`/`FOREACH`/`reduce` positions are out of scope and tracked separately. A variable-length expansion from a write-bound variable reports an internal-sounding runtime error where the comprehension form gives a clear planner diagnostic; that asymmetry predates this PR and is reachable on master.

Closes #3385
Closes #4485


imilinovic:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - `EXISTS { ... }` and the pattern form `exists((a)-[]->(b))` now work in a `WITH`/`RETURN` projection, in `WITH ... WHERE`, in `ORDER BY` and inside an aggregate argument rather than only in a `MATCH`'s `WHERE`, so an existence check can be returned, ordered on or aggregated directly — write it inline as `WITH p, EXISTS { MATCH (p)-[:ACTED_IN]->() } AS hasActed` instead of rewriting it as an `OPTIONAL MATCH` plus a count. [#4504](https://github.com/memgraph/memgraph/pull/4504)

tests/gql_behave/tests/memgraph_V1/features/memgraph_exists.feature
```diff
@@ -865,3 +883,243 @@ Feature: WHERE exists
...
+          CREATE (:Person {name: 'Regina King'})-[:ACTED_IN]->(:Movie {title: 'Jerry Maguire'})
+          CREATE (:Person {name: 'Bob'})
+          """
+      When executing query:
+          """
+          MATCH (p:Person)
+          RETURN p.name AS name, count(*) AS c, EXISTS { MATCH (p)-[:ACTED_IN]->(:Movie) } AS h
+          ORDER BY name;
+          """
+      Then the result should be:
```
colinbarry, line 945: Because each aggregated group comprises one row, these tests doesn't feel like they are actually testing aggregated `EXISTS`. Maybe add a few more input rows so we get actual grouping?

colinbarry (approved):
Approved (with one minor comment about how aggregated the GQLBehave aggregated tests are) 🥇

PR #4508, opened 2026-08-04
perf(storage): skip the indexes a write cannot have dirtied

### LINKED ISSUE #4512: Skip GC index cleanup for indexes a write cannot have dirtied
GC index cleanup walks every entry of an index rather than only the obsolete ones, so visiting an index costs its whole size whether or not it holds anything to collect. Every index is visited whenever any write has happened, so most of that work cannot find anything. Cost is governed by index count and index size, neither of which the writes touch.

### Scope

Writes record what they touched, labels and property ids per index family, and GC visits only the indexes those could have dirtied. Vertex and edge index families are decided separately, since no label or property is shared between their keys: a write to an edge property cannot dirty a vertex index, and an edge type never dirties an edge index because an edge's type cannot change.

Aborts get the same narrowing. Which indexes a label or property appears in is derived once per set of indexes and shared by the aborts running against it, rather than rebuilt per abort.

### Acceptance criteria

- A write naming one label causes only the indexes on that label to be visited. Indexes on other labels are not visited.
- The same holds for a vertex property write, an edge property write, and a unique constraint: only the indexes keyed on the written property are visited.
- An edge property write visits no edge-type index, since such an entry holds no property and an edge's type cannot change.
- A vertex write visits no edge index, and an edge write visits no vertex index.
- Deleting a vertex or an edge still visits every index of that family.
- An index that is visited collects everything a full sweep of it would have. No entry that was collectable before this change is left behind.
- An abort undoes its index entries in an index created after an earlier abort already ran.
- `GCIndexSweeps` in `SHOW METRICS INFO` reports how many indexes each run visited.

### Out of scope

Deleted objects still force every index of the family to be visited, because no delta names an entry pointing at memory about to be freed, and leaving one behind is a dangling pointer rather than a missed saving. A workload deleting faster than the GC interval therefore sees no benefit from this. Narrowing that correctness gate needs the deleted ids to carry their labels, which is a separate piece of work.

### Observability

A counter reports how many individual indexes each GC run visited, exposed as `GCIndexSweeps` in `SHOW METRICS INFO`.

Ignition:
GC index cleanup walks every entry of an index rather than only the obsolete ones,
so visiting an index costs its whole size whether or not it holds anything to
collect. Every index was visited whenever any write had happened, so most of that
work could not have found anything.

Writes now record what they touched, labels and property ids per index family, and
GC visits only the indexes those could have dirtied. Deleted objects still force
every index to be visited, since no delta names an entry pointing at memory about
to be freed.

Aborts get the same narrowing: which indexes a label or property appears in is
derived once per set of indexes and shared by the aborts running against it, rather
than rebuilt per abort. A counter reports the individual indexes each GC run visits.


Ignition:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Garbage collection now cleans only the indexes your writes could have affected, instead of all of them, which reduces time spent in GC on databases with many indexes. [#4508](https://github.com/memgraph/memgraph/pull/4508)

src/utils/id_bitmap.hpp
```diff
@@ -0,0 +1,73 @@
+// Copyright 2026 Memgraph Ltd.
```
imilinovic, line 1: could this be replaced with `boost::dynamic_bitset`?

imilinovic (approved):
this + things we discussed on slack

src/utils/id_bitmap.hpp
Ignition: Decided to stay with dedicated type. Minor performance benefit using IdBitmap.

PR #4510, opened 2026-08-04
feat: HA analytical mode

as51340:
## What

Enables the analytical bulk-import workflow on an HA data instance: switch to `IN_MEMORY_ANALYTICAL`, bulk import, switch back to `IN_MEMORY_TRANSACTIONAL`, register replicas back. Previously this was a flat refusal — `Data instances cannot use analytical mode`.

Design and the evidence behind each choice are in `specs/ha-analytical-import.md`.

## Why removing the check wasn't enough

Two independent blockers, not one:

1. **Analytical writes never reach the WAL.** A replica attached across the episode silently misses the entire import.
2. **Unregistering a replica does not wipe it.** So a re-registered replica is non-empty and stale, and its WAL chain still looks intact — an analytical episode punches a timestamp hole without disturbing sequence numbering. Recovery would walk the chain, find no gap, skip the snapshot, and never deliver the imported data.

The first is fixed by gating. The second needed recovery itself to change.

## What changed

**Gates.** Analytical is reachable only when the instance holds MAIN and has zero registered replication clients; registration and unregistration are rejected up front while any database is analytical. Instance-wide and all-or-nothing. Rejection happens before any replication state is mutated — persisting the instance-level client while no per-database client can be created would leave the replica permanently unattached, since every retry then fails with `NAME_EXISTS`.

Enforced in depth, ending with a hard window close: both the final mode store and the replica client insert re-check under `replication_storage_clients_`' own lock, which makes "analytical with a live client" unrepresentable rather than merely unlikely.

**Self-correcting recovery.** The WAL is finalized when *entering* analytical, so the episode leaves a file-level signature. `GetRecoverySteps` then treats "the latest snapshot's `durable_timestamp` lies inside no WAL file's `[from, to]` range" as proof that no WAL can reproduce that data, and forces snapshot-based recovery. Derived entirely from the durability files, so it survives a main restart — an earlier in-memory flag did not.

**Switch-back durability.** The switch-back snapshot is a new durability base, not an increment on the old one. Superseded snapshots and WALs are archived to `.old` (or deleted when `--storage-backup-dir-enabled=false`) and WAL sequence numbering restarts at 0. The two are inseparable: wiping alone breaks recovery, since a chain whose first file has a non-zero sequence number is only accepted when some WAL predates the snapshot — and those are exactly the files just wiped. Restarting alone is worse, because surviving files share this UUID and the new numbers would collide with theirs, which no check anywhere catches.

If the snapshot fails, the switch now aborts and throws with the mode, the cached `ldt`, `wal_seq_num_` and every durability file untouched. Previously it carried on and published a timestamp for data that had no durable record, which the next recovery dropped silently.

**RPC client lifetime.** Came out of reviewing the teardown reordering above. `rpc::Client`'s socket, serialization lock and endpoint now live in a shared `Connection`, so a `StreamHandler` can outlive the `Client` that produced it — which it genuinely can, because `~TransactionReplication` releases the storage-clients read lock before destroying its streams. The socket is published through an `atomic<shared_ptr>` so a reconnect never destroys one that an interrupt or an in-flight RPC still holds.

This closes a use-after-free where unregistering a replica destroyed the shared `rpc_client_` under a live SYNC commit on another database. `Shutdown()` is now safe to call with an RPC in flight, and its declaration says so — the old contract ("call `Abort()`, then join the worker thread") was unsatisfiable on the SYNC path, where the RPC runs inline on the committing thread and there is no thread to join.

## Notes for reviewers

- `--storage-mode=IN_MEMORY_ANALYTICAL` at startup stays forbidden for data instances; analytical is reachable only via the runtime query from a transactional start.
- `REGISTER INSTANCE` against an analytical main still returns `SUCCESS` and is healed by the reconciliation loop. The data instance logs the cause on reject. No `StateCheckRes` version bump, no new `SHOW INSTANCES` column.
- A manual `CREATE SNAPSHOT` between import and switch-back was dropped from the recipe: the switch-back already writes a correctly stamped snapshot, while a manual one taken while analytical skips the digest check and carries a pre-import timestamp.
- Operator constraint, stated in the docs: register and unregister replicas only while in transactional mode.

### DIFF
--- a/src/replication_handler/replication_handler.cpp
+++ b/src/replication_handler/replication_handler.cpp
@@ -23,6 +23,8 @@
 
 #include <spdlog/spdlog.h>
 
+#include <algorithm>
+
 namespace memgraph::replication {
 
 using namespace std::chrono_literals;
@@ -364,36 +366,48 @@ auto ReplicationHandler::UnregisterReplica(std::string_view name) -> query::Unre
 auto ReplicationHandler::UnregisterReplica_(std::string_view name) -> query::UnregisterReplicaResult {
   try {
     auto locked_repl_state = repl_state_.TryLock();
-
-    auto const replica_handler = [](RoleReplicaData const &) -> query::UnregisterReplicaResult {
-      return query::UnregisterReplicaResult::NOT_MAIN;
-    };
-    auto const main_handler =
-        [this, name, &locked_repl_state](RoleMainData &mainData) -> query::UnregisterReplicaResult {
-      if (!locked_repl_state->TryPersistUnregisterReplica(name)) {
-        return query::UnregisterReplicaResult::COULD_NOT_BE_PERSISTED;
-      }
-      // Remove database specific clients
-      dbms_handler_.ForEach([name](dbms::DatabaseAccess db_acc) {
-        db_acc->storage()->repl_storage_state_.replication_storage_clients_.WithLock([&name](auto &clients) {
-          std::erase_if(clients, [name](const auto &client) { return client->Name() == name; });
-        });
-      });
-      // Remove instance level clients
-      auto const n_unregistered =
-          std::erase_if(mainData.registered_replicas_, [name](auto const &client) { return client.name_ == name; });
-
-      // Drop the per-instance replication throughput series so the metric maps don't grow unbounded.
-      metrics::Metrics().RemoveReplicationThroughput(name);
-
-      return n_unregistered != 0 ? query::UnregisterReplicaResult::SUCCESS
-                                 : query::UnregisterReplicaResult::CANNOT_UNREGISTER;
-    };
-
-    return std::visit(utils::Overloaded{main_handler, replica_handler}, locked_repl_state->ReplicationData());
+    return UnregisterReplicaLocked_(locked_repl_state, name);
   } catch (const utils::TryLockException & /* unused */) {
     return query::UnregisterReplicaResult::NO_ACCESS;
   }
+}
+
+auto ReplicationHandler::UnregisterReplicaLocked_(LockedReplState &locked_repl_state, std::string_view name)
+    -> query::UnregisterReplicaResult {
+  auto const replica_handler = [](RoleReplicaData const &) -> query::UnregisterReplicaResult {
+    return query::UnregisterReplicaResult::NOT_MAIN;
+  };
+  auto const main_handler = [this, name, &locked_repl_state](RoleMainData &mainData) -> query::UnregisterReplicaResult {
+    if (!locked_repl_state->TryPersistUnregisterReplica(name)) {
+      return query::UnregisterReplicaResult::COULD_NOT_BE_PERSISTED;
+    }
+    // Stop the instance-level client before destroying any storage client below. Its thread pool holds raw
+    // pointers to them, and only Shutdown aborts the in-flight recovery, drops the queue and joins the
+    // worker; the destructor further down would do it after the storage clients are already gone.
+    // ClientsShutdown observes the same order for the role transition.
+    auto const client_it = std::ranges::find_if(mainData.registered_replicas_,
+                                                [name](auto const &client) { return client.name_ == name; });
+    if (client_it != mainData.registered_replicas_.end()) {
+      client_it->Shutdown();
+    }
+    // Remove database specific clients
+    dbms_handler_.ForEach([name](dbms::DatabaseAccess db_acc) {
+      db_acc->storage()->repl_storage_state_.replication_storage_clients_.WithLock([&name](auto &clients) {
+        std::erase_if(clients, [name](const auto &client) { return client->Name() == name; });
+      });
+    });
+    // Remove instance level clients
+    auto const n_unregistered =
+        std::erase_if(mainData.registered_replicas_, [name](auto const &client) { return client.name_ == name; });
+
+    // Drop the per-instance replication throughput series so the metric maps don't grow unbounded.
+    metrics::Metrics().RemoveReplicationThroughput(name);
+
+    return n_unregistered != 0 ? query::UnregisterReplicaResult::SUCCESS
+                               : query::UnregisterReplicaResult::CANNOT_UNREGISTER;
+  };
+
+  return std::visit(utils::Overloaded{main_handler, replica_handler}, locked_repl_state->ReplicationData());
 }
 
 auto ReplicationHandler::GetRole() const -> replication_coordination_glue::ReplicationRole {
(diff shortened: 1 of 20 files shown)


src/storage/v2/inmemory/replication/recovery.cpp
```diff
@@ -137,6 +139,23 @@ std::optional<std::vector<RecoveryStep>> GetRecoverySteps(uint64_t replica_commi
...
 
+    // A snapshot timestamp inside no WAL file's range means its data was never written to a WAL
+    // (analytical-mode writes bypass it), so WAL-only recovery would silently skip it. The chain looks
+    // intact because an analytical episode punches a timestamp hole without disturbing seq numbering.
+    // The > replica_commit guard is load-bearing: a replica already past the snapshot would otherwise
+    // enter the branch below with first_useful_wal pointing past the snapshot and trip its
+    // "Broken data chain" assert.
+    if (latest_snapshot && latest_snapshot->durable_timestamp > replica_commit &&
+        !SnapshotTsCoveredByAnyWal(
+            wal_files, current_wal_from_timestamp, current_wal_timestamp, latest_snapshot->durable_timestamp)) {
```
andrejtonev, line 150: I am not sure this is safe across the transactional <-> analytical switching. 
Till now we never supported a broken wal chain and the ts checks would be enough, but now we are allowing the chain to be broken by analytical mode switch.
Example: 
a) we are in transactional and have wal1, snapshotA, wal2 (wal1 is required to have a ts before snapshot so we know we are recovering the complete chain). 
b) we switch to analytical, do data modification, create snapshotB.
c) we switch back to transactional. we now have wal1, snapshotA, wal2, (break in wal chain) snapshotB, wal3, wal4...
d) we create another snapshot now and we see that snapshot as being covered by wals. but there is a break in the wal chain and we can't replay it completely.

andrejtonev: Possible solution is to bump the wal number when switching storage modes and force the chain to break.

src/storage/v2/inmemory/storage.cpp
andrejtonev, line 2893: I know this is not part of the PR, but will these be converted to transitional errors or client errors at the client level?

src/replication_handler/replication_handler.cpp
```diff
@@ -352,38 +354,60 @@ auto ReplicationHandler::RegisterReplica(const ReplicationClientConfig &config)
...
+auto ReplicationHandler::UnregisterReplicaLocked_(LockedReplState &locked_repl_state, std::string_view name)
+    -> query::UnregisterReplicaResult {
+  auto const replica_handler = [](RoleReplicaData const &) -> query::UnregisterReplicaResult {
+    return query::UnregisterReplicaResult::NOT_MAIN;
+  };
+  auto const main_handler = [this, name, &locked_repl_state](RoleMainData &mainData) -> query::UnregisterReplicaResult {
+    if (!locked_repl_state->TryPersistUnregisterReplica(name)) {
+      return query::UnregisterReplicaResult::COULD_NOT_BE_PERSISTED;
+    }
+    // Stop the instance-level client before destroying any storage client below. Its thread pool holds raw
```
andrejtonev, line 384: lets talk about this 1-1

src/replication_handler/replication_handler.cpp
```diff
@@ -352,38 +354,60 @@ auto ReplicationHandler::RegisterReplica(const ReplicationClientConfig &config)
...
+      db_acc->storage()->repl_storage_state_.replication_storage_clients_.WithLock([&name](auto &clients) {
+        std::erase_if(clients, [name](const auto &client) { return client->Name() == name; });
       });
-      // Remove instance level clients
-      auto const n_unregistered =
-          std::erase_if(mainData.registered_replicas_, [name](auto const &client) { return client.name_ == name; });
+    });
+    // Remove instance level clients
+    auto const n_unregistered =
+        std::erase_if(mainData.registered_replicas_, [name](auto const &client) { return client.name_ == name; });
```
andrejtonev: we already found it, just reuse the itr

src/storage/v2/inmemory/storage.cpp
as51340: no idea

as51340: out of scope for now

src/storage/v2/inmemory/replication/recovery.cpp
as51340: fixed by wiping

src/replication_handler/replication_handler.cpp
as51340: fixed

src/rpc/client.cpp
```diff
@@ -13,27 +13,55 @@
...
-  // Latch the aborted flag first so any thread that subsequently tries to open a stream fails fast instead of
-  // reconnecting. Then shut down the socket to abort any pending read, write, or connect operations on another thread.
-  // Does NOT destroy the client object, so it is safe to call concurrently with an in-flight RPC.
-  aborted_.store(true, std::memory_order_release);
+Connection::Connection(io::network::Endpoint endpoint, communication::ClientContext *context,
+                       std::chrono::milliseconds const connect_timeout_ms)
+    : endpoint_(std::move(endpoint)), context_(context), connect_timeout_ms_(connect_timeout_ms) {}
+
+void Connection::Interrupt() {
   if (!client_) return;
```
andrejtonev: How is client_ protected in multi-threaded workloads?

src/storage/v2/inmemory/storage.cpp
andrejtonev, line 2936: snapshot can fail and we need to either try again or fail the switch

src/rpc/client.hpp
```diff
@@ -224,18 +298,18 @@ class Client {
         });
 
         if (!maybe_message_header) {
-          self_->client_->ShiftData(response_data_size);
+          conn_->client_->ShiftData(response_data_size);
           throw SlkRpcFailedException();
           ;
```
andrejtonev: random `;`

src/rpc/client.cpp
as51340: it wasn't, now it is by the usage of atomic_shared_ptr\

src/storage/v2/inmemory/storage.cpp
as51340: fixed

andrejtonev (approved):

as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Enables the analytical bulk-import workflow on an HA data instance: switch to IN_MEMORY_ANALYTICAL, bulk import, switch back to IN_MEMORY_TRANSACTIONAL, register replicas back. Previously this was a flat refusal — Data instances cannot use analytical mode. [#4510](https://github.com/memgraph/memgraph/pull/4510)
- [x] **[ Documentation PR link ](https://github.com/memgraph/documentation/pull/1724)**

PR #4518, opened 2026-08-04
fix(storage): hold vector-search pins for the lifetime of their results

### LINKED ISSUE #4517: Engine SIGSEGV: InMemoryEdgeTypePropertyIndex dereferences an Edge freed by the edges_ skiplist GC (two arrivals; #4459/#4475 pins present and insufficient)
> **Updated 2026-08-04** with full symbolized stacks for both crashing-thread
> classes. The original report had only frame #0 and speculated about the call
> path; that speculation was wrong (see "History" at the bottom). This body now
> describes what the stacks actually show.

## Summary

`InMemoryEdgeTypePropertyIndex` dereferences a raw `Edge *` that the `edges_`
skiplist GC has already freed. Two different callers reach the same freed
object, which is why this presents as two crashing-thread classes:

| | `low prior` (query executor) | `Storage GC` |
|---|---|---|
| #0 | `HasExpectedProperty` | `HasExpectedProperty` |
| #1 | `PropertyStore::IsPropertyEqual` | `PropertyStore::IsPropertyEqual` |
| #2 | `CurrentEdgeVersionHasProperty(Edge const&, …)` | `AnyVersionHasProperty(Edge const&, …)` |
| #3 | `InMemoryEdgeTypePropertyIndex::Iterable::Iterator::AdvanceUntilValid()` | `InMemoryEdgeTypePropertyIndex::RemoveObsoleteEntries(…)` |
| reached from | `EdgesIterable::begin()` ← `ScanAllByEdgeTypePropertyValue` ← `PullPlan::Pull` ← Bolt | `Indices::RemoveObsoleteEdgeEntries` ← `InMemoryStorage::CollectGarbage` ← `Scheduler::ThreadRun` |

Across 10 cores captured in one day the frame #0 is identical every time — same
symbol, same instruction offset, differing only in PIE base — and every core is
the engine process (`execfn=/usr/lib/memgraph/memgraph`), SIGSEGV,
`oom_killed=false`, container exit 139 not 137. One defect, two arrivals.

**Build:** `master` @ `0e8aa326e` (2026-08-04), aarch64, Release with `.symtab`
retained, `IN_MEMORY_TRANSACTIONAL`, heavy edges (`storage_light_edge=false`,
`properties_on_edges=true`).

## Both arrivals are already pinned, and both still crash

This is the part that makes the existing mitigations insufficient:

- **GC side** — `InMemoryEdgeTypePropertyIndex::RemoveObsoleteEntries` takes
  `MakeEdgePin()` before its sweep loop, commented *"Pin the edge store while
  sweeping: the loop dereferences raw Edge\* the epoch GC could free."* It
  crashes inside that pinned loop. (This is #4475, relocated into the index
  method.)
- **Read side** — `Iterable` holds `pin_accessor_edge_` from the same
  `MakeEdgePin()` (#4459). It crashes inside that pinned iteration.
- In heavy-edge mode `MakeEdgePin()` is `edges_.access()`, a real
  `SkipListDb<Edge>::ConstAccessor` that calls `gc_.AllocateId()` in its
  constructor and `ReleaseId()` in its destructor — so neither pin is a no-op.

The locking is not the problem either: both `AnyVersionHasProperty`
(`indices_utils.hpp:170`) and `CurrentEdgeVersionHasProperty` (`:227`) read
`edge.properties` under `std::shared_lock{edge.lock}`, and the writer
(`EdgeAccessor::SetProperty`) takes `std::unique_lock{edge_.ptr->lock}` before
`properties.SetProperty(...)`. Reader and writer are properly excluded. The
memory is freed, not raced.

## Why a pin does not help here (hypothesis, from the GC contract)

I want to be explicit that this part is inference from reading
`utils/skip_list.hpp`, not something I have instrumented. The class doc for
`SkipListGc` states the contract:

> Each accessor is given a monotonically increasing ID. When a node is collected
> (**after the skip list has already unlinked it so no new accessor can reach
> it**) the ID of the newest currently-alive accessor is recorded. The node can
> be freed once that accessor has been destroyed.

`Collect()` tags the node with `accessor_id_.load() - 1`; `Run()` frees every
pending node whose tag is `< live_horizon`, the smallest still-alive accessor
id. So a pin protects a node **only if the pin's id was allocated before the
node was collected**.

The parenthetical is the load-bearing assumption — *"so no new accessor can
reach it"* — and it holds for anything that reaches nodes by traversing
`edges_`. It does not hold for `InMemoryEdgeTypePropertyIndex`, whose entries
hold a raw `Edge *` and reach the node without traversal. An index entry can
therefore outlive the unlink+collect of its edge, and a pin acquired *at
iteration or sweep start* is too late: its id is newer than the tag, so
`live_horizon` can advance past the tag and `Run()` frees the node while the
stale entry still points at it.

That would explain every observation: both arrivals crash, both pins are
genuine yet useless, the locking is correct, `data_` is garbage rather than
null, and the rate tracks edge churn against edge-index scans.

## Frequency and reproduction

Not rare for us. On a 16-shard aarch64 CI fleet running an extraction test
suite: **11 engine cores in ~2.5 hours** on 2026-08-04 (9 `low prior`, 2
`Storage GC`), and one run produced both classes from a single attempt. Longer
baseline: 89 engine deaths over 7.5 days, 52 of which our harness auto-retried
into green jobs, so the visible failure rate badly understates it.

Reproduction is probabilistic — concurrent edge property writes against
edge-type-property index scans — not a minimal case. **I am happy to run an
ASAN/TSAN build of any branch against this workload**; given it fires several
times an hour here, that is likely a faster route to confirmation than shrinking
a repro from your end. Just point me at a branch or an image tag.

## History / corrections

- Originally filed with only frame #0 (`HasExpectedProperty`), recovered by
  byte-matching a stripped binary against an unstripped one, plus a guess that
  it came in via vector search. **That guess was wrong** — I opened #4518 on
  it; it does not touch this path and is not a fix for this issue. Treat it as
  independent hardening or close it.
- Our CI now runs a symbols-bearing build of the exact binary, so the stacks
  above are direct reads, not reconstructions.

strichter:
Follow-on to #4475, same class of defect: an index hands out a raw pointer whose lifetime is not covered by anything the reader holds.

## The gap

`VectorIndexSearchOnNodes` / `VectorIndexSearchOnEdges` pin the object store for the duration of the **search call**, then return raw `Vertex*` / `Edge*` wrapped in accessors:

```cpp
// we have to take vertices accessor to be sure no vertex is deleted while we are searching
auto acc = mem_storage->vertices_.access();
const auto search_results = ...SearchNodes(...);
// ... wrap raw Vertex* into VertexAccessors ...
return result;   // ← pin dies here; the caller reads through the pointers afterwards
```

The comment states the intent, but the scoping does not deliver it: it is the *results*, not the lookup, that outlive the pin. Every other index read path in the tree gets this right — `InMemoryEdgeTypePropertyIndex::Iterable` and friends keep `EdgePin pin_accessor_edge_` and a `SkipListDb<Vertex>::ConstAccessor` as **members**, so the pin lives exactly as long as anything the caller can iterate. Vector search is the outlier.

This moves the pins onto the storage accessor, taken lazily on the first search of each kind. An object a search returned then stays reachable for as long as the accessors wrapping it can be used, and a transaction that never searches holds nothing.

## Two further problems in the edge search

- It took `edges_.access()` directly rather than `MakeEdgePin()`. Under `storage_light_edge` that pins **nothing relevant**: light edges are not skiplist nodes, and their memory is released by the graveyard drain against the epoch `MakeEdgePin()` acquires. The header says as much where `MakeEdgePin` is defined.
- It never pinned `vertices_` at all, although the `EdgeAccessor` it constructs carries both endpoint `Vertex*` next to the edge.

## Scope — what this is not

**I am not claiming this fixes #4517.** That crash is an engine SIGSEGV in a query-executor thread reading a freed `PropertyStore`, and I have only frame #0 — the release build will not unwind further, so the call path is unknown. This change is in the right family and touches a read path that hands out raw pointers, but I have not demonstrated that it is *that* path, and I would rather send you a correctly-scoped hardening change than a guess wearing a "Fixes" tag.

Specifically, it does **not** rescue the case where the index still holds an entry for a vertex freed before the reading transaction began; no pin taken at search time can help there. If that turns out to be #4517, the fix belongs in index cleanup, not here.

We are switching our CI to an unstripped build so the next occurrence yields a full symbolized stack, and I will attach it to #4517.

## Testing

Added `VectorIndexTest.SearchResultsSurviveGcWhileAccessorIsAliveTest`, which searches, runs `FreeMemory()`, and then reads a property back through the returned accessor — the pattern the scoping above left unprotected.

In fairness: I have not been able to build memgraph locally (no toolchain), so this has not been compiled on my side and the test has not been executed — your CI will be the first to do both. Happy to iterate on anything it turns up.

### DIFF
--- a/src/storage/v2/inmemory/storage.cpp
+++ b/src/storage/v2/inmemory/storage.cpp
@@ -5177,8 +5177,13 @@ std::vector<std::tuple<VertexAccessor, double, double>> InMemoryStorage::InMemor
   auto *mem_storage = static_cast<InMemoryStorage *>(storage_);
   std::vector<std::tuple<VertexAccessor, double, double>> result;
 
-  // we have to take vertices accessor to be sure no vertex is deleted while we are searching
-  auto acc = mem_storage->vertices_.access();
+  // The pin has to outlive the search, not just cover it. SearchNodes hands
+  // back raw Vertex*, we wrap them in VertexAccessors, and the caller reads
+  // through those long after this function has returned — while a pin scoped to
+  // this call is already gone and vertices_ is free to reclaim a node the
+  // moment its last accessor drops. Hold it on the storage accessor instead, so
+  // it covers the whole window in which the results are usable.
+  if (!vector_search_vertex_pin_) vector_search_vertex_pin_.emplace(mem_storage->MakeVertexPin());
   const auto search_results = storage_->indices_.vector_index_.SearchNodes(
       index_name, number_of_results, vector, mem_storage->name_id_mapper_.get());
   std::transform(search_results.begin(), search_results.end(), std::back_inserter(result), [&](const auto &item) {
@@ -5194,8 +5199,15 @@ std::vector<std::tuple<EdgeAccessor, double, double>> InMemoryStorage::InMemoryA
   auto *mem_storage = static_cast<InMemoryStorage *>(storage_);
   std::vector<std::tuple<EdgeAccessor, double, double>> result;
 
-  // we have to take edges accessor to be sure no edge is deleted while we are searching
-  auto acc = mem_storage->edges_.access();
+  // Same lifetime rule as the node search above. Both pins are needed: the
+  // EdgeAccessor holds the Edge and its two endpoint Vertex*, so pinning only
+  // edges_ would still let the endpoints be reclaimed underneath it. Take the
+  // edge pin via MakeEdgePin rather than edges_.access(): under
+  // storage_light_edge the edges are not in edges_ at all, and an accessor on
+  // it pins nothing — the graveyard drain frees light edges on the epoch the
+  // pin acquires.
+  if (!vector_search_edge_pin_) vector_search_edge_pin_.emplace(mem_storage->MakeEdgePin());
+  if (!vector_search_vertex_pin_) vector_search_vertex_pin_.emplace(mem_storage->MakeVertexPin());
   const auto search_results = storage_->indices_.vector_edge_index_.SearchEdges(index_name, number_of_results, vector);
   std::transform(search_results.begin(), search_results.end(), std::back_inserter(result), [&](const auto &item) {
     const auto &[entry, distance, score] = item;
--- a/src/storage/v2/inmemory/storage.hpp
+++ b/src/storage/v2/inmemory/storage.hpp
@@ -731,6 +731,15 @@ class InMemoryStorage final : public Storage {
                              IndexPerformanceTracker &impact_tracker);
     SalientConfig::Items config_;
 
+    // Vector search is the one read path that hands raw Vertex*/Edge* out of
+    // storage and lets the caller dereference them after the call returns, so
+    // its pins live here rather than in the search's own scope. Taken lazily on
+    // the first search of each kind and released with the accessor: an object a
+    // search returned stays reachable for as long as the accessors wrapping it
+    // can be used. Transactions that never search hold nothing.
+    std::optional<utils::SkipListDb<Vertex>::ConstAccessor> vector_search_vertex_pin_;
+    std::optional<EdgePin> vector_search_edge_pin_;
+
     // Bookkeeping
     durability::WalTxnDataPos wal_txn_positions_;
     bool needs_wal_update_{false};
--- a/tests/unit/vector_index.cpp
+++ b/tests/unit/vector_index.cpp
@@ -165,6 +165,32 @@ TEST_F(VectorIndexTest, DeleteVertexTest) {
   }
 }
 
+TEST_F(VectorIndexTest, SearchResultsSurviveGcWhileAccessorIsAliveTest) {
+  // The accessors a vector search returns wrap raw Vertex*, and the caller reads
+  // through them after the search call has returned. Pinning only for the
+  // duration of the search leaves that read unprotected, so the pin is held on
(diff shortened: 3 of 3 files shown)


strichter:
Housekeeping so this PR is judged on its own merits rather than as a candidate for #4517.

The body already said I was not claiming this fixes #4517. That is now settled rather than hedged: an ASAN build of the same commit caught #4517 and named it precisely — `InMemoryEdgeTypePropertyIndex::RemoveObsoleteEntries` dereferencing an `Edge *` freed by `SkipListGc<Edge>::Run()` driven inline from a concurrent `InMemoryAccessor::CreateEdge` on a query thread. Report and a candidate fix are in [#4517](https://github.com/memgraph/memgraph/issues/4517#issuecomment-5186792308). Different index, different path; nothing in that investigation touches vector search.

So this PR is **not** blocking on #4517 and should not be held for it. It stands or falls on the lifetime argument in the description alone: the pins are scoped to the search call while the raw `Vertex *` / `Edge *` in the returned accessors outlive it, `edges_.access()` rather than `MakeEdgePin()` pins nothing relevant under `storage_light_edge`, and `vertices_` is not pinned at all despite the `EdgeAccessor` carrying both endpoints.

Happy to close it if you would rather not carry the change — no argument from me either way.

PR #4524, opened 2026-08-05
feat: Atomic auth queries

colinbarry:
Auth queries could not run inside `BEGIN..COMMIT`. Each one took the system lock, wrote through, and replicated on
its own, so a sequence of them had no isolation and no rollback: a crash or a mistake halfway through left the
instance in a state that was neither the before nor the after.

This makes them transactional. Writes buffer in an overlay over the auth store, so other sessions see nothing until
COMMIT and a ROLLBACK simply discards them. The replication actions collect alongside and drain into a single system
transaction at COMMIT, which means the system lock is held for the flush rather than for as long as the user leaves
the transaction open, and a batch of auth changes reaches a replica whole or not at all. Two transactions that touch
the same users conflict, and the loser is told rather than silently overwriting.

Single auth statements outside a transaction behave exactly as before.

User profiles stay out of scope and are rejected inside an auth transaction. `UserProfiles` answers reads from an
in-memory cache rather than the store, so the overlay cannot isolate them, roll them back, or detect a conflict;
making that work is a separate piece of work.


colinbarry:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist

PR #4528, opened 2026-08-06
fix: Durability perf and logic improvements

as51340:
Measured improvement is around 35% when loading a single large WAL file/multiple WALs (https://docs.google.com/spreadsheets/d/1UD2OS90u6mNtnJNBjwUOdLBq5RHd07xUKT2Qz134GsU/edit?gid=0#gid=0).

Nothing in a WAL file stated its own timestamps or how much it held, so the only way to find out was `ReadWalInfo`, which walks every delta and CRCs every byte. Both *discovering* which files to recover from and *replaying* them therefore paid for a full parse — discovery entirely needlessly, and replay twice over.

`FinalizeWal` now back-patches `from_timestamp`, `to_timestamp` and the delta count into the header, the same way the constructor already back-patches the offsets. `ReadWalHeader` stops after the metadata section.

Design notes, including the alternatives that were tried and rejected, are in `specs/wal-header-summary.md`.

## What got faster

| Path | Runs | Before | After |
|---|---|---|---|
| `GetWalFiles` → `RecoverData`, `GetRecoverySteps` | every recovery, every replica recovery | full parse per file | one page per file |
| `EnsureNecessaryWalFilesExist` | after every snapshot, default every 300 s | full parse per file | one page per file |
| `durability::LoadWal` | per file replayed | parse twice | parse once |
| `InMemoryReplicationHandlers::LoadWal` | per file received | parse twice | parse once |

`RecoverData`'s snapshot-less pre-pass is also gone. It parsed the whole directory only to learn the newest file's uuid and epoch id, and `GetWalFiles` then parsed it again. With no uuid known yet, that branch now reads every header, adopts the newest file's identity, and filters the rest in memory. "Newest" is by path rather than `seq_num`, because sequence numbers restart from 0 whenever the storage uuid changes, so `back()` on the seq-num-sorted vector would pick a stale file in a mixed-uuid directory.

Snapshot-less recovery went from four full parses per WAL file to one; recovery from a snapshot went from three to one.

## Format

```
SECTION_METADATA marker, uuid, epoch_id, seq_num
metadata CRC trailer        <- written once by the constructor, never again
from_timestamp, to_timestamp, num_deltas
summary CRC trailer
offset_deltas -> first delta
```

The summary rides on v37, which is unreleased, so no WAL in the wild claims that version without one and replicas need not negotiate a new format. A version guard is still required: pre-v37 files have no bytes there.

A zero delta count marks the placeholder the constructor reserved — `FinalizeWal` only ever writes a positive count. `from_timestamp` could not serve as the marker, since `kTimestampInitialId` is 0 and a commit timestamp of zero is legitimate.

### Finalizing cannot damage the file's identity

The summary sits *outside* the metadata's CRC-protected region, with a CRC of its own. That matters because the first revision put it inside and had `WriteSummary` rewrite the whole section to recompute the CRC, which opened a window where a crash could cost an entire WAL file: every `GetPosition()` and `SetPosition()` reaches `OutputFile::SetPosition`, which flushes, so the new values reached the page cache several writes before the new CRC. A crash in between left a header carrying new values with a stale CRC, `ReadWalHeader` threw, and `GetWalFiles` dropped the file — up to `--storage-wal-file-size-kib` of committed transactions. An ordinary process crash sufficed, the earlier write already being in the page cache, and rotation runs routinely under write load. Before this PR `FinalizeWal` never touched the header at all, so a crash during rotation cost nothing.

Separating the regions scopes the damage. A torn summary write cannot invalidate uuid/epoch/seq or their CRC, and the summary's own CRC covers exactly the bytes `FinalizeWal` rewrites.

`WriteSummary` is seek-once and write-forward, since `WriteCrc()` emits the marker and value at the current position, so nothing flushes between the values and their CRC. It dropped five members it no longer needs: the uuid, epoch id, metadata offset, CRC offset and prefix CRC.

### A summary that fails its own CRC is fatal

The region has exactly two legitimate states, and **both verify**: the zeros the constructor reserves and syncs at creation, meaning the file was never finalized, and the values `FinalizeWal` fills in. It is 36 bytes inside a single sector, written in one go, so it cannot tear into a third state that leaves intact deltas behind. Anything that does not verify is media rot, and `DecodeWalHeader` throws.

The tolerant alternative — read a damaged summary as absent and re-derive the count by parsing — was tried and rejected, because it is indistinguishable from an unfinalized file and so hands a rotted *finalized* file straight to the truncate-on-damage path this PR set out to close. It also made those 36 bytes the one region of a v37 WAL where a single flipped bit changed nothing any reader reported: `ReadWalInfo` never consults the summary. `WalFileModification` now sweeps every byte of a finalized file and every one of them is accounted for.

The distinction stays sharp for `LoadWal`: `header.summary` is empty only for the placeholder state, i.e. genuinely unfinalized, which is the only case where a torn tail is legitimate.

## Replay uses the summary, and damage in a finalized file is fatal

A finalized WAL file is fsynced before it is renamed, so its transactions are durable and were acknowledged. If replay comes up short of the count its header states, the bytes rotted — a write was not interrupted. A file with no summary was never finalized and its tail may legitimately be torn, so that one is still parsed by the dry run, which stops at the last whole transaction.

| File | Extent from | Damage means |
|---|---|---|
| finalized (has summary) | the header | media rot → fail |
| unfinalized, or pre-v37 | `ReadWalInfo` dry run | interrupted write → truncate |

That split is what makes the strict policy safe. A crash leaves the file *unfinalized*, so crash recovery still degrades gracefully. Replay reads a finalized file once instead of twice and verifies each transaction's CRC as it applies it.

### This fixes a pre-existing bug

Recovering a prefix is only sound for the **last** file in the chain. The gap check catches a *missing* WAL file, never a *truncated* one:

```cpp
if (previous_seq_num && (wal_file.seq_num - *previous_seq_num) > 1) throw ...;  // missing file only
```

So a truncated file N left `last_loaded_timestamp` at the last transaction it applied, and file N+1 then applied in full on top of the hole. If N+1 touched anything from N's lost tail the apply threw, but where its transactions were independent, recovery **succeeded silently with an incomplete dataset**.

On a replica it was worse: the replica reported the later ldt, so main compared timestamps, considered it caught up, and the divergence never healed.

Failing on damage in a finalized file removes the case. A mid-chain file cannot be unfinalized in a way that matters, because whichever session recovered its prefix already continued from there, so the later files were built on that state rather than on the lost tail.

### The replica stops the chain rather than the process

There is no process to abort and no operator watching, so `InMemoryReplicationHandlers::LoadWal` catches the failure, reports it, and the caller stops before the remaining files — those build on the transactions that just went missing. The in-flight transaction's accessor is destroyed while unwinding, which aborts it, so nothing half-applied is ever committed.

The response carries no commit timestamp, so main does not advance its view of the replica, retries later, and the replica skips whatever it already has via `to_timestamp <= ldt`. Committed progress is kept; nothing partial is.

### Operational consequence

Bit rot in a WAL file *after* the newest snapshot now stops the tenant instead of bringing it up stale. Recovery is `--storage-allow-recovery-failure` plus `RECOVER SNAPSHOT`, or a restore. Files the newest snapshot already covers are skipped without being read, so rot in those stays harmless.

## Tests

New:

- `WalMixedUUID` — WAL-only recovery with two storages' files present must adopt the newest by path and ignore the rest. Nothing covered this; `SnapshotAndWalMixedUUID` always has a snapshot and so takes the other branch.
- `FinalizedHeaderSummaryMatchesScan` — the back-patched summary must equal what parsing derives. This is the invariant everything else rests on.
- `UnfinalizedFileHasNoSummary` — pins the placeholder state: zeros verify, and the reader treats them as no summary rather than as damage.
- `DamagedSummaryFailsRecovery` — flips the last byte of the summary's CRC on a finalized file; both `ReadWalHeader` and `ReadWalInfo` must refuse.

Changed:

- `WalCorruptLastTransaction` → `WalFinalizedFileCorruptLastTransactionCrashes`. It damages a *finalized* file and used to expect the prefix to recover; it now expects the refusal. This is the one test whose contract changed, and the thing worth reviewing hardest.
- `WalFileModification` now finalizes its file, so its per-byte sweep runs against a real summary rather than the zeroed placeholders, and every byte of the file is covered by some CRC.

Unchanged but relevant: clean shutdown finalizes, so `WalBasic` and friends exercise the summary path, and `WalDeathResilience` SIGKILLs — which is what covers graceful truncation of a crash-torn tail.


as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Changed WAL's file header format so it now also contains cached range of timestamps and the number of deltas. This should significantly improve recovery speed on restarts and on replica recovery. Fixed a bug where a WAL file could've been partially loaded although it wasn't the last one in the chain causing a possible data loss. [#4528](https://github.com/memgraph/memgraph/pull/4528)
- [x] **[ Documentation PR link ](https://github.com/memgraph/documentation/pull/1723)**

src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -1099,32 +1099,46 @@ InMemoryReplicationHandlers::LoadWalStatus InMemoryReplicationHandlers::LoadWal(
     slk::Builder *res_builder, uint32_t const start_batch_counter) {
   spdlog::trace("Received WAL saved to {}", wal_path);
 
-  std::optional<storage::durability::WalInfo> maybe_wal_info;
+  // A finalized file states how much it holds, so it is replayed straight from its header. Only main's current WAL,
+  // which it is still writing, has no summary; that one is parsed so a transaction main had not finished is not
+  // replayed. Either way each transaction's CRC is verified below as it is applied.
+  std::optional<storage::durability::WalHeader> maybe_wal_header;
```
andrejtonev: why do we need this to be optional? we either emplace or throw and return

src/storage/v2/durability/durability.cpp
```diff
@@ -173,39 +173,53 @@ std::optional<std::vector<WalDurabilityInfo>> GetWalFiles(const std::filesystem:
...
   std::error_code error_code;
 
-  // TODO: (andi) Inefficient to use I/O again, you already read infos.
   for (const auto &item : std::filesystem::directory_iterator(wal_directory, error_code)) {
     if (!ValidateDurabilityFile(item)) continue;
 
     try {
-      auto info = ReadWalInfo(item.path());
+      auto header = ReadWalHeader(item.path());
+      if ((!uuid.empty() && header.uuid != uuid) || (current_seq_num && header.seq_num >= *current_seq_num)) {
```
andrejtonev, line 182: is the >= seq number check correct? shouldn't we want a seq number to be equal or +1 to the current one?

src/storage/v2/durability/durability.cpp
```diff
@@ -173,39 +173,53 @@ std::optional<std::vector<WalDurabilityInfo>> GetWalFiles(const std::filesystem:
...
+                      item.path(),
+                      uuid,
+                      header.uuid,
+                      current_seq_num,
+                      header.seq_num);
+        continue;
+      }
+
+      // A file that finished with something in it states its own timestamps. Anything else - one the writer never
+      // finalized, one predating k37, or one holding no complete transaction - has its deltas parsed to derive
```
andrejtonev: k37?

src/storage/v2/durability/durability.cpp
```diff
@@ -173,39 +173,53 @@ std::optional<std::vector<WalDurabilityInfo>> GetWalFiles(const std::filesystem:
...
+      // them, and ReadWalInfo throwing for the last of those is how it gets dropped here.
+      uint64_t from_timestamp{0};
+      uint64_t to_timestamp{0};
+      if (header.summary) {
+        from_timestamp = header.summary->from_timestamp;
+        to_timestamp = header.summary->to_timestamp;
+      } else {
+        auto const info = ReadWalInfo(item.path());
```
andrejtonev: can this ReadWalInfo fail?

src/storage/v2/durability/snapshot.cpp
```diff
@@ -13820,12 +13820,20 @@ void EnsureNecessaryWalFilesExist(const std::filesystem::path &wal_directory, co
   for (const auto &item : std::filesystem::directory_iterator(wal_directory, error_code)) {
     if (!item.is_regular_file()) continue;
     try {
-      auto info = ReadWalInfo(item.path());
-      if (info.uuid != uuid) {
+      // Only from_timestamp decides what gets deleted below, and a finalized file states it in its header, so the
+      // deltas are parsed only for a file that doesn't - one the writer never finalized, or one predating k37.
+      auto header = ReadWalHeader(item.path());
```
andrejtonev, line 13825: this can fail. should we move the back compatibility logic (header or walinfo) into this function so we don't keep dragging the if/else logic everywhere?

src/storage/v2/durability/version.hpp
```diff
@@ -46,14 +46,16 @@ constexpr uint64_t kExtendedEdgeSetProperty{33};
 constexpr uint64_t kDescriptionAndDescIndexSupport{34};
 constexpr uint64_t kVectorIndexMultiLabel{35};
 constexpr uint64_t kCrcProtection{36};
-constexpr uint64_t kVertexPropertyIndex{37};
+// Carries two changes: the vertex property index, and the transaction summary the WAL header gained,
+// back-patched when the file is finalized.
+constexpr uint64_t k37{37};
```
andrejtonev: Why this name k37 and why not bump the version to 38?

as51340: because version was already bumped for this release and was named kVertexPropertyIndex. Since this version will contain 2 changes, I decided to rename it.

src/dbms/inmemory/replication_handlers.cpp
as51340: you're right

src/storage/v2/durability/durability.cpp
as51340: it's correct because this is not checking the chain continuity, rather seq num is only passed from recovery.cpp as the current main seq num so we don't want WAL files with >= seq nums. This is mostly for robustness

src/storage/v2/durability/durability.cpp
as51340: yes and then we catch the exception at line 225

src/storage/v2/durability/snapshot.cpp
as51340: Merged together into one function

src/storage/v2/durability/version.hpp
andrejtonev: there are already places we have multiple changes within the same version and we just made a new variable with the same number. example:
constexpr uint64_t kSridCartesian3DCorrected{21};
constexpr uint64_t kEdgeSetDeltaWithVertexInfo{21};

as51340: ok

src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -1155,15 +1156,29 @@ InMemoryReplicationHandlers::LoadWalStatus InMemoryReplicationHandlers::LoadWal(
...
+      // commit_txn_immediately is set true because when loading WAL files, we should commit immediately
+      deltas_res = ReadAndApplyDeltasSingleTxn(storage,
+                                               &wal_decoder,
+                                               *version,
+                                               res_builder,
+                                               /*two_phase_commit*/ false,
+                                               /*loading_wal*/ true,
+                                               deltas_batch_progress_size,
+                                               local_batch_counter);
+    } catch (const utils::BasicException &e) {
```
andrejtonev, line 1175: should we limit this to replication related exception only?

src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -1155,15 +1156,29 @@ InMemoryReplicationHandlers::LoadWalStatus InMemoryReplicationHandlers::LoadWal(
...
-                                                        deltas_batch_progress_size,
-                                                        local_batch_counter);
+    // A delta that won't parse, or a transaction whose CRC doesn't match, throws out of here. The in-flight
+    // transaction's accessor is destroyed while unwinding, which aborts it, so nothing half-applied is ever
+    // committed. Report failure so the caller stops before the WAL files that follow this one: those build on the
+    // transactions that just went missing, and applying them would leave a wrong dataset rather than a stale one.
+    std::optional<storage::SingleTxnDeltasProcessingResult> deltas_res;
+    try {
+      // commit_txn_immediately is set true because when loading WAL files, we should commit immediately
+      deltas_res = ReadAndApplyDeltasSingleTxn(storage,
```
andrejtonev, line 1167: does this assert the last delta as tx end (like you added in wal.cpp)?

src/storage/v2/durability/wal.cpp
```diff
@@ -1031,6 +1065,39 @@ WalInfo ReadWalInfo(const std::filesystem::path &path) {
...
+
+// Function used to read information about the WAL file.
+WalInfo ReadWalInfo(const std::filesystem::path &path) {
+  Decoder wal;
+  uint64_t version{};
+  auto header = DecodeWalHeader(wal, path, version);
+  return ScanWalDeltas(wal, version, std::move(header));
+}
+
+WalInfo ReadWalContents(const std::filesystem::path &path) {
```
andrejtonev: pass an optional header (in case the user already read it) and call DecodeWalHeader only if nullopt

andrejtonev (approved):

src/dbms/inmemory/replication_handlers.cpp
as51340: we're throwing BasicException all around ReadAndApplyDeltasSingleTxn so I think this needs thinking a bit more about this

src/dbms/inmemory/replication_handlers.cpp
as51340: not in the same way but it does: if there is a delta from the new txn while the previous one wasn't finished, we throw. I add a hardening to explicitly check that we processed all deltas from the transaction

src/storage/v2/durability/wal.cpp
as51340: ok

PR #4530, opened 2026-08-06
feat: complete the path module config — node filters, limit, end/terminator nodes, uniqueness, and abortable traversals

imilinovic:
Extends the `path` module's config map to the full set of keys it advertises, adds a config-map form of
`path.expand`, makes the traversals abortable, and fixes a set of pre-existing bugs — two of which take
the server down.

Addresses three reported gaps: blacklisting nodes by identity on `path.subgraph_all`, per-type direction
markers in a relationship filter, and the full config-map expand call.

## New

- `path.expand_config(start, config)`. The compatibility name mapping already pointed the config-map name
  at `path.expand`, whose positional signature rejects a map, so that entry was unreachable.
- `allowlistNodes` / `denylistNodes` (deprecated aliases `whitelistNodes` / `blacklistNodes`), and
  `endNodes` / `terminatorNodes` — the identity counterparts of the `>` and `/` label prefixes. Both are
  allowlisted implicitly, so an allowlist that does not name them cannot starve the filter it is paired
  with.
- `limit`, which caps the rows **and stops the walk**. Under breadth-first expansion that returns the
  shortest paths without paying for the rest of the graph; a trailing Cypher `LIMIT` cannot, since a
  non-batched procedure materialises every row before the first is read.
- `uniqueness`: `RELATIONSHIP_PATH` (default), `NODE_PATH`, `NODE_GLOBAL`. The last returns one path per
  reachable node and needed its own walk — the breadth-first driver reaches a depth by re-walking from the
  start, so a mark that survived would block the next pass. It walks a queue over the tree the traversal
  builds and rebuilds each path from it. Every start is seeded before any is expanded, so a start reached
  from another is returned once, as its own root; `bfs: false` marks them all up front for the same reason.
  Being iterative, it also makes an absent `maxHops` safe — `bfs: false` still recurses, so the depth
  ceiling applies there.
- Repeating **sequences** of labels and relationship types: a comma separates the steps, and the depth a
  node or relationship sits at chooses its step. `sequence` spells both kinds in one alternating string and
  supersedes the two filter keys. On `path.expand_config`, `path.subgraph_nodes` and `path.subgraph_all`;
  the positional form takes lists of alternatives, which cannot spell one. A comma-free filter is the
  single-step case, so what it matched before, it matches still.
- `beginSequenceAtStart` acts rather than being ignored: with it false the first relationship step is spent
  on the hop out of the start node and the rest repeat from there, and the start node has no step to be
  tested against, so the label filter does not apply to it however `filterStartNode` is set.
- `*` matches every label under any of the four prefixes — how a sequence says "any label here". It reaches
  the positional `path.expand` too; see **Breaking**. It is deliberately not a relationship wildcard, where
  a bare marker already means "any type".
- `minLevel` / `maxLevel` aliases; a `|`-separated string as well as a list for both filters; `bfs` read
  rather than ignored (`path.expand_config` defaults to breadth-first, the positional form stays
  depth-first since its signature cannot express it); a null `start` yields no rows.

## Breaking

`<TYPE>` no longer means "a reciprocal pair of relationships exists between the same two nodes" — it now
means an incoming `TYPE`, and the reciprocal filter is **removed with no replacement**. Every entry
containing `<` is now an incoming filter. The reference reads it that way and has no reciprocal mode, so
the old reading turned a filter written against it into one matching almost nothing; one spelling cannot
carry both meanings. The mode also held per-node state to pair the two relationships up and depended on
in-relationships being iterated first. If the capability is wanted back it can return as a local predicate
— when crossing X→Y of type T, ask whether Y has a T back to X — under a spelling that does not collide.

`*` becoming a wildcard changes results for any label filter containing one: `'-*'` was inert and now
denies every node; `'>*'` and `'/*'` now put **every** step into end-nodes-only mode, which decides which
nodes come back rather than only which the walk passes through. This reaches the positional `path.expand`,
which the sequence work does not otherwise touch, through its `labels` argument.

Otherwise nothing that behaved correctly before behaves differently now. The only remaining way an
existing call stops working is passing a key the reader now names instead of dropping.

## Fixes a crash or a hang

- The walks never polled the abort signal, so once started they ran to completion regardless: the query
  timeout did not fire, `TERMINATE TRANSACTIONS` reported success and dropped the transaction from
  `SHOW TRANSACTIONS` while the walk kept a core pinned, and graceful shutdown needed `SIGKILL`. Nothing
  else could stop them — the memory tracker only fires on allocation, and a walk with a `minHops` above
  the reachable depth traverses the whole graph without emitting a record.
- `path.create` dereferenced a missing `rel` key and segfaulted the server. The declared default argument
  has no `rel` key, so plain `path.create(n)` crashed it.
- The parsed label filter held `string_view`s into a list that dies with the statement, so it compared
  against freed memory for the whole traversal — correct under the small-string boundary, silently wrong
  above it, across all four filter categories.
- The expand walk recurses once per hop, so an absent or `-1` upper bound let a long chain exhaust the
  stack. It now refuses beyond depth 5000.
- `ParseLabels` read `front()` on an entry without checking it was non-empty.

## Changes results without raising

- A `/` terminator ended the walk however shallow the node was, but below the lower bound it cannot be
  returned yet, so the walk has to pass through it.
- Relationship entries are read by which markers they contain, wherever they appear, `<` winning: `'<R>'`,
  `'R<'` and `'<>R'` all name the incoming filter `'<R'` names, `'>R'` the outgoing one, `'<>'` every type
  incoming. Three of those matched nothing before.
- Two entries for the same type merge into any-direction instead of overwriting, so `['R>', '<R']` is no
  longer an order-dependent subset.
- `maxHops` was wrong in opposite directions in the two readers: `-1` meant "match nothing" positionally,
  and any other negative was dropped into an unbounded scan. `-1` is now no limit everywhere.
- `path.expand` no longer applies the label filter to an unfiltered start node — and, the other half
  of the same rule, no longer returns the zero-length start path when the filter carries a `>` or `/`
  prefix. Those prefixes decide which nodes are returned at all, and an exempt start node is not one
  of them however its own labels read; `filterStartNode: true` is what returns a start node carrying
  an end or termination label. `subgraph_nodes` already behaved this way, and so does the reference.
- Start nodes are walked in the order the caller listed them, which decides what a `limit` returns.
- A null start no longer makes the subgraph procedures skip config validation.

## Now rejected instead of silently wrong

An unrecognized key, a hop bound passed with its alias, a degenerate filter entry, a `limit` below `-1`, a
`uniqueness` value the walk cannot honour, a sequence step naming no filter (blank, trailing comma, or
all-empty alternatives), a `,` inside a list entry, `*` in a relationship filter, and
`beginSequenceAtStart: false` with a single-step relationship filter — each named in the error. The reader
used to keep the keys it knew and drop the rest, so the failure mode throughout was a silently different
result set. `labelFilter` and `relationshipFilter` are now type-checked even when `sequence` supersedes
them.

## Accepted and ignored, and not offered

`bfs` and `uniqueness` on the subgraph procedures, whose queue visits each node once — which is what the
reference does with them, overwriting whatever was passed with `NODE_GLOBAL`. Not offered:
`RELATIONSHIP_GLOBAL`, the `*_LEVEL` and `*_RECENT` modes, `NONE`, and `optional`. Which paths survive the
modes left out depends on the order relationships happen to be iterated in, which is not a rule a caller
can reason about.

## Deliberate divergences

Measured against the reference case by case, not assumed; everything else agrees up to row order.

- `beginSequenceAtStart: false` with a single-step relationship filter is rejected, naming the key. The
  reference divides by the length left after spending the first step and throws
  `ArithmeticException: / by zero`.
- A step naming no filter is rejected — blank (`'Post,,Reply'`), a trailing comma, or alternatives that are
  all empty (`'Post,|,Reply'`). The reference reads a blank step as matching everything and drops a
  trailing one.
- `*` matches every label. The reference builds an accepts-all matcher and then discards the return value
  (`LabelMatcherGroup.addLabel`), so there `'Person|*'` allowlists only `Person` and `-*` is inert — both
  contradicting its own docs. Under `>` and `/` it still flips its end-nodes-only flag with no matcher left
  to satisfy, so `'>*'` and `'/*'` return **nothing** there, where here they return every node and the
  first hop respectively.
- A `,` inside a *list* entry is rejected, pointing at the string form. A list entry is one alternative;
  reading commas there would change what the list form has always meant.
- A `sequence` whose label or relationship half ends up with no step is rejected, naming the key — for
  instance `sequence: 'Foo'`. The reference passes the empty half on and dies in the same `/ by zero`.
- Row order within a level differs throughout, since the reference iterates a node's relationships the
  other way. Pre-existing — it shows on `RELATIONSHIP_PATH`, which shipped earlier here. Under
  `NODE_GLOBAL` it also decides which of two equal-length paths a node gets, and which node a `limit` cuts
  at. Node sets, path lengths and the one-path-per-node property match.


imilinovic:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - **Added:** `path.expand_config(start, config)` — expansion configured by a map instead of a fixed argument list, so hop bounds, relationship and label filters (including repeating sequences of either), `uniqueness`, `limit`, `bfs` and node-identity filters can all be given. `path.subgraph_nodes` and `path.subgraph_all` accept the new filter keys too. [#4530](https://github.com/memgraph/memgraph/pull/4530)
    - **Breaking:** `<TYPE>` in a relationship filter now means an incoming `TYPE`, not "a reciprocal pair of relationships exists between the same two nodes"; the reciprocal-pair filter is **removed with no replacement**. A marker counts wherever it appears and `<` wins over `>`, so `'<R>'`, `'R<'` and `'<>R'` all mean incoming `R`. A call relying on the old reading returns **different results without raising an error**. On `path.expand`, `path.expand_config`, `path.subgraph_nodes` and `path.subgraph_all`. [#4530](https://github.com/memgraph/memgraph/pull/4530)
    - **Breaking:** `*` in a label filter is now a wildcard rather than a literal label name, so a filter containing one changes results: `'-*'` was inert and now denies every node, and `'>*'` or `'/*'` put every step into end-nodes-only mode, which decides which nodes are returned rather than only which the walk passes through. On `path.expand`, `path.expand_config`, `path.subgraph_nodes` and `path.subgraph_all` — the positional form is affected through its `labels` argument. [#4530](https://github.com/memgraph/memgraph/pull/4530)
    - **Breaking:** a config key the procedure cannot honour is now an error naming the key, where it was previously ignored, so a call passing `optional` or a misspelling such as `maxhops` stops working. On `path.expand_config`, `path.subgraph_nodes` and `path.subgraph_all`. [#4530](https://github.com/memgraph/memgraph/pull/4530)
    - **Changed:** a filter entry naming nothing usable is now rejected instead of producing a wrong answer — an empty label entry, a bare `+`/`-`/`>`/`/`, or a relationship entry with neither a type nor a direction, such as `':'`, which previously filtered out every relationship in the graph. A bare `'<'` or `'>'` stays valid and means every type in that direction. On `path.expand`, `path.expand_config`, `path.subgraph_nodes` and `path.subgraph_all`. [#4530](https://github.com/memgraph/memgraph/pull/4530)
    - **Changed:** `*` is not a wildcard in a relationship filter, where a bare `'>'`, `'<'` or `'<|>'` already means "any type"; an entry containing one is now rejected naming those, instead of being read as a type name that matches nothing. On `path.expand`, `path.expand_config`, `path.subgraph_nodes` and `path.subgraph_all`. [#4530](https://github.com/memgraph/memgraph/pull/4530)
    - **Changed:** a sequence step naming no filter is rejected, with the key and the step's 1-based position — blank (`'Post,,Reply'`), a trailing comma, or alternatives that are all empty (`'Post,|,Reply'`). A `,` inside a *list* entry is rejected too, since a list entry is one alternative; give the whole filter as a string to spell a sequence. On `path.expand_config`, `path.subgraph_nodes` and `path.subgraph_all`; the positional `path.expand` cannot spell a sequence and still reads a comma in a list entry as part of the label. [#4530](https://github.com/memgraph/memgraph/pull/4530)
    - **Changed:** `bfs` and `uniqueness` are accepted and ignored on `path.subgraph_nodes` and `path.subgraph_all`, which always visit each node once, breadth-first, so neither key can ask for anything they do not already do. `sequence` and `beginSequenceAtStart` do apply on both. [#4530](https://github.com/memgraph/memgraph/pull/4530)
    - **Fixed:** `beginSequenceAtStart` now shifts the sequence, rather than only exempting the start node from the label filter as it did before: with it false the first relationship step is spent on the hop out of the start node and the rest repeat from there. Passing it alongside a relationship filter of a single step is now an error, since that step is consumed by the first hop. On `path.expand_config`, `path.subgraph_nodes` and `path.subgraph_all`. [#4530](https://github.com/memgraph/memgraph/pull/4530)
    - **Fixed:** two relationship filter entries for the same type merge into an any-direction filter, so `['R>', '<R']` no longer returns an order-dependent subset. On `path.expand`, `path.expand_config`, `path.subgraph_nodes` and `path.subgraph_all`. [#4530](https://github.com/memgraph/memgraph/pull/4530)
    - **Fixed:** a `/` termination label no longer ends the walk below the lower hop bound, where the node cannot be returned yet; combining `/` with a `minHops` above 1 previously lost every path beyond the first match. On `path.expand`, `path.expand_config`, `path.subgraph_nodes` and `path.subgraph_all`. [#4530](https://github.com/memgraph/memgraph/pull/4530)
    - **Fixed:** a max hop bound of `-1` means no limit on `path.expand`, which previously returned nothing for it. On `path.expand_config`, `path.subgraph_nodes` and `path.subgraph_all`, a bound below `-1` now matches nothing instead of being dropped into an unbounded scan, and a `limit` below `-1` is rejected. [#4530](https://github.com/memgraph/memgraph/pull/4530)
    - **Fixed:** `path.expand` no longer applies the label filter to its start node, which it has no argument to ask for. [#4530](https://github.com/memgraph/memgraph/pull/4530)
    - **Fixed:** `path.subgraph_nodes` and `path.subgraph_all` validate the config before reading the start node, so a null start no longer hides the error for a bad key — the same call previously succeeded or failed depending on the data. [#4530](https://github.com/memgraph/memgraph/pull/4530)
    - **Fixed:** the traversals observe the query timeout, `TERMINATE TRANSACTIONS` and graceful shutdown. One previously ran to completion regardless: the timeout did not fire, `TERMINATE TRANSACTIONS` reported success while the walk kept a core pinned, and shutdown needed `SIGKILL`. On `path.expand`, `path.expand_config`, `path.subgraph_nodes` and `path.subgraph_all`. [#4530](https://github.com/memgraph/memgraph/pull/4530)
    - **Fixed:** `path.create` no longer aborts the server process when the relationship map has no `rel` key — including plain `path.create(n)`, whose own declared default has none. [#4530](https://github.com/memgraph/memgraph/pull/4530)
    - **Fixed:** the label filters no longer read freed memory, which made a label filter silently wrong for labels longer than 15 characters. On `path.expand_config`, `path.subgraph_nodes` and `path.subgraph_all`. [#4530](https://github.com/memgraph/memgraph/pull/4530)
    - **Fixed:** a deep expansion raises a query error past a depth of 5000 instead of exhausting the stack and taking the instance down. `maxHops` is still unbounded by default, so set `--memory-limit` to keep an unbounded traversal a recoverable query error. On `path.expand` and `path.expand_config`; `uniqueness: NODE_GLOBAL` with `bfs: true` is exempt, being iterative, and the subgraph procedures never recursed. [#4530](https://github.com/memgraph/memgraph/pull/4530)
    - **Fixed:** a type error for `minLevel` or `maxLevel` names the key the caller wrote, not the `minHops`/`maxHops` spelling it aliases. On `path.expand_config`, `path.subgraph_nodes` and `path.subgraph_all`. [#4530](https://github.com/memgraph/memgraph/pull/4530)
    - **Fixed:** a null `start` yields no rows instead of raising a type error, so a start node from an `OPTIONAL MATCH` needs no guard. On `path.expand`, `path.expand_config`, `path.subgraph_nodes`, `path.subgraph_all` and `path.create`; `path.subgraph_all` returns its single row with both lists empty. [#4530](https://github.com/memgraph/memgraph/pull/4530)
- [x] **[ Documentation PR link memgraph/documentation#1718 ]**

DavIvek (approved):

PR #4531, opened 2026-08-06
feat: add export module for JSON graph export

imilinovic:
Adds a new C++ query module `export` with three read procedures for exporting a graph as JSON, and re-points the JSON export aliases in `config/mappings.json` at them:

```
export.json_data(nodes, rels, file, config = {})
export.json_all(file, config = {})
export.json_graph(graph, file, config = {})
```

All three return the standard 12-column export row (`file, source, format, nodes, relationships, properties, time, rows, batchSize, batches, done, data`) and either write the payload to `file` or stream it back in `data`.

**How.** One shared serializer core plus a thin entry point per procedure, so the rest of the export namespace (csv, graphml, cypher) can reuse it:

| file | responsibility |
|---|---|
| `algorithm/json_writer.{hpp,cpp}` | value → JSON, ISO-8601 temporals, srid → crs, the three output shapes (`JSON_LINES`, `JSON`, `JSON_ID_AS_KEYS`), and the sink |
| `algorithm/export.{hpp,cpp}` | config parsing, result row |
| `export_module.cpp` | registration |

`nlohmann::ordered_json` is used so the emitted key order matches the interchange format's; this is cosmetic rather than semantic, since the acceptance bar is parsed-JSON equality. Registration goes through the low-level `mgp` path because `nodes`/`rels` must accept `NULL`, and the two nullable result columns (`file`, `data`) are inserted through the low-level record API since `mgp::Record::Insert` has no `Type::Null` case.

**Why.** `export_util.json` / `export_util.json_graph` emit an *internal round-trip format*: `import_util.json` parses it back and restores temporal types by dispatching on the `date(` / `localTime(` / `localDateTime(` / `duration(` prefixes it writes. Existing e2e coverage depends on that round trip, so switching those procedures to the interchange format would break it. The interchange format therefore lands as a sibling module; both remain callable under their own names.

### Breaking change to two aliases

Re-pointing the aliases changes the contract for anyone already calling through the `apoc.*` names — the alias mapper is a pure rename, so arguments and result columns are validated against the new target:

| alias | was | now |
|---|---|---|
| `apoc.export.json.all` | `export_util.json(path, config)` → `(path, data)` | `export.json_all(file, config)` → 12 columns, **no `path`** |
| `apoc.export.json.graph` | `export_util.json_graph(nodes, relationships, path, config)` | `export.json_graph(graph :: MAP, file, config)` |

So an existing `YIELD path` fails at plan time, and a `.graph` call passing two lists fails argument validation. The `.graph` case was already broken — the alias resolved to a procedure taking two lists where the real one takes a single graph map — so ported calls errored before this change too.

The `apoc.export.json.all` → `apoc.import.json` pairing also stops round-tripping: `apoc.import.json` still resolves to `import_util.json`, which expects a top-level array with integer `start`/`end` and a `properties` key on every element, while this module emits JSON-Lines by default, object endpoints, and omits `properties` when empty. `export_util.json` / `export_util.json_graph` stay callable under their native names and remain the migration path for both cases. Nothing in-tree depends on the old mapping.

### Output rules

Every rule was measured against the reference implementation rather than assumed. Element ids are strings, relationship endpoints are inlined in full even when absent from the exported node set, labels are sorted, `labels` and `properties` are each omitted entirely when empty, temporals elide zero components (with different sub-second rules for the local-time family and for durations), WGS-84 points use `latitude`/`longitude`/`height` keys, non-finite doubles carry their textual forms, and `writeRelationshipProperties` defaults to `writeNodeProperties`. An empty element set yields `""` under `JSON_LINES` but still emits the wrapper (`{"nodes":[],"rels":[]}` / `{"nodes":{},"rels":{}}`) under the object shapes, so `json.loads(data)["nodes"]` works on an empty result.

`jsonFormat` is matched case-insensitively, boolean config values accept `true`/`false`/`yes`/`no`/`1`/`0` as booleans, integers or strings, and `json_graph` accepts `edges` as an alias for `relationships` because that is the key `project()` produces.

Deliberate divergences from the reference: `compression` and `charset` are rejected rather than ignored, so a caller asking for gzip cannot silently receive plaintext; an unrecognised boolean value throws, where the reference coerces it to *true* (so `{stream: 'ture'}` there quietly turns streaming on) — a typo is reported rather than guessed at in either direction; property key order is sorted (`mgp` exposes properties only as an `unordered_map`); float notation is nlohmann's; element id *values* differ per database; durations of 24 h or more grow a day component, since the stored value is a single microsecond count; a duplicated element under `JSON_ID_AS_KEYS` collapses to one entry rather than emitting the same object key twice; `datetime('…Z')` keeps its zone name because Memgraph binds a trailing `Z` to `Etc/UTC` while the reference leaves the offset anonymous; sub-minute zone offsets are truncated to whole minutes, which is where `utils::Timezone::OffsetDuration` stops; and `YIELD *` returns the 12 columns alphabetically rather than in declared order, because the kernel keeps a procedure's results in an ordered map.

An enum property is refused with an explicit error: the property path hands enums over with empty type and value names, so the only thing that could be written is `"::"`.

### Memory and throughput

Elements are serialized as they arrive and written straight to the sink rather than accumulated as a document and rendered at the end. Measured on a 500k-node graph (89 MB payload), peak RSS over baseline:

| | before | after |
|---|---|---|
| `json_all` → file | 637 MB, 1.68 s | **0.2 MB**, 1.19 s |
| `json_all`, counters only | 445 MB, 1.20 s | **0.2 MB**, 0.34 s |
| `json_all` → `data` column | ~637 MB | 308 MB |

A file export is now effectively constant-memory. A streamed payload still has to exist as one string, because a procedure returns it as a single column, so only the file sink becomes constant. Against `export_util.json`, the Python exporter these aliases used to resolve to, the same 500k-node export takes 1.68 s versus 10.47 s — about 3.6× per byte written, since the two formats differ in size.

A file is written through a temporary renamed into place on success, so a failure part-way through leaves any previous export intact rather than the truncated file a straight truncate-then-write would leave. Targets that cannot be renamed into place — a device or a fifo — are written directly.

Argument lists are read as non-owning views: `mgp::List(args)` would deep-copy the whole argument list and `Value::ValueList()` the element list again, which cost 2N vertex copies per call before anything was serialized.


imilinovic:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - **Added:** the `export` query module — `export.json_data`, `export.json_all` and `export.json_graph` write or stream a graph as JSON in the standard interchange format (string ids, inlined relationship endpoints, ISO-8601 temporals, and the `JSON_LINES` / `JSON` / `JSON_ID_AS_KEYS` shapes), so migrated queries that export JSON now produce the expected output; use `export.*` for interchange and keep using `export_util.*` where the output is fed back through `import_util`. Writing to a file uses constant memory regardless of graph size, and replaces the target only once the export has finished, so a failure part-way through leaves the previous export intact. `export.json_graph` accepts `edges` as well as `relationships` in its graph map, so it composes with `project()`. [#4531](https://github.com/memgraph/memgraph/pull/4531)
    - **Changed:** the `apoc.export.json.all` and `apoc.export.json.graph` compatibility aliases now resolve to the new module, so they return the 12-column export row instead of `(path, data)` and `.graph` takes a single graph map instead of two lists — update any query that yields `path` from those aliases to yield `file`, or call `export_util.json` / `export_util.json_graph` by name to keep the previous behaviour. For the same reason `apoc.export.json.all` no longer produces a file that `apoc.import.json` can read back; use `export_util.json` with `import_util.json` for that round trip. [#4531](https://github.com/memgraph/memgraph/pull/4531)
- [x] **[ Documentation PR link memgraph/documentation#1717 ]**

DavIvek (approved):

PR #4534, opened 2026-08-07
feat(query): terminate all transactions with TERMINATE TRANSACTIONS "*"

as51340:
### Description

Adds `TERMINATE TRANSACTIONS "*"`, which terminates every transaction the caller is authorized to kill, so an operator no longer has to copy 19-digit ids out of `SHOW TRANSACTIONS` one at a time.

**No grammar or AST change.** `transactionId : literal` already accepts a quoted string, so the wildcard is detected in `HandleTransactionQueueQuery` and handled by a new `InterpreterContext::TerminateAllTransactions`, which shares its per-victim core (CAS `ACTIVE → VERIFYING`, status restore, authorization) with the existing id-list path. `DROP DATABASE ... FORCE`'s call site is untouched.

#### Semantics

| Decision | Behavior |
| --- | --- |
| Candidate set | The `SHOW TRANSACTIONS` set — all databases, not scoped to the caller's current database |
| Authorization | Per victim, identical to the id-list form: owns it, **or** holds `TRANSACTION_MANAGEMENT` on the victim's database |
| Unprivileged caller | Reaches only its own transactions; never errors on authorization, just returns fewer rows |
| Unauthorized victims | Absent from the output rather than reported, matching `SHOW` — no id is leaked |
| Self | Skipped |
| Output | Existing `{transaction_id, killed}` header, one row per terminated transaction, ascending id, zero rows when nothing matched |
| Mixing | `"*"` must be the sole argument; `TERMINATE TRANSACTIONS "*", '123'` is an error |
| Non-`ACTIVE` victims | Skipped, as today — a committing or aborting transaction cannot be pinned |

Self is skipped because terminating the issuing transaction makes its own `Commit()` throw, so the caller would get an error instead of the rows and would never learn what it killed. Two sessions sweeping concurrently can each kill the other; the `ACTIVE → VERIFYING` CAS serializes them and the loser sees fewer victims.

Query stripping turns the literal into a `ParameterLookup`, so a parameterized wildcard is indistinguishable from a literal one — `TERMINATE TRANSACTIONS $id` with `$id = "*"` is the same statement, and is covered by a test.

### Behavior changes to existing `TERMINATE TRANSACTIONS '<id>'` (for release notes)

Both in the first commit, both cases where the statement reported an outcome that did not happen:

1. **A matched-but-unauthorized victim now reports `killed: false` instead of `true`.** The authorization check ran *after* the id had been moved into the killed partition, so an unprivileged user naming someone else's id got `killed: true` back while the transaction kept running — a false positive that also confirmed the id existed. An unauthorized match now stays in the not-found partition, indistinguishable from a missing id.

2. **Ids must parse in full.** `std::stoul` accepted a numeric prefix, so `TERMINATE TRANSACTIONS '9223372036854775809abc'` silently terminated transaction `9223372036854775809`; anything unparseable degraded to `UINT64_MAX` and was reported as a kill attempt on `18446744073709551615`, an id the user never typed. Malformed ids now raise an error. This is what gives `"*"` an unambiguous home — exact wildcard match first, strict parse second, no sentinel.

### Known wart (pre-existing, unchanged)

An in-flight **system transaction** (`CREATE DATABASE`, `GRANT`) is swept and reported `killed: true`, but system transactions are not abortable — `Interpreter::Commit()` says so explicitly — so it runs to completion. Reachable today by naming such a transaction's id; the wildcard makes it routine. Not fixed here: `system_transaction_` is assigned *after* the release-store that publishes `ACTIVE`, so probing it from the terminating thread is a data race, and excluding these would need new synchronized state on `Interpreter` to produce a more honest report about a transaction we still cannot kill. Making the kill real is a separate feature with an existing TODO.

Out of scope: `SHOW TRANSACTIONS ON *` / `ON 'db1'` database scoping. `SHOW` already returns everything the caller can see, so the wildcard needs nothing there, and explicit scoping composes cleanly with this later.

### DIFF
--- a/src/query/interpreter.cpp
+++ b/src/query/interpreter.cpp
@@ -7328,6 +7328,14 @@ std::vector<TypedValue> BuildGcTransactionRow(storage::GcRunInfoView const &info
 }
 
 namespace {
+// TERMINATE TRANSACTIONS "*" terminates every transaction the caller may kill. Matched
+// exactly: near-misses are rejected by ParseTransactionId rather than silently ignored.
+constexpr std::string_view kTerminateAllWildcard = "*";
+
+bool IsTerminateAllWildcard(TypedValue const &value) {
+  return value.IsString() && std::string_view{value.ValueString()} == kTerminateAllWildcard;
+}
+
 // Transaction ids arrive as string literals, so the whole string must parse; a partial
 // parse would silently target a different transaction than the one named.
 uint64_t ParseTransactionId(TypedValue const &value) {
@@ -7347,7 +7355,7 @@ uint64_t ParseTransactionId(TypedValue const &value) {
 
 Callback HandleTransactionQueueQuery(TransactionQueueQuery *transaction_query,
                                      std::shared_ptr<QueryUserOrRole> user_or_role, const Parameters &parameters,
-                                     InterpreterContext *interpreter_context) {
+                                     InterpreterContext *interpreter_context, Interpreter const *self) {
   auto privilege_checker = [](QueryUserOrRole *user_or_role, std::string const &db_name) {
     return user_or_role &&
            user_or_role->IsAuthorized(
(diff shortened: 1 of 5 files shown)


colinbarry (approved):
Approved 🔥

as51340:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Adds TERMINATE TRANSACTIONS "*", which terminates every transaction the caller is authorized to kill, so an operator no longer has to copy 19-digit ids out of SHOW TRANSACTIONS one at a time. [#4534](https://github.com/memgraph/memgraph/pull/4534)
- [x] **[ Documentation PR link ](https://github.com/memgraph/documentation/pull/1716)**

PR #4537, opened 2026-08-07
fix: user agent fail on MAGE e2e

mattkjames7:
Test examples for `xml_module` now require a User-Agent header to be set otherwise we get a 403 when loading from https://www.w3schools.com.


colinbarry (approved):
Approved ✅

PR #4538, opened 2026-08-07
fix: Resolve IN-list cardinality and cost estimation

colinbarry:
Fixes cardinality estimations when using `IN` expressions with `label+properties` indices.

Previously, `MATCH (x:L) WHERE x.p IN ["a", "b"]` would unwind the `IN` to two rows with an anonymous symbol that could not be resolved by `ResolveAtPlantime`. It would instead fall back to an estimate based on total vertex count of that label and property.

We now pass the underlying membership list through to the estimator, where it can iterate the list members and sum the estimated cardinality for each property value.

This also supports multiple properties against a composite index, e.g., for `WHERE x.a IN [...] AND x.b IN [...])`, we estimate each slot's marginal cardinality independently and combine them.

The same unresolved-symbol problem existed in the index rewriter's `EstimateIndexedScanCardinality`, which is used when comparing candidate indices (including `USING` index hints). Both paths now share the same `EstimateInListSum` helper.

### DIFF
--- a/src/query/plan/cost_estimator.hpp
+++ b/src/query/plan/cost_estimator.hpp
@@ -771,12 +771,13 @@ class CostEstimator : public HierarchicalLogicalOperatorVisitor {
       -> std::optional<double> {
     auto *mapper = db_accessor_->GetStorageAccessor()->GetNameIdMapper();
     utils::OnScopeExit restore([&, saved = resolved_ranges[slot]]() { resolved_ranges[slot] = saved; });
+    auto pvrs = resolved_ranges | ranges::views::transform([](auto const &opt) { return *opt; }) | ranges::to_vector;
     double sum = 0.0;
     for (auto *elem : list.elements_) {
       auto resolved = ExpressionRange::Equal(elem).ResolveAtPlantime(parameters, mapper);
       if (!resolved) return std::nullopt;
       resolved_ranges[slot] = *resolved;
-      auto pvrs = resolved_ranges | ranges::views::transform([](auto const &opt) { return *opt; }) | ranges::to_vector;
+      pvrs[slot] = *resolved;
       sum += db_accessor_->VerticesCount(label, properties, pvrs);
     }
     return sum;
--- a/src/query/plan/cost_estimator.hpp
+++ b/src/query/plan/cost_estimator.hpp
@@ -747,7 +747,7 @@ class CostEstimator : public HierarchicalLogicalOperatorVisitor {
       case Type::IS_NOT_NULL:
         return db_accessor_->VerticesCount(property);
       case Type::EQUAL:
-      case Type::IN: {  // IN is lowered to EQUAL + Unwind at plan-build time, so fallthrough here is defensive
+      case Type::IN: {
         if (auto val = ConstPropertyValue(range.lower_->value())) {
           return db_accessor_->VerticesCount(
               property, storage::ToPropertyValue(*val, db_accessor_->GetStorageAccessor()->GetNameIdMapper()));
@@ -769,6 +769,8 @@ class CostEstimator : public HierarchicalLogicalOperatorVisitor {
     return EstimateInListSum(db_accessor_, label, properties, list, slot, pvrs, parameters);
   }
 
+  // Helper function to estimate cardinality for label properties queries.
+  // Used by both single-threaded and parallel scan operators.
   double EstimateLabelPropertiesCardinality(storage::LabelId label,
                                             std::vector<storage::PropertyPath> const &properties,
                                             std::vector<ExpressionRange> const &expression_ranges) {
--- a/src/query/plan/cost_estimator.hpp
+++ b/src/query/plan/cost_estimator.hpp
@@ -762,9 +762,9 @@ class CostEstimator : public HierarchicalLogicalOperatorVisitor {
     std::unreachable();
   }
 
-  // For an expression of the form `x IN [lst]`, resolve the IN membership list
-  // by summing per-element VerticesCount. Returns `nullopt` if any element
-  // cannot be resolved at plan time.
+  // Compute the marginal sum for a single IN slot: temporarily mark the slot as
+  // IsNotNull for other callers, iterate its elements, sum VerticesCount.
+  // All other slots in resolved_ranges must already be non-nullopt.
   auto EstimateInListCardinality(storage::LabelId label, std::vector<storage::PropertyPath> const &properties,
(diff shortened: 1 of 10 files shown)


colinbarry:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Improve query plans by using correct cardinality estimations for IN expressions with label+properties indices. The cost estimator will now correctly cost queries with expressions of the form `MATCH (x:L) WHERE x.p IN ["a", "b"]`, where we have a label+properties index on :L(p)`. [#4538](https://github.com/memgraph/memgraph/pull/4538)

imilinovic:
## 1. The list length is counted twice

`cost_estimator.hpp:441-446` (Unwind factor) + `:809-813` (scan factor)

Every `PostVisit` contributes a **rows-per-input-row** factor, and the `Unwind` is the scan's *input* (`rewrite/general.cpp:28`), so both apply. This branch made the Unwind factor `elements_.size()` **and** made the scan factor the *total sum over all elements*. Measured on the real plan shape over the existing `AddVertices(100, 30, 20)` fixture:

| list | true rows | estimate |
|---|---|---|
| `[5,10]` | 2 | **4** |
| `[2,4,6,8]` | 4 | **16** |
| `[1..8]` | 8 | **64** |
| `[5,5,5]` | 1 | **9** |

Those five rows are measured. Everything numeric below them — the `n/10` figures in #2, the post-fix test expectations in #3, and the edge/plan-flip examples — is derived from the cost constants, not measured.

`estimate = n × Σ` — the over-estimate factor *is* the list length. Master is a flat 50 here for any `n`, so **from `n = 8` on this fixture the new estimate is worse than what it replaces** (generally once `n²·m > 2.5·V`). That's the case the PR targets, and on a multi-pattern query it flips the start node.

**Fix:** return the per-element average — `Σ / n_distinct` at `:809-813`, each marginal at `:825-828` over its own list's distinct count. Not a fudge: every other scan estimator here already returns a per-input-row factor. Three conditions:

- The divisor must equal the Unwind's factor. Use *distinct* on both sides and this also fixes the dedup bug (`toSet` dedups at runtime, `rewrite/general.hpp:44-56`, but neither factor does — hence `[5,5,5]` → 9, and `[1, 1.0]` → 4 vs a true 1).
- Divide in the caller, not in `EstimateInListSum` — `EstimateIndexedScanCardinality` has no Unwind factor multiplied in and needs the raw total.
- Guard the empty list: `IN []` is currently 0 (correct); a division makes it `0/0` → NaN.

Note the inverse fix — neutralise the Unwind factor and keep `Σ` — doesn't work: `n` is genuinely correct for `ScanAllById` (factor 1) and for `ScanAll`/`ScanAllByLabel` under an Unwind (they really do re-scan per element). It has to stay.

**Worth benchmarking rather than unit-testing alone.** The competing plan is `ScanAllByLabel → Filter(x.p IN list)`, and that filter is *much* cheaper than the estimator thinks: `PrepareCaching` (`frame_change_caching.hpp:28`) registers the IN list with `FrameChangeCollector`, so `Visit(InListOperator&)` (`eval.hpp:419`, cache at `:453-454`) materialises it once into a `flat_hash_set` and does O(1) membership per row. The estimator has no model of that — `CardParam::kFilter` is a flat 0.25 whatever the list length. So today's `n × Σ` inflation is partly *compensating* for an unmodelled optimisation and pushing the planner toward the cached-filter plan, which for long lists may well be the better choice. Fixing this makes the `n`-seek plan look `n×` cheaper, so please check long-IN-list queries end to end, not just the estimate. (Hypothesis — depends on how close the two plans land in the current model; I haven't measured it.)

## 2. Two scan estimators get multiplied by `n` without knowing about the list

`ExtractListFromInUnwind` matches the Unwind's own expression, and every IN lowering uses the same `UnwindMembershipList`, so the `n` factor applies universally — but only `EstimateLabelPropertiesCardinality` learned to read `membership_list_`. It hurts where the scan factor is a **stand-in for the filter's selectivity** rather than a real per-row cost, since those don't scale with the input:

- **`cost_estimator.hpp:744-762`** — global vertex-property index. `Type::IN` shares the `EQUAL` body, `ConstPropertyValue` fails on the anon identifier → `VerticesCount(property) * kFilter`. The list is *already on the operator* (`index_lookup.hpp:1722-1734`); the branch's consumer at `:1621-1630` was wired to `EstimateIndexedScanCardinality`, which only feeds `ShouldUseSTShortestPath`. Should be a few lines to fix.
- **`cost_estimator.hpp:184-201`, `:222-236`** — edge indices. `edge_index_lookup.hpp:1083`/`:1189` pass `unwound.element` as the operator's `expression_`, a bare `Expression *` (`operator.hpp:828`), so there's nowhere to hang the list.

On both the change reduces to `10 → n` with the other factor pinned: branch/master `= n/10`, pivot at `n = 10` only because that's `kUnwindNoLiteral`. 1000 valued edges + a 50-value IN → truth ~50, master 2500, branch 12,500. So it degrades on long lists, on two index families instead of the one that was fixed.

Not affected: `ScanAll`/`ScanAllByLabel`/`ScanAllById` — their factors are real per-row costs (or 1), so `n` composes correctly; where they sit under an IN Unwind the branch is *more* accurate than master.

Suggest fixing the vertex-property estimator here, and for edges either mirroring this branch's own pattern (a non-owning `ListLiteral *` beside `expression_`, cloned in `Clone`, read in the two `PostVisit`s) or stating in the PR description that edge IN keeps a flat factor. The trade is defensible; being silent isn't. Neither path has any test coverage today.

## 3. Nothing in the new tests fails if the change is reverted

- All four new `...InList*` tests pass `nullptr` as the scan input → `Once`, not `Unwind` (`operator.cpp:1047-1048`), so **the composed shape is never built**
- `ExpectScanAllByLabelProperties` ignores `membership_list_` (`tests/unit/query_plan_checker.hpp:601-626`), which is why `In(fake_identifier, nullptr)` passes at `query_plan.cpp:2508,2532`. A `capture_membership_list` stuck at `nullptr` would leave everything green.
- Both e2e tests use 500 `:Big` vs 5 `:Small`; master already prefers `:Small` by ~40×, so both pass on master.

Give all four scans their `Unwind` input and the expectations become `1, 2, 0, 1` — the true row counts, hand-checkable. Two of the four change value, which is the diagnostic, not an inconvenience: today's values encode "the scan factor is the total". Also worth having `compare_expression_range` check `membership_list_`, so something asserts the rewriter captured it.

imilinovic (approved):

PR #4546, opened 2026-08-10
File IO improvements

Ignition:
Snapshot writes and snapshot recovery each push a large file through the page cache with no hint to the kernel, so a big write throttles every writer on the box and the file stays resident afterwards, evicting the working set. Writes now hand each window to writeback and drop it once clean, and both a finished snapshot and a recovered one are released from the cache when done. Off for the WAL. The release happens after recovery rather than behind the read point on purpose: recovery revisits ranges it has already read, so dropping during the load sends them back to the device a second time, which doubles physical reads while leaving elapsed time unchanged on fast local storage. No format change. The window is settable with --storage-snapshot-writeback-window-mib, default 32 and 0 to disable, because pacing does nothing on a filesystem that ignores the calls it is built from and protects nobody on an instance that is not serving anything while it writes; it applies per snapshot writer, so a parallel snapshot holds it once per thread. One behaviour change: a writeback error during a paced snapshot is now fatal like a failed fsync. Also included: an unrelated allocation fix in the GC's index sweep.

### DIFF
--- a/src/utils/file.cpp
+++ b/src/utils/file.cpp
@@ -12,8 +12,11 @@
 #include "utils/file.hpp"
 
 #include <fcntl.h>
+#include <sys/sendfile.h>
 #include <unistd.h>
 
+#include <algorithm>
+#include <cerrno>
 #include <chrono>
 #include <cstdint>
 #include <cstring>
@@ -22,6 +25,7 @@
 #include <ranges>
 #include <thread>
 #include <type_traits>
+#include <utility>
 #include <vector>
 
 #include "utils/exceptions.hpp"
@@ -145,6 +149,9 @@ bool RenamePath(const std::filesystem::path &src, const std::filesystem::path &d
 }
 
 bool HasReadAccess(const std::filesystem::path &path) { return access(path.c_str(), R_OK) == 0; }
+
+// `len == 0` means "to the end of the file".
+void DropCachedPages(int fd) { ::posix_fadvise(fd, 0, 0, POSIX_FADV_DONTNEED); }
 
 static_assert(std::is_same_v<off_t, ssize_t>, "off_t must fit into ssize_t!");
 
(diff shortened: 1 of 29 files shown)


Ignition:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Writing and loading a snapshot no longer fills up the operating system's file cache, so queries stay faster and more memory stays free while it happens. This works automatically; `--storage-snapshot-writeback-window-mib` tunes how much of a snapshot is cached before being written to disk, or set it to 0 for the old behaviour. [#4546](https://github.com/memgraph/memgraph/pull/4546)
- [x] **https://github.com/memgraph/documentation/pull/1720**

src/utils/file.cpp
```diff
@@ -956,6 +972,110 @@ void NonConcurrentOutputFile::FlushBufferInternal(size_t to_write) {
...
+  // What a failed `sync_file_range` means. EIO and ENOSPC are the same class of failure that makes
+  // a failed `fsync` fatal in this file, and they must not be swallowed: Linux reports a given
+  // writeback error to an open file once, so a pacing call that consumes one and drops it leaves
+  // the later `fsync` in `Sync` free to report success for data that never reached the disk. Every
+  // other failure says pacing does not apply to this descriptor rather than that anything is wrong
+  // with the data; the file is written exactly as it was before pacing, so pacing turns itself off
+  // and the file carries on. A descriptor pacing cannot work on will not become one.
+  auto const failed = [this](int rc) {
+    if (rc == 0) return false;
+    auto const err = errno;
```
as51340: Just check if this needs to be fatal

as51340 (approved):
just check MG_ASSERT

src/utils/file.cpp
Ignition: Fatal because a consumed writeback error is invisible to the later fsync, so continuing would report a snapshot as durable when part of it never reached the device.

PR #4550, opened 2026-08-11
fix(storage): treat an unresolvable nested index path as a null key

### LINKED ISSUE #4549: Nested index segmentation fault on creation
Version: Memgraph v3.12

```
step 1: STORAGE MODE IN_MEMORY_ANALYTICAL
  ok
step 2: CREATE INDEX ON :T(m.k)
  ok
step 3: CREATE (:T {m: 1})
  >>> FAILED at step 3: ServiceUnavailable: Failed to read from defunct connection
ExitCode=139 Running=false OOMKilled=false
```

Reproduced successfully locally, every time and reliably by @Josipmrden

Ignition:
An index on a nested property, such as `m.k`, keys a vertex on the value found by following the path into its `m` property. Writing `m` rebuilds the key the vertex held before that write, so the stale entry can be erased, and that key is read out of the value `m` held before. Nothing constrains that value: it may not be a map, and a map need not carry `k`. A path that does not resolve now yields null, the value the index already keys a vertex on when it has nothing there.

Both writers of a nested index key reach this. IN_MEMORY_ANALYTICAL reclaims the stale entry on the write itself, and index population in the transactional modes replays the property deltas of writers admitted once its scan downgrades to a read hold. In both, a vertex created with a value at `m` that has no `k` is the common case, the value before a first write being null.

### DIFF
--- a/src/storage/v2/indices/label_property_index.cpp
+++ b/src/storage/v2/indices/label_property_index.cpp
@@ -82,7 +82,9 @@ void PropertiesPermutationHelper::Update(PropertyId outer_prop_id, PropertyValue
   auto const &sorted_positions = it->second;
   for (auto const &pos : sorted_positions) {
     auto const *nested_value = ReadNestedPropertyValue(value, sorted_properties_[pos] | rv::drop(1));
-    extracted_values[pos] = *nested_value;
+    // The path need not resolve: `value` may not be a map, or may be one without the property.
+    // No value at the path means a null index key.
+    extracted_values[pos] = nested_value ? *nested_value : PropertyValue{};
   }
 }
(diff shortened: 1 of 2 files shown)


Ignition:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Fixed a crash when writing a property covered by a nested index with a value that has nothing at the indexed path, for example CREATE (:T {m: 1}) under an index on :T(m.k). [#4550](https://github.com/memgraph/memgraph/pull/4550)

colinbarry (approved):
All good 🚀

PR #4552, opened 2026-08-11
fix(mage): clamp path.slice bounds and accept null paths

imilinovic:
`path.slice` clamps out-of-range bounds instead of throwing, and `path.slice`, `path.elements` and `path.combine` accept a null path.

**What**

- An `offset` below `0` clamps to `0`. An `offset` above the path length clamps to it, giving the final node as a zero-length path.
- A `length` is capped at the relationships remaining after `offset`. `-1` still means "to the end of the path"; any other negative length takes no relationships.
- `path.slice` and `path.elements` return null for a null path. `path.combine` returns whichever path it was given, or null when both are null.

**How**

`Path::Slice` mixed signed arguments into unsigned arithmetic (`src/mage/cpp/path_module/algorithm/path.cpp`). A negative `offset` reached `GetNodeAt` as a large `size_t` and threw `Index out of range!`, and `offset + length` wrapped for a negative `length`, so `slice(p, 0, -5)` returned the whole path rather than the start node. Both bounds are now clamped in `int64_t` — the offset against the node index range `[0, path.Length()]`, the hop count against `path.Length() - offset` — so the loop bound can neither overflow nor read past the last relationship.

The three functions move to the low-level registration API so their path arguments can be declared nullable; `mgp::Parameter` cannot express a nullable type. The procedures in this module already use it (`src/mage/cpp/path_module/path_module.cpp`).

**Testing**

Behaviour was diffed against the reference implementation: an exhaustive sweep of path lengths 0–5 against offsets and lengths in `{-9,-3,-2,-1,0,1,2,3,4,5,6,7,9}` (1014 combinations), plus null arguments, omitted arguments and `INT64_MIN`/`INT64_MAX` bounds — identical throughout, no errors on either side. 10 E2E tests under `tests/mage/e2e/path_test/` cover the clamping branches and the null arguments.

### DIFF
--- a/src/mage/cpp/path_module/algorithm/path.cpp
+++ b/src/mage/cpp/path_module/algorithm/path.cpp
@@ -700,6 +700,11 @@ void Path::Elements(mgp_list *args, mgp_func_context * /*ctx*/, mgp_func_result 
   auto result = mgp::Result(res);
 
   try {
+    if (arguments[0].IsNull()) {
+      result.SetValue();
+      return;
+    }
+
     const auto path{arguments[0].ValuePath()};
     const size_t path_length = path.Length();
     mgp::List split_path((path_length * 2) + 1);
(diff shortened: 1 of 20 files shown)


imilinovic:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - `path.slice` clamps an out-of-range `offset` or `length` instead of throwing, and `path.slice`, `path.elements` and `path.combine` accept a null path; a negative `length` other than the `-1` "to the end of the path" sentinel now returns a subpath with no relationships rather than the remainder of the path. [#4552](https://github.com/memgraph/memgraph/pull/4552)
- [x] **[ Documentation PR link memgraph/documentation#1722 ]**

DavIvek (approved):

PR #4556, opened 2026-08-12
Omit vector-index properties from returned graph objects

DavIvek:
## What

Adds a runtime-configurable flag `--storage-omit-vector-index-properties-on-return` (setting `storage.omit_vector_index_properties_on_return`, default `false`, Community) that omits vector-index-backed properties (e.g. embeddings) from whole nodes/edges returned to a client.

When it's on, `RETURN n` no longer ships the embedding — removing the token/payload bloat that high-dimensional vectors cause for LLM agents and for Lab graph exploration — while the value stays fully accessible:

- `n.embedding` and `properties(n)` still return it (explicit access is the escape hatch)
- `DUMP DATABASE`, snapshot, WAL, and replication are byte-for-byte unaffected

"What counts as an embedding" is decided by **vector-index membership**, not by property name or type, so a string named `embedding` or a plain list of floats is never guessed at. The flag is toggleable at runtime via `SET DATABASE SETTING '...' TO 'true'` — no restart — and reading live data at serialization time means there is no staleness.

```cypher
SET DATABASE SETTING 'storage.omit_vector_index_properties_on_return' TO 'true';
MATCH (n:Doc) RETURN n;             -- comes back without `embedding`
MATCH (n:Doc) RETURN n.embedding;   -- still returns the vector
```

## Implementation

- The filter lives in one place — the glue conversion `ToBoltVertex`/`ToBoltEdge`, beside the existing per-property auth skip — so `RETURN n`, edges, paths, and graphs are covered while explicit access, dump, and replication take different paths and are untouched.
- Membership is read from the per-transaction active-indices snapshot (no per-row allocation), respecting each index's label / edge-type filter (including `WILDCARD`).


Josipmrden:
:fire:

colinbarry (approved):
Approved 🚀

DavIvek:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    -  Adds a runtime-configurable option to omit vector-index-backed properties from whole nodes and edges returned to clients, reducing response size and payload overhead for high-dimensional vectors. Vectors remain fully accessible through explicit property access, while storage, snapshots, WAL, replication, and database dumps are unaffected. [#4556](https://github.com/memgraph/memgraph/pull/4556)

Issue #4555, opened 2026-08-12
Deletion of 1M edges in HA turns replica to invalid state


Josipmrden:
Steps to reproduce:

1. Create a cluster

Queries on main:
```
CREATE INDEX ON :Supernode;
CREATE INDEX ON :Node;
CREATE INDEX ON :Node(id);
CREATE (:Supernode);
UNWIND RANGE(1, 1000000) as x CREATE (:Node {id: x});
MATCH (s:Supernode) MATCH (n:Node) CREATE (s)-[:HAS]->(n);

Query that's poor: MATCH (s:Supernode) DETACH DELETE n;
```

Observations:
1. Very slow delta execution, around 100 deltas per second -> based on this assumption, query should take around ~3 hours to complete, which is unnaceptable.
2. After some time, RPC timeout is hit

<img width="2762" height="544" alt="Image" src="https://github.com/user-attachments/assets/3dee2561-c3e6-430d-ab9a-aeccc20da72b" />
3. After RPC timeout is hit, replica goes to an invalid state, it's still replicating though, and will never even go to recovery state after some time, because it takes so much time to apply delta changes

<img width="2762" height="544" alt="Image" src="https://github.com/user-attachments/assets/17806d79-8777-4e69-9cfa-837b08ceaefa" />

<img width="2762" height="544" alt="Image" src="https://github.com/user-attachments/assets/446fa250-67f1-4c39-a1d8-6e9e972c0c57" />

-> So one workaround is that we set a lower parameter, but another one should be doing some smarter things with detach delete, which can figure out that the whole node needs to be deleted 🤔 

4. Because of slow delta application, MVCC on the REPLICA also poses problems to latency, with queries stalling: 

<img width="1725" height="811" alt="Image" src="https://github.com/user-attachments/assets/442a9e18-a9af-436a-addd-e5bc210bef3c" />

Flags: 
--logger-type=async
--log-level=TRACE -> this shouldn't degrade performance as it's async

Josipmrden:
hints from agent

Why it's fixable, not fundamental
The API already supports the fix: DetachDelete/PrepareDeletableEdges accept a vector of edges and will aggregate all target gids for a given vertex into one unordered_set, then do a single std::partition pass classifying the whole vector against that set — O(D) total instead of O(D²). That's exactly what happens today when you DETACH DELETE a node normally (ClearEdgesOnVertices, one pass).

[...]

PR #4557, opened 2026-08-12
fix(replication): keep recovery RPCs alive while work is still progressing

as51340:
### Problem

A single WAL delta can occupy a replica for minutes. Index population walks every vertex, so on a large tenant one `WalLabelIndexCreate` delta ran for **over three minutes** while the replica reported nothing — progress was only emitted at delta boundaries, and the transaction had three deltas:

```
04:22:02  Delta 1. Create label index on :Organisation
04:22:02  Vertices size when creating index: 2162077445
          ... 3 minutes of silence ...
04:25:05  Delta 2. Transaction end
04:25:05  Socket write error (fd=128, errno=32): Broken pipe
04:25:05  Applying deltas failed because of Failed to write RPC response
```

The main hit its 30s `CurrentWalReq` budget, shut the socket, and the replica then failed on a broken pipe — discarding the whole build and aborting the transaction. The main reconnected and the cycle repeated, so the replica never converged.

### Approach

`rpc::ProgressHeartbeat` emits `InProgressRes` on a fixed interval, **but only when the handler recorded work since the previous tick**. A handler that stops progressing stops heartbeating, so a genuinely wedged replica is still caught by the peer's timeout rather than masked by an unconditional keepalive — that property is what the design turns on, and it has a dedicated test.

The heartbeat thread is the sole writer of `InProgressRes` for its builder (the socket write behind `slk::Builder` is unsynchronized), and `Stop()` joins before any final response is written.

Progress is now reported from every path that can run long: each delta, each vertex visited while populating an index or validating a constraint, and storage teardown — `SkipList::clear` and `InMemoryStorage::Clear`, which previously ran entirely unobserved, in `SnapshotHandler`'s case before its observer was even constructed.

Index and constraint creation also **abandon early once the peer is gone**, instead of finishing a build whose result can no longer be delivered — the second half of the log above, where the replica ground on for 2.5 minutes after the main had hung up.

### Notable changes beyond the fix

- **Constraints gain the cancellation support indices already had**, including `ConstraintDefinitionCancelationError`. Cancellation is threaded through both the single-threaded and parallel validators and folded into their existing `Synchronized<expected<...>>` accumulation.
- **A latent process-kill is fixed**: a throwing cancel check inside `PopulateIndexOnMultipleThreads` (and the constraint validators' workers) would escape the thread function and call `std::terminate`. Unreachable before this change, and this change makes cancelling checks common on the replica, so the per-worker catch is added now.
- **`SnapshotObserverInfo` is replaced by a plain `ProgressCallback`.** Its only consumer was the replication snapshot handler, and its `ResettableAtomicCounter` ran a CAS loop plus a runtime modulo on *every vertex and edge* of a snapshot load. This removes that cost and leaves one progress mechanism from the skip list up to the RPC handlers. `ResettableAtomicCounter` and the `Observer<void>` specialization had no production users left and are deleted.
- **`deltas_batch_progress_size` no longer has an effect.** The heartbeat paces sends; dividing by the setting would have let a large value starve the heartbeat and reproduce this very timeout through configuration alone. It stays readable and settable so upgrades are unaffected — removing it is a follow-up.


as51340:
Partially fixes https://github.com/memgraph/memgraph/issues/4555. Timeouts shouldn't occur but doesn't fix the replication speed

src/dbms/inmemory/replication_handlers.cpp
```diff
@@ -884,23 +871,30 @@ void InMemoryReplicationHandlers::WalFilesHandler(
   // Snapshot lock needs to be hold for the whole duration of the WalFilesHandler
   auto snapshot_guard = std::unique_lock(storage->snapshot_lock_, std::defer_lock);
 
+  // Covers everything that can run long enough for the main to give up on us: the force-reset teardown below and the
+  // WAL application further down. Stopped before any response is written, since both write to res_builder.
+  rpc::ProgressHeartbeat heartbeat{res_builder};
```
andrejtonev: **Heartbeat is constructed *after* a potentially-slow `AbortPrevTxnIfNeeded`, so its first tick can land after the peer's 30s read timeout** — `src/dbms/inmemory/replication_handlers.cpp:876` (WalFiles; abort at `:857`) and `:1029` (CurrentWal; abort at `:1011`). First `InProgressRes` fires at `D + 25s` (`Run()` waits a full interval); with only ~5s headroom, a large interrupted-2PC `Abort()` (`storage.cpp:1486`, O(deltas)) can blow the budget. Low severity: interrupted-large-2PC path only, self-healing (single spurious timeout, not a thrash loop), not a regression.
   Fix: do **both** — (1) move `ProgressHeartbeat` construction *above* `AbortPrevTxnIfNeeded`, **and** (2) thread a `RecordProgress` callback into `Abort()`'s delta walk. Each alone is insufficient (move-only → nothing calls `RecordProgress` during abort so the tick stays silent; callback-only → no live heartbeat object during the abort). Both are sole-writer-safe (`Abort()` never touches `res_builder`).

src/storage/v2/inmemory/unique_constraints.cpp
```diff
@@ -692,7 +711,30 @@ void InMemoryUniqueConstraints::DropGraphClearConstraints() {
   container_.WithLock([](ContainerPtr &container) { container = std::make_shared<Container const>(); });
 }
 
+void InMemoryUniqueConstraints::RetireConstraint(IndividualConstraintPtr evicted) {
+  if (!evicted) return;
+  // Release the gauge now rather than letting ~ScopedGauge do it: the constraint stops being active the moment the
+  // DROP commits, whereas the object below outlives that until GC reclaims it. Leaving them coupled would report a
+  // dropped constraint as still active for as long as reclamation is outstanding.
+  evicted->gauge_ = {};
+  retired_.WithLock([&evicted](auto &retired) { retired.push_back(std::move(evicted)); });
```
andrejtonev, line 729: **`retired_` not reaped on `Clear()` / `DropGraphClearConstraints()`** 
No leak (freed in dtor), pure retention: with periodic GC off and no `FREE MEMORY` they linger until teardown. Fix: reap `retired_` on DropGraph.

src/storage/v2/constraints/utils.hpp
```diff
@@ -31,11 +36,13 @@ void do_per_thread_validation(ResultType &result, Func &&func,
     auto vertex_curr = vertices.find(gid_start);
     DMG_ASSERT(vertex_curr != vertices.end(), "No vertex was found with given gid");
     for (auto i{0U}; i < batch_size; ++i, ++vertex_curr) {
+      if (cancel_check()) {
+        cancelled.store(true, std::memory_order_relaxed);
+        return;
+      }
       const auto validation_result = func(*vertex_curr, std::forward<Args>(args)...);
```
andrejtonev: **Parallel constraint validator lacks the OOM catch the index path has** —  vs guarded `src/storage/v2/indices/indices_utils.hpp:283-293`. Latent/pre-existing (workers never arm OOM-throwing today), but this PR added the guard on the index path so the asymmetry is worth closing. Fix: mirror the `try/catch(OutOfMemoryException)` into `do_per_thread_validation`, or document the "workers throw only OOM/PopulateCancel" invariant.

src/rpc/progress_heartbeat.hpp
```diff
@@ -0,0 +1,76 @@
...
+  ~ProgressHeartbeat();
+
+  ProgressHeartbeat(ProgressHeartbeat const &) = delete;
+  ProgressHeartbeat &operator=(ProgressHeartbeat const &) = delete;
+  ProgressHeartbeat(ProgressHeartbeat &&) = delete;
+  ProgressHeartbeat &operator=(ProgressHeartbeat &&) = delete;
+
+  // Records that the handler made progress. Sits on per-item paths (every delta, every vertex visited while
+  // populating an index or validating a constraint.
+  void RecordProgress() noexcept { work_done_.exchange(true, std::memory_order_relaxed); }
```
andrejtonev: `RecordProgress` uses `exchange(true, relaxed)` and discards the result on a hot path; plain `store(true, relaxed)` is equivalent and cheaper

andrejtonev (approved):

as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Replicas will now notify main about its progress on an async thread every 25s if there was a progress made. Previously, replicas were sending progress heartbeat every N deltas but this approach was not precise enough. Users shouldn't see anymore replication exceptions occurring due to timeouts. [#4557](https://github.com/memgraph/memgraph/pull/4557)
- [x] **[ Documentation PR link ](https://github.com/memgraph/documentation/pull/1728)**

PR #4559, opened 2026-08-12
feat: support a filter lambda in KSHORTEST path expansion

imilinovic:
**What.** `-[*KSHORTEST (r, n | pred)]->` now returns the k shortest paths of the subgraph the predicate leaves behind. The three-argument `(r, n, p | ...)` form raises a semantic error: KSHORTEST searches bidirectionally and reuses inner searches across Yen's deviations, so there is no one path leading to the edge being tested.

**How.** Everything below the parser already carried `filter_lambda_` — the planner built it, `Clone` copied it, the plan JSON emitted it — and only the cursor ignored it, so the change is a semantic throw in `cypher_main_visitor.cpp` plus `KShortestPathsCursor::ShouldExpand` in `operator.cpp`. The cursor copies `STShortestPathCursor`'s node-binding convention: the source-side pass binds the vertex the edge leads to, the target-side pass the vertex it expands from, which is what makes a bidirectional search test the same `(edge, node)` pairs a forward walk of the finished path would. Yen runs the inner search `O(k·L)` times over the same edges, so the verdict is memoised per input row, keyed on `(edge, node, pass)`; the pass matters only when access checks are on, since that is when the two passes check opposite endpoints.

**Why.** KSHORTEST was the only variable expansion without a filter lambda, so narrowing a k-shortest search meant post-filtering whole paths — which changes the answer, since a path Yen already counted toward `k` cannot be replaced.

### Behaviour changes

- **`|k` now applies per input row.** `n_returned_paths_` was missing from `ResetState`, so it accumulated across rows while `limit_` meant "k paths". Once the running total reached `k`, the check at the top of `Pull` returned false — the cursor's "permanently exhausted" signal — before the input loop ran, so later rows were never pulled: `-[*KSHORTEST|1]->` over N pairs returned one row for the whole query. The count was also non-monotonic; over two pairs with two paths each, `|2` gave 2 rows but `|3` gave 4.
- **`|k` may reference the current row, and a negative `|k` errors.** Identifiers inside the limit were never resolved, so `-[*KSHORTEST|n.k]->` reached the cursor with an unresolved symbol and died on an unwrapped `std::out_of_range` ("An unknown exception occurred"). `SymbolGenerator::PreVisit(EdgeAtom &)` now visits `limit_`, and the cursor evaluates it once per input row beside the depth bounds rather than once per `Pull`. A limit of `0` skips its row instead of ending the pull; a negative limit raises a query error rather than returning no rows.
- **An invalid depth bound now errors** instead of silently disabling the check (both bounds are compared against a `size_t`, so a negative one wrapped). `upper_bound < 1` matches WSHORTEST/ALLSHORTEST; the lower bound rejects only negatives, since no sibling has a lower-bound floor and `0` is inert here.
- **An inlined property map is applied during the search.** With a limit this changes answers, not just cost: `-[*KSHORTEST|1 {p:1}]->` returned 0 rows on a graph whose only matching path was one hop longer than a non-matching one.

### Also fixed

`SymbolGenerator::PreVisit(EdgeAtom &)` held a `Scope &` across its children's `Accept` calls. A pattern comprehension in an edge pattern pushes a scope, reallocating the capacity-1 `scopes_`, so the writes after the traversal landed in freed memory and left the edge identifier unbound — a valid query answered `Unbound variable: r`. Reachable on master from a plain `MATCH`, independent of KSHORTEST.

`in_edges_`/`out_edges_` were the only members missing from the constructor's memory-resource init list and now use the query arena. A consistency fix, not a limit-escape fix — with jemalloc the malloc/extent hooks already charged those bytes. `EdgeVertexAccessorResult` is not allocator-aware, so the cursor caches an arena vector rather than the struct storage returns.

`GenerateCandidatesFromDeviation` compared the *spur* length against the upper bound, but a candidate's total is `deviation_index + spur_len`; the spur now gets `upper_bound_ - deviation_index`. Results are preserved, hop accounting is not — `number_of_hops` drops and a query under `USING HOPS LIMIT` can keep budget it previously spent.

### DIFF
--- a/src/query/frontend/ast/cypher_main_visitor.cpp
+++ b/src/query/frontend/ast/cypher_main_visitor.cpp
@@ -3427,8 +3427,7 @@ antlrcpp::Any CypherMainVisitor::visitRelationshipPattern(MemgraphCypher::Relati
         } else {
           // Other variable expands only have the filter lambda.
           edge->filter_lambda_ = visit_lambda(relationshipLambdas[0]);
-          // Bidirectional search reusing inner searches across deviations has no single
-          // well-defined path leading to the edge being tested.
+          // A bidirectional search reusing inner searches has no one path leading to the tested edge.
           if (edge->type_ == EdgeAtom::Type::KSHORTEST && edge->filter_lambda_.accumulated_path) {
             throw SemanticException("KSHORTEST expansion does not support the accumulated path in a filter lambda.");
           }
@@ -3534,8 +3533,7 @@ antlrcpp::Any CypherMainVisitor::visitVariableExpansion(MemgraphCypher::Variable
     --n_expressions;  // Last expression is the limit
     limit = std::any_cast<Expression *>(ctx->k->accept(this));
   }
-  // After discounting the limit: the grammar puts `| k` in the same `expression` list as the
-  // bounds, so `-[*KSHORTEST 1..3 |2]-` has three of them.
+  // Counted after discounting `| k`, which the grammar puts in the same `expression` list.
   DMG_ASSERT(n_expressions <= 2U, "Expected 0, 1 or 2 bounds in range literal.");
 
   if (n_expressions == 0U) {
--- a/src/query/frontend/ast/cypher_main_visitor.cpp
+++ b/src/query/frontend/ast/cypher_main_visitor.cpp
@@ -3408,9 +3408,6 @@ antlrcpp::Any CypherMainVisitor::visitRelationshipPattern(MemgraphCypher::Relati
         }
         break;
       case 1:
-        if (edge->type_ == EdgeAtom::Type::KSHORTEST) {
-          throw SemanticException("KSHORTEST expansion does not support filter lambda.");
-        }
         if (edge->type_ == EdgeAtom::Type::WEIGHTED_SHORTEST_PATH ||
             edge->type_ == EdgeAtom::Type::ALL_SHORTEST_PATHS) {
           // For wShortest and allShortest, the first (and required) lambda is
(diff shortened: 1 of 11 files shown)


imilinovic:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - feat: `-[*KSHORTEST (r, n | pred)]->` accepts a filter lambda, returning the k shortest paths of the subgraph the predicate leaves behind, so a k-shortest search can be narrowed during expansion instead of by filtering whole paths afterwards. The three-argument `(r, n, p | ...)` form is not supported and raises a query error. [#4559](https://github.com/memgraph/memgraph/pull/4559)
    - fix: `-[*KSHORTEST|k {prop: value}]->` with an inlined edge property map now spends its path limit only on paths that match the map; previously the limit was consumed by the unfiltered search, so the query could return fewer than `k` matching paths, or none at all. [#4559](https://github.com/memgraph/memgraph/pull/4559)
    - fix: the `|k` path limit in `*KSHORTEST` now applies per input row. Previously the expansion stopped producing rows entirely once any one row had returned `k` paths, so `-[*KSHORTEST|1]->` over several source/target pairs returned a single row for the whole query. [#4559](https://github.com/memgraph/memgraph/pull/4559)
    - fix: the `|k` path limit in `*KSHORTEST` may now be an expression that reads the current row, such as `-[*KSHORTEST|n.k]->`. Previously a variable or property inside `|k` was left unresolved and the query failed with "An unknown exception occurred". [#4559](https://github.com/memgraph/memgraph/pull/4559)
    - fix: a negative minimum depth in `*KSHORTEST`, such as `-[*KSHORTEST (0-1)..3]->`, now raises a query error. Previously it enumerated every simple path in the graph before returning no rows. [#4559](https://github.com/memgraph/memgraph/pull/4559)
    - fix: a maximum depth below 1 in `*KSHORTEST`, such as `-[*KSHORTEST 1..0]->`, now raises a query error, matching `*WSHORTEST` and `*ALLSHORTEST`. Previously it returned no rows without raising one. [#4559](https://github.com/memgraph/memgraph/pull/4559)
    - fix: a negative `|k` path limit in `*KSHORTEST`, such as `-[*KSHORTEST|(0-1)]->`, now raises a query error. Previously it returned no rows without raising one. [#4559](https://github.com/memgraph/memgraph/pull/4559)
    - fix: an edge pattern whose inlined property map or variable-expansion depth bounds contain a pattern comprehension, such as `-[r *BFS 1..size([(n)-[e]->(m) | m])]->`, no longer raises a spurious `Unbound variable` error naming the pattern's own edge variable. Resolving such a pattern wrote through a dangling reference. These queries still raise other errors, so this fixes symbol resolution rather than adding support for them. [#4559](https://github.com/memgraph/memgraph/pull/4559)
- [x] **[ Documentation PR link memgraph/documentation#1725 ]**

DavIvek (approved):

PR #4560, opened 2026-08-13
fix: three defects in EXISTS subquery bodies (crash, missing test, false rejection)

### LINKED ISSUE #4478: Memgraph container crashes (SIGSEGV) when executing a list comprehension with an EXISTS subquery

imilinovic:
Three defects in EXISTS bodies. None is caused by #4504, but #4504 makes each reachable from a new
position.

Closes #4478.

### Crashes the server

An EXISTS body that plans to no operators is dereferenced by whatever consumes it. Three shapes,
each a SIGSEGV that drops the server for every session and every database:

```cypher
RETURN EXISTS { RETURN 1 UNION RETURN 2 };
MATCH (n) WHERE EXISTS { RETURN 1 UNION ALL RETURN 2 } RETURN n;
MATCH (n) WHERE EXISTS { MATCH (m) RETURN m AS c UNION CALL { RETURN 1 AS z } RETURN 1 AS c } RETURN n;
```

The empty-row substitution moves into the planner's per-query-part loop, so every part that plans to
nothing gets it before any consumer — union combinator, nested subquery — can dereference it.

### Falsely rejected

```cypher
MATCH (n) WITH n, EXISTS { RETURN 1 } AS e RETURN e;
MATCH (n) WITH n ORDER BY EXISTS { RETURN 1 } RETURN n;
```

Both threw *"Only variables can be non-aliased in WITH."*, though the EXISTS is aliased in the first
and isn't a return item at all in the second. The flag gating that rule stayed set while the visitor
descended into the body, so the body's own `RETURN 1` was judged by the enclosing WITH's rule. It is
now cleared for the body and restored afterwards, leaving the outer WITH's items checked as before.

Both spellings already fail this way on master. The aliasing check runs in the parser, before the
semantic gate that refuses `EXISTS` in these positions, so master reports the aliasing error and the
position error never surfaces — visible in that `EXISTS { RETURN 1 AS c }` reports the position error
while `EXISTS { RETURN 1 }` reports the aliasing one. Neither position is usable on master regardless;
#4504 legalises them, which is what turns this from a misleading message into a false rejection of a
valid query.

### DIFF
--- a/src/query/frontend/ast/cypher_main_visitor.cpp
+++ b/src/query/frontend/ast/cypher_main_visitor.cpp
@@ -3954,8 +3954,11 @@ antlrcpp::Any CypherMainVisitor::visitExistsSubquery(MemgraphCypher::ExistsSubqu
     // Curly-brace subquery form: { cypherQuery }
     // Set the flag to indicate we are parsing an EXISTS subquery
     auto old_flag = parsing_exists_subquery_;
+    // The body's clauses are its own, so the enclosing WITH's "everything must be aliased" rule does not reach them.
+    auto old_in_with = std::exchange(in_with_, false);
     parsing_exists_subquery_ = true;
     auto *cypher_query = std::any_cast<CypherQuery *>(ctx->cypherQuery()->accept(this));
+    in_with_ = old_in_with;
     parsing_exists_subquery_ = old_flag;
     exists->content_ = cypher_query;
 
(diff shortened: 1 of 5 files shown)


imilinovic:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - An `EXISTS { ... }` subquery whose body plans to no operators, such as `EXISTS { RETURN 1 }`, no longer crashes the server and now correctly evaluates to `true`, so queries that previously dropped the connection just work with no changes needed. [#4560](https://github.com/memgraph/memgraph/pull/4560)

colinbarry (approved):
Approved, with no changes requested 🚀

PR #4561, opened 2026-08-13
fix: use index for filters on variables imported into CALL subqueries

imilinovic:
Inside a scoped `CALL (v) { ... }`, a predicate on the imported variable never selected an index. The branch planned as `ScanAll` + `Filter`, and since `Apply` re-runs the branch per input row, that is a full label scan for every row entering the subquery — `node_count x row_count` vertices visited where an index seek would do.

`PreVisit(Apply)` in both index rewriters calls `RewriteBranch`, which builds a **fresh** rewriter for the branch — deliberately, so a branch-internal `SetOnParent` cannot overwrite `Apply::input_`, i.e. the outer plan. That fresh rewriter starts with an empty bound set, so the `are_bound` gate rejects every filter referencing an imported symbol. The symbol is live on the frame, just not produced by the branch's own subtree, which is all `ModifiedSymbols` can see.

Both rewriters now carry an `inherited_bound_symbols_`, set from the `Apply`/`PeriodicSubquery` input's `ModifiedSymbols` when `RewriteBranch` constructs the child, and folded into `are_bound`. `RewriteBranch` also folds in whatever the current rewriter inherited, so `Union`/`Merge`/`Optional` branches and nested subqueries keep the enclosing scope. It stays out of the plan itself: `ModifiedSymbols` has consumers that read it as "produced by this subtree", which inherited symbols are not.

Making more filters indexable reaches two places that decide *plan shape* rather than a seek key, so both are tightened here:

- **`Cartesian` -> `IndexedJoin`** now requires a *removed* filter whose symbols reach into both `left_symbols_` and `right_symbols_`. A predicate confined to one side, or on an inherited symbol which belongs to neither, leaves the branches independent; converting those re-executes the sub-branch once per main-branch row and, because `JoinRewriter` runs afterwards and only rewrites `Cartesian`, forfeits the hash join. The decision reads the removed filters rather than the leftover ones; the stale pre-removal set is left as is, since other conversions depend on it.
- **`id(e)` on a constrained edge pattern** no longer takes the id path. `ScanAllByEdgeId` cannot carry an edge type — its constructor passes `{}` for `edge_types` — nor a node identity, and the id filter is erased, so a typed, multi-type or repeated-node pattern would match the wrong edge. Those patterns keep their `Expand`; untyped ones still get the scan. This is pre-existing behaviour, reachable from `WITH id(t) AS v MATCH (x)-[e:T]->(y) WHERE id(e) = v` and from legacy `CALL { WITH v ... }`, both of which this closes.

Effect, inside a scoped subquery:

| predicate | before | after |
|---|---|---|
| `n.p IN ids` | `ScanAll`+`Filter` | `Unwind`+`ScanAllByLabelProperties` |
| `n.p = v`, `MATCH (n:L {p: v})` | `ScanAll`+`Filter` | `ScanAllByLabelProperties` |
| `id(n) = v` | `ScanAll`+`Filter` | `ScanAllById` |
| `id(e) = v`, `id(e) IN ids` (untyped pattern) | `ScanAll`+`Filter` | `ScanAllByEdgeId` |
| two patterns joined by an equality, one filtered on `v` | `HashJoin`+`Filter` | `HashJoin` + an indexed scan |
| two independent patterns, one filtered on `v` | `Cartesian`+`Filter` | `Cartesian` + an indexed scan |
| both branches of a `UNION` | one branch only | both |

Applies to `CALL (v) { ... }`, `CALL (*) { ... }` and the `CALL ... IN TRANSACTIONS` forms. Legacy `CALL { WITH v ... }` is unaffected — its leading `WITH` already re-binds the symbol inside the branch — as is the edge-type property path, whose candidate selection never consults bound symbols.

One behaviour change: once a predicate drives an index, comparing the property to a value that cannot be a property — a node, edge or path — raises `'vertex' cannot be used as a property value` where the unindexed plan silently matched nothing. Every other bound-symbol path already behaves this way.

Key files: `src/query/plan/rewrite/index_lookup.hpp`, `src/query/plan/rewrite/edge_index_lookup.hpp`.

### DIFF
--- a/src/query/plan/rewrite/edge_index_lookup.hpp
+++ b/src/query/plan/rewrite/edge_index_lookup.hpp
@@ -169,7 +169,7 @@ class EdgeIndexRewriter final : public HierarchicalLogicalOperatorVisitor {
                             .node2_symbol = op.common_.node_symbol,
                             .direction = op.common_.direction,
                             .edge_types = op.common_.edge_types};
-    auto indexed_scan = GenScanByEdgeIndex(op, common);
+    auto indexed_scan = GenScanByEdgeIndex(op, common, /*can_scan_by_edge_id=*/!op.common_.existing_node);
     if (indexed_scan) {
       // We know (and checked) that the input of Expand was a Scan
       // GenScanByEdgeIndex replaced the Expand, yet maintained the input
@@ -979,8 +979,10 @@ class EdgeIndexRewriter final : public HierarchicalLogicalOperatorVisitor {
 
   bool DefaultPreVisit() override { throw utils::NotYetImplemented("Operator not yet covered by EdgeIndexRewriter"); }
 
+  /// @param can_scan_by_edge_id false when the destination node is already bound on the frame.
   template <class TOperator>
-  auto GenScanByEdgeIndex(const TOperator &op, const ScanByEdgeCommon &common) -> std::shared_ptr<LogicalOperator> {
+  auto GenScanByEdgeIndex(const TOperator &op, const ScanByEdgeCommon &common, bool can_scan_by_edge_id = true)
+      -> std::shared_ptr<LogicalOperator> {
     const auto &input = op.input();
     const auto &view = op.view_;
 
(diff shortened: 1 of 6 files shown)


imilinovic:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - fixes: A filter on a variable imported into a scoped `CALL (v) { ... }` subquery now uses a label-property or `id()` index. Previously it scanned every node or edge, once per input row. [#4561](https://github.com/memgraph/memgraph/pull/4561)
    - fixes: `MATCH ()-[e:TYPE]->() WHERE id(e) = ...` no longer returns an edge whose type does not match the pattern. The same applies to multi-type patterns such as `-[e:A|B]->` and to patterns that repeat a node, such as `(x)-[e]->(x)`. [#4561](https://github.com/memgraph/memgraph/pull/4561)
    - fixes: A filter that is not a join condition no longer turns a `Cartesian` into an `IndexedJoin`. That conversion re-ran one branch once per row of the other and prevented a hash join from being used. [#4561](https://github.com/memgraph/memgraph/pull/4561)
    - changed: Comparing an indexed property to a node, edge or path now raises `'vertex' cannot be used as a property value`, where a plan without the index silently matched nothing. [#4561](https://github.com/memgraph/memgraph/pull/4561)

DavIvek (approved):

PR #4564, opened 2026-08-13
fix: Speed up delete edge replication

as51340:
Deleting edges on a replica was quadratic where main is linear.

## The problem

Main's `Delete` operator buffers and calls `DetachDelete` once with the whole edge set, so `DetachRemainingEdges` runs its `std::partition` **once per endpoint vertex**.

The replica got one WAL `EDGE_DELETE` record per edge and applied each in isolation: `FindVertex` ×2, then `FindEdge` (which materializes a filtered, delta-applied `vector<EdgeAccessor>` and linear-searches it), then `DeleteEdge` → `DetachDelete` with a single edge, partitioning `from->out_edges` *and* `to->in_edges` for that one edge.

Deleting D edges off a vertex of degree D: **O(D) on main, O(D²) on the replica.**

The local durability recovery path in `wal.cpp` was already fine (swap-and-pop, single scan) — only replication was affected.

## 1. Batch the deletions

`WalEdgeDelete` now buffers `{edge_gid, from_gid, to_gid, edge_type}` rather than applying immediately, and the batch goes through a single `DetachDelete`.

This is safe and complete because of how the WAL encoder orders a transaction's deltas — five fixed passes (vertex writes → edge creates → edge property sets → **edge deletes** → vertex deletes). Every edge deletion of a transaction therefore arrives as one contiguous run, and the ordering constraints fall out for free: property sets always precede deletes, vertex deletes always follow.

The flush guard sits at the `std::visit` dispatch site rather than inside each handler, so the transaction-end delta (which commits) and the vertex-delete deltas (which require already-detached vertices) are both covered, as is any handler added later. The buffer is capped at 100k edges, because main's `DELETE` buffer size can split one transaction across several `DetachDelete` calls — the replica can see more deletions in a transaction than main ever held at once.

## 2. Resolve edges from the GID, per storage config

New `InMemoryAccessor::DeleteEdgesEx`. A WAL record already names the edge type and both endpoints, so the only open question is which `EdgeRef` the GID denotes:

| config | resolution | per edge |
|---|---|---|
| properties off | no `Edge` object is ever allocated — the `EdgeRef` carries the GID | no lookup at all |
| heavy | `edges_` skip list, keyed by GID | O(log E) |
| light | adjacency — nothing else holds the `Edge*` | see below |

Endpoint lookups are memoized, so a hub's whole fan-out costs one skip-list lookup instead of one per edge.

Deliberately **not** routed through `InMemoryStorage::FindEdge`: that has to recover the edge type and to-vertex from adjacency (`ExtractEdgeInfo` / `ScanOutEdgesForGid`) because its callers know only a GID. Reusing it would put an O(degree) scan back on every edge, including heavy, undoing the fix.

## 3. Light edges resolve from the cheaper endpoint

An edge is linked from both of its endpoints, so each is resolved from whichever adjacency list is shorter (`CheaperScanSide`), with edges landing on the same vertex bucketed to share one scan (`ScanEdgesForGids`, the plural either-direction form of the existing `ScanOutEdgesForGid`).

Both halves matter:

| case | before | after |
|---|---|---|
| delete 10 edges off a 1M-degree hub | 1M-entry scan | ~10 × O(1) |
| delete all D edges of a D-degree hub | one D-entry scan | one D-entry scan |
| D edges from D hubs into one leaf | D × O(hub degree) | one D-entry scan of the leaf |

`CheaperScanSide` reads the two degrees one lock at a time; holding only one lock means no ordering rule is needed and a self-loop needs no special case, unlike `FindEdges` which holds both at once.

## Correctness

`FindEdge` performed a `View::NEW` visibility check that caught a WAL naming an edge the replica does not have. Resolving by GID skips that, so the flush now compares `DeleteEdgesEx`'s returned count against the batch size and fails the transaction on mismatch. That covers all three paths, including a stale `Edge*` read from `edges_` for an already-deleted-but-not-yet-GC'd edge — it won't match adjacency, so it shows up as a short count.

Delta *ordering* within the replica's transaction changes (all out-edge deltas per vertex, then in-edge). Main already has this property, and the WAL encoder re-normalizes into its passes when re-encoding, so cascaded replication is unaffected.

### DIFF
--- a/src/dbms/inmemory/replication_handlers.cpp
+++ b/src/dbms/inmemory/replication_handlers.cpp
@@ -1295,6 +1295,43 @@ std::optional<storage::SingleTxnDeltasProcessingResult> InMemoryReplicationHandl
 
   std::unordered_map<EdgeSetPropertyCacheKey, EdgeAccessor, EdgeSetPropertyCacheKeyHash> edge_set_property_cache;
 
+  // Edge deletions are buffered and applied as one batch. DeleteEdgesEx resolves the whole batch from gids and
+  // hands it to a single DetachDelete, which groups by endpoint vertex and scans each vertex's adjacency once —
+  // so the batch costs O(degree) per vertex where a delta-at-a-time loop costs O(degree) per edge. Main is
+  // already on the batched path (the Delete operator buffers, then calls DetachDelete once), which is why
+  // deleting a dense vertex's edges is linear there and was quadratic here.
+  //
+  // The WAL encoder emits deltas in fixed passes (vertex writes, edge creates, edge property sets, edge deletes,
+  // vertex deletes), so every edge delete of a transaction arrives as one contiguous run and a single batch
+  // absorbs all of them.
+  std::vector<storage::InMemoryStorage::EdgeDeleteSpec> pending_edge_deletes;
+
+  // Bounds the buffer's memory: main may split one transaction's deletions across several DetachDelete calls
+  // (the Delete operator's buffer size), so a transaction can carry far more edge deletes than main ever held
+  // at once. Flushing mid-run only costs one extra adjacency scan per vertex per chunk.
+  constexpr size_t kMaxPendingEdgeDeletes = 100'000;
+
+  auto flush_pending_edge_deletes = [&]() {
+    if (pending_edge_deletes.empty()) return;
+    // A non-empty buffer implies the accessor was created when the first edge delete was buffered.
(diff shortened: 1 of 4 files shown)


colinbarry (approved):
All good - approved 💪

as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Replicating edge deletions will now be orders of magnitude faster since replicas were using different, non-optimized code path. [#4564](https://github.com/memgraph/memgraph/pull/4564)

PR #4565, opened 2026-08-13
fix(triggers): don't fire AFTER COMMIT triggers for aborted txns; don't silently drop queued ones on shutdown

andrejtonev:
AFTER COMMIT triggers fired for STRICT_SYNC transactions that failed at commit, so a trigger acted on rolled-back state; dispatch is now gated on whether the transaction actually committed. A SYNC replica that fails *after* the main finalized the commit still fires its triggers (the data is on main); a STRICT_SYNC 2PC transaction that aborts does not. Separately, ThreadPool::ShutDown() silently discarded queued triggers on shutdown/DROP — it now returns and logs the discarded count and rejects post-shutdown tasks.

### DIFF
--- a/src/storage/v2/storage_error.cpp
+++ b/src/storage/v2/storage_error.cpp
@@ -13,6 +13,7 @@
 
 #include <iterator>
 #include <ranges>
+#include <type_traits>
 
 namespace memgraph::storage {
 
@@ -100,6 +101,18 @@ auto FormatReplicationError(ReplicationError const &error) -> std::string {
   }
 
   return msg;
+}
+
+auto TransactionWasCommitted(StorageManipulationError const &error) -> bool {
+  return std::visit(
+      []<typename T>(T const &arg) {
+        if constexpr (std::is_same_v<std::remove_cvref_t<T>, ReplicationError>) {
+          return arg.transaction_committed;
+        } else {
+          return false;
+        }
+      },
+      error);
 }
 
 }  // namespace memgraph::storage
--- a/src/storage/v2/storage_error.cpp
+++ b/src/storage/v2/storage_error.cpp
@@ -13,7 +13,6 @@
 
 #include <iterator>
 #include <ranges>
-#include <type_traits>
 
 namespace memgraph::storage {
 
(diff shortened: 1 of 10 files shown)


src/storage/v2/storage_error.cpp
```diff
@@ -102,4 +102,9 @@ auto FormatReplicationError(ReplicationError const &error) -> std::string {
   return msg;
 }
 
+auto TransactionWasCommitted(StorageManipulationError const &error) -> bool {
+  auto const *replication_error = std::get_if<ReplicationError>(&error);
+  return replication_error != nullptr && replication_error->transaction_committed;
+}
```
as51340: I think no need for this function, you can reuse boolean replication_error_committed

tests/e2e/high_availability/strict_sync.py
```diff
@@ -431,5 +432,99 @@ def test_mt_strict_sync_commit(test_name):
...
+
+def read_log_since(log_path, offset):
+    with open(log_path, "r") as f:
+        f.seek(offset)
+        return f.read()
+
+
+def log_contains_any(log_path, offset, substrings):
+    content = read_log_since(log_path, offset)
+    return any(substring in content for substring in substrings)
```
as51340: good test but I don't like grepping logs for verification, you can check the state of the instance to see if the trigger was executed or not

tests/unit/utils_thread_pool.cpp
```diff
@@ -116,3 +116,83 @@ TEST(ThreadPool, MoveOnlyLambdaHighConcurrency) {
...
+    pool.AddTask([&] { ran.fetch_add(1); });
+  }
+
+  std::atomic<size_t> discarded{0};
+  std::jthread shutdown_thread([&] { discarded.store(pool.ShutDown()); });
+
+  release = true;
+  shutdown_thread.join();
+
+  ASSERT_TRUE(finished.load());
```
as51340, line 190: I am not sure if this is always true, you rely on the fact that pool will finish its task and set finished = true

as51340 (changes requested):

tests/unit/utils_thread_pool.cpp
andrejtonev: shouldn't be possible, because we only discard tasks that are not in progress. I'll clean up the test code a bit

as51340 (approved):

PR #4567, opened 2026-08-13
feat(multi-tenancy): release the session's DB accessor between queries so idle tenants can drain

andrejtonev:
An idle Bolt session held its tenant's Accessor for the whole connection, so DROP DATABASE could never drain. The session now releases db_acc_ between queries and re-acquires on the next, keeping only the database identity — measured DROP…FORCE convergence went 10.1 s → 0.007 s.


andrejtonev:
Superseded by https://github.com/memgraph/memgraph/pull/4583

PR #4568, opened 2026-08-13
fix(replication): key the 2PC commit cache by uuid and abort the cached accessor before storage teardown

andrejtonev:
The two-phase-commit commit-accessor cache was keyed by the accessor and outlived its storage, so a plain `DROP DATABASE` (and process exit) could free a tenant's storage out from under a live cached accessor — a use-after-free.

### Changes
- **Key the cache by uuid.** The slot records the tenant uuid captured at `Store` time and is aborted for the right tenant via `TakeForTenant(uuid)`, instead of relying on accessor identity.
- **Abort in `~Database`.** The tenant's cached 2PC is aborted before `storage_` is torn down — the one choke point every teardown path (drop, suspend, drop-recreate, process exit) passes through.
- **Tenant-scope the pre-abort call sites** (`PrepareCommitHandler`, `AbortPrevTxnIfNeeded`, and the `CreateDatabase` drop-recreate apply path) so an RPC for one tenant can't strand another tenant's still-pending 2PC.
- **Extract a testable `TwoPCCommitCache` component**, with a unit test covering the slot contract and the `~Database` use-after-free regression (fails on a revert of the `~Database` abort).
- **Own the slot's lifetime with an RAII guard in `main()`** (`TwoPCCommitCache::Owner`) so it is freed deterministically before static destruction — the same `main()`-stack-owned process-singleton pattern as `utils::PageCacheReleaser` — replacing the previously leaked-forever slot and the manual shutdown clear.

The single-slot / ts-only-on-finalize design is safe because the replica replication RPC server is single-threaded, so `Prepare`/`Finalize`/`CreateDatabase` handlers are serialized; the only concurrent slot access is the `defer_pool_` background `~Database`, which is serialized against the RPC thread by the slot mutex.

### DIFF
--- a/src/dbms/inmemory/two_pc_commit_cache.cpp
+++ b/src/dbms/inmemory/two_pc_commit_cache.cpp
@@ -40,11 +40,6 @@ auto TwoPCCommitCache::Slot() -> utils::Synchronized<Record, std::mutex> & {
   return *slot;
 }
 
-auto TwoPCCommitCache::Instance() -> TwoPCCommitCache & {
-  static auto *instance = new TwoPCCommitCache{};
-  return *instance;
-}
-
 void TwoPCCommitCache::Store(std::unique_ptr<storage::ReplicationAccessor> accessor,
                              uint64_t durability_commit_timestamp, utils::UUID uuid) {
   // Extract any pre-existing accessor instead of move-assigning over it in place: destroying it
--- a/src/dbms/inmemory/two_pc_commit_cache.cpp
+++ b/src/dbms/inmemory/two_pc_commit_cache.cpp
@@ -59,7 +59,7 @@ TwoPCCommitCache::Owner::~Owner() {
       spdlog::error(
           "TwoPCCommitCache::~Owner: slot still populated at shutdown -- leaking the accessor rather than aborting "
           "it against freed storage.");
-      (void)cache.commit_accessor_.release();
+      [[maybe_unused]] auto *const leaked = cache.commit_accessor_.release();
     }
   });
   delete slot;
--- a/src/dbms/inmemory/two_pc_commit_cache.cpp
+++ b/src/dbms/inmemory/two_pc_commit_cache.cpp
@@ -1,0 +1,89 @@
+// Copyright 2026 Memgraph Ltd.
+//
+// Use of this software is governed by the Business Source License
+// included in the file licenses/BSL.txt; by using this file, you agree to be bound by the terms of the Business Source
+// License, and you may not use this file except in compliance with the Business Source License.
+//
+// As of the Change Date specified in that file, in accordance with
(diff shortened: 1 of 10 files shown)


src/dbms/replication_handlers.cpp
```diff
@@ -82,6 +82,18 @@ void CreateDatabaseHandler(system::ReplicaHandlerAccessToState &system_state_acc
...
+  // deferred, not synchronous: Update() keeps its own accessor alive across Delete_(), so DeferDelete's
+  // try_delete() sees count_ > 1 and hands the Gatekeeper to defer_pool_, which frees it on a background
+  // thread with no ordering against this RPC thread -- the actual reason to abort here, up front. The
+  // cached 2PC commit accessor is storage-level, not gatekeeper-counted, so nothing drains it before that
+  // free. Same-uuid Update is a no-op salient refresh, so *local != req.config.uuid avoids aborting an
+  // in-flight 2PC for the very tenant being created/updated. Defence-in-depth: reachability here is
+  // unproven.
+  if (const auto local = dbms_handler.GetHotUuid(*req.config.name.str_view()); local && *local != req.config.uuid) {
+    InMemoryReplicationHandlers::AbortTwoPCForTenant(*local);
+  }
```
as51340, line 95: this is only on create, shouldn't we call abort on drop as well?

src/dbms/inmemory/two_pc_commit_cache.cpp
```diff
@@ -0,0 +1,107 @@
...
+void TwoPCCommitCache::Store(std::unique_ptr<storage::ReplicationAccessor> accessor,
+                             uint64_t durability_commit_timestamp, utils::UUID uuid) {
+  // Extract any pre-existing accessor instead of move-assigning over it in place: destroying it
+  // under the lock would run ~InMemoryAccessor's Abort(), which takes engine_lock_.
+  auto stale = Slot().WithLock([&](Record &cache) {
+    auto stale = std::move(cache.commit_accessor_);
+    cache.commit_accessor_ = std::move(accessor);
+    cache.durability_commit_timestamp_ = durability_commit_timestamp;
+    cache.uuid_ = uuid;
+    return stale;
```
as51340, line 84: can the lambda return void since stale is not used

as51340 (changes requested):

src/dbms/inmemory/two_pc_commit_cache.cpp
andrejtonev: stale is returned and stored in the top level stale. it is then destroyed after the WithLock.

src/dbms/replication_handlers.cpp
andrejtonev: already handled there, this is just a rename db bug fix

as51340 (approved):

PR #4573, opened 2026-08-13
feat(dbms): non-blocking DROP DATABASE ... FORCE via deferred teardown worker

andrejtonev:
## What

Make `DROP DATABASE <name> FORCE` **non-blocking**. The physical teardown of a tenant — stopping its
background tasks and destroying its storage — now runs on a single shared background worker instead of
synchronously under the process-wide DBMS lock. Plain `DROP DATABASE` (non-FORCE) is unchanged.

## Why

On master, `... FORCE` runs the whole teardown under `DbmsHandler::lock_`, and destroying the storage
must wait until the drop is the **sole (exclusive) accessor** of the tenant. A stale connection or a
long-running query still holding an accessor makes that wait unbounded. While it waits it holds `lock_`,
so a single stuck FORCE drop:

- **freezes every other database on the instance** (all DBMS operations serialize on `lock_`), and
- **head-of-line-blocks other pending drops**.

## How it works

`DbmsHandler::Delete_` (the FORCE path) now does only cheap, non-blocking work under `lock_` and returns
in microseconds:

1. capture the tenant UUID, retire its durability key;
2. advisory-**seal** the gatekeeper (`Gatekeeper::seal()`) as the last, non-throwing step;
3. hand the teardown to the shared deferred worker (`Handler::DeferDelete`), then release `lock_`.

**Deferred worker** (generic `Handler<T>`): one self-paced `utils::Scheduler` (~50 ms tick, started
lazily on the first drop) walks a `std::list` of pending teardowns. Per node it runs the one-time stop
step (`StopAllBackgroundTasks` + `streams()->DropAll()`), then **polls** `try_delete(10 ms)` — a bounded
trylock, never a blocking wait — until it becomes the sole accessor, then destroys the storage and removes
the on-disk directory. It never blocks on exclusive access: a still-pinned tenant is simply retried on the
next tick and WARN-logged once after 5 minutes, and one stuck drop never blocks the others in the queue.
`~Handler` (shutdown) drains any remaining nodes inline.

**Advisory seal.** The seal marks "this tenant is going away". It is advisory — `access()` still mints
accessors (the worker itself must, to tear the tenant down). It is consumed cooperatively:

- **Replication** (`replication_client.cpp` / `replication_transaction.cpp`): per-tenant recovery/finalize
  tasks bail to `MAYBE_BEHIND` and retire the RPC connection for a sealed tenant, so they stop re-arming
  and pinning it. The drop still propagates to replicas independently, as the replicated `DropDatabase`
  system action — both sites converge on "tenant gone".
- **After-commit triggers** (`database.*`, `interpreter.cpp`): a per-`Database`
  `after_commit_trigger_status_` atomic is set to `TERMINATED` before the trigger pool is joined, so an
  in-flight trigger aborts cooperatively (these triggers run outside the interpreter's transaction
  registry, so the normal terminate path can't reach them).

## Observability

While a tenant is draining it appears as a **`DROPPING`** row in `SHOW DATABASES`, sourced directly from
the worker's pending list. If the name is recreated while the old tenant is still draining, the draining
row is disambiguated as `name (uuid)` so the husk is distinguishable from the live tenant.

## Out of scope

Forcing stale accessors **off** a sealed tenant (session/connection reaping) is a separate change
(#4583). So a pinned tenant stays `DROPPING` until its holder releases — it never blocks anything else,
it just isn't instantaneous. There is no unbounded blocking anywhere on the hot path.

## Testing

- Unit: deferred-worker convergence, `~Handler` drain, advisory seal, `DROPPING` visibility incl. the
  `name (uuid)` disambiguation when a dropped name is retaken.
- E2E: `graph_free_queries/test_dropped_database.py` (query on a force-dropped DB) and
  `system_replication/multitenancy.py` (drop replicates + converges).
- Stress: a workload that force-drops many tenants concurrently — including tenants pinned by live
  sessions — asserting the control database stays responsive throughout and all husks eventually drain.


andrejtonev:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Reworked `DROP DATABASE ... FORCE`: the tenant's teardown — stopping its background tasks and destroying it — now runs on a single non-blocking background worker instead of under the process-wide database lock, so a slow or stuck drop no longer freezes other databases on the instance and one blocked drop no longer head-of-line starves other pending drops. A tenant still draining is shown as `DROPPING` in `SHOW DATABASES` so an in-progress drop is observable; if the name is recreated while the old tenant is still draining, the draining row is disambiguated as `name (uuid)`. Plain `DROP DATABASE` (non-FORCE) is unchanged and no configuration is required. [#4573](https://github.com/memgraph/memgraph/pull/4573)

src/dbms/handler.hpp
```diff
@@ -181,42 +217,56 @@ class Handler {
...
+
+    auto gk = std::move(itr->second);
     items_.erase(itr);
+
+    // splice is noexcept; if emplace_back throws (OOM), node allocation fails before any arg is moved,
+    // so gk and id are intact for the fallback.
+    std::list<PendingDeletion> node;
+    try {
+      node.emplace_back(
+          std::move(gk), std::string{name}, std::move(id), std::move(stop_step), std::move(post_delete_step));
```
colinbarry: Use-after-free here with `name`. `DbmsHandler::Delete_` calls this function with a `string_view` over the `items_` storage key. When we call `items_.erase(itr);` above, the `string_view` becomes dangling. (`DeleteCold_` does the right thing.)

tests/stress/drop_force.py
```diff
@@ -0,0 +1,562 @@
...
+    deadline = time.monotonic() + CONVERGENCE_TIMEOUT_S
+    drv = _make_driver(endpoint, username, password, ssl)
+    remaining: list[str] = []
+    try:
+        while time.monotonic() < deadline:
+            with drv.session() as sess:
+                rows = _run(sess, "SHOW DATABASES")
+            # Memgraph returns a "name" column; guard against driver variants.
+            remaining = [
+                r.get("name", r.get("Database Name", ""))
```
colinbarry: 1. Think this should be `"Name"` rather than `"name"`
2. If the column doesn't exist, `get` returns an empty string. This gets dropped from the `remaining` list comprehension because it fails the `.startwith("tenant_")` check below. Convergence checks will always then pass on the first try. Maybe change to something like:

```python
remaining = [r["Name"] for r in rows if r["Name"].startswith("tenant_")]
```

src/dbms/dbms_handler.cpp
```diff
@@ -839,28 +842,35 @@ DbmsHandler::DeleteResult DbmsHandler::Delete_(std::string_view db_name) {
...
-    (void)std::filesystem::remove_all(storage_path, ec);
-    if (ec) {
-      spdlog::error(R"(Failed to clean disk while deleting database "{}" stored in {})", db_name, storage_path);
-    }
-  });
+  // Non-throwing suffix: advisory seal, then hand ownership to the deferred worker. The worker records
+  // the husk (name + uuid) — surfaced as the DROPPING row in SHOW DATABASES — and tears down off lock_.
+  if (auto *gk = db_handler_.GetGatekeeper(db_name)) gk->seal();
+  db_handler_.DeferDelete(
+      db_name,
```
colinbarry: Because this is deferred, the name is free in `items_` for up to `defer_interval` (10 s) whilst the TTL and async-indexer threads are still alive. If a new tenant is created within that window using the same name, the db being deleted will get the new tenant's `Protector` (from `MakeDatabaseProtectorFactory`), whose `sealed()` returns `false`.

tests/unit/dbms_handler.cpp
```diff
@@ -904,6 +905,294 @@ TEST(DBMS_Handler, RenameMovesTenantDurabilityRecordVerbatim) {
...
+  while (!std::forward<Pred>(pred)()) {
+    if (clk::now() >= deadline) return false;
+    std::this_thread::sleep_for(std::chrono::milliseconds{10});
+  }
+  return true;
+}
+
+}  // namespace
+
+// While the holder is alive the worker's own accessor bumps count to 2, so try_delete(0ms)
```
colinbarry: Minor point, but the comments are incorrect. `try_delete` is now called with `kDeferTryTimeout`, which is `30ms`.

colinbarry (changes requested):
All good apart from a few stale comments, a use-after-free, and a test that never fails. Fix those up, and I'll approve. 💪

src/storage/v2/replication/replication_client.cpp
```diff
@@ -329,7 +336,28 @@ void ReplicationStorageClient::LogRpcFailure() const {
...
+  // error and re-enter MAYBE_BEHIND, from which the heartbeat loop re-establishes the stream.
+  // This disruption is deliberate and bounded: MAYBE_BEHIND -> heartbeat -> re-stream.
+  spdlog::warn(
+      "Retiring shared RPC connection to replica {} for sealed tenant drop. All other tenants "
+      "sharing this connection will transiently enter MAYBE_BEHIND and self-heal within ~1 recheck "
+      "interval (replica_check_frequency_, default 1 s). On SYNC/STRICT_SYNC replicas any "
+      "in-flight commit from an unrelated tenant on this connection is interrupted and the client "
+      "will receive an error — no data loss; the client should retry.",
+      client_.name_);
+  AbortRpcClient();
```
colinbarry: `AbortRpcClient()` only needs to run if `stream` is valid. Without one, there is nothing half-written on the socket, and the abort unnecessarily kills another tenant's in-flight RPC.

`if (stream) { ... }`

src/storage/v2/replication/replication_client.cpp
```diff
@@ -329,7 +336,28 @@ void ReplicationStorageClient::LogRpcFailure() const {
       utils::MessageWithLink("Couldn't replicate data to {}.", client_.name_, "https://memgr.ph/replication"));
 }
 
+void ReplicationStorageClient::RetireForSealedTenant(std::optional<ReplicaStream> &stream) const {
+  stream.reset();
```
colinbarry: Call `AbortRpcClient()` before `stream.reset()`. The reset releases the RPC lock, and this leaves a tiny gap where another RPC can write onto the half-framed socket before the abort kills it.

src/dbms/handler.hpp
```diff
@@ -181,42 +217,61 @@ class Handler {
...
-   *
-   * @param name Name associated with the context to delete
-   * @param post_delete_func What to do after deletion has happened
+   * @brief Defer teardown of @p name (erased from items_) to the background worker.
+   * @param id Opaque caller tag (e.g. UUID) surfaced by PendingItems().
+   * @param stop_step Before gatekeeper destruction (e.g. StopAllBackgroundTasks); @param post_delete_step after.
    */
-  template <typename Func>
-  void DeferDelete(std::string_view name, Func &&post_delete_func) {
+  void DeferDelete(std::string_view name, std::string id, std::move_only_function<void(T &)> stop_step,
```
colinbarry: I think we should take `name` by value as `std::string`.

The problems with the `string_view` are:

- It's copied after `Delete_` has sealed the tenant, so a `bad_alloc` leaves a sealed tenant in `items_`
- `emplace_back` passes it as an lvalue, so `PendingDeletion`'s by-value `name_` parameter is a copy of `name_owned` that can throw. The catch fallback then runs on the moved-from originals.

So I suggest taking `std::string name` by value, like `id`. Build it in `Delete_` next to `husk_id`, before the seal, and pass `std::move(name)` through to `emplace_back` and the fallback.

tests/unit/dbms_handler.cpp
```diff
@@ -904,6 +905,294 @@ TEST(DBMS_Handler, RenameMovesTenantDurabilityRecordVerbatim) {
...
+  //    via PendingItems().
+  {
+    const auto statuses = dbms.AllWithHotColdStatus();
+    const bool found = std::any_of(statuses.begin(), statuses.end(), [](const auto &p) {
+      return p.first == "dropping_visible" && p.second == "DROPPING";
+    });
+    EXPECT_TRUE(found) << "a pinned FORCE-dropped tenant must appear as DROPPING in AllWithHotColdStatus() "
+                          "while the pinning accessor is still held";
+  }
+
```
colinbarry, line 1054: `ALREADY_DROPPING` has no test. `DroppingTenantIsVisibleInShowDatabases` already holds a pinned husk between steps 3 and 4, so it only needs:

```cpp
EXPECT_EQ(dbms.Delete("dropping_visible", static_cast<memgraph::system::Transaction *>(nullptr)), std::unexpected{memgraph::dbms::DeleteError::ALREADY_DROPPING});
EXPECT_EQ(dbms.TryDelete("dropping_visible"), std::unexpected{memgraph::dbms::DeleteError::ALREADY_DROPPING});
```

src/dbms/dbms_handler.cpp
```diff
@@ -463,6 +463,12 @@ DbmsHandler::DeleteResult DbmsHandler::TryDelete(std::string_view db_name, syste
   // Get DB config for the UUID and disk clean up
   const auto conf = db_handler_.GetConfig(db_name);
   if (!conf) {
+    // A name erased from items_ by DeferDelete but still draining is a known husk (its DROPPING row
```
colinbarry: Word for word duplication of the same code in `Delete_`. It's only six lines (including comments), so not a huge deal, but it could be de-duplicated. Also, could use `std::ranges::contains` here.

colinbarry (approved):
A few other suggestions, some being nits.

colinbarry (changes requested):
Changes requested

src/dbms/dbms_handler.cpp
```diff
@@ -843,31 +852,55 @@ DbmsHandler::DeleteResult DbmsHandler::Delete_(std::string_view db_name) {
...
+  // Placed after the seal + DeferDelete handoff (the drop's point of no return): any earlier failure
+  // leaves the tenant HOT and unsealed, so a not-actually-dropped tenant never loses its rows.
+  // The drop is already committed at this point — an exception from on_uuid_retired_ must not unwind
+  // Delete_ (which would falsely surface failure to the client while the husk drains in pending_).
+  // Swallow and log instead; the uuid's param rows will be cleaned up on next startup or by GC.
+  if (on_uuid_retired_) {
+    try {
+      on_uuid_retired_(*db_uuid);
+    } catch (const std::exception &e) {
+      spdlog::error("on_uuid_retired_ threw after committed drop of '{}': {}", db_name, e.what());
```
colinbarry: Use-after-free here, because we take `db_name` by `string_view` and the underlying data points at the key of the map entry `DeferDelete` removes. I think it's worth swapping the signature to:

```cpp
DbmsHandler::DeleteResult DbmsHandler::Delete_(std::string db_name)
```

paying the cost of a single copy on `Delete_` (which will probably be a string small enough to fit in `std::string`'s small buffer) and making it easier for us to reason about this stuff in the future since we won't have to worry about `string_view` dangling.

colinbarry (changes requested):
Just an (unlikely) use-after-free to fix-up, and we're good. 🤞

colinbarry (approved):
Approved 🎆

PR #4574, opened 2026-08-13
feat(dbms): drain DROP DATABASE in three phases off the process-wide lock (DRAINING state)

andrejtonev:
`DbmsHandler::Delete_` held the process-wide `lock_` across the entire FORCE-drop teardown — including two unbounded thread joins (`StopAllBackgroundTasks`, `streams()->DropAll()`) — so a slow drop stalled every other tenant on the instance. This splits the drop into three phases and runs the blocking part holding nothing.

### What this PR delivers on its own
- **Teardown no longer holds `lock_`.** Phase 1 does in-memory bookkeeping under the lock; Phase 2 releases it, runs the two joins, and drops the operation's own accessor; Phase 3 re-locks to hand off to `DeferDelete`. From the caller's side the lock contract is unchanged: `lock` is held exclusive on entry and on every return, including error paths.
- **A tenant mid-drop is visible, not vanished.** It shows `DETACHED` for the whole drain window (published once, in Phase 1) instead of silently disappearing from `SHOW DATABASES`.
- **Honest errors on a racing drop.** `TryDelete` / `Delete` / `Delete_` / `Rename` report retriable `USING` for a tenant that is draining or mid-suspend, instead of a misleading `NON_EXISTENT`.

### What `DRAINING` is here — and what it is not
`draining_` is an orthogonal flag on the HOT gatekeeper state (not a fifth state), set by a single-flight `begin_drain()` and rolled back by `abort_drain()`. Its role in **this** PR is narrow and load-bearing: once Phase 2 releases `lock_`, `access()` must refuse **new** accessors, or a fresh query would re-arm the background work the joins just stopped. `drain_bypass` lets only the teardown itself still mint.

It is deliberately **not** an active drain here. This PR does not force existing holders off, does not wait for the tenant to become sole-held, and does not close the clone-based re-arm path. That policy — cooperative cancel, a bounded off-lock wait, and the recovery re-arm gate — is **#4575**, which builds directly on the phase split and the marker introduced here.

### Not in scope
- Forcing existing sessions / transactions off the tenant → **#4575** (FORCE ABORT), **#4577** (session eviction).
- Any change to interpreter DROP dispatch or grammar.

Stacked on **#4573**.

### DIFF
--- a/src/dbms/database_handler.hpp
+++ b/src/dbms/database_handler.hpp
@@ -48,7 +48,14 @@ class DatabaseHandler : public Handler<Database> {
   ~DatabaseHandler() override {
     for (auto &db : *this) {
       try {
-        if (auto db_acc = db.second.access()) {
+        // utils::drain_bypass: at process shutdown a tenant can still be mid-drain (draining_ ==
+        // true, value_ still HOT, plain access() would refuse). It must not be skipped here --
+        // this loop's job is to proactively StopAllBackgroundTasks() on every live tenant before
+        // the map (and thus every Gatekeeper in it) is destroyed, and a draining tenant is no
+        // exception. Calling it on a tenant whose drop already stopped its tasks is a harmless
+        // no-op: Scheduler::Stop() / ThreadPool::ShutDown() / Streams::Shutdown() all guard on
+        // their own stop/empty state, so a second call is a no-op rather than a double-join.
+        if (auto db_acc = db.second.access(utils::drain_bypass)) {
           (*db_acc)->StopAllBackgroundTasks();
         }
       } catch (std::exception const &e) {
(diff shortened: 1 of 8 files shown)


andrejtonev:
Superseded by the Option-T defer-the-teardown redesign, folded into #4573.

The whole teardown (StopAllBackgroundTasks + streams DropAll) now runs on the defer worker, off the process-wide lock_ — so the lock is never held across the unbounded joins that this PR's three-phase split existed to release. That makes the machinery here unnecessary:

- three-phase `Delete_` (unlock/relock) — deleted; `Delete_` is fast+atomic under `lock_` (mark, retire key, publish DETACHED, erase→hand off), destruction deferred to the worker.
- the `draining_` marker — deleted; erasing from the map is itself the access gate.
- `rollback_drain` — deleted; there is no failable off-lock window to unwind.

FORCE ABORT (this stack's #4575) is reimplemented on top of #4573 as a fire-and-forget cooperative-cancel (no `AwaitDrain_`/bounded wait). #4575's base has been re-pointed to #4573.

Built + `DBMS_Handler` 17/17 green locally. Closing.

PR #4572, opened 2026-08-13
fix(utils): destroy the managed value after releasing the gatekeeper mtx

andrejtonev:
Gatekeeper::try_delete() destroyed the value under GKInternals::mutex_, deadlocking against ~InMemoryStorage joining TTL/async-indexer threads that block on that same mutex. value_ becomes a unique_ptr<T> moved out under the mutex and destroyed after release — the deadlock repro goes from a hang to 0 ms.

PR #4576, opened 2026-08-13
fix(gatekeeper): mint the DB protector by handle (no off-lock items_ read) + deadlock repros

andrejtonev:
MakeDatabaseProtectorFactory looked the tenant up by name in Handler::items_ with no lock held, from TTL/async-indexer threads (TSan-confirmed). A new Gatekeeper::access_via mints the protector by handle instead, all mint routes share one mint_locked predicate, and the branch ships the deadlock repros validating the value-lifetime base — TSan 10 races → 0.


andrejtonev:
Closing as superseded by #4572.

The load-bearing change here — moving `Gatekeeper::try_delete()`'s value destruction out from under `GKInternals::mutex_` — is functionally identical to #4572, which is a clean, master-based, already-approved PR (`fix/gatekeeper-value-lifetime`). #4572 was verified TSan-clean, and the async-indexer AB-BA deadlock this PR targets is reproduced on master and fixed on #4572 by a regression test now added to #4572 (`TryDeleteDeadlocksUnderRealAsyncIndexerViaAccess`).

The remainder of this PR (the `DRAINING` marker, `begin_drain`/`drain_bypass`, three-phase drop, protector-by-handle/`access_via`) is the abandoned #4574 draining-state lineage — #4574 is closed and its approach was replaced by the deferred-drop design in #4573; that machinery is on neither master nor #4573. Nothing is stacked on this branch.

→ Take the gatekeeper fix via **#4572**. No action needed here.

PR #4073, opened 2026-04-24
Feature: CALL with resources

Josipmrden:
Implement `CALL ()` syntax

PR #4578, opened 2026-08-14
fix: keep scoped CALL subquery imports bound across an intermediate WITH

imilinovic:
Variables imported by `CALL (v1, v2) { ... }` and `CALL (*) { ... }` stay in scope for the whole
subquery body, so a `WITH` inside the body that does not mention them must not drop them. The
planner dropped them anyway, and planned a fresh scan for a variable that was already bound:

```cypher
MATCH (a:Movie {id: "1"})
MATCH (b:Movie) WHERE b.id IN ["3", "5"]
CALL (a, b) {
  MATCH (a)--(x)
  WITH x            // does not mention b
  MATCH (b)--(y)    // planned as ScanAll (b), not an expand from the imported b
  RETURN x, y
}
RETURN DISTINCT x.id, collect(DISTINCT y.id);
```

**What.** Because `Apply` shares the frame with the outer chain, that spurious `ScanAll` overwrote
the slot the outer chain had filled. Three consequences, all silent — no error, no warning:

- wrong rows (the query above collected every node id instead of the two reachable ones);
- the corruption escaped the subquery, so the outer query read the scanned node after the `CALL`;
- `CREATE`/`MERGE` on an import fabricated a brand-new node instead of attaching to the bound one,
  diverging the graph.

`OPTIONAL MATCH` was affected like `MATCH`. Expression-position reads of an import were always
correct, which is why this stayed hidden.

**How.** `PlanningContext` now carries the active import set. `HandleSubquery` seeds it from the
scoped variables it already computes, clears it on the legacy `CALL { WITH v ... }` path, and
saves/restores it around the nested plan so nested subqueries cannot leak imports either way.
`impl::GenWith` re-adds those imports after narrowing `bound_symbols`, skipping any whose name a
named expression redeclared — mirroring the equivalent guard in `SymbolGenerator::VisitReturnBody`,
so `WITH x AS b` still shadows correctly.

The legacy leading-`WITH` import form is deliberately unchanged: it imports nothing, so ordinary
`WITH` narrowing still applies there and the name written afterwards is a fresh variable.

**Why.** Semantic analysis has modelled this scoping rule since the feature landed
(`50aa3fcd6`, #4073); the planner never learned it, so scoped `CALL` has never worked across an
intermediate `WITH`. Affects v3.10.0 onwards.

Covered by four plan-shape unit tests in `tests/unit/query_plan.cpp` (including the legacy form as
a guard against over-broad preservation) and five `subqueries.feature` scenarios for the rows, the
outer-scope leak and the `CREATE` divergence.

### DIFF
--- a/src/query/plan/planner.hpp
+++ b/src/query/plan/planner.hpp
@@ -122,6 +122,9 @@ class PostProcessor final {
 template <template <class> class TPlanner, class TDbAccessor>
 auto MakeLogicalPlanForSingleQuery(QueryParts query_parts, PlanningContext<TDbAccessor> *context) {
   context->bound_symbols.clear();
+  // Only a subquery body has imports, and `HandleSubquery` restores them; clear anyway so the
+  // invariant holds locally rather than by induction over every planner entry point.
+  context->scoped_call_imports.clear();
   return TPlanner<PlanningContext<TDbAccessor>>(context).Plan(query_parts);
 }
 
--- a/src/query/plan/planner.hpp
+++ b/src/query/plan/planner.hpp
@@ -122,8 +122,7 @@ class PostProcessor final {
 template <template <class> class TPlanner, class TDbAccessor>
 auto MakeLogicalPlanForSingleQuery(QueryParts query_parts, PlanningContext<TDbAccessor> *context) {
   context->bound_symbols.clear();
-  // Only a subquery body has imports, and `HandleSubquery` restores them; clear anyway so the
-  // invariant holds locally rather than by induction over every planner entry point.
+  // Only a subquery body has imports; keep each entry point's start state local.
   context->scoped_call_imports.clear();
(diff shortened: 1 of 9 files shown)


imilinovic:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Variables imported by `CALL (v1, v2) { ... }` and `CALL (*) { ... }` now stay bound after a `WITH` inside the subquery body, so these subqueries no longer return wrong rows, corrupt the outer variable, or create a disconnected node instead of attaching to the imported one; if you worked around this by re-listing the imported variables in the inner `WITH`, that is no longer necessary. [#4578](https://github.com/memgraph/memgraph/pull/4578)

DavIvek (approved):

PR #4580, opened 2026-08-14
fix(storage): keep index entries from outliving the objects they name

### LINKED ISSUE #4523: Use-after-free: edge-type+property index dereferences an Edge freed by the edges_ skiplist GC (two sites)
`InMemoryEdgeTypePropertyIndex` stores a raw `Edge *` in each index `Entry`. The
epoch GC can free that `Edge` while the index still holds the pointer, and both
the GC sweep and query-time scans then dereference it. Reproduced under ASAN on
`master` (`0e8aa326`), aarch64.

### The two stacks

Read (query-time scan):

```
InMemoryEdgeTypePropertyIndex::Iterable::Iterator::AdvanceUntilValid()
 ← EdgesIterable::Iterator::operator++()
 ← ScanAllByEdgeCursor<ScanAllByEdgeTypePropertyValue::MakeCursor(...)::$_0>::Pull()
 ← Filter → Produce → PullPlan::Pull → SessionHL::Pull
```

Free (different thread, concurrent write):

```
utils::SkipListGc<Edge, DbAwareAllocator<char>>::Run()
 ← InMemoryStorage::InMemoryAccessor::CreateEdge(VertexAccessor*, VertexAccessor*, EdgeTypeId)
 ← plan::CreateExpand::CreateExpandCursor::Pull()
 ← plan::Merge::MergeCursor::Pull()
```

So: a `MERGE`/`CREATE` drives the edge GC, which frees the skiplist node while a
concurrent `ScanAllByEdgeTypePropertyValue` is walking the index.

ASAN reports it as `heap-use-after-free` while the block is quarantined and
`heap-buffer-overflow` once it has been recycled — one bug, two presentations, a
single faulting PC. In the overflow form the "allocated by" stack names the
block's *new* owner and is a red herring.

### Why the existing pin doesn't cover it

`Iterable`/`ChunkedIterable` already hold `pin_accessor_edge_` for the iterator's
lifetime, and it still faults. The pin protects nodes reachable through the
skiplist from the moment its accessor id was allocated, whereas an `Entry` holds
an `Edge *` captured when the entry was *indexed*. A node already collected —
tagged with an older accessor id — before the pin existed is freed underneath a
scan that holds one.

### Two dereference sites, not one

| site | reached from |
|---|---|
| `RemoveObsoleteEntries` | the background `CollectGarbage` sweep |
| `AdvanceUntilValid_` | query scans, via both `Iterable` and `ChunkedIterable` |

Worth stating explicitly because fixing only the first **moved** the crash
rather than removing it: on an engine patched at the sweep alone, all 20 ASAN
aborts over 96 CI jobs landed in the scan, versus all 3 in the sweep on the
unpatched control.

### Suggested fix

Carry the edge's `gid` on the `Entry` (not part of `operator<`, so index ordering
is unchanged) and re-resolve it through the pinned edge store instead of
dereferencing the stored pointer. At the scan site, require the resolved address
to equal `entry.edge`: a miss means the edge was reclaimed, a hit elsewhere means
the gid was reused, and both mean the entry is stale and must be skipped. Also
read `max_gid` off the entry's own gid — `entry.edge->gid` was the first touch,
requiring the edge to be alive merely to learn its identity.

Light mode keeps no edge skiplist to resolve against and can retain the current
behaviour; the crashes we see are all heavy-edge.

A root-cause alternative worth considering instead: make the pin cover entries
captured before it, so no consumer has to re-resolve.

I have this as a patch against `0e8aa326` covering both sites and am happy to
open it as a PR — say the word, or take the approach and implement it your way.

### Reproduction

Concurrent `MERGE`/`CREATE` traffic against an edge-type+property index while
scans run. It is rare on a release build (the freed read usually hits mapped
memory and passes unnoticed); an ASAN build turns it into a reliable abort — ~21%
of the affected CI jobs, versus roughly 2% observable as SIGSEGV without ASAN.

---

Supersedes #4517, which has the full investigation log but is too long to read.

Ignition:
An index entry holds a raw pointer to the vertex or edge it names, and nothing
keeps that object alive on the entry's behalf, so a collection pass has to
remove the entry before the object leaves the store. Analytical mode broke that
rule three ways: its sweeps spared entries at or past the oldest active start
timestamp, which is sound only where an older transaction can still reach the
version an entry names; its arms removed whatever was deleted at the moment they
ran, having swept earlier in the same pass; and it decided collectability by
reading an object's deleted flag and delta pointer without holding its lock. The
collectable set is now fixed before the sweeps, exactly that set is removed
afterwards, and those reads are synchronised.

Sweeps were armed off transaction deltas, which an analytical transaction does
not produce, so its indexes grew with every write and collected only when a
deletion forced a sweep; the analytical write path now notes the label or
property it touched. Removal moves entirely to the garbage collector: an
aborting transaction used to remove its own objects on its own thread, which
holds back neither a concurrent sweep nor a reader that started afterwards, and
now hands them over as committed deletes already did. The pass no longer scans
for deleted light edges, which cannot be reached from the store at all, skips
whatever the handover has already claimed so that nothing is retired twice, and
holds waiting objects in batches rather than one list node each.


Ignition:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - Fixed use-after-free crashes and unbounded index-entry growth in IN_MEMORY_ANALYTICAL with indexes. No user action needed. [#4580](https://github.com/memgraph/memgraph/pull/4580)

colinbarry (approved):
Approved 🚀

PR #4586, opened 2026-08-15
fix(query): measure and cut strings in code points

### LINKED ISSUE #4428: `size()` counts UTF-8 bytes instead of characters for Unicode strings
### Memgraph version

Memgraph `3.12.0`

### Environment
- Host OS: Linux x86_64
- Deployment: Docker
- Query interface: `mgconsole` via `docker exec`
- Query language: Cypher

### Description

`size()` returns the UTF-8 byte length of a string instead of the character count. For ASCII strings (1 byte per character), the result is correct. For Unicode characters (2-4 bytes in UTF-8), the result is inflated.

Neo4j and FalkorDB both return the character count.

### To reproduce

No graph data required.

```cypher
RETURN size('é') AS v;
```

### Expected behavior

Should return `1` — `'é'` is a single character (U+00E9).

Neo4j 2026.06.0: `1`
FalkorDB 4.20.1: `1`

### Actual behavior

```text
2
```

Memgraph returns the UTF-8 byte count: `'é'` = 2 bytes (0xC3 0xA9), `'中'` = 3 bytes (0xE4 0xB8 0xAD).

### Controls (work correctly)

```cypher
-- ASCII strings: 1 byte = 1 char, so results are accidentally correct
RETURN size('a') AS v;    -- 1
RETURN size('abc') AS v;  -- 3
RETURN size('') AS v;     -- 0
```

### Cross-engine comparison

| String | Memgraph 3.12 | Neo4j 2026.06 | FalkorDB 4.20 |
|--------|:---:|:---:|:---:|
| `size('a')` (ASCII) | 1 | 1 | 1 |
| `size('é')` (U+00E9) | **2** | 1 | 1 |
| `size('中')` (U+4E2D) | **3** | 1 | 1 |
| `size('aéb')` (mixed) | **4** | 3 | 3 |

### Isolated recheck

Fresh Docker container, Memgraph 3.12.0, 2026-07-16:

```
Run 1: size('é')  → 2; size('中') → 3; size('aéb') → 4
Run 2: identical
Run 3: identical
```

Reproduction rate: 3/3 (100%).

### Why this looks like a bug

Neo4j documents `size()` on strings as returning the number of **characters**. Memgraph returns the number of UTF-8 **bytes** instead, which gives different results for any string containing non-ASCII characters. FalkorDB matches Neo4j's character-counting behavior. The byte-count behavior is consistent with measuring the string's length in UTF-8 code units rather than decoding to Unicode code points.

### Impact

Any query using `size()` on strings containing non-ASCII characters will compute incorrect results on Memgraph. This affects `ORDER BY size(name)`, `WHERE size(name) > N`, `substring()` with computed lengths, and any application logic that depends on character counts.

Ignition:
Length, substring, left and right now work in code points, so a position and a
length mean the same number of characters throughout and a result is always
whole characters. Length reported the size of the buffer, and the other three
cut at byte offsets, so a position landing inside a character split it and
returned a fragment that is not valid UTF-8.

The unit is the code point rather than the grapheme cluster: a combining mark
counts on its own, and a cut can still separate one from the character it
follows. Clusters are defined by data that changes between Unicode versions,
which would let the length of a stored value change under an upgrade. Results
change for text outside ASCII.

### DIFF
--- a/src/utils/string.hpp
+++ b/src/utils/string.hpp
@@ -166,6 +166,21 @@ inline std::string ReverseUtf8(const std::string_view s) {
 }
 
 /**
+ * Count the UTF-8 code points of `s`.
+ *
+ * A code point counts once however many bytes encode it, so this is the length
+ * a reader would give the text rather than the size of its buffer. Note that a
+ * combining mark is a code point of its own, so a decomposed character counts
+ * as more than one.
+ *
+ * Bytes that do not introduce a well-formed sequence each count once, so a
+ * malformed string still yields a length rather than an error.
+ */
+inline size_t CountUtf8CodePoints(const std::string_view s) {
+  return static_cast<size_t>(std::ranges::count_if(s, std::not_fn(IsUtf8Continuation)));
+}
+
+/**
  * Uppercase all characters of a string and store the result in `out`.
  * Transformation is locale independent.
  * @return pointer to `out`.
--- a/src/utils/string.hpp
+++ b/src/utils/string.hpp
@@ -168,10 +168,8 @@ inline std::string ReverseUtf8(const std::string_view s) {
 /**
  * Count the UTF-8 code points of `s`.
  *
- * A code point counts once however many bytes encode it, so this is the length
- * a reader would give the text rather than the size of its buffer. Note that a
- * combining mark is a code point of its own, so a decomposed character counts
- * as more than one.
+ * A code point counts once however many bytes encode it. A combining mark is a
+ * code point of its own, so a decomposed character counts as more than one.
  *
  * Bytes that do not introduce a well-formed sequence each count once, so a
  * malformed string still yields a length rather than an error.
@@ -184,26 +182,19 @@ inline size_t CountUtf8CodePoints(const std::string_view s) {
  * Byte offset at which the code point numbered `index` starts, or the size of
  * `s` when it holds fewer than `index + 1` of them.
  *
- * Counting from the front is linear, which is inherent to a variable-width
- * encoding: there is no way to reach the n-th code point without reading what
- * precedes it.
+ * Reaching the n-th code point means reading what precedes it, which a
+ * variable-width encoding leaves no way around.
  */
 inline size_t Utf8OffsetOfCodePoint(const std::string_view s, const size_t index) {
   size_t seen = 0;
   for (size_t offset = 0; offset != s.size(); ++offset) {
-    if ((static_cast<unsigned char>(s[offset]) & 0xC0U) == 0x80U) continue;
+    if (IsUtf8Continuation(s[offset])) continue;
     if (seen == index) return offset;
     ++seen;
   }
   return s.size();
 }
 
-/**
- * Substring of `s` starting at code point `pos` and running for at most `count`
- * code points, so a multi-byte character is never cut in half.
- *
- * Out-of-range positions and lengths clamp rather than throw, as `Substr` does.
- */
 /**
  * Byte offset at which the last `count` code points of `s` begin, or 0 when it
  * holds no more than that many.
@@ -223,6 +214,12 @@ inline size_t Utf8OffsetOfLastCodePoints(const std::string_view s, const size_t 
   return 0;
 }
 
+/**
+ * Substring of `s` starting at code point `pos` and running for at most `count`
+ * code points, so a multi-byte character is never cut in half.
+ *
+ * Out-of-range positions and lengths clamp rather than throw.
+ */
 inline std::string_view SubstrUtf8(const std::string_view s, const size_t pos,
                                    const size_t count = std::string_view::npos) {
   const auto begin = Utf8OffsetOfCodePoint(s, pos);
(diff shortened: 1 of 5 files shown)


Ignition:
### On using standard algorithms for the string helpers

Taken seriously rather than waved off, since three of these are hand-written loops.

**Where a standard algorithm names the operation, it is used.** `CountUtf8CodePoints` is
`std::ranges::count_if`. `ReverseUtf8`'s inner loop is `std::find_if_not` and `std::copy` over a
buffer from `resize_and_overwrite`. `SubstrUtf8` is a composition of the two seeking helpers rather
than a third loop.

**The two seeking helpers ask for "the n-th element satisfying a predicate", which no standard
algorithm names.** The two ways to say it are `views::enumerate | views::filter | views::drop`, or
`find_if` in a loop. Both were written and measured against the explicit loop (ms, lower is
better):

| finding the n-th sequence start | loop | ranges | repeated find_if |
| --- | --- | --- | --- |
| 4 KiB ASCII, index 3000 | **25.3** | 38.4 | 57.1 |
| 3 KiB CJK, index 900 | 22.7 | **19.1** | 20.2 |
| short string, index 6 | **1.77** | 2.26 | 3.38 |

| finding the last n | loop | ranges |
| --- | --- | --- |
| 4 KiB ASCII, last 3 | **0.54** | 0.92 |
| 3 KiB CJK, last 3 | **1.13** | 1.18 |
| 4 KiB ASCII, last 3000 | **25.5** | 50.7 |

All three agree on every input; only the cost differs. The adaptor version loses by 1.5x to 2.3x on
ASCII-dominant text, which is the common case for string properties, and wins only on text that is
entirely multi-byte. It also reads less directly: the offset arrives as `std::get<0>` of an
enumerated pair, filtered through a predicate on `std::get<1>`.

So the loops stay, and they are the two places where the standard library has nothing to say. The
last row also shows why `right()` walks backwards: taking three characters off a 4 KiB string costs
0.54ms against 25.5ms for taking three thousand — it reads only what it returns.

Ignition:
Reported as #4428.

Ignition:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - size, substring, left and right now count characters rather than bytes, so size('中') is 1 instead of 3; recheck non-English text you saved with them, which may be cut mid-character. [#4586](https://github.com/memgraph/memgraph/pull/4586)

tests/unit/query_expression_evaluator.cpp
```diff
@@ -1955,6 +1955,42 @@ TYPED_TEST(FunctionTest, Last) {
   ASSERT_THROW(this->EvaluateFunction("LAST", 5), QueryRuntimeException);
 }
 
+TYPED_TEST(FunctionTest, SizeCountsCodePoints) {
```
colinbarry: Very minor niggle, but would it be worth placing the `SizeCountsCodePoints` directly after the `Size` tests, and the `SubstringLeftRightCountCodePoints` directly after the `Subtring` tests? It would make locating related tests easier to navigate.

src/utils/string.hpp
```diff
@@ -165,6 +165,69 @@ inline std::string ReverseUtf8(const std::string_view s) {
...
+/**
+ * Byte offset at which the code point numbered `index` starts, or the size of
+ * `s` when it holds fewer than `index + 1` of them.
+ *
+ * Reaching the n-th code point means reading what precedes it, which a
+ * variable-width encoding leaves no way around.
+ */
+inline size_t Utf8OffsetOfCodePoint(const std::string_view s, const size_t index) {
+  size_t seen = 0;
+  for (size_t offset = 0; offset != s.size(); ++offset) {
```
colinbarry: Potential optimisation here would be to use the lead byte's high bits to compute the code point sequence length and jump forward that many bytes. Fewer comparisons for multi-byte code points, but wouldn't work with malformed UTF-8 sequences.

colinbarry (approved):
Approved (with some comments) 💪

src/utils/string.hpp
Ignition: Measured it. Made to agree on malformed input (a stray continuation byte advances without counting, and a lead byte whose sequence isn't well formed falls back to length 1) the jump still has to read the continuation bytes in order to validate them, so it lands slower than the byte scan rather than faster: 752 MiB/s against 3.56 GiB/s on 4KB of ASCII, and 1.64 GiB/s against 3.56 GiB/s on CJK.

That validation is not optional. Without it the jump disagrees with the byte scan on malformed input, and since size() and substring() share one predicate today, they would stop agreeing on a truncated string, so left(s, size(s)) could stop returning s.

There is a faster option that keeps the predicate intact: evaluate it a word at a time with SWAR, counting the continuation bytes in eight bytes at once. That measured around 4x on long strings, at the cost of about a nanosecond on strings below ~16 bytes. I'm leaving it out here to keep the implementation simple, since nothing has shown this counting to be hot. It is a self-contained change if that ever turns up in a profile.

PR #4588, opened 2026-08-15
fix(query): match a real literal only with a fraction or exponent

### LINKED ISSUE #4521: Leading-zero literals containing 8 or 9 are parsed as DOUBLE, bypassing INT64 range validation and losing precision
### Memgraph version

Memgraph `3.12.0`
Also reproduced on Memgraph `3.11.0`.

### Environment

* Host OS: Linux x86_64
* Deployment: Docker (`memgraph/memgraph:3.12.0`)
* Query language: Cypher
* Reference engines:

  * Neo4j Community `2026.06.0`
  * Neo4j Community `5.26.27`

### Description

Digit-only literals with a leading zero take different lexer paths depending on their digit content, silently changing their value or runtime type:

* `RETURN 09` produces the DOUBLE value `9.0`, rather than an INTEGER value or a syntax error. This is demonstrated behaviorally by `09 / 2 = 4.5`, while `9 / 2 = 4`.
* `RETURN 9223372036854775808` is rejected with `Integer literal exceeds 64 bits.`, but adding a leading zero causes the same decimal digit sequence to be silently accepted as DOUBLE:

```cypher
RETURN 009223372036854775808 AS x;
```

This bypasses INT64 range validation and loses precision. The two distinct literals below produce the same floating-point result:

```cypher
RETURN
  009223372036854775808 AS x,
  009223372036854775809 AS x2;
```

Both are returned as approximately `9.22337e+18`, making `2^63` and `2^63 + 1` indistinguishable.

`010 -> 8` is included as a lexer-path control. It shows that a leading zero enters an octal-compatible path. When the literal contains `8` or `9`, however, Memgraph does not reject the invalid octal form and instead silently accepts it through a floating-point path.

Current Cypher uses the explicit `0o` prefix for octal literals. Regardless of whether Memgraph intentionally retains legacy `0`-prefixed octal syntax, a digit-only literal such as `09` should not be silently reclassified as DOUBLE.

Memgraph does not support the explicit octal form:

```cypher
RETURN 0o10;
-- parse error
```

The `0[89]...` to DOUBLE fallback is not documented.

### Steps to reproduce

Run the following queries in `mgconsole`:

```cypher
RETURN 010 AS x;

RETURN 9 / 2 AS x, 09 / 2 AS z;

RETURN 9223372036854775808 AS x;

RETURN 009223372036854775808 AS x;

RETURN
  009223372036854775808 AS x,
  009223372036854775809 AS x2;

RETURN 0177 / 2 AS x, 018 / 5 AS x2;
```

### Expected behavior

Digit-only literals containing no decimal point or exponent should not silently become DOUBLE values because they contain a leading zero.

Literals such as `09` should either:

1. be rejected as invalid numeric literals; or
2. be parsed according to a documented integer-literal rule and subjected to normal INT64 range validation.

In particular, adding a leading zero must not allow an otherwise rejected out-of-range integer value to bypass INT64 validation and become an imprecise DOUBLE.

Neo4j rejects `010` and `09` as invalid numeric literals. ArcadeDB also rejects them during parsing.

### Actual behavior

```text
> RETURN 010 AS x;
+----+
| x  |
+----+
| 8  |
+----+
```

```text
> RETURN 9 / 2 AS x, 09 / 2 AS z;
+-----+-----+
| x   | z   |
+-----+-----+
| 4   | 4.5 |
+-----+-----+
```

```text
> RETURN 9223372036854775808 AS x;
Failed query: RETURN 9223372036854775808 AS x
Client received query exception: Integer literal exceeds 64 bits.
```

```text
> RETURN 009223372036854775808 AS x;
+-------------+
| x           |
+-------------+
| 9.22337e+18 |
+-------------+
```

```text
> RETURN 009223372036854775808 AS x,
         009223372036854775809 AS x2;
+-------------+-------------+
| x           | x2          |
+-------------+-------------+
| 9.22337e+18 | 9.22337e+18 |
+-------------+-------------+
```

The `09 / 2` versus `9 / 2` result is a behavioral type witness:

* INTEGER `9 / 2` returns `4`.
* DOUBLE `9.0 / 2` returns `4.5`.

Because `mgconsole` may display both `9` and `9.0` similarly, the division result provides the observable type distinction.

### Controls

```cypher
RETURN 9 / 2 AS x;
-- 4: canonical INTEGER division
```

```cypher
RETURN 10 / 2 AS x;
-- 5: canonical INTEGER division
```

```cypher
RETURN 9 + 0.1 AS x;
-- 9.1: explicit floating-point promotion
```

```cypher
RETURN 0177 / 2 AS x;
-- 63: octal-compatible INTEGER path, 0177 = 127
```

```cypher
RETURN 018 / 5 AS x;
-- 3.6: literal containing 8 is silently routed to DOUBLE
```

### Cross-engine comparison

| Query           | Memgraph 3.12 |       Neo4j 5.26      |     Neo4j 2026.06     | ArcadeDB 26.7.2 |    FalkorDB 4.18.11   |
| --------------- | :-----------: | :-------------------: | :-------------------: | :-------------: | :-------------------: |
| `RETURN 9 / 2`  |       4       |           4           |           4           |        4        |           4           |
| `RETURN 09 / 2` |    **4.5**    | Invalid literal error | Invalid literal error |  Parsing error  | Invalid numeric value |
| `RETURN 010`    |     **8**     | Invalid literal error | Invalid literal error |  Parsing error  |           8           |
| `RETURN 0o10`   |  Parse error  |           8           |           8           |        8        |      Parse error      |

FalkorDB also accepts the legacy `0`-prefixed octal form, so `010 -> 8` alone is not the core defect.

The Memgraph-specific failures are:

1. literals matching the `0[89]...` shape are silently reclassified as DOUBLE; and
2. this reclassification bypasses INT64 range validation and causes precision loss.

### Isolated recheck

Tested on a fresh `memgraph/memgraph:3.12.0` container with no prior state.

The full witness set was executed three consecutive times on 2026-08-05:

```text
010                         -> 8
9 / 2                       -> 4
09 / 2                      -> 4.5
9223372036854775808          -> Integer literal exceeds 64 bits
009223372036854775808        -> 9.22337e+18
009223372036854775809        -> 9.22337e+18
0177 / 2                    -> 63
018 / 5                     -> 3.6
```

Reproduction rate: 3/3.

The same behavior was reproduced on Memgraph `3.11.0`.

### Why this looks like a bug

1. **Range-validation bypass:** the canonical out-of-range integer literal is rejected, while adding a leading zero silently routes the token to DOUBLE and bypasses the INT64 check.
2. **Silent precision loss:** the distinct values `2^63` and `2^63 + 1` become indistinguishable.
3. **Type inconsistency:** `9` behaves as INTEGER, while `09` behaves as DOUBLE despite containing no decimal point or exponent.
4. **Undocumented fallback:** even if legacy leading-zero octal syntax is intentional, an invalid octal digit should not silently trigger floating-point parsing.
5. **Cross-engine evidence:** Neo4j, ArcadeDB, and FalkorDB reject `09`; none of the comparison engines silently accepts it as DOUBLE.

### Impact

* Integer-looking literals such as `09` and `018` silently acquire a floating-point type.
* Arithmetic semantics change: `09 / 2` performs floating-point division, while `9 / 2` performs integer division.
* Leading zeros allow out-of-range integer values to bypass INT64 validation.
* Large distinct values can collapse to the same DOUBLE representation, silently losing data and changing comparisons, calculations, and stored values.

Ignition:
A run of digits is no longer matched as a real literal, leaving it to the
integer matchers. Literals are lexed both by the grammar and by the query
stripper, and the stripper's real matcher accepted a bare run of digits, which
under longest match beat the decimal matcher whenever a leading zero cut that
one short. A value written with a leading zero therefore became a double,
taking part in floating point arithmetic, and skipped the range check an
integer literal gets, so a number too large for an integer was kept as an
approximation.

Recording where the last complete real ended keeps a trailing partial exponent
out of the match, as the previous back-off did by stepping over it.

### DIFF
--- a/src/query/frontend/stripped.cpp
+++ b/src/query/frontend/stripped.cpp
@@ -457,12 +457,11 @@ int StrippedQuery::MatchReal(int start) const {
   enum class State { START, BEFORE_DOT, DOT, AFTER_DOT, E, E_MINUS, AFTER_E };
   State state = State::START;
   auto i = start;
-  // A real literal needs a fractional part or an exponent. A bare run of digits
-  // is an integer literal, and matching it here too would hand it to the real
-  // parser under the longest-match rule: the decimal-integer matcher stops
-  // after a leading zero, so `09` would be longer as a real than as an integer.
-  // Recording where the last complete real ended also drops a trailing partial
-  // exponent, as in `1.5e`.
+  // A real literal needs a fractional part or an exponent, so tracking where
+  // the last complete one ended keeps a bare run of digits out of the real
+  // parser. Longest match would otherwise hand it over: the decimal-integer
+  // matcher stops at a leading zero, making `09` longer as a real than as an
+  // integer. It also drops a partial trailing exponent, as in `1.5e`.
   auto end_of_real = start;
   while (i < static_cast<int>(original_.size())) {
     if (original_[i] == '.') {
--- a/src/query/frontend/stripped.cpp
+++ b/src/query/frontend/stripped.cpp
@@ -457,6 +457,13 @@ int StrippedQuery::MatchReal(int start) const {
   enum class State { START, BEFORE_DOT, DOT, AFTER_DOT, E, E_MINUS, AFTER_E };
   State state = State::START;
   auto i = start;
+  // A real literal needs a fractional part or an exponent. A bare run of digits
+  // is an integer literal, and matching it here too would hand it to the real
+  // parser under the longest-match rule: the decimal-integer matcher stops
+  // after a leading zero, so `09` would be longer as a real than as an integer.
+  // Recording where the last complete real ended also drops a trailing partial
+  // exponent, as in `1.5e`.
+  auto end_of_real = start;
   while (i < static_cast<int>(original_.size())) {
     if (original_[i] == '.') {
       if (state != State::BEFORE_DOT && state != State::START) break;
@@ -479,11 +486,9 @@ int StrippedQuery::MatchReal(int start) const {
       break;
     }
     ++i;
+    if (state == State::AFTER_DOT || state == State::AFTER_E) end_of_real = i;
   }
-  if (state == State::DOT) --i;
-  if (state == State::E) --i;
-  if (state == State::E_MINUS) i -= 2;
-  return i - start;
+  return end_of_real - start;
 }
 
 int StrippedQuery::MatchParameter(int start) const {
--- a/tests/unit/cypher_main_visitor.cpp
+++ b/tests/unit/cypher_main_visitor.cpp
@@ -1282,6 +1282,67 @@ TEST_P(CypherMainVisitorTest, StringLiteralEscapedUtf32) {
   CheckRWType(query, kRead);
 }
 
+TEST_P(CypherMainVisitorTest, NumericLiteralForms) {
+  // Literals are lexed twice, once by the grammar and once by the query
+  // stripper, and this suite runs every case through both. Each form below is
+  // therefore an assertion that the two agree as well as that the value is
+  // right.
+  auto &ast_generator = *GetParam();
+
+  auto literal_of = [&ast_generator](const std::string &query_string) {
+    auto *query = dynamic_cast<CypherQuery *>(ast_generator.ParseQuery(query_string));
+    EXPECT_TRUE(query) << query_string;
+    auto *return_clause = dynamic_cast<Return *>(query->single_query_->clauses_[0]);
+    return return_clause->body_.named_expressions[0]->expression_;
+  };
+
+  // Integers, in each base the grammar accepts.
+  ast_generator.CheckLiteral(literal_of("RETURN 0"), 0, 1);
+  ast_generator.CheckLiteral(literal_of("RETURN 42"), 42, 1);
+  ast_generator.CheckLiteral(literal_of("RETURN 9223372036854775807"), 9223372036854775807L, 1);
+  ast_generator.CheckLiteral(literal_of("RETURN 010"), 8, 1);
+  ast_generator.CheckLiteral(literal_of("RETURN 0177"), 127, 1);
+  ast_generator.CheckLiteral(literal_of("RETURN 0x1f"), 31, 1);
+  ast_generator.CheckLiteral(literal_of("RETURN 0xFF"), 255, 1);
+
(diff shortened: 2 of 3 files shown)


Ignition:
### Checked: has any class of valid parsing broken

Worth being explicit about the safety net, because a change to a lexer is the kind that breaks
something quietly.

**The parser suite already compares both lexers.** `CypherMainVisitorTest` is parameterised over
four AST generators, one of which is `CachedAstGenerator` — it runs the query through
`StrippedQuery` and parses the *stripped* text with the literals plugged back in, exactly as the
interpreter does when caching. So every one of the 1284 cases asserts that the hand-written matcher
and the grammar agree on the literals it contains, not merely that the query parses.

**Added a numeric corpus to that harness** (`NumericLiteralForms`, `NumbersAdjacentToDots`), so the
forms are stated rather than incidental:

- integers in each base the grammar accepts: `0`, `42`, `9223372036854775807`, `010` (8), `0177`
  (127), `0x1f`, `0xFF`
- every arrangement of fraction and exponent: `3.5`, `0.5`, `.5`, `1e5`, `1E5`, `1e-5`, `1E-5`,
  `1.5e3`, `1.5e-3`, `0.1e-2`, `.1e-2`
- numbers with a dot immediately after them, where the change is most likely to bite: list slicing
  `[1..2]`, variable-length patterns `[*1..2]`, `[*2..]`, `[*..3]`, `[*2]`, and reals inside a map

All pass under all four generators, including the stripped one.

The forms most at risk were the ones the removed back-off existed for, and they are pinned in
`tests/unit/stripped.cpp` as well: `1.5e` and `1.5e-` are the real `1.5` followed by something
else, `42.foo` is the integer `42` and a dot, `.5` still begins a real.

Suites: `cypher_main_visitor` 1284, `stripped` 82.

Ignition:
Reported as #4521.

Ignition:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - A number starting with 0 is octal, so 09 and 018 are now parse errors instead of 9.0 and 18.0; drop the leading zero, though 010 still means 8. [#4588](https://github.com/memgraph/memgraph/pull/4588)

colinbarry (approved):
Approved 👍

PR #4589, opened 2026-08-15
fix(query): convert whole-number strings to integers exactly

### LINKED ISSUE #4166: toInteger(large positive float) may wrap to INT64_MIN instead of preserving sign
**Memgraph version**
Observed on:

- Memgraph `3.9.0`
- Memgraph `3.10.1`

Docker images used locally:
- `memgraph/memgraph:3.9.0`
- `memgraph/memgraph:3.10.1` 

**Environment**
- Host OS: Windows 10
- Architecture: x86_64
- Deployment: Docker
- Connection method: Bolt
- Client used for direct reproduction: Python `neo4j` Bolt driver

**Describe the bug**
Memgraph may convert a large **positive** floating-point value to the most negative 64-bit integer when `toInteger(...)` is applied near or above the signed 64-bit upper bound.

In the minimized repro below, Neo4j returns the maximum signed 64-bit integer:

```text
9223372036854775807
```

Memgraph instead returns:

```text
-9223372036854775808
```

This looks like an overflow wraparound, because the sign flips even though the input value is positive.

**To Reproduce**
2. Run:

```cypher
RETURN toInteger(9223372036854775808.0) AS x;
```

**Expected behavior**
Observed Neo4j result:

```text
x = 9223372036854775807
```

At minimum, the result should not become negative for a positive input.

**Actual behavior**
Observed Memgraph result:

```text
x = -9223372036854775808
```

So Memgraph wraps the positive input to `INT64_MIN`.

**Control query 1**
At the negative boundary, both engines agree:

```cypher
RETURN toInteger(-9223372036854775808.0) AS x;
```

```text
x = -9223372036854775808
```

So the issue is not simply “large magnitude floats are unsupported”.

**Control query 2**
The same positive-wrap behavior also appears for nearby large positive values:

```cypher
RETURN toInteger(9223372036854775807.0) AS x;
RETURN toInteger(9223372036854775806.0) AS x;
```

Observed Neo4j result in both cases:

```text
x = 9223372036854775807
```

Observed Memgraph result in both cases:

```text
x = -9223372036854775808
```

This suggests the issue is not limited to a single exact threshold literal.

### LINKED ISSUE #4391: `toInteger()` maps positive overflow, Infinity, and NaN to `INT64_MIN`
### Memgraph version

Observed on the following local Docker image:

```text
docker images memgraph/memgraph --format "{{.ID}} {{.CreatedAt}}"

docker exec mg_verify mgconsole --version
Memgraph 3.11.0

Interface: Bolt, accessed via docker exec -i mg_verify mgconsole
```

Also reproduced in a fresh isolated Docker container:

```text
container: mg_verify
image: memgraph/memgraph:latest
pulled: 2026-07-11
version: 3.11.0
reproduction rate: 3/3
```

### Environment

* Host OS: Linux x86_64
* Deployment: Docker
* Query language: Cypher
* Reference engine: Neo4j Community `2026.05.0`

### Description

When `toInteger()` receives a floating-point value outside the signed 64-bit integer range, Memgraph returns:

```text
-9223372036854775808
```

This is `INT64_MIN`.

The same result is returned for:

* positive values above `INT64_MAX`
* positive infinity
* NaN
* negative values below `INT64_MIN`

This produces an incorrect sign for positive overflow. For example, converting a large positive floating-point value returns the minimum negative integer.

Neo4j `2026.05.0` instead returns:

```text
positive overflow: INT64_MAX
positive infinity: INT64_MAX
negative overflow: INT64_MIN
NaN: 0
```

Normal finite values within range are converted correctly by both engines.

### To reproduce

No graph data is required. Run each query separately.

#### Positive value just above the int64 range

```cypher
RETURN toInteger(9.223372036854776e18) AS result;
```

Memgraph returns:

```text
-9223372036854775808
```

Expected according to Neo4j:

```text
9223372036854775807
```

#### Larger positive value

```cypher
RETURN toInteger(1e19) AS result;
```

Memgraph returns:

```text
-9223372036854775808
```

Expected according to Neo4j:

```text
9223372036854775807
```

#### Positive infinity

```cypher
RETURN toInteger(1.0 / 0.0) AS result;
```

Memgraph returns:

```text
-9223372036854775808
```

Expected according to Neo4j:

```text
9223372036854775807
```

#### NaN

```cypher
RETURN toInteger(0.0 / 0.0) AS result;
```

Memgraph returns:

```text
-9223372036854775808
```

Neo4j returns:

```text
0
```

#### Negative value below the int64 range

```cypher
RETURN toInteger(-1e19) AS result;
```

Both Memgraph and Neo4j return:

```text
-9223372036854775808
```

### Expected behavior

Positive and negative out-of-range values should not produce the same result.

Following Neo4j behavior, conversion should be bounded according to the direction of overflow:

```text
large positive finite value: 9223372036854775807
positive infinity:           9223372036854775807
large negative finite value: -9223372036854775808
negative infinity:           -9223372036854775808
NaN:                         0
```

At minimum, a positive out-of-range value must not silently become `INT64_MIN`.

### Actual behavior

Memgraph returns `INT64_MIN` for all tested exceptional floating-point inputs:

```text
toInteger(9.223372036854776e18) = -9223372036854775808
toInteger(1e19)                 = -9223372036854775808
toInteger(1.0 / 0.0)            = -9223372036854775808
toInteger(0.0 / 0.0)            = -9223372036854775808
toInteger(-1e19)                = -9223372036854775808
```

The positive overflow and non-finite cases are therefore indistinguishable from a valid conversion to `INT64_MIN`.

### Control queries

Normal finite values are converted correctly:

```cypher
RETURN toInteger(3.14) AS result;
```

Result:

```text
3
```

```cypher
RETURN toInteger(-0.5) AS result;
```

Result:

```text
0
```

```cypher
RETURN toInteger(-3.14) AS result;
```

Result:

```text
-3
```

### Cross-engine comparison

| Expression                        | Memgraph `3.11.0` | Neo4j `2026.05.0` |
| --------------------------------- | ----------------: | ----------------: |
| `toInteger(9.223372036854776e18)` |       `INT64_MIN` |       `INT64_MAX` |
| `toInteger(1e19)`                 |       `INT64_MIN` |       `INT64_MAX` |
| `toInteger(1.0 / 0.0)`            |       `INT64_MIN` |       `INT64_MAX` |
| `toInteger(0.0 / 0.0)`            |       `INT64_MIN` |               `0` |
| `toInteger(-1e19)`                |       `INT64_MIN` |       `INT64_MIN` |
| `toInteger(3.14)`                 |               `3` |               `3` |
| `toInteger(-0.5)`                 |               `0` |               `0` |

### Isolated recheck

The behavior was reproduced in a fresh isolated Docker container:

```text
container: mg_verify
image: memgraph/memgraph:latest
pulled: 2026-07-11
version: 3.11.0

Run 1: toInteger(1e19) = -9223372036854775808
Run 2: toInteger(1e19) = -9223372036854775808
Run 3: toInteger(1e19) = -9223372036854775808

Reproduction rate: 3/3
```

Additional results in the same container:

```text
toInteger(9.223372036854776e18) = -9223372036854775808
toInteger(1.0 / 0.0)            = -9223372036854775808
toInteger(0.0 / 0.0)            = -9223372036854775808
toInteger(-1e19)                = -9223372036854775808
toInteger(3.14)                 = 3
```

### Why this appears to be a bug

A large positive floating-point value and a large negative floating-point value should not both convert to the minimum negative integer.

For example:

```cypher
RETURN toInteger(1e19);
```

returns:

```text
-9223372036854775808
```

This changes the sign of the input and silently produces a valid-looking integer value.

The result is especially problematic because `INT64_MIN` may also be produced by a legitimate in-range conversion. Applications therefore cannot distinguish a valid minimum integer from a failed positive overflow conversion.

The consistent `INT64_MIN` result for positive overflow, infinity, NaN, and negative overflow suggests that the floating-point to integer conversion path does not handle range and non-finite cases separately.

This is a semantic correctness issue because positive out-of-range values are silently converted into incorrect negative results.

### LINKED ISSUE #4431: `toInteger()` loses precision for large decimal strings and maps `LONG_MAX` to `LONG_MIN`
### Memgraph version

Memgraph `3.12.0`

### Environment
- Host OS: Linux x86_64
- Deployment: Docker
- Query interface: `mgconsole` via `docker exec`
- Query language: Cypher
- Reference engine: Neo4j Community `2026.05.0`

### Description
`toInteger()` produces incorrect results for decimal string inputs near and beyond 2^53:

1. Precision loss: decimal strings representing integers larger than 2^53 (~9e15) are returned with incorrect low-order digits
2. Valid `LONG_MAX` (`9223372036854775807`) is returned as `LONG_MIN` (`-9223372036854775808`)
3. Out-of-range decimal strings are silently mapped to `LONG_MIN` instead of raising an error or returning `null`

The observed results are consistent with the decimal string being converted through a floating-point representation before being cast to an integer.

Related: #4391 covers overflow for numeric/float inputs. This report focuses specifically on decimal string inputs, including precision loss for valid int64 values and incorrect handling of `LONG_MAX`.

### To reproduce
```cypher
// 2^53 boundary — exact up to this point
RETURN toInteger('9007199254740992') AS exact_boundary;

// First value that cannot be represented exactly in float64
RETURN toInteger('9007199254740993') AS first_inexact;

// Precision loss (off by 216)
RETURN toInteger('9223372036854775000') AS r;

// Valid LONG_MAX returned as LONG_MIN
RETURN toInteger('9223372036854775807') AS r;

// Out-of-range decimal string silently mapped to LONG_MIN
RETURN toInteger('9999999999999999999') AS r;
```

### Expected behavior
- `toInteger('9007199254740992')` should return `9007199254740992` (exact — within range)
- `toInteger('9007199254740993')` should return `9007199254740993` (exact)
- `toInteger('9223372036854775000')` should return `9223372036854775000` (exact)
- `toInteger('9223372036854775807')` should return `9223372036854775807` (`LONG_MAX`)
- Out-of-range decimal strings should produce an error or `null`; they must not silently return an unrelated valid integer value

Neo4j 2026.05 returns exact correct values for all valid int64 decimal strings and raises a type error on overflow.

### Actual behavior
```
// 2^53 boundary
RETURN toInteger('9007199254740992') AS exact_boundary;
// Result: 9007199254740992 (correct)

RETURN toInteger('9007199254740993') AS first_inexact;
// Result: 9007199254740992 (off by 1)

// Precision loss
RETURN toInteger('9223372036854775000') AS r;
// Result: 9223372036854774784 (off by 216)

// Valid LONG_MAX returned as LONG_MIN
RETURN toInteger('9223372036854775807') AS r;
// Result: -9223372036854775808

// Out-of-range mapped to LONG_MIN
RETURN toInteger('9999999999999999999') AS r;
// Result: -9223372036854775808
```

### Controls (work correctly)
```cypher
RETURN toInteger('42') AS r;    // 42
RETURN toInteger('-100') AS r;  // -100
RETURN toInteger('0') AS r;     // 0
RETURN toInteger(3.14) AS r;    // 3
```

### Cross-engine comparison
| Query | Memgraph 3.12 | Neo4j 2026.05 | ArcadeDB 26.8.1 |
|-------|:---:|:---:|:---:|
| `toInteger('9007199254740993')` | **9007199254740992** | 9007199254740993 | 9007199254740992 |
| `toInteger('9223372036854775000')` | **9223372036854774784** | 9223372036854775000 | 9223372036854774784 |
| `toInteger('9223372036854775807')` | **-9223372036854775808** | 9223372036854775807 | 9223372036854775807 |
| `toInteger('9999999999999999999')` | **-9223372036854775808** | ERROR | null |
| `toInteger('42')` | 42 | 42 | 42 |

### Isolated recheck
Fresh Docker container `memgraph/memgraph:latest` (v3.12.0), 3 runs, 2026-07-16. Reproduction rate: 3/3 (deterministic).

```
Run 1: 2^53_boundary=correct, first_inexact=off_by_1, precision_loss=off_by_216, LONG_MAX=LONG_MIN, overflow=LONG_MIN
Run 2: 2^53_boundary=correct, first_inexact=off_by_1, precision_loss=off_by_216, LONG_MAX=LONG_MIN, overflow=LONG_MIN
Run 3: 2^53_boundary=correct, first_inexact=off_by_1, precision_loss=off_by_216, LONG_MAX=LONG_MIN, overflow=LONG_MIN
```

### Why this looks like a bug
1. `toInteger('9223372036854775807')` is exactly `LONG_MAX` — a perfectly valid int64 value representable without any precision issues. Returning `LONG_MIN` for it is an unambiguous bug.
2. Out-of-range decimal strings silently return a valid-looking integer (`LONG_MIN`) instead of signaling an error, violating the principle that invalid input should not silently succeed with an unrelated value.
3. The precision loss pattern (off by 1 at 2^53+1, off by 216 near LONG_MAX) shows that integer parsing is not handling the full int64 range correctly, even for values well within the valid range.

Ignition:
Whole-number strings now convert to integers directly rather than through a
double. A double carries fewer significant digits than an integer, so a value
near the limits rounded onto a neighbouring integer: the largest came back as
the smallest, and neighbouring inputs collapsed onto one value.

Text naming no number converts to null, while text naming a number outside the
range is an error quoting it, since returning some other number would be worse
than saying so. A floating point argument saturates at the nearest end instead,
and NaN converts to zero. The range is checked before every conversion, so none
of them is undefined.

The bounds are written out rather than taken from boost's range checker, which
reports the exact lowest value as an underflow even though it converts, and
reports NaN as in range. Both would need working around here, where every input
has to yield an answer.

### DIFF
--- a/src/query/interpret/awesome_memgraph_functions.cpp
+++ b/src/query/interpret/awesome_memgraph_functions.cpp
@@ -684,22 +684,23 @@ TypedValue ToFloat(const TypedValue *args, int64_t nargs, const FunctionContext 
   }
 }
 
+// int64's maximum is not representable as a double, so the bound is the
+// smallest double above the range, 2^63, and it is exclusive.
+constexpr double kInt64UpperBoundExclusive = 9223372036854775808.0;
+constexpr auto kInt64LowerBoundInclusive = static_cast<double>(std::numeric_limits<int64_t>::min());
+
+// False for NaN, which compares false against both bounds.
+bool IsWithinInt64Range(const double value) {
+  return value >= kInt64LowerBoundInclusive && value < kInt64UpperBoundExclusive;
+}
+
 // Truncates toward zero, saturating at either end of the range and taking NaN
 // to zero. This is how a floating point argument converts: the C++ cast is
 // undefined outside the range, so the bounds are applied before it.
-//
-// The bounds are spelled out rather than taken from boost's range checker,
-// which reports the exact lowest value as an underflow even though it converts,
-// and reports NaN as in range. Both would need working around here, where every
-// input has to yield a number.
-int64_t TruncateToInteger(double value) {
-  // int64's maximum is not representable as a double; the smallest double at or
-  // above the range is 2^63, which is.
-  constexpr double kExclusiveUpperBound = 9223372036854775808.0;
-  constexpr auto kInclusiveLowerBound = static_cast<double>(std::numeric_limits<int64_t>::min());
+int64_t TruncateToInteger(const double value) {
   if (std::isnan(value)) return 0;
-  if (value >= kExclusiveUpperBound) return std::numeric_limits<int64_t>::max();
-  if (value < kInclusiveLowerBound) return std::numeric_limits<int64_t>::min();
+  if (value >= kInt64UpperBoundExclusive) return std::numeric_limits<int64_t>::max();
+  if (value < kInt64LowerBoundInclusive) return std::numeric_limits<int64_t>::min();
   return static_cast<int64_t>(value);
 }
 
@@ -732,11 +733,7 @@ std::pair<StringToInteger, int64_t> ParseInteger(std::string_view text) {
   } catch (const utils::BasicException &) {
     return {StringToInteger::kNotANumber, 0};
   }
-  constexpr double kExclusiveUpperBound = 9223372036854775808.0;
-  constexpr auto kInclusiveLowerBound = static_cast<double>(std::numeric_limits<int64_t>::min());
-  if (std::isnan(as_double) || as_double >= kExclusiveUpperBound || as_double < kInclusiveLowerBound) {
-    return {StringToInteger::kOutOfRange, 0};
-  }
+  if (!IsWithinInt64Range(as_double)) return {StringToInteger::kOutOfRange, 0};
   return {StringToInteger::kOk, static_cast<int64_t>(as_double)};
 }
 
--- a/src/query/interpret/awesome_memgraph_functions.cpp
+++ b/src/query/interpret/awesome_memgraph_functions.cpp
@@ -684,16 +684,53 @@ TypedValue ToFloat(const TypedValue *args, int64_t nargs, const FunctionContext 
   }
 }
 
-// Truncates toward zero, or reports that no int64 corresponds to the value.
-// Converting out-of-range doubles is undefined behaviour, so the range has to
-// be checked before the cast rather than after.
-std::optional<int64_t> TruncateToInteger(double value) {
+// Truncates toward zero, saturating at either end of the range and taking NaN
+// to zero. This is how a floating point argument converts: the C++ cast is
+// undefined outside the range, so the bounds are applied before it.
+int64_t TruncateToInteger(double value) {
   // int64's maximum is not representable as a double; the smallest double at or
   // above the range is 2^63, which is.
   constexpr double kExclusiveUpperBound = 9223372036854775808.0;
   constexpr auto kInclusiveLowerBound = static_cast<double>(std::numeric_limits<int64_t>::min());
-  if (std::isnan(value) || value < kInclusiveLowerBound || value >= kExclusiveUpperBound) return std::nullopt;
+  if (std::isnan(value)) return 0;
+  if (value >= kExclusiveUpperBound) return std::numeric_limits<int64_t>::max();
+  if (value < kInclusiveLowerBound) return std::numeric_limits<int64_t>::min();
   return static_cast<int64_t>(value);
+}
+
+// A string can fail to yield an integer in two ways that are reported
+// differently: one naming no number at all is null, while one naming a number
+// too large for the type is an error in toInteger and null in toIntegerOrNull.
+enum class StringToInteger : std::uint8_t { kOk, kNotANumber, kOutOfRange };
+
+std::pair<StringToInteger, int64_t> ParseInteger(std::string_view text) {
+  const auto trimmed = utils::Trim(text);
+  // A whole-number string is parsed as an integer directly. A double holds
+  // fewer significant digits than int64, so going through one would round large
+  // values onto a neighbouring integer or off the end of the range.
+  int64_t parsed{};
+  const auto *const begin = trimmed.data();
+  const auto *const end = begin + trimmed.size();
+  if (const auto [stopped_at, ec] = std::from_chars(begin, end, parsed); stopped_at == end) {
+    if (ec == std::errc{}) return {StringToInteger::kOk, parsed};
+    return {StringToInteger::kOutOfRange, 0};
+  }
+
+  // Anything else is only meaningful as a floating point number, and is
+  // truncated toward zero. Unlike a floating point argument it does not
+  // saturate: the text named an exact value, and no integer stands for it.
+  double as_double{};
+  try {
+    as_double = utils::ParseDouble(trimmed);
+  } catch (const utils::BasicException &) {
+    return {StringToInteger::kNotANumber, 0};
+  }
+  constexpr double kExclusiveUpperBound = 9223372036854775808.0;
+  constexpr auto kInclusiveLowerBound = static_cast<double>(std::numeric_limits<int64_t>::min());
+  if (std::isnan(as_double) || as_double >= kExclusiveUpperBound || as_double < kInclusiveLowerBound) {
+    return {StringToInteger::kOutOfRange, 0};
+  }
+  return {StringToInteger::kOk, static_cast<int64_t>(as_double)};
 }
 
 TypedValue ToInteger(const TypedValue *args, int64_t nargs, const FunctionContext &ctx) {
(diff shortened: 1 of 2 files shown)


Ignition:
### Why `TruncateToInteger` is written out rather than delegated

Fair question, and the codebase does already have an idiom for this —
`boost::numeric::converter<int64_t, double>` in `plan_v2/cost/builtin_estimator.cpp`. I tried it
here and it does not fit this use, for two measured reasons:

| input | `boost::numeric::converter<int64_t, double>::out_of_range` | truth |
| --- | --- | --- |
| `NaN` | `cInRange` | not convertible; casting it is undefined |
| exactly `int64_min` as a double | `cNegOverflow` | representable, and converts |
| `2^63` | `cPosOverflow` | correct |
| `1e30`, `-1e30`, infinities | correct | correct |

The NaN classification means the cast would still be undefined without a separate guard. The
lower-bound one matters more: `-9223372036854775808` is exactly representable, so the string naming
it should convert rather than raise, and the range checker would have it raise. Working around both
is more code than the bounds check, and subtler.

The existing use gets away with it because it only *declines to optimise* when the checker says out
of range; here every input has to yield a number or a defined error.

For completeness on the standard library: `std::saturate_cast` is C++26 and integer-to-integer
only, so it does not cover float to integer at all. There is no standard facility for this.

Both boundaries are now asserted, including the one boost would have got wrong, and the reasoning
is recorded next to the function.

Ignition:
Reported as #4431.

Ignition:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - toInteger('9223372036854775807') no longer returns the smallest number, and text too big to fit now errors rather than giving a wrong value; use toIntegerOrNull() for null instead of an error. [#4589](https://github.com/memgraph/memgraph/pull/4589)

imilinovic (approved):

PR #4590, opened 2026-08-15
fix(query): return null from single() when a predicate is null

### LINKED ISSUE #4430: `single()` returns `true` when one predicate result is `true` and another is `null`
### Memgraph version

Memgraph `3.12.0`

### Environment

* Host OS: Linux x86_64
* Deployment: Docker
* Query language: Cypher
* Reference engines:

  * Neo4j Community `2026.06.0`
  * FalkorDB `4.20.1`

### Description

When exactly one predicate evaluation is `true` and another is `null`, Memgraph's `single()` function returns `true`.

The result should be `null`, because the element whose predicate evaluates to `null` may or may not satisfy the predicate. It is therefore unknown whether exactly one element matches.

Neo4j and FalkorDB both return `null` for the same expressions.

### Steps to reproduce

No graph data is required.

```cypher
RETURN single(x IN [null, 1] WHERE x > 0) AS result;
```

Reversing the element order produces the same result:

```cypher
RETURN single(x IN [1, null] WHERE x > 0) AS result;
```

### Expected behavior

Both queries should return:

```text
null
```

One element definitely satisfies the predicate, while the result for the other element is unknown. Therefore, it cannot be determined whether exactly one element satisfies the predicate.

### Actual behavior

Memgraph returns:

```text
true
```

The result suggests that Memgraph treats the `null` predicate result as a definite non-match when exactly one other element evaluates to `true`.

### Controls

The following cases work correctly:

```cypher
// Exactly one match
RETURN single(x IN [1] WHERE x > 0) AS result;
// true

// No matches
RETURN single(x IN [0] WHERE x > 0) AS result;
// false

// More than one match
RETURN single(x IN [1, 2] WHERE x > 0) AS result;
// false

// Empty list
RETURN single(x IN [] WHERE x > 0) AS result;
// false

// Predicate explicitly matches null
RETURN single(x IN [null] WHERE x IS NULL) AS result;
// true

// Only an unknown predicate result
RETURN single(x IN [null] WHERE x > 0) AS result;
// null

// More than one definite match, plus null
RETURN single(x IN [null, 1, 2] WHERE x > 0) AS result;
// false
```

### Cross-engine comparison

| Expression                              | Memgraph 3.12.0 | Neo4j 2026.06.0 | FalkorDB 4.20.1 |
| --------------------------------------- | :-------------: | :-------------: | :-------------: |
| `single(x IN [null, 1] WHERE x > 0)`    |    **`true`**   |      `null`     |      `null`     |
| `single(x IN [1, null] WHERE x > 0)`    |    **`true`**   |      `null`     |      `null`     |
| `single(x IN [null] WHERE x > 0)`       |      `null`     |      `null`     |      `null`     |
| `single(x IN [null, 1, 2] WHERE x > 0)` |     `false`     |     `false`     |     `false`     |
| `single(x IN [1] WHERE x > 0)`          |      `true`     |      `true`     |      `true`     |

### Isolated reproduction

The behavior was reproduced in a fresh Docker container running Memgraph `3.12.0`:

```text
Run 1:
single(x IN [null, 1] WHERE x > 0) -> true
single(x IN [null] WHERE x > 0)    -> null

Run 2:
single(x IN [null, 1] WHERE x > 0) -> true

Run 3:
single(x IN [null, 1] WHERE x > 0) -> true
```

Reproduction rate: `3/3`.

### Why this looks like a bug

Under three-valued logic, a predicate result of `null` is unknown rather than false.

For `single()`:

* zero definite matches and at least one unknown result should produce `null`;
* one definite match and at least one unknown result should also produce `null`;
* two or more definite matches should produce `false`, regardless of unknown results.

Memgraph handles the first and third cases correctly, but incorrectly returns `true` for the second case.

### Impact

Queries using `single()` on lists containing null or otherwise indeterminate values may return a definite `true` result when the correct result is unknown.

This can affect validation predicates, filtering logic, and queries that rely on exactly-one matching semantics.

Ignition:
single() now returns null whenever the predicate evaluates to null for any
element. One element matching for certain alongside one that may match means the
count is either one or two, so the previous answer of true claimed a certainty
it did not have.

Two or more definite matches still return false, that being the only case where
the null elements cannot change the result whichever way they resolve.

### DIFF
--- a/src/query/interpret/eval.hpp
+++ b/src/query/interpret/eval.hpp
@@ -1006,8 +1006,8 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
     if (non_empty_list && !has_value) {
       return TypedValue(ctx_->memory);
     } else if (has_null_elements) {
-      // An unknown could be the second match, so with fewer than two definite
-      // matches the count is not settled. Two or more already returned false.
+      // Two matches already returned false above, so at most one definite match
+      // reached here and a null element could have been a second one.
       return TypedValue(ctx_->memory);
     } else {
       return TypedValue(predicate_satisfied, ctx_->memory);
--- a/src/query/interpret/eval.hpp
+++ b/src/query/interpret/eval.hpp
@@ -1005,7 +1005,9 @@ class ExpressionEvaluator : public ExpressionVisitor<TypedValue> {
     }
     if (non_empty_list && !has_value) {
       return TypedValue(ctx_->memory);
-    } else if (has_null_elements && !predicate_satisfied) {
+    } else if (has_null_elements) {
+      // An unknown could be the second match, so with fewer than two definite
+      // matches the count is not settled. Two or more already returned false.
       return TypedValue(ctx_->memory);
     } else {
       return TypedValue(predicate_satisfied, ctx_->memory);
--- a/tests/gql_behave/tests/memgraph_V1/features/functions.feature
+++ b/tests/gql_behave/tests/memgraph_V1/features/functions.feature
@@ -883,7 +883,7 @@ Feature: Functions
             """
         Then the result should be:
             | a    |
-            | true |
+            | null |
 
     Scenario: Single test 06:
         When executing query:
@@ -910,7 +910,7 @@ Feature: Functions
             """
         Then the result should be:
             | a     |
-            | true  |
+            | null  |
 
     Scenario: Single test 09:
         When executing query:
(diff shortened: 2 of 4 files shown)


Ignition:
Reported as #4430.

imilinovic (approved):

Ignition:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - single(x IN [null, 1] WHERE x > 0) now returns null rather than true, so filters over data containing nulls may match different rows; use coalesce(single(...), false) for the old answer. [#4590](https://github.com/memgraph/memgraph/pull/4590)

PR #4592, opened 2026-08-15
fix(query): report an unusable timezone as an argument error

### LINKED ISSUE #4529: Internal error (An unknown exception occurred) when constructing datetime with an offset-string timezone in a map
### Memgraph version

Memgraph 3.11.0 (also reproduced on 3.12.0)

### Environment
- Host OS: Linux x86_64
- Deployment: Docker (`memgraph/memgraph:3.11.0`, fresh container)
- Query interface: Bolt driver via `docker exec`-equivalent session
- Query language: Cypher

### Description

`datetime({... timezone: '+01:00'})` — constructing a datetime with a
numeric UTC offset string as the `timezone` map field — raises an
internal exception (`An unknown exception occurred, this is unexpected.
Real message should be in database logs.`) instead of returning a value
or a user-facing error.

The documented `timezone` map field accepts a named-zone string or an
integer minute offset (both work correctly, see Controls). An offset
string is not documented, but it is a plausible user input (the same
format is accepted by the string constructor `datetime("...T...+01:00")`
and by the reference engine's map constructor). When such a string is
passed, Memgraph raises an internal error rather than a user-facing
argument error. Any unsupported input should produce a user error, never
an internal exception.

### To reproduce

Run on a fresh Memgraph 3.11 container:

```cypher
RETURN datetime({year: 1984, month: 10, day: 11, hour: 12, timezone: '+01:00'}) AS r
```

### Expected behavior

If offset strings are supported for the `timezone` map field, the function
should construct the corresponding zoned datetime.

If this input form is unsupported, Memgraph should reject it with a
user-facing argument or timezone-format error. It must not expose a generic
internal exception stating that the real message is only available in the
database logs.

### Actual behavior

```text
{code: Memgraph.DatabaseError.MemgraphError.MemgraphError}
{message: An unknown exception occurred, this is unexpected. Real message should be in database logs.}
```

### Controls (work correctly)

| Query | Result |
|-------|--------|
| `datetime({... timezone: 60})` (documented int minutes) | `1984-10-11T12:00:00.000000000+01:00` |
| `datetime({... timezone: -330})` (documented int minutes) | `1984-10-11T12:00:00.000000000-05:30` |
| `datetime({... timezone: 0})` | `1984-10-11T12:00:00.000000000+00:00` |
| `datetime({... timezone: 'Europe/Stockholm'})` (documented named string) | `1984-10-11T12:00:00.000000000+01:00` |
| `datetime('1984-10-11T12:00:00+01:00')` (string constructor, offset) | `1984-10-11T12:00:00.000000000+01:00` |

The same input class on the reference engine:

```cypher
RETURN datetime({year: 1984, month: 10, day: 11, hour: 12, timezone: '+01:00'})
-- 1984-10-11T12:00:00+01:00
```

### Cross-engine comparison

| Query | Memgraph 3.11 | Neo4j 2026.06 |
|-------|:---:|:---:|
| `datetime({... timezone: '+01:00'})` | **internal error** | `1984-10-11T12:00:00+01:00` |
| `datetime({... timezone: 'Z'})` | **internal error** | `1984-10-11T12:00:00Z` |
| `datetime({... timezone: '-05:30'})` | **internal error** | `1984-10-11T12:00:00-05:30` |
| `datetime({... timezone: 'Europe/Stockholm'})` | correct | correct |

### Isolated recheck

Fresh `memgraph/memgraph:3.11.0` container, no prior state, 3 runs:
witness 3/3 internal error (stable). Also reproduced on a fresh
`memgraph/memgraph:3.12.0` container (3/3 stable). Controls 3/3 correct
on both versions.

### Why this looks like a bug

1. An internal exception ("Real message should be in database logs") is
   not a user-facing error for an input-value problem; the error-type
   contract is violated.
2. The engine is internally asymmetric: the `timezone` map field
   accepts strings (named zones work), but a different string form of
   the same field raises an internal exception instead of an argument
   error. Users cannot distinguish "supported named string" from
   "unsupported offset string" at the type level.
3. The same offset format is accepted by Memgraph's own string
   constructor and by the reference engine's map constructor, making
   the input a natural user expectation.

Ignition:
A timezone that cannot be understood is now reported as a problem with the
argument. The field was handed to the constructor that looks a name up in the
time zone database, which reports an unknown name by throwing out of the
standard library, so a bad argument reached the caller as an internal error
naming no cause.

A UTC offset is accepted as well, in the forms the string constructor already
takes. The same construction in the module interface reports an invalid
argument now too, which is what its own documentation promises for a parameter
it cannot parse.

### DIFF
--- a/src/utils/temporal.cpp
+++ b/src/utils/temporal.cpp
@@ -778,8 +778,8 @@ Timezone ParseTimezoneFromUserString(std::string_view timezone) {
   try {
     return Timezone{std::chrono::locate_zone(timezone)};
   } catch (const std::runtime_error &) {
-    // The only failure locate_zone reports, and narrow on purpose: catching
-    // everything would report a failed allocation as an unknown zone name.
+    // The only failure locate_zone reports. Catching everything here would
+    // report a failed allocation as an unknown zone name.
     throw temporal::InvalidArgumentException("Timezone name is not in the IANA time zone database.");
   }
 }
--- a/src/utils/temporal.cpp
+++ b/src/utils/temporal.cpp
@@ -756,6 +756,31 @@ std::pair<Timezone, uint64_t> ParseTimezoneFromOffset(std::string_view timezone_
   return {Timezone(compute_offset(sign, maybe_hours.value(), maybe_minutes.value())), timezone_offset_string.length()};
 }
 
+Timezone ParseTimezoneFromUserString(std::string_view timezone) {
+  if (timezone.empty()) {
+    throw temporal::InvalidArgumentException("The timezone must not be empty.");
+  }
+
+  if (timezone == "Z" || timezone == "z") {
+    return Timezone{std::chrono::minutes{0}};
+  }
+
+  if (timezone.front() == '+' || timezone.front() == '-') {
+    const auto [parsed, unconsumed_length] = ParseTimezoneFromOffset(timezone);
+    if (unconsumed_length != 0) {
+      throw temporal::InvalidArgumentException("Extra characters after the timezone offset. {}",
+                                               kSupportedZonedDateTimeFormatsHelpMessage);
+    }
+    return parsed;
+  }
+
+  try {
+    return Timezone{std::chrono::locate_zone(timezone)};
+  } catch (...) {
+    throw temporal::InvalidArgumentException("Timezone name is not in the IANA time zone database.");
+  }
+}
+
 ZonedDateTimeParameters ParseZonedDateTimeParameters(std::string_view string) {
   // https://en.wikipedia.org/wiki/ISO_8601#Time_zone_designators
 
--- a/src/utils/temporal.cpp
+++ b/src/utils/temporal.cpp
@@ -19,6 +19,7 @@
 #include <ctime>
 #include <format>
 #include <limits>
+#include <stdexcept>
 #include <string>
 #include <string_view>
 #include <utility>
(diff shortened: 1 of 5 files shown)


Ignition:
Reported as #4529.

Ignition:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - A bad timezone now says what is wrong instead of "An unknown exception occurred", and offsets such as +01:00 and Z work as well as names. [#4592](https://github.com/memgraph/memgraph/pull/4592)

src/utils/temporal.cpp
```diff
@@ -756,6 +756,31 @@ std::pair<Timezone, uint64_t> ParseTimezoneFromOffset(std::string_view timezone_
...
+    if (unconsumed_length != 0) {
+      throw temporal::InvalidArgumentException("Extra characters after the timezone offset. {}",
+                                               kSupportedZonedDateTimeFormatsHelpMessage);
+    }
+    return parsed;
+  }
+
+  try {
+    return Timezone{std::chrono::locate_zone(timezone)};
+  } catch (...) {
```
imilinovic: catch exact exception if possible

Ignition: Done, and it turned out to be worth more than tidiness.

`locate_zone` reports its one failure as a `std::runtime_error`, which the standard specifies and I confirmed against this toolchain: an unknown name, an empty name, an offset string and invalid bytes all produce exactly `std::runtime_error`, never a derived type.

The reason to narrow it is that `std::bad_alloc` does not derive from `runtime_error`. With `catch (...)` an allocation failure while loading the database would have been reported to the user as a timezone name missing from the IANA database, which is both wrong and hard to diagnose. It now propagates.

One thing worth deciding separately: `ParseTimezoneFromName` just above does the same lookup with the same `catch (...)`, and predates this change. Happy to narrow it in this pull request if you would rather the file were consistent, or leave it out to keep the diff to the reported problem.

imilinovic (approved):

PR #4593, opened 2026-08-15
fix(query): correct which temporal components may be omitted, and what they may hold

### LINKED ISSUE #4527: localtime({minute: 30}) and date({year: 1984, day: 5}) silently default omitted higher-significance fields instead of erroring
### Memgraph version

Memgraph 3.12.0 (image digest `sha256:e57cdd6a27c909d16b38d69bf1dc744b9707f966bc82c139ebe2c76e7187de14`)

Also reproduced on Memgraph 3.11.0.

### Environment
- Host OS: Linux x86_64
- Deployment: Docker (`memgraph/memgraph:3.12.0`, fresh container)
- Query interface: mgconsole
- Query language: Cypher

### Description

Constructing a temporal value from a map while omitting a field of
higher significance than a specified one is accepted silently: the
omitted field defaults to its minimal value.

The openCypher CIP 2015-08-06 (constructing temporal instant values)
states: "Omitted components are assumed to have their minimal possible
value. It is however not permitted to omit a value of higher
significance than one that has been specified."

`localtime({minute: 30})` omits `hour` (higher significance than
`minute`) but returns `00:30:00.000000000` instead of an error.
`date({year: 1984, day: 5})` omits `month` but returns `1984-01-05`.

### To reproduce

```cypher
RETURN localtime({minute: 30}) AS r
RETURN date({year: 1984, day: 5}) AS r
```

### Expected behavior

An error (missing required field of higher significance), matching the
CIP rule and the reference engine.

### Actual behavior

```text
00:30:00.000000000
1984-01-05
```

### Controls (work correctly)

| Query | Memgraph 3.11 | Neo4j 2026.06 |
|-------|:---:|:---:|
| `localtime({minute: 30})` | **00:30:00.000000000** | ERROR "hour must be specified" |
| `date({year: 1984, day: 5})` | **1984-01-05** | ERROR "day cannot be specified without month" |
| `localtime({second: 30})` | **00:00:30.000000000** | ERROR "hour must be specified" |

### Isolated recheck

Fresh `memgraph/memgraph:3.11.0` container, 3 runs: 3/3 reproduce
(stable).

### Why this looks like a bug

The CIP rule is explicit ("It is however not permitted to omit a value
of higher significance than one that has been specified"). Silent
defaulting of the omitted higher-significance field contradicts the
rule and diverges from the reference engine, which errors with a
specific message.

Ignition:
Building a date or a local time from a map now requires the leading component,
the year or the hour respectively. Every other component is an offset within
that one, so a map naming only lesser components leaves the value anchored
nowhere; it used to be accepted silently, taking 1970 or midnight.

Components below the leading one may be left out in any combination and take
their lowest value, so `localtime({hour: 5, second: 32})` is five o'clock
exactly. openCypher forbids omitting a component more significant than one that
has been specified, and Neo4j rejects that query, but the gap is unambiguous and
refusing it only makes callers pad the map with zeroes. The difference is
deliberate, and it leaves `date({year: 1984, day: 5})` accepted where the
reference engine reports an error.

The error is raised after the components are read, so an unrecognised name is
still reported as one rather than as a missing component.

A sub-second value beyond its own scale now carries into the coarser field,
making 1500 microseconds a millisecond and a half rather than an error, since
requiring each field to fill only its own three digits made the caller convert a
measurement they already held. The fraction is stored canonically, one
representation per instant, which the field-wise comparison and the hash both
depend on, and it must still stop short of a whole second.

### DIFF
--- a/src/query/interpret/awesome_memgraph_functions.cpp
+++ b/src/query/interpret/awesome_memgraph_functions.cpp
@@ -1663,9 +1663,12 @@ bool MapNumericParameters(auto &parameter_mappings, const auto &input_parameters
 }
 
 // Instant components are named from most to least significant, and an omitted
-// one takes its lowest value. That default is only meaningful for a trailing
-// run: naming a component while a more significant one is missing would leave
-// part of the instant to a default the caller was in the middle of replacing.
+// one takes its lowest value. That default only makes sense for a trailing run:
+// naming a component while a more significant one is missing would mix an
+// explicit value with a defaulted one above it.
+//
+// Call this after the components have been mapped, so that an unrecognised name
+// is reported as such rather than as a gap between the ones understood.
 void EnsureNoOmittedSignificantComponent(std::initializer_list<std::string_view> components_by_significance,
                                          const auto &input_parameters) {
   std::string_view first_omitted;
@@ -1719,8 +1722,6 @@ TypedValue Date(const TypedValue *args, int64_t nargs, const FunctionContext &ct
                                            std::pair{"month"sv, &date_parameters.month},
                                            std::pair{"day"sv, &date_parameters.day}};
 
-  // After the mapping, so an unrecognised key is reported as such rather than
-  // as a gap between the components that were understood.
   MapNumericParameters<Integer>(parameter_mappings, args[0].ValueMap());
   EnsureNoOmittedSignificantComponent({"year", "month", "day"}, args[0].ValueMap());
   return TypedValue(utils::Date(date_parameters), ctx.memory);
--- a/src/query/interpret/awesome_memgraph_functions.cpp
+++ b/src/query/interpret/awesome_memgraph_functions.cpp
@@ -1667,20 +1667,29 @@ bool MapNumericParameters(auto &parameter_mappings, const auto &input_parameters
 // naming a component while a more significant one is missing would mix an
 // explicit value with a defaulted one above it.
 //
+// Each level holds the names sharing one significance. The sub-second fields
+// are alternative scales for a single fraction of a second rather than steps of
+// their own, so any combination of them may be given together, and the level
+// counts as supplied when any one of them is.
+//
 // Call this after the components have been mapped, so that an unrecognised name
 // is reported as such rather than as a gap between the ones understood.
-void EnsureNoOmittedSignificantComponent(std::initializer_list<std::string_view> components_by_significance,
+void EnsureNoOmittedSignificantComponent(std::initializer_list<std::initializer_list<std::string_view>> levels,
                                          const auto &input_parameters) {
+  const auto is_supplied = [&input_parameters](const std::string_view component) {
+    return std::ranges::any_of(input_parameters,
+                               [component](const auto &entry) { return std::string_view{entry.first} == component; });
+  };
+
   std::string_view first_omitted;
-  for (const auto component : components_by_significance) {
-    const auto supplied = std::ranges::any_of(
-        input_parameters, [component](const auto &entry) { return std::string_view{entry.first} == component; });
-    if (!supplied) {
-      if (first_omitted.empty()) first_omitted = component;
+  for (const auto level : levels) {
+    const auto *const supplied = std::ranges::find_if(level, is_supplied);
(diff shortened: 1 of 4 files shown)


Ignition:
Reported as #4527.

Ignition:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    -  date() and localtime() now reject a missing bigger unit, so localtime({minute: 30}) is an error instead of 00:30; write localtime({hour: 0, minute: 30}) [#4593](https://github.com/memgraph/memgraph/pull/4593)

src/query/interpret/awesome_memgraph_functions.cpp
```diff
@@ -1677,6 +1700,7 @@ TypedValue LocalTime(const TypedValue *args, int64_t nargs, const FunctionContex
   };
 
   MapNumericParameters<Integer>(parameter_mappings, args[0].ValueMap());
+  EnsureNoOmittedSignificantComponent({"hour", "minute", "second", "millisecond", "microsecond"}, args[0].ValueMap());
```
imilinovic: PR in general seems OK but please double check that his is all required, seems a bit weird to require microseconds as mandatory

Ignition: Checked, and you were right that something was off, though nothing here was mandatory: the check only rejects a gap, so `localtime({hour: 10})` and `date({year: 1984})` are fine and the trailing components default.

The real problem was treating the sub-second fields as chain steps. Neo4j accepts every subset of them once `second` is present, so `localtime({hour:1, minute:2, second:3, microsecond:5})` is valid there and this PR was rejecting it. They are now one level below second. The hour/minute/second chain stays strict, matching Neo4j: `{hour:1, second:3}` is rejected.

While there, I also lifted the 0..999 cap on microsecond, since refusing `microsecond: 1500` was imposing our storage layout on the caller. It now carries into the coarser field, and the fraction is stored canonically so an instant keeps one representation, which the field-wise comparison and the hash rely on. It must still stay under a whole second. That was briefly a separate PR and is now folded in here.

Ignition: Correcting myself on one detail above: I said the sub-second fields are "freely combined" and not bounded like chain components, citing `microsecond:1500` giving .001500. That is only true when microsecond is the coarsest sub-second field given. Neo4j 2026.06 in full:

```
microsecond:999999                 -> 01:02:03.999999
microsecond:1000000                -> out of range
millisecond:4, microsecond:1500    -> out of range, 'Microsecond' expected 0 to 999
millisecond:999, microsecond:999999 -> out of range
```

So whichever sub-second field is coarsest may carry the whole remaining fraction, and any finer one given alongside it is bounded to its own three digits. Memgraph bounds both to 0..999 unconditionally (`temporal.cpp:394`), so `microsecond:1500` is an error for us where Neo4j reads it as 1.5ms. That divergence predates this PR and is untouched by it.

What matters for this check is unchanged, and it is only about which combinations are accepted: every subset of the sub-second fields is valid provided second is present, so an omission among them is never a gap. All seven subsets confirmed. The hour/minute/second chain is strict, as you would expect: `localtime({hour:1, second:3})` and `localtime({minute:2, second:3})` are both rejected by Neo4j and by this PR.

I have also reworded the comment on the helper. It previously justified the grouping by calling the fields alternative scales of one fraction, which reads as though they were interchangeable. It now says what the check is actually for: an omission among the sub-second fields does not signal the mistake that an omitted hour or minute does.

Ignition: Following on from the bounds point above: rejecting `microsecond: 1500` was us imposing a storage detail on the caller, so #4614 relaxes it. A sub-second value beyond its own scale now carries into the coarser field, making 1500 microseconds a millisecond and a half, and the fraction is stored canonically so an instant keeps one representation. It still has to stop short of a whole second, since carrying into the second would silently change a component the caller gave.

That is a separate change from this one and does not block it: this PR decides which combinations are accepted, #4614 decides what values they may hold.

imilinovic (approved):

PR #4594, opened 2026-08-15
fix(query): see through double negation when choosing an index

### LINKED ISSUE #4459: Double negation prevents an indexed equality from using the property index
# Double negation prevents an indexed equality from using the property index

I found that adding a double negation to an indexed equality makes the query much slower. Both queries return the same rows, but the double-negated form scans all nodes instead of using the property index.

- Memgraph Version: 3.10.1
- Operating System: macOS 27.0 arm64
- Installation Method: Docker image `memgraph/memgraph:latest`

### Steps to reproduce

1. Create the property index:

```cypher
CREATE INDEX ON :Item(key);
```

2. Create 500,000 `:Item` nodes. Run the following statement in batches, changing the range by 5,000 each time until the last range is `495000, 499999`:

```cypher
UNWIND range(0, 4999) AS i
CREATE (:Item {id: i, key: i % 1000});
```

3. Create 100 `:Item` nodes without `key`:

```cypher
UNWIND range(500000, 500099) AS i
CREATE (:Item {id: i});
```

4. Compare the complete matching ID lists:

```cypher
MATCH (n:Item)
WHERE n.key = 42
RETURN n.id AS id
ORDER BY id;

MATCH (n:Item)
WHERE NOT NOT (n.key = 42)
RETURN n.id AS id
ORDER BY id;
```

5. Run the following count queries several times in alternating order.

Query A:

```cypher
MATCH (n:Item)
WHERE n.key = 42
RETURN count(*) AS c;
```

Query B adds only `NOT NOT`:

```cypher
MATCH (n:Item)
WHERE NOT NOT (n.key = 42)
RETURN count(*) AS c;
```

6. Inspect both execution plans:

```cypher
EXPLAIN MATCH (n:Item)
WHERE n.key = 42
RETURN count(*) AS c;

EXPLAIN MATCH (n:Item)
WHERE NOT NOT (n.key = 42)
RETURN count(*) AS c;
```

### Expected behavior

Double negation preserves the result of a boolean expression, including `null`. Both queries should use the `:Item(key)` index and have comparable execution time.

### Actual behavior

Both queries return `c = 500`, and their complete matching ID lists are equal. Two fresh runs produced:

| Run | Query A | Query B |
|---|---:|---:|
| 1 | 1.53 ms | 100.12 ms |
| 2 | 1.43 ms | 105.85 ms |

`EXPLAIN` for query A uses the property index:

```text
* ScanAllByLabelProperties (n :Item {key})
```

`EXPLAIN` for query B scans all nodes:

```text
* Filter (n :Item), Generic {n}
* ScanAll (n)
```

Ignition:
A predicate negated twice is now analysed as the predicate itself when an index
is chosen, so a comparison written under two negations can be served by an index
rather than by a scan and a filter. Negating twice leaves a predicate unchanged,
null included, but analysis stopped at the outer negation and recorded the whole
expression as a generic filter, hiding the comparison inside.

Analysing what the pair wraps re-enters the filter collector, so a conjunction
inside it is split as usual. A single negation is unchanged, being no comparison
an index can answer.


Ignition:
Reported as #4459.

src/query/plan/preprocess.cpp
```diff
@@ -988,6 +988,14 @@ void Filters::AnalyzeAndStoreFilter(Expression *expr, const SymbolTable &symbol_
...
     }
   } else if (auto *is_not = utils::Downcast<NotOperator>(expr)) {
+    // Negating twice leaves a predicate unchanged, `null` included, so analyse
+    // what the pair wraps. Left as it is, the outer negation would only ever
+    // yield a generic filter and hide a comparison an index could answer.
+    // Going back through the splitter also breaks up any conjunction inside.
+    if (auto *inner_not = utils::Downcast<NotOperator>(is_not->expression_)) {
+      CollectFilterExpression(inner_not->expression_, symbol_table);
+      return;
+    }
```
imilinovic, line 998: does it make sense to allow N-negation (triple, quadruple, etc)?

Ignition:
### Tracking

### Standard development
- [ ] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    -  WHERE NOT NOT n.prop = 42 now uses the property index rather than scanning the label, with the same results. [#4594](https://github.com/memgraph/memgraph/pull/4594)

src/query/plan/preprocess.cpp
Ignition: Yes, and it already does, which I had reasoned but not shown. Thanks for asking, because the claim was resting on my reading of the recursion rather than on anything in the suite.

`CollectFilterExpression` goes back through the splitter and reaches this branch again, so each pass strips one pair and re-analyses what is left. Depth unwinds a pair at a time with no special case:

- `NOT NOT NOT NOT x` strips twice and ends at `x`
- `NOT NOT NOT x` strips once and ends at `NOT x`, which is not a comparison an index can answer

So an even count leaves the comparison exposed to index selection and an odd count leaves a single negation, which is the right answer for both rather than a lucky one.

Both are now asserted in `MatchFilterDoubleNegatedPropertyUsesIndex`: four negations plan the index scan, three plan a scan and a filter. The odd case matters more than it looks, since the failure mode there would be treating `NOT x` as if it were `x` and returning the wrong rows quickly.

No change to the code, only the tests.

colinbarry (approved):
NOT NOT approved 😆

PR #4596, opened 2026-08-16
feat: allow EXISTS subqueries inside a CASE expression

imilinovic:
> **Stacked PR.** Base is `fix/exists-empty-body-segv` (#4560), which is itself stacked on `feat/exists-in-projection` (#4504). Review only the two commits above the base; merge the stack bottom-up.

Lets an `EXISTS` subquery appear inside a `CASE` expression.

**What.** `CASE WHEN cond THEN EXISTS { ... } ELSE ... END`, and `EXISTS` in the `CASE` condition itself, are now planned and evaluated. They previously threw `IF operator cannot be used with exists, but only during matching!` in any position, including ones where a bare `EXISTS` was already accepted.

**How.** The whole production change is deleting that refusal from `SymbolGenerator::PreVisit(Exists&)` (`symbol_generator.cpp`). A `CASE` holds no position of its own — it carries whichever position it sits in — so the position allow-list (`IsSupportedExistsPosition`) added later already answers for every shape the `num_if_operators` gate used to catch, and still refuses the ones with no splice point, `CASE`-wrapped or not. No planner change is needed: the planner's collectors are expression-tree visitors that already descend through `IfOperator`.

The counter itself stays. `PreVisit(Aggregation&)` reads it to refuse an aggregation inside a `CASE`, which is untouched, and a `CASE`-wrapped `SET` stays in the refused set — so the two gates remain distinguishable now that the `CASE` message is gone.

**Why.** The gate predates the position allow-list and was added with no planner change and no recorded hazard; it had nothing left to protect. It made `CASE` an unexplained hole in `EXISTS` support — the same subquery was accepted directly in a `RETURN` but refused one `CASE` deeper.

### DIFF
--- a/src/query/frontend/ast/cypher_main_visitor.cpp
+++ b/src/query/frontend/ast/cypher_main_visitor.cpp
@@ -3954,8 +3954,11 @@ antlrcpp::Any CypherMainVisitor::visitExistsSubquery(MemgraphCypher::ExistsSubqu
     // Curly-brace subquery form: { cypherQuery }
     // Set the flag to indicate we are parsing an EXISTS subquery
     auto old_flag = parsing_exists_subquery_;
+    // The body's clauses are its own, so the enclosing WITH's "everything must be aliased" rule does not reach them.
+    auto old_in_with = std::exchange(in_with_, false);
     parsing_exists_subquery_ = true;
     auto *cypher_query = std::any_cast<CypherQuery *>(ctx->cypherQuery()->accept(this));
+    in_with_ = old_in_with;
     parsing_exists_subquery_ = old_flag;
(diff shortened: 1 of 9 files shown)


imilinovic:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - An `EXISTS` subquery can now be used inside a `CASE` expression, so conditional existence checks no longer have to be rewritten around the `CASE`; write them inline as `CASE WHEN EXISTS { ... } THEN ... END`. [#4596](https://github.com/memgraph/memgraph/pull/4596)

colinbarry (approved):
Approved! 💪

PR #4597, opened 2026-08-16
fix: plan an EXISTS body's RETURN instead of discarding it

imilinovic:
> Formerly stacked on #4560 and #4596; both have merged, so this now targets `master` directly. #4598 (`COUNT { }`) stacks on top of it.

Makes an `EXISTS` subquery's body `RETURN` take effect instead of being thrown away.

**What.** `GenReturn` discarded the whole `ReturnBody` inside an exists subquery, so everything on it that decides the row count went with it — `DISTINCT`, `SKIP`, `LIMIT` and aggregation — and the fold answered from a branch that no longer represented the query:

| body | said | should say |
|---|---|---|
| `MATCH (p)-[:KNOWS]->(f) RETURN count(f)` | `false` for a person with no friends | `true` — an ungrouped aggregate emits one row even on empty input |
| `MATCH (p)-[:KNOWS]->(f) RETURN f LIMIT 0` | `true` | `false` — the table is empty |
| `MATCH (p)-[:KNOWS]->(f) RETURN f SKIP 1` | `true` for a one-row table | `false` |

The same query written with a `WITH` always behaved correctly, so the two spellings disagreed on the same graph. A `UNION` arm was wrong the same way: `EXISTS { RETURN 1 AS c SKIP 1 UNION RETURN 2 AS c SKIP 1 }` said `true`, because both arms planned to a bare `Once`.

**How.** Deletes the four-line early return from `GenReturn` (`rule_based_planner.cpp`). The `in_exists_subquery` parameter stays — `GenWith` still reads it to preserve outer-scope variables across a body `WITH` — but `GenReturn` no longer takes it.

Planning the body's `RETURN` means the branch now reaches all of `GenReturnBody`, including two operators that have no business inside a filter predicate: `PeriodicCommit`, which would finalize and re-timestamp the caller's transaction from inside an expression, and `Accumulate`, which drains its input on the first pull, below the fold, turning one write into one per matched row. Neither can reach a body once the clause validation below is closed. So rather than pass narrowed arguments — which would silently turn a write executing N times into a write executing once, if that validation ever regained a hole — `CheckExistsBodyInvariants` states both invariants and throws. A throw and not an assert — an assert would abort the process from an `EXPLAIN` alone.

`USING PERIODIC COMMIT` and `USING PARALLEL EXECUTION` in a body are refused in `visitExistsSubquery`, next to the existing memory-limit check. A body parses as a full `cypherQuery`, so it carries the pre-query directives with it, but only the enclosing query's are ever read — so a body directive was accepted and then silently dropped. Periodic commit was worse than inert: the body's own `QueryParts` picks up `commit_frequency`, so it would have reached the planner.

Also corrects two comments in `rule_based_planner.hpp`. The null-`input_op` guard is now unreachable and kept as defence; the `EmptyResult` suppression stays **load-bearing** — a body with no `RETURN` at all (`EXISTS { MATCH (n:Person) }`) still has empty output symbols, and wrapping it would make it universally false.

**The `UNION` allow-list hole.** `visitExistsSubquery` validated its clause rules against `cypher_query->single_query_` only, so a `UNION`'s later branches — each its own `SingleQuery` — were never checked, and every clause the first branch forbids was legal in the second. On master, a write in a later branch executed from inside a `WHERE` predicate:

```cypher
MATCH (person:Person)
WHERE EXISTS { MATCH (person)-[:HAS_CAT]->(cat:Cat) RETURN cat AS d
               UNION MATCH (other:Dog) CREATE (:Marker) RETURN other AS d }
RETURN person.name;
-- accepted; one :Marker created by the predicate
```

The earlier branch has to match nothing for it to fire, since the fold stops at the first row, so it was intermittent as well as wrong. The clause checks are now factored into one validator applied to `single_query_` and to every `cypher_unions_[i]->single_query_`. The directive checks need no such loop: `memory_limit_` and `pre_query_directives_` live on `CypherQuery`, one per body.

**Why.** Accepted openCypher settles which spelling is right: only the number of rows in the body's result table is relevant, and the `RETURN` is what produces that table. What the spec says to ignore is the *columns*, not the clause — which is what the removed comment had misread. Every dependent construct inherits the bug, so this has to land before `COUNT { … }`, where four of five body shapes would otherwise be wrong.

**Behaviour changes.** None of these is breaking: each either fixes an answer that was already wrong or rejects input that used to produce one.

- A body `RETURN` now takes effect: `EXISTS { … RETURN count(*) }` becomes always-true, and `SKIP`/`LIMIT` in a body start being honoured. Affected queries return different results without raising an error.
- **A body `RETURN`'s expressions are now evaluated**, where the discarded clause used to evaluate nothing. So a body expression can raise where the query previously answered — `EXISTS { MATCH (n)-[]->(m) RETURN 1/0 }` now throws — and it does so data-dependently, since an outer row whose body `MATCH` yields no rows never evaluates. The equivalent `WITH` spelling already threw, so this makes the two agree.
- `USING PERIODIC COMMIT` or `USING PARALLEL EXECUTION` inside an `EXISTS` body is now a syntax error rather than being silently ignored.
- A clause outside `MATCH`/`WHERE`/`WITH`/`RETURN` in a `UNION` branch of an `EXISTS` body is now a syntax error rather than being accepted, with a write there executing inside the predicate.

**Cost.** A body `ORDER BY` now plans a real `OrderBy` inside the branch, and both folds `Reset()` the branch per outer row, so the sort is redone for every one — where the deferred fold's `Limit(1)` previously stopped at the first row. Discardable work, since a sort cannot change whether a table is empty, but removing it is not a one-line change: `GenReturnBody` is shared with `GenWith`, whose `WHERE` is built above `SKIP`/`LIMIT` and is therefore order-sensitive. Left as a follow-up — a rewrite over the branch root, not a flag.

### DIFF
--- a/tests/gql_behave/tests/memgraph_V1/features/memgraph_exists.feature
+++ b/tests/gql_behave/tests/memgraph_V1/features/memgraph_exists.feature
@@ -739,6 +739,24 @@ Feature: WHERE exists
           """
       Then an error should be raised
 
+  Scenario: Test invalid parallel execution inside EXISTS
+      Given an empty graph
+      And having executed:
+          """
+          CREATE (:Person {name: 'John'})-[:HAS_DOG]->(:Dog {name: 'Rex'})
+          """
+      When executing query:
+          """
+          MATCH (person:Person)
+          WHERE EXISTS {
+            USING PARALLEL EXECUTION
+            MATCH (person)-[:HAS_DOG]->(dog:Dog)
+            RETURN dog
+          }
+          RETURN person.name AS name;
+          """
+      Then an error should be raised
+
   Scenario: Test invalid SET inside EXISTS
       Given an empty graph
       And having executed:
@@ -1777,7 +1795,6 @@ Feature: WHERE exists
           CREATE (a:Person {name: 'Alice'}), (b:Person {name: 'Bob'}), (c:Person {name: 'Carol'})
           CREATE (f1:Friend {name: 'F1'}), (f2:Friend {name: 'F2'})
           CREATE (a)-[:KNOWS]->(f1)
-          CREATE (a)-[:KNOWS]->(f1)
           CREATE (a)-[:KNOWS]->(f2)
           CREATE (b)-[:KNOWS]->(f1)
           """
@@ -1800,7 +1817,6 @@ Feature: WHERE exists
           """
           CREATE (a:Person {name: 'Alice'}), (b:Person {name: 'Bob'}), (c:Person {name: 'Carol'})
           CREATE (f1:Friend {name: 'F1'}), (f2:Friend {name: 'F2'})
-          CREATE (a)-[:KNOWS]->(f1)
           CREATE (a)-[:KNOWS]->(f1)
           CREATE (a)-[:KNOWS]->(f2)
           CREATE (b)-[:KNOWS]->(f1)
@@ -1829,7 +1845,6 @@ Feature: WHERE exists
           CREATE (a:Person {name: 'Alice'}), (b:Person {name: 'Bob'}), (c:Person {name: 'Carol'})
           CREATE (f1:Friend {name: 'F1'}), (f2:Friend {name: 'F2'})
           CREATE (a)-[:KNOWS]->(f1)
-          CREATE (a)-[:KNOWS]->(f1)
           CREATE (a)-[:KNOWS]->(f2)
           CREATE (b)-[:KNOWS]->(f1)
           """
@@ -1851,7 +1866,6 @@ Feature: WHERE exists
           """
           CREATE (a:Person {name: 'Alice'}), (b:Person {name: 'Bob'}), (c:Person {name: 'Carol'})
           CREATE (f1:Friend {name: 'F1'}), (f2:Friend {name: 'F2'})
-          CREATE (a)-[:KNOWS]->(f1)
           CREATE (a)-[:KNOWS]->(f1)
           CREATE (a)-[:KNOWS]->(f2)
           CREATE (b)-[:KNOWS]->(f1)
@@ -1875,7 +1889,6 @@ Feature: WHERE exists
           CREATE (a:Person {name: 'Alice'}), (b:Person {name: 'Bob'}), (c:Person {name: 'Carol'})
           CREATE (f1:Friend {name: 'F1'}), (f2:Friend {name: 'F2'})
           CREATE (a)-[:KNOWS]->(f1)
-          CREATE (a)-[:KNOWS]->(f1)
           CREATE (a)-[:KNOWS]->(f2)
           CREATE (b)-[:KNOWS]->(f1)
           """
@@ -1891,7 +1904,8 @@ Feature: WHERE exists
           | 'Bob'   | false |
           | 'Carol' | false |
 
-  # Alice is the discriminating row: 3 rows, 2 distinct, so the SKIP clears the table only if DISTINCT was planned.
+  # Alice is the discriminating row: her doubled KNOWS edge to F1 makes 3 rows but 2 distinct, so the SKIP clears
+  # the table only if DISTINCT was planned. It is the one fixture here that needs a parallel edge.
   # Dave, with 3 distinct, keeps a row either way, so an all-false table cannot pass by accident.
   Scenario: Test EXISTS subquery whose body RETURN is DISTINCT
       Given an empty graph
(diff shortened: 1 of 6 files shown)


imilinovic:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - fix: A `RETURN` inside an `EXISTS { … }` subquery is now planned instead of discarded, so `DISTINCT`, `SKIP`, `LIMIT` and aggregation in the body decide the answer. Affected queries **return different results without raising an error** — `EXISTS { MATCH (p)-[:KNOWS]->(f) RETURN count(f) }` is now true even when nothing matched, `RETURN f LIMIT 0` is now false, and a body expression that throws now throws, so `RETURN 1/0` raises a query error only where the body matched, letting the same query pass on one dataset and fail on another. Recheck any query that relied on a body `RETURN` being ignored; the equivalent `WITH` spelling was already correct, so the two now agree. [#4597](https://github.com/memgraph/memgraph/pull/4597)
    - fix: Clauses in a `UNION` branch of an `EXISTS { … }` body are now validated like the first branch's. A write clause there — `CREATE`, `SET`, `DELETE`, `REMOVE`, `MERGE` — was accepted and executed from inside the predicate, creating or modifying data during what looked like a read-only query; it now raises a query error. [#4597](https://github.com/memgraph/memgraph/pull/4597)
    - fix: `USING PERIODIC COMMIT` and `USING PARALLEL EXECUTION` inside an `EXISTS { … }` body now raise a query error instead of being silently accepted and ignored. Only the enclosing query's directives were ever read, so a directive written in the body had no effect. [#4597](https://github.com/memgraph/memgraph/pull/4597)

tests/gql_behave/tests/memgraph_V1/features/memgraph_exists.feature
```diff
@@ -1699,3 +1766,347 @@ Feature: WHERE exists
...
+  # DISTINCT alone cannot - it never empties a non-empty table - so it appears only composed with a SKIP.
+
+  Scenario: Test EXISTS subquery whose body RETURN aggregates
+      Given an empty graph
+      And having executed:
+          """
+          CREATE (a:Person {name: 'Alice'}), (b:Person {name: 'Bob'}), (c:Person {name: 'Carol'})
+          CREATE (f1:Friend {name: 'F1'}), (f2:Friend {name: 'F2'})
+          CREATE (a)-[:KNOWS]->(f1)
+          CREATE (a)-[:KNOWS]->(f1)
```
as51340: is this a bug, two same relationships?

tests/gql_behave/tests/memgraph_V1/features/memgraph_exists.feature
```diff
@@ -1699,3 +1766,347 @@ Feature: WHERE exists
...
+
+  # Invariance pin: the WITH spelling was always planned, so this agrees either way. It guards the two from diverging.
+  Scenario: Test EXISTS subquery whose body RETURN aggregates agrees with the WITH spelling
+      Given an empty graph
+      And having executed:
+          """
+          CREATE (a:Person {name: 'Alice'}), (b:Person {name: 'Bob'}), (c:Person {name: 'Carol'})
+          CREATE (f1:Friend {name: 'F1'}), (f2:Friend {name: 'F2'})
+          CREATE (a)-[:KNOWS]->(f1)
+          CREATE (a)-[:KNOWS]->(f1)
```
as51340: same

tests/gql_behave/tests/memgraph_V1/features/memgraph_exists.feature
```diff
@@ -1699,3 +1766,347 @@ Feature: WHERE exists
...
+  # the body's RETURN discarded - so this pair guards a future rewrite that prunes an unread projection column from
+  # silently turning the grouped answer into the ungrouped one. Both measured against the reference engine.
+  Scenario: Test EXISTS subquery whose body RETURN aggregates with a grouping key
+      Given an empty graph
+      And having executed:
+          """
+          CREATE (a:Person {name: 'Alice'}), (b:Person {name: 'Bob'}), (c:Person {name: 'Carol'})
+          CREATE (f1:Friend {name: 'F1'}), (f2:Friend {name: 'F2'})
+          CREATE (a)-[:KNOWS]->(f1)
+          CREATE (a)-[:KNOWS]->(f1)
```
as51340: same

as51340: same at places below, check please

as51340 (approved):

Ignition (approved):
just one change for SyntaxException for a sub-query parallel execution

PR #4598, opened 2026-08-16
feat: add the COUNT {} subquery

imilinovic:
> **Stacked PR.** Base is `fix/exists-body-return` (#4597) → `feat/exists-in-case` (#4596) → `fix/exists-empty-body-segv` (#4560) → `feat/exists-in-projection` (#4504). Merge bottom-up. The base is not optional: without it four of the five body shapes below answer wrongly.

`COUNT { … }` counts the rows a correlated subquery's body produces, in both spellings `EXISTS` accepts — a bare pattern, `COUNT { (p)-[:KNOWS]->() }`, and a full body, `COUNT { MATCH (p)-[:KNOWS]->(f) RETURN f }`, including `WHERE`, `WITH`, `UNION`, `DISTINCT`, `SKIP`/`LIMIT` and aggregation. It works in every position `EXISTS` does — projection, `WHERE`, `ORDER BY`, `CASE`, an aggregate's argument — because it reaches the planner through the same splice points.

## How

The two constructs differ only in what the branch's rows reduce to, so this adds a fold axis rather than a second machinery.

| layer | change |
|---|---|
| grammar | `COUNT '{' countSubquery '}'` in `atom`, routed ahead of the existing arm, which is `COUNT(*)` |
| AST | one `Exists` node with a `Fold` field (`kBool`/`kCount`), not a class per spelling. Scoping, the position allow-list and the planner's collectors keep reading one node type; only diagnostics branch, through `FoldName()`, so a `COUNT` is never reported as an `EXISTS` |
| operators | `Fold` gains `kCount` and moves to namespace scope, so `EvaluatePatternFilter` — declared earlier — can name it too |
| planner | the fold is threaded through `preprocess`, both `RollUpApply` splice sites, and `variable_start_planner`, which rebuilds matchings field by field |

Two deliberate asymmetries with the bool fold:

**The count branch drains into a counter, not a vector.** Collecting the rows and taking the length would allocate an N-element list per input row, which is the cost a count over a supernode exists to avoid — a 200M-row drain grows RSS by 644 kB. The branch `Limit` is likewise not planted above a count, since it would truncate the drain. (Under the bool fold that `Limit` caps nothing — `EvaluatePatternFilterCursor` pulls exactly once whatever sits above it — and removing it is tracked separately.)

**In a pattern-level `WHERE` the fold is deferred** into the closure `EvaluatePatternFilter` hands the evaluator, as the bool fold already is. That buys two things a forced fold cannot: a disjunct the evaluator never reaches costs nothing, which for a count is a whole drain rather than one pull; and a count works inside a per-element lambda (`all`/`any`/`none`) at all, because the closure re-runs the branch from `Reset()` per element while an operator can only write one value per row.

## Conjunct ordering

`ExtractFilters` sorts a conjunct carrying a subquery branch last, so a cheaper one decides the row first and a deferred fold whose value is never read never runs its branch. Two things were wrong with that:

- It keyed on `FilterInfo::Type::Pattern`, which is set only when the conjunct **is** an `Exists` — so `COUNT { … } > 1`, an `Exists` wrapped in a comparison, was tagged `Generic` and kept whatever position collection order gave it. `exists_matchings` already comes from walking the whole filter expression, so it sees the wrapped case the type tag misses. `NOT EXISTS { … }` is fixed by the same change.
- Sorting them last says nothing about the order of two such conjuncts, and `SplitExpression` is a LIFO stack, so they arrived reversed: the cheap one landed behind the expensive one and paid its whole drain. On 50 hubs owning 2000 edges each, `WHERE COUNT { cheap } > 0 AND COUNT { 4M rows } > 1` goes **3.55 s → 0.00 s**, same answer. Only the moved conjuncts are reversed — reversing the whole collection breaks range-filter index selection.

Conjunct order is observable when an operand throws, so which exception surfaces first can change for existing wrapped-`EXISTS` queries. Answers cannot change: three-valued `AND` is commutative.

## Hardening

A missed fold site does not fail closed, as the one-node design assumed. A bool left in a count's frame slot answers `false` to `=` and `Null` to `>`, so rows quietly vanish and nothing is raised. `ToOperatorFold` is now a `switch`, so a third fold fails to compile rather than folding to `kBool`; `EvaluatePatternFilter`'s `Fold` has no default argument, matching `RollUpApply`'s; and the evaluator checks the frame value's type against the fold in Debug.

## Out of scope

The `COUNT { … } > 0 → EXISTS { … }` rewrite (early exit instead of a drain), and the same ordering gap under `OR`, where operands live in one expression rather than separate `FilterInfo`s so the partition never sees them.

Measured against Neo4j 2026.02.2 and refused here — each shared with `EXISTS` rather than new:

- a bare-pattern body that *declares* a variable (`COUNT { (p)-[:X]->(d:Dog) }`), and a bare pattern with a trailing `WHERE`. Both are expressible by adding `MATCH`, so this is syntax surface, not semantics.
- `SET p.x = COUNT { … }` — the forced fold has nowhere to sit above a `SET`. Adjacent: `CALL proc YIELD x WHERE COUNT { … }` passes the position gate and then internal-errors.
- a body outside the `MATCH`/`WHERE`/`WITH`/`RETURN` allow-list, e.g. `UNWIND` or a nested `CALL {}`. `UNWIND` is safe to admit; `CALL {}` needs a recursive read-only check, since that allow-list is what keeps writes out of a body.


imilinovic:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - feat: `COUNT { … }` counts the rows a subquery body produces, written either as a full body, `COUNT { MATCH (p)-[:KNOWS]->(f) RETURN f }`, or as the bare-pattern short form, `COUNT { (p)-[:KNOWS]->() }`, in which any node or relationship the pattern introduces must be unnamed. It goes in a `WITH`/`RETURN` projection, a `WHERE` or an `ORDER BY`, and the body's own clauses decide the number, so a `DISTINCT`, `SKIP` or `LIMIT` inside it applies — and a body ending in `RETURN count(f)` counts 1, the aggregate's own single row, not the rows that fed it. [#4598](https://github.com/memgraph/memgraph/pull/4598)

DavIvek (approved):

PR #4601, opened 2026-08-17
feat: report SYNC replication failure as a notification instead of an exception

as51340:
A transaction that cannot reach every SYNC replica is still committed on the main and on every reachable replica, so failing the query misrepresents what happened. Report it as a `WARNING` notification carrying the same message instead, and keep `ReplicationException` only for the case where the transaction was actually rolled back everywhere.

### Behaviour

`Interpreter::Commit()` branches on `ReplicationError::transaction_committed`:

- **committed on main** (no STRICT_SYNC replica took part in the txn, so the failures are SYNC-only) → notification, code `SyncReplicationFailure`, severity `WARNING`, title = the text the exception used to carry (`Failed to replicate to SYNC replica '…': … Replica will be recovered automatically. Transaction is still committed on the main instance and other alive replicas. Check the status of the replicas using 'SHOW REPLICAS' query.`)
- **aborted everywhere** (STRICT_SYNC 2PC rollback) → `ReplicationException`, unchanged

`transaction_committed` is the discriminator rather than the failure's `mode` field on purpose: in a STRICT_SYNC cluster a *SYNC*-mode failure also aborts the 2PC transaction, and keying on mode alone would report an aborted write to the client as a success. This also lines up the commit path with `HandlePeriodicCommitError`, which already used `transaction_committed` the same way.

The notification reaches the client for both implicit and explicit transactions:

- autocommit → appended to the query summary after `Commit()` (the pre-existing notification block has to stay *before* `Commit()`, which clears every query execution's runtime data)
- explicit `COMMIT` → `Interpreter::CommitTransaction()` returns the notification, `SessionHL::CommitTransaction()` turns it into bolt metadata, and `HandleCommit` sends it in the SUCCESS reply (which previously always carried an empty map)


Ignition (approved):

as51340:
### Tracking

### Standard development
- [x] Update unit/E2E tests

### CI Testing Labels

### Documentation checklist
    - A transaction that cannot reach every SYNC replica is still committed on the main and on every reachable replica, so failing the query misrepresents what happened. Reporting it now as a WARNING notification carrying the same message instead, and keeping ReplicationException only for the case where the transaction was actually rolled back everywhere. [#4601](https://github.com/memgraph/memgraph/pull/4601)
- [x] **[ Documentation PR link ](https://github.com/memgraph/documentation/pull/1745)**

PR #4614, opened 2026-08-17
fix(utils): carry sub-second values into the coarser field

Ignition:
A local time built from components now reads a sub-second field beyond its own
scale as the time it names rather than refusing it, so 1500 microseconds is a
millisecond and a half. Requiring each field to fill only its own three digits
made the caller convert a measurement they already held, for no gain.

The fraction is stored canonically, one representation per instant, which the
field-wise comparison and the hash both depend on. It must still stop short of a
whole second, since carrying into the second would silently change a component
the caller gave.


Ignition:
Folded into #4593, which is the same area of the temporal map handling and is still open for changes. The commit is unchanged apart from adding the minute the gap check in that PR now requires.

PR #4616, opened 2026-08-17
test(query): add benchmarks for aggregation and value handling

### LINKED ISSUE #4608: Aggregation query shapes have no performance regression coverage
## Problem

No benchmark in the repository exercises aggregation query shapes end to end. The benchmarks we have measure either individual components below the operators, or whole-system throughput. A change that slows aggregation by a noticeable margin can land without anything failing.

Detecting such a regression today means standing up a separate harness with a large generated dataset and a long run, which does not fit inside a normal review cycle. In practice that means it is not done per change.

## Acceptance criteria

- [ ] A benchmark in the repository runs the common aggregation shapes in process: grouping at several key cardinalities, filtered aggregation, conditional aggregation, computed grouping keys, aggregation over relationship properties, and aggregation into non-primitive results
- [ ] It builds its own dataset, needs no running server and no external dependency, and completes in a couple of minutes
- [ ] Cost per row is representative of a large graph, so a change that matters at scale is visible at test size
- [ ] It includes control cases that must **not** move, so a change can be shown to be targeted rather than broad
- [ ] It runs under the existing benchmark test label
- [ ] Documented: how to compare two builds, and what makes a result trustworthy on a noisy machine

## Notes

The control cases are the part most easily left out and the part that gives the rest its meaning. A benchmark suite where every case improves tells you nothing about whether a change was targeted.

Worth extending the same treatment to traversal shapes, which can double as controls for anything property-related.

Ignition:
Add benchmarks for the aggregation query shapes, and for what a query value
costs to copy, move, destroy and compare. A sequence of performance changes to
this area is coming and there is currently no way to measure one: the benchmarks
here sit either below the operators or above the whole engine, and asking the
question any other way means starting a server and generating a large dataset.
These run in the build, against a graph they build themselves.

Some of the cases are there to stay still rather than to improve. One reads no
property at all, one reads a single property once, one reads two properties once
each, and the value cases cover every type rather than only the ones a change
would be aimed at. They are what shows a change to be narrow rather than broad,
so they matter as much as the cases that move.

### DIFF
--- a/tests/benchmark/CMakeLists.txt
+++ b/tests/benchmark/CMakeLists.txt
@@ -36,6 +36,9 @@ target_link_libraries(${test_prefix}execution mg-query mg-communication)
 add_benchmark(query/aggregation.cpp)
 target_link_libraries(${test_prefix}aggregation mg-query)
 
+add_benchmark(query/typed_value_ops.cpp)
+target_link_libraries(${test_prefix}typed_value_ops mg-query)
+
 add_benchmark(query/planner.cpp)
 target_link_libraries(${test_prefix}planner mg-query)
 
--- a/tests/benchmark/CMakeLists.txt
+++ b/tests/benchmark/CMakeLists.txt
@@ -33,6 +33,9 @@ target_link_libraries(${test_prefix}eval mg-query)
 add_benchmark(query/execution.cpp ${CMAKE_SOURCE_DIR}/src/glue/communication.cpp)
 target_link_libraries(${test_prefix}execution mg-query mg-communication)
 
+add_benchmark(query/aggregation.cpp)
+target_link_libraries(${test_prefix}aggregation mg-query)
+
 add_benchmark(query/planner.cpp)
 target_link_libraries(${test_prefix}planner mg-query)
 
--- a/tests/benchmark/query/aggregation.cpp
+++ b/tests/benchmark/query/aggregation.cpp
@@ -17,8 +17,17 @@
 // graph in one process, so a single commit can be held against a single number.
 //
 // The graph is the one those queries were written for: a hub of degree N, whose neighbours carry a
-// skewed category, a uniform region and bucket, and the numbers the aggregations add up. It is built
-// once and shared, because building it costs far more than running any query over it.
+// skewed category, a uniform region and bucket, and the numbers the aggregations add up, with the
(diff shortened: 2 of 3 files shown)


Ignition:
Evidence, kept out of the description because that becomes the squash commit.

**Verification.** `ctest -L benchmark` is 22/22 green on these commits. Both binaries build and run against `master` with no product change behind them.

**Noise floor.** Baseline captured at three repetitions on an idle, core-pinned host: the query shapes vary by under 1.4% except `CountAllItems` at 2.75%, and the value operations by under 0.6%. A 2% effect is therefore measurable on the query shapes, and well under 1% on the value ones.

**Per-row cost across size.** At an eightfold larger graph (`MG_AGG_BENCH_ITEMS=2000000`), per-row throughput holds to within 2-19% and the relative ordering of the shapes is preserved:

| shape | 250k | 2M | drift |
|---|---:|---:|---:|
| SumAvgNumeric | 4.25 | 4.17 Mrow/s | -1.7% |
| SinglePropertyRead | 6.51 | 6.28 | -3.6% |
| CountNoGroupKey | 19.20 | 17.49 | -8.9% |
| RepeatedPropertyRead | 2.33 | 2.05 | -12.1% |
| GroupBySingleKey | 2.79 | 2.30 | -17.5% |
| FilteredAggregate | 3.26 | 2.63 | -19.3% |

Worth knowing when reading a result: the grouped and filtered shapes degrade with size more than the rest, so a change worth under about 20% on those two is worth confirming at a second size before believing it.

**Why `ConstructDestroy` passes its value to `DoNotOptimize`.** Without it, a type whose destructor does nothing and whose construction the compiler can see through is dead code: the loop body is deleted and the case measures an empty loop. On a build where those destructors are trivial it read about 0.4 ns against 3.7 ns here, which would present as a ninefold gain from a change that did nothing of the sort. It is the failure mode this kind of benchmark is most prone to, and it flatters exactly the changes the benchmark exists to judge.

**On reading any result from these.** Compare a build against a build in one sitting, and pin to a core. An early set of numbers taken while the same machine was compiling read `CompareNumbers` at 85.7 ns against 44.8 ns idle, on a case whose variation is otherwise 0.35%.

colinbarry (approved):
Looks good to me 💪

PR #4617, opened 2026-08-17
build(conan): build memgraph's dependencies on FreeBSD

seuros:
Two conan profiles and a CMake toolchain file describe a FreeBSD host: the
compiler is the system clang with libc++, and cmake, ninja, flex, antlr4 and
openjdk come from ports rather than being built. Several vendored recipes
needed portability fixes to compile there. nuraft recognises FreeBSD in its
attribute and include guards, and includes stddef.h for the size_t its own
declarations use. librdtsc falls back to CLOCK_MONOTONIC where
CLOCK_MONOTONIC_RAW is absent. libbcrypt skips the GNU compatibility
declarations in bcrypt.c, which uses only the OW interface, as crypt_blowfish
already does for itself. usearch passes -1 as the descriptor of an anonymous
mapping, which is what POSIX requires and FreeBSD enforces. Boost tells its
consumers to link pthread, without which anything using its threads fails to
link. Two more differences are settled in the profile: rocksdb's io_uring
support stays off, and the two dependencies with GNU-only makefiles are pointed
at gmake, scoped to them because that setting also reaches CMAKE_MAKE_PROGRAM.

Building on a host that has no bundled toolchain is now expressible.
Everything that knew the toolchain's directory layout moved into one CMake
module, so consumers link mg::toolchain rather than naming a runtime library by
path, and turning MG_BUNDLED_TOOLCHAIN off builds against the host's C++
runtime and installs nothing beside the binary. The build scripts identify
FreeBSD, describe its packages in an adapter beside the twelve Linux ones, and
derive both profiles and that option from the platform they detect. Two things
stand between this and memgraph running there: src/auth calls into libseccomp
unconditionally, and a FreeBSD graph needs partial locking, because tools taken
from the platform appear as references the lockfile does not carry. Linux is
unaffected, with every profile resolving to what it resolved to before, though
the refreshed lockfile does move the recipe revisions of the dependencies it
touches.


environment/toolchain/v7/toolchain-freebsd.cmake
```diff
@@ -0,0 +1,57 @@
...
+    set(CMAKE_OBJCOPY "${MG_TOOLCHAIN_ROOT}/bin/llvm-objcopy")
+    set(CMAKE_OBJDUMP "${MG_TOOLCHAIN_ROOT}/bin/llvm-objdump")
+    set(CMAKE_STRIP "${MG_TOOLCHAIN_ROOT}/bin/llvm-strip")
+    set(CMAKE_CXX_COMPILER_CLANG_SCAN_DEPS "${MG_TOOLCHAIN_ROOT}/bin/clang-scan-deps" CACHE STRING "" FORCE)
+
+    list(APPEND CMAKE_PREFIX_PATH "${MG_TOOLCHAIN_ROOT}")
+
+    # Exclude OpenSSL from toolchain search paths to force use of Conan-provided OpenSSL
+    list(APPEND CMAKE_IGNORE_PATH "${MG_TOOLCHAIN_ROOT}/lib/cmake/OpenSSL")
+    list(APPEND CMAKE_IGNORE_PATH "${MG_TOOLCHAIN_ROOT}/lib64/cmake/OpenSSL")
```
arrowd: We don't have `lib64` dir anywhere.

environment/toolchain/v7/toolchain-freebsd.cmake
```diff
@@ -0,0 +1,57 @@
...
+else()
+    message(STATUS "Using system LLVM toolchain (no MG_TOOLCHAIN_ROOT set)")
+endif()
+
+# FreeBSD installs third-party packages under /usr/local
+include_directories(SYSTEM /usr/local/include)
+link_directories(/usr/local/lib)
```
arrowd: Like Copilot said, this is a bad idea, also because `user.localbase` may be set to a value different than `/usr/local`. But luckily these two lines shouldn't be needed at all.

environment/freebsd-bin/make
```diff
@@ -0,0 +1 @@
+/usr/local/bin/gmake
```
arrowd: `/usr/bin/env gmake` would be more portable.

conan_recipes/recipes/openjdk/system/conanfile.py
```diff
@@ -0,0 +1,22 @@
...
+    version = "21.0.1"
+    description = "System OpenJDK wrapper for FreeBSD"
+    package_type = "application"
+    settings = "os", "arch"
+
+    def validate(self):
+        pass  # Accept any OS - the system provides Java
+
+    def package_info(self):
+        java_home = os.environ.get("JAVA_HOME", "/usr/local")
```
arrowd: Could we call `localbase` via C ffi there, instead of hardcoding "/usr/local" please?

conanfile.py
```diff
@@ -142,8 +142,9 @@ def requirements(self):
         self.requires("snappy/1.2.1", override=True)
 
     def build_requirements(self):
-        self.tool_requires("cmake/[>=4 <5]")
-        self.tool_requires("ninja/[>=1.13 <2]")
+        if self.settings.os != "FreeBSD":
```
Ignition: Lets try and keep conanfile.py OS agnostic as much as possible, unless needed.
You can use [replace_tool_requires] and [platform_tool_requires] in the freebsd profiles

environment/freebsd-bin/make
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+exec gmake "$@"
```
Ignition: nope, just tell conan directly in the profile
```
jemalloc/*:tools.gnu:make_program=gmake
libbcrypt/*:tools.gnu:make_program=gmake
```

And modify the libbcrypt recipe to not be hardcoded 
```
make = self.conf.get("tools.gnu:make_program", default="make")
self.run(f"{make} -C {self.source_folder} CC={cc}")
```

conan_recipes/recipes/mgclient/all/conanfile.py
```diff
@@ -0,0 +1,123 @@
...
+        # TCP_KEEPIDLE is Linux-specific; FreeBSD uses TCP_KEEPALIVE.
+        mgsocket = os.path.join(self.source_folder, "src", "linux", "mgsocket.c")
+        replace_in_file(self, mgsocket, '#include "mgsocket.h"',
+                        '#include "mgsocket.h"\n\n'
+                        '#ifndef SOL_TCP\n'
+                        '#define SOL_TCP IPPROTO_TCP\n'
+                        '#endif\n'
+                        '#ifndef TCP_KEEPIDLE\n'
+                        '#define TCP_KEEPIDLE TCP_KEEPALIVE\n'
+                        '#endif\n')
```
Ignition: I would rather this was fixed upstream than requiring vendoring here

conan_recipes/recipes/boost/all/conanfile.py
```diff
@@ -2095,6 +2095,9 @@ def filter_transform_module_libraries(names):
                 self.cpp_info.components["_libboost"].system_libs.append("rt")
                 if self.options.multithreading:
                     self.cpp_info.components["_libboost"].system_libs.append("pthread")
+            elif self.settings.os == "FreeBSD":
+                if self.options.multithreading:
+                    self.cpp_info.components["_libboost"].system_libs.append("pthread")
```
Ignition: This is only vendored ATM because of boost use of b2 needed fixing, at some point that will be fix, or maybe already fixed and we will stop using this vendored recipe. Please apply fix the the upstream recipe as well.

Ignition: Also maybe refactor to use `elif self.settings.os in ["Linux", "FreeBSD"]:`

conan_recipes/recipes/usearch/all/patches/0003-fix-mmap-anonymous-fd-freebsd.patch
```diff
@@ -0,0 +1,11 @@
+--- a/include/usearch/index_plugins.hpp
++++ b/include/usearch/index_plugins.hpp
+@@ -854,7 +854,7 @@ class page_allocator_t {
+ #if defined(USEARCH_DEFINED_WINDOWS)
+         return (byte_t*)(::VirtualAlloc(NULL, count_bytes, MEM_COMMIT | MEM_RESERVE, PAGE_READWRITE));
+ #else
+-        return (byte_t*)mmap(NULL, count_bytes, PROT_WRITE | PROT_READ, MAP_PRIVATE | MAP_ANONYMOUS, 0, 0);
++        return (byte_t*)mmap(NULL, count_bytes, PROT_WRITE | PROT_READ, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
```
Ignition, line 8: This has been fixed upstream, we should check the other patches we have, if all fixed we can bump the version we use and cleanup.

conan_config/profiles/memgraph_freebsd
```diff
@@ -0,0 +1,28 @@
...
+compiler=clang
+compiler.cppstd=gnu23
+compiler.libcxx=libc++
+compiler.version=21
+os=FreeBSD
+
+[conf]
+tools.cmake.cmaketoolchain:generator=Ninja
+tools.cmake.cmaketoolchain:user_toolchain=["{{ os.path.join(os.getcwd(), 'environment', 'toolchain', 'v7', 'toolchain-freebsd.cmake') }}"]
+tools.build:compiler_executables={"c": "/usr/bin/cc", "cpp": "/usr/bin/c++"}
```
Ignition, line 19: Looks like we are not using toolchain, in that case simplify.
CMake add a new option to guard the relevant code
```
option(MG_BUNDLED_TOOLCHAIN "Build against the bundled toolchain and package its libstdc++" ON)
```

wire it up with the conanfile
```
options = {
   "bundled_toolchain": [True, False],
}
default_options = {
   "bundled_toolchain": True, ...
}
tc.cache_variables["MG_BUNDLED_TOOLCHAIN"] = "ON" if self.options.bundled_toolchain else "OFF"
```

Fix memgraph_freebsd profile
```
[options]
&:bundled_toolchain=False
```

Ignition (changes requested):

conan_config/profiles/memgraph_freebsd_build
```diff
@@ -0,0 +1,25 @@
+{% set target_arch = {"amd64": "x86_64", "aarch64": "armv8"}.get(platform.machine(), platform.machine()) %}
+
+[settings]
+arch={{ target_arch }}
+build_type=Release
+compiler=clang
+compiler.cppstd=gnu17
```
Ignition, line 7: What about cstd?

Ignition (dismissed):
Also maybe need FreeBSD: distro entry in settings_user.yml

conan_config/profiles/memgraph_freebsd_build
seuros: cstd=gnu17   it was default, but let set it.

seuros:
Toolchain V8 just landed, i have to retest in both.

Ignition:
### Tracking
    - No issue exists for FreeBSD support: a search of the tracker for "FreeBSD" returns nothing, open or closed. Worth filing one to hold the remaining pieces, since this branch is the first of several.

### Standard development
- [ ] Update unit/E2E tests
    - No test covers the FreeBSD configuration, because no CI runner can build it. The dependency graph was resolved and the affected dependencies built by hand on FreeBSD 16.0-CURRENT, and the recipes that changed for every platform were rebuilt on Linux.
    - Nothing on the query or storage paths changes.

### CI Testing Labels
    - Release, debug and community core, plus one package build, since a refreshed lockfile changes the recipe revisions every build resolves.

### Documentation checklist
    - Memgraph's dependencies build on FreeBSD, using the system clang and tools from ports, although memgraph itself does not configure there yet. [#4617](https://github.com/memgraph/memgraph/pull/4617)

Ignition:
**The toolchain's layout is decided in one place.** `CMakeLists.txt` no longer
names `lib64`, `lib` or `libstdc++.so.6`: `cmake/MgToolchain.cmake` decides
them, and consumers link `mg::toolchain`, which carries the runtime and the
`-nostdlib++` that has to accompany it. Turning `MG_BUNDLED_TOOLCHAIN` off
builds against the host's C++ runtime and installs nothing beside the binary.
On Linux, configure and build are unchanged and the memgraph link line still
carries the flag, the rpath and the runtime.

**FreeBSD is a platform the build scripts know.** It is identified by its major
version, described by a dependency script beside the twelve Linux ones, and
`build.sh` derives both profiles and the bundled-toolchain option from what it
detects, so no hand-written conan command is needed. The shared machinery
learned `pkg` alongside apt and dnf. Three portability defects turned up on the
way: the architecture check rejected `amd64`, `install_deps.sh` could not
identify a host without `/etc/os-release`, and the shared scripts assumed bash
sits in `/bin`, which on FreeBSD it does not.

**The profiles are composed rather than copied.** A base profile per family
holds what its members agree on, so the pinned port versions are stated once
and a bump edits one file. Every profile renders to exactly what it rendered to
before, so no package id moves.

**The source fixes are patch files** in the recipe folders, where the next
reader can see what upstream is missing. nuraft's FreeBSD guards are absent
from upstream master and from the v3.0.0 tag, so the planned bump will not
retire that patch. usearch's mmap fix is upstream as of 2.24.0, so that one
goes away with the next version bump instead.

**One trap worth passing on.** Recipe revisions are part of `conan.lock`, so an
edited vendored recipe has no effect until the lockfile is refreshed; conan
keeps resolving the previously locked revision and the edit is silently inert.
That is why the lockfile is part of this branch.

**Verified on FreeBSD 16:** the adapter lists, checks and installs packages,
`./build.sh --graph-info` resolves the graph with no invalid nodes, and
libbcrypt, librdtsc, nuraft and jemalloc build with the system clang. On Linux
the same command still selects the v8 toolchain profile, and the patches that
apply on every platform were rebuilt there, since a patch that fails to apply
would break the Linux build.

**Still open.** `src/auth` calls into libseccomp unconditionally, so a FreeBSD
build now reaches the compiler and stops there, and that needs its owner.
Nothing in CI exercises any of this; the graph resolves in about twelve
seconds, which is cheap enough to be a job and is what would stop these
profiles going stale between the pull requests in this series.

Ignition:
@seuros I made some changes, can you retry it in your workflow and ensure it still works for you

Ignition:
Three more commits, which change what is open. The earlier comment on this pull request said memgraph itself could not configure on FreeBSD and that nothing selected these profiles; both are now addressed, so it is worth reading that comment as history rather than as current state.

**The toolchain's layout is decided in one place.** `CMakeLists.txt` no longer names `lib64`, `lib` or `libstdc++.so.6`: `cmake/MgToolchain.cmake` decides them, consumers link `mg::toolchain`, which carries the runtime and the `-nostdlib++` that has to accompany it, and the two global properties that used to pass the path around had no other readers and are gone. Turning `MG_BUNDLED_TOOLCHAIN` off builds against the host's C++ runtime and installs nothing beside the binary. On Linux, configure and build are unchanged and the memgraph link line still carries the flag, the rpath and the runtime; with the option off, configure succeeds instead of failing on a missing path.

**FreeBSD is a platform the build scripts know.** Detection returned an error for anything that was not glibc Linux or macOS, and `build.sh` calls it before choosing a profile, so these profiles were reachable only by invoking conan by hand. FreeBSD is now identified by its major version, described by a dependency script beside the twelve Linux ones, and `build.sh` derives both profiles and the bundled-toolchain option from what it detects. The shared machinery learned `pkg` alongside apt and dnf. Three further portability defects turned up while doing it: the architecture check rejected `amd64`, `install_deps.sh` could not identify a host without `/etc/os-release`, and the shared scripts assumed bash sits in `/bin`, which on FreeBSD it does not.

**The profiles are composed rather than copied.** Each was a full copy of its sibling, so the five pinned port versions and both explanatory comments were stated twice and a bump meant editing two files. A base profile per family now holds what its members agree on. Every profile renders to exactly what it rendered to before, so no package id moves.

Verified on FreeBSD 16: the adapter lists, checks and installs packages, and `./build.sh --graph-info` computes the graph with no hand-written conan command. On Linux the same command still selects the v8 toolchain profile. libbcrypt, librdtsc, nuraft and jemalloc build on FreeBSD, and the three patches that apply on every platform were rebuilt on Linux too.

What is still open: `src/auth` calls into libseccomp unconditionally, so a FreeBSD build now reaches the compiler and stops there, and that needs its owner. Nothing in CI exercises any of this yet; the dependency graph resolves in about twelve seconds, which is cheap enough to be a job and is what would stop these profiles going stale between the pull requests in this series.

seuros:
Will do.

seuros:
@Ignition This is compilable now in FreeBSD-15/16 .  Linux is also correct because the CI use it.

environment/os/check-packages.py
```diff
@@ -84,6 +99,21 @@ def install_packages_dnf(packages, dry_run=False):
...
+    """Install missing pkg packages"""
+    # Check which packages are missing
+    missing = check_packages_pkg(packages)
+
+    if missing and not dry_run:
+        print(f"Installing missing packages: {' '.join(missing)}")
+        result = subprocess.run(["pkg", "install", "-y"] + missing)
+        if result.returncode != 0:
+            print(f"Error installing packages: {result.stderr}")
+            return False
```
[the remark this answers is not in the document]
seuros: I get nitted by a bot :D

Ignition (approved):

